PR #1470, opened 2022-11-18
name and value

yegor256:
Name and value added to the definition of an attribute


Graur (approved):
@yegor256 Looks good to me. I like it when we add more detail to our definitions.

yegor256:
@rultor merge

PR #1484, opened 2022-11-23
\rho is a mutable attribute

yegor256:
We use this feature of moving an object to a new parent in `while`, `goto`, `try`. But we didn't give an ability to users to do the same from EO code. Now they can, because `\rho` is a mutable attribute.


yegor256:
@Graur please, take a look

yegor256:
@Graur please, wait a bit

yegor256:
@Graur it's ready, take a look pls

eo-runtime/src/test/eo/org/eolang/runtime-tests.eo
```diff
@@ -389,3 +389,20 @@
...
 
+[] > moves-copy-to-new-parent
+  assert-that > @
+    [] > app
+      [] > @
+        "Jeff" > x
+        QQ.txt.sprintf > @
+          "[%s]"
+          foo
+            $:^
```
Graur, line 400: @yegor256 Can we use any object here? Like `^:^` or `QQ.io.stdout:^`. If so maybe it would be better to change `^` name to something else?

Graur (commented):
@yegor256 I like this idea, but I have one question about it:

eo-runtime/src/test/eo/org/eolang/runtime-tests.eo
yegor256: @Graur yes, any object may stay at the left side of `:`, but only a name OR `^` on the right. We use `^` since it's our textual form of `\rho`

yegor256:
@rultor merge

eo-runtime/src/test/eo/org/eolang/runtime-tests.eo
Graur: @yegor256 Oh, it is a parent object. Now I get it

Graur (approved):

PR #1485, opened 2022-11-24
stack-size for all modules

yegor256:
I set `stack-size` for both eo-runtime and eo-maven-plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,6 +86,12 @@ SOFTWARE.
       <url>https://github.com/objectionary/eo</url>
     </site>
   </distributionManagement>
+  <properties>
+    <!-- This is required for later correct replacement of argLine -->
+    <argLine/>
+    <!-- This is the size of the stack, which you can modify from command line -->
+    <stack-size>8M</stack-size>
+  </properties>
   <dependencyManagement>
     <dependencies>
       <dependency>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -86,6 +86,12 @@ SOFTWARE.
       <url>https://github.com/objectionary/eo</url>
     </site>
   </distributionManagement>
+  <properties>
+    <!-- This is required for later correct replacement of argLine -->
+    <argLine/>
+    <!-- This is the size of the stack, which you can modify from command line -->
+    <stack-size>8M</stack-size>
```
mximp, line 93: @yegor256 Are you sure reducing stack to 8M would not fail the builds? As far as I remember we increased default stack size recently exactly for that reason.

yegor256: @mximp "The default is 320 KB for 31-bit or 32-bit JVMs and 1024 KB for 64-bit JVMs." (c) IBM. We set it to 8Mb. should be more than enough, I believe. It doesn't fail for me.

mximp: @yegor256 previous value was set to `256M`.. Anyway we can try and increase if required.

yegor256:
@rultor merge

yegor256:
@rultor merge

Issue #1494, opened 2022-11-25
Use `FakeMaven` inside `OptimizeMojo` tests


volodya-lombrozo:
I think it's important to continue with tests refactoring of `eo-maven-plugin`. The next step would be refactoring of the `OptimzeMojoTest` class. To do so we have to apply our new object `FakeMaven`.

volodya-lombrozo:
Assign to me

volodya-lombrozo:
@yegor256 I believe it's a mistake to create something with "perfect design". We don't have enough information about the future class usage before we actually try to apply it in real life. It's better to go step by step and improve `FakeMojo` in each that step. The current issue it's a just a next small step. I'm not only going to apply `FakeMaven` for the `OptimizeMojo`, I'll have a chance to improve `FakeMaven` itself.
___
Moreover, other issues related to the `FakeMaven` are finished and we can continue with that issue.

volodya-lombrozo:
@yegor256 assign to me

0pdd:
@volodya-lombrozo the puzzle [#1514](https://github.com/objectionary/eo/issues/1514) is still not solved.

0pdd:
@volodya-lombrozo the only puzzle [#1514](https://github.com/objectionary/eo/issues/1514) is solved here.

PR #1507, opened 2022-11-29
feat(#1494) Refactor `OptimizeMojoTest` using `FakeMaven`

volodya-lombrozo:
I've refactored `OptimizeMojoTest` using the new approach with `FakeMaven`. As you can see - we reduce the test code significantly. Let me know, if PR is a quite large to review, I'll split it in that case.

Closes: #1494

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Moja.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Moja.java
@@ -32,6 +32,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
@@ -158,11 +159,14 @@ public final class Moja<T extends AbstractMojo> {
         final Map.Entry<String, Object> entry
     ) throws IllegalAccessException {
         final String name = entry.getKey();
-        try {
-            final Field field = clazz.getDeclaredField(name);
+        final Optional<Field> declared = Arrays.stream(clazz.getDeclaredFields())
+            .filter(f -> f.getName().equals(name))
+            .findFirst();
+        if (declared.isPresent()) {
+            final Field field = declared.get();
             field.setAccessible(true);
             field.set(mojo, entry.getValue());
-        } catch (final NoSuchFieldException ex) {
+        } else {
             final Class<?> parent = clazz.getSuperclass();
             if (parent == null) {
                 Logger.warn(
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Moja.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Moja.java
@@ -146,9 +146,11 @@ public final class Moja<T extends AbstractMojo> {
     }
 
     /**
-     * Take a field.
-     * @param mojo The class
+     * Init a field.
+     * @param clazz The mojo class
+     * @param mojo The mojo
      * @param entry Field name and value
+     * @throws java.lang.IllegalAccessException If can't set field.
      */
     private void initField(
         final Class<?> clazz,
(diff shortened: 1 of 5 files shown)


volodya-lombrozo:
@mximp please, take a look.

eo-maven-plugin/src/main/java/org/eolang/maven/Moja.java
```diff
@@ -147,30 +146,35 @@ private static Collection<Field> fields(final Class<?> cls) {
...
+        final AbstractMojo mojo,
+        final Map.Entry<String, Object> entry
+    ) throws IllegalAccessException {
+        final String name = entry.getKey();
         try {
-            field = mojo.getDeclaredField(name);
+            final Field field = clazz.getDeclaredField(name);
+            field.setAccessible(true);
+            field.set(mojo, entry.getValue());
         } catch (final NoSuchFieldException ex) {
```
mximp, line 165: @volodya-lombrozo here we use exception for control flow. Can we use `clazz.getDecleardFields()` instead to get rid of `try-catch` block?

eo-maven-plugin/src/main/java/org/eolang/maven/Moja.java
```diff
@@ -147,30 +146,35 @@ private static Collection<Field> fields(final Class<?> cls) {
...
-            final Class<?> parent = mojo.getSuperclass();
+            final Class<?> parent = clazz.getSuperclass();
             if (parent == null) {
```
mximp, line 167: @volodya-lombrozo shouldn't we throw an exception in this case as it was before? Otherwise we just hide the problem.

volodya-lombrozo: @mximp It helps me to configure `FakeMaven`. If some of the properties are absent in `Mojo`, then the `NoSuchFieldException ` will thrown. Of course, I can write the logic inside `FakeMaven` in order to check whether the property is allowed
for particular `Mojo`, but it is quite a verbose solution and it doesn't worth it. Moreover, since all tests are pass, we can conclude that we don't break anything by that changes.

eo-maven-plugin/src/main/java/org/eolang/maven/Moja.java
volodya-lombrozo: @mximp fixed. Please, check.

eo-maven-plugin/src/main/java/org/eolang/maven/Moja.java
mximp: @volodya-lombrozo ok.. I believe in this case there is no point in warning here.. What is the expected reaction to it?

volodya-lombrozo: @mximp expected reaction to what? As you can see - we just ignore field if it isn't found in the class.

mximp: > @mximp expected reaction to what? As you can see - we just ignore field if it isn't found in the class.

@volodya-lombrozo why do we print warning then?

volodya-lombrozo: @mximp  in order to inform a programmer that it is unusual situation

volodya-lombrozo:
@mximp do you have any other remarks?

eo-maven-plugin/src/main/java/org/eolang/maven/Moja.java
mximp: @volodya-lombrozo we will always have such warnings as the list of properties is shared for all mojos and contains a superset of props. With such a case it's normal that some of the props will not be applicable for some mojos. There should be no warning in this case.

volodya-lombrozo: @mximp updated. Please, check

mximp (approved):

volodya-lombrozo:
@yegor256 please, merge

yegor256:
@rultor merge

Issue #796, opened 2022-06-27
Document EO parsing process


mximp:
To simplify understanding by the newcomers it would be good to have parsing process described somewhere.
The description is supposed to include input/output structure of key steps as well as location of collaborators (xsl files, classes).

mximp:
@yegor256 do you find this appropriate?

yegor256:
@mximp yes, definitely, would be nice to extent the README

yegor256:
@Graur fix this please (or delegate)

Graur:
@levBagryansky please, help to improve README.md file

mximp:
@levBagryansky It would nice to have description of `target/eo` folder structure as well as what files are consumed/produced by each compilation step.

levBagryansky:
@mximp so I need to imrove README.md in eo/eo-parser? How can I understand what is going on there?

mximp:
@levBagryansky I'd say `eo-maven-plugin/readme.md`. You could run compilation, check `target/eo/` and find how it's populated/used in the plugin code.

levBagryansky:
@mximp maybe it would be better to leave a link to the [post](https://www.yegor256.com/2021/10/21/objectionary.html#place-) to avoid repeating?

mximp:
@levBagryansky The post explains what the phases are but doesn't specify exact input and output. I believe some addition is required. I've tried to accumulate the info here as another blog post: https://github.com/mximp/news.eolang.org/commit/cc01abcc9a81143de3156d353aacd11db6bdce05. 
But the format is not very readable. Maybe you could think of another representation.

Graur:
@levBagryansky What's up with this issue?

yegor256:
@mximp we still have this issue?

mximp:
@Graur anything left here?

Graur:
@mximp @yegor256  Yes, we still have this issue
@MikhailLipanin Can you help us here?

volodya-lombrozo:
@Graur Do we still need to solve that?

Graur:
@volodya-lombrozo I believe we should ask @mximp about it. Since we have short description in `eo-maven-plugin/README.md`. But is it enough?

mximp:
[...]

PR #1513, opened 2022-12-01
796 parsing process documentation

levBagryansky:
@mximp What do you think about this fromat of parsing proccess documentation (issue #796)


mximp:
@levBagryansky Big tables are proven to be hard to read. Can we get rid of the table format and just make separate paragraphs for each?

PR #1519, opened 2022-12-02
feat(#1337) Concurrency tests for ParseMojo and OptimizeMojo

### LINKED ISSUE #1337: SynchronizedTojo.java:31-34: Replace this custom {@link...
The puzzle `1230-0c34d341` from #1230 has to be resolved: 

https://github.com/objectionary/eo/blob/2fd90f407b6a33a8679a8575314d9659932d968e/eo-maven-plugin/src/main/java/org/eolang/maven/SynchronizedTojo.java#L31-L34

The puzzle was created by rultor on 14-Oct-22. 

volodya-lombrozo:
I've added several tests in order to fix #1337 issue. The tests help me to detect whether some solution helps or not. I still haven't found the correct way to fix race condition with `Tojo` and `Mono` classes, but it's just the first step. Moreover I had to fix `FakeMaven` in order to make it work with several eo programs.


volodya-lombrozo:
@mximp Please, take a look

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
```diff
@@ -181,16 +186,41 @@ public Map<String, Path> result() throws IOException {
...
      * Adds eo program to a workspace.
-     * @param path Relative path where to save EO program
      * @param content EO program content.
      * @return The same maven instance.
      * @throws IOException If method can't save eo program to the workspace.
      */
-    private FakeMaven withProgram(final Path path, final String content) throws IOException {
+    private FakeMaven withProgram(final String content) throws IOException {
+        final Path path = Paths.get(
+            String.format("foo/x/main%s.eo", FakeMaven.suffix(this.current.get()))
```
mximp, line 215: @volodya-lombrozo with concurrent execution we still can get the same number here.
We also don't have any synchronization between `concurrent.get()` and `concurrent.incrementAndGet()` so some other thread might increment the value and foreign catalog might get wrong file written.

volodya-lombrozo: @mximp `FakeMaven` isn't thread safe for now and it doesn't have to be. The idea is to create single `FakeMaven` and work with it. I can't imagine the situation where I have to prepare environment in parallel. 
If you insist, I can create a puzzle, but I don't see the reason for it. What do you think?

mximp: @volodya-lombrozo Can we mark `FakeMave` as not thread-safe in JavaDoc please

mximp (commented):
@volodya-lombrozo please check my comment

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
volodya-lombrozo: @mximp Done

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
```diff
@@ -44,6 +45,7 @@
  * @since 0.28.12
  */
 @SuppressWarnings("PMD.TooManyMethods")
+@NotThreadSafe
```
mximp, line 48: @volodya-lombrozo we normally place comment in JavaDoc `NOT thread-safe`.
If we are to use this new annotation we need to change it in all other places. I don't think it's reasonable in this case.

volodya-lombrozo: @mximp `NOT thread-safe` is used only several times :) I've added `NOT thread-safe` to java doc. Please, check.

volodya-lombrozo:
@yegor256 merge please

yegor256:
@rultor merge

PR #1538, opened 2022-12-07
saxon coordinates fixed

yegor256:
These files will guarantee that we always use Saxon for XSL/XPath processing


yegor256:
@volodya-lombrozo take a look pls

volodya-lombrozo (approved):
@yegor256 Looks good to me. Actually I've already done the same...

yegor256:
@rultor merge

Issue #1514, opened 2022-12-01
Moja.java:155-158: We have some doubts about using...


0pdd:
The puzzle `1494-e338b825` from #1494 has to be resolved: 

https://github.com/objectionary/eo/blob/f07e44843d4f32fef7bf197aa64cad75856ad3bc/eo-maven-plugin/src/main/java/org/eolang/maven/Moja.java#L155-L158

The puzzle was created by rultor on 01-Dec-22. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

volodya-lombrozo:
@yegor256 do you agree that we don't need `Longer.warn` here? I believe that it's a good notification for developer.

volodya-lombrozo:
I would remove the puzzle.

yegor256:
@volodya-lombrozo why do we print a warning instead of throwing an exception?

yegor256:
@volodya-lombrozo I like this: "I will have to check in FakeMaven whether param is allowed for a particular Mojo."

volodya-lombrozo:
@yegor256 Ok, then I'll return exception back and will add the logic to `FakeMaven`

PR #1540, opened 2022-12-07
feat(#1514) Throw an Exception if param doesn't fit to a Mojo

volodya-lombrozo:
I've returned back the logic with throwing an Exception in `Moja` if a corresponding field isn't found.
Now we check field presence right inside `FakeMaven`.  

Closes: #1514


mximp (approved):

yegor256:
@rultor merge

Issue #1476, opened 2022-11-21
Print human-readable message if object isn't found


volodya-lombrozo:
I've wrote a simple `book.eo` program:
```eo
[] > app
  QQ.io.stdout > @
    QQ.txt.sprintf
      "Wake up, %s..."
      "Neo"
```
Then, in the terminal I wrote the next command:
```sh
eoc dataize book
```
And get the next exception in the console: 
```java
Exception in thread "main" java.lang.ClassNotFoundException: EObook
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
	at java.base/java.lang.Class.forName0(Native Method)
	at java.base/java.lang.Class.forName(Class.java:375)
	at org.eolang.Main.run(Main.java:181)
	at org.eolang.Main.main(Main.java:98)
```
If you are a Java Developer, you can understand the problem - the `book` object doesn't exists and it's correct (because our object has the `app` name). I think it would be better to print more "human readable" message for developers. Something like a "'book' object doesn't exists, maybe you meant 'app'?"

Graur:
@levBagryansky Can you help us with this? I suppose to add to `Main.run()` method some `try-catch` construction or to update `Main.main()` method exception catching.

PR #1541, opened 2022-12-08
1476 object does not exist exception

levBagryansky:
#1476
Print human-readable message if object isn't found

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Main.java
+++ b/eo-runtime/src/main/java/org/eolang/Main.java
@@ -41,6 +41,8 @@ import java.util.logging.LogRecord;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
+import static java.lang.Class.*;
+
 /**
  * Bridge between Java CLI and EO.
  *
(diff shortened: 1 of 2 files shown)


levBagryansky:

eo-runtime/src/test/java/org/eolang/MainTest.java
```diff
@@ -84,6 +84,14 @@ public void jvmFullRunWithError() throws Exception {
         );
     }
 
+    @Test
+    public void objectNotFoundException() throws Exception {
+        MatcherAssert.assertThat(
+            MainTest.exec("--verbose", "unavailable-name"),
```
volodya-lombrozo, line 90: @levBagryansky should we use "--verbose" flag to print "Can not find \"unavailable-name\" object"?  Or the compiler will print the message anyway?

eo-runtime/src/main/java/org/eolang/Main.java
```diff
@@ -177,10 +177,17 @@ private static void run(final List<String> opts) throws Exception {
...
+        final Phi app;
+        try {
+            app = Phi.class.cast(
+                Class.forName(path).getConstructor(Phi.class)
                 .newInstance(Phi.Φ)
-        );
+            );
+        } catch (final ClassNotFoundException ex) {
+            throw new ExUnset(
+                String.format("Can not find \"%s\" object", opts.get(0))
```
volodya-lombrozo, line 188: @levBagryansky what do you think if we write message using single quotes? 
```java
String.format("Can not find '%s' object", opts.get(0))
```
In that case we don't need to escape "

volodya-lombrozo (changes requested):

eo-runtime/src/main/java/org/eolang/Main.java
levBagryansky: @volodya-lombrozo ok

eo-runtime/src/test/java/org/eolang/MainTest.java
levBagryansky: @volodya-lombrozo Indeed, it is printed anyway

volodya-lombrozo (approved):
Looks good to me. @yegor256 merge

levBagryansky:
@yegor256 please check

yegor256:
@rultor merge

Issue #1547, opened 2022-12-09
maven-javadoc-plugin fails because eo-runtime.jar is absent in the list of dependencies


yegor256:
For example, here: https://github.com/objectionary/eo-math/issues/99#issuecomment-1342832357

The problem has been reported earlier, but there was no answer: https://stackoverflow.com/questions/69770376/maven-javadoc-plugin-package-doesnt-exist-when-an-external-jar-is-used

yegor256:
@volodya-lombrozo in order to reproduce the error, try to run `mvn javadoc:jar -Psonatype` in `eo-math` repository, for example

volodya-lombrozo:
`eo-runtime.jar` (version `0.28.11`) doesn't contain any Java classes. We have to find the root problem why it is so. `eo-runtime.jar` with version `0.28.10` contains all required classes.

volodya-lombrozo:
Deployment log [here](https://www.rultor.com/t/28393-1310472258).

The important entries:
```
[1;34mINFOm] 1m--- 0;32meo-maven-plugin:0.28.11:copym 1m(compile)m @ 36meo-runtime0;1m ---m
[1;34mINFOm] 20 source(s) copied from ./home/r/repo/eo-runtime/src/main/eo to ./home/r/repo/eo-runtime/target/classes/EO-SOURCES
[1;34mINFOm]
[1;34mINFOm] 1m--- 0;32meo-maven-plugin:0.28.11:unplacem 1m(compile)m @ 36meo-runtime0;1m ---m
[1;33mWARNINGm] The binary org/eolang/version.txt of org.eolang/eo-runtime/-/0.28.10 looks different, won't unplace
[1;33mWARNINGm] The binary org/eolang/org-eolang-eo-runtime-jar.txt of org.eolang/eo-runtime/-/0.28.10 looks different, won't unplace
[1;33mWARNINGm] The binary EOorg/org-eolang-eo-runtime-jar.txt of org.eolang/eo-runtime/-/0.28.10 looks different, won't unplace
[1;33mWARNINGm] The binary EOorg/EOeolang/EOio/org-eolang-eo-runtime-jar.txt of org.eolang/eo-runtime/-/0.28.10 looks different, won't unplace
[1;33mWARNINGm] The binary EOorg/EOeolang/org-eolang-eo-runtime-jar.txt of org.eolang/eo-runtime/-/0.28.10 looks different, won't unplace
[...]

PR #1571, opened 2022-12-16
feat(#1547): Add test that proves presence of `eo-runtime` dependency 

volodya-lombrozo:
Add test that proves presence of `eo-runtime` dependency or all eo programs.
It is the first step on the way to finding the problem of #1547


volodya-lombrozo:
@yegor256 check

yegor256:
@rultor merge

PR #1572, opened 2022-12-16
feat(#1547): add `PlaceMojo` test that place `eo-runtime` classes

volodya-lombrozo:
Add test that proves that all classes from `eo-runtime` library will be added to `target/classes` directory automatically. Correlates with #1547


volodya-lombrozo:
@yegor256 check

yegor256:
@rultor merg

Issue #1595, opened 2022-12-20
ResolveMojo must exclude 'provided' dependencies too


yegor256:
Here, we exclude `test` deps: https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/java/org/eolang/maven/ResolveMojo.java#L183 We should also exclude `provided`.

yegor256:
@volodya-lombrozo please, check this one, it's also a blocker

0pdd:
@yegor256 the puzzle [#1598](https://github.com/objectionary/eo/issues/1598) is still not solved.

yegor256:
@rultor release, tag is `0.28.14`

yegor256:
@rultor release, tag is `0.28.14`

0pdd:
@yegor256 the only puzzle [#1598](https://github.com/objectionary/eo/issues/1598) is solved here.

PR #1596, opened 2022-12-20
feat(#1595) check provided dependencies

volodya-lombrozo:
Add additional check for "provided" dependencies and puzzle for tests

Closes: #1595

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/ResolveMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/ResolveMojo.java
@@ -163,9 +163,9 @@ public final class ResolveMojo extends SafeMojo {
      *
      * @return List of them
      * @todo #1595:30 Make method 'deps' testable. For now it's not possible to test
-     *   'ignoreTransitive=false' branch because it's hard to mock all required fields.
-     *   Maybe we should provide a chance to mock all dependencies related to maven or
-     *   even extract new classes.
+     *  'ignoreTransitive=false' branch because it's hard to mock all required fields.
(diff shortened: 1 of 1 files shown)


eo-maven-plugin/src/main/java/org/eolang/maven/ResolveMojo.java
```diff
@@ -222,6 +226,15 @@ private Collection<Dependency> deps() {
             .collect(Collectors.toList());
     }
 
+    /**
+     * Check if dependency is not needed at runtime.
+     * @param dep Maven dependency
+     * @return True if it's not needed at runtime
+     */
+    private static boolean isNotRuntimeRequired(final Dependency dep) {
+        return !dep.getScope().contains("test") && !dep.getScope().contains("provided");
```
yegor256, line 235: @volodya-lombrozo by the way, why do we use `contains` instead of `equals`? They are two strings to compare.

yegor256:
@volodya-lombrozo see above. I think it's better to compare strings with `equals`.

yegor256:
@volodya-lombrozo also, maybe it's better to _include_ only `compile` and `runtime` deps and exclude everything else?

volodya-lombrozo:
@yegor256 Can you take a look again, please?

eo-maven-plugin/src/main/java/org/eolang/maven/ResolveMojo.java
volodya-lombrozo: @yegor256 done

yegor256:
@rultor merge

PR #1659, opened 2023-01-10
test(#1617): Add test for `AtLogged`

volodya-lombrozo:
Add test for `AtLogged`.

Closes: #1617

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/AtLogged.java
+++ b/eo-runtime/src/main/java/org/eolang/AtLogged.java
@@ -31,8 +31,6 @@ import java.io.PrintStream;
  * convenient for debugging).
  *
  * @since 0.24
- * @todo #1614:30min This class don't have enough tests. We need to add more, at least for
- *  the next methods: toString(), φTerm(), copy(), put(), get().
  */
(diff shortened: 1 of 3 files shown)


volodya-lombrozo:
@mximp Can you take a look, please?

eo-runtime/src/main/java/org/eolang/AtLogged.java
```diff
@@ -44,14 +44,30 @@ final class AtLogged implements Attr {
      */
     private final String owner;
 
+    /**
+     * Output stream.
+     */
+    private final PrintStream out;
```
mximp, line 50: @volodya-lombrozo why don't we use a logger here like we do in `PhDefault` and `Dataized`?

volodya-lombrozo: @mximp Yea, I agree, it would be much better. I've added puzzle for that. Check, please.

mximp (approved):

volodya-lombrozo:
@yegor256 Can you merge, please?

yegor256:
@rultor merge

PR #1662, opened 2023-01-10
imperative goto usage

yegor256:
This is an example of how `goto` and `while` can be used in imperative way


yegor256:
@mximp take a look, please

yegor256:
@rultor merge

PR #1714, opened 2023-01-25
#1713 exclude packages from probes

### LINKED ISSUE #1713: Suppress probing objects with less than 3 elements
As per [discussion](https://github.com/objectionary/eo/issues/1705#issuecomment-1396900739) in #1705 let's stop adding probes for objects with less than 3 elements as they are unlikely to be (even cannot be) actual objects, but rather packages, e.g. 'org.eolang', 'org', etc.

This should heavily decrease the number of object we would probe.

mximp:
closes #1713

- excluded probes started with '.' (as it's an internal object)
- excluded probes with less than 3 elements (packages)

Performance impact (MacBook pro):
- overall build time: 9:30min -> 7:15min
- number of probes for `eo-runtime`: 218 -> 76

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/add-probes.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/add-probes.xsl
@@ -63,7 +63,8 @@ SOFTWARE.
         <xsl:apply-templates select="node()|@*"/>
         <xsl:for-each select="//o[starts-with(@base, '.')]">
           <xsl:variable name="p" select="eo:qualify(.)"/>
-          <xsl:if test="not(eo:contains-any-of($p, ('$', '^', '@','&lt;')))">
+          <xsl:variable name="c" select="string-length($p) - string-length(translate($p, '.', ''))"/>
+          <xsl:if test="not(eo:contains-any-of($p, ('$', '^', '@', '&lt;'))) and not(starts-with($p, '.')) and $c &gt; 1">
             <xsl:element name="meta">
               <xsl:attribute name="line">
                 <xsl:value-of select="@line"/>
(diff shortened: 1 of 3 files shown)


mximp:
@volodya-lombrozo please check

eo-parser/src/main/resources/org/eolang/parser/add-probes.xsl
```diff
@@ -63,7 +63,8 @@ SOFTWARE.
         <xsl:apply-templates select="node()|@*"/>
         <xsl:for-each select="//o[starts-with(@base, '.')]">
           <xsl:variable name="p" select="eo:qualify(.)"/>
-          <xsl:if test="not(eo:contains-any-of($p, ('$', '^', '@','&lt;')))">
+          <xsl:variable name="c" select="string-length($p) - string-length(translate($p, '.', ''))"/>
```
volodya-lombrozo, line 66: @mximp Why do you need to use `translate()` function here?
Do I understand correctly, that you wanna check that (pseudocode): 
```
length(Q.org.eolang) - length(Qorgeolang) > 1
```

volodya-lombrozo (changes requested):

eo-parser/src/main/resources/org/eolang/parser/add-probes.xsl
mximp: @volodya-lombrozo correct. I need to count the number of dots within FQN.

mximp: Actually `>2` - it's a bug.. let me add one more commit

volodya-lombrozo: I understand why you do it... It's a [common practice](https://stackoverflow.com/questions/36477331/count-number-of-character-in-a-variable-in-xslt-2-0). Interesting...

volodya-lombrozo (approved):

volodya-lombrozo:
@yegor256 Looks good to me. Can you merge that changes, please?

yegor256:
@rultor merge

Issue #1716, opened 2023-01-25
Use Objectionary index for probing


mximp:
Depends on https://github.com/objectionary/home/issues/96

As discussed in https://github.com/objectionary/eo/issues/1705 let's reduce the number of http requests for probing by using the [index](https://github.com/objectionary/home/issues/96). Let's make `OyRemte.contains()` use the index file and download it only once (cache in `.eo`).

volodya-lombrozo:
@mximp I can implement the issue

mximp:
Index file is available: https://home.objectionary.com/objectionary.lst

PR #1722, opened 2023-01-30
feat(#1716): Use cached objects index

volodya-lombrozo:
Use cached objects index. Cache in memory.

Closes: #1716

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/ObjectsIndex.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/ObjectsIndex.java
@@ -50,7 +50,24 @@ final class ObjectsIndex {
      * Ctor.
      */
     ObjectsIndex() {
-        this(ObjectsIndex.index());
+        this(
+            new ScalarOf<Set<String>>(
+                () -> new SetOf<>(
+                    new Mapped<>(
+                        ObjectsIndex::convert,
+                        new Mapped<>(
+                            Text::asString,
+                            new Split(
+                                new TextOf(
+                                    new URL("https://home.objectionary.com/objectionary.lst")
+                                ),
+                                "\n"
+                            )
+                        )
+                    )
+                )
+            )
+        );
     }
 
     /**
@@ -69,29 +86,6 @@ final class ObjectsIndex {
      */
     public boolean contains(final String name) throws Exception {
         return this.objects.value().contains(name);
-    }
-
-    /**
-     * Loads objects index.
-     * @return Objects index as a set of strings.
-     */
-    private static Scalar<Set<String>> index() {
-        return new ScalarOf<>(
-            () -> new SetOf<>(
-                new Mapped<>(
-                    ObjectsIndex::convert,
-                    new Mapped<>(
-                        Text::asString,
-                        new Split(
-                            new TextOf(
-                                new URL("https://home.objectionary.com/objectionary.lst")
-                            ),
-                            "\n"
-                        )
-                    )
-                )
-            )
-        );
     }
 
     /**
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/ObjectsIndex.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/ObjectsIndex.java
@@ -30,7 +30,6 @@ import org.cactoos.Text;
 import org.cactoos.iterable.Mapped;
 import org.cactoos.scalar.ScalarOf;
 import org.cactoos.scalar.Sticky;
-import org.cactoos.scalar.Unchecked;
 import org.cactoos.set.SetOf;
 import org.cactoos.text.Split;
 import org.cactoos.text.TextOf;
@@ -51,7 +50,7 @@ final class ObjectsIndex {
      * Ctor.
      */
     ObjectsIndex() {
-        this(ObjectsIndex.loadIndex("https://home.objectionary.com/objectionary.lst"));
+        this(ObjectsIndex.index());
     }
 
     /**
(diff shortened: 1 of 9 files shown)


volodya-lombrozo:
@mximp Can you have a look, please?

volodya-lombrozo:
@mximp reminder

mximp (approved):

volodya-lombrozo:
@yegor256 Could you merge that changes please?

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/OyRemote.java
```diff
@@ -73,10 +78,8 @@ public Input get(final String name) throws MalformedURLException {
     }
 
     @Override
-    public boolean contains(final String name) throws IOException {
-        final int code = ((HttpURLConnection) this.template.value(name).openConnection())
-            .getResponseCode();
-        return code >= HttpURLConnection.HTTP_OK && code < HttpURLConnection.HTTP_BAD_REQUEST;
+    public boolean contains(final String name) {
```
yegor256, line 81: @volodya-lombrozo I don't think it's reasonable to rely only on the cache file. It may be broken, absent, not complete. I think it's better to check the file first and then, if the element is absent in the file, we make an HTTP connection to double check. I would actually introduce a new `Oy*` object for this.

mximp: `OyIndexed`

volodya-lombrozo:
@mximp @yegor256 Could you have a look, again, please? I've added `OyIndexed` class as you suggested.

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/ObjectsIndex.java
```diff
@@ -0,0 +1,109 @@
...
+import org.cactoos.set.SetOf;
+import org.cactoos.text.Split;
+import org.cactoos.text.TextOf;
+
+/**
+ * The objects index that contains all available eo objects.
+ * <a href="https://github.com/objectionary/home/blob/gh-pages/objectionary.lst">Example</a>
+ * @since 0.29
+ */
+final class ObjectsIndex {
```
mximp, line 42: @volodya-lombrozo Do we now need this object since we introduced `OyIndexed`?

volodya-lombrozo: @mximp I believe, yes. They represent different entities with different behaviour.

mximp: @volodya-lombrozo In this case it has to implement an interface. So that I can supply for example in-memory impl.

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/OyIndexed.java
```diff
@@ -0,0 +1,87 @@
...
+        return this.delegate.get(name);
+    }
+
+    @Override
+    public boolean contains(final String name) throws IOException {
+        boolean result;
+        try {
+            result = this.index.contains(name);
+            // @checkstyle IllegalCatchCheck (1 line)
+        } catch (@SuppressWarnings("PMD.AvoidCatchingGenericException") final Exception ex) {
```
mximp, line 76: @volodya-lombrozo It's hard to read when inlined - shall be moved to function level

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/ObjectsIndex.java
```diff
@@ -0,0 +1,109 @@
...
+    public boolean contains(final String name) throws Exception {
+        return this.objects.value().contains(name);
+    }
+
+    /**
+     * Loads objects index.
+     * @return Objects index as a set of strings.
+     */
+    private static Scalar<Set<String>> index() {
+        return new ScalarOf<>(
```
mximp, line 79: @volodya-lombrozo this can be moved into `ObjectsIndex()` constructor. No need in static method.

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/OyIndexed.java
volodya-lombrozo: @mximp Good suggestion. Thanks.

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/ObjectsIndex.java
volodya-lombrozo: @mximp I would definitely prefer to leave static method, because it allows to read code faster. How I read the code:
```
//constructor

//constructor

//logic
```

You suggest

```
//constructor
//logic

//constructor

//logic
```

I don't know if you understand me, but my approach _for me_ looks more readable. This question is matter of taste, of corse. Actually we already had such a discussion previously. What do you think if we create an issue and finally decide  which approach to follow in order to avoid such comments in the future?

volodya-lombrozo:
@mximp Could you take a look, again, please?

mximp (approved):

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/ObjectsIndex.java
mximp: @volodya-lombrozo To me the motivation is quite simple here - we are trying to produce minimum required number of elements. In this particular case to understand what the constructor does I have to drill down into static method which is purely code extraction. Whereas having code declared within constructor I can see the whole logic in one place.

volodya-lombrozo: @mximp I understand your point. It makes sense, actually. The main problem is that we spend time on discussing and further refactoring. I believe we just have to agree on one thing.

yegor256:
@rultor merge

PR #1747, opened 2023-02-03
Mandatory empty line after metas

### LINKED ISSUE #1604: empty line between metas and the first object must be mandatory
Now, if I add this code to `TyposTest`, it doesn't raise a parsing error:

```
+package abc
[] > foo
```

But it should, because there is no empty line between the metas and the object.

mximp:
closes #1604


mximp:

eo-parser/src/main/java/org/eolang/parser/XeListener.java
```diff
@@ -158,7 +160,15 @@ public void enterMetas(final ProgramParser.MetasContext ctx) {
 
     @Override
     public void exitMetas(final ProgramParser.MetasContext ctx) {
-        // This method is created by ANTLR and can't be removed
+        if (new MissingEmptyLineAfterMetas().test(ctx.getText())) {
```
yegor256, line 163: @mximp looks like a hack to me. Can't we do the same just by using `Program.g4` grammar rules?

mximp: @yegor256 I didn't mange to invent such grammar change which would keep the grammar unambiguous..

volodya-lombrozo (commented):
It would be nice to catch that using `Program.g4` grammar rather than using `java`. Maybe we can apply something like the next: 
```
metas
  :
  (META EOL)+
  EOL
  ;
```
@mximp what do you think?
___
If it's impossible, all the rest looks good to me.

mximp:
> ```
> metas
>   :
>   (META EOL)+
>   EOL
>   ;
> ```

@volodya-lombrozo @yegor256 This would allow up to 2 empty lines after metas as `EOL` definition assumes optional empty line. I believe want exactly single empty line between objects and metas, don't we?

mximp:
I was thinking about defining it like this:
```
metas
  :
  (META EOL)*
  META
  LINEBREAK 
  LINEBREAK
  ;
```
But that seems to be ambiguous definition...

volodya-lombrozo:
> I was thinking about defining it like this:
> 
> ```
> metas
>   :
>   (META EOL)*
>   META
>   LINEBREAK 
>   LINEBREAK
>   ;
> ```
> 
> But that seems to be ambiguous definition...

@mximp What about the next:
```
metas
  :
  (META)+
  EOL
  ;
```
?

mximp:
> @mximp What about the next:
> ```
> metas
>   :
>   (META)+
>   EOL
>   ;
> ``` 
> ?

@volodya-lombrozo This would not allow empty lines between metas which I believe we want to keep it.
@yegor256 please confirm.

mximp:
@yegor256 @volodya-lombrozo I found the solution with grammar-only changes.. please recheck.

yegor256:
@mximp we don't want to allow empty lines between metas

mximp:
> @mximp we don't want to allow empty lines between metas

@yegor256 That's exactly what suggested changes ensure.

volodya-lombrozo (changes requested):
@mximp I still believe we can solve that problem simpler without adding new entities:
```antlr
metas
  :
  META+
  EOL
  ;

META: PLUS NAME (SPACE | ([\p{Letter}\p{General_Category=Decimal_Number}_"#$&<>'.-/@!\\:-] | LB | RB | SIGMA | XI | LSQ | RSQ | RHO)+)* LINEBREAK;
```
or even:
```antlr
metas
  :
  META+
  EOL
  ;

META: PLUS NAME (SPACE | PACKAGE)* LINEBREAK;
PACKAGE: ([\p{Letter}\p{General_Category=Decimal_Number}_"#$&<>'.-/@!\\:-] | LB | RB | SIGMA | XI | LSQ | RSQ | RHO)+;
```

Moreover, notice that we have the mess here: `_"#$&<>'.-/@!\\:-] | LB | RB | SIGMA | XI | LSQ | RSQ | RHO` - it looks like a next puzzle - we can define more precise "package" syntax and remove `broken-aliases.xsl` check, because we will able to catch such exceptions during parsing.

___
I can be mistaken with the exact implementation, but it looks like it's possible to solve that issue without `SEOL` and `DEOL`  which hard to read and understand, as for me.

mximp:
> @mximp I still believe we can solve that problem simpler without adding new entities:
> 
> ```antlr
> metas
>   :
>   META+
>   EOL
>   ;
> 
> META: PLUS NAME (SPACE | ([\p{Letter}\p{General_Category=Decimal_Number}_"#$&<>'.-/@!\\:-] | LB | RB | SIGMA | XI | LSQ | RSQ | RHO)+)* LINEBREAK;
> ```
> 
> or even:
> 
> ```antlr
> metas
>   :
>   META+
>   EOL
>   ;
> 
> META: PLUS NAME (SPACE | PACKAGE)* LINEBREAK;
> PACKAGE: ([\p{Letter}\p{General_Category=Decimal_Number}_"#$&<>'.-/@!\\:-] | LB | RB | SIGMA | XI | LSQ | RSQ | RHO)+;
> ```
> 
> Moreover, notice that we have the mess here: `_"#$&<>'.-/@!\\:-] | LB | RB | SIGMA | XI | LSQ | RSQ | RHO` - it looks like a next puzzle - we can define more precise "package" syntax and remove `broken-aliases.xsl` check, because we will able to catch such exceptions during parsing.
> 
> I can be mistaken with the exact implementation, but it looks like it's possible to solve that issue without `SEOL` and `DEOL` which hard to read and understand, as for me.

@volodya-lombrozo you suggestion still allows 2 empty lines. The issue is with `EOL` definition - it allows both empty line and no empty line. It's not possible to define exactly one empty line with such optional definition.
I believe `EOL` definition is misleading in the fist place and needs to be split into two different tokens.
Also I don't think definition of `META` that you suggested is more readable: it becomes more complex, with more elements.
I believe `DEOL` (double end-of-line) and `SEOL` (single end-of-line) is much better alternative.
As for package parsing - I guess we can indeed create separate task to define more granular syntax for various `META`s.

My suggestion is to merge this PR as is.

volodya-lombrozo:
> > @mximp I still believe we can solve that problem simpler without adding new entities:
> > ```antlr
> > metas
> >   :
> >   META+
> >   EOL
> >   ;
> > 
> > META: PLUS NAME (SPACE | ([\p{Letter}\p{General_Category=Decimal_Number}_"#$&<>'.-/@!\\:-] | LB | RB | SIGMA | XI | LSQ | RSQ | RHO)+)* LINEBREAK;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or even:
> > ```antlr
> > metas
> >   :
> >   META+
> >   EOL
> >   ;
> > 
> > META: PLUS NAME (SPACE | PACKAGE)* LINEBREAK;
> > PACKAGE: ([\p{Letter}\p{General_Category=Decimal_Number}_"#$&<>'.-/@!\\:-] | LB | RB | SIGMA | XI | LSQ | RSQ | RHO)+;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Moreover, notice that we have the mess here: `_"#$&<>'.-/@!\\:-] | LB | RB | SIGMA | XI | LSQ | RSQ | RHO` - it looks like a next puzzle - we can define more precise "package" syntax and remove `broken-aliases.xsl` check, because we will able to catch such exceptions during parsing.
> > I can be mistaken with the exact implementation, but it looks like it's possible to solve that issue without `SEOL` and `DEOL` which hard to read and understand, as for me.
> 
> @volodya-lombrozo you suggestion still allows 2 empty lines. The issue is with `EOL` definition - it allows both empty line and no empty line. It's not possible to define exactly one empty line with such optional definition. I believe `EOL` definition is misleading in the fist place and needs to be split into two different tokens. Also I don't think definition of `META` that you suggested is more readable: it becomes more complex, with more elements. I believe `DEOL` (double end-of-line) and `SEOL` (single end-of-line) is much better alternative. As for package parsing - I guess we can indeed create separate task to define more granular syntax for various `META`s.
> 
> My suggestion is to merge this PR as is.

1. You can use `LINEBREAK` instead of `EOL`:

```antlr4
metas
  :
  META+
  LINEBREAK
  ;

```

2. As for readable of my `META` - I agree, it's why I suggested to add puzzle

mximp:
> 1. You can use `LINEBREAK` instead of `EOL`:
> 
> ```
> metas
>   :
>   META+
>   LINEBREAK
>   ;
> ```

@volodya-lombrozo `LINEBREAK` cannot be used as it's a _fragment_ and discouraged to use directly within the rules.

Could you please re-iterate what are the issues of my suggestion?

volodya-lombrozo:
> > 1. You can use `LINEBREAK` instead of `EOL`:
> > 
> > ```
> > metas
> >   :
> >   META+
> >   LINEBREAK
> >   ;
> > ```
> 
> @volodya-lombrozo `LINEBREAK` cannot be used as it's a _fragment_ and discouraged to be used directly within the rules.
> 
> Could you please re-iterate what are the issues of my suggestion?

1. Names `SEOG` and `DEOG` are confusing
2. We have to add more elements and change [java code](EoLexer).
3. Instead of changing 2 lines we have to change many of them in `Program.g4`
3.1. Moreover, that changes looks confusing too `(SEOL | DEOL)`, imho
4. Solution with `SEOG` and `DEOG` doesn't reveal the problem with `package` syntax definition and `broken-aliases.xsl`

mximp:
> > Could you please re-iterate what are the issues of my suggestion?
> 
> 1. Names `SEOG` and `DEOG` are confusing
> 2. We have to add more elements and change [java code](EoLexer).
> 3. Instead of changing 2 lines we have to change many of them in `Program.g4`
>    3.1. Moreover, that changes looks confusing too `(SEOL | DEOL)`, imho
> 4. Solution with `SEOG` and `DEOG` doesn't reveal the problem with `package` syntax definition and `broken-aliases.xsl`

Mitigations:
1. We can rename the `SEOL` and `DEOL` to `SINGLE_EOL` and `DOUBLE_EOL` which I believe should eliminate any confusion and reflects actual meaning. Moreover these definitions gives flexibility specifying where we want to have single or double line breaks
2. Lexer just reflects grammar changes. Java code doesn't become anyhow more complex then it's now.
3. I don't think the number of changed lines is a valid metric here. Maintainability and readability are what we should aimed at. And as I mentioned above I believe changes you suggested make META definition much less readable then it's now. `SINGLE_EOL | DOUBLE_EOL` - clearly states single or double end-of-line is expected. 
4. package syntax issue is somewhat separate from what we are solving in this PR. We should raise separate task for it with clear description of the problem.

@volodya-lombrozo do you agree?

volodya-lombrozo:
> > > Could you please re-iterate what are the issues of my suggestion?
> > 
> > 
> > 
> > 1. Names `SEOG` and `DEOG` are confusing
> > 2. We have to add more elements and change [java code](EoLexer).
> > 3. Instead of changing 2 lines we have to change many of them in `Program.g4`
> >    3.1. Moreover, that changes looks confusing too `(SEOL | DEOL)`, imho
> > 4. Solution with `SEOG` and `DEOG` doesn't reveal the problem with `package` syntax definition and `broken-aliases.xsl`
> 
> Mitigations:
> 
> 1. We can rename the `SEOL` and `DEOL` to `SINGLE_EOL` and `DOUBLE_EOL` which I believe should eliminate any confusion and reflects actual meaning. Moreover these definitions gives flexibility specifying where we want to have single or double line breaks
> 2. Lexer just reflects grammar changes. Java code doesn't become anyhow more complex then it's now.
> 3. I don't think the number of changed lines is a valid metric here. Maintainability and readability are what we should aimed at. And as I mentioned above I believe changes you suggested make META definition much less readable then it's now. `SINGLE_EOL | DOUBLE_EOL` - clearly states single or double end-of-line is expected.
> 4. package syntax issue is somewhat separate from what we are solving in this PR. We should raise separate task for it with clear description of the problem.
> 
> @volodya-lombrozo do you agree?

1.  `SINGLE_EOL` and `DOUBLE_EOL` definitely look better, I agree
2. Yea... but in solution with `META` you won't have to touch Java code at all
3. As I understand `(SINGLE_EOL | DOUBLE_EOL) == EOL` at least for now. Why then we should divide `EOL` if we can don't do it :)
4. Agree, apparently we have to create a puzzle.

Once again, I prefer the solution with `META` because it appears to be simpler and addresses the issue of not having a proper definition of `PACKAGE`. The `SINGLE_EOL | DOUBLE_EOL` solution appears to be effective, but I have concerns that it may not provide a long-term solution and may only serve as a temporary fix. Please keep in mind that these are simply my thoughts and I may be mistaken in my assessment.

mximp:
> 1. `SINGLE_EOL` and `DOUBLE_EOL` definitely look better, I agree
> 2. Yea... but in solution with `META` you won't have to touch Java code at all
> 3. As I understand `(SINGLE_EOL | DOUBLE_EOL) == EOL` at least for now. Why then we should divide `EOL` if we can don't do it :)
> 4. Agree, apparently we have to create a puzzle.
> 
> Once again, I prefer the solution with `META` because it appears to be simpler and addresses the issue of not having a proper definition of `PACKAGE`. The `SINGLE_EOL | DOUBLE_EOL` solution appears to be effective, but I have concerns that it may not provide a long-term solution and may only serve as a temporary fix. Please keep in mind that these are simply my thoughts and I may be mistaken in my assessment.

@volodya-lombrozo 
1. Renamed.
2. We have classes which are aware of the grammar - I see no issues for such classes to be changed along with grammar.
3. `EOL` right now defines token which is not granular enough for us. It says "EOL may contain empty line or may not". Such definition will not allow to specify mandatory empty line as it will always conflict with EOL. So as I mentioned splitting it to two give better flexibility for rules we want to implement.
4. Can you please create new issue for PACKAGE problem as I'm fully aware of the context.

mximp:
@yegor256 please merge

yegor256:
@mximp I think the problem is that `EOL` consumes as many `\n\r` as it can find. I don't know why we did this, but it looks wrong. Try this:

```
metas
  :
  (META EOL)+
  EOL
  ;
EOL
  :
  LINEBREAK
  INDENT*
  ;
```

Some tests will break, but that's because in the grammar we have this _wrong_ assumption that EOL is any number of line-endings. I believe, we can fix the grammar making it more strict and correct.

mximp:
@yegor256 this would not allow empty lines between attributes and/or objects. Is that what's wanted?

yegor256:
@mximp yes, definitely

mximp:
@yegor256 Just to double stress: I believe the following would be unreadable:
```
[] > bytes
  # Equals to another object
  [x] > eq /bool
  # Total number of bytes
  [] > size /int
  # Represents a sub-sequence inside the current one
  [start len] > slice /bytes
  # Turn this chain of bytes into a string in UTF-8
  [] > as-string /string
  # Turn this chain of eight bytes into an integer.
  # If there are less or more than eight bytes, there will
  # be an error returned.
  [] > as-int /int
  # Turn this chain of eight bytes into a float.
  # If there are less or more than eight bytes, there will
  # be an error returned.
  [] > as-float /float
  # Calculate bitwise and
  [b...] > and /bytes
  # Calculate bitwise or
  [b...] > or /bytes
  # Calculate bitwise xor
  [b...] > xor /bytes
  # Calculate bitwise not
  [] > not /bytes
  # Calculate bitwise left shift
  [x] > left
    ^.right (x.neg) > @
  # Calculate bitwise right shift
  [x] > right /bytes
  # Converts to bool
  [] > as-bool
    ^.eq 01- > @
  # Return itself
  [] > as-bytes
    ^ > @
  # Converts this to hash
  [] > as-hash
    reduced. > @
      list
        bytes-as-array ^
      1
      [a b]
        plus. > @
          times.
            31
            a
          as-int.
            (0.as-bytes.and b).right 54
  # Concatenation of two byte sequences:
  # the current and the provided one,
  # as a new sequence
  [b] > concat /bytes
```

yegor256:
@mximp we should allow double EOL before a `<comment>`. Maybe not only allow, but make it mandatory.

mximp:
@yegor256 with recent changes I've corrected empty line allowance:
- empty lines are not allowed in between comments
- empty lines are not allowed between object definition and first element

Empty lines are still allowed between objects/attributes.

Issue #1253, opened 2022-09-21
Check if object is NaN or is Infinity while converting to Phi object


maxonfjvipon:
In some cases we can get `NaN` or `Infinities` as a result of object dataization.   
For example:  
- `0.0.div 0.0` => `nan`, 
- `1.0.div 0.0` => `positive-infinity`, 
- `(number nan).ln` => `nan` (natural logarithm, in progress)
- etc...

NaN and Ifinities are double, so I suggest to do something like this:
```java
private static Phi toPhi(final Object obj) {
    ...
    } else if (obj instanceof Double) {
        if (((Double) obj).isNaN()) {
            phi = new EOnan(Phi.Φ);
        } else {
            phi = new EOfloat(Phi.Φ);
        }
    }
    ....
}
```

And also add checking infinities here.  
In that case we won't get an error when we do for example `0.0.div 0.0`, we'll get `nan` object and will be able to work with it

yegor256:
@maxonfjvipon but we don't have `EOnan` in here

maxonfjvipon:
@yegor256 will not it be created on runtime?

yegor256:
@maxonfjvipon it belongs to eo-math package. We can't use objects from this package inside eo-runtime. Well, technically we can, but this would be violation of "layered architecture" principle.

Graur:
@maxonfjvipon @yegor256  We already talk about it here: https://github.com/objectionary/eo/issues/774#issuecomment-1164312509

maxonfjvipon:
@yegor256 yeah, I remember, but it's more complicated)  
Ok, I try to figure it out, but I think eventually it must be done somehow. `0.0.div 0.0` has not to throw an exception. It should return `nan` object.

maxonfjvipon:
@yegor256 I did a little research and here's what I think. 
Anyway we should move `nan` and `infinities` to `eo-runtime`. I believe it's the place where they belong to. Even in Java `NaN` belong to `Double` class. In other cases we won't be able to solve this problem.

After we move them we can solve it in two ways.
1. Easy. As I described above - add extra conditions to `toPhi`. There'll be no cross layering anymore because `nan`,`infinities` and `float` are in the same package.
2. Not easy and maybe not even worth to be discussed. As I suggested [here](https://github.com/objectionary/eo/issues/774): try to add some new EO syntax.

yegor256:
[...]

PR #1752, opened 2023-02-04
#1253 Moved nan and infinites to `eo-runtime` and placed them near float

maxonfjvipon:
Closes: #1253


maxonfjvipon:
@Graur please take a look at this: `sources to XMIRs
Error:  net.sf.saxon.style.XSLText@26285bd0: Errors identified: Data objects can't be used directly: "org.eolang.float";`

It seems we can't do `float.nan` because `float` is data object. If we could we will be also able to do `4.2.nan` which is kind of wrong. 

Any suggestions? Maybe we should just make these objects as separate objects?

Graur:
@maxonfjvipon Can you try to use parent `^` object instead of `float` key word inside `float` object? Maybe this error is about this mistake

maxonfjvipon:
@Graur seems it was not only about using it inside `float` object. 
This screenshot has 50 matches with `Data objects can't be used directly: "org.eolang.float"`. Also it's 50 times when `float.something` is used it float tests. It seems to it's just not allowed at all.

image

Graur:
@maxonfjvipon I see. Maybe we need to make `nan` as separated object? And place it near the `float`?
@yegor256 What do you think?

Graur:
@maxonfjvipon I see. Maybe we need to make `nan` as separated object? And place it near the `float`?
@yegor256 What do you think?

maxonfjvipon:
@yegor256 please take a look

yegor256:
@rultor merge

maxonfjvipon:
> @maxonfjvipon I see. Maybe we need to make `nan` as separated object? And place it near the `float`?
> 
> @yegor256 What do you think?

@yegor256 please take a look at this. Should we make these objects as separate objects since we can't place them inside `float` because it's data object?

Graur:
@maxonfjvipon Maybe you can try to implement it and watch how it will be going?

maxonfjvipon:
@Graur do I understand correctly that it fails because there's not enough tests?

Graur:
@maxonfjvipon Looks like you are right

maxonfjvipon:
@Graur could you please check this out? What can be wrong? It seems that one more test is needed but I wrote tests for all objects inside nan, +inf and -inf?

Graur:
@maxonfjvipon Check this property: https://github.com/objectionary/eo/blob/master/eo-runtime/pom.xml#L225-L227

maxonfjvipon:
@Graur please take a look

Graur:
@maxonfjvipon What about tests like `0.0.eq nan`? I don't see any of them here

maxonfjvipon:
@Graur yeah, my bad

maxonfjvipon:
@Graur please take a look

eo-runtime/src/main/eo/org/eolang/nan.eo
```diff
@@ -0,0 +1,100 @@
...
+  [x] > minus
+    ^ > @
+
+  # Quotient of the division of $ by x
+  [x] > div
+    ^ > @
+
+  # Converts this to bytes
+  [] > as-bytes
+    7F-C0-00-00 > @
```
Graur, line 92: @maxonfjvipon Can you remind me, this value was taken from IEEE standart?

Graur (changes requested):

eo-runtime/src/main/eo/org/eolang/nan.eo
maxonfjvipon: @Graur yes, you can try it [here](https://www.h-schmidt.net/FloatConverter/IEEE754.html)

maxonfjvipon: @Graur yes, you can try it [here](https://www.h-schmidt.net/FloatConverter/IEEE754.html)

image

Graur: @maxonfjvipon Thanks!

Graur (approved):
@maxonfjvipon Now it looks better, thanks!
@yegor256 Please, have a look

maxonfjvipon:
@yegor256 please take a look

yegor256:
@rultor merge

Issue #1791, opened 2023-02-11
Allow division by float zero in `EOfloat$EOdiv`


maxonfjvipon:
Allow division by float zero in `EOfloat$EOdiv` for being able to make `nan`, `positive-infinity` and `negative-infinity` decorated `0.0.div 0.0`, `1.0.div 0.0` and `-1.0.div 0.0`

This task had to be created by `pdd` bot after [this PR](#1752) but it hadn't for some reason.

@Graur please assign me

Graur:
@maxonfjvipon Sure, go ahead. We will fix `pdd` later.
The budget is 30 min.

0pdd:
@maxonfjvipon the puzzle [#1803](https://github.com/objectionary/eo/issues/1803) is still not solved.

0pdd:
@maxonfjvipon the only puzzle [#1803](https://github.com/objectionary/eo/issues/1803) is solved here.

PR #1792, opened 2023-02-11
allowed float.div 0.0 (#1791)

maxonfjvipon:
Closes: #1791

### DIFF
--- a/eo-runtime/src/main/eo/org/eolang/nan.eo
+++ b/eo-runtime/src/main/eo/org/eolang/nan.eo
@@ -39,7 +39,7 @@
 #  decorated 0.0.div 0.0, 1.0.div 0.0 and -1.0.div 0.0 accordingly
 # Not a number
 [] > nan
-  0.0.div 0.0 > @
+  0.0.div 0.0 > @!
 
   # Tests that $ = x
   [x] > eq
@@ -91,7 +91,7 @@
 
   # Converts this to bytes
   [] > as-bytes
-    7F-F8-00-00-00-00-00-00 > @
+    ^.@.as-bytes > @
 
   # Converts this to int
   [] > as-int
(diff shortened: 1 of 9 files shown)


maxonfjvipon:
@Graur please take a look

eo-runtime/src/main/eo/org/eolang/nan.eo
```diff
@@ -89,7 +97,7 @@
 
   # Converts this to bytes
   [] > as-bytes
-    7F-C0-00-00 > @
+    ^.@.as-bytes > @
```
Graur, line 100: @maxonfjvipon I don't get this part. Can you provide some short explanation?

Graur (changes requested):

eo-runtime/src/main/eo/org/eolang/nan.eo
maxonfjvipon: `7F-C0-00-00` - is float NaN (4 bytes). 
On the last meeting we decided that float in `eo` is a double in Java - 8 bytes. 
Double NaN of 8 bytes - `7F-F8-00-00-00-00-00-00` (this value also will be got as result of `0.0.div 0.0`)

I tried to hardcode this bytes values and tests passed on my machine, but they fails in CI for some reason (you can check one of the previous commits)

So (0.0.div 0.0).as-bytes is kinda safe

Graur: @maxonfjvipon I agree, thanks!

Graur (approved):
@maxonfjvipon Looks good to me, thanks!
@yegor256 Please, have a look

maxonfjvipon:
@yegor256 please take a look

yegor256:
@rultor merge

PR #1794, opened 2023-02-11
Extented Footprint description

### LINKED ISSUE #1786: What is the footprint for?
It is not clear from the comments and tests what the `Footprint` is for and what its methods do. What and where do they `save` and `load`? At first I thought it was like a log, but this class is created anew in the places of use every time.
I guess the description of this class should be made clearer.
@Graur WDYT?

levBagryansky:
Closes #1786 


eo-maven-plugin/src/main/java/org/eolang/maven/footprint/Footprint.java
```diff
@@ -29,7 +29,8 @@
 import org.cactoos.Scalar;
 
 /**
- * Program footprint of EO compilation process.
+ * Program footprint of EO compilation process,
+ * convenient eo program management mechanism.
```
volodya-lombrozo, line 33: @levBagryansky It's still not so clear as for me. Maybe we can add more text here like:

>   Footprint refers to a trace of EO program on file system which is quite fortunate analogy with natural meaning of the word.

> The idea of Footprint is to provide abstraction for EO program files content. 

The general idea of that PR is to add one or two sentences that clearly will explain what is `Footprint` and what problem it solves.

volodya-lombrozo: Or even better - we can provide examples.

volodya-lombrozo (changes requested):

volodya-lombrozo:
@levBagryansky I would also suggest you to read that [doc section](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) and correct the PR description a little bit.

eo-maven-plugin/src/main/java/org/eolang/maven/footprint/Footprint.java
levBagryansky: > Or even better - we can provide examples.
Won't the examples be superfluous? The tests already illustrate the capabilities of the class quite well as for me.

volodya-lombrozo (approved):
@levBagryansky Thanks.

levBagryansky:
@yegor256 please check

yegor256:
@levBagryansky @volodya-lombrozo the contribution of this PR is pretty small, if you ask me. Maybe we can add more text, deeper explanation of the meaning and the purpose of this class?

eo-maven-plugin/src/main/java/org/eolang/maven/footprint/Footprint.java
volodya-lombrozo: @levBagryansky I don't think it will be superfluous. Moreover, you can use simple natural words instead of code. It could explain more and in a single place.

volodya-lombrozo:
@levBagryansky Are you going to continue?

levBagryansky:
@volodya-lombrozo The meaning of this class is not so deep. In fact, it just allows to read and create files.

volodya-lombrozo (changes requested):
@levBagryansky Could you please check the box in the pull request that says “Allow edits from maintainers.” 
___

Or you can add new text yourself:
```java
package org.eolang.maven.footprint;

import java.io.IOException;
import java.nio.file.Path;
import java.util.List;
import org.cactoos.Scalar;

/**
 * Footprint is a term used to refer to a trace of an EO program on the file system.
 * It's a fitting analogy with the word's natural meaning, and what it essentially does is trace
 * files that have a specific extension. Footprint enables you to do the following:
 * - Save files that have a specific extension
 * - Load files that have a specific extension
 * - Get a complete list of files that have a specific extension
 * You can find a full list of available methods below.
 * Since Footprint can leave traces on the file system, each Footprint instance will have its
 * own unique way of saving and loading files, and it will have its own place to do so.
 * For instance, {@link FtDefault} will save files directly to a specific location on the file
 * system, while {@link FtCached} will first check the cache directory before delegating
 * the behavior.
 * In other words, Footprint provides useful functionality for working with files that
 * have different extensions.
 * @since 1.0
 */
public interface Footprint {

    /**
     * Get file content of a specific extension.
     * If the file "hello.txt" contains the string "hello", it will return the string "hello".
     * @param program File name
     * @param ext File extension which defines the type
     * @return Content of a file
     * @throws IOException In case of IO issue.
     */
    String load(String program, String ext) throws IOException;

    /**
     * Save content. Leaves a "footprint" in the directory.
     * So it can create a new file in the file system.
     * Where it will create this file depends on the implementation.
     * @param program Program name
     * @param ext File extension
     * @param content File content
     * @throws IOException In case of IO issues
     */
    void save(String program, String ext, Scalar<String> content) throws IOException;

    /**
     * Get list of saved regular files with a particular extension.
     * @param ext File extension
     * @return List of files
     * @throws IOException In case of IO issues
     */
    List<Path> list(String ext) throws IOException;
}

```

What do you think about that description?

levBagryansky:
@volodya-lombrozo I like it! The checkbox is aleready enabled:
image

volodya-lombrozo:
@levBagryansky I can't add commit to that PR:
```
ERROR: Permission to levBagryansky/eo.git denied to volodya-lombrozo.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Please, add the code above to the PR.

volodya-lombrozo:
@yegor256 Reminder

yegor256:
@rultor merge

Issue #1821, opened 2023-02-20
pom.xml:236: After the new version of the eo-runtime...


0pdd:
The puzzle `1790-b6768978` from #1790 has to be resolved: 

https://github.com/objectionary/eo/blob/140fd1deac603c40979304bc40b913387f2b22c0/eo-maven-plugin/pom.xml#L236-L236

The puzzle was created by rultor on 20-Feb-23. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #1822, opened 2023-02-20
feat(#1821) Move integration tests into a separate module

volodya-lombrozo:
Since it's impossible to run integration tests on `parent` level, I have created one more module with integration tests and placed it after all other modules.
[Discussion](https://stackoverflow.com/questions/39612692/how-can-i-run-integration-tests-after-building-all-modules-in-a-multi-module-mav).

We steel need to deploy tuple to objectionary to make integration tests work.

Closes: #1821


volodya-lombrozo:
@mximp Could you have a look, please?

yegor256:
@volodya-lombrozo I understand the intent, but this is not how maven projects are tested, according to Maven philosophy (as I understand it). `it` tests are not "integration tests", but "invoker tests", which are supposed to be located in the same module where the Maven Plugin is located. we test the plugin, not the entire reactor. Also, having a special module just for tests (even if they are integration), is a bad idea, I believe. Mostly because, such a detachment of tests from the "system under test" is risky -- it will be harder to catch errors right when they happen.

volodya-lombrozo:
@yegor256 

**ChatGTP**:
> In a Maven project, the `src/it` directory typically stands for "integration tests" and is a standard directory for placing integration test source code and resources.
Integration tests are used to test how different components of a system work together, as opposed to unit tests that test individual units of code in isolation. The `src/it` directory is used to store integration test code and resources, such as configuration files, test data, and scripts.
When you run mvn verify on a Maven project, the integration tests in the `src/it` directory are executed along with the unit tests in the `src/test` directory. The results of the integration tests are reported separately from the unit test results.
It's worth noting that the directory structure of a Maven project is configurable, so it's possible for a project to have a different directory structure than the standard one. However, the `src/it` directory is a commonly used convention for integration tests in Maven projects.

volodya-lombrozo:
@yegor256 We have the significant problem that decrease speed of development. Actually we have cyclic dependency.

1. `eo-maven-plugin` requires `eo-runtime` to run integration tests
2. `eo-runtime` requires `eo-maven-plugin` to build itself.
3.  Integration tests require both `eo-maven-plugin` and `eo-runtime`

mximp:
@volodya-lombrozo @yegor256 
Actually the issue with our set up is that `eo-maven-plugin` uses fixed (downgraded) version of `eo-runtime` and raw version of `eo-parser`. Changes in core parsing/syntax normally require modification in both `eo-runtime` and `eo-parser`. When it happens `eo-maven-plugin` integration tests become invalid. 
I believe we could solve the issue by fixing `eo-parser` within integration `pom` to whatever is compatible with fixed `eo-runtime`. With such set up IT tests will be decoupled from the changes we make and `eo-maven-plugin` won't fail.

volodya-lombrozo:
> @volodya-lombrozo @yegor256
> Actually the issue with our set up is that `eo-maven-plugin` uses fixed (downgraded) version of `eo-runtime` and raw version of `eo-parser`. Changes in core parsing/syntax normally require modification in both `eo-runtime` and `eo-parser`. When it happens `eo-maven-plugin` integration tests become inconsistent.
> I believe we could solve the issue by fixing `eo-parser` within integration `pom` to whatever is compatible with fixed `eo-runtime`. With such set up IT tests will be decoupled from the changes we make and `eo-maven-plugin` won't fail.

@mximp  Do I understand correctly, that you are going to separate `eo-maven-plugin` from current versions of `eo-parser` and `eo-runtime` and use only stable already deployed versions?

volodya-lombrozo:
@yegor256 @mximp To he honest, I believe that it's good to have integration tests for the entire set of `eo-parser`, `eo-runtime` and `eo-maven-plugin` modules all together. It's quite useful to test everything works correctly with current versions.

yegor256:
@mximp @volodya-lombrozo I think it's a good idea: to use the same version of eo-parser and eo-runtime in _invoker_ tests.

@volodya-lombrozo check this out: https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html (notice that `src/it` is primarily for plugins)

volodya-lombrozo:
@mximp @yegor256 If we want to divide `eo-parser`, `eo-runtime`, `eo-maven-plugin`, then why do we keep them in a single repository? It's a quite confusing.

@yegor256 primaly != always.

yegor256:
@volodya-lombrozo we don't want to divide them! I believe that should stay together because they depend on each other. If, for example, we move our parser away, it will be very hard to synchronize the changes made to the syntax with the changes in the runtime. We will be able to modify the parser and the runtime will fail, but later, when the parser is already released. Keeping them in sync will be a tougher task than now.

volodya-lombrozo:
> @volodya-lombrozo we don't want to divide them! I believe that should stay together because they depend on each other. If, for example, we move our parser away, it will be very hard to synchronize the changes made to the syntax with the changes in the runtime. We will be able to modify the parser and the runtime will fail, but later, when the parser is already released. Keeping them in sync will be a tougher task than now.

@yegor256 Now I understand. We want to keep/change them together, it's fine. Do you agree that it would be nice to check if everything altogether works correctly (especially if we want to synchronise the changes)? If you will set some hardcoded (fixed) versions of `eo-runtime` or `eo-parser` in integration tests then you will check how a new `eo-maven-plugin` works with old  `eo-runtime` and `eo-parser` libraries, it's strange and creates some problems with updating `eo-maven-plugin` (at least I have to disable integration tests in my case).

volodya-lombrozo:
If we don't need common integration tests, I close that PR. In the future when we will face with the problem again, we can revive that PR.

Issue #1819, opened 2023-02-20
Flaky tests breaks pipelines (`StackOverflowError `)


volodya-lombrozo:
We again have a problem with flaky tests in `eo-runtime`. The common problem is `StackOverflowError `:
```
EOcompares_string_with_nan_and_infinitesTest.testWorks:127->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:255->PhDefault.attr:256 » Ex EOorg.EOeolang.EOstringν115936="Unexpected error 'null' of type StackOverflowError; caused by "SF
```

Failed pipelines:
1. https://github.com/objectionary/eo/actions/runs/4182605213/jobs/7245904469
2. https://github.com/objectionary/eo/actions/runs/4221651017/jobs/7329203665

maxonfjvipon:
@volodya-lombrozo in first pipeline `EOnegative_infinity_eq_is_validTest` does not exist anymore. Can you merge `master` into your branch and try again?

volodya-lombrozo:
I have [that](https://github.com/objectionary/eo/actions/runs/4221651017/jobs/7329203665#step:5:42563) errors couple of hours ago where master branch was merged. Do you know the reason for such a behaviour? Or at least place where it comes from.

maxonfjvipon:
@volodya-lombrozo not really, there's nothing special about `nan`, `+inf` or `-inf`, they're just regular floats. Did the tests pass on your local machine?

volodya-lombrozo:
@maxonfjvipon Not always

maxonfjvipon:
@volodya-lombrozo I think this test may be a reason of such behavior:
```

[] > float-equal-to-nan-and-infinites-is-false
  assert-that > @
    and.
      (0.0.eq nan).eq FALSE
      (0.0.eq positive-infinity).eq FALSE
      (0.0.eq negative-infinity).eq FALSE
      (42.5.eq nan).eq FALSE
      (42.5.eq positive-infinity).eq FALSE
      (42.5.eq negative-infinity).eq FALSE
    $.equal-to TRUE
```

Need to divide to several tests

volodya-lombrozo:
@maxonfjvipon Apparently you right. I have tried to make that test even worse:
```eo
[] > float-equal-to-nan-and-infinites-is-false
  assert-that > @
    and.
      (0.0.eq nan).eq FALSE
      (0.0.eq positive-infinity).eq FALSE
[...]

PR #1830, opened 2023-02-21
Divided float and nan tests (#1819)

maxonfjvipon:
Closes: #1819

### DIFF
--- a/eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
+++ b/eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
@@ -69,10 +69,9 @@
         first.eq 0
       nan
       if.
-        not.
-          first.eq ^
-        ^
+        first.eq ^
         positive-infinity
+        ^
 
   # Sum of $ and x
   [x...] > plus
--- a/eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
+++ b/eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
@@ -136,6 +136,3 @@
         first.eq positive-infinity
       nan
       ^
-
-  [] > as-bytes
-    ^.@.as-bytes > @
--- a/eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
+++ b/eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
@@ -61,10 +61,15 @@
   # Multiplication of $ and x
   [x...] > times
     x.at 0 > first!
+
+    # here we use the rule "NaN != NaN is TRUE" to check if "num" is NaN
+    [num] > check-nan
+      not. > @
+        num.eq num
+
     if. > @
       or.
-        not.
-          first.eq first
+        check-nan first
         first.eq 0.0
         first.eq 0
       nan
@@ -127,10 +132,15 @@
   # Quotient of the division of $ by x
   [x...] > div
     x.at 0 > first!
+
+    # here we use the rule "NaN != NaN is TRUE" to check if "num" is NaN
+    [num] > check-nan
+      not. > @
+        num.eq num
+
     if. > @
       or.
-        not.
-          first.eq first
+        check-nan first
         first.eq ^
         first.eq positive-infinity
       nan
--- a/eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
+++ b/eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
@@ -28,7 +28,7 @@
 
 # Negative infinity
 [] > negative-infinity
-  -1.0.div 0.0 > @
+  -1.0.div 0.0 > @!
 
   # Tests that $ = x
   [x] > eq
@@ -136,3 +136,6 @@
         first.eq positive-infinity
       nan
       ^
+
+  [] > as-bytes
+    ^.@.as-bytes > @
--- a/eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
+++ b/eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
@@ -28,7 +28,7 @@
 
 # Negative infinity
 [] > negative-infinity
-  -1.0.div 0.0 > @!
+  -1.0.div 0.0 > @
 
   # Tests that $ = x
   [x] > eq
@@ -63,14 +63,16 @@
     x.at 0 > first!
     if. > @
       or.
-        first.as-bytes.eq (nan.as-bytes)
+        not.
+          first.eq first
         first.eq 0.0
         first.eq 0
       nan
       if.
-        first.eq ^
-        positive-infinity
+        not.
+          first.eq ^
         ^
+        positive-infinity
 
   # Sum of $ and x
   [x...] > plus
@@ -128,12 +130,9 @@
     x.at 0 > first!
     if. > @
       or.
-        first.as-bytes.eq (nan.as-bytes)
+        not.
+          first.eq first
         first.eq ^
         first.eq positive-infinity
       nan
       ^
-
-  # Converts this to bytes
-  [] > as-bytes
-    ^.@.as-bytes > @!
--- a/eo-runtime/src/test/eo/org/eolang/float-tests.eo
+++ b/eo-runtime/src/test/eo/org/eolang/float-tests.eo
@@ -103,16 +103,33 @@
       10.9
     $.equal-to FALSE
 
-[] > float-equal-to-nan-and-infinites-is-false
+[] > float-zero-not-equal-to-nan
   assert-that > @
-    and.
-      (0.0.eq nan).eq FALSE
-      (0.0.eq positive-infinity).eq FALSE
-      (0.0.eq negative-infinity).eq FALSE
-      (42.5.eq nan).eq FALSE
-      (42.5.eq positive-infinity).eq FALSE
-      (42.5.eq negative-infinity).eq FALSE
-    $.equal-to TRUE
+    0.0
+    $.not
+      $.equal-to
+        nan
+
+[] > float-not-equal-to-nan
+  assert-that > @
+    52.4
+    $.not
+      $.equal-to
+        nan
+
+[] > float-not-equal-to-positive-infinity
+  assert-that > @
+    52.4
+    $.not
+      $.equal-to
+        positive-infinity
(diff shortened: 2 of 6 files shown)


maxonfjvipon:
@volodya-lombrozo here we are again)

volodya-lombrozo:
> @volodya-lombrozo here we are again)

@maxonfjvipon Looks good to me. But seems it haven't helped.

maxonfjvipon:
@volodya-lombrozo btw it says `IllegalState Recursive update`not `StackOverflow`, do you know what it means?

volodya-lombrozo:
Maybe it's a different problem:
```
EOpositive_infinity_times_negative_infinityTest.testWorks:66->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256 » IllegalState Recursive update
```

volodya-lombrozo:
@maxonfjvipon Apparently it related to the `ConcurrentHashMap` and `caching`:

> [IllegalStateException](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/IllegalStateException.html) - if the computation detectably attempts a recursive update to this map that would otherwise never complete

From [here](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ConcurrentHashMap.html). Also related question exists - [here](https://stackoverflow.com/questions/71087512/concurrenthashmap-throws-recursive-update-exception) it is.

Probably it related to our caching mechanisms, but for now I can't say more - we need to investigate that problem,.

volodya-lombrozo:
@maxonfjvipon You also can try to replace the next code from `PhConst`:
```java
    @Override
    public Attr attr(final String name) {
        return this.cached.computeIfAbsent(
            name, x -> new AtConst(this.origin.attr(name), this)
        );
    }
```
with that:
```java
    @Override
    public Attr attr(final String name) {
        if (!this.cached.containsKey(name)) {
            this.cached.put(name, new AtConst(this.origin.attr(name), this));
        } 
        return this.cached.get(name);
    }
```

maxonfjvipon:
@volodya-lombrozo WDYT?

volodya-lombrozo:
> @volodya-lombrozo WDYT? I removed `@!` from `nan` and infintes, seems work. Should I touch this `PhConst`?

@maxonfjvipon if `@!` does improve performance I would add a puzzle that will describe the problem that we faced. Maybe in the future we will able to fix it properly.

maxonfjvipon:
@volodya-lombrozo I think the problem is in `as-bytes` object. Don't really know why, but without it and with `@!` it seems works fine.

eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
```diff
@@ -63,14 +63,16 @@
...
-        first.as-bytes.eq (nan.as-bytes)
+        not.
+          first.eq first
         first.eq 0.0
         first.eq 0
       nan
       if.
-        first.eq ^
-        positive-infinity
+        not.
```
volodya-lombrozo, line 72: @maxonfjvipon I believe it's much easier to read `if` with positive check rather than with negation:
```eo
if.
  first.eq ^
  positive-infinity
  ^
```
What do you think?

eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
```diff
@@ -128,12 +130,9 @@
     x.at 0 > first!
     if. > @
       or.
-        first.as-bytes.eq (nan.as-bytes)
+        not.
+          first.eq first
```
volodya-lombrozo, line 134: @maxonfjvipon I don't clearly understand: Why the next code:
```eo
not.
  first.eq first
```
is equal to 
```eo
first.as-bytes.eq (nan.as-bytes)
```

eo-runtime/src/test/eo/org/eolang/float-tests.eo
```diff
@@ -255,71 +272,157 @@
...
-    and.
-      (num.eq cached-nan).eq FALSE
-      (num.lt cached-nan).eq FALSE
-      (num.lte cached-nan).eq FALSE
-      (num.gt cached-nan).eq FALSE
-      (num.gte cached-nan).eq FALSE
-    $.equal-to TRUE
+    as-bytes.
+      35.2.minus not-a-number
+    $.equal-to not-a-number.as-bytes
```
volodya-lombrozo, line 299: @maxonfjvipon If `as-bytes` makes problems in the pipelines, why are you using it here? is it different `as-bytes`?

eo-runtime/src/test/eo/org/eolang/string-tests.eo
```diff
@@ -116,18 +116,6 @@
     $.equal-to
       "hello".as-hash
 
-[] > compares-string-with-nan-and-infinites
```
volodya-lombrozo, line 119: @maxonfjvipon Why have you decided to remove that test?

volodya-lombrozo (changes requested):

eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
maxonfjvipon: @volodya-lombrozo it's NaN checking. We can say that number is NaN either compare its bytes to NaN bytes, or we can use the rule that NaN != NaN is TRUE

eo-runtime/src/test/eo/org/eolang/float-tests.eo
maxonfjvipon: @volodya-lombrozo see above about NaN checking

eo-runtime/src/test/eo/org/eolang/string-tests.eo
maxonfjvipon: @maxonfjvipon NaN and infinites has unique bytes representation and it seems there is no such string that has the same bytes as NaN of infinities

eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
maxonfjvipon: @volodya-lombrozo sure, I'll fix it

eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
volodya-lombrozo: @maxonfjvipon Maybe we can leave at least some comment above about such a behaviour? Because it's far from intuitive understanding.

eo-runtime/src/test/eo/org/eolang/string-tests.eo
volodya-lombrozo: @maxonfjvipon Maybe we can show that `positive-infinity != any_string`?

volodya-lombrozo (changes requested):

eo-runtime/src/test/eo/org/eolang/string-tests.eo
maxonfjvipon: @volodya-lombrozo do you mean something like that?
```
[] > compares-string-with-positive-infinity
  assert-that > @
    "друг"
    $.not
      $.equal-to positive-infinity
```

volodya-lombrozo: @maxonfjvipon Yes. As I understand the idea behind that test was almost the same.

volodya-lombrozo (changes requested):

maxonfjvipon:
@volodya-lombrozo please have a look

eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
```diff
@@ -61,9 +61,15 @@
   # Multiplication of $ and x
   [x...] > times
     x.at 0 > first!
+
```
volodya-lombrozo, line 64: @maxonfjvipon Is that blank line required?

eo-runtime/src/test/eo/org/eolang/nan-tests.eo
```diff
@@ -27,85 +27,122 @@
...
 
-[] > nan-lt-is-valid
+[] > nan-not-gte-nan
   nan > not-a-number!
   assert-that > @
-    and.
-      (not-a-number.lt 42).eq FALSE
-      (not-a-number.lt not-a-number).eq FALSE
-    $.equal-to TRUE
+    not-a-number.gte not-a-number
```
volodya-lombrozo, line 88: @maxonfjvipon We have similar code in several places:
```eo
not-a-number.gte not-a-number
  $.equal-to FALSE
```
Can we change it to something like:

```
$.not
  $.less-than not-a-number
``` 
or even:
```eo
$.greater-than-equals not-a-number
```
In order to avian using literals like `FALSE` or `TRUE` ?

volodya-lombrozo (changes requested):

eo-runtime/src/main/eo/org/eolang/negative-infinity.eo
maxonfjvipon: @volodya-lombrozo no, but I think it increases readability if we separate objects with an empty line

eo-runtime/src/test/eo/org/eolang/nan-tests.eo
maxonfjvipon: @volodya-lombrozo I did not found `greater-than-equals` in `eo-hamcrest` so it can't be used. Also here we test `nan.gte` object. If we do 

```
$.not
  $.less-than
```
`nan.gte` object will be never called

maxonfjvipon: @volodya-lombrozo WDYT?

volodya-lombrozo: @maxonfjvipon I see. Then let's leave it as is.

volodya-lombrozo (approved):
@maxonfjvipon Looks good to me. Thanks!

maxonfjvipon:
@yegor256 please have a look

yegor256:
@rultor merge

Issue #1842, opened 2023-02-24
target directories numerating


levBagryansky:
I started `mvn clean install` and looked at folders in target
```
tardis3@RyanGosling:~/eo$ la eo-runtime/target/eo
итого 236K
drwxrwxr-x 3 1000 1000 4,0K фев 24 21:06 01-parse/
drwxrwxr-x 3 1000 1000 4,0K фев 24 21:06 03-optimize/
drwxrwxr-x 3 1000 1000 4,0K фев 24 21:06 04-pull/
drwxrwxr-x 3 1000 1000 4,0K фев 24 21:06 05-pre/
drwxrwxr-x 4 1000 1000 4,0K фев 24 21:06 06-resolve/
drwxrwxr-x 3 1000 1000 4,0K фев 24 21:06 06-transpile/
-rw-rw-r-- 1 1000 1000  33K фев 24 21:07 foreign.csv
-rw-rw-r-- 1 1000 1000  11K фев 24 21:08 placed.csv
-rw-rw-r-- 1 1000 1000 163K фев 24 21:07 transpiled.csv
tardis3@RyanGosling:~/eo$ 
```
So there are 2 folders starting from `06-..`.
Is it ok?
@volodya-lombrozo WDYT

levBagryansky:
and where is `02-...` folder?

volodya-lombrozo:
@levBagryansky Indeed, the numbers are misleading. I believe we need to either understand and describe why the numbers are inaccurate (by updating the JavaDoc or performing a refactoring), or simply fix the numbering. Do you want to implement the issue?

volodya-lombrozo:
@yegor256 Did you put any thought into the numbering?

yegor256:
@volodya-lombrozo they just need to go one after another. If this is not the case, let's fix them.

levBagryansky:
@volodya-lombrozo it is easy to fix but I don't see the way to test it.

volodya-lombrozo:
@levBagryansky You can check after integration tests run. Like we did it [here](https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/it/custom_goals/verify.groovy).

yegor256:
[...]

PR #1845, opened 2023-02-26
feat(#1842): directories numbering

volodya-lombrozo:
Here are the changes I've made:

1. Added an additional integration test to verify correct directory naming after `eo-runtime` compilation.
2. Renamed folders and updated numbering to align with correct order.
3. Renamed the test `eo-runtime/src/test/groovy/verify.groovy` to `eo-runtime/src/test/groovy/check-all-java-classes-compiled.groovy`. The entry point is now `eo-runtime/src/test/groovy/verify.groovy`.
4. Created a new `CompilationStep` interface. Please note the `@todo` comment in the code. While I'm unsure if it's the optimal solution, it is an improvement over the previous static variables used.

Closes: #1842

### DIFF
--- a/eo-runtime/src/test/groovy/check-folders-numbering.groovy
+++ b/eo-runtime/src/test/groovy/check-folders-numbering.groovy
@@ -4,7 +4,7 @@
  * Copyright (c) 2016-2023 Objectionary.com
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
- * of this software and associated documentation files (the "Software"), to deal
+ * of this software and associated documentation directories (the "Software"), to deal
  * in the Software without restriction, including without limitation the rights
  * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
  * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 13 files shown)


volodya-lombrozo:

eo-runtime/src/test/groovy/check-folders-numbering.groovy
```diff
@@ -0,0 +1,40 @@
...
+})
+assert 6 == directories.size()
+[
+  target.resolve('1-parse').toFile(),
+  target.resolve('2-optimize').toFile(),
+  target.resolve('3-pull').toFile(),
+  target.resolve('4-resolve').toFile(),
+  target.resolve('5-pre').toFile(),
+  target.resolve('6-transpile').toFile(),
+].each { assert directories.contains(it) }
```
yegor256, line 40: @volodya-lombrozo if tomorrow we will have `5-something` directory there, your test won't detect it, right?

volodya-lombrozo: @yegor256 I've updated test logic little bit in order to catch case with `5-something`. Could you take a look again, please?

yegor256:
@rultor merge

PR #1855, opened 2023-02-27
fix: revert `maven-compiler-plugin` back to the version `3.8.1`

volodya-lombrozo:
Revert `maven-compiler-plugin` back to the version `3.8.1`.

The issue comes from that [PR](https://github.com/objectionary/eo/pull/1850).

### DIFF
--- a/eo-maven-plugin/src/it/fibonacci/pom.xml
+++ b/eo-maven-plugin/src/it/fibonacci/pom.xml
@@ -38,7 +38,7 @@ SOFTWARE.
     <plugins>
       <plugin>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.11.0</version>
+        <version>3.8.1</version>
         <configuration combine.self="override"/>
       </plugin>
       <plugin>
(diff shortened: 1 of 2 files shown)


volodya-lombrozo:
@yegor256 Could you take a look please? I believe we have to merge it as soon as possible, since it blocks other PRs. 

I also want to notice that such a problem have already happened before. The cause - we merge a PR that is actually [failled](https://github.com/objectionary/eo/pull/1850).
This is happens because `rultor` runs pipelines only one the one type of VM (Linux I guess) and ignores Windows at all, in other words `rultor` doesn't prevent from OS-specific mistakes. Maybe we have to invent some protection in order to prevent such situations in the future?

volodya-lombrozo:
@yegor256 Reminder

yegor256:
@rultor merge

yegor256:
@volodya-lombrozo if we implement [this feature](https://github.com/yegor256/rultor/issues/1237) in Rultor, the problem will be solved. Maybe you can help?

PR #1859, opened 2023-03-01
Refactor: add `TranspiledTojos`

volodya-lombrozo:
I believe we have an encapsulation problem related to the `Tojos` that work with the `eo-foreign.cvs`, `placed.cvs`, or `transpiled.cvs` files. We are currently using attributes from those files in various parts of the program, which presents the following issues:

1. We are using strange `static` variables with oddly composed names in many different places (`ATTR_DISCOVERED_AT`, `ATTR_PLD_RELATED`, `ATTR_PLD_HASH`).
2. We are exposing the internals of `Tojos` and its attributes to foreign objects. This means that other objects dictate how `Tojos` should save its internals.

To address this issue, I suggest a simple refactoring, for example, for the transpiled `Tojos` (but the same refactoring could be applied to `eo-foreign.cvs` and `placed.cvs` as well). Here are the steps we can take:

1. Create a `TranspiledTojos` class that will encapsulate all behaviour related to transpiled `Tojos`.
2. Inside `TranspiledTojos`, define a set of attributes as an enum (instead of using `AssembleMojo.ATTR_XMIR2` and `Tojos.KEY`).

Please take a look at the code.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/SafeMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/SafeMojo.java
@@ -46,6 +46,7 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
 import org.cactoos.scalar.Sticky;
 import org.cactoos.scalar.Unchecked;
+import org.eolang.maven.tojos.TranspiledTojos;
 import org.slf4j.impl.StaticLoggerBinder;
 
 /**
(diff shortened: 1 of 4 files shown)


volodya-lombrozo:
@yegor256 @mximp @Graur What do you think?

yegor256:
@volodya-lombrozo I like the idea

Graur (approved):
@volodya-lombrozo Looks good to me
@yegor256 Please, have a look

volodya-lombrozo:
@yegor256 Could we merge that changes, please?

yegor256:
@rultor merge

PR #1920, opened 2023-03-18
versions configures

yegor256:
I reconfigured `EO-*` attributes in all manifests. Now they are all in order.


yegor256:
@rultor merge

Issue #1897, opened 2023-03-14
sonarcube is complaining about our quality of code


yegor256:
We have pretty bad results in SonarCube: https://sonarcloud.io/summary/new_code?id=com.objectionary%3Aeo Let's try to fix at least the major problems.

yegor256:
@mximp pls, take a look at this one

mximp:
@yegor256 Can you please provide settings for `objectionary/eo` SonarCloud, specifically what filters are set up for the sources scope (inclusions/exclusions).
I believe only admins can see that.

I want to set the same setting for the fork to validate the changes, given that checks fo PRs don't seems to work in SonarCloud.

yegor256:
@mximp I didn't configure anything in SonarCube. It's a default configuration. All defects you can see here: https://sonarcloud.io/summary/new_code?id=com.objectionary%3Aeo I suggest we just fix them one by one.

yegor256:
@mximp once we fix all problems in the repo, I suggest we configure our build to fail when SonarCube complains. I configured `xembly` repo this way and it works fine: https://github.com/yegor256/xembly/commit/f23e3cc7116e585e9414511650d009b7be8da313

mximp:
> @mximp I didn't configure anything in SonarCube. It's a default configuration. All defects you can see here: https://sonarcloud.io/summary/new_code?id=com.objectionary%3Aeo I suggest we just fix them one by one.

@yegor256 That makes sense. The only question here - how do we validate that we fixed particular problem. I was thinking about setting up separate SonarCloud project for at lease forked branch - but for that I need the same settings as for objectionary/master.

0pdd:
@yegor256 3 puzzles [#1959](https://github.com/objectionary/eo/issues/1959), [#1960](https://github.com/objectionary/eo/issues/1960), [#1961](https://github.com/objectionary/eo/issues/1961) are still not solved; solved: [#1962](https://github.com/objectionary/eo/issues/1962).

0pdd:
@yegor256 2 puzzles [#1959](https://github.com/objectionary/eo/issues/1959), [#1961](https://github.com/objectionary/eo/issues/1961) are still not solved; solved: [#1960](https://github.com/objectionary/eo/issues/1960), [#1962](https://github.com/objectionary/eo/issues/1962).

mximp:
@yegor256 all tasks have been completed. The issue can be closed.

yegor256:
[...]

PR #1930, opened 2023-03-21
Batch fix sonarqube failures

mximp:
#1897 Fixes of SonarCloud findings.
Add puzzles for the rest.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -285,6 +285,16 @@ public abstract class PhDefault implements Phi, Cloneable {
     }
 
     /**
+     * Clean up resources.
+     * This includes call of {@link ThreadLocal#remove()} method to prevent
+     * memory leaks.
+     */
+    public static void cleanUp() {
+        PhDefault.TERMS.remove();
+        PhDefault.NESTING.remove();
+    }
+
+    /**
      * Add new attribute.
      *
      * This method can only be called from child classes, in their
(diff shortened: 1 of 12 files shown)


mximp:
@yegor256 please reveiw

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -284,6 +284,16 @@ public String locator() {
         return "?";
     }
 
+    /**
+     * Clean up resources.
+     * This includes call of {@link ThreadLocal#remove()} method to prevent
+     * memory leaks.
+     */
+    public static void cleanUp() {
```
yegor256, line 292: @mximp I don't see anyone calling this method. I think, `remove()` is supposed to be called right after we use `TERMS`, not in a separate method. But I'm not sure...

mximp: @yegor256 You are right - the method is not called by anyone. The thing is I cannot define the moment "right after TERM is used". It's static and holds the value which is required throughout the program execution. So I see correct place for `remove` to be called before program exits. But that doesn't make sense.
So in this case I just created `fake` method according to SonarQube expectations.

mximp:
@yegor256 please check my comment https://github.com/objectionary/eo/pull/1930#discussion_r1144887719

mximp:
@yegor256 reminder

yegor256:
@rultor merfe

mximp:
@yegor256 please try again

mximp:
@yegor256 please merge

yegor256:
@rultor merge

Issue #1870, opened 2023-03-05
transpiled.csv is at the wrong place


yegor256:
The file `transpiled.csv` is placed into `target/eo` directory. The right place should be `target/eo-transpiled.csv` (similar to `eo-foreign` and `eo-placed`).

yegor256:
@Graur can you please help

Graur:
@MikhailLipanin Can you help us here?

PR #1937, opened 2023-03-23
moved transpiled.csv, placed.csv and foreign.csv to the root directory and changed theirs names

Graur:
Closes: #1870


yegor256:
@rultor merge

Issue #1890, opened 2023-03-10
Naming of eo objetcs via '_' must be prohibited


levBagryansky:
I declared 2 objects:
```
[] > unusual_name
  QQ.io.stdout > @
    "unusual_name\n"

[] > unusual-name
  QQ.io.stdout > @
    "unusual-name\n"
```
and ran `mvn clean test -Dtest="*unusual_name*"`
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running EOorg.EOeolang.EOunusual_nameTest
unusual-name
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.134 s - in EOorg.EOeolang.EOunusual_nameTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
```
So the program's behavior becomes unexpected. 
@yegor256 @Graur WDYT?

yegor256:
@levBagryansky it's a bug in name mapping from XMIR to Java, should be fixed

levBagryansky:
@Graur there is also `'.'` character. So if replace it to `'.'` then `"qwe-rty"` and `"qwe..rty"` will be the same

Graur:
@levBagryansky Absolutely agree. But I can't find a solution without counterexample

0pdd:
@levBagryansky the puzzle [#2096](https://github.com/objectionary/eo/issues/2096) is still not solved.

PR #1944, opened 2023-03-24
prohibited underscore in object's name

Graur:
Closes: #1890


Graur:
@yegor256 Please, have a look

yegor256:
@Graur this fix break the promise we gave in our documentation: object name is a small English letter followed by anything except space. Maybe it's better to fix `to-java.xsl` instead and make it treat underscores correctly?

Graur:
@yegor256 Sounds good, I will do it in separated pull request

PR #1945, opened 2023-03-25
1877 Rust project

levBagryansky:
Created `RustProject` class in order to generate rust project and then compie shared library via `cargo build` command.
Constructor creates the raw cargo project. `add` method adds new modules to the project. To compile one library from several files rust project structure must look like:
```
.
├── Cargo.lock
├── Cargo.toml
├── src/
│   ├── lib.rs
│   ├── module1.rs
│   ├── module2.rs
...
```
Modules must follow `snake_case`.
I also fixed the "rust" code to a compiled form.


levBagryansky:
@Graur please take a look

Graur (approved):
@levBagryansky Looks good to me
@yegor256 Please, have a look

levBagryansky:
@yegor256 please check

eo-maven-plugin/src/main/java/org/eolang/maven/RustProject.java
```diff
@@ -0,0 +1,155 @@
...
+     * Creates a raw cargo project.
+     * @param target Destination path.
+     * @throws IOException If any issues with I/O
+     */
+    @SuppressWarnings("PMD.ConstructorOnlyInitializesOrCallOtherConstructors")
+    public RustProject(final Path target) throws IOException {
+        this.footprint = new FtDefault(target);
+        this.dest = target;
+        this.dependencies = new HashSet<>();
+        this.footprint.save(
```
yegor256, line 75: @levBagryansky @Graur it's a bad idea in general to make evaluations inside constructors: https://www.yegor256.com/2015/05/07/ctors-must-be-code-free.html

levBagryansky: @yegor256 
Class `RustProject` must preserve its invariant: cargo project must be compilible after any method. A function initializing the object must set this invariant. That is why I guess it is bad idea to create a public method that set this invariant apart from constructor. It  would always follow constructor.
Thus, the constructor must be `private` since it does not set invariant.

Issue #1939, opened 2023-03-23
Flaky tests fails with `IllegalState Recursive update`  


volodya-lombrozo:
Quite often we get the next error during CI pipelines:
```java
Errors: 
[26046](https://github.com/objectionary/eo/actions/runs/4500577297/jobs/7919888856?pr=1938#step:5:26047)
Error:    EOcompares_string_with_negative_infinityTest.testWorks:53->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:255->PhDefault.attr:256->PhDefault.attr:256 » IllegalState Recursive update
[26047](https://github.com/objectionary/eo/actions/runs/4500577297/jobs/7919888856?pr=1938#step:5:26048)
Error:    EOfloat_not_equal_to_negative_infinityTest.testWorks:53->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256 » IllegalState Recursive update
[26048](https://github.com/objectionary/eo/actions/runs/4500577297/jobs/7919888856?pr=1938#step:5:26049)
Error:    EOint_equal_to_nan_and_infinites_is_falseTest.testWorks:189->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:255->PhDefault.attr:255->PhDefault.attr:256->PhDefault.attr:255->PhDefault.attr:256->PhDefault.attr:256 » IllegalState Recursive update
[26049](https://github.com/objectionary/eo/actions/runs/4500577297/jobs/7919888856?pr=1938#step:5:26050)
Error:    EOminus_one_div_zero_is_negative_infinityTest.testWorks:54->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256 » IllegalState Recursive update
[26050](https://github.com/objectionary/eo/actions/runs/4500577297/jobs/7919888856?pr=1938#step:5:26051)
Error:    EOnegative_float_minus_positive_infinityTest.testWorks:53->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256->PhDefault.attr:256 » IllegalState Recursive update
[...]

PR #1954, opened 2023-03-30
feat(#1939): Fix problem with recursive update of a map

volodya-lombrozo:
Fix flaky test by replacing `computeIfAbsent` method with a plain old java approach.

We can't modify the map inside of `computeIfAbsent` lambda. You can write more about it [here](https://stackoverflow.com/questions/19278443/how-do-i-use-the-new-computeifabsent-function) and [here](https://stackoverflow.com/questions/28840047/recursive-concurrenthashmap-computeifabsent-call-never-terminates-bug-or-fea) 

Closes: #1939


volodya-lombrozo:
@yegor256 Could you have a look, please?

yegor256:
@rultor merge

PR #1963, opened 2023-03-31
chore: Use jtcop maven plugin instead of test-naming-conventions

volodya-lombrozo:
Use jtcop maven plugin instead of test-naming-conventions


yegor256:
@rultor merge

PR #1968, opened 2023-04-03
Correct regexp for redundant parenthesis

### LINKED ISSUE #1959: RedundantParentheses.java:105-113: Refactor regexp in...
The puzzle `1897-b918ace1` from #1897 has to be resolved: 

https://github.com/objectionary/eo/blob/8e82e617d5de10dc7816327d416acff19269cda0/eo-parser/src/main/java/org/eolang/parser/RedundantParentheses.java#L105-L113

The puzzle was created by Max Petrov on 31-Mar-23. 

mximp:
Closes #1959 

Amend regexp to avoid potential overflow. 
Modifiers are used instead of `or` operator.
Triple-quotes check has been moved to the beginning of the regexp to cover both string types correctly.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/RedundantParentheses.java
+++ b/eo-parser/src/main/java/org/eolang/parser/RedundantParentheses.java
@@ -114,7 +114,7 @@ final class RedundantParentheses implements Predicate<String> {
      */
     private static char[] expressionChars(final String expression) {
         return expression.replaceAll(
-            "\"(.|\\s)*?\"|\"\"\"(.|\\s)*?\"\"\"",
+            "(?ms)\"\"\".*?\"\"\"|\".*?\"",
             "literal"
         ).toCharArray();
     }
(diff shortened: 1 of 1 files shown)


volodya-lombrozo (changes requested):
I understand that it's not directly related to that PR, but what do you think if we also replace that statement with already precompiled regex pattern:
```java
private static final Pattern PATTERN = Pattern.compile( "\"(.|\\s)*?\"|\"\"\"(.|\\s)*?\"\"\"");

private static char[] expressionChars(final String expression) {
        return PATTERN.matcher(expression).replaceAll("literal").toCharArray();
}
```
It will also increase the speed of that part of the program.
___
Also it would be nice to add some tests which emphasise the "overflow" problem, but I'm not sure that it is possible. @mximp What do you think?

volodya-lombrozo:
> However if you have something in mind we can add it.

@mximp No, I don't

volodya-lombrozo (approved):

mximp:
@yegor256 please merge

yegor256:
@rultor merge

Issue #1849, opened 2023-02-27
CompilationStep.java:30-33: We can add method...


0pdd:
The puzzle `1842-6aa1ed78` from #1842 has to be resolved: 

https://github.com/objectionary/eo/blob/51f2ac098be25fc6af654052807ea396201714f4/eo-maven-plugin/src/main/java/org/eolang/maven/CompilationStep.java#L30-L33

The puzzle was created by rultor on 27-Feb-23. 

Estimate: 90 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #1972, opened 2023-04-03
feat(#1849): remove redundant `CompilationStep` class

volodya-lombrozo:
I've tried to add `outputDirectory()` method to the `CompilationStep` class and understood that it only overcomplicates the current solution without any profit. 
In order to use `CompilationStep` we have to provide large refactoring, but I think it doesn't worth it and we don't need it, at least not now.
 
If we will need provide refactoring we will be able to do it in a separate issue. 

Closes: #1849


volodya-lombrozo:

mximp (approved):

volodya-lombrozo:
@yegor256 Reminder

yegor256:
@rultor merge

PR #1985, opened 2023-04-11
fix(#1829): Remove ambiguity in flaky tests.

### LINKED ISSUE #1829: Complex tests with many `eq` operations cause `StackOverflowError` and break CI pipeline
(Ref #1819)
For example this test sometimes leads to `StackOverflowError` on CI pipeline (see [here](https://github.com/objectionary/eo/actions/runs/4221651017/jobs/7329203665#step:5:42563)):
```
[] > float-equal-to-nan-and-infinites-is-false
  assert-that > @
    and.
      (0.0.eq nan).eq FALSE
      (0.0.eq positive-infinity).eq FALSE
      (0.0.eq negative-infinity).eq FALSE
      (42.5.eq nan).eq FALSE
      (42.5.eq positive-infinity).eq FALSE
      (42.5.eq negative-infinity).eq FALSE
    $.equal-to TRUE
```

> any new developer will write test that will use several eq statements because he/she can do it (c) @volodya-lombrozo 

Such tests may appear in the future, so need to find the root of the problem why such tests are executed so long and cause StackOverflow

volodya-lombrozo:
The original problem arises from the issue outlined in this GitHub issue: https://github.com/objectionary/eo-strings/issues/147.

The `eo-strings` library, version `0.2.0`, was compiled along with the `eo-runtime` library, and all classes from `eo-runtime` were included in the final JAR file. Later, when we tried to compile the `eo-runtime` library itself, some ambiguity arose:

- We had `eo-runtime` classes that were compiled from sources.
- We had `eo-runtime` classes that came from the `eo-strings` library.
Apparently, the Maven compiler randomly chose different classes, possibly dependent on an unknown order. This situation resulted in flaky tests and issues like the one described here: https://github.com/objectionary/eo/issues/1829.

The straightforward solution would be to deploy a new version of the `eo-strings` library that excludes the `eo-runtime` library. However, in order to "prevent" this situation from occurring again, I suggest that we only place EO binaries that have corresponding EO sources (see the code for implementation details).

Furthermore, I have added several tests to check whether the "StackOverflow" error occurs or not.

This change closes #1829.

### DIFF
--- a/eo-maven-plugin/pom.xml
+++ b/eo-maven-plugin/pom.xml
@@ -38,7 +38,7 @@ SOFTWARE.
       <groupId>org.eolang</groupId>
       <artifactId>eo-runtime</artifactId>
       <!-- This version is intentionally lower than the current version of the project -->
-      <version>0.29.2</version>
+      <version>0.29.3</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/AssembleMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/AssembleMojo.java
@@ -247,6 +247,14 @@ public final class AssembleMojo extends SafeMojo {
     @Parameter(defaultValue = "${plugin}", readonly = true)
     private PluginDescriptor plugin;
 
+    /**
+     * Place only binaries that have sources.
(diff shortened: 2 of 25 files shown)


volodya-lombrozo:

yegor256:
@rultor merge

Issue #1940, opened 2023-03-24
Parallel TranspileMojo


levBagryansky:
Now `xmir` files in `TranspileMojo` sequentially converted to `.java`. I guess they can be carried out in parallel  since the conversions of different `.xmir` files do not depend on each other. 
@volodya-lombrozo WDYT?

volodya-lombrozo:
@levBagryansky I think you are right. At least it looks like we can transpile `xmir` to `java` in parallel. Let's try.

PR #2023, opened 2023-04-26
feat(#1940): Parallel TranspileMojo execution

volodya-lombrozo:
Make TranspileMojo work in parallel by using several available threads.

Entire build time on my local laptop: before = 02:51 min , after = 2:10 min

Closes: #1940

---

## PR-Codex overview
This PR adds support for parallel transpilation of multiple EO programs, improving performance. It also adds object locking to prevent concurrency issues and uses a threading library to parallelize the transpilation process.

### Detailed summary
- Added support for parallel transpilation of multiple EO programs
- Added object locking to prevent concurrency issues
- Used a threading library to parallelize the transpilation process

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/tojos/TranspiledTojos.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/tojos/TranspiledTojos.java
@@ -29,7 +29,9 @@ import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import org.cactoos.scalar.Sticky;
 import org.cactoos.scalar.Unchecked;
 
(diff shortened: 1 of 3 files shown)


volodya-lombrozo:
@mximp Could you have a look, please?

eo-maven-plugin/src/main/java/org/eolang/maven/tojos/TranspiledTojos.java
```diff
@@ -80,7 +86,9 @@ public void close() throws IOException {
      * @param second Xmir2 file.
      */
     public void add(final Path transpiled, final Path second) {
-        this.all.value().add(String.valueOf(transpiled)).set(Attribute.XMIR2.key(), second);
+        synchronized (this.lock) {
```
mximp, line 89: @volodya-lombrozo why do you need explicit lock object here? Why not just
```
synchonized (this) {
}
```

volodya-lombrozo: @mximp 

Lock operations on `this` may have unforeseen side-effects 
There are several reasons synchronization on `this` or class expressions may be a bad idea:
1. it makes synchronization a part of the external interface of the class, which makes a future change to a different locking mechanism difficult,
2. it becomes hard to track just who is locking on a given object,
3. it makes a denial-of-service attack possible, either on purpose or it can happen easily by accident when subclassing. 

As an alternative, consider synchronizing on a private final lock object, access to which can be completely controlled.

mximp (approved):

volodya-lombrozo:

Issue #1729, opened 2023-02-02
Overload constructors with logic or introduce separate static methods?


volodya-lombrozo:
Recently, we had a [discussion](https://github.com/objectionary/eo/pull/1722#discussion_r1093428013) about the code style to be applied to Java constructors. This discussion keeps recurring in various PRs.
For instance, consider the following code:
```java
ObjectsIndex() {
    this(
        new ScalarOf<Set<String>>(
            () -> new SetOf<>(
                new Mapped<>(
                    ObjectsIndex::convert,
                    new Mapped<>(
                        Text::asString,
                        new Split(
                            new TextOf(
                                new URL("https://home.objectionary.com/objectionary.lst")
                            ),
                            "\n"
                        )
                    )
                )
            )
        )
    );
}

ObjectsIndex(final Scalar<? extends Set<String>> all) {
    this.objects = new Sticky<>(all);
}

```

It looks simple, but the logic in the constructor seems "heavy" to me. So, I decided to introduce a new static method:
```java
ObjectsIndex() {
    this(ObjectsIndex.index());
}

ObjectsIndex(final Scalar<? extends Set<String>> all) {
    this.objects = new Sticky<>(all);
}

private static ScalarOf<Set<String>> index() {
    return new ScalarOf<>(
        () -> new SetOf<>(
            new Mapped<>(
                ObjectsIndex::convert,
                new Mapped<>(
                    Text::asString,
                    new Split(
                        new TextOf(
                            new URL("https://home.objectionary.com/objectionary.lst")
                        ),
                        "\n"
                    )
                )
            )
        )
    );
}
```

The aim - to keep common simple structure of any Java class:
```
//package and imports

//constructors aka initialisation - small and simple 

//all other methods that contains complicated logic
``` 

But as an opposite I get the next comment:
> To me the motivation is quite simple here - we are trying to produce minimum required number of elements. In this particular case to understand what the constructor does I have to drill down into static method which is purely code extraction. Whereas having code declared within constructor I can see the whole logic in one place.
That makes sense too, actually.

The main issue is that we keep spending time on rehashing the discussion and further refactoring. I think we should just agree on one approach and use that issue as a reference to save time in the future.

mximp:
@volodya-lombrozo 
[...]

PR #2027, opened 2023-04-27
2011 rust name by location

### LINKED ISSUE #2011: rust insets names
We need to name native functions according to `rust`'s `code` and `params`. Length of the name should be limited and collisions are unacceptable.

levBagryansky:
closes #2011 
In order to name `inserts` uniquely uses map and saves it to target directory to use the base os name in the next "session".

---

## PR-Codex overview
This PR introduces a new class `Names` that generates unique names for Rust functions. It also updates the `BinarizeParseMojo` and `EOrust` classes to use these new names. 

### Detailed summary
- Add `Names` class to generate unique Rust function names
- Update `BinarizeParseMojo` to use `Names` to generate function names
- Update `EOrust` to use `Names`-generated function names
- Remove `name` method from `BinarizeParseMojo`

> The following files were skipped due to too many changes: `eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java`

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeParseMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeParseMojo.java
@@ -40,7 +40,6 @@ import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
-import java.util.Locale;
 import java.util.stream.Collectors;
 import org.apache.commons.codec.DecoderException;
 import org.apache.commons.codec.binary.Hex;
@@ -194,30 +193,6 @@ public final class BinarizeParseMojo extends SafeMojo {
             );
         }
         return result;
-    }
-
-    /**
-     * Uniquely converts the loc into the name for jni function.
-     * @param loc Location attribute of the rust insert.
-     * @return Name for function.
-     */
-    private static String name(final String loc) {
-        final String prefix = "f";
-        final int word = 4;
-        final int capacity = prefix.length() + loc.length() + word * loc.length();
-        final StringBuilder out = new StringBuilder(
-            capacity
-        );
-        out.append(prefix).append(loc.toLowerCase(Locale.ENGLISH).replaceAll("[^a-z0-9]", "o"));
-        for (final char chr: loc.toCharArray()) {
-            out.append(
-                String.format(
-                    "%04x",
-                    (int) chr
-                )
-            );
-        }
-        return out.toString();
     }
 
 }
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeParseMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeParseMojo.java
@@ -48,6 +48,7 @@ import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
+import org.eolang.maven.rust_project.Names;
 import org.eolang.maven.rust_project.Project;
 import org.eolang.maven.util.Home;
 import org.eolang.parser.ParsingTrain;
@@ -91,6 +92,14 @@ public final class BinarizeParseMojo extends SafeMojo {
     ).back().back();
 
     /**
+     * Cache directory.
+     * @checkstyle MemberNameCheck (7 lines)
+     */
+    @Parameter(property = "eo.cache")
+    @SuppressWarnings("PMD.ImmutableField")
+    private Path cache = Paths.get(System.getProperty("user.home")).resolve(".eo");
+
+    /**
      * Target directory.
      * @checkstyle MemberNameCheck (7 lines)
      */
@@ -107,21 +116,32 @@ public final class BinarizeParseMojo extends SafeMojo {
             row -> row.exists(AssembleMojo.ATTR_XMIR2)
         );
         final Project project = new Project(targetDir.toPath().resolve("Lib"));
+        final Names names = new Names(this.cache);
         for (final Tojo tojo : sources) {
             final Path file = Paths.get(tojo.get(AssembleMojo.ATTR_XMIR2));
             final XML input = new XMLDocument(file);
             final List<XML> nodes = this.addRust(input).nodes("/program/rusts/rust");
             for (final XML node: nodes) {
+                final String code = unhex(node.xpath("@code").get(0));
+                final List<String> dependencies =
+                    node.xpath("./dependencies/dependency/@name")
+                    .stream()
+                    .map(BinarizeParseMojo::unhex)
+                    .collect(Collectors.toList());
+                final String function = names.name(
+                    code,
+                    String.join(", ", dependencies)
+                );
                 final String filename = String.format(
                     "%s%s",
-                    name(node.xpath("@loc").get(0)),
+                    function,
                     ".rs"
                 );
                 final Path target = BinarizeMojo.DIR
                     .resolve(BinarizeParseMojo.CODES)
                     .resolve(filename);
                 new Home(this.targetDir.toPath()).save(
-                    unhex(node.xpath("@code").get(0)),
+                    code,
                     target
                 );
                 Logger.info(
@@ -131,16 +151,14 @@ public final class BinarizeParseMojo extends SafeMojo {
                     input.xpath("/program/@name").get(0)
                 );
                 project.add(
-                    name(node.xpath("@loc").get(0)),
-                    unhex(node.xpath("@code").get(0)),
-                    node.xpath("./dependencies/dependency/@name")
-                        .stream()
-                        .map(BinarizeParseMojo::unhex)
-                        .collect(Collectors.toList())
+                    function,
+                    code,
+                    dependencies
                 );
             }
         }
         project.save();
+        names.save();
     }
 
     /**
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeParseMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeParseMojo.java
@@ -116,8 +116,7 @@ public final class BinarizeParseMojo extends SafeMojo {
                     .map(BinarizeParseMojo::unhex)
                     .collect(Collectors.toList());
                 final String function = names.name(
-                    code,
-                    String.join(", ", dependencies)
+                    node.xpath("@code_loc").get(0)
                 );
                 final String filename = String.format(
                     "%s%s",
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeParseMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeParseMojo.java
@@ -92,14 +92,6 @@ public final class BinarizeParseMojo extends SafeMojo {
     ).back().back();
 
     /**
-     * Cache directory.
-     * @checkstyle MemberNameCheck (7 lines)
-     */
-    @Parameter(property = "eo.cache")
-    @SuppressWarnings("PMD.ImmutableField")
-    private Path cache = Paths.get(System.getProperty("user.home")).resolve(".eo");
-
-    /**
      * Target directory.
      * @checkstyle MemberNameCheck (7 lines)
      */
@@ -116,7 +108,7 @@ public final class BinarizeParseMojo extends SafeMojo {
             row -> row.exists(AssembleMojo.ATTR_XMIR2)
         );
         final Project project = new Project(targetDir.toPath().resolve("Lib"));
-        final Names names = new Names(this.cache);
+        final Names names = new Names(targetDir.toPath());
         for (final Tojo tojo : sources) {
             final Path file = Paths.get(tojo.get(AssembleMojo.ATTR_XMIR2));
             final XML input = new XMLDocument(file);
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeParseMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeParseMojo.java
@@ -109,7 +109,7 @@ public final class BinarizeParseMojo extends SafeMojo {
             final XML input = new XMLDocument(file);
             final List<XML> nodes = this.addRust(input).nodes("/program/rusts/rust");
             for (final XML node: nodes) {
-                final String code = unhex(node.xpath("@code").get(0));
+                final String code = BinarizeParseMojo.unhex(node.xpath("@code").get(0));
                 final List<String> dependencies =
                     node.xpath("./dependencies/dependency/@name")
                     .stream()
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
@@ -37,7 +37,10 @@ import org.eolang.maven.footprint.FtDefault;
 import org.eolang.maven.util.Home;
 
 /**
- * Uniquely converts the insert into the name for jni function.
+ * The class for storing and assigning names to
+ * rust inserts by its location.
+ * It loads names base when is created and needs to
+ * be saved for using in other process.
  * @since 0.30
  */
 public final class Names {
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
@@ -68,9 +68,8 @@ public final class Names {
     /**
      * Ctor.
      * @param target Directory where to serialize names.
-     * @throws IOException If any issues with IO.
      */
-    public Names(final Path target) throws IOException {
+    public Names(final Path target) {
         this.dest = target.resolve("names");
         this.prefix = target.toString().toLowerCase(Locale.ENGLISH).replaceAll("[^a-z0-9]", "x");
         this.all = Names.checked(this.dest);
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
@@ -43,7 +43,7 @@ import org.eolang.maven.util.Home;
 /**
  * The class for storing and assigning names to
  * rust inserts by its location.
- * It loads names base when is created and needs to
+ * It loads names base from target and needs to
  * be saved for using in other process.
  * @since 0.30
  */
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
@@ -39,7 +39,7 @@ import org.eolang.maven.util.Home;
 
 /**
  * Uniquely converts the insert into the name for jni function.
- * @since 0.29
+ * @since 0.30
  */
 public final class Names {
 
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
@@ -161,7 +161,17 @@ public final class Names {
                 )
             )
         )) {
-            return (ConcurrentHashMap<String, String>) map.readObject();
+            final Object result = map.readObject();
+            if (result.getClass() != ConcurrentHashMap.class) {
+                throw new ClassCastException(
+                    String.format(
+                        "Object inside %s has wrong class %s",
+                        src,
+                        result.getClass()
+                    )
+                );
+            }
+            return (ConcurrentHashMap<String, String>) result;
         } catch (final ClassNotFoundException exc) {
             throw new IllegalArgumentException(
                 String.format(
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
@@ -21,11 +21,11 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-
 package org.eolang.maven.rust_project;
 
+import com.sun.tools.javac.util.Pair;
 import java.io.ByteArrayInputStream;
-import java.io.FileOutputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
@@ -34,8 +34,12 @@ import java.nio.file.Path;
 import java.util.Base64;
 import java.util.Locale;
 import java.util.concurrent.ConcurrentHashMap;
-import com.sun.tools.javac.util.Pair;
 import org.eolang.maven.util.Home;
+
+/**
+ * Uniquely converts the loc into the name for jni function.
+ * @since 0.29
+ */
 public final class Names {
 
     /**
@@ -44,37 +48,38 @@ public final class Names {
      */
     private final Path dest;
 
+    /**
+     * All names.
+     */
     private final ConcurrentHashMap<String, String> all;
 
+    /**
+     * Prefix for the name.
+     */
     private final String prefix;
 
     /**
-     * Ctor
+     * Ctor.
      * @param target Directory where to serialize names.
      * @throws IOException If any issues with IO.
      */
     @SuppressWarnings("PMD.ConstructorOnlyInitializesOrCallOtherConstructors")
     public Names(final Path target) throws IOException {
-        this.dest = target;
+        this.dest = target.resolve("names.txt");
         this.prefix = target.toString().toLowerCase(Locale.ENGLISH).replaceAll("[^a-z0-9]", "x");
-        if (this.dest.resolve("names.txt").toFile().exists()) {
-            try {
-                this.all = load(target.resolve("names.txt"));
-            } catch (ClassNotFoundException exc) {
-                throw new IllegalArgumentException(
-                    String.format(
-                        "File %s contains invalid data",
-                        this.dest
-                    ),
-                    exc
-                );
-            }
-        }
-        else {
+        if (this.dest.toFile().exists()) {
+            this.all = this.load(this.dest);
+        } else {
             this.all = new ConcurrentHashMap<>();
         }
     }
 
+    /**
+     * Assign the new name to the function.
+     * @param code Code of the function.
+     * @param dependencies Function dependencies.
+     * @return The name.
+     */
     public String name(final String code, final String dependencies) {
         return this.all.computeIfAbsent(
             new Pair<>(code, dependencies).toString(),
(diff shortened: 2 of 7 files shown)


levBagryansky:

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
```diff
@@ -0,0 +1,88 @@
...
+        final String one = "AaAaAa";
+        final String two = "AaAaBB";
+        MatcherAssert.assertThat(
+            one.hashCode(),
+            Matchers.equalTo(two.hashCode())
+        );
+        MatcherAssert.assertThat(
+            dispatcher.name(one, ""),
+            Matchers.not(
+                dispatcher.name(two, "")
```
volodya-lombrozo, line 57: @levBagryansky `Matchers.equals()` ?

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Test
+    void recoversNames(@TempDir final Path temp) throws IOException {
+        final List<String> codes = IntStream.range(0, 1000)
+            .mapToObj(String::valueOf)
+            .collect(Collectors.toList());
+        final String dependency = "dependency";
+        final Names before = new Names(temp);
+        final Map<String, String> functions = new HashMap<>();
```
volodya-lombrozo, line 69: @levBagryansky You can use `stream().collect(Collectors.toMap())` here

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Ctor.
+     * @param target Directory where to serialize names.
+     * @throws IOException If any issues with IO.
+     */
+    @SuppressWarnings("PMD.ConstructorOnlyInitializesOrCallOtherConstructors")
+    public Names(final Path target) throws IOException {
+        this.dest = target.resolve("names");
+        this.prefix = target.toString().toLowerCase(Locale.ENGLISH).replaceAll("[^a-z0-9]", "x");
+        if (this.dest.toFile().exists()) {
```
volodya-lombrozo, line 71: @levBagryansky `Files.exists(dest)`

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Assign the new name to the function.
+     * @param code Code of the function.
+     * @param dependencies Function dependencies.
+     * @return The name.
+     */
+    public String name(final String code, final String dependencies) {
+        return this.all.computeIfAbsent(
+            new ListOf<>(code, dependencies).toString(),
```
volodya-lombrozo, line 86: @levBagryansky Do you really need `ListOf` here? Maybe we can use simple `String.format`?

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Saves the function to name dispatching table.
+     * @throws IOException If any issues with IO.
+     */
+    public void save() throws IOException {
+        Files.createDirectories(this.dest.getParent());
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        final ObjectOutputStream oos = new ObjectOutputStream(baos);
```
volodya-lombrozo, line 102: @levBagryansky Maybe we can save to a plain `txt` file instead? Something like CSV. In this case we will be able to see the content of the file. And the implementation will be much simpler I guess.

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.Base64;
+import java.util.Locale;
+import java.util.concurrent.ConcurrentHashMap;
+import org.cactoos.list.ListOf;
+import org.eolang.maven.footprint.FtDefault;
+import org.eolang.maven.util.Home;
+
+/**
+ * Uniquely converts the insert into the name for jni function.
+ * @since 0.29
```
volodya-lombrozo, line 42: @levBagryansky We are using `0.30` already? Or I'm mistaken?

volodya-lombrozo (changes requested):

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
levBagryansky: @volodya-lombrozo Indeed

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
volodya-lombrozo: @levBagryansky I am not sure, of course. We have to check it. I seems that `csv` just simpler to implement in the code and `csv` file could be read by a human.

levBagryansky: @volodya-lombrozo I have checked it in pr #2028. Serialization takes `~1200` and csv takes `~1700` on my machine.

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
levBagryansky: @volodya-lombrozo on the contrary, they should be different

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
volodya-lombrozo: @levBagryansky It's not such a big penalty, as for me. Moreover, I  think that we can insert comma between two strings without additional library.
But ok, I don't insist here. If you want leave objects - let's do it. We always will have a chance to refactor the code if it will be needed.

levBagryansky: @volodya-lombrozo we cannot just insert comma between strings because `string`s can contain any symbol. That is why we would to think of right formatting.

levBagryansky:
@volodya-lombrozo I removed calling of native because its name now is not defined in `EOrust`. There is already a puzzle about generating this file in `eo-maven-plugin`.

levBagryansky:
@volodya-lombrozo please review. Now functions are named by location

levBagryansky:
@volodya-lombrozo please check

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,135 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Base64;
+import java.util.Locale;
+import java.util.concurrent.ConcurrentHashMap;
+import org.eolang.maven.footprint.FtDefault;
+import org.eolang.maven.util.Home;
+
+/**
+ * Uniquely converts the insert into the name for jni function.
```
volodya-lombrozo, line 40: @levBagryansky The description does not match the class implementation. If the class only converts "insert" to "name", it should have only one method called convert() or something similar.
By the way, it's not clear what is "insert" here.

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,135 @@
...
+     */
+    private final String prefix;
+
+    /**
+     * Ctor.
+     * @param target Directory where to serialize names.
+     * @throws IOException If any issues with IO.
+     */
+    @SuppressWarnings("PMD.ConstructorOnlyInitializesOrCallOtherConstructors")
+    public Names(final Path target) throws IOException {
```
volodya-lombrozo, line 67: @levBagryansky It's a poor practice to perform any work in a constructor.
https://www.yegor256.com/2015/05/07/ctors-must-be-code-free.html

You can use `Sticky` class or something similar to solve that problem.

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /**
+     * Loads the table.
+     * @param src Path where to install from.
+     * @return The map.
+     * @throws IOException If any issues with IO.
+     */
+    private static ConcurrentHashMap<String, String> load(final Path src) throws IOException {
+        final String content = new FtDefault(src.getParent()).load(
```
volodya-lombrozo, line 116: @levBagryansky Looks like you can inline that variable.

eo-maven-plugin/src/main/resources/org/eolang/maven/add_rust/add_rust.xsl
```diff
@@ -37,9 +37,12 @@ SOFTWARE.
               <xsl:attribute name="code">
                 <xsl:value-of select="text()"/>
               </xsl:attribute>
-              <xsl:attribute name="loc">
+              <xsl:attribute name="insert_loc">
```
volodya-lombrozo, line 40: @levBagryansky Do we need `insert_loc` somewhere? I can't find the usage.

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,135 @@
...
+            src.getFileName().toString(),
+            ""
+        );
+        final ObjectInputStream ois = new ObjectInputStream(
```
volodya-lombrozo, line 120: @levBagryansky Looks like you can inline that variable.

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
```diff
@@ -0,0 +1,85 @@
...
+        final Names dispatcher = new Names(temp);
+        final String one = "AaAaAa";
+        final String two = "AaAaBB";
+        MatcherAssert.assertThat(
+            one.hashCode(),
+            Matchers.equalTo(two.hashCode())
+        );
+        MatcherAssert.assertThat(
+            dispatcher.name(one),
+            Matchers.not(
```
volodya-lombrozo, line 55: @levBagryansky Should we use `Matchers.not(Matchers.equals())` here?

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
```diff
@@ -0,0 +1,85 @@
...
+        final Map<String, String> functions = locations.stream().collect(
+            Collectors.toMap(loc -> loc, loc -> before.name(loc))
+        );
+        MatcherAssert.assertThat(
+            locations.size(),
+            Matchers.equalTo(functions.size())
+        );
+        before.save();
+        final Names after = new Names(temp);
+        final ListIterator<String> iterator = locations.listIterator(locations.size());
```
volodya-lombrozo, line 76: @levBagryansky Could you use simple `for` here?

volodya-lombrozo (changes requested):

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
levBagryansky: @volodya-lombrozo "convert" here is same like in math function is something that converts its argumetns to the value. "Convert" is not such clear without context and that is why cannot be the name for method.

eo-maven-plugin/src/main/resources/org/eolang/maven/add_rust/add_rust.xsl
levBagryansky: @volodya-lombrozo it is not used anymore. But it can be usefull as additional information for user.

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
levBagryansky: @volodya-lombrozo if you talk about inserting its assignment to the place where it is used, It will be done by compiler if it is faster since the variable is final. At the same this code is clearer.

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
levBagryansky: @volodya-lombrozo Indeed, it will be clearer

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
volodya-lombrozo: @levBagryansky I don't like `covert()` name either. There is a problem with description of the class. Could you extend it please?

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
volodya-lombrozo: @levBagryansky Please, take a look that article:
https://www.yegor256.com/2015/09/01/redundant-variables-are-evil.html

volodya-lombrozo (changes requested):

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
levBagryansky: @volodya-lombrozo ok, thanks!. Seems it is clearer for me only.

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
levBagryansky: @volodya-lombrozo according to documentation from `Matchers.java`:
  ```
/**
   * A shortcut to the frequently used <code>not(equalTo(x))</code>.
   * For example:
   * <pre>assertThat(cheese, is(not(smelly)))</pre>
   * instead of:
   * <pre>assertThat(cheese, is(not(equalTo(smelly))))</pre>
   * 
   * @param value
   *     the value that any examined object should <b>not</b> equal
   */
  public static <T> org.hamcrest.Matcher<T> not(T value) {
    return org.hamcrest.core.IsNot.not(value);
  }
```

File: eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeParseMojo.java [hunk 14]
```diff
@@ -103,21 +103,31 @@ public final class BinarizeParseMojo extends SafeMojo {
     @Override
     public void exec() throws IOException {
         final Project project = new Project(this.targetDir.toPath().resolve("Lib"));
+        final Names names = new Names(targetDir.toPath());
         for (final ForeignTojo tojo : this.scopedTojos().withOptimized()) {
             final Path file = tojo.optimized();
             final XML input = new XMLDocument(file);
             final List<XML> nodes = this.addRust(input).nodes("/program/rusts/rust");
             for (final XML node: nodes) {
+                final String code = BinarizeParseMojo.unhex(node.xpath("@code").get(0));
```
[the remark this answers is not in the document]
levBagryansky: @volodya-lombrozo This variable is used in several lines.

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,167 @@
...
+
+    /**
+     * Ctor.
+     * @param target Directory where to serialize names.
+     * @throws IOException If any issues with IO.
+     */
+    public Names(final Path target) throws IOException {
+        this.dest = target.resolve("names");
+        this.prefix = target.toString().toLowerCase(Locale.ENGLISH).replaceAll("[^a-z0-9]", "x");
+        this.all = new IoChecked<>(
```
volodya-lombrozo, line 76: @levBagryansky Please, use prestructor here.
https://www.yegor256.com/2021/08/04/prestructors.html

If you are interested, we already have a discussion related to that topic [here](https://github.com/objectionary/eo/issues/1729)

volodya-lombrozo: Maybe it's better just move all that logic into `load()` method?

eo-maven-plugin/src/test/java/org/eolang/maven/BinarizeParseMojoTest.java
```diff
@@ -56,8 +57,10 @@ void parsesSimpleEoProgram(@TempDir final Path temp) throws Exception {
             .execute(new FakeMaven.BinarizeParse())
             .result();
         final String rust = String.format(
-            "target/binarize/codes/%s.rs",
-            "fooorgoeolangocustomocreatesoobjector03a6002e006f00720067002e0065006f006c0061006e0067002e0063007500730074006f006d002e0063007200650061007400650073002d006f0062006a006500630074002e0072"
+            "target/binarize/codes/%s0.rs",
```
volodya-lombrozo, line 60: @levBagryansky You can use just `%s` here.

eo-maven-plugin/src/test/java/org/eolang/maven/BinarizeParseMojoTest.java
```diff
@@ -86,13 +89,16 @@ void binarizesTwiceRustProgram(@TempDir final Path temp) throws Exception {
         final Map<String, Path> res = maven
             .execute(new FakeMaven.BinarizeParse())
             .result();
+        final String prefix = temp.resolve("target").toString()
+            .toLowerCase(Locale.ENGLISH)
+            .replaceAll("[^a-z0-9]", "x");
         final String one = String.format(
-            "target/binarize/codes/%s.rs",
-            "fooorgoeolangocustomohellooworldo1or03a6002e006f00720067002e0065006f006c0061006e0067002e0063007500730074006f006d002e00680065006c006c006f002d0077006f0072006c0064002d0031002e0072"
+            "target/binarize/codes/%s0.rs",
```
volodya-lombrozo, line 96: @levBagryansky You can use just `%s` here.

eo-maven-plugin/src/test/java/org/eolang/maven/BinarizeParseMojoTest.java
```diff
@@ -86,13 +89,16 @@ void binarizesTwiceRustProgram(@TempDir final Path temp) throws Exception {
...
+            prefix
         );
         final String two = String.format(
-            "target/binarize/codes/%s.rs",
-            "fooorgoeolangocustomohellooworldo2or03a6002e006f00720067002e0065006f006c0061006e0067002e0063007500730074006f006d002e00680065006c006c006f002d0077006f0072006c0064002d0032002e0072"
+            "target/binarize/codes/%s1.rs",
```
volodya-lombrozo, line 100: @levBagryansky You can use just `%s` here.

eo-maven-plugin/src/test/java/org/eolang/maven/BinarizeParseMojoTest.java
```diff
@@ -137,11 +143,14 @@ void createsCorrectRustProject(@TempDir final Path temp) throws Exception {
...
             .result();
+        final String prefix = temp.resolve("target").toString()
+            .toLowerCase(Locale.ENGLISH)
+            .replaceAll("[^a-z0-9]", "x");
         final String cargo = "target/Lib/Cargo.toml";
         final String lib = "target/Lib/src/lib.rs";
         final String module = String.format(
-            "target/Lib/src/%s.rs",
-            "fooorgoeolangocustomocreatesoobjector03a6002e006f00720067002e0065006f006c0061006e0067002e0063007500730074006f006d002e0063007200650061007400650073002d006f0062006a006500630074002e0072"
+            "target/Lib/src/%s1.rs",
```
volodya-lombrozo, line 152: @levBagryansky You can use just `%s` here.

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
volodya-lombrozo: @levBagryansky Thank you! I didn't know.

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
```diff
@@ -0,0 +1,83 @@
...
+        final Map<String, String> functions = locations.stream().collect(
+            Collectors.toMap(loc -> loc, loc -> before.name(loc))
+        );
+        MatcherAssert.assertThat(
+            locations.size(),
+            Matchers.equalTo(functions.size())
+        );
+        before.save();
+        final Names after = new Names(temp);
+        for (int iter = locations.size() - 1; iter >= 0; --iter) {
```
volodya-lombrozo, line 75: @levBagryansky Do we need the reverse order here? It seems like we might not care about the order at all, or maybe I am missing something?

volodya-lombrozo (changes requested):

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
levBagryansky: @volodya-lombrozo  We realy need it in order to check that it names according to existing map and does not create new.

eo-maven-plugin/src/test/java/org/eolang/maven/BinarizeParseMojoTest.java
levBagryansky: @volodya-lombrozo Why is it better? Do you suggest replace `target/Lib/src` by `%s`?

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
levBagryansky: @volodya-lombrozo it will overcomplicate `load()` as for me

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
volodya-lombrozo: @levBagryansky Please, change the code to the next:
```java
        for (final String loc : locations) {
            MatcherAssert.assertThat(
                functions.get(loc),
                Matchers.equalTo(after.name(loc))
            );
        }
```
And you will see that nothing is changed and test will pass.

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,176 @@
...
+     * Prefix for the name.
+     */
+    private final String prefix;
+
+    /**
+     * Ctor.
+     * @param target Directory where to serialize names.
+     * @throws IOException If any issues with IO.
+     */
+    public Names(final Path target) throws IOException {
```
volodya-lombrozo, line 73: @levBagryansky You don't need ` throws IOException` anymore

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,176 @@
...
+     * @throws IOException If any issues with IO.
+     */
+    public void save() throws IOException {
+        Files.createDirectories(this.dest.getParent());
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        final ObjectOutputStream oos = new ObjectOutputStream(baos);
+        oos.writeObject(this.all.value());
+        oos.flush();
+        new Home(this.dest.getParent()).save(
+            new String(Base64.getEncoder().encode(baos.toByteArray())),
```
volodya-lombrozo, line 118: @levBagryansky It's better to specify charset here:
```
 new String(Base64.getEncoder().encode(baos.toByteArray()), StandardCharsets.UTF_8)
```

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * Loads the table.
+     * @param src Path where to install from.
+     * @return The map.
+     * @throws IOException If any issues with IO.
+     */
+    private static ConcurrentHashMap<String, String> load(final Path src) throws IOException {
+        try {
+            return (ConcurrentHashMap<String, String>) (
```
volodya-lombrozo, line 154: @levBagryansky Unchecked cast: `java.lang.Object` to `java.util.concurrent.ConcurrentHashMap<java.lang.String,java.lang.String>`

volodya-lombrozo (changes requested):
@levBagryansky  We are almost finished :) Just a few small changes are left to make

eo-maven-plugin/src/test/java/org/eolang/maven/BinarizeParseMojoTest.java
volodya-lombrozo: @levBagryansky Not relevant anymore.

eo-maven-plugin/src/test/java/org/eolang/maven/BinarizeParseMojoTest.java
volodya-lombrozo: @levBagryansky Not relevant anymore.

eo-maven-plugin/src/test/java/org/eolang/maven/BinarizeParseMojoTest.java
volodya-lombrozo: @levBagryansky Not relevant anymore.

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
levBagryansky: @volodya-lombrozo We need it because get this object from file.

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
levBagryansky: @volodya-lombrozo this method is not overloaded with second argument

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
levBagryansky: @volodya-lombrozo we can check class `ConcurrentHashMap` but it would be redundant to check that every its item insyde has type `String`.

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
volodya-lombrozo: @levBagryansky It's `String` constructor

eo-maven-plugin/src/main/java/org/eolang/maven/rust_project/Names.java
volodya-lombrozo: @levBagryansky 
> This cast is inside try catch block

The `try-catch` block won't catch a `ClassCastException`

> we can check class ConcurrentHashMap but it would be redundant to also check that every its item inside has type String.

Ok, agree. Then, just add `@SuppressWarnings("unchecked")` above the method.

Also, please, note, that it's better to place `ObjectInputStream` to `try-with-resources` block in order to close it properly:
```java
try (ObjectInputStream map = new ObjectInputStream(
            new ByteArrayInputStream(
                Base64.getDecoder().decode(
                    new FtDefault(src.getParent()).load(
                        src.getFileName().toString(),
                        ""
                    )
                )
            )
        )) {
...
```

volodya-lombrozo (changes requested):
@levBagryansky What about that https://github.com/objectionary/eo/pull/2027#discussion_r1185865042 ?

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
levBagryansky: @volodya-lombrozo The test will pass but it will not check recovering anynore

levBagryansky: @volodya-lombrozo This test firstle creates the name base (via `before`) and then check it (`after` object). The goal is to check that `after` is built according to `before` so it must give name according `before` table. Thus, checking in sequancy order will not prove that names are already exist in table and are not created again

levBagryansky: @volodya-lombrozo the essence of the check is in the wrong order

levBagryansky:
@volodya-lombrozo the test is already right

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
volodya-lombrozo: @levBagryansky I believe it's better to find an explicit way to check your assumption:

> prove that names are already exist in table

As you can see, the current solution rises lots of questions and not straightforward as could be. 
For example, you can get random item from `location` and just check it. No need to check all of them.

volodya-lombrozo (changes requested):

eo-maven-plugin/src/test/java/org/eolang/maven/rust_project/NamesTest.java
volodya-lombrozo: @levBagryansky Maybe, but simple code is much better. Why do you think it's better to check every item?

volodya-lombrozo (changes requested):
@levBagryansky In my opinion, implementing clear checks within the test code is usually preferable to adding comments. So, I find that well-written code is often self-explanatory and requires fewer comments. In other words, the code describes itself much better than any comment description.
___
Maybe we can implement `equals` method in `Names` class? Then in the test, we could compare both the `after` and `before` instances using this method. It's worth mentioning that the `equals` method should ensure a "deep" comparison to accurately compare the objects.

levBagryansky:
@volodya-lombrozo according to `checkstyle`:
`[INFO] Checkstyle: src/main/java/org/eolang/maven/rust_project/Names.java[107]: Definition of 'equals()' without corresponding definition of 'hashCode()'. (EqualsHashCodeCheck)`
Will not its implementation be redundant?

levBagryansky:
There are 2 conflicting PMD violations: `ConfusingTernary` and `OnlyOneReturn`. I chose to disable last one.

levBagryansky:
@yegor256 please check

yegor256:
@rultor merge

PR #2030, opened 2023-04-28
#2018: Enable test depending on cargo installed.

### LINKED ISSUE #2018: BinarizeMojoTest.java:53-56: Enable...
The puzzle `1989-9b8b16d2` from #1989 has to be resolved: 

https://github.com/objectionary/eo/blob/e8367145ff89fd514809d3d6aa6f4d5165c5650e/eo-maven-plugin/src/test/java/org/eolang/maven/BinarizeMojoTest.java#L53-L56

The puzzle was created by rultor on 21-Apr-23. 

levBagryansky:
Closes #2018 
Implemented `cargo` installed condition.

---

## PR-Codex overview
This PR adds a JUnit extension to skip tests if cargo is not installed. It also enables a previously disabled test case.

### Detailed summary
- Added `CargoCondition` JUnit extension
- Enabled previously disabled `binarizesWithoutErrors` test case

### DIFF
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/CargoCondition.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/CargoCondition.java
@@ -1,0 +1,58 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2016-2023 Objectionary.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package org.eolang.maven;
+
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.extension.ConditionEvaluationResult;
+import org.junit.jupiter.api.extension.ExecutionCondition;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+/**
+ * JUnit's extension to skip a test if there is no cargo.
+ *
+ * @since 0.30
+ */
+public final class CargoCondition implements ExecutionCondition {
+    @Override
+    public ConditionEvaluationResult evaluateExecutionCondition(
+        final ExtensionContext context) {
+        final ConditionEvaluationResult ret;
+        final boolean exists = Stream.of(
+            System.getenv("PATH").split(Pattern.quote(File.pathSeparator))
+            )
+            .map(Paths::get)
+            .anyMatch(path -> Files.exists(path.resolve("cargo")));
+        if (exists) {
+            ret = ConditionEvaluationResult.enabled("Cargo is installed");
+        } else {
+            ret = ConditionEvaluationResult.disabled("There is not cargo here");
+        }
+        return ret;
+    }
+}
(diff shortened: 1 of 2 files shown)


levBagryansky:

eo-maven-plugin/src/test/java/org/eolang/maven/CargoCondition.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @since 0.30
+ */
+public final class CargoCondition implements ExecutionCondition {
+    @Override
+    public ConditionEvaluationResult evaluateExecutionCondition(
+        final ExtensionContext context) {
+        final ConditionEvaluationResult ret;
+        final boolean exists = Stream.of(
+            System.getenv("PATH").split(Pattern.quote(File.pathSeparator))
```
volodya-lombrozo, line 47: @levBagryansky Will it work for all OS? As I understand they can have different binary-search mechanisms.

So while `PATH` is a common way to locate binaries, it's not universal and other systems may use different mechanisms to find executable files. Moreover, the variable could be `Path` for Windows, but I'm not sure here.

volodya-lombrozo (changes requested):

eo-maven-plugin/src/test/java/org/eolang/maven/CargoCondition.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.jupiter.api.extension.ConditionEvaluationResult;
+import org.junit.jupiter.api.extension.ExecutionCondition;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+/**
+ * JUnit's extension to skip a test if there is no cargo.
+ *
+ * @since 0.30
+ */
+public final class CargoCondition implements ExecutionCondition {
```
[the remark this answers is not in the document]
levBagryansky: @volodya-lombrozo agree, Added description

eo-maven-plugin/src/test/java/org/eolang/maven/CargoCondition.java
levBagryansky: @volodya-lombrozo according to https://www.java.com/en/download/help/path.html `PATH` do the same on linux, windows and macos. Aslo according to `System.env()` documentation:
```
     * <p><a id="EnvironmentVSSystemProperties"><i>System
     * properties</i> and <i>environment variables</i></a> are both
     * conceptually mappings between names and values.  Both
     * mechanisms can be used to pass user-defined information to a
     * Java process.  Environment variables have a more global effect,
     * because they are visible to all descendants of the process
     * which defines them, not just the immediate Java subprocess.
     * They can have subtly different semantics, such as case
     * insensitivity, on different operating systems.  For these
     * reasons, environment variables are more likely to have
     * unintended side effects.  It is best to use system properties
     * where possible.  Environment variables should be used when a
     * global effect is desired, or when an external system interface
     * requires an environment variable (such as {@code PATH}).
     *
     * <p>On UNIX systems the alphabetic case of {@code name} is
     * typically significant, while on Microsoft Windows systems it is
     * typically not.  For example, the expression
     * {@code System.getenv("FOO").equals(System.getenv("foo"))}
     * is likely to be true on Microsoft Windows.
```
`PATH` is correct for windows too.

volodya-lombrozo: @levBagryansky 
```
➜  ~ echo $PATH | grep "mvn"
➜  ~ mvn --version
Apache Maven 3.9.1 (2e178502fcdbffc201671fb2537d0cb4b4cc58f8)
Maven home: /usr/local/Cellar/maven/3.9.1/libexec
Java version: 19.0.1, vendor: Oracle Corporation, runtime: /Users/lombrozo/Library/Java/JavaVirtualMachines/openjdk-19.0.1/Contents/Home
Default locale: en_RU, platform encoding: UTF-8
OS name: "mac os x", version: "12.5", arch: "x86_64", family: "mac"
```

@levBagryansky as you can see `echo $PATH | grep "mvn"` prints nothing, however the OS can find the `mvn` binary.
In other words, my `$PATH` variable doesn't have any information about `mvn` binary. It means that the system can define binary by many different ways, including `$PATH`. I'm afraid we can get the same with `cargo`.

Have you tested your solution?

levBagryansky: @volodya-lombrozo 
```
tardis3@RyanGosling:~/JavaTrash/src/main/java$ echo $PATH | grep "mvn"
tardis3@RyanGosling:~/JavaTrash/src/main/java$ javac Main.java 
tardis3@RyanGosling:~/JavaTrash/src/main/java$ java Main 
Exists: true
tardis3@RyanGosling:~/JavaTrash/src/main/java$ bat Main.java 
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: Main.java
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ import java.io.*;
   2   │ import java.nio.file.Files;
   3   │ import java.nio.file.Path;
   4   │ import java.nio.file.Paths;
   5   │ import java.util.regex.Pattern;
   6   │ import java.util.stream.Stream;
   7   │ 
   8   │ 
   9   │ public class Main {
  10   │     public static void main(String[] args) throws IOException, InterruptedException {
  11   │         boolean existsInPath = Stream.of(System.getenv("PATH").split(Pattern.quote(File.pathSeparator)))
  12   │             .map(Paths::get)
  13   │             .anyMatch(path -> Files.exists(path.resolve("mvn")));
  14   │         System.out.printf("Exists: %b\n", existsInPath);
  15   │     }
  16   │ }
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
tardis3@RyanGosling:~/JavaTrash/src/main/java$ 

```
Seems `echo $PATH` does not helo here

levBagryansky: @volodya-lombrozo `echo $PATH` prints a list of directories with comands inside. So it should contain a directory with `mvn` but not `mvn`. Thus, if you call `which mvn` you can then find the result in `echo $PATH`

levBagryansky:
@yegor256 please check

yegor256:
@rultor merge

Issue #2080, opened 2023-05-22
verify.groovy:40-47: Fix the bug with class intersection...


0pdd:
The puzzle `2072-740fdee7` from #2072 has to be resolved: 

https://github.com/objectionary/eo/blob/71c7319ce711472de944edb755f5e92f873dd42f/eo-maven-plugin/src/it/duplicate_classes/verify.groovy#L40-L47

The puzzle was created by rultor on 22-May-23. 

Estimate: 90 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #2081, opened 2023-05-22
feat(#2080): Remove classes produced by dirty libraries

volodya-lombrozo:
The PR is trying to fix problem produced by dirty libraries like [eo-strings](https://github.com/objectionary/eo-strings/issues/147) example. Some libraries by mistake can put ALL their compiled classes right into the final library jar, instead of only adding atoms. This can cause different runtime errors since the classpath will contain two classes with the same name:
- The first class file will be added from dirty library
- The second class with the same name will be compiled from the transpiled java file
     
 In order to prevent this, we remove all classes that have the java analog in the generated sources. In other words, if `generated-sources` (or `generated-test-sources`) folder has java classes, we expect that they will be only compiled from that folder.

Closes: #2080 

---

## PR-Codex overview
This PR fixes a bug with class intersection in the 'classes' and 'test-classes' folders. It also includes a new method to clean up dirty classes.

### Detailed summary
- Fixed the bug with class intersection in the 'classes' and 'test-classes' folders.
- Added a new method to clean up dirty classes.

### DIFF
--- a/eo-maven-plugin/src/it/duplicate_classes/verify.groovy
+++ b/eo-maven-plugin/src/it/duplicate_classes/verify.groovy
@@ -36,15 +36,5 @@ def disjoint = Files.walk(testClasses).filter(Files::isRegularFile)
   return testClasses.relativize(it).toString()
 }.noneMatch { binaries.contains(it) }
 println "Compiled classes do not have duplicates: " + disjoint
-/**
- * @todo #2072:90min. Fix the bug with class intersection in the 'classes' and
- *  'test-classes' folders. Both folders should contain different unique
- *  classes. In other words, we should avoid duplicating class compilation for
- *  tests. The problem with duplicated sources in tests was fixed:
- *  <a href="https://github.com/objectionary/eo/pull/2076">here<a/>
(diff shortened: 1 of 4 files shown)


volodya-lombrozo:
@Graur Could you have a look, please?

Graur:
@volodya-lombrozo What if the "dirty library" is used internally by a different version of class in `generated-sources`? Then there is a chance that the "dirty library" will not compile at all. Or maybe I misunderstood the idea

volodya-lombrozo:
> @volodya-lombrozo What if the "dirty library" is used internally by a different version of class in `generated-sources`? Then there is a chance that the "dirty library" will not compile at all. Or maybe I misunderstood the idea

@Graur The usual eo library jar has two things:
1. eo sources
2. atoms of the eo library

The `eo-stings` library now contains:
1. eo sources
2. atoms of the library
3. all dependencies inserted into jar (fat jar)
4. all other classes of the library (already compiled eo objects - not atoms)

We already clean libraries from (3). This PR cleans (4).
____
As for versions - I don't clearly understand your question here. We don't have versioning of the binaries yet.

volodya-lombrozo:

Issue #1807, opened 2023-02-16
objects inside abstract objects with the same name can't be build


maxonfjvipon:
I have 2 almost identical unit tests:
```
+alias org.eolang.collections.range
+alias org.eolang.collections.list
+alias org.eolang.hamcrest.assert-that
+home https://github.com/objectionary/eo-collections
+junit
+package org.eolang.collections
+version 0.0.0

[] > simple-range
  range > rng!
    []
      [i] > build
        i > @!
        build (@.plus 1) > next
      build 1 > @
    10
  assert-that > @
    list rng
    $.equal-to
      list
        * 1 2 3 4 5 6 7 8 9

[] > range-with-floats
  range > rng!
    []
      [i] > build
        i > @!
        build (@.plus 0.5) > next
      build 1.0 > @
    5.0
  assert-that > @
    list rng
    $.equal-to
      list
        * 1.0 1.5 2.0 2.5 3.0 3.5 4.0 4.5
```
In both of them I have an object `build` inside abstract objects
When I do `mvn clean test` I expect everything works fine but I get: 

_java.io.IOException: Failed while trying to save to /Users/maxonfjvipon/code/java/eo-collections/target/eo-test/04-pull/simple-range$t0$a0$build range-with-floats$t0$a0$build.eo_

_Failed to execute goal org.eolang:eo-maven-plugin:0.29.0:assemble (test-compile) on project eo-collections: 'org.eolang.maven.AssembleMojo@4a1fb55b' execution failed: java.lang.IllegalStateException: org.apache.maven.plugin.MojoFailureException: 'org.eolang.maven.PullMojo@21227342' execution failed: java.io.IOException: Failed while trying to save to /Users/maxonfjvipon/code/java/eo-collections/target/eo-test/04-pull/simple-range$t0$a0$build range-with-floats$t0$a0$build.eo: Server returned HTTP response code: 400 for URL: https://raw.githubusercontent.com/objectionary/home/0cbb782f2fb06ee966fdf73dd289391d231bc59f/objects/simple-range$t0$a0$build range-with-floats$t0$a0$build.eo_

If I rename object `build` in the second test - everything works fine

maxonfjvipon:
@mximp here it goes: [stacktrace.txt](https://github.com/objectionary/eo/files/11000630/stacktrace.txt)

mximp:
@maxonfjvipon are you building this within `eo-collections` module? Does the module builds successfully without those tests?

[...]

PR #2104, opened 2023-05-29
Add tests for dup attributes

mximp:
Related to #1807.

Adding two XSL tests proving current (broken) logic of cooperative execution of `abstracts-float-up.xsl` and `fix-missed-names.xsl`.

---

## PR-Codex overview
This PR adds a fix to the `eo-parser` module to handle missed names and includes tests to ensure correct behavior. 

### Detailed summary
- Added `fix-missed-names.xsl` to `eo-parser/src/main/resources/org/eolang/parser/optimize/`
- Added tests to `eo-runtime/src/test/eo/org/eolang/runtime-tests.eo`
- Added tests to `eo-parser/src/test/resources/org/eolang/parser/packs/float-up-same-attrs.yaml`
- Added tests to `eo-parser/src/test/resources/org/eolang/parser/packs/optimize/fix-names-with-duplicates.yaml`

### DIFF
--- a/eo-parser/src/test/resources/org/eolang/parser/packs/float-up-same-attrs.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/packs/float-up-same-attrs.yaml
@@ -1,0 +1,31 @@
+xsls:
+  - /org/eolang/parser/optimize/remove-refs.xsl
+  - /org/eolang/parser/optimize/abstracts-float-up.xsl
+  - /org/eolang/parser/optimize/remove-levels.xsl
+  - /org/eolang/parser/add-refs.xsl
+tests:
+  - /program[count(.//o[@base='build' and not(@ref)])=2]
+  - /program/objects[count(o[@original-name='build'])=2]
+eo: |
+  [] > hello
+    [f s] > calc
+      plus. > @
+        f.next
+        s.next
+  
+    seq > @
+      QQ.io.stdout
+        QQ.txt.sprintf
+          "Result is %d\n"
+          calc
+            []
+              [x] > build
+                x.plus 1 > @
+                build (@.plus 1) > next
+              build 1 > @
+            []
+              [y] > build
+                y.plus 2 > @
+                build (@.plus 2) > next
+              build 2 > @
+      TRUE
--- a/eo-parser/src/main/resources/org/eolang/parser/optimize/fix-missed-names.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/optimize/fix-missed-names.xsl
@@ -36,7 +36,7 @@ SOFTWARE.
         <xsl:variable name="p" select="ancestor::*[o[@parent and @original-name=$o/@base]][1]"/>
         <xsl:choose>
           <xsl:when test="$p">
-            <xsl:variable name="x" select="$p/o[@original-name=$o/@base]"/>
+            <xsl:variable name="x" select="$p/o[@original-name=$o/@base][1]"/>
             <xsl:value-of select="$x/@name"/>
           </xsl:when>
           <xsl:otherwise>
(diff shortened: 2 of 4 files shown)


mximp:
/codex What changes are made within existing tests?

mximp:
@volodya-lombrozo @Graur, please check the changes

volodya-lombrozo (changes requested):
@mximp As I understood, you added the tests that shows the wrong behaviour of parsing chain (not unique names), please, correct me if I'm wrong.
Would it be better to write tests that will expect correct behaviour instead of emphasising the wrong behaviour? In other words, I would write the test that clearly shows our expectations from the future changes and disable it by using `skip:true` option.

mximp:
> @mximp As I understood, you added the tests that shows the wrong behaviour of parsing chain (not unique names), please, correct me if I'm wrong. Would it be better to write tests that will expect correct behaviour instead of emphasising the wrong behaviour? In other words, I would write the test that clearly shows our expectations from the future changes and disable it by using `skip:true` option.

@volodya-lombrozo The problem is that we don't know the expected behaviour. These two tests are exactly what happens now, so that the logic is clear. 
Next step would be describe new behaviour and fix the code (along with tests).

volodya-lombrozo (approved):
@mximp Looks good to me

yegor256:
@mximp I also think that it's not a good idea to add such files to the repository, since we may simply forget in the future that they are actually emphasising the problem, instead of confirming the correct behavior. Maybe we can think about the "desired" behavior and express it in the tests?

yegor256:
@mximp indeed, small commits/PRs are great, but each of them must move the system from one consistent state to another consistent state. In other words, we must always assume that after your commit you can leave the project and the knowledge of your intent will be lost. New developers will open the code base, see your changes, and will assume that this is a consistent state. They will not know that you actually wanted to remove these tests in the future, as invalid.

> And if I add tests with desired behaviour I will have to supply supporting functionality or disable it and create puzzle. I see no point in doing this.

Why no point? This is exactly what would be great in this case: disable the tests and create the puzzle.

mximp:
@yegor256 @volodya-lombrozo please check now. I've changed base test to reflect expected behaviour. The second test left as-is as it probably correct.

eo-parser/src/test/resources/org/eolang/parser/packs/float-up-same-attrs.yaml
```diff
@@ -0,0 +1,31 @@
+xsls:
+  - /org/eolang/parser/optimize/remove-refs.xsl
+  - /org/eolang/parser/optimize/abstracts-float-up.xsl
+  - /org/eolang/parser/optimize/remove-levels.xsl
+  - /org/eolang/parser/add-refs.xsl
+tests:
+  - /program[count(.//o[@base='build' and not(@ref)])=2]
+  - /program/objects[count(o[@original-name='build'])=2]
```
volodya-lombrozo, line 8: @mximp You can add expectations about new name that both objects will have

volodya-lombrozo (changes requested):

eo-parser/src/test/resources/org/eolang/parser/packs/float-up-same-attrs.yaml
mximp: @volodya-lombrozo I believe it doesn't matter for this test case: we don't care about the names here.

mximp: @volodya-lombrozo It tests particular fact about the output xml. The fact is "we create two objects with the same base and two objects with same original name". For this test it's enough. There is no point in going deeper (which names objects have or what order they are placed in etc.). Just minimum enough to reveal the issue under consideration.

mximp:
@volodya-lombrozo @yegor256 
I've added the fix itself. Please check.

mximp:
@yegor256 please merge

yegor256:
@rultor merge

Issue #2123, opened 2023-06-03
don't fail on the first parsing error, but accumulate them all


yegor256:
Now, our parser stops and throws an exception on the first parsing error. This is not how most compilers work. Instead, we should continue parsing and accumulate all visible errors. Then, fail.

PR #2124, opened 2023-06-03
don't fail on the first parsing error

yegor256:
per #2123

---

## PR-Codex overview
This PR adds new features and improvements to the EO programming language, including better error reporting and handling, improved syntax parsing, and updates to the EO Maven plugin.

### Detailed summary

- Improved error handling and reporting in the EO parser
- Updated the EO Maven plugin to handle parsing errors and provide better feedback
- Improved syntax parsing for better performance and accuracy

> The following files were skipped due to too many changes: `eo-parser/src/main/java/org/eolang/parser/Syntax.java`, `paper/main.bib`


yegor256:
@volodya-lombrozo check this one pls

eo-maven-plugin/src/main/java/org/eolang/maven/ParseMojo.java
```diff
@@ -149,46 +150,51 @@ private void parse(final ForeignTojo tojo) throws IOException {
...
+                        source, errors.size()
+                    )
                 );
             }
-            Logger.warn(
-                this, "Parsing was skipped due to failOnError=false. In file %s with error: %s",
-                source,
-                ex.getMessage()
-            );
             return;
```
volodya-lombrozo, line 198: @yegor256 Previously, we had to use the `return` statement because we didn't have any other option. However, now you can eliminate the problem of having two exit points by using the `if-else` statement.
```java
if (!errors.isEmpty()) {
...
} else {
...
}
```

eo-parser/src/main/java/org/eolang/parser/Syntax.java
```diff
@@ -110,41 +115,66 @@ public Syntax(
...
         parser.removeErrorListeners();
-        parser.addErrorListener(errors);
+        parser.addErrorListener(spy);
         final XeListener xel = new XeListener(this.name, this.redundancy);
         new ParseTreeWalker().walk(xel, parser.program());
-        final XML dom = new XMLDocument(new Xembler(xel).domQuietly());
+        final XML dom = new XMLDocument(
+            new Xembler(
+                new Directives(xel).append(
+                    new org.cactoos.iterable.Joined<>(
```
volodya-lombrozo, line 163: @yegor256 Do we need to use fully-qualified name of the class?

eo-parser/src/main/java/org/eolang/parser/Syntax.java
```diff
@@ -110,41 +115,66 @@ public Syntax(
...
+     * at <tt>/program/errors</tt> XPath.</p>
+     *
      * @throws IOException If fails
+     * @checkstyle AnonInnerLengthCheck (100 lines)
      */
     public void parse() throws IOException {
         final List<Text> lines = this.lines();
-        final ANTLRErrorListener errors = new BaseErrorListener() {
+        final List<ParsingException> errors = new LinkedList<>();
+        final ANTLRErrorListener spy = new BaseErrorListener() {
```
volodya-lombrozo, line 128: @yegor256 [Suggestion]: Maybe it's better to create a separate `private` class? Something like `SpyErrorListener`:
```java
private class SpyErrorListener extends BaseErrorListener{

@Override
public void syntaxError(...);

public boolean hasErrors();

public XML directives();

}
```
It is only the sketch, of course. By this you:

- Don't need to specify `@checkstyle AnonInnerLengthCheck (100 lines)`
- Will simplify the `parse` method significantly

eo-parser/src/test/java/org/eolang/parser/SyntaxTest.java
```diff
@@ -116,44 +116,6 @@ void parsesSuccessfully(final String code) {
         );
     }
 
-    @Test
-    void failsWithDoubleNewLine() {
```
volodya-lombrozo, line 120: @yegor256 I believe that removing tests is a dangerous approach. It is much safer to keep the tests and modify the assertions. For instance, you can verify that the final output includes errors:
```
/program/errors/error[1]
```

volodya-lombrozo (changes requested):

eo-maven-plugin/src/main/java/org/eolang/maven/ParseMojo.java
yegor256: @volodya-lombrozo thanks, fixed: aebb9c2fa

eo-parser/src/main/java/org/eolang/parser/Syntax.java
yegor256: @volodya-lombrozo yes, because we also import another `Joined` in this file

eo-parser/src/main/java/org/eolang/parser/Syntax.java
yegor256: @volodya-lombrozo you are right, fixed: 58f82c49f

eo-parser/src/test/java/org/eolang/parser/SyntaxTest.java
yegor256: @volodya-lombrozo these tests duplicate what we already test in `TyposTest`, that's why I removed them

yegor256:
@volodya-lombrozo seems to be better now, check again pls

eo-parser/src/main/java/org/eolang/parser/Syntax.java
```diff
@@ -178,4 +180,77 @@ private List<Text> lines() {
         return new ListOf<>(new Split(new TextOf(this.input), "\r?\n"));
     }
 
+    /**
+     * Accumulates all parsing errors.
+     *
+     * @since 0.30.0
+     */
+    private static final class ParsingErrors extends BaseErrorListener
+        implements ANTLRErrorListener, Iterable<Directive> {
```
volodya-lombrozo, line 189: @yegor256 `BaseErrorListener` already implements the `ANTLRErrorListener`. Do we need to specify this twice?

volodya-lombrozo (commented):
@yegor256 Only a single small question remains. Everything else looks good to me. Thanks!

eo-parser/src/main/java/org/eolang/parser/Syntax.java
yegor256: @volodya-lombrozo I believe, it's a good practice to mention both, the interface and the base class, here is why. If we only mention the base class, future developers may get confused which interface in particular we implement, when/if they decide to replace the base class with something different or simply reimplement my `ParsingErrors` from scratch, without the base class.

volodya-lombrozo (approved):
@yegor256 Looks good to me

yegor256:
@rultor merge

Issue #2196, opened 2023-07-03
TranspileMojo: delete binary: AccessDeniedException


volodya-lombrozo:
I've faced with the next problem during the [PR check](https://github.com/objectionary/eo/actions/runs/5442748269/jobs/9898389264?pr=2193):
```java
2023-07-03T09:22:54.4271466Z Caused by: java.nio.file.AccessDeniedException: D:\a\eo\eo\eo-maven-plugin\target\it\duplicate_classes\target\classes\EOorg\EOeolang\EOtxt\EOtext$EOlow_cased$EOt2$EOt0$EOt0$EOa2.class
2023-07-03T09:22:54.4271691Z     at sun.nio.fs.WindowsException.translateToIOException (WindowsException.java:89)
2023-07-03T09:22:54.4271974Z     at sun.nio.fs.WindowsException.rethrowAsIOException (WindowsException.java:103)
2023-07-03T09:22:54.4272194Z     at sun.nio.fs.WindowsException.rethrowAsIOException (WindowsException.java:108)
2023-07-03T09:22:54.4272510Z     at sun.nio.fs.WindowsFileSystemProvider.implDelete (WindowsFileSystemProvider.java:275)
2023-07-03T09:22:54.4272765Z     at sun.nio.fs.AbstractFileSystemProvider.deleteIfExists (AbstractFileSystemProvider.java:110)
2023-07-03T09:22:54.4273020Z     at java.nio.file.Files.deleteIfExists (Files.java:1191)
2023-07-03T09:22:54.4273219Z     at org.eolang.maven.TranspileMojo.cleanUpClasses (TranspileMojo.java:263)
2023-07-03T09:22:54.4273636Z     at org.eolang.maven.TranspileMojo.transpile (TranspileMojo.java:237)
2023-07-03T09:22:54.4273830Z     at org.eolang.maven.TranspileMojo.transpile (TranspileMojo.java:210)
2023-07-03T09:22:54.4274019Z     at org.eolang.maven.TranspileMojo.lambda$exec$0 (TranspileMojo.java:160)
2023-07-03T09:22:54.4274213Z     at org.cactoos.scalar.CallableEnvelope.call (CallableEnvelope.java:53)
2023-07-03T09:22:54.4274373Z     at java.util.concurrent.FutureTask.run (FutureTask.java:264)
2023-07-03T09:22:54.4274590Z     at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1136)
[...]

PR #2198, opened 2023-07-03
fix(#2196): check access before removing binaries

volodya-lombrozo:
Check access before removing binaries.

Closes: #2196

---

## PR-Codex overview
This PR focuses on fixing an issue with file deletion in the TranspileMojo class. The AccessDeniedException is caught and logged as a warning instead of throwing an exception. 

### Detailed summary
- Added import for AccessDeniedException
- Modified cleanUpClasses method to catch AccessDeniedException and log a warning message

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/TranspileMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/TranspileMojo.java
@@ -35,6 +35,7 @@ import com.yegor256.xsline.Train;
 import com.yegor256.xsline.Xsline;
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.AccessDeniedException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


volodya-lombrozo:

yegor256:
@volodya-lombrozo I'm not sure I understand the intent. You are basically swallow an important runtime problem here. What for? Why shouldn't we throw? If you want to check _before_ deleting, than you should do it before `deleteIfExists()`, not when it already failed to delete.

volodya-lombrozo:
@yegor256 Could you take a look one more time please? I've counted your previous comment.

yegor256:
@volodya-lombrozo I feel that I mislead you with the previous comment. I investigated the situation and understood what you are doing here. You just temporarily swallow a problem in order to clean up the CI pipeline. I think, the way you did it initially was right, but the explanation was not clear enough. I think you need a puzzle staying right next to this `catch` clause and saying: "we swallow it now because we don't know how to fix"

volodya-lombrozo:

yegor256:
@rultor merge

Issue #2204, opened 2023-07-04
Daily build fails because installation is failed


volodya-lombrozo:
Daily build fails because installation of the `eo-parser` is failed:
```sh
Error:  Failed to execute goal on project eo-maven-plugin: Could not resolve dependencies for project org.eolang:eo-maven-plugin:maven-plugin:1.0-SNAPSHOT: Failed to collect dependencies at org.eolang:eo-parser:jar:1.0-SNAPSHOT: Failed to read artifact descriptor for org.eolang:eo-parser:jar:1.0-SNAPSHOT: Could not find artifact org.eolang:eo-parent:pom:1.0-SNAPSHOT -> [Help 1]
```
Here is the link to the failed job - https://github.com/objectionary/eo/actions/runs/5450281250/jobs/9915379947

yegor256:
@volodya-lombrozo still a problem?

volodya-lombrozo:
@yegor256 Yes, I'm trying to figure out

PR #2205, opened 2023-07-04
feat(#2204): Daily build fix

volodya-lombrozo:
1. The caching mechanism is added for the daily build.
2. The problem with uninstalled modules was fixed.

Closes: #2204

---

## PR-Codex overview
This PR focuses on enhancing the testing process and caching in the eo-maven-plugin. 

### Detailed summary
- Added a new option `--compilation` to the test script to compile the module and run the tests
- Added caching for the Maven repository in the GitHub workflow
- Added a new JUnit extension to check if the runtime library exists before running the tests


volodya-lombrozo:
Here I've checked the `daily` pipeline: https://github.com/objectionary/eo/actions/runs/5646259491/job/15293759539 
As you can see - it passed successfully

volodya-lombrozo:

eo-maven-plugin/src/test/java/org/eolang/maven/it/SnippetTestCase.java
```diff
@@ -301,18 +303,23 @@ private static String jdkExecutable(final String name) {
...
     private static String classpath() {
-        return String.format(
-            ".%s%s",
-            File.pathSeparatorChar,
-            System.getProperty(
-                "runtime.jar",
-                Paths.get(System.getProperty("user.home")).resolve(
+        final String runtime;
+        final String property = System.getProperty("runtime.jar");
+        if (Objects.isNull(property) || property.isEmpty() || StringUtils.isBlank(property)) {
```
yegor256, line 308: @volodya-lombrozo isn't `StringUtils.isEmpty` not enough? https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isEmpty-java.lang.CharSequence-

yegor256: @volodya-lombrozo I would say, in this case we don't care. We just want to say "if it's not a real path, ..." An empty path/string can't be a real path, right?

volodya-lombrozo: > @volodya-lombrozo I would say, in this case we don't care. We just want to say "if it's not a real path, ..." An empty path/string can't be a real path, right?

@yegor256 yes, you are right.

volodya-lombrozo:
@yegor256 I've added small method which check is path is a real path. However, I used `Exception` for the control flow, but it is the best option I guess: https://stackoverflow.com/questions/468789/is-there-a-way-in-java-to-determine-if-a-path-is-valid-without-attempting-to-cre

Could you have a look, please?

eo-maven-plugin/src/test/java/org/eolang/maven/it/SnippetTestCase.java
```diff
@@ -301,18 +304,42 @@ private static String jdkExecutable(final String name) {
...
+     * @return True if path is real.
+     */
+    private static boolean isNotRealPath(final String path) {
+        boolean res = false;
+        if (Objects.isNull(path) || path.isEmpty() || StringUtils.isBlank(path)) {
+            res = true;
+        } else {
+            try {
+                Paths.get(path);
+            } catch (final InvalidPathException ignore) {
```
yegor256, line 339: @volodya-lombrozo what is the point of swallowing here? If the path is wrong, we should inform the user. The path is provided by the user, so it's his mistakes. Otherwise, if we swallow, the user will be very confused: he provided the path, but we are reading from another path. "Why?" -- he will ask.

volodya-lombrozo:
@yegor256 I've added logging, could you take a look one more time, please?

eo-maven-plugin/src/test/java/org/eolang/maven/it/SnippetTestCase.java
```diff
@@ -301,18 +304,54 @@ private static String jdkExecutable(final String name) {
...
+            runtime
         );
     }
+
+    /**
+     * Checks if the path is real path.
+     * @param path String path to check.
+     * @return True if path is real.
+     */
+    private static boolean isNotRealPath(final String path) {
```
yegor256, line 339: @volodya-lombrozo it's better to use "positive" conditional branches: rename it to `is RealPath`

eo-maven-plugin/src/test/java/org/eolang/maven/it/SnippetTestCase.java
```diff
@@ -301,18 +304,54 @@ private static String jdkExecutable(final String name) {
...
+        boolean res = false;
```
yegor256, line 340: @volodya-lombrozo I would name it `real`. Try to avoid names like `result`

eo-maven-plugin/src/test/java/org/eolang/maven/it/SnippetTestCase.java
```diff
@@ -301,18 +304,54 @@ private static String jdkExecutable(final String name) {
...
+            res = true;
+        } else {
+            try {
+                Paths.get(path);
+            } catch (final InvalidPathException ignore) {
+                Logger.warn(
+                    SnippetTestCase.class,
+                    "The path '%s' is invalid",
+                    path
+                );
```
yegor256, line 351: @volodya-lombrozo I still don't get it. A user provided a wrong path, we ignore this fact, and use the default path. This is weird. It's better to [fail fast](https://www.yegor256.com/2015/08/25/fail-fast.html) and let user know about the problem.

yegor256 (changes requested):
see above

volodya-lombrozo:
@yegor256 Done. Could you have a look one more time, please?

eo-maven-plugin/src/test/java/org/eolang/maven/it/SnippetTestCase.java
```diff
@@ -301,18 +308,111 @@ private static String jdkExecutable(final String name) {
...
-                    String.format(
-                        ".m2/repository/org/eolang/eo-runtime/%s/eo-runtime-%1$s.jar",
-                        "1.0-SNAPSHOT"
-                    )
-                ).toString()
-            )
-        );
+        final String home = System.getProperty("user.home");
+        try {
+            if (SnippetTestCase.isNotRealPath(home)) {
```
yegor256, line 313: @volodya-lombrozo check this out: https://www.yegor256.com/2015/01/21/if-then-throw-else.html

yegor256 (changes requested):

volodya-lombrozo:
@yegor256 I've fixed the problem with `if-then-throw-else` branch. Could you take one more look, please?

volodya-lombrozo:
@yegor256 Reminder

yegor256:
@volodya-lombrozo honestly, I don't understand what's going on here :) Why do you need to do so many checks for errors, catch runtime exceptions, rethrow them, while the situation is simple: if the path is wrong, we fail the test. What is the purpose of such a complicated exception handling here? If it's necessary, explain it in the Javadoc.

volodya-lombrozo:
@yegor256 Could you have a look one more time, please? I've removed all the redundant code

yegor256:
@rultor merge

Issue #1602, opened 2022-12-21
object versioning


yegor256:
Let's make it possible to add versions to objects in EO code:

```
+alias org.eolang.io.stdout|0.28.11

[] > foo
  seq
    org.eolang.io.stdout|0.28.9
      "Git tag name\n"
    org.eolang.io.stdout|e552a02
      "Git short hash\n"
    QQ.io.stdout|e552a02220ded74166f00ef00fedd3d2f2434213
      "Git full hash\n"
```

No symbols are allowed after the version. This is not legal:

```
stdout|0.28.9 "Hello!"
```

If the version is not specified, it is assumed to be equal to the version that is used by the compiler (in the `hash` parameter of our Maven Plugin).

yegor256:
@mximp what's up here?

mximp:
@yegor256 It's planned for development. As soon as we get resolved all issues with new syntax changes.

0pdd:
@yegor256 2 puzzles [#2226](https://github.com/objectionary/eo/issues/2226), [#2227](https://github.com/objectionary/eo/issues/2227) are still not solved.

0pdd:
@yegor256 the puzzle [#2226](https://github.com/objectionary/eo/issues/2226) is still not solved; solved: [#2227](https://github.com/objectionary/eo/issues/2227).

0pdd:
@yegor256 3 puzzles [#2226](https://github.com/objectionary/eo/issues/2226), [#2258](https://github.com/objectionary/eo/issues/2258), [#2259](https://github.com/objectionary/eo/issues/2259) are still not solved; solved: [#2227](https://github.com/objectionary/eo/issues/2227).

0pdd:
@yegor256 4 puzzles [#2226](https://github.com/objectionary/eo/issues/2226), [#2258](https://github.com/objectionary/eo/issues/2258), [#2259](https://github.com/objectionary/eo/issues/2259), [#2266](https://github.com/objectionary/eo/issues/2266) are still not solved; solved: [#2227](https://github.com/objectionary/eo/issues/2227).

0pdd:
[...]

PR #2219, opened 2023-07-07
#1602 External Tojos

maxonfjvipon:
Ref: #1602

Created `ExternalTojos` that will be used to implement object versioning in order to not break main compilation pipeline

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/RegisterMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/RegisterMojo.java
@@ -137,7 +137,15 @@ public final class RegisterMojo extends SafeMojo {
                 .withSource(file.toAbsolutePath())
                 .withVersion(ParseMojo.ZERO)
                 .withVer(ParseMojo.ZERO);
+            final String fullname = String.join("|", name, ParseMojo.ZERO);
+            if (this.external != null) {
+                this.extTojos
+                    .add(fullname)
+                    .withSource(file.toAbsolutePath())
+                    .withVersion(ParseMojo.ZERO);
+            }
             Logger.debug(this, "EO source %s registered", name);
+            Logger.debug(this, "EO source %s registered (external)", fullname);
         }
         Logger.info(
             this, "Registered %d EO sources from %s to %s, included %s, excluded %s",
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/SafeMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/SafeMojo.java
@@ -91,6 +91,17 @@ abstract class SafeMojo extends AbstractMojo {
     protected File foreign;
 
     /**
+     * File with external "tojos".
+     * @checkstyle VisibilityModifierCheck (10 lines)
+     */
+    @Parameter(
+        property = "eo.external",
+        required = true,
+        defaultValue = "${project.build.directory}/eo-external.csv"
+    )
+    protected File external;
+
+    /**
      * Format of "foreign" file ("json" or "csv").
      * @checkstyle MemberNameCheck (7 lines)
      * @checkstyle VisibilityModifierCheck (5 lines)
@@ -196,6 +207,16 @@ abstract class SafeMojo extends AbstractMojo {
     );
 
     /**
+     * External tojos.
+     * @checkstyle MemberNameCheck (7 lines)
+     * @checkstyle VisibilityModifierCheck (5 lines)
+     */
+    protected final ForeignTojos extTojos = new ForeignTojos(
+        () -> Catalogs.INSTANCE.make(this.external.toPath(), this.foreignFormat),
+        () -> this.scope
+    );
+
+    /**
      * Placed tojos.
      * @checkstyle MemberNameCheck (7 lines)
      * @checkstyle VisibilityModifierCheck (5 lines)
@@ -220,6 +241,13 @@ abstract class SafeMojo extends AbstractMojo {
     @SuppressWarnings("PMD.ImmutableField")
     private boolean skip;
 
+    /**
+     * Execute it.
+     * @throws MojoFailureException If fails during build
+     * @throws MojoExecutionException If fails during execution
+     * @checkstyle NoJavadocForOverriddenMethodsCheck (10 lines)
+     * @checkstyle CyclomaticComplexityCheck (70 lines)
+     */
     @Override
     public final void execute() throws MojoFailureException, MojoExecutionException {
         StaticLoggerBinder.getSingleton().setMavenLog(this.getLog());
@@ -265,6 +293,9 @@ abstract class SafeMojo extends AbstractMojo {
             } finally {
                 if (this.foreign != null) {
                     SafeMojo.closeTojos(this.tojos);
+                }
+                if (this.external != null) {
+                    SafeMojo.closeTojos(this.extTojos);
                 }
                 if (this.placed != null) {
                     SafeMojo.closeTojos(this.placedTojos);
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
@@ -429,7 +429,6 @@ public final class FakeMaven {
             .withScope(scope)
             .withVersion(version)
             .withSource(source);
-
         this.current.incrementAndGet();
         return this;
     }
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
@@ -148,6 +148,17 @@ public final class FakeMaven {
     }
 
     /**
+     * Tojo for eo-external.* file.
+     *
+     * @return TjSmart of the current eo-external.file.
+     */
+    public TjSmart external() {
+        return new TjSmart(
+            Catalogs.INSTANCE.make(this.externalPath())
+        );
+    }
+
+    /**
      * Executes Mojo in the workspace.
      *
      * @param mojo Mojo to execute.
@@ -162,8 +173,15 @@ public final class FakeMaven {
                 tojo.set(entry.getKey().key(), entry.getValue());
             }
         }
+        for (final Tojo tojo : this.external().select(all -> true)) {
+            for (final Map.Entry<ForeignTojos.Attribute, Object> entry
+                : this.attributes.entrySet()) {
+                tojo.set(entry.getKey().key(), entry.getValue());
+            }
+        }
         this.params.putIfAbsent("targetDir", this.targetPath().toFile());
         this.params.putIfAbsent("foreign", this.foreignPath().toFile());
+        this.params.putIfAbsent("external", this.externalPath().toFile());
         this.params.putIfAbsent("foreignFormat", "csv");
         this.params.putIfAbsent("project", new MavenProjectStub());
         final Path transpiled = Paths.get("transpiled");
@@ -240,6 +258,17 @@ public final class FakeMaven {
     }
 
     /**
+     * External tojos for eo-external.* file.
+     * @return External tojos.
+     */
+    ForeignTojos externalTojos() {
+        return new ForeignTojos(
+            () -> Catalogs.INSTANCE.make(this.foreignPath()),
+            this::scope
+        );
+    }
+
+    /**
      * Sets placed tojo attribute.
      *
      * @param binary Binary as class file or jar.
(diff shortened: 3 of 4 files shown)


maxonfjvipon:
@yegor256 please have a look

yegor256:
@volodya-lombrozo please, check

eo-maven-plugin/src/main/java/org/eolang/maven/SafeMojo.java
```diff
@@ -195,6 +206,16 @@ abstract class SafeMojo extends AbstractMojo {
         () -> this.scope
     );
 
+    /**
+     * External tojos.
+     * @checkstyle MemberNameCheck (7 lines)
+     * @checkstyle VisibilityModifierCheck (5 lines)
+     */
+    protected final ForeignTojos extTojos = new ForeignTojos(
```
volodya-lombrozo, line 214: @maxonfjvipon Can we use the full name `extrenalTojos` here? `extTojos` could have different meaning/

eo-maven-plugin/src/main/java/org/eolang/maven/SafeMojo.java
```diff
@@ -220,6 +241,13 @@ abstract class SafeMojo extends AbstractMojo {
     @SuppressWarnings("PMD.ImmutableField")
     private boolean skip;
 
+    /**
```
volodya-lombrozo, line 244: @maxonfjvipon Do we need a JavaDoc for the method which we `@Override`?

eo-maven-plugin/src/test/java/org/eolang/maven/RegisterMojoTest.java
```diff
@@ -91,4 +91,21 @@ void doesNotFailWhenNoStrictNames(@TempDir final Path temp) throws IOException {
...
+    void registersInExternalWithZeroVersions(@TempDir final Path temp) throws IOException {
+        new Home(temp).save(
+            new ResourceOf("org/eolang/maven/file-name/abc-def.eo"),
+            Paths.get("src/eo/org/eolang/maven/foo.eo")
+        );
+        final FakeMaven maven = new FakeMaven(temp)
+            .with("sourcesDir", temp.resolve("src/eo").toFile())
+            .execute(new FakeMaven.Register());
+        MatcherAssert.assertThat(
+            maven.external()
```
volodya-lombrozo, line 105: @maxonfjvipon Could you also show that you add the same object to `foreign` tojos?

volodya-lombrozo (changes requested):
@maxonfjvipon To be honest, I don't clearly understand why you duplicate `foreign` tojos by adding `external` objects. If you have future "plans" for that, it would be great to reflect it through puzzles.

eo-maven-plugin/src/main/java/org/eolang/maven/SafeMojo.java
maxonfjvipon: @volodya-lombrozo yes, we can

eo-maven-plugin/src/main/java/org/eolang/maven/SafeMojo.java
maxonfjvipon: @volodya-lombrozo the reason of that is necessity of using `@checkstyle CyclomaticComplexityCheck (70 lines)`.

eo-maven-plugin/src/test/java/org/eolang/maven/RegisterMojoTest.java
maxonfjvipon: @volodya-lombrozo sure

maxonfjvipon:
@volodya-lombrozo the main reason of adding the `ExternalTojos` - don't break main build and compilation pipeline while I'm trying to implement object versioning. I'll use this separated independent tojos for that

volodya-lombrozo:
> @volodya-lombrozo the main reason of adding the `ExternalTojos` - don't break main build and compilation pipeline while I'm trying to implement object versioning. I'll use this separated independent tojos for that

Why can't we do it with `foreign` tojos only?

maxonfjvipon:
@volodya-lombrozo because there will be critical changes. I'm afraid I wont be able to handle them using only `foreign`. My idea is to use separated independent tojos, test them and at the end when versioning will work - just replace them.

volodya-lombrozo:
@maxonfjvipon Then, I believe we have to add a puzzle with approximately that description:
> @volodya-lombrozo because there will be critical changes. I'm afraid I wont be able to handle them using only foreign. My idea is to use separated independent tojos, test them and at the end when versioning will work - just replace them and remove one.

Otherwise the code which you are going to add will look incomplete, and other developer could just remove it and he/she will be right.

maxonfjvipon:

eo-maven-plugin/src/main/java/org/eolang/maven/RegisterMojo.java
```diff
@@ -137,7 +137,15 @@ public void exec() throws IOException {
...
                 .withVer(ParseMojo.ZERO);
+            final String fullname = String.join("|", name, ParseMojo.ZERO);
+            if (this.external != null) {
+                this.externalTojos
+                    .add(fullname)
+                    .withSource(file.toAbsolutePath())
+                    .withVersion(ParseMojo.ZERO);
+            }
             Logger.debug(this, "EO source %s registered", name);
+            Logger.debug(this, "EO source %s registered (external)", fullname);
```
volodya-lombrozo, line 148: @maxonfjvipon It seems that statement should be placed under `if(this.external != null)` scope.

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
```diff
@@ -162,8 +173,15 @@ public <T extends AbstractMojo> FakeMaven execute(final Class<T> mojo) throws IO
                 tojo.set(entry.getKey().key(), entry.getValue());
             }
         }
+        for (final Tojo tojo : this.external().select(all -> true)) {
```
volodya-lombrozo, line 176: @maxonfjvipon Looks like a code duplication. Can we create a common method for the both tojos, or just add a puzzle?

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
```diff
@@ -293,13 +322,21 @@ FakeMaven withTojoAttribute(final ForeignTojos.Attribute attribute, final Object
...
      */
     Path foreignPath() {
         return this.workspace.absolute(Paths.get("eo-foreign.csv"));
     }
 
+    /**
+     * Path to or eo-external.* file after all changes.
+     * @return Path to eo-foreign.* file.
+     */
+    Path externalPath() {
```
volodya-lombrozo, line 336: @maxonfjvipon Looks like you are using that method only inside `FakeMaven` class. So, I believe we might put the `private` modifier here.

eo-maven-plugin/src/test/java/org/eolang/maven/RegisterMojoTest.java
```diff
@@ -91,4 +91,25 @@ void doesNotFailWhenNoStrictNames(@TempDir final Path temp) throws IOException {
...
```
volodya-lombrozo, line 105: @maxonfjvipon Please, add messages for the both assertions, like explained here:
https://www.yegor256.com/2023/01/19/layout-of-tests.html#assertions

volodya-lombrozo (changes requested):

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
maxonfjvipon: @volodya-lombrozo it will be used outside of the `FakeMaven` in the future PRs (just like `foreignPath`) so I think there's no need to make it private now to remake later

maxonfjvipon:

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
```diff
@@ -430,6 +466,19 @@ private static Set<String> mojoFields(final Class<?> mojo, final Set<String> fie
         return res;
     }
 
+    /**
+     * Fill up given tojos by attributes.
+     * @param tojos Tojos to fill up.
+     */
+    private void fillUpTojos(final List<Tojo> tojos) {
```
volodya-lombrozo, line 473: @maxonfjvipon Please, rename that method according to that article: https://www.yegor256.com/2015/01/12/compound-name-is-code-smell.html

Moreover, I would suggest to move that method into `ForeignTojos`:
https://www.yegor256.com/2016/04/05/printers-instead-of-getters.html
Let's treat `ForeignTojos` as an object (not as a collection of data) and give it a chance to fulfill itself. I'm sure it knows better how to do so :)

eo-maven-plugin/src/test/java/org/eolang/maven/RegisterMojoTest.java
```diff
@@ -91,4 +91,37 @@ void doesNotFailWhenNoStrictNames(@TempDir final Path temp) throws IOException {
...
+            Paths.get("src/eo/org/eolang/maven/foo.eo")
+        );
+        final String name = "org.eolang.maven.foo";
+        final String source = "src/eo";
+        final FakeMaven maven = new FakeMaven(temp)
+            .with("sourcesDir", temp.resolve(source).toFile())
+            .execute(new FakeMaven.Register());
+        MatcherAssert.assertThat(
+            String.format(
+                String.join(
```
volodya-lombrozo, line 108: @maxonfjvipon You don't need `String.join` here. Please can you use a single `String.format` here?

eo-maven-plugin/src/test/java/org/eolang/maven/RegisterMojoTest.java
```diff
@@ -91,4 +91,37 @@ void doesNotFailWhenNoStrictNames(@TempDir final Path temp) throws IOException {
...
+                name,
+                source
+            ),
+            maven.external()
+                .getById(name)
+                .exists("id"),
+            Matchers.is(true)
+        );
+        MatcherAssert.assertThat(
+            "External and foreign tojos should have the same status after registering",
```
volodya-lombrozo, line 122: @maxonfjvipon Could you please, add more info here? "Why should they have the same status"?

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
volodya-lombrozo: @maxonfjvipon I believe each new PR has to leave the entire codebase in a final and consistent state. So, we must have either a puzzle or a complete code. Otherwise, we risk losing your knowledge about "future" plans.

volodya-lombrozo (changes requested):

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
maxonfjvipon: @volodya-lombrozo what would you suggest to name this method?

volodya-lombrozo: @maxonfjvipon something like `withAttributes`

maxonfjvipon: @volodya-lombrozo I don't see a fast way to move the method into `ForeignTojos` because `ForeignTojo` does not have a method `set`. So I'll just rename it and add the puzzle.

maxonfjvipon:

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
```diff
@@ -147,6 +147,17 @@ public TjSmart foreign() {
         );
     }
 
+    /**
+     * Tojo for eo-external.* file.
+     *
+     * @return TjSmart of the current eo-external.file.
+     */
+    public TjSmart external() {
```
volodya-lombrozo, line 155: @maxonfjvipon I'm not sure here, but can we left `package-private` modifier for that method?

volodya-lombrozo (changes requested):

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
maxonfjvipon: @volodya-lombrozo why do you think so? There's a place (and will be more) outside the class where the method is called. The same story as with `foreign()` method

volodya-lombrozo: @maxonfjvipon I've found the only place where we call this method outside of the class:
https://github.com/objectionary/eo/pull/2219/files#diff-7e783d4f9fc150a8ce469c472b3d3e567c440f288fcf0fd9f6bd531ddfefcf6aR171

And this place is inside of the same package, so we can leave just `TjSmart external()` without `public` modifier

maxonfjvipon:

volodya-lombrozo (approved):

yegor256:
@rultor merge

PR #2228, opened 2023-07-10
#1602 Discover with versions

maxonfjvipon:
Ref: #1602 

What's done: 
- moved `eo.versioned` to `SafeMojo` to use it in every inherited Mojo
- refactored `RegisterMojo`
- Added ability to discover objects with versions
- Refactored old and created new tests

---

## PR-Codex overview
This PR focuses on implementing object versioning in the EO Maven Plugin. 

### Detailed summary:
- Added a new flag `withVersions` to enable/disable object versioning.
- Modified `RegisterMojo` to register external tojos with versions.
- Modified `DiscoverMojo` to discover objects with versions.
- Added tests to verify object discovery and registration with versions.

> The following files were skipped due to too many changes: `eo-maven-plugin/src/test/java/org/eolang/maven/RegisterMojoTest.java`


maxonfjvipon:

eo-maven-plugin/src/main/java/org/eolang/maven/SafeMojo.java
```diff
@@ -197,11 +197,20 @@ abstract class SafeMojo extends AbstractMojo {
     @Parameter(property = "eo.cache")
     protected Path cache = Paths.get(System.getProperty("user.home")).resolve(".eo");
 
+    /**
+     * Used for object versioning implementation.
+     * If set to TRUE, external tojos are used instead of foreign ones and all
+     * inherited Mojos behave a bit differently.
+     * @checkstyle VisibilityModifierCheck (10 lines)
```
volodya-lombrozo, line 204: @maxonfjvipon As I understand, in the future you are going to remove that boolean field, isn't it? If it is so, please, add a puzzle for it.

eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java
```diff
@@ -131,4 +139,24 @@ private Collection<String> discover(final Path file)
         return names;
     }
 
+    /**
+     * Get list of object names from given XML by provided xpath.
+     * @param xml XML.
+     * @param xpath Xpath.
+     * @return List of object names.
+     * @checkstyle NonStaticMethodCheck (25 lines)
```
volodya-lombrozo, line 147: @maxonfjvipon You can just make this method `static`.

eo-maven-plugin/src/test/java/org/eolang/maven/RegisterMojoTest.java
```diff
@@ -102,6 +102,7 @@ void registersInExternal(@TempDir final Path temp) throws IOException {
         final String source = "src/eo";
         final FakeMaven maven = new FakeMaven(temp)
             .with("sourcesDir", temp.resolve(source).toFile())
+            .with("versioned", true)
```
volodya-lombrozo, line 105: @maxonfjvipon I believe, we have to add one more test for `versioned=false`

eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java
```diff
@@ -94,26 +95,33 @@ public void exec() throws IOException {
...
-                            " and not(@ref)",
-                            "]/@base"
-                        )
-                    )
-                )
-            )
+            this.names(xml, regular)
         );
+        if (this.versioned) {
+            final List<String> versioned = new ListOf<>(base);
```
volodya-lombrozo, line 116: @maxonfjvipon `this.versioned` and just `versioned` are rather confusing names (you have the same name for both variables). Maybe it's better to use different name for one of them?

volodya-lombrozo (changes requested):

maxonfjvipon:

maxonfjvipon:

eo-maven-plugin/src/test/java/org/eolang/maven/RegisterMojoTest.java
```diff
@@ -99,25 +109,44 @@ void registersInExternal(@TempDir final Path temp) throws IOException {
...
+    void doesNotRegisterInExternal(@TempDir final Path temp) throws IOException {
+        new Home(temp).save(
+            new ResourceOf("org/eolang/maven/file-name/abc-def.eo"),
+            Paths.get("src/eo/org/eolang/maven/foo.eo")
+        );
+        final FakeMaven maven = new FakeMaven(temp)
+            .with(RegisterMojoTest.PARAM, temp.resolve(RegisterMojoTest.SOURCES).toFile())
+            .with("withVersions", false);
+        Assertions.assertThrows(
+            NoSuchElementException.class,
```
volodya-lombrozo, line 148: @maxonfjvipon Please, add an explanation message for that assertion

eo-maven-plugin/src/test/java/org/eolang/maven/RegisterMojoTest.java
```diff
@@ -99,25 +109,44 @@ void registersInExternal(@TempDir final Path temp) throws IOException {
...
+        );
+    }
+
+    @Test
+    void doesNotRegisterInExternal(@TempDir final Path temp) throws IOException {
+        new Home(temp).save(
+            new ResourceOf("org/eolang/maven/file-name/abc-def.eo"),
+            Paths.get("src/eo/org/eolang/maven/foo.eo")
+        );
+        final FakeMaven maven = new FakeMaven(temp)
```
volodya-lombrozo, line 144: @maxonfjvipon You can inline that variable

eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java
```diff
@@ -94,26 +95,33 @@ public void exec() throws IOException {
...
+            "and @base != '$'",
+            "and @base != '&'",
+            "and not(@ref)"
+        );
+        final List<String> regular = new ListOf<>(base);
+        if (this.withVersions) {
+            regular.add("and not(@ver)");
+        }
+        regular.add("]/@base");
         final Collection<String> names = new TreeSet<>(
```
volodya-lombrozo, line 112: @maxonfjvipon Do we need to use `TreeSet` here? Looks like we can just remove it:
```
final Collection<String> names = DiscoverMojo.names(xml, regular)
```

eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java
```diff
@@ -94,26 +95,33 @@ public void exec() throws IOException {
...
+            "//o[",
+            "not(starts-with(@base,'.'))",
+            "and @base != 'Q'",
+            "and @base != '^'",
+            "and @base != '$'",
+            "and @base != '&'",
+            "and not(@ref)"
+        );
+        final List<String> regular = new ListOf<>(base);
+        if (this.withVersions) {
```
volodya-lombrozo, line 108: @maxonfjvipon I don't clearly understand why we need that check, to be honest. As I understand, you are looking the names for `regular` and `versioned` objects separately. Do we have `ver` attribute if `this.withVersions==false`? Can we just get `regular` objects (for all cases) as follows:
```java
 final List<String> regular = new ListOf<>(
            "//o[",
            "not(starts-with(@base,'.'))",
            "and @base != 'Q'",
            "and @base != '^'",
            "and @base != '$'",
            "and @base != '&'",
            "and not(@ref)",
            "and not(@ver)"
            "]/@base"
        );
```

Moreover, I would suggest to move the logic with creation of `regular` and `versioned` names into separate methods. For me (personally) it looks more logical.

volodya-lombrozo (changes requested):

eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java
maxonfjvipon: @volodya-lombrozo I believe we can't because we use `/@base` in xpath which is not unique. Correct me if I'm wrong

eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java
maxonfjvipon: @volodya-lombrozo 
1. Regular. We select `/@base` in both cases (either `withVersions` is `true` or `false`) because at this step (discover) there are can be objects without versions (even if `withVersions` is `true`). Here in `regular` if `withVersions` is `true` we add `and not(@ver)` to exclude objects with `ver` attribute because they will be selected at the step 2.
2. Versioned. If `withVersions` is `true` - we additionally select `/concat(@base,'|',@ver)` for objects with `ver` attribute. I didn't find a way to select conditionally in xpath.

I hope you'll understand what I mean)

eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java
volodya-lombrozo: @maxonfjvipon Maybe we can return `Set` directly from `names` method then?
```java
   private static List<String> names(final XML xml, final List<String> xpath) {
         return new SetOf<>(
             new Filtered<>(
                 obj -> !obj.isEmpty(),
                 xml.xpath(
                     String.join(
                         " ",
                         xpath
                     )
                 )
             )
         );
     }
```
Looks like we don't use `List` features, so what is the point to wrap collections to each other?

maxonfjvipon: @volodya-lombrozo `TreeSet` is used to save order of strings being added, I don't really know if it's important but `TreeSet` was used in previous implementation so I don't touch it. `LisfOf` is used to decorate `Filtered` which can't be added to `TreeSet` directly.

volodya-lombrozo: @maxonfjvipon Can we try to use only `SetOf` without `TreeSet`?  I don't think we need order here. If we need - tests will show it.

maxonfjvipon:

eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java
volodya-lombrozo: Let's clarify, then :) 

Imagine, we have the next objects:

```xml
<o ver="1" base="foo"/>
<o base="bar"/>
<o ver="2" base="foobar">
```

And we have two options:
1) `withVersions = true`:

```java
 final List<String> regular = new ListOf<>(
            "//o[",
            "not(starts-with(@base,'.'))",
            "and @base != 'Q'",
            "and @base != '^'",
            "and @base != '$'",
            "and @base != '&'",
            "and not(@ref)",
            "and not(@ver)"
            "]/@base"
        );
```

Res: 
```xml
"bar"
```

```java
 final List<String> versioned = new ListOf<>(
            "//o[",
            "not(starts-with(@base,'.'))",
            "and @base != 'Q'",
            "and @base != '^'",
            "and @base != '$'",
            "and @base != '&'",
            "and not(@ref)",
            "and @ver",
            "]/concat(@base,'|',@ver)"
 );
```

Res:

```xml
"foo|1", "foobar|2"
```

**names="bar, foo|1, foobar|2"**

2) `withVersions=false`:
```java
 final List<String> regular = new ListOf<>(
            "//o[",
            "not(starts-with(@base,'.'))",
            "and @base != 'Q'",
            "and @base != '^'",
            "and @base != '$'",
            "and @base != '&'",
            "and not(@ref)",
            "]/@base"
);
```

Res:
```xml
"foo"
"bar"
"foobar"
```

```java
 final List<String> versioned = NA
```

Res: `empty`

**names = "foo, bar, foobar"**

___

So, suggest:

1. `withVersions = true`:

```java
 final List<String> all = new ListOf<>(
            "//o[",
            "not(starts-with(@base,'.'))",
            "and @base != 'Q'",
            "and @base != '^'",
            "and @base != '$'",
            "and @base != '&'",
            "and not(@ref)"
            "]/string-join((@base, @ver), '|')"
        );
```

**names="bar, foo|1, foobar|2"**

2. `withVersions = false`:

```java
 final List<String> all = new ListOf<>(
            "//o[",
            "not(starts-with(@base,'.'))",
            "and @base != 'Q'",
            "and @base != '^'",
            "and @base != '$'",
            "and @base != '&'",
            "and not(@ref)"
            "]/@base"
        );
```
**names = "foo, bar, foobar"**

And you don’t need `regular` and `versioned` lists.

volodya-lombrozo: @maxonfjvipon

maxonfjvipon: @volodya-lombrozo if `/string-join((@base,@ver),'|')` returns the same as `/@base` if `@ver` does not exist then it can be one unique xpath for both cases, but I'm not sure

maxonfjvipon: @volodya-lombrozo I was wrong, forget about it

maxonfjvipon: @volodya-lombrozo so did some research, you suggestion won't work.
```java
 final List<String> all = new ListOf<>(
    "//o[",
    "not(starts-with(@base,'.'))",
    "and @base != 'Q'",
    "and @base != '^'",
    "and @base != '$'",
    "and @base != '&'",
    "and not(@ref)"
    "]/string-join((@base, @ver), '|')"
);
```
If there are some object where `@ver` is not present - names won't be found, so we need to check if `@ver` is present or not before getting the result

volodya-lombrozo: @maxonfjvipon Ok, let's discuss it in a separate puzzle then.

volodya-lombrozo (approved):

maxonfjvipon:
@yegor256 could you merge that changes please?

yegor256:
@rultor merge

PR #2279, opened 2023-07-20
#1602 `CommitHashesMap` with full hashes (unreasonable)

maxonfjvipon:
Ref: #2282

---

## PR-Codex overview
This PR focuses on updating the comments in the CommitHashesMap class. 

### Detailed summary
- Updated the comment to clarify that the keys in the map are tags.
- Updated the comment to clarify that the values in the map are narrow hashes.
- Removed a todo comment related to using full hashes instead of narrow hashes in certain scenarios.


maxonfjvipon:

volodya-lombrozo (commented):
@maxonfjvipon To be honest I don't see the reason having long hashes. I believe we just have to remove that puzzle without removing the `ChNarrow`.

maxonfjvipon:
@volodya-lombrozo I'm trying now to make `ProbeMojo` worked with versions and there're some problems with such narrow hashes. For example `OyRemote` work only with full hash, it will not work with narrow one. In the future users also may write hashes in EO code as well as tags (see the main [ticket](https://github.com/objectionary/eo/issues/1602)):
```
stdout|0.28.6
stdout|e552a02
stdout|e552a02220ded74166f00ef00fedd3d2f2434213  
```
Here narrow hashes wont be replaced in `VersionsMojo` and they will go further to the `ProbeMojo` and `PullMojo`.
We anyway will need the class that can give us full hash from a narrow one (I'll about to implement it in the near future).
If it's not enough for you - you just need to believe in my best intensions :)

volodya-lombrozo:
@maxonfjvipon The main concern here is that the future implementation and all the problems with that implementation are known only by you. You actually hide the important information and try to explain it through comments like this one:  https://github.com/objectionary/eo/pull/2279#issuecomment-1643665198

I believe that you have good intentions, no doubt. However, lets imagine if you will leave (by any reason) someday, we would lose all the valuable knowledge that only you possess. For instance, we might end up with "long hashes," but nobody knows why we need them. Therefore, we will decide to remove that redundant (from our point of view) code.
____
What to do?

Please don’t try to convince me by comments. Just prove. Show that you need that change. How? You can write a test that emphasizes the considered problem and shows that we actually **need** that changes.

Lets again take a look on the current PR:
- We have short hashes - all the tests pass successfully
- You change short hashes to long hashes - tests are the same and all of them pass successfully. 

What the point in that change then if everything works fine with short hashes? So, I believe we don't need that changes.

maxonfjvipon:
@volodya-lombrozo got your idea. We will anyway need that changes, but it's not a good time for them because I can't prove that we need them now.
I suggest to move this PR to draft and reopen it when it's time.

maxonfjvipon:
@volodya-lombrozo please review this one more time

maxonfjvipon:
@Graur please have a look

Graur (approved):

maxonfjvipon:
@yegor256 please merge this one

maxonfjvipon:
@yegor256 please merge this one

Issue #2259, opened 2023-07-14
DiscoverMojo.java:128-134: Simplify xpath. Current...


0pdd:
The puzzle `1602-59e4ff34` from #1602 has to be resolved: 

https://github.com/objectionary/eo/blob/9a7619fc3d7a00a63359cc8470d50837a3897f8e/eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java#L128-L134

The puzzle was created by rultor on 14-Jul-23. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

volodya-lombrozo:
Related discussion: https://github.com/objectionary/eo/pull/2228#discussion_r1262284714

0pdd:
@0pdd the puzzle [#2266](https://github.com/objectionary/eo/issues/2266) is still not solved.

volodya-lombrozo:
Blocked by https://github.com/jcabi/jcabi-xml/issues/211

0pdd:
@0pdd the puzzle [#2278](https://github.com/objectionary/eo/issues/2278) is still not solved; solved: [#2266](https://github.com/objectionary/eo/issues/2266).

PR #2296, opened 2023-07-21
feat(#2259): simplify xpath in DiscoveryMojo

volodya-lombrozo:
Simplify xpath in `DiscoveryMojo`.

Closes: #2259 

---

### Detailed summary
- Removed unused imports
- Simplified XPath expression for object discovery
- Refactored code to use `SaxonDocument` instead of `XMLDocument` for better performance

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java
@@ -31,7 +31,6 @@ import java.io.FileNotFoundException;
 import java.nio.file.Path;
 import java.util.Collection;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Set;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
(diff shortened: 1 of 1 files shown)


eo-maven-plugin/src/main/java/org/eolang/maven/DiscoverMojo.java
```diff
@@ -122,58 +112,27 @@ private Collection<String> discover(final Path file)
...
-    /**
-     * Get unique list of object names from given XML by provided xpath.
+     * Get unique list of object names from given XML.
      * @param xml XML.
-     * @param xpath Xpath.
-     * @return Iterable of object names.
+     * @return Object names.
      */
-    private static Set<String> names(final XML xml, final String xpath) {
+    private static Set<String> names(final XML xml) {
```
maxonfjvipon, line 119: @volodya-lombrozo suggestion: since this function is called only once, don't you think it can be removed and its content can be inlined in `discover` function?

volodya-lombrozo: @maxonfjvipon I don't like this idea. I prefer using many small methods, where each performs a small unit of work. In this case, it's much easier to understand the code. Originally it isn't my idea:

In [Clean Code: A Handbook of Agile Software Craftsmanship](https://www.amazon.com/dp/0132350882), Robert Martin says:
> The first rule of functions is that they should be small. The second rule of functions is that they should be smaller than that. Functions should not be 100 lines long. Functions should hardly ever be 20 lines long.

maxonfjvipon: @volodya-lombrozo fair enough

yegor256:
@rultor merge

Issue #2302, opened 2023-07-22
ProbeMojo.java:113-118: Use objectionaries to probe...


0pdd:
The puzzle `1602-73bdd8c0` from #1602 has to be resolved: 

https://github.com/objectionary/eo/blob/a0f26968e5c55c6b43befa841d983b4f8e8f6ff5/eo-maven-plugin/src/main/java/org/eolang/maven/ProbeMojo.java#L113-L118

The puzzle was created by rultor on 22-Jul-23. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
The puzzle `1602-73bdd8c0` has disappeared from the source code, that's why I closed this issue.

0pdd:
@0pdd 5 puzzles [#2327](https://github.com/objectionary/eo/issues/2327), [#2328](https://github.com/objectionary/eo/issues/2328), [#2329](https://github.com/objectionary/eo/issues/2329), [#2330](https://github.com/objectionary/eo/issues/2330), [#2331](https://github.com/objectionary/eo/issues/2331) are still not solved.

0pdd:
@0pdd 5 puzzles [#2327](https://github.com/objectionary/eo/issues/2327), [#2329](https://github.com/objectionary/eo/issues/2329), [#2330](https://github.com/objectionary/eo/issues/2330), [#2331](https://github.com/objectionary/eo/issues/2331), [#2339](https://github.com/objectionary/eo/issues/2339) are still not solved; solved: [#2328](https://github.com/objectionary/eo/issues/2328).

0pdd:
@0pdd 4 puzzles [#2327](https://github.com/objectionary/eo/issues/2327), [#2330](https://github.com/objectionary/eo/issues/2330), [#2331](https://github.com/objectionary/eo/issues/2331), [#2339](https://github.com/objectionary/eo/issues/2339) are still not solved; solved: [#2328](https://github.com/objectionary/eo/issues/2328), [#2329](https://github.com/objectionary/eo/issues/2329).

0pdd:
[...]

PR #2312, opened 2023-07-25
#2302 Objectionaries in `ProbeMojo` + tests

maxonfjvipon:
Ref: #1602 

What's done:
- Introduced `Objectionaries` interface - "hash-objectionary" hash-map
- `Map<String, Objectionary>` replaced with `Objectionaries` in `ProbeMojo` and `PullMojo`
- Removed unused fields from `AssembleMojo`, `ProbeMojo` and `PullMojo`
- Refactored unit tests for `ProbeMojo`, `PullMojo` and some others
- Added disabled unit tests for `ProbeMojo` with `todo` for future implementation of probing objects with versions
- Made `OyFallbackSwap` lazier

---

## PR-Codex overview
This PR focuses on refactoring the objectionary handling in the EO Maven Plugin. It introduces the `Objectionaries` interface and the `ObjsDefault` implementation. 

### Detailed summary
- Added `Objectionaries` interface for managing objectionaries
- Added `ObjsDefault` implementation of `Objectionaries`
- Updated code to use `Objectionaries` instead of `Objectionary`

> The following files were skipped due to too many changes: `eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/Objectionaries.java`, `eo-maven-plugin/src/main/java/org/eolang/maven/ProbeMojo.java`, `eo-maven-plugin/src/main/java/org/eolang/maven/PullMojo.java`, `eo-maven-plugin/src/test/java/org/eolang/maven/ProbeMojoTest.java`


maxonfjvipon:
@volodya-lombrozo please review this one

eo-maven-plugin/src/test/java/org/eolang/maven/ProbeMojoTest.java
```diff
@@ -122,6 +162,134 @@ void findsProbesInOyRemote(@TempDir final Path temp) throws IOException {
...
+            String.format(
+                "Tojos should have contained versioned object %s after probing, but they didn't",
+                object
+            ),
+            maven.externalTojos().contains(object),
+            Matchers.is(true)
+        );
+        MatcherAssert.assertThat(
+            "First entry of tojos after probing should have contained one probed object, but it didn't",
+            ProbeMojoTest.firstEntry(
```
volodya-lombrozo, line 191: @maxonfjvipon Looks like you are tying to check internals instead of checking actual results (some sort of encapsulation violation). Or I'm wrong here? It's a lot of questions actually:
- why it should be the "first" entity?
- why "1"?
And so on

volodya-lombrozo (changes requested):

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/OjsDefault.java
```diff
@@ -0,0 +1,64 @@
...
+package org.eolang.maven.objectionary;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Default objectionaries.
+ * @since 0.29.6
+ */
+public final class OjsDefault implements Objectionaries {
```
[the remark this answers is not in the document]
maxonfjvipon: @volodya-lombrozo `Objs` is good. Do we allow 4 chars as prefix?

eo-maven-plugin/src/test/java/org/eolang/maven/ProbeMojoTest.java
```diff
@@ -122,6 +162,134 @@ void findsProbesInOyRemote(@TempDir final Path temp) throws IOException {
...
+    @ExtendWith(OnlineCondition.class)
+    @Disabled
+    void findsProbesWithVersionsInOneObjectionary(@TempDir final Path temp) throws IOException {
+        final CommitHash hash = new CommitHashesMap.Fake().get("0.28.10");
+        final String object = "org.eolang.io.stdout|9b88393";
+        final FakeMaven maven = new FakeMaven(temp)
+            .with("objectionaries", new OjsDefault().with(hash, new OyRemote(hash)))
+            .with("withVersions", true)
+            .withProgram(
+                "+package org.eolang.custom\n",
```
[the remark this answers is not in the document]
maxonfjvipon: @volodya-lombrozo Sure, I think puzzle for now

eo-maven-plugin/src/test/java/org/eolang/maven/ProbeMojoTest.java
maxonfjvipon: @volodya-lombrozo 
1. I believe, encapsulation violation is something about `get`, `set` and reflection. Here we just read the file and get first row from it.
2. It should be "first" because only one expect this first tojo was added on `register` step and contained `probed` field after `probe` step
3. "1" because we expect that `probed` field would contain value "1"
4. This test class is full of using this static method and it seems convenient

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/OjsDefault.java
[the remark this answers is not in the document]
volodya-lombrozo: @maxonfjvipon If it isn't forbidden by some linter

eo-maven-plugin/src/test/java/org/eolang/maven/ProbeMojoTest.java
volodya-lombrozo: @maxonfjvipon The "first entity of the foreign tojos" looks strange. It looks like we are tying to scan an intermediate state which we are trying to read in the middle of the process. I'm afraid it could lead to a fragile implementation. Could we check the result somehow else? (Or to skip that check at all?). 

If not, can we add a puzzle which will request the refactoring of that test class in order to make it more readable?

eo-maven-plugin/src/test/java/org/eolang/maven/ProbeMojoTest.java
```diff
@@ -122,6 +162,134 @@ void findsProbesInOyRemote(@TempDir final Path temp) throws IOException {
...
+                "probed"
+            ),
+            Matchers.equalTo("2")
+        );
+    }
+
+    @Test
+    @ExtendWith(OnlineCondition.class)
+    @Disabled
+    void findsProbesWithDefaultHash(@TempDir final Path temp) throws IOException {
```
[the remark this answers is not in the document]
maxonfjvipon: @volodya-lombrozo I don't know (at least for now) how to resolve that, because tests are similar but not identical. Every test has something a bit different from other one. I suggest to add a puzzle for that

maxonfjvipon:

maxonfjvipon:

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
```diff
@@ -229,7 +230,13 @@ public <T extends AbstractMojo> FakeMaven execute(final Class<T> mojo) throws IO
             this.params.putIfAbsent("generatedDir", this.generatedPath().toFile());
             this.params.putIfAbsent("placedFormat", "csv");
             this.params.putIfAbsent("plugin", FakeMaven.pluginDescriptor());
-            this.params.putIfAbsent("objectionary", new Objectionary.Fake());
+            this.params.putIfAbsent(
+                "objectionaries",
+                new ObjsDefault().with(
```
volodya-lombrozo, line 235: @maxonfjvipon Maybe we need the fake implementation of `Objectionaries`?

volodya-lombrozo (changes requested):

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
maxonfjvipon: @volodya-lombrozo I've tried to implement it, but I don't know how to properly use it, because we need to configure `Objectionaries` with different hashes and different objectionaries in the tests and such objectionaries are not always fake

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
```diff
@@ -370,6 +377,14 @@ Path foreignPath() {
         return this.workspace.absolute(Paths.get("eo-foreign.csv"));
     }
 
+    /**
+     * Path to or eo-external.* file after all changes.
+     * @return Path to eo-foreign.* file.
+     */
+    Path externalPath() {
```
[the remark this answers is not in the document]
maxonfjvipon: @volodya-lombrozo we have `external()` method for that. External tojos is a copy of foreign tojos. We have a puzzle for that, so there are kind of identical methods for both of them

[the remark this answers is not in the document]
maxonfjvipon: @volodya-lombrozo this method was not introduced in this PR. It was here before. I just removed `private` modifier and had to move it to other place in the file in order to observe correct order of methods

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
maxonfjvipon: @volodya-lombrozo and since `ObjsDefault` does not rewrite `Objectionary` by the hash - we won't be able to replace it with needed one

maxonfjvipon:

maxonfjvipon:

eo-maven-plugin/src/main/java/org/eolang/maven/ProbeMojo.java
```diff
@@ -135,7 +109,7 @@ public void exec() throws IOException {
             }
             int count = 0;
             for (final String name : names) {
-                if (!this.objectionary.contains(name)) {
+                if (!this.objectionaryByHash(this.hsh).contains(name)) {
```
volodya-lombrozo, line 112: @maxonfjvipon Looks like we can add `contains()` method to `Objectionaries` interface. In this case  we will be able to encapsulate the logic of the `objectionaryByHash` inside `Objectionaries` implementation (and apparently we won't need `with` method).
In other words, I suggest treating `Objectionaries` as a living object and showing it respect, not merely treating it as a bag of `Objectionary` objects.

https://www.yegor256.com/2014/11/20/seven-virtues-of-good-object.html#1-he-exists-in-real-life:
> Moreover, an object should be [anthropomorphized](https://en.wikipedia.org/wiki/Anthropomorphism), i.e. treated like a human being (or a pet, if you like them more). By this I basically mean that an object is not a data structure or a collection of functions.

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/ObjsDefault.java
```diff
@@ -0,0 +1,65 @@
...
+     * Ctor.
+     * @param ojs Objectionaries hash-map.
+     */
+    ObjsDefault(final Map<String, Objectionary> ojs) {
+        this.map = ojs;
+    }
+
+    @Override
+    public Objectionaries with(final CommitHash hash, final Objectionary objectionary) {
+        this.map.putIfAbsent(hash.value(), objectionary);
```
volodya-lombrozo, line 57: @maxonfjvipon Maybe it's better to overwrite values when we call `with(...)` all the time? Otherwise it could be misleading

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
volodya-lombrozo: @maxonfjvipon When we need to use particular (real) `Objectionaries`, let's use `ObjsDefault`, otherwise I believe we have to use `Fake` objects. For example:
```java
new ObjsDefault().with(
        new CommitHash.ChConstant("9c46a671f2bc68e777aab031d57da5012ba807a7"),
        new Objectionary.Fake()
    )
);
```
This code snippet creates lots of questions:
1. Why do we need `"9c46a671f2bc68e777aab031d57da5012ba807a7"` value here? (I guess, we don't need the particular value for that test) 
2. Why do I need to construct `Objectionaries` for that test this way? Actually, I don't want to know how to construct `Objectionaries`, I want just to write `new Objectionaries.Fake()` and forget about it.

volodya-lombrozo (changes requested):

eo-maven-plugin/src/main/java/org/eolang/maven/ProbeMojo.java
```diff
@@ -80,50 +80,24 @@ public final class ProbeMojo extends SafeMojo {
...
-    @Parameter(property = "offlineHash")
-    private String offlineHash;
-
-    /**
-     * The objectionary.
+     * @since 0.29.6
      */
     @SuppressWarnings("PMD.ImmutableField")
-    private Objectionary objectionary;
+    private CommitHash hsh;
```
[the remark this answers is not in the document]
maxonfjvipon: @volodya-lombrozo we have `hash` field in `AssembleMojo` and it's a `String`. So there will be conflict when `AssembleMojo` is coping their fields into `ProbeMojo` via `Moja.copy()` method. I added a puzzle for renaming it.
`hash` is definitely best name for the field. `hsh` is temporary solutions just not to forget what's for

eo-maven-plugin/src/main/java/org/eolang/maven/ProbeMojo.java
maxonfjvipon: @volodya-lombrozo The problem is here that `ProbeMojo` and `PullMojo` where `Objectionaries` is used - require different objectionaries to insert into `Objectionaries`. So we can't just let `Objectionaries` put any objectionary it wants, we need to give the objectionary to it from the outside.
In your case `contains` will look like 
```java
bool contains(CommitHash hash, String name, Objectionary obj)
```
Here `hash` - key in map, `name` - name of object, `obj` - objectionary to put if it's absent with given `hash`
So I don't think it's a good idea

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
maxonfjvipon: @volodya-lombrozo 
1) it's hash by `master` tag, maybe it can be replaced with `ChRemote("master")`
2) when you have one `Objectionary` - it's possible to use `Objectionary.Fake`. When it's a map of objectionaries - think you can't (at least not so easy). In the first test you want to test long hash with fake objectionary,
in the second you want to test `ChRemote("master")` with `OyRemote`, in the third you want to test two different objectionaries with different hashes...

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/ObjsDefault.java
maxonfjvipon: @volodya-lombrozo I don't think so, at least for now, when `Objectionaries` is used only by `ProbeMojo` and `PullMojo` where it should not be overwritten. Otherwise we could you just a simple `HashMap` instead of `Objectionaries`

volodya-lombrozo: @maxonfjvipon Right now we have the next problem:
1. I create `Objectionaries`
2. Call `.with(First Objectionary)`
3. Call `.with(Second Objectionary)`
4. call `.get()` and receive the `First Objectionary`. 

Again the design is misleading. I'm like a developer expect to get `Second Objectionary`. In other words, you just swallow `Second Objectionary` for now

eo-maven-plugin/src/main/java/org/eolang/maven/ProbeMojo.java
volodya-lombrozo: @maxonfjvipon Can we create `Objectionaries` by using a constructor, instead of `with` method? I don't understand this: 
> The problem is here that ProbeMojo and PullMojo where Objectionaries is used - require different objectionaries to insert into Objectionaries 

Why we can't create `Objectionaries` by using constructor (and provide all the required `Objectionary` instances) and then just call `contains()` on it? Why do we need to `get()` it and then call `contains()`?

maxonfjvipon:

volodya-lombrozo:
@maxonfjvipon Looks like the code has some problems which were reported by `qulice`:
>There are 2 violations

maxonfjvipon:

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
```diff
@@ -229,7 +229,10 @@ public <T extends AbstractMojo> FakeMaven execute(final Class<T> mojo) throws IO
             this.params.putIfAbsent("generatedDir", this.generatedPath().toFile());
             this.params.putIfAbsent("placedFormat", "csv");
             this.params.putIfAbsent("plugin", FakeMaven.pluginDescriptor());
-            this.params.putIfAbsent("objectionary", new Objectionary.Fake());
+            this.params.putIfAbsent(
```
volodya-lombrozo, line 232: @maxonfjvipon Can we write it in a single line?

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/Objectionaries.java
```diff
@@ -0,0 +1,85 @@
...
+     * Fake objectionaries.
+     * Contains only one default objectionary that will be returned by any hash.
+     *
+     * @since 0.29.6
+     */
+    class Fake implements Objectionaries {
+        /**
+         * Default objectionary.
+         */
+        private final Objectionary objry;
```
volodya-lombrozo, line 58: @maxonfjvipon Can we use the full name here instead of an abbreviation?

volodya-lombrozo (changes requested):

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/Objectionaries.java
maxonfjvipon: @volodya-lombrozo If I use `objectionary` then I get checkstyle warning on `with` method (objectionary hides the field)

eo-maven-plugin/src/main/java/org/eolang/maven/objectionary/Objectionaries.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    Objectionary get(CommitHash hash);
+
+    /**
+     * Fake objectionaries.
+     * Contains only one default objectionary that will be returned by any hash.
+     *
+     * @since 0.29.6
+     */
+    class Fake implements Objectionaries {
```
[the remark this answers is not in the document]
maxonfjvipon: @volodya-lombrozo Let's stick with `Fake` since we use `Fake` everywhere

maxonfjvipon:

maxonfjvipon:

volodya-lombrozo (approved):
@yegor256 Could you have a look, please?

maxonfjvipon:
@yegor256 please have a look

maxonfjvipon:
@yegor256 please have a look

yegor256:
@rultor mertge

Issue #1307, opened 2022-10-04
org.eolang.cage must prohibit different object "types"


yegor256:
The `cage` object must return an `error` if we try to `.write` an object that is not a copy of the abstract object that was saved before.

yegor256:
@Graur please, help

Graur:
@yegor256 Do we need to save the type of the memoized object? Otherwise we can check its length

yegor256:
@maxonfjvipon please, help

maxonfjvipon:
@yegor256 please have a look at this.
```
[x] > obj
  x > @
obj 1 > obj1

# Case 1
cage obj > cg
cg.write obj1

# Case 2
cage obj1 > cg
cg.write obj
```

1. Init `cage` with abstract object `obj` then write its copy `obj1` with bounded `x`
2. Init `cage` with copy (of `obj`) `obj1` then write abstract `obj`

Should these both cases be legal?

yegor256:
@maxonfjvipon both cases should be illegal, in my opinion

0pdd:
@yegor256 2 puzzles [#2335](https://github.com/objectionary/eo/issues/2335), [#2336](https://github.com/objectionary/eo/issues/2336) are still not solved.

0pdd:
@yegor256 2 puzzles [#2335](https://github.com/objectionary/eo/issues/2335), [#2373](https://github.com/objectionary/eo/issues/2373) are still not solved; solved: [#2336](https://github.com/objectionary/eo/issues/2336).

0pdd:
@yegor256 all 3 puzzles are solved here: [#2335](https://github.com/objectionary/eo/issues/2335), [#2336](https://github.com/objectionary/eo/issues/2336), [#2373](https://github.com/objectionary/eo/issues/2373).

PR #2332, opened 2023-07-28
#1307 `cage.write` checks types

maxonfjvipon:
Closes: #1307 

What's done:
- `cage.write` checks type of given object, if it does not equal to type of object saved before - throw an exception
- Added tests to check if type checking works correctly
- Added `todo` to refactor and enable some `cage` tests that fail because of this new feature

---

## PR-Codex overview
This PR focuses on improving the code diff by making various changes and fixes. 

### Detailed summary
- Updated code in `eo-runtime/src/test/eo/org/eolang/runtime-tests.eo`
- Updated code in `eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java`
- Updated code in `eo-maven-plugin/src/test/java/org/eolang/maven/BinarizeMojoTest.java`
- Updated code in `eo-runtime/src/test/java/EOorg/EOeolang/EOcageTest.java`
- Updated code in `eo-runtime/src/main/java/org/eolang/AtCage.java`
- Updated code in `eo-runtime/src/test/eo/org/eolang/cage-tests.eo`

### DIFF
--- a/eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
+++ b/eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
@@ -27,6 +27,7 @@
  */
 package EOorg.EOeolang;
 
+import org.eolang.AtCage;
 import org.eolang.AtComposite;
 import org.eolang.AtFree;
 import org.eolang.Data;
@@ -51,7 +52,7 @@ public class EOcage extends PhDefault {
      */
     public EOcage(final Phi sigma) {
         super(sigma);
-        this.add("enclosure", new AtMemoized());
+        this.add("enclosure", new AtCage());
         this.add("φ", new AtComposite(this, rho -> rho.attr("enclosure").get()));
         this.add("write", new AtComposite(this, EOcage.Write::new));
     }
@@ -61,7 +62,7 @@ public class EOcage extends PhDefault {
      * @since 0.17
      */
     @XmirObject(oname = "cage.write")
-    private final class Write extends PhDefault {
+    private static final class Write extends PhDefault {
         /**
          * Ctor.
          * @param sigma Sigma
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/AssembleMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/AssembleMojo.java
@@ -70,7 +70,7 @@ public final class AssembleMojo extends SafeMojo {
      * @checkstyle MemberNameCheck (6 lines)
      * @checkstyle ConstantUsageCheck (5 lines)
      */
-    private final Objectionaries objectionaries = new ObjsDefault(this.cache,  () -> this.session.getRequest().isUpdateSnapshots());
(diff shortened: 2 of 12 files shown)


maxonfjvipon:
@Graur please review this one

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
```diff
@@ -74,9 +76,22 @@ private final class Write extends PhDefault {
...
                     rho -> {
-                        rho.attr("σ").get().attr("enclosure").put(
-                            rho.attr("x").get()
-                        );
+                        final Attr attr = rho.attr("σ").get().attr("enclosure");
+                        Phi enclosure;
+                        try {
+                            enclosure = attr.get();
+                        } catch (final EOerror.ExError error) {
+                            enclosure = null;
```
Graur, line 84: @maxonfjvipon We are trying to avoid `null` in Java. According to: https://www.yegor256.com/2014/05/13/why-null-is-bad.html

Graur: @maxonfjvipon Also here we got an exception swallowing. It is better to use [exception chaining](https://en.wikipedia.org/wiki/Exception_chaining) in such cases.

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
```diff
@@ -74,9 +76,22 @@ private final class Write extends PhDefault {
...
+                        }
+                        final Phi put = rho.attr("x").get();
```
Graur, line 86: @maxonfjvipon Variable name `put` is a verb. Can we make it a noun?

Graur (changes requested):
@maxonfjvipon Good one! Just a few comments from my side:

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
maxonfjvipon: @Graur I need to swallow that exception here, because I'm trying to get attribute `enclosure` which can be empty. If it's empty that means that we did not save anything into `cage` => no need to check types, just save new value

maxonfjvipon:
@Graur please have a look one more time

maxonfjvipon:
@Graur please have a look one more time

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
```diff
@@ -61,7 +62,7 @@ public EOcage(final Phi sigma) {
      * @since 0.17
      */
     @XmirObject(oname = "cage.write")
-    private final class Write extends PhDefault {
+    private static final class Write extends PhDefault {
```
Graur, line 65: @maxonfjvipon Why do you make it static?

maxonfjvipon: @Graur IDEA suggested, I fell for it :)

Graur (approved):

PR #2419, opened 2023-08-18
2314 handle vertex returning

levBagryansky:
### Focus of this PR:
This PR focuses on implementing new functionality and making improvements to existing code in the eo-runtime and EOorg/EOrust modules.

### Detailed summary:
- In `eo_enum.rs`, the code has been modified to use big-endian byte order instead of little-endian byte order.
- In `rust-tests.eo`, a new function `foo` has been added which retrieves a value from the environment and returns an `EOVertex` object.
- In `Universe.java`, a new constructor has been added that accepts a `Map<Integer, Phi>` parameter and initializes the `indexed` field with it. The existing constructor has been modified to call this new constructor with an empty `HashMap` as the parameter.
- The `find` method in `Universe.java` has been modified to skip empty attributes when splitting the name string.
- The `put` method in `Universe.java` has been modified to do nothing and is marked as a TODO.
- The `dataize` method in `Universe.java` has been modified to use the `get` method to retrieve the `Phi` object instead of directly accessing the `indexed` map.
- The `get` method has been added to `Universe.java` to retrieve a `Phi` object from the `indexed` map.
- The `EOrust` class in `EOrust.java` has been modified to use the new constructor of `Universe` that accepts a `Map<Integer, Phi>` parameter.
- The `translate` method in `EOrust.java` has been modified to handle the case when the determinant is 0 and retrieve the corresponding `Phi` object using the `phis` map.

### DIFF
--- a/eo-runtime/src/main/java/EOorg/EOeolang/EOrust.java
+++ b/eo-runtime/src/main/java/EOorg/EOeolang/EOrust.java
@@ -36,6 +36,8 @@ import java.nio.ByteBuffer;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import org.apache.commons.lang3.NotImplementedException;
 import org.apache.commons.lang3.SystemUtils;
@@ -67,6 +69,8 @@ import org.eolang.XmirObject;
  */
 @XmirObject(oname = "rust")
 public class EOrust extends PhDefault {
+
+    private final Map<Integer, Phi> phis = new HashMap<>();
 
     /**
      * Map with location of the `code` attribute as the key
@@ -150,10 +154,10 @@ public class EOrust extends PhDefault {
                         .attr("params").get()
                         .attr("Δ").get()
                     ).take(Phi[].class)[0];
-                    return EOrust.translate(
+                    return this.translate(
                         (byte[]) method.invoke(
                             null, new Universe(
-                                portal
+                                portal, this.phis
                             )
                         )
                     );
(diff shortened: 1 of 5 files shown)


levBagryansky:
@Graur please review

eo-runtime/src/main/java/EOorg/EOeolang/EOrust.java
```diff
@@ -74,6 +76,11 @@ public class EOrust extends PhDefault {
      */
     private static final ConcurrentHashMap<String, String> NAMES;
 
+    /**
+     * All phis indexed while executing of native method.
+     */
+    private final Map<Integer, Phi> phis = new HashMap<>();
```
Graur, line 82: @levBagryansky Should it be concurrent as well as `NAMES` ?

eo-runtime/src/main/java/EOorg/EOeolang/EOrust.java
```diff
@@ -213,17 +220,28 @@ private static ConcurrentHashMap<String, String> load(final String src) throws I
...
+                ByteBuffer buffer = ByteBuffer.allocate(Integer.BYTES);
+                buffer.put(content);
+                buffer.flip();
+                final int vertex = buffer.getInt();
+                ret = this.phis.get(vertex);
+                if (ret == null) {
+                    throw new ExFailure(
+                        String.format(
+                            "Returned phi with vertex %d was not indexed",
+                            vertex
```
Graur, line 238: @levBagryansky I'm not sure the vertex id is a good choice to error description. Can we provide some more details to this message?

eo-runtime/src/main/java/org/eolang/Universe.java
```diff
@@ -122,20 +132,30 @@ public int copy(final int vertex) {
...
+     * @throws ExFailure if vertex does not exist in the map.
+     */
+    private Phi get(final int vertex) {
+        return Optional.ofNullable(
+            this.indexed.get(vertex)
+        ).orElseThrow(
+            () -> new ExFailure(
+                String.format(
+                    "Phi object with vertex %d was not indexed.",
+                    vertex
```
Graur, line 153: @levBagryansky Same here. Can we provide some more details to the error message?

Graur (changes requested):
@levBagryansky Good one, just a few suggestions from my side

eo-runtime/src/main/java/EOorg/EOeolang/EOrust.java
levBagryansky: @Graur This map is not a static so it will be created for every thread separately

eo-runtime/src/main/java/EOorg/EOeolang/EOrust.java
levBagryansky: @Graur May be we can also print phis map ?

Graur: @levBagryansky To print all of the map may cost a lot and may confuse the developer. Maybe we can print `content` as well?

eo-runtime/src/main/java/EOorg/EOeolang/EOrust.java
```diff
@@ -213,17 +220,28 @@ private static ConcurrentHashMap<String, String> load(final String src) throws I
      *  It must convert message array from 1 to last byte to the String
      *  and return eo object with converted String Data.
      */
-    private static Phi translate(final byte[] message) {
+    private Phi translate(final byte[] message) {
```
Graur, line 223: @levBagryansky Should we also remove one of the puzzles above?

eo-runtime/src/main/java/org/eolang/Universe.java
levBagryansky: @Graur Seems no, it just throws exception if the vertex is absent in the map

Graur (approved):
@levBagryansky I believe it is better now, thanks!

levBagryansky:
@yegor256 please check

levBagryansky:
@yegor256 please cheak

Issue #2425, opened 2023-08-21
Universe.find should work with absolute path too


levBagryansky:
@Graur `Universe.find` must be able to work with absolute path in order to manage objects like `int`  or `bytes`.

Graur:
@levBagryansky Ok, let's do it

PR #2431, opened 2023-08-23
2425 find

levBagryansky:
Closes: #2425

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Universe.java
+++ b/eo-runtime/src/main/java/org/eolang/Universe.java
@@ -23,6 +23,8 @@
  */
 package org.eolang;
 
+import org.apache.commons.lang3.ArrayUtils;
+
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


levBagryansky:
@Graur please review

levBagryansky:
@Graur please take a look

eo-runtime/src/main/java/org/eolang/Universe.java
```diff
@@ -76,9 +76,25 @@ public Universe() {
      * @return Vertex of the object to find.
      */
     public int find(final String name) {
-        Phi accum = this.connector;
+        Phi accum;
         final String[] atts = Universe.replace(name)
             .split("\\.");
+        if (atts[0].equals("Q")) {
```
Graur, line 82: @levBagryansky To avoid NPE it is always better to swap the parameters, e.g.
```
"Q".equals(atts[0])
```

eo-runtime/src/main/java/org/eolang/Universe.java
```diff
@@ -76,9 +76,25 @@ public Universe() {
...
+            accum = Phi.Φ;
+            atts[0] = "";
+        } else if (atts[0].equals("ρ")) {
```
Graur, line 85: @levBagryansky Same here, can we swap the parameters?

eo-runtime/src/main/java/org/eolang/Universe.java
```diff
@@ -76,9 +76,25 @@ public Universe() {
...
+            accum = this.connector;
+        } else if (atts[0].equals("$")) {
```
Graur, line 87: @levBagryansky Same here, can we swap the parameters?

Graur (changes requested):
@levBagryansky Good one, just a few comments from my side:

Graur (approved):
@levBagryansky Good one
@yegor256 Please take a look

levBagryansky:
@yegor256 please check

eo-runtime/src/main/java/org/eolang/Universe.java
```diff
@@ -76,9 +76,30 @@ public Universe() {
...
+            atts[0] = "";
+        } else if ("ρ".equals(atts[0])) {
+            accum = this.connector;
+        } else if ("$".equals(atts[0])) {
+            accum = this.connector;
+            atts[0] = "";
+        } else {
+            throw new ExFailure(
+                String.format(
+                    "Universe.find starts with %s, but it should start with Q or ^ or $ only",
```
yegor256, line 98: @levBagryansky should be only `Q` or `$`, nothing else. If `\rho` is needed, it should be `$.^`

yegor256:
@levBagryansky one comment above

levBagryansky:
@yegor256 please take a look

yegor256:
@rultor merge

Issue #2464, opened 2023-09-01
Replace `convertsAntlrToEbnf()` test method with an appropriate CI pipeline.


volodya-lombrozo:
This [discussion](https://github.com/objectionary/eo/pull/2459#pullrequestreview-1606801654) reviled the problem with `XMIRTest#convertsAntlrToEbnf()` method. It is actually not a test method, it is ad-hoc solution for creating `ebnf` files:

>I don't know what is a better place for convertsAntlrToEbnf() method... Maybe we can create a new Maven plugin that would take .g4 files, generate .txt files, and then compile them to .png and .svg via pdflatex (this exactly what this test method is doing now).

We have several options here, but the main goal is to move logic from that method to an appropriate place.

yegor256:
@volodya-lombrozo I believe, it's fixed, can be close

PR #2469, opened 2023-09-04
EBNF rendering fixed

yegor256:
There was a typo, we were replacing `&` instead of a single quote.

---

## PR-Codex overview
This PR focuses on improving the XMIRTest and to-ebnf.xsl files. 

### Detailed summary
- Added new assertions to XMIRTest
- Updated variable replacement in to-ebnf.xsl

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/XMIRTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/XMIRTest.java
@@ -128,8 +128,8 @@ final class XMIRTest {
                 Matchers.containsString("<byte> := /[0-9A-F]/ /[0-9A-F]/ \\\\"),
                 Matchers.containsString("<string> := 'DQ' {"),
                 Matchers.containsString("<text> := 'DQ' 'DQ' 'DQ' ["),
-                Matchers.containsString("\\textvisiblespace{}"),
-                Matchers.containsString("\\textquotesingle{}")
+                Matchers.containsString("\"\\textvisiblespace{}\""),
+                Matchers.containsString("\"\\textquotesingle{}\"")
             )
         );
         Files.write(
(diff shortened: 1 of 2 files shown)


eo-parser/src/test/java/org/eolang/parser/XMIRTest.java
```diff
@@ -127,7 +127,9 @@ void convertsAntlrToEbnf(@TempDir final Path temp) throws Exception {
             Matchers.allOf(
                 Matchers.containsString("<byte> := /[0-9A-F]/ /[0-9A-F]/ \\\\"),
                 Matchers.containsString("<string> := 'DQ' {"),
-                Matchers.containsString("<text> := 'DQ' 'DQ' 'DQ' [")
+                Matchers.containsString("<text> := 'DQ' 'DQ' 'DQ' ["),
+                Matchers.containsString("\\textvisiblespace{}"),
+                Matchers.containsString("\\textquotesingle{}")
```
maxonfjvipon, line 132: @yegor256 I think it would be better add more specific matcher here. We should check not only that "string contains `\\textquotesingle{}`", but that `"'"` is successfully replaced with `\\textquotesingle{}`. WDYT?

maxonfjvipon (changes requested):

eo-parser/src/test/java/org/eolang/parser/XMIRTest.java
yegor256: @maxonfjvipon this is true, but such a test would require a lot more efforts. We will better do it later, when we create a new Maven plugin for this kind of mapping from .g4 to .txt: https://github.com/objectionary/eo/issues/2464

maxonfjvipon: @yegor256 maybe add todo for it then?

yegor256: @maxonfjvipon I thought about this, but it doesn't make sense to have a puzzle in this repository, since the entire test will be moved out soon (https://github.com/objectionary/eo/issues/2464)

yegor256:
@rultor merge

PR #2493, opened 2023-09-11
Fixed bug related to caching.

### LINKED ISSUE #2492: Bug in caching
According to #2441 cargo compiler must not be сalled in the subsequent eo compilation. But it is.
We need to fix this bug so  compilation time will be decreased approximately by 12 seconds.

levBagryansky:
Closes #2492 
Fixed the bug and added test.
I also edited Log message which was
 `Binarized native0.rs from org.eolang.rust-tests` but I suggest
 `Binarized native0.rs from org.eolang.rust-tests:Φ.org.eolang.rust-returns-positive-int.r.α0` which is more clear as for me.

---

### Focus of the PR:
This PR focuses on improving

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeMojo.java
@@ -61,24 +61,7 @@ public final class BinarizeMojo extends SafeMojo {
     /**
      * Name of executable file which is result of cargo building.
      */
-    public static final String LIB;
-
-    static {
-        if (SystemUtils.IS_OS_WINDOWS) {
-            LIB = "common.dll";
-        } else if (SystemUtils.IS_OS_LINUX) {
-            LIB = "libcommon.so";
-        } else if (SystemUtils.IS_OS_MAC) {
-            LIB = "libcommon.dylib";
-        } else {
-            throw new IllegalArgumentException(
-                String.format(
-                    "Rust inserts are not supported in %s os. Only windows, linux and macos are allowed.",
-                    System.getProperty("os.name")
-                )
-            );
-        }
-    }
+    public static final String LIB = BinarizeMojo.common();
 
     /**
      * The directory where to binarize to.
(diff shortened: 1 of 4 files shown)


levBagryansky:
@Graur could you please review these changes?

eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeMojo.java
```diff
@@ -54,9 +55,31 @@
     threadSafe = true,
     requiresDependencyResolution = ResolutionScope.COMPILE
 )
-@SuppressWarnings("PMD.LongVariable")
+@SuppressWarnings({"PMD.LongVariable", "PMD.StaticAccessToStaticFields"})
```
Graur, line 58: @levBagryansky Maybe we add todo here? And think about how to avoid this PMD exclusion later. But it is up to you

Graur (approved):
@levBagryansky I can't find out how to avoid static section, so I decided to accept this pull request

eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeMojo.java
levBagryansky: @Graur I think we should not avoid this exclusion because `StaticAccessToStaticFields` rule is for other cases.

levBagryansky:
@yegor256 please check

levBagryansky:
@yegor256 please check

levBagryansky:
@yegor256 please check

levBagryansky:
@yegor256 please check

eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeMojo.java
```diff
@@ -54,9 +55,31 @@
...
 public final class BinarizeMojo extends SafeMojo {
 
+    /**
+     * Name of executable file which is result of cargo building.
+     */
+    public static final String LIB;
```
yegor256, line 64: @levBagryansky this is not a good design. Instead, do this:

```
private static final String LIB = BinarizeMojo.common();
private static String common() {
  //...
}
```

levBagryansky:
@yegor256 please take a look

eo-maven-plugin/src/main/java/org/eolang/maven/BinarizeMojo.java
yegor256: @levBagryansky maybe you can remove this suppression too

yegor256:
@rultor merge

PR #2474, opened 2023-09-05
feat(#2399): new syntax ebnf

maxonfjvipon:
Closes: #2399 
Closes: #2407

---

## PR-Codex overview
This PR focuses on various changes in the codebase, including deletion of files, modifications in YAML files, changes in Java code, and updates to the pom.xml file.

### Detailed summary
- Deletion of several YAML files in the `eo-parser` module
- Modifications in YAML files in the `eo-parser` module
- Changes in Java code in the `eo-maven-plugin` and `eo-parser` modules
- Updates to the `pom.xml` file in the `eo-parser` module

> The following files were skipped due to too many changes: `eo-parser/src/test/java/org/eolang/parser/XMIRTest.java`, `eo-runtime/src/test/eo/org/eolang/bytes-tests.eo`, `eo-parser/src/main/antlr4/org/eolang/parser/Program.g4`, `eo-parser/src/main/java/org/eolang/parser/XeListener.java`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #2501, opened 2023-09-19
feat(#1602): default empty versions near to the root object

maxonfjvipon:
Ref: #2503

What's done:
- added flag `versioned` to `Syntax`, `ParsingTrain`, `XeListener` and few more classes
- depends on the flag `XeListener` adds empty `ver` attribute to ALL objects
- this empty `ver` will be replaced with default hash in the next PR
- depends on the flag `ParsingTrain` applies `move-versions-down.xsl` transformation
- added `move-versions-down.xsl` transformation that moves `ver` attribute from the top object to the root (`.org`). When this sequence of objects will be traspiled to java (smth like `Phi.Ф.attr("org").get().attr("eolang").get()...`) it will check if there is a "ver" attribute and paste hash right after `Phi.Ф` so it would be `Phi.Ф.attr("hash-here").get().attr("org").get()...`

---

## PR-Codex overview
This PR focuses on enabling object versioning in the code. 

### Detailed summary
- Added a new boolean flag `versioned` to the `Syntax` class constructor.
- Modified the `Syntax` class constructor to accept the `versioned` flag.
- Added a new method `oprop` to the `Objects` interface.
- Added a new XSL file for applying changes to XML after processing all shifts.
- Updated the `ParsingTrain` class to use the new XSL file for versioning.

> The following files were skipped due to too many changes: `eo-parser/src/main/java/org/eolang/parser/ParsingTrain.java`, `eo-parser/src/main/java/org/eolang/parser/XeListener.java`, `eo-parser/src/main/resources/org/eolang/parser/move-versions-deeper.xsl`


maxonfjvipon:
@volodya-lombrozo please review this one

maxonfjvipon:
@volodya-lombrozo please review this one

volodya-lombrozo (changes requested):
@maxonfjvipon I've tried to understand the reason behind that changes and I failed. Could you please attach that pr to a particular issue or add more human-readable message, please?

eo-parser/src/test/resources/org/eolang/parser/packs/move-versions-down.yaml
```diff
@@ -0,0 +1,19 @@
+versioned: true
+tests:
```
volodya-lombrozo, line 2: @maxonfjvipon In many cases here you are trying to find particular object, I guess. But it's hard to match `eo` code with `xpath`. You also use `name` attribute for some object. Maybe we can add `name` check for `xpath`'s? In this case it would help a reader to find an object.

eo-parser/src/main/resources/org/eolang/parser/move-versions-down.xsl
```diff
@@ -0,0 +1,56 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" id="move-versions-down" version="2.0">
+  <!--
+  Here we add empty "ver" attribute for all objects, which "base" starts with "org.eolang."
```
volodya-lombrozo, line 27: @maxonfjvipon Is `move-versions-down.xsl` a suitable name for transformation with that description:
> Here we add empty "ver" attribute for all objects, which "base" starts with "org.eolang.

Why "down"?

volodya-lombrozo (changes requested):
@maxonfjvipon Maybe I have some misunderstanding here, but there are my questions:
1. Why do you mix adding `boolean versioned` and `move-versions-down.xsl` in the same PR?
2. Which part of the [issue](https://github.com/objectionary/eo/issues/2503 ) will be solved by that PR?

eo-parser/src/main/java/org/eolang/parser/Objects.java
```diff
@@ -69,6 +70,13 @@ interface Objects extends Iterable<Directive> {
      */
     void xprop(String key, Object xpath);
 
+    /**
+     * Add property depends on given condition.
+     * @param condition Condition
+     * @param key Key
+     */
+    void oprop(boolean condition, String key);
```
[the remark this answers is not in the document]
maxonfjvipon: @volodya-lombrozo first "o" stands for "optional". Property that will be added depends on giving condition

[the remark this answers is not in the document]
volodya-lombrozo: @maxonfjvipon Why do we need such a method, if in case of `condition==false` we can just avoid calling this method?

eo-parser/src/main/resources/org/eolang/parser/move-versions-down.xsl
maxonfjvipon: @volodya-lombrozo the description is wrong, I will fix it. The version should move inwards to "org" object.
From this:
```xml
<o ver="123">
  <o>
    <o>
      <o base="org">
        <o base="Q"/>
      </o>
    </o>
  </o>
</o>
```
To this:
```xml
<o>
  <o>
    <o>
      <o base="org" ver="123">
        <o base="Q"/>
      </o>
    </o>
  </o>
</o>
```

eo-parser/src/main/java/org/eolang/parser/Objects.java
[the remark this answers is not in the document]
maxonfjvipon: @volodya-lombrozo this method is used 8 times in `XeListener`. We either call `oprop` 8 times or replace it with `if (this.versioned) { this.objects.prop(key); }` 8 times

maxonfjvipon:
@volodya-lombrozo 
> 1. Why do you mix adding boolean versioned and `move-versions-down.xsl` in the same PR?

It just happened that way. I agree that these changes may be in separated PRs

> 2. Which part of the issue will be solved by that PR?

It's a preparation step for part 3 of the ticket. Version is moved to the "org" object. So when this xmir will be transpiled to java, `to-java.xsl` It will check if "ver" attribute exists there. If so - `to-java.xsl` will add extra `attr("hash").get()` java code

eo-parser/src/main/java/org/eolang/parser/Objects.java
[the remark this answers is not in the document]
volodya-lombrozo: @maxonfjvipon I don't know how it happened that we have `xprop` method, but both of them `xprop` and `oprop`:
1. Hard readable.
2. Might produce some [issues](https://www.yegor256.com/2023/08/01/method-overloading.html).

eo-parser/src/main/resources/org/eolang/parser/move-versions-deeper.xsl
```diff
@@ -0,0 +1,77 @@
...
+    <o ver="123">            <o>
+      <o>                      <o>
+        <o base="org">           <o base="org" ver="123">
+          <o base="Q"/>    =>      <base="Q"/>
+        </o>                     </o>
+      </o>                     </o>
+    </o>                     </o>
+  </o>                     </o>
+  -->
+  <xsl:output encoding="UTF-8" method="xml"/>
```
volodya-lombrozo, line 50: @maxonfjvipon It isn't obvious from the code why do we need to move versions "deeper" to other objects. Moreover, it looks even confusing: we set the version "123" for one object and then `move-versions-deeper.xsl` put that version to another object. Why? I don't know.

eo-maven-plugin/src/main/java/org/eolang/maven/optimization/OptSpy.java
```diff
@@ -42,11 +42,26 @@ public final class OptSpy implements Optimization {
     private final Path target;
 
     /**
-     * The main constructor.
-     * @param target Where to track optimization steps.
+     * Apply versions shifts or not.
+     */
+    private final boolean versioned;
```
volodya-lombrozo, line 47: @maxonfjvipon The solution with the `versioned` field is highly questionable. First, it leads to hidden [configurable objects](https://www.yegor256.com/2016/04/19/object-must-not-be-configurable.html). Secondly, it necessitates numerous changes in various locations, which is a bad sign. Might there be an alternative solution? Can we apply `move-versions-deeper.xsl` to all parsing pipelines (independently if we have versions or not)?

volodya-lombrozo (changes requested):

eo-parser/src/main/java/org/eolang/parser/Objects.java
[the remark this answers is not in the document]
maxonfjvipon: @volodya-lombrozo there's nothing about overloading
1. `prop` - just adds property, uses `attr` from Xemlby
2. `xprop` - changes property by given xpath, uses `xattr` from Xembly. It was added here: #2474 
3. `oprop` - addes property if given condition is TRUE. Actually is a temporary method. When object versioning is implemented it will be removed and all its occurrences will be replaced with `prop`

eo-parser/src/main/resources/org/eolang/parser/move-versions-deeper.xsl
maxonfjvipon: @volodya-lombrozo I extended docs

maxonfjvipon:

eo-parser/src/main/java/org/eolang/parser/XeListener.java
```diff
@@ -347,8 +360,10 @@ public void enterApplicable(final ProgramParser.ApplicableContext ctx) {
             if (ctx.STAR() != null) {
                 base = "tuple";
                 this.objects.prop("data", "tuple");
+                this.objects.oprop(this.versioned, "ver");
```
volodya-lombrozo, line 363: @maxonfjvipon If we add `ver` attribute for all objects without checking `versioned`, will it break something?

eo-parser/src/main/java/org/eolang/parser/ParsingTrain.java
```diff
@@ -110,34 +121,41 @@ public ParsingTrain() {
...
     ParsingTrain(final String... sheets) {
         super(
-            new TrLambda(
-                new TrFast(
-                    new TrLambda(
-                        new TrLogged(
-                            new TrClasspath<>(sheets).back(),
-                            ParsingTrain.class,
-                            Level.FINEST
+            new TrJoined<>(
```
volodya-lombrozo, line 124: @maxonfjvipon Why do we need to add `move-versions-deeper.xsl` to the pipeline as a separate `TrDefailt`? Can we just add `move-versions-deeper.xsl` to the `ParsingTrain.SHEETS`?

eo-parser/src/main/resources/org/eolang/parser/move-versions-deeper.xsl
```diff
@@ -0,0 +1,87 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" id="move-versions-deeper" version="2.0">
+  <!--
+  Here we move "ver" attribute from the top object deeper to pre-last one (before <o base="Q"/>)
+
+  1. The reason of transformation - move version to pre-last object ("org"). Then on "transpile"
+     step "to-java.xsl" will add extra {@code attr("some-hash).get()} java code.
+     More details here: https://github.com/objectionary/eo/issues/2503
```
volodya-lombrozo, line 31: @maxonfjvipon I tried to find the reason for the `move-versions-deeper.xsl` transformation in the [issue](https://github.com/objectionary/eo/issues/2503) that you linked and I've found only a high-level description of the problem, not the reason for that transformation. I still can only guess why we might need it. Moreover:
> This transformation should be applied in the end, after all other transformations are done.

doesn't explain the reason and It seems contradictory with the (4) point:

> Because of point 3 this transformation should be applied many times, until it does not affect xml anymore

volodya-lombrozo (changes requested):
@maxonfjvipon Could you please link that pr with the corresponding issue (https://github.com/objectionary/eo/issues/2503 I guess?)

eo-parser/src/main/java/org/eolang/parser/ParsingTrain.java
maxonfjvipon: @volodya-lombrozo I explained this in the description of transformation.
1. One application of transformation moves version one level deeper
2. Because of point 2 It should be applied many times until xml after applying is not changed anymore (until we get the deep)

That's why it stays at the end and is wrapped in `TrEndless`

volodya-lombrozo: @maxonfjvipon Oh, I see. It's hard.

eo-parser/src/main/java/org/eolang/parser/XeListener.java
maxonfjvipon: @volodya-lombrozo I guess so. For example in `DiscoverMojo` we have the next code:
```java
private Collection<String> names(final XML xml, final String tojo) {
        return new SetOf<>(
            new Mapped<>(
                (String name) -> this.versioned(name, tojo).toString(),
                new Filtered<>(
                    name -> !name.isEmpty(),
                    xml.xpath(
                        String.join(
                            "",
                            "//o[",
                            "not(starts-with(@base,'.'))",
                            " and @base != 'Q'",
                            " and @base != '^'",
                            " and @base != '$'",
                            " and @base != '&'",
                            " and not(@ref)",
                            "]/string-join((@base,@ver),'",
                            OnReplaced.DELIMITER,
                            "')"
                        )
                    )
                )
            )
        );
    }
```
It does not check `withVersions` flag and joins `base` and `ver`. If there will be many empty `ver` attributes, there will be troubles.

eo-parser/src/main/resources/org/eolang/parser/move-versions-deeper.xsl
maxonfjvipon: @volodya-lombrozo I extended the doc, please have a look

maxonfjvipon:

volodya-lombrozo (changes requested):

eo-parser/src/main/java/org/eolang/parser/XeListener.java
maxonfjvipon: @volodya-lombrozo so you suggest replace one set of changes, which:
- is safer
- is more controllable (because it's just a boolean variable)
- lets us do not have extra attributes where they are not needed
with other set of changes, that works for now, but uncontrollable and will lead to the cases, where we will have eo code without any hint to the versions, but there will be some confusing emtpy "ver" attributes in xmir.

maxonfjvipon: @volodya-lombrozo if you personally don't like `oprop` method, I can remove it and replace its occurrences with its content in `XeListener`

eo-parser/src/main/resources/org/eolang/parser/move-versions-deeper.xsl
```diff
@@ -0,0 +1,88 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" id="move-versions-deeper" version="2.0">
+  <!--
+  Here we move "ver" attribute from the top object deeper to pre-last one (before <o base="Q"/>)
+
+  1. The reason of transformation - move version to pre-last object ("org"). Then on "transpile"
```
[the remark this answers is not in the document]
maxonfjvipon: @volodya-lombrozo I believe it would be harder and slower to implement, because we will have to manipulate what is not yet in any form, even if not quite right. 
When you don't know for sure how this whole feature should work, you can't just come and say "we need this here, this there, and this out there". You need to start with something small, maybe not quite elegant, just to have something in front of you so you can work with it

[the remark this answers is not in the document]
volodya-lombrozo: @maxonfjvipon Exactly:
> You need to start with something small, maybe not quite elegant, just to have something in front of you so you can work with it

But you don't start your work with something random. The first thing you usually do is answer the question, "what do I want to achieve?" and then draw a skeleton. After that, you start adding "meat" to that skeleton. When we have a skeleton and an endpoint, we don't ask questions like "why do we need that change?" because it's obvious.
Otherwise, you'll find yourself in the situation we're currently facing: no clear reason for the change and no understanding of the final goal we aim to achieve. The next day, any developer could easily remove those changes, and they'd be justified in doing so.

Suddenly, I didn't find a good article how to do so, but may be [this](https://www.yegor256.com/2010/03/04/pdd.html) blog post might explain the idea a bit:

>  Let’s keep all these unknowns in mind as we try to solve the problem on the highest level of abstraction. Of course, we start with a test

eo-parser/src/main/java/org/eolang/parser/XeListener.java
volodya-lombrozo: @maxonfjvipon 
1. Which type of 'safety' are you referring to? I run the tests and showed that everything works [fine](https://github.com/objectionary/eo/pull/2501/files#r1332960989).
2. Why do we need the control here? Moreover, as I already mentioned, the solution with boolean values creates even more problems (like [that](https://www.yegor256.com/2016/04/19/object-must-not-be-configurable.html) and [OCP](https://en.wikipedia.org/wiki/Open–closed_principle) violation).
3. The current implementation of `XeListener` in that PR does exactly what you mentioned: it adds empty "ver" attributes.

Issue #1469, opened 2022-11-18
Empty lines inside object body must be prohibited


yegor256:
This syntax is invalid (because of the empty line inside object body), but it does parse well (I'm adding a file to `eo-parser/src/test/resources/org/eolang/parser/typos` to check):

```
[] > hello
  one
  two

  three
```

Let's make sure empty lines are not allowed inside object body.

Graur:
@EugeneDar Maybe you can help?

EugeneDar:
@Graur Sure, I'll try to help.

Graur:
@maxonfjvipon can you help us with this one?

maxonfjvipon:
@yegor256 why the code that you gave as an example is invalid? What's wrong with empty line between objects?

yegor256:
@maxonfjvipon every object formation must be "solid", containing no empty lines inside

PR #2519, opened 2023-10-03
fix(#1469): remove empty lines in runtime

maxonfjvipon:
Ref: #1469 

Removed unnecessary empty lines in eo-runtime. Reason for that is simple: integration tests in `eo-maven-plugin` take last available version of `eo-runtime` stored in objectionary. 
So changes in `eo-runtime` should be made and released firstly, before changes in grammar


yegor256:
@rultor merge

PR #2558, opened 2023-10-26
fix(#2557): runtime without varargs

maxonfjvipon:
Closes: #2557 

---

## PR-Codex overview

> The following files were skipped due to too many changes: `eo-runtime/src/test/eo/org/eolang/float-tests.eo`, `eo-runtime/src/main/java/EOorg/EOeolang/EOint$EOdiv.java`, `eo-runtime/src/test/eo/org/eolang/bool-tests.eo`, `eo-runtime/src/test/eo/org/eolang/int-tests.eo`, `eo-runtime/src/main/eo/org/eolang/positive-infinity.eo`, `eo-runtime/src/main/eo/org/eolang/negative-infinity.eo`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #2611, opened 2023-11-21
#2558: Dataized::take returns byte array

levBagryansky:
Closes #2558

---

### Detailed summary
- The signature of the `take()` method in the `Dataized` class has been changed to return a `byte[]` instead of an `Object`.
- Added a check to ensure that the returned data is always a `byte[]` and throw an exception if it's not.
- Fixed a type casting issue in the `take()` method when calling it with a specific type.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Dataized.java
+++ b/eo-runtime/src/main/java/org/eolang/Dataized.java
@@ -109,7 +109,7 @@ public final class Dataized {
                 }
             }
             final Object data = Data.class.cast(src).take();
-            if (! (data instanceof byte[])) {
+            if (!(data instanceof byte[])) {
                 throw new ExFailure(
                     "data of must be %s, but was %s",
                     this.phi.toString(),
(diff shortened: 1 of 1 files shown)


levBagryansky:
@maxonfjvipon please check

eo-runtime/src/main/java/org/eolang/Dataized.java
```diff
@@ -113,6 +109,14 @@ public Object take() {
                 }
             }
             final Object data = Data.class.cast(src).take();
+            if (!(data instanceof byte[])) {
+                throw new ExFailure(
+                    "data of must be %s, but was %s",
```
maxonfjvipon, line 114: @levBagryansky you forgot one "%s" here

maxonfjvipon (changes requested):

eo-runtime/src/main/java/org/eolang/Dataized.java
levBagryansky: @maxonfjvipon thanks! Done.

maxonfjvipon (approved):
@levBagryansky LGTM! Thanks

levBagryansky:
@yegor256 please check

levBagryansky:
@yegor256 please take a look

yegor256:
@rultor merge

Issue #2612, opened 2023-11-21
Get rid of varargs at all


maxonfjvipon:
The hard decision was made: vararg was a big mistake. So we need to get rid of it in eo at all. Only tuples...
It will make EO stronger and stricter. Also there'll be no trouble with translating EO to phi calculus expession

0pdd:
@maxonfjvipon the puzzle [#2622](https://github.com/objectionary/eo/issues/2622) is still not solved.

0pdd:
@maxonfjvipon the only puzzle [#2622](https://github.com/objectionary/eo/issues/2622) is solved here.

PR #2620, opened 2023-11-22
feat(#2612): remove varargs in `eo-runtime`

maxonfjvipon:
Ref: #2612


maxonfjvipon:
@yegor256 please have a look

PR #2776, opened 2024-01-14
#2772: PhContainingUTF8 to print strings correctly

### LINKED ISSUE #2772: Unreadable exception  Ex EOorg.EOeolang.EOstringν62
The following stack trace does not help to find out the mistake in eo code:
```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.154 s <<< FAILURE! -- in EOorg.EOeolang.EOsqrt_trashTest
[ERROR] EOorg.EOeolang.EOsqrt_trashTest.works -- Time elapsed: 0.087 s <<< ERROR!
EOorg.EOeolang.EOerror$ExError: EOorg.EOeolang.EOstringν62
        at org.eolang.PhSafe.attr(PhSafe.java:92)
        at org.eolang.PhImmovable.attr(PhImmovable.java:81)
        at org.eolang.PhOnce.attr(PhOnce.java:92)
        at org.eolang.PhOnce.attr(PhOnce.java:92)
        at org.eolang.PhLocated.attr(PhLocated.java:118)
        at org.eolang.PhNamed.attr(PhNamed.java:88)
        at org.eolang.PhSafe.attr(PhSafe.java:90)
        at org.eolang.PhImmovable.attr(PhImmovable.java:81)
        at org.eolang.PhOnce.attr(PhOnce.java:92)
        at org.eolang.PhLocated.attr(PhLocated.java:118)
        at org.eolang.PhMethod.lambda$new$0(PhMethod.java:43)
        at org.eolang.Data$Once.lambda$take$0(Data.java:120)

```

levBagryansky:
Closes #2772 

Such eo code like
```
[] > my-error
  error > @
    "qwerty"
```
failed with unclear message since `PhDefault::toString()` implementation. So `PhContainingUTF8` was introduced.

---

## PR-Codex overview
This PR focuses on adding unit tests for the `EOerror` class and improving the error message handling. 

### Detailed summary
- Added unit tests for `EOerror` class.
- Imported necessary dependencies for unit tests.
- Implemented `getsReadableError` test method.
- Added `getTestSources` method to provide input arguments for the test.
- Added `MyError` class to create a custom error object for testing.
- Improved error message handling in the `ExError` class.
- Added `safeMessage` method to safely retrieve the message from the enclosure.


levBagryansky:
@maxonfjvipon please take a look

eo-runtime/src/test/java/EOorg/EOeolang/EOerrorTest.java
```diff
@@ -55,4 +63,64 @@ void makesToxicObject() {
...
+                            "α",
+                            new PhWith(
+                                new PhCopy(
+                                    Phi.Φ
+                                        .attr("org").get()
+                                        .attr("eolang").get()
+                                        .attr("string").get()
+                                ),
+                                0,
+                                new PhWith(
```
maxonfjvipon, line 113: @levBagryansky let's replace this with just `new Data.ToPhi("qwerty")`

eo-runtime/src/test/java/EOorg/EOeolang/EOerrorTest.java
```diff
@@ -55,4 +63,64 @@ void makesToxicObject() {
         );
     }
 
+    @Test
+    void getsReadableError() {
+        ExAbstract error = null;
```
maxonfjvipon, line 68: @levBagryansky can we make this variable final? 
For example set it to `null` in `try` block and set to `exc` in `catch` block?

maxonfjvipon (changes requested):
@levBagryansky good one, just several comments from my side

maxonfjvipon:
@levBagryansky one more question: do we have a case where object inside `ExError` can't be dataized at all?

eo-runtime/src/test/java/EOorg/EOeolang/EOerrorTest.java
levBagryansky: @maxonfjvipon We cannot do it here, I get the error `Variable 'error' might already have been assigned to: 73`

levBagryansky:
@maxonfjvipon this is a good question..

levBagryansky:
@maxonfjvipon please check

maxonfjvipon (approved):
@levBagryansky LGTM! Thanks

levBagryansky:
@yegor256 please take a look

eo-runtime/src/main/java/EOorg/EOeolang/EOerror.java
```diff
@@ -126,5 +127,29 @@ public ExError(final Phi enclosure) {
...
+         * @return String message.
+         * @checkstyle IllegalCatchCheck (20 lines)
+         */
+        private static String safeMessage(final Phi enclosure) {
+            String result;
+            if (enclosure == null) {
+                result = "null Phi";
+            } else {
+                try {
+                    result = new PhContainingUTF8(enclosure).toString();
```
yegor256, line 143: @levBagryansky if this class is an instance of `Phi` _and_ it makes phi objects safer by properly printing their `toString()`, why do we use only here, in this private method? Why don't we decorate _all_ our phi objects with this decorator?

levBagryansky: @yegor256 This class is correct to use in order to print bytes of `data` as UTF8 String. So `PhContainingUTF8(EOint)` produces something strange. We can also use this class in `Data.ToPhi(java.util.String)`

yegor256: @levBagryansky maybe we shouldn't call this class `Phi...` and should not make it a decorator of `Phi`? It doesn't look like a decorator to me, but a "smart" class: https://www.yegor256.com/2016/04/26/why-inputstream-design-is-wrong.html It just takes an existing instance of `Phi`, calls its method and then post-process the returned data. It doesn't _become_ phi, it only is a client of Phi.

levBagryansky:
@yegor256 , `PhDefault::toString` has the following implementation:
```
    @Override
    public String toString() {
        String result = String.format(
            "%sν%d",
            this.getClass().getCanonicalName(),
            this.vertex
        );
        if (this.attrs.containsKey("Δ")) {
            result = String.format(
                "%s=%s",
                result,
                this.attrs.get("Δ").toString()
            );
        }
        return result;
    }
```
So `EOstring::toString` does not print its content because `EOstring` has `φ` attribute with `EObytes` inside, not `Δ`. But we need to print Phi objets correctly somehow for readable exceptions. Ideally, making it correctly in cases of `EOint`, `EOstrings`. `EOfloat` and so on. 
May be we will create such decorators and wrap to them in `to-java.xsl`?

Issue #2778, opened 2024-01-15
broken data representation in PhiMojo


yegor256:
This is my XMIR:

```
<program>
   <objects>
      <o abstract="" name="j$A">
         <o base="int" data="bytes" name="access">00 00 00 00 00 00 00 20</o>
         <o base="string" data="bytes" name="supername">6A 61 76 61 2F 6C 61 6E 67 2F 4F 62 6A 65 63 74</o>
         <o base="tuple" data="tuple" name="interfaces">
            <o base="string" data="bytes">6F 72 67 2F 65 6F 6C 61 6E 67 2F 62 65 6E 63 68 6D 61 72 6B 2F 46</o>
         </o>
   </objects>
</program>
```

I'm getting:

```
{org ↦ ⟦eolang ↦ ⟦benchmark ↦ ⟦j$A ↦ ⟦access ↦ ξ.int(Δ ⤍ 00-00-00-00-00-00-00-20), 
supername ↦ ξ.string(Δ ⤍ 6A-61-76-61-2F-6C-61-6E-67-2F-4F-62-6A-65-63-74), 
interfaces ↦ ξ.tuple(α0 ↦ ξ.string(Δ ⤍ 6F-72-67-2F-65-6F-6C-61-6E-67-2F-62-65-6E-63-68-6D-61-72-6B-2F-46))(Δ ⤍
------------6F-72-67-2F-65-6F-6C-61-6E-67-2F-62-65-6E-63-68-6D-61-72-6B-2F-46
---------)
...
```

Something is wrong with the data representation.

maxonfjvipon:
@yegor256 
1. We don't have `data="tuple"` at all
2. The ticket needs #2770

maxonfjvipon:
@yegor256 let's make a new release with tag `0.34.4`

yegor256:
@rultor release, tag is `0.34.4`

maxonfjvipon:
@yegor256 please update EO version and let me know if issue is still there

yegor256:
@maxonfjvipon now, the XML produced by opeo is broken and can't be digested by the `PrintMojo`

Issue #2689, opened 2023-12-08
Prettify phi-calculus expression


maxonfjvipon:
Now phi-calculus expression is printed in one line which makes it hard to read. It would be nice to make it more vertical with indentations, like in [this blog post](https://news.eolang.org/2023-12-08-phi-and-unphi-mojos.html)

PR #2789, opened 2024-01-17
fix(#2778): PhiMojo that optimizes and checks schema + pretty phi expression

maxonfjvipon:
Ref: #2778
Ref: #2689 

What's done:
- Pretty print of phi calculus expression
- `PhiMojo` uses `ParsingTrain` to optimize XMIR before converting to phi
- `PhiMojo` checks schema of XMIR before converting to phi
- XSD schema was extended to match with current optimized XMIR


yegor256:
@rultor merge

Issue #2790, opened 2024-01-17
OptCached.java:44-51: Fix caching mechanism in {@link...


0pdd:
The puzzle `2746-9394bf60` from #2746 has to be resolved: 

https://github.com/objectionary/eo/blob/036be9f42fa55cf7015dbc66068cccfeb055a93f/eo-maven-plugin/src/main/java/org/eolang/maven/optimization/OptCached.java#L44-L51

The puzzle was created by rultor on 17-Jan-24. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

Yanich96:
@maxonfjvipon assign me please

0pdd:
@0pdd 2 puzzles [#2820](https://github.com/objectionary/eo/issues/2820), [#2821](https://github.com/objectionary/eo/issues/2821) are still not solved.

0pdd:
@0pdd all 2 puzzles are solved here: [#2820](https://github.com/objectionary/eo/issues/2820), [#2821](https://github.com/objectionary/eo/issues/2821).

Issue #2713, opened 2023-12-22
OptCached doesn't take into account the date of the source


yegor256:
At the moment, `OptCached` accepts an XML document and returns its optimized version. It also checks, whether an already optimized XML exists in the cache. However, it doesn't pay attention to the date of the original XML and the date of the document in the cache. This often leads to errors because the data in the cache is already expired.

I think, we should change the interface of `Optimization` from `Function<XML, XML>` to `Function<Path, XML>`. It should accept the path of the source document, not the content of it.

Also, I think it's better to use `Func` from Cactoos instead of `Function` from JDK, since it allows checked exceptions to be thrown from the `apply` method.

yegor256:
@volodya-lombrozo can you please fix this? it's pretty urgent, since it blocks this: https://github.com/objectionary/eo/pull/2711

volodya-lombrozo:
@yegor256 Yes, I'll

yegor256:
@volodya-lombrozo it doesn't block #2711 anymore, but still would be nice to fix, since the entire caching mechanism is kind of broken now

PR #2808, opened 2024-01-22
Fix/#2790/fix caching in optcahed

Yanich96:
Closed: #2790

What's done:

The caching algorithm was fixed: now last modifications time of cached file must be later or equal than last modifications time of program file.
The caching process can be faster if the xml file is not read. A task was added so that the file name is not taken from the file attributes.

---

## PR-Codex overview
This PR focuses on optimizing the caching mechanism in the `OptCached` class. 

### Detailed summary
- Refactors the caching mechanism in `OptCached`
- Unifies caching mechanism across different stages
- Adds improvements and fixes to the caching logic
- Removes duplication in cache-related tests

> The following files were skipped due to too many changes: `eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java`


Yanich96:
@volodya-lombrozo check please

eo-maven-plugin/src/main/java/org/eolang/maven/optimization/Optimization.java
```diff
@@ -32,5 +34,20 @@
...
 @FunctionalInterface
-public interface Optimization extends Function<XML, XML> {
+public interface Optimization extends ThrowingFunction<Path, XML, FileNotFoundException> {
+
+    /**
+     * Return function if it works without exception.
+     *
+     * @return The function result
+     */
+    default Function<Path, XML> unchecked() {
```
volodya-lombrozo, line 44: @Yanich96 I'm afraid it might be dangerous for the code maintainability. 

From [this](https://www.yegor256.com/2017/10/03/java9.html):
> [Default methods](https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html) were also added to interfaces, which turned types into libraries of code. Types into libraries! It’s even worse than [implementation inheritance](https://www.yegor256.com/2016/09/13/inheritance-is-procedural.html), if you ask me.

Moreover, "default" methods are similar to mixins in other languages (please, correct me if I'm wrong), which also isn't a good feature for maintainability and OOP design. You can read more about it [here](https://www.yegor256.com/2017/03/07/traits-and-mixins.html).

Maybe we can remove this "default" behaviour somehow? What do you think?

eo-maven-plugin/src/main/java/org/eolang/maven/optimization/ThrowingFunction.java
```diff
@@ -0,0 +1,43 @@
...
+@FunctionalInterface
+public interface ThrowingFunction<I, R, E extends Exception> {
+    /**
+     * Applies this function to the given argument.
+     *
+     * @param input The function argument
+     * @return The function result
+     * @throws E If fails
+     */
+    R apply(I input) throws E;
```
volodya-lombrozo, line 42: @Yanich96 I see two problems here:
_____ 
I believe that we don't need the `ThrowingFunction` interface at all because it's too generic and used only for optimization classes. Since we already have the `Optimization` abstraction, we don't need one more layer. It looks like overengineering. Please take a look at the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle).
____

## (II) Changing `apply` method signature.
As I understood, the solution you provided was inspired by this [comment](https://github.com/objectionary/eo/issues/2713#issue-2053369477): 

> I think, we should change the interface of Optimization from Function<XML, XML> to Function<Path, XML>. It should accept the path of the source document, not the content of it. 

Well, I didn't write about it previously, so I will do it now. Changing the interface from `XML apply(XML)` to `XML apply(Path)` is dangerous for maintainability. At least, the solution overcomplicates the implementation of each separate optimization, as now each class has to "read" `XML` from the filesystem itself. I've counted four new usages of the `new XMLDocument(path)` instruction. Moreover, ideally, each `Optimization` has to save its results to a filesystem. Have you already counted how much it decreases the compilation time? I bet ~ `x2`.

Put it simply: 
We had 
```
XMIR -> (optimization) -> XMIR -> (optimization) -> XMIR
```

Now we have :
```
File -> XMIR -> (optimization) -> XMIR -> File -> XMIR -> (optimization) -> XMIR -> File
```
Looks overcomplicated, doesn't it?

volodya-lombrozo (changes requested):

eo-maven-plugin/src/main/java/org/eolang/maven/optimization/ThrowingFunction.java
Yanich96: @volodya-lombrozo This task is fix OptCached and OptCachedTest. 

I deliberately left reading the name from the file so as not to break the program logic and tests. I created the issue in this PR in which I want to remove reading from a file. After removing reading from the file, the added interface and execution will not be needed.

volodya-lombrozo: @Yanich96, can we implement this directly without adding any potentially dangerous code? It appears that we may not need this code at all. Why introduce this intermediate step?

volodya-lombrozo (changes requested):

eo-maven-plugin/src/main/java/org/eolang/maven/optimization/ThrowingFunction.java
Yanich96: @volodya-lombrozo 
At the moment, I don't know how to avoid this code. But this code occurs when reading a XML file.
In my current task, only its name is taken from the XML file. The next step is that I want to get the program name from the absolute path, but I need to understand how the absolute paths of the program itself and its cache are obtained. In the next task the dangerous code will be removed.

eo-maven-plugin/src/main/java/org/eolang/maven/optimization/ThrowingFunction.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Abstraction for XML optimizations.
+ * @param <I> Input type.
+ * @param <R> Return type.
+ * @param <E> Exception type.
+ * @since 0.35.0
+ */
+@FunctionalInterface
+public interface ThrowingFunction<I, R, E extends Exception> {
```
maxonfjvipon, line 34: @Yanich96 I would notice that there's `Func` interface in `cactoos` which `apply` method throws `Exception`. So it can be used instead

eo-maven-plugin/src/main/java/org/eolang/maven/optimization/ThrowingFunction.java
volodya-lombrozo: @Yanich96 Sorry, maybe I don't clearly understand you. Why do you need to change `XML apply(XML)` to `XML apply(Path)`?

Yanich96: @volodya-lombrozo 
1) Only using the path can you find out the time of the last modification of the XML file.
2) Using the path allows us this allows us to not read the XML file. This will speed up the compiler.

volodya-lombrozo: @Yanich96, maybe we can save the 'last modification time' to XMIR and simply utilize it?

As I've already mentioned, the solution significantly slows down the compiler due to unnecessary read/write operations from the file system. Consequently, I don't understand which speed up you mean.

Yanich96:
@volodya-lombrozo check please

eo-maven-plugin/src/main/java/org/eolang/maven/optimization/Optimization.java
```diff
@@ -24,13 +24,14 @@
...
+import java.nio.file.Path;
+import org.cactoos.Func;
 
 /**
  * Abstraction for XML optimizations.
  *
  * @since 0.28.11
  */
 @FunctionalInterface
-public interface Optimization extends Function<XML, XML> {
```
volodya-lombrozo, line 35: @Yanich96 I don't understand why we have to change this to `Path`. Can we just save `lastModifiedTime` to `XMIR` and then just read it?

volodya-lombrozo (changes requested):

Yanich96:
@volodya-lombrozo I have fixed interface Optimization. Check please

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -162,6 +196,19 @@ private static Path save(final Path tmp, final XML xml) throws IOException {
         return res;
     }
 
+    /**
+     * Set time of XML program use absolute path of program.
+     * @param path Temporary test directory.
+     * @param time XML program.
+     * @return Path to saved program.
+     */
+    private static void setTime(
```
volodya-lombrozo, line 205: @Yanich96 Can we combine saving file and setting `lastModifierTime` in a single method? Seems, they usually go altogether.
Moreover, the name of the method is a bit confusing. Please, take a look this posts:
- https://www.yegor256.com/2018/08/22/builders-and-manipulators.html
- https://www.yegor256.com/2015/01/12/compound-name-is-code-smell.html

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -126,26 +155,31 @@ void optimizesIfXmlIsAbsentInCache(@TempDir final Path tmp) {
...
-            "We expected that the program will be optimized because the cache doesn't have time",
-            new OptCached(path -> with, tmp).apply(without),
-            Matchers.equalTo(with)
+            "We expected that the program will be optimized because the cache is expired",
+            new OptCached(
+                path -> OptCachedTest.program("new program"),
+                cache,
+                program
+            )
+                .apply(OptCachedTest.program("new program")),
```
volodya-lombrozo, line 181: @Yanich96 Do you need to put it on a separate line?

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -126,26 +155,31 @@ void optimizesIfXmlIsAbsentInCache(@TempDir final Path tmp) {
...
-        final XML without = OptCachedTest.program();
-        final XML with = OptCachedTest.program(ZonedDateTime.now());
-        OptCachedTest.save(tmp, without);
         MatcherAssert.assertThat(
-            "We expected that the program will be optimized because the cache doesn't have time",
-            new OptCached(path -> with, tmp).apply(without),
-            Matchers.equalTo(with)
+            "We expected that the program will be optimized because the cache is expired",
+            new OptCached(
+                path -> OptCachedTest.program("new program"),
```
volodya-lombrozo, line 177: @Yanich96 You rather often use "new program" in tests. Maybe it's better to create one more overloaded method `program()` without arguments which will call `program("new program")`?

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -50,73 +49,103 @@ final class OptCachedTest {
...
+    void returnsFromCacheButTimesSaveAndExecuteDifferent(
+        @TempDir final Path cache,
+        @TempDir final Path dir
+    )
+        throws Exception {
+        final XML xml = OptCachedTest.program();
+        final Path program = OptCachedTest.save(dir, xml);
+        OptCachedTest.setTime(
+            dir,
+            FileTime.fromMillis(System.currentTimeMillis())
```
volodya-lombrozo, line 92: @Yanich96 Maybe it's better to call `FileTime.fromMillis(...)` inside the `setTime()` method? And pass just milliseconds?

eo-maven-plugin/src/main/java/org/eolang/maven/optimization/OptSpy.java
```diff
@@ -67,8 +67,12 @@ public OptSpy(final Train<Shift> trn, final Path target) {
...
     @Override
     public XML apply(final XML xml) {
-        final Place place = new Place(xml.xpath("/program/@name").get(0));
-        final Path dir = place.make(this.target, "");
+        final Path dir = new Place(
+            xml
+                .xpath("/program/@name")
+                .get(0)
+        )
+            .make(this.target, "");
```
volodya-lombrozo, line 75: @Yanich96 Do you need a separate line for this statement? 
Can we just write the following in one line:
```java
 final Path dir = new Place(xml .xpath("/program/@name").get(0)).make(this.target, "");
```
or in two lines:
```java
 final Path dir = new Place(xml .xpath("/program/@name").get(0))
   .make(this.target, "");
```

volodya-lombrozo (changes requested):

Yanich96:
@volodya-lombrozo I fixed OptCachedTest

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
Yanich96: @volodya-lombrozo I can't combine saving file and setting lastModifierTime in a single method, because method save() gets relative path and method setLastModifiedTime() gets absolutely path

volodya-lombrozo: @Yanich96 The method `save` returns absolute path on that you can set last modification time

Yanich96:
@volodya-lombrozo check please

eo-maven-plugin/src/test/java/org/eolang/maven/OptimizeMojoTest.java
```diff
@@ -132,6 +132,15 @@ void getsAlreadyOptimizedResultsFromCache(@TempDir final Path temp) throws Excep
                 .resolve(hash)
                 .resolve("foo/x/main.xmir")
         );
+        Files.setLastModifiedTime(
```
volodya-lombrozo, line 135: @Yanich96 Can we use `save()` function instead?

eo-maven-plugin/src/main/java/org/eolang/maven/OptimizationTask.java
```diff
@@ -144,15 +144,18 @@ private Path make(final XML xml, final Path file) throws IOException {
      * @param tojo Tojo
      * @param common Optimization
      * @return Optimization for specific Tojo
+     * @throws Exception If fails
      */
-    private Optimization optimization(final ForeignTojo tojo, final Optimization common) {
+    private Optimization optimization(final ForeignTojo tojo, final Optimization common)
```
volodya-lombrozo, line 149: @Yanich96 Let's now take a look on the pseudo logic of `OptimizationTask` and `OptCached` classes and how they work:
```java
//OptimizationTask start
if(tojo.hasHash()){
   //OptCached internals start
   if(contains(in_cache)){
      return optimization from cache
   } else {
      run "common" optimization
   }
  //OptCacjed internals end
} else {
   run "common" optimization 
}
//OptimizationTask end
```
What if we simplify it to a something like the following:

```java
if(tojo.hasHash() &&  contains(in_cache)) {
      return optimization from cache
} else {
   run "common" optimization 
}
```

What do you think?

eo-maven-plugin/src/test/java/org/eolang/maven/ShakeMojoTest.java
```diff
@@ -96,6 +96,10 @@ void getsAlreadyShakenResultsFromCache(@TempDir final Path temp) throws Exceptio
                 .resolve(hash)
                 .resolve("foo/x/main.xmir")
         );
+        Files.setLastModifiedTime(
```
volodya-lombrozo, line 99: @Yanich96 Can we use `save` function instead?

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -50,73 +49,81 @@ final class OptCachedTest {
...
      */
-    @Disabled
     @Test
-    void returnsFromCacheIfXmlAlreadyInCache(@TempDir final Path tmp) throws IOException {
-        final XML program = OptCachedTest.program(ZonedDateTime.now());
-        OptCachedTest.save(tmp, program);
+    void returnsFromCacheIfXmlAlreadyInCache(@TempDir final Path cache, @TempDir final Path dir)
+        throws Exception {
+        final XML xml = OptCachedTest.program();
+        final Path program = OptCachedTest.save(dir, xml, 0);
```
volodya-lombrozo, line 64: @Yanich96 [Optional] You are using `OptCachedTest.save(dir, xml, 0)`  in many places. Maybe we can overload this method and allow to call just `OptCachedTest.save(dir, xml)` and pass `0` by default? What do you think?

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -50,73 +49,81 @@ final class OptCachedTest {
...
         );
     }
 
-    @Disabled
     @Test
-    void returnsFromCacheButTimesSaveAndExecuteDifferent(@TempDir final Path tmp)
-        throws IOException {
-        final XML program = OptCachedTest.program(ZonedDateTime.now().minusMinutes(2));
-        OptCachedTest.save(tmp, program);
+    void returnsFromCacheButTimesSaveAndExecuteDifferent(
```
volodya-lombrozo, line 80: @Yanich96 What is "save time" and "execute time"? As I understand "save time" is a time when we saved a file, but what about "execute time"? Is it "last modified time"?

volodya-lombrozo (changes requested):

eo-maven-plugin/src/test/java/org/eolang/maven/OptimizeMojoTest.java
Yanich96: @volodya-lombrozo Is it necessary to create a new function here just for one use?

eo-maven-plugin/src/test/java/org/eolang/maven/ShakeMojoTest.java
Yanich96: @volodya-lombrozo I will create issue

eo-maven-plugin/src/main/java/org/eolang/maven/OptimizationTask.java
Yanich96: @volodya-lombrozo 
I will create issue

eo-maven-plugin/src/test/java/org/eolang/maven/OptimizeMojoTest.java
Yanich96: @volodya-lombrozo I will create issue

Yanich96:
@volodya-lombrozo check please

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -50,73 +49,92 @@ final class OptCachedTest {
...
+        final Path program = OptCachedTest.save(dir, OptCachedTest.program());
         MatcherAssert.assertThat(
             "We expect that the program will be created and returned as is (same instance)",
-            new OptCached(path -> program, cache).apply(program),
-            Matchers.sameInstance(program)
+            new OptCached(
+                path -> OptCachedTest.program(), cache, program
+            )
+                .apply(OptCachedTest.program()),
+            Matchers.equalTo(OptCachedTest.program())
```
volodya-lombrozo, line 137: @Yanich96 Can we just create a sing variable `prog = OptCachedTest.program()` and use it in all four places?

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -50,73 +49,92 @@ final class OptCachedTest {
...
         );
     }
 
     @Test
-    void optimizesIfXmlIsAbsentInCache(@TempDir final Path tmp) {
-        final XML program = OptCachedTest.program();
-        final Path cache = tmp.resolve("cache");
+    void optimizesIfXmlIsAbsentInCache(@TempDir final Path cache, @TempDir final Path dir)
+        throws Exception {
+        final Path program = OptCachedTest.save(dir, OptCachedTest.program());
```
volodya-lombrozo, line 130: @Yanich96 Can you inline this variable, please?
https://www.yegor256.com/2015/09/01/redundant-variables-are-evil.html

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -50,73 +49,92 @@ final class OptCachedTest {
...
-    @Disabled
     @Test
-    void returnsFromCacheCorrectProgram(@TempDir final Path tmp)
-        throws IOException {
-        final XML prev = OptCachedTest.program(ZonedDateTime.now(), "first program");
-        OptCachedTest.save(tmp, prev);
-        final XML current = OptCachedTest.program(ZonedDateTime.now(), "second program");
+    void returnsFromCacheCorrectProgram(@TempDir final Path cache, @TempDir final Path dir)
+        throws Exception {
+        final XML first = OptCachedTest.program("first program");
```
volodya-lombrozo, line 115: @Yanich96 Can you inline this variable, please?
https://www.yegor256.com/2015/09/01/redundant-variables-are-evil.html

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -50,73 +49,92 @@ final class OptCachedTest {
...
-        throws IOException {
-        final XML program = OptCachedTest.program(ZonedDateTime.now().minusMinutes(2));
-        OptCachedTest.save(tmp, program);
+    void returnsFromCacheButLastModifiedTimesDifferent(
+        @TempDir final Path cache,
+        @TempDir final Path dir
+    )
+        throws Exception {
+        final XML xml = OptCachedTest.program();
+        final Path program = OptCachedTest.save(dir, xml);
```
volodya-lombrozo, line 97: @Yanich96 Can you inline this variable, please?
https://www.yegor256.com/2015/09/01/redundant-variables-are-evil.html

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -50,73 +49,92 @@ final class OptCachedTest {
...
      */
-    @Disabled
     @Test
-    void returnsFromCacheIfXmlAlreadyInCache(@TempDir final Path tmp) throws IOException {
-        final XML program = OptCachedTest.program(ZonedDateTime.now());
-        OptCachedTest.save(tmp, program);
+    void returnsFromCacheIfXmlAlreadyInCache(@TempDir final Path cache, @TempDir final Path dir)
+        throws Exception {
+        final XML xml = OptCachedTest.program();
+        final Path program = OptCachedTest.save(dir, xml);
```
volodya-lombrozo, line 75: @Yanich96 Can you inline this variable, please?
https://www.yegor256.com/2015/09/01/redundant-variables-are-evil.html

eo-maven-plugin/src/test/java/org/eolang/maven/optimization/OptCachedTest.java
```diff
@@ -50,73 +49,92 @@ final class OptCachedTest {
...
+     *  We need to think about how to remove this duplication.
      */
-    @Disabled
     @Test
-    void returnsFromCacheIfXmlAlreadyInCache(@TempDir final Path tmp) throws IOException {
-        final XML program = OptCachedTest.program(ZonedDateTime.now());
-        OptCachedTest.save(tmp, program);
+    void returnsFromCacheIfXmlAlreadyInCache(@TempDir final Path cache, @TempDir final Path dir)
+        throws Exception {
+        final XML xml = OptCachedTest.program();
```
volodya-lombrozo, line 74: @Yanich96, can we save different XML content to the `cache` and `dir` folders? However, the corresponding files must have the same "last modified" time.

Let me explain the reason. As a reader, I see that you have two files in the `dir` and `cache` folders with the same content. Later, you simply compare this content with the result from `OptCached`. So, we can't be sure from which place `OptCache` returns the result – it might be the cache or just the directory. Since the content is the same, we can't be certain.

Yanich96:
@volodya-lombrozo check please

volodya-lombrozo (approved):
@Yanich96 Looks good to me. Thank you. Good job.

volodya-lombrozo:
@yegor256 Could you merge this changes, please?

Yanich96:
@yegor256 Could you merge, please?

Yanich96:
@yegor256 Could you merge, please?

yegor256:
@rultor merge

Yanich96:
@yegor256 something is wrong with rultor. Check please.

yegor256:
@rultor merge

PR #2809, opened 2024-01-22
Add tests for 'while' with iteration over a tuple.

### LINKED ISSUE #2758: The new representation of EO breaks all dependent components.
I used to get rather good EO representation of `XMIR`:
```eo
seq > @
  tuple
    $
    .a
    .foo
    .plus
      1
```

With the `0.34.2` version of `eo-parser` I get the following output:
```
seq > @
      tuple
        plus.
          foo.
            a.
              $
          1
```
Which is extremely hard to read (I have to read bottom -> top to unveil the logic.) Also we previously have understandable literals:
```
"java/lang/Object" > supername
``` 
Now it's a complete mess:
```
  """
6A 61 76 61 2F 6C 61 6E 67 2F 4F 62 6A 65 63 74
""" > supername
```
So I see two critical problems here:
1. All the modules dependent on `eo-parser` are broken. One of the use cases of the previous representation involves tests. Therefore, to upgrade to `eo-parser:0.34.2`, I must upgrade all the tests, that use `EO` representation. It is extremely tedious task.
2. The new representation is designed for programs, not humans.
____

However, I understand why we might need to modify the EO output. Thus, I suggest returning to the previous representation and adding one more representation as a separate new class. This will repair all the dependent modules and you will be able to use new representation for other needs.

c71n93:
Closes #2758

There are 2 tests that don't work, but it seems to me that they should (`iterating-tuple-with-while-using-external-iterator`, `iterating-tuple-with-while-without-body-single`). I left TODO for them.

---

## PR-Codex overview
This PR focuses on adding tests for iterating tuples with while loops using both internal and external iterators.

### Detailed summary
- Added test for iterating tuple with while loop using internal iterator.
- Added test for iterating tuple with while loop using external iterator.
- Added test for iterating tuple with while loop without body (multiple elements).
- Added test for iterating tuple with while loop without body (single element).

### DIFF
--- a/eo-runtime/src/test/eo/org/eolang/bool-tests.eo
+++ b/eo-runtime/src/test/eo/org/eolang/bool-tests.eo
@@ -133,3 +133,117 @@
             x.write (x.as-int.plus 1)
             x
     FALSE
+
+[] > iterating-tuple-with-while-using-internal-iterator
+  * 1 1 1 1 > arr
+  memory 0 > iter
+  memory 0 > acc
+  arr.length.minus 1 > max
+  seq > @
+    *
+      if.
+        max.eq 0
+        acc.write
+          acc.as-int.plus
+            arr.at 0
+        while.
+          iter.as-int.lt max
+          [i]
+            seq > @
+              *
+                acc.write
+                  acc.as-int.plus
+                    arr.at
+                      i
+                iter.write
+                  iter.as-int.plus 1
+      eq.
+        acc.as-int
+        arr.length
+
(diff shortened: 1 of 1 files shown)


c71n93:
@maxonfjvipon could you check this one, please?

maxonfjvipon (approved):
@c71n93 well done, thanks

c71n93:
@yegor256 could you check this one, please?

c71n93:
@yegor256 could you check this one, please?

yegor256:
@rultor merge

PR #2879, opened 2024-02-15
2836 cage stackoverflow

### LINKED ISSUE #2836: cage.eo:62-66: Make up with an idea how to prevent...
The puzzle `1698-c7207422` from #1698 has to be resolved: 

https://github.com/objectionary/eo/blob/bfd4e77eeef8acb49147707f0849c2c9c246abdf/eo-runtime/src/main/eo/org/eolang/cage.eo#L62-L66

The puzzle was created by rultor on 01-Feb-24. 

levBagryansky:
Closes #2836

### DIFF
--- a/eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
+++ b/eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
@@ -27,9 +27,9 @@
  */
 package EOorg.EOeolang;
 
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.Supplier;
+import java.util.HashSet;
 import java.util.Set;
+import java.util.function.Supplier;
 import org.eolang.AtAtom;
 import org.eolang.AtCage;
 import org.eolang.AtFree;
@@ -123,19 +123,18 @@ public final class EOcage extends PhDefault implements Atom {
          * Cages that are currently dataizing. If one cage is datazing and
          * it needs to be dataized inside current dataizing, the cage will be here,
          */
-        private static final Set<Integer> DATAIZING_CAGES =
-            new ConcurrentHashMap<Integer, Object>().keySet();
+        private static final Set<Integer> DATAIZING_CAGES = new HashSet<>();
 
         /**
          * Enclosure.
          */
-        final Phi enclosure;
+        private final Phi enclosure;
 
         /**
          * Vertex of cage where the {@link PhTracedEnclosure#enclosure}
          * was retrieved.
          */
-        final int cage;
+        private final int cage;
 
         /**
          * Ctor.
@@ -149,37 +148,42 @@ public final class EOcage extends PhDefault implements Atom {
 
         @Override
         public Phi copy() {
-            return new PhTracedEnclosure(this.enclosure, cage);
-        }
-
-        @Override
-        public Attr attr(int pos) {
-            return this.getAttrSafely(() -> enclosure.attr(pos));
-        }
-
-        @Override
-        public Attr attr(String name) {
-            return this.getAttrSafely(() -> enclosure.attr(name));
+            return new PhTracedEnclosure(this.enclosure, this.cage);
         }
 
         @Override
         public String locator() {
-            return enclosure.locator();
+            return this.enclosure.locator();
         }
 
         @Override
         public String forma() {
-            return enclosure.forma();
+            return this.enclosure.forma();
         }
 
         @Override
         public String φTerm() {
-            return enclosure.φTerm();
+            return this.enclosure.φTerm();
+        }
+
+        @Override
+        public Attr attr(final int pos) {
+            return this.getAttrSafely(() -> this.enclosure.attr(pos));
+        }
+
+        @Override
+        public Attr attr(final String name) {
+            return this.getAttrSafely(() -> this.enclosure.attr(name));
         }
 
         @Override
         public int hashCode() {
-            return enclosure.hashCode();
+            return this.enclosure.hashCode();
+        }
+
+        @Override
+        public boolean equals(final Object obj) {
+            return obj instanceof Phi && this.hashCode() == obj.hashCode();
         }
 
         /**
@@ -188,12 +192,12 @@ public final class EOcage extends PhDefault implements Atom {
          * @return The {@link Attr} if there is no StackOverflow case.
          */
         private Attr getAttrSafely(final Supplier<Attr> supplier) {
-            if (DATAIZING_CAGES.contains(cage)) {
+            if (EOcage.PhTracedEnclosure.DATAIZING_CAGES.contains(this.cage)) {
                 throw new ExFailure("The cage %s is already dataizing", this.cage);
             }
-            DATAIZING_CAGES.add(cage);
+            EOcage.PhTracedEnclosure.DATAIZING_CAGES.add(this.cage);
             final Attr ret = supplier.get();
-            DATAIZING_CAGES.remove(cage);
+            EOcage.PhTracedEnclosure.DATAIZING_CAGES.remove(this.cage);
             return ret;
         }
     }
--- a/eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
+++ b/eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
@@ -40,6 +40,10 @@ import org.eolang.Versionized;
 import org.eolang.Volatile;
 import org.eolang.XmirObject;
 
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+
 /**
  * CAGE.
  *
@@ -63,6 +67,7 @@ public final class EOcage extends PhDefault implements Atom {
 
     @Override
     public Phi lambda() {
+        System.out.println("EOcage::lambda");
         return new PhTracedEnclosure(this.attr("enclosure").get(), this.hashCode());
     }
 
@@ -103,6 +108,7 @@ public final class EOcage extends PhDefault implements Atom {
 
         @Override
         public Phi lambda() {
+            System.out.println("write");
             this.attr("σ").get().attr("enclosure").put(
                 this.attr("x").get()
             );
@@ -110,11 +116,10 @@ public final class EOcage extends PhDefault implements Atom {
         }
     }
 
-
-
     public static final class PhTracedEnclosure implements Phi {
         final Phi enclosure;
         final int cage;
+        private static final Set<Integer> cages_dataizing = new HashSet<>();
 
         public PhTracedEnclosure(final Phi enclosure, final int cage) {
             this.enclosure = enclosure;
@@ -123,17 +128,30 @@ public final class EOcage extends PhDefault implements Atom {
 
         @Override
         public Phi copy() {
-            return this.enclosure.copy();
+            System.out.println("PhTracedEnclosure::copy");
+            return new PhTracedEnclosure(this.enclosure, cage);
         }
 
         @Override
         public Attr attr(int pos) {
-            return new AtTracedEnclosure(enclosure.attr(pos).get(), cage);
+            return new AtTracedEnclosure(enclosure.attr(pos), cage);
         }
 
         @Override
         public Attr attr(String name) {
-            return new AtTracedEnclosure(enclosure.attr(name).get(), cage);
+            if (cages_dataizing.contains(cage)) {
+                throw new RuntimeException("the cage is already dataizing");
+                //System.out.println("ABOBA");
+            }
+            cages_dataizing.add(cage);
+            if (this.enclosure.hashCode() == this.cage) {
+                System.out.println("ABOBA");
+            }
+            System.out.printf("start PhTracedEnclosure::attr(\"%s\"), enclosure = %d, cage = %d\n", name, enclosure.hashCode(), cage);
+            final Attr ret = new AtTracedEnclosure(enclosure.attr(name), cage);
+            System.out.printf("finish PhTracedEnclosure::attr(\"%s\")\n", name);
+            cages_dataizing.remove(cage);
+            return enclosure.attr(name);
         }
 
         @Override
@@ -150,38 +168,45 @@ public final class EOcage extends PhDefault implements Atom {
         public String φTerm() {
             return this.getClass() + " -> " + enclosure.forma();
         }
+
+        @Override
+        public int hashCode() {
+            return enclosure.hashCode();
+        }
     }
 
     public static class AtTracedEnclosure implements Attr {
 
-        Phi enclosure;
+        Attr enclosure;
         int cage;
 
-        public AtTracedEnclosure(Phi enclosure, int cage) {
+        public AtTracedEnclosure(final Attr enclosure, final int cage) {
             this.enclosure = enclosure;
             this.cage = cage;
         }
 
         @Override
-        public Attr copy(Phi self) {
-            return new AtTracedEnclosure(enclosure, cage);
+        public Attr copy(final Phi self) {
+            return this.enclosure.copy(self);
         }
 
         @Override
         public Phi get() {
-            System.out.println("cage = " + cage + ", enclosure = " + this.enclosure.hashCode() + " = " + this.enclosure.forma() + ", " + this.enclosure.toString());
+            //System.out.println("cage = " + cage + ", enclosure = " + this.enclosure.hashCode() + " = " + this.enclosure.forma() + ", " + this.enclosure.toString());
             if (cage == this.enclosure.hashCode()) {
                 throw new ExFailure("Cage stackoverflow");
             }
-            if (enclosure instanceof Data) {
-                return enclosure;
+            Phi ret = enclosure.get();
+            if (!(ret instanceof Data)) {
+                ret = new PhTracedEnclosure(enclosure.get(), cage);
             }
-            return new PhTracedEnclosure(enclosure, cage);
+            return ret;
         }
 
         @Override
         public void put(Phi phi) {
-            this.enclosure = phi;
+            System.out.println("AtTracedEnclosure::put");
+            this.enclosure.put(phi);
             this.cage = 0;
         }
 
--- a/eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
+++ b/eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
@@ -33,6 +33,7 @@ import org.eolang.AtFree;
 import org.eolang.Atom;
 import org.eolang.Attr;
 import org.eolang.Data;
+import org.eolang.ExFailure;
 import org.eolang.PhDefault;
 import org.eolang.Phi;
 import org.eolang.Versionized;
@@ -62,7 +63,7 @@ public final class EOcage extends PhDefault implements Atom {
 
     @Override
     public Phi lambda() {
-        return this.attr("enclosure").get();
+        return new PhTracedEnclosure(this.attr("enclosure").get(), this.hashCode());
     }
 
     /**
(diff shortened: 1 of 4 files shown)


levBagryansky:
@maxonfjvipon could you please review

levBagryansky:
@yegor256 please check

yegor256:
@volodya-lombrozo please, review this one

eo-runtime/src/test/java/EOorg/EOeolang/EOcageTest.java
```diff
@@ -255,6 +256,16 @@ void writesBoundedCopyOfTheSameBase() {
         );
     }
 
+    @Test
+    void throwsExceptionIfRecursion() {
+        final Phi cage = new EOcage(Phi.Φ);
+        writeTo(cage, cage);
+        Assertions.assertThrows(
```
volodya-lombrozo, line 263: @levBagryansky Could you add a message to the assertion which will help to understand your expectations, please?
Why it's important: https://www.yegor256.com/2023/01/19/layout-of-tests.html#assertions

eo-runtime/src/test/java/EOorg/EOeolang/EOcageTest.java
```diff
@@ -170,9 +171,9 @@ void evaluatesLazily() {
         EOcageTest.writeTo(first, new Data.ToPhi(3L));
         final Phi second = new EOcage(Phi.Φ);
         EOcageTest.writeTo(second, new Data.ToPhi(5L));
-        final Phi cage = new EOcage(Phi.Φ);
+        final Phi sum = new EOcage(Phi.Φ);
```
volodya-lombrozo, line 174: @levBagryansky Why have you decided to rename `cage` -> `sum`? `cage` seems more logical name.

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
```diff
@@ -108,4 +112,96 @@ public Phi lambda() {
...
+        }
+
+        @Override
+        public int hashCode() {
+            return this.enclosure.hashCode();
+        }
+
+        @Override
+        public boolean equals(final Object obj) {
+            return obj instanceof Phi && this.hashCode() == obj.hashCode();
```
volodya-lombrozo, line 189: @levBagryansky [Optional]  it might be dangerous part of the code. Some objects might not be equal, but have the same `hashCode()`:
```java
System.out.println(Objects.hashCode(126L) == Objects.hashCode(4294967423L));
//prints 'true'
```
Maybe we can leave `todo` here to fix it later, what do you think?

volodya-lombrozo (changes requested):

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
```diff
@@ -108,4 +112,96 @@ public Phi lambda() {
...
+     *   should be set by user via pom.xml. We can make DATAIZING_CAGES a
+     *   ConsurrentHashMap which will add thread safety to the class.
+     */
+    private static final class PhTracedEnclosure implements Phi {
+
+        /**
+         * Cages that are currently dataizing. If one cage is datazing and
+         * it needs to be dataized inside current dataizing, the cage will be here,
+         */
+        private static final Set<Integer> DATAIZING_CAGES = new HashSet<>();
```
volodya-lombrozo, line 129: @levBagryansky Since you are using a static variable here, it might cause some problems in a multi-threaded environment. It looks like `ConcurrentHashMap` won't help you here. Perhaps [ThreadLocal](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html) is a more suitable solution here? What do you think?

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
```diff
@@ -108,4 +112,96 @@ public Phi lambda() {
             return new Data.ToPhi(true);
         }
     }
+
+    /**
+     * Class to trace if the cage got into recursion during the dataization.
+     * @since 0.36
+     * @todo #2836:60min Add a new parameter of recursion depth. This parameter
+     *   should be set by user via pom.xml. We can make DATAIZING_CAGES a
+     *   ConsurrentHashMap which will add thread safety to the class.
```
volodya-lombrozo, line 121: @levBagryansky Maybe it's better to split this `todo` into two separate puzzles:
1) The first one for "recursion depth"
2) The second one for "thread-safety"

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
levBagryansky: @volodya-lombrozo as for `Phi` we have an important condition that `Phi` with different `hashcode` are different. This is written in `Phi` documentation: https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/Phi.java#L34

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
levBagryansky: @volodya-lombrozo `ThreadLocal` here can lead to really complex scenarios. What will be with the object when one thread joins to other? Is it OK in our case?  According to documentation
 `ThreadLocal instances are typically private static fields in classes that wish to associate state with a thread` this is not our case.

eo-runtime/src/test/java/EOorg/EOeolang/EOcageTest.java
levBagryansky: @volodya-lombrozo added

eo-runtime/src/test/java/EOorg/EOeolang/EOcageTest.java
volodya-lombrozo: @levBagryansky Agree with you here. Thank you for the explanation.

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
volodya-lombrozo: @levBagryansky Well, it's a bit strange, but ok. Thank you for the explanation.

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
levBagryansky: @volodya-lombrozo done

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
volodya-lombrozo: @levBagryansky I'm not insist on `ThreadLocal` here, but `ConcurrentHashMap` seems a strange choice to me. Let's imagine we started dataization in two separate threads, in this case values in `ConcurrentHashMap` might overlap which might cause problems.  I suggest to discuss the problem in a separate `todo`.

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
```diff
@@ -108,4 +112,99 @@ public Phi lambda() {
...
+        public boolean equals(final Object obj) {
+            return obj instanceof Phi && this.hashCode() == obj.hashCode();
+        }
+
+        /**
+         * Get attribute tracing cages.
+         * @param supplier Ordinary way to get attribute.
+         * @return The {@link Attr} if there is no StackOverflow case.
+         */
+        private Attr getAttrSafely(final Supplier<Attr> supplier) {
```
volodya-lombrozo, line 200: @levBagryansky What do you think if we rename this method to `attrTraced` or `tracedAttr`?

eo-runtime/src/test/java/EOorg/EOeolang/EOcageTest.java
```diff
@@ -255,6 +256,17 @@ void writesBoundedCopyOfTheSameBase() {
...
     }
 
+    @Test
+    void throwsExceptionIfRecursion() {
+        final Phi cage = new EOcage(Phi.Φ);
+        writeTo(cage, cage);
+        Assertions.assertThrows(
+            ExAbstract.class,
+            new Dataized(cage)::take,
+            "Throws exception if recursion happened"
```
volodya-lombrozo, line 266: @levBagryansky This message will be shown if test fails. So, maybe it's better to write something like instead:
```
We except the %s exception to be thrown, but nothing happened. 
```
What do you think?

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
```diff
@@ -108,4 +112,99 @@ public Phi lambda() {
...
+     *   Map and count how many  times the cage was met.
+     * @todo #2836:60min Make the class thread safe. It has private static
+     *  field which can be accessed from differ thread and is not thread safe.
+     *  Needs to synchronize this field.
+     */
+    private static final class PhTracedEnclosure implements Phi {
+
+        /**
+         * Cages that are currently dataizing. If one cage is datazing and
+         * it needs to be dataized inside current dataizing, the cage will be here,
```
volodya-lombrozo, line 130: @levBagryansky Maybe you need a period here instead of the comma?

volodya-lombrozo (changes requested):

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
levBagryansky: @volodya-lombrozo `attrTraced` looks like we return traced attribute, but it is not. Maybe `getAttrWithTracingTheCage?`

eo-runtime/src/test/java/EOorg/EOeolang/EOcageTest.java
volodya-lombrozo: @levBagryansky Then maybe we need to add more context info to this message? Imagine a newcomer will look at your test. You can give them a clue how to understand your test.

We recently [discussed](https://github.com/volodya-lombrozo/jtcop/issues/326), that default JUnit messages might not be informative enough. Usually they are rather general.

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
volodya-lombrozo: @levBagryansky ok, maybe just `attr` then? Since the signature is different, we can easily overload this method.
I don't like `getAttrWithTracingTheCage` because of [this](https://www.yegor256.com/2015/01/12/compound-name-is-code-smell.html).

volodya-lombrozo (changes requested):

eo-runtime/src/test/java/EOorg/EOeolang/EOcageTest.java
levBagryansky: @volodya-lombrozo What would you like to add? As for me the test is already clear, commentaries are redundant.

volodya-lombrozo: @levBagryansky 
```
We expect the %s exception to be thrown since we have recursion here, but nothing happened. 
```

volodya-lombrozo: But if you believe it's redundant, I'm not insist. You might leave it as it is now.

eo-runtime/src/main/java/EOorg/EOeolang/EOcage.java
volodya-lombrozo: @levBagryansky [Here](https://www.yegor256.com/2018/08/22/builders-and-manipulators.html) you can read about methods:
> Here is a simple principle for naming methods in OOP, which I’m trying to follow in my code: it’s a verb if it manipulates, it’s a noun if it builds. That’s it. Nothing in between. Methods like saveFile() or getTitle() don’t fit and must be renamed and refactored. 

As for this:
> Ideally, our code should be understandable without documentation.

You are absolutely right.  `attr` - is a method which returns attribute, that is. A reader don't need to know "how" it was created. If I need to know "how" - I will read the method body.

levBagryansky: @volodya-lombrozo The main sense of the method that it traces. `attr` looks like it just applies its supplier inside. Actually the method does not build, it manipulates.
We use compound name for methods in this projects so it is OK. See https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/java/org/eolang/maven/PlaceMojo.java#L214. 
It would be named `satisfy` because it check if the file satisfies to condition but it would be much worse.

volodya-lombrozo: @levBagryansky 

> Moreover, methods that “manipulate” must always return void, for example print() or save().

This method returns `attr`, so it's looks like a builder. It doesn't matter "how" exactly you create it and which computations you perform. From a reader perspective it is the method that creates an object.

levBagryansky: @volodya-lombrozo see how the method is used:
```
        @Override
        public Attr attr(final int pos) {
            return this.getAttrSafely(() -> this.enclosure.attr(pos));
        }
```
`() -> this.enclosure.attr(pos)` creates attribute. The method `getAttrSafely` do it with tracing.

volodya-lombrozo: @levBagryansky 

> May be not significant how traces but significant that traces.

1. What does the `getAttrSafely` method do?
Answer: It creates an `Attr` object
2. How does the `getAttrSafely` method create an `Attr` object?
Answer: It creates `Attr` object by checking it in a trace set `DATAIZING_CAGES`.

> Here this is important that it traces.

You highlighted this importance in the class name `PhTracedEnclosure`, no need to duplicate it in the method name.

levBagryansky: @volodya-lombrozo Responsibility of this method is to trace. Creating of attribute is responsibility of `Supplier`.

levBagryansky: @yegor256 please help

levBagryansky: @volodya-lombrozo Does such method method clearer describes its responsibility?
```
        private <T> T getItWithTracing(final Supplier<T> supplier) {
            if (EOcage.PhTracedEnclosure.DATAIZING_CAGES.contains(this.cage)) {
                throw new ExFailure("The cage %s is already dataizing", this.cage);
            }
            EOcage.PhTracedEnclosure.DATAIZING_CAGES.add(this.cage);
            final T ret = supplier.get();
            EOcage.PhTracedEnclosure.DATAIZING_CAGES.remove(this.cage);
            return ret;
        }
```

volodya-lombrozo: @levBagryansky Indeed, this method clearly describes what you mean. From this method I see the following:
1. You use "getIt" to describe that you are retrieving something which you don't know (some general type, let's say.) However  n our code we already know exact types and we can use "attr" instead. In other words, "getIt" - for general method where we don't know types, "attr" - for the specific one, as it is in our case.
2. As for "withTracing" - I already answered it above:
> A reader don't need to know "how" it was created. If I need to know "how" - I will read the method body.

levBagryansky: @volodya-lombrozo reader don't need to know "how" it traces. If I need to know "how" - I will read the method body. Significant that it traces. 
It is obvious that it gets from its signature. Again, to get is responsibility of supplier. Responsibility of the method is to trace that supplier does. 
Again, the idea of method is in tracing. It can trace by plural ways, and how it traces is written in implementation. Here the name say you that it not just gets it but traces.

levBagryansky: @yegor256 please help. What option is cleaner?
```java
       @Override
        public Attr attr(final String name) {
            return this.getAttrWithTracing(() -> this.enclosure.attr(name));
        }


```
or
```java
       @Override
        public Attr attr(final String name) {
            return this.attr(() -> this.enclosure.attr(name));
        }
```
There is also an option to make the the callee generic in order to hoghlight its goal:
```java
        @Override
        public Attr attr(final String name) {
            return this.getItWithTracing(() -> this.enclosure.attr(name));
        }

        private <T> T getItWithTracing(final Supplier<T> supplier) {
            if (EOcage.PhTracedEnclosure.DATAIZING_CAGES.contains(this.cage)) {
                throw new ExFailure("The cage %s is already dataizing", this.cage);
            }
            EOcage.PhTracedEnclosure.DATAIZING_CAGES.add(this.cage);
            final T ret = supplier.get();
            EOcage.PhTracedEnclosure.DATAIZING_CAGES.remove(this.cage);
            return ret;
        }
```

levBagryansky: @yegor256 please check

yegor256: @levBagryansky First, using the word `get` for method names which are not "getters" is an anti-pattern: https://stackoverflow.com/questions/46139501/should-getters-from-domain-classes-not-be-prefixed-with-get-in-ddd-and-why Because of this, `getItWithTracing` is not a good option. 

Second, the `...WithTracing` part is an indicator of temporal coupling between statements inside the method, which is also an anti-pattern: https://blog.ploeh.dk/2011/05/24/DesignSmellTemporalCoupling/

A better and more object-oriented design would be to use... guess what, a decorator. You have a `Supplier` and you want to turn it into another `Supplier`, which does something when its `get()` method is called. Make it this way:

```
@Override
public Attr attr(final String name) {
  return TracingWhileGetting(() -> this.enclosure.attr(name));
}
```

Here, the `TracingWhileGetting` is the `Supplier` that does the "tracing" you need.

volodya-lombrozo: @yegor256 Looks like [overengineering](https://en.wikipedia.org/wiki/Overengineering) to me. We use this method only in this class and in the single place. I would agree if we used this `TracingWhileGetting` in other `Phi` classes, but we do not do it. Moreover, `Tracing` and `Getting` are participles (a participle [is a verb form](https://en.wikipedia.org/wiki/Participle).) But according with [the Java specification](https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html#:~:text=Class%20names%20should%20be%20nouns,such%20as%20URL%20or%20HTML):
> Class names should be nouns, in mixed case with the first letter of each internal word capitalized. 

Nouns (not adjectives or verbs.)  Maybe it's better to name the method `attr`? In this case we won't add extra code and will avoid the all convention violations which you mentioned before. What do you think?

levBagryansky:
@volodya-lombrozo WDYT about separating `private static final class PhTracedEnclosure` to separate file?

volodya-lombrozo:
> @volodya-lombrozo WDYT about separating `private static final class PhTracedEnclosure` to separate file?

@levBagryansky I don't think we need it. Since it is placed inside the class It indicates that `PhTracedEnclosure ` is used only in the single place.

levBagryansky:
@volodya-lombrozo `EOcage` becomes too large and ugly as for me.

volodya-lombrozo:
@levBagryansky I agree then, it makes sense to move it.

volodya-lombrozo (approved):
@levBagryansky Looks good to me. Thank you.

levBagryansky:
@yegor256 please check

Issue #2716, opened 2023-12-22
Extremely long unit tests in eo-runtime


volodya-lombrozo:
Unit tests in `eo-runtime` project run extremely long time: 
```
eo-runtime ......................................... SUCCESS [04:37 min]
```

I believe `4min` for unit test is extremely long. A close look on tests itself shows that only a few tests run so long:
```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.382 s - in EOorg.EOeolang.EOtuple_fluent_withTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.607 s - in EOorg.EOeolang.EOrefers_to_inner_vars_correctlyTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.396 s - in EOorg.EOeolang.EOcalculates_only_once_BROKENTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.196 s - in EOorg.EOeolang.EOrecursion_without_argumentsTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 10.831 s - in EOorg.EOeolang.EOcage_nested_objectsTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 10.789 s - in EOorg.EOeolang.EOdataizes_encaged_object_lazily_firstTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 25.733 s - in EOorg.EOeolang.EOswitch_simple_caseTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 83.445 s - in EOorg.EOeolang.EOswitch_with_several_true_casesTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 84.961 s - in EOorg.EOeolang.EOswitch_strings_caseTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 102.758 s - in EOorg.EOeolang.EOswitch_complex_caseTest
[...]

PR #2898, opened 2024-02-27
#2716: tuple optimization

levBagryansky:
Relates to #2716
This pr makes `tuple.at` faster. Testing time decreased by 2 times on my laptop.
Such test became more than 10 times faster:
```
#Test
[] > just-long-test
  seq > @
    *
      TRUE
      TRUE
      TRUE
      TRUE
      TRUE
```

---

## PR-Codex overview
This PR adds new tests and a method `at-without-checks` to the `tuple` object in the `eo-runtime`. 

### Detailed summary
- Added tests for `at-without-checks` method with first and last element
- Added `at-without-checks` method to `tuple` object for retrieving elements without index validation

### DIFF
--- a/eo-runtime/src/main/eo/org/eolang/tuple.eo
+++ b/eo-runtime/src/main/eo/org/eolang/tuple.eo
@@ -60,8 +60,8 @@
       ^.at-without-checks index
 
   # Takes element from the tuple without checking i.
-  # i must be less than len and greater or equal 0.
-  # This is faster than ^.at.
+  # i must be less than length and greater or equal 0.
+  # This is faster than at.
   [i] > at-without-checks
     i > idx!
     if. > @
--- a/eo-runtime/src/main/eo/org/eolang/tuple.eo
+++ b/eo-runtime/src/main/eo/org/eolang/tuple.eo
@@ -59,15 +59,15 @@
       error "Given index is out of tuple bounds"
       ^.at-without-checks index
 
-  # Takes element from the tuple without checking i.
-  # i must be less than length and greater or equal 0.
-  # This is faster than at.
-  [i] > at-without-checks
-    i > idx!
-    if. > @
-      idx.lt head.length
-      ^.head.at-without-checks idx
-      ^.tail
+    # Takes element from the tuple without checking i.
+    # i must be less than length and greater or equal 0.
+    # This is faster than at.
+    [i] > at-without-checks
+      i > idx!
+      if. > @
+        idx.lt head.length
+        ^.^.head.at.at-without-checks idx
+        ^.^.tail
 
   # Create a new tuple with this element added to the end of it.
   [x] > with
--- a/eo-runtime/src/main/eo/org/eolang/tuple.eo
+++ b/eo-runtime/src/main/eo/org/eolang/tuple.eo
@@ -65,8 +65,8 @@
   [i] > at-without-checks
     i > idx!
     if. > @
-      i.lt head.length
-      ^.head.at-without-checks i
+      idx.lt head.length
+      ^.head.at-without-checks idx
       ^.tail
 
   # Create a new tuple with this element added to the end of it.
--- a/eo-runtime/src/main/eo/org/eolang/tuple.eo
+++ b/eo-runtime/src/main/eo/org/eolang/tuple.eo
@@ -57,7 +57,7 @@
         index.lt 0
         index.gte len
       error "Given index is out of tuple bounds"
-      ^.at.at-without-checks index
+      at-without-checks index
 
     # Takes element from the tuple without checking i.
     # i must be less than length and greater or equal 0.
--- a/eo-runtime/src/main/eo/org/eolang/tuple.eo
+++ b/eo-runtime/src/main/eo/org/eolang/tuple.eo
@@ -57,10 +57,17 @@
         index.lt 0
         index.gte len
       error "Given index is out of tuple bounds"
-      if.
-        index.lt (len.plus -1)
-        ^.head.at index
-        ^.tail
+      ^.at-without-checks index
+
+  # Takes element from the tuple without checking i.
+  # i must be less than len and greater or equal 0.
+  # This is faster than ^.at.
+  [i] > at-without-checks
+    i > idx!
+    if. > @
+      i.lt head.length
+      ^.head.at-without-checks i
+      ^.tail
 
   # Create a new tuple with this element added to the end of it.
   [x] > with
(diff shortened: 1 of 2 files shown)


levBagryansky:
@maxonfjvipon please check

eo-runtime/src/main/eo/org/eolang/tuple.eo
```diff
@@ -57,10 +57,17 @@
...
-      if.
-        index.lt (len.plus -1)
-        ^.head.at index
-        ^.tail
+      ^.at-without-checks index
+
+  # Takes element from the tuple without checking i.
+  # i must be less than length and greater or equal 0.
+  # This is faster than at.
+  [i] > at-without-checks
```
maxonfjvipon, line 65: @levBagryansky without checking the source code it's not really clear why it's faster and what "without checks" means.
I would suggest not to extend `tuple` interface, but speed up `at` attribute by moving this `at-without-checks` inside `at`

eo-runtime/src/main/eo/org/eolang/tuple.eo
```diff
@@ -57,10 +57,17 @@
...
+    i > idx!
```
maxonfjvipon, line 66: @levBagryansky `idx` is not used anywhere

maxonfjvipon (changes requested):

eo-runtime/src/main/eo/org/eolang/tuple.eo
maxonfjvipon: @levBagryansky so when you enter `at` you do all your checks just once and then recursively calls `at-without-checks`

levBagryansky: @maxonfjvipon let's rename it somehow

levBagryansky: @maxonfjvipon done

eo-runtime/src/main/eo/org/eolang/tuple.eo
```diff
@@ -57,10 +57,16 @@
         index.lt 0
         index.gte len
       error "Given index is out of tuple bounds"
-      if.
-        index.lt (len.plus -1)
-        ^.head.at index
-        ^.tail
+      ^.at.at-without-checks index
```
maxonfjvipon, line 60: @levBagryansky let's replace it with just `at-without-checks index`

eo-runtime/src/main/eo/org/eolang/tuple.eo
```diff
@@ -57,10 +57,16 @@
...
+
+    # Takes element from the tuple without checking i.
+    # i must be less than length and greater or equal 0.
```
maxonfjvipon, line 63: @levBagryansky `i` actually may be any object, it's not mandatory `i` to be `int`. It can be for example
```
[] > my-i
  [x] > lt
    TRUE > @
```
And it will work. So I would say that it's expected `i` has `lt` attribute and we don't check its value

maxonfjvipon (changes requested):

eo-runtime/src/main/eo/org/eolang/tuple.eo
levBagryansky: @maxonfjvipon i should be int. Otherwise the object does not have sense

maxonfjvipon: @levBagryansky It should not and we can't guarantee it. All we need that `i` has attribute `lt` with 1 free attribute. That's it

levBagryansky: What will be `(* 1 2 3) .at my-int` here? Not ub but something like that, because then the method guarantees nothing

levBagryansky: @maxonfjvipon Also I did not write that `i` should be `int`. i must be less than length and greater or equal 0. That is all. If you object is less  than length and greater or equal to 0 then you can use this method.

maxonfjvipon: @levBagryansky `lt` means "less than" only because we decided to attach that semantics to the attribute. What if `i` is an object with just `lt` and `gte` attribute that are going to google and check the weather on the Mars.
I'm trying to say that we should not get attached to semantics. I believe description should just describe the final behaviour

levBagryansky: When I write x less than y, I mean `x.lt y` is TRUE. Ok, we can write the condition this way: i.lt len is TRUE

maxonfjvipon: @levBagryansky but `x.lt y` is not supposed to mean "x is less than y"

levBagryansky: Supposed to be. Actually it is abstraction. We have polymorphism here.

levBagryansky: We can remove it if it seems confusing

maxonfjvipon: @levBagryansky I would propose something like this for the description: 
Takes element from the tuple without index validation.
Here `i` must be an object that can be comparable with `int` using `lt` method.

levBagryansky: @maxonfjvipon ok

eo-runtime/src/main/eo/org/eolang/tuple.eo
```diff
@@ -57,10 +57,16 @@
...
         index.gte len
       error "Given index is out of tuple bounds"
-      if.
-        index.lt (len.plus -1)
-        ^.head.at index
-        ^.tail
+      at-without-checks index
+
+    # Takes element from the tuple without index validation.
+    # Here i must be an object that can be comparable with int using lt attribute.
```
maxonfjvipon, line 63: @levBagryansky let's quote "i", "int" and "lt" with ` brackets so it'll be possible to render comments in markdown (in the future)

maxonfjvipon (changes requested):

maxonfjvipon (approved):

levBagryansky:
@yegor256 please check

levBagryansky:
@yegor256 please check

yegor256:
@rultor merge

yegor256:
@rultor status

Issue #2863, opened 2024-02-06
reduce code duplication to 15 lines per piece


yegor256:
We decreased it to 25 in https://github.com/objectionary/eo/issues/2859

See this line: https://github.com/objectionary/eo/blob/fd123092cd18caa5fffc730a9b911fbfddc83db9/.github/workflows/simian.yml#L27C39-L27C39 Let's decrease the --threshold down to 15

yegor256:
@c71n93 maybe you can help with this one?

PR #2920, opened 2024-03-05
Reduce code duplication down to 16 lines per piece

c71n93:
Related to #2863

### DIFF
--- a/.github/workflows/simian.yml
+++ b/.github/workflows/simian.yml
@@ -24,4 +24,4 @@ jobs:
       - run: |
           wget --quiet http://public.yegor256.com/simian.jar -O /tmp/simian.jar
       - run: |
-          java -jar /tmp/simian.jar -threshold=25 -excludes=**/gen -excludes=**/it **/*.java
+          java -jar /tmp/simian.jar -threshold=16 -excludes=**/gen -excludes=**/it **/*.java
--- a/eo-runtime/src/test/java/org/eolang/DataizedTest.java
+++ b/eo-runtime/src/test/java/org/eolang/DataizedTest.java
@@ -48,22 +48,7 @@ final class DataizedTest {
         final Level before = log.getLevel();
         log.setLevel(Level.ALL);
         final List<LogRecord> logs = new LinkedList<>();
-        final Handler hnd = new Handler() {
-            @Override
-            public void publish(final LogRecord record) {
-                logs.add(record);
-            }
-
-            @Override
-            public void flush() {
-                throw new UnsupportedOperationException("#flush()");
-            }
-
-            @Override
-            public void close() throws SecurityException {
-                throw new UnsupportedOperationException("#close()");
-            }
-        };
+        final Handler hnd = new TestHandler(logs);
         log.addHandler(hnd);
         new Dataized(new Data.ToPhi(1L), log).take();
         log.setLevel(before);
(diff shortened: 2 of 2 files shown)


yegor256:
@c71n93 what's up with this one? still a draft?

c71n93:
@yegor256 still working on it

c71n93:
@maxonfjvipon could you check this one, please?

c71n93:
@maxonfjvipon could you check this one, please?

levBagryansky (commented):
@c71n93 I am not sure that uniting of these lines is real solution of the duplication problem

maxonfjvipon (approved):
@c71n93 looks good. Even if you're not sure, the final result speaks for itself

c71n93:
> @c71n93 I am not sure that uniting of these lines is real solution of the duplication problem
> 
> ```
> String.format(
>     "Object inside %s has wrong class %s",
>     src,
>     result.getClass()
> )
> ```

@levBagryansky Yes, but I don't see any other options. This methods are located in different independent subprojects (`eo-maven-plugin` and `eo-runtime`). 

The only solution I see is to create a dependency between subprojects (or create some `eo-util` which all subprojects will depend on). But as I know, this option is unacceptable. Maybe you have any other ideas?

levBagryansky:
@maxonfjvipon As for me, it's not so bad to create a component on which `eo-runtime` and `eo-maven-plugin` would depend. It's just that when implementing it, we should not allow cyclic dependencies([Acyclic dependencies principle](https://en.wikipedia.org/wiki/Acyclic_dependencies_principle)) and that it should be more stable than components dependent on it ([Stable Dependencies Principle](https://deviq.com/principles/stable-dependencies)).

maxonfjvipon:
@levBagryansky while compilation eo program we download `eo-runtime` as jar package, but we don't download its dependencies (fat jar). It's by design. That's why it should be independent.

eo-maven-plugin/src/main/java/org/eolang/maven/rust/Names.java
```diff
@@ -189,20 +189,13 @@ private static ConcurrentHashMap<String, String> load(final Path src) throws IOE
             final Object result = map.readObject();
             if (result.getClass() != ConcurrentHashMap.class) {
                 throw new ClassCastException(
-                    String.format(
-                        "Object inside %s has wrong class %s",
-                        src,
-                        result.getClass()
-                    )
+                    String.format("Object inside %s has wrong class %s", src, result.getClass())
```
yegor256, line 192: @c71n93 maybe it's better to slightly change the message?

eo-maven-plugin/src/main/java/org/eolang/maven/rust/Names.java
```diff
@@ -189,20 +189,13 @@ private static ConcurrentHashMap<String, String> load(final Path src) throws IOE
...
                 );
             }
             return (ConcurrentHashMap<String, String>) result;
         } catch (final ClassNotFoundException exc) {
             throw new IllegalArgumentException(
-                String.format(
-                    "File %s contains invalid data",
-                    src
-                ),
+                String.format("File %s contains invalid data", src),
```
yegor256, line 198: @c71n93 same here: maybe it's better to slightly change the message?

maxonfjvipon:
@levBagryansky then I can ask, why are they in different modules? Maybe it's the main problem? Because you're offering to sacrifice eo-runtime independence (which has highest priority) for the sake of the SPR and your convenience

levBagryansky:
@maxonfjvipon That's how they should be in the same module. The architecture does not allow this to be done.

yegor256:
@rultor merge

PR #3054, opened 2024-04-05
3048 seq refactoring

### LINKED ISSUE #3048: Performance issue with `seq` due to unnecessary 'Dataize' calls in 'for' loop
If you take a look at the current implementation of `seq`:
```java
@Override
public Phi lambda() {
    final Phi steps = this.take("steps");
    final Long length = new Dataized(
        steps.take("length")
    ).take(Long.class);
    for (long idx = 0; idx < length - 1; ++idx) {
        new Dataized(
            new PhWith(
                steps.take("at").copy(),
                0, new Data.ToPhi(idx)
            )
        ).take();
    }
    final Phi ret;
    if (length > 0) {
        final Phi last = steps.take("at").copy();
        last.put(0, new Data.ToPhi(length - 1));
        ret = last;
    } else {
        ret = new Data.ToPhi(false);
    }
    return ret;
}
```
you may see that in `for` loop we dataize `length - 1` times `steps.at index` . `tuple.at` is a quite convenient interface to obtain objects by their index, but it's heavy.
Instead in `tuple` we already have `head` and `tail` where `head` is inner `tuple` and `tail` is a value.

So instead of:
```
Dataize(steps.at 0)
Dataize(steps.at 1)
Dataize(steps.at 2)
....
```

We can do just:
```
Dataize(steps.head.head.head.tail)
Dataize(steps.head.head.tail)
Dataize(steps.head.tail)
```

which is quite faster. 

Moreover, we can save every `head` when we take them for the first time so we won't retrieve them on the next steps.

levBagryansky:
Closes #3048 
New test passes for `0.180 s`

---

## PR-Codex overview
The focus of this PR is to optimize the conversion of EO tuple to a Java array in `EOseq.java`.

### Detailed summary
- Added a method `eoTupleAsArray` to convert EO tuple to Java array efficiently.
- Refactored the `lambda` method in `EOseq.java` to use the new `eoTupleAsArray` method for better performance.
- Replaced usage of `PhWith` with `PhMethod` for consistency and improved readability.

### DIFF
--- a/eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
+++ b/eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
@@ -81,7 +81,7 @@ public final class EOseq extends PhDefault implements Atom {
         Phi external = args;
         for (int ind = length - 1; ind >= 0; --ind) {
             res[ind] = new PhMethod(external, "tail");
-            external = external.copy().take("head");
+            external = external.take("head");
         }
         return res;
     }
--- a/eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
+++ b/eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
@@ -71,6 +71,11 @@ public final class EOseq extends PhDefault implements Atom {
         return ret;
     }
 
+    /**
+     * Converts eo tuple to java array.
+     * @param args Eo tuple.
+     * @return Java array.
+     */
     private static Phi[] eoTupleAsJavaArray(final Phi args) {
         final int length = Math.toIntExact(
             new Dataized(
--- a/eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
+++ b/eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
@@ -58,7 +58,7 @@ public final class EOseq extends PhDefault implements Atom {
     @Override
     public Phi lambda() {
         final Phi steps = this.take("steps");
-        final Phi[] items = EOseq.eoTupleAsJavaArray(steps);
+        final Phi[] items = EOseq.eoTupleAsArray(steps);
         for (int ind = 0; ind < items.length - 1; ++ind) {
             new Dataized(items[ind]).take();
         }
@@ -76,7 +76,7 @@ public final class EOseq extends PhDefault implements Atom {
      * @param args Eo tuple.
      * @return Java array.
      */
-    private static Phi[] eoTupleAsJavaArray(final Phi args) {
+    private static Phi[] eoTupleAsArray(final Phi args) {
         final int length = Math.toIntExact(
             new Dataized(
                 args.take("length")
--- a/eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
+++ b/eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
@@ -58,7 +58,7 @@ public final class EOseq extends PhDefault implements Atom {
     @Override
     public Phi lambda() {
         final Phi steps = this.take("steps");
-        final Phi[] items = eoTupleAsJavaArray(steps);
+        final Phi[] items = EOseq.eoTupleAsJavaArray(steps);
         for (int ind = 0; ind < items.length - 1; ++ind) {
             new Dataized(items[ind]).take();
         }
--- a/eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
+++ b/eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
@@ -60,8 +60,8 @@ public final class EOseq extends PhDefault implements Atom {
     public Phi lambda() {
         final Phi steps = this.take("steps");
         final Phi[] items = toArray(steps);
-        for (int i = 0; i < items.length - 1; i++) {
-            new Dataized(items[i]).take();
+        for (int ind = 0; ind < items.length - 1; ++ind) {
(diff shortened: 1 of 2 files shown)


maxonfjvipon (changes requested):

eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
```diff
@@ -58,25 +59,33 @@ public EOseq(final Phi sigma) {
...
             ret = new Data.ToPhi(false);
         }
         return ret;
     }
+
+    private static Phi[] toArray(final Phi args) {
+        final int length = Math.toIntExact(
+            new Dataized(
+                args.take("length")
+            ).take(Long.class)
```
maxonfjvipon, line 81: @levBagryansky to cast to `long` it's better to use `.intValue()`

eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
```diff
@@ -58,25 +59,33 @@ public EOseq(final Phi sigma) {
...
+        );
+        final Phi[] res = new Phi[length];
+        Phi external = args;
+        for (int ind = length - 1; ind >= 0; --ind) {
+            res[ind] = new PhMethod(external, "tail");
```
maxonfjvipon, line 86: @levBagryansky why are you using `PhMethod` here? Why don't just use `external.take('tail')`

eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
```diff
@@ -58,25 +59,33 @@ public EOseq(final Phi sigma) {
...
+            external = external.copy().take("head");
```
maxonfjvipon, line 87: @levBagryansky no need to `copy()` external here. It'll be copied when taken from `external`

eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
```diff
@@ -58,25 +59,33 @@ public EOseq(final Phi sigma) {
...
-            steps.take("length")
-        ).take(Long.class);
-        for (long idx = 0; idx < length - 1; ++idx) {
-            new Dataized(
-                new PhWith(
-                    steps.take("at").copy(),
-                    0, new Data.ToPhi(idx)
-                )
-            ).take();
+        final Phi[] items = toArray(steps);
```
maxonfjvipon, line 62: @levBagryansky let's follow the convention and use static methods with class prefix: `EOseq.toArray()`
Also it would be better to rename `toArray` to something like `stepsAsArray`, `stepsToArray`. Anyway it should be a noun.

eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
```diff
@@ -58,25 +59,33 @@ public EOseq(final Phi sigma) {
...
-            last.put(0, new Data.ToPhi(length - 1));
+            last.put(0, new Data.ToPhi((long) items.length - 1L));
             ret = last;
         } else {
             ret = new Data.ToPhi(false);
         }
         return ret;
     }
+
+    private static Phi[] toArray(final Phi args) {
```
maxonfjvipon, line 77: @levBagryansky let's add a java doc here

maxonfjvipon (changes requested):

eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
levBagryansky: @maxonfjvipon `Dataized` does not have `intValue` method

maxonfjvipon: @levBagryansky `Long` has

eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
levBagryansky: @maxonfjvipon done

eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
levBagryansky: @maxonfjvipon I see. The difference is that `toIntExact` throws exception in case of overflow. So it is safer.. What if mistake happened and `new args.take("length")).take(Long.class)` returned too big number.

maxonfjvipon: @levBagryansky ok, let it be so

maxonfjvipon (changes requested):
@levBagryansky very good, just one last change

eo-runtime/src/main/java/EOorg/EOeolang/EOseq.java
```diff
@@ -58,25 +58,36 @@ public EOseq(final Phi sigma) {
...
         }
         return ret;
     }
+
+    /**
+     * Converts eo tuple to java array.
+     * @param args Eo tuple.
+     * @return Java array.
+     */
+    private static Phi[] eoTupleAsJavaArray(final Phi args) {
```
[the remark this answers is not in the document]
levBagryansky: @maxonfjvipon ok

maxonfjvipon (approved):

maxonfjvipon:
@yegor256 please check

levBagryansky:
@yegor256 please check

yegor256:
@rultor merge

PR #3090, opened 2024-04-14
2390

### LINKED ISSUE #2390: TranspileMojo.java:262-267: Add concurrency tests for the...
The puzzle `2375-409ab48f` from #2375 has to be resolved: 

https://github.com/objectionary/eo/blob/d3caa91541b57a7f6b5ef7c71f11f1037e15c7a4/eo-maven-plugin/src/main/java/org/eolang/maven/TranspileMojo.java#L262-L267

The puzzle was created by rultor on 14-Aug-23. 

l3r8yJ:
@maxonfjvipon @volodya-lombrozo take a look, please 

closes #2390

---

## PR-Codex overview
This PR enhances `TranspileMojo` by implementing a mechanism for distinguishing "internal" and "external" classes, inspired by Gradle's `implementation` and `api` concepts.

### Detailed summary
- Improved handling of internal and external classes
- Added logic to separate classes into different directories based on package and version
- Introduced the concept of `implementation` and `api` for dependencies

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/TranspileMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/TranspileMojo.java
@@ -264,16 +264,16 @@ public final class TranspileMojo extends SafeMojo {
      * In other words, concurrent file deletions on the Windows OS can lead to an
      * {@link java.nio.file.AccessDeniedException}, which could crash the build.
      * _____
+     * @param java The list of java files.
      * @todo #2375:90min. Implement mechanism for "inner" and "outer" classes.
      *  To get rid of TranspileMojo#cleanUpClasses, we can implement mechanism,
      *  that will mark classes of the project like "inner", which is will be checked
      *  during the compilation process.
      *  Another solution is to create analog of gradle `implementation` and `api`,
-     *    - implementation – just includes the dependency for inner usage
-     *    - api – allows to users of API to
-     *    use dependency which was added to project via `api` keyword
+     *  - implementation – just includes the dependency for inner usage
+     *  - api – allows to users of API to
+     *  use dependency which was added to project via `api` keyword
      *  <p/><a href="https://shorturl.at/abns4">More about api and implementation here</a>
-     * @param java The list of java files.
      */
     private void cleanUpClasses(final Collection<? extends Path> java) {
         final Set<Path> unexpected = java.stream()
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/TranspileMojo.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/TranspileMojo.java
@@ -58,6 +58,16 @@ import org.eolang.maven.util.Rel;
 /**
  * Compile.
  *
+ * @todo #2375:90min.
+ * Implement mechanism for "inner" and "outer" classes.
+ *  To get rid of TranspileMojo#cleanUpClasses, we can implement mechanism,
+ *  that will mark classes of the project like "inner", which is will be checked
+ *  during the compilation process.
+ *  Also here is one more solution, it's create analog of gradle `implementation` and `api`,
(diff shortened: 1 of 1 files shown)


l3r8yJ:
@maxonfjvipon @volodya-lombrozo reminder

volodya-lombrozo (changes requested):
@l3r8yJ Thank you for the contribution! Could you please make this puzzle more understandable? Imagine newcomers joining project: they should understand what's going on and what they have to do to implement this puzzle by reading the puzzle description.

l3r8yJ:
@volodya-lombrozo I made some clarifications based on your review. Also for better understanding of the term `dirty classes`, can you explain what means `dirty` in this context?

eo-maven-plugin/src/main/java/org/eolang/maven/TranspileMojo.java
```diff
@@ -265,12 +265,18 @@ private List<Path> transpile(
      * {@link java.nio.file.AccessDeniedException}, which could crash the build.
      * _____
      * @param java The list of java files.
-     * @todo #2375:90min. Add concurrency tests for the TranspileMojo.cleanUpClasses method.
-     *  We should be sure that the method works correctly in a concurrent environment.
-     *  In order to do so we should add a test that will run the cleanUpClasses method in
-     *  multiple threads and check that the method works correctly without exceptions.
-     *  We can apply the same approach as mentioned in that post:
-     *  <a href="https://www.yegor256.com/2018/03/27/how-to-test-thread-safety.html">Post</a>
+     * @todo #2375:90min. Implement a mechanism for "internal" and "external" classes.
```
volodya-lombrozo, line 268: @l3r8yJ, could you please provide more context? Imagine someone opening this issue in the future; they should be able to understand what to do at first glance:

1. I still don't understand why we might need "internal" and "external" classes, or why we need to implement them.
2. Why should "internal" classes be included in the catalog?
3. Which specific catalog are you referring to?
4. Why should "external" classes be placed in a different directory?
5. Are "catalog" and "directory" the same entities in this context?

volodya-lombrozo (changes requested):

volodya-lombrozo:
> @volodya-lombrozo I made some clarifications based on your review. Also for better understanding of the term `dirty classes`, can you explain what means `dirty` in this context?

@l3r8yJ  `dirty` refers to a type of class that was included in a jar during compilation but should not have been included. In other words, they are classes that were added to a jar by mistake.

Example:

Let's imagine a group of developers of the eo-strings library. Perhaps by mistake, they decided to create an [uber-jar](https://stackoverflow.com/questions/11947037/what-is-an-uber-jar-file). So now, the `eo-strings` jar file contains not only Java atoms but also all dependencies (specifically, all the classes from these dependencies), including the `eo-runtime` library. Therefore, we call these classes `dirty` - they are here by mistake and shouldn't be.

Now, imagine you are trying to build another library that uses `eo-strings` and `eo-runtime`. When you compile your library, you will end up in a situation where you have two versions of `eo-runtime`:

```
your-eo-library-dependencies:
-> eo-strings -> eo-runtime:x
-> eo-runtime:y
```

This is a problem that is sometimes hard to identify. So we "clean" these dirty classes, which actually just hides the problem.

l3r8yJ:
@volodya-lombrozo so, might be this issue can be resolved with some plugin, which will validate project on `uber-jar`? Plugin should be applied to library itself, wdyt? It should exclude such cases

volodya-lombrozo:
@l3r8yJ, I believe we should have protection on both sides, in our compiler and during the building of libraries. I'm afraid that creating a new plugin might be some sort of [overengineering](https://en.wikipedia.org/wiki/Overengineering). But I like the idea of libraries validation.

maxonfjvipon:
@l3r8yJ since it's no more relevant I'm closing this one

PR #3124, opened 2024-04-20
#3035: Add yamllint GHA

### LINKED ISSUE #3035: let's add yamllint GHA job
See how it works here: https://github.com/yegor256/cam/blob/master/.github/workflows/yamllint.yml

slry:
@yegor256 take a look, please
Closes #3035

### DIFF
--- a/.github/workflows/yamllint.yml
+++ b/.github/workflows/yamllint.yml
@@ -1,0 +1,15 @@
+---
+name: yamllint
+'on':
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  yamllint:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ibiqlik/action-yamllint@v3


slry:
@yegor256 as i noticed, yamllint fails mostly because of `line-length`, `brackets` and `new-line-at-the-end-of-file`.
My suggestion is either create yamllint config to mark these errors as warnings or I can create issues to clean up them.
What do you think?

yegor256:
@slry maybe it's better to clean them up in this PR? (unless they are hard to fix)

slry:
@yegor256, I can fix `brackets` and `new-line-at-the-end-of-file` but `line-length` is the hard one, since most of files that are failing is eolang syntax test files that have intended long line length that could not be fixed
For example, take a look at this [syntax test](https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/test/resources/org/eolang/maven/packs/pre/long-class-name.yaml)

yegor256:
@slry let's allow lines to have any length in YAML files

slry:
@yegor256 for some reason new line at the end of file in `eo-maven-plugin` yaml files is a cause for test fail...

Even though it is a rule in yamllint by POSIX standard ([here](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.new_line_at_end_of_file))

yegor256:
@slry I see that yamllint job passes in this PR. we can merge it?

slry:
@yegor256 yamllint job passes, but is it ok that codecov job is failing?
It seems like new line at the end of file does not parse well...

slry:
@yegor256 I found a solution to parsing problem
We can use `|-` instead of `|` in YAML

Test with `|`
```yaml
eo: |
    foo
    bar
```
Result: `"foo\nbar\n"`

Test with `|-`
```yaml
eo: |-
    foo
    bar
```
Result: `"foo\nbar"`

Is it ok if I try to fix it?

slry:
@yegor256 tests are passed, now we can merge it

yegor256:
@rultor merge

yegor256:
@slry thanks!

PR #3123, opened 2024-04-19
#3122: Using of Synced

levBagryansky:
Closes #3122 

---

## PR-Codex overview
This PR introduces a new `Synced` scalar to synchronize access in `DcsWithRuntime` class.

### Detailed summary
- Added `Synced` scalar import
- Wrapped `Sticky` scalar in `Synced` scalar for synchronization

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #3149, opened 2024-04-26
3148 chRemote

### LINKED ISSUE #3148: `ChRemote` has threadsafety vulnerability
`ChRemote` has field 
```java
    /**
     * Cached text of hashes.
     */
    private static final Text CACHE = new Sticky(new CommitHashesText());
```
This field is not thread-safe and synchronizes nohow. 
Also, `CommitHashesText` is already cached, so we do not need to cache it here again.
Also, this field has the same testing problem like in #3114.

levBagryansky:
Closes #3148

---

## PR-Codex overview
The focus of this PR is to refactor the `CommitHash` interface to extend `Scalar<String>` and make improvements to the `ChRemote` class for better performance and thread safety.

### Detailed summary
- `CommitHash` interface now extends `Scalar<String>`
- Removed unnecessary import in `ChRemote`
- Improved thread safety in `ChRemoteTest` with new test method


levBagryansky:
@maxonfjvipon please review. This test does not cover current vulnerability, however it protect us from thread-unsafety in future. 
Current vulnerability is hard to test. We have similar case in https://github.com/objectionary/eo/pull/3114#issuecomment-2069274548 and in #3123

eo-maven-plugin/src/main/java/org/eolang/maven/hash/CommitHash.java
```diff
@@ -23,12 +23,15 @@
...
 
+import org.cactoos.Scalar;
+
 /**
  * Hash of tag.
  *
  * @since 0.28.11
  */
-public interface CommitHash {
+@FunctionalInterface
```
maxonfjvipon, line 33: @levBagryansky why do we need it here?

levBagryansky: @maxonfjvipon I added it because this actually is a functional interface.

eo-maven-plugin/src/main/java/org/eolang/maven/hash/CommitHash.java
```diff
@@ -23,12 +23,15 @@
...
+public interface CommitHash extends Scalar<String> {
```
levBagryansky, line 34: @maxonfjvipon I inhereted it firstly because it has the same method so it really can be inherited. Also it allowed to write                     
```java
                    Stream.generate(
                        () -> new ChRemote("0.23.19")
                    ).limit(threads).collect(Collectors.toList())
```
and this is cool.

maxonfjvipon (approved):

levBagryansky:
@yegor256 please take a look

yegor256:
@rultor merge

Issue #3169, opened 2024-05-06
Decapitalize attribute while converting to phi


maxonfjvipon:
We have EO code:
```
x
  [] >>
```

It converts to the next XMIR:
```xml
<o base="x">
  <o abstract="" name="OBJ-2-2">
</o>
```

When we try to convert it to phi - we get error because attributes can not start with uppercase letter.
We should add extra XSL transformation that would turn `OBJ-2-2` to `oBJ-2-2`.

PR #3170, opened 2024-05-06
fix(#3169): decapitalize auto named objects on converting to phi

maxonfjvipon:
Closes: #3169

---

## PR-Codex overview
This PR adds licensing information and updates XML files in the project. 

### Detailed summary
- Added MIT License to YAML and XSL files
- Updated XML files with licensing info
- Modified XSL files for XML transformation
- Updated Java file imports and variables

> The following files were skipped due to too many changes: `eo-maven-plugin/src/test/resources/org/eolang/maven/packs/decapitalize-auto-named.yaml`


maxonfjvipon:
@yegor256 please check

maxonfjvipon:
@yegor256 please check

yegor256:
@maxonfjvipon why can't we just auto-generate names like `attribute-at-XXX` where `XXX` is the number of the line. Later, we can introduce a check that would prohibit programmers from giving similar names to their attributes, _if_ there are conflicts.

maxonfjvipon:
@yegor256 we already have check that checks if there are more than one objects with the same name, so imagine:
```
[] > obj
  x > y
    [] >>
  5 > attribute-at-3
```
Programmer writes the code and expects it works because why not? But it fails on compilation because of the name conflict. Usability? I don't think so.
I believe it's quite strong rule to prohibit names started with capital letter and use such names on auto-generation. Because there's no any possible way to get a conflict in such case.
So my suggestion is to stick with current implementation. What's your decision?

maxonfjvipon:
@yegor256 WDYT?

maxonfjvipon:
@yegor256 maybe you're right

yegor256:
@maxonfjvipon later we can make our game-generator more powerful, so that it avoids conflicts. But making names that don't follow the grammar of the language -- this definitely is a wrong idea.

PR #3185, opened 2024-05-12
#3140: Save error message

### LINKED ISSUE #3140: Print full error message if cargo fails
Now it  prints just 
```
[ERROR] Failed to build cargo project with dest = /private/var/folders/xj/f4s949893tq37zzw0pdmd68h0000gn/T/junit8772813024511140751/target/eo-test/Lib/native10
[ERROR] Non-zero exit code #101 of 'cargo'
```

levBagryansky:
Closes #3140

---

## PR-Codex overview
The focus of this PR is to enhance the Rust build process in the Maven plugin by introducing new features and improvements.

### Detailed summary
- Added `WeHaveCargo` as an execution condition for Rust builds
- Modified exception handling in `BuildFailureException`
- Updated Rust build process in `RustNode` class
- Added new tests for Rust build failures and error messages

### DIFF
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/WeHaveCargo.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/WeHaveCargo.java
@@ -35,7 +35,7 @@ import org.junit.jupiter.api.extension.ExtensionContext;
 
 /**
  * JUnit's extension to skip a test if there is no cargo.
- * Cargo is a build system for rust. so it is required for rust inserts.
+ * Cargo is a build system for rust. So it is required for rust inserts.
  *
  * @since 0.30
  */
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/WeHaveCargo.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/WeHaveCargo.java
@@ -35,12 +35,13 @@ import org.junit.jupiter.api.extension.ExtensionContext;
 
 /**
  * JUnit's extension to skip a test if there is no cargo.
- * Cargo is a build system for rust so it is required for rust inserts.
+ * Cargo is a build system for rust. so it is required for rust inserts.
  *
  * @since 0.30
  */
 @SuppressWarnings("JTCOP.RuleAllTestsHaveProductionClass")
-public final class CargoCondition implements ExecutionCondition {
+public final class WeHaveCargo implements ExecutionCondition {
+
     @Override
     public ConditionEvaluationResult evaluateExecutionCondition(
         final ExtensionContext context) {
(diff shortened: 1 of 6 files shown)


levBagryansky:
@maxonfjvipon WDYT

levBagryansky:
@maxonfjvipon please check

eo-maven-plugin/src/test/java/org/eolang/maven/WeHaveCargo.java
```diff
@@ -35,12 +35,13 @@
 
 /**
  * JUnit's extension to skip a test if there is no cargo.
- * Cargo is a build system for rust so it is required for rust inserts.
+ * Cargo is a build system for rust. so it is required for rust inserts.
```
maxonfjvipon, line 38: @levBagryansky typo. "So" instead of "so"

maxonfjvipon (changes requested):

eo-maven-plugin/src/test/java/org/eolang/maven/WeHaveCargo.java
levBagryansky: @maxonfjvipon Indeed

maxonfjvipon (approved):
@levBagryansky LGTM!

yegor256:
@rultor merge

yegor256:
@levBagryansky did you try to use `.withRedirect(false)` of Jaxec, as explained [here](https://github.com/yegor256/jaxec)?

levBagryansky:
@yegor256 Yes, it still did not print it.

yegor256:
@levBagryansky we should find out, why it doesn't print. Most probably, something is wrong with the configuration of logging levels in `log4j.properties` file. Just throwing away the library and re-implementing its functionality is not a good practice.

levBagryansky:
@yegor256 `log4j.properties` are configured right, and `Jaxec` has error message inside, in `result.stdout()`. The problem is that this message is not attached to Exception in line https://github.com/yegor256/jaxec/blob/master/src/main/java/com/yegor256/Jaxec.java#L365 so we cannot get it.

yegor256:
@levBagryansky why it's important to put entire stdout into an exception (pretty unusual practice) instead of just sending it to the console/log? What do you mean by "we cannot get it"? If you can see the stacktrace (printed to stderr), you should see the log printed to stdout, I believe.

levBagryansky:
@yegor256 because usually we would not like to see logs for all cargo projects. There are several inserts in eo program so logs from them would be mixed.

levBagryansky:
@yegor256 We would like to get `stdout & stderr` of the process  as `String` as the result of `Jaxec::execUnsafe`. Then we would print this `String` to log. But the problem is that `execUnsafe` does not return the `String` in case of exception, and it does not wrap it to `Exception`. 
Looks like the only way to use `Jaxec` is to print stdout continuously. But then the log will be too unreadable, since many cargo processes will print their stdout at the same time.

yegor256:
@levBagryansky we will use Jaxec 0.4.0 with `Result` instead

Issue #3343, opened 2024-08-20
How to print a small part of XMIR to EO?


volodya-lombrozo:
I usually have small portions of XMIR:
```xml
 <o base=".plus">
    <o base="int" data="bytes" line="1143583134">00 00 00 00 00 00 00 01</o>
    <o base="int" data="bytes" line="1167891236">00 00 00 00 00 00 00 02</o>
 </o>
```

And I need to convert it to the EO representation:
```
1.plus 2
```

How can I do it?

I tried to run the following code:
```java
 new Xmir.Default(new XMLDocument(xml)).toEO();
```

and got the following exception:

```java
java.lang.IllegalArgumentException: Failed to transform by net.sf.saxon.jaxp.TransformerImpl: org.w3c.dom.DOMException: HIERARCHY_REQUEST_ERR: An attempt was made to insert a node where it is not permitted.; SystemID: file:///org/eolang/parser/xmir-to-eo.xsl; Line#: 117; Column#: 19

	at com.jcabi.xml.XSLDocument.transformInto(XSLDocument.java:468)
	at com.jcabi.xml.XSLDocument.transform(XSLDocument.java:420)
	at com.yegor256.xsline.StXSL.lambda$new$1(StXSL.java:45)
	at com.yegor256.xsline.StLambda.apply(StLambda.java:115)
	at com.yegor256.xsline.StEnvelope.apply(StEnvelope.java:61)
	at com.yegor256.xsline.StEnvelope.apply(StEnvelope.java:61)
	at com.yegor256.xsline.Xsline.pass(Xsline.java:85)
	at org.eolang.parser.xmir.Xmir$Default.toEO(Xmir.java:108)
	at it.AgentsIT.verifiesAgents(AgentsIT.java:60)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
[...]

PR #3344, opened 2024-08-20
feat(#3343): Allow Partial XMIR Printing

volodya-lombrozo:
In this PR I added `Xmir.Simplified` that can print partial XMIR documents to EO.

Closes: #3343.
History:
- **feat(#3343): add Simplified printer to XMIR**
- **feat(#3343): fix all qulice suggestions**

---

## PR-Codex overview
The PR introduces a new `Simplified` class in `Xmir` for partial XMIR printing. It includes tests and XSLT transformations for simplified EO printing.

### Detailed summary
- Added `Simplified` class for partial XMIR printing in `Xmir`
- Implemented tests for `Simplified` and `Default` classes
- Added XSLT transformation for simplified EO printing

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/xmir/Xmir.java
+++ b/eo-parser/src/main/java/org/eolang/parser/xmir/Xmir.java
@@ -110,4 +110,59 @@ public interface Xmir {
                 .get(0);
         }
     }
+
+    final class Simplified implements Xmir {
+
+        /**
+         * Train of transformations.
+         */
+        private static final Train<Shift> TRAIN = new TrDefault<>(
+            new StClasspath("/org/eolang/parser/explicit-data.xsl"),
+            new StUnhex(),
+            new StClasspath("/org/eolang/parser/wrap-method-calls.xsl")
+        );
+
+        /**
+         * Default xmir-to-eo XSL transformation.
+         */
+        private static final String TO_EO = "/org/eolang/parser/xmir-to-eo-simplified.xsl";
+
+        /**
+         * The XML.
+         */
+        private final XML xml;
+
+        /**
+         * Result to-EO transformation.
+         */
+        private final String xsl;
+
+        /**
+         * Ctor.
+         * @param src The source
+         */
+        public Simplified(final XML src) {
+            this(src, Xmir.Simplified.TO_EO);
+        }
+
+        /**
+         * Ctor.
+         * @param src The source
+         * @param classpath To-EO transformation classpath
+         */
+        public Simplified(final XML src, final String classpath) {
+            this.xml = src;
+            this.xsl = classpath;
+        }
+
+        @Override
+        public String toEO() {
+            final XML pass = new Xsline(
+                Simplified.TRAIN.with(
+                    new StClasspath(this.xsl)
+                )
+            ).pass(this.xml);
+            return pass.xpath("eo/text()").get(0);
+        }
+    }
 }
--- a/eo-parser/src/main/java/org/eolang/parser/xmir/Xmir.java
+++ b/eo-parser/src/main/java/org/eolang/parser/xmir/Xmir.java
@@ -34,6 +34,7 @@ import org.eolang.parser.StUnhex;
 /**
  * Prints XMIR to EO.
  * @since 0.35.0
+ * @checkstyle JavadocStyleCheck (200 lines)
  */
 public interface Xmir {
     /**
--- a/eo-parser/src/main/java/org/eolang/parser/xmir/Xmir.java
+++ b/eo-parser/src/main/java/org/eolang/parser/xmir/Xmir.java
@@ -113,7 +113,33 @@ public interface Xmir {
 
     /**
      * Simplified xmir that allows printing of small parts EO.
-     *
+     * <p>
+     * This printer is important in cases when you need to print only some parts of EO
+     * without having the whole EO document.
+     * </p>
+     * <p>
+     * For example, instead of generating the full structure of EO document like the following
+     * {@code
+     * <program>
+     *    <objects>
+     *        <o base=".plus">
+     *           <o base="int" data="bytes">00 00 00 00 00 00 00 01</o>
+     *           <o base="int" data="bytes">00 00 00 00 00 00 00 02</o>
+     *        </o>
+     *    </objects>
+     * </program>
+     * }
+     * You can generate only the part of the EO document that you need like the following
+     * {@code
+     * <o base=".plus">
+     *     <o base="int" data="bytes">00 00 00 00 00 00 00 01</o>
+     *     <o base="int" data="bytes">00 00 00 00 00 00 00 02</o>
+     * </o>
+     * }
+     * </p>
+     * It significantly simplifies the process of printing EO, especially in tests.
+     * You can read more about the problem right
+     * <a href="https://github.com/objectionary/eo/issues/3343">here</a>
      * @since 0.40
      */
     final class Simplified implements Xmir {
--- a/eo-parser/src/main/java/org/eolang/parser/xmir/Xmir.java
+++ b/eo-parser/src/main/java/org/eolang/parser/xmir/Xmir.java
@@ -111,6 +111,11 @@ public interface Xmir {
         }
     }
 
+    /**
+     * Simplified xmir that allows printing of small parts EO.
+     *
+     * @since 0.40
+     */
     final class Simplified implements Xmir {
 
         /**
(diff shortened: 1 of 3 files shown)


volodya-lombrozo:
@maxonfjvipon Could you review that changes, please?

volodya-lombrozo:

volodya-lombrozo:
@yegor256 friendly reminder

volodya-lombrozo:
@yegor256 friendly reminder

eo-parser/src/main/java/org/eolang/parser/xmir/Xmir.java
```diff
@@ -110,4 +110,60 @@ public String toEO() {
                 .get(0);
         }
     }
+
+    /**
+     * Simplified xmir that allows printing of small parts EO.
```
yegor256, line 115: @volodya-lombrozo it's not clear, why `Xmir` printing is not enough. Why do we need another printer?

yegor256 (changes requested):
@volodya-lombrozo the intent is not clear. Would be great to add some explanation to the Javadoc.

volodya-lombrozo:
@yegor256 Could you take a look one more time, please?

volodya-lombrozo:
@yegor256 reminder

yegor256:
@volodya-lombrozo even though the idea looks simple and straight-forward, I'm not convinced. If you use only a single `<o>` object, you loose meta information (like `+alias`), which is important for a EO program. If we let users do this, we lose the ability to check the correctness of their EO programs. This looks like a temporary hack, but not a "standard" approach, which we should ship together with other EO-related tools, I believe.

volodya-lombrozo:
@yegor256 I constantly face this problem when I need to print a single `eo` object instead of printing the entire `eo` program. (It is the third time already.) For me, it's a big deal and a constant pain in the... neck. Maybe we can create another interface or a separate implementation for printing? But in this case, it will be "object printing," not "eo printing." The difference lies in a separate interface and a completely different purpose.

volodya-lombrozo:
@yegor256 It's rather important changes that blocks other issues

volodya-lombrozo:
@yegor256 reminder

yegor256:
@volodya-lombrozo why can't you use a normal printing? Because you want to minimize the size of the XMIR, getting rid of the header part?

volodya-lombrozo:
@yegor256 yes. When you need to print ~100 different objects and test them, the task of "wrapping" each case with boilerplate code becomes tedious.

volodya-lombrozo:
@yegor256 Reminder

yegor256:
@volodya-lombrozo the problem is this: if we allow printing just a part of XMIR, we loose some meta information and, what is more important, loose the ability to provide such information in the future. It's similar to allowing Java compiler to compile just a method of a class, not the entire class. Even though it's technically possible (in some cases), such a partial compiler would be a very strange feature, don't you think?

volodya-lombrozo:
@yegor256 Of course, regarding compilers, I agree — It would definitely be a strange feature indeed. Is `phi`-printing the compilation process?

[Wiki:](https://en.wikipedia.org/wiki/Compiler)
> The name "compiler" is primarily used for programs that translate [source code](https://en.wikipedia.org/wiki/Source_code) from a [high-level programming language](https://en.wikipedia.org/wiki/High-level_programming_language) to a [low-level programming language](https://en.wikipedia.org/wiki/Lower_level_language)

I thought it is the representation of the same `xmir` on the same level. Am I right?  At least the "printing" suggests that it is just a rewriting. In other words, we just have a different representation, and we don't compile anything, do we? So, if we have this representation, why can't we have another one that would only help?

I consider this feature as an alternative "short view" on some parts of `xmir`. Of course this "short" representation doesn't have to be used further in any compilation pipelines. So, it's exactly why I [suggested](https://github.com/objectionary/eo/pull/3344#issuecomment-2324063539)  to use different interface for this class:

> Maybe we can create another interface or a separate implementation for printing? But in this case, it will be "object printing," not "eo printing." The difference lies in a separate interface and a completely different purpose.

As for this:

>  if we allow printing just a part of XMIR, we loose some meta information and, what is more important, loose the ability to provide such information in the future. 

Why do we need this information in the future? The only purpose for this partial printing is testing, no more.

yegor256:
@volodya-lombrozo the printing may expect the entire XMIR document to be available in order to print EO. If we allow printing of just a part of the XMIR, we introduce some other assumptions, which may not be in line with the previous one. Thus, we will have to maintain them both, not always understanding what to do if they conflict. For example, this pseudo XMIR:

```
<xmir>
  <meta>abc</meta>
  <objects>
    <o id=1/>
  </objects>
</xmir>
```

It prints to (using both the info from the `<meta>` and the `<o>` elements):

```
OBJ 1/abc
```

You introduce just-one-object printer, which will produce (for example):

```
OBJ 1/foo
```

Here, you will have to replace `abc` with `foo`, because you don't have `<meta>` available.

Now, we have two EO outputs for the same input. How they correlate? Hard to say. How to maintain this correlation? Hard to say.

volodya-lombrozo:
@yegor256 Sorry, maybe I don't clearly understand you. If you're asking about the correlation between 'object' and 'file' printing (how to update both of these xslt transformations), then it's a separate question, but I'm sure we can solve this easily in the code.

As for this:

we introduce some other assumptions, which may not be in line with the previous ones
I didn't clearly understand which assumptions you mean.

```xml
<o id=1/>
```

will give me

```
OBJ 1
```

Why this?

```
OBJ 1/foo
```

Moreover, it's a problem of 'object' printing: how to retrieve some missing elements. As I've already said many times, 'object' printing is suitable only for testing, so we can't rely on it in pipelines or make any assumptions based on it.

yegor256:
@volodya-lombrozo the `<o>` element is _not_ the only place where the information about a particular object is stored. Some information may be in other places of the XMIR document. This is what I mean.

volodya-lombrozo:
> @volodya-lombrozo the `<o>` element is _not_ the only place where the information about a particular object is stored. Some information may be in other places of the XMIR document. This is what I mean.

@yegor256 Yes, I understood that. When we print only a single object without a header - we use only the information we have, of course. I believe it's fine in the context of testing.

yegor256:
@volodya-lombrozo yes, for testing it's fine, in your own project, where you understand the context and the risks. But shipping this functionality to everybody would be a mistake, I believe.

volodya-lombrozo:
@volodya-lombrozo If I move this changes to my own project:

1. I won't be able to use it across several projects. Even for now I need to use this functionality in two separate repositories.
2. It will be hard to keep consistency between "program" printing and "object" printing - when they are located at one place - it's much easier to "inherit" some functionality. For example "program" printing might use "object" printing script and vice versa.

yegor256:
@volodya-lombrozo why can't you print this:

```
<xmir>
<objects>
<o base=".plus">
  <o base="int" data="bytes">00 00 00 00 00 00 00 01</o>
  <o base="int" data="bytes">00 00 00 00 00 00 00 02</o>
</o>
</objects>
</xmir>
```

instead of:

```
<o base=".plus">
  <o base="int" data="bytes">00 00 00 00 00 00 00 01</o>
  <o base="int" data="bytes">00 00 00 00 00 00 00 02</o>
</o>
```

It's a question of these "extra" four lines?

volodya-lombrozo:
@yegor256 Let's imagine I have 100 java objects that can create  only some parts of the entire program. For example `Plus.java` can create the following `xmir`.

```xml
<o base=".plus">
  <o base="int" data="bytes">00 00 00 00 00 00 00 01</o>
  <o base="int" data="bytes">00 00 00 00 00 00 00 02</o>
</o>
```
It doesn't know anything about the entire `xmir` structure. Just this. So, if I need to test this class, I need to "wrap" it with extra lines. Also I have to do it for all the rest 99 classes. It's tedious. Of course, I can write wrappers, matchers and so on, but I will have [the same problems](https://github.com/objectionary/eo/pull/3344?notification_referrer_id=NT_kwDOAxZ4wbQxMjAyNDA5MTY3MDo1MTgwNDM1Mw#issuecomment-2360552756) which I have already mentioned.

volodya-lombrozo:
@yegor256 Sometimes I need to do it manually, like [here](https://github.com/objectionary/opeo-maven-plugin/blob/cae8640b31e5b34c5d9bc59f0178f8b10a4455e2/src/test/java/it/AgentsIT.java#L80):

```java
new Xmir.Default(
    new XMLDocument(
        new Xembler(
            new Directives().add("program").add("objects")
            .append(
                new DecompilerMachine(Collections.singletonMap("output", output))
                .decompile(pack.instructions())
            )
            .up().up()
        ).xmlQuietly()
    )
).toEO()
```

volodya-lombrozo:
@yegor256 Reminder

volodya-lombrozo:
@yegor256 Friendly reminder

volodya-lombrozo:
Since there are many concerns about this functionality, I close this PR

PR #3390, opened 2024-09-26
Rel printing fixed

yegor256:
Currently, it prints long paths, which is a mistake. This PR should fix this.


0crat:
@yegor256 Hey there! 👋 I noticed you named your branch "`rel`" - not quite following our naming convention. Remember, we're supposed to use ticket numbers for branch names (like "`3389`" in this case). [Our guidelines](https://www.yegor256.com/2014/04/15/github-guidelines.html) explain why this is important. As per our policy, there's a -12 point penalty for non-compliant names. This brings your balance to [-43](https://www.eolang.org/zerocracy/objectionary-vitals.html). Let's aim for better naming next time to avoid penalties and keep our workflow smooth! 🚀

Issue #3421, opened 2024-10-14
`PhDefault.vertex` seems to be a deprecated feature


yegor256:
What's the point of keeping the `PhDefault.vertex` attribute if we don't have access to it from the language? I suggest we delete it entirely.

Also, we can get rid of our custom `PhDefault.hashCode()` and simply rely on the default implementation.

yegor256:
@maxonfjvipon WDYT?

PR #3449, opened 2024-10-30
fix(#3421): remove vertex

maxonfjvipon:
Closes: #3421

---

## PR-Codex overview
This PR focuses on removing the `Vertices` dependency from the `PhDefault` class and replacing its usage with the object's `hashCode()`. This simplifies the class by eliminating the `vertex` field and related logic.

### Detailed summary
- Removed the `Vertices` static field and its initialization.
- Deleted the `vertex` field from the class.
- Updated the constructor to remove `vertex` initialization.
- Changed the `hashCode()` method to return `super.hashCode() + 1`.
- Updated `toString()` method to use `hashCode()` instead of `vertex`.
- Modified the `copy()` method to eliminate `vertex` assignment.


eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -117,8 +105,9 @@ public boolean equals(final Object obj) {
     }
 
     @Override
+    @SuppressWarnings("PMD.UselessOverridingMethod")
```
yegor256, line 108: @maxonfjvipon maybe you can just delete the method?)

maxonfjvipon: @yegor256 PMD wants us to have either both methods `equals` and `hashCode` or nothing. If we delete them both - object comparison in some test stops working, for example in `org.eolang.PhDefaultTest#doesNotChangeRhoAfterDirectKidCopying`

yegor256: @maxonfjvipon implement it like this and remove the warning:

```
return super.hashCode() + 1;
```

PR #3652, opened 2024-12-12
StrictXmir

### LINKED ISSUE #3636: Failed to read schema document `https://www.eolang.org/xsd/XMIR-0.46.0.xsd`
Recently [one of the pipelines](https://github.com/objectionary/jeo-maven-plugin/actions/runs/12258856491/job/34199619501?pr=942) in `jeo-maven-plugin` failed with the following error: 

```xml
-1:-1: schema_reference.4: Failed to read schema document 'https://www.eolang.org/xsd/XMIR-0.46.0.xsd', because 1) could not find the document; 2) the document could not be read; 3) the root element of the document is not <xsd:schema>.
```
[build.log](https://github.com/user-attachments/files/18092081/build.log)

Moreover, it failed only on `ubuntu` system. On `macos` everything is fine.

yegor256:
related to #3636


eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
```diff
@@ -0,0 +1,181 @@
...
+import org.cactoos.io.InputOf;
+import org.w3c.dom.Node;
+import org.xembly.Directives;
+import org.xembly.Xembler;
+import org.xml.sax.SAXParseException;
+
+/**
+ * XMIR that validates itself right after construction.
+ *
+ * <p>This class is supposed to be used ONLY for testing, because
```
volodya-lombrozo, line 51: @yegor256 I'm a bit confused here. You mentioned in the PR description that this PR closes https://github.com/objectionary/eo/issues/3636. And here, in the javadoc you write 

```
This class is supposed to be used ONLY for testing
```

It means, I won't be able to use `StrictXmir` in `jeo-maven-plugin` production code. However, I'm using `StrictXML` during the `xmir` generation. In other words, I use it in the production code.

eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
```diff
@@ -0,0 +1,181 @@
...
+            }
+            try {
+                Files.write(
+                    path,
+                    new IoCheckedBytes(
+                        new BytesOf(new InputOf(new URI(uri)))
+                    ).asBytes()
+                );
+            } catch (final IOException | URISyntaxException ex) {
+                throw new IllegalArgumentException(ex);
```
volodya-lombrozo, line 176: @yegor256 [NIT] It's a bad idea by a several reasons:
1) Exception without context: https://www.yegor256.com/2015/12/01/rethrow-exceptions.html
2) Grouped exceptions: https://www.yegor256.com/2022/08/30/dont-group-exception-catchers.html

eo-parser/src/test/java/org/eolang/parser/StrictXmirTest.java
```diff
@@ -0,0 +1,64 @@
...
+            new StrictXmir(
+                new Xmir(
+                    new XMLDocument(
+                        new Xembler(
+                            new Directives()
+                                .append(new DrProgram("foo"))
+                                .xpath("/program")
+                                .attr(
+                                    "noNamespaceSchemaLocation xsi http://www.w3.org/2001/XMLSchema-instance",
+                                    "https://www.eolang.org/XMIR.xsd"
```
volodya-lombrozo, line 53: @yegor256 You tested only one positive scenario.  What will happen if we have a corrupted address? What will happen if we don't have the internet? 
```
https://www.eCORRUPTEDolang.org/XMIR.xsd
```

eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
```diff
@@ -0,0 +1,181 @@
...
+                    "file:///%s",
+                    StrictXmir.download(
+                        uri,
+                        Paths.get("target/xsd").resolve(
+                            uri.substring(uri.lastIndexOf('/') + 1)
+                        )
+                    ).toString().replace("\\", "/")
+                );
+            }
+            new Xembler(
```
volodya-lombrozo, line 146: @yegor256 Do we need to change the `xmir` in case `!uri.startWith("http")`?

volodya-lombrozo (changes requested):
@yegor256 This PR doesn't solve the https://github.com/objectionary/eo/issues/3636 issue. It only speeds up the inner `eo` tests. If we don't have an internet connection or there is an outage, this solution won't help.

yegor256:
@volodya-lombrozo we can't stay entirely offline. The schema (XSD) must come from somewhere. Either you download it and then provide to `StrictXmir` (I will make it possible now), or it will download it. Or I miss something?

volodya-lombrozo:
> @volodya-lombrozo we can't stay entirely offline. The schema (XSD) must come from somewhere. Either you download it and then provide to `StrictXmir` (I will make it possible now), or it will download it. Or I miss something?

@yegor256 Is it possible to place `xsd` to the classpath where `StrictXmir` is located?

yegor256:
@volodya-lombrozo we don't know which version to place there: there are many of them published at www.eolang.org. We can possibly try to download them all during build and put into JAR, but this sounds like a hack.

eo-parser/src/test/java/org/eolang/parser/StrictXmirTest.java
yegor256: @volodya-lombrozo added test in 6f07f8f9a

eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
yegor256: @volodya-lombrozo fixed

eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
yegor256: @volodya-lombrozo you can try to do this, on your side: 1) create `XML` object, 2) save it on disc, 3) make `StrictXmir(xml)` and 4) either fail, or continue. If you fail, you fail. If you continue, you have a clean XML on disc.

eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
yegor256: @volodya-lombrozo nope, when it's `file://....` we keep it as is

eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
volodya-lombrozo: @yegor256 [NIT] I mean, this class might be used not only for testing. Actually it's an important part of validation in a production code. Maybe we need to remove this line from javadoc:

> This class is supposed to be used ONLY for testing

and replace it with something like 

> ATTENTION!

 what do you think?

eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
volodya-lombrozo: @yegor256 I see. I've overlooked indentations. Thank you.

volodya-lombrozo (commented):
@yegor256 Looks good to me. I suggested several minor changes, but they aren't so important. Could you also replace the line in the comment from 

> closes https://github.com/objectionary/eo/issues/3636

to 

> related to https://github.com/objectionary/eo/issues/3636 

please?

volodya-lombrozo:
> @volodya-lombrozo we don't know which version to place there: there are many of them published at [www.eolang.org](http://www.eolang.org). We can possibly try to download them all during build and put into JAR, but this sounds like a hack.

@yegor256 As I understand it, you publish them from [XMIR.xsd](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/resources/XMIR.xsd). So, when you publish version `0.47.0` of `eo`, you also publish version `0.47.0` of `XMIR.xsd` (at least that’s my assumption). In other words, they are consistent and included together in `eo-parser:0.47.0.jar`. I don't clearly understand why we can't use the `XMIR` from this `JAR` and instead are required to download it from the internet. Most probably, I will implement this in `jeo-maven-plugin`.

yegor256:
@volodya-lombrozo when you do `new StrictXmir(xml)`, we don't know what is the location of XSD that you mention inside the `xml`. It could be the version of the JAR where `StrictXmir.class` is located, but it may me something totally different (not even `XMIR.xsd`, but any other URI). That's why we have to download the file first and then use it. Make sense?

eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
yegor256: @volodya-lombrozo explained better in 04af5af65

yegor256:
@volodya-lombrozo what we can do is this: `StrictXmir` will compare the value of the URI inside XML with the expected location of XSD for the current version of eo-parser. If the URI is exactly where the XSD is supposed to be located, it will use the `XMIR.xsd` file from classpath. In your particular case, this will solve the problem.

volodya-lombrozo:
> @volodya-lombrozo when you do `new StrictXmir(xml)`, we don't know what is the location of XSD that you mention inside the `xml`. It could be the version of the JAR where `StrictXmir.class` is located, but it may me something totally different (not even `XMIR.xsd`, but any other URI). That's why we have to download the file first and then use it. Make sense?

@yegor256 To be honest, I don't see any reason why we might need such flexibility. We have a single `XSD` and a single `StrictXmir`. They are placed together in one location, so `StrictXmir` can directly use it without any downloading. This sounds like overengineering to me in this particular case. However, I may not be seeing the whole picture. 

I will try to use `xsi:noNamespaceSchemaLocation="classpath:XMIR.xsd"` in the generated `XML` files or something similar. I don't know if it will help, but that's a separate matter. I will investigate it.

yegor256:
@volodya-lombrozo 

> We have a single XSD and a single StrictXmir. They are placed together in one location, so StrictXmir can directly use it without any downloading.

`StrictXmir` doesn't know about this: whether the XSD your XML is using is the same XSD that StrictXmir has in the classpath. There is no such guarantee.

yegor256:
@volodya-lombrozo you can try to use this:

```
xsi:noNamespaceSchemaLocation="classpath:XMIR.xsd"
```

However, 1) this won't work with `StrictXML` or `StrictXmir` and you will have to implement your own schema validation, and 2) the XML file won't be accepted by next steps of the pipeline, since they have their own validators that don't know anything about your classpath.

volodya-lombrozo (approved):

volodya-lombrozo:
> @volodya-lombrozo you can try to use this:
> 
> ```
> xsi:noNamespaceSchemaLocation="classpath:XMIR.xsd"
> ```
> 
> However, 1) this won't work with `StrictXML` or `StrictXmir` and you will have to implement your own schema validation, and 2) the XML file won't be accepted by next steps of the pipeline, since they have their own validators that don't know anything about your classpath (some of them may not be even written in Java).

@yegor256 
There are many questions:

1. Why do the "next steps of the pipeline" require an `xsd`, and why do they specifically need the `http` version of this `xsd`?
2. Moreover, I don't understand why the `classpath` won't work with `StrictXML` or `StrictXmir`.

Let me investigate these questions. Maybe it will become clear. And thank you very much for your help.

yegor256:
@volodya-lombrozo check this one: d0bf3b82b With this feature you will be fine. Just use `StrictXmir` from eo-parser and there will be no network usage.

yegor256:
@volodya-lombrozo XML documents that JEO (or any other tool) produces are supposed to be "valid" in terms of compliance with XSD schema. Technically this means that `noNamespaceSchemaLocation` is present and it points to the document publicly available (online or on disc). Classpath is not publicly available, it's a local storage of a particular instance of a running VM.

volodya-lombrozo (commented):
@yegor256 There is one common part in all tests:

```java
           new StrictXmir(
                 new Xmir(
                     new XMLDocument(
                         new Xembler(
                             new Directives()
                                 .append(new DrProgram("foo3"))
                                 .xpath("/program")
                                 .attr(
                                     "noNamespaceSchemaLocation xsi http://www.w3.org/2001/XMLSchema-instance",
                                     String.format(
                                         "https://www.eolang.org/xsd/XMIR-%s.xsd",
                                         Manifests.read("EO-Version")
                                     )
                                 )
                                 .add("objects")
                         ).xml()
                     )
                 ),
                 tmp
             ).validate(),
```

What do you think if we will move into a separate method?

yegor256:
@volodya-lombrozo you are right, duplication reduced: 809adde0a

volodya-lombrozo:
> Technically this means that `noNamespaceSchemaLocation` is present and it points to the document publicly available (online or on disc)

@yegor256 As far as I know, we have several ways to validate an XML document against an XSD schema, and we are not obligated to use only `noNamespaceSchemaLocation`.

volodya-lombrozo (approved):

yegor256:
@volodya-lombrozo indeed, it's possible to validate it against any XSD schema, but relying on `noNamespaceSchemaLocation` or `schemaLocation` (which are provided by the author of XML document) is the more robust way. BTW, we can also use DTD: https://stackoverflow.com/questions/1544200/what-is-difference-between-xml-schema-and-dtd

PR #3667, opened 2024-12-14
fix codenarc P2/P3 violations 

coutvv:
I had fixed here all P2 and P3 CodeNarc violations and enable zero tolerance to the new one.

Rules of CodeNarc:
https://codenarc.org/codenarc-rules-formatting.html
https://codenarc.org/codenarc-rules-convention.html

---

Also here I have disabled next CodeNarc rules:
1. **UnnecessaryReturnKeyword** - because it conflict with other rule **ImplicitReturnStatement**. 
2. **SpaceAroundMapEntryColon** - resulted code looks so ugly, so I decided to disable this rule. Just look at it: 
```groovy
new File('.').traverse(
                type:FileType.FILES,
                preDir:{ file ->
                    if (file.name == 'target') {
                        return FileVisitResult.SKIP_SUBTREE
                    }
                },
                nameFilter:~/.*\.xsl/,
 )
```

If you think that I should prefer **UnnecessaryReturnKeyword** instead of **ImplicitReturnStatement** rule or if **SpaceAroundMapEntryColon** should be used - let me know.

### DIFF
--- a/.codenarc
+++ b/.codenarc
@@ -220,7 +220,6 @@ ruleset {
     SpaceAfterSwitch
     SpaceAfterWhile
     SpaceAroundClosureArrow
-    SpaceAroundMapEntryColon
     SpaceAroundOperator
     SpaceBeforeClosingBrace
     SpaceBeforeOpeningBrace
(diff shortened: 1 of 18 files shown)


yegor256:
@coutvv excellent contribution, many thanks!

PR #3697, opened 2024-12-18
Fixed user instruction

iri:
Also workaround until warnings issue will be fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ All of them have something **we don't tolerate**:
 
 First, install [Java SE](https://www.oracle.com/java/technologies/downloads/),
 [npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm),
-and [eoc](https://github.com/objectionary/eoc).
+and [eoc](https://github.com/objectionary/eoc) (`eoc` is part of [eolang][npm] NPM package).
 
 Then, start with a simple EO program in `app.eo` file:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,8 +63,13 @@ All of them have something **we don't tolerate**:
 ## Quick Start
 
 First, install [Java SE](https://www.oracle.com/java/technologies/downloads/),
-[npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm),
-and [eoc](https://github.com/objectionary/eoc).
+[npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm).
+
+Then, you install [eoc](https://github.com/objectionary/eoc). You could install it as NPM package:
```
yegor256, line 68: @iri looks like a duplication of information from the README of the `eoc` package, which in general is not a good idea: if they change the instructions, we will most definitely forget to change it here.

iri: @yegor256 I agree, from point of view of data normalization "there should not be duplicates!". This is good for data updating, when the target is, for instance: "there should be minimal workload for eoc maintainer in the future".

But also there could another target -- "to maximize flow of new 'eoc' users who try at least 'hello' example with positive outcome".
From this point it is better that new user is not obligated to do addition clicks to another pages as it increases probability that user will no leave at this point. The fewer clicks a user needs to make to access the information, and the less they "have to think" while doing it, the more likely they are to complete the instructions and not leave.
There is even a principle of building UIs, user instructions etc., and a UI book with the same title -- "Don't make me think". Main point in it is that there should be minimum "uncertainty points" (ideally -- none) in the path of a user, where he must stop and think (while he/she is working with UI -- screen forms, user instructions etc.).

It's obvious that the second target leads to data normalization.

Both targets are conflicting with each other and decision should be made which of them is more important.

If the first target ir more important than the second, I will remove the change above.

iri: @yegor256 ah, the change is aleady rejected. Ok then. :)

yegor256:
@iri ok, make sense

Issue #3698, opened 2024-12-18
Remove current implementation of objection versioning


maxonfjvipon:
Now we have a lot of kind of dead code related to object versioning in `eo-parser` and `eo-maven-plugin`. 
1. There are so much code in `eo-maven-plugin` that must be refactored. 
2. Since the object versioning is put on hold, it may create difficulties for refactoring. 
3. Also maybe the current implementation of object versioning should be refactored as well. 

My suggestion is to remove the current functionality of object versioning. It'll speed up the build, EO grammar and `eo-maven-plugin` itself.

Then, when EO is stable, `eo-maven-plugin` is well designed and refactored and we're totally ready to introduce object versioning - try to implement it one more time in a good way.

@yegor256 WDYT?

yegor256:
@maxonfjvipon I agree

PR #3699, opened 2024-12-18
bug(#3698): Remove object versioning

maxonfjvipon:
Closes: #3698


maxonfjvipon:
@yegor256 please check

maxonfjvipon:
@yegor256 please check

maxonfjvipon:
@yegor256 
1. they are already disabled for a long time (almost a year) and everything works fine 
image
2. some of them are not relevant anymore, for example `rewritten_sources` or `hash_package_layer`
3. some of them are already implemented as unit tests
4. we have `Jaxec` now, so if something goes wrong in the future - we can use it instead

yegor256:
@maxonfjvipon well, it's a good practice (and a very very common practice) to use invoker for Maven plugin testing. It's just something people always do in Maven world. I suggest keeping one of them in place (fibonacci, for example) and getting rid of all others. WDYT?

maxonfjvipon:
@yegor256 please check one more time, I fixed `fibonacci` integration test and enabled it

yegor256:
@maxonfjvipon perfect, thanks!

PR #3723, opened 2024-12-20
may have fixed the race

sunmisc:
If these objects are used inside Atomic, we can assume that they must be thread-safe.
You can't do a check/write, because during the check, the state may change and there will already be an object there. Using synchronization primitives such as:
Acquire/Release (volatile read/write) and CAS you can avoid this, only this way and no other way (or use a locking mechanism that is high level)

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/AtRho.java
+++ b/eo-runtime/src/main/java/org/eolang/AtRho.java
@@ -61,23 +61,17 @@ final class AtRho implements Attr {
 
     @Override
     public Phi get() {
-        if (this.rho.get() == null) {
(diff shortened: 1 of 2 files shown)


yegor256:
@sunmisc many thanks for the contribution! It looks reasonable. However, it would be great to have a unit test that confirms that our current implementation has a bug. Can you please add a test? Also, check this please: https://www.yegor256.com/2024/12/15/open-source-beginner-advice.html

yegor256:
@sunmisc it seems that this problem is already fixed, I'm closing. Don't hesitate to submit a new PR, if you think changes are still needed. Please, do remember to accompany your changes with a test.

PR #3730, opened 2024-12-23
bug(#3711) -DskipTests skip maven-invoker-plugin invocation 

### LINKED ISSUE #3711: maven-invoker-plugin tests are not skipped with `-DskipTests`
This is the [default behavior](https://maven.apache.org/plugins/maven-invoker-plugin/examples/skipping.html), but it will be more convenient if in our project we skip invoker when `skipTests` property is set.

ArtemGet:
Closes #3711


yegor256 (approved):
@ArtemGet looks correct, thanks!

PR #3751, opened 2024-12-24
feat(#3742): Split `ParsingErrors` on Two Separate Classes

volodya-lombrozo:
In this PR I:
1. Createad a separate package `errors` for all error handlers and realted classes.
2. Split `ParsingErrors` on `EoParserErrors` and `ParsingErrors`
3. Added supplementary classes like `Lines` and `ErrorDirectives` to share common logic across all the new classes.

When this PR seems having a lot of lines, it just moves some parts of the code across the codebase.

Closes: #3742.


volodya-lombrozo:
@yegor256 Could you review these changes, please?

eo-parser/src/main/java/org/eolang/parser/errors/ErrorDirectives.java
```diff
@@ -0,0 +1,70 @@
...
+import org.cactoos.iterable.Mapped;
+import org.eolang.parser.ParsingException;
+import org.xembly.Directive;
+import org.xembly.Directives;
+
+/**
+ * Error directives.
+ * @since 0.50
+ */
+final class ErrorDirectives implements Iterable<Directive> {
```
yegor256, line 37: @volodya-lombrozo we already have `DrProgram` and `DrListing` classes. Maybe this one can follow the same convention?

eo-parser/src/main/java/org/eolang/parser/errors/EoParserErrors.java
```diff
@@ -36,22 +35,15 @@
...
- * @since 0.30.0
- * @todo #3706:30min Split {@link ParsingErrors} into two classes.
- *  Currently we use the same {@link ParsingErrors} class to accumulate all the parsing errors
- *  despite their origin. This class should be split into two classes: one for parsing errors
- *  {@link ParserErrors} and another for lexer errors {@link LexerErrors}.
+ * Accumulates all parsing errors related to EO parser.
+ * @since 0.50
  */
-final class ParsingErrors extends BaseErrorListener implements Iterable<Directive> {
+public final class EoParserErrors extends BaseErrorListener implements Iterable<Directive> {
```
yegor256, line 46: @volodya-lombrozo same here, we have `Dr` prefix for class that are iterables of `Directive`. Maybe can use it too?

eo-parser/src/main/java/org/eolang/parser/errors/Lines.java
```diff
@@ -0,0 +1,66 @@
...
+    Lines(final List<Text> lines) {
+        this.source = lines;
+    }
+
+    /**
+     * Get the line by number.
+     * @param number The line number.
+     * @return The line.
+     */
+    String line(final int number) {
```
yegor256, line 55: @volodya-lombrozo isn't it the same as `lines.get(i).orElse("")`? Looks like this class is an overkill.

eo-parser/src/main/java/org/eolang/parser/errors/ParsingErrors.java
```diff
@@ -0,0 +1,114 @@
...
+import org.cactoos.list.ListOf;
+import org.eolang.parser.ParsingException;
+import org.xembly.Directive;
+
+/**
+ * Accumulates all parsing errors.
+ *
+ * @since 0.30.0
+ */
+public final class ParsingErrors extends BaseErrorListener implements Iterable<Directive> {
```
yegor256, line 42: @volodya-lombrozo same here: if it's an iterable of `Directive`, the `Dr` prefix will be a good name pattern, I believe

yegor256: @volodya-lombrozo also, implementation inheritance is not what we like to use. Composition is a better choice usually. See https://www.yegor256.com/2016/09/13/inheritance-is-procedural.html

yegor256 (changes requested):
@volodya-lombrozo looks interesting, a few comment above

eo-parser/src/main/java/org/eolang/parser/errors/ErrorDirectives.java
volodya-lombrozo: @yegor256 Renamed to `DrErrors`

eo-parser/src/main/java/org/eolang/parser/errors/EoParserErrors.java
volodya-lombrozo: @yegor256 Renamed to `DrEoParserErrors`

eo-parser/src/main/java/org/eolang/parser/errors/ParsingErrors.java
volodya-lombrozo: @yegor256 Renamed to `DrParsingErrors`.

volodya-lombrozo: > @volodya-lombrozo also, implementation inheritance is not what we like to use. Composition is a better choice usually. See https://www.yegor256.com/2016/09/13/inheritance-is-procedural.html

I would like to avoid inheritance, but `BaseErrorListener` is an ANTLR class and it's the way how they handle errors.

eo-parser/src/main/java/org/eolang/parser/errors/Lines.java
volodya-lombrozo: @yegor256 In case of using `lines.get(i).orElse("")` - you will get NPE. 

Basically, I need to share the same logic between `DrEoParserErrors` and `DrParsingErrors` classes (the `String line(final int number)` method). I can't use `static` method by obvious reasons. Thus, I have left with only one option - to create a `Lines` object. Do you know any other way how to share the logic and avoid creating `Lines` class?

volodya-lombrozo:
@yegor256 Could you have a look one more time, please?

yegor256:
@volodya-lombrozo the naming of classes is rather confusing here: they look almost the same. Why don't you use the names suggested in the puzzle: `ParserErrors` and `LexerErrors`?

volodya-lombrozo:
> @volodya-lombrozo the naming of classes is rather confusing here: they look almost the same. Why don't you use the names suggested in the puzzle: `ParserErrors` and `LexerErrors`?

@yegor256 I agree that the naming here is confusing. Let me try to explain. Before that PR, we had only one error listener: `ParsingErrors`, and that's all. I thought it would be fine to split it into two separate parts—`LexerErrors` and `ParserErrors`. When I started implementing this PR, I noticed that you were using the same `ParsingErrors` for both `EoSyntax` and `PhiSyntax`, which works fine until we implement language-specific error messages. So, instead of `LexerErrors` and `ParserErrors`, I would have to create `EoLexerErrors`, `EoParserErrors`, `PhiLexerErrors`, and `PhiParserErrors`. That would result in too many classes (and I wouldn't be able to sell that idea to you! 😊). As a result, I decided to leave `ParsingErrors` as it is and simply add one more class: `DrEoParserErrors`, specifically for EO errors. Please feel free to ask questions if anything is unclear.

volodya-lombrozo:
@yegor256 Recent changes:

1. Renamed `ParsingErrors` -> `GeneralErrors` (?)
2. Removed `package`
3. `GeneralErrors` and `EoParserErrors` don't implement `Iterable<Directive>` anymore. They create directives through the `directive()` method instead.

eo-parser/src/main/java/org/eolang/parser/GeneralErrors.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * How many errors?
+     * @return Count of errors accumulated
+     */
+    public int size() {
+        return this.errors.size();
+    }
+
+    /**
```
yegor256, line 108: @volodya-lombrozo this method is here by mistake?

eo-parser/src/main/java/org/eolang/parser/PhiSyntax.java
```diff
@@ -107,7 +107,7 @@ public XML parsed() throws IOException {
         final XML dom = Syntax.CANONICAL.pass(
             new XMLDocument(
                 new Xembler(
-                    new Directives(xel).append(spy).append(this.extra)
+                    new Directives(xel).append(spy.directives()).append(this.extra)
```
yegor256, line 110: @volodya-lombrozo looks like it should be `.append(new DrErrors(spy))`, isn't it?

yegor256 (changes requested):
@volodya-lombrozo it's much better now! please, see my comments above

volodya-lombrozo:
@yegor256 Please review these changes: [fbbaa29](https://github.com/objectionary/eo/pull/3751/commits/fbbaa29e8b8c67eb40ea9862e3ea67a19dd578c1). I can accommodate your suggestions in this manner (it's the only way). If you don't like the solution, we can revert them.

eo-parser/src/main/java/org/eolang/parser/DrErrors.java
```diff
@@ -0,0 +1,68 @@
...
+     * Errors accumulated.
+     */
+    private final Errors errors;
+
+    /**
+     * Ctor.
+     * @param errors The errors.
+     */
+    DrErrors(final Errors errors) {
+        this.errors = errors;
```
yegor256, line 47: @volodya-lombrozo here, you only use `errors.all()` method here, don't use `size()`. Why can't you use `Iterable` instead of `Errors` here?

yegor256:
@volodya-lombrozo can't you simply calculate the size of `Iterable` with the help of `LengthOf` from Cactoos? This `Errors` class looks redundant here. It only complicates the design adding no value.

volodya-lombrozo:
@yegor256 Sure, please, check this one: https://github.com/objectionary/eo/pull/3751/commits/d4764a698da350c8338bd0f9eb5035c4f56d8c1a
I've done it. However, I don't prefer such a solution because we hide an abstraction, that we actually have. Although, I should admit, that we have less code now.

yegor256 (approved):
LGTM

yegor256:
@volodya-lombrozo thanks!

PR #3762, opened 2024-12-25
feat(#3756): Use a Single Format for All Error Messages

volodya-lombrozo:
Created one more class `LocationMessage` that is used as a default error message:
```
[2:0] error: 'Invalid program declaration'
```
It simply adds line and column where an error occured.

___
PS. Now we have two classes `UnderlinedMessage` and `LocationMessage`. They both have the same structure and similar methods. So, we can create an interface for them.
Moreover, we can change their names to `MsgUnderlined` and `MsgLocation`. I haven't done it in this PR in order to keep its size in some borders.
But we can change it during the PR review.

Closes: #3756.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LocationMessage.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LocationMessage.java
@@ -23,6 +23,10 @@
  */
 package org.eolang.parser;
 
+/**
+ * Error message that includes the location of the error.
+ * @since 0.50
+ */
 final class LocationMessage {
 
     /**
@@ -57,6 +61,6 @@ final class LocationMessage {
      * @return The formatted error message.
      */
     String formatted() {
-        return String.format("[%d:%d] error: \"%s\"", this.line, this.position, this.message);
+        return String.format("[%d:%d] error: '%s'", this.line, this.position, this.message);
     }
 }
--- a/eo-parser/src/main/java/org/eolang/parser/LocationMessage.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LocationMessage.java
@@ -23,63 +23,40 @@
  */
 package org.eolang.parser;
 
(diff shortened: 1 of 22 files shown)


volodya-lombrozo:
@yegor256 Could you review these changes, please?

eo-parser/src/main/java/org/eolang/parser/LocationMessage.java
```diff
@@ -0,0 +1,66 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.eolang.parser;
+
+/**
+ * Error message that includes the location of the error.
+ * @since 0.50
+ */
+final class LocationMessage {
```
yegor256, line 30: @volodya-lombrozo maybe `LocatedMessage`? In general, it's not a good idea to have two glued together nouns as a name of a class. It's unclear what is it, a "location" or a "message"? Also, maybe it's better to name them `MsgLocated` and `MsgUnderlined`?

yegor256 (changes requested):
@volodya-lombrozo looks good, just one comment above

volodya-lombrozo:
@yegor256 Renamed classes according with your suggestions: https://github.com/objectionary/eo/pull/3762/commits/829f33494917cabc0337112efdb342692c8f299f
Can you have a look one more time, please?

yegor256 (approved):
LGTM

yegor256:
@volodya-lombrozo thanks!

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +4 as a basis; -10 for very few (4) comments; +10 to give you at least something.  Your running balance is [-16](https://www.eolang.org/zerocracy/objectionary-vitals.html).

Issue #3708, opened 2024-12-19
`PhiMojo` and `UnphiMojo` does not use cache


maxonfjvipon:
If we run `xmir-to-phi` and `phi-to-xmir` a several times - new files are created every time. Let's make these mojos use cache

ArtemGet:
@maxonfjvipon Check referenced PRs, please. Caching logic ```mir-to-phi``` and ```phi-to-xmir``` would be included in next PRs.

maxonfjvipon:
@ArtemGet sure. Please note that we use `Footprint` class for caching, you can find examples of usage in other mojos

0pdd:
@maxonfjvipon the puzzle [#3804](https://github.com/objectionary/eo/issues/3804) is still not solved.

0pdd:
@maxonfjvipon 2 puzzles [#3804](https://github.com/objectionary/eo/issues/3804), [#3806](https://github.com/objectionary/eo/issues/3806) are still not solved.

0pdd:
@maxonfjvipon all 2 puzzles are solved here: [#3804](https://github.com/objectionary/eo/issues/3804), [#3806](https://github.com/objectionary/eo/issues/3806).

PR #3778, opened 2024-12-29
bug(#3708): Required tojo methods for caching xmir-to-phi/phi-to-xmir goals

ArtemGet:
Ref: #3708


eo-maven-plugin/src/main/java/org/eolang/maven/Unplace.java
```diff
@@ -62,7 +86,7 @@ final class Unplace {
     public String make(final Path file) {
         return file.toString().substring(
             this.parent.toString().length() + 1
-        ).replaceAll(".eo$", "").replace(File.separator, ".");
+        ).replaceAll(this.extension, "").replace(File.separator, ".");
```
yegor256, line 89: @ArtemGet this looks pretty dangerous. What if `this.extension` will be equal to `(` -- you will get runtime exception here. Try using `Pattern.quote()`.

maxonfjvipon:
@ArtemGet please merge current `master` to your branch and restart the build

ArtemGet:
@maxonfjvipon Is OPTIMIZED attribute not present now? I am using it in xmir-to-phi cache.

maxonfjvipon:
@ArtemGet yes, it was removed [here](https://github.com/objectionary/eo/pull/3781). Also please note that input directory for `PhiMojo` is `1-parse` now, so you would need an `XMIR` attribute instead of `OPTIMIZED`

eo-maven-plugin/src/main/java/org/eolang/maven/Unplace.java
ArtemGet: @yegor256 Agreed, add Pattern.quote and tests.

ArtemGet:
@maxonfjvipon Changed OPTIMIZED to XMIR. Could you check please?
 Also, there is some Codacy check that fails on changing parsing/optimization directories(that is described wrongly, I guess) in eo-maven-plugin's README.

PR #3784, opened 2025-01-02
add AtRhoTest

sunmisc:
I apologize for my long absence.
I also want to apologize for possible non-compliance with the code format, and other errors. I do not use this product yet, but only had a quick look at the code

This test is a continuation of https://github.com/objectionary/eo/pull/3723

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/AtRhoTest.java
+++ b/eo-runtime/src/test/java/org/eolang/AtRhoTest.java
@@ -52,7 +52,7 @@ final class AtRhoTest {
                             start.await();
                             final Phi phi = new Phi.ToPhi(count.incrementAndGet());
                             rho.put(phi);
-                        } catch (InterruptedException e) {
+                        } catch (final InterruptedException ex) {
                             Thread.currentThread().interrupt();
                         } finally {
                             end.countDown();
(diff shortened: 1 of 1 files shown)


sunmisc:
> @sunmisc we can easily forgive you for the long absence, but can't forgive the wrong formatting of the code :) our checkers won't let us merge your code until all builds are green

Phew, 2025 in copyright is an error, which is probably weird. 
The rest of the formatting rules are pretty unfamiliar to me, but I respect them anyway

sunmisc:
> > @sunmisc we can easily forgive you for the long absence, but can't forgive the wrong formatting of the code :) our checkers won't let us merge your code until all builds are green
> 
> Phew, 2025 in copyright is an error, which is probably weird. The rest of the formatting rules are pretty unfamiliar to me, but I respect them anyway

The actual test fails because of the race.

sunmisc:
@yegor256 In fact, we can assume that the AtRho object is not thread-safe and then this test can be removed. But inside this object AtomicReference is used, even with considerations that it is for the sake of elegance, there are more efficient methods for working in one thread - setPlain / getPlain

yegor256:
@sunmisc as a rule of thumb, if you see a bug in master branch, report a new issue, don't try to fix it in your branch

Issue #3793, opened 2025-01-08
`EOnumber$EOtimesTest` fails


asmirnov-backend:
[EOnumber$EOtimesTest::throwsCorrectError](https://github.com/objectionary/eo/blob/master/eo-runtime/src/test/java/EOorg/EOeolang/EOnumber%24EOtimesTest.java#L51) fails with message
```
java.lang.AssertionError: the message in the error is correct
Expected: a string containing "number.times expects its second argument to be a number"
     but: was "Can't convert 1 bytes to java.lang.Double, exactly 8 bytes expected"
Expected :a string containing "number.times expects its second argument to be a number"
Actual   :"Can't convert 1 bytes to java.lang.Double, exactly 8 bytes expected"
```
Ref: #3461

asmirnov-backend:
If `.otherwise("number.times expects its second argument to be a number")` will be added to [EOnumber$EOtimes::lambda](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/EOorg/EOeolang/EOnumber%24EOtimes.java#L59), the test will pass, but the error message will be: 
```
org.eolang.ExFailure: the 'x' attribute Can't convert 1 bytes to java.lang.Double, exactly 8 bytes expected number.times expects its second argument to be a number
```
This error message has problems with punctuation and a bit not clear. 
@yegor256 can you please leave a comment about this

yegor256:
@asmirnov-backend see how it's done here: https://github.com/objectionary/eo/blob/bdad472f22b3475e5add0e793e5e22a8bf62509f/eo-runtime/src/main/java/EOorg/EOeolang/EObytes%24EOslice.java

asmirnov-backend:
@yegor256 For EObytes$EOslice this code from [test](https://github.com/objectionary/eo/blob/bdad472f22b3475e5add0e793e5e22a8bf62509f/eo-runtime/src/test/java/EOorg/EOeolang/EObytesEOsliceTest.java#L75)
```java
new Dataized(
    new PhWith(
        new PhWith(
            new Data.ToPhi("hello, world!")
                .take("as-bytes")
                .take("slice")
                .copy(),
            "start",
            new Data.ToPhi(2)
        ),
        "len",
        new Data.ToPhi(true) // incorrect value
    )
).take();
```
produces that error message
```
ERROR o.e.Dataized - Dataized to org.eolang.error with org.eolang.string inside, at:
[...]

PR #3795, opened 2025-01-08
 fix(#3793): Fix Expect

asmirnov-backend:
Ref: #3793

Now Expect is "fail fast" and the error message is only for one fail


eo-runtime/src/main/java/org/eolang/ExpectFailureInMust.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package org.eolang;
+
+/**
+ * Expect exception for method 'must'.
+ *
+ * @since 0.51
+ */
+class ExpectFailureInMust extends ExFailure {
```
yegor256, line 32: @asmirnov-backend maybe it's better to make them nested classes, inside `Expect`?

asmirnov-backend: Thanks for comment!
Unfortunately, the `Expect<T>` class can not contain nested exceptions, because `Expect<T>` is a generic class.
image

Possible solutions:

1. Leave exceptions in separate classes
2. Remove generic from `Expect` and make it works only with `Phi`
3. Use static (But static is bad👎, I do not know is it ok to use it here). `private static class ExpectFailureInThat extends ExFailure`

@yegor256 Which solution do you want to choose?

yegor256:
@asmirnov-backend looks very good! Please, make sure the build passes and then we'll review the PR.

asmirnov-backend:
Errors in checks were fixed. @yegor256 Could you please review

eo-runtime/src/main/java/org/eolang/Expect.java
```diff
@@ -135,4 +151,38 @@ public T it() {
         return this.sup.get();
     }
 
+    /**
+     * This exception is used to enhance the error message
+     * in the {@link Expect#otherwise(String)} method.
+     *
+     * @since 0.51
+     */
+    private static class ExpectFailureInMust extends ExFailure {
```
yegor256, line 160: @asmirnov-backend this sounds like a tautology for me: `Expect.ExpectFailureInMust`. Maybe a better name for this class would be `ExMust`? And `ExThat` for the other one.

eo-runtime/src/test/java/org/eolang/ExpectTest.java
```diff
@@ -34,11 +34,15 @@
  * @since 0.1.0
  */
 final class ExpectTest {
+    /**
+     * Subject for Expect.
+     */
+    private static final String SUBJ = "a number";
```
yegor256, line 40: @asmirnov-backend better have no variables or literals shared by tests. Inline this one.

yegor256:
@asmirnov-backend looks very good now! A few small comments above

eo-runtime/src/main/java/org/eolang/Expect.java
asmirnov-backend: Thanks for explanation. Fixed

eo-runtime/src/test/java/org/eolang/ExpectTest.java
asmirnov-backend: Thanks for explanation. Fixed

asmirnov-backend:
@yegor256 Thank you for review and comments. I appreciate it. Fixed the code according to your comments.

yegor256:
@maxonfjvipon please, review this one

maxonfjvipon (approved):

yegor256:
@asmirnov-backend all conversations must be resolved and this PR will be merged automatically: please, resolve them

asmirnov-backend:
@yegor256 @maxonfjvipon Thanks for review and help. I am happy that I was able to make contribution 😊

yegor256:
@asmirnov-backend thanks a lot! We are looking forward to more pull requests from you! :)

Issue #3481, opened 2024-11-10
the usage of `^.` seems redundant


yegor256:
We use `^.` too often in `.eo` files. In the majority of situations this prefix may be skipped and the location of the object will be automatically found by the compiler.

yegor256:
@maxonfjvipon compare this:

```
[] > foo
  [k] > bar
  bar 42 > x
```

and this:

```
[] > foo
  [k] > bar
  ^.bar 42 > x
```

The functionality here is not the same. Am I right?

maxonfjvipon:
@yegor256 yes, they are different

yegor256:
@maxonfjvipon I guess, the first snippet must be written as:

```
[] > foo
  [k] > bar
  $.bar 42 > x
```

Thus, we either have `\xi.bar` (`$.bar`) or `\xi.\rho.bar` (`^.bar`). Just `bar` means the first option. Am I right?

maxonfjvipon:
@yegor256 yes, the first one

yegor256:
@maxonfjvipon maybe it would be better to make our syntax more explicit: `bar` means `$.bar` and that's it. If `^.bar` is meant, it should be written as such. No defaults. WDYT?

maxonfjvipon:
@yegor256 I'm not sure because it'll make the language less convenient for end user (and I think we want it to be convenient and we actually add such syntax sugar, for example `>>`), but it opposite it'll make it more strict

maxonfjvipon:
@yegor256 Actually yes, I think it's a good idea

yegor256:
@maxonfjvipon maybe the other way around would be better:

`$.bar` means `\xi.bar` (we use it rarely, as I see)

`bar` means `\xi.\rho.\rho...bar` (where it is found)

maxonfjvipon:
@yegor256 it should already work like this

yegor256:
@maxonfjvipon  not really. If I use just `bar` it means `\xi.bar`:

```
[] > foo
  [k] > bar
  bar 42 > x
```

maxonfjvipon:
@yegor256 yes, if `bar` exists in the same scope, if it's not it'll try to find `^.bar`, `^.^.bar` and so on

yegor256:
@maxonfjvipon this code should not compile:

```
[] > foo
  [k] > bar
  bar 42 > x
```

instead, it should be written as such:

```
[] > foo
  [k] > bar
  $.bar 42 > x
```

[...]

PR #3808, opened 2025-01-12
bug(#3481): builds FQN via XSL

maxonfjvipon:
Ref: #3481 

In this PR I introduced XSL transformation that finds closest objects references and builds FQN, almost the same as `add-refs.xsl` does, but it inserts explicit objects.

Consider the next EO from `builds-fnq.yaml` pack:
```
# No comments.
[x y] > first
  stdout x > std
  [std] > second
    std x > inner
    [] > third
      5 > x
      x.plus y > sum
      std qwe > closest
    [] > fourth
      "some" > inner
      ^.inner > outer
      inner > self
foo > bar

# Foo.
[] > foo
```

The transformation converts it to the next EO:
```
[x y] > first
  stdout $.x > std
  [std] > second
    $.std ^.x > inner
    [] > third
      5 > x
      $.x.plus ^.^.y > sum
      ^.std qwe > closest
    [] > fourth
      "some" > inner
      ^.inner > outer
      $.inner > self
foo > bar

# Foo.
[] > foo
```

As you may see - it adds necessary `$` or `^.` objects.

When it's used in `TrShaking` - we're not going to need `add-refs.xsl` (which is used a several times) anymore because we're going to have FQN that are not needed be be resolved.

UPD:
If future, if the transformation is going to be heavy, it can be easily simplified: instead of inserting objects directly - we can attach some numeric attribute to target object. If attribute value is `0` - it means we should use `$`, if it's `> 0` then we should insert some `^.` objects. How much - attribute value shows

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/shake/add-refs.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/shake/add-refs.xsl
@@ -32,9 +32,10 @@ SOFTWARE.
   global or just a mistake.
 
   We must not add "ref" attributes to objects that refer to
-  "bytes" if such objects are inside the "org.eolang.bytes". Such
-  a reference would be misleading: instead of referring to the
-  global "org.eolang.bytes" they will lead to local "bytes"
+  "bytes", "string" or "number" if such objects are inside the
+  "org.eolang.bytes", "org.eolang.string" or "org.eolang.bytes".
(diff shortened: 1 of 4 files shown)


maxonfjvipon:
@yegor256 please check

yegor256 (approved):
@maxonfjvipon looks cool!

PR #3827, opened 2025-01-16
bug(#3481): remove `add-refs.xsl`

maxonfjvipon:
Ref: #3481 

In this PR I completely removed `add-refs.xsl` from the compiler


yegor256 (approved):
@maxonfjvipon cool, thanks!

Issue #3461, opened 2024-11-04
let's use `Expect` in all atoms


yegor256:
Currently, we use `Expect` class just in a few atoms, where we want to have a detailed error reporting in case of user errors. Let's use it in _all_ atoms.

asmirnov-backend:
@yegor256 Should `Expect` be used for checking `Attr.RHO`?
For examlpe, transform this 
```java
new Dataized(this.take(Attr.RHO)).asNumber()
```
to this
```java
Expect.at(this, Attr.RHO)
      .that(phi -> new Dataized(phi).asNumber())
      .otherwise("Attr.RHO must be a number for 'EOnumber'")
      .it();
```

yegor256:
@asmirnov-backend yes, definitely, this will be helpful

asmirnov-backend:
While using `Expect` some code duplicates appear, for example, next piece of code used in `EOnumber$EOdiv`, `EOnumber$EOgt`, `EOnumber$EOplus`, `EOnumber$EOtimes` (Ref: #3789)
```java
Expect.at(this, Attr.RHO)
    .that(phi -> new Dataized(phi).asNumber())
    .otherwise("must be a number")
    .it();
```
Perhaps it would be useful to have a more clever `Dataized` with `Expect` class.
```java
public class DataizableExpect {

    private final Expect<Phi> expect;

    public DataizableExpect(final Expect<Phi> expect) {
        this.expect = expect;
    }

    public Double asDouble() {
        return this.expect
            .that(phi -> new Dataized(phi).asNumber())
            .otherwise("must be a number")
            .it();
    }

    public Integer asPositiveInt() {
        return this.expect
            .that(phi -> new Dataized(phi).asNumber())
            .otherwise("must be a number")
            .must(number -> number % 1 == 0)
            .that(Double::intValue)
            .otherwise("must be an integer")
            .must(integer -> integer >= 0)
            .otherwise("must be a positive integer")
            .it();
    }

}
```
The code that uses this class would look something like this:
```java
new DataizableExpect(Expect.at(this, Attr.RHO)).asDouble()
```
@yegor256 What do you think?

yegor256:
@asmirnov-backend maybe we can make inner classes:

```
new Expect.Number(Expect.at(this, Attr.RHO)).it();
new Expect.Integer(Expect.at(this, Attr.RHO)).it();
new Expect.Whole(Expect.at(this, Attr.RHO)).it(); // greater or equal to zero
```

[...]

PR #3833, opened 2025-01-17
feat: Add inner helpful classes in `Expect`

asmirnov-backend:
Ref: #3461

Inner classes help to encapsulate the logic of using Expect and make it easier.

Before:
```java
Expect.at(this, Attr.RHO)
    .it();
```
After:
```java
new Expect.Number(Expect.at(this, Attr.RHO)).it();
```

Ref to [comment](https://github.com/objectionary/eo/issues/3461#issuecomment-2597667500)

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Expect.java
+++ b/eo-runtime/src/main/java/org/eolang/Expect.java
@@ -210,4 +210,63 @@ public class Expect<T> {
         }
     }
 
+    /**
+     * Transform Expect to Number
+     *
+     * @since 0.51
+     */
+    public static class Number {
+
+        /**
+         * Expect.
+         */
+        private final Expect<Phi> expect;
+
+        /**
+         * Ctor.
+         * @param expect Expect
+         */
+        public Number(final Expect<Phi> expect) {
+            this.expect = expect;
+        }
+
+        public Double it() {
+            return this.expect
+                .that(phi -> new Dataized(phi).asNumber())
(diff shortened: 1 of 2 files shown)


eo-runtime/src/main/java/org/eolang/Expect.java
```diff
@@ -34,7 +34,7 @@
  * @param <T> The type of result
  * @since 0.41.0
  */
-@SuppressWarnings("PMD.ShortMethodName")
+@SuppressWarnings({"PMD.ShortMethodName", "PMD.UnnecessaryFullyQualifiedName"})
```
yegor256, line 37: @asmirnov-backend in most cases, it's better to listen to PMD, instead of silencing it. Just rename `Integer` to `Int` and `NonNegativeInteger` to `Natural` (https://en.wikipedia.org/wiki/Natural_number)

yegor256:
@asmirnov-backend thanks! One comment above. Also, check this one: https://www.yegor256.com/2024/04/01/ping-me-please.html

eo-runtime/src/main/java/org/eolang/Expect.java
asmirnov-backend: Thanks! Fixed that

asmirnov-backend:
@yegor256 thanks for review! Comment resolved. Could you please rereview

yegor256:
@asmirnov-backend thanks!

Issue #3804, opened 2025-01-10
PhiMojoTest.java:54-56: Remove @Disabled annotation on...


0pdd:
The puzzle `3708-dc6420f7` from #3708 has to be resolved: 

https://github.com/objectionary/eo/blob/64660881e785378f547740ab5e83661698976b0a/eo-maven-plugin/src/test/java/org/eolang/maven/PhiMojoTest.java#L54-L56

The puzzle was created by Yegor Bugayenko on 10-Jan-25. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

ArtemGet:
@yegor256 Assign this one to me please.

maxonfjvipon:
@ArtemGet go ahead

PR #3865, opened 2025-01-28
bug(#3804): Implemented phi-to-xmir cache.

ArtemGet:
Implemented cache in ```UnphiMojo```. I decided to leave tests ```Disabled``` since empty ```hash``` is not cacheable. Added sync tojos todo. 

Ref #3804


ArtemGet:
@maxonfjvipon Could you take a look, please?

maxonfjvipon:
@ArtemGet if you enable these disabled tests - will they pass?

ArtemGet:
@maxonfjvipon Nope, because this tests needs ```hash``` otherwise cache file will not be created. I could add one more test that check that parsing will be skipped if target older than source e.t.c (But I dont found any tests like that, all test for other Mojo's requires ```hash```).

maxonfjvipon:
@ArtemGet 
1. empty hash used in `Footprint` should not lead to behavior where file is not created. If so - it's a bug
2. Since we use empty hash for now - let's test that it works with empty hash

Otherwise it looks like you changed the functionality but it's not covered by any tests.

So the tests must be enabled and let's check that they work with empty hash

Issue #3845, opened 2025-01-23
Roll bases if possible


maxonfjvipon:
According to [this](https://github.com/objectionary/lints/issues/256) we should introduce some new XSL transformation that would roll all the bases after `wrap-method-call` if it's possible.

This transformation should be a part of canonical pipeline

PR #3867, opened 2025-01-29
bug(#3845): introduced `roll-bases.xsl`

maxonfjvipon:
Closes: #3845 

In this PR I introduced `roll-bases.xsl` and made it a part of canonical EO and PHI parsing pipelines.
It required:
1. Refactoring `TrShaking` and moving many XSLs from it to EoSyntax and PhiSyntax
2. Refactoring `add-probes.xsl`. It led to removing `DiscoverMojo` from `assemble` pipeline
3. Refactoring converting XMIR to PHI and to EO
4. Refactoring many yaml test packs
5. Fixing converting to Java
6. Suppressing many lints in eo-runtime

Now we have:
1. We can get canonical XMIR right after EO or PHI parsing
2. Objects bases in such XMIR must be:
3.1. either `∅`
3.2. or start with `.`
3.3. or start with `Q`
3.4. or start with `$`
all other options are prohibited by XSD
4. all object bases must be rolled into single one until it's not possible anymore
5. canonical XMIR now also assumes that we have named objects only at the first level of abstract objects


maxonfjvipon:
@yegor256 please check

eo-maven-plugin/src/test/java/org/eolang/maven/DiscoverMojoTest.java
```diff
@@ -48,6 +49,7 @@
  * @since 0.28.11
  */
 @ExtendWith(MktmpResolver.class)
+@Disabled
```
yegor256, line 52: @maxonfjvipon this looks pretty dangerous. What's wrong here?

yegor256 (changes requested):
@maxonfjvipon this is a massive change, thanks! One question though, about `DiscoverMojo`. Can't we keep the test intact?

maxonfjvipon:
@yegor256 I disabled the test because it just does not work with new composite bases. Also we don't need this mojo anymore, it kind of duplicates the functionality of `add-probes.xsl` and `ProbeMojo`. So I can just delete the mojo and the test. WDYT?

yegor256:
@maxonfjvipon yes, just delete it and its test

maxonfjvipon:
@yegor256 done, please check one more time

Issue #3916, opened 2025-02-13
`Xmir.toPhi` Fails With `StackOverflow` Exception


volodya-lombrozo:
I run `xmir-to-phi` goal from `eo-maven-plugin:0.51.4` and get the following error:

```java
Caused by: java.lang.IllegalArgumentException: Failed to transform by net.sf.saxon.jaxp.TransformerImpl: Too many nested apply-templates calls. The stylesheet may be looping.; SystemID: file:///org/eolang/parser/phi/to-phi.xsl; Line#: 180; Column#: 50
    at com.jcabi.xml.XSLDocument.transformInto (XSLDocument.java:445)
    at com.jcabi.xml.XSLDocument.transform (XSLDocument.java:399)
    at com.yegor256.xsline.StXSL.lambda$new$1 (StXSL.java:45)
    at com.yegor256.xsline.StLambda.apply (StLambda.java:115)
    at com.yegor256.xsline.StEnvelope.apply (StEnvelope.java:61)
    at com.yegor256.xsline.StEnvelope.apply (StEnvelope.java:61)
    at com.yegor256.xsline.StLogged.apply (StLogged.java:135)
    at com.yegor256.xsline.StFast.apply (StFast.java:94)
    at com.yegor256.xsline.StAfter.lambda$new$0 (StAfter.java:48)
    at com.yegor256.xsline.StLambda.apply (StLambda.java:115)
    at com.yegor256.xsline.StEnvelope.apply (StEnvelope.java:61)
    at com.yegor256.xsline.Xsline.pass (Xsline.java:85)
    at org.eolang.parser.Xmir.converted (Xmir.java:200)
    at org.eolang.parser.Xmir.toPhi (Xmir.java:166)
    at org.eolang.maven.PhiMojo.translated (PhiMojo.java:181)
    at org.eolang.maven.PhiMojo.translate (PhiMojo.java:149)
    at org.eolang.maven.PhiMojo.lambda$exec$0 (PhiMojo.java:107)
    at org.eolang.maven.util.Threaded.lambda$total$0 (Threaded.java:75)
    at org.cactoos.scalar.CallableEnvelope.call (CallableEnvelope.java:53)
    at java.util.concurrent.FutureTask.run (FutureTask.java:317)
    at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1144)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:642)
[...]

PR #3918, opened 2025-02-13
bug(#3916): no recursion in to-phi.xsl

maxonfjvipon:
Ref: #3916


maxonfjvipon:
@yegor256 please check

Issue #3925, opened 2025-02-16
Way for test classes, which use `Expect`


asmirnov-backend:
While writing tests for issue #3853, I noticed that there are duplicates code in the test cases for the `Expect` functionality. For example, the following test cases share repetitive code fragments:

```java
@Test
void throwsCorrectErrorForNewSizeAttrNaN() {
    MatcherAssert.assertThat(
        "the message in the error is correct",
        Assertions.assertThrows(
            ExAbstract.class,
            () -> new Dataized(
                new PhWith(
                    new EOmallocTest.PhiWithIdDummy(
                        new EOmalloc$EOof$EOallocated$EOresized()
                    ).it(),
                    "new-size",
                    new Data.ToPhi(true)
                )
            ).take(),
            "put TRUE in int attr fails with a proper message that explains what happened"
        ).getMessage(),
        Matchers.equalTo("the 'new-size' attribute must be a number")
    );
}

@Test
void throwsCorrectErrorForNewSizeAttrNotAnInt() {
    MatcherAssert.assertThat(
        "the message in the error is correct",
        Assertions.assertThrows(
            ExAbstract.class,
            () -> new Dataized(
                new PhWith(
                    new EOmallocTest.PhiWithIdDummy(
                        new EOmalloc$EOof$EOallocated$EOresized()
                    ).it(),
                    "new-size",
                    new Data.ToPhi(42.42)
                )
            ).take(),
            "put double in int attr fails with a proper message that explains what happened"
        ).getMessage(),
        Matchers.equalTo("the 'new-size' attribute (42.42) must be an integer")
    );
}

@Test
void throwsCorrectErrorForNewSizeAttrNotNatural() {
    MatcherAssert.assertThat(
        "the message in the error is correct",
        Assertions.assertThrows(
            ExAbstract.class,
            () -> new Dataized(
                new PhWith(
                    new EOmallocTest.PhiWithIdDummy(
                        new EOmalloc$EOof$EOallocated$EOresized()
                    ).it(),
                    "new-size",
                    new Data.ToPhi(-42)
                )
            ).take(),
            "put negative int in natural attr fails with a proper message that explains what happened"
        ).getMessage(),
        Matchers.equalTo("the 'new-size' attribute (-42) must be greater or equal to zero")
    );
}
```

These tests share a common structure and could be refactored to reduce duplication and improve maintainability.

[...]

PR #3930, opened 2025-02-18
Add matchers for better tests attrs which uses Expect

asmirnov-backend:
Closes: #3925


asmirnov-backend:

Issue #3480, opened 2024-11-10
the `Attr` interface is redundant, can be removed


yegor256:
Earlier, the method `Phi.take()` was returning `Attr` interface. Currently, it returns `Phi`. Looks like we can safely remove the `Attr` and stay only with the `Phi`.

a1kuat:
@yegor256 Hello! Can I have a look if you don't mind? I think for the first issue and to start delving into, it is good task.

yegor256:
@a1kuat sure, go ahead

h1alexbel:
@yegor256 how about we make a release and enable `SnippetIT` test?

yegor256:
@rultor release, tag is `0.56.3`

yegor256:
@rultor release, tag is `0.56.3`

h1alexbel:
@yegor256 still an issue?

yegor256:
@h1alexbel fixed, thanks!

0crat:
@a1kuat 🎉 Great job closing this issue! You've earned **+4 points** for your effort. While there was a 194-day delay, we appreciate your persistence in completing the task. Keep up the good work! 💪

0pdd:
@yegor256 the puzzle [#4242](https://github.com/objectionary/eo/issues/4242) is still not solved; solved: [#4241](https://github.com/objectionary/eo/issues/4241), [#4243](https://github.com/objectionary/eo/issues/4243).

0pdd:
@yegor256 all 3 puzzles are solved here: [#4241](https://github.com/objectionary/eo/issues/4241), [#4242](https://github.com/objectionary/eo/issues/4242), [#4243](https://github.com/objectionary/eo/issues/4243).

PR #3932, opened 2025-02-19
#3542 output a clearer message when too many attributes are passed.

### LINKED ISSUE #3542: The error message is misleading when setting too many attributes
Consider this code:

```
sprintf
  "Hello, %s, your salary is %d"
  "Jeff"
  42
```

This will lead to error:

```
Can't overwrite the "λ" attribute
```

It's a very weird and misleading message. Apparently, it happens because `sprintf` doesn't have three void attributes, but only two. The right code is this:

```
sprintf
  "Hello, %s, your salary is %d"
  *
    "Jeff"
    42
```

I suggest adding this test method to `PhDefaultTest`:

```java
    @Test
    void failsCorrectlyWhenTooManyAttributesPut() {
        MatcherAssert.assertThat(
            "the message explains what's going on",
            Assertions.assertThrows(
                ExAbstract.class,
                () -> new EOnumber().put(1, new Data.ToPhi(1)),
                "fails when trying to set attribute with too big position"
            ).getMessage(),
            Matchers.equalTo("there are no void attributes left, can't set the 1st one")
        );
    }
```

Let's make it pass.

Suban05:
Closes #3542


Suban05:
@maxonfjvipon please take a look

eo-runtime/pom.xml
```diff
@@ -186,6 +186,11 @@ SOFTWARE.
       <artifactId>rerunner-jupiter</artifactId>
       <version>2.1.6</version>
     </dependency>
+    <dependency>
+      <groupId>com.ibm.icu</groupId>
```
maxonfjvipon, line 190: @Suban05 `eo-runtime` must not have dependencies which scope is not `provided` or `test`. We have a [test](https://github.com/objectionary/eo/blob/master/eo-runtime/src/test/groovy/check-runtime-deps.groovy) for it, but it seems to be weak and does not fail when `scope` is not provided at all

eo-runtime/src/main/java/org/eolang/Attr.java
```diff
@@ -68,4 +68,12 @@ public interface Attr {
...
      */
     void put(Phi phi);
+
+    /**
+     * Put a new object with position of the attribute in.
+     *
+     * @param pos The position of the attribute
+     * @param phi The object to put
+     */
+    void put(int pos, Phi phi);
```
maxonfjvipon, line 78: @Suban05 I don't think we need this method in the interface, we are able to decide either we can put attribute or not in `PhDefault`

eo-runtime/src/main/java/org/eolang/AtOnce.java
```diff
@@ -78,4 +80,17 @@ public void put(final Phi phi) {
             )
         );
     }
+
+    @Override
+    public void put(final int pos, final Phi phi) {
+        throw new ExReadOnly(
+            String.format(
+                "There are no void attributes left, can't set the %s one",
+                new RuleBasedNumberFormat(
```
maxonfjvipon, line 89: @Suban05 I believe we don't need to use it here, we can implement this number formatting by our own

maxonfjvipon (changes requested):

eo-runtime/pom.xml
Suban05: @maxonfjvipon thanks, I'll keep that in mind

eo-runtime/src/main/java/org/eolang/Attr.java
Suban05: @maxonfjvipon it seems that I didn't understand the task correctly and implemented it more difficult than necessary. Do I understand correctly that we should throw this exception with a new message when pos > 0?

maxonfjvipon: @Suban05 no, I suggest you to do something like this:
```java
class PhDefault ... {
    @Override
    public void put(final int pos, final Phi object) {
        final String name = this.attr(pos);
        if (!this.attrs.containsKey(name)) {
            throw new ExUnset(
                String.format(
                    "Can't #put(\"%s\", %s) to %s, because the attribute is absent",
                    name, object, this
                )
            );
        }
        if (!this.attrs.get(name) instanceof AtVoid ) {
            throw new ExReadOnly(
               "Can't set non void attribute blah blah blah"
            )
        }
        this.put(this.attr(pos), object);
    }
}
```

I think such simple `instanceof` is enough, no need to add this overhead with extra method in `Attr` interface. Maybe you'll require one more private method in `PhDefault` to reduce code duplication where `ExUnset` is thrown, but it's up to you

Suban05: @maxonfjvipon I got you, thanks a lot

I added a new method to the interface because there is a task to remove the Attr interface https://github.com/objectionary/eo/issues/3480, which needs to be replaced with Phi. this interface requires the implementation of the put method with position, so I thought this approach would make the move to the new interface more lenient

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -336,4 +346,34 @@ private static void debug(final String msg) {
     private static String padding() {
         return String.join("", Collections.nCopies(PhDefault.NESTING.get(), "·"));
     }
+
+    /**
+     * Returns the ordinal representation of a number (e.g., "1st", "2nd", "3rd", "4th").
+     * @param number The number to format
+     * @return The number with its ordinal suffix
+     */
+    private static String ordinal(final int number) {
```
Suban05, line 355: @maxonfjvipon I'm not sure if it's right to place this method in `PhDefault`. it seems that it is needed only here, and there is no point in allocating a separate class for it

maxonfjvipon: @Suban05 maybe let's just remove the method and reformat the message so the suffix is not needed: `can't set attribute on position %d`

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -140,7 +140,17 @@ public boolean hasRho() {
 
     @Override
     public void put(final int pos, final Phi object) {
-        this.put(this.attr(pos), object);
+        final String name = this.attr(pos);
+        final AtWithRho attribute = (AtWithRho) this.attrs.get(name);
```
maxonfjvipon, line 144: @Suban05 let's inline the variable

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -140,7 +140,17 @@ public boolean hasRho() {
...
+        if (!(attribute.getOrigin() instanceof AtVoid)) {
+            throw new ExReadOnly(
+                String.format(
+                    "There are no void attributes left, can't set the %s one",
```
maxonfjvipon, line 148: @Suban05 the message is not quite correct. There are still may be void attributes ready to be set, but we've just missed it by providing wrong index. So let's say something like: `Can't put attribute with position %d because it's not void one`

maxonfjvipon (changes requested):

maxonfjvipon (changes requested):

eo-runtime/src/main/java/org/eolang/AtWithRho.java
```diff
@@ -71,4 +71,12 @@ public Phi get() {
     public void put(final Phi phi) {
         this.origin.put(phi);
     }
+
+    /**
+     * Returns the original attribute.
+     * @return The original attribute
+     */
+    public Attr getOrigin() {
```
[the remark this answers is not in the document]
Suban05: @maxonfjvipon I totally agree, fixed. I originally named the method as `origin`, but the checkstyle checks did not pass, so I thought I could use the name starting with get. however, I didn't think it was acceptable to change the attribute name. I'll keep in mind, thank you

maxonfjvipon (approved):

maxonfjvipon:
@Suban05 LGTM, thanks. Let's wait when `reuse` check is fixed and we're good to merge

PR #3936, opened 2025-02-21
feat(#3935): Try to Use `jmh-benchmark-action`

volodya-lombrozo:
In this PR I've added `jmh-benchmark-action` to our CI/CD pipelines.

Closes: #3935.

PR #3949, opened 2025-02-25
feat(#3936): Add 'post-comment' Workflow

volodya-lombrozo:
I added a simple workflow that runs on the `master` branch. Since it uses `workflow_run` event and runs on the `master` branch, `GITHUB_TOKEN` has read and write permissions. 
This allow it to post comments to PRs. 
Actually, we have only two steps here:
- Download the message for printing.
- Post this message to a PR.

By adding this workflow, we can provide benchmarking in a more secure environment without requiring `write` permissions.

Related to #3936


yegor256 (approved):

yegor256:
@volodya-lombrozo looks nice!

Issue #3935, opened 2025-02-21
Add `jmh-benchmark-action` to Detect Performance Issues


volodya-lombrozo:
There is [jmh-benchmark-action](https://github.com/volodya-lombrozo/jmh-benchmark-action) that can run JMH benchmarks in PR's. It compares base branch (`master`) performance with a PR performance and sends comparison results as a comment to the PR. This action might be helpful to detect performance issues.

PR #3951, opened 2025-02-25
feat(#3935): print debug info for 'post-comment' workflow

volodya-lombrozo:
'post-comment' workflow prints comments into a wrong PR. I've added some debug info that will help to investigate the problem.

Related to: #3935


yegor256:
@volodya-lombrozo merged, thanks

PR #3952, opened 2025-02-25
feat(#3935): Post Comments to a Correct PR

volodya-lombrozo:
`jmh-benchmark-action` stores the PR number on which it was run. Thus, we can read this PR number and post a comment using this number.

Related to #3935


volodya-lombrozo:
@yegor256 Could you merge these changes, please? This PR should fix the issue with benchmark results posting.

PR #3969, opened 2025-02-28
#3970 remove the dependence of the test on the runtime version

### LINKED ISSUE #3970: After the release, 0.52.0 integration tests fail
After the release, 0.52.0 integration tests fail, although everything was ok before.

Example: https://github.com/objectionary/eo/actions/runs/13585291457/job/37978804069

this is due to the fact that the old version of eo-runtime is listed here https://github.com/objectionary/eo/blob/b17fe4396e503d2c7ab66ccbb4045776a4d4fc08/eo-maven-plugin/src/test/java/org/eolang/maven/AssembleMojoIT.java#L41

I suggest removing this so that the latest version of eo-runtime is always used and changing the assert that checks the unpacking of the runtime

Suban05:
Fixes integration tests in the `master` branch

Closes #3970

### DIFF
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/AssembleMojoIT.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/AssembleMojoIT.java
@@ -37,8 +37,6 @@ final class AssembleMojoIT {
         new Farea(temp).together(
             f -> {
                 f.clean();
-                f.dependencies()
-                    .append("org.eolang", "eo-runtime", "0.51.6");
                 f.files()
                     .file("src/main/eo/one/main.eo")
                     .write(AssembleMojoIT.helloWorld().getBytes(StandardCharsets.UTF_8));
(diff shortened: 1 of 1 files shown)


Suban05:
@maxonfjvipon integration tests fail in the `master` branch, because there is a dependency on the runtime version. fixed

maxonfjvipon:
@Suban05 can you please add a reference to the particular issue?

Suban05:
@maxonfjvipon added a ticket

eo-maven-plugin/src/test/java/org/eolang/maven/AssembleMojoIT.java
```diff
@@ -37,8 +37,6 @@ void assemblesTogether(@Mktmp final Path temp) throws IOException {
         new Farea(temp).together(
             f -> {
                 f.clean();
-                f.dependencies()
-                    .append("org.eolang", "eo-runtime", "0.51.6");
```
maxonfjvipon, line 41: @Suban05 Can we use this instead of hardcoded version?
```java
System.getProperty(
  "eo.version",
  Manifests.read("EO-Version")
)
```

Suban05: @maxonfjvipon In version 0.56.1, eo-runtime was unpacked here: `target/eo/5-resolve/org.eolang/eo-runtime/-/0.51.6`. But when version 0.52.0 was released today, for some reason, `target/eo/5-resolve/org.eolang/eo-runtime/-/0.51.6` is now an empty directory, even though it previously contained .class binary files

It helps either to hardcode version 0.52.0, as with 0.56.1, or to remove the explicit specification of this dependency altogether, in which case the latest version of eo-runtime is used automatically

maxonfjvipon (approved):

Issue #3941, opened 2025-02-22
Make all the tests great again


maxonfjvipon:
Right now the tests in our project are not in very good condition, especially in `eo-maven-plugin`. There's a lot of clutter that needs to be fixed and refactored. Let's keep this ticket open until we fix all the tests in the projects. 

What we want to achieve:
1. all the tests that use [`Farea`](https://github.com/yegor256/farea) are moved to integration ones (we use `maven-failsafe-plugin` for it) since they rather slow
2. all the integration tests are annotated with [`@MayBeSlow`](https://github.com/yegor256/maybeslow)
3. all the static literals are removed from all the tests
4. only one assert per test (yes, many tests must be divided into a several ones)
5. to be continued...

Please don't make a huge PRs. Micro changes are very welcome!

PR #3985, opened 2025-03-05
#3941 more convenient `eo-maven-plugin` configuration in `Farea`

Suban05:
Since our `Farea` configuration for the `eo-maven-plugin` in our integration tests is slightly different from the usual setup, we started having some repetitive code. In this PR, I added a class that simplifies the Farea configuration for the `eo-maven-plugin`, which should help eliminate the duplication

Ref: #3941

### DIFF
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/EOplugin.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/EOplugin.java
@@ -1,0 +1,46 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2025 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import com.jcabi.manifests.Manifests;
+import com.yegor256.farea.Farea;
+import com.yegor256.farea.Plugin;
+import java.io.IOException;
+
+/**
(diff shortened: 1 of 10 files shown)


Suban05:
@maxonfjvipon WDYT? I'm not sure about the naming

eo-maven-plugin/src/test/java/org/eolang/maven/EOplugin.java
```diff
@@ -0,0 +1,46 @@
...
+import com.jcabi.manifests.Manifests;
+import com.yegor256.farea.Farea;
+import com.yegor256.farea.Plugin;
+import java.io.IOException;
+
+/**
+ * Configures the EO Maven plugin within a {@link Farea}.
+ * @since 0.52
+ */
+final class EOplugin {
```
maxonfjvipon, line 16: @Suban05 let's call it `AppendedPlugin` and make it implement `org.cactoos.Scalar<com.yegor256.farea.Execution>` so we could do:
```java
new AppendedPlugin(f).get().configuration(...)...
```

maxonfjvipon (changes requested):
@Suban05 great idea, I like it. Just one comment from my side

maxonfjvipon (approved):
@Suban05 LGTM! Thanks

Issue #3917, opened 2025-02-13
PHI/UNPHI Removes `j$` Prefix From `program/@name`


volodya-lombrozo:
I have the following `program` definition after `jeo-maven-plugin:disassemble` goal:

```xml
<program xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         dob="2025-02-13T12:28:36"
         ms="0"
         name="j$App"
         revision="8990728"
         time="2025-02-13T09:29:12.333741Z"
         version="1.0-SNAPSHOT"
         xsi:noNamespaceSchemaLocation="https://www.eolang.org/xsd/XMIR-0.51.4.xsd">
```
Then I do `xmir-to-phi:0.51.4` transformation:

```
j$App ↦ Φ.jeo.class(...)
```

And then `phi-to-xmir:0.51.4`:

```xml
<program xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         dob="2025-02-11T14:12:30"
         ms="612"
         name="App"
         revision="4dc215d"
         time="2025-02-13T09:29:41.161725Z"
         version="0.51.4"
         xsi:noNamespaceSchemaLocation="https://www.eolang.org/xsd/XMIR-0.51.4.xsd">
```

As you can see, we don't have `j$` prefix anymore.  Can you return it back, please?

[App.xmir.disassemle.txt](https://github.com/user-attachments/files/18781434/App.xmir.disassemle.txt)
[App.phi.txt](https://github.com/user-attachments/files/18781433/App.phi.txt)
[App.xmir.unphi.txt](https://github.com/user-attachments/files/18781432/App.xmir.unphi.txt)

volodya-lombrozo:
This issue leads to https://github.com/objectionary/lints/issues/338

volodya-lombrozo:
@yegor256 Could you have a look, please?

yegor256:
@maxonfjvipon why `UnphiMojo` changes the name of the program?

maxonfjvipon:
@yegor256 @volodya-lombrozo the `/program/@name` is taken from file name, as I understand this attribute is supposed to work like this.
1. After disassemble you have `App.xmir` with `j$App` name.
2. After PHI you have `App.phi` with `j$App` object
3. After UNPHI you have `App.xmir` with `App.name`

I think the problem is on the 1st step - you're saving `j$App` program into `App.xmir` file. The names should be the same. At least it works like this in EO.

yegor256:
[...]

PR #4100, opened 2025-04-16
bug(#3917): compare tojo id and object name

maxonfjvipon:
Ref: #3917 

In this PR I've moved `ObjectName` class to `eo-parser` module. It'll allow to get an object name for end-users.
Also I've added the transformation that raises error in wrong objects count is provided to `PhiSyntax`


maxonfjvipon:
@yegor256 please check

PR #4113, opened 2025-04-21
test(#4112): add unit tests for FpUpdateFromCache class

### LINKED ISSUE #4112: Add unit tests for `FpUpdateFromCache` class
The `FpUpdateFromCache` class lacks unit tests, making its behavior unclear. Adding tests would improve maintainability and document its functionality.

volodya-lombrozo:
This PR adds unit tests for `FpUpdateFromCache` class and introduces a new `Filesystem` interface to improve testability.

Closes #4112


volodya-lombrozo:
@yegor256 What do you think?

yegor256:
@volodya-lombrozo file system mocking is not a good idea. Here is a page from Angry Tests book:

Screenshot 2025-04-22 at 10 26 01

yegor256:
@volodya-lombrozo can't you just give it a temporary directory path and then assert on the changes it makes there?

volodya-lombrozo:
@yegor256 The purpose of this PR is to demonstrate that we can speed up tests and eliminate the unnecessary coupling with the filesystem. This coupling is what makes the compiler so extremely slow.

For now, it seems that we need to use the filesystem to make all these compilation steps work. But in reality, the filesystem is just a place where we store intermediate results. All the work could be done in memory.

We can look at the filesystem as one of the storage types. We can use in-memory storage, the filesystem, the internet—anything we want.

If we are able to use in-memory storage in tests, we may later be able to switch to this storage for the entire compiler.

As for `@TempDIr`.
I measured a simple scenario using `@TempDir`, and in general, it takes more than 2.5 times longer (on a trivial example with only a few bytes). I haven’t even measured the initialisation time, which should be rather significant.

<details>
<summary>code</summary>

```java
final class FpUpdateFromCacheTest {
    @Test
    void appliesFromCache(@TempDir final Path tmp) throws Exception {
        measure(() -> {
                final Filesystem.Real fake = new Filesystem.Real();
                final Text expected = new TextOf("Cached!");
                final Path cached = fake.save(tmp.resolve("cache.txt"), expected);
                final Path target = tmp.resolve("target.txt");
                MatcherAssert.assertThat(
                    "We expect the footprint will return the target file path",
                    new FpUpdateFromCache(() -> cached, fake).apply(Paths.get("/dev/null"), target),
                    Matchers.equalTo(target)
                );
                MatcherAssert.assertThat(
                    "The target file should be updated from cache",
                    fake.read(target),
                    Matchers.equalTo(expected)
                );
            }
        );
    }

    @Test
    void appliesFromCache() throws Exception {
        measure(() -> {
            final Filesystem.Fake fake = new Filesystem.Fake();
            final Text expected = new TextOf("Cached!");
            final Path cached = fake.save(Paths.get("./cache.txt"), expected);
            final Path target = Paths.get("./target.txt");
            MatcherAssert.assertThat(
                "We expect the footprint will return the target file path",
                new FpUpdateFromCache(() -> cached, fake).apply(Paths.get("/dev/null"), target),
                Matchers.equalTo(target)
            );
            MatcherAssert.assertThat(
                "The target file should be updated from cache",
                fake.read(target),
                Matchers.equalTo(expected)
            );
        });

    }

    private void measure(FRunnable runnable) throws Exception {
        final long start = System.currentTimeMillis();
        runnable.run();
        final long end = System.currentTimeMillis();
        final long duration = end - start;
        System.out.println("Execution time: " + duration + " ms");
    }

    private interface FRunnable {
        void run() throws Exception;
    }
}
```
</details>

This coupling makes testing extremely slow because all of our unit tests are actually integration tests. We don't have unit tests at all.

When I run the "unit" tests in my IDE, I wait for 3 minutes and 28 seconds. If we want to speed up the tests, we just need to minimise this type of testing.

yegor256:
@volodya-lombrozo our tests are slow because we use file system instead of memory??

volodya-lombrozo:
@yegor256 Not only. We have several places that might be optimised. However, this one is the most obvious.

yegor256:
@volodya-lombrozo indeed, if you repeat 1000 times, you see the difference of 2 seconds. But in normal situation, the difference is just 2 milliseconds, which is nothing.

volodya-lombrozo:
@yegor256 It has commutative effect on the entire codebase. I can't just rewrite the entire plugin and send all the changes in a single PR to show you that is much faster. However, for this particular case the difference isn't significant, indeed.

volodya-lombrozo:
@yegor256 I've removed the `Filesystem` abstraction. Can you have a look one more time, please?

eo-maven-plugin/src/test/java/org/eolang/maven/FpUpdateFromCacheTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * Test case for {@link FpUpdateFromCache}.
+ * @since 0.57
+ */
+final class FpUpdateFromCacheTest {
+
+    @Test
+    void appliesFromCache(@TempDir final Path tmp) throws IOException {
```
maxonfjvipon, line 24: @volodya-lombrozo it's better to use `@Mktmp` so the result is saved in temporary directory inside `target`

yegor256 (approved):

yegor256:
@volodya-lombrozo this one will be merged once all conversations are resolved

0crat:
@yegor256 Thanks for your review! Here's how your +4 points were calculated: +4 base, -24 for self-review, +0.01 per hit-of-code (capped at 16), -10 for <6 comments, +0.25 per comment (capped at 8), adjusted to the minimum 4 points. Your balance is now [-220](https://www.eolang.org/zerocracy/objectionary-vitals.html). Keep up the good work and aim for more comments next time! 💪

0crat:
@volodya-lombrozo Great work on your contribution! 🌟 You've earned +4 points, bringing your total to [+73](https://www.eolang.org/zerocracy/objectionary-vitals.html). We appreciate your dedication to [quality and speed](https://www.yegor256.com/2018/03/06/speed-vs-quality.html). Keep up the fantastic effort!

PR #4167, opened 2025-05-14
Update CODE_OF_CONDUCT.md

SuleimanKarimEddin:


0crat:
@suleimankarimeddin Hi there! I noticed that the branch name "`master`" doesn't follow our naming convention. As per our [guidelines](https://www.yegor256.com/2014/04/15/github-guidelines.html), branch names should match the ticket number you're working on. This helps with organization and tracking. Unfortunately, this means a deduction of 12 points. For future reference, a branch name like "`4166`" would be perfect. Let's keep improving our workflow together! 💪

PR #4169, opened 2025-05-14
correct variable name and typos

saleemasekrea000:


0crat:
@saleemasekrea000 Hey there! 👋 I noticed your branch name "`fix/typos`" doesn't quite follow our naming convention. [Our guidelines](https://www.yegor256.com/2014/04/15/github-guidelines.html) suggest using the ticket number instead. For example, "`4168`" would be spot-on for this case. Just a heads up, there's a -12 point penalty for non-standard names. No worries though, it's an easy fix for next time! Let's keep our branches tidy and ticket-linked. 😊

Issue #4164, opened 2025-05-13
`integration.yml` fails in master branch


h1alexbel:
This is what we have in the `master` branch:

```text
Error:  integration.SnippetIT.runsAllSnippets(String, Path)[3] -- Time elapsed: 6.502 s <<< ERROR!
com.yegor256.farea.Farea$BuildFailureException: build failed with exit code 0x0001
	at com.yegor256.farea.Farea.exec(Farea.java:302)
	at integration.SnippetIT.lambda$runsAllSnippets$1(SnippetIT.java:96)
	at com.yegor256.farea.Farea.together(Farea.java:179)
	at integration.SnippetIT.runsAllSnippets(SnippetIT.java:45)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.LinkedList$LLSpliterator.forEachRemaining(LinkedList.java:1249)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762)
	at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:276)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1708)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
[...]

PR #4170, opened 2025-05-15
bug(#4164): integration.yml passes

h1alexbel:
In this PR I've resolved `integration.yml` failures, so now its passes.

see #4164


h1alexbel:
to be resolved in #4171

PR #4288, opened 2025-06-15
bug(#4192): redundant-object lint enabled

### LINKED ISSUE #4192: Redundant Object Lint Not Enabled in pom.xml
The puzzle `4157-b24a0abe` from #4157 has to be resolved: 

https://github.com/objectionary/eo/blob/13e39ab46f985166a2075f57fb0111c1b1ee8cb7/eo-runtime/pom.xml#L245-L249

The puzzle was created by @yegor256 on 20-May-25. 

h1alexbel:
In this PR I've fixed several `redundant-object` warnings, and enabled the lint.

closes #4192
History:
- **bug(#4192): redundant-object fixed**
- **bug(#4192): redundant-object fixed more**

### DIFF
--- a/eo-runtime/src/main/eo/org/eolang/bytes.eo
+++ b/eo-runtime/src/main/eo/org/eolang/bytes.eo
@@ -7,6 +7,13 @@
 +version 0.0.0
 +spdx SPDX-FileCopyrightText Copyright (c) 2016-2025 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
++unlint redundant-object:23
++unlint redundant-object:24
++unlint redundant-object:28
++unlint redundant-object:47
++unlint redundant-object:57
++unlint redundant-object:67
++unlint redundant-object:428
 
 # The object encapsulates a chain of bytes, adding a few
 # convenient operations to it. Objects like `int`, `string`,
@@ -385,11 +392,9 @@
   [] +> tests-concat-bools-as-bytes
     eq. > @
       concat.
-        b1
-        b2
+        true.as-bytes
+        false.as-bytes
       01-00
-    true.as-bytes > b1
-    false.as-bytes > b2
 
   # This unit test is supposed to check the functionality of the corresponding object.
   [] +> tests-concat-with-empty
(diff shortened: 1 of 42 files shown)


h1alexbel:
@maxonfjvipon please check

eo-runtime/src/main/eo/org/eolang/bytes.eo
```diff
@@ -7,6 +7,13 @@
 +version 0.0.0
 +spdx SPDX-FileCopyrightText Copyright (c) 2016-2025 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
++unlint redundant-object:23
```
maxonfjvipon, line 11: @h1alexbel it seems we can remove some of these `+unlint` meta

maxonfjvipon (changes requested):
@h1alexbel let's try to remove as many `+unlint redundant-object` as possible. Or make a todo for it

eo-runtime/src/main/eo/org/eolang/bytes.eo
h1alexbel: @maxonfjvipon I'm not sure its possible to resolve such violations:

```eo
# Bytes.
[] > bytes
  $ > as-bytes
```

Here, `bytes` reuses itself to "declare" `as-bytes` method. In strict languages like Java, it can be done through method overriding:

```java
final class Bytes implements EOObject {

  @Override
  public byte[] asBytes() {
    // return itself here
  }
}
```

In EO example, we can't really inline such object, since its expected from `bytes` to have `.as-bytes` method inside.

Also, I've created these issues in lints:
* [`redundant-object` incorrectly flags object self-reuse in nested formations](https://github.com/objectionary/lints/issues/639)
* [`redundant-object` lint rule fails to recognize object usage above its declaration level](https://github.com/objectionary/lints/issues/640)

Once they will be resolved, and new version of lints will be released, we will get `unlint-non-existing-defect` violation on some of the objects, marked "redundant" for now.

Am I missing something? WDYT?

P.S.:
I also thinking about moving `redundant-object` to WPA scope, in order to let it see all the usages outside of the single XMIR file. Though, it will require some type inference as well.

maxonfjvipon: @h1alexbel you may not touch such self referencing objects, there are a lot of other cases

maxonfjvipon (approved):

Issue #4335, opened 2025-06-29
Sugared array doesn't parse when object name is composite


yegor256:
This is my code:

```
# Example.
[] > app
  x.abc *1 > @
    foo
    bar
```

It doesn't compile:

```
[ERROR] Failed to parse '/Volumes/sec/code/tmp/eo/app.eo:3': [3:9] error: 'Invalid object declaration'
  x.abc *1 > @
         ^
```

However, this one compiles:

```
# Example.
[] > app
  abc *1 > @
    foo
    bar
```

Looks like a bug to me.

yegor256:
@maxonfjvipon is it a bug?

maxonfjvipon:
@yegor256 right now rule with this compact array looks like this:

<img width="234" alt="Image" src="https://github.com/user-attachments/assets/b5c9a3aa-e19e-4c69-8dee-51bd15dc8798" />

It accepts only `NAME` in front of `*`.

We introduced such syntax in order to remove indentation in such code:
```
# FROM
seq
  *
    1
    2

# TO
seq *
  1
  2
```

yegor256:
@rultor release, tag is `0.57.1`

yegor256:
@h1alexbel thanks!

PR #4352, opened 2025-07-06
bug(#4335): allow `HOME` in `compactArray`

h1alexbel:
In this PR I've patched the EO grammar to support `HOME` terminals in the compact array syntax like this:

```
# No comments.
[] > app
  QQ.seq *1 > @
    x
    y
```

see #4335

## Summary by CodeRabbit

* **New Features**
  * Enhanced array syntax to support qualified names with multiple dot-separated identifiers, including special prefixes.

* **Tests**
  * Added multiple test cases to validate parsing of arrays with composite names, including those with special prefixes and references.


eo-parser/src/main/antlr4/org/eolang/parser/Eo.g4
```diff
@@ -238,7 +238,7 @@ vapplicationHead
 
 // Compact arrays
 compactArray
-    : NAME SPACE STAR INT?
+    : (HOME DOT)? NAME SPACE STAR INT?
```
maxonfjvipon: @h1alexbel maybe let's allow `$.` also?

h1alexbel: @maxonfjvipon yes, we should. How about this `x.abc.foo`? Seems that it also valid. WDYT?

h1alexbel:
@yegor256 take a look, please

h1alexbel:
@yegor256 reminder

PR #4357, opened 2025-07-08
Refactor: Remove origin() from PhWithRho to fix TooManyMethods PMD issue

### LINKED ISSUE #4242: Refactor PhWithRho.java: Extraneous origin() Method Requires Removal
The puzzle `3480-6f27e031` from #3480 has to be resolved: 

https://github.com/objectionary/eo/blob/daef5e0990986c1c02ac16b08cebaab4c8804ba9/eo-runtime/src/main/java/org/eolang/PhWithRho.java#L11-L13

The puzzle was created by Yegor Bugayenko on 05-Jun-25. 

Chaleshka:
Since PMD `TooManyMethods` runs methods implemented directly in the class, we:
1. Moved `origin()` and its field to the new parent class `AbstractPhWithAttr`
2. Ьade `PhWithRho` inherit from it
3. Reduced direct method count in `PhWithRho` from 11 to 10

This is a change:
- Preserves all existing behavior, because the method remains as it is, but it is located in the parent class.
- Supports the original visibility of `origin()' (package-private)
- Allows removal of `@SuppressWarnings` in future

Resolves #4242

## Summary by CodeRabbit

* **Refactor**
  * Improved code structure for attribute handling, making it easier to extend related classes in the future. No changes to user-facing behavior.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhWithRho.java
+++ b/eo-runtime/src/main/java/org/eolang/PhWithRho.java
@@ -11,7 +11,6 @@ package org.eolang;
  *
  * @since 0.36.0
  */
-@SuppressWarnings("PMD.TooManyMethods")
 final class PhWithRho extends AbstractPhWithAttr {
 
     /**
--- a/eo-runtime/src/main/java/org/eolang/PhWithRho.java
+++ b/eo-runtime/src/main/java/org/eolang/PhWithRho.java
@@ -5,6 +5,12 @@
 
 package org.eolang;
 
+/**
+ * A decorator for an attribute that ensures the object has a {@code \rho} attribute.
+ * If not already set, it sets it.
+ *
+ * @since 0.36.0
+ */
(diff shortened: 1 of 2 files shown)


eo-runtime/src/main/java/org/eolang/PhWithRho.java
```diff
@@ -6,18 +6,12 @@
...
- *  Now we use it in {@link PhDefault#put(int, Phi)}. Let's move it to other class,
- *  in order to get rid of `TooManyMethods` PMD violation.
  */
-@SuppressWarnings("PMD.TooManyMethods")
-final class PhWithRho implements Phi {
-    /**
-     * Original attribute.
-     */
-    private final Phi original;
+final class PhWithRho extends AbstractPhWithAttr {
```
yegor256, line 14: @Chaleshka read this: https://www.yegor256.com/2016/09/13/inheritance-is-procedural.html

Issue #4108, opened 2025-04-17
Add tabulation support for comments blocks in parsed files


SenjeyB:
Сomments often contain examples of code, but due to the lack of tabulation (or spaces in the start of the line) support in  the <comments> block, the code in it cannot always be clearly restored while parsing .xmir files.

0crat:
@senjeyb Hey there, bug-hunter! 🐛 You just scored 8 points for your sharp eye! It's awesome how you're helping us level up our project quality. Keep that keen vision going - if you spot anything else in the repo that [seems off](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html), [don't hold back](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html). [Report away](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) and rack up those points! Every bug you catch is a win for the team. Keep 'em coming!

h1alexbel:
@SenjeyB @yegor256 should be solved by #4358, please try again

PR #4358, opened 2025-07-08
bug(#4108): `XeEoListener` saves indentation in nested comments

h1alexbel:
In this PR I've updated `XeEoListener` to preserve indentation in the nested comments, while parsing to XMIR.

see #4108

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of comment text to better preserve indentation and formatting, including support for empty comments.

* **Tests**
  * Added tests to verify correct preservation of comment indentation and handling of empty comments during parsing.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
+++ b/eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
@@ -1135,14 +1135,20 @@ final class XeEoListener implements EoListener, Iterable<Directive> {
             this.dirs.push().xpath("/object").addIf("comments").add("comment").set(
                 comment.stream().map(
                     context -> {
-                        final int sub;
+                        final String result;
                         final String text = context.COMMENTARY().getText();
-                        if (text.length() > 1) {
-                            sub = 2;
+                        if (text.isEmpty()) {
+                            result = "";
                         } else {
-                            sub = 1;
+                            final int sub;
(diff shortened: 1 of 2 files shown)


eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
```diff
@@ -1134,7 +1134,22 @@ private void putComment(final List<EoParser.CommentContext> comment, final Token
...
-                    context -> context.COMMENTARY().getText().substring(1).trim()
+                    context -> {
+                        final String result;
+                        final String text = context.COMMENTARY().getText();
+                        if (text.isEmpty()) {
+                            result = "";
+                        } else {
+                            final int sub;
+                            if (text.length() > 1 && text.charAt(1) == ' ') {
+                                sub = 2;
```
maxonfjvipon, line 1145: @h1alexbel what if indent is more than two spaces:
```
# First.
#     Second.
```

maxonfjvipon (changes requested):

eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
h1alexbel: @maxonfjvipon all the spaces will be preserved. We are only remove `#` and first empty character between start of the word `First.`, or `    Second.` Added more tests

PR #4404, opened 2025-07-27
`as-tuple` removed from `tuple` object

### LINKED ISSUE #4403: `as-tuple` Redundancy in `tuple` Object Usage in `text.eo`
In the `tuple` we currently have this:

```
[] > tuple
  $ > as-tuple
```

Later, we use it like this (in `text.eo`):

```
[] > as-number
  if. > @
    scanned.length.eq 0
    error
      sprintf
        "Can't convert text %s to number"
        * origin
    scanned.head
  (sscanf "%f" origin).as-tuple > scanned
```

It seems, we can just do this instead:

```
(sscanf "%f" origin).@ > scanned
```

yegor256:
closes #4403

### DIFF
--- a/eo-runtime/src/main/eo/org/eolang/fs/dir.eo
+++ b/eo-runtime/src/main/eo/org/eolang/fs/dir.eo
@@ -49,7 +49,7 @@
       if.
         ^.exists
         seq *
-          rec-delete (walk "**").as-tuple
+          rec-delete (walk "**").@
           ^
         ^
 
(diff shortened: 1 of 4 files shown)


maxonfjvipon:
@yegor256 it works, but slower than before

h1alexbel:
@yegor256 @maxonfjvipon can we merge this one? Looks like these changes required as part of `$` syntax removal

Issue #4386, opened 2025-07-20
`>> [...]` Syntax Results in Code Compilation Hang


yegor256:
This is my code:

```
[n] > app
  malloc.of > @
    8
    [m]
      while > @
        m.get.eq 42 > [i]
        m.put 42 > i
```

It compiles and then hangs forever. This is the code that works:

```
[n] > app
  malloc.of > @
    8
    [m]
      while > @
        [i] >>
          m.get.eq 42 > @
        [i] >>
          m.put 42 > @
```

Would be nice to have a more compact syntax:

```
[n] > app
  malloc.of > @
    8
    [m]
      while > @
        m.get.eq 42 >> [i]
        m.put 42 >> i
```

Here, the `>> [i]` would mean: "give this formation a random name _and_ assign object to its `@`".

yegor256:
@h1alexbel maybe you can try to help?

h1alexbel:
@yegor256 yes, I’m in

PR #4412, opened 2025-08-04
bug(#4386): auto named formation with `\phi` object inside

h1alexbel:
In this PR I've added syntax possibility to create auto name formations with inner object assigned to `@`, from the application.

see #4386

## Summary by CodeRabbit

* **New Features**
  * Introduced support for automatic phi formation in EO code, enabling automatic recognition and naming of specific object patterns based on source position.

* **Tests**
  * Added test cases to verify correct parsing, formation, and error handling of automatically named phi objects and invalid object declarations in EO code.

### DIFF
--- a/eo-parser/src/main/antlr4/org/eolang/parser/Eo.g4
+++ b/eo-parser/src/main/antlr4/org/eolang/parser/Eo.g4
@@ -436,7 +436,12 @@ onameOrTname
 
 // Object name
 oname
-    : suffix CONST?
+    : (suffix CONST?) | aphi
+    ;
+
+// Auto named assigned to PHI (`@`).
+aphi
+    : SPACE ARROW ARROW SPACE voids
     ;
 
 tname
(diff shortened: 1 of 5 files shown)


h1alexbel:
@maxonfjvipon please check

eo-parser/src/main/antlr4/org/eolang/parser/Eo.g4
```diff
@@ -152,10 +152,14 @@ happlicationReversedHead
 // Extended horizontal application
 // The head can contain elements in horizontal or vertical notations
 happlicationExtended
-    : happlicationHeadExtended happlicationTail
+    : happlicationHeadExtended happlicationTail aphi?
```
maxonfjvipon: @h1alexbel are you sure this is the right place for `aphi`? I see `happlicationExtended` is used in `tapplication` which allows `tname` after it. So it's technically possible to write something like
```
5.plus 5 >> [] +> test
```
which is wrong

I think this `aphi` concept should be a standalone rule like `onlyPhi`

maxonfjvipon (approved):

Issue #4411, opened 2025-08-04
Object idempotency is not tracked, but can be


yegor256:
What if we let EO compiler label some objects as **idempotent**. If an object is [idempotent](https://en.wikipedia.org/wiki/Idempotence), its lambda _must not_ be called more than once, since it always returns the same object.

A programmer can make any object idempotent:

```
random > r'
```

It translates to:

```
random.xi🌵 > r
```

All formations have this synthetic attribute in XMIR:

```
[] > random
  $ > xi🌵
```

Here, if `r` is idempotent, `x` too:

```
r.plus r > x
```

Finally, we can entirely get rid of `$`, in the syntax.

yegor256:
@maxonfjvipon WDYT?

h1alexbel:
@yegor256 yes, I'm in

yegor256:
@h1alexbel yes, correct

yegor256:
@h1alexbel see how it's done in https://github.com/objectionary/eo/pull/4404 The same should be done everywhere.

yegor256:
@h1alexbel yes, `$` should not be present in EO anymore. The `ξ` stays in XMIR.

h1alexbel:
@yegor256 got it, but I'm not sure how the removal of `$` will contribute to object idempotency idea

yegor256:
@h1alexbel in #4404 you see how `tuple` was using `$ > as-tuple` in order to optimize performance. We do the same with the idempotence concept. Imagine this object:

```
5.plus 6 > x
```

Now I do this:

```
x.div x > y
```

How many times `.plus` will be called when I dataize `a`? Twice. First, when we do `x.div`. Second, when `div` will use `x` to get its value.

In order to get rid of the redundant calculation, we sometimes do this, in current version of EO (sometimes it's `as-number`, or `@`, or `as-tuple`):

```
(5.plus 6).as-number > x
```

Here, when we _first time_ dataize `x`, we dispatch `.as-number` and it becomes a _new_ object that is assigned to `x`. Then, all other requests to `x` deal with this object, not with the `(5.plus 6)`. Simply put, the dot is caching in EO. Once an object is dispatched, it's remembered.

Works good, but looks ugly.

The new syntax will make the same, but more elegant:

```
5.plus 6 > x'
```

yegor256:
@h1alexbel we don't need to add `xi🌵` attribute to every formation during parsing. Only to those who have apostrophe.

[...]

PR #4452, opened 2025-08-15
bug(#4411): removed `$` usages from runtime objects

h1alexbel:
In this PR I've removed usages of `$` from runtime objects, as part of #4411. Now objects will be slower, until we implement object idempotency on runtime level.

## Summary by CodeRabbit

* **New Features**
  * Added mkdir and touch helpers, file.exists query, map entry type, and PRNG now exposes .next; added aliases for formatting and tmpdir utilities.
* **Bug Fixes**
  * More robust cross‑OS tmpdir resolution with Windows fallbacks; tightened existence and access checks across file/console/stdin/socket I/O.
* **Refactor**
  * Simplified dir/file/path/tmpdir APIs, unified port-based wiring, removed legacy conversion aliases, and streamlined read/write/data-flow propagation.

### DIFF
--- a/eo-runtime/src/main/eo/org/eolang/txt/regex.eo
+++ b/eo-runtime/src/main/eo/org/eolang/txt/regex.eo
@@ -85,6 +85,11 @@
       # - `next`          - to get next matched block
       # - `text`          - to get matched subsequence as `string`
       # - `group {index}` - to get string subsequence captured by `index`-th group.
+      # @todo #4411:35min Replace all left usages of `$ > oname`.
+      #  We should remove all the usages like `$ > matched` from the syntax, and
+      #  from the runtime objects. Currently, it is not possible because syntax sugar for object
+      #  idempotency is not fully implemented, and we cannot remove some edge case usages, like
+      #  here, in `matched`. Don't forget to remove this puzzle.
       [position start from to groups] > matched
         $ > matched
         groups.length > groups-count


h1alexbel:
@yegor256 I will remove it right after I get rid of all `$` from runtime objects. For now, only few problematic places left: `bytes`, `regex`, and `console`.

h1alexbel:
@codereabbitai is it possible to rewrite `regex.matched` object without `$ > matched`? See this:

```eo
[position start from to groups] > matched
    $ > matched
    groups.length > groups-count
    start.gte 0 > exists
    if. > next
      exists
      matched.
        matched-from-index
          position.plus 1
          to
      error "Matched block does not exist, can't get next"
    if. > text
      exists
      group 0
      error "Matched block does not exist, can't get text"
    
    # Returns the string subsequence captured by the group
    # by `index` during the `match` operation.
    groups.at index > [index] > group
```

maxonfjvipon:
@h1alexbel the `$ > matched` is caching crutch which we want to get rid of. I think we need to write a blog post but I'll give you a quick summary why and how it works. Consider this code:
```
[] > random
  [] > @ ? # returns number

[bytes] > number
  bytes > @
```

Here we have `random` object that returns random `number`, `random.@` is atom which actually calculates random number.

Then you use this object:
```
random > x
x.plus x > sum
```

When you dataize `sum`, it will dataize `x` twice, `x` is `random` which is not dataizable for now, under the hood our complier will temporary "rewrite" this code to:

```
x.@.plus x.@ > sum
```

It means that `x.@` is dataized twice, which random number will actually calculated twice and you get some unpredictable result. Imagine you want to avoid it. How? Dispatch is your friend. Why? Thanks to caching `PhOnce` object each operation on object (we have only 'dispatch' and 'application') is executed only once. !!! Object itself does not save this cached state, this state has to be attached to some other object !!! So to avoid recalculation you need to modify `number`:
```
[] > random
  [] > @ ?

[bytes] > number
  bytes > @
  $ > self
```

Now you can use it:
```
random.self > x
x.plus x > sum
```

Before any dataization or execution `x` is attached to declarative dispatch `random.self`, but since this dispatch is attached to `x`, after the first execution it'll be rewritten to: `random.@.self` -> `number.self` -> `number`. So now `x` is attached to already calculated `number`, this is exactly what you wanted to achieve. When `x` is calculated the second time, it does not go to `random.@` atom, it's already `number`. This is how this 'dispatch-caching' works. 

Right now, to use such caching you need manually to introduce such `$ > self` attributes, some objects have them, some don't. We want this feature is available by default for each formation. How? We need to add such synthetic attribute to each formation automatically, during parsing or XSL transformations. We use special `🌵` character which user can't write in EO code but attribute with such symbol is valid. So we need to add `xi🌵` to each formation:

```
[as-bytes] > number
  $ > xi🌵

[data] > bytes
  $ > xi🌵

[] > string
  $ > xi🌵
```

Then we introduce syntax sugar with apostrophe `'`:
```
random > x'
x.plus x > sum
```

This sugar does exactly the same as `.self` attribute but with `.xi🌵`:
```
random.xi🌵 > x
x.plus x > sum
```

So we get 'dispatch-caching' out of the box with pre reserved `xi🌵` attribute which is the same for each object and user don't need to think what exactly caching attribute this formation uses `as-number`, `as-tuple` or anything else. The user just does: `x'` and he knows that it uses 'dispatch-caching'

maxonfjvipon:
@h1alexbel when it's done, we should carefully look at each object and try to replace `$ > self` (or `$ > matched` in this case) with apostrophe syntax sugar. When it's done - there will be no more `$` is used in the code because it's the only way this attribute is used.

h1alexbel:
@yegor256 please check

yegor256 (approved):

Issue #4557, opened 2025-09-17
Replace `ap🌵` in auto-phi-formation-restruct.xsl Prevents Puzzle Resolution


0pdd:
The puzzle `4444-58b56423` from #4444 has to be resolved: 

https://github.com/objectionary/eo/blob/c3860bfbafc22bcdbe713e16ec30e6d47789916f/eo-parser/src/main/resources/org/eolang/parser/parse/auto-phi-formation-restruct.xsl#L8-L12

The puzzle was created by @yegor256 on 17-Sep-25. 

Estimate: 90 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
The puzzle `4444-58b56423` has disappeared from the source code, that's why I closed this issue.

Issue #4559, opened 2025-09-17
eo-parser incorrectly parses objects with synthetic names


yegor256:
Take a look: https://github.com/objectionary/eo/issues/4556#issuecomment-3302946177

The weird part is this:

```
"Can't #take("(text orgn)"), the attribute is absent among other 7 attrs of ...
```

How come `(text orgn)` is the name of the attribute?

This seems to be a bug in eo-parser.

PR #4560, opened 2025-09-17
bug(#4559): auto phi formation correctly parsed with application precedences

h1alexbel:
In this PR I've updated parsing of applications with precedences in auto Phi formations.

see #4559

## Summary by CodeRabbit

- New Features
  - Improved handling of dotted and fully-qualified application names to derive correct bases during automatic φ‑formation.

- Bug Fixes
  - Fixed base resolution and precedence for application chains (including hyphenated identifiers and ρ-related cases), avoiding incorrect ξ/ρ linking and respecting positional attributes.

- Tests
  - Added parser, transpile and integration tests covering application precedence, nested/FQN/QQ applications, hyphens, ρ scenarios, and related transpilation checks.


yegor256:
@h1alexbel some nitpick comments from Rabbit make sense, maybe you can use them

h1alexbel:
@maxonfjvipon @yegor256 this PR is currently in progress. Need to polish a few things

h1alexbel:
@maxonfjvipon reminder

eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
```diff
@@ -1210,10 +1217,19 @@ private String alphaAttr(final ParserRuleContext ctx, final String msg) {
...
      */
     private void startAutoPhiFormation(final ParserRuleContext ctx, final String application) {
+        final Matcher matcher = XeEoListener.LAST_APPLICATION.matcher(application);
+        String abase;
+        if (matcher.find()) {
+            do {
+                abase = String.format(".%s", matcher.group(1));
+            } while (matcher.find());
+        } else {
+            abase = String.format("ξ.ρ.%s", application);
```
maxonfjvipon: @h1alexbel what about this?
```
Q.org.eolang.number >> [x]
```

Are you sure it'll be parsed correctly?

maxonfjvipon (changes requested):

h1alexbel:
@maxonfjvipon reminder

eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
```diff
@@ -1210,10 +1217,25 @@ private String alphaAttr(final ParserRuleContext ctx, final String msg) {
...
+        final Matcher matcher = XeEoListener.LAST_APPLICATION.matcher(application);
+        String abase;
+        if (matcher.find()) {
+            do {
+                abase = String.format(".%s", matcher.group(1));
+            } while (matcher.find());
+        } else {
+            if (application.startsWith("Q.")) {
+                abase = String.format(
+                    "ξ.ρ.%s", application.substring(application.lastIndexOf('.') + 1)
```
maxonfjvipon: @h1alexbel are you sure we need to do `$.^.` if application starts with `Q.`? It seems to be wrong

eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
```diff
@@ -1210,10 +1217,25 @@ private String alphaAttr(final ParserRuleContext ctx, final String msg) {
      * @param application Application base
      */
     private void startAutoPhiFormation(final ParserRuleContext ctx, final String application) {
+        final Matcher matcher = XeEoListener.LAST_APPLICATION.matcher(application);
+        String abase;
+        if (matcher.find()) {
+            do {
+                abase = String.format(".%s", matcher.group(1));
+            } while (matcher.find());
+        } else {
```
maxonfjvipon, line 1233: @h1alexbel instead of 2 level `if-else` let's stick with one

maxonfjvipon (changes requested):

eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
h1alexbel: @maxonfjvipon we are removing `Q.` part till last `.`, so `Q.org.eolang.number` turns to `number`, and then to `ξ.ρ.number`, that contains `Q.org.eolang.number` inside.

maxonfjvipon: @h1alexbel I'm not sure this is the right way to parse. If I write `Q.org.eolang.number` in EO, I want to see it as is in XMIR, what do xi or rho deal with it?

h1alexbel: @maxonfjvipon yes, you are right, `Φ.org.eolang.number` is the correct variant

eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
```diff
@@ -1210,10 +1217,21 @@ private String alphaAttr(final ParserRuleContext ctx, final String msg) {
      * @param application Application base
      */
     private void startAutoPhiFormation(final ParserRuleContext ctx, final String application) {
+        final Matcher matcher = XeEoListener.LAST_APPLICATION.matcher(application);
+        String abase;
+        if (matcher.find()) {
+            do {
+                abase = String.format(".%s", matcher.group(1));
```
maxonfjvipon: @h1alexbel what if the base would be something like this: `^.x.^.y.^.z`? You'll reduce it to `z` and append `$.^.` in front of it?

h1alexbel: @maxonfjvipon we won't reduce it, but will append the base "as is" to the `ξ.ρ.`, so it looks like this: `ξ.ρ.ρ.x.ρ.y.ρ.z`. Should we remove first `ρ.` part from the start in this case, so the final one will look like `ξ.ρ.x.ρ.y.ρ.z`?

maxonfjvipon: @h1alexbel all this stuff with appending `$.^` is not correct by default, we have a corresponding XSL transformation which does all of this if it's necessary. The main part of parsing this block is to properly create formation with auto name and with `@` attribute and save everything else unchanged and leave it to the further processing

h1alexbel: @maxonfjvipon sometimes we need add redundant `ξ.ρ` in auto Phi formations to correctly processed by `auto-phi-formation-restruct.xsl`, especially on this line:

```xsl
<!-- Remove redundant ξ.X for void attributes, located inside the auto-phi formation -->
<xsl:template match="o[starts-with(@base,'ξ.') and not(@pos) and ancestor::o[contains(@name,'ap🌵')][descendant::o[@base='∅' and @name=substring-after(current()/@base,'ξ.')]]]"/>
```

Once we solve [this](https://github.com/objectionary/eo/issues/4557) puzzle, we can get rid of this sheet and appending `ξ.ρ` to the application base. WDYT? Am I missing something?

h1alexbel:
@maxonfjvipon updated. Take a look, please

eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
```diff
@@ -1210,13 +1226,44 @@ private String alphaAttr(final ParserRuleContext ctx, final String msg) {
...
+        } else if (transformed.startsWith("Q.")) {
+            base = transformed.replace("Q.", "Φ.");
+        } else {
+            base = transformed;
+        }
+        final String result;
+        if (base.startsWith("ρ.") || base.startsWith("Φ.")) {
+            result = base;
+        } else {
+            result = String.format("ξ.ρ.%s", base);
```
maxonfjvipon, line 1262: @h1alexbel again, I strongly believe there's no need to append `ξ.ρ.` before `base`, imagine such code:
```
[] > foo
  5 > x
  [] > bar
    [] > baz
      x.some >> []
```

You'll translate it to

```
[] > foo
  5 > x
  [] > bar
    [] > baz
      [] > some-auto-name
        $.^.x.some > @
```

But this is wrong, should be `$.^.^.^.x.some`, and it'll be as such, but later, on XSL transformations step

h1alexbel: @maxonfjvipon please check [this](https://github.com/objectionary/eo/pull/4560#discussion_r2378804338). How about we add this test case as disabled with puzzle?

eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
maxonfjvipon: @h1alexbel I think all of this stuff is over engineered and we're shooting to our leg by introducing such crutches. I strongly believe all of this auto naming processing can be done in XeListener without any regexp and further XSL processing, because you have two entry points for each AST element: enter and exit. It should be implemented much simpler in general

maxonfjvipon: @h1alexbel Let's resolve that puzzle first so we create right XML in XeListener for auto named formation with phi without further XSL processing and `ap` prefixes

h1alexbel: > Let's resolve that puzzle first so we create right XML in XeListener for auto named formation with phi without further XSL processing and `ap` prefixes

@maxonfjvipon in this PR, right? Or we can merge this one, and resolve #4557 in parallel?

maxonfjvipon: @h1alexbel let's resolve #4557 first

h1alexbel:
Waiting for #4557

Issue #4502, opened 2025-08-31
`QQ.structs.list.shifted` Functionality Not Supporting Tuple Element Skipping


yegor256:
Would be nice to have a way to turn a tuple into a new one, without a few upfront elements:

```
shifted. > x
  QQ.structs.list
    * "banana" "apple" "orange"
  2
```

Here, `x` equals to `* "orange"` (since two elements skipped).

pratyushkumar211:
Hello @yegor256 , I am a new open source contributor and interested in helping to resolve this issue. Can I take this up?

yegor256:
@pratyushkumar211 sure, please help us

PR #4596, opened 2025-09-30
ISSUE-4502 Add shifted functionality to list.eo

pratyushkumar211:
Add a shifting feature to address issue https://github.com/objectionary/eo/issues/4502

- Leveraged the existing slice function
- Add two tests to verify the behavior for the shift variable is less than length and another test when the shift variable is greater than length

## Summary by CodeRabbit

- New Features
  - Added a shifted operation for lists that returns a sublist after removing the first N elements.
  - Supports edge cases: when N is greater than or equal to the list length, it returns an empty list.

- Tests
  - Added tests verifying shifted with N less than the list length returns remaining elements.
  - Added tests verifying shifted with N greater than or equal to the list length returns an empty result.

### DIFF
--- a/eo-runtime/src/main/eo/org/eolang/structs/list.eo
+++ b/eo-runtime/src/main/eo/org/eolang/structs/list.eo
@@ -934,7 +934,7 @@
         list
           * 8 2 3 4 5
         2
-      list * 3 4 5
+      * 3 4 5
   
   # Tests that shifted will work if the shift parameter is greater than the list length.
   [] +> tests-shifted-with-shift-greater-than-length
@@ -943,4 +943,4 @@
         list
           * 8 2 3 4 5
         6
-      list *
+      *


yegor256:
@pratyushkumar211 thanks! In the future, don't forget to ping me when you submit a PR, see: https://www.yegor256.com/2024/04/01/ping-me-please.html

0crat:
@pratyushkumar211 Hi there! I noticed that your branch name "`add-shifting`" doesn't follow our [naming convention](https://www.yegor256.com/2014/04/15/github-guidelines.html). According to our team policy, this results in a **-6 point** deduction. Please name your branches using the ticket number you're working on - in this case, "`4595`" would be perfect. Your current balance is [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html), and don't forget to check your [Zerocracy balance](https://www.zerocracy.com/) as well. Thanks for your attention to this going forward!

Issue #4533, opened 2025-09-07
The dollar sign is still in the grammar, by may/should be removed


yegor256:
Since #4411 is fixed, let's get rid of `$` from our grammar. Let's remove it.

h1alexbel:
@maxonfjvipon @yegor256 the last place we have `$` in runtime objects is in `txt/regex.pattern.match.matched`:

```eo
[position start from to groups] > matched
  $ > matched
  groups.length > groups-count
  start.gte 0 > exists
  if. > next
    exists
    matched.
      matched-from-index
        position.plus 1
        to
    error "Matched block does not exist, can't get next"
  if. > text
    exists
    group 0
    error "Matched block does not exist, can't get text"
```

In XMIR it looks like this:

```xml
<o name="matched">
  <o base="ξ" name="matched"/>
```

After we implemented #4411, we have this attribute:

```xml
<o name="matched">
  <o base="ξ" name="xi🌵"/>
```

However, it won't replace `$ > matched` as is, since we don't have "attribute redirect" from `xi🌵` to `matched`. Am I correct about this?

maxonfjvipon:
@h1alexbel it should look like this:
```
[position start from to groups] > matched
  groups.length > groups-count
  start.gte 0 > exists
  if. > next'
    exists
    matched-from-index
      position.plus 1
      to
    error "Matched block does not exist, can't get next"
  if. > text
    exists
    group 0
    error "Matched block does not exist, can't get text"
```

Under the hood it looks like this:
```
[position start from to groups] > matched
  groups.length > groups-count
  start.gte 0 > exists
  xi🌵. > next
    if.
      exists
      matched-from-index
        position.plus 1
        to
      error "Matched block does not exist, can't get next"
  if. > text
    exists
    group 0
    error "Matched block does not exist, can't get text"
```

maxonfjvipon:
@h1alexbel inside this atom we do
```java
final Phi match = this.take(Phi.RHO);
...
result = match.take("matched");
```

Here we take `regex.pattern.match.matched` object which is fine

maxonfjvipon:
Waits for #5165

VasilevNStas:
@yegor256 

This can be closed as obsolete: the premise no longer holds after the spec-driven parser rewrite.

## Why

- The issue (Sep-2025) predates commit `769675b1a` ("replace ANTLR parser with spec-driven EO→XMIR converter", #5078).
[...]

Issue #4228, opened 2025-06-02
its not clear how to disable EO test


h1alexbel:
Consider the following EO test:

```eo
# Test.
[] +> executes-objects
  eq.
    foo "x"
    "foo x"
```

However, its not clear how to disable this test. I suggest the following syntax:

```eo
# Test.
# @todo #42:30min Enable this `executes-objects`...
# ...
[] !+> executes-objects
  eq.
    foo "x"
    "foo x"
```

Here, `!+>` means disabled test attribute, which will be annotated with `@Disabled` during the Java transpilation.

yegor256:
@h1alexbel maybe just like this:

```
# Test.
[] +> disabled
  "@todo #42:30min Enable this `executes-objects`..."
  executes-objects
    eq.
      foo "x"
      "foo x"
```

In EO, we try to get rid of any kinds of annotations and meta-syntax. Just objects.

h1alexbel:
@yegor256 agree, looks better

h1alexbel:
@yegor256 should we add this one to the backlog?

yegor256:
@h1alexbel yes, indeed

sherenkov:
@yegor256 I'd like to pick this up. Before I start, can you confirm how the transpiler should detect the disabled marker? Two readings of your proposal:

**(A) By attribute name** — the test attribute is named `disabled` (or prefixed, e.g. `disabled-executes-objects`). `to-java.xsl` adds a branch next to the existing `throws*` special-case at [`to-java.xsl:815`](https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl#L815) and emits `@Disabled` instead of `@Test`. No new EO object needed; the `@todo` text lives as a free-form string child of the attribute.

**(B) By wrapper object** — there is a real `org.eolang.disabled` object that the test binds to via `@base`. The XSLT inspects the base and emits `@Disabled` accordingly. Requires a new `eo-runtime/src/main/eo/disabled.eo` and likely a small change in `tests.xsl` to expose the wrapper.

(A) is a ~15-line XSLT change, no runtime additions. (B) is more work but feels closer to "just objects, no meta-syntax". Which one do you have in mind?

yegor256:
@sherenkov it's option (B). We should create a new object. However, I suggest naming it `noop` instead of `disabled`. This way it may be used not only in tests, but in other places. It should accept one argument:

```
[] +> noop
[...]

PR #4604, opened 2025-10-01
bug(#4533): remove `$` from syntax

h1alexbel:
In this PR I've removed `$` (`\xi`) from the syntax.

see #4533

## Summary by CodeRabbit

- Refactor
  - Removed the deprecated “$” (xi) token from the language and simplified parser behavior and path handling.
  - Consolidated dotted/chain access forms in examples and expected outputs.

- Runtime
  - Adjusted regex matcher flows and test invocation flags; no user-visible runtime behavior expected.

- Tests
  - Temporarily disabled several integration tests.
  - Updated parser/transpiler test fixtures to reflect syntax changes.
  - Added typo-validation tests for xi and removed an obsolete syntax case.
  - Standardized test visibility across runtime modules.


eo-integration-tests/src/test/java/integration/JarIT.java
```diff
@@ -19,17 +19,24 @@
...
+ *  For now these tests fail, since they fetch old objects with `$` removed syntax:
+ *  `JarIT#runsProgramFromJar`, `SnippetIT#runsAllSnippets`,
+ *  `ReadmeSnippetsIT#validatesReadmeSnippets`. We need to release new EO version without `$`
+ *  (`\xi`) syntax, and then enable tests. They all should pass.
  */
 @SuppressWarnings("JTCOP.RuleAllTestsHaveProductionClass")
 @ExtendWith(MktmpResolver.class)
 final class JarIT {
 
+    @Disabled
```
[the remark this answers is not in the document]
h1alexbel: @coderabbitai `printsErrorToStderr` is enabled

h1alexbel:
@maxonfjvipon please check

eo-runtime/src/main/eo/org/eolang/txt/regex.eo
```diff
@@ -60,7 +60,7 @@
 
     # Creates a `matcher` that will match the given input against the pattern.
     [txt] > match
-      (matched-from-index 1 0).matched > next
+      matched-from-index 1 0 > next
```
maxonfjvipon, line 63: @h1alexbel let's use `next'` here also

eo-runtime/src/main/eo/org/eolang/fs/path.eo
```diff
@@ -600,7 +600,7 @@
       "C:"
 
   # Tests that Windows path normalization replaces forward slashes with backslashes.
-  [] +> tests-normalizes-win32-path-with-replacing-slashes
+  [] > tests-normalizes-win32-path-with-replacing-slashes
```
maxonfjvipon, line 603: @h1alexbel I think it's not a good way to disable test.

eo-parser/src/main/java/org/eolang/parser/CompactArrayFqn.java
```diff
@@ -51,12 +51,6 @@ public String asString() {
...
             fqn = String.format("Φ.org.eolang.%s", name);
         }
-        final String base;
-        if (this.context.XI() == null) {
-            base = fqn;
-        } else {
-            base = String.format("ξ.%s", fqn);
-        }
-        return base;
+        return fqn;
```
maxonfjvipon, line 54: @h1alexbel It's better to leave xi in XMIR since it's IR for PHI expression also so it must be as full as possible

maxonfjvipon (changes requested):

eo-runtime/src/main/eo/org/eolang/fs/path.eo
h1alexbel: @maxonfjvipon I've created `todo` that mentions all disabled tests.

eo-runtime/src/main/eo/org/eolang/txt/regex.eo
h1alexbel: @maxonfjvipon it does not work here:

```text
Ex org.eolang.Data$ToPhi@10d23e54(Δ = [0x454F6F72-672E454F-656F6C61-6E672E45-4F747874-2E454F72-65676578-24454F70-61747465-726E2445-4F6D6174-63682445-4F6D6174-63686564-5F66726F-6D5F696E-64657840-37363031-30383063-20686173-206A7573-74203220-61747472-69627574-65287329-2C206361-6E277420-72656164-20746865-20322D74-68206F6E-65] = "EOorg.EOeolang.EOtxt.EOregex$EOpattern$EOmatch$EOmatched_from_index@7601080c has just 2 attribute(s), can't read the 2-th one")
```

Seems that it happens due to idempotency attribute `xi🌵` is not available in runtime for `Phi.take("xi🌵")`

eo-parser/src/main/java/org/eolang/parser/CompactArrayFqn.java
h1alexbel: @maxonfjvipon to me, it does not look correct to append `\xi` to base in `CompactArrayFqn`. This snippet:

```eo
[] > foo
  bar *1 > x
    y
    z
```

Will be parsed as such:

```eo
[] > foo
  ξ.bar *1 > x
    y
    z
```

However, `bar` object also can be global one, so we can't say it should be either `Q.org..bar` or `ξ.bar`. If we keep just `bar`, later, special XSL transformation will resolve that. WDYT?

eo-runtime/src/main/eo/org/eolang/txt/regex.eo
maxonfjvipon: @h1alexbel it must be available because this attribute is added before converting to Java

eo-runtime/src/main/eo/org/eolang/fs/path.eo
maxonfjvipon: @h1alexbel I saw, but disabling tests by replacing `+>` with `>` is not a good idea in general

h1alexbel: @maxonfjvipon should I comment them then?

P.S. https://github.com/objectionary/eo/issues/4228 to be implemented.

maxonfjvipon: @h1alexbel just make them decorate `true` for now

eo-runtime/src/main/eo/org/eolang/txt/regex.eo
h1alexbel: @maxonfjvipon the attribute `xi🌵` available in XMIR, but in the translation phase, we are doing this:

```xslt
<xsl:if test="not(bound/o[eo:idempotent(.)])">
  <xsl:variable name="name" select="eo:attr-name(@name, false())"/>
  <xsl:if test="not(@name)">
    <xsl:message terminate="yes">
      <xsl:text>Unnamed attribute found in </xsl:text>
      <xsl:value-of select="parent::*/@loc"/>
    </xsl:message>
  </xsl:if>
  <xsl:if test="not(contains($name, '+'))">
    <xsl:value-of select="eo:eol($indent)"/>
    <xsl:if test="$context!='this'">
      <xsl:text>((PhDefault) </xsl:text>
    </xsl:if>
    <xsl:value-of select="$context"/>
    <xsl:if test="$context!='this'">
      <xsl:text>)</xsl:text>
    </xsl:if>
    <xsl:text>.add("</xsl:text>
    <xsl:value-of select="$name"/>
    <xsl:text>", </xsl:text>
    <xsl:apply-templates select="void|bound|atom|abstract">
      <xsl:with-param name="indent" select="$indent"/>
      <xsl:with-param name="name" select="$name"/>
      <xsl:with-param name="parent" select="$parent"/>
      <xsl:with-param name="context" select="$context"/>
    </xsl:apply-templates>
    <xsl:text>);</xsl:text>
  </xsl:if>
</xsl:if>
```

If we remove this `if` check, then in runtime, we will get troubles with `put()` attributes into `xi🌵`. Is it the right direction?

maxonfjvipon: @h1alexbel why do we get troubles? We put only void attributes which must be upper than any other attributes. This xi is already bound so I don't see any troubles

Issue #4505, opened 2025-09-01
`OyRemote` Fails to Recognize Packages in Directories


yegor256:
In the `OyRemote` we return FALSE for `contains("org.eolang.io")`. However, it's a package and it does exist: https://github.com/objectionary/home/tree/master/objects/org/eolang/io

Let's fix the class, to make it find packages (directories) too.

tihiyn:
@yegor256, in the `OyRemote` class, a link to `raw.githubusercontent.com` is hardwired in its constructor:
```
OyRemote(final CommitHash hash) {
        this.template = new UrlOy(
            "https://raw.githubusercontent.com/objectionary/home/%s/objects/%s",
            hash
        );
}
```
`raw.githubusercontent.com` always returns a 404 error for directories.

yegor256:
@tihiyn maybe we can use this URL: `https://github.com/objectionary/home/tree/master/objects/org/eolang/fs` (it returns 200 for directories)

tihiyn:
@yegor256, I'm in.

yegor256:
@tihiyn I don't understand why it's "OR" in your question :) The `contains` should check whether the object is in the repository, whether it's a package or a file.

tihiyn:
@yegor256, do I need to change the logic of the `contains` in the `OyCached` classes (for example, add a separate cache for directories) and `OyIndexed` (I don't know how yet, but we can check whether the name is contained in the `objectionary.lst`. If so, is it definitely not a directory)?

yegor256:
@tihiyn this is fixed already?

tihiyn:
@yegor256, yes, in https://github.com/objectionary/eo/pull/4608. For some reason, it is not displayed in the history.

yegor256:
@tihiyn thanks!

PR #4608, opened 2025-10-03
feat(objectionary#4505): OyRemote.contains can work with directories too

tihiyn:
In this PR, I made it so that the `contains` method from `OyRemote` can accept not only programs, but also directories. I also added the `isDirectory` method, which accepts an object as input and determines whether it is a directory or not. Since `OyRemote` is used in `OyCached` and in `OyIndexed`, I have refined the logic for them as well. Tests have been written for all changes.
**Resolves**: #4505

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Objectionary.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Objectionary.java
@@ -35,6 +35,15 @@ interface Objectionary {
     boolean contains(String name) throws IOException;
 
     /**
+     * Checks whether a provided object is a directory.
+     *
+     * @param name Object name.
+     * @return Boolean: "true" if directory, "false" if not.
+     * @throws IOException If fails to fetch.
+     */
+    boolean isDirectory(String name) throws IOException;
+
+    /**
      * Objectionary with lambda-function Ctor-s for testing.
      *
      * @since 0.28.11
@@ -52,6 +61,11 @@ interface Objectionary {
          * Function that emulates 'contains()' method in {@link Objectionary}.
          */
         private final Func<? super String, Boolean> container;
+
+        /**
+         * Function that emulates 'isDirectory' method in {@link Objectionary}.
+         */
+        private final Func<? super String, Boolean> checker;
 
         /**
          * Ctor.
@@ -76,7 +90,8 @@ interface Objectionary {
         Fake(final Func<? super String, ? extends Input> gett) {
             this(
                 gett,
-                s -> true
+                s -> true,
+                s -> false
             );
         }
 
@@ -85,13 +100,16 @@ interface Objectionary {
          *
          * @param gett Lambda func for get()
          * @param cont Lambda func for contains()
+         * @param chckr Lambda func for isDirectory()
          */
         Fake(
             final Func<? super String, ? extends Input> gett,
-            final Func<? super String, Boolean> cont
+            final Func<? super String, Boolean> cont,
+            final Func<? super String, Boolean> chckr
         ) {
             this.getter = gett;
             this.container = cont;
+            this.checker = chckr;
         }
 
         @Override
@@ -102,6 +120,11 @@ interface Objectionary {
         @Override
         public boolean contains(final String name) {
             return new Unchecked<>(() -> this.container.apply(name)).value();
+        }
+
+        @Override
+        public boolean isDirectory(final String name) {
+            return new Unchecked<>(() -> this.checker.apply(name)).value();
         }
 
         @Override
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/OyCached.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/OyCached.java
@@ -21,9 +21,14 @@ final class OyCached implements Objectionary {
     private final Objectionary origin;
 
     /**
-     * The cache.
+     * The cache for programs.
      */
-    private final Map<String, Input> cache;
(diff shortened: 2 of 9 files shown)


eo-maven-plugin/src/main/java/org/eolang/maven/OyCached.java
```diff
@@ -61,6 +78,24 @@ public Input get(final String name) throws IOException {
...
+
+    @Override
+    public boolean isDirectory(final String name) throws IOException {
+        return this.directories.computeIfAbsent(
+            name, key -> {
+                try {
+                    return this.origin.isDirectory(name);
+                } catch (final IOException exception) {
+                    throw new IllegalStateException(
+                        "An error occurred during the access to the origin objectionary",
```
yegor256: @tihiyn read this: https://www.yegor256.com/2015/12/01/rethrow-exceptions.html

eo-maven-plugin/src/main/java/org/eolang/maven/OyRemote.java
```diff
@@ -67,10 +80,21 @@ public Input get(final String name) throws MalformedURLException {
     }
 
     @Override
-    @RetryOnFailure(delay = 1L, unit = TimeUnit.SECONDS)
     public boolean contains(final String name) throws IOException {
-        final int code = ((HttpURLConnection) this.template.value(name).openConnection())
-            .getResponseCode();
+        final URL file = this.program.value(name);
+        final URL dir = this.directory.value(name);
+        return exists(file) || exists(dir);
```
maxonfjvipon: @tihiyn let's do `OyRemote.exists(file)` instead of just `exists(file)`.
@yegor256 shouldn't checkstyle or PMD catch it?

eo-maven-plugin/src/main/java/org/eolang/maven/OyRemote.java
```diff
@@ -67,10 +80,21 @@ public Input get(final String name) throws MalformedURLException {
...
+    }
+
+    @Override
+    public boolean isDirectory(final String name) throws IOException {
+        final URL url = this.directory.value(name);
+        return exists(url);
```
maxonfjvipon: @tihiyn let's do `OyRemote.exists(file)` instead of just `exists(file)`.

eo-maven-plugin/src/main/java/org/eolang/maven/OyIndexed.java
```diff
@@ -66,4 +66,23 @@ public boolean contains(final String name) throws IOException {
...
     }
+
+    // @checkstyle IllegalCatchCheck (7 line)
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    @Override
+    public boolean isDirectory(final String name) throws IOException {
+        boolean result;
+        try {
+            result = !this.index.contains(name) && this.delegate.isDirectory(name);
+        } catch (final Exception ex) {
```
maxonfjvipon, line 77: @tihiyn read this: https://www.yegor256.com/2015/12/01/rethrow-exceptions.html
Also it's not a good idea to throw just an `Exception`, it's better to use some specific exception instead

eo-maven-plugin/src/main/java/org/eolang/maven/Objectionary.java
```diff
@@ -104,6 +122,11 @@ public boolean contains(final String name) {
             return new Unchecked<>(() -> this.container.apply(name)).value();
         }
 
+        @Override
+        public boolean isDirectory(final String name) {
+            return new Unchecked<>(() -> this.checker.apply(name)).value();
```
maxonfjvipon: @tihiyn instead of creating `new Unchecked()` each time on `isDirectory` method call let's create this `Unchecked` in constructor and just call `.value()`

eo-maven-plugin/src/main/java/org/eolang/maven/Objectionary.java
```diff
@@ -53,6 +62,11 @@ final class Fake implements Objectionary {
          */
         private final Func<? super String, Boolean> container;
 
+        /**
+         * Function that emulates 'isDirectory' method in {@link Objectionary}.
+         */
+        private final Func<? super String, Boolean> checker;
```
maxonfjvipon: @tihiyn `checker` is a bad name for an object, read this: https://www.yegor256.com/2015/03/09/objects-end-with-er.html

maxonfjvipon (changes requested):

eo-maven-plugin/src/main/java/org/eolang/maven/OyRemote.java
yegor256: @maxonfjvipon it should...

eo-maven-plugin/src/main/java/org/eolang/maven/Objectionary.java
tihiyn: @maxonfjvipon, it's impossible, because lambda in `Unccheked` ctor depends on `name` which won't exist in `OyFake` ctor

eo-maven-plugin/src/main/java/org/eolang/maven/OyIndexed.java
tihiyn: @maxonfjvipon, I agree, but `this.index.contains(name)` throws exactly `Exception`. To compensate for this, I tried to make the exception message more detailed

yegor256:
@rultor merge

PR #4766, opened 2025-12-17
fix(#4765): redirect `fibonacci` integration test outputs to correct directory

### LINKED ISSUE #4765: `fibonacci` Integration Test Outputs to Wrong Directory, Hindering `mvn clean` and `qulice` Operations
Currently, the `fibonacci` integration test compiles all the classes into it's own directory:

```
eo-integration-tests/src/it/fibonacci/target/
```

it should compile all the classes to the following dir instead:

```
eo-integration-tests/target/it/
```

It will allow us to clean it's results using `mvn clean` and what is the most important, it will prevent `qulice` from checking `java` files generated by `eo-maven-plugin`.

volodya-lombrozo:
This PR updates the `fibonacci` integration test to compile outputs to the correct directory, ensuring compatibility with `mvn clean` and `qulice`.

Fixes #4765

The key change in this pr is:

```
-        <configuration combine.self="override">
+        <configuration combine.self="append">
```

by doing this we append default settings that are set in the `jcabi-paernt.xml`. The default settings from `jcabi-parent.xml` redirect `maven-invoker-plugin` output to the `target/it` folder. 

This changes allow to use `mvn qulice:check -Pqulice` without getting `TimeoutException`.

### DIFF
--- a/eo-integration-tests/pom.xml
+++ b/eo-integration-tests/pom.xml
@@ -207,7 +207,7 @@
       </plugin>
       <plugin>
         <artifactId>maven-invoker-plugin</artifactId>
-        <configuration combine.self="override">
+        <configuration combine.self="append">
           <skipInstallation>${skipITs}</skipInstallation>
           <skipInvocation>${skipITs}</skipInvocation>
         </configuration>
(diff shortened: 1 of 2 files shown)


volodya-lombrozo:
@yegor256 would you mind having a look?

volodya-lombrozo:
> @volodya-lombrozo you removed many checks. By mistakes?

@yegor256 no I did it intentionally, because they rely on specific placement of some artefacts and these checks rather fragile and check some intermediate results rather than the final output. That said, I can return them back. Would it be better?

PR #4915, opened 2026-03-05
fix(#4103): Use Xnav.strict for class/@name lookup

### LINKED ISSUE #4103: the `.strict()` method from Xnav is not used
A new method has been introduced: https://github.com/volodya-lombrozo/xnav/issues/100 Let's use it right after `.path()`, to fail faster.

nikdenalb:
## Summary
- Use Xnav.strict("/object/class/@name", 1) to fail fast when XMIR contains multiple matches.

Fixes #4103

## Test plan
- mvn -pl eo-parser -Pqulice test

## Summary by CodeRabbit

* **Bug Fixes**
  * Parsing now uses stricter validation for deriving object names: it fails fast when multiple class names are present and yields null when no name can be derived, improving error clarity.

* **Tests**
  * Added and updated tests to verify detection and reporting of missing or multiple class-name occurrences, asserting error messages include the actual count found.

### DIFF
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjParseTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjParseTest.java
@@ -242,7 +242,7 @@ final class MjParseTest {
             ),
             XhtmlMatchers.hasXPaths(
                 "//error[@severity='critical']",
-                "//error[text()=\"XMIR should have either '/object/o/@name' or '/object/class/@name' attribute\"]"
+                "//error[@check='mandatory-object-name'][contains(text(), 'found 0')]"
             )
         );
     }
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjTranspileTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjTranspileTest.java
@@ -117,9 +117,7 @@ final class MjTranspileTest {
             "TranspileMojo should throw an exception with detailed message on invalid EO code",
             writer.toString(),
             Matchers.allOf(
-                Matchers.containsString(
-                    "XMIR should have either '/object/o/@name' or '/object/class/@name' attribute"
-                ),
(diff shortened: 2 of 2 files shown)


eo-parser/src/main/java/org/eolang/parser/OnDefault.java
```diff
@@ -84,10 +84,21 @@ private Optional<String> name() {
...
-                        .orElse(null)
+                    () -> {
+                        final String xpath = "/object/class/@name";
+                        String result = null;
+                        try {
+                            result = this.xnav.strict(xpath, 1)
+                                .findFirst()
+                                .flatMap(Xnav::text)
+                                .orElse(null);
+                        } catch (final IllegalStateException ex) {
```
volodya-lombrozo: @nikdenalb something strange is happening here. Why do we need this `try-catch` here? Can't we just replace `path` with `strict`?

nikdenalb: @volodya-lombrozo you’re right — I overthought it trying to preserve the old behavior/tests. I followed your suggestion and replaced .path() logic with Xnav.strict("/object/class/@name", 1) directly. Because this changes the “0 matches” case to fail fast with strict()’s message, I also updated the dependent tests (including eo-maven-plugin ones). Could you please re-review?

volodya-lombrozo:
@nikdenalb Looks good to me, thanks!

yegor256:
@nikdenalb thanks!

0crat:
@nikdenalb **Great work on your contribution!** 🎉 You've earned **+12 points** (+16 base, -4 for 35 hits-of-code). [Keep the momentum going](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) - your running score is now [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html). Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates!

PR #5023, opened 2026-04-22
#4673 Make AtWithRho.get() thread-safe

### LINKED ISSUE #4673: Weird behaviour in PhRho.take method
While working on the task #4672 and investigating code in eo-runtime I found this weird peace of code in PhRho.java

```java
    @Override
        if (this.rho.get() == null) {
            throw new ExUnset(
                String.format("The \"%s\" attribute is not set", Phi.RHO)
            );
        }
        return this.rho.get();
    }
```

This is the method that is supposed to take attribute by name from PhRho object, but just `this.rho.get()` is returned instead. Looks confusing

@yegor256 @maxonfjvipon WDYT?

idrisoffrinat-cpu:
## Summary

Fixes the race condition in `AtWithRho.get()` described in #4673. Previously, when the underlying object lacked RHO, concurrent callers would each:
1. Pass the `!ret.hasRho()` check
2. Create their own copy via `ret.copy()`
3. Set RHO on their private copy

…so each thread received a **different** `Phi` instance, violating the expectation of a consistent view.

## Fix

- Cache the first fixed-up result in an `AtomicReference<Phi>` guarded by a `ReentrantLock` (double-checked locking — same pattern used in `PhCached` / `PhOnce`)
- `put()` invalidates the cache (`cached.set(null)`) so a subsequent `get()` reflects the new underlying value
- Removed the `This attribute is NOT thread safe!` note from the class javadoc

## Test plan

- [x] Added `returnsSameInstanceToConcurrentCallers` — spawns 16 threads hitting `get()` simultaneously from a shared start-gate and asserts exactly one distinct instance is returned
- [ ] All existing `AtWithRhoTest` cases still pass
- [ ] `mvn test -pl eo-runtime` passes with no regressions

Closes #4673

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrency so simultaneous callers reliably receive a single initialized cached instance.
  * Cache invalidation now clears prior cached values before updates so subsequent reads reinitialize correctly.

* **Documentation**
  * Updated class documentation to reflect the new safe, cached behavior and removed the prior thread-safety warning.

* **Tests**
  * Added a concurrency test verifying simultaneous callers observe the same instance and utilities to coordinate concurrent checks.


idrisoffrinat-cpu:
Thanks @coderabbitai for flagging the issue-link mismatch. To clarify:

- Issue #4673 is about the confusing `PhRho.take()` method and is closed.
- This PR fixes a different concern — a data race in `AtWithRho.get()` where, under concurrent access, the first-time lazy `\rho` assignment could be observed partially initialized or executed twice on the same attribute, producing two distinct `Phi` instances for what should be one attribute value.

The `#4673` in the title/branch is carried over from where the issue was first observed while working adjacent to that area, but the fix scope is `AtWithRho`, not `PhRho`. Happy to retitle or open a dedicated tracking issue if preferred — please advise.

Also just pushed a follow-up commit fixing the 8 qulice violations that broke the previous build (nested-if depth in `AtWithRho`, bracket formatting / too-many-asserts / unclosed `ExecutorService` in the test). The test now uses try-with-resources on the pool and a private helper to collect the distinct instances observed across threads.

eo-runtime/src/main/java/org/eolang/AtWithRho.java
```diff
@@ -49,16 +56,48 @@ public Attr copy(final Phi self) {
...
+        Phi ret = this.cached.get();
+        if (ret == null) {
+            ret = this.initialize();
         }
         return ret;
     }
 
     @Override
     public void put(final Phi phi) {
+        this.cached.set(null);
```
gemshrine, line 68: Order is invalidate-then-store; that races with a concurrent `get()` and can leave the cache permanently populated with the pre-put value:

```
T1  put(newPhi)
    cached.set(null)                 -- old cache dropped
                                        (context switch)
T2  get()
    cached.get() → null              -- observes T1's invalidation
    initialize() locks
      cached.get() → null            -- DCL sees null
      original.get() → oldPhi        -- T1 hasn't stored newPhi yet
      cached.set(withRho(oldPhi))    -- caches oldPhi under lock
    unlock, returns oldPhi
T1  original.put(newPhi)             -- store lands now

T3  get() → cached.get() → oldPhi    -- stale, forever, until the
                                        next put()
```

Either the whole `put()` needs to serialise with `initialize()` on `this.lock`, or the invalidation has to trail the store (and even then a transient stale read is possible until the invalidate lands — but it converges after `put()` returns, unlike the current order which doesn't):

```java
    this.lock.lock();
    try {
        this.original.put(phi);
        this.cached.set(null);
    } finally {
        this.lock.unlock();
    }
}
```

eo-runtime/src/test/java/org/eolang/AtWithRhoTest.java
```diff
@@ -98,4 +103,61 @@ void copiesWithNewRhoMustCallCopyOnOriginal() {
...
+        MatcherAssert.assertThat(
+            "AtWithRho.get() must return the same instance for all concurrent callers",
+            AtWithRhoTest.distinctConcurrentGets(
+                new AtWithRho(
+                    new AtComposite(new PhDefault(), phi -> phi),
+                    new PhDefault()
+                ),
+                16
+            ),
+            Matchers.is(1)
```
gemshrine, line 118: The concurrent test exercises only the read-side race that #4673 describes. The write-side race (`put()` interleaving with `initialize()` — see comment on `AtWithRho.java:68`) has no coverage. A companion test would race a `put(newPhi)` against a burst of `get()` calls and assert that once `put()` returns, every subsequent `get()` observes the new value:

```java
@Test
void reflectsPutAfterConcurrentGets() throws InterruptedException {
    final Phi first  = new PhDefault();
    final Phi second = new PhDefault();
    final AtWithRho attr = new AtWithRho(new AtSimple(first), new PhDefault());
    // race N gets against a put(second), wait for the executor to close,
    // then assert one further attr.get() returns something with the same
    // identity as second (not first).
}
```

Without that, the very ordering bug the fix introduces sits under the current test's blind spot.

eo-runtime/src/main/java/org/eolang/AtWithRho.java
```diff
@@ -49,16 +56,48 @@ public Attr copy(final Phi self) {
...
+    /**
+     * Initialize the cached value under the lock, using double-checked locking.
+     * @return Cached {@link Phi}
+     */
+    private Phi initialize() {
+        this.lock.lock();
+        try {
+            Phi ret = this.cached.get();
+            if (ret == null) {
+                ret = this.withRho(this.original.get());
```
gemshrine, line 81: `this.original.get()` runs while `this.lock` is held. If `this.original` is itself a lock-taking `Attr` — another `AtWithRho`, or any `Attr` backed by mutex-guarded state — a call path that acquires those locks in the reverse order (e.g. some other thread taking `original`'s lock first and then trying to reach this `AtWithRho.get()`) deadlocks. Publishing via CAS keeps the fast path lock-free and avoids the nested-lock exposure at the cost of possibly running `withRho(...)` more than once under contention:

```java
private Phi initialize() {
    final Phi computed = this.withRho(this.original.get());
    if (this.cached.compareAndSet(null, computed)) {
        return computed;
    }
    return this.cached.get();
}
```

The `ReentrantLock` and its field then drop out entirely. Trade-off is duplicated `withRho` work under contention vs. the deadlock exposure — worth deciding rather than defaulting.

gemshrine (commented):
The DCL-cache closes #4673's read race, but `put()` opens a symmetric write race that can leave the cache permanently populated with the pre-put value. Details plus a coverage-gap note in the inline comments.

yegor256:

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+792](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5042, opened 2026-04-23
refactor(#4750): implement number.floor in EO

syncended:
Closes: #4750

Replaces the Java atom `EOnumber$EOfloor` with an EO-level implementation that reuses existing atoms:

```eo
[] > floor
  as-i64.as-number > @
```

Semantically identical to the removed Java code (`double → longValue() → double`, i.e. truncation toward zero — matches the pre-existing `tests-div-with-remainder` test that expects `floor(13/-5) = -2`).

See also: #5043 (separate behavioral fix for `floor` rounding semantics).

## Changes
- `eo-runtime/src/main/eo/number.eo`: replace abstract atom `[] > floor ?` with the EO expression; add three unit tests (`tests-floor-positive-fraction`, `tests-floor-negative-fraction`, `tests-floor-of-integer`).
- `eo-runtime/src/main/java/org/eolang/EOnumber$EOfloor.java`: deleted.
- `eo-runtime/src/test/java/org/eolang/EOnumberTest.java`: drop `EOnumber$EOfloor.class` from the `throwsCorrectErrorForRhoAttr` parameterized list (the atom no longer exists).
- `eo-integration-tests/pom.xml`: temporarily override `maven-invoker-plugin` with `skipInstallation=true` / `skipInvocation=true` (with `@todo #4750:30min`). Same pattern as PR #5035 for issue #4944: the invoker sandbox pulls eo-runtime EO sources from the remote objectionary, which still carries the old atom-based `floor`, so its transpilation references the removed `EOnumber$EOfloor` class and fails to compile.
- `eo-integration-tests/src/test/java/integration/JarIT.java`: add `@Disabled` on the class with the same `@todo #4750:30min` rationale — JarIT's sandbox hits the exact same issue via the remote pull.

Both TODOs are expected to be reverted by a follow-up PR once a fresh release catches the remote up with the EO-level `floor`.

## Test plan
- [x] `mvn -pl eo-runtime clean test -Deo.skipLinting=true` → 1244 tests, 0 failures, 0 errors
- [x] New floor tests pass in `EOnumberEOAtomTest`
- [x] Existing `tests-div-with-remainder` and `is-integer` tests pass
- [x] `positive-infinity.floor` / `negative-infinity.floor` tests pass
- [x] `eo-integration-tests` compiles with `@Disabled` applied

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * The floor function is now fully implemented and available for numeric operations.

[...]

PR #5048, opened 2026-04-24
fix: number.floor rounds toward -infinity (not toward zero)

### LINKED ISSUE #5043: `number.floor` wrongly truncates towards zero for negative decimal numbers, contrary to documentation
### Summary

`number.floor` performs truncation toward zero, but its own documentation and its name imply mathematical floor (rounding toward -∞). The behaviors differ for negative non-integer inputs.

### Documentation vs. implementation

The doc comment at [`eo-runtime/src/main/eo/number.eo:112-114`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number.eo#L112-L114) says:

> The object rounds down the original `number` to the nearest whole `number` that is **less than or equal to it**

— i.e. mathematical floor, same as Python's `math.floor`, Java's `Math.floor`, JS's `Math.floor`, Ruby's `Float#floor`, etc.

The current implementation (both the Java atom that existed before #4750, and the EO `as-i64.as-number` replacement introduced by #5042) uses `longValue()` under the hood, which truncates toward zero. For negative non-integers these differ:

| Input                  | Expected (mathematical floor) | Actual (truncate toward zero) |
|------------------------|-------------------------------|-------------------------------|
| `3.7`                  | `3`                           | `3`                           |
| `-3.7`                 | `-4`                          | `-3` ← wrong                  |
| `-0.1`                 | `-1`                          | `0`  ← wrong                  |
| `13.div -5` (= `-2.6`) | `-3`                          | `-2` ← wrong                  |

### Impact on existing tests

The test `tests-div-with-remainder` at [`number.eo:349-354`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number.eo#L349-L354) currently asserts `floor(13 / -5) = -2`, which documents and locks in the incorrect behavior. Fixing floor requires updating this test to `-3`.

### Suggested fix

```eo
[] > floor
  if. > @
    trunc.gt $
    trunc.minus 1
    trunc
  as-i64.as-number > trunc
```

Idea: truncation (`as-i64.as-number`) equals mathematical floor for every input *except* negative non-integers, where it rounds one step toward zero — i.e. the truncated value is strictly greater than the original. In that single case, subtract 1.

Sanity check:

| Input  | `trunc` | `trunc > $`? | Result |
|--------|---------|--------------|--------|
| `3.7`  | `3`     | no           | `3`    |
| `-3.7` | `-3`    | yes          | `-4`   |
| `-5`   | `-5`    | no           | `-5`   |
| `-5.0` | `-5`    | no           | `-5`   |
| `0`    | `0`     | no           | `0`    |
| `-0.1` | `0`     | yes          | `-1`   |

Note: do **not** reuse `is-integer` here — it is defined as `is-finite and eq floor`, which would recurse into `floor` infinitely.

### Non-finite inputs

NaN and ±Infinity need explicit handling. The existing `tests_*_infinity_floor_is_equal_to_self` tests assume `(+∞).floor = +∞` and `(-∞).floor = -∞`. Today they pass only by accident (`longValue()` on an infinite double produces `Long.MIN/MAX_VALUE`, which survives round-trip only because the tests happen to compare in a specific way). The fix should short-circuit: if the input is not finite, return it unchanged; if it's NaN, return NaN.

This is a behavioral fix, so it should be a separate PR from #5042 (which is a pure refactor).

mvanhorn:
## Summary

Changes `number.floor` in `eo-runtime/src/main/eo/number.eo` from truncation-toward-zero to mathematical floor (round toward -∞), matching the existing doc comment ("less than or equal to it") and the behavior of `Math.floor` in Java / Python / JS / Ruby.

Fixes #5043.

## Why this matters

The issue (syncended) traces it to the doc/implementation mismatch:

> `number.floor` performs truncation toward zero, but its own documentation and its name imply mathematical floor (rounding toward -∞).

| Input                  | Expected | Before  |
|------------------------|----------|---------|
| `3.7.floor`            | `3`      | `3`     |
| `-3.7.floor`           | `-4`     | `-3` ← wrong |
| `-0.1.floor`           | `-1`     | `0`  ← wrong |
| `floor. 13.div -5`     | `-3`     | `-2` ← wrong |

PR #5042 (merged this morning) refactored the Java atom into an EO expression (`as-i64.as-number`) but preserved the truncation-toward-zero behavior — that PR's description explicitly notes "See also: #5043 (separate behavioral fix for `floor` rounding semantics)." This PR is that separate behavioral fix.

## Change

`eo-runtime/src/main/eo/number.eo` floor body from:

```eo
[] > floor
  as-i64.as-number > @
```

to:

```eo
[] > floor
  if. > @
    is-finite.not
    $
    if.
      trunc.gt $
      trunc.minus 1
      trunc
  as-i64.as-number > trunc
```

Non-finite inputs (NaN, +∞, -∞) short-circuit via `is-finite.not` and are returned unchanged — this preserves the behavior the existing `tests-positive-infinity-floor-is-equal-to-self` and `tests-negative-infinity-floor-is-equal-to-self` tests codify. `is-nan` is not a separate case because `is-finite` already returns false for NaN, and the `$`-return branch returns NaN unchanged.

For finite inputs: `trunc` is the integer part (toward zero). `trunc > $` is true only for negative non-integers (-3.7 → trunc=-3 > -3.7); in that one case subtract 1 to round toward -∞. For positive non-integers, zero, and integers, `trunc` is already the correct floor.

`is-finite` is safe to reuse here because its definition at `number.eo:23-28` (`is-nan.not and. not. or. eq positive-infinity eq negative-infinity`) does not recurse through floor. The issue explicitly warns against reusing `is-integer` because that one is defined as `is-finite and eq floor` and would recurse.

## Tests

Two existing tests codified the buggy truncation-toward-zero behavior — both needed updating:

- `tests-div-with-remainder`: `floor(13 / -5)` now expects `-3` (was `-2`)
- `tests-floor-negative-fraction`: `-3.7.floor` now expects `-4` (was `-3`); comment updated from "truncates toward zero" to "rounds toward -infinity"

Two new regression tests added next to the existing floor tests:

- `tests-floor-small-negative-fraction`: `-0.1.floor = -1` (guards the case where truncation would hit zero)
- `tests-floor-of-negative-integer`: `-42.floor = -42` (guards against over-eager subtraction in the integer case)

Existing `tests-floor-positive-fraction` (`3.7.floor = 3`) and `tests-floor-of-integer` (`42.floor = 42`) continue to pass unchanged.

## Testing

I was not able to complete a full local `mvn test -Dtest=EOnumberEOAtomTest` run in a reasonable window on my environment (~10+ min into the build without finishing). Trusting upstream CI to catch any EO syntax issues in the new `if.` nesting. Happy to iterate on the review.

## AI disclosure

This contribution was developed with AI assistance (Claude Code).

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the `floor` function to correctly handle edge cases and improve mathematical accuracy. Non-finite numbers are now returned unchanged, while negative numbers and fractions are properly rounded toward negative infinity, ensuring correct behavior across all input types.

### DIFF
--- a/eo-runtime/src/main/eo/number.eo
+++ b/eo-runtime/src/main/eo/number.eo
@@ -118,9 +118,17 @@
       is-finite.not
       $
       if.
-        trunc.gt $
-        trunc.minus 1
-        trunc
+        or.
+          gte 9223372036854775808.0
+          lte -9223372036854775808.0
+        $
+        if.
+          as-bytes.eq -0.as-bytes
+          $
+          if.
+            trunc.gt $
+            trunc.minus 1
+            trunc
     as-i64.as-number > trunc
 
   # Tests that less-than comparison returns true when first number is smaller.
@@ -390,6 +398,26 @@
     eq. > @
       -42.floor
       -42
+
+  # Tests that floor of NaN returns NaN.
+  [] +> tests-floor-of-nan
+    nan.floor.is-nan > @
+
+  # Tests that floor of positive infinity returns positive infinity.
+  [] +> tests-floor-of-positive-infinity
+    positive-infinity.floor.eq positive-infinity > @
+
+  # Tests that floor of negative infinity returns negative infinity.
+  [] +> tests-floor-of-negative-infinity
+    negative-infinity.floor.eq negative-infinity > @
+
+  # Tests that floor of a large positive number returns the number itself.
+  [] +> tests-floor-of-large-number
+    9223372036854775808.0.floor.eq 9223372036854775808.0 > @
+
+  # Tests that floor of negative zero preserves its sign.
+  [] +> tests-floor-of-negative-zero
+    -0.0.floor.as-bytes.eq -0.0.as-bytes > @
 
   # Tests that division result can be less than one.
   [] +> tests-div-less-than-one
--- a/eo-runtime/src/main/eo/number.eo
+++ b/eo-runtime/src/main/eo/number.eo
@@ -112,8 +112,16 @@
   # The object rounds down the original `number` to the nearest
   # whole `number` that is less than or equal to it
   # and returns the result as `org.eolang.number`.
+  # For non-finite inputs (NaN, ±infinity) the input is returned unchanged.
   [] > floor
-    as-i64.as-number > @
+    if. > @
+      is-finite.not
+      $
+      if.
+        trunc.gt $
+        trunc.minus 1
+        trunc
+    as-i64.as-number > trunc
 
   # Tests that less-than comparison returns true when first number is smaller.
   [] +> tests-int-less-true
@@ -351,7 +359,7 @@
     eq. > @
       floor.
         13.div -5
-      -2
+      -3
 
   # Tests that floor of a positive fractional number truncates toward zero.
   [] +> tests-floor-positive-fraction
@@ -359,17 +367,29 @@
       3.7.floor
       3
 
-  # Tests that floor of a negative fractional number truncates toward zero.
+  # Tests that floor of a negative fractional number rounds toward -infinity.
   [] +> tests-floor-negative-fraction
     eq. > @
       -3.7.floor
-      -3
+      -4
+
+  # Tests that floor of a small negative fraction rounds toward -infinity, not zero.
+  [] +> tests-floor-small-negative-fraction
+    eq. > @
+      -0.1.floor
+      -1
 
   # Tests that floor of a whole number returns the same number.
   [] +> tests-floor-of-integer
     eq. > @
       42.floor
       42
+
+  # Tests that floor of a negative whole number returns the same number.
+  [] +> tests-floor-of-negative-integer
+    eq. > @
+      -42.floor
+      -42
 
   # Tests that division result can be less than one.
   [] +> tests-div-less-than-one


eo-runtime/src/main/eo/number.eo
```diff
@@ -112,8 +112,16 @@
   # The object rounds down the original `number` to the nearest
   # whole `number` that is less than or equal to it
   # and returns the result as `org.eolang.number`.
+  # For non-finite inputs (NaN, ±infinity) the input is returned unchanged.
   [] > floor
-    as-i64.as-number > @
+    if. > @
+      is-finite.not
```
gemshrine, line 118: This `is-finite.not` early-return branch has no test. The four new/updated tests below all use finite inputs (`3.7`, `-3.7`, `-0.1`, `42`, `-42`, `13.div -5`), so the branch is exercised only by fall-through when a caller happens to feed it NaN or ±Inf. Any regression here — condition flipped, `.not` dropped, wrong operand order — routes non-finite inputs straight into `as-i64.as-number`, where `NaN.as-i64` returns `0` and `±Inf.as-i64` saturates to `Long.MIN_VALUE`/`MAX_VALUE`. Result: `NaN.floor` would silently become `0` and `+inf.floor` would flip sign, both without any test noticing.

The infra is already there — see `simple-number-is-finite` at line 436 and the `positive-infinity.as-bytes.is-finite.not` case at line 444 for the exact shape. Three assertions cover the branch:

```eo
# Tests that floor preserves NaN because floor of NaN is NaN.
[] +> tests-floor-of-nan-is-nan
  (number nan.as-bytes).floor.as-bytes.eq nan.as-bytes > @

# Tests that floor preserves positive infinity.
[] +> tests-floor-of-positive-infinity-is-positive-infinity
  (number positive-infinity.as-bytes).floor.as-bytes.eq positive-infinity.as-bytes > @

# Tests that floor preserves negative infinity.
[] +> tests-floor-of-negative-infinity-is-negative-infinity
  (number negative-infinity.as-bytes).floor.as-bytes.eq negative-infinity.as-bytes > @
```

Without this, the docstring at line 115 ("For non-finite inputs (NaN, ±infinity) the input is returned unchanged") is an unenforced comment, not a spec.

eo-runtime/src/main/eo/number.eo
```diff
@@ -112,8 +112,16 @@
...
+      $
+      if.
+        trunc.gt $
+        trunc.minus 1
+        trunc
+    as-i64.as-number > trunc
```
gemshrine, line 132: `as-i64.as-number` silently overflows for `|x| > 2^63 ≈ 9.22e18`. Java `(long) 1e19` saturates to `Long.MAX_VALUE = 9223372036854775807L`, so for a large-magnitude finite input the returned `trunc` is wildly off, and the follow-up `trunc.gt $` comparison at line 121 becomes noise — it may or may not fire the `.minus 1` branch depending on which side of the saturation the input lands. Either result is bogus.

The pre-existing code had the same silent bug, but this PR is the natural place to at least fence it because now there's already a non-finite guard right above. Two ways to plug it:

1. **Widen the guard** — treat `|x| ≥ 2^63` the same as non-finite ("the floor is representationally the input") and return `$`:

   ```eo
   [] > floor
     if. > @
       or.
         is-finite.not
         is-out-of-i64-range
       $
       if.
         trunc.gt $
         trunc.minus 1
         trunc
     as-i64.as-number > trunc
     or.
       $.gte 9223372036854775808.0
       $.lte -9223372036854775808.0
       > is-out-of-i64-range
   ```

2. **Document + pin** — leave the behaviour but add a test at the boundary (`(2.0.pow 63).floor` or a hard-coded near-max double) that asserts the *current* return value, so anyone later refactoring `as-i64` semantics sees the trip-wire.

Unpinned, this is a landmine: the same input that used to truncate-to-Long.MAX will now sometimes flip to Long.MAX − 1 because the `trunc.gt $` branch will fire on saturation, and no test will notice.

eo-runtime/src/main/eo/number.eo
```diff
@@ -112,8 +112,16 @@
...
   # whole `number` that is less than or equal to it
   # and returns the result as `org.eolang.number`.
+  # For non-finite inputs (NaN, ±infinity) the input is returned unchanged.
   [] > floor
-    as-i64.as-number > @
+    if. > @
+      is-finite.not
+      $
+      if.
+        trunc.gt $
```
gemshrine: `-0.0.floor` regresses to positive zero under this shape: `trunc = as-i64.as-number = 0` (Java `long` has no signed zero), then `0.gt -0.0` is `false` because IEEE-754 defines `0.0 == -0.0`, so the else branch fires and returns `trunc = 0`. Java's `Math.floor(-0.0d)` deliberately returns `-0.0` — the sign is a load-bearing bit for downstream `atan2` / `copysign` / `divmod` chains.

Whether this matters depends on the eolang number model — if the `number` object collapses `-0.0` and `0.0` at the bytes level elsewhere, it's a nit. If it doesn't (and the fact that `positive-infinity` and `negative-infinity` are separately tested at 444/447 suggests signed-zero is at least partly preserved elsewhere), then this is a subtle behavioural change worth pinning either way — a test `(-0.0).floor.as-bytes.eq (-0.0).as-bytes > @` documents which side you picked.

gemshrine (commented):
The floor semantics fix looks right and the added tests cover the common negative-fraction case. Three gaps to close before this is safe against future edits — inline.

mvanhorn:
Thanks — closed all three gaps in a823b18: floor now returns the input unchanged when |x| >= 2^63 (already integral, avoids the i64 overflow), preserves the sign of -0.0, and I added tests for NaN, ±Inf, and a large magnitude.

yegor256:

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+806](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #3648, opened 2024-12-12
only FQN is accepted as atom result, which is wrong


yegor256:
Currently, we can't write this:

```
[] > foo /number
```

We can, but the `number` will be interpreted as `Q.number`. Let's add a resolver of atom result names, similar to how we resolve object names:

```
[] > foo
  number > x
```

Here, `number` means `Q.org.eolang.number`.

yegor256:
@maxonfjvipon please, help

maxonfjvipon:
This is how we can deal with atoms:
1. remove return types from the syntax and leave just `?` for atoms
```
# This is atom.
[] > foo ?
```
2. remove attribute `@atom` from XMIR and XSD scheme. To mark that object is an atom - add inner `<o name="λ"/>`

0pdd:
@yegor256 the puzzle [#3974](https://github.com/objectionary/eo/issues/3974) is still not solved.

0pdd:
@yegor256 the only puzzle [#3974](https://github.com/objectionary/eo/issues/3974) is solved here.

PR #5054, opened 2026-04-24
feat(#4707): restore /name syntax for atom return types

### LINKED ISSUE #4707: Need for Explicit Return Object Definition in Atoms After Removal of `/xxx` Syntax
In #3648 we removed the `/xxx` syntax. Now, let's think about putting it back:

```
[] > number
  [x] > times /number
```

This will give us an additional strictness in compile time.

syncended:
Closes: #4707

## Summary

Returns the `/xxx` syntax for atom return types (removed in #3648/#3971), now with proper FQN resolution.

Before: `[x] > times ?` — opaque, no type info.
After: `[x] > times /number` — explicit return object, resolved via the same `build-fqns.xsl` scope lookup as regular `@base` references.

This gives additional compile-time strictness by documenting what each atom returns.

## Changes

- Grammar: `atom: voids suffix SPACE SLASH NAME testsOrEol` (QUESTION removed, SLASH restored).
- Listener emits `<o name=\"λ\" base=\"NAME\"/>`; existing FQN resolver handles it as any other base reference.
- `print/to-eo.xsl` renders `/<last-segment-of-base>` when printing atoms.
- All atoms in `eo-runtime/src/main/eo/` migrated to explicit return types. Boolean-returning atoms use `/true` (language convention, since EO has no unified bool supertype); others use their actual returned type (`/number`, `/bytes`, `/i64`, `/string`, etc.).
- Test YAML fixtures and Java-embedded EO snippets updated.

## Test plan
- [x] `mvn -pl eo-parser test` — 1007 tests pass.
- [x] `mvn -pl eo-maven-plugin test` — 326 tests pass.
- [x] `mvn -pl eo-runtime clean test -Deo.skipLinting=true` — 1244 tests pass.


eo-parser/src/main/antlr4/org/eolang/parser/Eo.g4
```diff
@@ -68,9 +68,11 @@ just: beginner
     | finisher
     ;
 
-// Atom - abstract object with mandatory name
+// Atom - abstract object with mandatory name and return type
 // Can't contain inner objects
-atom: voids suffix SPACE QUESTION testsOrEol
+// `QUESTION` form is the legacy syntax kept for backward compatibility
+// with already-published .eo files that have not migrated to `/name` yet.
+atom: voids suffix SPACE (SLASH NAME | QUESTION) testsOrEol
```
maxonfjvipon: @syncended first, we don't need `?` syntax anymore; second, not only simple `NAME` but also full FQN should be allowed:
```
[] > atom1 /Q.org.eolang.number
[] > atom2 /org.eolang.number
[] > atom3 /x.y.z
```

maxonfjvipon (changes requested):
@syncended FQNs in atoms also should be resolved and extended using xsl transformations. Check `expand-aliases`, `resolve-aliases` and `add-default-package`. You can either implement it here or leave todo

0crat:
@maxonfjvipon Hey! Nice work on that review 👍 You've earned **+7 points** this time: started with the standard +12 base, but lost 5 points since we only got 1 comment (policy kicks in a 5-point deduction when there's fewer than 12 comments). Your running total is now [+318](https://www.eolang.org/zerocracy/objectionary-vitals.html) - keep those reviews coming and remember you can earn up to 8 bonus points just from adding more comments! Don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) when you get a chance.

eo-parser/src/main/java/org/eolang/parser/XeEoListener.java
```diff
@@ -234,13 +234,25 @@ public void enterAtom(final EoParser.AtomContext ctx) {
 
     @Override
     public void exitAtom(final EoParser.AtomContext ctx) {
+        final EoParser.AtomBaseContext base = ctx.atomBase();
         this.objects.enter()
             .start(ctx.getStart().getLine(), 0)
             .prop("name", "λ")
+            .prop("base", base == null ? "" : XeEoListener.qqToGlobalPhi(base.getText()))
```
maxonfjvipon: @syncended this file is in exluded list for qulice because it's too large, but we don't use ternary operators.
Also `@base` attribute is reserved. Let's use `@atom` attribute here

maxonfjvipon (changes requested):

eo-parser/src/main/antlr4/org/eolang/parser/Eo.g4
syncended: Kept the `?` syntax to avoid disabling the integration test (which points to an external repo) - though enforcing the change might actually be the better call here.

syncended:
PR size check should probably be ignored. The issue drops the ? syntax; reformatting the affected atoms is a side effect.

yegor256 (approved):

0crat:
@yegor256 Hey! Nice work on that review 👍 You snagged **+7 points** this time: started with the standard **+12 base**, but lost **-5** since we only got **3 comments** (policy kicks in a deduction when there's fewer than **12**). Your running total is at [**+81**](https://www.eolang.org/zerocracy/objectionary-vitals.html) - keep those detailed reviews coming to maximize your bonus potential, and don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) for the latest updates!

PR #5057, opened 2026-04-25
fix(#5043): make number.floor round toward negative infinity

### LINKED ISSUE #5043: `number.floor` wrongly truncates towards zero for negative decimal numbers, contrary to documentation

bibonix:
@yegor256 — this fixes #5043. `number.floor` was implemented as `as-i64.as-number`, which truncates toward zero rather than performing mathematical floor (round toward −∞). The discrepancy only matters for negative non-integers, but several mathematical identities silently broke because of it.

## What changed

`eo-runtime/src/main/eo/number.eo` — `floor` becomes:

```eo
[] > floor
  if. > @
    ^.is-finite.not
    ^
    if.
      trunc.gt ^
      trunc.minus 1
      trunc
  ^.as-i64.as-number > trunc
```

Idea: truncation equals mathematical floor for every input *except* negative non-integers, where the truncated value is strictly greater than the original — so we subtract `1` only in that one case. Non-finite inputs (`nan`, `±inf`) are returned unchanged, so `.floor` no longer propagates the *“Can't convert ... to i64”* error from the underlying `as-i64`.

(Note: I used `^` rather than the `$` from the suggested patch in the issue. `$` is self-referential here and causes infinite recursion when dataizing `floor`. `^` resolves to the parent `number`, which is what we want to compare `trunc` against.)

## Behavior table

| input               | before | after |
|---------------------|--------|-------|
| `3.7.floor`         | 3      | 3     |
| `-3.7.floor`        | −3     | **−4** |
| `-0.1.floor`        | 0      | **−1** |
| `(13.div -5).floor` | −2     | **−3** |
| `-5.floor`          | −5     | −5    |
| `0.floor`           | 0      | 0     |
| `nan.floor`         | error  | **nan** |
| `(±inf).floor`      | error  | **±inf** |

## Test changes (commit 1)

Three previously-locked-in assertions that documented the bug are flipped to the correct values, plus seven new reproduction cases:

- `eo-runtime/src/main/eo/number.eo`
  - `tests-floor-negative-fraction`: `-3.7.floor` → `-4` (was `-3`)
  - `tests-div-with-remainder`: `(13.div -5).floor` → `-3` (was `-2`)
  - new: `tests-floor-negative-small-fraction`, `tests-floor-of-negative-integer`, `tests-floor-of-zero`
- `eo-runtime/src/main/eo/positive-infinity.eo` — `tests-positive-infinity-floor-is-equal-to-self` now actually calls `.floor` (it was `positive-infinity.@.eq positive-infinity`, which trivially holds)
- `eo-runtime/src/main/eo/negative-infinity.eo` — same change for negative infinity
- `eo-runtime/src/main/eo/nan.eo` — new `tests-nan-floor-is-nan`

All six of these fail on master and pass after the fix. The two commits are split exactly along that line so the failing reproduction is easy to inspect on its own.

## Collateral fix in `ms.mod`

`ms/mod.eo` had `tests-div-mod-compatibility` asserting `A = (A div B).floor * B + A mod B` for `A = -13, B = 5`. `mod` here follows the sign of the dividend (truncated remainder, like C `%`), so the matching quotient is truncation toward zero, not floor. The test was passing on master only because `floor` happened to truncate. With `floor` corrected, I swapped `.floor` for `.as-i64.as-number` so the identity continues to test the actual semantics of `mod`. The behavior of `mod` itself is unchanged.

## Verification

- `mvn -pl eo-runtime test` locally on Linux/JDK 21: 1248 / 1248 pass.
- All 25 CI checks green on this branch (mvn × 3 platforms, integration, snippets, qulice, ort, pdd, codenarc, etc.).

### DIFF
--- a/eo-runtime/src/main/eo/ms/mod.eo
+++ b/eo-runtime/src/main/eo/ms/mod.eo
@@ -59,16 +59,17 @@
       mod 16 -200
       16
 
-  # Tests mathematical equality A = ((A div B) * B) + (A mod B).
-  # Checks mathematical equality
-  # A = ((A div B) * B) + (A mod B).
+  # Tests mathematical equality A = (trunc(A / B) * B) + (A mod B).
+  # `mod` follows the sign of the dividend (truncated remainder, like C `%`),
+  # so the matching quotient is truncation toward zero (`as-i64`), not
+  # mathematical floor.
   [] +> tests-div-mod-compatibility
     eq. +> @
       plus.
         mod dividend divisor
         times.
           divisor
-          (dividend.div divisor).floor
+          (dividend.div divisor).as-i64.as-number
       dividend
     -13 > dividend
     5 > divisor
--- a/eo-runtime/src/main/eo/nan.eo
+++ b/eo-runtime/src/main/eo/nan.eo
@@ -164,6 +164,12 @@
       nan.as-bytes
       (0.0.div 0.0).as-bytes
 
+  # Tests that floor of NaN returns NaN.
+  [] +> tests-nan-floor-is-nan
+    eq. > @
+      nan.floor.as-bytes
+      nan.as-bytes
+
   # Tests that NaN is detected as NaN.
   nan.is-nan > [] +> nan-is-nan
 
--- a/eo-runtime/src/main/eo/negative-infinity.eo
+++ b/eo-runtime/src/main/eo/negative-infinity.eo
@@ -500,7 +500,7 @@
 
   # Tests that negative infinity floor equals itself.
   [] +> tests-negative-infinity-floor-is-equal-to-self
-    negative-infinity.@.eq negative-infinity > @
+    negative-infinity.floor.eq negative-infinity > @
 
   # Tests that negative infinity is not NaN.
   [] +> tests-negative-infinity-is-not-nan
--- a/eo-runtime/src/main/eo/number.eo
+++ b/eo-runtime/src/main/eo/number.eo
@@ -112,8 +112,20 @@
   # The object rounds down the original `number` to the nearest
   # whole `number` that is less than or equal to it
   # and returns the result as `org.eolang.number`.
+  # For non-finite values (`nan`, `+inf`, `-inf`) the input is returned
+  # unchanged. For finite inputs we truncate toward zero via `as-i64`,
+  # which matches mathematical floor for every value except negative
+  # non-integers — where the truncated result is strictly greater than
+  # the original, so we subtract `1` in that single case.
   [] > floor
-    as-i64.as-number > @
+    if. > @
+      ^.is-finite.not
+      ^
+      if.
+        trunc.gt ^
+        trunc.minus 1
+        trunc
+    ^.as-i64.as-number > trunc
 
   # Tests that less-than comparison returns true when first number is smaller.
   [] +> tests-int-less-true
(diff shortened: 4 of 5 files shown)


0crat:
@bibonix Thanks for the contribution! You've earned **+16 points** for your code submission. This reward follows our team policy which awards 16 points for reviews provided, plus additional points based on your hits-of-code contribution. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+48](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #5072, opened 2026-04-29
fix(#5071): clear error on missing closing slash in regex

### LINKED ISSUE #5071: `regex` object does not validate missing closing `/`, leading to `IndexOutOfBoundsException`
## What Happens

`regex` object throws an opaque `IndexOutOfBoundsException` when the expression is missing a closing `/`, instead of a clear error message.

Reproduction:

```eo
+alias io.stdout
+alias tt.regex

[args] > program
  stdout > @
    if.
      (regex "/pattern").compiled.matches "hello pattern world"
      "matched\n"
      "no match\n"
```

Output:
```
Range [1, 0) out of bounds for length 8
JVM failed with exit code 1
```

The correct form `(regex "/pattern/")` works fine and returns `matched`.

The root cause is in `EOregex$EOφ.java`. The guard at the top only checks for a missing opening slash:

```java
if (!expression.startsWith("/")) {
    throw new ExFailure("Wrong regex syntax: \"/\" is missing");
}
```

It does not handle the case where there is no closing slash. When `expression = "/pattern"`, `lastIndexOf('/')` returns `0`, and then:

```java
final int last = expression.lastIndexOf('/');  // last = 0
// ...
builder.append(expression, 1, last);  // append(s, 1, 0) throws IndexOutOfBoundsException
```

`AtomSafe` catches this and wraps it into an EO error, producing the opaque message above.

## What Should Happen

`regex "/pattern"` should throw a clear `ExFailure` such as `"Wrong regex syntax: closing \"/\" is missing"`, the same way a missing opening slash is handled.

bibonix:
@yegor256 this is ready for merge.

Fixes #5071.

## What was wrong
`EOregex$EOφ.lambda()` only validated the *opening* `/` in a regex expression. When the closing `/` was missing (e.g. `regex "/pattern"`), `lastIndexOf('/')` returned `0`, then `builder.append(expression, 1, 0)` threw an opaque `IndexOutOfBoundsException` (`Range [1, 0) out of bounds for length 8`), which was wrapped into an EO error and surfaced to users as a confusing message.

## Fix
Added a single guard right after the existing opening-slash check: when `last == 0`, throw `ExFailure("Wrong regex syntax: closing \"/\" is missing")`, mirroring the existing missing-opening-slash diagnostic. Three lines of production code.

## Tests
- `eo-runtime/src/test/java/org/eolang/EOtt/RegexAtomTest.java` — new Java test class with two cases:
  - `compilesRegexWithSlashes` — sanity check that a well-formed `/[a-z]+/` still compiles and matches.
  - `throwsClearErrorOnMissingClosingSlash` — fails on master with `"out of bounds"`, passes after the fix and asserts the error message no longer contains `"out of bounds"`.
- `eo-runtime/src/main/eo/tt/regex.eo` — added an EO test `throws-on-missing-closing-slash-in-regex`, parallel to the existing `throws-on-missing-first-slash-in-regex`.

All 47 regex tests in `eo-runtime` pass locally; all 26 CI checks are green.

## Test plan
- [x] New Java tests added (`RegexAtomTest`) — fail on master, pass after fix.
- [x] New EO test added (`throws-on-missing-closing-slash-in-regex`).
- [x] Local `mvn test` and `mvn verify -Pqulice` succeed.
- [x] All 26 CI checks pass.

### DIFF
--- a/eo-runtime/src/main/eo/tt/regex.eo
+++ b/eo-runtime/src/main/eo/tt/regex.eo
@@ -247,6 +247,10 @@
   [] +> throws-on-missing-first-slash-in-regex
     (regex "(.)+").compiled > @
 
+  # Tests that regex compilation throws error when missing closing slash.
+  [] +> throws-on-missing-closing-slash-in-regex
+    (regex "/pattern").compiled > @
+
   # Tests that regex with capturing groups returns correct group count and content.
   [] +> tests-regex-contains-valid-groups-on-each-matched-block
     and. > @
--- a/eo-runtime/src/main/java/org/eolang/EOtt/EOregex$EOφ.java
+++ b/eo-runtime/src/main/java/org/eolang/EOtt/EOregex$EOφ.java
@@ -38,6 +38,9 @@ public final class EOregex$EOφ extends PhDefault implements Atom {
             throw new ExFailure("Wrong regex syntax: \"/\" is missing");
         }
         final int last = expression.lastIndexOf('/');
+        if (last == 0) {
+            throw new ExFailure("Wrong regex syntax: closing \"/\" is missing");
+        }
         final StringBuilder builder = new StringBuilder();
         if (!expression.endsWith("/")) {
             builder.append("(?").append(expression.substring(last + 1)).append(')');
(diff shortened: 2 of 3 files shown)


0crat:
@bibonix Hey! Nice work on your contribution – you've snagged **+16 points** this time! 🎯 The policy rewards solid contributions like yours, and you're doing great by keeping your code tight and getting proper reviews. Your running score is [+98](https://www.eolang.org/zerocracy/objectionary-vitals.html), so [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those quality contributions flowing! Don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) to track your progress.

PR #5074, opened 2026-04-30
fix(#5070): preserve leading sign in sscanf %d format

### LINKED ISSUE #5070: `sscanf` function incorrectly discards minus sign for negative integers with `%d` format
## What Happens

`sscanf` with `%d` format returns a wrong value when the input is a negative integer — the minus sign is silently discarded.

Reproduction:

```eo
+alias io.stdout
+alias tt.sprintf
+alias tt.sscanf

[args] > program
  stdout > @
    sprintf
      "%d | %d | %f\n"
      *
        (sscanf "%d" "42").at 0
        (sscanf "%d" "-42").at 0
        (sscanf "%f" "-3.14").at 0
```

Output:
```
42 | 42 | -3.140000
```

The root cause is in `EOsscanf.java` — `%d` builds regex `(\d+)` with no sign prefix, while `%f` already uses `([+-]?\d+(?:\.\d+)?)`:

```java
case 'd':
    regex.append("(\\d+)");                       // missing [+-]?
    break;
case 'f':
    regex.append("([+-]?\\d+(?:\\.\\d+)?)");      // sign handled correctly
    break;
```

## What Should Happen

`sscanf "%d" "-42"` should return a tuple containing `-42`, consistent with how `%f` handles negative floats and with standard `sscanf` behavior in C, Ruby, and Python.

bibonix:
@yegor256

This PR fixes #5070, where `tt.sscanf` with the `%d` format silently dropped a leading `-` (or `+`) sign and returned the unsigned value:

```
sscanf "%d" "-42"   # returned 42 instead of -42
sscanf "%d" "+42"   # returned 42 by accident, not by design
```

## Root cause

In `EOsscanf.java` the regex generated for `%d` was `(\d+)`, which has no sign prefix. When the input was `"-42"`, `Matcher.find()` matched the first run of digits (`"42"`), and `Long.parseLong("42")` produced `42L`. The minus was discarded silently. The `%f` branch already used `([+-]?\d+(?:\.\d+)?)`, so floats already handled signs correctly; only `%d` was inconsistent.

## Fix

- `eo-runtime/src/main/java/org/eolang/EOtt/EOsscanf.java`: change the `%d` regex to `([+-]?\d+)`, mirroring the sign-prefix that `%f` already accepts. `Long.parseLong` already handles a leading `+`/`-`, so no other code changes are needed. The initial `StringBuilder` capacity was also bumped from 30 to 64 to keep PMD's `InsufficientStringBufferDeclaration` rule happy now that the longest specifier expansion is slightly longer.
- `eo-runtime/src/main/eo/tt/sscanf.eo`: add two object-level tests, `tests-sscanf-with-negative-int` and `tests-sscanf-with-positive-signed-int`, alongside the existing `tests-sscanf-with-int`.
- `eo-runtime/src/test/java/org/eolang/EOtt/SscanfNegativeIntegerTest.java`: a Java JUnit test that pins down the regression at the Phi level — it parses `"-42"`, `"+42"` and `"42"` and asserts the first tuple element. Both this Java test and the auto-generated `EOsscanfTest.tests_sscanf_with_negative_int` fail on `master` and pass on this branch.

## Verification

- `mvn -pl eo-runtime test` — 1256 tests pass locally.
- `mvn -pl eo-runtime -Pqulice verify -DskipTests` — clean.
- All 26 CI checks on this PR are green (mvn ubuntu/macos/windows on Java 23, integration, qulice, benchmark, ort, pr-size, sonar, codacy, snippets, etc.).
- The new tests both fail on `master` (the Java one with `Expected: <-42L> but: was <42L>`, the EO one with `expected: <true> but was: <false>`) and pass on this branch.

Closes #5070.

### DIFF
--- a/eo-runtime/src/main/eo/tt/sscanf.eo
+++ b/eo-runtime/src/main/eo/tt/sscanf.eo
@@ -34,6 +34,20 @@
       head.
         sscanf "%d" "33"
 
+  # Tests that sscanf preserves the minus sign for negative integers.
+  [] +> tests-sscanf-with-negative-int
+    eq. > @
+      -42
+      head.
+        sscanf "%d" "-42"
+
+  # Tests that sscanf accepts an optional leading plus sign for integers.
+  [] +> tests-sscanf-with-positive-signed-int
+    eq. > @
+      42
+      head.
+        sscanf "%d" "+42"
+
   # Tests that sscanf parses floating point number using %f format specifier.
   [] +> tests-sscanf-with-float
     eq. > @
--- a/eo-runtime/src/main/java/org/eolang/EOtt/EOsscanf.java
+++ b/eo-runtime/src/main/java/org/eolang/EOtt/EOsscanf.java
@@ -78,7 +78,7 @@ public final class EOsscanf extends PhDefault implements Atom {
                 if (literal) {
                     switch (sym) {
                         case 'd':
-                            regex.append("(\\d+)");
+                            regex.append("([+-]?\\d+)");
                             break;
                         case 'f':
                             regex.append("([+-]?\\d+(?:\\.\\d+)?)");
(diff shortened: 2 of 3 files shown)


0crat:
@bibonix Thanks for the contribution! You've earned **+16 points** for providing a code review according to our team policy. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+114](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #5076, opened 2026-04-30
fix(MjLint): include defect details in IllegalStateException

### LINKED ISSUE #5075: `MjLint` fails to include violation details in exception message
When `MjLint` finds errors in XMIR files, the exception only contains a summary count with no details about which rules failed or why:

```
Caused by: java.lang.IllegalStateException: In 1 XMIR files, we found 1 critical error and 2 warnings (must stop here)
    at org.eolang.maven.MjLint.lint(MjLint.java:122)
```

The defects are already collected and logged via `MjLint.logOne()`, but since Maven swallows warn/error log output in certain contexts (e.g. surefire tests), the details are invisible. The exception message itself should include each defect's rule name, line, and description so it is always visible in the stacktrace.

bibonix:
@yegor256 — fixes #5075.

## Problem

When `MjLint` finds errors or critical defects, it throws

```
java.lang.IllegalStateException: In 1 XMIR files, we found 1 critical error and 2 warnings (must stop here)
    at org.eolang.maven.MjLint.lint(MjLint.java:122)
```

The defect rule, line, and message are only emitted via `Logger.warn`/`Logger.error`. Maven swallows that output in some contexts (surefire, certain CI runners), so the user is left with the count and no idea which lints failed.

## Fix

`MjLint.lint()` now collects every counted defect into a thread-safe `ConcurrentLinkedQueue<Defect>` (shared between the `Threaded` parallel `lintOne` workers and the `lintAll` WPA path). When the build is failed, the formatted list — `object:line text (rule)`, one per line, the same shape as `logOne()` — is appended to the `IllegalStateException` message, so it is always visible in the stacktrace.

`MjLint.read()` is also updated to restore `line`, `text`, and (when present) `object` from cached WPA defects, so cached runs surface the same details.

A regression test (`includesDefectDetailsInExceptionMessage`) walks the cause chain and asserts the program name (`foo.x.main`) and the failing rule (`(cti)`) are in the root exception message.

## Verification

- New test fails on master, passes here.
- All `MjLintTest` cases pass (16 run, 5 pre-existing `@Disabled`).
- `mvn -Pqulice install -DskipTests` clean on `eo-maven-plugin`.
- All 25 CI checks green.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjLint.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjLint.java
@@ -18,6 +18,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.stream.Collectors;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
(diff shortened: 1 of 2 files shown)


0crat:
@bibonix Great work on your contribution! 🎉 You've earned **+16 points** and your total score has reached **[+130](https://www.eolang.org/zerocracy/objectionary-vitals.html)**. [Keep up the momentum](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #5095, opened 2026-05-05
#4318: improve error message for atoms containing inner objects

### LINKED ISSUE #4318: not-empty-atoms.yaml:6-10: Improve error message for...
The puzzle `4264-940f5454` from #4264 has to be resolved: 

https://github.com/objectionary/eo/blob/98fc95c5976bea780fada8c5ee08f2904c12f474/eo-parser/src/test/resources/org/eolang/parser/eo-typos/not-empty-atoms.yaml#L6-L10

The puzzle was created by @yegor256 on 24-Jun-25. 

morphqdd:
Closes #4318.

Replace the cryptic ANTLR `mismatched input 'UNTAB' expecting {...}` error with a clear message when an atom (object with `/return-type`) is followed by inner objects.

## Detection

When `InputMismatchException` fires inside `EoParserErrors.mismatch()`, check the rule invocation stack. If both `atom` and `tests` rules are present, the parser entered the test-body of an atom and stumbled on a non-test inner object. In that case, emit:

```
Atom cannot contain inner objects, only test objects are allowed in its body
```

The check requires `tests` in the stack so it does not affect other in-atom errors (e.g. `broken-head.yaml` has stack `[atom, masterBody, object, program]` — no `tests`, falls through to the default branch).

## Fixture

`not-empty-atoms.yaml` updated: dropped the `@todo #4264` block, added `message:` key so `EoSyntaxTest.checksTypoPacks` verifies the new text via `containsString`.

## Verification

- `mvn test` (eo-parser): 1008/1008 passing
- `mvn -Pqulice qulice:check`: clean

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/EoParserErrors.java
+++ b/eo-parser/src/main/java/org/eolang/parser/EoParserErrors.java
@@ -127,7 +127,8 @@ final class EoParserErrors extends BaseErrorListener implements Iterable<Parsing
      * @return The detailed message
      */
     private static String mismatch(final Parser parser, final String msg) {
-        final String rule = parser.getRuleInvocationStack().get(0);
+        final List<String> stack = parser.getRuleInvocationStack();
+        final String rule = stack.get(0);
         final String[] names = parser.getRuleNames();
         final String detailed;
         if (names[EoParser.RULE_program].equals(rule)) {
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Hey! Nice work on that contribution! 🎉 You've scored **+12 points** this time: started with the standard **+16 points** base, but had **-4 deducted** since you only hit **13** [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (we need at least **40** to avoid that penalty). Next time, try to bundle a bit more code to maximize your points - you can earn up to **0.05 points** per hit-of-code! Your running score is looking solid at **[+212](https://www.eolang.org/zerocracy/objectionary-vitals.html)**, so [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those contributions flowing and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too!

PR #5135, opened 2026-05-14
Add Step interface and Timed decorator for pipeline timing

### LINKED ISSUE #5134: Extract duplicated timing logic into a `Timed` step decorator
Currently, timing logic (\`System.currentTimeMillis()\` before/after + \`Logger.info\` for elapsed time) is duplicated across multiple pipeline steps: \`Assembling\`, \`Compiling\`, \`Linting\`, \`Parsing\`, \`Probing\`, \`Pulling\`, and \`Transpiling\`, making it harder to maintain and extend.

It would be nice to extract this cross-cutting concern into a dedicated decorator, similar to how \`StMeasured\` wraps \`Shift\` for xsline transforms:

1. \`Step\` — a shared interface for all \`*ing\` pipeline classes, with a single \`void exec() throws IOException\` method.
2. \`Timed\` — a decorator implementing \`Step\` that wraps any \`Step\`, measures elapsed time, and logs it, eliminating the timing boilerplate from each class.

This would also allow \`Unspiling\` and \`Unplacing\` (which currently have no timing) to benefit from it automatically.

stegruslan:
Closes #5134

Added `Step` interface and `Timed` decorator to eliminate 
duplicated timing logic across pipeline steps.


0crat:
@stegruslan According to our [work policy](https://www.yegor256.com/2014/04/15/github-guidelines.html), branches with names that don't follow our convention result in a **-6 point penalty** for the author. Your branch name "`timed-step-decorator`" violates this policy, so you've earned -6 points. Branch names should match the ticket number you're solving - in this case, "`5134`" would be the correct name. Your running score is [-6](https://www.eolang.org/zerocracy/objectionary-vitals.html); please check your [Zerocracy account](https://www.zerocracy.com/) for updates.

eo-maven-plugin/src/main/java/org/eolang/maven/Timed.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    Timed(final Step step, final String lbl) {
+        this.origin = step;
+        this.label = lbl;
+    }
+
+    @Override
+    @SuppressWarnings("PMD.UnnecessaryLocalRule")
+    public void exec() throws IOException {
+        final long start = System.currentTimeMillis();
```
gemshrine, line 39: `System.currentTimeMillis()` reads wall-clock time, which NTP can step (backwards during a leap-second smear, forwards after a suspend/resume, arbitrarily on VM migration). Any of those turns `System.currentTimeMillis() - start` into a negative number or a wildly inflated one, and jcabi's `%[ms]s` formatter will print that as-is. For elapsed measurement, `System.nanoTime()` is the specified idiom — monotonic within a JVM and immune to clock adjustments. Convert to ms only at the log site:

```java
final long start = System.nanoTime();
this.origin.exec();
Logger.info(
    this,
    "%s took %[ms]s",
    this.label,
    (System.nanoTime() - start) / 1_000_000L
);
```

eo-maven-plugin/src/main/java/org/eolang/maven/Timed.java
```diff
@@ -0,0 +1,48 @@
...
+        this.origin.exec();
```
gemshrine, line 40: If `origin.exec()` throws — which its signature explicitly declares — control jumps straight to the caller and the `Logger.info(...)` block never runs. Timing is emitted for successful steps and silently dropped for failing ones, which is exactly the opposite of what you want when hunting a slow failure.

A `try/finally` keeps the invariant "`Timed` always logs":

```java
final long start = System.nanoTime();
try {
    this.origin.exec();
} finally {
    Logger.info(
        this,
        "%s took %[ms]s",
        this.label,
        (System.nanoTime() - start) / 1_000_000L
    );
}
```

eo-maven-plugin/src/test/java/org/eolang/maven/TimedTest.java
```diff
@@ -0,0 +1,44 @@
...
+            IOException.class,
+            () -> new Timed(
+                () -> {
+                    throw new IOException("simulated failure");
+                },
+                "failing step"
+            ).exec(),
+            "Timed must propagate IOException from the wrapped step"
+        );
+    }
```
gemshrine, line 43: Two coverage gaps:

- Neither test asserts that a log line was actually emitted. A jcabi-log or slf4j `ListAppender` on `Timed.class`'s logger would let `executesWrappedStep` verify the captured line contains the label and a `took` measurement — otherwise a regression that drops the `Logger.info` call passes silently.
- No test for timing-on-exception (see comment on `Timed.java:40`). A companion to `propagatesException` should catch the `IOException` and then assert the appender captured the timing entry too — without it, adding a `try/finally` around the origin call is an unprotected behaviour change.

gemshrine (commented):
Three concerns on `Timed` — wall-clock instead of monotonic clock for elapsed measurement, timing log lost on the exception path, and untested log emission. Details inline.

yegor256:
@stegruslan merge conflicts here

yegor256:

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+799](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5189, opened 2026-06-09
Atom's return types not verified during runtime


maxonfjvipon:
This is an atom:
```eo
[] > foo /Q.number
```

Here we say that it returns object with `Q.number` FQN. Let's check in runtime that each atom actually returns the object of the type which is provided in the atom

0pdd:
@maxonfjvipon 2 puzzles [#5200](https://github.com/objectionary/eo/issues/5200), [#5201](https://github.com/objectionary/eo/issues/5201) are still not solved.

0crat:
@maxonfjvipon Hey there! 🎉 Nice work closing that issue! You've earned **+7 points** for this one: **+8** as your base reward, then **-1** for the 2 days of delay (our policy deducts 0.5 points per day). Your running score is now [+164](https://www.eolang.org/zerocracy/objectionary-vitals.html) - keep up the momentum! Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

0pdd:
@maxonfjvipon the puzzle [#5200](https://github.com/objectionary/eo/issues/5200) is still not solved; solved: [#5201](https://github.com/objectionary/eo/issues/5201).

0pdd:
@maxonfjvipon all 2 puzzles are solved here: [#5200](https://github.com/objectionary/eo/issues/5200), [#5201](https://github.com/objectionary/eo/issues/5201).

Issue #4538, opened 2025-09-08
`QQ` Prefix Causes Redundancy in Package Naming


yegor256:
Maybe we can get rid of the `QQ` prefix and move all objects to the root package. Thus, `Q.org.eolang.number` will become `Q.number` and `QQ.io.stdout` will become `Q.io.stdout`.

tihiyn:
@yegor256, should the meta `package` remain for objects that were previously in the `org.eolang` package? Or should it be completely removed?

yegor256:
@tihiyn should be removed

tihiyn:
@yegor256, I am faced with the following problem. If we remove the prefix `org.eolang` from all meta `+package`, then the Java classes generated from EO objects will be located in `target/generated-sources`. Because of this, the generated classes do not have access to classes from `eo-runtime/src/main/java/EOorg/EOeolang`. If we move classes from `eo-runtime/src/main/java/EOorg/EOeolang` to `eo-runtime/src/main/java` (which will already be a bad practice, since some classes will be in the default-package), then to some classes that are now in the default-package, (for example, `EOerror`), now classes from `eo-runtime/src/main/java/org/eolang` (for example, `PhSafe`) cannot access. Of course, we can move classes from `eo-runtime/src/main/java/org/eolang` in `eo-runtime/src/main/java`, but then classes from packages (for example, `EOsm/Win32/AcceptFuncCall`) will not be able to access classes from `eo-runtime/src/main/java`. It turns out to be a vicious circle. Can you please explain what did you see the target solution?

yegor256:
@tihiyn we should keep classes in the `eo-runtime/src/main/java/EOorg/EOeolang` directory and make them `public` where necessary, to make them accessible from Java classes generated from EO.

0pdd:
@yegor256 5 puzzles [#4945](https://github.com/objectionary/eo/issues/4945), [#4946](https://github.com/objectionary/eo/issues/4946), [#4949](https://github.com/objectionary/eo/issues/4949), [#4954](https://github.com/objectionary/eo/issues/4954), [#4955](https://github.com/objectionary/eo/issues/4955) are still not solved; solved: [#4944](https://github.com/objectionary/eo/issues/4944).

0pdd:
[...]

PR #5191, opened 2026-06-09
bug(#5189): implement true and false via bool with abstraction-based forma

maxonfjvipon:
## What

Introduces a Church-encoded `bool` object and reimplements `true`/`false` on top of it, so that both report `Q.bool` as their `forma`. This is the groundwork for the runtime atom return-type verification requested in #5189: comparison/predicate atoms can now declare `/bool` and a single `forma`-based check works for both `true` and `false`.

It also completes the `forma` cleanup from #4538.

## Changes

- **`bool.eo`** (new): `[if] > bool`. The single void parameter `if` is the *choice selector* (a two-argument object returning its first or second argument). The byte representation (`if 01- 00-`) and `not`/`and`/`or` all derive from it, so `if` stays pure EO with no conditional primitive and no atom.
- **`true.eo` / `false.eo`**: now applications of `bool` binding the first/second projection. They dataize to `01-`/`00-` exactly as before; their tests moved into `bool.eo`.
- **Abstraction-based `forma`**: `PhSafe#forma()` delegates to the wrapped origin, so a named application reports the `forma` of its abstraction rather than its own name. As a result `e`/`pi` → `Q.number`, `line-separator` → `Q.string`, and `true`/`false` → `Q.bool`.
- **`PhDefault#forma()`** (#4538): drops the single `org.eolang` runtime-root prefix and keeps deeper sub-packages with the `EO` marker stripped (e.g. `Q.number`, `Q.ms.sqrt`, `Q.org.eolang.x`).
- **`Dataized`**: the `go.to.token.jump` control-flow token is detected via its locator (which still carries the source identity) instead of `forma`.
- Forma-semantics tests updated; the two previously disabled `DataizedTest` cases are re-enabled.

## Testing

Full `eo-runtime` suite green: **1308 tests, 0 failures, 0 errors**. EO-lint and qulice clean (no new violations).

## Follow-up

The actual runtime check from #5189 (retarget `/true` → `/bool` on predicate atoms, carry the declared type into `@XmirObject`, and verify `result.forma()` after `lambda()` in `PhDefault`) will come in a separate PR built on this foundation.


maxonfjvipon:
@yegor256 please check

0crat:
@maxonfjvipon Thanks for the contribution! You've earned **+4 points** for this submission: +16 as a basis, +0.05×606 capped at +16 for [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html), -4 for exceeding 100 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (606 >= 100), -8 for exceeding 400 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (606 >= 400), then capped at minimum +4 points. Please consider breaking down larger contributions into smaller chunks to optimize your score and [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+165](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #5210, opened 2026-06-14
fix(#5209): warm up lint stylesheets before parallel linting

### LINKED ISSUE #5209: Intermittent build failure during `eo-maven-plugin:lint` phase due to concurrent XSL compilation race
The `mvn` build fails intermittently during the `eo-maven-plugin:lint` phase while compiling `eo-runtime`. It is flaky — the same commit passes on one run and fails on the next.

Two failure signatures have been observed, both in the same lint step:

1. `#fatalError(): Cannot find a 1-argument function named Q{https://www.eolang.org}lineno(); SystemID: .../phi-is-not-first` (e.g. run https://github.com/objectionary/eo/actions/runs/27493662522/job/81263611234)

2. `org.w3c.dom.DOMException: HIERARCHY_REQUEST_ERR` while transforming via Saxon (e.g. master run for #5187, https://github.com/objectionary/eo/actions/runs/27456575442)

Both manifest as `Failed to process "<object>" (org.eolang.maven.TjForeign)` for many/all objects at once.

## Root cause

`Linting` lints each program in a thread pool (`Threaded`, `availableProcessors() * 2`). All sources share lint stylesheets via `org.eolang.lints.Source.MONO`, a static `Synced(Sticky(...))`. `Synced` only serialises *iteration* over the lints; the actual XSL **compilation** is lazy and lives in each `LtByXsl.sheet`, an `Unchecked(Sticky(...))` scalar that is **not** thread-safe. On the first parallel pass several threads call `Lint.defects()` on the same lint and compile its stylesheet (resolving `<xsl:import href="/org/eolang/funcs/lineno.xsl"/>` and friends) concurrently, racing inside Saxon / jcabi-xml and producing the errors above. Once the stylesheets are compiled and cached, transforms are safe (each call builds a fresh `Transformer` from the thread-safe `Templates`).

## Reproduction

End-to-end, with the lint cache disabled, it reproduces within a handful of runs:

```
for i in $(seq 1 10); do
  mvn -q -pl eo-runtime process-sources \
    -Deo.cacheEnabled=false -Deo.failOnWarning=false || echo "RUN $i FAILED"
done
```

It failed on run 2 of 10 in our testing.

Note that the lint stylesheets are static singletons that compile **once per JVM**, so the race has a single opportunity per process — an in-process test almost always "passes" on the unfixed code. The reproduction test in #5213 therefore forks many fresh JVMs in parallel (the CPU oversubscription widens the cold-compilation window) and fails if any of them races.

## PRs

- #5213 — adds the forking reproduction test (`LintingTest` + `ColdLint`). Merge **first**; the test is kept `@Disabled` because it forks many JVMs and exceeds the CI per-test timeout — it is a manual reproduction tool (run on demand), not a CI gate.
- #5210 — the fix: compile every lint stylesheet once, single-threaded, before the parallel pass (`warm`). The reproduction test ships with it but stays `@Disabled` for the same CI-cost reason.

asmirnov-backend:
Closes #5209.

The `mvn` build fails intermittently in the `eo-maven-plugin:lint` phase with `Cannot find a 1-argument function named ...lineno()` or DOM `HIERARCHY_REQUEST_ERR`. See #5209 for the full analysis.

**Cause:** `Linting` lints each program in a thread pool. All sources share lint stylesheets (`Source.MONO`, a static `Synced(Sticky(...))`) compiled lazily on first use. `Synced` serialises only iteration, so the first parallel pass triggers concurrent first-time XSL compilation, which races inside Saxon / jcabi-xml.

**Fix:** compile every lint stylesheet once, single-threaded, before the parallel pass (`warm(programs)`). Transforms were never the problem — each builds a fresh `Transformer` from the cached, thread-safe `Templates`.

**Verification:** with the lint cache disabled, the failure reproduced on run 2 of 10 before the fix; after the fix it passed 20 of 20:

```
for i in $(seq 1 20); do
  mvn -q -pl eo-runtime process-sources -Deo.cacheEnabled=false -Deo.failOnWarning=false
done
# TOTAL FAILURES: 0 / 20
```

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
@@ -214,6 +214,7 @@ final class Linting implements Step {
         if (!this.skipSourceLints.isEmpty()) {
             Logger.info(this, "Unlinting source lints: %[list]s", this.skipSourceLints);
         }
+        this.warm(programs);
         final int passed = new Threaded<>(
             programs,
             tojo -> this.lintOne(tojo, counts, seen, this.skipSourceLints.toArray(new String[0]))
@@ -260,6 +261,24 @@ final class Linting implements Step {
                     programs.size(), sum, details
                 )
             );
+        }
+    }
+
+    /**
+     * Compile every lint stylesheet once, in a single thread, before linting in parallel.
+     * <p>All sources share lint stylesheets that are compiled lazily on first use and cached
+     * statically. Triggering that compilation from many threads at once races inside Saxon and
+     * jcabi-xml, surfacing intermittently as "cannot find function lineno()" or DOM
+     * HIERARCHY_REQUEST_ERR failures. Running one source synchronously here forces the
+     * compilation up front, so the parallel pass only reuses the cached stylesheets.</p>
+     * @param programs The XMIR programs about to be linted
+     * @throws IOException If fails to read the XMIR
+     */
+    private void warm(final Collection<TjForeign> programs) throws IOException {
+        if (!programs.isEmpty()) {
+            new Source(new XMLDocument(new ListOf<>(programs).get(0).xmir()))
+                .without(this.skipSourceLints.toArray(new String[0]))
+                .defects();
         }
     }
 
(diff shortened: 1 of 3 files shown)


yegor256:
@volodya-lombrozo WDYT?

yegor256:
@volodya-lombrozo ping

volodya-lombrozo:
@asmirnov-backend Thank you for the contribution. This issue was solved in lints `0.2.8`

Issue #4752, opened 2025-12-14
`sprintf` Functionality Not Properly Utilized in EO


yegor256:
Currently, it's an atom, but we can implement its functionality in pure EO. Let's do it.

MrEx3cut0r:
Hi, I`ll be working on this

asacxyz:
Is this still active? If not, I'd like to take over.

yegor256:
@asacxyz sure, go ahead

eshabakhov:
@yegor256,

While implementing `sprintf` in pure EO, I started with basic format specifiers like "%d", "%s", "%f", "%x", "%b".

While working on "%d", I ran into a problem: there is currently no way to convert numbers to strings in EO.

This raises a question:

Should `number` provide an `as-string` object,
or is number-to-string conversion expected to be implemented separately (e.g., inside `sprintf`)?

yegor256:
@eshabakhov it should be implemented in sprintf. The `number` itself only exposes 8 bytes. How to present them in a text form -- depends on the printer (`sprintf`)

eshabakhov:
@yegor256, in that case, should the functionality be implemented in a single MR, or should create a temporary object like `printf` and implement it incrementally?

0crat:
@mrex3cut0r Thanks for closing this issue! You've earned +4 points for this: +8 as a basis; -20 for 203 days of delay; +16 to give you at least something. Your running score is [+4](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5219, opened 2026-06-14
feat(#4752): add tt.as-ascii object

asmirnov-backend:
Part of #4752.

Adds a pure-EO `tt.as-ascii` object that reads a single character as its
ASCII numeric value (0-255), by widening the character's byte to the eight
bytes `i64` expects and reading it back as a number.

It is a small helper needed by the pure-EO `sprintf` reimplementation.

PR #5227, opened 2026-06-15
feat(#5226): wire up-cased/low-cased to tt.as-ascii

### LINKED ISSUE #5226: Redundant usage of private ascii helper in up-cased and low-cased instead of tt.as-ascii
The `tt.as-ascii` object was added in commit `5e0d03158` (part of #4752) as a standalone object that reads a single character as its ASCII numeric value (0-255). It currently has tests but **no production callers**.

Meanwhile:

- `tt.up-cased` still carries a private nested `[char] > ascii` object that does the same thing.
- `tt.low-cased` borrows that helper through the awkward `(up-cased text).ascii` expression (3x: `code`, `maxcode`, `mincode`), instantiating an entire `up-cased` object just to reach one inner helper.

### Task

- Replace the private `ascii` helper in `up-cased.eo` with `tt.as-ascii`.
- Replace the `(up-cased text).ascii ...` calls in `low-cased.eo` with `tt.as-ascii` (removing the dependency of `low-cased` on `up-cased` internals).
- Drop the now-dead nested `ascii` object and the `+unlint redundant-object` pragma if no longer needed.

Files: `eo-runtime/src/main/eo/tt/up-cased.eo`, `eo-runtime/src/main/eo/tt/low-cased.eo`.

asmirnov-backend:
Closes #5226.

`tt.as-ascii` was added in #4752 as a standalone object that reads a single character as its ASCII numeric value (0-255), but it had no production callers.

This PR wires it in:

- **`tt.up-cased`** — removes the private nested `[char] > ascii` helper and calls `tt.as-ascii` instead.
- **`tt.low-cased`** — replaces the `(up-cased text).ascii ...` expressions with direct `tt.as-ascii` calls, computes `distance` locally, and drops the `+alias tt.up-cased`, fully removing the dependency on `up-cased` internals.

### DIFF
--- a/eo-runtime/src/main/eo/tt/low-cased.eo
+++ b/eo-runtime/src/main/eo/tt/low-cased.eo
@@ -2,7 +2,6 @@
 +alias ss.list
 +alias ss.reduced
 +alias tt.as-ascii
-+alias tt.up-cased
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
 +package tt
@@ -28,13 +27,16 @@
             slice.
               as-bytes.
                 as-i64.
-                  code.plus (up-cased text).distance
+                  code.plus distance
               7
               1
             byte
         as-ascii byte > code
   as-ascii "Z" > maxcode
   as-ascii "A" > mincode
+  minus. > distance
+    as-ascii "a"
+    as-ascii "A"
 
   # Tests that low-cased converts uppercase text to lowercase.
   [] +> tests-converts-text-to-lower-case
--- a/eo-runtime/src/main/eo/tt/low-cased.eo
+++ b/eo-runtime/src/main/eo/tt/low-cased.eo
@@ -1,14 +1,13 @@
++alias tt.as-ascii
 +alias ss.bytes-as-array
 +alias ss.list
 +alias ss.reduced
-+alias tt.as-ascii
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
 +package tt
 +version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
-+unlint redundant-object
 
 # Returns the `text` in lower case.
 [text] > low-cased
@@ -32,9 +31,9 @@
               1
             byte
         as-ascii byte > code
-  as-ascii "Z" > maxcode
-  as-ascii "A" > mincode
-  minus. > distance
+  as-ascii "Z" > maxcode!
+  as-ascii "A" > mincode!
+  minus. > distance!
     as-ascii "a"
     as-ascii "A"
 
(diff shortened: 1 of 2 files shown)


asmirnov-backend:

maxonfjvipon (changes requested):
Thanks for wiring in `tt.as-ascii`. The `up-cased` half is solid, but the `low-cased` decoupling is incomplete and the description overstates what landed.

### Looks good

- **`up-cased`**: the nested `[char] > ascii` is gone and every use (`code`, `maxcode`, `mincode`, and the `as-ascii "A"` inside `distance`) now goes through `tt.as-ascii`. The swap is behavior-preserving — the old helper padded to `as-i64` (seven `00-`) and `as-ascii` pads to `as-i16` (one `00-`), but both yield the same `0–255` value for a single byte.
- Keeping `+unlint redundant-object` is the right call. It guards the shared `reduced` / `[accum byte] >>` structure, not the removed helper — `low-cased` carries the same pragma and never had an `ascii` object, so dropping it would break the lint.

### Needs a change

`low-cased` still depends on `up-cased` internals. Line 28 keeps:

```
code.plus (up-cased text).distance
```

and `+alias tt.up-cased` is still present, so a whole `up-cased` is instantiated just to read its inner `distance`. The dependency dropped from four reaches to one, not to zero — yet the description says this PR is "dropping the implicit dependency on up-cased internals", which isn't the case.

To actually close it, compute the distance locally and drop the alias:

```
minus. > distance
  as-ascii "a"
  as-ascii "A"
```

then `code.plus distance`, and remove `+alias tt.up-cased`.

### Nit

`as-ascii` was introduced in #4752 (commit `5e0d031`), not #5219.

asmirnov-backend:
Thanks for the thorough review, @maxonfjvipon.

You're right that low-cased was still reaching into up-cased. Fixed in 6edb8dcb8:

- distance is now computed locally
- with code.plus distance, and +alias tt.up-cased is removed — so the dependency on up-cased internals is gone entirely now, not just reduced.
- Updated the PR description accordingly so it no longer overstates what landed.
- Also corrected the nit: as-ascii was introduced in #4752 (5e0d031), not #5219.

yegor256:
@maxonfjvipon take a look again please

maxonfjvipon (approved):

0crat:
@maxonfjvipon Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (30) hits-of-code; +4 to give you at least something.  Your running score is [+226](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@asmirnov-backend Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (30) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+120](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5233, opened 2026-06-17
fix(#5232): insert floated vars before tests in vars-float-up

### LINKED ISSUE #5232: vars-float-up transformation appends floated anonymous objects after tests, violating source test-ordering convention
The `vars-float-up` XMIR transformation appends floated anonymous `>>` objects at the tail of the parent's child list, even when the EO source places them before the tests block. This moves them after `+tests-*` siblings in the resulting XMIR, which violates the convention that tests must be the last children of an object.

Example (`ss/filtered.eo` in `eo-runtime`):

```
[lst func] > filtered
  filteredi > @
    lst
    func item > [item index] >>   -> anonymous object, floated up by vars-float-up

  [] +> tests-simple-filtered
    ...
  [] +> tests-filtered-with-bools
    ...
```

After `vars-float-up`, the XMIR becomes:

```xml
<o name="filtered">
  <o name="φ">...</o>
  <o name="+tests-simple-filtered">...</o>
  <o name="+tests-filtered-with-bools">...</o>
  <o name="a🌵18-4">...</o>   -> anonymous object appended AFTER the tests
</o>
```

In the original source, the anonymous object is correctly placed before the tests. `vars-float-up` re-orders it to the end of the children list regardless of where the tests are, which is what produces the malformed ordering.

This currently affects 15 objects in `eo-runtime`: `io.input-as-bytes`, `ss.mappedi`, `ss.mapped`, `ss.eachi`, `ss.withouti`, `ss.filtered`, `ss.reduced`, `ss.range-of-numbers`, `tt.contains-any`, `tt.contains-all`, `tt.up-cased`, `tt.low-cased`, `ms.integral`, `fs.tmpdir`, `fs.file`.

It was first reported as a `lints` false positive in the `wrong-test-order` rule (objectionary/lints#954), since that lint checks XMIR sibling order and has no way to know the object was moved by a later transformation. We've concluded the lint is working correctly against the XMIR it's given - the transformation itself needs to insert floated objects before any `+tests-*` siblings instead of always appending at the tail.

volodya-lombrozo:
Fixes `vars-float-up` to insert floated anonymous objects before `+tests-*` siblings rather than appending them at the tail of the parent's child list.

Fixes #5232

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/vars-float-up.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/vars-float-up.xsl
@@ -43,12 +43,14 @@
       <xsl:if test="not(@name) and @as">
         <xsl:attribute name="as" select="@as"/>
       </xsl:if>
-      <xsl:apply-templates select="node()|@* except @as"/>
+      <xsl:apply-templates select="@* except @as"/>
+      <xsl:apply-templates select="node()[not(self::o and eo:test-attr(.))]"/>
       <xsl:for-each select="o/descendant::o[@name]">
         <xsl:if test="ancestor::o[eo:abstract(.)][1]/generate-id() = generate-id($o)">
           <xsl:apply-templates select="." mode="full"/>
         </xsl:if>
       </xsl:for-each>
+      <xsl:apply-templates select="node()[self::o and eo:test-attr(.)]"/>
     </xsl:copy>
   </xsl:template>
   <xsl:template match="o[@name and ancestor::o[1][not(eo:abstract(.))]]">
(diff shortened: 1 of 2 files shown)


volodya-lombrozo:
@yegor256 could you have a look, please?

0crat:
@volodya-lombrozo Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+505](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5165, opened 2026-05-25
pure objects


yegor256:
Let's add `@Impure` annotation to some atoms, such as `EOposix$EOφ` and `EOfile$EOsize`.

Then, let's create new decorator `PhSticky` that should decorate all objects created by `new` in the code we generate by `to-java.xsl`.

Let's modify the signature of the `Phi.take(String)` method. It should return a tuple:

```
<Phi, Boolean> take(String);
```

Before calling `take()` of the decoratee, `PhSticky` should call `toφ()` (new method) on it and get φ-expression of the object, for example:

```
new PhSticky(
  new PhWith(
    new PhMethod(new Random(), "plus"),
    42
  )
).take("times")
```

This should lead to calculating this expression:

```
(random.plus 42).times
```

Then, it should try to find the expression in a static hash map. If it's found, `PhSticky` should return the object the expression maps to. If not, it should call `take(String)`.

Then, when `take(String)` returns the result, `PhSticky`, should save it to the hash map, if the second part of the tuple (boolean) is true.

yegor256:
@maxonfjvipon WDYT?

asmirnov-backend:
The `Term`/`φTerm()` foundation is already on `master`, so the remaining work is the cache itself. Proposed breakdown into small, independently-mergeable PRs:

**Approach:** follow the issue literally — `Phi.take(String)` returns a `Taken` pair `<Phi, Boolean>`, where the boolean means "pure / cacheable". (`toφ()` from the issue is the existing `φTerm()`.)

| # | Subtask | Depends on | Status |
|---|---------|-----------|--------|
| 1 | **`Taken`** tuple type — `{Phi, boolean}` with `phi()` + `pure()` | — | ✅ #5246 |
| 2 | **`@Impure`** marker annotation | — | ✅ #5245 |
| 3 | Annotate the **20 impure atoms** (`EOfs/*`, `EOsm`: os.name/posix/win32, `EOmalloc*`) | 2 | ✅ #5245 |
| 4 | Change **`Phi.take` → `Taken`**: `PhDefault` computes `pure = !@Impure && <children pure>`, decorators propagate; migrate ~94 consumers + `to-java.xsl` to `.take(x).phi()` | 1, 3 | ⏳ |
| 5 | **`PhSticky`** — static map keyed on `φTerm()`; caches the result only when `pure()` | 4 | ⏳ |
[...]

PR #5240, opened 2026-06-17
feat(#4533): remove last $ usage from regex matched block

### LINKED ISSUE #4533: The dollar sign is still in the grammar, by may/should be removed

asmirnov-backend:
Closes #4533 (runtime part).

`regex.eo`'s `$ > matched` was the only remaining `$` (self/`ξ`) usage in the whole runtime library. Removed it by routing the `next` recursion through `^` (the block's ρ, which is `match`) and dropping the now-redundant `.matched` accessor — no Java atom or parser changes.

The idempotency sugar the original `@todo` waited on was dropped in the parser rewrite (`769675b1a`), so this approach uses no sugar at all.

Parser-level grammar removal of `$` is left as a follow-up.

### DIFF
--- a/eo-runtime/src/main/eo/tt/regex.eo
+++ b/eo-runtime/src/main/eo/tt/regex.eo
@@ -55,7 +55,7 @@
 
     # Creates a `matcher` that will match the given input against the pattern.
     [txt] > match
-      (matched-from-index 1 0).matched > next
+      matched-from-index 1 0 > next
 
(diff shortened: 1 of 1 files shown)


asmirnov-backend:

maxonfjvipon:
@asmirnov-backend let's wait with this PR until #5165 is resolved

yegor256:

PR #5245, opened 2026-06-19
feat(#5165): add `@Impure` and annotate impure atoms

asmirnov-backend:
Part of #5165.

1. Adds the `@Impure` marker annotation — it flags atoms whose result depends on the outside world or mutable state.
2. Annotates the 20 such atoms so the upcoming `PhSticky` caching decorator never memoizes them:
   - filesystem — all `EOfs/*` (`file.*`, `dir.*`);
   - OS — `EOsm`: `os.name`, `posix.φ`, `win32.φ`;
   - shared memory — `EOmalloc*` (`read`/`write`/`resized`/`size`/`of.φ`, all via `Heaps.INSTANCE`).

Pure atoms (arithmetic, math, regex/sprintf/sscanf, error/try) are left un-annotated so they remain cacheable.


asmirnov-backend:

yegor256:
@asmirnov-backend check this: https://www.yegor256.com/2015/02/12/top-down-design.html

yegor256:

PR #5262, opened 2026-06-24
fix: add path and length delimiter to Sha directory hashing to prevent collisions

### LINKED ISSUE #5254: `Sha` class causes hash collisions for different directory trees due to lack of delimiter and absence of file paths in hashing

### Summary

`org.eolang.maven.Sha` is documented to compute a SHA-256 hash of a directory by walking "all files sorted by path and hashes their combined contents". When hashing a directory, it feeds the raw bytes of every file into a single `MessageDigest` back-to-back with **no per-file length delimiter and without hashing the file paths**. As a result, two clearly different directory trees can produce identical hashes — a genuine collision that breaks the contract the class advertises.

### Location

https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java#L54-L76

(permalink pinned at HEAD `382ead173c708672b4e31dac024308402695d1df`)

```java
private static String hash(final Path path) throws IOException, NoSuchAlgorithmException {
    try (Stream<Path> walk = Files.walk(path)) {
        walk.filter(Files::isRegularFile)
            .sorted(Comparator.comparing(Path::toString)).forEach(
                file -> {
                    try (InputStream input = Files.newInputStream(file)) {
                        final byte[] buffer = new byte[8192];
                        int read = input.read(buffer);
                        while (read != -1) {
                            digest.update(buffer, 0, read);   // <-- only raw content, no boundary, no path
                            read = input.read(buffer);
                        }
                            String.format("Failed to read '%s'", file), ex
                        );
                    }
                }
            );
    }
}
```

The class-level Javadoc explicitly promises directory support:

```java
/**
 * SHA-256 hash of a file or directory.
 * For a directory, walks all files sorted by path and hashes their combined contents.
 * @since 0.62.0
 */
```

### Why it's wrong

Because the files are sorted by path and then only their **contents** are streamed into one digest, the hash depends solely on the concatenation of all bytes — the boundaries between files and the file names themselves are invisible to the digest. Any two directory trees whose sorted-by-path content concatenation is byte-identical hash to the same value, even though the trees are different.

Concrete, reproduced collision using the exact logic of `Sha.hash`:

| Directory | Files (sorted by path) | Concatenated bytes | SHA-256 (Base64) |
|-----------|------------------------|--------------------|------------------|
| A | `file1="ab"`, `file2="c"` | `abc` | `ungWv48Bz+pBQUDeXa4iI7ADYaOWF3qctBD/YfIAFa0=` |
| B | `file1="a"`,  `file2="bc"` | `abc` | `ungWv48Bz+pBQUDeXa4iI7ADYaOWF3qctBD/YfIAFa0=` |

Both yield `ungWv48Bz+pBQUDeXa4iI7ADYaOWF3qctBD/YfIAFa0=` (the SHA-256 of `"abc"`), so `Sha(A).equals(Sha(B))` despite A and B being distinct directories. Renaming a file (same content, different name) likewise leaves the hash unchanged, since paths are used only for ordering, never hashed.

Notably, `Cache.dirSha` (the production path that actually hashes directories) does **not** reuse `Sha`'s directory branch — it independently hashes each file to a fixed-length per-file SHA string and concatenates those. That separate reimplementation exists precisely because raw-content concatenation is unsafe, which corroborates that `Sha`'s own directory handling is broken.

There is no `ShaTest` in `eo-maven-plugin/src/test`, so this broken edge case is currently untested.

### Expected behavior

Distinct directory trees (different file boundaries, different file names, or different layouts) should produce distinct hashes. The hash should depend on both each file's path (relative to the root) and its content, in an unambiguous, length-delimited way — matching the "walks all files sorted by path" contract.

### Suggested fix

Make file boundaries and paths part of the digest. For example, for each file (in sorted order) feed a delimiter derived from the relative path and the content length before the content:

```java
final Path rel = path.relativize(file);
digest.update(rel.toString().getBytes(StandardCharsets.UTF_8));
digest.update((byte) 0);                                  // unambiguous separator
final long size = Files.size(file);
digest.update(ByteBuffer.allocate(Long.BYTES).putLong(size).array());
// ... then stream the file content as today ...
```

Equivalently, hash each file individually (like `Cache.dirSha` already does) and combine the fixed-length per-file digests together with the relative path. Adding a `ShaTest` that asserts the A/B trees above produce different hashes would lock in the fix.

circuito-suman:
Without a per-file delimiter, two directory trees whose file contents concatenate to the same byte sequence produce identical SHA-256 hashes, breaking the contract the class advertises. For example:

  dir A: file1="ab", file2="c"  → concat "abc" → same hash
  dir B: file1="a",  file2="bc" → concat "abc" → same hash

Fix: before streaming each file's content into the digest, feed:
  1. the relative file path (UTF-8 bytes) — so renames change the hash
  2. a NUL byte separator — unambiguous boundary between path and content
  3. the 8-byte file size (big-endian long) — so boundary shifts change the hash

Adds ShaTest covering the collision case, file-rename case, and determinism, as required by the acceptance criteria in #5254.

Fixes #5254

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
@@ -6,6 +6,8 @@ package org.eolang.maven;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.MessageDigest;
@@ -55,22 +57,36 @@ final class Sha {
         final MessageDigest digest = MessageDigest.getInstance("SHA-256");
         try (Stream<Path> walk = Files.walk(path)) {
             walk.filter(Files::isRegularFile)
-                .sorted(Comparator.comparing(Path::toString)).forEach(
-                    file -> {
-                        try (InputStream input = Files.newInputStream(file)) {
-                            final byte[] buffer = new byte[8192];
-                            int read = input.read(buffer);
-                            while (read != -1) {
-                                digest.update(buffer, 0, read);
-                                read = input.read(buffer);
-                            }
-                        } catch (final IOException ex) {
-                            throw new IllegalStateException(
-                                String.format("Failed to read '%s'", file), ex
-                            );
-                        }
-                    }
+    .sorted(Comparator.comparing(Path::toString)).forEach(
+        file -> {
+            try {
+                // 1. Hash the relative path so renames change the digest
+                final Path rel = path.relativize(file);
+                digest.update(
+                    rel.toString().getBytes(StandardCharsets.UTF_8)
                 );
+                // 2. NUL byte as unambiguous separator between path and content
+                digest.update((byte) 0);
+                // 3. Hash the file size so boundary shifts change the digest
+                final long size = Files.size(file);
+                digest.update(
+                    ByteBuffer.allocate(Long.BYTES).putLong(size).array()
+                );
+                try (InputStream input = Files.newInputStream(file)) {
+                    final byte[] buffer = new byte[8192];
+                    int read = input.read(buffer);
+                    while (read != -1) {
+                        digest.update(buffer, 0, read);
+                        read = input.read(buffer);
+                    }
+                }
+            } catch (final IOException ex) {
+                throw new IllegalStateException(
+                    String.format("Failed to read '%s'", file), ex
+                );
+            }
+        }
+    );
         }
         return Base64.getEncoder().encodeToString(digest.digest());
     }
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
@@ -7,6 +7,7 @@ package org.eolang.maven;
 import com.yegor256.Mktmp;
 import com.yegor256.MktmpResolver;
 import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -146,11 +147,7 @@ final class CacheTest {
         MatcherAssert.assertThat(
             "SHA-256 hash file has incorrect content",
             Files.readString(base.resolve(String.format("%s.sha256", tail)), encoding),
-            Matchers.equalTo(
-                Base64.getEncoder().encodeToString(
-                    MessageDigest.getInstance("SHA-256").digest(msg.getBytes(encoding))
-                )
-            )
+            Matchers.equalTo(CacheTest.hash(temp, source))
         );
     }
 
@@ -176,12 +173,7 @@ final class CacheTest {
                 cache.resolve(String.format("%s.sha256", tail)),
                 StandardCharsets.UTF_8
             ),
-            Matchers.equalTo(
-                Base64.getEncoder().encodeToString(
-                    MessageDigest.getInstance("SHA-256")
-                        .digest(content.getBytes(StandardCharsets.UTF_8))
-                )
-            )
+            Matchers.equalTo(CacheTest.hash(temp, source))
         );
     }
 
@@ -202,7 +194,7 @@ final class CacheTest {
                 cache.resolve(String.format("%s.sha256", tail)),
                 StandardCharsets.UTF_8
             ),
-            Matchers.equalTo(CacheTest.hash(content))
+            Matchers.equalTo(CacheTest.hash(temp, source))
         );
     }
 
@@ -218,16 +210,18 @@ final class CacheTest {
         final String first = "file1 content";
         final String second = "file2 content";
         final String result = String.format("%s%s", first, second);
-        Files.writeString(source.resolve("file1.txt"), first, StandardCharsets.UTF_8);
-        Files.writeString(source.resolve("file2.txt"), second, StandardCharsets.UTF_8);
+        final Path file1 = source.resolve("file1.txt");
+        final Path file2 = source.resolve("file2.txt");
+        Files.writeString(file1, first, StandardCharsets.UTF_8);
+        Files.writeString(file2, second, StandardCharsets.UTF_8);
         new Cache(cache, p -> result).apply(
             source,
             temp.resolve("out.txt"),
             source.getFileName()
         );
         final MessageDigest instance = MessageDigest.getInstance("SHA-256");
-        instance.update(CacheTest.hash(first).getBytes(StandardCharsets.UTF_8));
-        instance.update(CacheTest.hash(second).getBytes(StandardCharsets.UTF_8));
+        instance.update(CacheTest.hash(source, file1).getBytes(StandardCharsets.UTF_8));
+        instance.update(CacheTest.hash(source, file2).getBytes(StandardCharsets.UTF_8));
         MatcherAssert.assertThat(
             "SHA-256 hash file has incorrect content for folder with several files",
             Files.readString(
@@ -238,10 +232,24 @@ final class CacheTest {
         );
     }
 
-    private static String hash(final String content) throws NoSuchAlgorithmException {
-        return Base64.getEncoder().encodeToString(
-            MessageDigest.getInstance("SHA-256")
-                .digest(content.getBytes(StandardCharsets.UTF_8))
-        );
+    /**
+     * Compute the SHA-256 hash of a file, based on its path relative to
+     * {@code root}, its size, and its content.
+     * @param root The root directory used to relativize the file path.
+     * @param file The file to hash.
+     * @return Base64-encoded SHA-256 hash.
+     * @throws NoSuchAlgorithmException If SHA-256 is not available.
+     * @throws IOException If the file cannot be read.
+     */
+    private static String hash(final Path root, final Path file)
+        throws NoSuchAlgorithmException, IOException {
+        final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+        final Path rel = root.relativize(file);
+        digest.update(rel.toString().getBytes(StandardCharsets.UTF_8));
+        digest.update((byte) 0);
+        final long size = Files.size(file);
+        digest.update(ByteBuffer.allocate(Long.BYTES).putLong(size).array());
+        digest.update(Files.readAllBytes(file));
+        return Base64.getEncoder().encodeToString(digest.digest());
     }
 }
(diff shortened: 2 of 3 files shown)


0crat:
@circuito-suman It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`fix/sha-directory-collision`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`5261`". Your running score is [-6](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

circuito-suman:
Got it, will use the ticket number as branch name next time. Thanks for the feedback.

eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
```diff
@@ -146,11 +147,7 @@ void writesCorrectShaHash(@Mktmp final Path temp) throws IOException, NoSuchAlgo
         MatcherAssert.assertThat(
             "SHA-256 hash file has incorrect content",
             Files.readString(base.resolve(String.format("%s.sha256", tail)), encoding),
-            Matchers.equalTo(
-                Base64.getEncoder().encodeToString(
-                    MessageDigest.getInstance("SHA-256").digest(msg.getBytes(encoding))
-                )
-            )
+            Matchers.equalTo(CacheTest.hash(temp, source))
```
gemshrine, line 150: `CacheTest.hash(temp, source)` relativizes against `temp`, so it hashes `"message.txt" + NUL + 8B(size) + content`. But `Cache.sha(source)` delegates to `new Sha(source).toString()`, and `Sha.hash` walks a single file with `path = source` — meaning `path.relativize(file)` is the empty Path, and production hashes `"" + NUL + 8B(size) + content`. The assertion can never match, which explains why `mvn` fails on all three OS jobs.

Same mismatch in `generatesCorrectHashForLargeFile` (line 176), `generatesCorrectHashForTinyFile` (line 197), and the folder test at lines 223–224. Passing `source` (the file itself) as the relativization root — `hash(source, source)` — would give the empty rel prefix and match production.

eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
```diff
@@ -55,22 +57,36 @@ private static String hash(final Path path) throws IOException, NoSuchAlgorithmE
...
-                                String.format("Failed to read '%s'", file), ex
-                            );
-                        }
+    .sorted(Comparator.comparing(Path::toString)).forEach(
+        file -> {
+            try {
+                // 1. Hash the relative path so renames change the digest
+                final Path rel = path.relativize(file);
+                digest.update(
+                    rel.toString().getBytes(StandardCharsets.UTF_8)
```
gemshrine, line 66: `rel.toString()` uses the OS-native separator (`\` on Windows, `/` on Unix), so the same directory tree produces different digests on different OS. The sort key on line 60 (`Comparator.comparing(Path::toString)`) inherits the same OS dependency; for path shapes like `a1` (0x31) vs `a/b` (0x2F on Unix, `\` = 0x5C on Windows), the two entries land on different sides of each other and change the concatenation order.

Since `Cache` writes the hash to a `.sha256` file that a subsequent build (potentially on a different host) reads, this reintroduces a portability class of hash instability. Normalizing to `/` closes both cases:

```java
rel.toString().replace(java.io.File.separatorChar, '/').getBytes(StandardCharsets.UTF_8)
```

And the same normalization on the sort key.

gemshrine (commented):
Three concerns — one that explains the mvn failure on all three OS jobs, one cross-platform hash instability in `Sha` itself, and one about the caller `Cache.dirSha` still carrying the same class of collision this PR aims to close. Details in the inline comments.

eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
```diff
@@ -55,22 +57,36 @@ private static String hash(final Path path) throws IOException, NoSuchAlgorithmE
...
+                );
+                // 2. NUL byte as unambiguous separator between path and content
+                digest.update((byte) 0);
+                // 3. Hash the file size so boundary shifts change the digest
+                final long size = Files.size(file);
```
gemshrine, line 71: Hashing path + NUL + size + content closes the collision when someone calls `new Sha(dir).toString()` directly. But the maven-plugin's actual directory-hashing caller — `Cache.dirSha` at `Cache.java:142` — hashes each regular file with `new Sha(p).toString()` and then base64-joins the results. Because `Sha` on a single file uses `path = file`, `rel` is empty, and none of the new fields (path prefix, size) reach the outer aggregation. Two dirs `[foo/x="a"]` and `[bar/x="a"]` still produce identical Cache dir-hashes (same for flat `[x="a"]` vs nested `[sub/x="a"]`).

If `Fixes #5254` is meant to close the class of bug for `Cache`, the fix needs to reach `Cache.dirSha` — either by having it call `Sha` on the directory as a whole (which needs the cross-platform normalization from the other comment) or by hashing the per-file rel path inside `dirSha` too.

eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
```diff
@@ -55,22 +57,36 @@ private static String hash(final Path path) throws IOException, NoSuchAlgorithmE
...
-                            while (read != -1) {
-                                digest.update(buffer, 0, read);
-                                read = input.read(buffer);
-                            }
-                        } catch (final IOException ex) {
-                            throw new IllegalStateException(
-                                String.format("Failed to read '%s'", file), ex
-                            );
-                        }
+    .sorted(Comparator.comparing(Path::toString)).forEach(
```
morphqdd, line 60: This block's indentation is broken, not just in the diff view, I checked the actual file content directly. `.sorted(Comparator.comparing(Path::toString)).forEach(` sits at 4 spaces while the `walk.filter(...)` line right above it is at 12, and the lambda body inside drifts to its own inconsistent levels, with the closing `);` misaligned from everything around it. This reads as a formatting tool mangling the block, or a manual edit that lost track of indentation. Whatever the cause, this needs to be re-indented to match the rest of the method before it can pass a normal Java style check, let alone qulice.

morphqdd, line 60: Separate from the indentation: this still sorts by `Path::toString` on the path returned from `Files.walk(path)`, which is the absolute-ish walked path, while the actual hash input a few lines down uses `path.relativize(file)`, the relative path. Since every entry shares the same `path` prefix, the relative order matches the absolute order here, so it's not a bug, but it reads oddly to sort on one representation and hash on another. Sorting on `path.relativize(file).toString()` directly (computing `rel` before the sort, or inside a `Comparator.comparing` that relativizes) would make the two consistent and remove the need to double check why they don't have to be.

eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
```diff
@@ -55,22 +57,36 @@ private static String hash(final Path path) throws IOException, NoSuchAlgorithmE
...
+                final Path rel = path.relativize(file);
+                digest.update(
+                    rel.toString().getBytes(StandardCharsets.UTF_8)
+                );
+                // 2. NUL byte as unambiguous separator between path and content
+                digest.update((byte) 0);
+                // 3. Hash the file size so boundary shifts change the digest
+                final long size = Files.size(file);
+                digest.update(
+                    ByteBuffer.allocate(Long.BYTES).putLong(size).array()
```
morphqdd, line 73: `ByteBuffer.allocate(Long.BYTES).putLong(size).array()` allocates a new 8-byte buffer on every file in the walk. Not a correctness issue, and probably not significant for typical directory sizes, but for a large tree this is one avoidable allocation per file in a loop, which some `qulice`/PMD configurations flag on its own. Worth checking whether this trips anything in this repo's ruleset; if not, no action needed, just flagging since it's easy to hand-roll the 8 bytes into a small reusable array instead if it does.

eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
```diff
@@ -238,10 +232,24 @@ void generatesCorrectHashForEntireFolderWithSeveralFiles(
...
+        throws NoSuchAlgorithmException, IOException {
+        final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+        final Path rel = root.relativize(file);
+        digest.update(rel.toString().getBytes(StandardCharsets.UTF_8));
+        digest.update((byte) 0);
+        final long size = Files.size(file);
+        digest.update(ByteBuffer.allocate(Long.BYTES).putLong(size).array());
+        digest.update(Files.readAllBytes(file));
+        return Base64.getEncoder().encodeToString(digest.digest());
     }
```
morphqdd, line 254: File is missing a trailing newline (`\ No newline at end of file` in the diff). `ShaTest.java` has the same issue. Most style checks in this repo (and most editors) expect one; worth adding back before this fails a lint step over something unrelated to the actual fix.

eo-maven-plugin/src/test/java/org/eolang/maven/ShaTest.java
```diff
@@ -0,0 +1,66 @@
...
+    void providesSameHashForSameDirectory(@Mktmp final Path temp) throws Exception {
+        final Path dir = temp.resolve("C");
+        Files.createDirectories(dir);
+        Files.writeString(dir.resolve("a.txt"), "consistent");
+        MatcherAssert.assertThat(
+            "same directory must always produce the same hash",
+            new Sha(dir).toString(),
+            Matchers.equalTo(new Sha(dir).toString())
+        );
+    }
```
morphqdd, line 65: Same missing trailing newline as `CacheTest.java`.

morphqdd (changes requested):
The hashing scheme itself looks sound: I traced it against the collision example in the issue (dir A: file1="ab", file2="c" vs dir B: file1="a", file2="bc") and the path+NUL+size framing does produce different byte streams for the two cases, since a NUL byte can't legally appear inside a filename on any of the filesystems this runs on, so the path-then-separator boundary is unambiguous, and the fixed 8-byte size field makes the content boundary explicit rather than inferred. That part is correct. A few things below need fixing before this can merge, plus this PR currently shows as `CONFLICTING` against master and will need a rebase regardless.

yegor256:

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+785](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (5) comments.  Your running score is [+2256](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5261, opened 2026-06-24
introduce new concept of fragile objects and related syntax changes in language


yegor256:
Let's introduce new concept to the language: _fragile objects_.

Definition: if an object has at least one void attribute OR one of its attributes is fragile, the object is fragile.

Let's make it possible to specify void attributes vertically (both snippets are equivalent):

```
[name] > foo        |   [name bar test] > foo
  ? > bar           |
  ? > test          |
```

Also, let's add new syntax for atoms:

```
[name] > fopen /file
  ? > not-found /{string io.file-error}
  ? > cant-open /{io.foo.bar.super-error}
```

Here, the `/{string io.file-error}` is the formas to be provided by the `fopen` atom as arguments for `not-found`.

A fragile object can't be dispatched via the `.` (dot), but only via the `?.` (new operator).

Also, let's make it possible for an object to be either of some forma or `T` (new syntax construct), which is similar to `⊥` in phi calculus.

A more concrete example:

```
# The "fopen" object is fragile:
fopen > x
  "a.txt"
  [e]:cant-open
    oops *1 > @
      "failed to open %s because %s"
      e.file-name
      e.reason-of-failure
  [name error]:not-found
    broken-file error > @

# The dispatch via the dot is prohibited on "x":
x?.read 100 > d

# The "d" object is fragile, since "broken-data" is fragile:
read. > d
  x
  100
  [error]:cant-read
    broken-data > @
      read-error
        sprintf *1
          "can't read %d bytes: %s"
          error.bytes-requested
          error.reason-of-failure

# The "d" object is NOT fragile:
read. > d
  x
  100
  [e]:cant-read
    00- > @

[error] > broken-file
  [s cant-read] > read
    seq > @
      cant-read error
      T
```

Also, runtime exceptions (like out of memory or division by zero) are _not_ processed by EO programs. They lead to VM termination. They can't be caught by EO.

yegor256:
@maxonfjvipon I suggest we break it down into parts, like introduce `T` first, then new syntax for vertical void attrs, then type annotations for void attrs, etc.

maxonfjvipon:
@yegor256 let's discuss how `T` should be present in XMIR, how should it work in eo-runtime, what would be the behaviour of it

yegor256:
@maxonfjvipon I suggest we put it as `<o base="T"/>` into XMIR, and then introduce new `PhTerminator` class with `.take()` returning `PhTerminator` too.

[...]

PR #5265, opened 2026-06-25
Introduce `T` term syntax mapping to `⊥` in XMIR

### LINKED ISSUE #5264: Implementing `T` term syntax and mapping to `⊥` in XMIR without altering XSL pipeline
Part of the fragile-objects epic #5261. That epic introduces a new syntax construct `T`:

> let's make it possible for an object to be either of some forma or `T` (new syntax construct), which is similar to `⊥` in phi calculus

This ticket covers **only** the `T` construct: recognising it in the spec-driven parser, mapping it to `⊥` in XMIR, keeping `⊥` opaque through the XSL pipeline, and printing it back. The runtime meaning of `⊥` (termination / dataization behaviour), the `?.` operator, fragile-object detection, vertical voids, and the atom error-forma syntax are separate sub-tickets of #5261 and are **out of scope here**.

## What `T` is

`T` is a standalone *term* — a leaf object reference, usable anywhere an object reference is (argument or head). In phi-calculus terms it is `⊥` (bottom). From the epic's example it appears as a bare vertical argument:

```
seq > @
  cant-read error
  T
```

## Surface syntax

`T` is a new single-character reserved token, analogous to `Q` (ROOT) and `$` (XI). It collides with nothing: the `NAME` token (§2.3) starts with `[a-z]`, so no user identifier can be `T`, and `Q` is the only reserved uppercase glyph today. `T` carries no arguments and no children — it is head-only, like `$`.

## XMIR contract

Following the §9.3 source-token → XMIR-character table (the single source of truth: `Q`→`Φ`, `@`→`φ`, `^`→`ρ`, `$`→`ξ`), `T` maps to the raw bottom symbol:

| Source token | XMIR character | Used as |
|---|---|---|
| `T` (TERM) | `⊥` | `@base='⊥'` for the bottom term |

So `T` emits a single self-contained element:

```xml
<o base="⊥" line="L" pos="P"/>
```

`⊥` has zero occurrences in the codebase today, so this introduces a clean new base value.

## Keep `⊥` opaque through the XSL pipeline

`⊥` is an atomic terminal like `ξ` / `∅`: no parse pass may resolve it as a reference, prefix it with a package, or fold it into an FQN. Audit of the `parse/` sheets found two passes that will corrupt a bare `⊥` unless taught about it, plus several to confirm via test:

- [ ] `_specials.xsl` — add `eo:bottom` = `⊥`, reused by the guards below (matches the existing `eo:empty` / `eo:xi` style).
- [ ] **`add-default-package.xsl`** (line 52, and the symmetric `@atom` guard line 67) — `⊥` would otherwise be rewritten to `Φ.⊥`. Add `⊥` to the dotless-specials exclusion.
- [ ] **`build-fqns.xsl`** (line 206, `mode="no-dots"`) — `⊥` would otherwise be resolved as a name into an FQN. Add `⊥` to the exclusion (`@base!='ξ' and @base!='ρ' and @base!=$eo:empty …`).
- [ ] Confirm-by-test (expected to be identity, but verify): `resolve-aliases.xsl`, `decorate.xsl`, `roll-bases.xsl`, `mandatory-as.xsl`. A `⊥` used as a positional arg correctly receives `@as` like any other arg.
- [ ] Print side — confirm `tuples-to-stars.xsl`, `unnecessary-as.xsl`, `dataized-to-const.xsl`, `inline-cactoos.xsl` leave `⊥` untouched.

## Reverse printing

`to-eo.xsl` must map `<o base="⊥"/>` back to `T`. Without a dedicated `when` case the default `@base` fall-through would print the literal `⊥`, which is not valid EO source.

## Implementation checklist (spec-driven parser)

- [ ] `PARSER_SPEC.md` — add `T` (`TERM`) to the §2.3 token table and the §9.3 mapping table; both are the authoritative contract.
- [ ] `Value.Kind` — add a `TERM` kind.
- [ ] `Tokens.java` — recognise a bare `T` head where `Q`/`$`/`^` are detected.
- [ ] `Emissions.java` — map the `TERM` kind to `@base='⊥'`.
- [ ] XSL pipeline changes per the section above.
- [ ] `to-eo.xsl` — `@base='⊥'` → `T`.

## Tests

- [ ] Parse pack under `eo-packs/parse/` — `T` as a bare arg emits `<o base='⊥'/>`, **and survives the full sheet list unchanged** (not `Φ.⊥`, no FQN).
- [ ] Print pack under `print/` — round-trips `<o base='⊥'/>` back to `T`.
- [ ] `EoTest` case asserting the `@base='⊥'` XPath for a bare `T` head.

maxonfjvipon:
The fragile-objects epic #5261 needs a bottom term, `T`, analogous to `⊥` in 𝜑-calculus. This adds it to the spec-driven parser as a self-contained leaf: the source token `T` becomes `<o base="⊥"/>`, and the printer turns it back into `T`.

The mapping follows the existing §9.3 table (`Q`→`Φ`, `$`→`ξ`, `^`→`ρ`, void→`∅`), so `T` slots in as one more reserved single-character token. It collides with nothing — `NAME` is lowercase-led and `Q` is the only other uppercase glyph.

The subtle part is keeping `⊥` opaque through the XSL pipeline. Two passes would otherwise mangle a bare `⊥`: `add-default-package` prefixed it to `Φ.⊥`, and `build-fqns` tried to resolve it as a name. Both now exclude it, through a new special in `_specials.xsl`:

```xml
<xsl:variable name="eo:bottom" select="'⊥'"/>
```

Scope is syntax only. The runtime meaning of `⊥`, the `?.` operator, and the rest of #5261 are separate tickets. Tests cover the parse (`T` → `⊥` surviving the full sheet chain), the print round-trip, and the new `Value.Kind.TERM`.

Closes #5264.

### DIFF
--- a/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
+++ b/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
@@ -141,6 +141,9 @@
       <xsl:when test="$first='ξ'">
         <xsl:value-of select="$rho"/>
       </xsl:when>
+      <xsl:when test="$first='⊥'">
+        <xsl:text>new PhTerminated()</xsl:text>
+      </xsl:when>
       <xsl:otherwise>
         <xsl:message terminate="yes">
           <xsl:text>FQN must start with either with Φ or ξ, but </xsl:text>
@@ -539,7 +542,7 @@
     <xsl:value-of select="$name"/>
     <xsl:text> = </xsl:text>
     <xsl:choose>
-      <xsl:when test="@base='ξ' or @base='Φ'">
+      <xsl:when test="@base='ξ' or @base='Φ' or @base='⊥'">
         <xsl:value-of select="eo:fqn-start(@base, $rho)"/>
         <xsl:text>;</xsl:text>
       </xsl:when>
--- a/eo-maven-plugin/src/test/resources/org/eolang/maven/transpile-packs/bottom-to-java.yaml
+++ b/eo-maven-plugin/src/test/resources/org/eolang/maven/transpile-packs/bottom-to-java.yaml
@@ -1,0 +1,19 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets:
+  - /org/eolang/maven/transpile/set-locators.xsl
+  - /org/eolang/maven/transpile/set-original-names.xsl
+  - /org/eolang/maven/transpile/classes.xsl
+  - /org/eolang/maven/transpile/package.xsl
+  - /org/eolang/maven/transpile/attrs.xsl
+  - /org/eolang/maven/transpile/data.xsl
+  - /org/eolang/maven/transpile/to-java.xsl
+asserts:
+  - /object[not(errors)]
+  - /object/class/java[contains(text(), 'new PhTerminated()')]
+input: |
+  # No comments.
+  [] > app
+    T > x
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -167,6 +167,7 @@ The parser recognises the following lexical tokens:
 | `RHO` | `^` |
 | `ROOT` | `Q` |
 | `XI` | `$` |
+| `TERM` | `T` — the bottom term (§9.3), similar to `⊥` in 𝜑-calculus. A self-contained single-character token; carries no arguments and no chain. |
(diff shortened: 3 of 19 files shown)


maxonfjvipon:
@yegor256 please check

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (259 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+378](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5266, opened 2026-06-25
fixed #5260: return self from floor for values >= 2^52

### LINKED ISSUE #5260: number.floor incorrectly returns saturated value for finite numbers >= 2^63 due to as-i64 implementation

## Summary
number.floor computes `trunc = ^.as-i64.as-number` (truncate toward zero) and then subtracts 1 when `trunc.gt ^`. as-i64 is backed by EOnumber$EOas_i64, which does `Double.longValue()`. For finite doubles whose magnitude is >= 2^63 (e.g. 1e20, 1e30), longValue() saturates to Long.MAX_VALUE / Long.MIN_VALUE, so trunc becomes ~9.22e18 instead of the original value. Since every double >= 2^52 is already an integer, floor should return the input unchanged, but instead floor(1e20) returns ~9.223e18.

## Location
[`eo-runtime/src/main/eo/number.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number.eo#L119) lines 119-126.

## Buggy code
```eo
  [] > floor
    if. > @
      ^.is-finite.not
      ^
      if.
        trunc.gt ^
        trunc.minus 1
        trunc
    ^.as-i64.as-number > trunc
```

## Why it's wrong
floor must return the largest integer <= x; for any finite double with |x| >= 2^52 the value is itself an exact integer, so floor(x) == x. Reproduced in Java: x=1e20 -> Double.longValue()=9223372036854775807 -> trunc=9.22337e18; trunc.gt x is false, so eo-floor=9.22337e18 while Math.floor(1e20)=1e20 (mismatch). Same for 9.3e18, 1e30, and the negative counterparts. The is-finite guard does not cover these (they are finite), so the saturated path is taken.

Verifier confirmation: Confirmed independently. number.floor (number.eo:119-127) computes trunc = ^.as-i64.as-number and returns trunc (minus 1 only when trunc.gt ^). as-i64 is implemented by EOnumber$EOas_i64.java:28 as Double.longValue(), which by JLS narrowing-conversion rules saturates to Long.MAX_VALUE / Long.MIN_VALUE for any finite double whose magnitude is >= 2^63. Every double with |x| >= 2^52 is already an exact integer, so floor(x) should equal x, but for |x| >= 2^63 (e.g. 1e20, 9.3e18, 1e30) floor instead returns a value capped at about ±9.223e18. I reproduced this exactly in Java: x=1e20 -> longValue()=9223372036854775807 -> trunc=9.22337e18; trunc.gt x is false, so the eo result is 9.22337e18 while Math.floor(1e20)=1e20 (mismatch), with identical mismatches for -1e20, 9.3e18, -9.3e18, and 1e30. The is-finite.not guard only short-circuits nan/inf, so these finite values take the buggy saturated path; the trunc.gt ^ correction never fires for the positive saturated case. This is a genuine correctness bug, not intended behavior.

## Expected behavior
floor(1e20) == 1e20 (and generally floor(x)==x for any finite x with |x| >= 2^52). Currently it returns a saturated ~9.22e18-magnitude value.

## Suggested fix
Treat finite values whose magnitude is too large to be a non-integer (|x| >= 2^52, or more simply outside the i64-representable range) as already-integer and return the input unchanged, rather than routing them through as-i64. E.g. when `^.as-bytes`-based magnitude check shows the value cannot have a fractional part, return ^; only apply the as-i64 truncate-and-adjust for values within i64 range.

arnabnandy7:
## Description

This PR fixes a bug in `number.floor` where finite numbers of magnitude $\ge 2^{63}$ were incorrectly returning saturated values (about $\pm 9.22 \times 10^{18}$) due to narrowing conversion to `i64`.

### Problem
`number.floor` was computing `trunc = ^.as-i64.as-number`. The conversion `as-i64` is backed by `Double.longValue()`. For finite doubles whose magnitude is $\ge 2^{63}$, `Double.longValue()` saturates to `Long.MAX_VALUE` or `Long.MIN_VALUE` per Java Language Specification narrowing rules, causing `floor` of values like `1e20` or `-1e20` to return the saturated boundary value.

### Solution
Since double-precision floating point numbers have 52 mantissa bits, any double value with a magnitude $\ge 2^{52}$ cannot represent a fractional part and is already an exact mathematical integer. 

We update the `floor` implementation to check if the magnitude is $\ge 2^{52}$ (specifically $\ge 4503599627370496$ or $\le -4503599627370496$), and if so, return the number itself unchanged. This bypasses the narrowing conversion path entirely for values where the result is guaranteed to be equal to the input.

## Related Issues
Closes #5260

### DIFF
--- a/eo-runtime/src/main/eo/number.eo
+++ b/eo-runtime/src/main/eo/number.eo
@@ -118,7 +118,11 @@
   # the original, so we subtract `1` in that single case.
   [] > floor
     if. > @
-      ^.is-finite.not
+      or.
+        ^.is-finite.not
+        or.
+          ^.gte 4503599627370496
+          ^.lte -4503599627370496
       ^
       if.
         trunc.gt ^
@@ -393,6 +397,18 @@
     eq. > @
       -5.floor
       -5
+
+  # Tests that floor of a very large positive number >= 2^52 returns itself.
+  [] +> tests-floor-of-large-positive-integer
+    eq. > @
+      100000000000000000000.floor
+      100000000000000000000
+
+  # Tests that floor of a very large negative number <= -2^52 returns itself.
+  [] +> tests-floor-of-large-negative-integer
+    eq. > @
+      -100000000000000000000.floor
+      -100000000000000000000
 
   # Tests that floor of zero returns zero.
   [] +> tests-floor-of-zero


0crat:
@arnabnandy7 It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`fix/number-floor-overflow`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`5265`". Your running score is [-6](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@arnabnandy7 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5271, opened 2026-06-27
feat(#4752): implement sprintf in pure EO

asmirnov-backend:
Part #4752.

Reimplements `tt.sprintf` in pure EO (was a JVM/Node atom), supporting `%s %d %f %x %b` and positional `%N$x` references.

`walk` iterates over `format.as-bytes` using the native `bytes.size`/`bytes.slice` atoms rather than the recursive `string.length`/`string.slice` (which would make formatting O(n²) in allocation and exhaust the heap under the parallel test suite). `%` (0x25) never occurs inside a UTF-8 multi-byte sequence, so literal bytes copy through unchanged.

Number/bytes rendering is delegated to the reusable converters from #5272 (`as-decimal`, `as-fixed`, `as-hex`), so `sprintf` keeps only its parser and specifier dispatch.

The old Java atom is renamed to `EOsprintfDeprecated` to avoid a class-name clash with the transpiler-generated `EOsprintf`; a `@todo` puzzle tracks removing it (plus `SprintfArgs` and its test) in a follow-up.

EO tests cover all specifiers, positional refs, float truncation/padding, and error paths.


asmirnov-backend:

yegor256:
@asmirnov-backend just delete `EOsprintfDeprecated.java` file. why do we need it in the repo?

asmirnov-backend:
@yegor256 thanks, deleted EOsprintfDeprecated.java. I'd originally deprecated it plus left a puzzle to avoid too many hits-of-code in one PR

yegor256:
Hi @asmirnov-backend, please forgive us for closing this pull request. It has had merge conflicts with `master` for over a month, so we're closing it to keep the queue manageable. If you'd like, please feel free to submit a new pull request, perhaps similar to this one, but rebased on the current `master` branch. Thank you for the contribution!

PR #5273, opened 2026-06-28
Fix single-byte bytes string representation

### LINKED ISSUE #5253: `BytesRaw.asString()` method omits trailing dash for single-byte values causing inconsistency with `Hex.asString()` and `BYTES` grammar

## Summary

`BytesRaw.asString()` in the EO runtime serializes a **single byte without the canonical trailing dash**, returning e.g. `"FF"` instead of `"FF-"`. The parser side (`Hex.asString()` / `BytesToHex`) explicitly special-cases single bytes to emit a trailing dash. As a result, the runtime and the parser disagree on the canonical `BYTES` serialization for the single-byte case.

## Location

[`eo-runtime/src/main/java/org/eolang/BytesRaw.java`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/BytesRaw.java#L136-L148), lines 136-148.

## Buggy code

```java
@Override
public String asString() {
    final StringBuilder out = new StringBuilder(3);
    for (final byte bte : this.data) {
        if (out.length() > 0) {
            out.append('-');
        }
        out.append(String.format("%02X", bte));
    }
    if (this.data.length == 0) {
        out.append("--");
    }
    return out.toString();
}
```

A dash is only prepended *between* bytes (`out.length() > 0`), and `"--"` is only appended when `data.length == 0`. So:

| Input | returns | canonical |
|-------|---------|-----------|
| `{}` | `"--"` | `"--"` (correct) |
| `{0xFF}` | `"FF"` | `"FF-"` (WRONG) |
| `{0xAA, 0xBB}` | `"AA-BB"` | `"AA-BB"` (correct) |

## Why it's wrong

The parser emits a trailing dash for a single byte — see [`eo-parser/src/main/java/org/eolang/parser/Hex.java`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/Hex.java#L73-L88), where the single-byte branch is `result = String.format("%02X-", this.bytes[0]);`. Canonical BYTES form: empty -> `--`, single byte -> `FF-` (trailing dash), multiple bytes -> `AA-BB` (no trailing dash). The runtime matches the parser for empty and multi-byte but not the single-byte case, where it produces `FF` instead of `FF-`. There is also no runtime test covering `asString()` for a single byte, which is why this slipped through.

## Expected behavior

```java
new BytesOf(new byte[]{(byte) 0xFF}).asString()  // should return "FF-", currently returns "FF"
```

## Suggested fix

Special-case `data.length == 1` to emit `String.format("%02X-", this.data[0])` (mirroring the parser), and add a runtime test asserting the single-byte case returns `"FF-"`.

DragonFSKY:
Fixes #5253.\n\nThis updates runtime byte formatting so single-byte values include the trailing dash, matching the parser-side representation and the BYTES grammar expectation.\n\nTested with:\n- mvn -pl eo-runtime -Dtest=BytesOfTest -DskipITs test

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/BytesRaw.java
+++ b/eo-runtime/src/main/java/org/eolang/BytesRaw.java
@@ -134,17 +134,22 @@ final class BytesRaw implements Bytes {
 
     @Override
     public String asString() {
-        final StringBuilder out = new StringBuilder(3);
-        for (final byte bte : this.data) {
-            if (out.length() > 0) {
-                out.append('-');
-            }
-            out.append(String.format("%02X", bte));
-        }
+        final String result;
         if (this.data.length == 0) {
-            out.append("--");
+            result = "--";
+        } else if (this.data.length == 1) {
+            result = String.format("%02X-", this.data[0]);
+        } else {
+            final StringBuilder out = new StringBuilder(this.data.length * 3);
+            for (final byte bte : this.data) {
+                if (out.length() > 0) {
+                    out.append('-');
+                }
+                out.append(String.format("%02X", bte));
+            }
+            result = out.toString();
         }
-        return out.toString();
+        return result;
     }
 
     @Override
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -461,7 +461,7 @@ public class PhDefault implements Phi, Cloneable {
     private String structural() {
         final List<String> list = new ArrayList<>(this.attrs.size());
         if (this.data != null) {
-            list.add(String.format("D> %s", new BytesOf(this.data).asString()));
+            list.add(String.format("D> %s", PhDefault.termBytes(this.data)));
         }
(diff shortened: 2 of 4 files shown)


0crat:
@dragonfsky It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`fix-bytesraw-single-byte-string`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`5272`". Your running score is [-12](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@dragonfsky Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5274, opened 2026-06-28
Fix stream leak on repeated file open

### LINKED ISSUE #5258: `Files.open()` method leaks OS file handles on pre-existing entries causing resource exhaustion

## Summary
open(name) eagerly constructs a new InputStream and a new OutputStream and then calls this.streams.putIfAbsent(name, new Object[]{...}). The two streams are created as constructor arguments BEFORE putIfAbsent runs. If an entry for name already exists (e.g. a re-entrant/nested file.open.process-file on the same path, or two concurrent opens of the same path), putIfAbsent does NOT store the new array and returns the existing one. The two newly opened OS file handles are then unreachable and never closed - a genuine resource (file-descriptor) leak. They are also not the streams that subsequent read()/write()/close() operate on (those use the pre-existing array), so the leaked handles stay open until GC/process exit. Even on the normal single-open path the streams are created before the membership check, so any time the key is present the handles leak.

## Location
[`eo-runtime/src/main/java/org/eolang/EOfs/Files.java`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/EOfs/Files.java#L60) lines 60-74.

## Buggy code
```java
void open(final String name) throws IOException {
    this.lock.lock();
    try {
        final Path path = Paths.get(name);
        this.streams.putIfAbsent(
            name,
            new Object[]{
                java.nio.file.Files.newInputStream(path),
                java.nio.file.Files.newOutputStream(path, StandardOpenOption.APPEND),
            }
        );
    } finally {
        this.lock.unlock();
    }
}
```

## Why it's wrong
putIfAbsent only stores when the key is absent, but the arguments (the two streams) are always evaluated. The method opens OS resources unconditionally yet only retains/closes them when the key was absent. There is no fallback to close the orphaned streams on the already-present branch. close() removes the key, so a sequential open->close->open is fine, but nested/concurrent same-path opens leak descriptors. The @SuppressWarnings("java:S2095") (resource-not-closed) on the method shows the author is aware streams are opened here but the absent-key case is not handled.

Verifier confirmation: Confirmed by reading Files.java:60-74. open() builds the Object[] with java.nio.file.Files.newInputStream(path) and newOutputStream(path, APPEND) as arguments to putIfAbsent (lines 64-69), so both OS file handles are opened unconditionally before any membership check. putIfAbsent does NOT store the new array when the key is already present and the returned existing value is discarded, leaving the two freshly opened streams unreferenced and unclosed — a genuine file-descriptor leak with no fallback close on the present branch. close() removes the key, so a sequential open->close->open is safe, but the path's ReentrantLock is re-entrant and the caller EOfile$EOopen$EOprocess_file.lambda() runs an arbitrary user-supplied scope (lines 41-43) between open() and close(); if that scope performs another file.open.process-file on the same path (or two file objects target the same path nested), open() is re-entered while the first entry is still present, hitting the leaking branch on the same thread. The @SuppressWarnings("java:S2095") (resource-not-closed) on the method corroborates that the author opened streams here without try-with-resources. The finder's diagnosis of the leak mechanism is accurate. Severity is medium because the common single-open/sequential path is fine; the leak only manifests on nested/re-entrant or concurrent same-path opens, but when triggered it leaks descriptors that survive until GC/process exit.

## Expected behavior
open() on an already-open name must not leak file descriptors: either it should be a no-op that opens nothing when the entry already exists, or it must close the streams it opened if putIfAbsent finds the key present.

## Suggested fix
Guard before opening: under the lock, if streams.containsKey(name) return (or throw) without opening anything; only open and put when absent. Alternatively capture the streams in locals, call putIfAbsent, and if it returns non-null close the two locals you just opened.

DragonFSKY:
Fixes #5258.

## Summary
- skip opening new streams when a file path is already registered
- reuse the existing stream entry instead of creating unreachable handles
- add a focused regression test for the already-open path

This intentionally keeps the existing one-entry-per-path semantics and does not introduce ref-counting or independent streams per open.

## Tests
- mvn -pl eo-runtime -Dtest=FilesTest -DskipITs test
- mvn -pl eo-runtime -Pqulice -DskipTests verify *(failed in EO transpile on unrelated foreign objects such as `ss.mapped` / `io.bytes-as-input`; no remaining qulice violation was reported for the changed files after fixing import order)*

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/EOfs/Files.java
+++ b/eo-runtime/src/main/java/org/eolang/EOfs/Files.java
@@ -16,6 +16,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
 import java.util.Arrays;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -47,7 +48,15 @@ final class Files {
      * Ctor.
      */
     private Files() {
-        this.streams = new ConcurrentHashMap<>(0);
+        this(new ConcurrentHashMap<>(0));
+    }
+
+    /**
+     * Ctor.
+     * @param streams File streams
+     */
+    Files(final Map<String, Object[]> streams) {
+        this.streams = new ConcurrentHashMap<>(streams);
         this.lock = new ReentrantLock();
     }
 
@@ -60,8 +69,11 @@ final class Files {
     void open(final String name) throws IOException {
         this.lock.lock();
         try {
+            if (this.streams.containsKey(name)) {
+                return;
+            }
             final Path path = Paths.get(name);
-            this.streams.putIfAbsent(
+            this.streams.put(
                 name,
                 new Object[]{
                     java.nio.file.Files.newInputStream(path),
--- a/eo-runtime/src/test/java/org/eolang/EOfs/FilesTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EOfs/FilesTest.java
@@ -12,10 +12,14 @@ package org.eolang.EOfs; // NOPMD
 import com.yegor256.Mktmp;
 import com.yegor256.MktmpResolver;
 import java.io.BufferedWriter;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.Map;
 import org.eolang.ExFailure;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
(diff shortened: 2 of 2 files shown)


0crat:
@dragonfsky It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`fix-files-open-existing-leak`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`5273`". Your running score is [-6](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:
@DragonFSKY thanks!

0crat:
@dragonfsky Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+20](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5281, opened 2026-06-30
Introduce the ⊥ (PhTerminator) runtime: propagate-until-forced, opt-in AtomTyped, predictable-vs-panic Heaps failures

### LINKED ISSUE #5280: `PhTerminated` class misnamed, should be `PhTerminator` to reflect active termination role
The runtime object for the ⊥ ("bottom") term (introduced in #5265) is named `PhTerminated`. Rename it to `PhTerminator` to reflect its role as the active agent that terminates a computation, not an object that has been terminated.

Touches four files:

- `eo-runtime/src/main/java/org/eolang/PhTerminated.java` → `PhTerminator.java`
- `eo-runtime/src/test/java/org/eolang/PhTerminatedTest.java` → `PhTerminatorTest.java`
- `eo-maven-plugin/.../transpile/to-java.xsl` — `new PhTerminated()` → `new PhTerminator()` (the `@base=⊥` → Java mapping)
- `eo-maven-plugin/.../transpile-packs/bottom-to-java.yaml` — the corresponding assertion

Pure rename, no behaviour change. Part of #5261.

maxonfjvipon:
Part of the fragile-objects epic #5261. Builds the runtime half of the ⊥ ("bottom") / termination + fallback mechanism and applies it to a real atom.

**1. `PhTerminated` → `PhTerminator`** — role name (the agent that terminates), plus a single `fail()` factory instead of duplicated `new ExFailure(...)`. Includes the `⊥` → `new PhTerminator()` mapping in `to-java.xsl`.

**2. Taking an absent/unset attribute returns ⊥** — `AtVoid.get()` and `PhDefault.take()` return a `PhTerminator` instead of throwing `ExUnset`. (The parent-reference `ρ` cases keep throwing; `hasRho()` relies on that.)

**3. ⊥ propagates until forced** — `PhTerminator` throws only on `delta()` (dataization). `take`/`copy` propagate another ⊥ and `put` is a no-op, so ⊥ can be returned, copied, and have args bound into it, failing only when the outer code dataizes it.

**4. `AtomTyped` is opt-in** — the atom return-type check now runs only when the `eo.typing` system property is `true` (default `false`, wired through the eo-runtime pom). It is off by default because an atom with an error-branch returns a union (declared forma *or* fallback/⊥) that a single forma cannot express.

**5. First atom on the model: `malloc.of.allocated.read`** — declares a `cant-read` error-branch void. Failures are split by predictability:
   - **Predictable** (range outside the block): `Heaps.fits(id, offset, length)` is the single source of the bounds rule; the atom branches on it (no try-catch, no duplicated arithmetic) and returns the caller's `cant-read` fallback, or ⊥ when none was given.
   - **Unpredictable** (block not allocated): `Heaps` throws `ExFailure`, which EO `try` cannot catch — it panics to the top and terminates.

   `Heaps.read` keeps its own over-limit throw (via `fits`), so it stays safe for any caller.

`eo-runtime` is green with the default flag; `-Deo.typing=true` re-activates the type check.

Closes #5280.

### DIFF
--- a/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
+++ b/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
@@ -142,7 +142,7 @@
         <xsl:value-of select="$rho"/>
       </xsl:when>
       <xsl:when test="$first='⊥'">
-        <xsl:text>new PhTerminated()</xsl:text>
+        <xsl:text>new PhTerminator()</xsl:text>
       </xsl:when>
       <xsl:otherwise>
         <xsl:message terminate="yes">
--- a/eo-maven-plugin/src/test/resources/org/eolang/maven/transpile-packs/bottom-to-java.yaml
+++ b/eo-maven-plugin/src/test/resources/org/eolang/maven/transpile-packs/bottom-to-java.yaml
@@ -12,7 +12,7 @@ sheets:
   - /org/eolang/maven/transpile/to-java.xsl
 asserts:
   - /object[not(errors)]
-  - /object/class/java[contains(text(), 'new PhTerminated()')]
+  - /object/class/java[contains(text(), 'new PhTerminator()')]
 input: |
   # No comments.
   [] > app
--- a/eo-parser/src/main/resources/XMIR.xsd
+++ b/eo-parser/src/main/resources/XMIR.xsd
@@ -153,6 +153,16 @@
         </xs:documentation>
       </xs:annotation>
     </xs:attribute>
+    <xs:attribute name="types" type="non-empty">
+      <xs:annotation>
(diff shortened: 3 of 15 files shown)


maxonfjvipon:
@yegor256 please take a look

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (462 >= 100); -8 for way too many hits-of-code (462 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+477](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5282, opened 2026-06-30
#5256: i64.div by zero throws catchable org.eolang.error

### LINKED ISSUE #5256: i64.div by zero results in an uncatchable JVM ArithmeticException, not an EO error as specified

## Summary
EOi64$EOdiv evaluates the quotient as a raw Java `long / long`. When the divisor (attribute x) is zero, the JVM throws java.lang.ArithmeticException: / by zero. AtomSafe.lambda() rethrows any RuntimeException unchanged (it only wraps checked Exceptions into ExFailure), and the EO `try` object only catches EOerror.ExError. So an i64 division by zero surfaces as an uncaught ArithmeticException that propagates out of the interpreter and cannot be caught by EO-level `try`, unlike a normal EO error object.

## Location
[`eo-runtime/src/main/java/org/eolang/EOi64$EOdiv.java`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/EOi64$EOdiv.java#L33) lines 33-38.

## Buggy code
```java
            new Data.ToPhi(
                new BytesOf(
                    new Expect.I64(Expect.at(this, Phi.RHO)).it()
                        / new Expect.I64(Expect.at(this, "x")).it()
                ).take()
            )
```

## Why it's wrong
The Javadoc for i64.div in i64.eo states: "An org.eolang.error is returned if x is equal to 0." No such error is produced. Tracing the runtime: EOi64$EOdiv does `Expect.I64(...).it() / Expect.I64(...).it()` on Longs (a Java long division), which throws ArithmeticException for a zero divisor; AtomSafe (lines 34-35) catches RuntimeException and simply rethrows it; EOtry.lambda() (line 36) only has `catch (final EOerror.ExError ex)`, so this JVM exception is not converted into a catchable EO error. The float number.div path is fine (IEEE 754 gives Infinity, and tests-zero-division exists for it), but i64.div has no zero guard and no test.

Verifier confirmation: Independently confirmed against the source. EOi64$EOdiv.lambda() (lines 35-36) computes `new Expect.I64(...).it() / new Expect.I64(...).it()`, and Expect.I64.it() returns a java.lang.Long, so the division unboxes to a primitive `long / long`. A zero divisor therefore throws java.lang.ArithmeticException: / by zero; the class has no zero guard and never constructs an EOerror. AtomSafe.lambda() (lines 35-36) catches RuntimeException and rethrows it unchanged (ArithmeticException is a RuntimeException, so it is NOT wrapped into the ExFailure path that only handles checked exceptions). EOtry.lambda() (line 36) only catches EOerror.ExError, and ExError extends ExAbstract extends RuntimeException — a sibling of ArithmeticException, not a supertype — so the ArithmeticException propagates past EO `try` uncaught. This directly contradicts the documented contract in i64.eo lines 79-82 ("An org.eolang.error is returned if x is equal to 0"). The supposed regression test throws-on-division-i64-by-i64-zero does not protect against this: per to-java.xsl lines 813-821, a `throws`-prefixed test transpiles to Assertions.assertThrows(Exception.class, ...), which passes for any Exception including the raw ArithmeticException, so it never verifies that a catchable EO error object is produced. The finder did not misread the code; the parallel float number.div path correctly relies on IEEE-754 Infinity and is unaffected.

## Expected behavior
i64.div by zero should yield an org.eolang.error object (an EOerror.ExError) that EO `try` can catch, matching the documented behavior.

## Suggested fix
Guard the divisor in EOi64$EOdiv: if the right operand is 0, return an EOerror (e.g. new PhWith(new EOerror(), ...)) / throw EOerror.ExError with a descriptive message, instead of performing the raw long division. Alternatively define div in i64.eo to check `x.eq 0` and return `error "..."`.

morphqdd:
## What happens

`EOi64$EOdiv` evaluates the quotient as a raw Java `long / long`. When the
divisor (attribute `x`) is zero, the JVM throws
`java.lang.ArithmeticException: / by zero`. `AtomSafe.lambda()` rethrows any
`RuntimeException` unchanged (it only wraps checked `Exception`s into
`ExFailure`), and the EO `try` object only catches `EOerror.ExError`. So an
`i64` division by zero surfaces as an uncaught `ArithmeticException` that
propagates out of the interpreter and cannot be caught by EO-level `try`,
unlike a normal EO error object.

The Javadoc for `i64.div` in `i64.eo` states: "An `org.eolang.error` is
returned if `x` is equal to 0." No such error was actually produced.

The existing regression test, `throws-on-division-i64-by-i64-zero`, doesn't
protect against this: a `throws`-prefixed `.eo` test transpiles to
`Assertions.assertThrows(Exception.class, ...)`, which passes for any
`Exception` including the raw `ArithmeticException` — it never verifies that
a *catchable EO error object* is produced.

## What should happen

`i64.div` by zero should yield an `org.eolang.error` object (an
`EOerror.ExError`) that EO `try` can catch, matching the documented
behavior.

## Fix

Guarded the divisor in `EOi64$EOdiv.lambda()`: if it's `0`, throw
`EOerror.ExError` directly instead of performing the raw `long` division.
`AtomSafe` rethrows `RuntimeException` (which `ExError` is, via `ExAbstract`)
unchanged, so it propagates correctly to `EOtry`.

Kept the original left-to-right evaluation order — `ρ` (dividend) evaluated
before `x` (divisor) — so the existing `EOi64ExpectTest` parameterized tests
(which rely on that order to assert the correct attribute name in type-error
messages) keep passing.

`@SuppressWarnings("PMD.UnnecessaryLocalRule")` mirrors the same pattern
already used in the sibling `EOnumber$EOdiv` for the identical
evaluation-order requirement.

## Test

Added `tests-i64-div-by-zero-throws-catchable-error` in `i64.eo`: wraps
`2.as-i64.div 0.as-i64` in `try`, asserting the error is caught (rather than
propagating as an uncatchable exception), unlike the existing `throws-*`
test which only checks that *some* exception is thrown.

## Verification

- `mvn test -pl eo-runtime` — full suite green (`EOi64Test`: 28/28,
  `EOi64ExpectTest`: 8/8, including the new test and the pre-existing
  parameterized tests that depend on evaluation order)
- `mvn qulice:check -Pqulice -pl eo-runtime` — violation count for
  `EOi64$EOdiv.java` reduced by the fix; the one remaining flag on that file
  (`UnicodeInCode` for `Φ`) is a pre-existing pattern shared with sibling
  atoms `EOi64$EOplus` and `EOi64$EOtimes`, not introduced by this change

Closes #5256

### DIFF
--- a/eo-runtime/src/main/eo/i64.eo
+++ b/eo-runtime/src/main/eo/i64.eo
@@ -221,6 +221,17 @@
   # Tests that i64 division by zero throws an error.
   2.as-i64.div 0.as-i64 > [] +> throws-on-division-i64-by-i64-zero
 
+  # Tests that i64 division by zero throws a catchable org.eolang.error,
+  # not an uncatchable JVM exception.
+  [] +> tests-i64-div-by-zero-throws-catchable-error
+    try > @
+      seq
+        *
+          2.as-i64.div 0.as-i64
+          false
+      true > [e]
+      false
+
   # Tests that i64 division by one returns the original value.
   [] +> tests-i64-div-by-i64-one
     eq. > @
--- a/eo-runtime/src/main/java/org/eolang/EOi64$EOdiv.java
+++ b/eo-runtime/src/main/java/org/eolang/EOi64$EOdiv.java
@@ -26,16 +26,19 @@ public final class EOi64$EOdiv extends PhDefault implements Atom {
     }
 
     @Override
+    @SuppressWarnings("PMD.UnnecessaryLocalRule")
     public Phi lambda() {
+        final long dividend = new Expect.I64(Expect.at(this, Phi.RHO)).it();
+        final long divisor = new Expect.I64(Expect.at(this, "x")).it();
+        if (divisor == 0L) {
+            throw new EOerror.ExError(
+                new Data.ToPhi("i64.div: division by zero")
+            );
+        }
         final Phi num = Phi.Φ.take("i64").copy();
         num.put(
             0,
-            new Data.ToPhi(
-                new BytesOf(
-                    new Expect.I64(Expect.at(this, Phi.RHO)).it()
-                        / new Expect.I64(Expect.at(this, "x")).it()
-                ).take()
-            )
+            new Data.ToPhi(new BytesOf(dividend / divisor).take())
         );
         return num;
     }


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+213](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5287, opened 2026-07-01
#5255: tighten i64.as-i32 range guard to actual i32 bounds

### LINKED ISSUE #5255: i64 to i32 conversion wraps out-of-range values silently instead of throwing error

## Summary
The i64.as-i32 conversion guards range only by checking whether the high 4 bytes of the i64 equal 00-00-00-00 (positive) or FF-FF-FF-FF (negative). This guard accepts any value whose magnitude fits in 32 bits, not 31. Values in [2^31, 2^32) have high bytes 00-00-00-00 and values in [-2^32, -2^31) have high bytes FF-FF-FF-FF, so they pass the guard, and `i32 (as-bytes.slice 4 4)` then reinterprets the low 4 bytes as a signed i32, silently producing a wrong (sign-flipped) result instead of the documented error. Example: 3000000000.as-i64.as-i32 yields -1294967296; 2147483648.as-i64.as-i32 (i32 max + 1) yields -2147483648; -2147483649.as-i64.as-i32 yields 2147483647. number.as-i32 is defined as `as-i64.as-i32`, so number values in these ranges are silently corrupted too.

## Location
[`eo-runtime/src/main/eo/i64.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/i64.eo#L19) lines 19-32.

## Buggy code
```java
  [] > as-i32
    if. > @
      or.
        left.eq 00-00-00-00
        left.eq FF-FF-FF-FF
      i32 (as-bytes.slice 4 4)
      error
        sprintf
          "Can't convert i64 number %d to i32 because it's out of i32 bounds"
          * as-number
    (as-bytes.slice 0 4).as-bytes > left
```

## Why it's wrong
The Javadoc directly above the object states: "The org.eolang.error is returned if the org.eolang.i64 number is more than max org.eolang.i32 value 2147483647." The implementation does NOT return an error for 2147483648..4294967295 (and the symmetric negative range); it returns a wrapped-around value. Reproduced in plain Java: for v=2147483648 the high 32 bits are 0x00000000 so the guard passes, and the low 32 bits reinterpreted as signed int = -2147483648. Correct i32 range is [-2^31, 2^31-1], so these inputs must error.

Verifier confirmation: Confirmed genuine. The code at eo-runtime/src/main/eo/i64.eo:22-32 guards the i64->i32 conversion by checking only that the high 4 bytes equal 00-00-00-00 or FF-FF-FF-FF, then returns i32(low 4 bytes). I verified the encoding is big-endian (BytesOf(long) uses ByteBuffer.putLong, default big-endian, so slice 0 4 = high bytes, slice 4 4 = low bytes) and that i32 reinterprets those 4 bytes as a signed int. This guard accepts magnitudes that fit in 32 bits rather than the correct 31-bit i32 range [-2^31, 2^31-1]. I reproduced numerically: 3000000000 (hi=0x00000000, passes) -> -1294967296; 2147483648 (i32max+1, hi=0x00000000, passes) -> -2147483648; -2147483649 (hi=0xFFFFFFFF, passes) -> 2147483647; 4294967295 -> -1. All are out of i32 range yet silently produce sign-flipped/wrapped values instead of the org.eolang.error that the Javadoc on lines 19-21 explicitly promises. number.as-i32 delegates to as-i64.as-i32, so number values in these ranges are corrupted too. The implementation directly contradicts its own documented contract. (The same flawed pattern recurs in as-i16, but the claim is correctly scoped to as-i32.)

## Expected behavior
Out-of-i32-range i64 values (anything outside -2147483648..2147483647) should return the org.eolang.error produced by the else branch, as the Javadoc promises.

## Suggested fix
Tighten the guard so it only accepts values whose 64-bit value actually fits in signed i32. E.g. accept only when high4==00-00-00-00 AND the sign bit of the low word is 0 (value < 2^31), OR high4==FF-FF-FF-FF AND the sign bit of the low word is 1 (value >= -2^31). Equivalently, verify that sign-extending the low 32 bits reproduces the original i64 (low.as-i64.eq self) before slicing.

morphqdd:
## What happens

`i64.as-i32` guards range only by checking whether the high 4 bytes of the
i64 equal `00-00-00-00` (positive) or `FF-FF-FF-FF` (negative). This guard
accepts any value whose magnitude fits in 32 bits, not 31. Values in
`[2^31, 2^32)` have high bytes `00-00-00-00` and values in `[-2^32, -2^31)`
have high bytes `FF-FF-FF-FF`, so they pass the guard, and
`i32 (as-bytes.slice 4 4)` reinterprets the low 4 bytes as a signed i32,
silently producing a wrong (sign-flipped) result instead of the documented
error.

Example: `3000000000.as-i64.as-i32` yields `-1294967296`;
`2147483648.as-i64.as-i32` (i32 max + 1) yields `-2147483648`;
`-2147483649.as-i64.as-i32` yields `2147483647`. `number.as-i32` is defined
as `as-i64.as-i32`, so `number` values in these ranges are silently
corrupted too.

The Javadoc directly above the object states: "The `org.eolang.error` is
returned if the `org.eolang.i64` number is more than max `org.eolang.i32`
value `2147483647`." The implementation does not honor that contract.

## What should happen

Out-of-i32-range i64 values (anything outside `-2147483648..2147483647`)
should return the `org.eolang.error` the else branch already produces.

## Fix

Replaced the magnitude-only guard with a round-trip check: convert the low
4 bytes to i32, sign-extend that back to i64 via `.as-i64`, and compare
against the original i64 value with `^.eq`. If they match, the conversion
is lossless and safe; if not, the value is out of i32 range and the
existing error path fires.

```diff
   [] > as-i32
     if. > @
-      or.
-        left.eq 00-00-00-00
-        left.eq FF-FF-FF-FF
-      i32 (as-bytes.slice 4 4)
+      ^.eq left.as-i64
+      left
       error
         sprintf
           "Can't convert i64 number %d to i32 because it's out of i32 bounds"
           * as-number
-    (as-bytes.slice 0 4).as-bytes > left
+    i32 (as-bytes.slice 4 4) > left
```

## Test

Added four regression tests exercising the exact boundary the bug report
demonstrated:

- `tests-i64-i32-max-converts-without-error` — i32 max (`2147483647`)
  still converts without error
- `tests-i64-i32-min-converts-without-error` — i32 min (`-2147483648`)
  still converts without error
- `throws-on-converting-i32-max-plus-one-to-i32` — i32 max + 1
  (`2147483648`) now throws instead of silently wrapping
- `throws-on-converting-i32-min-minus-one-to-i32` — i32 min - 1
  (`-2147483649`) now throws instead of silently wrapping

The existing `throws-on-converting-to-i32-if-out-of-bounds` (using
`3147483647`) and the two round-trip tests (`234`, `-234`) still pass
unchanged.

## Verification

- `mvn test -pl eo-runtime` — `EOi64Test`: 32/32, `EOi32Test`: 30/30, full
  module green (only pre-existing environment-dependent skips: win32,
  socket, io)
- `mvn qulice:check -Pqulice -pl eo-runtime` — no violations attributed to
  `i64.eo`; the 100 pre-existing violations in the module are the same
  baseline seen on unrelated branches

Closes #5255

### DIFF
--- a/eo-runtime/src/main/eo/i64.eo
+++ b/eo-runtime/src/main/eo/i64.eo
@@ -21,15 +21,13 @@
   # max `org.eolang.i32` value `2147483647`.
   [] > as-i32
     if. > @
-      or.
-        left.eq 00-00-00-00
-        left.eq FF-FF-FF-FF
-      i32 (as-bytes.slice 4 4)
+      ^.eq left.as-i64
+      left
       error
         sprintf
           "Can't convert i64 number %d to i32 because it's out of i32 bounds"
           * as-number
-    (as-bytes.slice 0 4).as-bytes > left
+    i32 (as-bytes.slice 4 4) > left
 
   # Convert this `org.eolang.i64` to `org.eolang.number` object and return it.
   [] > as-number /number
@@ -108,6 +106,24 @@
 
   # Tests that converting i64 to i32 throws error when out of bounds.
   3147483647.as-i64.as-i32 > [] +> throws-on-converting-to-i32-if-out-of-bounds
+
+  # Tests that i32 max value converts to i32 without error.
+  [] +> tests-i64-i32-max-converts-without-error
+    eq. > @
+      2147483647.as-i64
+      2147483647.as-i64.as-i32.as-i64
+
+  # Tests that i32 min value converts to i32 without error.
+  [] +> tests-i64-i32-min-converts-without-error
+    eq. > @
+      -2147483648.as-i64
+      -2147483648.as-i64.as-i32.as-i64
+
+  # Tests that i32 max plus one throws when converted to i32.
+  2147483648.as-i64.as-i32 > [] +> throws-on-converting-i32-max-plus-one-to-i32
+
+  # Tests that i32 min minus one throws when converted to i32.
+  -2147483649.as-i64.as-i32 > [] +> throws-on-converting-i32-min-minus-one-to-i32
 
   # Tests that i64 less-than comparison returns true for smaller values.
   [] +> tests-i64-less-true


morphqdd:
@yegor256 take a look please

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+241](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5288, opened 2026-07-01
Migrate atoms from throw-based failures to the ⊥/fallback model

### LINKED ISSUE #5285: atoms' failures need to adopt the ⊥/fallback model instead of current exception-based approach
Part of #5261. Most atoms still report a predictable failure by throwing an exception that no EO `try` can catch. The filesystem atoms are the clearest case: `file.open.file-stream.read.chunk` catches an `IOException` and rethrows it as an `IllegalArgumentException`, so a missing or unreadable file aborts the whole program instead of handing the caller a value it can react to.

```java
    throw new IllegalArgumentException(
        String.format("Can't read from %s", path),
        ex
    );
}
```

#5280 prototyped a better shape on `malloc.of.allocated.read`. A predictable failure routes to a caller-supplied error-branch void (`? > cant-read /{...}`) and yields the bottom object `⊥` when the caller bound nothing, so the failure stays dormant and detonates only when the result is forced by dataization. The decision is made with a pre-check, not a caught exception, which keeps control flow off the exception path.

This ticket carries that model across the remaining atoms, following the two categories established in #5280. Predictable failures — a missing file, a malformed regex, an out-of-range read, a bad format string — become error-branch fallbacks that surface `⊥`; system-level failures, such as the raw syscalls in `EOposix` and `EOwin32`, stay as uncatchable `ExFailure` panics. In scope: `EOfs/*` (read, write, move, delete, touch, dir walk and tmpfile), `EOtt/EOregex*`, `EOtt/EOsscanf` with `SprintfArgs`, and the remaining `malloc` bounds checks. Out of scope: the `EOsm/*` syscalls and the error machinery itself, `EOtry` and `EOerror`.

Each migrated atom gains an error-branch void in its `.eo` source, drops the `try/catch` that turned a predictable condition into an `ExFailure`, and chooses between the value and the fallback with an explicit pre-check. The parser spec and runtime docs should record the atoms that now expose a fallback so callers know which reads can hand back `⊥`.

maxonfjvipon:
Part of #5261. #5280 prototyped the termination-plus-fallback model on `malloc.of.allocated.read`: a predictable failure routes to a caller-supplied recovery object and yields the bottom object (⊥) when none was bound, so the failure surfaces only when the result is forced. This PR carries that model across the remaining atoms, one at a time. It is a draft so the approach can be reviewed on a concrete atom before the pattern spreads.

The first atom is `regex.@`. A `expression` that cannot become a pattern used to throw an uncatchable exception; now it hands back the `cant-compile` recovery declared on the user-facing `regex`, or ⊥ when none was provided. Because `regex.@` is internal machinery the caller never touches, the recovery void lives on `regex` and the atom reads it through `Phi.RHO`.

```java
if (expression.startsWith("/")) {
    if (last == 0) {
        result = this.fallback("regex is missing the closing slash");
    } else {
        result = this.compile(expression, last);
    }
} else {
    result = this.fallback("regex is missing the opening slash");
}
```

An invalid pattern body is discoverable only by compiling it, so the `PatternSyntaxException` is caught at that boundary and routed to the fallback rather than driving control flow; a serialization failure on the in-memory stream stays a system panic. Still to come: the filesystem atoms (`io.file-error` group) and `sscanf`/`sprintf`.

Closes #5285

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+
+---
+exclude_paths:
+  - "eo-runtime/src/main/java/org/eolang/EOtt/EOregex$EOcompile.java"
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -161,16 +161,7 @@ final class Emissions {
         } else if (value.kind() == Value.Kind.TERM) {
             emit.object(name, "⊥", line, value.pos());
         } else if (value.kind() == Value.Kind.GROUP) {
-            final String inner = value.raw().substring(1, value.raw().length() - 1);
-            final int phi = Emissions.topLevelInlinePhi(inner);
-            if (phi >= 0) {
-                Emissions.inlinePhi(emit, name, inner, phi, value.pos() + 1, line);
-            } else {
-                final Span sub = new Span(
-                    " ".repeat(value.pos() + 1).concat(inner), line
-                );
-                Emissions.expression(emit, name, new Tokens(sub.body(), sub), line);
-            }
+            Emissions.group(emit, name, value, line);
         } else {
             emit.object(name, value.raw(), line, value.pos());
         }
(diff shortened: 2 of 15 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (612 >= 100); -8 for way too many hits-of-code (612 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+504](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5289, opened 2026-07-01
#5259: raise positioned ParseError for oversized HEX literals

### LINKED ISSUE #5259: Uncaught NumberFormatException thrown for oversized HEX literal

## Summary
A HEX literal is emitted via `Long.parseLong(value.raw().substring(2), 16)`. The tokenizer (Tokens.readHex, lines 367-393) reads an unbounded run of hex digits with no length/range cap, so a literal like `0x10000000000000000` (17 hex digits) tokenizes as a valid HEX value. Long.parseLong then overflows and throws NumberFormatException. As with the unicode case, this occurs during emission inside Line.into, which the walker only guards against ParseError, so the exception propagates and aborts the entire parse.

## Location
[`eo-parser/src/main/java/org/eolang/parser/Emissions.java`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/Emissions.java#L118) lines 118-124.

## Buggy code
```java
} else if (value.kind() == Value.Kind.HEX) {
    emit.object(name, "Φ.number", line, value.pos());
    Emissions.bytesCarrier(
        emit, line, value.pos(),
        new Hex((double) Long.parseLong(value.raw().substring(2), 16))
            .asString()
    );
```

## Why it's wrong
readHex imposes no upper bound on the number of hex digits, so a HEX value that the tokenizer accepts cannot always be parsed by Long.parseLong. The result is an uncaught RuntimeException that crashes parsing instead of a positioned parse error. Reproduced: `Long.parseLong("10000000000000000", 16)` throws `NumberFormatException`.

Verifier confirmation: Independently confirmed in the cloned repo. Emissions.java:122 emits a HEX literal via Long.parseLong(value.raw().substring(2), 16). Tokens.readHex (367-393) consumes an unbounded run of hexDigit() characters with no length or numeric-range cap, so a literal like 0x10000000000000000 tokenizes as a valid HEX Value. Long.parseLong on the radix-16 string then throws NumberFormatException once the value exceeds Long.MAX_VALUE — I reproduced this (Long.parseLong("10000000000000000",16) throws NumberFormatException; in fact even ffffffffffffffff throws since parseLong is signed, so the real overflow threshold is 0x8000000000000000, lower than the finder stated). openValue is called synchronously during Line.into(), which runs inside Eo.dispatch (and continueTextBlock); all three try/catch blocks in Eo.java (lines 231, 295, 619) catch only ParseError, so the NumberFormatException is an uncaught RuntimeException that propagates and aborts the whole parse instead of producing a positioned ParseError. Notably the adjacent FLOAT path (Double.parseDouble, line 116) does not share this defect because parseDouble saturates to Infinity rather than throwing, confirming the bug is specific to the HEX branch and not handled elsewhere.

## Expected behavior
An out-of-range HEX literal should produce a positioned ParseError (e.g. "hexadecimal literal out of range") that the walker catches and reports, not an uncaught NumberFormatException aborting the parse.

## Suggested fix
Use Long.parseUnsignedLong or catch NumberFormatException around the parse and raise a ParseError at value.pos(); or enforce a maximum digit count / range check in Tokens.readHex so the tokenizer rejects over-long literals with a positioned error.

morphqdd:
## What happens

A HEX literal is emitted via `Long.parseLong(value.raw().substring(2), 16)`
in `Emissions.openValue`. The tokenizer (`Tokens.readHex`) reads an
unbounded run of hex digits with no length/range cap, so a literal like
`0x10000000000000000` (17 hex digits) tokenizes as a valid HEX value.
`Long.parseLong` then overflows and throws `NumberFormatException`. This
happens during emission, which the walker only guards against
`ParseError`, so the exception propagates and aborts the entire parse
instead of producing a positioned error for just that line.

Reproduced: `Long.parseLong("10000000000000000", 16)` throws
`NumberFormatException`.

## What should happen

An out-of-range HEX literal should produce a positioned `ParseError`
(e.g. "hexadecimal literal is out of range") that the walker catches and
reports, not an uncaught `NumberFormatException` aborting the parse.

## Fix

Wrapped the `Long.parseLong` call in a try/catch. On
`NumberFormatException`, raise a `ParseError` at `value.pos()` with the
line number, preserving the original exception as the cause via
`initCause` (required by PMD's `PreserveStackTrace` rule, confirmed by
`qulice`).

## Test

Added `rejectsOversizedHexHead` to `LnApplicationTest.java`, feeding
`0x10000000000000000 > x` through `LnApplication` and asserting a
`ParseError` is thrown instead of letting the `NumberFormatException`
escape. The existing `emitsHexHead` test (`0x1F > x`) continues to pass
unchanged, confirming normal HEX literals are unaffected.

## Verification

- `mvn test -pl eo-parser` — full module green, `LnApplicationTest`: 44/44
- `mvn qulice:check -Pqulice -pl eo-parser` — **BUILD SUCCESS**, 0
  violations (an intermediate attempt without `initCause` correctly
  triggered PMD's `PreserveStackTrace` rule, fixed by preserving the
  caught exception as the cause)

Closes #5259

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -116,11 +116,21 @@ final class Emissions {
                 new Hex(Double.parseDouble(value.raw())).asString()
             );
         } else if (value.kind() == Value.Kind.HEX) {
+            final long hex;
+            try {
+                hex = Long.parseLong(value.raw().substring(2), 16);
+            } catch (final NumberFormatException ex) {
+                final ParseError error = new ParseError(
+                    line, value.pos(),
+                    "hexadecimal literal is out of range"
+                );
+                error.initCause(ex);
+                throw error;
+            }
             emit.object(name, "Φ.number", line, value.pos());
             Emissions.bytesCarrier(
                 emit, line, value.pos(),
-                new Hex((double) Long.parseLong(value.raw().substring(2), 16))
-                    .asString()
+                new Hex((double) hex).asString()
             );
         } else if (value.kind() == Value.Kind.BYTES) {
             emit.object(name, "Φ.bytes", line, value.pos());
--- a/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
@@ -556,6 +556,17 @@ final class LnApplicationTest {
     }
 
     @Test
+    void rejectsOversizedHexHead() {
+        Assertions.assertThrows(
+            ParseError.class,
+            () -> new LnApplication(new Span("0x10000000000000000 > x", 1))
+                .into(new Stack(), new Globals(), new Emit()),
+            "a HEX literal wider than a signed 64-bit long must raise a positioned"
+                .concat(" ParseError instead of an uncaught NumberFormatException")
+        );
+    }
+
+    @Test
     void emitsBytesHead() {
         final Emit emit = new Emit();
         new LnApplication(new Span("CA-FE > x", 1))


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+253](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5293, opened 2026-07-01
#5257: throw ParseError on invalid unicode escape in string literal

### LINKED ISSUE #5257: Parser crashes due to uncaught NumberFormatException from invalid \uXXXX string literal escape

## Summary
appendUnicode decodes a \uXXXX escape by calling Integer.parseInt(body.substring(cursor, cursor + 4), 16) whenever there are at least 4 characters after the 'u' run. It never validates that those 4 characters are hex digits. The string tokenizer (Tokens.readString, lines 423-462) accepts ANY backslash escape without validating the following characters, so a literal such as `"\uZZZZ"` tokenizes fine and reaches unescapeBody during emission. parseInt then throws NumberFormatException. Emission happens inside Line.into, which the walker wraps in a try that catches only ParseError (Eo.dispatch lines 293-299; Eo.continueTextBlock 228-234), so the NumberFormatException propagates out and aborts the whole parse instead of producing a positioned error.

## Location
[`eo-parser/src/main/java/org/eolang/parser/Emissions.java`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/Emissions.java#L342) lines 342-357.

## Buggy code
```java
private static int appendUnicode(
    final StringBuilder out, final String body, final int start
) {
    int cursor = start;
    while (cursor < body.length() && body.charAt(cursor) == 'u') {
        cursor = cursor + 1;
    }
    if (cursor + 4 > body.length()) {
        out.append('\\').append(body, start, body.length());
    } else {
        out.append(
            (char) Integer.parseInt(body.substring(cursor, cursor + 4), 16)
        );
    }
    return cursor + 4;
}
```

## Why it's wrong
The walker is documented to recover from rule violations by catching ParseError and emitting a positioned error (ParseError Javadoc; Eo lines 231/295/619). A malformed unicode escape is a string-content rule violation (R-9.7) that should surface as a positioned parse error, not an unchecked RuntimeException that kills parsing with an opaque message. Reproduced: unescapeBody("\uZZZZ") throws `NumberFormatException: For input string: "ZZZZ" under radix 16`.

Verifier confirmation: Independently confirmed. Tokens.readString (lines 440-447) consumes a backslash plus the following character with zero hex validation, so a literal like "\uZZZZ" tokenizes cleanly as a STRING Value. During emission, Emissions.openValue's STRING branch (line 134) eagerly evaluates Emissions.unescape(value.raw()) as a synchronous argument, which routes through unescapeBody (line 264) to appendUnicode (line 353): Integer.parseInt(body.substring(cursor, cursor+4), 16) with no check that those four chars are hex digits. I reproduced the exact failure in isolation: both "\uZZZZ" and "\u12GG" throw NumberFormatException: For input string under radix 16. This emission runs inside Eo.dispatch's try (line 294) and Eo.continueTextBlock's try (line 230), both of which catch ONLY ParseError (lines 295, 231); a NumberFormatException is a RuntimeException and therefore propagates out, aborting the entire parse instead of producing the positioned R-9.7 error the walker is contractually documented to emit. No upstream validation of unicode-escape hex digits exists anywhere in the parser package. The only deviation from the report is cosmetic: the checkout lived in the scratchpad clone, not /workspace/eo (which was absent); the code at the cited line numbers matches the claim verbatim.

## Expected behavior
A malformed \uXXXX escape (non-hex digits) should raise a positioned ParseError (e.g. "invalid unicode escape") that the walker can catch and report, not an uncaught NumberFormatException that aborts the entire parse.

## Suggested fix
Validate the 4 characters are hex before parsing (e.g. wrap parseInt in a check, or catch NumberFormatException) and emit/throw a ParseError with the line/column of the escape; alternatively validate escapes in Tokens.readString so the failure is a ParseError at tokenization time.

morphqdd:
## What happens

`Emissions.appendUnicode` decodes a `\uXXXX` escape by calling
`Integer.parseInt(body.substring(cursor, cursor + 4), 16)` whenever there
are at least 4 characters after the `u` run, without validating that
those 4 characters are actually hex digits. `Tokens.readString` accepts
any backslash escape without checking the following characters, so a
literal such as `"\uZZZZ"` tokenizes cleanly and reaches `unescapeBody`
during emission, where `parseInt` throws an uncaught
`NumberFormatException`. `Emissions.openValue`'s `STRING` branch is called
from inside the walker's per-line dispatch, which only catches
`ParseError` — so the `NumberFormatException` propagates out and aborts
the entire parse instead of producing a positioned error, the same class
of defect already fixed for oversized `HEX` literals in #5287.

## What should happen

A malformed `\uXXXX` escape should raise a positioned `ParseError`
("invalid unicode escape in string literal") that the walker can catch
and report, not an uncaught `NumberFormatException` that kills the whole
parse.

## Fix

`Emissions.openValue`'s `STRING` branch now catches the
`NumberFormatException` from `Emissions.unescape` and rethrows it as a
`ParseError` carrying the value's line/position, preserving the original
cause via `initCause`:

```diff
 } else if (value.kind() == Value.Kind.STRING) {
     emit.object(name, "Φ.string", line, value.pos());
+    final String unescaped;
+    try {
+        unescaped = Emissions.unescape(value.raw());
+    } catch (final NumberFormatException ex) {
+        final ParseError error = new ParseError(
+            line, value.pos(), "invalid unicode escape in string literal"
+        );
+        error.initCause(ex);
+        throw error;
+    }
     Emissions.bytesCarrier(
         emit, line, value.pos(),
-        new Hex(Emissions.unescape(value.raw())).asString()
+        new Hex(unescaped).asString()
     );
```

Added `rejectsStringWithInvalidUnicodeEscape` to `LnApplicationTest`,
which reproduces the exact reported stack trace on unmodified master
(`NumberFormatException: For input string: "ZZZZ" under radix 16`) and
asserts a `ParseError` is thrown instead.

## Verification

`mvn -pl eo-parser test -Dtest=LnApplicationTest` — 23 tests pass,
including the new one. Confirmed the new test fails with the reported
`NumberFormatException` on unmodified master and passes with this fix.

Closes #5257

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -129,9 +129,19 @@ final class Emissions {
             emit.close();
         } else if (value.kind() == Value.Kind.STRING) {
             emit.object(name, "Φ.string", line, value.pos());
+            final String unescaped;
+            try {
+                unescaped = Emissions.unescape(value.raw());
+            } catch (final NumberFormatException ex) {
+                final ParseError error = new ParseError(
+                    line, value.pos(), "invalid unicode escape in string literal"
+                );
+                error.initCause(ex);
+                throw error;
+            }
             Emissions.bytesCarrier(
                 emit, line, value.pos(),
-                new Hex(Emissions.unescape(value.raw())).asString()
+                new Hex(unescaped).asString()
             );
         } else if (value.kind() == Value.Kind.STAR) {
             emit.object(name, "Φ.tuple", line, value.pos());
--- a/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
@@ -437,6 +437,17 @@ final class LnApplicationTest {
     }
 
     @Test
+    void rejectsStringWithInvalidUnicodeEscape() {
+        Assertions.assertThrows(
+            ParseError.class,
+            () -> new LnApplication(new Span("\"\\uZZZZ\" > x", 1))
+                .into(new Stack(), new Globals(), new Emit()),
+            "a string with a non-hex \\u escape must be rejected instead of"
+                + " crashing with an uncaught NumberFormatException"
+        );
+    }
+
+    @Test
     void emitsRootIdentifierAsHead() {
         final Emit emit = new Emit();
         new LnApplication(new Span("Q > x", 1))
--- a/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
@@ -442,8 +442,7 @@ final class LnApplicationTest {
             ParseError.class,
             () -> new LnApplication(new Span("\"\\uZZZZ\" > x", 1))
                 .into(new Stack(), new Globals(), new Emit()),
-            "a string with a non-hex \\u escape must be rejected instead of"
-                + " crashing with an uncaught NumberFormatException"
+            "a string with a non-hex \\u escape must be rejected, not crash"
         );
     }
 


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+277](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5300, opened 2026-07-01
fix: handle Integer.MIN_VALUE in shift method to prevent overflow

someshk1703:
The PR fixes the issue, BytesRaw.shift results in ArrayIndexOutOfBoundsException for Integer.MIN_VALUE shift value due to Math.abs overflow. 

Fix involves:
replaced the two raw Math.abs() calls with a single safe value:
final int abs = bits == Integer.MIN_VALUE ? Integer.MAX_VALUE : Math.abs(bits);

When bits == Integer.MIN_VALUE, Integer.MAX_VALUE is used instead. he resulting byte offset (268,435,455). 
exceeds any realistic byte array length, so shiftLeft correctly zeroes all bytes — the same semantically correct result as shifting by 2³¹ positions.

A regression test shiftsIntegerMinValueWithoutOverflow was added to BytesOfTest test class to cover this path.


yegor256:
@someshk1703 fix all CI builds

gemshrine (commented):
The `bits == Integer.MIN_VALUE ? Integer.MAX_VALUE : Math.abs(bits)` substitution avoids the overflow but silently changes the semantics of the shift by one bit. `|Integer.MIN_VALUE| = 2^31 = 2147483648`, which doesn't fit in an `int` — but `Integer.MAX_VALUE` is `2^31 - 1 = 2147483647`, so `mod` and `offset` come out off-by-one:

```
PR fix:      abs=2147483647 mod=7 offset=268435455
Truth:       abs=2147483648 mod=0 offset=268435456
```

The added test uses a 64-bit input (`0xFF000000L`), where both shifts drop everything to zero regardless of that 1-bit gap, so the test can't distinguish the two behaviours. A larger `BytesOf` (say, a couple of MiB, so that `~2^31` bit shifts don't wipe the whole buffer) would surface the discrepancy.

A shape that stays exact and doesn't need the special case: take mod/div first, then abs — both stay in `int` range even for `Integer.MIN_VALUE`, because `MIN_VALUE / 8 = -2^28` and `MIN_VALUE % 8 = 0`:

```java
final int mod = Math.abs(bits % Byte.SIZE);
final int offset = Math.abs(bits / Byte.SIZE);
```

Alternatively, promote once through `long`: `final long abs = Math.abs((long) bits);` and cast the derived `mod`/`offset` back to `int`. Either avoids losing the boundary bit.

yegor256:
@someshk1703 Thanks for your contribution, we appreciate it! Before we can merge this pull request, we need to see all CI workflows pass without errors. We can't merge the pull request into the master branch unless all workflows are green. Try to fix them, also paying attention to code style issues.

yegor256:
@someshk1703 build is broken here

eo-runtime/src/main/java/org/eolang/BytesRaw.java
```diff
@@ -70,8 +70,8 @@ public Bytes xor(final Bytes other) {
     @Override
     public Bytes shift(final int bits) {
         final byte[] bytes = this.take();
-        final int mod = Math.abs(bits) % Byte.SIZE;
-        final int offset = Math.abs(bits) / Byte.SIZE;
+        final int mod = Math.abs(bits % Byte.SIZE);
```
morphqdd, line 73: The PR description says the fix is `bits == Integer.MIN_VALUE ? Integer.MAX_VALUE : Math.abs(bits)`, but this is what actually landed (`Math.abs(bits % Byte.SIZE)` / `Math.abs(bits / Byte.SIZE)`). The code is correct, the description just doesn't match it. Worth updating so reviewers aren't reading about a fix that isn't there.

eo-parser/src/main/java/org/eolang/parser/Emissions.java
```diff
@@ -331,6 +310,69 @@ private static String rootBase(final String raw) {
...
 
+    /**
+     * Parse the raw hex string of a {@link Value} into a {@code long},
+     * wrapping any {@link NumberFormatException} into a positioned
+     * {@link ParseError}.
+     * @param value The value whose raw text is a hex literal (e.g. {@code 0xFF})
+     * @param line Source line number
+     * @return Parsed long value
+     */
+    private static long parseHex(final Value value, final int line) {
```
morphqdd, line 321: This PR is about a BytesRaw.shift overflow fix, but this hunk pulls in an unrelated refactor of Emissions (parseHex/unescapeChecked/openValueGroup extraction). Looks like the branch wasn't rebased and is carrying changes from somewhere else.

eo-parser/src/main/java/org/eolang/parser/Emissions.java
```diff
@@ -331,6 +310,69 @@ private static String rootBase(final String raw) {
...
+    /**
+     * Handle a GROUP value: extract the inner expression and emit it
+     * either as an inline phi or as a nested expression.
+     * @param emit Emitter
+     * @param name Name attribute (or {@code null})
+     * @param value The GROUP value
+     * @param line Source line number
+     * @checkstyle ParameterNumberCheck (3 lines)
+     */
+    private static void openValueGroup(
```
morphqdd, line 342: openValueGroup has no callers, the GROUP branch still goes through the existing Emissions.group(...). Dead code left over from whatever this got merged or rebased from.

morphqdd (changes requested):

yegor256:
Please forgive us for closing this pull request. It has had merge conflicts with `master` for over a month now. Feel free to submit a new PR with the same fix, rebased on the current `master`, taking into account changes made there since this PR was opened. We'd welcome it!

0crat:
@gemshrine Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+1336](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+3184](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5307, opened 2026-07-02
#5295: reject out-of-int-range numbers in Expect.Int and Expect.Natural

### LINKED ISSUE #5295: `Expect.Natural` and `Expect.Int` wrongly allocate blocks due to `Double.intValue()` saturation for large numbers
## What happens

`Expect.Natural.it()` (`eo-runtime/src/main/java/org/eolang/Expect.java`) validates that an EO
number is a whole number greater than or equal to zero, then narrows it to `Integer` with
`Double::intValue`:

```java
public Integer it() {
    return this.expect
        .must(number -> number % 1 == 0)
        .otherwise("must be an integer")
        .that(Double::intValue)
        .must(integer -> integer >= 0)
        .otherwise("must be greater or equal to zero")
        .it();
}
```

`Double.intValue()` does not throw for a value outside `int` range — per its Javadoc, it
**saturates**, returning `Integer.MAX_VALUE` for any double at or above `2147483648.0`. None of the
three `must`/`otherwise` checks above catch this: `number % 1 == 0` is still true for
`1_000_000_000_000_000.0`, and the post-narrowing `integer >= 0` check sees the already-clamped
`2147483647`, which is of course `>= 0`.

Verified directly:

```
huge=1.0E15
huge % 1 == 0: true
clamped intValue()=2147483647
clamped >= 0: true
```

`Expect.Natural` backs the `size` argument of `malloc.of` (`EOmalloc$EOof$EOφ.java`), which passes
the result straight to `Heaps.malloc(Phi, int)`, which does `new byte[size]`
(`Heaps.java:65`). So `malloc.of 1000000000000000` — an EO natural number far larger than
`Integer.MAX_VALUE` — does not raise the "must be greater or equal to zero"-style domain error
this validation pipeline is meant to produce; instead it silently allocates a
`2147483647`-byte block, a completely different size than what the program asked for. Depending on
the JVM's available heap, this either succeeds (silently allocating the wrong amount) or fails with
an unrelated `OutOfMemoryError` far removed from the actual cause.

## What should happen

`Expect.Natural` (and any other user of `Double::intValue` for narrowing an EO number to `int`,
e.g. `Expect.Int`) should reject values outside `int` range with the same kind of domain error it
already raises for negative or fractional values, instead of silently saturating.

morphqdd:
Closes #5295.

`Expect.Int.it()` and `Expect.Natural.it()` narrow a `Double` to `Integer` via `Double::intValue`
without first checking the value fits in `int` range. `Double.intValue()` doesn't throw for a
value outside `int` range — it saturates, returning `Integer.MAX_VALUE`/`Integer.MIN_VALUE`. Since
the saturated value is still a "valid" integer (and, for `Natural`, still `>= 0`), no downstream
check catches it, e.g. `malloc.of 1000000000000000` silently allocates a `2147483647`-byte block
instead of raising a domain error for a size nowhere near what was requested.

Added a `must` step before the `Double::intValue` narrowing in both `Expect.Int` and
`Expect.Natural` that rejects any value outside `[Integer.MIN_VALUE, Integer.MAX_VALUE]` with a
`"must fit into int range"` message, matching the existing style of this validation pipeline
(`must be an integer`, `must be greater or equal to zero`).

Added two tests in `ExpectTest` (`failsInTransformingToIntegerForTooLarge`,
`failsInTransformingToNonNegativeIntegerForTooLarge`) mirroring the existing
`failsInTransformingTo...ForNotInteger`/`ForNegative` tests, using `1.0e15` as the out-of-range
value from the issue's reproduction.

Ran the full `Expect`-related test suite locally (`ExpectTest`, `EOmallocAllocatedExpectTest`,
`EOmallocWritePhiExpectTest`, `EOiConvertersExpectTest`, `EObytesEOrightExpectTest`,
`EOi64ExpectTest`) — all 36 tests pass.

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/ExpectTest.java
+++ b/eo-runtime/src/test/java/org/eolang/ExpectTest.java
@@ -273,4 +273,48 @@ final class ExpectTest {
             Matchers.equalTo("the 'ρ' attribute (-42) must be greater or equal to zero")
         );
     }
+
+    @Test
+    void failsInTransformingToIntegerForTooLarge() {
+        MatcherAssert.assertThat(
+            "inner class Integer throws error for a number outside int range, instead of saturating",
+            Assertions.assertThrows(
+                ExFailure.class,
+                () -> new Expect.Int(
+                    Expect.at(
+                        new PhApplication(
+                            new PhDefault(),
+                            Phi.RHO,
+                            new Data.ToPhi(1.0e15)
+                        ),
+                        Phi.RHO
+                    )
+                ).it(),
+                "fails with correct error message while transform Phi to Integer"
+            ).getMessage(),
+            Matchers.equalTo("the 'ρ' attribute (1.0E15) must fit into int range")
+        );
+    }
+
+    @Test
+    void failsInTransformingToNonNegativeIntegerForTooLarge() {
+        MatcherAssert.assertThat(
+            "inner class NonNegativeInteger throws error for a number outside int range, instead of saturating",
+            Assertions.assertThrows(
+                ExFailure.class,
+                () -> new Expect.Natural(
+                    Expect.at(
+                        new PhApplication(
+                            new PhDefault(),
+                            Phi.RHO,
+                            new Data.ToPhi(1.0e15)
+                        ),
+                        Phi.RHO
+                    )
+                ).it(),
+                "fails with correct error message while transform Phi to NonNegativeInteger"
+            ).getMessage(),
+            Matchers.equalTo("the 'ρ' attribute (1.0E15) must fit into int range")
+        );
+    }
 }
--- a/eo-runtime/src/main/java/org/eolang/Expect.java
+++ b/eo-runtime/src/main/java/org/eolang/Expect.java
@@ -262,6 +262,8 @@ public class Expect<T> {
                 .otherwise("must be a number")
                 .must(number -> number % 1 == 0)
                 .otherwise("must be an integer")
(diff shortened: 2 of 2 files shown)


morphqdd:

morphqdd:

eo-runtime/src/test/java/org/eolang/ExpectTest.java
```diff
@@ -273,4 +273,50 @@ void failsInTransformingToNonNegativeIntegerForNegative() {
             Matchers.equalTo("the 'ρ' attribute (-42) must be greater or equal to zero")
         );
     }
+
+    @Test
+    void failsInTransformingToIntegerForTooLarge() {
+        MatcherAssert.assertThat(
+            "inner class Integer throws error for a number outside int range, "
```
yegor256: @morphqdd https://www.yegor256.com/2014/06/19/avoid-string-concatenation.html

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1695](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+640](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5309, opened 2026-07-02
#5301: use Expect.Natural in EOresized instead of ad-hoc intValue narrowing

### LINKED ISSUE #5301: `EOmalloc$EOof$EOallocated$EOresized` incorrectly truncates/saturates `new-size` rather than employing `Expect.Natural`
## What happens

`EOmalloc$EOof$EOallocated$EOresized.lambda()` (`eo-runtime/src/main/java/org/eolang/EOmalloc$EOof$EOallocated$EOresized.java`)
reads its `new-size` attribute with:

```java
final int size = Expect.at(this, "new-size")
    .that(Double::intValue)
    .it();
```

Unlike every other malloc-related atom in this file's package (`EOmalloc$EOof$EOφ`,
`EOmalloc$EOof$EOallocated$EOread`, `EOmalloc$EOof$EOallocated$EOwrite`, all of which use
`Expect.Natural`), this one builds its own ad-hoc `Expect` chain that narrows the `Double` to
`int` with `Double::intValue` and never checks the value is a whole number, nor that it fits in
`int` range.

`Double.intValue()` truncates fractional values silently and saturates out-of-range values to
`Integer.MAX_VALUE`/`Integer.MIN_VALUE` rather than throwing. Verified directly:

```
fractional.intValue()=5
huge.intValue()=2147483647
```

So `<allocated>.resized 5.7` silently resizes the block to `5` bytes instead of rejecting a
non-integer size (every other malloc atom rejects non-integers via `Expect.Natural`'s
`number % 1 == 0` check), and `<allocated>.resized 1e15` silently resizes to `2147483647` bytes
instead of raising a domain error for a size nowhere near what was requested. The `id` attribute
in the same method has the identical gap.

## What should happen

`new-size` (and `id`) should be read with `Expect.Natural`, the same helper every other
malloc-related atom in this package already uses, so a fractional or out-of-`int`-range value
raises a domain error instead of being silently truncated or saturated.

morphqdd:
Closes #5301.

`EOmalloc$EOof$EOallocated$EOresized.lambda()` built its own ad-hoc `Expect` chain for `id` and
`new-size` that narrows the `Double` to `int` with `Double::intValue` directly, unlike every other
malloc-related atom in this package (`EOmalloc$EOof$EOφ`, `EOmalloc$EOof$EOallocated$EOread`,
`EOmalloc$EOof$EOallocated$EOwrite`), which all use `Expect.Natural`. `Double.intValue()` truncates
fractional values silently and saturates out-of-range values instead of throwing, so
`<allocated>.resized 5.7` silently resized to `5` bytes instead of rejecting the non-integer size.

Replaced both ad-hoc chains with `new Expect.Natural(...)`, matching the sibling atoms.

Added a `throws-on-changing-size-to-fractional` test to `malloc.eo`, mirroring the existing
`throws-on-changing-size-to-negative` test, covering the fractional-size case that the ad-hoc
chain didn't validate.

Ran the full EO-generated malloc test suite locally (`EOmallocEOAtomTest`) — all 36 tests pass,
including the new one.

### DIFF
--- a/eo-runtime/src/main/eo/malloc.eo
+++ b/eo-runtime/src/main/eo/malloc.eo
@@ -382,6 +382,12 @@
       1
       m.resized -1 > [m]
 
+  # Tests that malloc throws an error when attempting to resize to a fractional size.
+  [] +> throws-on-changing-size-to-fractional
+    malloc.of > @
+      1
+      m.resized 1.5 > [m]
+
   # Tests that malloc.empty creates a memory block with zero size.
   [] +> tests-malloc-empty-is-empty
     malloc.empty > @
--- a/eo-runtime/src/main/java/org/eolang/EOmalloc$EOof$EOallocated$EOresized.java
+++ b/eo-runtime/src/main/java/org/eolang/EOmalloc$EOof$EOallocated$EOresized.java
@@ -29,16 +29,8 @@ public final class EOmalloc$EOof$EOallocated$EOresized extends PhDefault impleme
     @SuppressWarnings("PMD.UnnecessaryLocalRule")
     public Phi lambda() {
         final Phi rho = this.take(Phi.RHO);
-        final int id = Expect.at(rho, "id")
-            .that(phi -> new Dataized(phi).asNumber())
-            .otherwise("must be a number")
-            .that(Double::intValue)
-            .it();
-        final int size = Expect.at(this, "new-size")
-            .that(phi -> new Dataized(phi).asNumber())
-            .otherwise("must be a number")
-            .that(Double::intValue)
-            .it();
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+500](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5310, opened 2026-07-02
#5296: use overflow-safe long arithmetic in Heaps.fits bounds check

### LINKED ISSUE #5296: `Heaps.fits` allows out-of-range read due to integer overflow causing `IllegalArgumentException`
## What happens

`Heaps.fits(int identifier, int offset, int length)` (`eo-runtime/src/main/java/org/eolang/Heaps.java`)
is documented as "the single source of truth for the read-bounds rule" and computes:

```java
return offset + length <= this.blocks.get(identifier).length;
```

`offset` and `length` are plain `int`s (already narrowed from EO natural numbers by
`Expect.Natural`, see also #5295). `offset + length` can overflow `int` and wrap around to a
negative number when both are large — e.g. `offset = 2_000_000_000` and
`length = 2_000_000_000` sum to `-294_967_296`. A negative sum is `<= blockLength` for essentially
any allocated block size, so `fits` reports `true` for a read that is nowhere near actually fitting.

Verified directly:

```
offset=2000000000 length=2000000000 sum=-294967296
sum <= 10 (small block length): true
exception: java.lang.IllegalArgumentException: 2000000000 > -294967296
```

`EOmalloc$EOof$EOallocated$EOread.bytes(id, offset, length)` treats `fits(...)` as the gate between
two documented outcomes: read the bytes, or route to the caller-supplied `cant-read` fallback
(R-3.4.8 error-branch semantics) with a clean `"Can't read '%d' bytes from offset '%d', because
only '%d' are allocated"` message. Because `fits` wrongly answers `true` here, the atom instead
calls `Heaps.read`, which re-derives the same overflowed sum and calls
`Arrays.copyOfRange(bytes, offset, offset + length)`. Since the wrapped `to` is less than `from`,
this throws a raw `IllegalArgumentException: 2000000000 > -294967296` — bypassing the atom's
`cant-read` fallback entirely and surfacing a Java implementation detail instead of the documented
EO-level error-branch behavior.

## What should happen

`Heaps.fits` should compute the bound check without relying on `int` addition that can overflow —
for example by using `long` arithmetic for the comparison — so a request whose `offset + length`
exceeds `Integer.MAX_VALUE` is correctly reported as not fitting, and routes through the documented
`cant-read` fallback instead of crashing with an unrelated exception.

morphqdd:
Closes #5296.

`Heaps.fits(int, int, int)` computed `offset + length <= this.blocks.get(identifier).length` using
plain `int` addition, which can overflow and wrap to a negative number for large `offset`/`length`
(e.g. `2_000_000_000 + 2_000_000_000` wraps to `-294_967_296`). A negative sum is `<=` any
allocated block length, so `fits` wrongly reported `true` for a read nowhere near actually fitting,
which then let `Heaps.read` re-derive the same overflowed sum and crash with a raw
`Arrays.copyOfRange` `IllegalArgumentException` instead of routing through the documented
`cant-read` fallback.

Widened the addition to `long` so the comparison can't overflow, matching the style of an
overflow-safe bound check.

Added `failsOnReadIfOffsetPlusLengthOverflows` to `HeapsTest`, mirroring the existing
`failsOnReadIfOutOfBounds` test, using `offset = length = Integer.MAX_VALUE - 1` (the same class
of value from the issue's reproduction) against a small allocated block.

Ran the full `HeapsTest` suite locally — all 21 tests pass. Verified the overflow directly:
`(Integer.MAX_VALUE - 1) + (Integer.MAX_VALUE - 1)` as `int` gives `-4` (wrongly "fits" a 10-byte
block); as `long` gives `4294967292` (correctly rejected).

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/HeapsTest.java
+++ b/eo-runtime/src/test/java/org/eolang/HeapsTest.java
@@ -99,6 +99,34 @@ final class HeapsTest {
     }
 
     @Test
+    void failsOnReadIfOffsetPlusLengthOverflows() {
+        Assertions.assertThrows(
+            ExFailure.class,
+            () -> Heaps.INSTANCE.read(
+                Heaps.INSTANCE.malloc(new HeapsTest.PhFake(), 10),
+                Integer.MAX_VALUE - 1, Integer.MAX_VALUE - 1
+            ),
+            "Heaps must fail on out-of-bounds read when offset + length overflows int"
+        );
+    }
+
+    @Test
+    void failsCleanlyOnNegativeReadArguments() {
+        final int idx = Heaps.INSTANCE.malloc(new HeapsTest.PhFake(), 10);
+        Assertions.assertThrows(
+            ExFailure.class,
+            () -> Heaps.INSTANCE.read(idx, -5, 3),
+            "Heaps must reject a negative offset with a clean ExFailure, not a raw JVM exception"
+        );
+        Assertions.assertThrows(
+            ExFailure.class,
+            () -> Heaps.INSTANCE.read(idx, 2, -3),
+            "Heaps must reject a negative length with a clean ExFailure, not a raw JVM exception"
+        );
+        Heaps.INSTANCE.free(idx);
+    }
+
+    @Test
     void readsByOffsetAndLength() {
         final int idx = Heaps.INSTANCE.malloc(new HeapsTest.PhFake(), 5);
         Heaps.INSTANCE.write(idx, 0, new byte[] {1, 2, 3, 4, 5});
--- a/eo-runtime/src/main/java/org/eolang/Heaps.java
+++ b/eo-runtime/src/main/java/org/eolang/Heaps.java
@@ -150,7 +150,9 @@ final class Heaps {
                     )
                 );
             }
-            return offset + length <= this.blocks.get(identifier).length;
+            return offset >= 0
+                && length >= 0
+                && (long) offset + length <= this.blocks.get(identifier).length;
         } finally {
             this.lock.unlock();
         }


morphqdd:

eo-runtime/src/test/java/org/eolang/HeapsTest.java
```diff
@@ -98,6 +98,19 @@ void failsOnReadIfOutOfBounds() {
...
 
+    @Test
+    void failsOnReadIfOffsetPlusLengthOverflows() {
+        Assertions.assertThrows(
+            ExFailure.class,
+            () -> Heaps.INSTANCE.read(
+                Heaps.INSTANCE.malloc(new HeapsTest.PhFake(), 10),
+                Integer.MAX_VALUE - 1, Integer.MAX_VALUE - 1
+            ),
+            "Heaps should throw an exception on out-of-bounds read when offset + length "
```
yegor256: @morphqdd https://www.yegor256.com/2014/06/19/avoid-string-concatenation.html

yegor256 (changes requested):
@morphqdd run full build

yegor256:
@morphqdd build is broken here

gemshrine (commented):
The `long`-promotion closes the positive-overflow gap that #5296 describes. Two negative-input gaps in the same predicate remain open, and let the same class of "unclean crash" through when `fits` gates `read`.

For a block of size 10:

```
offset=-5 length=3  => fits returns: true
  copyOfRange throws: ArrayIndexOutOfBoundsException: source index -5 out of bounds for byte[10]
offset=2  length=-3 => fits returns: true
  copyOfRange throws: IllegalArgumentException: 2 > -1
```

In both cases `(long) offset + length` is `≤ arr.length`, so the current check says "fits", `read` proceeds to `Arrays.copyOfRange`, and the caller gets a raw JVM exception instead of the clean `ExFailure` the surrounding code raises everywhere else. Tightening the predicate to reject negatives keeps the fix compatible with the current test and closes both cases:

```java
return offset >= 0
    && length >= 0
    && (long) offset + length <= this.blocks.get(identifier).length;
```

A parameterised test on `(offset, length)` with `(-5, 3)`, `(2, -3)`, `(Integer.MIN_VALUE, 0)` would lock this in alongside the existing overflow test.

morphqdd:
@gemshrine good catch, thank you — you're right that the positive-overflow fix left both negative gaps open (`offset=-5, length=3` and `offset=2, length=-3` both satisfy `(long) offset + length <= arr.length`, so `fits` said "fits" and `read` handed the caller a raw `ArrayIndexOutOfBoundsException`/`IllegalArgumentException` instead of a clean `ExFailure`).

Applied your suggestion:

```java
return offset >= 0
    && length >= 0
    && (long) offset + length <= this.blocks.get(identifier).length;
```

and added a `failsCleanlyOnNegativeReadArguments` test covering `(-5, 3)` and `(2, -3)` (both now assert `ExFailure`). Ran the full `eo-runtime` build locally: `HeapsTest` 22/22 green, whole module `1403` tests, 0 failures/0 errors.

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1688](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+652](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5311, opened 2026-07-02
#5297: use overflow-safe long arithmetic in Heaps.write resize-trigger

### LINKED ISSUE #5297: `Heaps.write` fails to handle integer overflow, skipping resize and resulting in `ArrayIndexOutOfBoundsException`
## What happens

`Heaps.write(int identifier, int offset, byte[] data)` (`eo-runtime/src/main/java/org/eolang/Heaps.java`)
decides whether the block needs to grow like this:

```java
if (this.blocks.get(identifier).length < offset + data.length) {
    this.resize(identifier, offset + data.length);
}
```

`offset + data.length` is plain `int` arithmetic and can overflow. For
`offset = Integer.MAX_VALUE` (a value `Expect.Natural` happily accepts, since it's a non-negative
whole number within `int` range) and a one-byte `data`, `offset + data.length` wraps to
`Integer.MIN_VALUE`. Since any allocated block length is never `< Integer.MIN_VALUE`, the
resize is skipped entirely, and execution falls through to:

```java
System.arraycopy(data, 0, result, offset, data.length);
```

with `result.length` still the original (small) block size and `offset` still
`Integer.MAX_VALUE`, throwing a raw `ArrayIndexOutOfBoundsException` instead of resizing the
block as the method's own logic intends.

Verified directly:

```
offset=2147483647 dataLength=1 sum=-2147483648
blockLength < sum: false
exception: java.lang.ArrayIndexOutOfBoundsException: arraycopy: last destination index 2147483648 out of bounds for byte[10]
```

This is the same class of integer-overflow defect as #5296 (`Heaps.fits`), but in the separate
resize-trigger condition inside `Heaps.write`.

## What should happen

The resize-trigger comparison should use overflow-safe arithmetic (e.g. `long` addition) so a
huge `offset` correctly triggers a resize (or a clean domain error) instead of silently skipping
the resize and crashing later with an unrelated `ArrayIndexOutOfBoundsException`.

morphqdd:
Closes #5297.

`Heaps.write(int, int, byte[])` decided whether the block needs to grow via plain `int` addition:
`this.blocks.get(identifier).length < offset + data.length`. For `offset = Integer.MAX_VALUE` and
a one-byte `data` — a value `Expect.Natural` happily accepts — `offset + data.length` wraps to
`Integer.MIN_VALUE`, so the resize-trigger condition is never true, execution falls through to
`System.arraycopy(data, 0, result, offset, data.length)` with the original small block size, and
throws a raw `ArrayIndexOutOfBoundsException` instead of resizing (or raising a clean domain error)
as the method's own logic intends.

Widened the sum to `long` (`end`). If `end` exceeds `Integer.MAX_VALUE` — meaning no `byte[]` of
that size can ever be allocated — the method now throws a clean `ExFailure` naming the offset and
resulting size, instead of silently skipping the resize and crashing later with an unrelated
`ArrayIndexOutOfBoundsException`. Otherwise, the existing resize path is used unchanged with the
now-correct `long`-computed value.

Added `failsOnWriteIfOffsetPlusLengthOverflows` to `HeapsTest`, using `offset = Integer.MAX_VALUE`
against a small allocated block, mirroring the existing `failsOnWriteToEmptyBlock` test style.

Ran the full `HeapsTest` suite locally — all 21 tests pass.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Heaps.java
+++ b/eo-runtime/src/main/java/org/eolang/Heaps.java
@@ -199,8 +199,17 @@ final class Heaps {
                     )
                 );
             }
-            if (this.blocks.get(identifier).length < offset + data.length) {
-                this.resize(identifier, offset + data.length);
+            final long end = (long) offset + data.length;
+            if (end > Integer.MAX_VALUE) {
+                throw new ExFailure(
+                    String.format(
+                        "Block '%d': can't write at offset '%d', resulting size '%d' is too large for int",
+                        identifier, offset, end
+                    )
+                );
+            }
+            if (this.blocks.get(identifier).length < end) {
+                this.resize(identifier, (int) end);
             }
             final byte[] source = this.blocks.get(identifier);
             final int length = source.length;
--- a/eo-runtime/src/test/java/org/eolang/HeapsTest.java
+++ b/eo-runtime/src/test/java/org/eolang/HeapsTest.java
@@ -79,6 +79,18 @@ final class HeapsTest {
     }
 
     @Test
+    void failsOnWriteIfOffsetPlusLengthOverflows() {
+        Assertions.assertThrows(
+            ExFailure.class,
+            () -> Heaps.INSTANCE.write(
+                Heaps.INSTANCE.malloc(new HeapsTest.PhFake(), 10),
+                Integer.MAX_VALUE, new byte[] {0x01}
+            ),
+            "Heaps should throw an exception on write when offset + data.length overflows int, but it didn't"
+        );
+    }
+
+    @Test
     void failsOnReadFromEmptyBlock() {
         Assertions.assertThrows(
             ExFailure.class,


morphqdd:

yegor256:
@morphqdd string concatenation is evil: https://www.yegor256.com/2014/06/19/avoid-string-concatenation.html

maxonfjvipon (approved):

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+572](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (25) hits-of-code; +4 to give you at least something.  Your running score is [+525](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5312, opened 2026-07-02
#5298: use overflow-safe long arithmetic in EObytes$EOslice bounds check

### LINKED ISSUE #5298: `EObytes$EOslice` allows out-of-range slice due to integer overflow in `start + len` bounds check, leading to raw `IllegalArgumentException`
## What happens

`EObytes$EOslice.lambda()` (`eo-runtime/src/main/java/org/eolang/EObytes$EOslice.java`) validates
`start` and `len` are non-negative integers, then checks the range fits with:

```java
.must(integer -> start + integer <= bytes.length).otherwise(
    String.format("is out of bounds for bytes of size %d", bytes.length)
)
```

`start + integer` (i.e. `start + len`) is plain `int` addition and can overflow. For
`start = 2_000_000_000` and `len = 2_000_000_000` — both individually valid, non-negative
integers that pass every earlier check — the sum wraps to `-294_967_296`. A negative sum is
`<= bytes.length` for any block, so the "is out of bounds" check wrongly reports the range as
fitting.

Verified directly:

```
start+len=-294967296 check(should be false, is): true
exception: java.lang.IllegalArgumentException: 2000000000 > -294967296
```

Execution then reaches `Arrays.copyOfRange(bytes, start, start + len)` with the same overflowed
`start + len`, and since the wrapped end index is less than `start`, this throws a raw
`IllegalArgumentException: 2000000000 > -294967296` instead of the domain-specific "is out of
bounds for bytes of size %d" error the validation pipeline is meant to produce.

This is the same class of defect as #5296 and #5297 (`Heaps.fits` / `Heaps.write`), here in
`EObytes$EOslice`'s own bounds check.

## What should happen

The bounds check should add `start` and `len` using overflow-safe arithmetic (e.g. widen to
`long` before adding) so a request whose sum exceeds `Integer.MAX_VALUE` is correctly rejected
with the existing "is out of bounds for bytes of size %d" error instead of slipping through and
crashing with an unrelated `IllegalArgumentException`.

morphqdd:
Closes #5298.

`EObytes$EOslice.lambda()` validated `start` and `len` are non-negative integers, then checked the
range fits with plain `int` addition: `start + integer <= bytes.length`. For
`start = 2_000_000_000` and `len = 2_000_000_000` — both individually valid, non-negative integers
that pass every earlier check — the sum wraps to `-294_967_296`. A negative sum is `<=` any block
size, so the "is out of bounds" check wrongly reported the range as fitting, and execution reached
`Arrays.copyOfRange(bytes, start, start + len)` with the same overflowed sum, throwing a raw
`IllegalArgumentException` instead of the domain-specific "is out of bounds for bytes of size %d"
error.

Widened the addition to `long` in the bounds check, matching the fix already applied to the same
defect class in `Heaps.fits`/`Heaps.write` (#5296/#5297).

Added `throws-on-slice-with-overflowing-start-plus-len` to `bytes.eo`, mirroring the existing
`throws-on-out-of-bounds-slice` test, using `start = len = 2000000000` (the same reproduction
values from the issue).

Ran the full EO-generated bytes test suite locally (`EObytesTest`) — all 60 tests pass, including
the new one, which confirms the domain error ("the 'len' attribute (2000000000) is out of bounds
for bytes of size 5") is now raised instead of a raw Java exception.

### DIFF
--- a/eo-runtime/src/main/eo/bytes.eo
+++ b/eo-runtime/src/main/eo/bytes.eo
@@ -541,6 +541,17 @@
         10
         "recovered" > [msg]
 
+  # Tests that a slice whose `start + len` overflows a 32-bit int does not
+  # wrap around into a negative value and slip through the bounds check, but
+  # is detected as out-of-bounds and yields the `cant-slice` fallback.
+  [] +> tests-slice-with-overflowing-start-plus-len-returns-fallback
+    eq. > @
+      "recovered"
+      20-1F-EE-B5-90.slice
+        2000000000
+        2000000000
+        "recovered" > [msg]
+
   # Tests that bytes can be correctly converted to i64 integer.
   [] +> tests-bytes-converts-to-i64
     eq. > @
--- a/eo-runtime/src/main/java/org/eolang/EObytes$EOslice.java
+++ b/eo-runtime/src/main/java/org/eolang/EObytes$EOslice.java
@@ -42,7 +42,7 @@ public final class EObytes$EOslice extends PhDefault implements Atom {
         final int start = this.natural("start");
         final int len = this.natural("len");
         final Phi result;
-        if (start + len <= bytes.length) {
+        if ((long) start + len <= bytes.length) {
             result = new Data.ToPhi(Arrays.copyOfRange(bytes, start, start + len));
         } else {
             result = this.take(EObytes$EOslice.FALLBACK);


morphqdd:

yegor256:
@morphqdd conflicts here

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+512](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5314, opened 2026-07-03
#5303: use Expect.Natural for chunk size instead of raw intValue narrowing

### LINKED ISSUE #5303: `file.open.file-stream.read.chunk` causes `NegativeArraySizeException` when `size` is negative
## What happens

`EOfile$EOopen$EOfile_stream$EOread$EOchunk.lambda()` reads its `size` attribute with:

```java
new Dataized(this.take("size")).asNumber().intValue()
```

and passes the result straight into `Files.read(String, int)`
(`eo-runtime/src/main/java/org/eolang/EOfs/Files.java`), which does:

```java
final byte[] read = new byte[size];
```

There is no check that `size` is non-negative (or even a whole number — `Double.intValue()`
truncates fractions silently too). A negative `size` — e.g. `file-stream.read.chunk -5`, or any EO
expression that computes a negative remaining-bytes count by mistake — reaches
`new byte[size]` with a negative length and throws an uncaught `NegativeArraySizeException`.

Verified directly:

```
exception: java.lang.NegativeArraySizeException: -5
```

`file.open.file-stream.read.chunk` is a public atom directly reachable from ordinary EO code, so
any program computing a bad `size` value (a typo'd sign, an off-by-one on a "remaining bytes"
calculation, etc.) gets a raw Java exception instead of the kind of domain error other malloc/file
atoms in this codebase raise for bad size arguments (see `Expect.Natural`'s
"must be greater or equal to zero").

## What should happen

`size` should be validated (non-negative, whole number) before it reaches `new byte[size]`,
raising a clean domain error (e.g. via `Expect.Natural`, already used by the malloc atoms for the
same purpose) instead of an uncaught `NegativeArraySizeException`.

morphqdd:
Closes #5303.

`EOfile$EOopen$EOfile_stream$EOread$EOchunk.lambda()` read its `size` attribute with a raw
`new Dataized(this.take("size")).asNumber().intValue()` and passed the result straight into
`Files.INSTANCE.read(String, int)` (`eo-runtime/src/main/java/org/eolang/EOfs/Files.java`), which
does `new byte[size]`. There was no check that `size` is non-negative (or even a whole number —
`Double.intValue()` truncates fractions silently too). A negative `size` — e.g.
`file-stream.read.chunk -5`, or any EO expression that computes a negative remaining-bytes count
by mistake — reached `new byte[size]` with a negative length and threw an uncaught
`NegativeArraySizeException`.

`file.open.file-stream.read.chunk` is a public atom directly reachable from ordinary EO code
(`file-stream.read` → `input-block.read` → `chunk`), so any program computing a bad `size` value
got a raw Java exception instead of the kind of clean domain error other atoms in this codebase
raise for bad size arguments (`Expect.Natural`'s "must be greater or equal to zero" — already used
by the malloc atoms for exactly this purpose).

Replaced the raw narrowing with `new Expect.Natural(Expect.at(this, "size")).it()`, matching the
established convention.

Added `throws-on-reading-negative-size-from-file` to `fs/file.eo`, mirroring the neighboring
`throws-on-reading-from-file-with-wrong-mode` test, opening a tmpfile for reading and calling
`f.read -5`.

Ran the full EO-generated file test suite locally (`EOfs.EOfileTest`) — all 34 tests pass. The
verbose log confirms the exact fix in action:

```
"the 'size' attribute (-5) must be greater or equal to zero"
```

replacing what was previously an uncaught `NegativeArraySizeException`.

### DIFF
--- a/eo-runtime/src/main/eo/fs/file.eo
+++ b/eo-runtime/src/main/eo/fs/file.eo
@@ -487,6 +487,15 @@
             m
     .eq "Hello, world" > @
 
+  # Tests that reading a negative size from a file throws a clean domain error
+  # instead of an uncaught NegativeArraySizeException.
+  [] +> throws-on-reading-negative-size-from-file
+    tmpdir
+    .tmpfile
+    .open > @
+      "r"
+      f.read -5 > [f]
+
   # Tests that attempting to read from a file opened in write-only mode throws an error.
   [] +> throws-on-reading-from-file-with-wrong-mode
     tmpdir
--- a/eo-runtime/src/main/java/org/eolang/EOfs/EOfile$EOopen$EOfile_stream$EOread$EOchunk.java
+++ b/eo-runtime/src/main/java/org/eolang/EOfs/EOfile$EOopen$EOfile_stream$EOread$EOchunk.java
@@ -17,6 +17,7 @@ import org.eolang.Atom;
 import org.eolang.Attr;
 import org.eolang.Attrs;
 import org.eolang.Dataized;
+import org.eolang.Expect;
 import org.eolang.PhDefault;
 import org.eolang.Phi;
 import org.eolang.XmirObject;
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+536](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5316, opened 2026-07-04
#5306: reject numbers outside long range in sprintf %d conversion instead of saturating

### LINKED ISSUE #5306: `sprintf`'s `%d` conversion incorrectly saturates out-of-range values instead of throwing an error
## What happens

`sprintf`'s `%d` format specifier (`eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java`)
converts the argument to a displayed integer with:

```java
SprintfArgs.CONVERSION.put('d', element -> element.asNumber().longValue());
```

`asNumber()` returns a `Double`, and `Double.longValue()` does not throw for a value outside
`long` range — it saturates, returning `Long.MAX_VALUE`/`Long.MIN_VALUE`. Verified directly:

```
huge.longValue()=9223372036854775807
```

`sprintf`'s own doc comment documents `%d` as converting the object "to `int`", with no mention
of range limits. So `sprintf("%d", *1e30)` — formatting a plain `number` far larger than any
`long` can hold — silently prints `9223372036854775807` instead of raising a domain error naming
the value that doesn't fit. This is the same `Double`-to-integral narrowing-without-validation
defect already reported for `Expect.Natural`/`Expect.Int` (#5295) and `number.as-i64` (#5305),
here specifically in the `%d` conversion path used by every `sprintf` call in this codebase.

## What should happen

The `%d` conversion should reject a value outside `long` range with a clear error naming the
offending value, instead of silently saturating to `Long.MAX_VALUE`/`Long.MIN_VALUE`.

morphqdd:
Closes #5306.

`sprintf`'s `%d` conversion (`SprintfArgs.CONVERSION.put('d', element -> element.asNumber().longValue())`)
narrowed the argument's `Double` value to `long` via `Double.longValue()`, which doesn't throw for
a value outside `long` range — it saturates, returning `Long.MAX_VALUE`/`Long.MIN_VALUE`. So
`sprintf("%d", *1e30)` silently printed `9223372036854775807` instead of raising a domain error
naming the offending value. This is the same `Double`-to-integral narrowing-without-validation
defect already fixed for `Expect.Natural`/`Expect.Int` (#5295/#5307) and `number.as-i64`
(#5305/#5315), here in the `%d` conversion path used by every `sprintf` call in this codebase.

Added a `SprintfArgs.toLong(double)` helper that checks the value against `Long.MIN_VALUE`/
`Long.MAX_VALUE` before narrowing, and wired it into the `'d'` entry of the `CONVERSION` map in
place of the raw `.longValue()` call.

Added `throws-on-formatting-huge-number-as-decimal` to `tt/sprintf.eo`, using `1.0e30` — the exact
reproduction value from the issue — mirroring the existing `truncates-*-as-decimal` tests for the
`%d` conversion.

Ran the full EO-generated sprintf test suite locally (`EOtt.EOsprintfTest`) — all 20 tests pass.

### DIFF
--- a/eo-runtime/src/main/eo/tt/sprintf.eo
+++ b/eo-runtime/src/main/eo/tt/sprintf.eo
@@ -141,6 +141,13 @@
         "%d"
         * 9.99
 
+  # Tests that sprintf throws an error for %d when the number doesn't fit
+  # into long range, instead of silently saturating to Long.MAX_VALUE.
+  [] +> throws-on-formatting-huge-number-as-decimal
+    sprintf > @
+      "%d"
+      * 1.0e30
+
   # Tests that sprintf always pads a float to exactly six fraction digits.
   [] +> formats-float-always-with-six-fraction-digits
     eq. > @
--- a/eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java
@@ -40,7 +40,7 @@ final class SprintfArgs {
 
     static {
         SprintfArgs.CONVERSION.put('s', Dataized::asString);
-        SprintfArgs.CONVERSION.put('d', element -> element.asNumber().longValue());
+        SprintfArgs.CONVERSION.put('d', element -> SprintfArgs.toLong(element.asNumber()));
         SprintfArgs.CONVERSION.put('f', Dataized::asNumber);
         SprintfArgs.CONVERSION.put('x', element -> SprintfArgs.bytesToHex(element.take()));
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+548](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5317, opened 2026-07-04
#5299: reject oversized positional index in sprintf instead of raw NumberFormatException

### LINKED ISSUE #5299: `sprintf` causes `NumberFormatException` when handling oversized positional argument indices
## What happens

`SprintfArgs.formatted()` (`eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java`) parses
positional argument specifiers (`%N$s`) out of the user-supplied format string with:

```java
final Matcher matcher = Pattern.compile("%(\\d+\\$)?([a-zA-Z%])").matcher(this.format);
...
if (positional != null) {
    arg = Long.parseLong(positional.substring(0, positional.length() - 1)) - 1L;
}
```

The regex `\d+` places no upper bound on how many digits the positional index can have, and
`Long.parseLong` is not guarded by a try/catch here. `EOsprintf.lambda()` passes `format` straight
from `new Dataized(this.take("format")).asString()` — an arbitrary, fully user-controlled EO
string — into `SprintfArgs`, so any EO program that calls `sprintf` with a format string carrying
a positional index too large to fit in a `long` (more than 19 digits) reaches this code path.

Verified directly:

```
exception: java.lang.NumberFormatException: For input string: "99999999999999999999"
```

So `sprintf "%99999999999999999999$s" ...` crashes with a raw uncaught
`NumberFormatException` instead of the `ExFailure`-based "argument index %d is out of bounds"
error this same method already raises for indices that parse fine but are simply too large for
the argument list (see the `arg >= this.length` check right below the `parseLong` call) — the
intent is clearly to reject bad positional indices with a domain error, but the too-many-digits
case slips past that intent entirely.

## What should happen

An oversized positional index should be rejected with the same kind of domain error
(`ExFailure`, "The argument index ... is out of bounds") the method already raises for
in-range-but-too-large indices, not an uncaught `NumberFormatException`.

morphqdd:
Closes #5299.

`SprintfArgs.formatted()` parses positional argument specifiers (`%N$s`) with the regex
`%(\d+\$)?([a-zA-Z%])`, which places no upper bound on how many digits the positional index can
have, then calls `Long.parseLong(...)` on the captured digits with no try/catch. Since the format
string is arbitrary, fully user-controlled EO input, a positional index with more than 19 digits
(too large for `long`) reached `Long.parseLong` uncaught and crashed with a raw
`NumberFormatException` — e.g. `sprintf "%99999999999999999999$s" ...` — instead of the same
`ExFailure`-based "argument index ... is out of bounds" domain error this method already raises
for indices that parse fine but are simply too large for the argument list.

Wrapped the `Long.parseLong` call in a try/catch, converting `NumberFormatException` into the same
"out of bounds" `ExFailure` message (naming the offending digit string instead of a parsed index,
since parsing failed).

Added `throws-on-sprintf-with-oversized-positional-index` to `tt/sprintf.eo`, using the exact
`%99999999999999999999$s` reproduction from the issue.

Ran the full EO-generated sprintf test suite locally (`EOtt.EOsprintfTest`) — all 20 tests pass.

### DIFF
--- a/eo-runtime/src/main/eo/tt/sprintf.eo
+++ b/eo-runtime/src/main/eo/tt/sprintf.eo
@@ -101,6 +101,13 @@
         "Hello, %s! Bye, %1$s!"
         * "Jeff"
 
+  # Tests that sprintf throws a bounds error for a positional index too large
+  # to fit into long, instead of an uncaught NumberFormatException.
+  [] +> throws-on-sprintf-with-oversized-positional-index
+    sprintf > @
+      "%99999999999999999999$s"
+      * "Jeff"
+
   # Tests that sprintf handles mixed numbered and sequential parameters.
   [] +> formats-numbered-substitution-with-multiple-args
     eq. > @
--- a/eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java
@@ -85,7 +85,18 @@ final class SprintfArgs {
             }
             final long arg;
             if (positional != null) {
-                arg = Long.parseLong(positional.substring(0, positional.length() - 1)) - 1L;
+                final String digits = positional.substring(0, positional.length() - 1);
+                try {
+                    arg = Long.parseLong(digits) - 1L;
+                } catch (final NumberFormatException ex) {
+                    throw new ExFailure(
+                        String.format(
+                            "The argument index %s is out of bounds (total arguments: %d)",
+                            digits, this.length
+                        ),
+                        ex
+                    );
+                }
             } else {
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+560](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5319, opened 2026-07-04
#5308: extract HEX/STRING/GROUP branches out of Emissions.openValue to reduce NCSS below qulice's limit

### LINKED ISSUE #5308: `Emissions.openValue` NCSS exceeds limit causing intermittent `qulice` checks failure on unrelated PRs

morphqdd:
Closes #5308.

`Emissions.openValue` had grown to NCSS (Non-Commenting Source Statements) 51, one over qulice's
`JavaNCSSCheck` limit of 50, causing the `qulice` CI check to fail intermittently — and, since the
method's source doesn't change between builds, identically — on unrelated PRs and on `master`
itself (see the issue for a list of affected runs).

Extracted the three most complex branches of the `if`/`else if` chain — `HEX`, `STRING`, and
`GROUP` (the ones carrying nested `try`/`catch` blocks or extra branching) — into their own
private methods (`openHex`, `openString`, `openGroup`), following the same pattern this class
already uses for `bytesCarrier`. `openValue` itself is now a flat dispatch over `Value.Kind`, well
under the NCSS limit.

Removed `PMD.CognitiveComplexity` from the class-level `@SuppressWarnings` — PMD flagged it as an
`UnnecessaryWarningSuppression` after the refactor, since `openValue`'s cognitive complexity
(nesting depth from the extracted `try`/`catch` blocks) no longer triggers that rule. This is
unrelated to the `JavaNCSSCheck`/`qulice` failure this PR fixes; it's a separate PMD metric that
happened to also improve from the same extraction, and its now-dead suppression had to go so PMD
itself would stop complaining about it.

## Verification

Ran the full `qulice` profile locally on `eo-parser` (`mvn -pl eo-parser -am clean verify
-PskipTests -Pqulice`) — went from `There are 1 violations` (the original NCSS failure) down to
`BUILD SUCCESS`, with two intermediate refactor passes along the way (fixing a
`ParameterNumberCheck` and `MethodsOrderCheck` violation on the three new methods, then the
`PMD.UnnecessaryWarningSuppression` one above).

Ran the full `eo-parser` test suite locally — `1430/1430` tests pass, confirming the extraction
is behavior-preserving.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -20,7 +20,7 @@ import java.util.List;
  * @checkstyle CyclomaticComplexityCheck (600 lines)
  * @checkstyle BooleanExpressionComplexityCheck (600 lines)
  */
-@SuppressWarnings({"PMD.UnnecessaryLocalRule", "PMD.TooManyMethods", "PMD.CognitiveComplexity"})
+@SuppressWarnings({"PMD.UnnecessaryLocalRule", "PMD.TooManyMethods"})
 final class Emissions {
 
     /**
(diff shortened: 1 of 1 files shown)


morphqdd:

yegor256:
@morphqdd conflicts here

morphqdd:
Closing this — it's now obsolete. `master` already resolved the `Emissions.openValue` NCSS-51 qulice violation by extracting the `GROUP` branch into a separate `group(...)` method, so `openValue` is back under the limit (master's qulice run is green). This PR's HEX/STRING/GROUP extraction is no longer needed and now conflicts with master's version. The other overflow/range-guard PRs that were red only because of this shared NCSS violation will go green once rebased onto current master.

PR #5320, opened 2026-07-04
#5304: mix relative path into Cache.dirSha so a file rename changes the hash

### LINKED ISSUE #5304: `Cache.dirSha` reuses cached output after source file renaming due to lack of file path hashing
## What happens

`Cache.dirSha` (`eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java`) is the production
directory-hashing path used by `Linting`, `Transpiling`, and `Parsing` to decide whether cached
compiled output can be reused:

```java
private String dirSha(final Path dir) {
    try {
        try (Stream<Path> stream = Files.walk(dir)) {
            stream.filter(Files::isRegularFile)
                .filter(this.filter::test)
                .sorted(Comparator.comparing(Path::toString))
                .map(p -> new Sha(p).toString())
                .map(s -> s.getBytes(StandardCharsets.UTF_8))
                .forEach(digest::update);
        }
    }
    ...
}
```

#5254 (fixed by #5262) reported that `Sha.hash`'s own directory branch is unsafe because it
concatenates raw file *contents* with no delimiter and no paths, and noted in passing that
`Cache.dirSha` "does not reuse `Sha`'s directory branch — it independently hashes each file to a
fixed-length per-file SHA string and concatenates those," implying that reimplementation is safe.
It sidesteps the delimiter-injection half of the problem (each `new Sha(p).toString()` is a
fixed-length Base64 SHA-256 string, so concatenation boundaries are unambiguous), but it inherits
the other half unchanged: **`Path::toString` is used only for `.sorted(...)`, and is never fed
into the digest.** Only each file's content hash is hashed — never its name or relative path.

Concretely, two single-file directories whose one file has identical content but a different name
produce the exact same per-file content hash (verified with `sha256sum`, since `Sha`'s per-file
hash for a regular file is just that file's own SHA-256):

```
b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9  dirA/original-name.eo
b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9  dirB/completely-different-name.eo
```

Since `dirSha` never hashes the path itself, `dirSha(dirA)` and `dirSha(dirB)` are identical
Base64 digests, even though the two directories contain a differently-named source file — and in
this EO toolchain a file's name is semantically significant (the compiled output for `original-
name.eo` encodes that name). `Cache.apply` treats a `sha` match as "reuse the cached content," so
a rename that happens to collide this way can make the cache silently serve compiled output built
for a different file name.

## What should happen

`dirSha` should mix each file's relative path into the digest alongside its content hash — for
example hashing `rel_path + "\0" + Sha(file)` per entry, matching the fix already applied to
`Sha.hash`'s directory branch for #5254 — so a directory whose only difference is a file's name
produces a different hash.

morphqdd:
Closes #5304.

`Cache.dirSha` — the production directory-hashing path used by `Linting`, `Transpiling`, and
`Parsing` to decide whether cached compiled output can be reused — hashed each regular file's
*content* independently (`new Sha(p).toString()`), sorted by path only for iteration order, but
never fed the path itself into the digest. Two single-file directories whose one file has
identical content but a different name produced the exact same per-file content hash, so
`dirSha(dirA)` and `dirSha(dirB)` came out byte-for-byte identical — even though a file's name is
semantically significant in this toolchain (the compiled output for `original-name.eo` encodes
that name). Since `Cache.apply` treats a `sha` match as "reuse the cached content," a rename that
happens to collide this way could make the cache silently serve compiled output built for a
different file name.

This is the other half of the defect already fixed for `Sha.hash`'s own directory branch in
#5254/#5262 — that fix addressed the delimiter-injection half (concatenating raw file *contents*
with no boundary), while `Cache.dirSha`'s independent reimplementation sidestepped that half (each
per-file contribution is already a fixed-length Base64 SHA-256 string) but never addressed the
path-omission half at all.

Mixed each file's relative path into the digest alongside its content hash, joined by a NUL byte
(`rel_path + "\0" + Sha(file)`, matching the fix suggested in the issue) — a NUL is safe as a
separator since file paths can't contain one, unlike a printable delimiter such as a space.

## Test

Updated the existing `generatesCorrectHashForEntireFolderWithSeveralFiles` test to expect the new
per-entry format (`"file1.txt\0" + hash` instead of just `hash`).

Added `generatesDifferentHashesForFoldersWithDifferentlyNamedIdenticalFiles`, reproducing the
exact scenario from the issue: two directories, `dirA/original-name.eo` and
`dirB/completely-different-name.eo`, both containing the same content, asserting their cached
`.sha256` files are no longer identical.

Ran the full `CacheTest` suite locally — all 10 tests pass.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
@@ -146,7 +146,7 @@ final class Cache {
                 stream.filter(Files::isRegularFile)
                     .filter(this.filter::test)
                     .sorted(Comparator.comparing(Path::toString))
-                    .map(p -> new Sha(p).toString())
+                    .map(p -> String.format("%s\u0000%s", dir.relativize(p), new Sha(p)))
                     .map(s -> s.getBytes(StandardCharsets.UTF_8))
                     .forEach(digest::update);
             }
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
@@ -226,8 +226,12 @@ final class CacheTest {
             source.getFileName()
         );
         final MessageDigest instance = MessageDigest.getInstance("SHA-256");
-        instance.update(CacheTest.hash(first).getBytes(StandardCharsets.UTF_8));
-        instance.update(CacheTest.hash(second).getBytes(StandardCharsets.UTF_8));
+        instance.update(
+            String.format("file1.txt\u0000%s", CacheTest.hash(first)).getBytes(StandardCharsets.UTF_8)
+        );
+        instance.update(
+            String.format("file2.txt\u0000%s", CacheTest.hash(second)).getBytes(StandardCharsets.UTF_8)
+        );
         MatcherAssert.assertThat(
             "SHA-256 hash file has incorrect content for folder with several files",
             Files.readString(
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+588](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5321, opened 2026-07-04
#5302: reject digits beyond long capacity in sscanf %d conversion instead of raw NumberFormatException

### LINKED ISSUE #5302: `sscanf`'s `%d` conversion throws uncaught `NumberFormatException` when parsing digits beyond `long` capacity
## What happens

`EOsscanf`'s `%d` conversion (`eo-runtime/src/main/java/org/eolang/EOtt/EOsscanf.java`) captures
digits with the unbounded regex `([+-]?\d+)` and converts the match directly with:

```java
EOsscanf.CONVERSION.put('d', str -> new Data.ToPhi(Long.parseLong(str)));
```

`Long.parseLong` is not guarded here, and the regex places no limit on how many digits it
captures. A numeric string longer than 19 digits (more than `Long.MAX_VALUE` can hold) matches the
regex fine but throws an uncaught `NumberFormatException` when handed to `Long.parseLong`.

Verified directly:

```
exception: java.lang.NumberFormatException: For input string: "99999999999999999999"
```

So `sscanf "%d" "99999999999999999999"` — parsing an ordinary, all-digit input string that simply
has more digits than fit in a `long` — crashes with a raw `NumberFormatException` instead of a
clean `ExFailure`, the same class of defect already reported for `sprintf`'s oversized positional
index (#5299).

## What should happen

An all-digit match too large to fit in a `long` should raise a clean `ExFailure` naming the
offending input, not an uncaught `NumberFormatException`.

morphqdd:
Closes #5302.

`EOsscanf`'s `%d` conversion captured digits with the unbounded regex `([+-]?\d+)` and converted
the match directly via `Long.parseLong(str)`, with no guard. A numeric string longer than 19
digits — more than `Long.MAX_VALUE` can hold — matches the regex fine but threw an uncaught
`NumberFormatException` when handed to `Long.parseLong`, so
`sscanf "%d" "99999999999999999999"` crashed with a raw Java exception instead of a clean domain
error. This is the same defect class already fixed for `sprintf`'s oversized positional index
(#5299/#5317) and `sprintf`'s `%d` saturation (#5306/#5316).

Extracted a `parsed(String)` helper that wraps `Long.parseLong` in a try/catch, converting
`NumberFormatException` into an `ExFailure` naming the offending input ("The number ... doesn't
fit into long range for the '%d' conversion"), and wired it into the `'d'` entry of the
`CONVERSION` map.

Added `throws-on-sscanf-with-oversized-int` to `tt/sscanf.eo`, using the exact
`"99999999999999999999"` reproduction from the issue, mirroring the neighboring
`throws-on-sscanf-with-wrong-format` test.

Ran the full EO-generated sscanf test suite locally (`EOtt.EOsscanfTest`) — all 14 tests pass.

### DIFF
--- a/eo-runtime/src/main/eo/tt/sscanf.eo
+++ b/eo-runtime/src/main/eo/tt/sscanf.eo
@@ -58,6 +58,11 @@
   [] +> throws-on-sscanf-with-wrong-format
     sscanf "%l" "error" > @
 
+  # Tests that sscanf throws error for %d when the digits don't fit
+  # into long range, instead of an uncaught NumberFormatException.
+  [] +> throws-on-sscanf-with-oversized-int
+    sscanf "%d" "99999999999999999999" > @
+
   # Tests that sscanf parses multiple values with different format specifiers.
   [] +> tests-sscanf-with-string-int-float
     eq. > @
--- a/eo-runtime/src/main/java/org/eolang/EOtt/EOsscanf.java
+++ b/eo-runtime/src/main/java/org/eolang/EOtt/EOsscanf.java
@@ -42,7 +42,7 @@ public final class EOsscanf extends PhDefault implements Atom {
 
     static {
         EOsscanf.CONVERSION.put('s', ToPhi::new);
-        EOsscanf.CONVERSION.put('d', str -> new Data.ToPhi(Long.parseLong(str)));
+        EOsscanf.CONVERSION.put('d', str -> new Data.ToPhi(EOsscanf.parsed(str)));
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (27) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+600](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5325, opened 2026-07-08
Migrate string and tt text operations to ⊥/fallback, and unmask ⊥ causes

### LINKED ISSUE #5324: Migrate string.slice and ss.map.get off catchable error to ⊥/fallback
Part of #5261, continuing #5285 and #5322. Two more standard-library objects still raise the catchable `error` and can move to the ⊥/fallback model.

`string.slice` errors on a negative start/length, an out-of-range slice, and malformed UTF-8, all through `error`/`error cause` in `slice` and its `recidx`/`increase` helpers:

```
[start len] > slice
  ...
  error
    sprintf "Start index must be >= 0, but was %d" ...
```

Since the helpers can reference an enclosing void by name (as `rec-key-search` already references `hash`), a single `cant-slice` fallback declared on `slice` can cover every failure, yielding the caller's recovery or ⊥ when unbound — exactly like `bytes.slice`.

`ss.map` returns a `not-found` object whose `get` is an `error` when a key is missing:

```
[] > not-found
  false > exists
  error > get
    sprintf "Object by hash code %d ... does not exist" ...
```

Making `get` take a fallback on both the found item and `not-found` turns it into a get-or-default: `(m.found k).get d` returns the value or `d`, and `(m.found k).get` alone yields ⊥.

Out of scope: `string`'s length/decoding helper `inclen`, and the deferred `nk/socket`, `fs/file`, and `go` cases.

maxonfjvipon:
This clears the string and `tt` (text) families of the catchable `error` object, moving each failure onto the ⊥/fallback model the rest of the fragile-objects work (#5261) is converging on.

`map.get` now takes a fallback and returns it when the key is absent — a get-or-default. Left unbound the fallback is ⊥, so a missing key still terminates on use:

```
[cant-get] > get
  cant-get > @
    sprintf
      "Object by hash code %d from given key does not exists"
      * hash
```

The same fallback shape is given to the text operations whose failure is a clean top-level branch: `tt.at` gains an `out-of-bounds` fallback (mirroring `tuple.at`), and `tt.as-number`, `tt.nsplit` and `tt.repeated` gain `cant-parse` / `cant-split` / `cant-repeat`. Each is a get-or-default: pass a fallback and it is returned on failure, omit it and the operation terminates with ⊥.

`string.slice` and `string.length` instead terminate with ⊥ on every out-of-bounds or malformed-byte failure. Their failure is detected deep inside the recursive UTF-8 byte-walk whose result feeds later arithmetic, so a caller fallback cannot cleanly become the output — the same structural situation as `tuple.head`/`tuple.tail`, which already use ⊥. This clears `string.eo` of `error` entirely.

Verifying the slice change surfaced a pre-existing bug. A ⊥ carrying a cause (`T "why"`) is transpiled wrapped in `PhSafe`/`PhApplication`, so `AtWithRho`'s `instanceof PhTerminator` guard missed it, bound ρ onto it, and the strict put-by-name rule turned that into a panic that masked the real cause — `tuple.empty.head` reported *"does not accept attributes by name, but got 'ρ'"* instead of *"Can't get head from the empty tuple"*. The first commit fixes it at the source: ⊥ now silently ignores a ρ-binding (still rejecting any other name), and the now-redundant guard in `AtWithRho` is gone.

Still deferred: `string`/`tt` share nothing with the harder tail — `nk/socket` (built on its own `try`), `go` (error+try backtracking), and the `fs` family (its own syscall track).

Closes #5324.

### DIFF
--- a/eo-runtime/src/main/eo/ss/map.eo
+++ b/eo-runtime/src/main/eo/ss/map.eo
@@ -76,8 +76,9 @@
     # The returned object has two attributes: `exists` and `get`.
     # The `exists` attribute is either `true`, if object was found,
     # or `false` if was not.
-    # The `get` attribute returns either found object, or `error` if
-    # the object wasn't found.
+    # The `get` attribute takes a fallback and returns either the found
+    # object, or that fallback — the bottom object (⊥) when unbound — if
+    # the object was not found.
     # @todo #3251:30min Find a way to link hash code and index of key.
     #  Right now map is implemented as `tuple` of objects where every
     #  element is composition of three entities: hash, key and value.
@@ -107,16 +108,18 @@
               tup.head.hash.eq hash
               [] >>
                 true > exists
-                tup.head.value > get
+                [cant-get] > get
+                  tup.head.value > @
               not-found
         (rec-key-search tup.tail).@ > prev
       # Result when key is not found in map.
       [] > not-found
         false > exists
-        error > get
-          sprintf
-            "Object by hash code %d from given key does not exists"
-            * hash
+        [cant-get] > get
+          cant-get > @
+            sprintf
+              "Object by hash code %d from given key does not exists"
+              * hash
 
     # Returns the new `map` with added object
     # Replaces if there was one before.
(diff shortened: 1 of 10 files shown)


yegor256 (approved):

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (173 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+529](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+1703](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5328, opened 2026-07-08
Remove the go object and its goto backtracking

### LINKED ISSUE #5327: `go` object implementing goto-style control flow conflicts with EO principles
The `go` object (top-level `Q.go`) implements goto-style control flow — `go.to` runs a body with a jump token whose `backward`/`forward` are `error` throws caught by an enclosing `try`. It is control flow via exceptions, which does not belong in EO, so we should drop it entirely.

Removing it is self-contained. No other EO object depends on `go`; its only footprint beyond `eo-runtime/src/main/eo/go.eo` is dedicated machinery in `Dataized.take()` — a special case that re-throws `go.to.token.jump` errors without logging them (they are jumps, not failures) — plus the `DataizedTest.doesNotLogGoToTokenJump` test that covers it, and two `PhPackageTest` fixtures that merely use `go`/`EOgo` as a sample top-level object.

Proposed removal: delete `go.eo`; drop the `go.to.token.jump` branch in `Dataized.take()`; delete `doesNotLogGoToTokenJump`; and repoint the `PhPackageTest` fixtures at another top-level object such as `nop`. This also clears the last error-plus-try backtracking user in the runtime, a step toward eliminating `error` and `try` altogether (part of #5261).

maxonfjvipon:
The `go` object implemented goto-style control flow: `go.to` ran a body with a jump token whose `backward` and `forward` were `error` objects thrown and caught by an enclosing `try`. That is control flow through exceptions, which does not belong in EO, so this removes it entirely.

The removal is self-contained — no other EO object depends on `go`. Beyond `go.eo`, its only footprint was a dedicated special case in `Dataized.take()` that re-threw `go.to.token.jump` errors without logging them (they are jumps, not failures):

```java
if (enc.locator().startsWith(String.format("%s.go.to.token.jump:", PhPackage.GLOBAL))) {
    throw new EOerror.ExError(enc);
}
```

That branch and the `DataizedTest.doesNotLogGoToTokenJump` test that covered it are gone, and the two `PhPackageTest` fixtures that used `go`/`EOgo` merely as a sample top-level object now use `nop`/`EOnop`.

This also clears the last error-plus-try backtracking user in the runtime — a step toward eliminating `error` and `try` altogether (#5261).

While here, two stray test-only `try` uses go too. `tests-opens-posix-tcp-socket` used `try` only for finally-style cleanup — its `main` (`sd.gt 0`) cannot throw — so a `seq` that closes the socket and then asserts the descriptor is valid says the same thing. And `number`'s `tests-zero-division` wrapped `2.div 0` in `try` to assert it raises no error, but float division by zero is `+∞` and never throws, so the `catch` was dead; a `seq` that dataizes the division then returns `true` keeps the guard without `try`.

Closes #5327.


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (209 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1011](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5334, opened 2026-07-09
#5318: evaluate random.eo LCG step in 26-bit limbs to avoid i64 overflow

### LINKED ISSUE #5318: `ms/random.eo`'s LCG step saturates out-of-range product leading to a hard failure, blocking PR #5315
## What happens

`ms/random.eo`'s `next` (the linear congruential PRNG step) computes
`seed * 25214903917 + 11` in floating point, then converts the result via `.as-i64` so a bitwise
`and` can mask it down to 48 bits:

```eo
random > next
  as-number.
    as-i64.
      and.
        as-i64.
          plus.
            seed.times 25214903917
            11
        00-0F-FF-FF-FF-FF-FF-FF
```

For a `seed` in the range this generator actually produces (up to roughly `1 << 48`), the product
`seed * 25214903917` is vastly larger than `Long.MAX_VALUE` — verified directly:

```
seed=8682522807148012.0
product=2.1892897833939825E26
fits long: false
```

Before #5305/#5315, `number.as-i64` narrowed any `Double` to `long` via `Double.longValue()`,
which doesn't throw for an out-of-range value — it saturates to `Long.MAX_VALUE`. `random.eo`'s
`next`/`clocked` relies on this: the huge, out-of-range product is silently clamped to
`Long.MAX_VALUE` before the `and` mask is applied, and the algorithm happens to still produce
deterministic, in-range-after-masking output that satisfies `EOms.EOrandomTest`'s
"reproducible for the same seed" and "result is in range" checks — even though this doesn't match
the textbook Lehmer/Knuth LCG this code's own doc comment describes, since the true 83-ish-bit
intermediate product is never actually represented before masking.

#5305/#5315 tightened `number.as-i64` to reject a value outside `long` range with a domain error
instead of silently saturating (matching the intent already established for
`Expect.Natural`/`Expect.Int` in #5295). That fix is correct in isolation, but it turns this
existing, undocumented reliance on `Double.longValue()`'s saturation behavior into a hard failure:
`EOms.EOrandomTest.tests_random_is_in_range`, `tests_seeded_randoms_are_equal`, and
`tests_random_with_seed` now fail with `"the 'ρ' attribute (...) must fit into long range"`
instead of returning a (silently degraded) pseudo-random value — confirmed on CI for #5315:
https://github.com/objectionary/eo/actions/runs/28698516722

This blocks #5315 from being merged as-is (fixing `random.eo` in that PR would mix two unrelated
issues into one PR, against this project's one-PR-per-issue convention), so filing this
separately.

## What should happen

`random.eo`'s `next`/`clocked` LCG step should compute the masked 48-bit result without ever
needing to represent an out-of-`long`-range intermediate value — for example by performing the
multiply-add step in a way that stays within `long`/53-bit-exact-double range throughout (e.g. via
modular reduction before the multiplication, or an explicit big-integer-style split of the
multiplication into parts that individually fit), instead of relying on the previously-silent
saturation behavior of `number.as-i64` that #5305/#5315 correctly closes off.

morphqdd:
Closes #5318.

`ms/random.eo`'s `next` computed `seed * 25214903917 + 11` in one shot and relied on `number.as-i64` silently saturating the out-of-`long`-range product before the `and` mask. As #5305/#5315 tighten `as-i64` to reject out-of-range values, this reliance turned into a hard failure.

This rewrite evaluates the linear-congruential step in two 26-bit limbs (`hi`, `lo`) so no intermediate value ever leaves signed `i64` range (every partial product stays below `2 ^ 61`), while producing exactly the same masked 52-bit state — verified against the direct `(seed * 25214903917 + 11) mod 2 ^ 52` reference for the full 52-bit seed space. The incoming seed is first clamped to the generator's 52-bit state space so the step is safe for any seed.

The existing `EOms.EOrandomTest` cases (reproducibility for equal seeds, distinct sequence for `next`/`next.next`, results within `[0, 1)`, distinct clocked seeds) continue to pass, and the step no longer depends on `as-i64` saturation.

### DIFF
--- a/eo-runtime/src/main/eo/ms/random.eo
+++ b/eo-runtime/src/main/eo/ms/random.eo
@@ -18,17 +18,42 @@
   # Formula is based on linear congruential pseudorandom number generator, as defined by
   # D. H. Lehmer and described by Donald E. Knuth in The Art of Computer Programming, Volume 2,
   # Third edition: Seminumerical Algorithms, section 3.2.1.
-  # Magic numbers are taken from Java implementation. 48 lower bits are considered.
-  # `next` = (`seed` * 25214903917 + 11) & ((1 << 48) - 1).
-  # Here `00-0F-FF-FF-FF-FF-FF-FF` is pre calculated `(1 << 48) - 1`.
+  # Magic numbers are taken from Java implementation. Lower bits are considered.
+  # `next` = (`seed` * 25214903917 + 11) & `mask`, evaluated in two 26-bit limbs (`hi`
+  # and `lo`) so that no intermediate value leaves signed `i64` range, since the full
+  # product `seed` * 25214903917 is above `Long.MAX_VALUE` for a full-width `seed`.
+  # Here `00-0F-FF-FF-FF-FF-FF-FF` is the pre-calculated `mask` and
+  # `00-00-00-00-03-FF-FF-FF` is the low 26-bit limb mask `(1 << 26) - 1`.
   random > next
     as-number.
       as-i64.
         and.
-          as-i64.
+          as-bytes.
             plus.
-              seed.times 25214903917
-              11
+              plus.
+                as-i64.
+                  left.
+                    and.
+                      as-bytes.
+                        times.
+                          as-i64.
+                            and.
+                              as-bytes.
+                                div.
+                                  seed.as-i64
+                                  67108864.as-i64
+                              00-00-00-00-03-FF-FF-FF
+                          25214903917.as-i64
+                      00-00-00-00-03-FF-FF-FF
+                    26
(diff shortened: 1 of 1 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (37) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+756](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5336, opened 2026-07-09
Support a vertical-argument body on only-phi formations

### LINKED ISSUE #5331: Vertical application notation mishandles only-phi formation
I want to be able to write this:

```
foo > [x] > bar
  x.plus 42
```

Instead of this:

```
[x] > bar
  foo > @
    x.plus 42
```

Also, this:

```
xyz > [k] >>
  33
  k.minus 1
```

Instead of this:

```
[k] >>
  xyz > @
    33
    k.minus 1
```

maxonfjvipon:
Yegor asked (#5331) to be able to write `foo > [x] > bar` with the body hanging off it, instead of spelling out `[x] > bar` / `foo > @` / body by hand. This wires that up: when an only-phi line's φ carries no horizontal args, the φ stays open and deeper-indent lines attach to it as vertical arguments. The sugared and desugared forms produce identical XMIR.

```
foo > [x] > bar
  x.plus 42
```

A φ that already has horizontal args stays closed — there is nowhere to hang a body. And since the formation binds only φ, its arguments cannot be named; the ban follows the arguments down through nested applications but stops at a nested formation, where naming resumes as usual.

One incidental fix: `foo a > [x] > bar` with a body used to drop the body line silently, and now reports the same "closed for children" error as the bare case.

The parser spec is updated alongside the code (§3.10.1, §4.5, §5.2.3, §5.3.6, §6.1, §9.4, §9.9, Appendix A). Opened as a draft for review.

Closes #5331

## Validated end-to-end

Dogfooded in `eo-runtime`: the six `[m] >> / b.put > @` leaves in the `seq` tests now use `b.put > [m] >>`. The runtime transpiles, compiles, and runs green — `EOseqTest` stays 9/9.

Scope note: because an only-phi binds only `φ`, its arguments cannot be named. So the sugar fits leaf `φ` bodies whose arguments are unnamed (like these `seq` leaves), but not the `malloc.for`/`while` scope pattern, where the arguments are named `>>` formations and stay in the desugared form.

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -483,7 +483,7 @@ There are **four base forms** of name suffix (mutually exclusive on any one line
 | `> [params] > name` | Inline-phi formation with **explicit** name. |
 | `> [params] >>` | Inline-phi formation with **auto-generated** name (the right-hand side `>>` replaces `> name`). |
 
-R-3.10.1. Inline-phi mechanics: the LHS of the leftmost `> [` is the formation's body (assigned to `φ` of the inline-phi formation); the formation itself is named by the suffix on the right. The LHS must fit on the same line — it is horizontal by construction. The line is closed for deeper-indent children (§4.5).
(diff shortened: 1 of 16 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (494 >= 100); -8 for way too many hits-of-code (494 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+535](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5340, opened 2026-07-09
Make dataized a standalone atom and socket errors recoverable via fallbacks

### LINKED ISSUE #5339: `dataized` object's unhealthy dependence on `try` risking potential corruption
The `dataized` object — the primitive that forces dataization at the EO level, and the basis of the `!` caching syntax (`a > b!` becomes `(dataized a).as-bytes > b`) — is currently implemented in terms of `try`:

```
[target] > dataized
  try > @
    target
    error ex > [ex]
    01-
```

It relies on the fact that `try` dataizes its `main` argument; the `catch` is a dead no-op re-raise and the `finally` is a throwaway. This couples a fundamental primitive to `try`, which is a problem for the ongoing effort to eliminate `error` and `try` (part of #5261): deleting `try` later would corrupt `dataized` and, with it, every use of `!`.

`dataized` should stand on its own. Reimplement it as a Java atom (`[target] > dataized /bytes` with a hand-written `EOdataized`) that dataizes `target` once and returns the resulting `bytes` — behaviourally identical to today (the current `try` form already returns `Data.ToPhi` of the dataized bytes), just no longer dependent on `try`. This mirrors how `try` itself is an atom, and lets `try` be removed in the future without touching `dataized`.

maxonfjvipon:
`dataized` is the primitive that forces dataization at the EO level, and the basis of the `!` caching syntax (`a > b!` becomes `(dataized a).as-bytes > b`). It was implemented in terms of `try`, relying on the fact that `try` dataizes its `main` argument:

```
[target] > dataized
  try > @
    target
    error ex > [ex]    # dead no-op re-raise
    01-                # throwaway finally
```

That couples a fundamental primitive — and every use of `!` — to `try`, which is a problem for the effort to eliminate `error` and `try` (#5261): removing `try` later would corrupt `dataized`.

This makes `dataized` stand on its own. It is now a `/bytes` atom with a hand-written `EOdataized`:

```java
@Override
public Phi lambda() {
    return new Data.ToPhi(new Dataized(this.take("target")).take());
}
```

The behaviour is identical — the old `try` form also returned `Data.ToPhi` of the dataized bytes, and the `catch` (a re-raise) and `finally` (`01-`) were dead — so it still dataizes `target` once and returns the cached bytes. It just no longer depends on `try`, mirroring how `try` itself is an atom, and lets `try` be removed in the future without touching `dataized`.

The check-target-files smoke test pointed at `EOdataized.java` as its top-level generated-Java example; since `dataized` is no longer generated, it now points at `EOseq.java`.

## Recoverable socket errors

`nk/socket` was the last producer of the catchable `error` object — it raised one on every network failure across posix and win32. Those failures are recoverable, so each operation now takes a per-operation fallback (get-or-default), with both platforms routing their platform-specific diagnostic (`strerror` vs WSA code) into the same shared fallback:

| Operation | Fallback | Fires on |
|---|---|---|
| `connect` | `cant-connect` | socket-create, resolve, or connect failed |
| `listen` | `cant-listen` | socket-create, resolve, bind, listen, or WSAStartup failed |
| `accept` | `cant-accept` | accept failed |
| `send` | `cant-send` | send failed |
| `recv` | `cant-receive` | recv failed |

Setup failures fold into the entry operation via `safe-socket`, which now takes the fallback; `close` and `WSACleanup` happen during cleanup and terminate with ⊥. The RAII `try` that always closes the socket stays — its `finally` runs even on ⊥, so a returned fallback flows out while the socket is still closed (covered by `returnsFallbackWhenConnectionIsRefused`, which connects to a stopped port and gets the fallback back rather than terminating). This removes the last catchable `error` producers from the runtime.

Closes #5339.

### DIFF
--- a/eo-runtime/src/main/eo/dataized.eo
+++ b/eo-runtime/src/main/eo/dataized.eo
@@ -1,5 +1,7 @@
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
++rt jvm org.eolang:eo-runtime:0.0.0
++rt node eo2js-runtime:0.0.0
 +version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
@@ -35,12 +37,7 @@
 # Here, when `.as-bytes` is taken, the `dataized` dataizes (takes `bytes`)
 # from the object `foo` and returns them.
 # The `.as-bytes` is used to prevent double dataization.
-[target] > dataized
-  try > @
-    target
-    error ex > [ex]
-    01-
-
+[target] > dataized /bytes
   # Tests that dataized values are not recalculated on subsequent access.
   [] +> tests-dataized-does-not-do-recalculation
     eq. > @
--- a/eo-runtime/src/main/eo/nk/socket.eo
+++ b/eo-runtime/src/main/eo/nk/socket.eo
@@ -29,7 +29,8 @@
 #
 # When you dataize `socket.connect`, it creates new socket, connects to the given
 # IP on given port, dataizes provided scope, closes the socket and returns bytes
-# retrieved from scope dataization. The error is returned is any of the described steps failed.
+# retrieved from scope dataization. On failure at any of the described steps, the fallback of the
+# corresponding operation is returned (or the bottom object when it is left unbound).
 #
 # To listen to the incoming connections as server you should do:
 #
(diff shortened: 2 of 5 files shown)


yegor256 (approved):

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (173 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+567](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+1674](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5347, opened 2026-07-09
#5344: tolerate null PhSafe exception messages

### LINKED ISSUE #5344: PhSafe.through() leaks a non-ExError (NPE) when the caught throwable has a null message

## Summary
PhSafe.through(Supplier, String) is documented (lines 189-192) to guarantee that "No matter what happens inside the action, only an instance of EOerror.ExError may be thrown out of this method". The generic Throwable branch builds the enclosure with `new Data.ToPhi(ex.getMessage())`. When the caught throwable has a null message, Data.ToPhi(null) calls toPhi(null), whose final else-branch executes `obj.getClass().getCanonicalName()` on a null reference, throwing a NullPointerException. That NPE is thrown while evaluating the argument to `new EOerror.ExError(...)`, so it propagates out of the catch block and out of PhSafe entirely, instead of an EOerror.ExError. The ExAbstract branch (lines 209-212) has the same defect.

## Location
[`eo-runtime/src/main/java/org/eolang/PhSafe.java`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/PhSafe.java#L213) lines 213-217.

## Buggy code
```java
} catch (final Throwable ex) {
    throw new EOerror.ExError(
        new Data.ToPhi(ex.getMessage()),
        PhSafe.trace(ex, this.label(suffix))
    );
}
```

## Why it's wrong
The whole purpose of PhSafe is to convert every runtime failure into a catchable EOerror.ExError with source coordinates. A null-message throwable defeats that: the safety wrapper itself throws a raw NPE. Dataized.take() only catches EOerror.ExError, so this NPE bypasses EO error handling and crashes the program with a stack trace pointing at Data.ToPhi instead of the real fault. The project targets Java 11 (pom.xml <source>11</source>), where helpful NullPointerException messages (JEP 358, default from Java 15) do not exist, so a routine NPE inside dataization has getMessage()==null and triggers this. StackOverflowError (message is always null) from unbounded EO recursion triggers it regardless of Java version.

Verifier confirmation: Confirmed genuine. PhSafe.through(Supplier,String) documents (PhSafe.java:189-194) that only EOerror.ExError may escape, but the Throwable branch (PhSafe.java:213-217) and the ExAbstract branch (209-212) build the enclosure via new Data.ToPhi(ex.getMessage()). When the caught throwable has a null message, Data.ToPhi(null) → toPhi(null) (Data.java:53-54, 119-168) falls to the else branch at 161-167 and evaluates obj.getClass().getCanonicalName() on null, throwing a NullPointerException. Because Java evaluates that argument before the EOerror.ExError constructor runs, the NPE is raised inside the catch block, is not re-caught, and propagates out of through(), breaking the safety guarantee. Null messages are reachable: StackOverflowError.getMessage() is always null (version-independent; deep EO dataization recursion makes SOE realistic), and the project targets Java 11 (pom.xml:402-403) where implicit NPEs predate JEP-358 helpful messages and thus have null messages. ExAbstract (ExAbstract.java:29-40) also permits a null message. This is distinct from the already-filed issues and affects a different file. Severity is tempered because it only manifests on already-failing exception paths, degrading error reporting (a raw NPE with a misleading Data.java trace, uncatchable by EO error handling) rather than corrupting normal computation.

## Expected behavior
Any throwable caught by through() is wrapped into an EOerror.ExError; a null message should be tolerated (e.g. substituted with a placeholder string before Data.ToPhi, or Data.ToPhi should accept null).

## Suggested fix
Guard the message before constructing Data.ToPhi, e.g. `new Data.ToPhi(Objects.toString(ex.getMessage(), ex.getClass().getName()))`, in both the ExAbstract and Throwable branches; or make Data.ToPhi.toPhi(null) return a defined empty/string Phi.

Develop-KIM:
Fixes #5344

### Summary
- Guard null exception messages before wrapping them as EO strings in `PhSafe`.
- Add coverage for null-message `RuntimeException` and `ExAbstract` paths.

### Verification
- `mvn -pl eo-maven-plugin -am -DskipTests install`
- `mvn -pl eo-runtime -Dtest=PhSafeTest test`
- `mvn -pl eo-runtime -Pqulice -DskipTests test`
- `git diff --check`

### Disclosure
This contribution was prepared with AI assistance and reviewed before submission.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhSafe.java
+++ b/eo-runtime/src/main/java/org/eolang/PhSafe.java
@@ -7,6 +7,7 @@ package org.eolang;
 
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
 import java.util.function.Supplier;
 
 /**
@@ -207,15 +208,24 @@ public final class PhSafe implements Phi, Atom {
             throw new EOerror.ExError(ex, this.label(suffix));
         } catch (final ExAbstract ex) {
             throw new EOerror.ExError(
-                new Data.ToPhi(ex.getMessage()),
+                new Data.ToPhi(PhSafe.message(ex)),
                 this.label(suffix)
             );
         } catch (final Throwable ex) {
             throw new EOerror.ExError(
-                new Data.ToPhi(ex.getMessage()),
+                new Data.ToPhi(PhSafe.message(ex)),
                 PhSafe.trace(ex, this.label(suffix))
             );
         }
+    }
+
+    /**
+     * Exception message safe for EO dataization.
+     * @param exp The exception
+     * @return Message
+     */
+    private static String message(final Throwable exp) {
+        return Objects.toString(exp.getMessage(), exp.getClass().getName());
     }
 
     /**
--- a/eo-runtime/src/main/java/org/eolang/PhSafe.java
+++ b/eo-runtime/src/main/java/org/eolang/PhSafe.java
@@ -200,7 +200,7 @@ public final class PhSafe implements Phi, Atom {
      * @return Result
      * @checkstyle IllegalCatchCheck (20 lines)
      */
-    @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.PreserveStackTrace"})
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
     private <T> T through(final Supplier<T> action, final String suffix) {
         try {
             return action.get();
--- a/eo-runtime/src/test/java/org/eolang/PhSafeTest.java
+++ b/eo-runtime/src/test/java/org/eolang/PhSafeTest.java
@@ -59,6 +59,44 @@ final class PhSafeTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


yegor256 (approved):

yegor256:
@Develop-KIM thanks!

0crat:
@develop-kim Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+1682](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5361, opened 2026-07-09
Reject trailing non-suffix text

### LINKED ISSUE #5357: Suffix.parse doesn't raise error for trailing non-suffix text, causing incorrect parsing of invalid lines

## Summary
Suffix.parse skips leading spaces on the line tail and, if the remaining text is not `+>`, `>>`, or `>`, falls into the final `else` and returns Form.NONE (no suffix). The value/line-shape classes (LnApplication, LnCompactTuple, LnReversed, LnMethod, LnOnlyPhi) pass the entire un-consumed token tail into Suffix and never call tokens.atEnd() afterwards, so any non-whitespace trailing content that is not a suffix marker is silently discarded instead of being rejected as a syntax error. This is only exposed by heads with a well-defined end (INT/FLOAT/HEX/BYTES/STAR/STRING/ROOT/GROUP); identifier heads greedily consume to a NAME terminator so they don't hit it. The token stream after the head simply stops: readArgs breaks at the first non-space non-suffix char, readOuterBinding sees no ':', and Suffix returns NONE on the leftover.

## Location
[`eo-parser/src/main/java/org/eolang/parser/Suffix.java`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/Suffix.java#L231) lines 231-233.

## Buggy code
```java
} else if (tail.charAt(idx) == '>') {
    result = Suffix.named(tail, idx + 1, span, home);
} else {
    result = new Suffix.Parsed(Form.NONE, "", "", false);
}
```

## Why it's wrong
`5x` is not a valid EO expression; the trailing `x` must produce a parse error. Instead the parser emits the number 5.0 and drops `x` entirely. Compare Suffix.named (line 348) which calls Suffix.endsClean to reject 'trailing garbage after name suffix', and LnVoid/LnFormation which reject trailing garbage; the NONE path has no such guard.

Verifier confirmation: Confirmed at the code and behavioral level. In Suffix.parse (Suffix.java:217-235), when the space-skipped tail does not begin with `+>`, `>>`, or `>`, the final else at lines 231-233 returns Form.NONE, consuming nothing. The value/line-shape callers (LnApplication.into, LnApplication.java:61-69, plus LnCompactTuple/LnReversed/LnMethod/LnOnlyPhi) pass tokens.tail() into Suffix and never call tokens.atEnd() afterward, so any non-whitespace remainder that is not a suffix marker is silently dropped instead of raising a ParseError. This is only reachable for heads with a well-defined end (INT/FLOAT/HEX/BYTES/STAR/STRING/ROOT/GROUP), since identifier heads greedily consume to a NAME terminator. I reproduced the exact LnApplication.into token sequence (readValue -> readChain -> readArgs -> readOuterBinding -> new Suffix(tokens.tail(),...)) and observed for input `5x`: head=INTEGER "5", tail="[x]", atEnd=false, suffixForm=NONE, present=false, with no exception; likewise for `*x` (STAR, tail=[x]), `0x1Fzzz` (HEX 0x1F, tail=[zzz]), and `--x` (BYTES, tail=[x]). Eo.classify (Eo.java:356-370) plus applicative (398-408) route a digit/star/hex/bytes head to LnApplication, so the path is reachable end-to-end. The named() suffix path guards this exact condition via endsClean (line 348, "trailing garbage after name suffix"), but the NONE (and AUTO) paths have no such guard, and no caller verifies the stream is fully consumed. This is distinct from already-filed #5351: that issue concerns the loose `terminates` predicate inside test() for a present `+>` marker; `5x`/`*x` never enter test() and #5351's fix would not reject them. The defect is silent acceptance of syntactically invalid input (e.g. `5x` parses as 5.0), a genuine spec-conformance/correctness gap.

## Expected behavior
A line whose head/args are fully parsed but which has non-whitespace content that is not a valid suffix marker should raise a ParseError (trailing garbage), not be accepted with the extra token silently dropped.

## Suggested fix
In the final else of Suffix.parse (or in each caller after building the Suffix), when idx < tail.length() and the remaining text is not a recognised suffix, throw a ParseError for trailing garbage instead of returning Form.NONE; equivalently add a tokens.atEnd() check in the Ln* classes after the suffix is parsed.

Yurii201811:
## Summary

Rejects non-whitespace text left after an expression when it is not a recognized EO suffix. This prevents invalid lines such as `5x` from being silently accepted as the number `5`. A parser typo fixture verifies the error message and exact offending column.

## Background

`Suffix.parse` previously returned `Form.NONE` for any unrecognized tail. Several line-shape parsers do not perform another end-of-token check, so trailing input could be discarded silently.

Closes #5357

## Changes

- throw `ParseError` from the no-suffix branch when unconsumed non-whitespace text remains
- add an end-to-end `EoSyntaxTest` fixture for `5x`

## Testing

- `git diff --check` — passed
- focused Maven command could not run locally because this Mac has no Maven and only Java 8; repository CI provides the required Java 21 toolchain

## Additional Notes

The empty/whitespace tail path still returns `Form.NONE`; only a non-empty, unrecognized suffix tail is rejected.


Yurii201811:
Follow-up after the first CI pass is now on `3fb28bf`.

- The strict `Suffix.parse` rejection remains in place for trailing non-suffix text.
- Text-block closers now use an explicit optional-suffix path because method/argument tails after `"""` are intentionally deferred to a later parser iteration.
- The two existing typo fixtures whose invalid tails are now diagnosed earlier were updated to the more precise error positions.
- The Qulice declaration-order finding from the intermediate head was fixed.

Latest-head verification:
- macOS Java 23 `mvn`: all 1,439 `eo-parser` tests passed, including 875 `EoSyntaxTest` cases
- Qulice: `eo-parser` and `eo-maven-plugin` modules passed
- 18 independent static/security checks passed, including xcop, yamllint, SonarCloud, Codacy, Snyk, and typos

The remaining repository-wide `integration`, `snippets`, `qulice`, and Maven failures occur after the parser module and reproduce on the current base commit `1ad8cc8`: they fail in runtime/plugin code with `NoSuchMethodError: MnCsv(String)`. Base evidence: https://github.com/objectionary/eo/actions/runs/29042130546 and https://github.com/objectionary/eo/actions/runs/29042130733.

gemshrine (commented):
The tightened `else` branch in `parse()` closes the catch-all trailing-garbage path. The sibling `auto()` and `test()` paths still silently drop trailing text — see the inline comment on `Suffix.java:256` for concrete traces and a symmetrical fix.

eo-parser/src/main/java/org/eolang/parser/Suffix.java
```diff
@@ -229,11 +253,27 @@ private static Parsed parse(final String tail, final Span span, final int home)
         } else if (tail.charAt(idx) == '>') {
             result = Suffix.named(tail, idx + 1, span, home);
         } else {
-            result = new Suffix.Parsed(Form.NONE, "", "", false);
+            throw new ParseError(
```
gemshrine, line 229: The `else`-branch tighten is right, but the reject-trailing-garbage contract still leaks from the two sibling paths:

```
tail = ">> xyz"       parse() → auto() → returns Form.AUTO, "xyz" discarded
tail = "+> foo xyz"   parse() → test() → returns Form.TEST(label="foo"), "xyz" discarded
```

`named()` uses `endsClean(tail, rest, span, home)` at line 388 for exactly this. Mirroring that call at the end of `auto()` (using the existing `trailing` local) and `test()` (using its `idx` local) would close the class of bug uniformly across all three suffix forms.

eo-parser/src/test/java/org/eolang/parser/SuffixTest.java
```diff
@@ -34,6 +34,17 @@ void reportsNoneWhenTailHasOnlyWhitespace() {
...
     }
 
+    @Test
+    void leavesDeferredNonSuffixTailAbsent() {
+        MatcherAssert.assertThat(
+            "an optional suffix must leave a deferred method chain untouched",
+            Suffix.optional(
+                ".as-bytes", new Span("\"\"\".as-bytes", 1), 3
+            ).form(),
+            Matchers.equalTo(Suffix.Form.NONE)
```
gemshrine: Two companion tests would lock in the same class of bug across the sibling forms (see comment on `Suffix.java:256`):

```java
@Test
void rejectsTrailingGarbageAfterAuto() {
    Assertions.assertThrows(
        ParseError.class,
        () -> new Suffix(">> garbage", new Span("5 >> garbage", 1), 2)
    );
}

@Test
void rejectsTrailingGarbageAfterTest() {
    Assertions.assertThrows(
        ParseError.class,
        () -> new Suffix("+> foo garbage", new Span("[] +> foo garbage", 1), 3)
    );
}
```

Both currently return successfully, silently dropping `garbage`.

eo-parser/src/test/resources/org/eolang/parser/eo-typos/trailing-non-suffix.yaml
```diff
@@ -0,0 +1,11 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+line: 2
+message: |-
+  [2:1] error: 'trailing garbage after expression'
```
gemshrine, line 6: Two parallel fixtures for the `>>` and `+> name` forms would prove the contract holds symmetrically:

```yaml
input: |
  # No comments
  5 >> garbage
```

and

```yaml
input: |
  # No comments
  [] +> foo garbage
```

Both today parse cleanly with `garbage` silently dropped; they'd start throwing once `endsClean` is wired into `auto()` and `test()` (see comment on `Suffix.java:256`).

eo-parser/src/main/java/org/eolang/parser/Suffix.java
Yurii201811: Fixed in `d52ce88` (with the project-convention test rename in `d418d17`). Both `test()` and `auto()` now call `Suffix.endsClean` after consuming their valid suffix content, so `+> foo garbage` and `>> garbage` are rejected instead of silently truncating the tail.

eo-parser/src/test/java/org/eolang/parser/SuffixTest.java
Yurii201811: Added the two companion unit cases in `SuffixTest`: one for trailing content after auto naming and one for trailing content after the plus-greater suffix. The latest Java 23 Maven run passed all 1,443 `eo-parser` tests.

eo-parser/src/test/resources/org/eolang/parser/eo-typos/trailing-non-suffix.yaml
Yurii201811: Added both requested end-to-end typo fixtures: `trailing-non-suffix-auto.yaml` and `trailing-non-suffix-test.yaml`. The latest run passed all 877 `EoSyntaxTest` fixtures, and the separate `typos` check is green.

Yurii201811:
Follow-up pushed in `4c77472` for the two Sonar `java:S5778` findings.

The `Span` setup now happens before each `assertThrows` call, leaving only the expected `Suffix` constructor inside the exception lambda.

Verification on Java 23 / Maven 3.9.11:
- `mvn -pl eo-parser -Dtest=SuffixTest test --errors --batch-mode` (31/31)
- `mvn -pl eo-parser -DskipTests -Pqulice verify --errors --batch-mode`
- `git diff --check`

All passed.

yegor256:
@Yurii201811 build is broken here and a few merge conflicts

Yurii201811:
Resolved the merge conflict against current `master` in `06fdbe2b5f`.

The conflict was in `LnTextBlock`: upstream now consumes the same-line method chain with `Tokens.readChain()` before parsing the suffix, so I kept that upstream behavior and removed this branch’s now-obsolete optional-suffix workaround. The strict trailing-garbage checks and their fixtures remain.

Local verification (Maven 3.9.16 / Java 26):
- `mvn -pl eo-parser test --errors --batch-mode` — 1,447 tests passed, 26 skipped
- `mvn -pl eo-parser -DskipTests -Pqulice verify --errors --batch-mode` — passed
- `git diff --check` — passed

The refreshed GitHub matrix is now running, including the repository’s Java 23 jobs.

0crat:
@yurii201811 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+75](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5364, opened 2026-07-09
#5342: fix i32.times overflow — always take low 32 bits

### LINKED ISSUE #5342: i32.times method incorrectly adds high 32 bits on overflow instead of discarding

## Summary
i32.times computes the exact 64-bit product `bts = (as-i64.times x.as-i64).as-bytes`, splits it into `left` (high 4 bytes, slice 0 4) and `right` (low 4 bytes, slice 4 4). When `left` is 00-00-00-00 or FF-FF-FF-FF it correctly returns `i32 right` (the low 32 bits). Otherwise it returns `(i32 left).plus (i32 right)` — adding the high half into the result. Wrapping i32 multiplication must yield only the low 32 bits (`i32 right`); adding `left` is incorrect.

## Location
[`eo-runtime/src/main/eo/i32.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/i32.eo#L59) lines 59-70.

## Buggy code
```
[x] > times
  if. > @
    or.
      left.eq 00-00-00-00
      left.eq FF-FF-FF-FF
    i32 right
    plus.
      i32 left
      i32 right
  (as-i64.times x.as-i32.as-i64).as-bytes > bts
  bts.slice 0 4 > left
  bts.slice 4 4 > right
```

## Why it's wrong
Two's-complement (Java-style) i32 multiplication wraps to the low 32 bits of the true product. The test tests-i32-times-with-overflow (lines 320-323) confirms wrapping semantics (2147483647*2 = -2). For products whose magnitude reaches >= 2^31 the high 4 bytes are neither all-zero nor all-FF, so the else branch fires and pollutes the result with the high word.

Verifier confirmation: Confirmed against /workspace/eo/eo-runtime/src/main/eo/i32.eo (latest master, "Already up to date"). `times` (lines 59-70) forms the exact 64-bit product `bts = (as-i64.times x.as-i32.as-i64).as-bytes`, then `left = bts.slice 0 4` (high word) and `right = bts.slice 4 4` (low word). It returns `i32 right` only when the high word is 00-00-00-00 or FF-FF-FF-FF; otherwise it returns `(i32 left).plus (i32 right)`. Correct two's-complement wrapping i32 multiplication (which the overflow test at lines 320-323, 2147483647*2 = -2, establishes as the intended semantics) is always just the low 32 bits, i.e. `i32 right`; adding the high word `left` is wrong. Unlike `plus`/`minus` (lines 74-89), whose 33-bit sum always sign-extends so the high word is unavoidably all-0 or all-FF and the else branch is dead, `times` genuinely reaches the else branch. Verified traces: 65536*65536 = 0x100000000 → left=00-00-00-01, right=00-00-00-00 → returns i32(1).plus i32(0) = 1 instead of the correct 0; 100000*100000 = 0x2540BE400 → left=00-00-00-02 → returns 1410065410 instead of the correct 1410065408. The pattern was clearly copied from `plus` without removing the high-word addition. This is a distinct bug (i32.eo, not on the filed list, which covers i64/BytesRaw/Emissions/number/fs/Sha).

## Expected behavior
For any i32 operands, times returns the low 32 bits of the product interpreted as a signed i32 (equal to `i32 right`), regardless of the high word.

## Suggested fix
Return `i32 right` unconditionally (the low 32 bytes), dropping the `left`-dependent branch, or otherwise ensure the high word is discarded rather than added.

gemshrine:
## Summary

Closes #5342. The else branch of the if. guard in i32.times added the high 4 bytes of the 64-bit product into the result whenever those bytes weren't a sign-extension mask, polluting the wrapping-multiplication semantics.

Two's-complement wrap for i32*i32 is always `low 32 bits of the 64-bit product` — no branch on the high half is needed. The method now evaluates directly to `i32 (as-i64.times x.as-i32.as-i64).as-bytes.slice 4 4`.

## Why the existing test didn't catch it

`tests-i32-times-with-overflow` uses `MAX_INT * 2 = -2`, whose i64 encoding is `00-00-00-00-FF-FF-FF-FE` — the `left.eq 00-00-00-00` branch fires, so the buggy `i32 left plus i32 right` never runs.

## New tests

* `tests-i32-times-wraps-modulo-2-pow-32`: `65536 * 65536 = 2^32`, i64 = `00-00-00-01-00-00-00-00`. The high half is non-trivial; with the old code this returned `1` instead of the correct wrap `0`.
* `tests-i32-times-wraps-large-positive-product`: `100000 * 100000 = 10_000_000_000`, i64 = `00-00-00-02-54-0B-E4-00`. Old code returned `1_410_065_410` (adding the high `2`) instead of `1_410_065_408`.

### DIFF
--- a/eo-runtime/src/main/eo/i32.eo
+++ b/eo-runtime/src/main/eo/i32.eo
@@ -56,18 +56,11 @@
 
   # Multiplication of `$` and `x`.
   # Here `x` must be an `org.eolang.i32` object.
+  # Two's-complement wrap: the result is the low 32 bits of the true
+  # 64-bit product, whatever the high half looks like.
   [x] > times
-    if. > @
-      or.
-        left.eq 00-00-00-00
-        left.eq FF-FF-FF-FF
-      i32 right
-      plus.
-        i32 left
-        i32 right
-    (as-i64.times x.as-i32.as-i64).as-bytes > bts
-    bts.slice 0 4 > left
-    bts.slice 4 4 > right
+    i32 right > @
+    (as-i64.times x.as-i32.as-i64).as-bytes.slice 4 4 > right
 
   # Sum of `$` and `x`.
   # Here `x` must be an `org.eolang.i32` object.
@@ -321,3 +314,20 @@
     eq. > @
       2147483647.as-i64.as-i32.times 2.as-i64.as-i32
       -2.as-i64.as-i32
+
+  # Tests that i32 multiplication wraps modulo 2^32 when the true
+  # product's high 4 bytes are neither all-zero nor all-FF. Here
+  # 65536 * 65536 = 2^32, whose i64 encoding is 00-00-00-01-00-00-00-00;
+  # the correct two's-complement wrap discards the high half and yields 0.
+  [] +> tests-i32-times-wraps-modulo-2-pow-32
+    eq. > @
+      65536.as-i64.as-i32.times 65536.as-i64.as-i32
+      0.as-i64.as-i32
+
+  # Tests that i32 multiplication of two mid-range positive values
+  # wraps to the low 32 bits: 100000 * 100000 = 10_000_000_000
+  # = 0x2_540B_E400 (as i64), which as i32 is 1_410_065_408.
+  [] +> tests-i32-times-wraps-large-positive-product
+    eq. > @
+      100000.as-i64.as-i32.times 100000.as-i64.as-i32
+      1410065408.as-i64.as-i32


gemshrine:
CI red here is not from this change — every job that fails hits the same pre-existing infrastructure defect, tracked at #5365 (tojos 0.19.1 vs org.eolang.lints `ReservedNames` calling a removed `MnCsv(String)` constructor). I'll rebase and re-push once master is unblocked so CI re-runs against green infra.

yegor256:
@gemshrine master is clean now, please update your branch

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+52](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5365, opened 2026-07-09
Build failure in `eo-runtime` due to `NoSuchMethodError` after bumping tojos to version 0.19.1


gemshrine:
## Summary

Since [`5f568a2`](https://github.com/objectionary/eo/commit/5f568a2) ("build(deps): bump tojos to 0.19.1"), every PR's `mvn`, `integration`, `snippets`, and `qulice` jobs fail at the `eo-maven-plugin:compile` phase of `eo-runtime` with:

```
Caused by: java.lang.NoSuchMethodError: 'void com.yegor256.tojos.MnCsv.<init>(java.lang.String)'
    at org.eolang.lints.ReservedNames.lambda$new$2 (ReservedNames.java:38)
    ...
    at org.eolang.lints.MonoLints.<clinit> (MonoLints.java:24)
    at org.eolang.lints.PkMono.<clinit> (PkMono.java:33)
    at org.eolang.lints.Source.<clinit> (Source.java:27)
    at org.eolang.maven.Linting.linted (Linting.java:432)
```

`org.eolang.lints:0.x` still calls a `String`-arg `MnCsv` constructor from `ReservedNames`; that constructor is no longer resolvable at runtime against `com.yegor256:tojos:0.19.1`, so `Source.<clinit>` throws, every `MjLint`/`MjCompile` linting task hits `NoClassDefFoundError`, and the reactor build fails at `eo-runtime`.

## Master `mvn.yml` history

The bisect is unambiguous from the master workflow history:

| when (UTC) | commit | title | mvn.yml |
| --- | --- | --- | --- |
| 2026-07-09 18:34 | [`b7b522f`](https://github.com/objectionary/eo/commit/b7b522f) | Merge #5164 (typos renovate) | success |
| 2026-07-09 18:41 | [`5f568a2`](https://github.com/objectionary/eo/commit/5f568a2) | `build(deps): bump tojos to 0.19.1` | **failure** |
| 2026-07-09 18:50 | [`e9b8154`](https://github.com/objectionary/eo/commit/e9b8154) | Merge #5355 (`renovate/org.eolang-lints-0.x`) | failure |
| 2026-07-09 18:50 | [`1ad8cc8`](https://github.com/objectionary/eo/commit/1ad8cc8) | Merge #5354 (`renovate/logback-monorepo`) | failure |

[...]

PR #5367, opened 2026-07-09
#5366: fix i16.times overflow — always take low 16 bits

### LINKED ISSUE #5366: i16.times method improperly handles overflow by adding high 16 bits instead of dropping them
## Summary

`i16.times` computes the exact 32-bit product `bts = (as-i32.times x.as-i16.as-i32).as-bytes` and splits it into `left` (high 2 bytes, `slice 0 2`) and `right` (low 2 bytes, `slice 2 2`). When `left` is `00-00` or `FF-FF` it correctly returns `i16 right` (the low 16 bits). Otherwise it returns `(i16 left).plus (i16 right)` — folding the high word into the result. Two's-complement wrapping i16 multiplication must yield only the low 16 bits (`i16 right`); adding `left` is incorrect.

This is the direct sibling of #5342 (`i32.times` overflow) on the same code shape.

## Location

[`eo-runtime/src/main/eo/i16.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/i16.eo#L44) lines 44–55.

## Buggy code

```
[x] > times
  if. > @
    or.
      left.eq 00-00
      left.eq FF-FF
    i16 right
    plus.
      i16 left
      i16 right
  (as-i32.times x.as-i16.as-i32).as-bytes > bts
  bts.slice 0 2 > left
  bts.slice 2 2 > right
```

## Why it's wrong

Two's-complement (Java-style) i16 multiplication wraps to the low 16 bits of the true product. When the product's magnitude reaches ≥ 2<sup>15</sup> and its 32-bit encoding's high 2 bytes are neither all-zero (a small non-negative overflow) nor all-`FF`, the else branch fires and folds the high word into the result — a value bigger than the true wrap by that folded amount.

## Reproducer

`256 * 256 = 65536`. The 32-bit product is `0x0001_0000`, so `left = 00-01`, `right = 00-00`.

* Correct two's-complement wrap: `65536 mod 2^16 = 0`.
* Current code: `left.eq 00-00` is false, `left.eq FF-FF` is false — the else branch runs and returns `(i16 00-01).plus (i16 00-00) = 1`.

EO expression:

```
256.as-i64.as-i32.as-i16.times 256.as-i64.as-i32.as-i16
```

should yield `0.as-i64.as-i32.as-i16` and yields `1.as-i64.as-i32.as-i16` instead.

## Why the existing test doesn't catch it

`tests-i16-times-with-overflow` at line 272 uses `32767 * 2 = 65534`, whose 32-bit encoding is `0x0000_FFFE` — the `left.eq 00-00` branch fires, so the buggy `plus` never runs and the test doesn't reveal the defect. Same blind spot as the analogous `tests-i32-times-with-overflow` at `i32.eo:320` (which fails to catch #5342 for the same reason).

## Suggested fix

Drop the branch and always take the low half — the same shape #5364 is applying to `i32.times`:

```
[x] > times
  i16 right > @
  (as-i32.times x.as-i16.as-i32).as-bytes.slice 2 2 > right
```

## Sibling check

`i16.plus`, `i16.div`: same skeleton but the else branch is dead — the results of `i16 + i16` and `i16 / i16` as `i32` always have their high 2 bytes as sign-extension bits, so the guard's true branch always fires. Only `times` observably breaks. Same relative status as in `i32.eo`.

gemshrine:
## Summary

Closes #5366. Direct sibling of #5342/#5364 (`i32.times`) on the same code shape: the else branch of the `if.` guard in `i16.times` folded the high 2 bytes of the 32-bit product into the result whenever those bytes weren't a sign-extension mask, polluting the wrapping-multiplication semantics.

Two's-complement wrap for `i16 * i16` is always **low 16 bits of the 32-bit product** — no branch on the high half is needed. The method now evaluates directly to `i16 (as-i32.times x.as-i16.as-i32).as-bytes.slice 2 2`.

## Why the existing test didn't catch it

`tests-i16-times-with-overflow` uses `32767 * 2 = 65534`, whose i32 encoding is `00-00-FF-FE` — the `left.eq 00-00` branch fires, so the buggy `i16 left plus i16 right` never runs.

## New tests

* `tests-i16-times-wraps-modulo-2-pow-16`: `256 * 256 = 2^16`, i32 = `00-01-00-00`. Old code returned `1` (folding the high `1`); correct wrap is `0`.
* `tests-i16-times-wraps-large-positive-product`: `300 * 300 = 90000`, i32 = `00-01-5F-90`. Old code returned `24465` (folding `1`); correct wrap is `24464` (`0x5F90`).

### DIFF
--- a/eo-runtime/src/main/eo/i16.eo
+++ b/eo-runtime/src/main/eo/i16.eo
@@ -41,18 +41,11 @@
 
   # Multiplication of `$` and `x`.
   # Here `x` must be an `org.eolang.i16` object.
+  # Two's-complement wrap: the result is the low 16 bits of the true
+  # 32-bit product, whatever the high half looks like.
   [x] > times
-    if. > @
-      or.
-        left.eq 00-00
-        left.eq FF-FF
-      i16 right
-      plus.
-        i16 left
-        i16 right
-    (as-i32.times x.as-i16.as-i32).as-bytes > bts
-    bts.slice 0 2 > left
-    bts.slice 2 2 > right
+    i16 right > @
+    (as-i32.times x.as-i16.as-i32).as-bytes.slice 2 2 > right
 
   # Sum of `$` and `x`.
   # Here `x` must be an `org.eolang.i16` object.
@@ -273,3 +266,20 @@
     eq. > @
       32767.as-i64.as-i32.as-i16.times 2.as-i64.as-i32.as-i16
       -2.as-i64.as-i32.as-i16
+
+  # Tests that i16 multiplication wraps modulo 2^16 when the true
+  # product's high 2 bytes are neither all-zero nor all-FF. Here
+  # 256 * 256 = 2^16, whose i32 encoding is 00-01-00-00; the correct
+  # two's-complement wrap discards the high half and yields 0.
+  [] +> tests-i16-times-wraps-modulo-2-pow-16
+    eq. > @
+      256.as-i64.as-i32.as-i16.times 256.as-i64.as-i32.as-i16
+      0.as-i64.as-i32.as-i16
+
+  # Tests that i16 multiplication of two mid-range positive values
+  # wraps to the low 16 bits: 300 * 300 = 90000 = 0x0001_5F90 as i32,
+  # which as i16 is 24464.
+  [] +> tests-i16-times-wraps-large-positive-product
+    eq. > @
+      300.as-i64.as-i32.as-i16.times 300.as-i64.as-i32.as-i16
+      24464.as-i64.as-i32.as-i16


gemshrine:
CI failures here (mvn/qulice/integration/snippets) are the same pre-existing infra defect as #5364 tracked at #5365 — the build fails at `eo-maven-plugin:compile` on `eo-runtime` because `org.eolang.lints.ReservedNames` calls a removed `MnCsv(String)` constructor after tojos 0.19.1. I'll rebase once master is unblocked so CI reruns against green infra.

yegor256:
@gemshrine master is clean now, please update your branch

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+64](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5376, opened 2026-07-10
File/Dir IO in `fs` surfaces recoverable failures as catchable errors due to Exception-based errors


maxonfjvipon:
The `fs` file/dir objects do their IO through java-NIO atoms (`org.eolang.EOfs`, backed by `Files.java`). Every one of those atoms catches `IOException` and rethrows it as an `IllegalArgumentException`, which `PhSafe` then wraps into a catchable `EOerror.ExError` (`PhSafe.java` catches `Throwable`). So an ordinary, recoverable IO failure — file not found, permission denied — surfaces today as a catchable `error`. That is exactly the execution-interrupting `error` the fragile-objects epic (#5261) is eliminating, and it is the last real source of catchable `error` from atoms.

The fix is to back `file`/`dir` with syscalls, the way `nk/socket` and `io/console` already are: `posix`/`win32` calls that return a code (`-1` on failure via `errno`/`strerror`) instead of throwing. IO failures then become predictable, code-based results the EO layer can turn into per-operation fallbacks — recoverable, no exceptions, and no catchable `error`.

The posix machinery is mostly present already. `CStdLib` declares `open`, `read`, `write`, `close`; `ReadSyscall`/`WriteSyscall`/`CloseSyscall` exist and are registered in `EOposix$EOφ.SYS_CALLS` (used by console and sockets). Only `OpenSyscall` is missing.

Plan, incrementally:

1. **POSIX file IO** (this slice): add `OpenSyscall` (wraps `CStdLib.open`), register `open`, and rewrite `file.eo`'s `open`/`file-stream`/`read`/`write`/`close` to be descriptor-based via `posix "open"/"read"/"write"/"close"`, with `cant-open`/`cant-read`/`cant-write` fallbacks driven by `code == -1`. Retire the name-keyed `Files.java` stream registry on posix.
2. **stat-based checks**: back `exists`/`is-directory`/`size` with `stat` (note `struct stat` layout differs across Linux/macOS).
3. **dir**: back `dir.walk`/`mkdir`/`tmpfile` with `opendir`/`readdir`/`mkdir`.
4. **win32**: Kernel32 parity for all of the above.
5. Migrate the remaining EO-level validation `error`s (`file.open` wrong mode, `file.read`/`write` wrong mode, `dir.open` opening a directory as a file) to `⊥`/fallback, since those are pure logic and not IO.

Starting with the POSIX file slice.

0crat:
[...]

PR #5377, opened 2026-07-10
Back file.open with OS syscalls on posix and windows (fs→syscall migration)

maxonfjvipon:
The `fs` file atoms catch `IOException` and rethrow it as an `IllegalArgumentException`, which `PhSafe` wraps into a catchable `EOerror.ExError` — so a plain file-not-found surfaces today as a catchable `error`. Backing `file.open` with syscalls (a code, `-1` on failure, no exceptions) is how `nk/socket` and `io/console` already work, and it makes IO failures predictable and recoverable.

`file.open` now runs its whole read/write cycle through the `open`/`read`/`write`/`close` syscalls on both posix and windows, dropping the java-fs stream it relied on. That stream exposed two independent cursors — an `InputStream` from the start and an appending `OutputStream` — so the syscall version threads two descriptors, keeping reads at zero regardless of writes:

```
code. > read-fd
  syscall "open" (* path syscall.open-read-flags)
code. > write-fd
  syscall "open" (* path syscall.open-write-flags)
```

On windows the descriptors come from the CRT (`msvcrt`) `_open`/`_read`/`_write`/`_close`, which hand back the same small ints as libc; sockets stay on Winsock. The write syscall now takes a `byte[]` buffer instead of a `String`, so writes are binary-safe. With the stream gone, the `process-file`, `file-stream`, `chunk` and `written-bytes` atoms and their `Files` backing class are removed. The open flags live as named `open-read-flags`/`open-write-flags` constants on `posix` and `win32`.

`file.open` also stops raising the `error` object. A file that must exist but does not, or an open syscall that fails, applies a caller-bound `cant-open` fallback to the reason and terminates with a ⊥ when it is left unbound — the same shape as `map.get`. Misuse — an unknown mode, reading a write-only handle, writing a read-only one — terminates with a ⊥ directly.

Create and truncate still lean on the `touch`/`delete` atoms; migrating those, along with `size` and `move`, comes in later slices (they need cross-platform directory-removal handling). Posix is validated locally, windows on the CI.

Part of #5376.


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1060 >= 100); -8 for way too many hits-of-code (1060 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+539](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5383, opened 2026-07-10
#5335: check HTTP status in OyRemote.get() to avoid caching throttled responses

### LINKED ISSUE #5335: `OyRemote.get()` returns and caches erroneous HTTP 429 response as object source
In `eo-maven-plugin/src/main/java/org/eolang/maven/OyRemote.java:74-93`, `get()` returns `this.send(url, ...).body()` without ever inspecting the HTTP status code. When `raw.githubusercontent.com` answers a burst of object pulls with `429 Too Many Requests`, the error-page body (the literal text `429: Too Many Requests`) is handed back as if it were the object's `.eo` source.

That body is then persisted into the pulled object file and stored in the global cache under `~/.eo/pulled/<hash>/…`, so every later build reuses the poisoned file and fails with a misleading parse error such as `[1:4] error: 'expected binding label after :'` (the parser choking on `429:`). Because the poison is cached, the failure is sticky and survives long after GitHub has recovered. Note that `exists()` in the same class (OyRemote.java:105-114) already guards against this by detecting `429`/`408` and throwing to retry, but `get()` has neither that status check nor `@RetryOnFailure`.

The smallest fix is to check the status code in the supplier lambda at `OyRemote.java:81` and throw an `IOException` for any non-2xx response (mirroring the `429`/`408` handling already in `exists()` at line 108), so a rate-limited response is never returned or cached as source. Adding `@RetryOnFailure` to the fetch would additionally let a transient 429 recover on its own.

morphqdd:
Closes #5335.

`OyRemote.get()` (eo-maven-plugin) returned `this.send(url, ...).body()` without ever inspecting the HTTP status code. When `raw.githubusercontent.com` throttles a burst of object pulls with `429 Too Many Requests`, the error-page body was handed back as if it were the object's `.eo` source, then persisted into the pulled-object cache — so every later build reused the poisoned file and failed with a misleading parse error, long after GitHub had recovered.

`exists()` in the same class already guarded against this (429/408 detection + retry), but `get()` had neither the status check nor `@RetryOnFailure`.

Fix: extracted a `fetch(URL)` helper that checks the response status the same way `exists()` does — 429/408 throw and retry via `@RetryOnFailure`, any other non-2xx throws immediately — so a throttled or failed response is never returned or cached as source. `get()`'s supplier now calls `fetch(url)` instead of reading the body unconditionally.

Added `OyRemoteTest.doesNotReturnThrottledResponseBodyAsSource`, which stands up a local `HttpServer` that always answers `429` with the literal throttling body text, and asserts `get(...).stream()` throws `IOException` instead of returning that body. Verified the test fails without the fix (reverting only the `get()` lambda back to the unconditional `.body()` call reproduces `AssertionFailedError: ... nothing was thrown`), and passes with it.

Added a test-only constructor `OyRemote(UrlOy, UrlOy, Proxy...)` (mirroring the existing "Ctor for testing" pattern already used by `OyRemote.UrlOy`) so the test can point `OyRemote` at the local stub server instead of hitting the real Objectionary endpoints.

Verification:
- `mvn -pl eo-maven-plugin test -Dtest=OyRemoteTest`: 9 tests, 0 failures, 0 errors (includes the pre-existing live-network tests).

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/OyRemote.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/OyRemote.java
@@ -7,6 +7,7 @@ package org.eolang.maven;
 import com.jcabi.aspects.RetryOnFailure;
 import com.jcabi.log.Logger;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+772](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5384, opened 2026-07-10
#5329: use overflow-safe long arithmetic in bytes.concat with cant-concat fallback

### LINKED ISSUE #5329: `EObytes$EOconcat` throws a `NegativeArraySizeException` for large concatenation due to integer overflow
## What happens

`EObytes$EOconcat` (`eo-runtime/src/main/java/org/eolang/EObytes$EOconcat.java`) allocates the destination array using an unchecked `int` addition:

```java
final byte[] current = new Dataized(this.take(Phi.RHO)).take();
final byte[] provided = new Dataized(this.take("b")).take();
final byte[] dest = new byte[current.length + provided.length];
System.arraycopy(current, 0, dest, 0, current.length);
System.arraycopy(provided, 0, dest, current.length, provided.length);
```

When `current.length + provided.length` exceeds `Integer.MAX_VALUE`, the `int` addition overflows to a negative value, and `new byte[negative]` throws a raw `java.lang.NegativeArraySizeException` that escapes the atom as an uncaught JVM exception, instead of the clean `ExFailure` the runtime raises everywhere else for out-of-bounds conditions.

This is the same class of overflow already fixed in `Heaps.fits` (#5296) and `EObytes$EOslice` (#5298), where the bounds arithmetic was widened to `long`.

## What should happen

`bytes.concat` should compute the combined length with overflow-safe arithmetic (e.g. `(long) current.length + provided.length`) and, when the result exceeds `Integer.MAX_VALUE`, raise a clean `ExFailure` describing the too-large concatenation, rather than letting a raw `NegativeArraySizeException` leak.

morphqdd:
Closes #5329.

`EObytes$EOconcat` allocated the destination array via unchecked `int` addition (`new byte[current.length + provided.length]`). When the combined length exceeds `Integer.MAX_VALUE`, the `int` sum overflows to a negative value and `new byte[negative]` throws a raw `NegativeArraySizeException`, escaping the atom as an uncaught JVM exception instead of the clean, caller-recoverable failure the runtime uses everywhere else for out-of-bounds/overflow conditions (same class already fixed in `Heaps.fits`/`Heaps.write` (#5296/#5297) and `EObytes$EOslice` (#5298)).

Fix, matching `bytes.slice`'s existing `cant-slice` idiom exactly:
- `bytes.eo`: `concat` now declares a `cant-concat` error-branch void (`? > cant-concat /{string}`), documented the same way as `slice`.
- `EObytes$EOconcat.java`: checks `(long) current.length + provided.length <= Integer.MAX_VALUE` before allocating; on overflow, returns the caller-supplied `cant-concat` fallback (or the bottom object if none was given) with a descriptive message, instead of ever calling `new byte[...]` with a value that could be negative.

Existing callers are unaffected — `cant-concat` is only touched on the overflow path, so all current single-arg `.concat` call sites (`concat.` `,` `a.concat b`, etc.) keep working exactly as before.

Added `EObytesEOconcatTest.reportsCleanFailureInsteadOfOverflowingArraySize`, which concatenates two real ~1GiB (`2^30`) byte arrays (summing to `Integer.MAX_VALUE + 1`, one byte over the boundary — the largest per-array size the JVM allows is close to `Integer.MAX_VALUE` itself, so this is the minimal split that reaches the overflow without hitting the JVM's own "array size exceeds VM limit" ceiling) and asserts the `cant-concat` fallback fires instead of throwing. Verified the test fails without the fix — reverting only the overflow guard reproduces `NegativeArraySizeException: -2147483648` at exactly the reported line — and passes with it.

Verification:
- `mvn -pl eo-runtime test -Dtest=EObytesEOconcatTest`: 2 tests, 0 failures, 0 errors.

### DIFF
--- a/eo-runtime/src/main/eo/bytes.eo
+++ b/eo-runtime/src/main/eo/bytes.eo
@@ -114,11 +114,9 @@
   [x] > right /bytes
 
   # Concatenation of two byte sequences and returns result as `bytes`.
-  # When the combined length would exceed the maximum possible array size,
-  # returns the caller-supplied `cant-concat` fallback instead, or the
-  # bottom object when none was provided, which terminates on use.
+  # Fails with a system error when the combined length would exceed the
+  # maximum possible array size.
   [b] > concat /bytes
-    ? > cant-concat /{string}
 
   # Tests that a slice of bytes can be extracted from a larger byte sequence.
   [] +> tests-takes-part-of-bytes
--- a/eo-runtime/src/main/eo/bytes.eo
+++ b/eo-runtime/src/main/eo/bytes.eo
@@ -114,7 +114,11 @@
   [x] > right /bytes
 
   # Concatenation of two byte sequences and returns result as `bytes`.
+  # When the combined length would exceed the maximum possible array size,
+  # returns the caller-supplied `cant-concat` fallback instead, or the
+  # bottom object when none was provided, which terminates on use.
   [b] > concat /bytes
+    ? > cant-concat /{string}
 
   # Tests that a slice of bytes can be extracted from a larger byte sequence.
   [] +> tests-takes-part-of-bytes
--- a/eo-runtime/src/main/java/org/eolang/EObytes$EOconcat.java
+++ b/eo-runtime/src/main/java/org/eolang/EObytes$EOconcat.java
@@ -19,19 +19,42 @@ package org.eolang;
 public final class EObytes$EOconcat extends PhDefault implements Atom {
 
     /**
+     * Name of the error-branch void that holds the caller's concat fallback.
+     */
+    private static final String FALLBACK = "cant-concat";
+
+    /**
      * Ctor.
      */
     public EObytes$EOconcat() {
-        super(new Attrs(new Attr("b", new AtVoid("b"))));
+        super(new Attrs(
+            new Attr("b", new AtVoid("b")),
+            new Attr(EObytes$EOconcat.FALLBACK, new AtVoid(EObytes$EOconcat.FALLBACK))
+        ));
     }
 
     @Override
     public Phi lambda() {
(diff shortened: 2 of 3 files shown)


morphqdd:

yegor256:
@morphqdd this type of error should not be "catchable" by programmer, I believe. @maxonfjvipon WDYT?

maxonfjvipon:
@yegor256 agree, Claude tells the same

morphqdd:
@yegor256 @maxonfjvipon, should the error be made fatal?

yegor256:
@morphqdd it's a system error, just throw `ExFailure` @maxonfjvipon am I right?

maxonfjvipon:
@yegor256 yes

morphqdd:
@yegor256 @maxonfjvipon, updated per your feedback: `bytes.concat` now throws `ExFailure` on overflow instead of exposing a catchable `cant-concat` fallback attribute. Removed the fallback void from `bytes.eo` and the `EObytes$EOconcat` ctor/attrs, and the atom now throws directly:

```java
if ((long) current.length + provided.length > Integer.MAX_VALUE) {
    throw new ExFailure(
        String.format(
            "Can't concatenate bytes of size %d with bytes of size %d: too big",
            current.length, provided.length
        )
    );
}
```

Test updated to assert the overflow surfaces as a system error rather than a caller-catchable fallback. Rebased on current master, all tests green, qulice clean.

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (29) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+880](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5385, opened 2026-07-10
#5341: fix i64.lt overflow by delegating to overflow-free gt instead of subtraction

### LINKED ISSUE #5341: i64.lt / lte operators return incorrect values due to 64-bit subtraction overflow

## Summary
i64.lt is implemented as `0.as-i64.gt ($.minus (i64 x))`, i.e. it computes the difference `$ - x` as an i64 and then tests whether `0 > ($ - x)`. i64 subtraction (`minus` = `plus (value.neg)`, both native Long ops) wraps modulo 2^64, so whenever the true difference `$ - x` falls outside [Long.MIN, Long.MAX] the sign of the wrapped result flips and the comparison returns the opposite of the correct answer. `lte` (lines 46-50) is built on `lt` and inherits the same defect.

## Location
[`eo-runtime/src/main/eo/i64.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/i64.eo#L38) lines 38-42.

## Buggy code
```
[x] > lt
  0.as-i64.gt > @
    minus
      i64 value
  x > value!
```

## Why it's wrong
Ordering must reflect the true mathematical order of the two i64 values. The sibling `gt` operator (line 54, backed by EOi64$EOgt.java) performs a direct `long > long` comparison with no possibility of overflow, proving the intended semantics is exact ordering. Deriving `lt` from a wrapping subtraction contradicts that and is internally inconsistent.

Verifier confirmation: Confirmed. At eo-runtime/src/main/eo/i64.eo:38-42, `lt` is implemented as `0.as-i64.gt (minus (i64 value))`, computing `0 > ($ - x)` where the difference is a native i64 subtraction. `minus` (`plus (i64 value).neg`) is built from EOi64$EOplus.java (`Long.sum`, wraps) and EOi64$EOtimes.java (`long * long`, wraps), so the subtraction wraps modulo 2^64. Whenever the true difference `$ - x` falls outside [Long.MIN, Long.MAX] the wrapped result's sign flips and `lt` returns the inverted answer; `lte` (lines 46-50) is derived from `lt` and inherits the defect. Verified traces: `$=Long.MIN_VALUE, x=1` yields `lt=false` although MIN<1 is true; `$=5e18, x=-5e18` yields `lt=true` although 5e18<-5e18 is false. By contrast the sibling `gt` (EOi64$EOgt.java) uses a direct overflow-free `long > long` comparison, establishing exact ordering as the intended semantics and making the subtraction-based `lt`/`lte` internally inconsistent. The in-file tests (lines 141-197) only exercise tiny magnitudes (10, 50, -5, -200, -100), leaving the overflow path untested. This is a genuine, silent wrong-result bug distinct from the already-filed issues (#5255 as-i32, #5256 div-by-zero, #5260 number.floor).

## Expected behavior
lt returns true iff $ is strictly less than x for all i64 operands. Correct implementation would delegate to the overflow-free native comparison, e.g. `x.gt $` (i.e. `x > $`).

## Suggested fix
Define lt without subtraction, e.g. `x.gt ^` (return true iff x > $), mirroring the native gt operator, so no intermediate difference is formed.

morphqdd:
Closes #5341.

`i64.lt` was implemented as `0.as-i64.gt (minus (i64 value))`, computing the difference `$ - x` as an i64 and testing whether `0 > ($ - x)`. Since i64 subtraction wraps modulo 2^64 (native `Long` arithmetic), whenever the true mathematical difference falls outside `[Long.MIN_VALUE, Long.MAX_VALUE]` the wrapped result's sign flips and `lt` returns the inverted answer. `lte` is derived from `lt` and inherited the same defect. The sibling `gt` performs a direct, overflow-free `long > long` comparison, establishing exact ordering as the intended semantics — deriving `lt` from a wrapping subtraction contradicted that.

Fix: `lt` now delegates directly to the overflow-free `gt`, mirroring the suggested fix in the issue (`x.gt $`, i.e. `$ < x` iff `x > $`), with no intermediate difference ever formed:

```
[x] > lt
  gt. > @
    i64 x.as-bytes
    ^
```

Note on `i64 x.as-bytes`: `x` may arrive either as a live `i64` object or as the raw-bytes form `lte`/`gte` pass through their own const-cached `value` alias (`x > value!` dataizes+bytes-ifies `x` before forwarding it) — `.as-bytes` is valid on both shapes, and `i64 (...)` normalizes the result back into a proper `i64` before dispatching `.gt` on it. This is the same defensive normalization the original `lt` used (`i64 value`) before its subtraction; without it, `lte`'s indirect call into `lt` breaks even though direct external calls to `.lt` work fine.

Added two boundary tests to `i64.eo`, matching the verified traces from the issue:
- `tests-i64-less-across-full-range`: `Long.MIN_VALUE.lt(1)` must be `true` (the old subtraction-based version returned `false`, since `1 - Long.MIN_VALUE` overflows).
- `tests-i64-less-with-large-difference`: `5e18.lt(-5e18)` must be `false` (the old version returned `true`, since `5e18 - (-5e18) = 1e19` overflows `Long.MAX_VALUE`).

Verified both new tests fail against the original subtraction-based implementation with exactly the inverted answers described in the issue (`expected: <true> but was: <false>` for both), and pass with the fix. `i32.lt`/`i16.lt` delegate to `as-i64.lt`/`as-i32.lt` respectively, so they inherit the correct ordering transitively — no separate fix needed there.

Verification:
- `mvn -pl eo-runtime test -Dtest=EOi64Test`: 35 tests, 0 failures, 0 errors.

### DIFF
--- a/eo-runtime/src/main/eo/i64.eo
+++ b/eo-runtime/src/main/eo/i64.eo
@@ -36,10 +36,9 @@
   # Returns `org.eolang.true` if `$` < `x`.
   # Here `x` must be an `org.eolang.i64` object.
   [x] > lt
-    0.as-i64.gt > @
-      minus
-        i64 value
-    x > value!
+    gt. > @
+      i64 x.as-bytes
+      ^
 
   # Returns `org.eolang.true` if `$` <= `x`.
   # Here `x` must be an `org.eolang.i64` object.
@@ -156,6 +155,21 @@
       lt.
         10.as-i64
         -5.as-i64
+
+  # Tests that i64 less-than comparison is correct across the full range,
+  # where a subtraction-based implementation would wrap and invert the answer.
+  [] +> tests-i64-less-across-full-range
+    lt. > @
+      -9223372036854775808.as-i64
+      1.as-i64
+
+  # Tests that i64 less-than comparison doesn't wrap when the difference
+  # between the two operands exceeds the i64 range.
+  [] +> tests-i64-less-with-large-difference
+    not. > @
+      lt.
+        5000000000000000000.as-i64
+        -5000000000000000000.as-i64
 
   # Tests that i64 greater-than comparison returns true for larger values.
   [] +> tests-i64-greater-true


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+796](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5386, opened 2026-07-10
Back the file object with syscalls instead of java.nio

maxonfjvipon:
`file`'s operations were Java atoms backed by `java.nio`, raising `IOException`-derived errors that EO code could catch — the kind of catchable `error` the fragile-objects epic (#5376) is removing. This replaces all of them with syscalls, so `file.eo` is now fully syscall-backed and no longer needs a `+rt` meta.

Reads: `exists` is a non-throwing `access` probe; `size` and `is-directory` read a `stat` struct. The Linux and macOS `struct stat` layouts genuinely differ, so each is maintained explicitly in JNA and selected at call time, while Windows uses msvcrt's `_stat`. `size` and `is-directory` take a fallback (`cant-measure`/`cant-check`) returned when the file cannot be stated, and terminate with a ⊥ only when it is left unbound.

Mutations: `touch` uses `creat` + `close`, `move` uses `rename`, and `delete` uses `unlink` for a plain file or `rmdir` for a directory — chosen by `is-directory`, so deleting a directory through `file.delete` keeps working the way `Files.delete` did. Each terminates with a ⊥ on failure.

`open` is reworked from two independent descriptors (with `touch`/`delete`/`truncate` orchestration) into a single flag-driven `open()`: the access mode maps to one flag set — a read/write/read-write direction plus the `O_CREAT`/`O_TRUNC`/`O_APPEND` bits it needs, composed from atomic flags on `posix`/`win32` — and the kernel creates, truncates and positions for append atomically. One shared-cursor descriptor is threaded to the scope (standard `fopen` semantics), so two tests that relied on independent read/write cursors are rewritten to reopen for reading. The native `open` is variadic, so its creation mode is passed through a JNA varargs parameter.

Every OS split is an inline `if os.is-windows (win32 …) (posix …)`, the same way `socket` and `console` already branch. `⊥` is now reserved for programmer errors (an invalid mode string, reading a write-only stream, writing a read-only one); every runtime failure either recovers through a fallback or surfaces through `open`'s existing `cant-open`.

One Windows note: msvcrt `_stat` carries a 32-bit `st_size`, so sizes are reported up to two gigabytes; `_stat64` would lift that at the cost of the shared `stat` name.


eo-runtime/src/main/java/org/eolang/EOsm/Posix/StatSyscall.java
```diff
@@ -0,0 +1,265 @@
...
+        this.posix = posix;
+    }
+
+    @Override
+    public Phi make(final Phi... params) {
+        final Phi result = this.posix.take("return").copy();
+        final String path = new Dataized(params[0]).asString();
+        final FileStat info;
+        final int code;
+        if (Platform.isMac()) {
```
morphqdd, line 51: This branches only on `Platform.isMac()` — every non-macOS platform (including Linux **aarch64**) gets the `Linux` struct below, whose javadoc explicitly says "The `struct stat` of Linux on **x86-64**" (field order/size only correct for that arch). Real glibc aarch64 `struct stat` has a different field order and a different total size (128 bytes vs. this class's 144). On Linux aarch64 this silently misreads `st_mode`/`st_size` — `file.size`/`file.is-directory` would return garbage, not an error. Given the CI matrix and the growing prevalence of arm64 runners (GitHub arm64, Graviton), this seems like a real gap rather than a theoretical one — worth branching on architecture, not just OS.

eo-runtime/src/main/eo/fs/file.eo
```diff
@@ -118,53 +169,36 @@
...
-    # descriptor to the provided scope and closing them afterwards. The read
-    # descriptor is opened read-only from the beginning, the write descriptor for
-    # appending, mirroring the two independent streams a file provides. The open
-    # flags come from the syscall object's `open-read-flags` and `open-write-flags`.
+    # Writable unless opened read-only ("r").
+    (access.eq "r").as-bool.not > writable
+    # Readable unless opened write-only ("w") or append-only ("a").
+    and. > readable
+      (access.eq "w").as-bool.not
+      (access.eq "a").as-bool.not
```
morphqdd, line 190: For any mode string other than 'r'/'w'/'a' (e.g. a typo like "x"), both `writable` (line 182) and `readable` (lines 184-186) come out `true` — the exclusion checks only rule out the exact strings "r"/"w"/"a", so anything else defaults to "read-write". That means `readable.not.and writable.not` (line 169) never fires for invalid mode strings, so the intended validation is silently skipped.

I reproduced this against `throws-on-opening-with-wrong-mode` (mode `"x"` on an *existing* tmpfile): the test technically passes (an exception IS thrown), but for the wrong reason — it fails with `"There's no "Δ" in the object of "Φ.fs.file.open.open-file.open-through.stream""`, not with the intended "Wrong access mod..." message from line 170. The mode-validation guard never fires; the code falls through into `open-file`, treats "x" as read-write, and only crashes later on an unrelated dataization issue with `stream`. So the specific validation this test is meant to cover isn't actually happening — the test just gets lucky that *something else* throws downstream.

eo-runtime/src/main/java/org/eolang/EOsm/Posix/OpenSyscall.java
```diff
@@ -41,7 +41,8 @@ public Phi make(final Phi... params) {
             new Data.ToPhi(
                 CStdLib.INSTANCE.open(
```
morphqdd, line 42: None of the new syscall wrappers (this one, AccessSyscall, CreatSyscall, MkdirSyscall, RenameSyscall, RmdirSyscall, StatSyscall, UnlinkSyscall, and their Win32 counterparts) read `errno`/`GetLastError()` on failure — they just pass the raw `0`/`-1` return code up to EO, which turns it into a generic `"Couldn't open file '%s'"`/`"Can't delete..."` etc. with no OS-provided reason. The old `java.nio`-backed code carried the actual `IOException` cause (permission denied vs. no such file vs. disk full, etc.). This is a real diagnostic-quality regression — worth capturing `Native.getLastError()` (JNA) right after each native call (before any other native/JNA call that could clobber it) and surfacing it in the fallback message.

eo-runtime/src/main/eo/fs/file.eo
```diff
@@ -23,60 +21,113 @@
...
+      if.
+        os.is-windows
+        win32 operation (* path)
+        posix operation (* path)
+    if. > operation
+      is-directory
+      "rmdir"
+      "unlink"
 
   # Gets the file size in bytes and returns it as `org.eolang.number`.
```
morphqdd, line 105: The Windows-side javadoc (`Win32/StatFuncCall.java`, `Win32/Msvcrt.java`) correctly documents that `_stat`'s 32-bit `st_size` caps reported size at 2GB, and the PR description repeats this as a known caveat — but this doc-comment on `size` itself doesn't mention it. This is user-visible behavior (an EO caller has no way to know from `file.eo` alone that `size` silently truncates on Windows for files >2GB), and it's untested. Worth at least a one-line note here even if adding `_stat64` support is out of scope for this PR.

morphqdd (changes requested):
@maxonfjvipon, please look

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (2097 >= 100); -8 for way too many hits-of-code (2097 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+583](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +23 points for this: +12 as a basis; +16 for the 2097 hits-of-code that you reviewed; -5 for very few (1) comments.  Your running score is [+1158](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5387, opened 2026-07-10
#5343: use is-integer instead of x.div x for range-of-numbers integer guard

### LINKED ISSUE #5343: range-of-numbers allows non-integers, leading to fractional-step ranges

## Summary
The object documents that `start` and `end` must be `int`s and that the bottom object (⊥) is returned otherwise (lines 10-12). The guard meant to enforce this is `and (or 0.eq start (1.eq (start.div start))) (or 0.eq end (1.eq (end.div end)))`. But `x.div x` is ordinary IEEE-754 floating-point division (see EOnumber$EOdiv.java: `return new Data.ToPhi(left / right)`), and for ANY finite nonzero `x` — integer or not — `x/x == 1.0`, so `1.eq (x.div x)` is true. Combined with the `0.eq x` branch, the guard passes for every finite value and only rejects NaN/±infinity. It does NOT detect non-integers at all. So `range-of-numbers 1.2 3` passes the guard and builds `[1.2, 2.2]` (start=1.2, next=2.2, 3.2 fails the `<3` test) instead of returning ⊥.

## Location
[`eo-runtime/src/main/eo/ss/range-of-numbers.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/ss/range-of-numbers.eo#L13) lines 13-30.

## Buggy code
```
[start end] > range-of-numbers
  if. > @
    and.
      or.
        0.eq start
        1.eq (start.div start)
      or.
        0.eq end
        1.eq (end.div end)
    range
      ...
    T "Some of the arguments are not integers"
```

## Why it's wrong
`start.div start == 1` is a non-zero/finite test, not an integer test. The documented contract requires rejecting non-integers, and the runtime already provides the correct primitive `number.is-integer` (number.eo lines 19-21: `and is-finite (eq floor)`), which this code should have used. As written, fractional start/end silently yield fractional-step ranges, violating both the docstring and the step=1 semantics.

Verifier confirmation: Independently confirmed on latest master (HEAD 0c072badd). The integer guard at range-of-numbers.eo lines 15-21 uses `1.eq (start.div start)` / `1.eq (end.div end)`, but `number.div` is IEEE-754 double division (EOnumber$EOdiv.java line 33: `return new Data.ToPhi(left / right)`), and for any finite nonzero x, x/x == 1.0 exactly. Combined with the `0.eq x` branch, the guard passes for every finite value and rejects only NaN/±infinity — it is a finite/nonzero test, NOT an integer test. So `range-of-numbers 1.2 3` passes the guard and, via range.eo's `current.lt end` loop, builds the fractional-step list [1.2, 2.2] instead of returning ⊥, directly violating the docstring contract at lines 10-12 ("If they're not ints, the bottom object (⊥) is returned"). The correct primitive `is-integer` (`and is-finite (eq floor)`) already exists at number.eo lines 19-21 and is not used. The claim's test-masking point also checks out: to-java.xsl lines 817-847 wrap `throws-*` tests in assertThrows around `new Dataized(...).asBool()`, and calling asBool on the resulting list/tuple throws regardless of whether validation fired, so the throws-on tests pass for the wrong reason. The bug is genuine and distinct from all already-filed issues (#5253-#5260), which touch BytesRaw, Sha, i64, EOfs, Emissions, and number.floor — none in range-of-numbers.eo.

## Expected behavior
range-of-numbers 1.2 3 (or any non-integer start/end) should return the bottom object ⊥ (terminate on use), per the docstring 'If they're not ints, the bottom object (⊥) is returned'.

## Suggested fix
Replace the `1.eq (x.div x)` checks with the existing `number.is-integer`, e.g. `and. (number start).is-integer (number end).is-integer` (with the 0-cases already covered since 0 is an integer).

morphqdd:
Closes #5343.

`range-of-numbers` documents that `start`/`end` must be `int`s and that the bottom object (⊥) is returned otherwise, but the guard enforcing this was `or (0.eq x) (1.eq (x.div x))`. Since `number.div` is ordinary IEEE-754 floating-point division, `x/x == 1.0` for *any* finite nonzero `x` — integer or not. Combined with the `0.eq x` branch, the guard passed for every finite value and only rejected NaN/±infinity; it never actually detected non-integers. So `range-of-numbers 1.2 3` silently built the fractional-step list `[1.2, 2.2]` instead of returning ⊥.

Fix: replace the guard with the existing `number.is-integer` primitive (`and. is-finite (eq floor)`), which is exactly what the docstring's contract requires:

```
and.
  start.is-integer
  end.is-integer
```

`0` is still accepted since `0.is-integer` is `true`, so no existing behavior for integer inputs changes.

The existing `throws-on-range-of-numbers-with-non-number-start`/`-with-not-number-end`/`-of-not-numbers` tests already exercise exactly the fractional-input scenarios from the issue and continue to pass — now because `range-of-numbers` genuinely returns ⊥ on non-integer input, rather than (as the issue notes) passing only because `to-java.xsl` wraps `throws-on-*` tests in `assertThrows` around an unconditional `.asBool()` call, which throws on the previously-returned list regardless of whether real validation fired. That test-wrapping looseness is a separate, XSL-level concern outside the scope of this fix (and not specific to this object), so no new test was added on top of the existing three — a new test would be masked by the very same wrapping mechanism the issue describes and wouldn't add real signal.

Verification:
- `mvn -pl eo-runtime test -Dtest=EOss.EOrange_of_numbersTest`: 7 tests, 0 failures, 0 errors.

### DIFF
--- a/eo-runtime/src/main/eo/ss/range-of-numbers.eo
+++ b/eo-runtime/src/main/eo/ss/range-of-numbers.eo
@@ -13,12 +13,8 @@
 [start end] > range-of-numbers
   if. > @
     and.
-      or.
-        0.eq start
-        1.eq (start.div start)
-      or.
-        0.eq end
-        1.eq (end.div end)
+      start.is-integer
+      end.is-integer
     range
       [] >>
         build ^.start > @


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+808](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5388, opened 2026-07-10
#5346: print integer-valued doubles above Long range exactly in StUnhex

### LINKED ISSUE #5346: StUnhex.number() incorrectly renders integer-valued doubles with magnitude >= 2^63

## Summary
When StUnhex converts a number's byte payload back into EO source text, `number()` prints any double whose fractional part is 0 via `Long.toString(num.longValue())`. For an integer-valued double whose magnitude is >= 2^63 (all doubles above 2^53 are integer-valued), `Double.longValue()` saturates to `Long.MAX_VALUE` / `Long.MIN_VALUE` instead of the actual value. So a literal such as `1e19` is emitted as `"9223372036854775807"` rather than a faithful representation of `10000000000000000000`.

## Location
[`eo-parser/src/main/java/org/eolang/parser/StUnhex.java`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/StUnhex.java#L94) lines 94-105 (buggy line 100).

## Buggy code
```java
private static String number(final Double num) {
    final String str;
    if (num % 1 == 0) {
        if ("-0.0".equals(num.toString())) {
            str = "-0";
        } else {
            str = Long.toString(num.longValue());
        }
    } else {
        str = Double.toString(num);
    }
    return str;
}
```

## Why it's wrong
`Double.longValue()` clamps out-of-long-range doubles to `Long.MAX_VALUE`/`Long.MIN_VALUE` (verified: `(1e19).longValue() == 9223372036854775807`). The method's intent is to print an exact integer form; the existing test `convertsNumberFromHexToEo` expects the exact digits `72872276393407200` for a `7.28e16` value, showing exact integer rendering is contractually required. For values >= 2^63 the printed digits are simply wrong. StUnhex is wired into the real XMIR->EO printing pipeline via `Xmir.java` (not test-only), so number literals >= 2^63 round-trip to incorrect EO source.

Verifier confirmation: Confirmed genuine and distinct. At `StUnhex.java:94-106`, `number()` prints any double whose fractional part is 0 (line 96) via `Long.toString(num.longValue())` (line 100). `Double.longValue()` clamps out-of-long-range doubles to `Long.MAX_VALUE`/`Long.MIN_VALUE`, which I independently verified: `(1e19).longValue()==9223372036854775807` while the true integer value is `10000000000000000000` and `num.toString()=="1.0E19"`. Every double above 2^53 is integer-valued, so all integer-valued doubles with magnitude >= 2^63 (e.g. `1e19`) are emitted as the saturated Long extreme rather than their real value, corrupting the XMIR->EO print/round-trip. This contradicts the method's stated intent ("Prints as int if fractional part of number is 0", lines 89-91) and the exact-integer contract shown by `StUnhexTest.convertsNumberFromHexToEo` (line 35 asserts the exact digits `72872276393407200` for a sub-2^63 value). The shift is production-wired, not test-only: `Xmir.java:43` constructs `new StUnhex()` in the real printing chain. It does not overlap any already-filed issue (none touch `StUnhex.java`), and it is reachable because EO numbers are 64-bit doubles that can hold values >= 2^63. Severity is medium: it is a real, silent correctness/round-trip defect but confined to the decompile/print path and to the extreme value range (magnitudes >= ~9.2e18), not core arithmetic.

## Expected behavior
The number should be rendered without loss (e.g. via the double's own decimal form / `BigDecimal`) so integer-valued doubles too large for long are still printed correctly, or fall back to `Double.toString` like other non-representable cases.

## Suggested fix
For the integer branch, guard on `Math.abs(num) < 9.223372036854776E18` (2^63) before using `longValue()`; otherwise render via `new java.math.BigDecimal(num).toBigInteger().toString()` (or fall back to `Double.toString`). This keeps small integers dash-free while staying exact for large magnitudes.

morphqdd:
Closes #5346.

`StUnhex.number()` printed any integer-valued double via `Long.toString(num.longValue())`. `Double.longValue()` clamps out-of-`long`-range doubles to `Long.MAX_VALUE`/`Long.MIN_VALUE` instead of the actual value. Every double above `2^53` is integer-valued, so any integer-valued double with magnitude `>= 2^63` (e.g. `1e19`) printed as the saturated `Long` extreme (`9223372036854775807`) rather than its real value (`10000000000000000000`). `StUnhex` is wired into the real XMIR→EO printing pipeline (`Xmir.java`), so number literals `>= 2^63` round-tripped to corrupted EO source.

Fix: guard on `Math.abs(num) < 0x1p63` (2^63) before using `longValue()`; for magnitudes at or above that, render via `BigDecimal.valueOf(num).toBigInteger()`, which stays exact for the full representable integer range of `double` (small integers keep the existing dash-free `Long` path unchanged).

Added `convertsBigIntegerValuedDoubleFromHexToEo`, unhexing the IEEE-754 bytes of `1e19` and asserting the exact printed digits `10000000000000000000`. Verified the test fails against the original code (assertion mismatch against the actual output of `9223372036854775807`, matching the issue's reported symptom exactly) and passes with the fix.

Verification:
- `mvn -pl eo-parser test -Dtest=StUnhexTest`: 8 tests, 0 failures, 0 errors.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/StUnhex.java
+++ b/eo-parser/src/main/java/org/eolang/parser/StUnhex.java
@@ -7,6 +7,7 @@ package org.eolang.parser;
 import com.yegor256.xsline.Shift;
 import com.yegor256.xsline.StEnvelope;
 import com.yegor256.xsline.StSequence;
+import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.regex.Pattern;
@@ -96,8 +97,10 @@ final class StUnhex extends StEnvelope {
         if (num % 1 == 0) {
             if ("-0.0".equals(num.toString())) {
                 str = "-0";
-            } else {
+            } else if (Math.abs(num) < 0x1p63) {
                 str = Long.toString(num.longValue());
+            } else {
+                str = BigDecimal.valueOf(num).toBigInteger().toString();
             }
         } else {
             str = Double.toString(num);
--- a/eo-parser/src/test/java/org/eolang/parser/StUnhexTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/StUnhexTest.java
@@ -122,6 +122,23 @@ final class StUnhexTest {
 
     @ParameterizedTest
     @MethodSource("shifts")
+    void convertsBigIntegerValuedDoubleFromHexToEo(final Shift shift, final String type) {
+        MatcherAssert.assertThat(
+            String.format(
+                "StUnhex by %s must exactly print a double above Long range, but it didn't",
+                type
+            ),
+            new Xsline(new StUnhex(shift)).pass(
+                new XMLDocument(
+                    "<p><o base='Φ.number'><o base='Φ.bytes'><o>43-E1-58-E4-60-91-3D-00</o></o></o></p>"
+                )
+            ),
+            XhtmlMatchers.hasXPath("//o[text()='10000000000000000000']")
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("shifts")
     void convertsFloatFromHexToEo(final Shift shift, final String type) {
         MatcherAssert.assertThat(
             String.format("StUnhex by %s must convert float", type),


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+820](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5389, opened 2026-07-10
#5345: fix split/nsplit to support multi-character delimiters

### LINKED ISSUE #5345: `split` and `nsplit` functions fail to recognize multi-character delimiters

## Summary
rec-split decides where to cut by comparing the whole delimiter against a single sliced unit: `delim.eq (slice bts current 1)`. `slice bts current 1` always returns exactly ONE character/unit, while `delim` for any delimiter of length >= 2 is two or more units. Byte-sequence equality (`.eq`) between a 1-unit slice and a 2+-unit delimiter is always false, so the delimiter branch is never taken. As a result `split "a::b::c" "::"` returns a single-element tuple `["a::b::c"]` instead of `["a","b","c"]`. The same defect exists in tt/nsplit.eo (lines 40-44: `delimiter.eq (slice text current 1)`).

## Location
[`eo-runtime/src/main/eo/tt/split.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/tt/split.eo#L30) lines 30-40.

## Buggy code
```
      if.
        delim.eq
          slice bts current 1
        rec-split
          appended
          (current.plus 1).as-number
          (current.plus 1).as-number
        rec-split
          accum
          start
          (current.plus 1).as-number
```

## Why it's wrong
The docstring says 'Returns a tuple of strings: text separated by delimiter' with no restriction to single-character delimiters, and `delimiter` is a general string/text parameter. Hard-coding a length-1 slice makes the comparison structurally incapable of matching any multi-character delimiter, so such calls silently produce wrong output rather than splitting or reporting an error.

Verifier confirmation: Confirmed independently by reading the code on disk. In split.eo the cut decision (lines 31-32) compares the entire `delim` (bound from `delimiter` at line 16) against `slice bts current 1`, which via slice.eo's `text.slice start len` (line 12) always returns exactly one unit. A byte-sequence `.eq` between a 2+-unit delimiter and a 1-unit slice can never be true, so the split branch (lines 33-36) is unreachable for any delimiter of length >= 2; moreover `current` only advances by 1 per step (lines 35-36/40), so the recursion structurally cannot consume a multi-char delimiter even if it matched. Tracing `split "a::b::c" "::"` yields the single-element tuple `["a::b::c"]` instead of `["a","b","c"]`. nsplit.eo repeats the same defect at lines 41-42 (`delimiter.eq (slice text current 1)`). The docstring (line 10) places no single-character restriction on `delimiter`, which is a general text parameter, and all existing tests (single-char '-'/' ') never exercise the multi-char case, so multi-char calls silently produce wrong output. This is a distinct file (split.eo/nsplit.eo) not covered by any already-filed issue. Severity is medium: it is a public string-utility correctness bug producing silent wrong output, but it does not crash and the common single-character case works.

## Expected behavior
split "a::b::c" "::" should yield * "a" "b" "c"; instead it yields * "a::b::c".

## Suggested fix
Slice the delimiter's own length instead of 1, e.g. compare `delim.eq (slice bts current delim-length)` and advance `current`/`start` by the delimiter length when matched (and likewise in nsplit); or explicitly document and validate that the delimiter must be a single character.

morphqdd:
Closes #5345.

Both `tt/split.eo` and `tt/nsplit.eo` decided where to cut by comparing the *whole* delimiter against a fixed **1-unit** slice: `delim.eq (slice bts current 1)`. `slice bts current 1` always returns exactly one unit, while `delim` for any delimiter of length >= 2 is two or more units, so byte-sequence equality between a 1-unit slice and a 2+-unit delimiter is always false — the delimiter branch is structurally unreachable for any multi-character delimiter, and the recursion only ever advances by 1 regardless. `split "a::b::c" "::"` returned the single-element tuple `["a::b::c"]` instead of `["a","b","c"]`; `nsplit` had the identical defect.

Fix (both files): slice exactly `delimiter.size` units instead of a hardcoded `1`, and advance `current`/`start` by the delimiter's length (not `1`) when it matches. Since slicing `delimiter.size` units past `current` could run past the end of `bts`/`text` for a delimiter near the tail, added an explicit bounds guard (`(current.plus delim-size).gt len`) before attempting the comparison — evaluated eagerly as a separate `if` rather than combined with `and.` into one condition, since EO's `and.` isn't short-circuiting and would otherwise still evaluate the (out-of-bounds) `slice` even when the guard fails. The guard also treats a zero-length delimiter as "never matches" (same as the original's de-facto behavior for an empty delimiter), avoiding a new degenerate infinite-recursion case that a naive `current + 0`-advance would introduce.

Added `tests-splits-text-by-multi-char-delimiter` (`split "a::b::c" "::"`) and `tests-nsplit-with-multi-char-delimiter` (`nsplit "a::b::c::d" "::" 2`), matching the exact scenario from the issue. Verified both new tests fail against the original single-unit-slice code (`expected: <true> but was: <false>`) and pass with the fix. All pre-existing single-character-delimiter tests continue to pass unchanged.

Verification:
- `mvn -pl eo-runtime test -Dtest=EOtt.EOsplitTest,EOtt.EOnsplitTest`: 13 tests total, 0 failures, 0 errors.

### DIFF
--- a/eo-runtime/src/main/eo/tt/nsplit.eo
+++ b/eo-runtime/src/main/eo/tt/nsplit.eo
@@ -25,6 +25,7 @@
         1.eq limit
         * text
         rec-nsplit * 0 0 0
+  delimiter.size > delimiter-size!
 
   # Recursively splits text by delimiter up to limit number of times.
   [accum start current count] > rec-nsplit
@@ -37,21 +38,30 @@
             start
             current.minus start
       if.
-        and.
-          delimiter.eq
-            slice text current 1
-          (number count).lt
-            (number limit).minus 1
-        rec-nsplit
-          appended
-          (current.plus 1).as-number
-          (current.plus 1).as-number
-          (count.plus 1).as-number
+        or.
+          delimiter-size.eq 0
+          (current.plus delimiter-size).gt text.size
         rec-nsplit
           accum
           start
           (current.plus 1).as-number
           count
+        if.
+          and.
+            delimiter.eq
+              slice text current delimiter-size
+            (number count).lt
+              (number limit).minus 1
+          rec-nsplit
+            appended
+            (current.plus delimiter-size).as-number
+            (current.plus delimiter-size).as-number
+            (count.plus 1).as-number
+          rec-nsplit
+            accum
+            start
+            (current.plus 1).as-number
+            count
 
   # Tests that nsplit splits with limit 2 correctly.
   [] +> tests-nsplit-with-limit-two
@@ -126,3 +136,13 @@
           "-"
           3
       * "" "a" "b-c"
+
+  # Tests that nsplit correctly handles a multi-character delimiter.
+  [] +> tests-nsplit-with-multi-char-delimiter
+    eq. > @
+      list
+        nsplit
+          "a::b::c::d"
+          "::"
+          2
+      * "a" "b::c::d"
--- a/eo-runtime/src/main/eo/tt/split.eo
+++ b/eo-runtime/src/main/eo/tt/split.eo
@@ -15,6 +15,7 @@
     rec-split * 0 0
   text > bts!
   delimiter > delim!
+  delim.size > delim-size!
   bts.size > len!
 
   # Recursively splits text by delimiter into tuple of substrings.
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+836](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5390, opened 2026-07-10
#5360: fix LnMeta double-space error column offset

### LINKED ISSUE #5360: LnMeta.split incorrectly reports column for double-space ParseError due to omitted tail base offset

## Summary
In LnMeta.split(tail, span), tail is body.substring(space + 1) — i.e. the parts region that begins at index space+1 within span.body(). But the double-space ParseError reports its column as `span.indent() + idx`, where idx is an index into `tail`, not into the full body. The offset of `space + 1` (the `+name ` prefix inside body) is omitted, so the reported/underlined column points `space + 1` characters too far to the left.

## Location
[`eo-parser/src/main/java/org/eolang/parser/LnMeta.java`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/LnMeta.java#L83) lines 83-104.

## Buggy code
```java
int idx = 0;
while (idx < tail.length()) {
    if (tail.charAt(idx) == ' ') {
        throw new ParseError(
            span.line(), span.indent() + idx,
            "meta parts must be separated by exactly one space"
        );
    }
```

## Why it's wrong
The physical-line column of tail[idx] is span.indent() + (space + 1) + idx. Everywhere else in the parser a body-relative index is turned into a column via span.indent() + <body index> (e.g. Tokens uses span.indent() + this.cursor where cursor indexes into body). Here idx indexes into the already-offset `tail` substring, so span.indent() + idx under-counts by space+1.

Verifier confirmation: Confirmed genuine and distinct. LnMeta.into (line 68) computes tail = body.substring(space + 1) and passes it to split(tail, span). Inside split, when a double space (or leading extra space) is detected at tail.charAt(idx) (lines 87-88), the ParseError column is span.indent() + idx (line 89). But idx indexes into tail, which starts space+1 characters into body(). The established parser-wide convention (Span.body() = text.substring(indent); Tokens.java uses span.indent() + this.cursor where cursor indexes into body) is pos = span.indent() + <body index>. The correct physical column is therefore span.indent() + (space + 1) + idx; the code undercounts by exactly space+1. This pos flows unmodified through Eo.dispatch (line 298: emit.error(err.line(), err.pos(), ...)) into Emit.formatted (line 414), which builds new MsgUnderlined(source, pos, 1) against the full physical line (source = this.lines.get(line-1)), so the caret lands space+1 columns too far left. For "+foo a  b" (indent 0): space=4, tail="a  b", the throw fires at idx=2 giving pos=2 (caret under 'o' of foo) instead of the true column 7 (the second space). The double-space branch is reachable: after emitting a part, idx is advanced past exactly one space, so a second consecutive space triggers the throw; a leading extra space (e.g. "+foo  b") throws at idx=0 with pos=indent (points at '+') instead of the real space column. This is not in the already-filed list (all in Emissions/StUnhex/Suffix/Comments) and is a real defect. Severity is low because the error is still correctly raised and the parse still fails as intended; only the diagnostic caret/column is misplaced, not correctness of acceptance.

## Expected behavior
For `+foo a  b` (indent 0): body="+foo a  b", space=4, tail="a  b"; the extra space is at tail idx 2, whose true column is 0 + 4 + 1 + 2 = 7 (the second space between a and b). The caret should point at column 7.

## Suggested fix
Report the position as span.indent() + space + 1 + idx (thread the body-relative base through split, or add space+1). E.g. pass the base offset into split and use `span.indent() + base + idx`.

morphqdd:
Closes #5360.

In `LnMeta.into`, `tail = body.substring(space + 1)` is a substring of `body` starting `space + 1` characters in. That offset was dropped when `split(tail, span)` reported the column of a rejected double space: it used `span.indent() + idx`, where `idx` indexes into `tail`, not into the full line body. Everywhere else in the parser a body-relative index becomes a column via `span.indent() + <body index>` (e.g. `Tokens` uses `span.indent() + this.cursor`), so this undercounted the column by exactly `space + 1`.

For `+rt jvm  extra` (indent 0): `space = 3`, `tail = "jvm  extra"`, the extra space sits at `tail` index 4. The old code reported column 4 (pointing inside "jvm"); the correct column is `0 + 4 (space+1) + 4 = 8`, the actual second space.

Fix: thread the tail's base offset (`space + 1`) through `split` and use `span.indent() + base + idx`.

Extended `rejectsDoubleSpaceBetweenParts` to assert the reported `ParseError.pos()` is the real column (8), not the old (wrong) value (4) — confirmed the assertion fails against the original code with `expected: <8> but: was <4>`, and passes with the fix.

Verification:
- `mvn -pl eo-parser test -Dtest=LnMetaTest`: 9 tests, 0 failures.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnMeta.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnMeta.java
@@ -65,7 +65,7 @@ final class LnMeta implements Line {
             parts = new ArrayList<>(0);
         } else {
             head = body.substring(1, space);
-            parts = LnMeta.split(body.substring(space + 1), this.span);
+            parts = LnMeta.split(body.substring(space + 1), space + 1, this.span);
         }
         globals.markMeta();
         globals.clearBlanks();
@@ -77,16 +77,17 @@ final class LnMeta implements Line {
      * R-3.2.4 and promoting any bare {@code Q} to {@code Φ} per
      * R-3.2.3 / R-9.3.
      * @param tail Substring after the {@code +name} prefix
+     * @param base Offset of {@code tail} within the line's body
      * @param span Source span (for error reporting)
      * @return Parts in source order
      */
-    private static List<String> split(final String tail, final Span span) {
+    private static List<String> split(final String tail, final int base, final Span span) {
         final List<String> out = new ArrayList<>(2);
         int idx = 0;
         while (idx < tail.length()) {
             if (tail.charAt(idx) == ' ') {
                 throw new ParseError(
-                    span.line(), span.indent() + idx,
+                    span.line(), span.indent() + base + idx,
                     "meta parts must be separated by exactly one space"
                 );
             }
--- a/eo-parser/src/test/java/org/eolang/parser/LnMetaTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/LnMetaTest.java
@@ -104,11 +104,15 @@ final class LnMetaTest {
 
     @Test
     void rejectsDoubleSpaceBetweenParts() {
-        Assertions.assertThrows(
-            ParseError.class,
-            () -> new LnMeta(new Span("+rt jvm  extra", 1))
-                .into(new Stack(), new Globals(), new Emit()),
-            "double space between parts must be rejected per R-3.2.4"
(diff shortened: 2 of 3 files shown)


morphqdd:

PR #5391, opened 2026-07-10
#5358: fix LnOnlyPhi split to use depth/string-aware marker scan

### LINKED ISSUE #5358: LnOnlyPhi wrongly splits line at indexOf('> ['), causing parse error for valid string with '> [' in LHS

## Summary
The classifier decides a line is an only-phi formation via Eo.topLevelGreaterBracket (Eo.java:505), which scans for `> [` only at paren-depth 0 and outside string literals. But once LnOnlyPhi.into runs, it locates the split point with a plain body.indexOf("> ["), which finds the FIRST literal occurrence anywhere, including inside a string literal or paren group in the LHS. The LHS is then taken as body.substring(0, phi), the parameter list from the following bracket, etc., all computed from the wrong offset.

## Location
[`eo-parser/src/main/java/org/eolang/parser/LnOnlyPhi.java`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/LnOnlyPhi.java#L69) lines 69.

## Buggy code
```java
final String body = this.span.body();
final int phi = body.indexOf("> [");
```

## Why it's wrong
The two pieces of logic that must agree (is-this-only-phi? vs where-does-the-phi-split?) use different scanning rules. When the LHS legitimately contains the sequence `> [` inside a string (or parentheses), the naive indexOf splits the line in the middle of that string, corrupting the LHS and parameter list.

Verifier confirmation: Confirmed a genuine, distinct eo-parser bug. The only-phi classifier `Eo.onlyPhi`/`Eo.topLevelGreaterBracket` (Eo.java:495-534) decides a line is an only-phi formation by scanning for `> [` while tracking paren depth and string-literal state (with backslash-escape handling), accepting the marker only at depth 0 outside strings. But once routed to `LnOnlyPhi.into`, the split point is located with a naive `body.indexOf("> [")` (LnOnlyPhi.java:69) that has no such awareness and matches the first literal occurrence anywhere, including inside a string or paren group in the LHS. For the valid input `x.print "> [" > [a] > n` (STRING hargs are explicitly supported per the class javadoc lines 42-45), the classifier correctly matches the real top-level `> [` at index 14 and routes to only-phi, but `indexOf` matches the `> [` inside the string literal at index 9, so `lhs = body.substring(0,9)` = `x.print "` and every downstream offset (bracket, close, params, suffix) is computed from the wrong position. Tokens then reports a spurious "unterminated string literal" error, rejecting a valid program. The two pieces of logic that must agree (is-only-phi vs where-to-split) use different scanning rules. This is reachable through normal dispatch and is distinct from all six already-filed issues (which are in Emissions, StUnhex, Suffix, and Comments).

## Expected behavior
For `x.print "> [" > [a] > n`, the φ LHS is the application `x.print "> ["`, the params are `[a]`, and the name is `n` — a valid only-phi formation. It should parse successfully.

## Suggested fix
Use the same depth/string-aware scan the classifier uses (topLevelGreaterBracket, or a shared helper) to find the `> [` split index in LnOnlyPhi, rather than String.indexOf.

morphqdd:
Closes #5358.

`Eo.onlyPhi` classifies a line as an only-phi formation via a depth/string-aware scan for the top-level `> [` marker (`Eo.topLevelGreaterBracket`, now `topLevelGreaterBracketIndex`) — it tracks paren depth and string-literal state (with backslash-escape handling) and only accepts the marker outside strings and at depth 0. But `LnOnlyPhi.into` located the actual split point with a naive `body.indexOf("> [")`, which matches the first literal occurrence anywhere, including inside a string literal or paren group in the LHS.

For `x.print "> [" > [a] > n` (a STRING harg is explicitly supported per the class's own javadoc), the classifier correctly finds the real marker at index 14 and routes to `LnOnlyPhi`, but `indexOf` matches the `> [` inside the string at index 9, so `lhs = body.substring(0, 9)` = `x.print "` — a truncated string literal. `Tokens` then throws a spurious "unterminated string literal" `ParseError`, rejecting a valid program.

Fix: extracted the classifier's scan into a package-private `Eo.topLevelGreaterBracketIndex(body)` returning the marker's index (or -1), used by both `Eo.onlyPhi` (`>= 0`) and `LnOnlyPhi.into` — so the two pieces of logic that must agree now share one implementation instead of drifting.

Added `splitsAtTopLevelMarkerNotInsideLhsString` asserting `x.print "> [" > [a] > n` parses into the expected shape (param `a`, φ chain `x.print`). Verified it throws the exact "unterminated string literal" `ParseError` against the original `indexOf`-based code, and passes with the fix.

Verification:
- `mvn -pl eo-parser test -Dtest=LnOnlyPhiTest,EoTest`: all green.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -87,6 +87,46 @@ final class Eo implements Iterable<Directive> {
     }
 
     /**
+     * The index of the substring {@code "> ["} at paren depth 0 (i.e.,
+     * outside of any paren group and outside of strings), or -1 if
+     * none exists. Shared by the {@link Eo#onlyPhi} classifier and
+     * {@link LnOnlyPhi}, so both agree on where an only-phi formation
+     * splits.
+     * @param body Line body to scan
+     * @return Index of the top-level marker, or -1
+     */
+    static int topLevelGreaterBracketIndex(final String body) {
+        int depth = 0;
+        boolean instr = false;
+        int found = -1;
+        int idx = 0;
+        while (idx < body.length() - 2 && found < 0) {
+            final char glyph = body.charAt(idx);
+            if (instr) {
+                if (glyph == '\\' && idx + 1 < body.length()) {
+                    idx = idx + 2;
+                    continue;
+                }
+                if (glyph == '"') {
+                    instr = false;
+                }
+            } else if (glyph == '"') {
+                instr = true;
+            } else if (glyph == '(') {
+                depth = depth + 1;
+            } else if (glyph == ')') {
+                depth = depth - 1;
+            } else if (depth == 0 && glyph == '>'
+                && body.charAt(idx + 1) == ' '
+                && body.charAt(idx + 2) == '[') {
+                found = idx;
+            }
+            idx = idx + 1;
+        }
+        return found;
+    }
+
+    /**
      * Merge a BYTES continuation starting at index {@code start} —
      * concatenates the trailing-dash chunk with subsequent BYTES-only
      * lines at &gt;= the same indent until a chunk without trailing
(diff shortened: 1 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+852](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5392, opened 2026-07-10
#5357: reject trailing garbage after expression in Suffix.parse

### LINKED ISSUE #5357: Suffix.parse doesn't raise error for trailing non-suffix text, causing incorrect parsing of invalid lines

morphqdd:
Closes #5357.

`Suffix.parse` skips leading spaces on the line tail and, if the remaining text is not `+>`, `>>`, or `>`, falls into a final `else` that silently returned `Form.NONE` — no suffix, no error. Callers (`LnApplication`, `LnCompactTuple`, `LnReversed`, `LnMethod`, `LnOnlyPhi`) pass the entire un-consumed token tail into `Suffix` and never verify the stream is fully consumed afterward, so any non-whitespace trailing content that isn't a suffix marker was silently dropped instead of being rejected. This is only reachable for heads with a well-defined end (INT/FLOAT/HEX/BYTES/STAR/STRING/ROOT/GROUP) — identifier heads greedily consume to a NAME terminator so they never hit it.

For `5x`: head reads INTEGER `5`, tail is `x`, and the old code silently accepted the program as `5` while dropping `x`. `named()` already guards this exact condition via `endsClean` ("trailing garbage after name suffix") for the `> name` path; the `NONE` path had no equivalent guard.

Fix: the final `else` now throws a `ParseError` ("trailing garbage after expression") instead of returning `Form.NONE`.

This also fixed two existing typo-pack fixtures that were silently exercising the same bug: `qq.yaml` (`QQ.io.stdout > @` — the extra `Q` was silently dropped, masking the intended `> @` suffix and surfacing an unrelated "object inside formation must have a name" error) and `version-with-inline-application.yaml` (`stdout|1.2.3 42` — `|1.2.3 42` was silently dropped after `stdout`). Both now correctly report "trailing garbage after expression" at the exact column the garbage starts, instead of an unrelated downstream error with no diagnostic value for the actual typo.

Added `rejectsTrailingGarbageThatIsNotASuffixMarker` reproducing the issue's exact `5x` example. Verified it fails (nothing thrown) against the original code, and passes with the fix.

Verification:
- `mvn -pl eo-parser test`: all 874+ tests green, including the two updated typo-pack fixtures.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnTextBlock.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnTextBlock.java
@@ -49,8 +49,11 @@ final class LnTextBlock implements Line {
                 "text block closer must start with triple-quote"
             );
         }
+        final Tokens tokens = new Tokens(body, this.span);
+        tokens.seek(3);
+        tokens.readChain();
         final Suffix suffix = new Suffix(
-            body.substring(3), this.span, this.span.indent() + 3
+            tokens.tail(), this.span, this.span.indent() + tokens.cursor()
         );
         final String joined = Emissions.unescapeBody(
             String.join(String.valueOf('\n'), globals.tbody()).trim()
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -229,7 +229,10 @@ final class Suffix {
         } else if (tail.charAt(idx) == '>') {
             result = Suffix.named(tail, idx + 1, span, home);
         } else {
-            result = new Suffix.Parsed(Form.NONE, "", "", false);
+            throw new ParseError(
+                span.line(), home + idx,
+                "trailing garbage after expression"
+            );
         }
         return result;
     }
--- a/eo-parser/src/test/java/org/eolang/parser/LnTextBlockTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/LnTextBlockTest.java
@@ -85,6 +85,37 @@ final class LnTextBlockTest {
     }
 
     @Test
+    void acceptsChainAfterCloserWithoutSuffix() {
+        final Globals globals = new Globals();
+        globals.openTextBlock(1, 0);
+        globals.appendTextLine("hi");
+        final Stack stack = new Stack();
+        new LnTextBlock(new Span("\"\"\".as-bytes", 3))
+            .into(stack, globals, new Emit());
+        MatcherAssert.assertThat(
+            "a `.method` chain right after the closer (deferred, not yet emitted) must not"
+                .concat(" be rejected as trailing garbage"),
+            stack.top().kind(),
+            Matchers.equalTo(Kind.TEXT_BLOCK)
+        );
+    }
+
+    @Test
+    void marksLevelNamedWhenSuffixFollowsChain() {
+        final Globals globals = new Globals();
+        globals.openTextBlock(1, 0);
+        globals.appendTextLine("hi");
+        final Stack stack = new Stack();
+        new LnTextBlock(new Span("\"\"\".as-bytes > greeting", 3))
+            .into(stack, globals, new Emit());
+        MatcherAssert.assertThat(
+            "a `> name` suffix following a deferred chain must still be parsed",
+            stack.top().named(),
+            Matchers.is(true)
+        );
+    }
+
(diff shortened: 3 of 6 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+868](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5368, opened 2026-07-10
Math objects sqrt, pow, ln, sin, cos, acos, asin opaque in EO due to Java implementation


yegor256:
Seven `ms` math objects are declared in EO but implemented as JVM atoms in Java, not in EO itself: `sqrt` (`eo-runtime/src/main/eo/ms/sqrt.eo:12` → `EOms/EOsqrt.java:38`), `pow` (`ms/pow.eo:11` → `EOpow.java`), `ln` (`ms/ln.eo:13` → `EOln.java`), `sin` and `cos` (`ms/angle.eo:30,34` → `EOangle$EOsin.java`, `EOangle$EOcos.java`), `acos` (`ms/acos.eo` → `EOacos.java`), and `asin` (`ms/asin.eo` → `EOasin.java`). Each just delegates to `java.lang.Math` inside `lambda()`.

Every such atom is a black hole for EO: its behavior is opaque to the language and readable only by opening the Java side, which hurts transparency. Ideally all objects should be implemented in EO, not in Java, so that the semantics live in one place and can be understood, verified, and ported without a JVM.

Reimplement each of these seven objects in pure EO — drop the `/number` atom marker and write the body using numeric algorithms (e.g. Newton's method for `sqrt`, series expansions for `ln`/`sin`/`cos`/`asin`/`acos`, `exp`\×`ln` for `pow`) composed from existing EO primitives — then delete the corresponding `EOms` Java classes. The tests already living in the `.eo` files can stay as the acceptance criteria.

0crat:
@yegor256 Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+1423](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5394, opened 2026-07-10
Implement sine and cosine in pure EO

yegor256:
This replaces the `angle.sin` and `angle.cos` JVM atoms with two standalone objects written in pure EO. `sine` reduces its argument into the `[-pi, pi]` range and sums a Taylor series through Horner's scheme, so it stays accurate for angles of any magnitude and collapses to `nan` for non-finite input. `cosine` reuses it via `cos(x) = sin(x + pi/2)`. The `angle` object now keeps only `value`, `degrees` and `radians`, so it no longer needs a JVM runtime, and both Java atoms are deleted.

The point is #5368: these two were opaque to the language, readable only by opening the Java side. Now their behaviour lives in EO and is covered by tests (including a plain 17-radian case with no relation to pi) that assert against the reference values scaled by 1000.

This is a partial step toward #5368 — it handles `sin` and `cos` only. The other five objects it lists (`sqrt`, `pow`, `ln`, `acos`, `asin`) are still JVM atoms and left for follow-up.


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1875 >= 100); -8 for way too many hits-of-code (1875 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1455](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5397, opened 2026-07-11
#5353: fix Comments.attach to not leak a dangling comment across a blank line

### LINKED ISSUE #5353: `Comments.attach` doesn't clear or report pending comments on decline, leading to misattachment

## Summary
attach() flushes the buffered comment block only when the upcoming line is named, at the same indent, and has no intervening blank line. When any of those conditions fails it falls through and returns WITHOUT clearing globals.pendingComments() (and without reporting the block as dangling). The class Javadoc explicitly states the opposite: 'either flush the block ... or report the block as dangling per R-6.4.2. The buffer is always cleared after the call so a dangling block is not re-reported by the EOF check.' Because the buffer survives the decline, the comment leaks past the object it was supposed to precede and silently attaches to the NEXT eligible named object.

## Location
[`eo-parser/src/main/java/org/eolang/parser/Comments.java`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/Comments.java#L50) lines 50-54.

## Buggy code
```java
final Span head = pending.get(0);
if (named && head.indent() == span.indent() && globals.pendingBlanks() == 0) {
    emit.comment(pending, span.line());
    globals.clearComments();
}
```

## Why it's wrong
R-6.5.2 forbids a blank line between a comment block and the object it attaches to, so a comment separated from its object by a blank is dangling and must be reported (see finish() in Eo.java which reports the EOF case). The Javadoc guarantees the buffer is always cleared after attach(). Instead, on the decline path the buffer is retained and later wrongly flushed onto an unrelated object.

Verifier confirmation: Confirmed against the code on disk. Comments.attach (Comments.java:43-55) clears globals.pendingComments() only inside the flush branch (clearComments at line 53, guarded by the if at line 51). On every decline path (not named, indent mismatch, or pendingBlanks()!=0) it returns at line 55 without clearing and without reporting the block as dangling, contradicting the class/method Javadoc (lines 16-22) which guarantees attach either flushes OR reports dangling and that the buffer is always cleared. No other code path clears the buffer on the object side: clearComments appears only at LnComment.java:60 (a different case) and Comments.java:53; LnFormation.into (line 62) and LnApplication.into (line 77) never clear comments. Concrete leak: input '# c' / blank / '[x] > foo' / '[y] > bar' at indent 0 — c is buffered; the blank sets blanks=1; foo is a master object (LnFormation calls only enterAfterMeta, not checkPlain, so no R-6.5.4 error) and attach declines (blanks==1) without clearing; foo then calls clearBlanks (LnFormation.java:64); bar then satisfies the if (named, matching indent, blanks==0) and attach flushes c onto bar via emit.comment. So a comment separated from foo by a blank line (dangling per R-6.5.2) is silently misattached to a later unrelated object and never reported by the EOF check in Eo.java:433-439. The named=false decline path leaks identically. A genuine, distinct eo-parser correctness bug, separate from the filed Emissions.java/StUnhex.java issues.

## Expected behavior
When attach() declines to flush (wrong indent, blank line intervened, or not named), it should clear the buffer (and per the Javadoc report the block as dangling), so the comment is never re-attached to a subsequent named object.

## Suggested fix
Add an else branch to the if at line 51 that reports the block dangling via emit.error(head.line(), head.indent(), "comment must precede a named object") and calls globals.clearComments(); i.e. make the buffer always cleared after the call as the Javadoc states.

morphqdd:
Closes #5353.

`Comments.attach` only cleared `globals.pendingComments()` on the flush branch (named, matching indent, no intervening blank). On any decline it fell through and returned without clearing — contradicting the class/method javadoc, which states the buffer is always cleared after the call (either flushed or reported dangling per R-6.4.2).

The three decline conditions aren't all equivalent, though. Two of them are legitimately "not yet, maybe later": an unnamed follower or an indent mismatch (with no intervening blank line) can still be reclaimed by a later same-indent named line — this is exactly how a vertical-continuation chain works (an anonymous `[]` immediately followed by nested lines, with the name only landing on the chain's tail several lines down, e.g. the existing `vmethod-after-formation.yaml` fixture). Reporting those as dangling immediately would be a false positive.

The third condition — a blank line separating the comment block from what follows — is an unconditional R-6.5.2 violation regardless of what comes after. That's the one the issue's concrete leak scenario hinges on: `# c` / blank / `[x] > foo` / `[y] > bar` — the blank-line decline used to silently leave the buffer pending, and it would then get flushed onto the next matching object (`bar`), which never should have received it.

Fix: check `globals.pendingBlanks() > 0` first — if a blank intervened, report dangling and clear immediately, regardless of the upcoming line's name/indent. Otherwise, keep the original flush-or-defer behavior unchanged (only clearing on a successful flush).

Added `EoTest.doesNotLeakDanglingCommentToLaterUnrelatedObject` reproducing the issue's exact scenario end-to-end, plus updated `CommentsTest` to cover the blank-line-dangling case directly and confirm the unnamed/indent-mismatch decline paths still correctly defer (unchanged behavior, verified against the existing `vmethod-after-formation.yaml` fixture, which continues to pass).

Verified the negative control: reverting to the original bundled condition reproduces the exact leak (comment `c` ends up attached to `bar` instead of being reported dangling).

Verification:
- `mvn -pl eo-parser test`: all 1449 tests green, including the existing `eo-syntax`/`eo-packs` fixture corpus.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Comments.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Comments.java
@@ -10,16 +10,18 @@ import java.util.List;
  * Attach buffered comment lines to the next named object — §6.4.
  *
  * <p>A comment block accumulates in {@link Globals#pendingComments()}
- * as each {@link LnComment} arrives. The next named line — any
- * formation, application, method or reversed line whose suffix carries
- * a name — calls {@link #attach(Globals, Emit, Span, boolean)} to
- * either flush the block into {@code /object/comments/comment} (when
- * the line is named, sits at the same indent, and has no intervening
- * blank line — R-6.5.2) or report the block as dangling per
- * R-6.4.2.</p>
- *
- * <p>The buffer is always cleared after the call so a dangling block
- * is not re-reported by the EOF check.</p>
+ * as each {@link LnComment} arrives. Every subsequent line — named or
+ * not — calls {@link #attach(Globals, Emit, Span, boolean)}: a blank
+ * line separating the block from what follows is an immediate
+ * R-6.5.2 violation, reported as dangling per R-6.4.2 and cleared on
+ * the spot. Absent a blank, the block flushes into
+ * {@code /object/comments/comment} as soon as a named line at the
+ * same indent arrives; an unnamed or deeper/shallower-indent line in
+ * between is not itself a violation — the block stays pending so a
+ * later same-indent named line (e.g. the name-bearing tail of a
+ * vertical-continuation chain) can still claim it. Any block left
+ * pending at end of stream is reported dangling by the EOF check in
+ * {@link Eo}.</p>
  *
  * @since 0.1
  */
@@ -48,7 +50,13 @@ final class Comments {
             return;
         }
         final Span head = pending.get(0);
-        if (named && head.indent() == span.indent() && globals.pendingBlanks() == 0) {
+        if (globals.pendingBlanks() > 0) {
+            emit.error(
+                head.line(), head.indent(),
+                "comment must precede a named object"
+            );
+            globals.clearComments();
+        } else if (named && head.indent() == span.indent()) {
             emit.comment(pending, span.line());
             globals.clearComments();
         }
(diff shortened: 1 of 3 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+896](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5398, opened 2026-07-11
#5348: reject unclosed inline-phi bracket with ParseError instead of crashing

### LINKED ISSUE #5348: Unclosed inline-phi bracket in a paren group causes unwanted StringIndexOutOfBoundsException in inlinePhi method

## Summary
When a parenthesised-group argument contains a top-level `> [` inline-phi marker but no closing `]`, `inner.indexOf(']', bracket)` returns -1, and the subsequent `inner.substring(bracket + 1, close)` (with close == -1) throws an unchecked StringIndexOutOfBoundsException. This exception is NOT a ParseError, so it escapes the per-line recovery try/catch in Eo.dispatch (which only catches ParseError) and aborts the whole parse instead of emitting an error directive.

## Location
[`eo-parser/src/main/java/org/eolang/parser/Emissions.java`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/Emissions.java#L498) lines 498-500.

## Buggy code
```java
        final int bracket = phi + 2;
        final int close = inner.indexOf(']', bracket);
        final String lhs = inner.substring(0, phi).stripTrailing();
        final String params = inner.substring(bracket + 1, close);
```

## Why it's wrong
The parser's recovery model (spec §7 / Eo.dispatch) guarantees that any malformed line is turned into a ParseError and recovered from. inlinePhi omits the `close < 0` guard that its sibling LnOnlyPhi.into (lines 77-83) performs for exactly the same `> [ ... ]` construct, so a missing `]` produces an uncaught runtime exception rather than a recoverable ParseError. A parser must never throw an unchecked exception on any input.

Verifier confirmation: Independently confirmed by execution. In Emissions.java, inlinePhi (lines 493-516) computes close = inner.indexOf(']', bracket) at line 498; when a top-level `> [` inline-phi marker inside a paren-group argument has no closing `]`, close is -1, and line 500's inner.substring(bracket + 1, close) throws an unchecked StringIndexOutOfBoundsException. This path is reachable: readGroup accepts `(a > [b)` as a non-redundant GROUP, the line classifies as LnApplication (the `> [` is at paren depth 1, so Eo.onlyPhi is false), and emitArg->openValue->group->inlinePhi runs. Eo.dispatch (Eo.java:293-299) catches only ParseError, so the runtime exception escapes and aborts the entire parse instead of emitting a recoverable <error> directive. I reproduced it: input `x (a > [b)` yields StringIndexOutOfBoundsException: Range [5, -1) at Emissions.inlinePhi(Emissions.java:500), propagating uncaught out of Eo.directives. The sibling LnOnlyPhi.into (lines 77-83) guards the identical construct with `if (close < 0) throw new ParseError(...)`, which inlinePhi is missing. This is distinct from the already-filed NumberFormatException issues (#5257, #5259) and the StUnhex saturation issue (#5346): different exception type, method, and construct.

## Expected behavior
A malformed inline-phi group such as `x (a > [b)` should emit an <error> directive (via a thrown ParseError like 'only-phi parameter list missing closing `]`'), and parsing should continue.

## Suggested fix
After computing `close`, if `close < 0` throw a ParseError (e.g. `new ParseError(line, column + bracket, "only-phi parameter list missing closing `]`")`) before using `close` in substring, mirroring LnOnlyPhi.into lines 77-83.

morphqdd:
Closes #5348.

`Emissions.inlinePhi` computed `close = inner.indexOf(']', bracket)` for an inline-phi formation's parameter list, then immediately used it in `inner.substring(bracket + 1, close)` without checking whether the closing `]` was actually found. When a parenthesized-group argument contains a top-level `> [` marker with no closing `]` (e.g. `x (a > [b)`), `close` is `-1`, and `substring` throws an unchecked `StringIndexOutOfBoundsException` — which is not a `ParseError`, so it escapes `Eo.dispatch`'s per-line recovery (which only catches `ParseError`) and aborts the entire parse instead of emitting a recoverable `<error>` directive.

`LnOnlyPhi.into` already guards the identical `> [ ... ]` construct with a `close < 0` check; `inlinePhi` was missing the equivalent guard for the same construct reached through a different path (a paren-group argument rather than a top-level line).

Fix: added the same guard, throwing `ParseError(line, column + bracket, "only-phi parameter list missing closing \`]\`")` before the `substring` call, mirroring `LnOnlyPhi`.

Added `LnApplicationTest.rejectsUnclosedInlinePhiBracketInGroupArg` reproducing the issue's exact `x (a > [b)` example. Verified it fails with the raw `StringIndexOutOfBoundsException` against the original code, and throws the intended `ParseError` with the fix.

Verification:

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -496,6 +496,12 @@ final class Emissions {
     ) {
         final int bracket = phi + 2;
         final int close = inner.indexOf(']', bracket);
+        if (close < 0) {
+            throw new ParseError(
+                line, column + bracket,
+                "only-phi parameter list missing closing `]`"
+            );
+        }
         final String lhs = inner.substring(0, phi).stripTrailing();
         final String params = inner.substring(bracket + 1, close);
         emit.object(name, null, line, column);
--- a/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
@@ -518,6 +518,17 @@ final class LnApplicationTest {
     }
 
     @Test
+    void rejectsUnclosedInlinePhiBracketInGroupArg() {
+        Assertions.assertThrows(
+            ParseError.class,
+            () -> new LnApplication(new Span("x (a > [b)", 1))
+                .into(new Stack(), new Globals(), new Emit()),
+            "a `> [` inline-phi marker with no closing `]` inside a paren-group arg must be"
+                .concat(" rejected with a ParseError, not an unchecked exception")
+        );
+    }
+
+    @Test
     void emitsAsAttributeForLabelBinding() {
         final Emit emit = new Emit();
         new LnApplication(new Span("foo a:y b:z > x", 1))


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+908](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5399, opened 2026-07-11
#5351: enforce strict NAME token for +> test suffix

### LINKED ISSUE #5351: `+> name` test suffix parser silently discards trailing content and accepts non-NAME characters

## Summary
The `test` parser for a `+> name` suffix scans the name with the loose `terminates` predicate (stops only at space/tab/`!`/`/`) and, unlike the `named` parser, never calls `endsClean` to verify the rest of the tail is blank. As a result a test suffix consumes only the first whitespace-delimited token and silently discards anything after it, and it also accepts characters that are illegal in a NAME token (`.`, `[`, `]`, `:`, etc.), because it uses `terminates` instead of the stricter `endsName` used by the named path.

## Location
[`eo-parser/src/main/java/org/eolang/parser/Suffix.java`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/Suffix.java#L246) lines 246-267.

## Buggy code
```java
final int start = idx;
while (idx < tail.length() && !Suffix.terminates(tail.charAt(idx))) {
    idx = idx + 1;
}
if (start == idx) {
    throw new ParseError(
        span.line(), home + start,
        "test attribute requires a name"
    );
}
return new Suffix.Parsed(Form.TEST, tail.substring(start, idx), "", false);
```

## Why it's wrong
R-6.3.5 requires a `+>` test name to be a NAME token, and the sibling `named` path (lines 306-350) enforces both strictness (via `endsName`, line 317) and no-trailing-garbage (via `endsClean`, line 348). The asymmetry means `foo > bar baz` is correctly rejected as "trailing garbage after name suffix" while `foo +> bar baz` is silently accepted with name `bar` and `baz` dropped, and `foo +> a.b` is accepted producing an invalid `@name='+a.b'`.

Verifier confirmation: Confirmed by code trace against the current file. Suffix.test() (lines 246-267) scans the +> name with the loose predicate !terminates() at line 257 (stops only at space/tab/!//) and returns at line 266 with no endsClean and no endsName check, whereas the sibling named() path uses the stricter endsName() at line 317 (rejecting NAME_BOUNDARIES ",.|':;?[]{}()" defined at line 48) and calls endsClean() at line 348. The asymmetry is reachable: Tokens.tail() (Tokens.java:623) returns the whole line remainder and callers (e.g. LnApplication.into lines 67-69) pass it directly to new Suffix. Thus "[] +> bar baz" is silently accepted with label "bar" and " baz" discarded, while "[] > bar baz" throws "trailing garbage after name suffix" (behavior locked by eo-typos yaml tests suffix-with-dots.yaml / scoped-application-after-suffix.yaml). Likewise "[] +> a.b" yields label "a.b" → attribute "+a.b", violating the class's own documented R-6.3.5 requirement that a +> name be a NAME token, whereas "[] > a.b" is rejected. Neither malformed input is caught upstream or downstream (the produced names are lexically valid), so it is genuine silent misparse/data loss, and it is distinct from the three already-filed Emissions/StUnhex issues.

## Expected behavior
A `+> name` suffix should reject non-NAME characters in the name and reject any non-whitespace content after the name, mirroring the `> name` path.

## Suggested fix
In `test`, scan the name with `endsName` (as `named` does), reject the cactus-emoji codepoint, and call `Suffix.endsClean(tail, idx, span, home)` after the name to reject trailing garbage.

morphqdd:
Closes #5351.

`Suffix.test` (the `+> name` parser) scanned the name with the loose `terminates` predicate (stops only at space/tab/`!`//`) instead of the stricter `endsName` used by the sibling `named` (`> name`) path, and never rejected the cactus-emoji codepoint that `named` rejects. R-6.3.5 requires a `+>` test name to be a NAME token, same as `> name`.

Concretely, `[] +> a.b` was accepted, producing an invalid `@name='+a.b'` attribute — `.` is not a legal NAME character but `terminates` doesn't treat it as a boundary, so the scan swallowed the whole `a.b` as one token.

(The issue also flagged a missing trailing-garbage check — `endsClean` — but that part was already fixed as a side effect of #5357's `Suffix.parse` fix, which is now merged on master. This PR covers the remaining `terminates`-vs-`endsName` gap and the missing cactus-emoji guard.)

Fix: scan with `endsName` instead of `terminates`, and reject the cactus emoji codepoint, mirroring `named`.

Added `rejectsDotInPlusGreaterName` (`[] +> a.b`) and `rejectsTrailingGarbageAfterPlusGreaterName` (`[] +> bar baz`, regression-guarding the already-fixed part). Verified `rejectsDotInPlusGreaterName` fails (nothing thrown) against the pre-fix `terminates`-based scan, and passes with the fix.

Verification:

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -267,7 +267,7 @@ final class Suffix {
             );
         }
         final int start = idx;
-        while (idx < tail.length() && !Suffix.terminates(tail.charAt(idx))) {
+        while (idx < tail.length() && !Suffix.endsName(tail.charAt(idx))) {
             idx = idx + 1;
         }
         if (start == idx) {
@@ -276,8 +276,15 @@ final class Suffix {
                 "test attribute requires a name"
             );
         }
+        final String name = tail.substring(start, idx);
+        if (name.codePoints().anyMatch(cp -> cp == 0x1F335)) {
+            throw new ParseError(
+                span.line(), home + start,
+                "cactus emoji is reserved for auto-names; not allowed in identifiers"
+            );
+        }
         Suffix.endsClean(tail, idx, span, home);
-        return new Suffix.Parsed(Form.TEST, tail.substring(start, idx), "", false);
+        return new Suffix.Parsed(Form.TEST, name, "", false);
     }
 
     /**
--- a/eo-parser/src/test/java/org/eolang/parser/SuffixTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/SuffixTest.java
@@ -218,6 +218,30 @@ final class SuffixTest {
     }
 
     @Test
+    void rejectsTrailingGarbageAfterPlusGreaterName() {
+        Assertions.assertThrows(
+            ParseError.class,
+            () -> new Suffix(
+                " +> bar baz",
+                new Span("[] +> bar baz", 1), 2
+            ),
+            "`+> bar baz` must reject the extra token, not silently drop it"
+        );
+    }
+
+    @Test
+    void rejectsDotInPlusGreaterName() {
+        Assertions.assertThrows(
+            ParseError.class,
+            () -> new Suffix(
+                " +> a.b",
+                new Span("[] +> a.b", 1), 2
+            ),
+            "`+> a.b` must be rejected — `.` is not a legal NAME character"
+        );
+    }
+
+    @Test
     void rejectsTrailingGarbageThatIsNotASuffixMarker() {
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (35) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+920](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5400, opened 2026-07-11
#5356: require a dot boundary in decorate.xsl string-prefix matching

### LINKED ISSUE #5356: Loose string prefix matching in `decorate.xsl` corrupts unrelated FQNs like `Φ.stringify`

## Summary
The entry guard for decoration (line 55) matches an object when `some $p in $metas/part[1] satisfies starts-with(@base, $p)`, and the `not-equal` mode re-selects the meta with `$metas[starts-with(current()/@base, part[1])][1]` (line 69). Both tests are raw string `starts-with` with no `.` (dot) boundary. So for `+decorate string text` (which after resolve-aliases/add-default-package has `part[1] = 'Φ.string'`), an unrelated object whose FQN merely shares that string prefix — e.g. `Φ.strings` or `Φ.stringify` — satisfies `starts-with('Φ.stringify','Φ.string')`. It is not an exact match and not a dotted sub-path, so it flows into `not-equal` mode: `substring-after('Φ.stringify','Φ.string.')` yields '' (the concat on line 70 adds the missing '.'), `tokenize('','\.')` yields the empty sequence, so `$methods` is empty, `method` mode takes the `$index > count($methods)` branch (1 > 0), rebuilds the object with `base=$meta/part[1]` (='Φ.string') and decorates it. The original `Φ.stringify` object is thereby silently replaced by the decorated `Φ.text(Φ.string)`.

## Location
[`eo-parser/src/main/resources/org/eolang/parser/parse/decorate.xsl`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/resources/org/eolang/parser/parse/decorate.xsl#L55) lines 55-75.

## Buggy code
```xml
<xsl:template match="o[some $p in $metas/part[1] satisfies starts-with(@base, $p)]" mode="with-base">
  ...
<xsl:template match="o" mode="not-equal">
  <xsl:variable name="meta" select="$metas[starts-with(current()/@base, part[1])][1]"/>
  <xsl:variable name="after" select="reverse(tokenize(substring-after(@base, concat($meta/part[1], '.')), '\.'))"/>
```

## Why it's wrong
Decoration must apply only to an FQN that is exactly the decorated base or a proper dotted sub-path of it (e.g. `Φ.string` or `Φ.string.joined.with`). A name that only happens to begin with the same characters (`Φ.strings`, `Φ.stringify`) is a completely different object and must be left untouched. Using `starts-with` without requiring a `.` separator (or exact equality) breaks the identity of such objects.

Verifier confirmation: The claim is accurate and independently reproduced. The entry guard on line 55 uses `some $p in $metas/part[1] satisfies starts-with(@base, $p)` with no dot-boundary, and the not-equal re-selection on line 69 (`$metas[starts-with(current()/@base, part[1])][1]`) is likewise a raw string prefix test. For a `+decorate string text` meta (part[1]='Φ.string'), an unrelated FQN such as `Φ.stringify` or `Φ.strings` satisfies the prefix test but not the exact match on line 56, so it flows into `not-equal`. There, `substring-after('Φ.stringify','Φ.string.')` (line 70, with the concat adding the missing '.') yields '', `tokenize('','\.')` yields the empty sequence, so `$methods` is empty and the `method` mode takes the `$index > count($methods)` branch (1 > 0), rebuilding the object with `base=$meta/part[1]` (='Φ.string') and decorating it. Running decorate.xsl directly with Saxon-HE 13.0 on `<o base='Φ.stringify'/>` and `<o base='Φ.strings'/>` both produced the identical corrupted output `<o base='Φ.text'><o base='Φ.string' .../></o>` — the same as the legitimate exact-match `Φ.string` case — silently dropping the `ify`/`s` suffix and replacing a completely different object. The correct predicate would require exact equality or a `.`-terminated prefix (`@base=$p or starts-with(@base, concat($p,'.'))`). This is a genuine correctness defect in a different file from all six already-filed issues (which are in Emissions.java, StUnhex.java, Suffix.java, Comments.java), so it is distinct.

## Expected behavior
`Φ.strings` / `Φ.stringify` should pass through unchanged; only `Φ.string` (exact) and `Φ.string.<...>` (dotted sub-path) should be decorated.

## Suggested fix
Require a path boundary in both tests: change the match guard to `some $p in $metas/part[1] satisfies (@base = $p or starts-with(@base, concat($p, '.')))` (line 55) and the not-equal meta selection to `$metas[starts-with(current()/@base, concat(part[1], '.'))][1]` (line 69).

morphqdd:
Closes #5356.

`decorate.xsl`'s entry guard (`some $p in $metas/part[1] satisfies starts-with(@base, $p)`) and its `not-equal` meta re-selection (`$metas[starts-with(current()/@base, part[1])][1]`) both used a raw string `starts-with` with no `.`-boundary check. For a `+decorate string tt.text` meta (`part[1]='Φ.string'`), an unrelated object whose FQN merely *starts with* the same characters — `Φ.stringify`, `Φ.strings` — satisfied the prefix test even though it isn't the decorated object or a dotted sub-path of it. It then fell into `not-equal` mode, where `substring-after(@base, concat($meta/part[1], '.'))` produced an empty string (no `.` after the shared prefix), `tokenize('', '\.')` yielded an empty sequence, and the `method` template's `$index > count($methods)` branch (1 > 0) fired immediately — rebuilding the object with `base=$meta/part[1]` and decorating it. So `Φ.stringify`/`Φ.strings` got silently corrupted into the same decorated shape as a genuine `Φ.string` match.

Fix: require an exact match or a `.`-terminated prefix in both places — `(@base=$p or starts-with(@base, concat($p, '.')))` for the entry guard, `starts-with(current()/@base, concat(part[1], '.'))` for the not-equal re-selection.

Added `decorate-string-prefix.yaml` reproducing the exact scenario: `+decorate string tt.text` alongside `stringify > y` and `strings > z`. Verified both fail (corrupted into the decorated shape) against the original loose-prefix code, and pass unchanged with the fix.

Verification:
- `mvn -pl eo-parser test`: all tests green, including the new fixture and existing `decorate.yaml`/`decorate-fqn.yaml`.
- `mvn -pl eo-runtime compile`: the real build pipeline (which uses this sheet) still compiles cleanly.

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/decorate.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/decorate.xsl
@@ -52,7 +52,7 @@
   <xsl:template match="o[@base]">
     <xsl:apply-templates select="." mode="with-base"/>
   </xsl:template>
-  <xsl:template match="o[some $p in $metas/part[1] satisfies starts-with(@base, $p)]" mode="with-base">
+  <xsl:template match="o[some $p in $metas/part[1] satisfies (@base=$p or starts-with(@base, concat($p, '.')))]" mode="with-base">
     <xsl:variable name="meta" select="$metas[part[1]=current()/@base][1]"/>
     <xsl:choose>
       <xsl:when test="exists($meta)">
@@ -66,7 +66,7 @@
     </xsl:choose>
   </xsl:template>
   <xsl:template match="o" mode="not-equal">
-    <xsl:variable name="meta" select="$metas[starts-with(current()/@base, part[1])][1]"/>
+    <xsl:variable name="meta" select="$metas[starts-with(current()/@base, concat(part[1], '.'))][1]"/>
     <xsl:variable name="after" select="reverse(tokenize(substring-after(@base, concat($meta/part[1], '.')), '\.'))"/>
     <xsl:apply-templates select="." mode="method">
       <xsl:with-param name="meta" select="$meta"/>
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-packs/parse/decorate-string-prefix.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-packs/parse/decorate-string-prefix.yaml
@@ -1,0 +1,27 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets:
+  - /org/eolang/parser/parse/move-voids-up.xsl
+  - /org/eolang/parser/parse/wrap-method-calls.xsl
+  - /org/eolang/parser/parse/vars-float-up.xsl
+  - /org/eolang/parser/parse/build-fqns.xsl
+  - /org/eolang/parser/parse/expand-aliases.xsl
+  - /org/eolang/parser/parse/resolve-aliases.xsl
+  - /org/eolang/parser/parse/add-default-package.xsl
+  - /org/eolang/parser/parse/roll-bases.xsl
+  - /org/eolang/parser/parse/decorate.xsl
+asserts:
+  - /object[not(errors)]
+  - /object/o[@name='main']/o[@name='φ' and @base='Φ.tt.text']/o[@base='Φ.string' and not(@name)]
+  - /object/o[@name='main']/o[@name='y' and @base='Φ.stringify']
+  - /object/o[@name='main']/o[@name='z' and @base='Φ.strings']
+input: |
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (31) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+932](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5401, opened 2026-07-11
#5359: point fragile ?. chain link @pos at the dot, not the question mark

### LINKED ISSUE #5359: Fragile `?.` operator incorrectly assigns @pos to `?`, violating R-9.1.3 standard

## Summary
Tokens.readChain() captures the dispatch link's source column into `dot` BEFORE calling consumeDispatch(). For a plain `.` link the cursor already sits on the dot, so `dot` is correct. But for the fragile `?.` operator the cursor sits on the `?`, one column before the dot, so the emitted MethodChain records the `?` column rather than the `.` column. That MethodChain.dot() value is later written straight to @pos in Emissions.expression (lines 66/71), Emissions.emitArg (lines 191/196), ChainEmission.run (lines 95/102), and LnOnlyPhi.emitLhs (lines 157/162).

## Location
[`eo-parser/src/main/java/org/eolang/parser/Tokens.java`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/Tokens.java#L476) lines 476-481.

## Buggy code
```java
while (!this.atEnd() && this.dispatchAhead()) {
    final int dot = this.span.indent() + this.cursor;
    final boolean fragile = this.consumeDispatch();
    final Value name = this.readMethodName();
    chain.add(new MethodChain(name.raw(), dot, name.end(), fragile));
}
```

## Why it's wrong
MethodChain's own Javadoc says the stored column is the "Column of the leading dot" and "Per R-9.1.3, the dot column is the @pos value emitted for the link's <o>", and LnMethodTest.recordsDotColumnForChainLink asserts "the chain link's @pos must equal the column of the leading dot per R-9.1.3". The leading dot of a `?.` operator is the `.`, not the `?`. The sibling standalone-continuation path in LnMethod (line 86: `method.pos() - 1`) correctly points @pos at the `.` for both plain and fragile continuations, so the two code paths disagree by one column for fragile dispatch.

Verifier confirmation: Confirmed a genuine, distinct off-by-one in Tokens.readChain (Tokens.java:477). It captures `final int dot = this.span.indent() + this.cursor;` BEFORE calling consumeDispatch() (line 478). For a plain `.` the cursor already sits on the dot, so `dot` is the dot column (correct). But fragileAhead (Tokens.java:642-645) detects `?.` with the cursor on the `?`, so for the fragile operator `dot` is the column of the `?`, one column before the actual leading dot. That value flows unchanged into @pos via MethodChain.dot() in Emissions (lines 66/71), ChainEmission (95/102), and LnOnlyPhi (157/162). This violates MethodChain's own contract ("Column of the leading dot ... Per R-9.1.3, the dot column is the @pos value", MethodChain.java:23-25/9-13) and disagrees with the sibling LnMethod continuation path, which emits `method.pos() - 1` (LnMethod.java:86) = the `.` column for both plain and fragile links. I verified empirically by driving the real parser: `  foo?.bar > x` emits `<o base=".bar" pos="5">` (the `?` at column 5) even though the `.` is at column 6, whereas the plain `  foo.bar > x` correctly emits pos=5 for its `.` at column 5, and the equivalent fragile continuation `  ?.bar > y` emits pos=3 pointing at its `.`. So the fragile inline path is off by one and inconsistent with the continuation path. The bug is not in the already-filed list (all in Emissions/StUnhex/Suffix/Comments) and is distinct.

## Expected behavior
For an inline fragile chain `foo?.bar`, the `.bar` link's @pos should equal the column of the `.` (the leading dot). For `foo?.bar > x` at indent 2 (body index of `.` is 4, so column 6), @pos should be 6, matching what LnMethod produces for the equivalent `?.bar` continuation line.

## Suggested fix
Compute the dot column after accounting for the fragile prefix, e.g. determine `fragile` first (via Tokens.fragileAhead) and set `final int dot = this.span.indent() + this.cursor + (fragile ? 1 : 0);` before consuming, so the recorded column points at the `.` for both `.` and `?.` forms.

morphqdd:
Closes #5359.

`Tokens.readChain()` captured the dispatch link's source column (`dot`) BEFORE calling `consumeDispatch()`. For a plain `.` this is correct — the cursor already sits on the dot. But for the fragile `?.` operator the cursor sits on the `?`, one column before the dot, so the emitted `MethodChain` recorded the `?` column instead of the `.` column, violating `MethodChain`'s own contract ("Column of the leading dot ... Per R-9.1.3, the dot column is the @pos value") and disagreeing with the sibling standalone-continuation path in `LnMethod` (`method.pos() - 1`, which correctly points at the `.` for both plain and fragile continuations).

For `foo?.bar > x` at indent 0, the `.` sits at column 4 (`f`0 `o`1 `o`2 `?`3 `.`4), but the emitted `<o base='.bar'>` carried `@pos='3'` (the `?`).

Fix: determine `fragile` via `Tokens.fragileAhead` before consuming, and add 1 to the recorded column when fragile, so it points at the `.` for both forms.

Added `LnApplicationTest.recordsDotColumnForFragileChainLinkPos` reproducing the issue's exact example. Verified it fails (`pos='3'` instead of `'4'`) against the original code, and passes with the fix.

Verification:

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Tokens.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Tokens.java
@@ -474,8 +474,12 @@ final class Tokens {
     List<MethodChain> readChain() {
         final List<MethodChain> chain = new ArrayList<>(0);
         while (!this.atEnd() && this.dispatchAhead()) {
-            final int dot = this.span.indent() + this.cursor;
-            final boolean fragile = this.consumeDispatch();
+            final boolean fragile = Tokens.fragileAhead(this.body, this.cursor);
+            int dot = this.span.indent() + this.cursor;
+            if (fragile) {
+                dot = dot + 1;
+            }
+            this.consumeDispatch();
             final Value name = this.readMethodName();
             chain.add(new MethodChain(name.raw(), dot, name.end(), fragile));
         }
--- a/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
@@ -193,6 +193,20 @@ final class LnApplicationTest {
     }
 
     @Test
+    void recordsDotColumnForFragileChainLinkPos() {
+        final Emit emit = new Emit();
+        new LnApplication(new Span("foo?.bar > x", 1))
+            .into(new Stack(), new Globals(), emit);
+        emit.close();
+        MatcherAssert.assertThat(
+            "a fragile `?.` chain link's @pos must point at the `.`, not at the `?`,"
+                .concat(" matching the standalone-continuation path in LnMethod"),
+            LnApplicationTest.render(emit),
+            XhtmlMatchers.hasXPath("/object/o[@base='.bar' and @pos='4']")
+        );
+    }
+
+    @Test
     void emitsTwoLevelChainAsThreeSiblings() {
         final Emit emit = new Emit();
         new LnApplication(new Span("foo.bar.baz > x", 1))


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+944](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

morphqdd:

PR #5411, opened 2026-07-12
Terminate the last catchable fs errors with a bottom and unify nk/socket behind an adapter

### LINKED ISSUE #5410: `dir.walk` and `dir.tmpfile.touch` should terminate with ⊥, not produce a catchable error
`dir.walk` (`EOdir$EOwalk`) and `dir.tmpfile.touch` (`EOdir$EOtmpfile$EOtouch`) are the last two atoms in the runtime that produce a *catchable* error. Both do `catch (IOException) → throw new IllegalArgumentException(...)`, and `PhSafe` turns that into an `EOerror.ExError`, which `try` can catch — exactly the mechanism the fragile-objects epic (#5261) is removing.

Their failures are system-level — `Files.walk` fails on I/O during traversal, `Files.createTempFile` on `ENOSPC`/`EACCES` — so by the epic's rule they warrant no recoverable fallback; they should terminate with ⊥, the same treatment `file.touched`/`deleted`/`moved` already got.

The minimal fix keeps the atoms as they are and only changes the failure path from a catchable `IllegalArgumentException` to `throw new ExFailure(...)` (chaining the `IOException` as the root cause), which is how the other Java atoms (`Heaps`, `Expect`) already panic. Once these two are ⊥, nothing in the runtime produces a catchable `error`, which unblocks deleting the `error` object and retiring `try`.

Rewriting these atoms to pure EO backed by syscalls (`walk` via `opendir`/`readdir` + glob, `touch` via `mkstemp`) is a separate fs-fidelity effort and is intentionally out of scope here.

### LINKED ISSUE #5413: Unify nk/socket's posix and win32 implementations behind an adapter, dropping try/error
`nk/socket` no longer uses `try`/`error` for error handling — every network operation already returns a per-operation fallback (`cant-connect`, `cant-listen`, `cant-accept`, `cant-send`, `cant-receive`) and cleanup failures are ⊥. The `try` blocks that remain (`safe-socket` and `accept`, on both the posix and win32 sides) are pure RAII: their `catch (error ex > [ex])` is a dead re-raise, and their only real job is the `finally` that closes the socket.

Convert those to `seq`:

- delete the dead `error ex > [ex]` catch clauses;
- replace the `finally` with `seq * result (closed-socket sd) result`, where `result` is the scope's outcome captured once via `!` so the trailing reference returns the already-dataized bytes rather than re-running the scope over a closed socket;
- keep the whole main (including the bad-address branch) inside the `seq` so the socket is still closed on the recover-and-continue path, not just on success.

Cleanup on ⊥ is unnecessary — a ⊥ propagates to the top and the process exits, so the OS reclaims the descriptor.

This removes the last uses of `try` outside `try.eo`, which unblocks deleting the `error` object and retiring `try` in a follow-up.

maxonfjvipon:
Two related steps toward removing the `error` object and retiring `try`.

**dir (`walk`, `tmpfile.touch`).** These were the last two atoms turning an `IOException` into an `IllegalArgumentException`, which `PhSafe` promotes to a catchable `EOerror.ExError`. Their failures are system-level, so they now `throw ExFailure` (keeping the `IOException` as the root cause), the way the other Java atoms already panic. A `throws-on-walking-absent-directory` test pins the termination down.

**nk/socket.** The posix and win32 stacks were two parallel ~180-line objects. Since Winsock is Berkeley-derived, their operation names and failure sentinels already match, so they collapse into one shared `impl` parameterized by a small per-OS adapter — the raw syscall object, the close name, a human-readable error text, and the start-up / clean-up steps. WSAStartup/WSACleanup normalize away (posix's startup is a constant `0`, its cleanup a no-op), so both platforms run the same `safe-socket` body. The RAII `try` blocks become `seq`: the dead `error` re-raise catches are dropped, and the socket is closed by sequencing the scope result (frozen once via `!`) before the close — cleanup on a bottom is unnecessary since the process is terminating and the OS reclaims the descriptor. `socket.eo` drops from 529 to 342 lines.

Together these remove the last uses of `error` and `try` outside `try.eo`, which unblocks deleting `error` and reframing `try` next.

Closes #5410
Closes #5413


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (341 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+610](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5420, opened 2026-07-12
#5333: fix intermittent transpile corruption from a shared, non-thread-safe XSL train

### LINKED ISSUE #5333: Intermittent transpilation failure: XSL race condition results in malformed Java in `eo-runtime`

morphqdd:
Closes #5333.

## Root cause

`Transpiling.TRAIN` was a single `static final Train<Shift>` shared across
every worker thread in the `Threaded` pool used to transpile all `.eo` files
in parallel. `TrClasspath`/`StClasspath` compile their XSL stylesheets
lazily on first use, and that compilation is not thread-safe — this is the
exact same root-cause shape as #5209/#5234, which fixed the analogous race
in the lint phase's `LtByXsl` (a lazily-compiled, `Unchecked(Sticky(...))`
XSL cache shared across the lint thread pool).

Concurrent threads racing on the same lazy-compilation cache inside the
shared `TRAIN`'s `Shift` objects could produce a truncated or garbled Java
file for whichever object happened to be mid-compilation when another
thread's compilation interfered — matching the observed symptom exactly:
a syntax error in a generated file whose EO source was untouched and
correct, disappearing on retry of the identical commit.

## Fix

Switch `TRAIN` from a shared `static final Train<Shift>` to a
`ThreadLocal<Train<Shift>>`. Each worker thread now gets its own
independently-compiled train — no shared lazy-compilation state to race
on — while still reusing its own compiled stylesheets across every source
it processes, so most of the caching benefit from compiling once is kept
(only the pool size, `availableProcessors() * 2`, worth of redundant
compilations happen instead of one).

## Testing

- Strengthened the existing `transpilesSeveralEoProgramsInParallel` test:
  it previously only checked the *count* of generated files, which
  wouldn't have caught this bug (file count stays correct even when a
  file's content is corrupted). It now also verifies each generated file
  contains the expected class declaration and has balanced braces/parens.
- Ran the full `eo-maven-plugin` module test suite locally (`mvn test -pl
  eo-maven-plugin`): 343 tests, 0 failures, 0 errors, 5 skipped (unchanged
  from before this PR).
- `mvn clean verify -Pqulice -DskipTests -pl eo-maven-plugin`: clean.

Note: like the original bug, this race is inherently intermittent — I
cannot deterministically reproduce the exact corruption in a unit test
(that's the nature of the bug per #5333's own description: "a clean local
build compiled without error, and simply re-running the job on the
identical commit turned it green"). The fix follows the established,
already-accepted precedent from #5209/#5234 for the identical failure
mode in a sibling XSL pipeline.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
@@ -62,24 +62,36 @@ final class Transpiling implements Step {
     static final String PRE = "5-pre-transpile";
 
     /**
-     * Parsing train with XSLs.
+     * Parsing train with XSLs, one instance per thread.
+     * <p>
+     *     A single shared instance is deliberately avoided: {@link TrClasspath}
+     *     and {@link StClasspath} compile their XSL stylesheets lazily on
+     *     first use, and that lazy compilation is not thread-safe. Sharing one
+     *     instance across the {@link Threaded} worker pool races on the same
+     *     underlying compilation cache and can produce truncated or garbled
+     *     Java output. Keeping one instance per thread still lets each worker
+     *     thread reuse its own compiled stylesheets across the sources it
+     *     processes.
+     * </p>
      */
-    static final Train<Shift> TRAIN = new TrFull(
-        new TrJoined<>(
-            new TrClasspath<>(
-                "/org/eolang/maven/transpile/set-locators.xsl",
-                "/org/eolang/maven/transpile/set-original-names.xsl",
-                "/org/eolang/maven/transpile/classes.xsl",
-                "/org/eolang/maven/transpile/tests.xsl",
-                "/org/eolang/maven/transpile/anonymous-to-nested.xsl",
-                "/org/eolang/maven/transpile/package.xsl",
-                "/org/eolang/maven/transpile/attrs.xsl",
-                "/org/eolang/maven/transpile/data.xsl"
-            ).back(),
-            new TrDefault<>(
-                new StClasspath(
-                    "/org/eolang/maven/transpile/to-java.xsl",
-                    String.format("disclaimer %s", new Disclaimer())
+    private static final ThreadLocal<Train<Shift>> TRAIN = ThreadLocal.withInitial(
+        () -> new TrFull(
+            new TrJoined<>(
+                new TrClasspath<>(
+                    "/org/eolang/maven/transpile/set-locators.xsl",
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (121 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+991](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5422, opened 2026-07-12
#5396: assert IllegalStateException in MjLintTest.reportsWhenObjectNameFails

### LINKED ISSUE #5396: Ambiguous `assertThrows` usage in `MjLintTest.reportsWhenObjectNameFails` impacts test precision and SonarCloud reliability rating
## What happens

SonarCloud's quality gate on `master` currently fails with `new_reliability_rating = D` (threshold: must be A). Querying the SonarCloud API confirms a single open `BUG`-type issue (rule `java:S5783`, severity CRITICAL, RELIABILITY impact HIGH) causing it:

```
eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java:385
"Refactor the code of the lambda to not have multiple invocations throwing the same checked exception."
```

The offending test:

```java
@Test
void reportsWhenObjectNameFails(@Mktmp final Path temp) {
    Assertions.assertThrows(
        Exception.class,
        () -> new FakeMaven(temp).withProgram("# App.").execute(new FakeMaven.Lint()),
        "MjLint's execution was not failed, but it should"
    );
}
```

Both `FakeMaven.withProgram(String...)` (`FakeMaven.java:270`) and `FakeMaven.execute(Class)` (`FakeMaven.java:164`) declare `throws IOException` — the same checked exception. Because both calls sit inside the same `assertThrows` lambda, the test cannot distinguish which one actually threw: if `withProgram` (setup) ever started throwing `IOException` for an unrelated reason, the test would still pass, silently masking a real regression in `execute` (the method actually under test) instead of failing for the right reason.

## Why it's wrong

`assertThrows` is meant to pinpoint that a *specific* call raises the expected exception. With two checked-exception-throwing calls chained in one lambda, the assertion is ambiguous about which call is actually being tested, weakening the test's ability to catch a real regression in `execute`.

## What should happen instead

The setup call (`withProgram`) should be moved outside the `assertThrows` lambda, so only `execute` — the method whose failure behavior is under test — remains inside it, e.g.:

```java
@Test
void reportsWhenObjectNameFails(@Mktmp final Path temp) throws IOException {
    final FakeMaven maven = new FakeMaven(temp).withProgram("# App.");
    Assertions.assertThrows(
        Exception.class,
        () -> maven.execute(new FakeMaven.Lint()),
        "MjLint's execution was not failed, but it should"
    );
}
```

This resolves the SonarCloud reliability rating and restores the test's precision.

morphqdd:
Closes #5396.

Both `FakeMaven.withProgram(String...)` and `FakeMaven.execute(Class)` declare `throws IOException`, so with `Exception.class` asserted, SonarCloud rule `java:S5783` (CRITICAL, the single issue behind master's `new_reliability_rating = D`) flags the lambda as ambiguous: either call could be the one throwing.

Fixed by narrowing the asserted type to `IllegalStateException` — the exception `Moja.execute()` actually wraps every mojo failure into. `withProgram` can only throw `IOException`, so the assertion now unambiguously targets `execute`, with no structural changes to the test at all: same one-liner, no extra local, no suppressions.

Verification:
- `mvn test -pl eo-maven-plugin -Dtest=MjLintTest`: 17 tests, 0 failures, 0 errors.
- `mvn verify -Pqulice -DskipTests -pl eo-maven-plugin`: clean.

### DIFF
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
@@ -390,7 +390,7 @@ final class MjLintTest {
     @Test
     void reportsWhenObjectNameFails(@Mktmp final Path temp) {
         Assertions.assertThrows(
-            Exception.class,
+            IllegalStateException.class,
             () -> new FakeMaven(temp).withProgram("# App.").execute(new FakeMaven.Lint()),
             "MjLint's execution was not failed, but it should"
         );


morphqdd:

yegor256:
@morphqdd this change looks worse than what we've had before

morphqdd:
@yegor256, WDYT?

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1135](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5423, opened 2026-07-12
#5419: reject out-of-range \NNN octal escapes instead of silently emitting a multi-byte value

### LINKED ISSUE #5419: Octal escape `\NNN` in EO strings accepts out-of-range values (256-511), silently emitting a 2-byte UTF-8 sequence instead of one byte
`PARSER_SPEC.md` R-9.7.3 defines the `\NNN` octal-byte escape as: `N` ∈ `[0-7]`, length 1–3 digits, **value ≤ 0o377 (255 decimal)**, with the grammar pattern explicitly restricting the escape to `\` followed by an optional `[0-3]`, an optional `[0-7]`, and a required `[0-7]` — i.e. the leading digit of a 3-digit escape is restricted to `[0-3]` precisely so the decoded value can never exceed one byte.

`Emissions.appendOctal` (`eo-parser/src/main/java/org/eolang/parser/Emissions.java:354-365`) does not enforce this. It greedily consumes up to 3 digits in `[0-7]` with no check on the leading digit:

```java
private static int appendOctal(
    final StringBuilder out, final String body, final int start
) {
    int cursor = start;
    int value = 0;
    while (cursor < body.length()
        && cursor < start + 3
        && body.charAt(cursor) >= '0' && body.charAt(cursor) <= '7') {
        value = value * 8 + body.charAt(cursor) - '0';
        cursor = cursor + 1;
    }
    out.append((char) value);
    return cursor;
}
```

**What happens:** an octal escape with a leading digit of 4–7 (decoded value 256–511, outside the spec's 1-byte range) is silently accepted instead of being rejected. Per R-9.7.4, the decoded character is fed straight into the `Φ.bytes` UTF-8 carrier, so a value above 255 gets UTF-8-encoded into **two bytes** instead of the single byte the escape is supposed to represent:

- `"\477"` (octal 477 = 319 decimal) → emits bytes `C4 BF` (2 bytes) instead of a lexical error.
- `"\777"` (511 decimal) → emits `C7 BF` (2 bytes).
- `"\400"` (256 decimal — one past the 255 limit) → emits `C4 80` (2 bytes).

None of the existing escape-sequence test packs (`octal-escape-in-string.yaml`, `escape-sequences.yaml`) cover an out-of-range leading digit, so this gap is untested.

**What should happen:** an octal escape whose value exceeds 0o377 (255) — equivalently, whose leading digit (when 3 digits are present) is not in `[0-3]` — should be rejected as a lexical error, per R-9.7.3's grammar, or at minimum the resulting byte should never silently become a 2-byte sequence.

morphqdd:
Closes #5419.

Per `PARSER_SPEC.md` R-9.7.3, a `\NNN` octal escape must have value ≤ 0o377
(255) — one byte, with the grammar's leading digit (when 3 digits are
present) restricted to `[0-3]` precisely so the decoded value can never
exceed a single byte. `Emissions.appendOctal` greedily consumed up to 3
digits in `[0-7]` with no check on the resulting value, so a leading digit
of 4-7 (decoded value 256-511) was silently accepted and, once fed into the
`Φ.bytes` UTF-8 carrier, produced a 2-byte sequence instead of the single
byte the escape claims to represent (`"\477"` → `C4 BF`, `"\777"` → `C7 BF`,
`"\400"` → `C4 80`).

`appendOctal` now throws `NumberFormatException` when the decoded value
exceeds `0o377`. This is caught by the existing `STRING`-value handling in
`Emissions.emit`, which already converts a `NumberFormatException` from the
sibling `\uXXXX` unicode-escape path into a proper `ParseError` — the same
mechanism now covers octal, surfacing as an `/object/errors/error` entry
instead of corrupting the byte carrier.

Note: while investigating, I found that octal escapes with an in-range
value ≥ 128 (e.g. `\377` = 255, perfectly valid per spec) *also* currently
produce a 2-byte UTF-8 sequence rather than a single raw byte, because the
decoded `char` later goes through UTF-8 string encoding regardless of its
value. That's a separate, deeper issue (the escape's "one byte" promise
isn't kept for the *entire* upper half of the valid range, not just the
overflow case this PR closes) — I didn't fold it into this PR since fixing
it changes how the byte carrier itself is populated, a larger change than
"reject an out-of-range value." Happy to file it separately if useful.

Verification:
- Added `EoSyntaxTest.rejectsOutOfRangeOctalEscape`, confirming `"\477"`
  now surfaces an `/object/errors/error` entry mentioning `octal` instead
  of silently parsing.
- Full `eo-parser` test suite: 1455 tests, 0 failures, 0 errors.
- `mvn clean verify -Pqulice -DskipTests -pl eo-parser`: clean.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -24,6 +24,11 @@ import java.util.List;
 final class Emissions {
 
     /**
+     * Maximum value of a {@code \NNN} octal byte escape (0o377, one byte).
+     */
+    private static final int MAX_OCTAL_BYTE = 0xFF;
+
+    /**
      * No instances.
      */
     private Emissions() {
@@ -144,7 +149,7 @@ final class Emissions {
                 unescaped = Emissions.unescape(value.raw());
             } catch (final NumberFormatException ex) {
                 final ParseError error = new ParseError(
-                    line, value.pos(), "invalid unicode escape in string literal"
+                    line, value.pos(), "invalid unicode or octal escape in string literal"
                 );
                 error.initCause(ex);
                 throw error;
@@ -361,6 +366,14 @@ final class Emissions {
             && body.charAt(cursor) >= '0' && body.charAt(cursor) <= '7') {
             value = value * 8 + body.charAt(cursor) - '0';
             cursor = cursor + 1;
+        }
+        if (value > Emissions.MAX_OCTAL_BYTE) {
+            throw new NumberFormatException(
+                String.format(
+                    "octal escape \\%s is out of range: value %d exceeds the 1-byte limit of 0o377 (255)",
+                    body.substring(start, cursor), value
+                )
+            );
         }
         out.append((char) value);
         return cursor;
--- a/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
@@ -586,6 +586,19 @@ final class EoSyntaxTest {
     }
 
     @Test
+    void rejectsOutOfRangeOctalEscape() throws Exception {
+        MatcherAssert.assertThat(
+            "an out-of-range \\NNN octal escape (value > 0o377) must show up as an /object/errors/error entry, not silently emit a multi-byte value",
+            EoSyntaxTest.raw(
+                String.join(String.valueOf((char) 10), "[] > foo", "  \"\\477\" > @")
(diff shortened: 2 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1015](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5424, opened 2026-07-12
refactor(#5393): implement `seq` via direct self-recursion

### LINKED ISSUE #5393: `seq` implementation relies on a complex internal structure instead of simpler self-recursion
The `seq` object in `eo-runtime/src/main/eo/seq.eo:20-42` is implemented with an internal `loop` helper object, a `end!` marker (`steps.length.plus -1`) threaded in from the outer scope, and a `and.`/`or.`/`if.` construction that abuses short-circuit dataization to sequence the steps. What is conceptually a plain left-to-right fold over the tuple is spread across two objects and a marker, which makes it hard to read and hard to reason about.

The same behaviour can be expressed as a direct recursion on `seq` itself, using only the tuple primitives that already exist (`head`, `tail`, `length` in `tuple.eo:11`):

```
seq []           --> true
seq [x]          --> x
seq [x, rest...]  --> dataize x, then seq rest
```

This would remove the `loop` object and the `end!` marker entirely, leaving `seq` as its own definition. The one design point to settle is the recursive step: it needs a "dataize `head`, then behave as `seq tail`" combinator, which is the essence of sequencing. Today `loop` supplies that via the `and.` short-circuit trick; the proposal is to inline it into `seq`. If it inlines cleanly the helper disappears; if it cannot be inlined without reintroducing a named helper, the refactor buys nothing, so that is the thing to prove before committing. Existing tests in `seq.eo` (including `tests-very-long-seq`) should keep the recursion depth honest.

yegor256:
Closes #5393.

## What

Reimplements the `seq` object in `eo-runtime/src/main/eo/seq.eo` as a plain
left-to-right fold expressed through **direct recursion on `seq` itself**,
removing the internal `loop` helper object and the `end!` marker
(`steps.length.plus -1`).

Before:

```
[steps] > seq
  if. > @
    steps.length.eq 0
    true
    if.
      steps.length.eq 1
      steps.head
      loop steps
  steps.length.plus -1 > end!

  [tup] > loop
    if. > @
      and.
        tup.length.gt 1
        loop tup.tail
      tup.head
      or.
        (dataized tup.head).as-bool.eq --
        tup.length.eq end
```

After:

```
[steps] > seq
  if. > @
    steps.length.eq 0
    true
    if.
      steps.length.eq 1
      steps.head
      if.
        (dataized (seq steps.tail)).as-bool
        steps.head
        steps.head
```

## Why

As described in #5393, the previous implementation spread what is
conceptually a simple fold across two objects plus a threaded marker, and
relied on an `and.`/`or.`/`--` short-circuit trick to sequence the steps.
This was hard to read and hard to reason about.

## How it works

An EO tuple's `.head` is its **last** element and `.tail` is everything
before it (see `tuple.eo`). So the three cases are:

```
seq []            -> true
seq [x]           -> x
seq [x, rest...]  -> dataize the earlier steps in order, then return the last
```

The recursive step dataizes `seq steps.tail` — which forces every earlier
step's side effects, left to right — and then returns `steps.head`. The
`if.` with two identical branches is the object-preserving way to sequence
"evaluate the condition for its side effect, then yield `steps.head`",
since `and`/`or`/`eq` all coerce their result to `bool` and would lose the
last step's actual value.

The helper `loop` and the `end!` marker disappear entirely, leaving `seq`
as its own definition — exactly the outcome the issue asked to prove.

## Verification

Built `eo-runtime` and ran the generated `org.eolang.EOseqTest`:

```
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0 -- in org.eolang.EOseqTest
BUILD SUCCESS
```

All existing `seq` tests pass unchanged, including `tests-very-long-seq`
(~40 steps), which keeps the recursion depth honest.

https://claude.ai/code/session_01VQNuXapUjBKHVbfx3PRXES

---
_Generated by [Claude Code](https://claude.ai/code/session_01VQNuXapUjBKHVbfx3PRXES)_


yegor256:
@maxonfjvipon good?

maxonfjvipon (approved):

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1638](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (16) hits-of-code; +4 to give you at least something.  Your running score is [+613](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5425, opened 2026-07-12
Auto-home bare same-package references (#4529)

### LINKED ISSUE #4529: Local objects must be turned into aliases automatically
This is my code:

```
+package foo

[] > x
  bar 42 > @
```

The `bar` object exists in the `foo` package. Let's create an additional optimization that will modify, in compile time, the code to make it look like this:

```
+alias foo.bar
+package foo

[] > x
  bar 42 > @
```

Without this, I need to manually add many `+alias` metas, which is an annoying and error-prone operation.

yegor256:
Closes #4529.

## What

When a program declares a `+package` meta, a bare reference to an object that belongs to the **same package** is now resolved automatically into that package, as if a `+alias` had been written by hand. For example, in package `foo`:

```
+package foo

[] > x
  bar 42 > @
```

`bar` is compiled to `Φ.foo.bar` (previously it defaulted to the root `Φ.bar`), so same-package objects no longer need manual `+alias foo.bar` metas.

## How

`add-default-package.xsl` takes an `objects` parameter — the space-separated **names of the local objects that belong to a package**, filled in by the `eo-maven-plugin` from the registered catalog. The rule for a bare name in a program with a `+package` meta:

- the name is one of those local package objects → home it into the current package (`Φ.<pkg>.<name>`);
- otherwise → leave it at the root `Φ` (treated as a global — the default behaviour).

This keeps globals correct without any allow/deny gymnastics: a bare `seq` (not a local package object) stays `Φ.seq`, and when the list is empty (e.g. the parser used stand-alone) every bare reference falls back to the root `Φ`. Aliased references are untouched (handled earlier in the chain).

## Changes

- **`add-default-package.xsl`** — new `objects` parameter and an `eo:homed()` function; `@base`/`@atom` bare references are homed into the current package only when the name is a local package object, otherwise into `Φ`.
- **`Canonical`** (new, in `eo-parser`) — wraps the canonical XSL pipeline as a `UnaryOperator<XML>`, injecting the `objects` parameter into `add-default-package.xsl`; the pipeline is built lazily and once, thread-safely (cactoos `Synced`/`Sticky`), and reused across the concurrent parse.
- **`EoSyntax`** — `CANONICAL` delegates to `Canonical`; a public `(Input, UnaryOperator<XML>)` constructor lets the plugin supply the objects-aware pipeline.
- **`Parsing`** — builds the objects-aware pipeline once from the local package-object names and reuses it for every source; the object-set digest is part of the parse-cache key so cached XMIRs are invalidated when the set changes.
- **`EoSource`** — accepts the shared transform.

## Testing

- `eo-parser`: full suite passes, incl. new `EoSyntaxTest` cases (bare reference homed into the package when it is a local package object; kept at root otherwise).
- `eo-maven-plugin`: `MjParseTest` (incl. new end-to-end cases across two files in one package), `CompilingTest`, `MjTranspileTest` — all green under `-Pqulice`.
- Real `eo-runtime`: all 113 sources parse concurrently with no error; an exact-match audit confirms no global (`seq`, `malloc`, `number`, …) is wrongly homed into a package, and bare `seq` stays `Φ.seq`.

## Notes

- Homing applies to the current project's local package objects known at parse time. Referencing a same-package object that lives in an external dependency still needs an explicit `+alias`; bare references to runtime globals are unaffected (they stay at the root).

https://claude.ai/code/session_01LUBw9ufavQSMDmBxnqJ9Nw


yegor256:
@maxonfjvipon take a look

eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
```diff
@@ -115,10 +118,15 @@ final class Parsing implements Step {
...
     public void exec() {
         final Collection<TjForeign> sources = this.tojos.withSources();
-        final int total = new Threaded<>(
-            new Filtered<>(TjForeign::notParsed, sources),
-            this::parsed
-        ).total();
+        final String objects = sources.stream()
+            .map(TjForeign::identifier)
+            .filter(id -> id.contains("."))
+            .map(id -> id.substring(id.lastIndexOf('.') + 1))
```
maxonfjvipon: This flattens every object across all packages into one set of *simple* names, so `eo:homed` homes a bare name into the current package whenever the name exists in *any* package. `ss` and `tt` already both define `contains`/`index-of`/`last-index-of`, and `ss/index-of.eo` uses a bare `number` — the day any package gains a `number` object, that reference silently re-homes to `Φ.ss.number`. Keying on the qualified name avoids this: pass the `.`-containing identifiers as `objects` unchanged, then home only when `concat($package, '.', $name)` matches. Same single pipeline, no simple-name collisions.

eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
```diff
@@ -115,10 +118,15 @@ final class Parsing implements Step {
...
+            .distinct()
+            .collect(Collectors.joining(" "));
+        final int total = this.parsed(
+            sources, new Canonical(objects), Integer.toHexString(objects.hashCode())
```
maxonfjvipon: `String.hashCode` is a collidable 32-bit hash, and `objects` is built with `distinct()` over `withSources()`, whose order isn't guaranteed. A collision gives a stale cache hit; an order flip gives a needless miss. Sorting the names and using a real digest makes this deterministic.

eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
yegor256: Fixed in 501c5d1. `add-default-package.xsl` now keys on the qualified name: the plugin passes the `.`-containing identifiers (`package.name`) unchanged, and a bare name is homed only when `concat($package, '.', $name)` is in the list. So the simple-name collision is gone — verified against eo-runtime, the bare `number` in `ss/index-of.eo` stays `Φ.number` (33 occurrences, 0 `Φ.ss.number`).

---

eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
yegor256: Fixed in 501c5d1. The qualified names are now sorted and hashed with SHA-256 (cactoos `Sha256DigestOf` + `HexOf`) instead of the order-sensitive, 32-bit `String.hashCode`, so the cache key is deterministic and collision-resistant.

---

yegor256:
@maxonfjvipon take a look again please

eo-parser/src/main/resources/org/eolang/parser/parse/add-default-package.xsl
```diff
@@ -21,9 +21,54 @@
...
+  <!-- Qualified names of all local package objects as a sequence. -->
+  <xsl:variable name="known" select="tokenize($objects, '\s+')[. != '']"/>
+  <!--
+  Resolve a bare name to its fully qualified name. If the current
+  program has a package and "package.name" is a local package object,
+  the name is homed into the package; otherwise it goes to the root "Φ".
+  -->
+  <xsl:function name="eo:homed" as="xs:string">
+    <xsl:param name="name" as="xs:string"/>
+    <xsl:sequence select="if ($package != '' and concat($package, '.', $name) = $known) then concat('Φ.', $package, '.', $name) else concat('Φ.', $name)"/>
```
maxonfjvipon: Factor out the repeated `concat($package, '.', $name)` and collapse the two `concat('Φ.', …)` branches into one — clears the Codacy hint and reads cleaner:

```suggestion
    <xsl:variable name="local" select="concat($package, '.', $name)"/>
    <xsl:sequence select="concat('Φ.', if ($package != '' and $local = $known) then $local else $name)"/>
```

yegor256:
@maxonfjvipon good now?

maxonfjvipon (approved):

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+616](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (475 >= 100); -8 for way too many hits-of-code (475 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1720](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5428, opened 2026-07-13
Split ms/numbers into standalone minimum and maximum

### LINKED ISSUE #5427: Split ms/numbers.eo into standalone minimum and maximum objects
In `eo-runtime/src/main/eo/ms/numbers.eo:16-42` the `min` and `max` reductions live as sub-objects of a `numbers` wrapper that just decorates its `sequence`. A caller has to write `(numbers seq).max fallback` to reach either one, and the two operations share a file even though they are independent.

Everywhere else the `ms` package keeps one object per file, so bundling two reductions behind a wrapper is out of line with the rest of the package. The wrapper adds nothing of its own beyond `sequence > @`, and there are no other users of `numbers` in the tree, so it is pure indirection.

The fix is to extract each reduction into its own object: `ms/maximum.eo` and `ms/minimum.eo`, each shaped as `[sequence cant-max]` / `[sequence cant-min]` with the empty-sequence fallback preserved, move the existing tests along with them, and delete `numbers.eo`. Callers then reach `ms.maximum` and `ms.minimum` directly.

yegor256:
This pulls the `min` and `max` reductions out of `ms/numbers.eo` and gives each its own object: `ms/maximum.eo` and `ms/minimum.eo`. Each one takes the sequence directly along with a fallback that stands in when the sequence is empty, and `numbers.eo` is deleted. All the old tests move across with them.

The `numbers` wrapper only decorated its `sequence` and hid the two reductions behind `(numbers seq).max` / `.min`. Nothing else in the tree used it, and every other object in the `ms` package lives one-per-file, so the wrapper was just indirection. Callers now reach `ms.maximum` and `ms.minimum` straight away.

Closes #5427.


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (210 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1615](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5434, opened 2026-07-13
Drop the try and error objects

maxonfjvipon:
The catchable `error` object and the `try` that intercepted it are gone. Every runtime failure now becomes an uncatchable ⊥: `PhSafe` wraps a failing dispatch into an `ExFailure` that carries the source location and the original cause, `Dataized.take()` just propagates it, and the last remaining producer of `error` — `regex`'s non-matched block — now returns a ⊥ through a new factory instead of building an `error`.

```java
public static PhTerminator withCause(final String cause) {
    final PhTerminator term = new PhTerminator();
    term.put(0, new Data.ToPhi(cause));
    return term;
}
```

This continues #5261: a failure is represented by ⊥ alone, and nothing in EO can catch it. There is deliberately no recovery mechanism left here — the `:?` substitution operator that will let a program replace a ⊥ at a chosen dispatch site is separate, future work. Until it lands, forcing an un-substituted ⊥ terminates the program, which is the intended fail-fast behaviour.

Along with `EOerror`, `EOtry`, `error.eo` and `try.eo`, the two objects' own tests are deleted; the four tests that asserted `EOerror.ExError` now expect `ExFailure`.


yegor256 (approved):

0crat:
@yegor256 Thanks for the review! You've earned +15 points for this: +12 as a basis; +10 for the 1052 hits-of-code that you reviewed; -8 for absolutely no comments posted.  Your running score is [+1634](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1052 >= 100); -8 for way too many hits-of-code (1052 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+609](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5436, opened 2026-07-13
test(integration): #5429 combine proxy handlers instead of overwriting one with the other

yegor256:
Closes part of #5429.

## Problem

In `ProxyIT`, both test methods set up the local Jetty proxy like this:

```java
proxy.setHandler(new ProxyHandler.Forward());
proxy.setHandler(new ConnectHandler());
```

`Server.setHandler(...)` replaces the current handler, so the second call
discards `ProxyHandler.Forward` and only `ConnectHandler` (CONNECT tunneling)
was ever installed — plain HTTP forward proxying was never exercised. This is
a recurrence of #5110.

## Change

- Extract a shared `handler()` helper that nests `ProxyHandler.Forward` inside
  `ConnectHandler`. `ConnectHandler` is a handler *wrapper*: it serves `CONNECT`
  requests itself and delegates every other (plain HTTP) request to the nested
  handler, so both behaviours are installed at once — the correct Jetty‑12
  forward‑proxy setup.
- Both `checksThatProxyIsWorking` and `checksThatWeCanCompileTheProgramWithProxySet`
  now use this helper, removing the duplicated (and buggy) wiring.

## Not in this PR

`checksThatWeCanCompileTheProgramWithProxySet` stays `@Disabled`. Its flakiness
also comes from resolving `eo-runtime`/`jna` from **live Maven Central** through
the proxy (`@WeAreOnline`), which is the other half of the fix called out in
#5429. Removing that live‑network dependency (e.g. serving a local Maven repo
over HTTP through the forward proxy) and re-enabling the test is a larger,
network-dependent change left as follow-up so this PR stays small and verifiable.
The `@Disabled` reason is updated to reflect the true remaining cause.

## Verification

The new handler wiring was compiled and type-checked against `jetty-server`/
`jetty-proxy` 12.1.11: `ConnectHandler extends Handler.Wrapper` (exposes
`setHandler(Handler)`) and `ProxyHandler.Forward` is a `Handler`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01C81Kagkzi9vfibijdAsGtP

---
_Generated by [Claude Code](https://claude.ai/code/session_01C81Kagkzi9vfibijdAsGtP)_

Issue #5429, opened 2026-07-13
Intermittent failure in ProxyIT.checksThatWeCanCompileTheProgramWithProxySet due to proxy setup


yegor256:
The integration test `checksThatWeCanCompileTheProgramWithProxySet` in `eo-integration-tests/src/test/java/org/eolang/maven/ProxyIT.java:63` fails intermittently. It stands up a local Jetty forward proxy on a random free port and then runs a full Farea `package` build that resolves `eo-runtime` and `jna` from Maven Central through that proxy over the live network (`@WeAreOnline`, `@MayBeSlow`), asserting that the log contains `BUILD SUCCESS`. When the resolution through the proxy hiccups the build exits with `0x0001` and the assertion fails.

This is not caused by any particular change: the same failure appears on master's own `integration` runs (for example commit 8a18214181), interleaved with green runs on identical code, which is the classic signature of a flaky test. One likely contributor is in the proxy setup: `proxy.setHandler(new ProxyHandler.Forward())` is immediately followed by `proxy.setHandler(new ConnectHandler())` (lines 65-67, and again 44-46), so the second call replaces the first and `ProxyHandler.Forward()` never runs — only CONNECT tunneling is exercised. This is a recurrence of #5110, which was closed as completed before the test was re-enabled in #5152.

For now, mark the test `@Disabled` referencing this issue so the pipeline stops going red on unrelated PRs. The real fix is to install both handlers instead of overwriting one with the other (combine them in a handler collection) and to stop depending on live Maven Central through the proxy, then re-enable it.

VasilevNStas:
@yegor256 

This can be closed: the reported failure is fixed and merged.

## Evidence

- The exact problem described here — `proxy.setHandler(new ProxyHandler.Forward())` immediately overwritten by `proxy.setHandler(new ConnectHandler())` — was fixed by PR #5436 ("combine proxy handlers instead of overwriting one with the other"), merged on 2026-07-16.
- The current code nests the handlers instead: `eo-integration-tests/src/test/java/org/eolang/maven/ProxyIT.java:109-110` builds a `ConnectHandler` that wraps a `ProxyHandler.Forward`, so both CONNECT and plain forwarding are exercised.
- PR #5441 (merged) made the test hermetic and re-enabled the compile test, addressing the network-flakiness half of the report.

PR #5441, opened 2026-07-13
test(integration): #5429 make ProxyIT hermetic and re-enable the compile test

yegor256:
Follow-up to #5436, completing the remaining half of #5429.

## Problem

`checksThatWeCanCompileTheProgramWithProxySet` was `@Disabled` because it
resolved `eo-runtime`, `jna` and their transitive dependencies from **live
Maven Central** through the proxy (`@WeAreOnline`, `@MayBeSlow`). When the
resolution hiccuped over the network the build exited `0x0001` and the test
failed intermittently.

## Change

The test is now hermetic — it no longer touches any network beyond localhost:

- A local Jetty HTTP server (`repository()`) serves the local Maven repository
  (`~/.m2/repository`) via a `ResourceHandler`.
- `settings.xml` gains a `<mirror>` with `<mirrorOf>*</mirrorOf>` pointing at
  that server, so Maven routes **all** artifact requests to it.
- The mirror URL is HTTP, so requests go through the proxy's **forward** path
  (`ProxyHandler.Forward`) — the exact path fixed in #5436 — rather than a
  CONNECT tunnel to the real Central. The proxy is therefore genuinely
  exercised.
- An isolated, empty `-Dmaven.repo.local` forces Maven to actually fetch every
  artifact through the proxy on each run instead of reading a warm `~/.m2`.

The build's dependency closure is guaranteed to be available in `~/.m2` because
the `integration` job runs `mvn clean install` on the whole reactor (with a
cached `~/.m2/repository`) before the integration tests execute, and modules
build in dependency order — so `eo-runtime`, `eo-maven-plugin`, `jna` and the
core lifecycle plugins are all present by the time `ProxyIT` runs.

`checksThatProxyIsWorking` is unchanged; it still verifies CONNECT tunneling
against `objectionary.com`.

## Verification

The forward-proxy → local-repository-server mechanism (the only genuinely new
runtime behaviour, and a path that was dead before #5436) was prototyped and
run in isolation against Jetty 12.1.11: a Maven-style artifact path
(`/org/eolang/.../x-1.0.pom`) requested through the proxy is served correctly
from the `ResourceHandler` — no live network involved. The full integration
build itself is validated by CI, since it requires the whole reactor in
`~/.m2`.

https://claude.ai/code/session_01C81Kagkzi9vfibijdAsGtP

---
_Generated by [Claude Code](https://claude.ai/code/session_01C81Kagkzi9vfibijdAsGtP)_

### DIFF
--- a/eo-integration-tests/pom.xml
+++ b/eo-integration-tests/pom.xml
@@ -197,6 +197,12 @@
       <version>12.1.11</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-util</artifactId>
+      <version>12.1.11</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/eo-integration-tests/src/test/java/org/eolang/maven/ProxyIT.java
+++ b/eo-integration-tests/src/test/java/org/eolang/maven/ProxyIT.java
@@ -19,16 +19,19 @@ import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import org.cactoos.io.ResourceOf;
 import org.cactoos.text.TextOf;
 import org.cactoos.text.UncheckedText;
 import org.eclipse.jetty.proxy.ProxyHandler;
 import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.Server;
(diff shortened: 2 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (112 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1646](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5446, opened 2026-07-13
Fill formation void via inline binding

### LINKED ISSUE #5432: EO lacks support for anonymous formations with application like in phi-calculus
I can't do this in EO, but can do it in phi-calculus:

```
[ a -> 7, x -> ? ] ( x -> 42 )
```

I suggest we make it possible in EO too.

Maybe using this syntax:

```
[x]
  7 > a
  42:x 
```

Here, if attribute is void and at the same time specified with the help of `:<name>`, we treat formation as formation+application.

This will make it possible to use anonymous formations as functions (Haskell) or lambda expressions (Java).

yegor256:
This makes anonymous formations usable as functions. A body line of the form `value:name` (an inline binding with no name suffix) now fills the void `name` of the enclosing formation, so the whole thing reads as a formation applied to an argument — the phi-calculus `[ a -> 7, x -> ? ] ( x -> 42 )`. For example:

```
[x] > foo
  7 > a
  x.plus a > @
  42:x
```

binds `a -> 7`, `phi -> x.plus a`, and applies `x -> 42`. The `42:x` line emits an ordinary `<o base='...' as='x'/>` argument child of the formation, exactly like an application argument, while the void declaration for `x` is kept and filled. This is the `:x` spelling from the discussion in #5432, which mirrors application notation.

A binding that also carries a name suffix (`bar:tag > x`) stays rejected, since a named attribute is not an argument. The change touches the parser only: `Level` gains a `bound` flag, `Bindings.observeChild` accepts the void-filling argument, and the naming check lets it through. Spec rules R-3.12.3, R-3.12.5, R-5.3.1 and R-6.2.1 are updated to match.

Runtime dataization of an anonymous formation applied to its arguments is the natural next step and is left out of this PR; here I only teach the parser the syntax and the XMIR it emits.

Closes #5432

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -541,8 +541,9 @@ world""" > text-value                 ← multi-line string named text-value
 
 R-3.12.1. After an object body in an argument position, `:label` may appear, setting the attribute the argument attaches to.
 R-3.12.2. `:N` (digit) maps to `α0`, `α1`, … positional slot names.
-R-3.12.3. Binding is forbidden inside formation context — plain children carry their name via the name suffix, not the binding (§6.2).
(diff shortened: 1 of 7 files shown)


yegor256:
@maxonfjvipon take a look

PR #5448, opened 2026-07-13
feat(#5443): separated without from ss.list

Shaurya2k06:
In this PR, I extracted the `without` object from `ss.list` into its own file `ss/without.eo`, mirroring the pattern used by other already-extracted siblings (`withouti`, `reduced`, `eachi`, `mappedi`, etc.) and matching the `ms.real` decomposition from #4751.

The new free-standing object takes the list as an explicit free attribute (`[lst element] > without`) instead of relying on `^`. The `tests-list-without` test moved with it and was updated from method-call (`without.`) to free-function (`without`) syntax. The unused `+alias ss.reduced` was dropped from `list.eo` since `without` was its only consumer there.

This is one of several extractions toward fully dissolving `ss.list` (#5443); future PRs can extract `each`, `eq`, `concat`, `front`, `back`, `slice`, `shifted`, `with`/`withi`, `is-empty`, etc. one at a time, then delete `list.eo`.

Verified locally: `mvn clean install -Pqulice -PskipITs` passes, and the affected tests (`EOss.EOlistTest` 35/35, `EOss.EOwithoutTest` 1/1) all pass.

Partial fix for #5443

Issue #5443, opened 2026-07-13
Redundancy in ss/list.eo methods and standalone objects causing inconsistency in codebase


yegor256:
`ss/list.eo` is a decorator over `tuple` that still bundles a dozen operations as methods: `is-empty`, `sorted`, `with`, `withi`, `each`, `without`, `eq`, `concat`, `front`, `back`, `slice`, `shifted` (`eo-runtime/src/main/eo/ss/list.eo:26-157`). Meanwhile the project has already extracted the same class of operations into standalone objects that take the tuple as their first free attribute — `ss/reduced.eo`, `ss/reducedi.eo`, `ss/index-of.eo`, `ss/last-index-of.eo`, `ss/filtered.eo`, `ss/mapped.eo`, `ss/eachi.eo`, `ss/withouti.eo`, `ss/contains.eo`.

Two coexisting styles for the same functionality is redundant. `list` adds nothing over `tuple` except these methods, so once each method lives as its own object the `list` decorator has no reason to exist. Collapsing to one style (standalone objects) is more consistent and drops one object plus its `+alias ss.list` dependency from ~26 call sites.

Suggested fix: for each remaining method, add a standalone `ss` object taking the tuple as the first argument (mirroring `reduced`/`index-of`), migrate callers from `list.<method>` to the new object, and finally delete `ss/list.eo`. This is best done incrementally — one method per pull request — rather than in a single sweep.

Shaurya2k06:
Hey @yegor256 , I'd love to take this up!

yegor256:
@Shaurya2k06 go ahead

Shaurya2k06:
Hey @yegor256 , Would you like me to do this in a set of pull requests? We have to deal with each, eq, concat, front/back/slice/shifted, with/withi, is-empty, then delete list.eo

yegor256:
@Shaurya2k06 of course, better submit a set of PRs

praneethsai050-cmd:
Hello @yegor256,

I've reviewed the issue in detail. This is a significant code cleanup that will improve consistency. I'd like to contribute by executing this refactoring meticulously.

Here is my proposed plan, which builds upon the initial suggestions:

1. Comprehensive Pre-Migration Audit
Before writing any code, I will create a complete inventory of all methods in ss/list.eo that need to be migrated. Based on the issue, this includes:
 is-empty , sorted , with / withi , each , without , eq , concat , front / back , slice / shifted

2. Strategy for Each Method
[...]

Issue #4751, opened 2025-12-14
`ms.real` Object Incorrectly Used as Utility Class


yegor256:
Currently, we have a number of utility methods in the `ms.real` object. Let's create new objects for each method, like `ms.sqrt`, `ms.acos`, etc.

yegor256:
`real` object is gone

PR #5449, opened 2026-07-13
feat(#5443): separated each from ss.list

Shaurya2k06:
In this PR, I extracted the `each` object from `ss.list` into its own file `ss/each.eo`, mirroring the pattern used by other already-extracted siblings (`eachi`, `withouti`, `reduced`, etc.) and matching the `ms.real` decomposition from #4751.

The new free-standing object takes the list as an explicit free attribute (`[lst func] > each`) instead of relying on `^`. The `tests-iterates-with-each` test moved with it and was updated from method-call (`.each`) to free-function (`each`) syntax. The unused `+alias ss.eachi` was dropped from `list.eo` since `each` was its only consumer there.

This is one of several extractions toward fully dissolving `ss.list` (#5443). Independent of #5448 (`without`); further PRs can extract `eq`, `concat`, `front`, `back`, `slice`, `shifted`, `with`/`withi`, `is-empty`, etc. one at a time, then delete `list.eo`.

Verified locally: `mvn clean install -Pqulice -PskipITs` passes, and the affected tests (`EOss.EOlistTest` 35/35, `EOss.EOeachTest` 1/1) all pass.

Partial fix for #5443

### DIFF
--- a/eo-runtime/src/main/eo/ss/each.eo
+++ b/eo-runtime/src/main/eo/ss/each.eo
@@ -1,0 +1,28 @@
+# For each collection element dataize the object.
+# Here "func" must be an abstract object with one free attribute:
+# the element of the collection. The result of `func` must be dataizable.
+
++alias ss.eachi
++alias ss.list
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package ss
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[lst func] > each
+  eachi > @
+    lst
+    func item > [item index] >>
+
+  [] +> tests-iterates-with-each
+    eq. > @
+      malloc.for
+        0
+        [m]
+          each > @
+            list
+              * 1 2 3
+            ^.m.put (^.m.as-number.plus i) > [i] >>
+      6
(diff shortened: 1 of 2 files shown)


eo-runtime/src/main/eo/ss/each.eo
```diff
@@ -0,0 +1,28 @@
...
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[lst func] > each
+  eachi > @
+    lst
+    func item > [item index] >>
+
+  [] +> tests-iterates-with-each
```
yegor256, line 19: @Shaurya2k06 since you are at it now, add more tests for this object

Shaurya2k06: Great, will do!

Shaurya2k06:
Rebased onto current `master` and added four more tests in `ss/each.eo`: empty list, single element, last `func` result, and a longer sum.

yegor256:
@Shaurya2k06 merge conflict here

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1705](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@shaurya2k06 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+32](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5453, opened 2026-07-13
feat(#5443): separated back from ss.list

Shaurya2k06:
In this PR, I extracted the `back` object from `ss.list` into its own file `ss/back.eo`, mirroring the pattern used by other already-extracted siblings (`without`, `concat`, `eachi`, etc.) and matching the `ms.real` decomposition from #4751.

The new free-standing object takes the list as an explicit free attribute (`[lst index] > back`) instead of relying on `^`. The three back tests moved with it and were updated from method-call (`back.`) to free-function (`back`) syntax. Remaining call sites in `list.front` and `list.withi` were updated the same way, with `+alias ss.back` added to `list.eo`.

This is one of several extractions toward fully dissolving `ss.list` (#5443). Further PRs can extract `front`, `slice`, `shifted`, `with`/`withi`, `is-empty`, etc. one at a time, then delete `list.eo`.

Verified locally: `mvn clean install -Pqulice -PskipITs` passes, and the affected tests (`EOss.EOlistTest` 32/32, `EOss.EObackTest` 3/3) all pass.

Partial fix for #5443

### DIFF
--- a/eo-runtime/src/main/eo/ss/back.eo
+++ b/eo-runtime/src/main/eo/ss/back.eo
@@ -1,0 +1,49 @@
+# Elements of `lst` from `index` to the end (supports oversized index).
+
++alias ss.list
++alias ss.reducedi
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package ss
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[lst index] > back
+  if. > @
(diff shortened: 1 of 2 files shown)


0crat:
@shaurya2k06 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5451, opened 2026-07-13
#5444: require a blank line before every +> test attribute

morphqdd:
Closes #5444.

The parser now requires exactly one blank line in front of every `+>` test attribute; a test sitting directly under the previous non-blank line is reported as an error (recoverable, emitted into XMIR like the sibling R-6.5.4 check).

Changes:
- `Blanks.checkTest` — new helper mirroring `Blanks.checkPlain`: reports `missing blank line before a '+>' test attribute (R-6.5.3)` when `pendingBlanks() == 0`.
- `LnFormation` and `LnOnlyPhi` call it when the suffix is a test, covering both forms: `[] +> name` and `false > [] +> throws-on`. The application form (`eq. +> name`) already forbids a preceding blank via `checkPlain`, so it is untouched.
- `PARSER_SPEC.md` — R-6.5.3 split: masters keep zero-or-one blank, `+>` tests require exactly one; the "Master child" definition in §1 updated accordingly.
- Five stdlib objects that violated the new rule got their blank inserted: `ms/sqrt.eo`, `ms/ln.eo`, `dataized.eo`, `tt/sprintf.eo`, `tt/sscanf.eo`.
- Tests: two new typo packs (formation and inline-phi forms, exact position and message), four unit tests in `LnFormationTest`/`LnOnlyPhiTest`, and ten existing fixtures (`eo-syntax`/`transpile-packs` yaml packs, `TranspileIT`) updated to satisfy the rule.

Verification:
- `mvn -pl eo-parser test`: 914 tests, 0 failures.
- `mvn -pl eo-maven-plugin test`: 343 tests, 0 failures.
- `mvn -pl eo-runtime install` (full transpile of the stdlib): BUILD SUCCESS.
- `mvn verify -Pqulice` clean on `eo-parser`, `eo-maven-plugin`, `eo-integration-tests`.

PR #5455, opened 2026-07-13
#5445: add ++> shorthand for [] +> test attribute declaration

### LINKED ISSUE #5445: Need for `++>` shorthand to replace repetitive `[] +>` test attribute declaration
Every unit test in the codebase opens with `[] +> test-name`, and the empty brackets are pure boilerplate: `[] +>` occurs 5703 times across 242 `.eo` files, and not one `+>` test carries non-empty void params (a grep for `[..] +>` with parameters returns zero matches). See `eo-runtime/src/main/eo/ms/sqrt.eo:18` and its siblings for a typical run. I propose a shorthand `++>` so `[] +> test-name` can be written `++> test-name`.

The empty head carries no information for a test — a test never takes void parameters — so repeating `[]` on every one of 5703 declarations only adds noise. Collapsing it to `++>` is more concise and reads cleaner, in the same spirit as the project's existing sugar (e.g. `!` for a dataized object, #3042). The literal `++>` appears nowhere in the repo today, so the token is free.

The change is front-end only. In `eo-parser/PARSER_SPEC.md` §3.1 the classifier lookahead (around lines 200-201) sends a line whose first non-space char is `+` to a signed-number application (digit follows) or a meta directive (otherwise); add one branch so `+` followed by `+` classifies as a test-attribute line. There is no ambiguity, since metas are legal only before any object is emitted (R-3.2.2) whereas `++>` always sits inside an object body. Treat `++> name` as desugaring to a bare `[]` formation carrying a `+> name` suffix, so the existing `Suffix`/`LnFormation` path in `eo-parser/src/main/java/org/eolang/parser/` and all XMIR emission (§9) stay unchanged.

morphqdd:
Closes #5445.

`++> name` is now accepted as a shorthand for `[] +> name`. The desugaring is front-end only:

- `Eo.classify` — a `+`-headed non-number line dispatches through a new private `plussed(span)`: `++>` classifies as a formation line, anything else stays a meta directive (extracting the method also keeps `classify` under the PMD NPath threshold).
- `LnFormation` — on a `++>` head the params list is empty and the suffix is parsed from `body.substring(1)`, which is exactly `+> name` at its true source column, so every error position and every `@line`/`@pos` in XMIR matches the source with no drift. From there the existing `Suffix`/emission path runs unchanged — the emitted XMIR is byte-for-byte the same as for `[] +> name`. The class-level `@SuppressWarnings("PMD.UnnecessaryLocalRule")` became unnecessary after the restructuring and was removed (qulice flags unused suppressions).
- `PARSER_SPEC.md` — a `+` / `++>` row in the §3.1 classification table and a new rule R-6.3.6 stating the equivalence; there is no ambiguity with metas since metas are legal only before the first object (R-3.2.2).

Tests: three unit tests in `LnFormationTest` (same emission as `[] +>`, same `BARE_FORMATION` level, `++> @` rejected per R-6.3.5), a positive `eo-syntax` pack with two `++>` tests inside an object, and a typo pack checking the exact error position for `++> @`.

Existing `[] +> name` declarations stay legal; nothing in the stdlib is touched.

Verification: `mvn -pl eo-parser verify -Pqulice`: 1460 tests, 0 failures, qulice clean.

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -198,6 +198,7 @@ Each non-blank, non-comment line is classified into exactly one shape, determine
 | First char (after indent) | Lookahead | Line shape |
 |---|---|---|
 | `+` | followed by digit | signed-number application (see §3.6) |
+| `+` | followed by `+>` | test-attribute shorthand — `++> name` desugars to `[] +> name` (§3.4 / R-6.3.6) |
 | `+` | otherwise | meta directive (§3.2) |
 | `#` | — | comment (§3.3) |
 | `.` | — | method-dispatch line (§3.5) |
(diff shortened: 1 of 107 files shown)


yegor256:
@morphqdd replace all `[] +>` in all .eo files too, while you are at this task

morphqdd:
@yegor256 done in 511ffbf: all 1099 line-form `[] +> name` declarations across 100 `.eo` files replaced with `++> name` (plus the doc example in `nop.eo`). The only occurrences left untouched are the inline-phi form (`expr > [] +> name`, e.g. `i64.eo:208`) — there `[]` is the only-phi parameter list, not a bare-formation head, so the `++>` shorthand cannot express it. Verified locally: full `eo-runtime` transpile is clean and the transpiled test suites pass.

morphqdd:
@yegor256 rebased on master and resolved the conflicts with #5451 (the mandatory-blank rule now covers the `++>` form as well). All checks are green except `pr-size`: the full replacement changes 2335 lines against the 200-line limit, which is inherent to converting all 1099 declarations in one PR. Please take a look.

0crat:
@morphqdd Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (2335 >= 100); -8 for way too many hits-of-code (2335 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1138](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5456, opened 2026-07-14
feat(#5443): separated slice from ss.list

Shaurya2k06:
In this PR, I extracted the `slice` object from `ss.list` into its own file `ss/slice.eo`, mirroring the pattern used by other already-extracted siblings (`without`, `back`, `front`, etc.) and matching the `ms.real` decomposition from #4751.

The new free-standing object takes the list as an explicit free attribute (`[lst start end] > slice`) instead of relying on `^`. All eleven slice tests moved with it and were updated from method-call (`slice.`) to free-function (`slice`) syntax. The remaining call site in `list.shifted` was updated the same way, with `+alias ss.slice` added to `list.eo`.

This is one of several extractions toward fully dissolving `ss.list` (#5443). Further PRs can extract `shifted`, `with`/`withi`, `is-empty`, etc. one at a time, then delete `list.eo`.

Verified locally: `mvn clean install -Pqulice -PskipITs` passes, and the affected tests (`EOss.EOlistTest` 24/24, `EOss.EOsliceTest` 11/11) all pass.

Partial fix for #5443


yegor256:
@Shaurya2k06 conflict here

yegor256:
@Shaurya2k06 more conflicts again, sorry

Shaurya2k06:
This feature was already introduced in #5457 .

PR #5458, opened 2026-07-14
Fix YAML linting rule comment and update message

praneethsai050-cmd:
Fixed variable spelling: DELETEITIONS → DELETIONS

Corrected syntax for accessing GitHub context: ${{ github.event.pull_request.additions }} (not ${(...)})

Fixed YAML indentation and placement of run under steps

Added exit 1 only when the limit is exceeded

But remember: This workflow will still fail if your PR exceeds 200 lines. To change that, you'd need to increase MAX_LINES_CHANGED – but that's against the project's policy and likely not allowed.


yegor256:
@praneethsai050-cmd some CI jobs fail here. also, don't suppress any workflows, fix code instead.

.github/workflows/pr-size.yml
```diff
@@ -1,9 +1,10 @@
-# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+---# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
```
morphqdd, line 1: The YAML document marker is glued directly to the comment (`---#`), so `#` no longer starts a comment here. Needs a space or its own line.

.github/workflows/pr-size.yml
```diff
@@ -1,9 +1,10 @@
...
 # SPDX-License-Identifier: MIT
----
-# yamllint disable rule:line-length
+# yamlint disable rule:line-length
+
 name: pr-size
-on: pull_request
+on: pull_request true
```
morphqdd, line 6: `on: pull_request true` isn't valid Actions trigger syntax, the trailing `true` breaks this.

.github/workflows/pr-size.yml
```diff
@@ -17,6 +18,7 @@ jobs:
           total=$((ADDITIONS + DELETIONS))
           echo "Total lines changed: ${total}"
           if [ "${total}" -gt "${MAX_LINES_CHANGED}" ]; then
-            echo "Pull request changes ${total} lines, exceeding the ${MAX_LINES_CHANGED}-line limit; consider breaking it down."
-            exit 1
+            run: echo "Total lines changed: ${{ github.event.pull_request.additions + github.event.pull_request.deletions }}"
```
morphqdd, line 21: `run: echo ...` inside a `run: |` bash block isn't a YAML key here, it's parsed as a shell command called `run:`, which doesn't exist.

.github/workflows/pr-size.yml
```diff
@@ -17,6 +18,7 @@ jobs:
...
+          if [ "${total}" -gt "${MAX_LINES_CHANGED}" ]; then echo "Pull request changes ${total} lines, exceeding the ${MAX_CHANGED} line limit"; exit 1; fi
```
morphqdd, line 22: This duplicates the `if [ "${total}" ... ]; then` check from a few lines up, with mismatched `fi`s, and references `${MAX_CHANGED}` which is never defined anywhere. The env var is `MAX_LINES_CHANGED`.

morphqdd (changes requested):

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (4) comments; -4 for too few (14) hits-of-code; +1 to give you at least something.  Your running score is [+2740](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:
Please forgive us for closing your pull request: it has had merge conflicts for over a month now. Feel free to submit a new one, perhaps similar to this one, but rebased against the current master branch.

PR #5462, opened 2026-07-14
#5460: rewrite i16.as-i32 and i32.as-i64 atoms in pure EO

### LINKED ISSUE #5460: `i16.as-i32` and `i32.as-i64` atoms unnecessarily complicate widening conversions
The `as-i32` object in `eo-runtime/src/main/eo/i16.eo:21` (`[] > as-i32 /i32`) and the `as-i64` object in `eo-runtime/src/main/eo/i32.eo:20` (`[] > as-i64 /i64`) are declared as atoms. Both are widening, sign-extension conversions: they pad 2→4 and 4→8 bytes respectively, prepending `00-00..` when the value is non-negative and `FF-FF..` when negative.

There is no need for them to be atoms. Every primitive required already exists as a `bytes` atom (`concat`, `and`, `eq`), and the narrowing counterparts are already written in pure EO — `i64.as-i32` at `eo-runtime/src/main/eo/i64.eo:20` and `i32.as-i16` at `eo-runtime/src/main/eo/i32.eo:22` both just slice off the high bytes with a bounds check. Widening is even simpler than narrowing because it never throws, so keeping it as an atom while narrowing stays pure EO is an unjustified asymmetry.

Suggested fix for `i16.as-i32`: build the result as `i32 (prefix.concat as-bytes)`, where `prefix` is `00-00` when `(as-bytes.and 80-00).eq 00-00` and `FF-FF` otherwise. Apply the same pattern to `i32.as-i64` with the 4-byte mask `80-00-00-00` and 4-byte prefixes. This matches the existing tests (`42` → `00-2A` → `00-00-00-2A`, `-200` → `FF-38` → `FF-FF-FF-38`). Related sibling concern: #5442.

morphqdd:
Closes #5460.

Both widening conversions are now pure EO, mirroring the suggested pattern: the result is `i32 (prefix.concat as-bytes)` (resp. `i64`), where `prefix` is the zero bytes when `(as-bytes.and 80-00).eq 00-00` (sign bit clear) and the `FF..` bytes otherwise. The Java atoms `EOi16$EOas_i32` and `EOi32$EOas_i64` are deleted, and since those were the last atoms in `i16.eo` and `i32.eo`, the now-pointless `+rt jvm` / `+rt node` metas are dropped too (the `rt-without-atoms` lint enforces exactly that).

Four new byte-level tests pin the widening semantics from the issue: `42` → `00-00-00-2A` / `00-..-2A` and `-200` → `FF-FF-FF-38` / `FF-..-38` for both conversions. The two entries for the deleted atoms are removed from `EOiConvertersExpectTest`.

Verification: full `eo-runtime` transpile clean; `EOi16Test` + `EOi32Test`: 70 tests, 0 failures (all arithmetic in these files routes through the rewritten conversions); `EOi64Test`, `EOnumberTest`, `EObytesTest`, `EOiConvertersExpectTest`: 111 tests, 0 failures; qulice clean. Net diff: −45 lines.


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (109 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1174](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5463, opened 2026-07-14
#5442: rewrite number.as-i64 atom in pure EO via IEEE-754 decode

### LINKED ISSUE #5442: `number.as-i64` Java atom poses consistency issue in float-to-integer conversion
`number.as-i64` (declared at `eo-runtime/src/main/eo/number.eo:30`, implemented in `eo-runtime/src/main/java/org/eolang/EOnumber$EOas_i64.java`) is the only remaining Java atom in the float-to-integer conversion chain. Its neighbours are already pure EO: `number.as-i32` and `number.as-i16` compose from it (`number.eo:16-17`), and `i64.as-i32`, `i32.as-number`, `i16.as-i64` are composed in EO too.

This matters because every atom must be reimplemented and kept in sync across runtimes (JVM `eo-runtime` and `eo2js`). The atom does only `Double::longValue` — truncate toward zero, error outside `[-2^63, 2^63)` — and a double's integer value is fully determined by the IEEE-754 layout already exposed through `as-bytes`, so the primitive is not strictly necessary.

Suggested fix: reimplement `number.as-i64` in `number.eo` in pure EO by decoding sign/exponent/mantissa from `as-bytes`, restoring the implicit leading bit, and shifting by `exponent-1023-52` using the bit operations `bytes` already provides (`left`/`right`/`and`/`or`/`slice`), keeping the existing out-of-range error, then delete the Java class. It is not trivial (a full IEEE-754 decode, ~50 lines) but it removes one cross-runtime atom.

morphqdd:
Closes #5442.

`number.as-i64` is now pure EO — a full IEEE-754 decode over the bit operations `bytes` already provides, removing the last Java atom in the float-to-integer chain:

- `exp` and `mantissa` (with the implicit leading bit restored via `.or 00-10-..`) are extracted from `as-bytes` with masks and a 52-bit shift;
- exponents below 1023 (|n| < 1, zeros, subnormals) truncate to `0`;
- otherwise the 53-bit significand is shifted by `exp − 1075` (`left` or `right`), which is exactly truncation toward zero — matching the old atom's `Double::longValue`;
- negative values are produced by two's complement over `bytes` (`not` + `plus 1`) rather than `i64.neg`, because `neg` is `times -1.as-i64` and would recurse into `number.as-i64` itself;
- `exp == 2047` (NaN/±Inf), `exp > 1086`, and the unrepresentable positive `2^63` are out of range; `−2^63` (exp 1086, bare mantissa, negative) maps to `80-00-..-00`.

The signature follows the established convention of its pure-EO siblings (`i64.as-i32`, `i32.as-i16`): a `cant-convert` void that receives the error message. Existing `x.as-i64` call sites are unaffected (the void stays unfilled and only errors when actually dataized on the out-of-range path), and the existing `throws-on-converting-huge-number-to-i64` / `throws-on-converting-long-max-boundary-to-i64` tests still pass unchanged; callers may now also supply a fallback, tested the same way `i32.as-i16` tests do.

Eight new tests pin the tricky spots: truncation toward zero for both signs (`3.7` → `03`, `-3.7` → `FF..FD`), fractions and `-0.0` to zero, a left-shift case (`1.0e18` → `0D-E0-B6-B3-A7-64-00-00`), the `−2^63` boundary, NaN and `pinf` rejection, and the fallback path.

Verification: full clean `eo-runtime` build and the entire transpiled+Java test suite: 1484 tests, 0 failures, 0 errors; qulice clean.


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (123 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1186](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5472, opened 2026-07-14
Add the recovered atom to intercept a bottom

maxonfjvipon:
Now that a failure is a ⊥ and nothing can catch it, there needs to be a sanctioned way to intercept one and keep going. This adds that primitive.

`recovered` takes an object and an alternative; it normalizes the object and, if that resolves to a ⊥, behaves as the alternative, otherwise as the object itself — returned as-is, not dataized, so a non-dataizable object (a plain formation with no Δ) survives. A ⊥ on both sides stays a ⊥, so an outer recovery can still intercept it.

Detecting the ⊥ without forcing it is the crux, since the object is usually a lazy application or dispatch. The new `Phi.normalized()` resolves an object through dispatch, λ and φ to reveal whether it is a ⊥: a ⊥ — written `T`, produced by an unset void, or returned by a failing atom (including one whose provided fallback itself ends in a ⊥) — comes back as a `PhTerminator` instance, any other object is returned unchanged, and a genuine, unrecoverable failure still throws `ExFailure`. No exception control flow, no dataization.

This is the runtime behind the `:?` substitution operator; the surface syntax is deliberately left for a later change. The eo2js runtime will need a matching atom in its own repo.

### DIFF
--- a/eo-runtime/src/main/eo/recovered.eo
+++ b/eo-runtime/src/main/eo/recovered.eo
@@ -1,0 +1,50 @@
+# Recover from a terminated computation (⊥).
+# Resolves `value` to its normal form; if that is a ⊥, behaves as
+# `alternative`, otherwise as `value`. This is the only way to intercept a
+# ⊥ and keep going: a ⊥ that is never recovered terminates the program.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++rt jvm org.eolang:eo-runtime:0.0.0
++rt node eo2js-runtime:0.0.0
++version 0.0.0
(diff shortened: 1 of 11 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (192 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+643](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5467, opened 2026-07-14
`i64.div` atom implementation lacks precision and performance


yegor256:
The `div` object in `eo-runtime/src/main/eo/i64.eo:59` (`[x] > div /i64`) is a native atom. Unlike `i16.div` and `i32.div`, which delegate upward to `i64.div` (`i32.eo:77`, `i16.eo:68`), `i64` is the widest integer and cannot delegate to a wider type; delegating to `number.div` (`number.eo:117`) is not an option either, since `number` is a floating-point double and would lose precision for magnitudes above 2^53. Hence the atom.

Even so, integer division is computable from the primitives that already exist, so `div` can be pure EO. A binary long-division (shift-and-subtract) over the 64 bits needs only: byte-level bit manipulation (`bytes.left`, `bytes.right`, `and`, `or`), the `i64` comparison and subtraction already available as `gte` and `minus`, and the `while` loop object for the iteration. The quotient sign is the XOR of the operand signs, and truncation-toward-zero (as required by the existing tests, e.g. `13.div -5` = `-2`) falls out of dividing magnitudes and then applying that sign.

This is more involved than the trivial widening atoms in #5460: it requires a 64-iteration loop, an unsigned comparison inside the loop (the raw bit pattern, not the signed `gt`), and careful handling of the `i64` minimum `-9223372036854775808`, whose magnitude is not representable as a positive `i64`. It also trades one native operation for ~64 iterations, so there is a real performance cost to weigh. Suggested approach: add a pure-EO `div` that loops with `while`, shifting the dividend bit by bit into a remainder, subtracting the divisor via `minus` when it fits, and setting the corresponding quotient bit with `bytes.or`; keep `cant-divide` for the zero-divisor case. Related: #5460, #5442.

PR #5475, opened 2026-07-14
#5469: rewrite i64.plus atom in pure EO via carry propagation

### LINKED ISSUE #5469: `i64.plus` atom should be pure EO, implemented using bitwise carry propagation
The `plus` object in `eo-runtime/src/main/eo/i64.eo:53` (`[x] > plus /i64`) is a native atom and sits at the very bottom of integer arithmetic: `minus` and `neg` are defined in terms of it (`i64.eo:55`, `i64.eo:18`), `i16`/`i32` addition delegates up to it, and the pure-EO `times`/`div` proposed in #5468 and #5467 would rely on it too.

Even so, addition is derivable from bitwise operations that already exist on `bytes`, so `plus` can be pure EO. The classic carry-propagation loop computes the sum without a native adder: the bitwise XOR of the operands is the sum ignoring carries, the bitwise AND is the carry bits, and shifting the carry one position left and repeating until it is zero yields the result. It needs only `bytes.xor` and `bytes.and`, a one-bit `bytes.left` shift, `bytes.eq` to test the carry against zero, and the `while` object. The loop terminates in at most 64 iterations because the carry shifts toward the high end and is dropped once it leaves the 8-byte width, which also gives the correct wrap-around modulo 2^64 with no sign handling.

The caveat here is sharper than in #5467 and #5468: `plus` is the most frequently used arithmetic primitive, and everything else (`minus`, `times`, `div`, and the two proposals above) is built on top of it, so replacing one native addition with a ~64-iteration loop compounds multiplicatively through the whole stack. Suggested approach: add a pure-EO `plus` that loops with `while`, keeping `sum` as `a.xor b` and `carry` as `(a.and b).left 1`, until the carry reaches zero. Related: #5467, #5468, #5460, #5442.

morphqdd:
Closes #5469.

`i64.plus` is now the classic carry-propagation loop in pure EO, expressed as a recursive sibling formation (the same shape as `factor` in `ms/sine.eo`):

```
[x] > plus
  carried > @
    as-bytes
    x.as-bytes
  if. > [a b] > carried
    b.eq 00-00-00-00-00-00-00-00
    i64 a
    carried
      a.xor b
      left.
        a.and b
        1
```

`a.xor b` is the sum without carries, `(a.and b).left 1` is the carry; the recursion terminates in at most 64 steps because the carry marches toward the high end and falls off the 8-byte width, which is also exactly the wrap-around modulo 2^64, so no sign handling is needed. The `EOi64$EOplus` atom is deleted and removed from `EOi64ExpectTest`'s parameter list.

Four new tests pin the carry behavior: a carry rippling through all eight bytes (`00-FF-..-FF + 1`), wrap-around at `i64` max (`7F-FF-..-FF + 1` -> `80-00-..-00`), `-1 + 1 = 0` (full-width carry-out), and mixed signs (`-100 + 42 = -58`). Note `minus`, `neg`, and all `i16`/`i32` arithmetic bottom out in this object, so the whole existing integer suite exercises it.

Verified:
- `EOi64Test`: 46/46 green.
- `EOi64ExpectTest`: 2/2 green.

Note for reviewers: this PR touches the same `i64.eo` region as #5467/#5471 (submitted separately, one per issue) - I'll rebase whichever lands later.


yegor256:
@morphqdd merge conflict here

morphqdd:
@yegor256 rebased on master after #5477 landed — conflicts resolved (plus and times now coexist in pure EO), 51 tests green locally.

morphqdd:
@yegor256 the CI failure here is worth attention beyond this PR: it only appears when the pure-EO `times` (#5477, already on `master`) and the pure-EO `plus` (this PR) are **combined**. Each is green in isolation — #5477 passed full CI with `plus` still native, and this branch passes `EOi64Test`/`EOi16Test`/`EOi32Test`/`EOnumberTest` locally (51 tests). But `EOss.EOhash_code_ofTest`, which stacks them (`31.as-i64.times acc`, then `plus.`), fails with:

```
... Error in "Φ.number.eq.xbts.ρ.α0.Δ" at .number:86:4; Error in "Φ.number.div.Δ" at .number:117:2; the 'ρ' attribute must be a number
```

Note the diagnostic dead end: `Expect.that(...).otherwise("must be a number")` swallows the underlying exception, so whatever really failed inside the dataization is replaced by the generic type message. I'm hitting the same pattern while implementing #5467 locally — isolated reproductions of every construct pass, while the composed pure-EO arithmetic fails with these masked errors, which smells like a runtime ρ/caching defect rather than a defect in either EO rewrite.

I'd hold this PR until that's understood — happy to file a separate issue with the evidence if you agree.

yegor256:
@morphqdd sure, file an issue

morphqdd:
@yegor256 filed: #5478 — with the full evidence matrix (what passes in isolation, what fails composed, the masked-exception dead end, and the ⊥-in-test-harness side finding). Repro: cherry-pick this PR's commit onto master and run EOss.EOhash_code_ofTest.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1549](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5476, opened 2026-07-14
fix(#5404): continue positional application into next unset void

### LINKED ISSUE #5404: Positional application incorrectly binds to first declared void attribute, blocking continuation of partial application
Positional application in EO binds an argument to the void attribute at its *declaration index*, not to the next unset one. `PhDefault.put(int, Phi)` at `eo-runtime/src/main/java/org/eolang/PhDefault.java:192` resolves the position straight to a name:

```java
@Override
public void put(final int pos, final Phi object) {
    this.put(this.attr(pos), object);
}
```

and `AtVoid.put` at `eo-runtime/src/main/java/org/eolang/AtVoid.java:77` rejects a second write to a slot that is already bound:

```java
if (!this.object.compareAndSet(null, phi)) {
    throw new ExReadOnly(
        String.format("This void attribute \"%s\" is already set, can't reset", this.name)
    );
}
```

The consequence is that partial application cannot be continued. Given `posix` declared as `[name args]`, the expression `posix "open"` binds `name`, so a later `(posix "open") (* path flags mode)` targets position 0 again — `name` — and fails with `This void attribute "name" is already set, can't reset`. To resume the application you have to name the slot explicitly: `(posix "open") (* path flags mode):args`, which is awkward for something as ordinary as feeding a curried object its next argument.

The proposal is to make positional application bind to the first *unset* void attribute in declaration order, so partial application composes the way it does in a curried language:

```haskell
foo :: Int -> Int -> Int
foo x y = x + y

bar = foo 5
bar 6  -- 11
```

Concretely, `α0` at application time would fill the first still-empty void rather than the fixed position-0 attribute. This only changes cases that error today (a positional write onto an already-bound leading slot is currently an exception), so existing well-formed programs keep their meaning while curried application starts working without the `:name` crutch.

maxonfjvipon:
Positional application resolved the argument position straight to the void attribute at that declaration index, so once a curried object had its first void set, feeding it the next argument hit the same already-bound slot and failed with "This void attribute is already set, can't reset".

`PhDefault.put(int, Phi)` now redirects a positional write to the first still-unset void that follows, in declaration order, when the requested position is a void that has already been set. Writes onto non-void attributes keep failing, so passing too many arguments is still reported. A new `Attribute.vacant()` reports whether an attribute is an unset void.

Closes #5404

Claude-Session: https://claude.ai/code/session_01Czy4BBgMrJbK2FuaFXsvwm

### DIFF
--- a/eo-runtime/src/main/eo/fs/file.eo
+++ b/eo-runtime/src/main/eo/fs/file.eo
@@ -177,10 +177,10 @@
             scope
               stream descriptor
             code.
-              syscall.close (* descriptor):args
+              syscall.close (* descriptor)
             true
         code. > descriptor
-          syscall.open (* path flags syscall.create-mode):args
+          syscall.open (* path flags syscall.create-mode)
         plus. > flags
           plus.
             plus.
@@ -224,7 +224,7 @@
                   seq *
                     chunk
                     input-block chunk
-                (syscall.read (* descriptor size):args).output > chunk!
+                (syscall.read (* descriptor size)).output > chunk!
 
           [buffer] > write
             output-block.write buffer > @
@@ -240,7 +240,7 @@
                       * access
                   seq *
                     code.
-                      syscall.write (* descriptor buffer buffer.size):args
+                      syscall.write (* descriptor buffer buffer.size)
                     output-block
 
   ++> tests-check-if-current-directory-is-directory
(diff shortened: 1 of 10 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (118 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+648](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5477, opened 2026-07-14
#5468: rewrite i64.times atom in pure EO via shift-and-add

### LINKED ISSUE #5468: `i64.times` atom unnecessarily native, possible implementation in pure EO with shift-and-add
The `times` object in `eo-runtime/src/main/eo/i64.eo:51` (`[x] > times /i64`) is a native atom, and like `div` it is the bottom of the chain: `i16.times` and `i32.times` delegate up to it (`i16.eo:32`, `i32.eo:42`), while `number` multiplication is a separate floating-point atom that cannot preserve full 64-bit precision. Hence the atom.

Still, multiplication is derivable from primitives that already exist, so `times` can be pure EO. A shift-and-add loop accumulates the product: for each set bit `i` of one operand, add the other operand shifted left by `i`. It needs only `i64.plus` (already an atom) for accumulation and doubling, byte-level `and` and `left` for bit testing and shifting, and the `while` object for the 64 iterations. Because `i64.plus` wraps modulo 2^64, the result naturally truncates to the low 64 bits, matching the documented wrap-around behavior (e.g. the `i32`/`i16` overflow tests that bottom out here).

This is notably simpler than the division case in #5467: the low 64 bits of a two's-complement product are identical whether the operands are read as signed or unsigned, so there is no sign handling, no unsigned-compare, and no `i64`-minimum edge case — only a 64-iteration loop. The one caveat shared with #5467 is performance: one native multiply becomes ~64 iterations. Suggested approach: add a pure-EO `times` that loops with `while`, testing successive bits of the multiplier with `and` and adding the left-shifted multiplicand into an accumulator via `plus`. Related: #5467, #5460, #5442.

morphqdd:
Closes #5468.

`i64.times` is now a shift-and-add loop in pure EO, as the issue suggests — a recursive sibling formation testing successive bits of the multiplier and accumulating the left-shifted multiplicand through `plus`:

```
[x] > times
  looped 00-00-00-00-00-00-00-00 as-bytes x.as-bytes > @
  [acc a b] > looped
    if. > @
      b.eq 00-00-00-00-00-00-00-00
      i64 acc
      looped
        if.
          (b.and 00-00-00-00-00-00-00-01).eq 00-00-00-00-00-00-00-01
          ((i64 acc).plus (i64 a)).as-bytes
          acc
        a.left 1
        b.right 1
```

`bytes.right` is a logical shift (zero-fill, `BytesRaw.shiftRight`), so the multiplier reaches zero in at most 64 steps for negative operands too, and since `plus` wraps modulo 2^64 the result is the low 64 bits of the product — identical for signed and unsigned readings, no sign handling. The `EOi64$EOtimes` atom is deleted and removed from both `EOi64ExpectTest` parameter lists.

Four new tests: two negatives (`-7 × -6 = 42`), mixed signs, wrap to zero (`2^62 × 4`), and the low bits of a large product (`(2^31-1)^2` → `3F-FF-FF-FF-00-00-00-01`). `neg` (`times -1`) and all `i16`/`i32` multiplication bottom out here, so the existing overflow/wrap tests exercise the loop as well.

Verification: `EOi64Test`, `EOi16Test`, `EOi32Test`, `EOnumberTest`, `EOi64ExpectTest`, `EObytesTest` — 189 tests, 0 failures; full transpile clean; qulice clean.

Note: same `i64.eo` region as #5469/#5467/#5471 (one PR per issue) — I'll rebase whichever lands later.


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1234](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5480, opened 2026-07-14
#5465: allow BYTES literals to carry a method chain on horizontal args

### LINKED ISSUE #5465: `BYTES` literal method chain rejected, parser suggests non-usable form
## What happens

A method chain on a `BYTES` literal in horizontal-argument position is rejected, while the same construct with any other data literal parses fine. This works:

```
[] > foo
  x.plus 1.as-i64 > @
```

but this:

```
[] > foo
  x.plus 00-00-00-00-00-00-00-01.as-i64 > @
```

fails with:

```
error: 'trailing garbage after expression'
```

The chain is simply not consumed: in `Tokens.readArgs()` (`eo-parser/src/main/java/org/eolang/parser/Tokens.java:563-568`) the chain is read only when `Emissions.chainable(bare)` is true, and `Emissions.chainable()` (`eo-parser/src/main/java/org/eolang/parser/Emissions.java:232-239`) lists `IDENTIFIER`, `ROOT`, `GROUP`, `INTEGER`, `FLOAT`, `STRING` — but not `BYTES`. Verified at the `Tokens` level: after parsing `x.plus 1.as-i64` the unconsumed tail is empty, while after `x.plus 00-01.as-i64` it is `.as-i64`, which then trips the suffix parser.

Worse, the parser blocks its own suggested workaround. Parenthesizing the argument:

```
  x.plus (00-00-00-00-00-00-00-01.as-i64) > @
```

fails with:

```
error: 'redundant parentheses around a single token — `(00-00-00-00-00-00-00-01.as-i64)` should be written as `00-00-00-00-00-00-00-01.as-i64`'
```

— actively recommending exactly the form that the first error rejects, so there is no legal horizontal spelling at all. In head position the very same chain is accepted (`00-00-00-00-00-00-00-2A.as-i64` on its own line, e.g. `eo-runtime/src/main/eo/bytes.eo:472`), because the head chain is read unconditionally. Both error messages were hit back-to-back while writing pure-EO code for #5442; the only escape is the vertical `plus.` form.

## What should happen

A `BYTES` literal with a method chain should be one argument, exactly like `1.as-i64` or `"hi".length` — i.e. `Emissions.chainable()` should accept `Value.Kind.BYTES`. Alternatively, if chains on `BYTES` arguments are deliberately unsupported, the redundant-parentheses error must stop recommending a form that is itself rejected.

gemshrine:
Closes #5465.

## What was wrong

`Tokens.readArgs()` (`eo-parser/src/main/java/org/eolang/parser/Tokens.java:565`) reads a `.method` chain on a horizontal argument only when `Emissions.chainable()` returns true. The whitelist covered `IDENTIFIER`, `ROOT`, `GROUP`, `INTEGER`, `FLOAT`, `STRING` — but not `BYTES`. So

```eo
[] > foo
  x.plus 00-00-00-00-00-00-00-01.as-i64 > @
```

failed with `trailing garbage after expression`, while

```eo
  x.plus 1.as-i64 > @
```

parsed fine. In head position the chain is read unconditionally, which is why `00-00-00-00-00-00-00-2A.as-i64` on its own line already works (see `eo-runtime/src/main/eo/bytes.eo:472`).

Worse, the parser blocked its own suggested workaround: `x.plus (00-01.as-i64) > @` was rejected by the redundant-parens lint, which recommended exactly the form the first error had refused. There was no legal horizontal spelling at all — this was hit while writing pure-EO code for #5442.

## What changed

- `Emissions.chainable()` now includes `Value.Kind.BYTES`.
- Two new `TokensTest` cases mirror the existing IDENTIFIER coverage: one asserts the chain is captured on a BYTES arg, the other asserts the reader consumes it fully with no dangling `.method` tail.

After the fix, the redundant-parens suggestion is also actually followable, so both errors surface a valid single spelling.

## Verification

- `mvn -pl eo-parser test -Dtest=TokensTest` — 45/45 green (43 existing + 2 new).
- `mvn -pl eo-parser test -Dtest='*Emissions*,*Tokens*,*Parser*'` — 45/45 green.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -235,7 +235,8 @@ final class Emissions {
             || head.kind() == Value.Kind.GROUP
             || head.kind() == Value.Kind.INTEGER
             || head.kind() == Value.Kind.FLOAT
-            || head.kind() == Value.Kind.STRING;
+            || head.kind() == Value.Kind.STRING
+            || head.kind() == Value.Kind.BYTES;
     }
 
     /**
--- a/eo-parser/src/test/java/org/eolang/parser/TokensTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/TokensTest.java
@@ -135,6 +135,34 @@ final class TokensTest {
     }
 
     @Test
+    void readsChainOnHorizontalBytesArg() {
+        final Tokens tokens = new Tokens(
+            "foo 00-01.as-i64", new Span("foo 00-01.as-i64", 1)
+        );
+        tokens.readName();
+        final List<Value> args = tokens.readArgs();
+        MatcherAssert.assertThat(
+            "a BYTES arg of the form `head.m1` must carry its chain — 1 link here",
+            args.get(0).chain(),
+            Matchers.hasSize(1)
+        );
+    }
+
+    @Test
+    void consumesEntireChainOnBytesArg() {
+        final Tokens tokens = new Tokens(
+            "foo 00-01.as-i64", new Span("foo 00-01.as-i64", 1)
+        );
+        tokens.readName();
+        tokens.readArgs();
+        MatcherAssert.assertThat(
+            "after reading a BYTES arg, its `.method` tail must not be left dangling",
+            tokens.tail(),
+            Matchers.equalTo("")
+        );
+    }
+
+    @Test
     void stopsArgsBeforeSuffixMarker() {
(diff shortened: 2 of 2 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (31) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+159](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5489, opened 2026-07-14
Unintended Exceptions in sprintf/sscanf due to ExFailure Error Message Handling


gemshrine:
## What happens

Several error paths in `sprintf`/`sscanf` are supposed to raise a descriptive `ExFailure` with a specific message — but instead raise an unrelated, uninformative `java.util.MissingFormatArgumentException` or `java.util.UnknownFormatConversionException`, because the intended error message never reaches the user: it gets format-parsed a second time by accident and blows up before construction completes.

## Root cause

`ExFailure` has two constructors (`eo-runtime/src/main/java/org/eolang/ExFailure.java`):

```java
public ExFailure(final String cause, final Object... args) {
    this(String.format(cause, args), (Throwable) null);   // formats internally
}

public ExFailure(final String cause, final Throwable root) {
    super(cause, root);                                    // no formatting
}
```

The idiomatic call site passes a raw format string plus its args directly, e.g. `new ExFailure("Unknown type of data: %s", obj.getClass().getCanonicalName())` (`Data.java:167-170`) — one format pass, inside the constructor.

Some call sites instead pre-format the message themselves and pass the *already-formatted* string as the sole argument:

```java
throw new ExFailure(
    String.format(
        "The number %s doesn't fit into long range for the '%%d' conversion",
        number
    )
);
```

With only one `String` argument and no `Throwable`, this resolves to `ExFailure(String cause, Object... args)` with an **empty** `args` array — so the constructor runs `String.format(cause, args)` on the *already-formatted* string, a second and unintended pass. If that pre-formatted string still contains a literal `%`, the second pass either throws (no args left to satisfy a real conversion) or silently mangles the text.

This bites in two independent ways:

**1. A literal `%%X` escape in the format string, meant to show `%X` to the user, survives the first pass as a live conversion for the second.**

`EOtt/SprintfArgs.java:154-158` (pre-existing `toLong`, used by `sprintf`'s `%d` conversion):
```java
"The number %s doesn't fit into long range for the '%%d' conversion"
```
First pass: `%s`→value, `%%`→literal `%`, `d`→literal → result contains literal `%d`. Second (accidental) pass on that result, with zero args, throws `MissingFormatArgumentException: Format specifier '%d'` instead of the intended message.

**2. A `%s`-substituted *value* that itself contains `%` characters (here, a literal list of format flags) survives as live conversions for the second pass.**

[...]

PR #5490, opened 2026-07-14
#5487: reject non-positive positional index in sprintf

### LINKED ISSUE #5487: `sprintf`'s `%0$s` silently returns the last argument instead of rejecting the invalid index
## What happens

`sprintf`'s positional-index syntax (`%N$s`) is documented and tested as 1-based — `%1$s` refers to the first argument (`formats-numbered-substitution`, `formats-with-ordered-numbered-substitutions` in `eo-runtime/src/main/eo/tt/sprintf.eo`). `%0$s` is a malformed reference (there is no "argument 0"), and I expected it to be rejected the same way `%99999999999999999999$s` is (`throws-on-sprintf-with-oversized-positional-index`).

Instead, `%0$s` silently returns the **last** argument — no error, no warning, just a formatted string built from the wrong value.

## Root cause

`SprintfArgs.formatted()` (`eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java:86-99`):

```java
if (positional != null) {
    final String digits = positional.substring(0, positional.length() - 1);
    try {
        arg = Long.parseLong(digits) - 1L;   // convert 1-based to 0-based
    } catch (final NumberFormatException ex) { ... }
} ...
if (arg >= this.length) {
    throw new ExFailure(...);   // only checks the UPPER bound
}
```

For `%0$s`, `digits = "0"`, so `arg = 0L - 1L = -1L`. The bounds check only tests `arg >= this.length` — it never rejects negative values. `-1 >= length` is false for any non-empty tuple, so the guard passes and `taken.put(0, new Data.ToPhi(-1L))` calls `retriever.at(-1)`.

The `retriever` is `args.at` (`eo-runtime/src/main/java/org/eolang/EOtt/EOsprintf.java:56`), i.e. `tuple.at` (`eo-runtime/src/main/eo/tuple.eo:18-29`), which supports **Python-style negative indexing on purpose** (see `tests-tuple-with-negative-index-gets-last` / `-gets-first` in `tuple.eo:216-226`, where `arr.at -1` returns the last element). That feature is correct and intentional for `tuple` in general — the bug is that `SprintfArgs` forwards an internally-computed `-1` into it without meaning to, because the positional-index guard never checked for the negative case at all.

## Verified reproduction

I confirmed this by running `SprintfArgs` directly against a 2-argument tuple shaped exactly like the existing `returnsCorrectArgumentsForMixedSubstitutions` test fixture (`length=2`, `head="second"`, `tail={length=1, head="first"}` — i.e., logical args `["first", "second"]`):

```java
new SprintfArgs("%0$s", 2L, tuple.take("at")).formatted()
// → ["second"]
```

Expected: an `ExFailure` ("argument index is out of bounds"), matching the existing behavior for `%99999999999999999999$s`. Got: silently the last argument, `"second"`, as if the user had written `%2$s`.

At the EO level:

```eo
sprintf > @
  "Hello, %0$s!"
  * "first" "second"
# ⇒ "Hello, second!"  — no error, wrong-looking success
```

## Suggested fix

Reject non-positive positional indices explicitly, alongside the existing oversized-index check:

```java
if (positional != null) {
    final String digits = positional.substring(0, positional.length() - 1);
    try {
        arg = Long.parseLong(digits) - 1L;
    } catch (final NumberFormatException ex) {
        throw new ExFailure(...);
    }
    if (arg < 0L) {
        throw new ExFailure(
            String.format(
                "The argument index %s must be a positive number (1-based) for the '%%N$' conversion",
                digits
            )
        );
    }
}
```

This closes the gap symmetrically with the existing `arg >= this.length` upper-bound check, and keeps `tuple.at`'s negative-indexing feature untouched for its intended (non-sprintf) callers.

Add a regression test alongside `throws-on-sprintf-with-oversized-positional-index` in `sprintf.eo`, e.g. `throws-on-sprintf-with-zero-positional-index` using `"%0$s"`.

gemshrine:
Closes #5487.

## What was wrong

`sprintf`'s `%N$` positional-index syntax is documented and tested as 1-based (`%1$s` refers to the first argument). `%0$s` is a malformed reference — there is no "argument 0" — but was silently accepted.

`SprintfArgs.formatted()` (`eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java`) converts the 1-based digit string to a 0-based index: `arg = Long.parseLong(digits) - 1L`. For `digits = "0"`, `arg = -1`. The only bounds check present was `arg >= this.length` (upper bound) — nothing rejected negative values, so `-1` sailed through.

That `-1` then reached `tuple.at` (`eo-runtime/src/main/eo/tuple.eo:18-29`), which supports Python-style negative indexing **on purpose** (`arr.at -1` returns the last element — see `tests-tuple-with-negative-index-gets-last`). So `%0$s` silently resolved to the **last** argument instead of raising the same "index out of bounds" error already thrown for `%99999999999999999999$s` (`throws-on-sprintf-with-oversized-positional-index`).

## Verified reproduction

Ran `SprintfArgs` directly against a 2-argument tuple shaped like the existing `returnsCorrectArgumentsForMixedSubstitutions` fixture (`length=2`, logical args `["first", "second"]`):

```java
new SprintfArgs("%0$s", 2L, tuple.take("at")).formatted()
```
Before the fix: returns `["second"]`, no error.
After the fix: throws `ExFailure: The argument index 0 must be a positive number (1-based) for the '%N$' conversion`.

## What changed

- Added a lower-bound check right after the positional-index parse: `arg < 0L` throws `ExFailure`, symmetric with the existing upper-bound check.
- Added `throws-on-sprintf-with-zero-positional-index` next to the existing oversized-index test in `sprintf.eo`.

## A second bug found in passing

While writing the fix, my first attempt used `new ExFailure(String.format(fmt, args))` (pre-formatting, then passing the single resulting string) — this crashed with `UnknownFormatConversionException` instead of throwing my intended message. Root cause: `ExFailure(String, Object...)` formats its `cause` argument internally; passing an *already-formatted* string as the sole argument makes it get formatted **again** with zero args, and since my message ended in a literal `'%N$'` (surviving the first pass's `%%` → `%` unescaping), the second pass tried to parse it as a live conversion.

I fixed my own code by using the idiomatic direct call (`ExFailure(fmt, args)`, one pass, matching `Data.java:167-170` and most of the codebase) — but found the exact same defect already present in `SprintfArgs.toLong` and `SprintfArgs.fmt` (plus `EOsscanf.converted`), silently turning their intended `ExFailure` messages into raw `MissingFormatArgumentException`/`UnknownFormatConversionException` leaks. That's a distinct, pre-existing bug unrelated to this one — filed separately as #5489 rather than folding an unrelated fix into this PR.

## Verification

`SprintfArgsTest` (2 existing tests) run green against the patched `SprintfArgs.java` — confirmed via direct JUnit Platform Launcher invocation, since the local `mvn` transpile is currently broken on an unrelated `tt.regex` XSL error on this machine (reproduces identically on a clean `master` checkout, not something this PR introduces).

### DIFF
--- a/eo-runtime/src/main/eo/tt/sprintf.eo
+++ b/eo-runtime/src/main/eo/tt/sprintf.eo
@@ -97,6 +97,11 @@
       "%99999999999999999999$s"
       * "Jeff"
 
+  ++> throws-on-sprintf-with-zero-positional-index
+    sprintf > @
+      "%0$s"
+      * "first" "second"
+
   ++> formats-numbered-substitution-with-multiple-args
     eq. > @
       "This is the first; This is the first as well; The second; and the 3"
--- a/eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java
@@ -97,6 +97,12 @@ final class SprintfArgs {
                         ex
                     );
                 }
+                if (arg < 0L) {
+                    throw new ExFailure(
+                        "The argument index %s must be a positive number (1-based) for the '%%N$' conversion",
+                        digits
+                    );
+                }
             } else {
                 arg = auto;
                 auto += 1L;


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+207](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5492, opened 2026-07-14
#5485: reject exactly 2^63 in sprintf's %d conversion instead of saturating

### LINKED ISSUE #5485: `SprintfArgs.toLong` incorrectly saturates `2^63` to `Long.MAX_VALUE`, contrary to method documentation
## What happens

`SprintfArgs.toLong` (`eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java:137-154`) documents itself as *"rejecting a value outside long range instead of silently saturating to Long.MAX_VALUE/Long.MIN_VALUE as Double.longValue() does"* — but for the one value `2^63 = 9223372036854775808.0` it does exactly what it promises not to: it passes the guard and then saturates.

Root cause is IEEE-754. `Long.MAX_VALUE = 9223372036854775807L` is *not* exactly representable as a `double`; `(double) Long.MAX_VALUE` rounds up to `9.223372036854776E18`, which is exactly `2^63`. So the check
```java
if (number < Long.MIN_VALUE || number > Long.MAX_VALUE) { throw ... }
```
promotes both sides to `double`, and the comparison `2^63 > 2^63` is `false`. The value passes the guard and hits `return (long) number`, where the double-to-long narrowing conversion saturates any value ≥ `2^63` to `Long.MAX_VALUE`. Result: the caller gets `9223372036854775807` back for an input of `2^63`, off by one and silently.

The negative side is fine: `Long.MIN_VALUE = -2^63` *is* exactly representable, so `number < (double) Long.MIN_VALUE` correctly rejects anything below it and passes `-2^63` through with no loss.

## Failure scenarios

Java repro (also `Double.MIN_NORMAL`-free — pure integer math):

```java
double n = 9223372036854775808.0;             // = 2^63
System.out.println(n > Long.MAX_VALUE);       // false — both promote to 9.223372036854776E18
System.out.println((long) n);                 // 9223372036854775807, i.e. Long.MAX_VALUE — saturation
```

At the EO level, any `number` whose byte representation is the double `2^63` reaches `toLong` untouched and gets rendered as `9223372036854775807` under `%d`:

```eo
sprintf > @
  "value is %d"
  * (2.0.pow 63)
# ⇒ "value is 9223372036854775807"  — user asked to format 2^63, got 2^63 - 1
```

Same shape with a hand-rolled hex-encoded double whose bit pattern is `43-E0-00-00-00-00-00-00` (= `2^63` in IEEE-754): the sprintf renders `Long.MAX_VALUE` instead of raising the "doesn't fit into long range" `ExFailure` that the method's javadoc promises.

Note the vulnerable range is a single value: `Math.nextUp(9.223372036854776E18) = 9223372036854777856.0` is `2^63 + 2^11`, which does compare `> (double) Long.MAX_VALUE` and gets rejected correctly. So the leak is exactly `2^63`, not a wide window — but it is a value users will legitimately reach (`2.0.pow 63`, bit-shift math, byte-level construction), and the failure is silent.

## Suggested fix

Compare against `2^63` directly on the positive side, so the guard uses the exact boundary rather than the rounded-up `(double) Long.MAX_VALUE`:

```java
private static final double POSITIVE_LONG_LIMIT = 0x1.0p63;   // = 2^63, one ulp above Long.MAX_VALUE

    if (number < Long.MIN_VALUE || number >= POSITIVE_LONG_LIMIT) {
        throw new ExFailure(
            String.format(
                number
            )
        );
    }
    return (long) number;
}
```

`Long.MIN_VALUE` stays as-is because it *is* exactly representable, so `<` still gives the correct half-open interval. On the positive side, `>=` catches `2^63` and everything above; every `double` that survives is in `[-2^63, 2^63)`, which is exactly the range of `long`. No saturation possible.

Add one regression test that `%d` on `2.0.pow 63` (or a hand-crafted `43-E0-00-00-00-00-00-00.as-number`) surfaces an error, not `9223372036854775807`.

Happy to open the PR — flagging as an issue first because the docstring on `toLong` claims exactly this guarantee and it's a small enough correctness fix that I want to confirm the suggested boundary before touching it.

gemshrine:
Closes #5485.

**Note:** this branch is stacked on top of #5491 (`fix/issue-5489-exfailure-double-format`), since both touch `SprintfArgs.toLong` and I wanted to avoid a guaranteed conflict. The diff below will show #5491's changes too until that one merges — once it does, this PR will automatically narrow down to just the `2^63` boundary fix.

## What was wrong

`toLong`'s docstring promises to *"reject a value outside long range instead of silently saturating to `Long.MAX_VALUE`/`MIN_VALUE` as `Double.longValue()` does"* — but the guard compared against `(double) Long.MAX_VALUE`, which itself rounds **up** to `2^63` (`Long.MAX_VALUE = 9223372036854775807` has no exact `double` representation at that magnitude — the nearest representable doubles are `2^63 - 2048` and `2^63`, and `9223372036854775807` is only `1` away from `2^63` vs. `2047` away from the other candidate, so it rounds up).

So for the single value `2^63`, `number > Long.MAX_VALUE` evaluated to `false` — both sides promote to the same `double`, `9.223372036854776E18` — the guard passed, and `(long) number` silently saturated to `Long.MAX_VALUE`. Exactly the behavior the docstring says this method avoids.

## What changed

- Compare against `2^63` directly (`0x1.0p63`, the exact hex double literal) with `>=`, instead of against the rounded-up `(double) Long.MAX_VALUE` with `>`.
- `Long.MIN_VALUE` needed no change: `-2^63` **is** exactly representable as a `double`, so the existing `<` comparison on that side was already exact.

## Verified reproduction / fix

```java
double n = 9223372036854775808.0;   // = 2^63
n > Long.MAX_VALUE;                  // false — both promote to the same double
(long) n;                            // 9223372036854775807, i.e. Long.MAX_VALUE — silent saturation
```

After the fix, `toLong(2^63)` throws `ExFailure: The number ... doesn't fit into long range for the '%d' conversion` instead.

## Tests added

- `rejectsExactlyTwoToThe63rdInsteadOfSaturating` — `2^63` is now rejected with the correct message (previously silently returned `Long.MAX_VALUE`).
- `acceptsLargestDoubleThatTrulyFitsInLongRange` — the largest double that truly fits below `2^63` (`2^63 - 2048`, the representable double one ulp below the boundary at this magnitude — note `9223372036854775807.0` as a literal already rounds up to `2^63`, so it can't be used to test the "just fits" side) still converts correctly.

## Verification

`SprintfArgsTest` — 6/6 green (4 existing/from #5491 + 2 new), run via direct JUnit Platform Launcher invocation since the local `mvn` transpile is currently broken on an unrelated `tt.regex` XSL error on this machine.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EOtt/SprintfArgs.java
@@ -38,6 +38,16 @@ final class SprintfArgs {
      */
     private static final char PERCENT = '%';
 
+    /**
+     * One past the largest {@code double} that still narrows to a valid
+     * {@code long} without saturating: {@code 2^63}, exactly one ulp above
+     * {@link Long#MAX_VALUE}. {@code (double) Long.MAX_VALUE} itself rounds
+     * up to this same value, so comparing against it directly (rather than
+     * against the rounded {@code Long.MAX_VALUE}) is what makes the {@code
+     * >=} guard in {@link #toLong(double)} exact.
+     */
+    private static final double LONG_UPPER_LIMIT = 0x1.0p63;
+
     static {
         SprintfArgs.CONVERSION.put('s', Dataized::asString);
         SprintfArgs.CONVERSION.put('d', element -> SprintfArgs.toLong(element.asNumber()));
@@ -144,7 +154,7 @@ final class SprintfArgs {
      * @return The number as {@code long}
      */
     private static long toLong(final double number) {
-        if (number < Long.MIN_VALUE || number > Long.MAX_VALUE) {
+        if (number < Long.MIN_VALUE || number >= SprintfArgs.LONG_UPPER_LIMIT) {
             throw new ExFailure(
                 "The number %s doesn't fit into long range for the '%%d' conversion",
                 number
--- a/eo-runtime/src/test/java/org/eolang/EOtt/SprintfArgsTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EOtt/SprintfArgsTest.java
@@ -77,6 +77,43 @@ final class SprintfArgsTest {
     }
 
     @Test
+    void rejectsExactlyTwoToThe63rdInsteadOfSaturating() {
+        final Phi tuple = Phi.Φ.take("tuple").copy();
+        tuple.put("length", new Data.ToPhi(1));
+        tuple.put("head", new Data.ToPhi(9_223_372_036_854_775_808.0));
+        final ExFailure ex = Assertions.assertThrows(
+            ExFailure.class,
+            () -> new SprintfArgs("%d", 1L, tuple.take("at")).formatted(),
+            "2^63 is one past Long.MAX_VALUE and must be rejected, not silently "
+                + "saturated to Long.MAX_VALUE via narrowing (double) Long.MAX_VALUE "
+                + "rounds up to exactly this value, so the boundary check must "
+                + "compare against 2^63 directly rather than against Long.MAX_VALUE"
+        );
+        MatcherAssert.assertThat(
+            "the ExFailure message must name the out-of-range number",
+            ex.getMessage(),
+            Matchers.containsString("doesn't fit into long range")
+        );
+    }
+
+    @Test
+    void acceptsLargestDoubleThatTrulyFitsInLongRange() {
+        // 2^63 - 2048: the representable double one ulp below 2^63 at this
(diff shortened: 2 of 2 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+239](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5496, opened 2026-07-14
#5495: fix index-of/last-index-of returning distance-from-end and wrong search direction

### LINKED ISSUE #5495: ss.index-of and ss.last-index-of return the wrong value for any non-symmetric list — hidden by coincidental test data
## What happens

`ss.index-of` and `ss.last-index-of` both return the wrong number for any list where the match isn't positioned symmetrically around the middle of the list. Every existing test in both files happens to use data where the bug's output coincides with the correct answer, which is why it's gone unnoticed.

## `ss.index-of` (`eo-runtime/src/main/eo/ss/index-of.eo:12-27`)

```eo
[lst wanted] > index-of
  number > @
    recidx lst

  [tup] > recidx
    if. > @
      tup.length.eq 0
      -1
      if.
        next.eq -1
        if.
          tup.head.eq wanted
          tup.tail.length
          -1
        next
    recidx tup.tail > next!
```

`next` is computed by recursing into `tup.tail` *before* this level looks at its own `tup.head`. If the recursive call already found a match (`next != -1`), this level returns `next` unchanged — it never even checks whether *its own* head also matches. Since the recursion bottoms out at the list's last element and unwinds outward, this means:

1. **It finds the *last* occurrence of `wanted`, not the first** — any level whose tail already matched short-circuits and never looks at its own (earlier) head, so an outer/earlier match can never override an inner/later one.
2. **The number it returns for a match is `tup.tail.length`** — the count of elements *after* the matching position — which is `list.length - position - 1`, i.e. the index counted **from the end**, not the standard 0-based index from the start.

Both bugs happen to cancel out (or coincide) in every existing test:

| Test | List | Wanted | Why it accidentally passes |
|---|---|---|---|
| `tests-returns-index-of` | `[1,2,3]` | `2` | match is the exact middle element of an odd-length list → distance-from-end equals distance-from-start |
| `tests-returns-first-index-of-element` | `[-1,2,-1]` | `-1` | occurs at both ends; the *last* occurrence found is the literal last element (0 elements after it), which numerically equals the *first* occurrence's true index (0) |
| `tests-finds-index-of-string` | `["asdfgh","qwerty",3]` | `"qwerty"` | again the middle element of a 3-element list |

Counter-example that breaks the symmetry:

```eo
index-of (list (* 10 20 30 40 50)) 20
```

The true (first, 0-based) index of `20` is `1`. I traced the recursion by hand and confirmed with an independent Python simulation of the exact algorithm: **it returns `3`** (`= 5 - 1 - 1`, i.e. `list.length - true_index - 1`).

## `ss.last-index-of` (`eo-runtime/src/main/eo/ss/last-index-of.eo:12-22`)

```eo
[lst wanted] > last-index-of
  reclast lst > @

  [tup] > reclast
    if. > @
      tup.length.eq 0
      -1
      if.
        tup.head.eq wanted
        tup.tail.length
        reclast tup.tail
```

This one checks `tup.head` *first*, before recursing, and returns as soon as it finds *any* match scanning from the head forward — i.e. it stops at the **first** occurrence, not the last, and it has the exact same "returns `tup.tail.length` instead of the true index" problem as above.

Same coincidental cancellation in every existing test:

| Test | List | Wanted | Why it accidentally passes |
|---|---|---|---|
| `tests-finds-last-index-of` | `["qwerty",2,3]` | `2` | middle element of a 3-element list |
| `tests-finds-last-index-of-repeated` | `[24,42,24]` | `24` | first occurrence (index 0) and desired last-occurrence index (2) sum to `length-1`, so `tup.tail.length` at the first match (2) equals the *true* last index by coincidence |
| `tests-last-index-of-unicode` | `["Hi","Привет","Hola","Привет","こんにちわ"]` | `"Привет"` | occurs at positions 1 and 3 in a 5-element list — symmetric around the middle (`1+3=length-1`), so the (wrong) first-occurrence-distance-from-end again equals the (correct) last-occurrence-index |

Counter-example:

```eo
last-index-of (list (* 10 20 30 40 20)) 20
```

The true (last, 0-based) index of `20` is `4`. Traced by hand and confirmed by simulation: **it returns `3`** — the *first* occurrence (index 1) reported as `list.length - 1 - 1 = 3`, an unrelated number that happens to look plausible.

## Verification

I simulated both recursive algorithms exactly (same structure, same order of operations) in Python and ran it against every existing `.eo` test case plus the two counter-examples above:

```
index_of([1,2,3], 2) = 1        (existing test, passes)
index_of([-1,2,-1], -1) = 0     (existing test, passes)
index_of([10,20,30,40,50], 20) = 3   <- should be 1

last_index_of(['qwerty',2,3], 2) = 1              (existing test, passes)
last_index_of([24,42,24], 24) = 2                 (existing test, passes)
last_index_of(['Hi','X','Hola','X','Y'], 'X') = 3 (existing test, passes)
last_index_of([10,20,30,40,20], 20) = 3   <- should be 4
```

All existing behavior reproduces exactly, confirming the model, and the two constructed counter-examples land on the predicted wrong answers.

## Suggested fix

Both need a standard forward (for `index-of`) or forward-with-last-write-wins (for `last-index-of`) linear scan carrying an explicit position counter from the head, instead of deriving the answer from `tup.tail.length` at whichever level happens to short-circuit. For example, `index-of` could thread an accumulator:

```eo
[lst wanted] > index-of
  number > @
    recidx lst 0

  [tup pos] > recidx
    if. > @
      tup.length.eq 0
      -1
      if.
        tup.head.eq wanted
        pos
        recidx tup.tail (pos.plus 1)
```

This scans left-to-right, returns the position counter (not a tail-length), and returns as soon as the *first* match is found — matching the docstring ("Returns index of the first occurrence"). `last-index-of` needs the mirror image: keep scanning to the end, remembering the *last* position where a match occurred, rather than stopping at the first.

Please add at least one non-symmetric regression test to each file (e.g. the two counter-examples above) so this class of bug can't hide behind coincidental test data again.

gemshrine:
Closes #5495.

## What was wrong

Both `ss.index-of` and `ss.last-index-of` derived their answer from `tup.tail.length` at whichever recursion level happened to short-circuit first, instead of threading a real position counter from the head. This produced two independent, compounding defects in each:

- **`index-of`** recursed into the tail *before* checking its own head, and propagated the tail's result unconditionally whenever it wasn't `-1` — so an earlier (outer) match could never override a later (inner) one. It silently found the **last** occurrence, not the first, and reported `list.length - position - 1` (distance from the end) instead of the position itself.
- **`last-index-of`** did the mirror-opposite: checked its own head *before* recursing, returning immediately on the first hit scanning head-first — so it found the **first** occurrence instead of the last, with the same "returns tail-length instead of index" defect.

Every existing test in both files happens to use data where these errors either don't matter (the match sits at the exact middle of an odd-length list, so distance-from-start equals distance-from-end) or cancel out (the wrong-direction match happens to land on a position whose distance-from-end numerically equals the intended answer). That's why this went unnoticed.

## Verified reproduction

I simulated both the old and new algorithms in Python (mirroring the exact recursive structure) against every existing `.eo` test and two constructed, non-symmetric counter-examples:

```
index_of([10,20,30,40,50], 20)      old: 3   correct: 1
last_index_of([10,20,30,40,20], 20) old: 3   correct: 4
```

All existing tests pass under *both* the old and new implementations (confirming the model matches the real recursion), while the counter-examples expose the bug precisely as analyzed.

## What changed

Rewrote both to thread an explicit `pos` counter from `0`:

- `index-of` now does a plain forward scan, returning `pos` as soon as `tup.head` matches — matching its docstring ("Returns index of the **first** occurrence").
- `last-index-of` keeps its original recurse-then-prefer-inner shape (a later match must win over an earlier one), but returns `pos` instead of `tup.tail.length`.

Added two non-symmetric regression tests to each file using the counter-examples above, so the direction *and* the numeric value are both pinned down independently of any coincidental list symmetry:

- `index-of.eo`: `tests-finds-first-index-of-non-symmetric-match`, `tests-finds-first-of-repeated-non-symmetric-match`
- `last-index-of.eo`: `tests-finds-last-index-of-non-symmetric-single-match`, `tests-finds-last-of-repeated-non-symmetric-match`

## Verification

- Both files parse correctly via `eo-parser`'s `EoSyntax` (invoked directly, standalone — this machine's full `mvn` transpile is broken on an unrelated `tt.regex` XSL error that reproduces identically on a clean `master` checkout).
- Re-ran the exact recursive structure of the new implementation against all existing + new test cases in a standalone Python simulation — all pass, including the two counter-examples.
- Full EO-level test execution (`mvn test`) will run in CI, since the local transpile pipeline is unavailable in this environment.

### DIFF
--- a/eo-runtime/src/main/eo/ss/index-of.eo
+++ b/eo-runtime/src/main/eo/ss/index-of.eo
@@ -11,20 +11,16 @@
 
 [lst wanted] > index-of
   number > @
-    recidx lst
+    recidx lst 0
 
-  [tup] > recidx
+  [tup pos] > recidx
     if. > @
       tup.length.eq 0
       -1
       if.
-        next.eq -1
-        if.
-          tup.head.eq wanted
-          tup.tail.length
-          -1
-        next
-    recidx tup.tail > next!
+        tup.head.eq wanted
+        pos
+        recidx tup.tail (pos.plus 1)
 
   ++> tests-returns-index-of
     eq. > @
@@ -57,3 +53,19 @@
         list
           * "asdfgh" "qwerty" 3
         "qwerty"
+
+  ++> tests-finds-first-index-of-non-symmetric-match
+    eq. > @
+      1
+      index-of
+        list
+          * 10 20 30 40 50
+        20
+
+  ++> tests-finds-first-of-repeated-non-symmetric-match
+    eq. > @
+      1
+      index-of
+        list
+          * 10 20 30 40 20
+        20
--- a/eo-runtime/src/main/eo/ss/index-of.eo
+++ b/eo-runtime/src/main/eo/ss/index-of.eo
@@ -11,16 +11,16 @@
 
 [lst wanted] > index-of
   number > @
-    recidx lst 0
+    recidx lst
 
-  [tup pos] > recidx
+  [tup] > recidx
     if. > @
       tup.length.eq 0
       -1
       if.
         tup.head.eq wanted
-        pos
-        recidx tup.tail (pos.plus 1)
+        lst.length.minus tup.length
+        recidx tup.tail
 
   ++> tests-returns-index-of
     eq. > @
--- a/eo-runtime/src/main/eo/ss/last-index-of.eo
+++ b/eo-runtime/src/main/eo/ss/last-index-of.eo
@@ -10,16 +10,20 @@
 +spdx SPDX-License-Identifier: MIT
 
 [lst wanted] > last-index-of
-  reclast lst > @
+  reclast lst 0 > @
 
-  [tup] > reclast
+  [tup pos] > reclast
     if. > @
       tup.length.eq 0
       -1
       if.
-        tup.head.eq wanted
-        tup.tail.length
-        reclast tup.tail
+        next.eq -1
+        if.
+          tup.head.eq wanted
+          pos
+          -1
+        next
+    reclast tup.tail (pos.plus 1) > next!
 
   ++> tests-finds-last-index-of
     eq. > @
@@ -59,3 +63,19 @@
         list
           * "Hi" "Привет" "Hola" "Привет" "こんにちわ"
         "Привет"
+
+  ++> tests-finds-last-index-of-non-symmetric-single-match
+    eq. > @
+      1
+      last-index-of
+        list
+          * 10 20 30 40 50
+        20
+
+  ++> tests-finds-last-of-repeated-non-symmetric-match
+    eq. > @
+      4
+      last-index-of
+        list
+          * 10 20 30 40 20
+        20
--- a/eo-runtime/src/main/eo/ss/last-index-of.eo
+++ b/eo-runtime/src/main/eo/ss/last-index-of.eo
@@ -10,9 +10,9 @@
 +spdx SPDX-License-Identifier: MIT
 
 [lst wanted] > last-index-of
-  reclast lst 0 > @
+  reclast lst > @
 
-  [tup pos] > reclast
+  [tup] > reclast
     if. > @
       tup.length.eq 0
       -1
@@ -20,10 +20,10 @@
         next.eq -1
         if.
           tup.head.eq wanted
-          pos
+          lst.length.minus tup.length
           -1
         next
-    reclast tup.tail (pos.plus 1) > next!
+    reclast tup.tail > next!
 
   ++> tests-finds-last-index-of
     eq. > @


gemshrine:
Closing — this was based on a mistaken analysis of EO's tuple construction order. `tuple.with(x)` makes `x` the new `head` and the *old* tuple its `tail` (see `tuple.eo:37-41`), so for `* a1 a2 ... an`, `tup.head` at the outermost level is `an` (the *last* literal element), and walking `.tail` moves toward `a1` (the first). I modeled this the other way around (as a plain left-to-right list) when I first analyzed #5495, which made the original, unmodified `index-of`/`last-index-of` look broken when it wasn't — re-tracing both algorithms by hand against `tuple.at`'s actual semantics (and my own constructed counter-examples) confirms the original code already returns the correct answer in every case. Apologies for the noise — closing without merging.

Issue #5471, opened 2026-07-14
`i64.as-number` could be implemented in EO, replicating `number.as-i64` approach


yegor256:
The `as-number` object in `eo-runtime/src/main/eo/i64.eo:30` (`[] > as-number /number`) is a native atom. The decisive point is that its inverse is already pure EO: `number.as-i64` in `eo-runtime/src/main/eo/number.eo:31-66` disassembles the IEEE-754 double by hand — reading the sign with `and 80-..`, extracting the biased exponent with `and 7F-F0-..` and `right 52`, rebuilding the mantissa with the implicit leading bit, and shifting it into an integer. The forward direction is the mirror image of that same bit surgery, so the primitives are demonstrably sufficient.

Converting an `i64` to a double is: take the sign and the magnitude, find the position of the highest set bit to get the unbiased exponent (a handful of `gt`/shift comparisons, no long loop needed), add the bias of 1023, drop the implicit leading one and place the remaining bits as the 52-bit mantissa, then assemble sign | exponent | mantissa into eight bytes and wrap them as `number <bytes>`. The `number` constructor already takes raw bytes (`number.eo:15`), so no floating-point arithmetic is needed to build the result and there is no recursion back into `as-number`. Values with magnitude up to 2^53 are exact.

This is the most involved of the related conversions (#5467, #5468, #5469, #5470): magnitudes at or above 2^53 must be rounded to nearest, ties to even, by inspecting the guard and sticky bits below bit 52, and a round-up can carry into the exponent — the one genuinely subtle case. Suggested approach: add a pure-EO `as-number` that computes the biased exponent from the magnitude’s bit length, positions and rounds the mantissa with `bytes` shifts and masks, and ORs the three fields into the eight-byte double. Related: #5470, #5469, #5468, #5467, and the now-closed inverse #5442.

PR #5497, opened 2026-07-14
#5478: make bool branch selector voids anonymous so if results do not capture their methods

### LINKED ISSUE #5478: Runtime failure on composing `i64` arithmetic in pure-EO and masking of actual exception by `Expect`
## What happens

Rewriting `i64` atoms in pure EO works in isolation but fails at runtime once the rewrites are **composed** — and the failure is undebuggable because `Expect` replaces the real exception with a generic type message.

Evidence, all reproducible:

1. `master` today: `i64.times` is pure EO (#5477), `i64.plus` is still a native atom — full CI green.
2. Branch of #5475 (pure-EO `plus` via carry propagation, rebased on that `master`): `EOi64Test`, `EOi16Test`, `EOi32Test`, `EOnumberTest`, `EOi64ExpectTest` all pass locally (51 tests), including new carry/wrap-around tests. But `EOss.EOhash_code_ofTest` — whose `ss/hash-code-of.eo` stacks the two rewrites (`31.as-i64.times acc`, then `plus.`) — fails on all three CI platforms with 7 errors of the form:

```
Error in "Φ.number.eq.Δ" at .number:68:2; ... Error in "Φ.number.eq.xbts.ρ.α0.Δ" at .number:86:4;
Error in "Φ.number.div.Δ" at .number:117:2; the 'ρ' attribute must be a number
```

3. The same pattern blocks #5467 (pure-EO `div`, long division) locally: the composed implementation fails inside with `Error in "Φ.i64.gt" ... the 'ρ' attribute must be an i64`, while **every isolated reproduction of every construct it uses passes** — I probed each one as a separate test on the same branch: sibling-formation recursion with number and bytes voids, a sibling helper applied to the enclosing object's void, a helper called from another helper's body with that helper's void as argument, three-level formation nesting, and the exact `gt.`-over-`i64 (v.xor 80-..)` shape at nesting level three. All green individually; only the composition fails.

Two aggravating observations:

- **The real exception is unreachable.** `Expect.that(phi -> new Dataized(phi).take(Long.class)).otherwise("must be an i64")` (`eo-runtime/src/main/java/org/eolang/Expect.java`) converts *any* failure inside that lambda — including a genuine error thrown five levels deep while dataizing the attribute — into `the 'ρ' attribute must be an i64`. Whatever actually breaks in the composed arithmetic is invisible; every diagnostic path dead-ends in this message.
- While probing I also found that a `++>` test which references a sibling attribute of its enclosing object directly (`probe-loop 3 00-..`) gets `⊥` instead of the sibling: the unit-test harness apparently wraps the test without the parent's full attribute set, and since `AtVoid`/`PhDefault` now return a silent `PhTerminator`, the first symptom is the bizarre `the ⊥ object only accepts a cause at position 0, not 1`. Routing the same call through an instance attribute (`(i64 ..).probe-x`) works.

## What should happen

Composing pure-EO `plus` and `times` (and eventually `div`) should work exactly like composing the native atoms — `EOss.EOhash_code_ofTest` should stay green on the #5475 branch. And when a dataization inside `Expect` fails, the original exception should be preserved (chained as a cause, or rethrown when it is not a type mismatch) instead of being replaced by the generic message, so failures like this one can be diagnosed at all.

To reproduce: cherry-pick the #5475 commit onto `master`, `mvn -pl eo-runtime clean test -Dtest=EOss.EOhash_code_ofTest`. Related: #5467, #5468, #5469, #5475, #5477.

morphqdd:
Closes #5478.

Root cause: the result of every `if.` is a copy of the branch selector formation from `true.eo` and `false.eo`, and its voids were the public names `left` and `right`. Attribute lookup finds them before the phi fallthrough, so any dispatch of `.left` or `.right` on an `if.` result hit the selector's own void instead of the method of the chosen branch. For a `bytes`-returning branch, `magnitude.right 1` took the void `right` (an already bound `bytes`) and then applied `1` to it, landing in the already set `data` void of that `bytes`: `This void attribute "data" is already set, can't reset`. Since `left` and `right` are the shift methods of `bytes`, every integer computation that normalizes with `if.` and then shifts was broken (this is what killed the composed pure-EO arithmetic in #5467, #5471 and the `hash-code-of` failure on #5475), with `Expect` masking it as a generic type error.

Fix, using the approach @yegor256 proposed in #5498 and the `>> handle` on voids from #5529 / #5535: the selector voids are now anonymous with file-local handles.

```
bool > false
  []
    ? >> left
    ? >> right
    right > @
```

Their external names are auto-generated (cactus names, unreachable from outside), so `(if. c a b).left` and `.right` can no longer capture them and reach the returned value's method instead. This is opt-in per void, so public accessor voids like `i64`'s `[as-bytes]` are untouched. Two regression tests in `bool.eo` pin the behavior: `.right` and `.left` on an `if.` result now reach the chosen branch's `bytes` shift.

Verification: full `eo-runtime` build with lints and the whole test suite: 1553 tests, 0 failures, 0 errors.


morphqdd:
@yegor256, please look

yegor256:
@maxonfjvipon look at this. It seems that we should get rid of the `bool` object, to fix this problem.

maxonfjvipon:
@yegor256 no, we just need other names instead of `left` and `right`

maxonfjvipon:
@yegor256 that's a known problem and should be caught by type inference

maxonfjvipon (approved):

0crat:
@maxonfjvipon Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (14) hits-of-code; +4 to give you at least something.  Your running score is [+633](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

morphqdd:
@maxonfjvipon @yegor256 this PR already does exactly that: it renames the selector voids from `left`/`right` to `left-branch`/`right-branch` in `true.eo` and `false.eo`, so an `if.` result no longer shadows the `left`/`right` methods of the value it returns. All callers bind them positionally, so nothing else changes, and the full runtime suite is green (1505 tests). The `bool` object itself stays.

On the "should be caught by type inference" point: I filed #5498 for exactly that follow-up (a lint that flags a formation whose void name collides with an attribute of the object it returns), since the same trap exists in other helpers like `while` with `[condition body]`.

morphqdd:
@yegor256 reworked to your suggestion. Now that #5535 landed, the `? >> handle` void form parses, so the selector voids in `true.eo` and `false.eo` are anonymous with file-local handles:

```
bool > false
  []
    ? >> left
    ? >> right
    right > @
```

Their external names are auto-generated, so `(if. c a b).left`/`.right` can no longer capture them. The public accessor voids (like `i64`'s `[as-bytes]`) stay untouched, since it is opt-in per void. Full suite is green: 1553 tests, 0 failures.

morphqdd:
@yegor256, review pls

morphqdd:

yegor256:
@morphqdd merge conflict here, please fix

morphqdd:
@yegor256, please review

morphqdd:
@yegor256, please check

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1521](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5466, opened 2026-07-14
Implement parameter 'coverage.file' for EO object coverage measurement in transpile goal


yegor256:
### Motivation

Today EO measures only *Java bytecode* coverage, via JaCoCo, and only under an opt-in profile (`eo-runtime/pom.xml`, the `jacoco` profile). We have no way to answer a different and more interesting question: **which EO objects were actually dataized while the tests ran?** An `.eo` object can transpile into covered Java and still never be reached by any test. A phi-object-level coverage metric would tell us how much of our EO code (the `.eo` sources, not the generated Java) our test suite really exercises. No such mechanism exists, and I could not find an open or closed issue proposing one.

### Proposal

Add a new configuration parameter `coverage.file` to the `transpile` goal (`MjTranspile`, `eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java`). Its default is `NULL` — nothing changes, no instrumentation. When it is set to a path, the transpiler wraps every located object into a new runtime decorator, `PhCoverage`, which records a "hit" to that file the first time the object is touched at runtime.

The wrapping point already exists. `to-java.xsl` wraps essentially every located object into `new PhSafe(...)` in one place — the `mode="located"` template (`eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl:607-628`), which already carries the source coordinates (`@line`, `@pos`, `@loc`, object name). When the new `coverage` XSL param is non-empty, that template emits `new PhCoverage(new PhSafe(...), ...)` instead. The param travels the same channel `disclaimer` already uses: a `StClasspath` vararg in `Transpiling.java:91` (`String.format("coverage %s", ...)`), declared as `<xsl:param name="coverage"/>` at the top of `to-java.xsl` (next to the existing `disclaimer` param on line 8). One caveat worth noting for whoever implements it: `Transpiling`'s `TRAIN` is a `static final ThreadLocal` built once, so the `to-java.xsl` `StClasspath` will need to be constructed per-run rather than statically to carry a per-build value.

### `PhCoverage` usage

`PhCoverage` is a plain `Phi` decorator, the same shape as the existing `PhLogged` (`eo-runtime/src/main/java/org/eolang/PhLogged.java`) and `PhSafe` (`eo-runtime/src/main/java/org/eolang/PhSafe.java`). Instead of printing to stdout like `PhLogged`, it appends a hit and delegates. Generated Java changes from:

```java
[...]

PR #5499, opened 2026-07-14
#5466: add PhCoverage decorator recording object coverage hits

morphqdd:
Part of #5466. This used to be split into two pull requests to respect the 200-line budget: this one for the `PhCoverage` runtime decorator, #5500 for the transpiler wiring that calls it. Yegor pointed out, correctly, that reviewing the decorator alone leaves it with zero real callers in the diff, so it reads as dead code. The two are merged back into this single pull request. `pr-size` will fail here on purpose, because connecting a new class to a real caller in the same pull request is worth more than staying under the line limit. #5500 is closed with a pointer back here.

What is included, from both former pull requests:

`PhCoverage` is a plain decorator in the shape of `PhSafe` and `PhLogged`. It records a hit of its source location on `delta()` and `take(name)` and delegates everything else; `copy()` re-wraps, so instrumentation survives copying. Recording is enabled by the `eo.coverageFile` system property: the first time an instrumented object is touched, one `loc:line:pos` line is appended to that file, deduplicated through a concurrent set, so each location is written at most once per JVM. Without the property every hit is a silent no-op.

A `coverageFile` parameter on the `transpile` goal, default null. When it is not set, nothing changes anywhere. When it is set, the `mode="located"` template in `to-java.xsl` emits `name = new PhCoverage(name, "loc", line, pos);` right after the existing `PhSafe` line, reusing the same source coordinates, and `Transpiling`'s per-thread compiled-stylesheet cache is keyed on both `trackLocations` and this new flag so the two settings do not collide.

Two bugs found while merging the two pull requests back together, both fixed here:

The property name did not match between the two. The Maven parameter is `eo.coverageFile` (not `eo.coverage.file` as in the issue text, because the project validator rejects dots after the `eo.` prefix), but `PhCoverage` still read the old dotted name. Even after both pull requests merged separately, the feature would have stayed silently disconnected. Renamed to `eo.coverageFile` in the class and its test.

`PhCoverage` cached its target file in a `static final` field resolved once at class load. That was harmless while nothing else touched the class, but now that it wraps every located object, whichever object is touched first in a shared test JVM would freeze a stale answer (usually null) for the rest of the run. It now re-reads the property on every touch instead.

Two `@todo` puzzles are left on `MjTranspile.coverageFile`, the entry point a user actually configures, instead of split across files: consuming the raw hits into an LCOV report, and enforcing a minimum coverage threshold in eo-runtime once that report exists. A third puzzle covers a gap this merge surfaced: the `coverageFile` value at build time is only used as a boolean gate and is never threaded to the process that runs the compiled program, so today the same path must be set twice by hand, once at build time and once at JVM launch.

Verification: `MjTranspileTest` 40 tests green, `PhCoverageTest` green together with other `eo-runtime` tests in the same JVM (checking exactly the shared-JVM staleness bug above), a full local build with `eo.coverageFile` set confirms the generated Java actually wraps located objects and that running an instrumented test with the same system property writes real hits to the file. `eo-maven-plugin` and `eo-runtime` qulice clean.


yegor256:
@morphqdd read this: https://www.yegor256.com/2015/02/12/top-down-design.html

morphqdd:
@yegor256 done. Added a `@todo #5466` puzzle on the `PhCoverage` class explaining that this decorator only writes the raw `loc:line:pos` file and that nothing consumes it yet, so the next step (an LCOV reporter that merges these hits against the full set of instrumented locations) is written into the code, not just the PR body. The full feature is #5466, split across this and #5500, with the reporter and the build-time threshold left as puzzles.

morphqdd:

yegor256:
@morphqdd again, this feature is not connected to any existing classes. If someone will find `PhCoverage` tomorrow in the codebase, he will simply delete it and _no tests_ will break. This feature will look like dead code to him. Read again: https://www.yegor256.com/2015/02/12/top-down-design.html

morphqdd:
@yegor256 you are right, and I want to be precise about why. The code that actually calls `PhCoverage` was in #5500, split out only to stay under the 200-line pull request limit. Looking at this pull request alone, there was no caller, so the objection was correct.

Merged #5500 back into this pull request. `pr-size` will fail here, on purpose: connecting the class to a real caller in the same diff matters more than the line limit. #5500 is closed with a pointer back here.

While merging the two I also found and fixed two real bugs the split had caused. The Maven property introduced in #5500 was `eo.coverageFile` (not `eo.coverage.file` as in the issue, because the project validator rejects dots after the `eo.` prefix), but `PhCoverage` here still read the old dotted name. Even if both pull requests had merged separately, the feature would have stayed silently disconnected. Fixed to match.

Also found that `PhCoverage` cached its target file in a `static final` field resolved once at class load. That was harmless while nothing else touched the class, but now that it wraps every located object, whichever object gets touched first in a shared test JVM would freeze a stale answer for the rest of the run. It now re-reads the property on every touch instead.

Verified locally: `MjTranspileTest` and `PhCoverageTest` green, and an actual end to end run with `eo.coverageFile` set on both the build and the test JVM writes real hits to the file.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (331 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1321](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5500, opened 2026-07-14
#5466: add coverageFile parameter wrapping located objects into PhCoverage

morphqdd:
Part of #5466: the transpiler side, split from #5499 to respect the 200-line budget. This one is independent of #5499: the tests assert the text of the generated Java, they do not compile it, so the two can be reviewed and merged in any order. The LCOV reporter and the eo-runtime threshold enforcement come next.

What is here, following the design in the issue:

- A `coverageFile` parameter on the `transpile` goal, default null. When it is not set, nothing changes anywhere. One deviation from the issue text: the property is `eo.coverageFile`, not `eo.coverage.file`, because the project validator rejects dots after the `eo.` prefix.
- `Transpiling`: the static per-thread `TRAIN` became a per-instance `ThreadLocal` built from the `coverage` value, exactly the caveat called out in the issue; each worker thread still reuses its own compiled stylesheets. The `to-java.xsl` step receives the `coverage` param only when the file is set.
- `to-java.xsl`: a new `coverage` param declared next to `disclaimer`; when it is not empty, the `mode="located"` template emits `name = new PhCoverage(name, "loc", line, pos);` right after the `PhSafe` line, reusing the same source coordinates.

Two tests: the generated Java contains `new PhCoverage(` when `coverageFile` is set, and never mentions it by default.

Verification: `MjTranspileTest` 37 tests green, `eo-maven-plugin` qulice clean.


eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -410,4 +394,50 @@ private Supplier<Path> cached(final String hsh, final String jname) {
...
+     */
+    private static Train<Shift> pipeline(final Path coverage) {
+        final String[] params;
+        if (coverage == null) {
+            params = new String[] {
+                String.format("disclaimer %s", new Disclaimer()),
+            };
+        } else {
+            params = new String[] {
+                String.format("disclaimer %s", new Disclaimer()),
```
yegor256: @morphqdd looks like code duplication to me

yegor256:
@morphqdd imagine, we merge this PR and never see you again. The development team will know what to do with the code? They will understand how to implement `PhCoverage` and what it's for, but just looking into the code base? Read this: https://www.yegor256.com/2015/02/12/top-down-design.html Also, read about PDD: https://www.yegor256.com/2010/03/04/pdd.html

morphqdd:
@yegor256 you are right, thanks. Fixed both points:

1. Top-down / PDD: added two `@todo #5466` puzzles on the `coverageFile` parameter in `MjTranspile` where a reader first meets the feature. The first marks the missing LCOV reporter (this parameter produces a raw hits file that nothing consumes yet), the second marks the eo-runtime threshold enforcement. So the team reading only the code knows what is unfinished and what `PhCoverage` is ultimately for.

2. Duplication at `Transpiling.java`: the `disclaimer` param was built in both branches of the if. Now the common part is added once into a list and the `coverage` part is appended only when set.

yegor256:
@volodya-lombrozo please, review this one

morphqdd:

yegor256:
@morphqdd waiting for review from @volodya-lombrozo

morphqdd:
Closing this. Yegor pointed out on #5499 that reviewing the `PhCoverage` decorator without its caller in the same diff makes it look like dead code. The two pull requests are merged back into #5499, which now includes this transpiler wiring plus a couple of fixes found while merging: the `eo.coverageFile` property name did not match between the two, and `PhCoverage` cached its target file at class load in a way that would go stale once it wraps every located object.

Sorry for the churn, @volodya-lombrozo, no need to review this one anymore.

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2536](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5505, opened 2026-07-15
Make packages behave as objects

### LINKED ISSUE #5486: Inconsistent behavior when calling methods on packages instead of objects
I suggest to change the semantic of packages.

When this happens (the call of `.take("power")` on the `number` object):

```
(number 42).power 3 > x
```

The runtime should return `number.power 42 3`, if `power` is absent in the `number` object, but the `power` object exists in the `number` package.

Primary motivation: this will help us get rid of meaningless package names like `ms`, `tt`, and `io`. Every object will have an ability to extend its features by adding objects to the package named after him. More examples available for us in the future:

```
"hello".contains "h"          instead of         tt.contains "hello" "h"
string.contains "hello" "h"

4.cosine                      instead of         ms.cosine 4
number.cosine 4

(* 1 2).with 5                instead of         ss.with (* 1 2) 5
tuple.with (* 1 2) 5

(* 3 4 5).maximum             instead of         ms.maximum (* 3 4 5)
tuple.maximum (* 3 4 5)
```

Here, we don't polute the `string` object with supplementary objects like `contains` or `regex`. The `string` stays small, while extendable by dozens of objects in the `string.*` package.

yegor256:
This teaches the runtime that a package can be an object. When an attribute is absent on a value, dispatch no longer stops at ⊥ — it looks for an object in the package named after the value's type and binds the receiver as the first argument. So `(number 42).power 3` reads as `number.power 42 3`, and `2.power 4` gives `16`. The explicit form `number.power 2 4` works too, because the bare package name now resolves to an object that hands out its extensions untouched.

The motivation is the one from the issue: get rid of meaningless package names like `ms`, `tt` and `ss`. Every object can now grow features by putting more objects into a folder named after it, so `ms.power` becomes `number.power`, `tt.sprintf` becomes `string.sprintf`, and so on — the base object stays small while its package is open for extension, in this repo or any other. Those three packages are folded into `number`, `string` and `tuple` and removed.

The one structural change worth flagging for review: an object and its package share a name, which Java forbids for a class and a package, so package name segments now carry an `EO_` prefix (`org.eolang.EO_number`) while the object keeps `EO` (`org.eolang.EOnumber`). `PhNest` is the new object-that-is-also-a-package, and `PhDefault.take` performs the fallback. The Node runtime (eo2js) needs the same treatment in its own repo, and a few parser test fixtures still use illustrative `tt.*` names.

Closes #5486


yegor256:
@maxonfjvipon could you review this when you get a chance? CI is green now — happy to walk through anything in the package-dispatch changes if it helps.

maxonfjvipon (approved):

0crat:
@maxonfjvipon Thanks for the review! You've earned +20 points for this: +12 as a basis; +15 for the 1586 hits-of-code that you reviewed; -8 for absolutely no comments posted.  Your running score is [+697](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1586 >= 100); -8 for way too many hits-of-code (1586 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1760](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5509, opened 2026-07-15
#5470: rewrite i64.gt atom in pure EO via sign-split comparison

### LINKED ISSUE #5470: `i64.gt` atom unnecessary as it can be expressed in pure EO with a sign-split comparison
The `gt` object in `eo-runtime/src/main/eo/i64.eo:43` (`[x] > gt /bool`) is a native atom and is the root of ordering: `lt`, `lte`, and `gte` are all defined in terms of it (`i64.eo:32`, `i64.eo:37`, `i64.eo:45`). Removing this atom therefore reduces the comparison surface to a single derived object.

Signed 64-bit `gt` is expressible in pure EO with a sign split, and unlike #5467/#5468/#5469 it needs no loop at all. When the two operands differ in sign, `a.gt b` is simply whether `a` is non-negative (its sign bit is clear). When they share a sign, `a.minus b` cannot overflow — same-sign subtraction always stays within the `i64` range — so `a.gt b` is whether that difference is non-zero and its sign bit is clear. The sign bit is read with `bytes.and 80-00-00-00-00-00-00-00` compared to zero via `bytes.eq`, and the two branches are selected with `if.`; the only arithmetic used is the existing `minus`.

Because there is no iteration, this conversion carries essentially no performance penalty, which makes `gt` the most attractive of these atoms to move into pure EO. The one dependency to note is that `minus` reduces to `plus` (and `neg` to `times`), so if those stay native this remains a single native add per comparison; there is no cycle, since neither `plus` nor `times` depends on `gt`. Suggested approach: add a pure-EO `gt` that returns `a`’s sign-clear flag when the sign bits differ, and otherwise the sign-clear-and-non-zero flag of `a.minus b`. Related: #5467, #5468, #5469, #5460, #5442.

gemshrine:
Closes #5470.

`i64.gt` was a native atom (`EOi64$EOgt.java`) and the root of all i64 ordering (`lt`, `lte`, `gte` derive from it). This moves it into pure EO using the sign-split approach from the issue, so it needs no loop at all.

### Logic
- **Signs differ** — `a.gt b` iff `a`'s sign bit is clear (i.e. `a >= 0`, so `a` positive-or-zero and `b` negative).
- **Signs equal** — `a.minus b` cannot overflow (same-sign subtraction stays in i64 range), so `a.gt b` iff that difference is non-zero and its sign bit is clear.

The sign bit is read with `bytes.and 80-00-00-00-00-00-00-00` compared to zero via `bytes.eq`. Crucially, the differing-sign branch never dataizes the subtraction (lazy `if.`), so the MAX/MIN case that would overflow a direct `a.minus b` is safe.

### Changes
- `eo-runtime/src/main/eo/i64.eo` — pure-EO `gt`; added tests for the MAX/MIN sign-split boundary (built from exact bytes, since `9223372036854775807.as-i64` is unrepresentable as a double literal) and a same-sign large-negatives case.
- Removed `eo-runtime/src/main/java/org/eolang/EOi64$EOgt.java`.
- `EOi64ExpectTest` — dropped `EOi64$EOgt.class` from the atom `@ValueSource` lists.

`i32.gt` delegates to `i64.gt`, so its existing suite exercises this too. Verified locally: `EOi64Test`, `EOi32Test`, `EOi64ExpectTest` all green.


0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+267](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5510, opened 2026-07-15
Add pipe syntax for applying anonymous formations

### LINKED ISSUE #5432: EO lacks support for anonymous formations with application like in phi-calculus

maxonfjvipon:
A `|` line applies arguments to the same-indent object declared just above it — a named formation or another pipe. That lets an anonymous formation be applied without inventing a call-site name, which is the phi-calculus formation-with-application `⟦…⟧(…)` that EO could not express before. There are three forms: horizontal `| a b > name`, vertical `|` with a deeper argument block, and consecutive pipes that build left-associated applications.

The pipe carries no new semantics. It desugars to an ordinary application that refers to the predecessor by name, and the predecessor stays in place, so nothing downstream (scope resolution, base rolling, the runtime) needs to know about it:

```
[x] > foo          |    [x] > foo
  x.plus x > @     |      x.plus x > @
| 5 > foo5         |    foo 5 > foo5
```

Both sides produce the same `<o base='ξ.foo' name='foo5'>`. A new first parse pass, `wrap-applications.xsl`, reads the preceding sibling's `@name` onto the pipe node and drops the `@pipe` marker; everything after it runs unchanged.

A pipe may follow only a formation or another pipe. Once a `.method` has taken an attribute off the formation the object is no longer in hand, so a pipe after a `.method` is rejected; a `.method` after a pipe is fine. The predecessor must be named (explicit or `>>`), since the pipe refers to it by name.

`PARSER_SPEC.md` gets §3.14 plus the classification and cross-line rules. Reverse printing has no `|` form — it renders the canonical application, same as other sugar.

Closes #5432

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -202,6 +202,7 @@ Each non-blank, non-comment line is classified into exactly one shape, determine
 | `+` | otherwise | meta directive (§3.2) |
 | `#` | — | comment (§3.3) |
 | `.` | — | method-dispatch line (§3.5) |
+| `\|` | — | pipe-application line (§3.14) |
 | `[` | — | formation line (§3.4) |
 | `"""` (start of line, no other content) | — | text-block opener (§3.11) |
 | identifier | ends with `.` followed by space/EOL | reversed-dispatch line (§3.8) |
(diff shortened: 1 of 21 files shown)


maxonfjvipon:
@yegor256 this implements the pipe `|` syntax we landed on in #5432 — happy to adjust the surface form or the rules if you want anything tweaked before review.

yegor256:
@maxonfjvipon try to add this test example too:

```
[] > app
    [x y] > add
      x.plus y > @
    | >>
      2 > foo
      3 > bar
```

Here, `foo` and `bar` are the attributes of the `app`. Should work.

yegor256:
@maxonfjvipon also, add this example:

```
[] > app
    [x y] > add
      x.plus y > @
    | >>
      "hello":foo
      42:bar
```

Should work.

maxonfjvipon:
@yegor256 both work as you described and are now covered as tests (`pipe-vertical-named`, `pipe-vertical-bound`). Named vertical args float up as attributes of `app` and the application refers to them; inline `:bindings` become `@as` slots. Pushed in f96de21.

yegor256 (approved):

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+1756](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (452 >= 100); -8 for way too many hits-of-code (452 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+711](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5403, opened 2026-07-11
File operation failures due to fs syscall wrappers ignoring errno from OS


maxonfjvipon:
The fs layer moved from `java.nio` to JNA syscalls in #5386, and each new wrapper hands EO only the raw return code the native call produced. `OpenSyscall.make` at `eo-runtime/src/main/java/org/eolang/EOsm/Posix/OpenSyscall.java:38` is representative: it calls `open`, stores the descriptor (or `-1`), and returns, without ever reading `errno`.

```java
final Phi result = this.posix.take("return").copy();
result.put(
    0,
    new Data.ToPhi(
        CStdLib.INSTANCE.open(
            new Dataized(params[0]).asString(),
            new Dataized(params[1]).asNumber().intValue(),
            new Dataized(params[2]).asNumber().intValue()
        )
    )
);
```

The same pattern holds for `AccessSyscall`, `CreatSyscall`, `MkdirSyscall`, `RenameSyscall`, `RmdirSyscall`, `StatSyscall`, `UnlinkSyscall`, and their Win32 counterparts in `EOsm/Win32`. So when one of these operations fails, EO sees only `-1` and produces a generic message such as `Couldn't open file '%s'` or `Cant delete the file '%s'`. The old `java.nio`-backed code carried the real `IOException` cause instead, distinguishing permission denied from no such file from disk full from a cross-device rename. Since these failures now terminate with a bottom whose only visible text is the panic message, dropping the reason is a real diagnostic-quality regression.

The fix is to capture the OS error immediately after each native call that returned a failure sentinel, before any other native or JNA call clobbers it: `Native.getLastError()` for POSIX `errno` and `GetLastError` on Win32, folded into the fallback reason so the bottom's message names the actual cause.

0crat:
@maxonfjvipon Thanks for closing this issue! You've earned +6 points for this: +8 as a basis; -2 for 5 days of delay. Your running score is [+733](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5515, opened 2026-07-15
fix(#5403): fold OS error reason into failed fs operation messages

maxonfjvipon:
The fs syscall wrappers moved from java.nio to JNA (#5386) but dropped errno on failure, so file operations panicked with generic messages like "Couldn't open file '%s'" or "Cant delete the file '%s'", losing the distinction between permission-denied, no-such-file, disk-full, etc.

Each POSIX and Win32 wrapper (open, creat, mkdir, rename, rmdir, unlink) now captures the OS error immediately after a failing native call, before any other native/JNA call can clobber it, and exposes a human-readable reason through the syscall return's output slot:
  - POSIX: Native.getLastError() + strerror.
  - Win32: the CRT per-thread errno via _errno() + strerror. A small Errno helper per package does the capture-and-format; it stays an empty output on success, exactly as the wrappers produced before.

fs/file.eo (open, touched, deleted, moved) and fs/dir.eo (made) fold the reason into their fallback messages, e.g.
"Couldn't open file '%s': No such file or directory".

Claude-Session: https://claude.ai/code/session_01FJe73q8kncgR4LXm5sDmHG

### DIFF
--- a/eo-runtime/src/main/eo/fs/dir.eo
+++ b/eo-runtime/src/main/eo/fs/dir.eo
@@ -31,13 +31,14 @@
           ^
           T
             sprintf
-              "Cant make the directory '%s'"
-              * ^.path
-    code. > outcome
+              "Cant make the directory '%s': %s"
+              * ^.path created.output
+    called. > created
       if.
         os.is-windows
         win32 "_mkdir" (* ^.path)
         posix "mkdir" (* ^.path posix.mkdir-mode)
+    created.code > outcome
 
   [glob] > walk /tuple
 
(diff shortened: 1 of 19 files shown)


yegor256:
@maxonfjvipon conflicts here

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (385 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+727](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5519, opened 2026-07-15
#5514: make PhNest.put fail fast to protect shared package state

### LINKED ISSUE #5514: `PhNest` shared instance mutates overall state when directly modified
## What happens

`PhPackage` hands out the cached `PhNest` **without copying it**:

```java
// eo-runtime/src/main/java/org/eolang/PhPackage.java:131-137
private static Phi dispatched(final Phi cached) {
    final Phi result;
    if (cached instanceof PhNest) {
        result = cached;          // shared instance, not copied
    } else {
        result = cached.copy();
    }
    return result;
}
```

So every `Φ.take("number")` returns the *same* `PhNest` instance. Meanwhile `PhNest` delegates mutation to a lazily-cached singleton object:

```java
// eo-runtime/src/main/java/org/eolang/PhNest.java
public void put(final int pos, final Phi object) { this.object().put(pos, object); }
public void put(final String name, final Phi object) { this.object().put(name, object); }
```

`object()` memoizes the underlying object via `computeIfAbsent`. Any code that calls `put` on a `PhNest` **without copying it first** therefore mutates state shared across every reference to that package.

## Why it is latent (not yet a live bug)

`copy()` collapses the `PhNest` into a fresh underlying object, and well-formed EO application copies before it writes arguments — so in practice the `put` lands on a copy, and the full test suite passes. The hazard is that `dispatched` deliberately returns a shared, mutable object whose `put` is unguarded. A future caller (or a code path that puts before copying) would silently corrupt the shared number/string/tuple package object, and the failure would be spooky and order-dependent.

## Suggested resolution

Either make `PhNest.put` fail fast (it arguably should never be written to directly — writes belong on a copy), or document at `dispatched` (`PhPackage.java:131`) precisely why the shared instance is safe to hand out unguarded, so the invariant is explicit rather than incidental.

Introduced in #5505.

gemshrine:
Closes #5514.

`PhPackage.dispatched` hands out the cached `PhNest` **without copying it**, so every `Φ.take("number")` (and `string`/`tuple`) returns the *same* instance. `PhNest.put` delegated to a lazily-cached singleton object, so any direct `put` mutated state shared across every reference to that package — a latent corruption hazard.

### Fix
Per @yegor256's direction, `put` now **fails fast**. There is exactly one legitimate direct write: the one-time `ρ` wiring that `PhPackage.take` performs on a freshly-loaded `PhNest` while resolving a package (`PhPackage.java:91`). So:

- `put(String, Phi)` accepts `ρ` (the wiring) and rejects every other name;
- `put(int, Phi)` is always rejected — positional argument binding must land on a copy, never on the shared instance.

A caller that needs to bind arguments must `copy()` first, which collapses the `PhNest` into a writable plain object (e.g. `(number 42)`).

> Note on scope: a blanket rejection of *all* puts breaks package resolution, because `PhPackage.take` sets `ρ` on the shared `PhNest` directly during first load. The `ρ` exception preserves that path; verified by running the full runtime suite.

### Changes
- `PhNest.java` — `put(int)` / `put(String)` fail fast (mirroring the existing `PhPackage.put(int)`), with the invariant documented on the class Javadoc.
- `PhNestTest.java` — new: rejects direct positional and named puts, and asserts a copy stays writable.

Verified locally: full `eo-runtime` suite is green (1521/1522; the single failure is the network-dependent `EOsocketTest.refusesConnectionViaSyscall`, unrelated).

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhNest.java
+++ b/eo-runtime/src/main/java/org/eolang/PhNest.java
@@ -21,6 +21,13 @@ import java.util.concurrent.ConcurrentHashMap;
  * {@code number.power} reads the extension untouched, while {@code (number 42)}
  * copies into a plain number value.</p>
  *
+ * <p>A single instance is shared across every reference to the package, so it
+ * must never be written to directly: {@link #put(int, Phi)} and
+ * {@link #put(String, Phi)} fail fast, the sole exception being the one-time
+ * {@code ρ} wiring done while the package is being resolved. A caller that
+ * needs to bind arguments must {@link #copy()} first, which collapses this
+ * into a writable plain object.</p>
+ *
  * @since 0.62
  */
 @SuppressWarnings("PMD.TooManyMethods")
@@ -69,12 +76,22 @@ final class PhNest implements Phi {
 
     @Override
     public void put(final int pos, final Phi object) {
-        this.object().put(pos, object);
+        throw new ExFailure(
+            "Can't #put(%d, %s) to package object \"%s\" directly, make a copy first",
+            pos, object, this.pkg
+        );
     }
 
     @Override
     public void put(final String name, final Phi object) {
-        this.object().put(name, object);
+        if (Phi.RHO.equals(name)) {
+            this.object().put(name, object);
+        } else {
+            throw new ExFailure(
+                "Can't #put(%s, %s) to package object \"%s\" directly, make a copy first",
+                name, object, this.pkg
+            );
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+283](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5524, opened 2026-07-15
#5512: raise a clear error for implicit dispatch to a nullary extension

### LINKED ISSUE #5512: Implicit method dispatch fails to terminate for nullary package extension
## What happens

The implicit package-extension fallback added in #5505 binds the receiver unconditionally:

```java
// eo-runtime/src/main/java/org/eolang/PhDefault.java:398-399
found = Phi.Φ.take(full.substring(PhPackage.GLOBAL.length() + 1));
found.put(0, this);
```

`put(0, this)` assumes the extension object has a free first positional attribute to receive the receiver. That assumption breaks for a nullary extension — an object under the package that declares no void attributes (a constant).

The method's own Javadoc promises the opposite:

> When there's no such object, the terminated computation stays terminated.

But when the object *does* exist and simply has no arg-0 slot, the fallback doesn't stay terminated — it throws.

## Failure scenario

`pi` is a nullary object living in the `number` package:

```
# eo-runtime/src/main/eo/number/pi.eo
3.14159265358979323846 > pi
```

Evaluating `42.pi`:
1. `take("pi")` on the number value misses → `PhTerminator` → `extension("pi", ⊥)` (`PhDefault.java:230, 392`).
2. `forma()` is `Φ.number`, so `full = "Φ.number.pi"`, whose class `org.eolang.EO_number.EOpi` exists → the guard passes.
3. `found.put(0, this)` → `attr(0)` → `ExFailure("There are no attributes here, can't read the 0-th one")`.

So `42.pi` raises `ExFailure` instead of degrading to `⊥`. The same shape hits any nullary extension (`e`, etc.) reached through the implicit `value.name` form. The explicit form `number.pi` is fine — it goes through `PhNest` and never calls `put(0, ...)`.

It is a niche misuse, but the contract stated in the Javadoc is not honored.

## Suggested resolution

Only bind the receiver when the resolved object actually has a vacant positional attribute; otherwise fall back to the original `⊥` (or raise a message that names the real problem — "`pi` takes no arguments" — rather than the internal "no attributes here").

Introduced in #5505.

gemshrine:
Closes #5512.

The implicit package-extension fallback (`PhDefault.extension`, added in #5505) binds the receiver with `found.put(0, this)`, assuming the extension has a free first positional attribute. That breaks for a **nullary** extension — a constant under a package, such as `number.pi`.

### What was wrong
Evaluating `42.pi`:
1. `take("pi")` misses → `PhTerminator` → `extension("pi", ⊥)`;
2. class `Φ.number.pi` exists → guard passes;
3. `found.put(0, this)` → the object has no free positional slot → `ExFailure`.

So `42.pi` surfaced an internal message instead of explaining the problem, contradicting the method's own Javadoc.

> Note: the actual low-level message is `void attribute "as-bytes" is already set, can't reset` (not `"There are no attributes here"` as the report guessed) — because `pi`'s single positional attribute is *already bound* to its value `3.14159…`. The fix handles both shapes by catching the failing `put` rather than pre-checking arity.

### Fix
Per @yegor256's direction ("raise a message that names the real problem"), the failing `put(0, this)` is caught and re-raised as a clear `ExFailure` that names the extension and states it takes no arguments (original cause chained). The happy path — extensions with a free `α0`, e.g. `power` — is untouched.

### Changes
- `PhDefault.java` — catch the receiver-binding failure and raise a clear message; Javadoc updated.
- `PhDefaultTest.java` — new case asserting `42.pi` fails with "takes no arguments".

Verified locally: full `eo-runtime` suite green (1546/1547; the single failure is the network-dependent `EOsocketTest.refusesConnectionViaSyscall`, unrelated).

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -383,7 +383,10 @@ public class PhDefault implements Phi, Cloneable {
      * after this object's forma. If it exists, it is returned with this object
      * bound as its first argument, so {@code (number 42).power 3} reads as
      * {@code number.power 42 3}. When there's no such object, the terminated
-     * computation stays terminated.</p>
+     * computation stays terminated. When the object exists but has no free
+     * positional attribute to receive the bound object (for example a nullary
+     * constant like {@code number.pi}), a clear error is raised instead of the
+     * low-level "attribute is already set / no attributes here" message.</p>
      *
      * @param name The name of the absent attribute
      * @param bottom The terminated computation to keep when there's no such object
@@ -395,8 +398,19 @@ public class PhDefault implements Phi, Cloneable {
         if (forma.startsWith(String.format("%s.", PhPackage.GLOBAL))) {
             final String full = String.join(".", forma, name);
             if (PhDefault.exists(new JavaPath(full).toString())) {
-                found = Phi.Φ.take(full.substring(PhPackage.GLOBAL.length() + 1));
-                found.put(0, this);
+                final Phi taken = Phi.Φ.take(full.substring(PhPackage.GLOBAL.length() + 1));
+                try {
+                    taken.put(0, this);
+                } catch (final ExFailure ex) {
+                    throw new ExFailure(
+                        String.format(
+                            "Object '%s' takes no arguments, so it can't be applied to '%s' via the implicit '%s' form",
+                            full, forma, name
+                        ),
+                        ex
+                    );
+                }
+                found = taken;
             }
         }
         return found;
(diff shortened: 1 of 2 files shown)


yegor256 (approved):

yegor256:
@rultor merge

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+1780](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (33) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+295](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5527, opened 2026-07-15
#5520: drop redundant same-package +alias metas after auto-homing

### LINKED ISSUE #5520: 'Redundant same-package `+alias` metas cluttering `eo-runtime` .eo files'
Since #5425 (merged, closes #4529) the compiler auto-homes a bare reference to an object in the *same* package into that package (`Φ.<pkg>.<name>`), exactly as a hand-written `+alias` used to. The homing rule lives in `eo-parser/src/main/resources/org/eolang/parser/parse/add-default-package.xsl` (the `objects` parameter and `eo:homed()`), driven by the `Canonical` pipeline. That PR already dropped the same-package aliases in the `number` package as a proof of concept, so `number` and `nk` are now clean.

Every `+alias <pkg>.<name>` whose `<pkg>` equals the file own `+package` is therefore dead weight today. For example `eo-runtime/src/main/eo/string/at.eo:5-9` declares `+package string` yet still carries `+alias string.slice` and `+alias string.sprintf`, both redundant because bare `slice`/`sprintf` now resolve to `Φ.string.*`. Around 119 such metas remain across ~66 files: `string` 31, `tuple` 54, `io` 17, `sm` 9, `fs` 8. Cross-package aliases must stay — e.g. `eo-runtime/src/main/eo/tuple/reduced.eo:6-8` may lose `+alias tuple.list` and `+alias tuple.reducedi` but must keep `+alias string.sprintf`, and the root-package files (`bytes.eo`, `string.eo`, `number.eo`, `i16.eo`, `i32.eo`, `i64.eo`) keep their `+alias string.sprintf` too.

Fix: for each `.eo` file that has a `+package` meta, delete the `+alias` lines whose target package prefix equals that `+package` value, leaving every cross-package alias untouched. A follow-up cleanup enabled by #5425.

morphqdd:
Closes #5520.

Since #5425 the compiler auto-homes a bare reference to an object in the same package, so a hand-written `+alias <pkg>.<name>` whose `<pkg>` equals the file's own `+package` is dead weight. This removes all of them.

For every `.eo` file that has a `+package P` meta, the `+alias P.<name>` lines are deleted, and every cross-package alias is left untouched. For example `string/at.eo` loses `+alias string.slice` and `+alias string.sprintf` (both same-package), while `tuple/reduced.eo` loses `+alias tuple.list` and `+alias tuple.reducedi` but keeps `+alias string.sprintf`. Root-package files like `string.eo` keep their `+alias string.sprintf` because their package is not `string`.

100 aliases removed across 62 files, pure deletion, no other changes:
string 31, tuple 35, io 17, sm 9, fs 8.

Verification: `mvn -pl eo-runtime clean install` reaches the Java compilation stage, which means register, parse, assemble, lint and transpile all pass with the aliases gone, so every bare reference still resolves via auto-homing (the lint stage would flag any unresolved reference).

### DIFF
--- a/eo-runtime/src/main/eo/fs/dir.eo
+++ b/eo-runtime/src/main/eo/fs/dir.eo
@@ -2,7 +2,6 @@
 # Apparently every directory is a file.
 
 +alias string.sprintf
-+alias fs.tmpdir
 +alias sm.os
 +alias sm.posix
 +alias sm.win32
--- a/eo-runtime/src/main/eo/fs/file.eo
+++ b/eo-runtime/src/main/eo/fs/file.eo
@@ -1,14 +1,11 @@
 # Represents a file in the filesystem for reading and writing.
 
 +alias string.contains
-+alias fs.file
 +alias string.joined
 +alias sm.os
-+alias fs.path
 +alias sm.posix
 +alias sm.win32
 +alias string.sprintf
-+alias fs.tmpdir
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
 +package fs
--- a/eo-runtime/src/main/eo/fs/path.eo
+++ b/eo-runtime/src/main/eo/fs/path.eo
@@ -1,8 +1,6 @@
 # Path `path` that is hierarchical and composed of a sequence of
 # directory and file name elements separated by a special separator or delimiter.
 
-+alias fs.dir
-+alias fs.file
 +alias sm.os
 +alias tuple.reduced
 +alias string.regex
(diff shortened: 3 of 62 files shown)


morphqdd:
@yegor256, please take a look when you have a moment. All 25 checks are green.

yegor256:
@morphqdd conflicts here, please resolve them

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (100 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1250](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5522, opened 2026-07-15
feat(#5433): self-reference `%` for anonymous formations

yegor256:
Closes #5433.

## What

Adds the `%` token — **syntactic sugar for the auto-generated name of the anonymous formation (introduced by `>>`) that surrounds it**. It is not a new reference mechanism: at compile time `%` is simply replaced by that name, which is otherwise untypeable (the cactus is reserved). This lets an anonymous helper call itself recursively while staying inlined and invisible to the enclosing scope — the encapsulation the issue asks for.

```
io.stdout > @
  tt.sprintf *1
    "The 5th Fibonacci number is %d\n"
    [n] >>                       # anonymous (auto-named) formation
      if. > @
        n.lt 2
        n
        plus.
          % (n.minus 1)          # % = the anonymous formation's auto-name
          % (n.minus 2)          # % = the anonymous formation's auto-name
    | 5                          # applies the formation to 5 (the | pipe, #5432)
```

A `%` used **outside** any anonymous formation is a compile-time error — there is no name to substitute.

## How

- **Parser.** `%` is recognised as a value (`Value.Kind.SELF`) wherever a value is expected — head, argument, inside a paren group — with the ordinary application shape (it may carry args and a `.method` chain). It is emitted as a base-less `<o self=''>` marker (`Emit.self()`), mirroring how the `>>` pipe emits `<o pipe=''>`.
- **Substitution.** A new first-pass `resolve-self.xsl` (wired into `Canonical` right after `wrap-applications`) replaces the marker with the name: it sets `@base` to the `@name` of the nearest enclosing anonymous formation (the closest cactus-prefixed ancestor, §9.2) and drops `@self`. From there it is an ordinary reference by the auto-name — no new semantics, no changes below the parser front-end.
- **Error.** A `%` with no enclosing anonymous formation is reported by the reshape as a `resolve-self` check error (a post-parse check, like the compact-tuple index check).

This deliberately follows the pattern of #5432 (`|` pipe): a base-less marker at parse time, resolved by a small first-pass XSL. The canonical use combines both — recursion via `%`, application via `| N`.

## Tests

- `snippets/self-fibo.yaml` — end-to-end runtime (`SnippetIT`): an anonymous formation recursing via `%`, applied via `| 6`, prints `8 is the 6th number`.
- `eo-packs/parse/resolve-self.yaml` — the reshape in isolation: `@self` → `@base='a🌵…'`, args preserved.
[...]

Issue #5433, opened 2026-07-13
anonymous formation can't make a copy of self, but should be possible to


yegor256:
I want to be able to do this (after #5432):

```
io.stdout > [] > app
  tt.sprintf *1
    "The 5th Fibonacci numbers is %d\n"
    [n] >>
      n:5
      if. > @
        n.lt 2
        n
        plus.
          % (n.minus 1)
          % (n.minus 2)
```

Here, the `%` means `a🌵xxx` introduced by the `>>` above.

Primary motivation: such an inlining is the encapsulation in EO. We don't want to make `fibo` object visible in the `app` -- so we inline it as anonymous formation, recursively calling itself.

yegor256:
@maxonfjvipon thoughts?

maxonfjvipon:
@yegor256 looks good, not sure about char `%`, but idea is great

yegor256:
@maxonfjvipon the `%` will be scoped inside a specific formation:

```
seq * > [] > app
  io.stdout
    tt.sprintf *1
      "The 5th Fibonacci numbers is %d\n"
      [n] >>    # if this is a🌵FIBO
        n:5
        if. > @
          n.lt 2
          n
          plus.
            % (n.minus 1)   # % becomes a🌵FIBO
            % (n.minus 2)   # % becomes a🌵FIBO
  io.stdout
    tt.sprintf *1
      "The 6th Fibonacci numbers is %d\n"
      % 6  #  compile time error!
```

The `%` is only accessible inside the formation, allowing recursive calls to it, nothing else.

yegor256:
this feature will be reverted, we'll do this instead: https://github.com/objectionary/eo/issues/5525

0crat:
@yegor256 Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+1804](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5528, opened 2026-07-15
feat(#5525): file-local name handles for anonymous `>>` objects via `>> foo`

### LINKED ISSUE #5525: Feature: Anonymous objects handling using file-local name via '>> foo' not working as expected
### Idea

Extend the auto-name suffix `>>` (`PARSER_SPEC.md` §3.10, §9.2 — the reserved
cactus name like `a🌵12-5`, which nobody can type because 🌵 is excluded from
`NAME`) to optionally carry a name:

```
[] > app
  [n] >> fibo
    ...
```

`>> fibo` keeps the object **anonymous** in the emitted XMIR (it still gets its
deterministic cactus `@name` per §9.2), but `fibo` becomes a **file-local
handle** usable anywhere in the same `.eo` file. The compiler builds a per-file
table `handle → cactus-name` and, during a parse pass, rewrites every reference
to `fibo` into that cactus name, then drops the marker. Since the cactus name is
reserved, the result is identical to hand-writing a name a human could never
type: the object never enters the visible namespace, yet stays reachable by a
friendly handle within the file. Pure syntactic sugar — no new semantics.

`>>` should accept the trailing name uniformly on any line that already allows
bare `>>` (bare formation `[...]`, inline-phi formation, application). `!` const
stays allowed; `/sig` stays forbidden (R-3.10.2).

### Fibonacci example

The canonical use — an anonymous helper that recurses by its own handle and is
also called from a sibling `@`:

```
+alias io.stdout
+alias string.sprintf
+package snippets

[] > app
  [n] >> fibo
    if. > @
      n.lt 2
      n
      plus.
        fibo (n.minus 1)
        fibo (n.minus 2)
  stdout > @
    sprintf
      "%d is the %dth number\n"
      *
        fibo 6
        6
```

`fibo` is referenced twice — recursively inside its own body and from `app`'s
`@` — and both rewrite to the same cactus `@name`. `app` exposes no visible
`fibo` attribute, in contrast to plain `> fibo`, which would put `fibo` on app's
public surface.

### Why this is better than `%` (#5433, reverted in #5522)

The `%` sugar (issue #5433, merged then reverted in PR #5522) only
self-referenced the **nearest enclosing** anonymous formation. `>> foo` gives a
typeable handle to **any** anonymous object, referenceable from anywhere the
cactus name is in scope — so it covers self-recursion, mutual recursion, and
calling an anonymous helper from a sibling, whereas `%` covered only immediate
self-reference. `>> foo` fully subsumes `%`. It keeps the same clean shape as
#5432 (`|` pipe) and #5433: a marker at parse time resolved by one small
first-pass XSL, with no change below the parser front-end.

### Implementation sketch

Mirror the reverted PR #5522:

- **Parser front-end** — extend name-suffix handling so `>>` accepts an optional
  trailing `NAME`. Same files #5522 touched:
  `eo-parser/src/main/java/org/eolang/parser/Tokens.java` (read the name after
  `>>`), `Value.java` / `Eo.java`, and `Emit.java` / `Emissions.java` (emit the
  object with its computed cactus `@name` **and** a marker attribute recording
  the handle, e.g. `local='fibo'`; ordinary references stay as plain
  `<o base='fibo'>`).
- **New first-pass XSL**, e.g. `resolve-local-names.xsl`, registered in
  `eo-parser/src/main/java/org/eolang/parser/Canonical.java:92` at the top of the
  parse train (right after `wrap-applications.xsl`, **before** `build-fqns.xsl` /
  scope resolution) — analogous to the reverted `resolve-self.xsl`. It (1)
  collects `@local → @name` for the whole document, (2) rewrites every `@base`
  equal to a handle into the matching cactus `@name`, (3) drops the `@local`
  markers. Downstream scope resolution then treats the cactus names as ordinary
  local references.
- **Naming caveat** — do *not* call it an "alias": `expand-aliases.xsl` /
  `resolve-aliases.xsl` already handle `+alias` **import** aliases, an unrelated
  concept. Prefer `resolve-local-names.xsl` (or "handles").
- **Spec + tests** — document under §3.10 and §9.2; add parse-pack tests
  mirroring the reverted `resolve-self.yaml` / `resolve-self-outside.yaml`, plus
  an integration snippet like the reverted `self-fibo.yaml`.

### Open questions

- Duplicate handle in one file → compile-time error (`duplicate local name 'foo'`).
- Collision between a `>> foo` handle and a visible `> foo` name in the same file
  → likely a compile-time error; needs a rule.
- A `foo` reference with no matching `>> foo` handle is left untouched (it may be
  a real name, else caught later by scope resolution) — the pass rewrites only
  names present in the table.
- Handle scope is the whole file (flat table), so a reference must still land
  where the cactus name is actually in scope; the sugar does name substitution
  only, normal scoping applies afterward.

Prior art: #5432 (`|` pipe, same marker + first-pass-XSL pattern), #5433 and
#5522 (the reverted `%` — this supersedes that approach).

yegor256:
Closes #5525.

## What

Extends the `>>` auto-name suffix to optionally carry a trailing `NAME` — a **file-local handle**:

```
[] > app
  [n] >> fibo
    if. > @
      n.lt 2
      n
      plus.
        fibo (n.minus 1)
        fibo (n.minus 2)
  stdout > @
    sprintf "%d is the %dth number\n" *(fibo 6) 6
```

`>> fibo` keeps the object **anonymous** — it still receives its deterministic cactus `@name` (§9.2) and never enters the visible namespace — but `fibo` becomes a **typeable alias** for that cactus name, usable anywhere in the same `.eo` file. Here `fibo` is referenced twice recursively inside its own body and once from `app`'s `@`, and all three rewrite to the same cactus `@name`. Unlike plain `> fibo`, the object is never put on `app`'s public surface. Pure syntactic sugar — no new semantics.

This supersedes the reverted `%` (#5433): `%` self-referenced only the *nearest enclosing* anonymous formation, whereas `>> foo` gives a typeable handle to *any* anonymous object reachable from anywhere the cactus name is in scope (self-recursion, mutual recursion, sibling calls).

## How

Mirrors the marker + first-pass-XSL pattern of #5432 / #5522.

- **Parser front-end.** `Suffix.auto()` reads an optional handle after `>>` (and the `!` const), surfaced via the new `Suffix.handle()`. `Emit.local()` records it as an `@local` marker on the object that carries the cactus `@name`; it is emitted by `LnFormation`, `LnOnlyPhi` and `ChainEmission`, so `>> name` is accepted uniformly on bare formations, inline-phi formations and applications. `!` const stays allowed (`>>! name`); `/sig` stays forbidden (R-3.10.2).
- **New first-pass `resolve-local-names.xsl`**, wired into `Canonical` right after `resolve-self.xsl`, before `build-fqns.xsl` / scope resolution. It (1) collects the per-file `@local → @name` table, (2) rewrites every `@base` equal to a handle into the matching cactus `@name`, (3) drops the `@local` markers, and (4) reports a duplicate handle as a `resolve-local-names` compile-time error (`duplicate local name 'name'`). A reference with no matching handle is left untouched for later scope resolution. Downstream passes see only ordinary references by the reserved cactus name — nothing changes below the parser front-end.
- **Spec.** `PARSER_SPEC.md` gains R-3.10.12 (file-local handles) and R-9.2.3 (emission + rewrite), plus the `>>` base-form table entry.

## Tests

- `eo-packs/parse/resolve-local-names.yaml` — the reshape in isolation: three `fibo` references rewrite to the single cactus `@name`, `@local` markers dropped, no errors.
- `eo-packs/parse/resolve-local-names-duplicate.yaml` — two `>> dup` in one file produce the `resolve-local-names` duplicate-handle error.
- `SuffixTest` — `>> fibo` yields `Form.AUTO` carrying the handle; `>>! fibo` combines const with the handle; trailing garbage after the handle is still rejected (`trailing-non-suffix-auto` typo pack updated).

All `eo-parser` tests pass (959, incl. the parse packs); qulice, xcop and yamllint are clean. I also verified the full parse train on the Fibonacci example end-to-end: all `fibo` references resolve to the single cactus `@name` with zero errors — identical to the resolved XMIR of the existing `snippets/fibo.yaml` (which uses `> f`), so runtime behaviour is unchanged.

The diff is kept under the 200-line pr-size limit (197). To stay within it, the runtime `SnippetIT` snippet was deferred to a follow-up — its program is the existing `snippets/fibo.yaml` with `> f` → `>> fibo`, and the resolution is already covered by the isolation packs and the full-train check above. Happy to add it (or split the duplicate-handle safeguard out) if you'd prefer a different balance.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TB27TwPXPJ3c7bKZbxerxj)_

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -482,7 +482,7 @@ There are **four base forms** of name suffix (mutually exclusive on any one line
 | Form | Meaning |
 | --- | --- |
 | `> name` | Explicit name. Optional trailing `!` for const. Optional ` /sig` to declare an atom. |
-| `>>` | Auto-generated name (deterministic, derived from line and column). Optional `!`. Atom signature forbidden. |
+| `>>` | Auto-generated name (deterministic, derived from line and column). Optional `!`. Optional trailing `NAME` — a file-local handle (R-3.10.12). Atom signature forbidden. |
 | `+> name` | Test attribute. `name` must be a `NAME` token, not `PHI` (`@`) — see R-6.3.5. Legal only at indent level 1 of a top-level object (§6.3). |
 | (none) | Allowed unless the object is a plain child of a formation (§6.2). |
 
@@ -526,6 +526,8 @@ R-3.10.10a. **Anonymous inline-phi as a paren-grouped value.** Inline-phi format
 
 R-3.10.10. `sig` declares the atom's return type. Examples: `/number`, `/bytes`, `/i32`. Dotted form (`/Q.org.eolang.number`) is permitted but rare. A bare `/Q` (root alone, no dot-name) is rejected — the grammar accepts it but the new parser does not. **Other malformed sigs** — bare `/` (empty signature), trailing dot `/Q.`, or sigs starting with `.` — are already rejected by the underlying grammar (`atomBase : (ROOT | NAME) (DOT NAME)*`) and need no separate spec rule. The new parser narrows the grammar in exactly one spot: forbidding `/Q` alone.
 R-3.10.11. The leading `Q` in a dotted `sig` is promoted to `Φ` in XMIR (the source→XMIR mapping table in §9.3 is the single source of truth for all Q→Φ / @→φ / ^→ρ promotions).
+
+R-3.10.12. **File-local handles — `>> name`.** A `>>` auto-name suffix may carry an optional trailing `NAME`: a *file-local handle*. The object stays **anonymous** — it still receives its cactus `@name` (§9.2) and never enters the visible namespace — but `name` becomes a typeable alias for that cactus name, usable anywhere in the same `.eo` file (`resolve-local-names`, §9.2, rewrites references to the cactus name). So an anonymous helper can recurse by its handle or be reached from a sibling — unlike plain `> name`, which would expose `name` on the enclosing object's public surface. Accepted uniformly wherever bare `>>` is (bare formation, inline-phi formation, application); `!` const stays allowed (`>>! name`), `/sig` stays forbidden (R-3.10.2). A handle declared twice in one file is a compile-time error (`duplicate local name 'name'`); a reference with no matching handle is left untouched for later scope resolution. See §9.2 for the emission and the `handle → cactus-name` rewrite.
 
 ### 3.11 Triple-quoted text block — `"""`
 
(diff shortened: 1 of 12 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (197 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1808](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5501, opened 2026-07-15
Inconsistent naming convention in the eo-runtime object layout


yegor256:
Currently, object names in the eo-runtime are pretty messy. We don't have a solid uniform naming concept. For example:

* The `as-hex` turns string into hexadecimal presentation, while `trimmed` removes spaces. Why it's not `as-trimmed`? 
* The `bytes-as-array` turns bytes into a tuple of single-byte items. Why it's not `bytes-as-tuple`? Also, why `tt.as-number` is not `tt.string-as-number`?
* The `ss.hash-code-of` turns `input` into `bytes`, while `io.input-length` turns `input` into `number`. Why it's not `io.length-of`?

How about we introduce this naming principle (and the classification of _all_ objects), after #5486 is implemented:

* **runtime** objects

  * `dataized`
  * `recovered`

* **data** objects
They represent bytes.
 
  * `bytes` (with only `.eq`, `.size`, `.slice`, `.and`, `.or`, `.not`, `.right`, and `.concat` - all atoms)
  * `number` (with only `.gt`, `.times`, `.plus`, and `.div` - all atoms)
  * `string` (with only `.length` and `.slice`)
  * `bool`
  * `char` (didn't exist)
  * `i64`, `i32`,`i16`, `i8`
  * `u64`, `u32`,`u16`, `u8` (didn't exist)
  * `nan`
  * `ninf`
  * `pinf`
  * `pi`
  * `e`
  * `eol`

* **platform** objects
They build a bridge between EO runtime and the underlying platform (Unix, iOS, JVM, etc.)

  * `os`
  * `malloc`
  * `posix`
  * `win32`

* **flow control** objects

  * `seq`
  * `while`
  * `switch`
  * `true`
  * `false`
  * `nop`

* **structure** objects
 
  * `tuple`
  * `map`
  * `set`
  * `range`
  * `input`
  * `output`
  * `buffer` (formerly `string-buffer`)

* **resource** objects

  * `console`
  * `clock` (formerly `system-clock`)
  * `random`
  * `file`
  * `getenv`
  * `directory` (formerly `dir`)
  * `socket`
  * `mktemp` (formerly `tmpdir`)
  * `path`
  * `uri` (didn't exist)
  * `stdin`
  * `stdout`
  * `stderr` (didn't exist)

* **supplementary** objects

  * `bytes.hex`: bytes as hexadecimal text -> `string`
  * `bytes.as-input`: bytes as readable `input` -> `input`
[...]

PR #5530, opened 2026-07-15
Introduce bytes package for supplementary objects

yegor256:
This carves the non-atom objects out of the core `bytes` object and into a new `bytes` package, following the layout proposed in #5501. The `bytes` object is left with only its eight atoms (`eq`, `size`, `slice`, `and`, `or`, `not`, `right`, `concat`), while `left`, `xor`, `as-bool`, `as-number`, `as-i16`, `as-i32`, and `as-i64` become standalone objects under `eo-runtime/src/main/eo/bytes/`, reached through the same package dispatch that already serves `number`, `string`, and `tuple`. The void data attribute is renamed to `@`, so a bytes literal is now `[@] > bytes` with no separate `data` alias.

The issue suggested dropping `as-bytes` entirely. It turns out to be load-bearing: the `eq` atom reads `b.as-bytes`, the const (`!`) operator expands to `(dataized ...).as-bytes`, and `bool` (and therefore `true`/`false`) reaches it by decoration. So instead of deleting it I moved it into the package as an identity object, which keeps all three paths working through dispatch while still emptying it out of the core object. Removing it for real would mean rewiring the `eq` atom and the const transform, which feels like its own issue. Renaming `data` to `@` also needed a one-line fix in the φ-term renderer, which used to look up the bytes value by the hard-coded `data` name and now follows the decoratee.

All 1546 eo-runtime tests pass, including the moved tests now living beside their objects in the package.

Refs #5501

### DIFF
--- a/eo-runtime/src/main/eo/bytes.eo
+++ b/eo-runtime/src/main/eo/bytes.eo
@@ -1,16 +1,17 @@
-# Object `bytes` encapsulates a chain of bytes, providing bitwise operations,
-# numeric conversions, and byte manipulation methods. Objects like `number`, `string`,
-# and integer types (`i16`, `i32`, `i64`) use `bytes` as their internal representation.
+# Object `bytes` encapsulates a chain of bytes, providing the bitwise and
+# structural atoms every data object relies on. Objects like `number`, `string`,
+# and the integer types (`i16`, `i32`, `i64`) use `bytes` as their internal
+# representation, while conversions and derived operations live in the
+# `bytes` package (e.g. `bytes.as-number`, `bytes.xor`).
 #
 # Usage examples:
 # ```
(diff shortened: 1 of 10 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (492 >= 100); -8 for way too many hits-of-code (492 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1824](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5531, opened 2026-07-15
#5511: cache the classpath probe on the package-extension dispatch path

### LINKED ISSUE #5511: Uncached `Class.forName` in package-as-object dispatch slows down every ⊥ resolution
## What happens

After #5505 ("packages are objects", closes #5486), attribute dispatch falls back to a package-extension lookup whenever a `take` would otherwise resolve to `⊥`. That fallback probes the classpath with `Class.forName`, and it does so **uncached**, on a hot path.

Two spots:

1. `PhDefault.take` calls `extension(...)` for **every** attribute that resolves to a `PhTerminator` (`eo-runtime/src/main/java/org/eolang/PhDefault.java:230`). `extension` then calls `exists(...)` → `Class.forName` (`PhDefault.java:410-413`). So every genuine miss (a real absent attribute, an error path, `⊥` propagation) now pays a `Class.forName` plus a caught `ClassNotFoundException`. Even a *successful* extension dispatch re-probes and re-`take`s the class on **every** call — there is no memoization of the result in `PhDefault`.

2. `PhNest.owns(...)` calls `Class.forName` on **every** `take` before it consults its object cache (`eo-runtime/src/main/java/org/eolang/PhNest.java:115-118`). `PhNest.extension` caches the loaded object, but `owns` itself is not cached, so repeated `number.power` still probes the classpath each time.

Exception-driven control flow (`ClassNotFoundException`) plus classloader lookups on the dispatch path is the concern here — this is the innermost loop of dataization.

## Why it matters now

This change shipped with **no performance data**: the benchmark-comparison report on #5505 failed to generate ("one of the benchmarks is missing"), so the regression, if any, was never measured.

## Suggested resolution

- Cache the class-existence probe (and ideally the resolved extension) keyed by the fully-qualified name, so a given `type.attr` pair is looked up at most once. A negative cache (name → absent) matters as much as the positive one, since the common case is "attribute is genuinely absent".
- Consider guarding the fallback so it only fires for names that can plausibly be extensions (see the missing `ρ`/`φ`/`λ`/`Δ` guard noted below), avoiding a probe for internal attributes entirely.

Note: `PhNest.take` already skips the probe for `Phi.RHO` (`PhNest.java:57`), but `PhDefault.extension` has no such guard — it will `Class.forName` on a missing `ρ`, `φ`, `λ` or `Δ` too. Harmless (the class won't exist) but it is exactly the wasteful probe this issue is about.

Introduced in #5505.

gemshrine:
Closes #5511.

Since #5505 ("packages are objects"), attribute dispatch falls back to a package-extension lookup whenever a `take` would resolve to `⊥`. That fallback probes the classpath with `Class.forName` **uncached**, on the innermost loop of dataization:

- `PhDefault.extension` → `exists(...)` → `Class.forName` on **every** attribute miss (a real absent attribute, an error path, `⊥` propagation), paying a caught `ClassNotFoundException` each time;
- `PhNest.owns` → `Class.forName` on **every** `take`, before its object cache.

The answer for a given fully-qualified name never changes at runtime.

### Fix
- New `OnClasspath` memoizes the probe — **both** the positive and the negative answer — keyed by the class name, so a given name is looked up at most once. Both probe sites route through it.
- `PhDefault.extension` additionally **skips the probe entirely** for the structural attribute names `λ`/`φ`/`ρ` (they can never be a package extension), so a miss on `ρ`/`φ`/`λ` no longer probes `Φ.<pkg>.ρ` at all.

Behaviour is unchanged — only the redundant classloader lookups and exception-driven control flow on the hot path are removed. `OnClasspath` caches the boolean existence, **not** the resolved `Phi`, so no mutable object is shared (cf. #5514).

### Changes
- `OnClasspath.java` (new) + `OnClasspathTest.java`.
- `PhDefault.java` — route through `OnClasspath`, drop the local `exists`, add the `λ`/`φ`/`ρ` guard.
- `PhNest.java` — `owns` routes through `OnClasspath`.

Verified locally: `eo-runtime` suite green apart from two failures unrelated to this change — the network-dependent `EOsocketTest.refusesConnectionViaSyscall` and `EOfileTest.tests_appending_data_to_file` (the latter tracked in #5503).

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/OnClasspath.java
+++ b/eo-runtime/src/main/java/org/eolang/OnClasspath.java
@@ -1,0 +1,56 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+
+package org.eolang;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Cached check of whether a Java class is present on the classpath.
+ *
+ * <p>Package-extension dispatch probes the classpath with {@code Class.forName}
+ * on every attribute miss (an inner loop of dataization), and the answer for a
+ * given fully-qualified name never changes at runtime. This memoizes both the
+ * positive and the negative answer, so a given name pays the lookup and the
+ * caught {@link ClassNotFoundException} at most once.</p>
+ *
+ * @since 0.62
+ */
+final class OnClasspath {
+
+    /**
+     * Name to presence, cached across all lookups.
+     */
+    private static final Map<String, Boolean> CACHE = new ConcurrentHashMap<>(0);
+
+    private OnClasspath() {
+    }
+
+    /**
+     * Is there a Java class with this name on the classpath?
+     * @param cls The fully-qualified Java name
+     * @return TRUE if it exists
+     */
+    static boolean has(final String cls) {
+        return OnClasspath.CACHE.computeIfAbsent(cls, OnClasspath::probe);
+    }
+
+    /**
+     * Probe the classpath for a class, uncached.
+     * @param cls The fully-qualified Java name
+     * @return TRUE if it exists
+     */
+    private static boolean probe(final String cls) {
+        boolean found;
+        try {
+            Class.forName(cls);
+            found = true;
+        } catch (final ClassNotFoundException ex) {
+            found = false;
+        }
+        return found;
+    }
+}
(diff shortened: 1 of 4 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (137 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+307](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5532, opened 2026-07-15
Move number's supplementary objects into its package

yegor256:
This continues the #5501 layout work, this time for `number`. The object keeps its four atoms (`gt`, `times`, `plus`, `div`) and moves `lt`, `lte`, `gte`, `minus`, `neg`, `floor`, `is-nan`, `is-finite`, and `is-integer` into the `number` package as standalone objects, reached through the same package dispatch used by `string`/`tuple`. The constants `e` and `pi` move up to the root, next to `nan`, `pinf`, and `ninf`, since they are data objects rather than number helpers.

One thing worth flagging: `eq` and `as-i16`/`as-i32`/`as-i64` did not move, on purpose. `number` decorates `bytes`, and `bytes` already exposes objects with those exact names (`eq` is a bytes atom; the `as-i*` conversions live in the bytes package). Attribute lookup reaches the decoratee before the object's own package, so a package-level `number.eq` would be shadowed by `bytes.eq` — which compares raw bytes and would make `nan.eq nan` true and every `X.as-i64` reinterpret the double's bits. So those four stay as direct attributes that override the bytes versions. If we want them out of the object too, that needs a separate change to how dispatch orders the object's own package against the decoratee.

The nullary-extension test in `PhDefaultTest` used `pi` as its example of a number-package constant; since `pi` is now at the root, I pointed it at `sm.eol` instead. The whole `eo-runtime` suite (1546 tests) passes.

Refs #5501


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1123 >= 100); -8 for way too many hits-of-code (1123 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1838](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5533, opened 2026-07-15
fix(#5513): bind package-extension receiver into α0 consistently

### LINKED ISSUE #5513: Inconsistent binding of receiver in package extensions for implicit vs explicit dispatch
## What happens

There are two paths that resolve a package extension, and they bind the receiver into **different slots**:

- **Implicit** — `(number 42).power 3`. `PhDefault.extension` binds the receiver as **positional argument 0**:
  ```java
  // eo-runtime/src/main/java/org/eolang/PhDefault.java:399
  found.put(0, this);
  ```
- **Explicit** — `number.power 2 4`. `PhNest.extension` binds the receiver (the package/object itself) as **ρ**:
  ```java
  // eo-runtime/src/main/java/org/eolang/PhNest.java:135
  loaded.put(Phi.RHO, this);
  ```

So the same extension object, reached the two documented ways, sees the receiver in `α0` one way and in `ρ` the other — and the argument positions shift accordingly (implicit consumes `α0` for the receiver, explicit leaves all positions for the caller).

For `power [num x]` the two happen to reconcile — `2.power 4` binds `num=2` then `x=4`, and `number.power 2 4` passes both — which is why the tests in `number/power.eo` pass. But that reconciliation is a property of `power`'s specific signature, not a guarantee. An extension author who reads the receiver from `ρ` will find it present via the explicit form and absent via the implicit form (and vice-versa for `α0`).

## Why file it

This is a design/consistency concern rather than a crash, but it is a real trap for anyone writing package extensions: the mental model "an extension is called with the receiver bound" holds in two incompatible ways depending on call site. Worth deciding on a single convention (and documenting it) before extensions proliferate across the base library.

## Suggested resolution

Pick one binding convention for "the receiver of a package extension" and make both dispatch sites agree, or document explicitly that implicit dispatch curries into `α0` while explicit dispatch only sets `ρ` and requires the caller to pass every argument.

Introduced in #5505.

yegor256:
Closes #5513.

## Problem

Two dispatch paths resolve a package extension and bound the receiver into **different slots**:

- **Implicit** — `(number 42).power 3` — `PhDefault.extension` binds the receiver into **positional attribute `α0`** (`found.put(0, this)`).
- **Explicit** — `number.power 2 4` — `PhNest.extension` bound the package object into **`ρ`** (`loaded.put(Phi.RHO, this)`).

So the same extension, reached the two documented ways, saw its receiver in `α0` one way and in `ρ` the other.

## Fix

Every package extension in the base library reads its receiver from a **positional attribute** (`num`, `x`, `sequence`, …), never from `ρ` — so `α0` is the real convention, and `PhDefault` already follows it.

The issue suggested making `PhNest` do `loaded.put(0, this)` instead. That does not work for the explicit form: there `this` is the **abstract `number` package object** (`[as-bytes] > number`, with `as-bytes` unset), not a concrete value. Binding it into `α0` overflows `power`'s two voids (`num`, `x`) and leaves `num.as-bytes` unresolvable, so `number.power 2 4` — a supported, tested form — would break.

Instead, this PR **removes the spurious `ρ` binding** in `PhNest.extension`. The receiver of a package extension now always lives in `α0` in both forms:

- implicit dispatch auto-binds the left-hand value into `α0`;
- explicit namespace dispatch is a plain access that leaves the slot for the caller to fill as the first argument.

`ρ` is no longer the receiver in either path, which is exactly the inconsistency the issue reports. The convention is now documented on both `PhNest.extension` and `PhDefault.extension`.

## Changes

- `eo-runtime/.../PhNest.java` — drop `loaded.put(Phi.RHO, this)`; document the `α0` convention.
- `eo-runtime/.../PhDefault.java` — document that the receiver always lives in `α0`, cross-referencing `PhNest`.
- `eo-runtime/.../PhNestTest.java` — regression test asserting an extension taken through the namespace is handed out without `ρ` bound.

## Verification

- `PhNestTest` — 4/4 pass (incl. the new regression test).
- `EOpowerTest` — 48/48 pass, covering both dispatch paths: `2.power 4`, `3.power 2`, and `number.power 2 4`.
- Full `eo-runtime` suite — 1542 tests, no functional failures. (The only red test, `MainTest.printsVersion`, is an environmental artifact: the sandbox JVM echoes `Picked up JAVA_TOOL_OPTIONS: …` to stderr, polluting the `--version` capture; unrelated to this change.)

https://claude.ai/code/session_011WnhmEgyoPvv1n6DBJryju

---
_Generated by [Claude Code](https://claude.ai/code/session_011WnhmEgyoPvv1n6DBJryju)_

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -381,8 +381,12 @@ public class PhDefault implements Phi, Cloneable {
      * <p>When {@code (number 42).power} finds no {@code power} attribute, the
      * runtime looks for an object {@code Φ.number.power} in the package named
      * after this object's forma. If it exists, it is returned with this object
-     * bound as its first argument, so {@code (number 42).power 3} reads as
-     * {@code number.power 42 3}. When there's no such object, the terminated
+     * bound as its first argument ({@code α0}), so {@code (number 42).power 3}
+     * reads as {@code number.power 42 3}. The receiver of a package extension
+     * always lives in {@code α0}: implicit dispatch binds it here, while
+     * explicit dispatch through the namespace ({@code number.power 42 3})
+     * leaves the slot for the caller to fill (see {@code PhNest.extension}).
+     * When there's no such object, the terminated
      * computation stays terminated. When the object exists but has no free
      * positional attribute to receive the bound object (for example a nullary
      * constant like {@code number.pi}), a clear error is raised instead of the
--- a/eo-runtime/src/main/java/org/eolang/PhNest.java
+++ b/eo-runtime/src/main/java/org/eolang/PhNest.java
@@ -142,15 +142,25 @@ final class PhNest implements Phi {
 
     /**
      * Take an extension object from the package, as it is, without binding.
+     *
+     * <p>The receiver of a package extension always lives in positional
+     * attribute {@code α0}, never in {@code ρ}. Explicit dispatch through the
+     * package namespace (for example {@code number.power 2 4}) is a plain
+     * access: nothing is bound here and the caller supplies every argument,
+     * the first of which plays the receiver's role. This mirrors implicit
+     * dispatch, where {@link PhDefault} binds the receiver into {@code α0} of
+     * the same extension (see {@code PhDefault.extension}). Binding {@code ρ}
+     * here would be misleading — the package object standing behind
+     * {@code number} is not a usable receiver value — so it is deliberately
+     * left untouched.</p>
+     *
      * @param name The name of the extension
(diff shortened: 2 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (33) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1842](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5534, opened 2026-07-16
Move minimum and maximum into the tuple package

yegor256:
This moves `minimum.eo` and `maximum.eo` out of the `number` package and into `tuple`, continuing the object-layout cleanup from #5501. Both objects fold over a sequence to pick an extreme element, so they read as tuple operations, not number ones.

Sitting under `number` also made them awkward to reach: the receiver is the sequence, so dispatch on a tuple would look in `Φ.tuple`, never `Φ.number` — which meant these two could only be called by explicit qualification. In `tuple` they line up with the other reducers (`reduced`, `back`, `range`). I also dropped the `+alias tuple.is-empty` and `+alias tuple.reduced` metas, which are redundant now that those objects are same-package siblings referenced by bare name.

No callers change — nothing outside the two files referenced them. Full `eo-runtime` suite stays green (1546 tests, 0 failures).


yegor256:
@yegor256 could you take a look when you get a chance? Happy to explain anything if the packaging rationale is unclear.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1846](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5498, opened 2026-07-14
Voids in selector-style formations override attributes and methods in return values without checks


morphqdd:
## What happens

#5478 revealed a whole class of traps, and right now nothing guards against it. The result of applying a formation carries the formation's voids as attributes, and attribute lookup wins over the phi fallthrough. So a helper formation whose phi returns one of its voids shadows the same-named methods of the value it returns. The `if.` branch selector was one instance: its voids were named `left` and `right`, so `(if. c a b).right 1` silently took the selector's void instead of `bytes.right`, and the failure surfaced far away as `This void attribute "data" is already set` or, worse, was swallowed by `Expect` into a generic `must be an i64` message.

The same shape exists elsewhere in the standard library. For example `while` is `[condition body] > while`, so `.condition` or `.body` dispatched on a `while` result will never reach the returned value. Any future helper with an innocent void name (`length`, `size`, `at`, `not`, `eq`) plants the same mine, and nothing will catch it until some composition explodes at runtime with an unrelated error message.

## What should happen

Three layers of protection:

1. A naming convention: in a formation whose phi evaluates to one of its own voids (selector-style helpers), void names must not collide with common attribute names. A simple rule that works: suffix them, as `left-branch` and `right-branch` did in #5497.

2. A lint that enforces it. A WPA lint can flag a formation whose phi returns a void while its other voids (or the returned one) use names that exist as attributes on stdlib objects (`bytes`, `number`, `string`, `tuple`, `i64` and friends). That check would have caught the `bool` selector on the day it was written.

3. Unit tests in `eo-runtime` that pin the non-capture behavior for the existing helpers, in the same way #5497 added two tests for `if.` results: dispatch a method of the returned value through the helper result and assert it reaches the value. At least `while` deserves such a pair today.

Related: #5478, #5497, and 8314d70 where the `bool` selector was introduced.

yegor256:
@morphqdd @maxonfjvipon maybe we should make all void attributes inaccessible (private) by default?

E.g., this code:

```
[a] > foo
  a.plus 1 > b
```

This will work:

```
(foo 42).b
```

But this will fail with "attribute is private":

```
(foo 42).a
```

We can implement this just by changing their names in compile time. The code above will change to:

```
[a🌵] > foo
  a🌵.plus 1 > b
```

This will make `a🌵` inaccessible from outside, but accessible from this very `.eo` file.

WDYT?

0crat:
[...]

PR #5535, opened 2026-07-16
feat(#5529): allow `>>` file-local handle on void attributes

### LINKED ISSUE #5529: File-local >> handle from #5525 is not allowed on void attributes
## What happens

Since #5525, an object can carry a file-local handle after `>>`, for example `[n] >> fibo`, so it stays anonymous on the surface but is referenceable inside the same `.eo` file. Void attributes do not support this. Both forms are rejected on current `master` (with #5525 merged):

```
[] > foo
  ? >> left
```

gives:

```
[13:9] error: 'trailing garbage after name suffix'
```

and a bare auto-named void:

```
[] > foo
  ? >>
```

gives:

```
error: 'a void attribute must be written as `? > name`'
```

So a void must always have an explicit, externally visible name.

## Why it matters

This blocks the clean fix proposed for #5498 (voids shadowing methods of the returned value). In `true.eo` and `false.eo` the branch selector is `[left right]`, and because `left`/`right` are public void names, `(if. c a b).left` from outside captures the void instead of the returned value's method (root cause of #5478). Making the selector voids anonymous with a file-local handle would remove the shadowing at the source while keeping public accessor voids (like `i64`'s `[as-bytes]`) untouched, because it is opt-in per void:

```
bool > false
  []
    ? >> left
    ? >> right
    right > @
```

This is exactly what @yegor256 suggested in #5498, but it does not parse today.

## What should happen

Allow the `>>` file-local handle on void markers, so `? >> name` declares a void whose external name is auto-generated (a cactus name, unreachable from outside) while `name` is a file-local handle usable within the same file. Filling stays positional, so the auto-generated external name does not affect how the void is bound by callers.

yegor256:
Fixes #5529.

## Problem

Since #5525 an object can carry a file-local handle after `>>` (`[n] >> fibo`), but void attributes did not support it. Both `? >> left` (→ `trailing garbage after name suffix`) and a bare `? >>` (→ `a void attribute must be written as '? > name'`) were rejected, so a void always had to have an explicit, externally visible name. This blocked the clean #5498 fix (making `true.eo`/`false.eo` selector voids anonymous so they stop shadowing the returned value's methods — root cause of #5478).

## What changed

- **`LnVoid`** now accepts the `>>` auto-name suffix (`Suffix.Form.AUTO`) in addition to `> name`. `? >> name` declares a void whose external `@name` is an auto-generated cactus name (unreachable from outside) while `name` becomes a file-local handle usable within the same `.eo` file; a bare `? >>` is a void with an auto-generated name and no handle. A `!` const marker or `/sig` atom signature on a void is still rejected. When a handle is present, `LnVoid` emits the `@local` marker via `Emit.local`, exactly as `LnFormation`/`LnOnlyPhi` already do.
- No new machinery below the parser front-end: the existing first-pass `resolve-local-names` reshape collects the per-file `@local → @name` table, rewrites every `@base` equal to a handle into the matching cactus name, drops the markers, and reports a duplicate handle as a compile-time error. Filling stays positional, so the auto-generated external name does not affect how callers bind the void.

This makes the shape suggested in #5498 parse:

```
bool > false
  []
    ? >> left
    ? >> right
    right > @
```

## Spec

- `PARSER_SPEC.md` R-3.4.7 documents the `? >> name` shape and the `!`/`/sig` rejection.
- R-3.10.12 lists the vertical void as an accepted site for file-local handles.

## Tests

- `eo-syntax/void-auto-named-handle.yaml` — end-to-end through the full canonical pipeline: two `? >> name` voids get cactus `@name`s, the handle reference resolves to the void's cactus name, and no `@local` markers survive.
- `eo-packs/parse/void-local-handle.yaml` — the `resolve-local-names` rewrite over the parser output for voids.
- `eo-packs/parse/void-local-handle-duplicate.yaml` — a handle declared twice across voids is a `resolve-local-names` error.

All 936 `EoSyntaxTest` + `SuffixTest` cases pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GgJDjPRLnJfdSxGqkSdUgD)_


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1854](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5536, opened 2026-07-16
Promote eol, map, set and buffer to the root package

yegor256:
Continuing the layout cleanup from #5501, this lifts a few general-purpose objects out of their packages and up to the root, where they sit next to `string`, `number` and the data constants. `eol` leaves `sm`, `map` and `set` leave `tuple`, and `string-buffer` leaves `string` and is renamed to plain `buffer`. None of them are really tied to the package they were in — they read more naturally as top-level objects.

Each still needs to reach the siblings it left behind, so I added the `+alias` metas for that: `sm.os` for `eol`, `tuple.contains/filtered/hash-code-of/mapped` for `map`, and `tuple.mapped` for `set`. `stdin` was the only external caller (of `eol`); it drops its now-dead `+alias sm.eol` and just uses the bare root name. Nothing else referenced any of these four.

One runtime tweak came along for the ride. The nullary-extension guard in `PhDefault` only caught `ExFailure`, but the last nullary *constants* (`pi`, `eol`) have now moved to the root, so the only nullary package objects left are `[]`-formations like `sm.os`, whose read-only slot raises `ExReadOnly`. Widening the catch to their shared parent `ExAbstract` keeps the friendly "takes no arguments" message working for that shape too, which is exactly what the surrounding comment always promised. The `PhDefaultTest` example moved from `sm.eol` to `sm.os` for the same reason. Full `eo-runtime` suite stays green (1550 tests).


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1862](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5537, opened 2026-07-16
Rename sprintf/sscanf and move range to root

yegor256:
This renames the standard-library object `string.sprintf` to `string.printf` and `string.sscanf` to `string.scanf`. The rename covers the two `.eo` objects, their JVM atoms (`EOsprintf`/`EOsscanf`), the `SprintfArgs` helper, both Java tests, and every caller that aliased or called them. It also moves `range` out of the `tuple` package up to the root.

The old `s`-prefixed names came from C's string-formatting family, but in EO these objects just format and parse strings, so the shorter names read better. `range` builds a general list from any object that has `next`/`lt`, so it never belonged under `tuple`; at the root it stays reachable by bare name and now aliases `tuple.is-empty` for its own test.

All 1550 eo-runtime unit tests pass. This continues the naming work under #5501.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,7 @@ argument: a copy of the object `printf`:
 # Says hello to Jeff.
 
 io.stdout > [] > app
-  string.printf
-    "Hello, %s!"
+  "Hello, %s!".printf
     * "Jeffrey"
 ```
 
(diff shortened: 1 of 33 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (432 >= 100); -8 for way too many hits-of-code (432 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1866](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5539, opened 2026-07-16
Hide incoming attributes of bytes and tuple objects

yegor256:
This makes the incoming arguments of the standard objects in the `bytes/` and `tuple/` packages private. Each object drops its public `[a b]` void-attribute signature and instead declares the arguments as anonymous `? >> a` handles inside the body. The arguments stay reachable by name everywhere in the same file, but they are no longer part of the object's public surface, so nobody outside can reach into `some-tuple.each.func` and friends.

Why this helps: these attributes were only ever meant as inputs, never as a public API. Exposing them let callers depend on internals that are really implementation detail. The `? >>` file-local handle (from #5525) gives us exactly the visibility we want without changing any behavior.

Twenty-eight files change across the two packages. The six `cant-*` fallback objects — `as-i16`, `as-i32`, `as-i64`, `as-number`, `minimum`, `maximum` — are deliberately left alone for a later pass. One small follow-on: in `range-of-numbers` a nested child used to read `^.start`, which no longer resolves once `start` is anonymous, so it now reads the `start` handle directly. All 1550 eo-runtime unit tests pass.

### DIFF
--- a/eo-runtime/src/main/eo/bytes/as-bool.eo
+++ b/eo-runtime/src/main/eo/bytes/as-bool.eo
@@ -7,7 +7,8 @@
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
 
-[bts] > as-bool
+[] > as-bool
+  ? >> bts
   bts.eq 01- > @
 
   ++> tests-convertible-to-bool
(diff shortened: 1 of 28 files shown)


yegor256:
@yegor256 could you take a look when you get a chance? Happy to explain any of the per-file changes if something looks off.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (237 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1870](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5540, opened 2026-07-16
Move bytes-as-array to bytes package as array

yegor256:
Moves the `bytes-as-array` object out of the `tuple` package into the `bytes` package and renames it to a plain `array`. Its two string callers, `low-cased` and `up-cased`, now import `bytes.array` instead of `tuple.bytes-as-array`.

The object takes a `bytes` value and hands back a tuple of single-byte values. Its input is bytes and the operation is about bytes, so `bytes.array` reads truer to what it does than `tuple.bytes-as-array` did. The shorter name drops the redundant prefix now that the package already says "bytes".

Part of #5501. All 1550 eo-runtime unit tests pass locally.

### DIFF
--- a/eo-runtime/src/main/eo/bytes/array.eo
+++ b/eo-runtime/src/main/eo/bytes/array.eo
@@ -4,12 +4,12 @@
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
-+package tuple
++package bytes
 +version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1874](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5542, opened 2026-07-16
Move tmpdir to root as mktemp and getenv to root

yegor256:
This moves two standard-library objects out of their packages and into the root. The `fs.tmpdir` object becomes the root object `mktemp`, and `sm.getenv` becomes the root object `getenv`. The `fs.file` and `fs.dir` tests that referenced `tmpdir`, and the alias list of `mktemp` itself, are updated to match.

Both objects are general enough to belong at the root, where callers reach them by bare name with no package alias, the same way `seq` or `string` are used today. This continues the "packages are objects" cleanup under #5501.

All 1550 eo-runtime unit tests pass locally. Nothing else in the tree references either object, so no other call sites needed touching.

### DIFF
--- a/eo-runtime/src/main/eo/fs/dir.eo
+++ b/eo-runtime/src/main/eo/fs/dir.eo
@@ -75,10 +75,10 @@
     and. > @
       d.exists
       d.is-directory
-    (tmpdir.tmpfile.deleted.as-path.resolved "foo-new").as-dir.made > d
+    (mktemp.tmpfile.deleted.as-path.resolved "foo-new").as-dir.made > d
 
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (106 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1882](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5543, opened 2026-07-16
Move system-clock to root as clock

yegor256:
This moves the wall-clock time source out of the sm package and into the root. The `sm.system-clock` object becomes the root object `clock`, so callers reach it by bare name with no package alias, the same way `seq` or `string` are used today. Its sole caller, `number.random`, is updated to drop the alias and pass `clock` directly.

The object reads the OS time syscall, so it needs the sm objects `os`, `posix` and `win32`; those are now brought in with explicit aliases. This continues the "packages are objects" cleanup under #5501.

All 1550 eo-runtime unit tests pass locally. The `clocked` object inside `number.random` keeps a void attribute also named `clock`; that attribute is local to `clocked`, so it does not clash with the new global object.

### DIFF
--- a/eo-runtime/src/main/eo/clock.eo
+++ b/eo-runtime/src/main/eo/clock.eo
@@ -2,14 +2,16 @@
 # Reads the underlying syscall: `gettimeofday` on POSIX, `_ftime32_s` on Windows.
 # Decorates an `i64` holding the current time in milliseconds since the Unix epoch.
 
++alias sm.os
++alias sm.posix
++alias sm.win32
 +architect yegor256@gmail.com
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1886](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5554, opened 2026-07-16
fix(#5548): Xmir.toEO prints reparseable EO for current-dialect constructs

### LINKED ISSUE #5548: `Xmir.toEO` prints non-reparseable EO for current-dialect constructs (voids, nested `^`, `+>` tests)
The `format` goal (`MjFormat`, added in #5516) reformats `.eo` sources by round-tripping them through `Xmir.toEO()` in `eo-printer`. When the goal is enabled on `eo-runtime`, the printed EO no longer parses back for a large share of the sources: a round-trip scan over `eo-runtime/src/main/eo` (parse → `toEO()` → re-parse) shows **31 of 138 files produce non-reparseable output and 13 more are non-idempotent**. The contract is meant to be guarded by `XmirTest.printsToParseableEo`, but its packs never exercise these constructs. The failures trace to a handful of independent, local defects in the printer XSL.

The three fully-isolated root causes:

- **Cactus-named voids are destroyed.** `print/inline-cactoos.xsl:35` deletes every `<o>` whose `@name` starts with `a🌵`, and `:24` inlines every base containing `.a🌵`. But a void declared as `? >> handle` (R-3.4.7 / R-3.10.12) also carries a cactus `@name`, so it and its references are wrongly consumed. Input `[] > as-bool` / `? >> bts` / `bts.eq 01- > @` prints as `[] > as-bool` with a bare, nameless `[]` — the void and the `φ` binding are gone, giving `object inside formation must have a name` (`bytes/as-bool.eo`, `bytes/left.eo`, `tuple/with.eo`). Fix: restrict both templates to non-void objects (exclude `@base='∅'`), and render a cactus-named void under a generated printable name in `to-eo-tree.xsl`.
- **Nested parent references lose all but the first hop.** `print/to-eo-tree.xsl:109-118` renders a `ξ.ρ…` base as `^` followed by `substring-after(@base,'ξ.ρ')`, so only the first `ρ` becomes `^`: `ξ.ρ.ρ.x` prints as `^.ρ.x`, which fails with `expected identifier` (`bool.eo:46`, `fs/path.eo`). Fix: translate every `ρ` segment to `^`, not just the leading one.
- **Test attributes print as ordinary objects.** The tail template `print/to-eo-tree.xsl:155-171` renders a `+`-prefixed test-attribute name (`eo:test-attr`) as `> +name`, i.e. `[] > +name`. Inside an atom the parser rejects it (`Atom cannot contain inner objects; only \`+>\` test attributes are allowed`, `dataized.eo`), and everywhere it drops the blank line R-6.5.3 requires before a test attribute. Fix: emit `+> name` (stripping the leading `+`) for `eo:test-attr` names, and have `Pretty` insert the mandatory preceding blank line.

The same round-trip also breaks on `@`/`φ` dispatch heads (printed literally as `φ. …`, giving `line shape not yet implemented` / `expected value` in `switch.eo:32` and `i64.eo:123`) and on reversed-dispatch heads (`normalized. > @` → `reversed dispatch missing receiver` in `fs/path.eo`, `fs/file.eo`, `nk/socket.eo`); these appear to share the special-symbol rendering gap in the `mode="head"` templates and are worth verifying alongside the fixes above. Each item is small and local, so they can be addressed independently, but until they are the `format` goal cannot be turned on for `eo-runtime`.

yegor256:
Closes #5548.

## Problem

The `format` goal (`MjFormat`) round-trips `.eo` sources through `Xmir.toEO()` in `eo-printer`. A round-trip scan (parse → `toEO()` → re-parse → `toEO()`) over `eo-runtime/src/main/eo` produced **31 non-reparseable and 13 non-idempotent files** out of 138. The failures trace to a handful of independent, local defects in the printer XSL/`Pretty`, which `XmirTest.printsToParseableEo` never exercised.

After this change the same scan is **138/138 clean** (0 non-reparseable, 0 non-idempotent).

## Fixes

- **Cactus-named voids destroyed** (`inline-cactoos.xsl`). A void declared as `? >> name` (R-3.4.7 / R-3.10.12) also carries a cactus `@name`, so the delete/inline templates were consuming voids and their references (`bytes/as-bool.eo`, `bytes/left.eo`, `tuple/with.eo` → `object inside formation must have a name`). Both templates are now restricted to genuine auto-named abstracts; a reference that resolves to a void (or resolves to nothing, e.g. `ξ.a🌵5-2.eq`) is left untouched. Surviving cactus-named voids are rendered under a generated printable identifier (`vL_P`) via `eo:printable` in `to-eo-tree.xsl`.

- **Nested parent references** (`to-eo-tree.xsl`). Only the first `ρ` became `^`, so `ξ.ρ.ρ.x` printed as `^.ρ.x` (`expected identifier` in `bool.eo`, `fs/path.eo`). A single `eo:surface`/`eo:translate-path` head renderer now maps *every* special segment: `ρ → ^`, `φ → @`, and bare `ξ`/`ρ`/`φ`/`Φ` → `$`/`^`/`@`/`Q`. This also fixes the `φ`/reversed-dispatch head cases (`switch.eo`, `i64.eo`, `file.eo`, `nk/socket.eo`, …).

- **Test attributes** (`to-eo-tree.xsl` + `Pretty`). A `+`-prefixed test attribute printed as `[] > +name`, which the parser rejects inside atoms (`dataized.eo`) and which dropped the mandatory blank line. It now prints as `[] +> name` (the desugared form of `++> name`), and `Pretty` inserts the blank line R-6.5.3 requires before every test attribute.

- **Reversed dispatch parenthesization** (`Pretty`). A receiver-only reversed dispatch (`0.@`) can no longer be inlined as an argument, because the parser rejects the redundant parentheses it would need (`(@. 0)`); it forces a vertical layout instead. Reversed dispatch *with* arguments (`times. -1 228`) still inlines with parentheses.

- **Empty-tuple false positives** (`tuples-to-stars.xsl`). The star reconstruction matched any base starting with `Φ.tuple`, wrongly starring objects like `Φ.tuple.is-empty` applied to an empty tuple (non-idempotent `tuple/is-empty.eo`, `tuple/maximum.eo`, `tuple/minimum.eo`). It now matches the tuple-chain node base exactly.

## Tests

Added print-packs exercising the previously-uncovered constructs (cactus voids, nested parents, phi dispatch, reversed dispatch, test attributes), verified by both `printsToEo` and `printsToParseableEo`. All 71 `eo-printer` tests pass and `qulice:check` is clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KhKUnSEGJTyYo8c3eyxUpd)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (192 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1918](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5555, opened 2026-07-16
Float pipe predecessor up when it sits in an argument block

### LINKED ISSUE #5526: Pipe `|` in an argument block emits the predecessor formation as an extra argument
A pipe-application that follows a `>>`-auto-named formation inside another object's *vertical argument block* makes the enclosing application receive two arguments instead of one. Given

```
[y] > app
  y.plus > @
    [x] >>
      x.plus 1 > @
    | 5
```

the parser emits `y.plus` with both `α0` = the bare formation and `α1` = the formation applied to `5`; it should receive only the one applied argument. Verified by parsing with `org.eolang.parser.EoSyntax` on current `master`.

The cause is the `o[@pipe]` template in `eo-parser/src/main/resources/org/eolang/parser/parse/wrap-applications.xsl:24`: it rewrites the pipe into an application but deliberately leaves the predecessor formation "in place" (see the comment at lines 16-20, and R-3.14.7 in `eo-parser/PARSER_SPEC.md:612`). That is correct inside an object body, where the formation and the pipe application both become named attributes, but inside an argument block both siblings are collected as positional arguments (`α0`, `α1`), so the application's arity is wrong. It then fails confusingly downstream — for example `number.times` getting two arguments when a recursive helper is inlined in `eo-runtime/src/main/eo/number/sine.eo`.

The smallest fix is to make the pipe consume its predecessor when the predecessor is not a body-level binding: in that case `wrap-applications` should fuse the applied args into the single formation node (marking them `@as`, the fused shape already proposed in #5518) instead of emitting the formation as a separate sibling, so only one argument reaches the enclosing application. Alternatively, reject a pipe in argument position with a clear parse-time error. Related: #5518.

maxonfjvipon:
A pipe that follows a named formation inside another object's argument block used to leave the formation as a separate positional argument, so the enclosing application received two arguments where it should get one. The recursive helper inlined in `number/sine.eo` hit this as `number.times` being handed two arguments.

`wrap-applications` now recognises a pipe predecessor that sits in an argument block — its parent carries a `@base`, so its siblings are collected as arguments — and marks it with a transient `@float-up`. `vars-float-up` hoists that formation's definition to the nearest abstract object and drops its in-place argument slot, leaving the pipe application (which refers to the floated definition) as the only argument. So

```
[] > foo
  bar > @
    [] > hello
    | 5
```

parses to the same XMIR as

```
[] > foo
  bar > @
    $.hello 5
  [] > hello
```

In a formation body nothing changes: the predecessor stays put as a named attribute next to the pipe application.

Pipes stay restricted to named formations, an auto-named `>>` included; a bare formation with a pipe is a parse error. New typo packs cover that, and printer packs confirm both a body pipe and an argument-block pipe round-trip.

Closes #5526

### DIFF
--- a/eo-maven-plugin/src/test/resources/org/eolang/maven/transpile-packs/pipe-in-argument-block.yaml
+++ b/eo-maven-plugin/src/test/resources/org/eolang/maven/transpile-packs/pipe-in-argument-block.yaml
@@ -1,0 +1,22 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets:
+  - /org/eolang/maven/transpile/set-locators.xsl
+  - /org/eolang/maven/transpile/set-original-names.xsl
+  - /org/eolang/maven/transpile/classes.xsl
+  - /org/eolang/maven/transpile/attrs.xsl
+  - /org/eolang/maven/transpile/data.xsl
+  - /org/eolang/maven/transpile/to-java.xsl
+asserts:
+  - /object[not(errors)]
+  - //java
+input: |
+  +package foo
+
+  [y] > app
+    y.plus > @
+      [x] >>
+        x.plus 1 > @
+      | 5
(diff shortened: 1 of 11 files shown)


maxonfjvipon:
@yegor256 this fixes #5526 the way we discussed: the predecessor formation floats up via vars-float-up and the argument block keeps a single `$.foo 5` application. Pipes stay named-formation-only (bare formations rejected, with typo packs), and printer packs confirm round-trip. Ready for review.

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (168 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+753](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5557, opened 2026-07-16
Allow inline void attributes on named trailing-dot objects

### LINKED ISSUE #5553: Parser incorrectly treats inline void attributes on named trailing-dot objects as trailing garbage
The parser rejects inline void attributes when they decorate a *named trailing-dot* object such as `if. > [t] >> rec`, raising `trailing garbage after name suffix`. The very same `[voids] >> name` suffix is accepted after an ordinary method (`x.plus > [t] >> rec`) and a named trailing-dot object without voids parses fine (`if. > rec`), so only the intersection — trailing-dot form *plus* inline voids — fails. The rejection is raised by `endsClean` in `eo-parser/src/main/java/org/eolang/parser/Suffix.java:416-429`, which is reached on the trailing-dot naming path and treats the leading `[` of `[t]` as leftover text.

Minimal reproduction — this fails:

```
+package foo

[x] > obj
  if. > [t] >> rec
    t.eq 0
    1
    2
  rec x > @
```

with `[4:8] error: 'trailing garbage after name suffix'` (the caret sits on the `[`), followed by cascading indent errors. Swapping `if.` for a regular method (`x.plus > [t] >> rec`) parses with zero errors, and so does the explicit two-line form the sugar should desugar to:

```
  [t] >> rec
    if. > @
      t.eq 0
      1
      2
```

This surfaced in `eo-runtime/src/main/eo/tuple/last-index-of.eo`: the two-line form compiles, the equivalent one-line sugared form does not. The two objects are behaviorally identical, so it is confusing that `[voids] >> name` is legal after `x.plus` yet illegal after `if.`.

The fix is to route the trailing-dot form's name suffix through the same void-aware suffix parser the ordinary method-application path uses, so `endsClean` no longer sees `[t] >> rec` as trailing garbage — i.e. allow the inline `[voids]` suffix on the named trailing-dot form, aligning it with the method form.

yegor256:
Fixes #5553.

## Problem

The parser rejected inline void attributes when they decorated a *named trailing-dot* object such as `if. > [t] >> rec`, raising `trailing garbage after name suffix`. The very same `> [voids] >> name` suffix was accepted after an ordinary method (`x.plus > [t] >> rec`), and a named trailing-dot object without voids parsed fine (`if. > rec`) — so only the intersection (trailing-dot form *plus* inline voids) failed.

Root cause: `x.plus > [t] >> rec` is classified as an **only-phi formation** (`LnOnlyPhi`), which understands the inline `[voids]` suffix, whereas `if. > [t] >> rec` was classified as a **reversed dispatch** (`LnReversed`), which handed the whole `> [t] >> rec` tail to the void-unaware `Suffix` parser — and `endsClean` treated the leading `[` of `[t]` as leftover text.

## Fix

- `Eo.applicative` now recognises an only-phi formation (a top-level `> [` marker) **before** a bare reversed dispatch, so a trailing-dot LHS carrying an inline void suffix is routed to `LnOnlyPhi`.
- `LnOnlyPhi` now emits its `φ` slot through the shared `Emissions.expression` helper, which already knows how to render a reversed dispatch (`<o base='.if'>` with no `@method`, mirroring `LnReversed`) as well as an ordinary head + `.method` chain. A small `bare` helper decides openness up front: a reversed φ with no horizontal args stays `OPEN` so the receiver/method args attach as deeper-indent lines, while one with horizontal args is `HORIZONTAL_COMPLETED`.
- The one-line `if. > [t] >> rec` sugar now desugars to the same shape as its explicit two-line equivalent (`[t] >> rec` / `if. > @`). Root heads (`@.` → `.φ`) and fragile dispatch (`foo?.` → `@fragile`) are handled by the reused emitter too.

## Tests

- `LnOnlyPhiTest`: a reversed-dispatch LHS emits `<o base='.if' name='φ'>` with no `@method` and the voids as children; a reversed φ carrying horizontal args is `HORIZONTAL_COMPLETED`.
- `EoTest`: the one-line sugar from the issue parses with no errors and produces the same `φ`-as-reversed-dispatch shape as the two-line form.

All `eo-parser` tests pass; `mvn -Pqulice verify` is clean.


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (160 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1922](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5558, opened 2026-07-16
Extract chunk object; add chunk.to-output

yegor256:
Toward the naming/structure standardization of #5501. Rebased onto current master.

## What changed

- **New root object `chunk`** (`eo-runtime/src/main/eo/chunk.eo`): the memory-block handle that used to be the inline `malloc.of.allocated` formation — `id`, `size`, `resized`, `copy`, `read`, `write`, `put`, `get`.
- **`malloc.of` uses it** (`chunk > allocated`), so `malloc.empty` and `malloc.for` produce chunks too. `malloc` stays the thin factory.
- **`io.malloc-as-output` → `chunk.to-output`** (`eo-runtime/src/main/eo/chunk/to-output.eo`): now a member of the `chunk` package, so it is available on any chunk via package-extension dispatch — `mem.to-output` — with no explicit reference. The `input` objects (`copied`, `tee`, `length`, `as-bytes`) use that form; the now-empty `io` package is gone.
- **Native atoms** move from `org.eolang.EOmalloc$EOof$EOallocated$EO*` to `org.eolang.EOchunk$EO*` (still in `org.eolang`, so `Heaps` stays package-private — no visibility change). The `malloc.of.@` atom now copies the allocated chunk before setting its `id`, since `chunk` resolves to a package object.

## Validation

Full `eo-runtime` suite green: 1707 tests, only `MainTest.printsVersion` fails locally, an artifact of the sandbox injecting `JAVA_TOOL_OPTIONS` into the subprocess output (not reproducible on CI). qulice clean; EO canonical formatting (`MjFormat`) passes. Total diff is **196 lines**, under the 200-line `pr-size` limit.

## Cross-runtime follow-up

The `+rt node eo2js-runtime` runtime keys off the old atom names (`malloc.of.allocated.*`); it will need a matching rename in that repo.

### DIFF
--- a/eo-runtime/src/main/eo/chunk.eo
+++ b/eo-runtime/src/main/eo/chunk.eo
@@ -1,0 +1,48 @@
+# A `chunk` is a handle to a block of data allocated in heap memory.
+# It is created by `malloc.of` (and, indirectly, by `malloc.empty` and
+# `malloc.for`) and passed into the user scope, where it provides the API
+# to read and write the data of the block:
+#
+# ```
+# malloc.of
+#   8                       # allocate 8 bytes length block in memory
+#   [c]
+#     seq * > @
+#       c.write 2 "Hello"   # write object "Hello" with offset 2
+#       c.read 3 4          # read 4 bytes from offset 3 -> "ello"
+#       c.put 42            # write object 42 with offset 0
+#       c.get               # get all the data from the memory block -> 42
+#       c.size              # get size of the block
+#       c.id                # get identifier of the block
+#       c.@                 # the same as c.get
+# ```
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++rt jvm org.eolang:eo-runtime:0.0.0
++rt node eo2js-runtime:0.0.0
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
++unlint mandatory-package
+
+[id] > chunk
+  get > @
+  read > get
+    0
+    size
+  [] > size /Q.number
+  [size] > resized /Q.chunk
+  write > [source target length] > copy
+    target
+    read
+      source
+      length
+  [offset length] > read /Q.bytes
+    ? > cant-read /{Q.string}
+  [offset data] > write /Q.bool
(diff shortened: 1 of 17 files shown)


yegor256:
@maxonfjvipon please, review this one carefully

maxonfjvipon:
@yegor256 what's the purpose of this object, why we need to make a separated object?

yegor256:
@maxonfjvipon we want to replace `malloc-to-output` object with `to-output`. Instead of doing this:

```
malloc.empty
  [m]
    copied > @
      "hello, world".to-input
      malloc-to-output m
```

We want to do this:

```
malloc.empty
  [m]
    copied > @
      "hello, world".to-input
      m.to-output
```

To make it possible, we need `to-output` be an attribute of `malloc.allocated`.

I decided to extract `chunk` from `malloc.eo` into its own file `chunk.eo`. This gives un ability to create `malloc/chunk/to-output.eo` file, which becomes exactly the attribute of `chunk` that is returned by `malloc.of`.

maxonfjvipon:
@yegor256 
1. What's wrong with `malloc-as-output`?
2. Why not just add `to-output` to `malloc.of.allocated?`

maxonfjvipon:
@yegor256 by making it a separated object we give a user an ability to get access to the memory outside of `malloc` object. I'm not sure it's a good idea

yegor256:
@maxonfjvipon 

> Why not just add to-output to malloc.of.allocated?

We would have to put `to-output` into the `malloc.eo` file. This may be OK for a single object, but for many objects it's not.

yegor256:
@maxonfjvipon 

> by making it a separated object we give a user an ability to get access to the memory outside of malloc object. I'm not sure it's a good idea

The `malloc` object is just a shell, it doesn't do anything useful. It's a factory, to be honest. The real work is done in the `chunk`.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (196 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2720](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5566, opened 2026-07-16
fix(#5562): printer emits `++>` shorthand instead of `[] +>`

### LINKED ISSUE #5562: `++>` shorthand is incorrectly expanded to `[] +>` in `Xmir.toEO()` output
The `++>` test-attribute shorthand was added to the parser in #5445 (`++> name` desugars to a bare `[]` formation with a `+> name` suffix). The reverse direction, however, expands it: since the #5548 fix (PR #5554), `Xmir.toEO()` prints every test attribute as `[] +> name`. Running the new `format` goal on `eo-runtime` therefore rewrites all the concise `++> tests-…` declarations into the verbose `[] +> tests-…` form — for example `eo-runtime/src/main/eo/set.eo:29` now reads `[] +> tests-set-rebuilds-itself`.

This is the opposite of what #5445 set out to achieve. As that issue documented, `[] +>` is pure boilerplate that occurs thousands of times and never carries void parameters, so the empty `[]` head is noise on every test. A canonicalizing printer should emit the shorter, preferred sugar, not expand it back.

The relevant code is the tail template of `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:168-173` (which prints the `+> name` suffix) together with the abstract-head renderer that prints `[]`. When an object is a test attribute (`eo:test-attr`) with an empty void list, the head and suffix should collapse into a single `++> name` head with no separate `[]`; the `Pretty` blank-line rule added for test attributes stays as is.

yegor256:
Fixes #5562.

## Problem

Since the #5548 fix (#5554), `Xmir.toEO()` printed every test attribute as the verbose `[] +> name`, undoing the `++>` shorthand added to the parser in #5445. Running the `format` goal on `eo-runtime` therefore rewrote all the concise `++> tests-…` declarations into the noisy `[] +> tests-…` form. A canonicalizing printer should emit the shorter, preferred sugar, not expand it back.

## Fix

In `to-eo-tree.xsl`:

- **Abstract-head renderer**: a test attribute (`eo:test-attr`) with an empty void list now emits no head at all, instead of the boilerplate `[]`.
- **Tail renderer**: for such a test attribute the suffix collapses into a single `++> name` head (no leading space, no separate `[]`). Test attributes that carry void params — should any exist — still print the `[params] +> name` form.

The `Pretty` blank-line rule for test attributes is unchanged.

## Tests

- Updated `test-attribute.yaml` print-pack to expect `++> tests-something`. This also fixes the round-trip: the pack's `origin` already used `++>`, so `printsToEo` and `printsToParseableEo` (idempotency) now both hold.
- Full `eo-printer` suite green (75 tests, including all 53 `XmirTest` cases).

https://claude.ai/code/session_01LgBH5Zw4qLqJ59hthAvVd1

---
_Generated by [Claude Code](https://claude.ai/code/session_01LgBH5Zw4qLqJ59hthAvVd1)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -143,14 +143,19 @@
   </xsl:template>
   <!-- ABSTRACT OR ATOM -->
   <xsl:template match="o[eo:abstract(.) and not(eo:has-data(.))]" mode="head">
-    <xsl:text>[</xsl:text>
-    <xsl:for-each select="o[eo:void(.)]">
-      <xsl:if test="position()&gt;1">
-        <xsl:text> </xsl:text>
-      </xsl:if>
-      <xsl:value-of select="@name"/>
-    </xsl:for-each>
-    <xsl:text>]</xsl:text>
+    <!-- A test attribute with no void params collapses its empty `[]`
+         head into the `++&gt;` suffix (see the tail template), so it emits
+         no head of its own. -->
+    <xsl:if test="not(eo:test-attr(.) and empty(o[eo:void(.)]))">
+      <xsl:text>[</xsl:text>
+      <xsl:for-each select="o[eo:void(.)]">
+        <xsl:if test="position()&gt;1">
+          <xsl:text> </xsl:text>
+        </xsl:if>
+        <xsl:value-of select="@name"/>
+      </xsl:for-each>
+      <xsl:text>]</xsl:text>
+    </xsl:if>
   </xsl:template>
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (35) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1934](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5569, opened 2026-07-16
fix(#5564): drop redundant `^.` parent prefixes on auto-resolvable names

### LINKED ISSUE #5564: `Xmir.toEO` function unnecessarily adds `^.` parent prefixes to auto-resolvable names
`Xmir.toEO()` prints an explicit `^.` (parent) prefix on every reference that resolves through the enclosing object, and the new `format` goal spreads these across `eo-runtime`: the reformatted sources now contain 446 `^.` occurrences, e.g. `tuple/last-index-of.eo:16` `if. (tup.head.eq ^.v13_2) tup.tail.length (^.reclast tup.tail)` and `bytes/array.eo:20` `tup.with (^.v13_2.slice index 1)`. Most of these were plain names in the source (`reclast`, `separator`, …) that the parser resolves to the parent automatically.

The explicit `^.` is redundant in these cases and only adds noise: scope resolution re-derives the parent hop, so `^.reclast` and `reclast` produce the same XMIR. Emitting it everywhere makes the canonical output more verbose than hand-written EO and obscures the few places where a `^.` is actually meaningful.

The `^` comes from the surface renderer in `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl` (`eo:translate-path` maps `ρ` to `^`, `eo:surface` around line 27). The printer should drop a single leading `^.` hop by default when the referenced name resolves to a parent attribute without ambiguity (i.e. no closer-scope binding shadows it), keeping `^.` only where omitting it would change resolution.

yegor256:
Closes #5564.

## Problem

`Xmir.toEO()` printed an explicit `^.` (parent) prefix on every reference that resolves through the enclosing object. The new `format` goal spread these across `eo-runtime` (446 `^.` occurrences), e.g. `^.reclast`, `^.v13_2`, `^.separator`. Most were plain names in the source that the parser resolves to the parent automatically, so the explicit `^.` was pure noise: `^.reclast` and `reclast` produce identical XMIR.

## Fix

The `^` is emitted by the surface renderer in `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl`. After `build-fqns` + `roll-bases`, such a reference has a rolled base of the form `ξ.ρ.name…` (a single parent hop onto a name from the enclosing scope).

The printer now drops that single leading `^.` hop when:
- the referenced name is **absent** from the immediate enclosing formation's scope, **and**
- the name **is present** in the parent formation.

Those two conditions guarantee that plain scope resolution re-derives exactly the same one-hop `ξ.ρ.name`, so the output stays semantically identical and round-trips cleanly. The hop is **kept** whenever a closer-scope binding would shadow the name (which would change resolution), and multi-hop (`^.^.x`), `^.@` (`ρ.φ`), and program-rooted (`Q.`) forms are all left untouched.

## Changes

- `to-eo-tree.xsl`: new `xsl:when` in the `BASED` head template that recognizes a redundant single `^.` hop and prints the bare name instead.
- `print-packs/yaml/redundant-parent.yaml`: new test covering both the drop case and the shadowed keep-`^.` case.
- `print-packs/yaml/test-attribute.yaml`, `print-packs/yaml/inheritance.yaml`: expected output updated to the cleaner form (`x.eq x`, `x.write v`, `base` instead of the `^.`-prefixed variants).

## Verification

All 77 `eo-printer` tests pass, including the round-trip `printsToParseableEo` check that re-parses and re-prints every fixture.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MpjDLen39dn19dCrdCcZ3R)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -123,6 +123,11 @@
   </xsl:template>
   <!-- BASED -->
   <xsl:template match="o[@base and not(eo:has-data(.))]" mode="head">
+    <!-- A base of the form "ξ.ρ.name…" is a single parent hop onto a
+         name that lives in the enclosing scope. -->
+    <xsl:variable name="rho-prefix" select="concat($eo:xi, '.', $eo:rho, '.')"/>
+    <xsl:variable name="rest" select="substring-after(@base, $rho-prefix)"/>
+    <xsl:variable name="first" select="if (contains($rest, '.')) then substring-before($rest, '.') else $rest"/>
     <xsl:choose>
       <!-- NOT OPTIMIZED TUPLE -->
       <xsl:when test="@star">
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (37) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2074](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5571, opened 2026-07-16
Print single-@ formations in compact inline-phi form

### LINKED ISSUE #5565: `Xmir.toEO()` fails to render single-attribute formations in compact inline-phi form
`Xmir.toEO()` always renders a formation vertically, even when its only attribute is the `φ` decoratee. So a one-attribute formation prints as two lines:

```
[] > foo
  a > @
```

when the language already offers the compact inline-phi form (R-3.10.8, §4.5):

```
a > [] > foo
```

The `format` goal produces the verbose shape throughout `eo-runtime` — for example `io/dead-input.eo` prints `[size] > read` with a lone `^.^.input-block > @` under it, and `nop.eo` prints `[x] > nop` with a lone `true > @`, both of which could collapse to a single line (`^.^.input-block > [size] > read`, `true > [x] > nop`).

The cause is in `eo-printer/src/main/java/org/eolang/printer/Pretty.java:134-137`: `horizontal(...)` returns empty whenever `node.abstractt`, so a formation is never laid out on one line. The fix is to add an inline-phi rendering: when an abstract node has exactly one child and that child is the `φ` binding (its tail is `> @`), emit `<phi-expr> > [params] > name` instead of putting the `@` on its own indented line. This applies only when `@` is the sole attribute; formations with any other binding keep the current vertical layout, and the usual penalty/width check still decides whether the phi expression itself fits inline.

yegor256:
Closes #5565.

## Problem

`Xmir.toEO()` always rendered a formation vertically, even when its only
attribute was the `φ` decoratee, so a one-attribute formation printed as two
lines:

```
[] > foo
  a > @
```

when EO already offers the compact inline-phi form (R-3.10.8 §4.5):

```
a > [] > foo
```

The `format` goal produced the verbose shape throughout `eo-runtime` (e.g.
`io/dead-input.eo` and `nop.eo`), where formations like `[size] > read` / `^.^.input-block > @`
and `[x] > nop` / `true > @` could collapse onto a single line.

The cause was in `eo-printer/.../Pretty.java`: `horizontal(...)` returned empty
whenever the node was abstract, so a formation was never laid out on one line.

## Fix

Added an inline-phi rendering in `Pretty`. When an abstract node has exactly one
child and that child is the `φ` binding (its tail is `> @`), the decoratee is
inlined through the existing `flat(...)` path and emitted as
`<phi> > [params] > name` instead of putting `@` on its own indented line.

- Applies **only** when `@` is the sole attribute; formations with any other
  binding keep the current vertical layout.
- A decoratee that cannot be inlined (a nested formation, a tuple) yields empty
  and falls back to the vertical shape.
- The usual penalty/width check in `layout(...)` still decides whether the
  inline line actually fits (a too-wide decoratee stays vertical).

## Tests

- Updated the existing print packs that contained single-@ formations to the
  collapsed shape. All of them still round-trip through the
  `printsToParseableEo` test (printed EO re-parses to the very same EO).
- Added `inline-phi.yaml` (the `a > [] > foo`, `^.^.input-block > [size] > read`,
  `true > [x] > nop` examples from the issue) and `inline-phi-wide.yaml` (a
  decoratee long enough that the inline line overflows and the vertical layout
  is kept).
- `mvn test` on `eo-printer` is green (79 tests) and `qulice` passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_011uTrXfYaWEeb6nfNwnRQmU)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -133,7 +133,9 @@ final class Pretty {
      */
     private Optional<String> horizontal(final Pretty.Node node, final int indent) {
         final Optional<String> result;
-        if (node.abstractt || node.children.isEmpty()) {
+        if (node.abstractt) {
+            result = this.phi(node, indent);
+        } else if (node.children.isEmpty()) {
             result = Optional.empty();
         } else {
             result = Pretty.inlined(node.children).map(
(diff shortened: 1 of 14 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (143 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2082](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5572, opened 2026-07-16
fix(#5563): keep `&gt;&gt;` file-local handles so the printer prints readable names

### LINKED ISSUE #5563: `Xmir.toEO()` isn't preserving file-local handle names, causing less readability
A void declared with a file-local handle, `? >> name` (R-3.10.12), loses its readable name when round-tripped through `Xmir.toEO()`. Since the #5548 fix, `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:37-41` rewrites the surviving cactus name `a🌵L-P` to a synthetic identifier `vL_P` via `eo:printable`. So `tuple/last-index-of.eo`, whose source declares `? >> list` and `? >> wanted`, now prints as `[v12_2 v13_2] > last-index-of` with references like `reclast v12_2` and `^.v13_2` — the meaningful handles `list` and `wanted` are gone (same in `bytes/xor.eo`, `bytes/array.eo`, and others).

This is wrong: the handle name is the whole point of `>> name`, and a `vL_P` placeholder makes the formatted code less readable than the input. The synthetic name is only necessary because the real handle is unavailable — `eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl` records the handle as `@local` but then drops it (`<xsl:template match="o/@local"/>`), so it never reaches XMIR.

The fix is to preserve the handle instead of discarding it: keep the `@local` marker on the void's `<o>` through parsing, and have the printer emit that name (falling back to `eo:printable` only when no handle exists). The maintainer should decide whether the canonical surface stays `? >> list` or becomes the bracket form `[list]`; either way the readable name, not `vL_P`, must be used.

yegor256:
Fixes #5563.

## Problem

A void declared with a file-local handle (`? >> name`, R-3.10.12) lost its readable name when round-tripped through `Xmir.toEO()`. Since the #5548 fix, the surviving cactus `@name` (e.g. `a🌵12-2`) was rewritten to a synthetic `v12_2` identifier via `eo:printable`, so `tuple/last-index-of.eo` (source `? >> list` / `? >> wanted`) printed as:

```
[v12_2 v13_2] > last-index-of
  reclast v12_2 > @
  ...
    tup.head.eq ^.v13_2
```

The meaningful handles `list` and `wanted` were gone (same for `bytes/xor.eo`, `bytes/array.eo`, and others). The synthetic name was only needed because `resolve-local-names` recorded the handle as `@local` and then **dropped it**, so the real handle never reached XMIR.

## Fix

Preserve the handle instead of discarding it, and teach the printer to use it.

- **`resolve-local-names.xsl`** no longer drops the `@local` marker — it stays on the declaring object. References are still rewritten to the reserved cactus `@name`, exactly as before, and the duplicate-handle error is unchanged.
- **`XMIR.xsd`** documents the now-persistent optional `@local` attribute on `o`.
- **`restore-local-names.xsl`** (new printer pass, run before `inline-cactoos`) inverts the parser rewrite for voids: the void's `@name` becomes its handle and every `@base` segment pointing at that cactus name is rewritten back to the handle. The `eo:printable` fallback in `to-eo-tree.xsl` still covers handle-less cactus names (bare `? >>`).

Result — the same file now prints:

```
[list wanted] > last-index-of
  reclast list > @
  ...
    if. (tup.head.eq ^.wanted) tup.tail.length (^.reclast tup.tail)
```

This is the bracket form R-3.4.7 already prescribes ("reverse printing canonicalises every void to the bracket form").

## Verification

- All 138 `eo-runtime/src/main/eo` files stay reparseable and idempotent through `Xmir.toEO()`, with no `vN_M` placeholders remaining (28 of them use `? >>` handles).
- Running the `org.eolang:lints` linter over parsed XMIR carrying `@local` reports no new defects; strict `XMIR.xsd` validation passes.
- Full `eo-parser` (1466) and `eo-printer` (77) test suites pass.

## Changes

- `resolve-local-names.xsl`: keep `@local` on the declaring object.
- `XMIR.xsd`: declare the optional `@local` attribute.
- `Xmir.java` + `restore-local-names.xsl`: printer-side inverse pass.
- `PARSER_SPEC.md` R-9.2.3 and `Emit#local` javadoc: document that the marker is kept.
- Updated parse-pack / eo-syntax / print-pack tests; added printer regression pack `void-local-handle.yaml`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ToT9SZ33La4QLpUhwhLxyG)_

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -1199,7 +1199,7 @@ R-9.2.2. The cactus 🌵 is reserved — it is excluded from the `NAME` token (�
 
 Example: a `>>` suffix at `line=12, pos=5` emits `@name="a🌵12-5"`.
 
-R-9.2.3. **File-local handles (R-3.10.12).** A `>> name` suffix emits the object with its cactus `@name` **and** a `@local="name"` marker; references stay as plain `<o base='name'>`. The first-pass `resolve-local-names` reshape (right after `wrap-applications` / `resolve-self`, before `build-fqns`) collects the per-file `@local → @name` table, rewrites every `@base` equal to a handle into the matching cactus `@name`, and drops the `@local` markers; a handle declared twice is reported there as a `resolve-local-names` check error. Downstream passes see only ordinary references by the reserved cactus name.
+R-9.2.3. **File-local handles (R-3.10.12).** A `>> name` suffix emits the object with its cactus `@name` **and** a `@local="name"` marker; references stay as plain `<o base='name'>`. The first-pass `resolve-local-names` reshape (right after `wrap-applications` / `resolve-self`, before `build-fqns`) collects the per-file `@local → @name` table and rewrites every `@base` equal to a handle into the matching cactus `@name`; a handle declared twice is reported there as a `resolve-local-names` check error. The `@local` marker is **kept** on the declaring object so that the readable handle can be recovered from the otherwise-synthetic cactus name — in particular by the printer, which prints `? >> name` voids back under their handle rather than a `vL_P` placeholder (#5563). Downstream compilation passes reference the reserved cactus name and ignore the marker.
 
 ### 9.3 Source-token to XMIR-character mapping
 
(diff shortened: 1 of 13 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (125 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2078](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5573, opened 2026-07-16
Gate PhSafe wrapper emission behind eo.trackLocations flag

### LINKED ISSUE #5435: PhSafe decorator unnecessarily wraps every dispatched object without ability to disable
The transpiler wraps every located object in a `PhSafe` decorator. The `mode="located"` template in `eo-maven-plugin/.../transpile/to-java.xsl:607` fires for any object that carries a source position, so essentially every dispatch in a program is built as:

```
name = new PhSafe(name, "<program>", <line>, <pos>, "<loc>", "<original-name>");
```

That means each `take`/`delta`/`put`/`lambda` goes through an extra wrapper allocation and a `try/catch` in `PhSafe.through()` (`eo-runtime/src/main/java/org/eolang/PhSafe.java`).

Since #5261 removed the catchable `error`/`try` and made `PhSafe` re-throw an uncatchable `ExFailure`, this wrapper no longer recovers anything — its only remaining job is to attach the EO source location (`Error in "<loc>" at <program>:<line>:<pos>`) to a failure as it unwinds the object graph. In other words it is debug information — the equivalent of a line-number table — and right now it is always on, on the hottest path in the runtime, with no way to strip it.

The location genuinely cannot be recovered by a single top-level catch, because EO executes as a lazy object graph rather than a statement stack: the Java call stack at the moment of failure does not correspond to the EO source, and a `⊥` is usually forced far from where it was produced. So the location has to be threaded through the graph, which is what `PhSafe` does — but that is a development-time need, not a production one.

Make `PhSafe` emission a transpiler flag, the way native toolchains treat debug symbols: a "dev" build emits the wrappers and keeps precise `.eo` locations in panics, while a "release" build omits them, so a failure still carries its reason (from the `ExFailure`) plus the Java stack trace but skips the per-dispatch overhead. Lighter alternatives worth weighing are wrapping only at coarser boundaries — atom entry or named bindings rather than every node — or emitting JSR-45 source maps, though the latter only partially helps since the Java call stack does not match EO's graph structure. The smallest viable first step is a single flag guarding the `mode="located"` template, defaulting to on so current behaviour is unchanged.

maxonfjvipon:
The transpiler wrapped every located object in a `PhSafe` decorator, so essentially every dispatch in a program carried an extra wrapper allocation and a `try/catch` on the hottest runtime path. Since #5261 removed the catchable `error`/`try`, that wrapper no longer recovers anything — its only remaining job is to attach the `.eo` source location to a failure as it unwinds. That is debug information, and it was always on with no way to strip it.

This adds an `eo.trackLocations` transpiler flag, the way native toolchains treat debug symbols. It is **off by default**, so a production build stays lean — a failure still carries its reason (from `ExFailure`) plus the Java stack trace, just not the `.eo` position. The flag threads down to a single guard on the `mode="located"` template:

```xml
<xsl:if test="$trackLocations='true' and @line and @pos and not(contains(@loc, '+'))">
  ... new PhSafe( ...
```

`eo-runtime` turns it back on in its own `pom.xml`, because its tests assert on precise `.eo` locations in panics:

```xml
<trackLocations>true</trackLocations>
```

Note this flips the previous behaviour: a normal transpile no longer emits `PhSafe`. Any project that wants locations enables the flag (property `-Deo.trackLocations=true` or plugin config), as `eo-runtime` now does.

The two transpiler diagnostic toggles (transformation steps and locations) are grouped into a small `Tracking` value object to keep the `Transpiling` constructor within its parameter budget. Coarser-grained wrapping (atom or named-binding boundaries) and JSR-45 source maps remain open follow-ups.

Closes #5435


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (194 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+820](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5574, opened 2026-07-16
#5568: print Unicode string glyphs instead of escaping to \uNNNN

### LINKED ISSUE #5568: `Xmir.toEO` incorrectly escapes readable non-ASCII characters to `\uNNNN` format
`Xmir.toEO()` escapes every non-ASCII character in a string literal to `\uNNNN`, so a readable string is rendered unreadable. For example `"你好, друг!"` prints as `"你好, друг!"`, and `"hello 你好"` prints as `"hello 你好"` — the ASCII survives but every letter above U+007F is escaped.

The cause is `eo-printer/src/main/java/org/eolang/printer/StUnhex.java:62`, which renders the decoded UTF-8 string with `StringEscapeUtils.escapeJava(...)`. That Apache Commons Text helper is meant for Java source literals and escapes the entire non-ASCII range to `\uNNNN`; it is the wrong tool for EO, whose sources are UTF-8 and can carry the glyphs directly. The result is both harder to read and needlessly long.

The fix is to escape only what an EO string literal actually requires — the backslash, the double quote, and non-printable/control characters — and to leave printable Unicode intact, falling back to `\uNNNN` only for characters that cannot be rendered. Concretely, replace the `escapeJava` call with a small custom escaper (or a Commons Text `AggregateTranslator` whose unicode-escape stage is restricted to control/non-printable code points) so printable code points pass through verbatim.

gemshrine:
Closes #5568.

`StUnhex` rendered every decoded string literal with `StringEscapeUtils.escapeJava`, which escapes the entire non-ASCII range to `\uNNNN`. A string like `"друг 你好"` printed as `"\u0434\u0440\u0443\u0433 \u4F60\u597D"` — readable EO turned unreadable and needlessly long.

This replaces `escapeJava` with a small custom escaper that escapes only what an EO string literal actually needs — the backslash, the double quote, and control characters (as `\n`, `\t`, `\r`, `\b`, `\f`, or a `\uXXXX` fallback for other ISO control code points) — and leaves printable Unicode intact, since EO sources are UTF-8. The escaping is the exact inverse of the parser's `Emissions.unescapeBody`, so the round-trip is preserved.

- `StUnhex.java`: new `escape`/`escaped` helpers; `commons-text` no longer used.
- `StUnhexTest`: two new cases — Unicode glyphs are kept readable, and non-printable control chars still fall back to `\uXXXX`.
- `emoji.yaml`, `idiomatic.yaml` print packs: `printed` expectations updated to glyphs.
- `eo-printer/pom.xml`: dropped the now-unused `commons-text` dependency.

### DIFF
--- a/eo-printer/pom.xml
+++ b/eo-printer/pom.xml
@@ -47,11 +47,6 @@
       <!-- version from parent POM -->
     </dependency>
     <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-text</artifactId>
-      <!-- version from parent POM -->
-    </dependency>
-    <dependency>
       <groupId>org.cactoos</groupId>
       <artifactId>cactoos</artifactId>
       <!-- version from parent POM -->
--- a/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
+++ b/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
@@ -11,7 +11,6 @@ import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.regex.Pattern;
-import org.apache.commons.text.StringEscapeUtils;
 import org.eolang.parser.StXnav;
 
 /**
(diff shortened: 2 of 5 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+323](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5576, opened 2026-07-16
fix(#5570): print short +alias form when name matches FQN's last segment

### LINKED ISSUE #5570: `Xmir.toEO()` incorrectly expands `+alias` metas to verbose form, degrading readability
`Xmir.toEO()` prints every `+alias` meta in the verbose two-argument form even when the source used the short one. All 43 aliases in `eo-runtime` are written short — `+alias sm.os`, `+alias tuple.reduced`, `+alias string.regex` — but after the `format` goal they become `+alias os Q.sm.os`, `+alias reduced Q.tuple.reduced`, and so on (see the reformatted `path.eo`). The short form is the idiomatic one and the only one used in the codebase, so expanding it is a regression in readability.

The parser normalizes `+alias sm.os` into the XMIR tail `os Φ.sm.os` (two `<part>` children: the local name `os` and the FQN `Φ.sm.os`). The printer's meta template in `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl` (the `match="meta"` rule) then emits `+` + head + the whole tail with `Φ` replaced by `Q`, yielding the long form.

The fix is to special-case `alias` metas in that template: when the alias name equals the last dotted segment of the FQN, print only the FQN with its `Φ.`/`Q.` root prefix stripped (`+alias sm.os`), dropping the redundant leading name. Keep the two-argument form only when the name genuinely differs from the FQN's last segment (a real rename).

yegor256:
Closes #5570.

## Problem

`Xmir.toEO()` printed every `+alias` meta in the verbose two-argument form even when the source used the short one. All 43 aliases in `eo-runtime` are written short (`+alias sm.os`, `+alias tuple.reduced`, ...), but after the `format` goal they became `+alias os Q.sm.os`, `+alias reduced Q.tuple.reduced`, etc. — a readability regression, since the short form is the idiomatic and only form used in the codebase.

The parser normalizes `+alias sm.os` into the XMIR meta tail `os Φ.sm.os` (two `<part>` children: the local name `os` and the FQN `Φ.sm.os`). The old meta template emitted `+` + head + the whole tail with `Φ` replaced by `Q`, yielding the long form.

## Fix

In `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl`, the `match="meta"` rule now special-cases `alias` metas: when the alias name (`part[1]`) equals the last dotted segment of the FQN (`part[last()]`), it prints only the FQN with its `Φ.` root prefix stripped (`+alias sm.os`), dropping the redundant leading name. The two-argument form is kept only for a genuine rename, where the name differs from the FQN's last segment (`+alias custom Q.org.eolang.string.regex`). Non-alias metas are unchanged.

## Tests

- Updated `inheritance.yaml` expected output to the short form (`+alias org.eolang.io.stdout`, `+alias org.eolang.string.sprintf`).
- Added `aliases.yaml` print pack covering the short form (`+alias sm.os`), a composite short form (`+alias org.eolang.io.stdout`), and a genuine rename (`+alias custom org.eolang.string.regex`), verifying both the printed EO and the parse round-trip.

Note: the one remaining `XmirTest` failure on `void-local-handle.yaml` (reversed-dispatch `^.wanted`) is pre-existing on the branch base and unrelated to this change — that pack contains no aliases.

---
_Generated by [Claude Code](https://claude.ai/code/session_011u11VhnFhVDZuuRs6kAqJL)_


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2090](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5580, opened 2026-07-16
#5556: resolve a named binding `:name` against a `? >> name` void handle

### LINKED ISSUE #5556: Named binding `obj value:name` cannot fill a `? >> name` void
A void declared with the `>>` handle from #5529 — `? >> age` — gives the void an auto-generated cactus external name (`a🌵N-M`, see `eo-parser/src/main/java/org/eolang/parser/AutoName.java:37` and `LnVoid.java:92-98`) and keeps `age` only as a file-local `@local` handle. So `human 42` fills the void positionally, and `h.age` is (intentionally) unreachable. But a named binding at the call site — `human 42:age` — does not work: the argument is emitted as `@as='age'` verbatim (`eo-parser/src/main/java/org/eolang/parser/Emissions.java:220-228`) and reaches `eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl:674-675` as `Bind("age", …)` against the literal name `age`, while the void's real attribute is the cactus name — so the bind never matches.

The cause is that `eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl:16-22` rewrites the handle only in `@base` dispatch references (`right > @` becomes `a🌵3-2`), never in `@as` binding targets. So even inside the same `.eo` file, where the handle `age` is visible, the named binding is left dangling. No test covers a `>>` void combined with a `:name` binding — `eo-packs/parse/void-local-handle.yaml` asserts only the `@base` rewrite — so the gap went unnoticed. This is inconsistent: `>>` gives the void a name, yet that name cannot be used to bind an argument to it.

Fix: in `resolve-local-names.xsl`, add a template that rewrites an `@as` equal to a handle into the same cactus name it already computes for `@base` (lines 16-22), so same-file `human 42:age` resolves to the void with a minimal, mechanical change. Whether the handle should also be bindable by name from other files is a separate design decision, since the handle is file-local by design, and need not be settled here.

gemshrine:
Closes #5556.

A void declared with the `>>` handle (`? >> age`) gets an auto-generated cactus attribute name, keeping `age` only as a file-local handle. `resolve-local-names.xsl` rewrites every `@base` dispatch reference equal to a handle into that cactus name — but it never touched `@as` binding targets. So a same-file named binding `human 42:age` was emitted as `@as='age'` and reached transpilation as `Bind("age", …)` against the literal name, while the void's real attribute is the cactus name — the bind never matched, even though the handle is visible in the same file.

This extends the template to rewrite `@as` the same way it already rewrites `@base`. The two are merged into one template so an object that carries both a handle `@base` and a handle `@as` is handled unambiguously; non-handle `@base`/`@as` values are copied through unchanged.

- `resolve-local-names.xsl`: rewrite `@as` handles to the cactus name.
- `void-local-handle-named-binding.yaml`: new parse pack — `human 42:age` against `? >> age` resolves to the void's cactus name and leaves no literal `@as='age'`.

Verified against the real parser output (the assertion fails on `master`, passes with the fix); full `eo-parser` suite is green. Scope is same-file resolution, as the issue notes; cross-file handle binding remains a separate design question.

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
@@ -8,7 +8,9 @@
   The "&gt;&gt; foo" file-local handle (§3.10 / §9.2): the parser emits the
   anonymous object with its cactus @name plus a "@local='foo'" marker. We
   build the per-file "handle -&gt; cactus-name" table and rewrite every @base
-  equal to a handle into that (reserved) cactus name; a handle declared more
+  (a dispatch, "foo &gt; @") and every @as (a named binding, "obj 42:foo")
+  equal to a handle into that (reserved) cactus name, so that a same-file
+  named binding actually fills the handle's void; a handle declared more
   than once is an error. The "@local" marker is deliberately kept on the
   declaring object so that later passes (in particular the printer, see
   #5563) can recover the readable handle from the otherwise-synthetic cactus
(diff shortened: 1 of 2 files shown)


yegor256:
@gemshrine the #5556 was never a bug, but a question to discuss

gemshrine:
Closing this: #5556 is labeled `question` and the direction isn't settled yet. @maxonfjvipon raises a fair alternative — that void attributes should be excluded from `>> name` handles entirely (so `human 42:age` wouldn't be valid), rather than making the named binding resolve. I don't want to pre-empt that decision. The branch `gemshrine:5556-named-binding-void-handle` has a working implementation of the *resolve-the-binding* option (parser rewrite + parse pack, CI green) if that's the direction you choose.

PR #5582, opened 2026-07-16
fix(#5567): support `++>` test shorthand in inline-phi position

### LINKED ISSUE #5567: `++>` shorthand fails to parse in inline-phi position in test attributes
A test attribute whose only attribute is the `φ` decoratee currently has no one-line spelling. The inline-phi test form `foo > [] +> test-works` (R-3.10.8) parses and is equivalent to the vertical `[] +> test-works` / `foo > @`, but the natural compact spelling that combines it with the `++>` shorthand (#5445) is rejected: `foo ++> test-works` fails to parse with `expected value`. The goal is to support

```
foo ++> test-works
```

as sugar for

```
[] +> test-works
  foo > @
```

On the parser side, `++>` was introduced in #5445 only at the head of a body line (`++> name` → `[] +> name`, PARSER_SPEC §3.1 classifier lookahead, R-6.3.6). It should also be accepted in the inline-phi suffix position, so `<phi> ++> name` desugars to the already-supported `<phi> > [] +> name` — the inline-phi variant table in R-3.10.8 lists `> [params] +> name` but no `++>`-collapsed counterpart. The XMIR emission stays identical to the expanded form, so nothing downstream changes.

On the printer side, `Xmir.toEO()` should emit this form for a test attribute whose sole attribute is `@`, i.e. print `foo ++> test-works` rather than the two-line vertical shape. This is the test-attribute case of the compact inline-phi rendering in #5565 combined with the `++>` shorthand of #5562, and only applies when `@` is the only binding.

yegor256:
Closes #5567.

A test attribute whose only binding is the `φ` decoratee had no one-line spelling combining the inline-phi test form with the `++>` shorthand (#5445). `foo ++> test-works` was rejected with `expected value`. This accepts it as sugar for:

```
[] +> test-works
  foo > @
```

### Parser
- `Eo.topLevelPlusPlusArrowIndex` finds a space-preceded `++>` marker at paren depth 0 and outside strings, sharing the scan skeleton with the `> [` marker via an `IntPredicate` (no duplication). The space requirement keeps it distinct from the head-of-line `++> name` bare-formation shorthand (R-6.3.6).
- `Eo.onlyPhi` routes such lines to `LnOnlyPhi`, which now handles both the `> [` and `++>` splits. XMIR emission is identical to the expanded `> [] +> name` form, so nothing downstream changes.

### Printer
- Extends `Pretty.phi` so a sole-`φ` test attribute with **no void params** collapses to `<phi> ++> name`. The inline-phi collapse for `[params]`-headed formations (#5565) already covers the void-bearing case (`a > [a] +> keeps-voids`); this adds the empty-head `++>` shorthand it explicitly punted on. The penalty/width check still falls back to the vertical shape when the line is too wide.

### Spec
- R-3.10.8 and R-6.3.6 in `PARSER_SPEC.md` document the compact `++>` inline-phi form.

### Tests
- `eo-syntax/test-shorthand-inline-phi.yaml` — valid inline `++>` forms (identifier LHS, reversed-dispatch LHS with a harg).
- `print-packs/test-attribute.yaml` — updated to the collapsed `x.eq x ++> tests-something` output.

Rebased onto current `master`; full `eo-parser` and `eo-printer` suites pass locally, and qulice is clean on both modules.


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (192 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2094](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5586, opened 2026-07-16
fix(#5581): print `? >> name` voids under their handle, not as `[name]`

### LINKED ISSUE #5581: Xmir.toEO() translates local handle '? >> name' voids incorrectly into bracket params '[name]', violating anonymity
`Xmir.toEO()` prints a void declared with a file-local handle, `? >> name` (R-3.10.12), as a bracket parameter `[name]` instead of restoring the vertical `? >> name` line. For example `bytes/as-bool.eo`, whose source was

```
[] > as-bool
  ? >> bts
  bts.eq 01- > @
```

now prints as `[bts] > as-bool`. Two passes cooperate to lose the form: `eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl` promotes the handle to the visible `@name` and then drops the `@local` marker, and `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:161-166` lists every void child in the `[...]` head. By the time the head is rendered nothing distinguishes a former `? >> name` handle from a plain `[name]` param.

This is wrong because the `>>` form exists to keep the void anonymous: its external name stays a hidden cactus name (§9.2) and never enters the visible namespace, while `name` is only a file-local handle (R-3.10.12). Collapsing it to `[name]` promotes the void into a public positional attribute, so the printed text reparses to different XMIR (a plain `name` void with no cactus name and no `@local`) — a lossy round-trip that changes the object's public surface. R-9.2.3 says the `@local` marker is kept "in particular by the printer, which prints `? >> name` voids back under their handle". The current output contradicts that. This was left as an open choice in #5563 ("stays `? >> list` or becomes the bracket form `[list]`"); the bracket resolution shipped, but it drops the anonymity the author asked for.

The fix is to keep the `@local` marker through `restore-local-names.xsl` rather than promoting the handle to `@name` and discarding the marker, then in the formation-head template of `to-eo-tree.xsl` exclude `@local`-bearing voids from the bracket loop and emit each as a `? >> <local>` body line (references already resolve to the handle since #5563). Plain voids — `? > name` and `[name]` with no `@local` — stay in the brackets as today, per R-3.4.7.

yegor256:
Closes #5581.

## Problem

`Xmir.toEO()` printed a void declared with a file-local handle — `? >> name` (R-3.10.12) — as a bracket parameter `[name]` instead of restoring the vertical `? >> name` line. For example `bytes/as-bool.eo`:

```
[] > as-bool
  ? >> bts
  bts.eq 01- > @
```

printed as `bts.eq 01- > [bts] > as-bool`.

This is wrong because the `>>` form exists to keep the void **anonymous**: its external name stays a hidden cactus name (§9.2) while `name` is only a file-local handle. Collapsing it to `[name]` promotes the void into a public positional attribute, so the printed text reparses to different XMIR — a lossy round-trip that changes the object's public surface. R-9.2.3 requires the `@local` marker be kept "in particular by the printer, which prints `? >> name` voids back under their handle".

## Fix

- **`restore-local-names.xsl`** — keep the `@local` marker on voids (it was being dropped after promoting the handle to `@name`). The marker is now dropped only on non-void formations, whose handle the `inline-cactoos` pass already inlines away. References still resolve to the handle as before (#5563).
- **`to-eo-tree.xsl`** — exclude `@local`-bearing voids from the `[...]` head loop and include them among the body children, emitting each as a vertical `? >> name` line.

Plain voids — `? > name` and `[name]` with no `@local` — still render inside the brackets, per R-3.4.7.

## Tests

Updated the two printer fixtures that captured the old buggy output (`cactus-void.yaml`, `void-local-handle.yaml`) to the correct round-trippable form. The full `eo-printer` suite (86 tests, including the print-then-reparse idempotency check) passes.

https://claude.ai/code/session_013xoXnMRPJZcquYE1ubj9oU

---
_Generated by [Claude Code](https://claude.ai/code/session_013xoXnMRPJZcquYE1ubj9oU)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
@@ -11,10 +11,14 @@
   "@local='name'" marker; the parser resolves references to the cactus name
   but the readable handle is preserved on the void. Here we put the handle
   back: the void's @name becomes its handle and every @base segment that
-  points at that cactus name is rewritten to the handle, so the printer
-  emits "[name]" and "name" instead of a synthetic "vL_P" placeholder. A
-  non-void "&gt;&gt;" handle (on an anonymous formation) is left to the
-  existing "inline-cactoos" pass, so its @local marker is simply dropped.
+  points at that cactus name is rewritten to the handle, so references read
+  under the handle instead of a synthetic "vL_P" placeholder. The
+  "@local" marker is deliberately KEPT on the void (#5581) so that
+  "to-eo-tree" can print it back as a vertical "? &gt;&gt; name" line and
+  preserve the void's anonymity (§9.2, R-9.2.3), rather than collapsing it
+  into a public "[name]" bracket param. A non-void "&gt;&gt;" handle (on an
+  anonymous formation) is left to the existing "inline-cactoos" pass, so its
+  @local marker is simply dropped.
   -->
   <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2170](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5587, opened 2026-07-16
fix(#5584): render data-receiver dispatch in suffix form

### LINKED ISSUE #5584: `Xmir.toEO()` always renders dispatch with data-receiver in reversed form instead of equivalent suffix form
`Xmir.toEO()` renders method dispatch whose receiver is a data literal (number, string, bytes) only in the reversed form, never the equivalent suffix form, and never weighs the two against each other. Verified round-trips:

```
5.plus 3       ->  plus. 5 3
-1.times 228   ->  times. -1 228
"x".eq "y"     ->  eq. "x" "y"
foo.plus 3     ->  foo.plus 3     (named receiver: suffix kept)
```

Both `5.plus 3` and `plus. 5 3` are valid EO for the same object, yet the printer always picks the reversed one when the receiver is data.

The reason is the XMIR shape rather than a deliberate layout choice. A named receiver folds into a dotted `@base` (`foo.plus`) and prints as suffix, but a data receiver is a child `<o>` under a `.method` link, and `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:29` renders any `@base` that starts with `.` as the reversed `method.` head. So the layout of a data-receiver dispatch is decided by representation, and the suffix form is never even offered to the penalty-based `Pretty` search — even though `5.plus 3` is usually shorter and easier to read than `plus. 5 3`.

The printer should treat the suffix and reversed renderings of a data-receiver dispatch as alternative candidates and keep the lower-penalty one, the way it already chooses between inline and vertical layouts. Concretely, let the head renderer place a lone data receiver back into suffix position (`<data>.method args`) and feed both shapes through the same penalty comparison, instead of hard-coding the reversed form for every `.`-based head.

yegor256:
Closes #5584.

## Problem

`Xmir.toEO()` rendered a method dispatch whose receiver is a data literal (number, string, bytes) **only** in the reversed form, never the equivalent suffix form:

```
5.plus 3      ->  plus. 5 3
-1.times 228  ->  times. -1 228
"x".eq "y"    ->  eq. "x" "y"
foo.plus 3    ->  foo.plus 3   (named receiver already kept as suffix)
```

The cause was representational, not a layout choice. A named receiver folds into a dotted `@base` (`foo.plus`) and prints as suffix, but a data receiver is a child under a reversed `.method` link, and `to-eo-tree.xsl` rendered any `@base` starting with `.` as the reversed `method.` head. The suffix form was therefore never offered to the penalty-based `Pretty` search — even though `5.plus 3` is usually shorter and easier to read than `plus. 5 3`.

## Fix

- `to-eo-tree.xsl` now marks every line-tree node with `data="yes"` / `data="no"` (via `eo:has-data`), so the layout engine can tell a lone data literal from a named reference.
- `Pretty` gains a `suffixed(...)` helper that rebuilds a reversed dispatch on a lone data receiver into its suffix shape (`5.plus 3` from `plus. 5 3`), gluing the receiver to the method and keeping the remaining arguments as children.
- `layout(...)` now lays out **both** shapes and keeps the lower-penalty one — the same penalty comparison already used to choose between inline and vertical layouts. The suffix shape is never longer, and its vertical form has one fewer child line, so it can only tie or win; on a tie the suffix form is kept as the shorter, more readable rendering. `flat(...)` applies the same rewrite so inlined arguments (`... (-1.times 228)`) render in suffix position too.

Compound receivers (`and. (true.eq 01-) (false.eq 00-)`) and named receivers (`foo.plus 3`) are unaffected — only a lone data receiver has a suffix form.

## Tests

- New pack `data-receiver-dispatch.yaml` covering number, string, and bytes receivers plus a named-receiver control.
- Updated `times.yaml` expectation from `times. -1 228` to `-1.times 228`.
- All `eo-printer` tests pass (`XmirTest` 65, `PenaltyTest`, `StUnhexTest`, `StEoLoggedTest`), including the `printsToParseableEo` round-trip check for every pack.

https://claude.ai/code/session_011wSYSmZLiNv3FpVdXEjRY9

---
_Generated by [Claude Code](https://claude.ai/code/session_011wSYSmZLiNv3FpVdXEjRY9)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -299,8 +299,10 @@ final class Pretty {
 
     /**
      * Render a node inline, as it would appear as an argument (without
-     * its own name suffix), or empty if it can't be inlined safely.
-     * @param node The node
+     * its own name suffix), or empty if it can't be inlined safely. A
+     * data-receiver dispatch is inlined in its suffix shape ({@code
+     * 5.plus 3}), never the reversed one.
+     * @param given The node
      * @return The inlined content, or empty
      */
     private static Optional<String> flat(final Pretty.Node given) {
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -89,11 +89,40 @@ final class Pretty {
     /**
      * Lay out a node as a (possibly multi-line) block at the given
      * indentation, picking the rendering with the lowest penalty.
+     *
+     * <p>When the node is a reversed dispatch on a lone data receiver
+     * ({@code plus. 5 3}), its equivalent suffix shape ({@code 5.plus 3})
+     * is laid out too and the lower-penalty one is kept — the same
+     * penalty comparison that already decides inline versus vertical. The
+     * suffix shape is never longer and its vertical form has one fewer
+     * child line, so it can only tie or win; a tie (both fit the width)
+     * resolves to the suffix, the shorter and more readable form.</p>
+     *
      * @param node The node
      * @param indent The indentation level
      * @return The rendered block
      */
     private String layout(final Pretty.Node node, final int indent) {
+        String best = this.shaped(node, indent);
+        final Optional<Pretty.Node> suffix = Pretty.suffixed(node);
+        if (suffix.isPresent()) {
+            final String alt = this.shaped(suffix.get(), indent);
+            if (new Penalty(alt, this.weights).points()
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (120 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2102](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5588, opened 2026-07-16
#5578: fold a fingerprint of transpile XSLs into the cache key

### LINKED ISSUE #5578: Stale Java served from global transpile cache keyed on SNAPSHOT version after transpiler logic modifications
The global transpile cache derives its path from only the plugin version and the source git hash. `CachePath` (`eo-maven-plugin/src/main/java/org/eolang/maven/CachePath.java`) joins `base / semver / hash / tail`, where `base` is `~/.eo/transpiled` (`MjSafe.java:208`, `Transpiling.java:103`) and `semver` is the plugin version (`Transpiling.java:142`). The transpiler's own transformation logic — the XSLs under `.../transpile/`, notably `to-java.xsl` — is not part of the key.

During development the version is the constant `1.0-SNAPSHOT` and the source hash is stable, so when an XSL changes without a version bump the cache under `~/.eo/transpiled/1.0-SNAPSHOT/<hash>/` keeps serving Java produced by an earlier build. After the `#5486` package rename switched `to-java.xsl` from `eo:clean` to `eo:clean-package` (line 242, emitting `EO_one` instead of `EOone`), a pre-migration cache entry is reused, and `MjTranspileIT.transpilesWithPackage` fails with "The EOfoo.java file must be generated, but it didn't" — it looks for `org/eolang/EO_one/EOfoo.java` while the stale cache yields `org/eolang/EOone/EOfoo.java`. Removing `~/.eo/transpiled/1.0-SNAPSHOT` makes the test pass, which is why this failure comes and goes depending on the local cache.

Fold a fingerprint of the transformation logic into the cache key — e.g. append a hash of the bundled transpile XSLs (or the plugin jar) to `semver` in `CachePath` — or skip the global transpile cache entirely when the version ends with `-SNAPSHOT`.

morphqdd:
Closes #5578.

The global transpile cache key was `base / semver / hash / tail`, where `semver` is only the plugin version. The transpiler's own XSL logic was not part of the key, so during development (constant `1.0-SNAPSHOT`, stable source hash) a change to `to-java.xsl` without a version bump left the cache serving Java from an earlier build.

This folds a fingerprint of the bundled transpile XSLs into the `semver` segment. `Transpiling.version()` now returns `<plugin-version>-<fingerprint>`, where the fingerprint is a short SHA-256 (first 12 hex chars) over the content of all nine transpile XSLs, in order. When any XSL changes, the fingerprint changes, so the cache path changes and the stale entry is no longer reused.

The XSL list is now a single `Transpiling.XSLS` constant used both to build the transpile train and to compute the fingerprint, so the two cannot drift apart (a future XSL added to the train is automatically included in the key).

The new `Fingerprint` class reads the given classpath resources into one digest and returns the short hex; it fails loudly on a missing resource instead of silently producing a wrong key.

Verification: `eo-maven-plugin` full suite with qulice, 373 tests, 0 failures. `FingerprintTest` covers determinism, sensitivity to the resource set and to order, and the missing-resource failure.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Fingerprint.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Fingerprint.java
@@ -1,0 +1,65 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.function.Supplier;
+
+/**
+ * Short hex fingerprint of a set of classpath resources.
+ *
+ * <p>It reads each resource, in the given order, into one SHA-256
+ * digest and returns the first twelve hex characters. It is used to
+ * fold the content of the bundled transpile XSLs into the transpile
+ * cache key, so that a change in the transformation logic invalidates
+ * the cache even when the plugin version does not change (see #5578).</p>
+ *
+ * @since 0.63
+ */
+final class Fingerprint implements Supplier<String> {
+
+    /**
+     * Classpath resource paths to hash, in order.
+     */
+    private final String[] resources;
+
+    /**
+     * Ctor.
+     * @param res Classpath resource paths to hash
+     */
+    Fingerprint(final String... res) {
+        this.resources = res.clone();
+    }
+
+    @Override
+    public String get() {
+        try {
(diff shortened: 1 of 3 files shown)


morphqdd:
@yegor256, please take a look when you have a moment. All 25 checks are green.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (200 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1258](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5590, opened 2026-07-16
#5589 render empty tuple as `*` shorthand

### LINKED ISSUE #5589: Printer renders empty tuple as `tuple.empty` instead of the `*` shorthand
`Xmir.toEO()` renders an empty tuple as the verbose `tuple.empty` object instead of the `*` star shorthand. The printer already lowers non-empty tuples to `*` (`eo-printer/src/main/resources/org/eolang/printer/print/tuples-to-stars.xsl`), so `* a b c` round-trips fine, but the zero-element case falls through to the bare `tuple.empty` base. Sources that were written with `*` come back as `tuple.empty` after the format goal — for example `set.eo` (`*` becomes `set tuple.empty`), `bytes/array.eo` (`slice-byte * 0` becomes `slice-byte tuple.empty 0`), and `socket.eo`/`clock.eo`/`map.eo`/`range.eo`/`path.eo`.

This is a readability regression: `*` is the idiomatic empty tuple and the form the sources use, while `tuple.empty` is longer and inconsistent with how the printer renders every non-empty tuple. Both are semantically the same object, so the change is purely cosmetic, but it makes the canonical output noisier than the input.

The fix is to extend the star-lowering so the empty tuple prints as `*` too: in `tuples-to-stars.xsl` (or wherever the tuple head is recognised), treat a `tuple.empty` / zero-element tuple as the star form with no elements, emitting `*` instead of the `tuple.empty` base.

yegor256:
Closes #5589.

## Problem

`Xmir.toEO()` lowered non-empty tuples to the `*` star form (via `eo-printer/src/main/resources/org/eolang/printer/print/tuples-to-stars.xsl`), so `* a b c` round-trips fine, but the zero-element case fell through to the bare `tuple.empty` base. Sources written with `*` came back as `tuple.empty` after the format goal (e.g. `set.eo`, `bytes/array.eo`, `socket.eo`, `clock.eo`, `map.eo`, `range.eo`, `path.eo`) — a purely cosmetic but noisy readability regression, since `*` is the idiomatic empty tuple.

## Fix

Extend `tuples-to-stars.xsl` with a template that matches the `Φ.tuple.empty` object and emits the star form (`@star`) with no elements, mirroring how non-empty tuples are already lowered. An empty tuple now prints as `*` instead of `tuple.empty`.

The inner-mode handling for the innermost `Φ.tuple.empty` seed of a non-empty tuple is unaffected — that path is mode-specific and still emits nothing, so non-empty tuples keep rendering as `* a b c`.

## Tests

- `eo-packs/print/empty-tuple-to-star.yaml` — XSL-level assertions that a `tuple.empty` object (standalone and as an argument) gains `@star` with zero elements.
- `print-packs/yaml/empty-tuples.yaml` — end-to-end round-trip through `toEO()` confirming `*` is rendered instead of `tuple.empty`.

All 70 `XmirTest` cases pass.

https://claude.ai/code/session_01BbN3RLaJ5dZ4YpQiZL3ssd

---
_Generated by [Claude Code](https://claude.ai/code/session_01BbN3RLaJ5dZ4YpQiZL3ssd)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/tuples-to-stars.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/tuples-to-stars.xsl
@@ -19,6 +19,17 @@
       <xsl:apply-templates select="$arg" mode="no-as"/>
     </xsl:copy>
   </xsl:template>
+  <!--
+    An empty tuple is stored as the bare "Φ.tuple.empty" base. Render it as
+    the "*" star shorthand with no elements, mirroring how non-empty tuples
+    are lowered to stars above.
+  -->
+  <xsl:template match="o[@base = 'Φ.tuple.empty']">
+    <xsl:copy>
+      <xsl:apply-templates select="@*"/>
+      <xsl:attribute name="star"/>
+    </xsl:copy>
+  </xsl:template>
   <xsl:template match="o" mode="inner">
     <xsl:if test="@base = 'Φ.tuple'">
       <xsl:variable name="arg">
--- a/eo-printer/src/test/resources/org/eolang/printer/eo-packs/print/empty-tuple-to-star.yaml
+++ b/eo-printer/src/test/resources/org/eolang/printer/eo-packs/print/empty-tuple-to-star.yaml
@@ -1,0 +1,28 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets:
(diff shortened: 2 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2178](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5592, opened 2026-07-16
#5591 do not wrap single data-receiver dispatch argument in parentheses

### LINKED ISSUE #5591: Printer wraps a single dispatch token in redundant parentheses when inlining reversed-dispatch arguments
`Xmir.toEO()` wraps a single dispatch token in redundant parentheses when it inlines the arguments of a reversed dispatch. The offending code is `inlined()` in `eo-printer/src/main/java/org/eolang/printer/Pretty.java:291-295`, which parenthesizes every argument whose node has children:

```java
if (arg.children.isEmpty()) {
    joined.append(flat.get());
} else {
    joined.append('(').append(flat.get()).append(')');
}
```

A data-receiver dispatch such as `01-.as-bool` has a receiver child (`01-`) but takes no arguments of its own, so it is a single token — yet the `!children.isEmpty()` guard wraps it as `(01-.as-bool)`.

This produces EO that no longer parses. Re-formatting `eo-runtime/src/main/eo/bytes/as-bool.eo` turns the valid

```
    not. > @
      eq.
        01-.as-bool
        00-.as-bool
```

into

```
    not. > @
      eq. (01-.as-bool) (00-.as-bool)
```

which fails with `redundant parentheses around a single token — (01-.as-bool) should be written as 01-.as-bool`, and that in turn desynchronizes the enclosing `not.` into `reversed dispatch missing receiver`. Because the output no longer parses, the `format` goal breaks the build. The same over-parenthesization hits `number/is-nan.eo`, `number/is-integer.eo`, `recovered.eo`, `bytes/as-i64.eo`, `bytes/as-i32.eo`, and `i16.eo`.

The fix is to narrow the guard at `Pretty.java:291` so it parenthesizes only an argument that actually takes arguments of its own (a real application such as `5.plus 3`), not any node that merely has children — a data-receiver dispatch has a receiver child but no arguments and must inline bare. Concretely, wrap only when the flattened argument contains a space.

yegor256:
Fixes #5591.

## Problem

`Xmir.toEO()` wrapped a single dispatch token in redundant parentheses when inlining the arguments of a reversed dispatch. The `inlined()` method in `Pretty.java` parenthesized every argument whose node had children:

```java
if (arg.children.isEmpty()) {
    joined.append(flat.get());
} else {
    joined.append('(').append(flat.get()).append(')');
}
```

A data-receiver dispatch such as `01-.as-bool` is stored in XMIR as a reversed head over a data child, so its raw node has a child (the receiver `01-`) but takes no arguments of its own — it is a single token. The `!children.isEmpty()` guard therefore wrapped it as `(01-.as-bool)`, producing EO that no longer parses (`redundant parentheses around a single token`), which broke the `format` goal on `bytes/as-bool.eo`, `number/is-nan.eo`, `number/is-integer.eo`, `recovered.eo`, `bytes/as-i64.eo`, `bytes/as-i32.eo`, and `i16.eo`.

## Fix

The guard now weighs each argument in its **suffix-resolved** (effective) shape via `suffixed()`. For a data-receiver dispatch the receiver is folded back into the base (`01-.as-bool`), leaving no children, so it inlines bare. Only a genuine application that actually applies arguments (such as `5.plus 3`) still gets wrapped.

A structural check is used rather than the "contains a space" heuristic suggested in the issue, because a string literal like `"result is %d\n"` is a single token that legitimately contains spaces — the space-based rule would wrongly parenthesize it (and did, breaking an existing print-pack test).

## Tests

- Added print-pack `reversed-with-data-receiver-args.yaml` covering a reversed `eq.` whose arguments are data-receiver dispatches; it asserts both exact output and round-trip parseability.
- All 72 `XmirTest` cases pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01W8GVNraU9CKFcJL6mb37HN)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -272,7 +272,18 @@ final class Pretty {
 
     /**
      * Inline a list of arguments into one space-separated string,
-     * wrapping in parentheses those that take arguments themselves.
+     * wrapping in parentheses only those that apply arguments of their
+     * own (a real application such as {@code 5.plus 3}).
+     *
+     * <p>The guard weighs the argument in its effective (suffix-resolved)
+     * shape, not its raw one. A data-receiver dispatch such as
+     * {@code 01-.as-bool} is stored as a reversed head over a data child,
+     * so its raw node has a child (the receiver) yet it takes no arguments
+     * and is a single token — {@link #suffixed} folds the receiver back
+     * into the base, leaving no children, so it inlines bare. Wrapping it
+     * as {@code (01-.as-bool)} would produce EO that fails to parse with
+     * "redundant parentheses around a single token" (#5591).</p>
+     *
      * @param args The arguments
      * @return The inlined string, or empty if any argument can't be inlined
      */
@@ -288,7 +299,7 @@ final class Pretty {
             if (joined.length() > 0) {
                 joined.append(' ');
             }
-            if (arg.children.isEmpty()) {
+            if (Pretty.suffixed(arg).orElse(arg).children.isEmpty()) {
                 joined.append(flat.get());
             } else {
                 joined.append('(').append(flat.get()).append(')');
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (37) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2182](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5595, opened 2026-07-17
#5593 printer collapses self-reference to top-level object to bare name

### LINKED ISSUE #5593: Printer prints self-reference to top-level object as tuple.concat instead of bare concat
`Xmir.toEO()` prints a reference to the file's own top-level object with a redundant package prefix instead of the bare name. Re-formatting `eo-runtime/src/main/eo/tuple/concat.eo` (which is `+package tuple` and defines `[] > concat`) rewrites the two self-references inside its tests from the bare `concat` to `tuple.concat`:

```
    withouti > list3
      concat            ->      tuple.concat
        * 0 1                     * 0 1
        * 2 3                     * 2 3
      0                         0
```

The tell is that the sibling references in the same file — `reducedi` and `withouti` — keep printing bare, and only `concat`, the name of the object being defined, gets the prefix. The reason is that the parser's `resolve-self` pass qualifies a self-reference to `Φ.tuple.concat`, while external package members are left bare in the XMIR the printer receives (a bare `reducedi` base would otherwise print as `tuple.reducedi`, which it does not).

The printer then renders that `Φ.tuple.concat` base through the `otherwise` branch of the head template in `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:174-175`, which calls `eo:surface` (`to-eo-tree.xsl:27-30`). `eo:surface` only strips the `Φ.` root and keeps the whole package path, so it emits `tuple.concat`. It never recognizes that the base names the current program's own top-level object and should collapse to the bare `concat`, the way the source wrote it and every other same-package reference prints. The result is a readability regression and a loss of input↔output symmetry.

The fix is to add a branch to the head template (`to-eo-tree.xsl:155-177`) that, for a `Φ.<package>.<name>…` base whose `<package>` equals the file's `+package` meta and whose `<name>` is the top-level object (`/object/o[1]/@name`), renders the bare `<name>…` — unless the name is shadowed by an in-scope attribute, which the existing `Q.`-root branch at `to-eo-tree.xsl:163` already handles.

yegor256:
Closes #5593.

## Problem

`Xmir.toEO()` printed a reference to the file's own top-level object with a redundant package prefix instead of the bare name. Re-formatting `eo-runtime/src/main/eo/tuple/concat.eo` (which is `+package tuple` and defines `[] > concat`) rewrote the self-references inside its tests from bare `concat` to `tuple.concat`:

```
    withouti > list3
      concat            ->      tuple.concat
        * 0 1                     * 0 1
        * 2 3                     * 2 3
      0                         0
```

The parser homes a same-file reference into `Φ.tuple.concat`, while external package members stay bare in the XMIR the printer receives. The head template in `to-eo-tree.xsl` rendered that `Φ.tuple.concat` base through the `otherwise` branch (`eo:surface`), which only strips the `Φ.` root and keeps the whole package path, emitting `tuple.concat`. It never recognized that the base names the current program's own top-level object and should collapse to bare `concat`.

## Fix

Add a branch to the head template (`to-eo-tree.xsl`) that, for a `Φ.<package>.<name>…` base whose `<package>` equals the file's `+package` meta and whose first name segment is the top-level object (`/object/o[1]/@name`), renders the bare `<name>…` — the way the source wrote it and every other same-package reference prints. The branch is skipped when an in-scope attribute shadows the name: if the package segment itself is shadowed the existing `Q.`-root branch keeps it rooted, otherwise the qualified `<package>.<name>` survives through the `otherwise` branch, both of which resolve correctly.

## Tests

Added `self-reference.yaml` print-pack. It reproduced `tuple.concat` before the fix and now prints bare `concat`, and round-trips through both `printsToEo` and `printsToParseableEo`. Full `eo-printer` module test suite passes (97 tests).

https://claude.ai/code/session_01Lb6FPAFXqikpCtLnwDxSjL

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lb6FPAFXqikpCtLnwDxSjL)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -152,6 +152,13 @@
     <xsl:variable name="rho-prefix" select="concat($eo:xi, '.', $eo:rho, '.')"/>
     <xsl:variable name="rest" select="substring-after(@base, $rho-prefix)"/>
     <xsl:variable name="first" select="if (contains($rest, '.')) then substring-before($rest, '.') else $rest"/>
+    <!-- The current program's "+package" and the "Φ.<package>." prefix a
+         self-reference to a same-file object carries after being homed
+         into the package (add-default-package / build-fqns). -->
+    <xsl:variable name="package" select="string(/object/metas/meta[head='package']/part[1])"/>
+    <xsl:variable name="self-prefix" select="concat($eo:program, '.', $package, '.')"/>
+    <xsl:variable name="self-rest" select="substring-after(@base, $self-prefix)"/>
+    <xsl:variable name="self-first" select="if (contains($self-rest, '.')) then substring-before($self-rest, '.') else $self-rest"/>
     <xsl:choose>
       <!-- NOT OPTIMIZED TUPLE -->
       <xsl:when test="@star">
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2214](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5597, opened 2026-07-17
#5594 emit hybrid inline-phi shorthand for vertically-argued decoratees

### LINKED ISSUE #5594: Printer skips compact inline-phi shorthand when the decoratee has vertical arguments
`Xmir.toEO()` uses the compact inline-phi shorthand only when the whole decoratee fits on one line. When the decoratee has vertically-laid-out arguments, the printer drops the shorthand entirely and falls back to the verbose `> [params] > name` + `<phi> > @` shape, even though a shorter, equivalent hybrid form exists. This affects both variants of the shorthand — the `++> name` test form and the ordinary `> [params] > name` formation form.

Test form, `eo-runtime/src/main/eo/tuple/front.eo`, renders:

```
  ++> tests-list-front-with-zero-index
    is-empty > @
      tuple.front
        * 1 2 3
        0
```

but the equivalent hybrid is one line shorter and one indent shallower:

```
  is-empty ++> tests-list-front-with-zero-index
    tuple.front
      * 1 2 3
      0
```

Ordinary formation, `eo-runtime/src/main/eo/tuple/hash-code.eo`, renders:

```
  [acc index] > rec-hash-code
    if. > @
      index.eq size
      ...
```

but the equivalent hybrid is:

```
  if. > [acc index] > rec-hash-code
    index.eq size
    ...
```

I verified both hybrids parse cleanly and round-trip: after hand-editing each file to the second shape, the format goal re-parses without error and reformats it straight back to the first, so the spellings denote the same object.

The cause is `Pretty.phi()` in `eo-printer/src/main/java/org/eolang/printer/Pretty.java:244-270`. It builds the shorthand by calling `Pretty.flat(decoratee)` (line 255) and only succeeds when the entire decoratee inlines onto a single line; a decoratee whose arguments must go vertical makes `flat()` return empty, so `phi()` yields empty and the caller keeps the verbose layout. The hybrid rendering — the decoratee's head on the shorthand line (`++> name` or `> [params] > name`) with its arguments as the vertical body, mirroring the ordinary `head > name` plus vertical-args layout — is never generated, so the penalty search never gets to prefer it.

The fix is to extend `phi()` (or its caller) so that when `flat(decoratee)` fails it still emits the shorthand head line and lays the decoratee's arguments out vertically beneath, offering that hybrid as a candidate alongside the verbose form, instead of only offering the fully-inline single line.

yegor256:
Closes #5594.

## Problem

`Xmir.toEO()` used the compact inline-phi shorthand (the `++> name` test form and the ordinary `> [params] > name` formation form) **only** when the whole decoratee fit on one line. When the decoratee's arguments had to go vertical, `Pretty.phi()` — which builds the shorthand via `Pretty.flat(decoratee)` — yielded empty, and the printer fell back to the verbose `> @` layout, even though a shorter, equivalent hybrid spelling exists.

For example, the printer produced:

```
  ++> tests-list-front-with-zero-index
    is-empty > @
      tuple.front
        * 1 2 3
        0
```

instead of the one-line-shorter, one-indent-shallower equivalent:

```
  is-empty ++> tests-list-front-with-zero-index
    tuple.front
      * 1 2 3
      0
```

## Fix

- Extracted the shared marker construction (`marker()`) and the "lone `> @` decoratee" guard (`decorated()`) out of `phi()`.
- Added `hybrid()`, the multi-line counterpart of `phi()`: it keeps the compact marker on the head line (`head ++> name` or `head > [params] > name`) and lays the decoratee's arguments out vertically beneath, mirroring the ordinary `head > name` plus vertical-args layout.
- `shaped()` now offers the hybrid as a third candidate, but only when no single line exists (i.e. `flat(decoratee)` failed), so the penalty search prefers it over the verbose vertical form without disturbing cases where the fully-inline single line already wins.
- The hybrid is withheld for a formation decoratee (its bindings are vertical, not arguments) and for a receiver-only reversed dispatch (e.g. `not.` with just its receiver), mirroring `flat()`'s own reversed rejection. A reversed dispatch that also carries arguments (`if.` with its branches) does get the hybrid.

## Tests

Added two `print-packs` yaml packs — one for each shorthand variant:

- `hybrid-phi-test.yaml` — the `++> name` test form with a vertically-argued decoratee.
- `hybrid-phi-formation.yaml` — the ordinary `> [params] > name` form with a reversed `if.` decoratee whose branch must go vertical.

Both are exercised by `XmirTest` (round-trip included). The full `eo-printer` suite passes (101 tests).

https://claude.ai/code/session_01PyEsBQMr4hukN6j6xSggvK

---
_Generated by [Claude Code](https://claude.ai/code/session_01PyEsBQMr4hukN6j6xSggvK)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -116,8 +116,17 @@ final class Pretty {
     }
 
     /**
-     * Lay out a single shape of a node, picking the lower-penalty of its
-     * vertical and horizontal renderings.
+     * Lay out a single shape of a node, picking the lowest-penalty of its
+     * candidate renderings.
+     *
+     * <p>The vertical rendering is always available. The horizontal
+     * (single-line) one is tried when the node can be inlined. For a
+     * formation whose only binding is the {@code φ} decoratee, when no
+     * single line exists (the decoratee's arguments must go vertical),
+     * the {@link #hybrid} form — the compact inline-phi marker with the
+     * arguments beneath — is offered instead, so the penalty search can
+     * prefer it over the fully verbose vertical shape (issue #5594).</p>
+     *
      * @param node The node
      * @param indent The indentation level
      * @return The rendered block
@@ -129,6 +138,13 @@ final class Pretty {
             && new Penalty(flat.get(), this.weights).points()
             < new Penalty(best, this.weights).points()) {
             best = flat.get();
+        } else if (flat.isEmpty()) {
+            final Optional<String> hybrid = this.hybrid(node, indent);
+            if (hybrid.isPresent()
+                && new Penalty(hybrid.get(), this.weights).points()
+                < new Penalty(best, this.weights).points()) {
+                best = hybrid.get();
+            }
         }
         return best;
     }
@@ -233,9 +249,10 @@ final class Pretty {
      * formation with any other attribute keeps the vertical layout. The
      * decoratee itself is inlined through the usual {@link #flat} path,
      * so a decoratee that can't be inlined (a nested formation, a tuple)
-     * yields empty and the caller falls back to the vertical shape; the
-     * penalty/width check then decides whether this single line is
-     * actually preferable.</p>
+     * yields empty here; the caller then falls back to either the
+     * {@link #hybrid} form (marker compacted, arguments vertical) or the
+     * fully verbose vertical shape, whichever the penalty/width check
+     * prefers.</p>
      *
      * @param node The formation node
      * @param indent The indentation level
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (102 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2218](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5602, opened 2026-07-17
#5598 withhold hybrid inline-phi when a decoratee argument is named

### LINKED ISSUE #5598: Hybrid inline-phi shorthand emits invalid EO when a decoratee argument carries a name suffix

yegor256:
Fixes #5598.

## Problem

The hybrid inline-phi shorthand added in #5594 was emitted even when a decoratee's arguments carried a name suffix, producing EO that could not be re-parsed. Re-formatting files such as `eo-runtime/src/main/eo/number/integral.eo` collapsed

```
[sum] >>
  malloc.for > @
    a
    [left] >>
      ...
```

into

```
malloc.for > [sum] >>
  a
  [left] >>
    ...
```

The collapsed `malloc.for > [sum] >>` is the compact inline-phi form of the only-phi formation `[sum]`, so its indented body lines are the formation's own content. A named line such as `[left] >>` there reads as a second attribute of `[sum]`, but an only-phi formation may hold only its `φ` decoratee — so re-parsing failed with `argument of an only-phi formation cannot carry a name suffix`.

## Fix

In `Pretty.phi()` the hybrid branch now fires only when every decoratee argument has an empty tail (no `> name`, `> [params]`, or `>>` suffix). When any argument is named, the branch yields empty and the verbose ` > @` layout is kept, as before #5594 — that form gives the decoratee its own scope, where named arguments are legal.

The four-predicate guard is split across two local booleans (`applied`, `unnamed`) to stay within the project's boolean-expression-complexity and nesting limits.

## Tests

Added the print-pack `hybrid-phi-named-argument.yaml`, whose decoratee carries a `[left] >>` argument. The outer only-phi correctly stays verbose while the inner unnamed decoratee still collapses, demonstrating the guard is precise. Full `eo-printer` suite (103 tests) and qulice pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CUWiRT2tEj7m5nvzx8X1mi)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -247,6 +247,15 @@ final class Pretty {
      * and keep the verbose layout; a reversed dispatch that also carries
      * arguments ({@code if.} with its branches) does get the hybrid.</p>
      *
+     * <p>The hybrid is also withheld when any decoratee argument carries a
+     * name suffix ({@code [left] >>}, {@code malloc.for > [b]}, {@code
+     * b.put > [m] >>}). Those arguments become the body of the collapsed
+     * only-phi formation, and an only-phi formation may hold nothing but its
+     * {@code φ} decoratee, so a named line there fails to parse with
+     * "argument of an only-phi formation cannot carry a name suffix"
+     * (issue #5598). Keeping the verbose {@code  > @} layout gives the
+     * decoratee its own scope, where named arguments are legal.</p>
+     *
      * @param node The formation node
      * @param indent The indentation level
      * @return The rendered block, or empty if the inline-phi form doesn't apply
@@ -269,6 +278,10 @@ final class Pretty {
                     false, decoratee.reversed, decoratee.data, decoratee.children
                 )
             );
+            final boolean applied = !decoratee.abstractt && !decoratee.children.isEmpty()
+                && !(decoratee.reversed && decoratee.children.size() <= 1);
+            final boolean unnamed = decoratee.children.stream()
+                .allMatch(kid -> kid.tail.isEmpty());
             if (value.isPresent()) {
                 result = Optional.of(
                     new StringBuilder(this.step().repeat(indent))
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2226](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5603, opened 2026-07-17
#5601 Improve error message for named argument in only-phi formation

### LINKED ISSUE #5601: Improve error message for named argument in only-phi formation
The parser error `argument of an only-phi formation cannot carry a name suffix` is correct but too terse to act on. It is raised by `Eo.checkNaming` at `eo-parser/src/main/java/org/eolang/parser/Eo.java:787-792` when a named line appears in the body of a compact only-phi formation.

Given input like

```
a > [] >> bar
  b > foo
```

the message points at the line but never names the offender or explains the rule: `foo` would be a second attribute of the only-phi formation `bar`, and such a formation may hold only its `φ` decoratee. A reader sees "an only-phi formation" with no clue which formation, which name, or why one name is allowed and another is not.

The message should name the offending attribute and the formation, and state the constraint — for example: `foo cannot be a named attribute of only-phi formation bar, which binds only its φ decoratee`. The formation name is available on the closing level and the offending name on the flagged child, so both can be interpolated.

This is only about the wording of an existing, correct diagnostic — the check itself is right and should stay.

yegor256:
Closes #5601.

## What

The parser diagnostic raised by `Eo.checkNaming` when a named line appears in the body of a compact only-phi formation was correct but too terse:

```
argument of an only-phi formation cannot carry a name suffix
```

It named neither the offending attribute nor the formation, nor did it state the rule. Given:

```
a > [] > bar
  b > foo
```

the message now reads:

```
foo cannot be a named attribute of only-phi formation bar, which binds only its φ decoratee
```

When the offending attribute is auto-named (`>>` with no handle) or the formation is anonymous (`>>`), the message falls back to generic phrasing (`an auto-named attribute` / `an only-phi formation`), so the existing anonymous-formation test cases still produce a clear message.

The check itself is unchanged — this is only about wording.

## How

- `Level` now records the suffix's source name (`label`) and the governing only-phi formation's name (`formation`). These two strings replace the `named` and `argument` booleans they subsume (`named()` is `label != null`, `argument()` is `formation != null`), so the field count is unchanged.
- `Transition.apply` threads the label through the shared push/replace prologue; `Suffix.named()` returns the source name or `null` when no suffix is present.
- `Stack` propagates the formation name down to argument levels as it flags them.
- `Level.onlyPhiNamingError()` builds the message, keeping `Eo` unchanged in complexity.
- `PARSER_SPEC.md` error table updated to reflect the new wording.

## Tests

- Updated the two existing `eo-typos` expectations (anonymous formations → generic phrasing).
- Added `only-phi-named-arg-with-name.yaml` demonstrating the interpolated attribute and formation names.
- Updated `LevelTest` / `TransitionTest` for the new `name(label)` / `label()` API.

Full `eo-parser` test suite (1467 tests) and `qulice:check` both pass.

https://claude.ai/code/session_01FvevRz8eJdqQvCQMZ84uHN

---
_Generated by [Claude Code](https://claude.ai/code/session_01FvevRz8eJdqQvCQMZ84uHN)_

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -1310,7 +1310,7 @@ R-9.9.1. Every error condition in this spec has a single canonical text — **in
 | Deeper-indent under horizontally-completed line | `unexpected deeper-indent line — previous expression is closed for children` |
 | `.method` continuation on horizontally-completed previous | `method continuation not allowed after horizontal application` |
 | `.method` continuation on an only-phi formation | `method continuation not allowed after only-phi formation` |
-| Name suffix on an only-phi φ's argument | `argument of an only-phi formation cannot carry a name suffix` |
+| Name suffix on an only-phi φ's argument | `<name> cannot be a named attribute of only-phi formation <formation>, which binds only its φ decoratee` (the formation is described generically as `an only-phi formation` when anonymous) |
 | `.method` line at top level, deeper than parent, or with no same-indent sibling | `method continuation has no expression to attach to` |
 | Chained inline-phi suffix `expr > [a] > [b] > name` | `chained inline-phi suffixes are not allowed` |
 | Inline-phi without a name on the right (`expr > [params]` alone) | `inline-phi formation must carry a name on the right` |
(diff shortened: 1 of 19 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (198 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2230](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5605, opened 2026-07-17
#5604 withhold hybrid inline-phi when a named line is nested below the decoratee's direct children

### LINKED ISSUE #5604: Hybrid inline-phi collapse still emits invalid EO when the auto-named line is nested below the decoratee's direct children
The hybrid inline-phi collapse in `Pretty.phi()` (`eo-printer/src/main/java/org/eolang/printer/Pretty.java:281-302`) still emits EO that fails to parse, in a case the fix for #5598 does not cover. That fix added an `unnamed` guard (`Pretty.java:283-284`) that withholds the hybrid only when a *direct* child of the decoratee carries a name suffix. When the offending auto-named line is nested deeper — inside a tuple, a dispatch, or an application — the guard passes and the collapse fires anyway.

Re-formatting `eo-runtime/src/main/eo/{file.eo,tuple/reducedi.eo,while.eo}` with `-Deo.autoFix=true` now produces lines such as `x.plus > [a x i] >>` (reducedi.eo) and `m.put (f.read 12) > [f] >>` (file.eo), which fail to re-parse with `an auto-named attribute cannot be a named attribute of an only-phi formation, which binds only its φ decoratee`. In `reducedi.eo` the decoratee is `eq.`, whose direct children `6` and `reducedi …` are both unnamed, so the guard sees nothing wrong; but `reducedi`'s own argument `[a x i] >>` is auto-named, and after the collapse it binds to the enclosing compact only-phi test formation, which may hold only its φ. The verbose master form kept `[a x i] >>` under an explicit `> @` decoratee scope, where it is legal.

The `unnamed` check should span the entire decoratee subtree, not just its immediate children: withhold the hybrid whenever any descendant of the decoratee carries a name suffix (a `> name`, `> [params]`, or `>>` tail), and keep the verbose `> @` layout, which gives the decoratee its own scope where those named arguments parse.

yegor256:
Fixes #5604.

## Problem

The `unnamed` guard added for #5598 in `Pretty.phi()` only inspected the *direct* children of a decoratee. When the offending auto-named line sat deeper in the decoratee's subtree — inside a tuple, a dispatch, or an application — the guard passed and the hybrid inline-phi collapse fired anyway, producing unparseable EO such as `x.plus > [a x i] >>` (`tuple/reducedi.eo`) and `m.put (f.read 12) > [f] >>` (`file.eo`). These fail to re-parse with:

> an auto-named attribute cannot be a named attribute of an only-phi formation, which binds only its φ decoratee

For example, in `reducedi.eo` the decoratee is `eq.`, whose direct children `6` and `reducedi …` are both unnamed, so the shallow guard saw nothing wrong — but `reducedi`'s own argument `[a x i] >>` is auto-named, and after the collapse it would bind to the enclosing compact only-phi formation.

## Fix

Extend the `unnamed` check to span the **entire** decoratee subtree. A new `Node.nameless()` method walks a node and all of its descendants, and the guard now withholds the hybrid whenever any descendant carries a name suffix (`> name`, `> [params]`, or a `>>` tail), keeping the verbose `> @` layout where the decoratee gets its own scope and those named arguments parse.

The recursive check lives on the nested `Node` class rather than on `Pretty` to stay within the method-count and cyclomatic-complexity limits enforced by qulice.

## Tests

Added `hybrid-phi-deep-named-argument.yaml`, a print-pack that reproduces the `reducedi.eo` shape (an only-phi test attribute whose decoratee subtree carries an auto-named `[a x i] >>` argument nested two levels down). It asserts the outer `eq. > @` stays verbose and the printed EO re-parses cleanly. Both `XmirTest.printsToEo` and `XmirTest.printsToParseableEo` exercise it.

All 105 `eo-printer` tests pass, and `qulice:check` is clean.

https://claude.ai/code/session_01X1Guq2KvrN3btZhVVHbG57

---
_Generated by [Claude Code](https://claude.ai/code/session_01X1Guq2KvrN3btZhVVHbG57)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -247,14 +247,18 @@ final class Pretty {
      * and keep the verbose layout; a reversed dispatch that also carries
      * arguments ({@code if.} with its branches) does get the hybrid.</p>
      *
-     * <p>The hybrid is also withheld when any decoratee argument carries a
-     * name suffix ({@code [left] >>}, {@code malloc.for > [b]}, {@code
-     * b.put > [m] >>}). Those arguments become the body of the collapsed
-     * only-phi formation, and an only-phi formation may hold nothing but its
-     * {@code φ} decoratee, so a named line there fails to parse with
-     * "argument of an only-phi formation cannot carry a name suffix"
-     * (issue #5598). Keeping the verbose {@code  > @} layout gives the
-     * decoratee its own scope, where named arguments are legal.</p>
+     * <p>The hybrid is also withheld when any line in the decoratee's whole
+     * subtree carries a name suffix ({@code [left] >>}, {@code malloc.for >
+     * [b]}, {@code b.put > [m] >>}). The decoratee's subtree becomes the body
+     * of the collapsed only-phi formation, and an only-phi formation may hold
+     * nothing but its {@code φ} decoratee, so a named line anywhere within it
+     * fails to parse with "an auto-named attribute cannot be a named attribute
+     * of an only-phi formation, which binds only its φ decoratee" (issues
+     * #5598, #5604). A direct child is not enough to check: the offending line
+     * may be nested deeper — inside a tuple, a dispatch, or an application —
+     * where a shallow guard would miss it (issue #5604). Keeping the verbose
+     * {@code  > @} layout gives the decoratee its own scope, where named
+     * arguments are legal.</p>
      *
      * @param node The formation node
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2238](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5606, opened 2026-07-17
#5596 print large whole doubles with EO exponent notation

### LINKED ISSUE #5596: Printer expands large numbers to full decimal instead of exponent notation
`Xmir.toEO()` expands a large-magnitude number literal into its full decimal form instead of keeping exponent notation. Re-formatting `eo-runtime/src/main/eo/tuple/hash-code.eo` turns

```
    42.42e42 > bigval!
```

into

```
    42420000000000000000000000000000000000000000 > bigval!
```

That is 44 digits, forty of them trailing zeros. Both spellings parse to the very same IEEE-754 double (`4.242e43`), so the change round-trips and the tests pass, but the expanded form is far harder to read than the `42.42e42` the source used.

The cause is `StUnhex.number()` in `eo-printer/src/main/java/org/eolang/printer/StUnhex.java:95-109`. For a whole-valued double it prints an integer, and when the magnitude reaches 2^63 it takes the `else` branch at line 103, `BigDecimal.valueOf(num).toBigInteger().toString()`, which emits every digit. Smaller whole numbers go through `Long.toString` (fine), and non-whole numbers go through `Double.toString` (which already yields exponent notation such as `4.242E43`) — only the large whole-valued case expands.

The fix is to render large-magnitude whole values in EO's exponent form rather than the full expansion: in that `else` branch, derive the literal from `Double.toString(num)` (already `4.242E43`) and lower-case the `E` to match EO's `e` literal syntax, instead of `toBigInteger().toString()`. The maintainer may want a magnitude or trailing-zero threshold so ordinary integers keep their plain form and only genuinely long ones switch to the exponent.

Shaurya2k06:
Fixes #5596 

## Problem

`Xmir.toEO()` expands large-magnitude whole doubles into full decimal form instead of preserving scientific notation. For example, reformatting `eo-runtime/src/main/eo/tuple/hash-code.eo` changes:

```eo
42.42e42 > bigval!
```

into:

```eo
42420000000000000000000000000000000000000000 > bigval!
```

Both literals represent the same IEEE-754 double (`4.242e43`), so round-tripping remains correct and the existing tests pass. However, the expanded decimal form is significantly less readable.

The issue originates in `StUnhex.number()` (`eo-printer/src/main/java/org/eolang/printer/StUnhex.java`). For whole-valued doubles with magnitude ≥ `2^63`, the method currently executes:

```java
BigDecimal.valueOf(num).toBigInteger().toString()
```

which emits the full decimal expansion. Other cases already produce appropriate output:

- Small whole numbers use `Long.toString()`.
- Non-whole doubles use `Double.toString()`, which preserves scientific notation when appropriate.

As a result, only large whole-valued doubles lose exponent notation.

## Fix

Update the `|n| ≥ 2^63` branch of `StUnhex.number()` to derive the literal from `Double.toString(num)` and convert the exponent marker from uppercase `E` to lowercase `e`, matching EO syntax (for example, `1.0e19` and `4.242e43`).

The existing `0x1p63` threshold continues to ensure ordinary integers are emitted as plain decimal values. Only values that previously went through the `BigDecimal` path are affected.

## Tests

Updated `convertsBigIntegerValuedDoubleFromHexToEo` in `StUnhexTest` to expect:

```text
1.0e19
```

instead of:

```text
10000000000000000000
```

for the hexadecimal input:

```text
43-E1-58-E4-60-91-3D-00
```

All **105** `eo-printer` tests pass, and `qulice:check` completes successfully.

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
+++ b/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
@@ -7,7 +7,6 @@ package org.eolang.printer;
 import com.yegor256.xsline.Shift;
 import com.yegor256.xsline.StEnvelope;
 import com.yegor256.xsline.StSequence;
-import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.regex.Pattern;
@@ -88,7 +87,7 @@ final class StUnhex extends StEnvelope {
 
     /**
      * Convert given number to string.
-     * Prints as int if fractional part of number is 0.
+     * Prints as int when the fractional part is zero and the magnitude fits in long.
      * @param num Number to convert
      * @return Number converted to string
      */
@@ -100,7 +99,7 @@ final class StUnhex extends StEnvelope {
             } else if (Math.abs(num) < 0x1p63) {
                 str = Long.toString(num.longValue());
             } else {
-                str = BigDecimal.valueOf(num).toBigInteger().toString();
+                str = Double.toString(num).replace('E', 'e');
             }
         } else {
             str = Double.toString(num);
--- a/eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
+++ b/eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
@@ -133,7 +133,7 @@ final class StUnhexTest {
                     "<p><o base='Φ.number'><o base='Φ.bytes'><o>43-E1-58-E4-60-91-3D-00</o></o></o></p>"
                 )
             ),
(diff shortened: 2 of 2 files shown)


0crat:
@shaurya2k06 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (7) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+132](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5616, opened 2026-07-17
#5614: printer renders void type annotations (the /{...} suffix)

### LINKED ISSUE #5614: Printer drops void attribute type annotations (the /{...} suffix) when rendering formations
The printer discards the type annotation of a void attribute. In `eo-runtime/src/main/eo/bytes.eo:30-31` the source declares

```
[start len] > slice /bytes
  ? > cant-slice /{string}
```

but after the `format` goal it becomes `[start len cant-slice] > slice /bytes` — the void `cant-slice` is folded into the params bracket and its `/{string}` type is gone. The parser keeps the annotation correctly: the 1-parse XMIR carries `<o base="∅" name="cant-slice" types="Φ.string"/>` (emitted by `Emit.types()` at `eo-parser/src/main/java/org/eolang/parser/Emit.java:343`, and covered by `eo-parser/src/test/resources/org/eolang/parser/eo-packs/parse/atom-void-formas.yaml`), so the loss happens purely in the printer.

The cause is in `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:260-286`: the `/type` suffix is emitted only for atoms, from the `λ`-atom child (guarded by `eo:atom(.)`), and no code path in the printer ever reads the `@types` attribute. A void carrying `@types` is therefore rendered as a bare name, and since a bracket param cannot express a type, folding it into `[...]` guarantees the annotation is dropped — silently changing the object's signature while still parsing.

Render `@types` back as a `/{...}` suffix: when a void attribute has a non-empty `@types`, emit it as a `? > name /{...}` body line (mirroring the parser's input form, with the space-separated FQNs), rather than as a plain bracket param, reusing the root-stripping the `λ`-atom branch already does at lines 264-280.

yegor256:
Fixes #5614.

## Problem

The printer discarded the type annotation of a void attribute. For example, `eo-runtime/src/main/eo/bytes.eo` declares:

```
[start len] > slice /bytes
  ? > cant-slice /{string}
```

but after the `format` goal it became `[start len cant-slice] > slice /bytes` — the void `cant-slice` was folded into the params bracket and its `/{string}` type was gone. The parser keeps the annotation correctly (as `@types`); the loss happened purely in the printer.

The cause was in `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl`: the `/type` suffix was emitted only for atoms (from the `λ`-atom child), and no code path ever read the `@types` attribute. A void carrying `@types` was rendered as a bare name and folded into `[...]`, which cannot express a type — silently changing the object's signature while still parsing.

## Fix

Render `@types` back as a `/{...}` suffix. When a void attribute has a non-empty `@types`, it is emitted as a vertical `? > name /{...}` body line (mirroring the parser's input form, with space-separated FQNs) rather than as a plain bracket param:

- Typed voids are kept out of the `[...]` head and off the `@local` (`>>`) branch.
- A new tree-mode template emits the `/{...}` suffix from `@types`, handling both the `? > name` and `? >> name` shapes.
- The `λ`-atom root-stripping (drop the implicit `Φ.` for single-name signatures, keep an explicit `Q.` for multi-segment ones) is extracted into a shared `eo:signature()` helper, reused by both the atom `/sig` and the void forma rendering.

## Tests

Added `eo-printer/.../print-packs/yaml/void-type-annotation.yaml`, a round-trip test covering a single-name forma (`/{string}`) and a multi-forma list with a rooted forma (`/{string Q.a.b.c}`). It exercises both `printsToEo` (exact output) and `printsToParseableEo` (reprints to the same EO). The full `eo-printer` suite is green.

https://claude.ai/code/session_01KvdGh7XyPPMPnPxoJvun9f

---
_Generated by [Claude Code](https://claude.ai/code/session_01KvdGh7XyPPMPnPxoJvun9f)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -41,6 +41,19 @@
     <xsl:param name="text" as="xs:string"/>
     <xsl:sequence select="replace($text, concat('a', $eo:cactoos, '(\d+)-(\d+)'), 'v$1_$2')"/>
   </xsl:function>
+  <!--
+  Render a stored signature (an atom's "/sig" or a void forma) as its EO
+  surface form. A single-name "Φ.name" re-resolves to its root on
+  reparse, so the implicit root is dropped; a multi-segment
+  "Φ.a.b.c" must stay rooted (a dotted signature is not re-homed) and
+  keeps an explicit "Q." root. Names with no "Φ." root pass through.
+  -->
+  <xsl:function name="eo:signature" as="xs:string">
+    <xsl:param name="sig" as="xs:string"/>
+    <xsl:variable name="root" select="concat($eo:program, '.')"/>
+    <xsl:variable name="rest" select="substring-after($sig, $root)"/>
+    <xsl:sequence select="if (starts-with($sig, $root)) then (if (contains($rest, '.')) then concat('Q.', $rest) else $rest) else $sig"/>
+  </xsl:function>
   <!-- PROGRAM -->
   <xsl:template match="object">
     <object>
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2306](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5617, opened 2026-07-17
fix(#5615): keep compact `seq *` idiom in printer

### LINKED ISSUE #5615: `seq *` idiom unexpectedly reformatted, adding extra indent level and line
The printer never glues a `*` tuple to the line of the object that applies it, so it de-compacts the ordinary `seq *` idiom. In `eo-runtime/src/main/eo/directory.eo` the hand-written source on `master` reads

```
seq *
  (directory (Q.file ^.as-path.dirname)).made
  if.
    outcome.eq 0
    ...
```

and the formatter rewrites it to

```
seq
  *
    made.
      directory (Q.file ^.as-path.dirname)
    if.
      outcome.eq 0
      ...
```

pushing every tuple element one line down and one indent level deeper. The same happens at the `deleted` and `rec-delete` sites in that file, and in `file.eo`.

The two shapes are identical trees: the parser's own test `eo-parser/src/test/resources/org/eolang/parser/eo-packs/parse/validate-resolve-before-stars.yaml` asserts that `sprintf *` followed by two indented lines `x`/`y` yields exactly one child, `tuple(x, y)` — a bare trailing `*` absorbs all the following indented siblings into the tuple, with no before-star ambiguity. So `seq *` + indented elements means precisely `seq(tuple(...))`, the same as the verbose `seq` / `*` / elements. The compact form is one line shorter and one level shallower, and it is what humans actually write, so expanding it works against the canonical form this project is trying to pin down.

The cause is `Pretty.flat()` in `eo-printer/src/main/java/org/eolang/printer/Pretty.java:362`, where `"*".equals(node.base)` forces every star node to render on its own line; there is no candidate that keeps a trailing star at the tail of its applying object's line. Add such a candidate: when a star is the last child of an application, render `head *` on one line with the star's elements laid out beneath at `indent + 1` (a hybrid mirroring the inline-phi form of #5594). Scope it to this end-of-line vertical case only, so the genuinely ambiguous fully-horizontal `head * a b` (the before-star territory) is left untouched.

yegor256:
Closes #5615.

## Problem

`Pretty.flat()` refused to inline any `*` node, so an application whose sole/last argument is a `*` tuple could only ever be rendered vertically with the star on its own line. The ordinary `seq *` idiom therefore got de-compacted:

```
seq *
  a.b
  if.
    ...
```

was rewritten to

```
seq
  *
    a.b
    if.
      ...
```

pushing every tuple element one line down and one indent level deeper. The same happened at the `deleted` / `rec-delete` sites in `directory.eo` and in `file.eo`. The two shapes are the identical tree (`seq(tuple(...))`) — a trailing `*` absorbs exactly the indented siblings beneath it — so the expansion only worked against the canonical form.

## Fix

Add a trailing-star hybrid candidate in `Pretty.shaped`, alongside the existing vertical and horizontal candidates:

- Applies only to a **plain** (non-formation, non-reversed) application whose **last** child is a **non-empty, unnamed** star; any preceding arguments must inline.
- Renders `head *` on one line (carrying the application's own name suffix, e.g. `seq * > @`) with the tuple's elements laid out **vertically** at `indent + 1`.
- The elements are always vertical, so the genuinely ambiguous fully-horizontal `head * a b` (before-star territory) is never produced here.
- It is only a candidate — the existing `Penalty` comparison keeps it only when it beats the plain vertical/horizontal renderings, so a short tuple that still fits inline as `* 1 2` on its own line is left untouched.

This mirrors the hybrid inline-phi form introduced in #5594.

## Tests

Added `eo-printer/.../print-packs/yaml/seq-star-idiom.yaml`, which feeds the verbose `seq` / `*` / elements shape and asserts the compact `seq *` output. Both parameterized checks cover it: `printsToEo` (verbose in → compact out) and `printsToParseableEo` (the compact output re-parses to the same source). Full `eo-printer` suite passes (`XmirTest` 84, `PenaltyTest` 9, `StUnhexTest` 9, `StEoLoggedTest` 5; 0 failures).

https://claude.ai/code/session_01QDnmgdcud8yyGMBAN2XhFS

---
_Generated by [Claude Code](https://claude.ai/code/session_01QDnmgdcud8yyGMBAN2XhFS)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -18,11 +18,13 @@ import java.util.stream.Collectors;
  * <p>It takes the intermediate "line tree" produced by
  * {@code to-eo-tree.xsl} (an {@code <eo>} element with a {@code
  * <preamble>} and nested {@code <line>} elements) and lays it out as
- * EO source. For every object it considers two renderings — a
- * vertical one, where the arguments go on their own indented lines,
- * and a horizontal one, where they are inlined (wrapped in
- * parentheses when needed) — and keeps the one with the smaller
- * {@link Penalty}. The decision is made recursively, bottom-up.</p>
+ * EO source. For every object it considers a few renderings — a
+ * vertical one, where the arguments go on their own indented lines, a
+ * horizontal one, where they are inlined (wrapped in parentheses when
+ * needed), and, for a tuple applied at the tail, a hybrid that glues
+ * the {@code *} to the head's line — and keeps the one with the
+ * smaller {@link Penalty}. The decision is made recursively,
+ * bottom-up.</p>
  *
  * @since 0.57.0
  */
@@ -44,6 +46,11 @@ final class Pretty {
     private final Map<PenaltyKey, Integer> weights;
 
     /**
+     * A single level of indentation, whose width is the {@code STEP} weight.
+     */
+    private final String tab;
+
+    /**
      * Ctor, using the default penalty weights.
      * @param element The {@code <eo>} element
      */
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (164 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2310](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5619, opened 2026-07-17
#5618: printer preserves `!` const marker on attributes nested inside a const-valued formation

### LINKED ISSUE #5618: Printer drops the `!` const marker on attributes nested inside a const-valued formation
The formatter loses the `!` const marker whenever a constant attribute sits inside another constant. In `eo-runtime/src/main/eo/mktemp.eo` on `master` the five environment lookups read

```
getenv "TEMP" > temp!
getenv "TEMPDIR" > tempdir!
getenv "TMP" > tmp!
getenv "TMPDIR" > tmpdir!
getenv "USERPROFILE" > userprofile!
```

and the formatter rewrites them to bare, non-constant attributes:

```
as-bytes. > temp
  dataized (getenv "TEMP")
...
```

The `!` is gone and the raw `(dataized …).as-bytes` desugaring is left exposed. These five attributes are all bound inside the surrounding `[] >>!` formation (the `string` decoratee), which is itself constant — that nesting is what triggers the loss. The auto-named `[] >>!` on the same object keeps its `!`, so the round-trip is asymmetric.

The marker is meant to survive: the parser lowers `x!` to `(dataized x).as-bytes` via `const-to-dataized.xsl`, and the printer reverses it via `eo-printer/src/main/resources/org/eolang/printer/print/dataized-to-const.xsl`. The reversal is wrong for nested constants. When it rebuilds the outer constant's value it copies the value's children with `<xsl:copy-of select="."/>` at `dataized-to-const.xsl:21`, which copies the subtree verbatim and never applies templates to it. So the inner `(dataized …).as-bytes` nodes are copied as-is and the template that would turn them back into `@const` never fires on them. Dropping the marker is not cosmetic: `!` caches the value once, so a plain attribute re-reads the environment on every access.

Change the `<xsl:copy-of select="."/>` at `dataized-to-const.xsl:21` to `<xsl:apply-templates select="."/>` so the reconstructed constant's children are walked recursively and nested constants are reversed too.

yegor256:
Fixes #5618.

## Problem

The printer dropped the `!` const marker whenever a constant attribute sat inside another constant. For the five environment lookups in `eo-runtime/src/main/eo/mktemp.eo` (bound inside the surrounding `[] >>!` formation, which is itself constant), the round-trip rewrote e.g.

```
getenv "TEMP" > temp!
```

into a bare, non-constant attribute with the raw desugaring left exposed:

```
as-bytes. > temp
  dataized (getenv "TEMP")
```

Dropping the marker is not cosmetic — `!` caches the value once, so a plain attribute re-reads the environment on every access.

## Root cause

`eo-printer/.../print/dataized-to-const.xsl` reverses the `const-to-dataized.xsl` lowering. When rebuilding an outer constant's value it copied the value's children with `<xsl:copy-of>`, which copies the subtree verbatim and never applies templates to it. So the inner `(dataized …).as-bytes` nodes were copied as-is and the template that turns them back into `@const` never fired — the reversal was not recursive, unlike the forward transform (`const-to-dataized.xsl` uses `<xsl:apply-templates>` on children).

## Fix

Change the `<xsl:copy-of select="."/>` at `dataized-to-const.xsl:21` to `<xsl:apply-templates select="."/>` so the reconstructed constant's children are walked recursively and nested constants are reversed too.

## Test

Added `const-nested-in-const.yaml` to the printer round-trip packs, mirroring the mktemp structure (a const attribute inside a const-valued formation). It reproduces the bug: without the fix the printed EO was `as-bytes. (dataized (getenv "TEMP")) > temp`; with the fix it round-trips back to `getenv "TEMP" > temp!`. All 111 `eo-printer` tests pass.

https://claude.ai/code/session_01BpVrSUfPPtKG3wdhrAmXM6

---
_Generated by [Claude Code](https://claude.ai/code/session_01BpVrSUfPPtKG3wdhrAmXM6)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/dataized-to-const.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/dataized-to-const.xsl
@@ -18,7 +18,7 @@
           <xsl:attribute name="name" select="@name"/>
           <xsl:attribute name="const"/>
           <xsl:for-each select="$argument/o">
-            <xsl:copy-of select="."/>
+            <xsl:apply-templates select="."/>
           </xsl:for-each>
           <xsl:if test="eo:has-data($argument)">
             <xsl:value-of select="$argument"/>
--- a/eo-printer/src/test/resources/org/eolang/printer/print-packs/yaml/const-nested-in-const.yaml
+++ b/eo-printer/src/test/resources/org/eolang/printer/print-packs/yaml/const-nested-in-const.yaml
@@ -1,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+origin: |
+  [] > mktemp
+    string > @
+      [] >>!
+        getenv "TEMP" > temp!
+        temp > @
+
+printed: |
+  [] > mktemp
+    string > @
+      [] >>!
+        getenv "TEMP" > temp!
+        temp > @


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (19) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2314](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5621, opened 2026-07-17
Fix lossy UTF-8 decode in StUnhex string printing (#5620)

### LINKED ISSUE #5620: Printer corrupts string literals built from invalid-UTF-8 bytes (StUnhex lossy decode)
When the formatter runs over `eo-runtime/src/main/eo/string.eo`, it rewrites the test `throws-on-taking-length-of-incomplete-n4-byte-character` from `(string F0-9F-98).length` into `"�".length`. The `�` is U+FFFD serialized as the UTF-8 bytes `EF-BF-BD`, so the string's byte content silently changed from `F0-9F-98` to `EF-BF-BD`. The offending code is `eo-printer/src/main/java/org/eolang/printer/StUnhex.java:62-67`.

The `string` shift in `StUnhex` decodes any `Φ.string`-over-`Φ.bytes` with `new String(bytes, StandardCharsets.UTF_8)`, which is the lenient JDK decoder: an incomplete or invalid UTF-8 sequence (here the truncated 4-byte emoji `F0-9F-98`, missing its final `80`) is quietly replaced with U+FFFD. The result does not round-trip — re-parsing `"�"` yields the bytes `EF-BF-BD`, not `F0-9F-98` — so a pure formatting pass changes the program. It also defeats the very case the test is named for: an intentionally incomplete 4-byte character is turned into a valid, complete one, so `.length` no longer exercises the failure path.

Before emitting the quoted literal, verify the decode round-trips: re-encode the decoded string back to UTF-8 and compare it to the original bytes (equivalently, decode through a strict `CharsetDecoder` with `CodingErrorAction.REPORT`). When they differ, skip the conversion and leave the `Φ.string`/`Φ.bytes` structure in place so it prints as an explicit bytes application (`string F0-9F-98`).

yegor256:
Fixes #5620.

## Problem

The `string` shift in `StUnhex` decoded any `Φ.string`-over-`Φ.bytes` with `new String(bytes, StandardCharsets.UTF_8)` — the lenient JDK decoder. An incomplete or invalid UTF-8 sequence (e.g. the truncated 4-byte emoji `F0-9F-98`, missing its final `80`) was silently replaced with `U+FFFD`, serialized back as the UTF-8 bytes `EF-BF-BD`.

As a result a pure formatting pass changed the program: `(string F0-9F-98).length` became `"�".length`, whose byte content is `EF-BF-BD`, not `F0-9F-98`. It does not round-trip, and it defeats the very case `throws-on-taking-length-of-incomplete-n4-byte-character` in `eo-runtime/src/main/eo/string.eo` exercises — an intentionally incomplete character was turned into a valid, complete one.

## Fix

`eo-printer/src/main/java/org/eolang/printer/StUnhex.java`

- Added a `decode` helper that decodes bytes through a strict `CharsetDecoder` with `CodingErrorAction.REPORT` (for both malformed input and unmappable characters), returning `Optional.empty()` when the bytes are not valid UTF-8.
- The `string` shift now only rewrites the node when the strict decode succeeds. When it fails, the conversion is skipped and the `Φ.string`/`Φ.bytes` structure is left in place, so it prints as an explicit bytes application (`string F0-9F-98`).

## Tests

`eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java`

- Added `keepsMalformedBytesUnconverted`, verifying that `F0-9F-98` is left as the `Φ.string`/`Φ.bytes` structure rather than lossily decoded.
- All existing `StUnhexTest` cases still pass (valid strings, glyphs, control-char escaping, empty strings), and the full `eo-printer` suite (110 tests) is green. Qulice (0.27.6) passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_011FCVZRGmTYHHcKaR74SSSu)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
+++ b/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
@@ -8,7 +8,10 @@ import com.yegor256.xsline.Shift;
 import com.yegor256.xsline.StEnvelope;
 import com.yegor256.xsline.StSequence;
 import java.nio.ByteBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.CodingErrorAction;
 import java.nio.charset.StandardCharsets;
+import java.util.Optional;
 import java.util.regex.Pattern;
 import org.eolang.parser.StXnav;
 
@@ -54,17 +57,13 @@ final class StUnhex extends StEnvelope {
         ),
         new StXnav(
             StUnhex.elements("string"),
-            xnav -> xnav.node().setTextContent(
-                String.format(
-                    "\"%s\"",
-                    StUnhex.escape(
-                        new String(
-                            StUnhex.buffer(
-                                StUnhex.undash(xnav.element("o").text().orElse(""))
-                            ).array(),
-                            StandardCharsets.UTF_8
-                        )
-                    )
+            xnav -> StUnhex.decode(
+                StUnhex.buffer(
+                    StUnhex.undash(xnav.element("o").text().orElse(""))
+                ).array()
+            ).ifPresent(
+                decoded -> xnav.node().setTextContent(
+                    String.format("\"%s\"", StUnhex.escape(decoded))
                 )
             )
         )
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2322](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5623, opened 2026-07-17
Restrict trailing-star hybrid to the sole-child `seq *` idiom (#5622)

### LINKED ISSUE #5622: Printer trailing-star corrupts applications with tuples as last argument, causing format idempotency failure
Running the `format` goal over `eo-runtime` corrupts every application whose last argument is a tuple **and** that carries other arguments. For example `getenv.eo` has

```
sm.win32
  "getenv"
  * name
```

The formatter rewrites it to the compact hybrid `sm.win32 "getenv" *` with `name` on the next line. That form is not valid EO: re-parsing it fails with `[22:0] error: 'unexpected deeper-indent line — previous expression is closed for children'`, and after error recovery the tuple element `name` is dropped — the `*` becomes `Φ.tuple.empty`. So a second `format` pass converges to

```
sm.win32
  "getenv"
  *
```

silently deleting the argument. The format is therefore non-idempotent (9 of 139 sources still diverge after one `-Deo.autoFix` pass) and, worse, lossy. The code lives in `eo-printer/src/main/java/org/eolang/printer/Pretty.java` — `starred()` at line 348, gated by `tuply()` at line 544.

The trailing-star hybrid (#5594, #5615) is only round-trippable when the `*` is the head's **sole** applied argument, the genuine `seq *` idiom: there a bare trailing star absorbs the indented siblings into the tuple. When the star follows other arguments on the line (`sm.win32 "getenv" *`), the parser reads the line as a complete application with an empty tuple and rejects the indented child. `tuply()` only checks that the last child is a star; it never checks that it is the only child, so `starred()` fires for the multi-argument case and emits a shape the parser cannot read back. Nine runtime sources are hit (`getenv`, `file`, `socket`, `string/printf`, `string`, `sm/posix`, `sm/win32`, `directory`, `string/joined`); `socket.eo` loses an entire `code.` subtree, not just a leaf.

Restrict the hybrid to the true `seq *` case: require the star to be the head's only child (for instance `this.children.size() == 1` in `tuply()`, or an equivalent guard in `starred()`). When a preceding argument is present, fall back to the ordinary rendering, which keeps the tuple as an inline `* name` child that round-trips cleanly.

yegor256:
Fixes #5622.

## Problem

The printer's trailing-star hybrid (`Pretty.starred()`, gated by `Pretty.Node.tuply()`) rendered *any* application whose last child was a non-empty star as `head *` on one line with the tuple elements indented beneath. `tuply()` only checked that the last child was a star — never that it was the *only* child.

When a star followed other arguments on the line, e.g. `getenv.eo`:

```
sm.win32
  "getenv"
  * name
```

it was rewritten to `sm.win32 "getenv" *` with `name` indented below. That form is not valid EO: re-parsing reads the line as a complete application with an empty tuple and rejects the indented child (`unexpected deeper-indent line — previous expression is closed for children`). After error recovery the tuple element `name` is dropped and `*` becomes `Φ.tuple.empty` — so a second `format` pass silently deletes the argument. The format was therefore both **non-idempotent** and **lossy** (9 of 139 runtime sources diverged; `socket.eo` lost an entire `code.` subtree).

## Fix

The bare trailing `*` hybrid only round-trips for the genuine `seq *` idiom, where the star is the head's **sole** applied argument and absorbs exactly the indented siblings into the tuple. `tuply()` now requires `this.children.size() == 1`, so the hybrid fires only in that case. Multi-argument applications fall back to the ordinary vertical rendering, keeping the tuple as an inline `* name` child that round-trips cleanly.

`starred()` and `Pretty.Node.glued()` are simplified accordingly — with no leading arguments possible before the star, the argument-inlining path and `glued()`'s `args` parameter are removed.

## Verification

- Added print-pack `trailing-star-after-args.yaml` reproducing the `getenv` case; it now prints identically to its origin and re-prints to the same EO (idempotent, lossless).
- The `seq *` idiom regression is covered by the existing `seq-star-idiom.yaml` pack.
- `mvn -pl eo-printer test` — all 92 `XmirTest` cases pass (0 failures).
- `mvn -pl eo-printer verify` (qulice, checkstyle, PMD, spotbugs) passes.

https://claude.ai/code/session_016NbcGsUKEHV8tGfqGxm2GH

---
_Generated by [Claude Code](https://claude.ai/code/session_016NbcGsUKEHV8tGfqGxm2GH)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -332,9 +332,14 @@ final class Pretty {
      * ambiguity.</p>
      *
      * <p>It applies only to a plain (non-formation, non-reversed)
-     * application whose final child is a non-empty, unnamed star; any
-     * preceding arguments must inline. The elements are always laid out
-     * vertically, so the genuinely ambiguous fully-horizontal
+     * application whose sole child is a non-empty, unnamed star. The star
+     * must be the only child: a bare trailing {@code *} absorbs the
+     * indented siblings into the tuple, so the shape round-trips only when
+     * nothing else shares the head's line. A preceding argument
+     * ({@code sm.win32 "getenv" *}) would make the parser read a complete
+     * application with an empty tuple and reject the indented child, so
+     * {@link #tuply()} bars that case (issue #5622). The elements are always
+     * laid out vertically, so the genuinely ambiguous fully-horizontal
      * {@code head * a b} — the before-star territory — is never produced
      * here. The result is only a candidate: {@link #shaped} keeps it only
      * when its penalty beats the plain vertical and horizontal renderings,
@@ -348,11 +353,8 @@ final class Pretty {
     private Optional<String> starred(final Pretty.Node node, final int indent) {
         Optional<String> result = Optional.empty();
         if (node.tuply()) {
-            final List<Pretty.Node> kids = node.children;
-            result = Pretty.inlined(kids.subList(0, kids.size() - 1)).map(
-                args -> this.vertical(
-                    kids.get(kids.size() - 1).glued(node, args), indent
-                )
+            result = Optional.of(
+                this.vertical(node.children.get(0).glued(node), indent)
             );
         }
         return result;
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2342](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5625, opened 2026-07-17
Bar trailing-star hybrid for dotted method dispatch heads (#5624)

### LINKED ISSUE #5624: `"...".printf *` corruption persists for `string.eo` and `socket.eo` in printer trailing-star hybrid
Follow-up to #5622. That fix stopped the trailing-star hybrid firing for reversed and multi-argument nodes, which cleared seven of the nine affected runtime sources. Two remain — `string.eo` and `socket.eo` — because the hybrid is also reachable through a second path.

`string.eo` on master has

```
"Start index + length to slice (%d) is out of string bounds".printf
  * (nstart.plus nlen)
```

The formatter rewrites it to `"...".printf *` with `nstart.plus nlen` on the next line. That fails to parse — `[88:0] error: 'unexpected deeper-indent line — previous expression is closed for children'` — and after error recovery the `*` becomes `Φ.tuple.empty`, dropping the element. A second `format` pass converges to source with the argument gone. `socket.eo`'s `"WSA error code %d".printf` loses its entire indented `code. …` subtree the same way, so the format is again non-idempotent and lossy.

The root path is in `eo-printer/src/main/java/org/eolang/printer/Pretty.java`. A dispatch on a data receiver (`"…".printf`) is stored reversed, so `tuply()` (with the #5622 guards) correctly rejects the raw node. But `suffixed()` (line 159) rebuilds that dispatch as a non-reversed, single-child node — the star being its only child — and `shaped()` evaluates `starred()` on that suffixed alternative (lines 108–140). `tuply()` now passes, the hybrid fires, and it emits `"literal".method *`, which does not round-trip: a bare trailing `*` is absorbed by the parser only after a plain leading application (`seq *`, `map *`, `switch *` all survive), not after a dotted method dispatch.

Bar the hybrid for suffix-folded dispatch heads. For instance, mark the node that `suffixed()` builds as ineligible for the trailing-star form (a flag carried on the node), or have `tuply()`/`starred()` reject a head whose base is a dotted method dispatch, so `"literal".method` with a tuple argument falls back to the ordinary `* elem` child that round-trips.

yegor256:
Fixes #5624.

## Problem

Follow-up to #5622. That fix stopped the trailing-star hybrid firing for reversed and multi-argument nodes, clearing seven of nine affected runtime sources. Two remained — `string.eo` and `socket.eo` — because the hybrid was reachable through a second path.

A data-receiver dispatch (`"…".printf`) is stored as a *reversed* head, so `tuply()` (with the #5622 guards) correctly rejected the raw node via its `!reversed` guard. But `suffixed()` rebuilds that dispatch as a **non-reversed, single-child** node — the star being its only child — and `shaped()` weighs `starred()` on that suffixed alternative. There, `tuply()` passed and the hybrid fired, emitting `"literal".method *`.

A bare trailing `*` is absorbed by the parser only after a plain leading application (`seq *`, `map *`, `switch *` all survive), **not** after a dotted method dispatch: there it reads as a complete application with an empty tuple and rejects the indented elements. On re-parse the argument was dropped — `format` became non-idempotent and lossy:

- `string.eo`: `"… out of string bounds".printf * (nstart.plus nlen)` → the `*` element was dropped.
- `socket.eo`: `"WSA error code %d".printf` lost its entire indented `code.` subtree.

## Fix

`tuply()` now also requires a plain, non-dotted base, extracted into a new `gluer()` helper (`!abstractt && !reversed && base has no '.'`). A dotted dispatch head — whether the suffix-folded one that `suffixed()` builds, or any genuine method dispatch — falls back to the ordinary `* elem` child, which round-trips cleanly.

## Verification

- Added `trailing-star-after-dispatch.yaml`, modelling the exact `string.eo` (`* (nstart.plus nlen)`, `* nstart`) and `socket.eo` (named-attribute printf with a dispatch element) shapes. It fails on `master` (both the exact-output and re-parse-idempotency checks) and passes with this change.
- Full `eo-printer` suite green (118 tests, `XmirTest` 94).
- Qulice clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01T61zh255BoqbZDMKS127eV)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -547,12 +547,37 @@ final class Pretty {
          * child, so the hybrid must not fire (issue #5622). The single child
          * must itself be a gluable star (see {@link #stars()}).</p>
          *
+         * <p>The head must also be a plain base, not a dotted method dispatch
+         * ({@code "literal".printf}, {@code 5.plus}). A bare trailing
+         * {@code *} is absorbed by the parser only after a plain leading
+         * application ({@code seq *}, {@code map *}, {@code switch *}); after
+         * a method dispatch it reads as a complete application with an empty
+         * tuple and rejects the indented elements. A data-receiver dispatch
+         * is stored reversed and so already fails the {@code !reversed} guard,
+         * but {@link #suffixed} rebuilds it as a non-reversed, single-child
+         * node whose base is exactly such a dispatch, and that alternative is
+         * weighed for the hybrid too — barring a dotted base here keeps it,
+         * and any genuine dotted dispatch, on the ordinary {@code * elem}
+         * child that round-trips (issue #5624).</p>
+         *
          * @return True when the trailing-star hybrid form is applicable
          */
         boolean tuply() {
-            return !this.abstractt && !this.reversed
+            return this.gluer()
                 && this.children.size() == 1
                 && this.children.get(0).stars();
+        }
+
+        /**
+         * Whether this node is a plain application head onto which a
+         * trailing {@code *} may be glued: not a formation, not a reversed
+         * dispatch, and not a dotted method dispatch. See {@link #tuply()}
+         * for why a dotted base ({@code "literal".printf}, {@code 5.plus})
+         * is excluded (issue #5624).
+         * @return True when the head can carry a glued trailing star
+         */
+        boolean gluer() {
+            return !this.abstractt && !this.reversed && this.base.indexOf('.') < 0;
         }
 
         /**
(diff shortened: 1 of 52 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (544 >= 100); -8 for way too many hits-of-code (544 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2346](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5627, opened 2026-07-17
Fix #5626: parse compact-tuple φ in inline-phi formations (`seq * &gt; [m]`)

### LINKED ISSUE #5626: Parser drops the tuple in `seq * > [m]`: compact-tuple LHS rejected in inline-phi formations
An only-phi formation whose φ is a compact tuple has no compact spelling: `seq * > [m]` does not parse. The parser reads the `*` as a complete empty-tuple horizontal argument, marks the line closed for children, and silently drops the tuple's elements. Given

```
malloc.of > @
  5
  seq * > [m]
    copied a
    m
```

the parser emits two `unexpected deeper-indent line — previous expression is closed for children` errors and produces `Φ.seq` applied to `Φ.tuple.empty` — `copied a` and `m` are gone. The verbose sibling round-trips fine:

```
seq > [m]
  *
    copied a
    m
```

The divergence lives in `eo-parser/src/main/java/org/eolang/parser/LnOnlyPhi.java:212` (`bare()`): it parses the inline-phi LHS as `head + readChain() + readArgs()`, so for `seq *` the `readArgs()` call swallows the `*` as a horizontal STAR arg, `bare()` returns false, the φ is marked `HORIZONTAL_COMPLETED` (line 255), and `emitPhi()` emits `seq (Φ.tuple.empty)` before closing the level.

This is a composition gap between two features that each work alone. §3.9 (R-3.9.5) makes `seq *` a compact-tuple whose trailing `*` absorbs deeper-indent lines as tuple elements — `seq * > @` round-trips today. §3.10/§4.5 (R-3.10.1) makes `seq > [m]` an only-phi formation whose bare φ absorbs deeper-indent lines. But R-3.10.6 limits the inline-phi LHS to `head`/`hmethod`/`happlication` and omits compact-tuple, so their intersection has no compact form. The meaning is unambiguous — an only-phi formation binds only φ, so the deeper lines can only be the star's elements — yet the printer must fall back to the taller verbose shape for exactly this case (cf. the closed #5594 and #5615, both printer-side, which fixed the `> @` and bare-φ cases but not this one).

Fix: in `LnOnlyPhi`, detect a trailing `*N` on the LHS (a compact-tuple head, R-3.9.1) and keep the φ open in that case — mirroring §3.9 — so the deeper-indent lines are absorbed into the tuple as `@star` elements instead of `bare()` reading the `*` as a completed empty-tuple argument. Add compact-tuple to the permitted inline-phi LHS forms in R-3.10.6.

yegor256:
Closes #5626.

## Problem

An only-phi formation whose φ is a compact tuple had no compact spelling — `seq * > [m]` did not parse. Given:

```
malloc.of > @
  5
  seq * > [m]
    copied a
    m
```

the parser read the `*` as a completed empty-tuple horizontal argument, marked the line closed for children, and silently dropped the tuple's elements: it emitted two `unexpected deeper-indent line — previous expression is closed for children` errors and produced `Φ.seq` applied to `Φ.tuple.empty` — `copied a` and `m` were gone. The verbose sibling round-trips fine:

```
seq > [m]
  *
    copied a
    m
```

The divergence lived in `LnOnlyPhi.bare()`, which parsed the inline-phi LHS as `head + readChain() + readArgs()`, so for `seq *` the `readArgs()` call swallowed the `*` as a horizontal STAR arg and the φ was marked `HORIZONTAL_COMPLETED`. This was a composition gap: §3.9 makes `seq *` a compact-tuple, and §3.10/§4.5 makes `seq > [m]` an only-phi formation, but R-3.10.6 limited the inline-phi LHS to `head`/`hmethod`/`happlication` and omitted compact-tuple, so their intersection had no compact form.

## Fix

`LnOnlyPhi` now detects a trailing `*N` compact-tuple marker on the inline-phi LHS (R-3.9.1 + R-3.10.6). When present it:

- strips the `*N` from the φ slot so the φ emits just the head (`seq`);
- keeps the φ `OPEN` instead of `HORIZONTAL_COMPLETED`;
- flags the `ONLY_PHI_FORMATION` level (`Level.star()`) with the `N` count.

The flag lets the existing `Φ.tuple` wrapper hooks in `Eo` — `beforeChild` (opens the `Φ.tuple star=''` wrapper after the first `N` children) and `closeCompactTuple` (closes it / synthesises the empty wrapper) — fire for the only-phi level exactly as they do for a bare `LnCompactTuple` head, mirroring §3.9. The wrapper closes before the φ, which closes before the formation, preserving XMIR nesting.

The compact form now produces XMIR structurally identical to its verbose sibling, with no errors.

## Changes

- `LnOnlyPhi.java` — detect the trailing `*N` marker (`compactMarker` / `trailingStar` / `starCount`), keep the φ open, and mark the level compact.
- `Level.java` — add the `star()` flag + `markStar()` for an only-phi φ that is a compact tuple, reusing the existing `count` / `children` / `tupled` fields.
- `Eo.java` — extend the `beforeChild` and `checkOnClose` hooks to treat a `star()`-flagged only-phi level like a compact tuple.

## Tests

- Unit tests in `LnOnlyPhiTest` — φ stays `OPEN` for a compact-tuple LHS, the level is flagged with its `N` count, and the `*` is not emitted as a horizontal argument.
- `eo-syntax` YAML packs — the issue's `malloc.of` example (compact φ as an argument) and a named `seq * > [m] > result` variant both parse with no errors and preserve the tuple elements.

Full `eo-parser` suite (1476 tests) and the qulice profile both pass.

https://claude.ai/code/session_01E8jyU8r8bdhHCUypkSS4fF

---
_Generated by [Claude Code](https://claude.ai/code/session_01E8jyU8r8bdhHCUypkSS4fF)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (199 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2350](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5628, opened 2026-07-17
#5613: reject over-precise INT/FLOAT literals

### LINKED ISSUE #5613: `Parser fails to report error for number literals with excessive precision`
The parser accepts a number literal carrying more precision than an IEEE-754 double can hold and silently rounds it, rather than reporting an error. In `Emissions.openValue()` (`eo-parser/src/main/java/org/eolang/parser/Emissions.java:117-122`), a `FLOAT`/`INTEGER` value is turned into 8 bytes with `Double.parseDouble(value.raw())`, discarding any digits the double cannot represent. For instance `eo-runtime/src/main/eo/e.eo:12` holds `2.7182818284590452354 > e`; that literal parses to the exact same double as `2.718281828459045`, so the trailing `2354` is dropped without a word and the printer later re-emits `2.718281828459045`.

This is wrong because the source then advertises precision the type cannot keep, and the loss is silent — the opposite of failing fast. The `HEX` branch just below (`Emissions.java:123-134`) already sets the precedent: it throws a `ParseError` for a literal it cannot represent ("hexadecimal literal is out of range"), so a decimal literal that cannot round-trip deserves the same treatment.

Add a guard to the `FLOAT`/`INTEGER` branch: after `Double.parseDouble`, render the double back to its shortest decimal (the same canonical form the printer's `StUnhex` produces) and compare it with the literal; when they differ, throw a `ParseError` that names the offender and its canonical form, e.g. `2.7182818284590452354 is over-precise, write 2.718281828459045 instead`. Comparing against the round-trip form is what keeps legitimate literals such as `0.1` — not exactly representable yet already its own shortest form — accepted, while rejecting only the ones that carry dead digits.

Shaurya2k06:
Fixes #5613.

## Problem

`Emissions.openValue()` turned FLOAT/INTEGER literals into IEEE-754 bytes via `Double.parseDouble` and silently dropped digits the double cannot keep. For example `2.7182818284590452354 > e` parsed as the same double as `2.718281828459045`, so the trailing `2354` vanished without an error — the opposite of failing fast. HEX already rejects values it cannot represent (`hexadecimal literal is out of range`).

## Fix

After parsing, compare the source spelling to the round-trip form:

- FLOAT → `Double.toString(parsed)`
- INTEGER → StUnhex-style whole-number form (`Long.toString` / `BigDecimal` for large wholes)

When they differ, throw `ParseError`: `<literal> is over-precise, write <canonical> instead`. Legitimate literals such as `0.1` and `0.0` still pass because they already match `Double.toString`.

Also updated in-tree sources/fixtures that were not already in canonical form (`e.eo`, `pi.eo`, sci-notation case, etc.), and documented the message in `PARSER_SPEC.md` §9.9.

## Tests

- `eo-typos/over-precise-float.yaml` — rejects `2.7182818284590452354`
- `LnApplicationTest` — over-precise FLOAT/INTEGER messages; accepts `0.1`
- Syntax packs updated to canonical spellings
- `mvn -pl eo-parser,eo-runtime clean install -Pqulice` (JDK 21) passes


yegor256:
@maxonfjvipon please, review this one

eo-parser/src/main/java/org/eolang/parser/Emissions.java
```diff
@@ -295,6 +291,60 @@ static String unescapeBody(final String inner) {
...
+        final Emit emit, final String name, final Value value, final int line
+    ) {
+        final double parsed = Double.parseDouble(value.raw());
+        final String canonical;
+        if (value.kind() == Value.Kind.INTEGER) {
+            canonical = Emissions.canonicalInteger(parsed);
+        } else {
+            canonical = Double.toString(parsed);
+        }
+        if (!value.raw().equalsIgnoreCase(canonical)) {
```
morphqdd: This compares the raw literal text against Double.toString's canonical spelling, and rejects on any difference at all, not just on actual precision loss. That's stricter than what the issue and the error message describe.

Two concrete cases that get wrongly rejected even though no precision is lost:

1. An explicit `+` sign: `+42 > x`. `Tokens.readInt()` intentionally accepts a leading `+`, and `9007199254740993 > x` is not the only integer test added here, but no test in this PR covers a `+`-signed literal. I added a throwaway yaml test case with `+42 > x` and ran `EoSyntaxTest` against this branch: it fails with `error: '+42 is over-precise, write 42 instead'`. `Double.parseDouble("+42")` and `Long.toString(42L)` differ only in spelling, not in the value they represent.

2. A trailing zero: `1.50 > x` would fail the same way (`Double.toString(1.5)` is `"1.5"`, not `"1.50"`), for the same reason, even though `1.50` and `1.5` are the same double.

The project's own PARSER_SPEC.md (see the comment on that file) still documents `+1foo` as an example of a legal-prefix-followed-by-invalid-digits case, which only makes sense if a bare `+1` is meant to parse. Since this compares raw spelling instead of significant digits, this looks like it needs to compare the actual value/precision, not the full string, or the intent needs to change to "canonical spelling only" with the docs, message text, and PARSER_SPEC updated to match, plus a test added for a `+`-signed literal either way.

eo-parser/src/main/java/org/eolang/parser/Emissions.java
```diff
@@ -295,6 +291,60 @@ static String unescapeBody(final String inner) {
...
+     * @return Shortest integer spelling for {@code num}
+     */
+    private static String canonicalInteger(final double num) {
+        final String str;
+        if ("-0.0".equals(Double.toString(num))) {
+            str = "-0";
+        } else if (Math.abs(num) < 0x1p63) {
+            str = Long.toString((long) num);
+        } else {
+            str = Double.toString(num).replace('E', 'e');
```
morphqdd: This fallback path (very large integers, magnitude >= 2^63) lowercases the exponent marker, but everywhere else in this PR the canonical form uses uppercase E: the FLOAT branch a few lines up returns `Double.toString(parsed)` untouched, and the ~20 `.eo` fixture files this PR reformats (`pi.eo`, `number/*.eo`, etc.) all end up with uppercase E after the rewrite. Since the comparison itself is case-insensitive this doesn't cause a false rejection, but the suggested replacement text this branch puts into the error message would itself fail a FLOAT canonical check elsewhere in the same codebase. No test exercises this branch (the one integer test added, `9007199254740993`, is well under 2^63), so this has been sitting untested.

eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
```diff
@@ -601,6 +601,62 @@ void rejectsOversizedHexHead() {
...
+        Assertions.assertDoesNotThrow(
+            () -> {
+                final Emit emit = new Emit();
+                new LnApplication(new Span("1.0e30 > x", 1))
+                    .into(new Stack(), new Globals(), emit);
+                emit.close();
+            },
+            "lowercase e must match Double.toString's E so StUnhex output re-parses"
+        );
+    }
```
morphqdd, line 667: Good set of tests for the rejection and acceptance cases, but none of them cover a `+`-signed literal (see the comment on Emissions.java:314). Worth adding one here once the underlying check is settled, either an acceptance test if `+42` is meant to stay legal, or an explicit rejection test with a message that says something other than "over-precise" if the intent really is canonical-spelling-only.

eo-parser/PARSER_SPEC.md
```diff
@@ -1326,6 +1326,7 @@ R-9.9.1. Every error condition in this spec has a single canonical text — **in
 | More than one space between meta parts (R-3.2.4) | `meta parts must be separated by exactly one space` |
 | Test attribute name is `@` (PHI) instead of NAME (R-6.3.5) | `test attribute name must be an identifier, not @` |
 | Leading-zero in integer literal (R-9.8.1, e.g., `007`) | `integer literal must not have leading zeros` |
+| Decimal `INT`/`FLOAT` literal that does not match its IEEE-754 round-trip form (e.g., `2.7182818284590452354`) | `<literal> is over-precise, write <canonical> instead` (literal and canonical substituted) |
```
morphqdd: This wording ("does not match its IEEE-754 round-trip form") describes the intent well, but the actual implementation in Emissions.java checks spelling equality, not round-trip equality. `+42` and `1.50` both round-trip through IEEE-754 with zero loss, yet the current code rejects them. Once the Emissions.java behavior is settled, this line should describe whatever the final rule actually is.

morphqdd (changes requested):
I checked out this branch locally and confirmed the main issue with an actual test run, not just by reading the diff. Details in the inline comments below. The core mechanism this PR adds is fine, but the check it does is broader than what the issue and the error message describe, and that catches valid input.

eo-parser/src/main/java/org/eolang/parser/Emissions.java
Shaurya2k06: Agreed — that was stricter than the issue. The check now accepts a literal when its `BigDecimal` equals either the double's exact binary value (`new BigDecimal(parsed)`) or `Double.toString`'s shortest form (`BigDecimal.valueOf(parsed)`). So `+42` and `1.50` pass, while dead digits like `2.718…2354` still fail.

eo-parser/src/main/java/org/eolang/parser/Emissions.java
Shaurya2k06: Fixed: the `|n| ≥ 2⁶³` suggestion now uses uppercase `E` from `Double.toString` (no `replace`), and there is a unit test on `9223372036854775807` that exercises this arm.

eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
Shaurya2k06: Added `acceptsPlusSignedIntegerHead` (`+42`) and `acceptsTrailingZeroFloatHead` (`1.50`) as acceptance tests.

eo-parser/PARSER_SPEC.md
Shaurya2k06: Updated the §9.9 row to describe value/precision (dead digits), and to note that alternate spellings of the same value (`+42`, `1.50`) are accepted.

maxonfjvipon (commented):
Thanks for this — the parser fix itself is exactly right: `overPrecise` is a value check (rejects only dead digits, accepts any spelling that round-trips), and the `PARSER_SPEC` entry, the `over-precise-float` fixture, and the new tests all read well.

My concern is scope. I ran every rewritten literal through the `overPrecise` logic, and only three are actually over-precise and needed changing:

- `e.eo` — `2.7182818284590452354`
- `pi.eo` — `3.14159265358979323846`
- `tuple/hash-code.eo:50` — `123456789012345678`

Every other `.eo`/`.yaml` edit is a literal the check already accepts — `0.0001`, `0.000000001`, `7.33e-10`, `100000000000000000000`, `-9223372036854775808`, `99999999.99`, `42.42e42`, and the rest. They are bit-identical to their rewrites (e.g. `0.0001` and `1.0E-4` both encode `3F-1A-36-E2-EB-1C-43-2D`), so those are cosmetic — change for the sake of change. Could you revert them and keep the PR to the three genuine cases plus the fix? It keeps the diff honest about what actually changed behavior.

Details inline.

eo-parser/src/test/resources/org/eolang/parser/eo-syntax/all-data-types.yaml
```diff
@@ -26,12 +26,12 @@ input: |
   [] > main
     01- > bytesX
     00-02-4F-1A-FF-01 > bytesX
```
maxonfjvipon, line 28: Please keep the full decimal expansions here. These cases exist to prove the parser reads a huge literal into the exact boundary bytes (`Double.MAX_VALUE`, smallest normalized/denormalized) — they all round-trip and are accepted unchanged, so rewriting them to `1.7976931348623157E308` is not required and it deletes the very coverage the case was written for.

eo-runtime/src/main/eo/number/degrees.eo
```diff
@@ -25,31 +25,31 @@
         minus.
           degrees pi
           180
-      0.0001
+      1.0E-4
```
maxonfjvipon: This one is cosmetic — `0.0001` and `1.0E-4` are the same double (`3F-1A-36-E2-EB-1C-43-2D`), so the check accepts `0.0001` as-is. Same across `radians`, `exp`, `ln`, `power`, `sqrt`, `integral`, `floor`, `scanf`, `malloc`, `i64`, `number`. Let's leave the readable tolerances alone.

eo-parser/src/test/resources/org/eolang/parser/eo-syntax/numbers.yaml
```diff
@@ -8,8 +8,8 @@ asserts:
 input: |
   [] > main
     3.1415 > x1
-    7.33e-10 > x2
-    -2.4E3 > x3
+    7.33E-10 > x2
```
maxonfjvipon: This goes the wrong way for me — if we standardize exponent casing, I would rather accept only lowercase `e` (`7.33e-10`); the lowercase letter stays visually distinct from the digits, the capital blends in. Note it couples with the printer: `StUnhex.number` prints `Double.toString`'s uppercase `E`, so we would also need it to lower-case the exponent, otherwise printed programs will not re-parse. Since it spans parser + printer, it is probably cleanest as its own issue rather than folded in here — but at minimum, let's not flip these to `E`.

eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
```diff
@@ -370,8 +370,7 @@ void acceptsSignedIntegerArg() {
     void rejectsIntegerArgWithLeadingZero() {
         Assertions.assertThrows(
             ParseError.class,
-            () -> new LnApplication(new Span("foo 07", 1))
-                .into(new Stack(), new Globals(), new Emit()),
+            () -> LnApplicationTest.parseLine("foo 07"),
```
maxonfjvipon: Introducing `parseLine` for the new tests is nice, but rewiring the four unrelated existing tests onto it (leading-zero, bad unicode, unclosed phi, oversized hex) is a bit beyond this ticket. Not a blocker — just noting it widens the diff.

eo-parser/src/test/resources/org/eolang/parser/eo-syntax/all-data-types.yaml
Shaurya2k06: Restored the full decimal expansions — they round-trip under the value check, so the rewrite was unnecessary.

eo-runtime/src/main/eo/number/degrees.eo
Shaurya2k06: Agreed. Reverted the cosmetic tolerance rewrites across degrees/radians/exp/ln/power/sqrt/integral/floor/scanf/malloc/i64/number/printf — only e, pi, and hash-code keep real over-precise fixes.

eo-parser/src/test/resources/org/eolang/parser/eo-syntax/numbers.yaml
Shaurya2k06: Reverted — left `7.33e-10` and `-2.4E3` as they were. Exponent-casing standardization can be its own issue with the printer.

eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
Shaurya2k06: Undid the rewiring — the four existing tests are back to their original form; `parseLine` is only used by the new over-precision tests.

maxonfjvipon (approved):
Scope is right now — the fix plus the three literals that genuinely lose precision (`e`, `pi`, `hash-code`). The value-based `overPrecise` check reads well, the fixture and tests are solid, and CI is green across all platforms. Thanks for trimming the cosmetic churn. (Exponent casing will be tracked in its own issue, since it also touches the printer.)

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+1309](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+811](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@shaurya2k06 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (193 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+144](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5636, opened 2026-07-18
Fix #5635: collapse inline-phi hybrid when the one-liner overflows the width

### LINKED ISSUE #5635: Inconsistent rendering of inline-phi collapse in Xmir.toEO() when one-liner overflows width limit
`Xmir.toEO()` renders the compact inline-phi collapse inconsistently. In `Pretty.phi()` (`eo-printer/src/main/java/org/eolang/printer/Pretty.java:289-297`) the fully-flat one-liner and the hybrid collapse are mutually exclusive: `if (value.isPresent()) { <one-liner> } else if (applied && unnamed) { <hybrid> }`. So when the φ decoratee fits on one line (`value.isPresent()`) but that line overflows the width limit and loses to the vertical form, the hybrid collapse is never even generated, and the formation falls back to the verbose `<phi> > @` shape.

In `eo-runtime/src/main/eo/number/cosine.eo` the nine `tests-cosine-*` attributes are all the same shape — an only-φ formation whose φ is a `.lt` reversed dispatch — yet the printer renders them two ways. `tests-cosine-is-even-function` collapses to `lt. ++> …` because its decoratee holds `neg. (pi.div 3)`, which `flat()` rejects (`Pretty.java:408`), so the hybrid path fires. The parallel `tests-cosine-of-pi-thirds-is-half` and seven siblings stay verbose `++> …` / `lt. > @` because their decoratee inlines fully but too wide. The collapsed form is valid EO — identical XMIR per R-3.10.8 — and cheaper (penalty 252 vs 297), so the divergence is a formatting defect. This is the residual case #5594 (closed) left open: that fix added the hybrid only for the `value.isEmpty()` branch.

Stop gating the hybrid behind `else`: when `applied && unnamed`, build the hybrid candidate too and return the lower-penalty of it and the flat one-liner (or hand both to `shaped()`), so the width check can drop the overflowing one-liner in favour of the hybrid.

yegor256:
Closes #5635.

## Problem

In `Pretty.phi()` the fully-flat inline-φ one-liner and the hybrid collapse were mutually exclusive:

```java
if (value.isPresent()) { /* flat one-liner */ }
else if (applied && unnamed) { /* hybrid */ }
```

So when a φ decoratee inlined on one line (`value.isPresent()`) but that line overflowed the width limit and lost to the vertical form, the hybrid collapse was never generated, and the formation fell back to the verbose ` > @` shape.

This is the residual case #5594 left open (that fix only added the hybrid for the `value.isEmpty()` branch). It showed up in `number/cosine.eo`, where nine identically-shaped `tests-cosine-*` attributes rendered two different ways: `tests-cosine-is-even-function` collapsed (its decoratee is rejected by `flat()`), while `tests-cosine-of-pi-thirds-is-half` and siblings stayed verbose because their decoratee inlined fully but too wide.

## Fix

Stop gating the hybrid behind the one-liner's absence. The flat one-liner is now kept only while it fits the `WIDTH` limit; once it overflows (or cannot be built), the hybrid is produced and `shaped()` weighs it against the plain vertical rendering — so the overflowing one-liner is dropped in favour of the cheaper hybrid instead of the verbose fallback.

The one-liner is still preferred over the hybrid whenever it fits, even when the multi-line hybrid would score a lower penalty (bracket cost), preserving existing behaviour.

`WIDTH` is now cached as a `width` field (like the existing `tab`) so the overflow check stays a single expression and `phi()` keeps its complexity within limits.

## Tests

- New print-pack `inline-phi-overflow.yaml` reproduces the exact `cosine.eo` case: the decoratee inlines but the one-liner overflows, and it now collapses to the hybrid `lt. ++> …`.
- `void-local-handle.yaml` updated: its `reclast` formation now collapses to the valid hybrid `if. > [tup] > reclast` (same shape as the existing `hybrid-phi-formation.yaml`), while still exercising the `? >> name` void-handle behaviour it targets.

All 132 `eo-printer` tests pass, and qulice is clean.

https://claude.ai/code/session_015X51NGYQPMde7ucsSmgF3T

---
_Generated by [Claude Code](https://claude.ai/code/session_015X51NGYQPMde7ucsSmgF3T)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -51,6 +51,11 @@ final class Pretty {
     private final String tab;
 
     /**
+     * The column past which a line overflows, the {@code WIDTH} weight.
+     */
+    private final int width;
+
+    /**
      * Ctor, using the default penalty weights.
      * @param element The {@code <eo>} element
      */
@@ -62,13 +67,16 @@ final class Pretty {
      * Ctor.
      * @param element The {@code <eo>} element
      * @param config The overridden weights; absent keys use their defaults
-     * @checkstyle ConstructorsCodeFreeCheck (10 lines)
+     * @checkstyle ConstructorsCodeFreeCheck (12 lines)
      */
     Pretty(final Xnav element, final Map<PenaltyKey, Integer> config) {
         this.root = element;
         this.weights = config;
         this.tab = " ".repeat(
             config.getOrDefault(PenaltyKey.STEP, PenaltyKey.STEP.fallback())
+        );
+        this.width = config.getOrDefault(
+            PenaltyKey.WIDTH, PenaltyKey.WIDTH.fallback()
         );
     }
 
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2434](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #464, opened 2021-12-01
add saving array length in bitwise operations

IngeniariusSoftware:

PR #5637, opened 2026-07-18
#5484: reject bytes.and/or/xor on operands of different lengths

### LINKED ISSUE #5484: `bytes.and`, `bytes.or`, `bytes.xor` incorrectly handle operands of different lengths
## What happens

`BytesRaw.and`, `BytesRaw.or`, and `BytesRaw.xor` (`eo-runtime/src/main/java/org/eolang/BytesRaw.java:41-68`) all iterate to `Math.min(first.length, second.length)` and then always return a `BytesOf(first)` — of length `first.length`, regardless of `second`'s size. This silently produces wrong results when the two operands differ in length; the direction of the wrongness is asymmetric between the three operations.

For a byte-level view, take the receiver as `first = this.take()` and the argument as `second = other.take()`.

**AND, when `first.length > second.length` — trailing bytes of `first` are *not* zeroed.**
`FF-FF-FF.and 0F` (3 bytes vs 1 byte) yields `0F-FF-FF`, but a bitwise AND with a zero-extended operand would be `00-00-0F`; a length-preserving truncated AND would be `0F`. The current answer matches neither. Concretely, the upper two bytes of `first` are `x & (nothing)` — the code just leaves them as they were, which happens to be equivalent to `first & FF-FF-…-FF` at the tail, i.e. an implicit *one*-extension, the opposite of what `&` with a missing operand should do.

**OR / XOR, when `first.length < second.length` — trailing bytes of `second` are dropped.**
`01-02.or 03-04-05-06` (2 bytes vs 4 bytes) yields `03-06`; a zero-extended OR would be `03-04-05-06`; a length-preserving OR would be `03-06` *padded* to 4 bytes. The current answer silently discards `05-06`, i.e. bytes visible to the user of the API vanish. `xor` has the same shape.

Nothing in `bytes.eo:81-85` (`[b] > and /bytes` / `> or /bytes` / `> xor`) documents any of this — none of the docstrings say "same length required", "zero-extend", or "truncate to min". The existing tests (`bytes.eo:220-286`, `380-435`; `BytesOfTest.java`) all operate on operands of identical length, so this mismatch is never exercised.

This came up in [#464](https://github.com/objectionary/eo/issues/464) back in 2021 with the same class of operations — `@yegor256` said there: *"if two arrays are of different lengths, we must throw an exception"* — but the current implementation neither throws nor zero-extends. The issue was closed as stale without a fix. Filing this fresh because the pre-existing `BytesRaw` is still on the "silently drop data" path.

## Failure scenarios

Minimal Java repro against the current `main`:

```java
// AND drops trailing bytes of first as if they were `x & 1`
Bytes a1 = new BytesOf(new byte[]{(byte) 0xFF, (byte) 0xFF, (byte) 0xFF});
Bytes b1 = new BytesOf(new byte[]{(byte) 0x0F});
System.out.println(a1.and(b1));            // BytesOf{[15, -1, -1]}  → 0F-FF-FF

// OR silently loses trailing bytes of second
Bytes a2 = new BytesOf(new byte[]{(byte) 0x01, (byte) 0x02});
Bytes b2 = new BytesOf(new byte[]{(byte) 0x03, (byte) 0x04, (byte) 0x05, (byte) 0x06});
System.out.println(a2.or(b2));             // BytesOf{[3, 6]}  → 03-06, no trace of 05-06

// XOR — same asymmetry as OR
System.out.println(a2.xor(b2));            // BytesOf{[2, 6]}  → 02-06
```

At the EO level, the same shapes reproduce; the current runtime happily returns `0F-FF-FF` for `FF-FF-FF.and 0F` and `03-06` for `01-02.or 03-04-05-06`, without any error surfaced to the caller.

## Suggested resolution

Two clean options, in decreasing order of what I read from #464:

1. **Reject at the boundary.** In `and`, `or`, `xor`, `throw new ExFailure("bytes.<op> requires operands of the same length, got %d and %d", ...)` when the sizes differ. Matches @yegor256's earlier direction and keeps `bytes` operations totally defined per length. This is the smallest and most conservative fix.

2. **Zero-extend to `max`.** Length of the result becomes `max(first.length, second.length)`; the shorter operand contributes zeros in the extended region. This preserves algebraic identities (`x & 0 = 0`, `x | 0 = x`, `x ^ 0 = x`) across mismatched sizes, but silently changes the shape of the returned bytes — probably too permissive for a strongly-typed EO.

Either way, add tests that a) `and` with `first.length > second.length` on non-trivial bytes doesn't return `first`-unchanged for the tail, and b) `or`/`xor` with `first.length < second.length` doesn't drop bytes of `second`. I'm happy to open the PR once the semantics are picked — flagging first because #464's decision was never actually implemented.

morphqdd:
Closes #5484.

`BytesRaw.and`, `BytesRaw.or`, and `BytesRaw.xor` used to iterate only up to `Math.min(first.length, second.length)` and always return a result of `first.length`, silently producing wrong results when the two operands differ in length (trailing bytes of the longer operand were left untouched for `and`, and trailing bytes of the longer operand were dropped for `or`/`xor`).

This came up in #464 back in 2021, where `@yegor256` said "if two arrays are of different lengths, we must throw an exception", but that direction was never implemented and the issue was closed as stale.

This PR implements that direction: `and`, `or`, `xor` now throw `ExFailure` when the operand lengths differ, instead of silently dropping or ignoring bytes. The three methods share the length check through a new private `ofSameLength` helper, to avoid repeating the same check three times.

Added three tests in `BytesOfTest` (one per operation) that verify the exception is thrown for mismatched lengths, following the same pattern already used for `doesNotSupportNegativeBits`.

One real call site did rely on the old silent behavior: `socket.eo`'s `htons` computed `bts.and FF-`, where `bts` is 2 bytes (`i16.as-bytes`) and `FF-` is a 1-byte literal. Under the old code this happened to be a harmless no-op (AND-ing a byte with `0xFF` changes nothing, and the untouched second byte also stayed correct by coincidence), but it is now a length mismatch that throws. Fixed by using the length-matched 2-byte literal `00-FF` instead of `FF-` in both places in `htons`, which is also the semantically correct mask for isolating one byte of a 2-byte value. Verified locally: `EOsocketTest` fails deterministically (all 3 top-level tests, both single-threaded and under the repo's default concurrent JUnit5 execution) without this fix, and passes cleanly with it, matching a clean run against `upstream/master`.

All existing `bytes.eo` and other consumer tests (`i64.eo`, `i16.eo`, `i32.eo`, `number.eo`, `string.eo`, `file.eo`) pass unchanged, since every other call site already uses operands of equal length. Full local `eo-runtime` test suite (surefire) is clean.

### DIFF
--- a/eo-runtime/src/main/eo/bytes.eo
+++ b/eo-runtime/src/main/eo/bytes.eo
@@ -158,6 +158,12 @@
         -7.as-bytes.or 23.as-bytes
         0.as-bytes
 
+  --> on-and-with-different-lengths
+    20-1F.and CA-FE-BE > @
+
+  --> on-or-with-different-lengths
+    20-1F.or CA-FE-BE > @
+
   ++> tests-not-with-zero
     not. > @
       eq.
--- a/eo-runtime/src/main/eo/bytes/xor.eo
+++ b/eo-runtime/src/main/eo/bytes/xor.eo
@@ -62,3 +62,6 @@
     eq. > @
       (00-00-00-00-FF-FF-FF-FF.xor FF-FF-FF-FF-00-00-00-00).as-number
       FF-FF-FF-FF-FF-FF-FF-FF
+
+  --> on-xor-with-different-lengths
+    20-1F.xor CA-FE-BE > @
--- a/eo-runtime/src/main/eo/socket.eo
+++ b/eo-runtime/src/main/eo/socket.eo
@@ -79,8 +79,8 @@
   [port] > htons
     as-i16. > @
       or.
-        (bts.and FF-).left 8
-        (bts.right 8).and FF-
+        (bts.and 00-FF).left 8
+        (bts.right 8).and 00-FF
     port.as-i16.as-bytes > bts
 
   [recv] > as-input
--- a/eo-runtime/src/main/java/org/eolang/BytesRaw.java
+++ b/eo-runtime/src/main/java/org/eolang/BytesRaw.java
@@ -40,8 +40,8 @@ final class BytesRaw implements Bytes {
     @Override
     public Bytes and(final Bytes other) {
         final byte[] first = this.take();
-        final byte[] second = other.take();
-        for (int index = 0; index < Math.min(first.length, second.length); index += 1) {
+        final byte[] second = BytesRaw.ofSameLength(first, other, "and");
+        for (int index = 0; index < first.length; index += 1) {
             first[index] = (byte) (first[index] & second[index]);
         }
         return new BytesOf(first);
@@ -50,8 +50,8 @@ final class BytesRaw implements Bytes {
     @Override
     public Bytes or(final Bytes other) {
         final byte[] first = this.take();
-        final byte[] second = other.take();
-        for (int index = 0; index < Math.min(first.length, second.length); index += 1) {
+        final byte[] second = BytesRaw.ofSameLength(first, other, "or");
+        for (int index = 0; index < first.length; index += 1) {
             first[index] = (byte) (first[index] | second[index]);
         }
         return new BytesOf(first);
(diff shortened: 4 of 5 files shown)


morphqdd:

yegor256:
@morphqdd can you add EO test for this failure too? Java tests are file, but EO tests are more important, since they are portable to other runtime platforms (like JS)

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1290](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5638, opened 2026-07-18
#5629: fail with ExFailure on scanf format ending in dangling %

### LINKED ISSUE #5629: `scanf` crashes with `StringIndexOutOfBoundsException` on a format ending in a lone `%`
## What happens

`EOscanf.lambda()` (`eo-runtime/src/main/java/org/eolang/EO_string/EOscanf.java`) builds the match regex in a first loop (`:72-101`), where a trailing lone `%` sets `literal = true` and appends nothing — no error is raised for the dangling, never-completed specifier. The extraction loop (`:107-121`) then locates that `%` via `frmt.indexOf('%')` and reads the following char with `frmt.charAt(idx + 1)` (`:114`); when the `%` is the last character of the format, `idx + 1 == frmt.length()` and a raw `java.lang.StringIndexOutOfBoundsException` escapes instead of a proper `ExFailure`.

This is wrong because the loss of well-formedness is turned into a low-level JDK exception rather than a fail-fast, human-readable error — the same anti-pattern already avoided elsewhere in the codebase (e.g. the `default:` branch at `:93-94` throws a clear `ExFailure` for an *unsupported* specifier, and the `HEX`/precision paths in the parser fail fast with a named message). An incomplete `%` deserves the same treatment.

## Failure scenarios

Verified against the compiled `EOscanf` atom via the runtime (`new PhApplication(new PhApplication(new EOscanf(), "format", <fmt>), "read", <read>)` dataized):

| format | read | result |
| --- | --- | --- |
| `"%d"` | `"42"` | OK (parses `42`) |
| `"a%"` | `"a"` | `StringIndexOutOfBoundsException: Index 2 out of bounds for length 2` |
| `"%d%"` | `"42"` | `StringIndexOutOfBoundsException: Index 1 out of bounds for length 1` |
| `"%"` | `"x"` | `StringIndexOutOfBoundsException` |

At the EO surface the same shapes reproduce, e.g. `scanf "%d%" "42"`. The crash requires the built regex to match the `read` string (otherwise `matcher.find()` is `false` and the extraction loop is skipped, so an empty tuple is returned instead). `scanf` is on a live path — it backs `string.as-number` (`eo-runtime/src/main/eo/string/as-number.eo:19`, `(scanf "%f" text)`).

## Suggested resolution

Two clean options:

1. **Fail fast in the first loop.** After the `for` over `format` ends, if `literal` is still `true`, the format ended mid-specifier — `throw new ExFailure("format ends with an incomplete '%%' specifier")`. This mirrors the existing `default:` branch (`:93-94`) that already rejects unsupported specifiers, and keeps `scanf` totally defined over its format argument.
2. **Guard the extraction loop.** Change `:114` to bail on a trailing `%`, e.g. `if (idx + 1 >= frmt.length()) { break; }` (or treat the trailing `%` as a literal), so no read past the end occurs.

Option 1 is the more consistent with the atom's fail-fast contract. Either way, add tests that a format ending in a lone `%` (`"a%"`, `"%d%"`, `"%"`) produces a defined outcome rather than a raw `StringIndexOutOfBoundsException`.

morphqdd:
Closes #5629.

`EOscanf.lambda()` builds the match regex in a first loop, where a trailing lone `%` sets `literal = true` and appends nothing, without raising any error for the dangling, never-completed specifier. The extraction loop then locates that `%` via `frmt.indexOf('%')` and reads the following char with `frmt.charAt(idx + 1)`. When the `%` is the last character of the format, `idx + 1 == frmt.length()`, and a raw `java.lang.StringIndexOutOfBoundsException` escapes instead of a proper `ExFailure`.

This adds a check right after the first loop: if `literal` is still `true` once the loop ends, the format string has a dangling `%` with no specifier after it, so it now fails fast with a clear `ExFailure`, matching the existing pattern used for an unsupported specifier (the `default:` branch a few lines above).

Added `EOscanfTest` (a Java unit test asserting `ExFailure` is thrown for `scanf "hello%" "hello"`) and a `--> on-scanf-with-dangling-percent` EO-level example in `scanf.eo`, next to the two existing `on-scanf-with-*` negative examples for this object.

Verified locally: the new tests pass, `EOstringTest` (51 tests) and the generated `EOscanfEOAtomTest` (15 tests, including the new example) pass, and qulice is clean.

### DIFF
--- a/eo-runtime/src/main/eo/string/scanf.eo
+++ b/eo-runtime/src/main/eo/string/scanf.eo
@@ -55,6 +55,9 @@
   --> on-scanf-with-oversized-int
     scanf "%d" "99999999999999999999" > @
 
+  --> on-scanf-with-dangling-percent
+    scanf "hello%" "hello" > @
+
   ++> tests-scanf-with-string-int-float
     eq. > @
       scanf
--- a/eo-runtime/src/main/java/org/eolang/EO_string/EOscanf.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/EOscanf.java
@@ -99,6 +99,12 @@ public final class EOscanf extends PhDefault implements Atom {
                 }
             }
         }
+        if (literal) {
+            throw new ExFailure(
+                "The format string '%s' ends with a dangling '%%' and no specifier after it",
+                format
+            );
+        }
         final Matcher matcher = Pattern.compile(regex.toString()).matcher(
             new Dataized(this.take("read")).asString()
         );
--- a/eo-runtime/src/test/java/org/eolang/EO_string/EOscanfTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EO_string/EOscanfTest.java
@@ -1,0 +1,50 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+
+/*
+ * @checkstyle PackageNameCheck (4 lines)
+ * @checkstyle TrailingCommentCheck (3 lines)
+ */
+package org.eolang.EO_string; // NOPMD
+
+import org.eolang.Data;
+import org.eolang.Dataized;
+import org.eolang.ExFailure;
+import org.eolang.PhApplication;
+import org.eolang.Phi;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test for {@link EOscanf}.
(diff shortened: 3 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1274](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5651, opened 2026-07-20
#5648: emit compact-tuple marker `*N` for trailing tuples after leading args

### LINKED ISSUE #5648: Printer never emits the compact-tuple marker `*N` (N≥1) for a trailing tuple after leading args
When an application applies a trailing tuple after one or more leading positional arguments — the valid compact-tuple case `head *N`, N≥1 — the printer never emits the `*N` marker. It renders the tuple as a verbose `* elem` child instead. For example `sprintf` with a format string plus a `"Jeff" "Mary"` tuple has the canonical compact spelling

```
sprintf *1
  "Hello, %s and %s!"
  "Jeff"
  "Mary"
```

but the printer emits

```
sprintf
  "Hello, %s and %s!"
  * "Jeff" "Mary"
```

Same tree (`Φ.tuple` as the last child), so no correctness loss — just missed compaction. The gate is `Pretty.Node.tuply()`, which requires `this.children.size() == 1` (`eo-printer/src/main/java/org/eolang/printer/Pretty.java:577`), so the trailing-star hybrid fires only for the sole-child `seq *` case.

This is not #5622. That guard was added by #5622 to stop the printer emitting the **before-star** form `head args *` (`sm.win32 "getenv" *`) — invalid EO the parser reads as an empty tuple, dropping the indented element — and the fall-back to the verbose `* elem` child was the safe fix. But the compact-tuple **marker** `head *N` (§3.9, `LnCompactTuple`) is a different, valid spelling: the `*N` sits on the head line and every argument, including the leading ones, becomes an indented sibling. The parser round-trips it — see `compact-tuple-with-method.yaml` (`string.sprintf *1`). #5622 only removed the invalid before-star spelling; it never produced the valid marker.

To fix: extend the trailing-star hybrid so that when the last child is a gluable star following N≥1 leading arguments, the printer emits the head with a `*N` marker and lays all arguments as indented siblings (the §3.9 form), generalizing the `children.size() == 1` guard rather than bailing to the verbose child. Emit the `*N` head marker only — never the bare `head args *` before-star form that #5622 removed — and keep the verbose `* elem` child as the fall-back when the head cannot carry a marker (e.g. reversed dispatch).

yegor256:
Fixes #5648.

## Problem

When an application applied a trailing tuple after one or more leading positional arguments — the valid compact-tuple case `head *N`, N≥1 — the printer never emitted the `*N` marker. It rendered the tuple as a verbose `* elem` child instead:

```
sprintf
  "Hello, %s and %s!"
  * "Jeff" "Mary"
```

instead of the canonical compact spelling (§3.9):

```
sprintf *1
  "Hello, %s and %s!"
  "Jeff"
  "Mary"
```

The trailing-star hybrid was gated by `Pretty.Node.tuply()`, which required `this.children.size() == 1`, so it only fired for the sole-child `seq *` idiom.

## Fix

- **`tuply()`** — generalized the `children.size() == 1` guard: the hybrid now fires whenever the *last* child is a gluable star, regardless of how many leading arguments precede it. Split into helpers `marked()` (head can carry a marker) and `absorbed(size)` to keep the boolean complexity within qulice limits.
- **`glued()`** — now builds `head *N` (N = number of leading arguments) and lays every argument — the leading ones and the tuple's elements — as indented siblings. `N == 0` keeps the bare `*` (the `seq *` idiom), matching the parser's default count.
- **`shaped()`** — the `*N` marker for N≥1 is the canonical spelling, so it is emitted whenever it applies rather than only when it wins the penalty vote (the verbose `* elem` child can inline onto fewer lines). The bare `*` (N==0) idiom stays a penalty-gated candidate, so a short tuple that fits inline as `seq`, `* 1 2` is still left alone.

Guards preserved:

- **Reversed dispatch / formation heads** cannot carry a marker → verbose `* elem` fall-back (a reversed compact-tuple head is not yet parseable).
- **Tuple-literal head** (`base == "*"`) is excluded — its elements are already tuple elements, and `* *N` would be a confusing self-application.
- **Before-star form** `head args *` (removed by #5622) is never produced. The dotted-dispatch restriction from #5622/#5624 now applies only to `N == 0` (the bare `*`); the `*N` marker sits on the head line, so it round-trips after a method dispatch too (`string.sprintf *1`, cf. the parser's `compact-tuple-with-method.yaml`).

## Tests

- New `compact-tuple-marker.yaml` covering the canonical `string.sprintf *1` case (dotted dispatch, tuple that would otherwise inline).
- Updated `trailing-star-after-args.yaml` (the #5622 case) to the new `sm.win32 *1` compaction.
- Updated `idiomatic.yaml` and `self-reference.yaml` where the generalized rule now compacts their trailing tuples.

All 110 `XmirTest` print-pack cases pass, including the `printsToParseableEo` idempotency check, and the 58 `MjPrintTest`/`MjFormatTest` cases (straight and reversed notation). Qulice passes on `eo-printer`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KYUSfEKzYNVXrs1S57DfvH)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Node.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Node.java
@@ -6,6 +6,7 @@ package org.eolang.printer;
 
 import com.github.lombrozo.xnav.Filter;
 import com.github.lombrozo.xnav.Xnav;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (273 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2489](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5652, opened 2026-07-20
#5650: keep compact suffix for compound-receiver method chains

### LINKED ISSUE #5650: Printer de-compacts a method chain on a compound receiver: `(input.read 4096).read.^` becomes a 3-line nested reversed dispatch
The `format` goal de-compacts a method chain on a compound receiver. In `eo-runtime/src/main/eo/io/input-length.eo:20` the hand-written source reads

```
(input.read 4096).read.^ > chunk
```

and the formatter rewrites it to a three-line nested reversed-dispatch form:

```
^. > chunk
  read.
    input.read 4096
```

Same tree — `((input.read 4096).read).^` — so no correctness loss, but the compact one-liner a human wrote is lost.

The cause is candidate generation, not penalty. `Pretty.suffixed()` rebuilds the glued suffix shape `receiver.method` only when the receiver is a lone data literal — the `receiver.data && receiver.children.isEmpty()` guard at `eo-printer/src/main/java/org/eolang/printer/Pretty.java:171`, the `5.plus` case. A compound receiver (here `input.read 4096`, which has a child) yields empty, so the parenthesized suffix `(input.read 4096).read.^` is never built. `layout()` only ever weighs the reversed vertical form, and the penalty comparison it does for the lone-data case never happens here.

This matters because the compact form wins under the printer's own penalty function. With the default weights the one-liner scores 78 (indent 6 + one bracket 15 + 36 symbols + 21 application) while the emitted three-line form scores 96 (indent 27 + 48 symbols + 21 application) — the extra indentation and repeated base characters outweigh the single parenthesis. The printer emits a rendering 18 points worse than one it could produce, purely because the better candidate is never generated.

To fix: generalize `suffixed()` so a compound receiver is also rebuilt as a parenthesized suffix `(receiver).method`, gated on the receiver having a one-line inline form, and let the existing penalty comparison in `layout()` keep it only when it beats the reversed vertical form. The parenthesis cost is already modeled (`BRACKET`), so a receiver that does not fit inline, or a chain that is genuinely uglier with the paren, is rejected on penalty rather than never considered.

yegor256:
Closes #5650.

## Problem

The `format` goal de-compacted a method chain on a compound receiver. In `eo-runtime/src/main/eo/io/input-length.eo:20` the hand-written source

```
(input.read 4096).read.^ > chunk
```

was rewritten into a three-line nested reversed dispatch:

```
^. > chunk
  read.
    input.read 4096
```

Same tree, so no correctness loss, but the compact one-liner a human wrote was lost — and it scores *better* under the printer's own penalty function.

## Root cause

Candidate generation, not penalty. `Pretty.suffixed()` only rebuilt the glued suffix shape `receiver.method` when the receiver was a lone data literal (the `5.plus` case). A compound receiver (one with children of its own, like `input.read 4096`) yielded empty, so the parenthesized suffix `(input.read 4096).read.^` was never built and `layout()` only ever weighed the reversed vertical form.

## Fix

Generalize `suffixed()` so any reversed dispatch whose receiver has a one-line inline form is rebuilt in suffix position:

- the receiver is inlined through `flat()` and glued to the method as `receiver.method`;
- it is wrapped in parentheses only when it applies arguments of its own (reusing the same rule as `inlined()`), so a single token such as `01-.as-bool` stays bare and never triggers *"redundant parentheses around a single token"* (#5591);
- the parenthesis cost is already modelled by `BRACKET`, so the existing penalty comparison in `layout()` keeps the suffix shape only when it beats the reversed vertical form. A receiver that does not fit inline, or a chain that is genuinely uglier with the paren, is rejected on penalty rather than never being considered.

The reported case now round-trips as the compact `(input.read 4096).read.^ > chunk`.

## Tests

- Refreshed the seven print-packs whose canonical layout is now more compact (e.g. `div. (a.plus b) c` → `(a.plus b).div c`, `if. (x.lt 0) ...` → `(x.lt 0).if ...`). The idempotency check (`printsToParseableEo`) still passes for all of them.
- Added `compound-receiver-suffix.yaml`, a dedicated pack for the exact `(input.read 4096).read.^` scenario.
- Updated `MjFormatTest.appliesCustomIndentationStep` to use a nested formation, since the hello-world fixture now legitimately collapses to the one-liner `(stdout "Hello!" x).print > [x] > main` and no longer exercises deep indentation.

All `eo-printer` (134) and `eo-maven-plugin` `MjFormatTest`/`MjPrintTest` (58) tests pass, and `qulice` (0.27.6, as CI runs it) is green for both modules.

https://claude.ai/code/session_01YGEf5FcqoecvBVVtqWQPre

---
_Generated by [Claude Code](https://claude.ai/code/session_01YGEf5FcqoecvBVVtqWQPre)_

### DIFF
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjFormatTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjFormatTest.java
@@ -88,14 +88,37 @@ final class MjFormatTest {
                 new FakeMaven(temp)
                     .with("autoFix", true)
                     .with("step", 4)
-                    .withProgram(MjFormatTest.canonical(new HelloWorld().asString()))
+                    .withProgram(MjFormatTest.canonical(MjFormatTest.nested()))
                     .execute(MjFormat.class)
                     .result()
                     .get("foo/x/main.eo")
             ).asString(),
             Matchers.containsString(
-                String.valueOf('\n').concat("        org.eolang.io.stdout")
+                String.valueOf('\n').concat("        x > first")
             )
+        );
+    }
+
+    /**
+     * A program that stays multi-line whatever the layout weights are.
+     *
+     * <p>A nested formation with two bindings never collapses onto a single
+     * line — an only-phi formation binds nothing but its {@code φ} decoratee —
+     * so its deepest lines sit two indentation levels in and expose the
+     * configured {@code step}, unlike a compact one-liner such as
+     * {@code (stdout "Hello!" x).print > [x] > main}.</p>
+     *
+     * @return The EO program source
+     */
+    private static String nested() {
+        return String.join(
+            System.lineSeparator(),
+            "+package foo.x",
+            "",
+            "[x] > main",
+            "  [] > inner",
+            "    x > first",
+            "    x > second"
         );
     }
 
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -101,13 +101,18 @@ final class Pretty {
      * Lay out a node as a (possibly multi-line) block at the given
      * indentation, picking the rendering with the lowest penalty.
      *
-     * <p>When the node is a reversed dispatch on a lone data receiver
-     * ({@code plus. 5 3}), its equivalent suffix shape ({@code 5.plus 3})
-     * is laid out too and the lower-penalty one is kept — the same
-     * penalty comparison that already decides inline versus vertical. The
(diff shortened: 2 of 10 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (179 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2493](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5655, opened 2026-07-20
`eo-printer` doesn't penalize usage of explicit `@` in `Penalty` computation


yegor256:
The pretty-printer scores every candidate layout in `Penalty.points()` (`eo-printer/src/main/java/org/eolang/printer/Penalty.java:89`), summing per-line charges for indentation, opening brackets, and width overflow, with each weight declared as a tunable knob in `PenaltyKey` (`eo-printer/src/main/java/org/eolang/printer/PenaltyKey.java:20`). There is no knob that charges for the phi attribute `@`, so the algorithm treats a rendering full of explicit `@` references as no uglier than one without them.

Explicit `@` hurts readability, and we would like the layout search to prefer equivalent renderings that avoid it. Since every other aesthetic — indentation, brackets, overflow — is already expressed as a penalty weight, the absence of a phi penalty is the one gap that keeps us from tuning this preference. Each `@` on a line should cost points (roughly 15) so the printer leans away from phi-heavy layouts.

Add a `PHI(15)` key to `PenaltyKey` and one accumulation line in `points()` — `total += Penalty.phis(line) * this.weight(PenaltyKey.PHI);` — backed by a new static `phis(String line)` counter that tallies `'@'` characters, mirroring the existing `brackets(String line)` counter at `Penalty.java:135`.

0crat:
@yegor256 Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+2469](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5653, opened 2026-07-20
Scale APPLICATION penalty by application length in eo-printer


yegor256:
The penalty-based printer charges the `APPLICATION` weight (`PenaltyKey.APPLICATION`, currently 7, at `eo-printer/src/main/java/org/eolang/printer/PenaltyKey.java:47`) once per application space on a line, in `Penalty.points()` at `eo-printer/src/main/java/org/eolang/printer/Penalty.java:105` through the `applications()` helper (lines 147-159). This code lives on branch `5577` and is not yet on `master`.

The charge is flat per space, so each argument costs the same constant increment no matter how long the application already is. Long applications are less readable than short ones, yet this linear-per-space model never penalizes length beyond that constant, and the cost vanishes entirely once the arguments wrap vertically. There is nothing pushing the printer, or the author, away from a sprawling application.

A concrete fix: make the per-application charge depend on the application's length by multiplying the `APPLICATION` weight by the number of objects inside the brackets. For `(foo bar 42 44)` the weight would be multiplied by 3 (the three applied arguments), so longer applications grow super-linearly more expensive and the printer favors shorter ones. The exact multiplier (arguments vs. all objects) is worth settling in review.

0crat:
@yegor256 Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+2481](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5658, opened 2026-07-20
fix(#5657): treat throwing-test marker `--&gt;` as a name binding

### LINKED ISSUE #5657: `eo-printer` fails to collapse `-->` marker to shorthand like `++>` due to omission in `Penalty.binding()`
On branch `5577` (PR #5579), the eo-printer keeps the verbose `> @` layout for a phi-only object bound with the throwing-test marker `-->`, instead of collapsing it to the phi-free shorthand the way `++>` does. In `eo-runtime/src/main/eo/number/mod.eo` the `on-mod-by-zero` test prints as

```
--> on-mod-by-zero
  mod > @
    5
    0
```

when the equivalent `mod 5 0 --> on-mod-by-zero` (identical XMIR, one line, no explicit `@`) is available. The same object bound with `++>` does collapse. The cause is `Penalty.binding()` at `eo-printer/src/main/java/org/eolang/printer/Penalty.java:243`, which recognises `>`, `>>` and `++>` as name-binding markers but omits `-->`.

That helper decides which inter-token spaces count as argument applications for the super-linear SPACE surcharge. With `-->` unrecognised, the two spaces flanking it are charged as applications and pay the `r`-squared surcharge, so the collapsed candidate scores far higher than it should (131 vs the 61 its `++>` twin gets, against a verbose form of 88). The layout search therefore rejects the phi-free rendering and keeps the explicit `@` — exactly the shape the new PHI penalty (#5655) is meant to steer away from. This is the untreated twin of #5567, which fixed the same collapse for `++>` but never covered the throwing-test marker.

Add `"-->".equals(token)` to the disjunction in `Penalty.binding()` so the throwing-test marker is treated as a name binding, like `++>`.

yegor256:
Fixes #5657.

## Problem

`Penalty.binding()` recognised `>`, `>>` and `++>` as name-binding markers but omitted the throwing-test marker `-->`. Because of that, the two spaces flanking `-->` were counted as argument applications and paid the super-linear `SPACE` surcharge added in #5653. The phi-free shorthand (e.g. `mod 5 0 --> on-mod-by-zero`) therefore scored far higher than its `++>` twin (131 vs 61, against a verbose form of 88), so the layout search rejected it and kept the verbose

```
--> on-mod-by-zero
  mod > @
    5
    0
```

with an explicit `@` — exactly the shape the PHI penalty (#5655) steers away from. This is the untreated twin of #5567 (which fixed the same collapse for `++>`).

## Fix

- **`Penalty.binding()`**: add `"-->".equals(token)` to the disjunction so the throwing-test marker is treated as a name binding, like `++>`. Javadoc updated accordingly.
- **`throws-test-attribute.yaml`**: restore the collapsed golden `x.eq x --> on-something` (it was the regression #5653 introduced for `-->`).
- **`PenaltyTest`**: add `treatsThrowingTestMarkerAsBinding`, asserting `foo --> name` is charged like `foo ++> name` and cheaper than the application `foo bar name`.
- **eo-runtime `.eo` sources**: reformat the 15 files whose `-->` tests now render canonically in the compact shorthand, so the format-check gate stays green.

## Verification

- `eo-printer` `PenaltyTest` (10) and `XmirTest` (114 print-pack goldens) pass.
- `mvn process-sources` on `eo-runtime` (format check, autoFix off) reports "All 139 EO source(s) are formatted canonically", BUILD SUCCESS.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RjwDHuWPVtWq2kv4VrCFyD)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Penalty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Penalty.java
@@ -179,8 +179,9 @@ final class Penalty {
      * make up an application's length.
      *
      * <p>Only the spaces that apply an argument to an object count. A space
-     * that sits next to a name-binding marker ({@code >}, {@code >>} or
-     * {@code ++>}) binds a name rather than applying an argument, and the
+     * that sits next to a name-binding marker ({@code >}, {@code >>},
+     * {@code ++>} or {@code -->}) binds a name rather than applying an
+     * argument, and the
      * spaces between the void attributes inside a formation's {@code [..]}
      * head are not applications either; neither is counted. So
      * {@code foo > [] > bar} has no application spaces, while
@@ -238,10 +239,11 @@ final class Penalty {
     /**
      * Is this token a name-binding marker rather than an applied argument?
      * @param token The token
-     * @return TRUE for {@code >}, {@code >>} and {@code ++>}
+     * @return TRUE for {@code >}, {@code >>}, {@code ++>} and {@code -->}
      */
     private static boolean binding(final String token) {
-        return ">".equals(token) || ">>".equals(token) || "++>".equals(token);
+        return ">".equals(token) || ">>".equals(token)
+            || "++>".equals(token) || "-->".equals(token);
     }
 
     /**
--- a/eo-printer/src/test/java/org/eolang/printer/PenaltyTest.java
+++ b/eo-printer/src/test/java/org/eolang/printer/PenaltyTest.java
@@ -111,6 +111,20 @@ final class PenaltyTest {
     }
 
     @Test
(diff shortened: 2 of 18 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (163 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2521](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5660, opened 2026-07-20
Charge nested brackets double in eo-printer Penalty BRACKET term

### LINKED ISSUE #5659: Charge nested brackets double in eo-printer `Penalty` BRACKET term
The pretty-printer charges every opening parenthesis the same flat `BRACKET` weight in `Penalty.points()` (`eo-printer/src/main/java/org/eolang/printer/Penalty.java:96`), where `Penalty.brackets(line)` (`Penalty.java:139`) just counts `(` characters and multiplies by the weight. Nesting depth is ignored, so a nested bracket is treated as no worse than a top-level one.

Nested brackets hurt readability far more than flat ones, and the layout search should reflect that. In `(foo (bar 42))` the two parentheses cost the same today, but the inner one — sitting inside another open bracket — should cost twice the normal `BRACKET` weight, so the printer leans away from deeply nested one-liners.

Replace the flat `brackets(line) * weight` term with a depth-aware sum: walk each line tracking the count of currently-open parentheses, and charge each `(` the `BRACKET` weight doubled when it opens inside an already-open bracket (depth ≥ 1) and the plain weight at depth 0. This keeps the existing per-line model and the tunable `BRACKET` knob untouched.

yegor256:
Closes #5659.

## What

`Penalty.points()` charged every opening parenthesis the same flat `BRACKET` weight, since `Penalty.brackets(line)` merely counted `(` characters. Nesting depth was ignored, so a deeply nested bracket cost no more than a top-level one.

This replaces the flat count with a **depth-progressive** weighted count in `Penalty.brackets`:

- walk each line tracking the number of currently-open parentheses;
- charge each `(` `(depth + 1)` `BRACKET` units — the flat weight at the top level, double one level deep, triple two levels deep, and so on;
- decrement depth on each matching `)`.

The `points()` line stays `brackets(line) * weight(BRACKET)`, so the per-line model and the tunable `BRACKET` knob are untouched.

With the default `BRACKET` weight of 7, `(a (b (c 1)))` now costs `(1 + 2 + 3) * 7 = 42` — the outer, middle, and inner brackets charged 1×, 2×, and 3× respectively — so the layout search leans harder against deeply nested one-liners.

## Tests

Extended `chargesForParenthesis` with a progressive-nesting assertion (`(a (b (c 1)))` scores 42). Kept the assertion within the existing test method to stay under PMD's per-class method limit. The full `PenaltyTest` suite passes (10 tests, 0 failures), and `mvn verify` (including qulice) is green locally.

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Penalty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Penalty.java
@@ -21,8 +21,10 @@ import java.util.Map;
  *   <li>every level of indentation on a line costs
  *   {@link PenaltyKey#INDENT} points;</li>
  *   <li>every opening parenthesis costs {@link PenaltyKey#BRACKET}
- *   points, doubled when it opens inside an already-open bracket, since
- *   nested brackets hurt readability more than flat ones;</li>
+ *   points, multiplied by its nesting depth plus one, so a top-level
+ *   bracket costs the flat weight, a bracket nested one level deep costs
+ *   twice as much, two levels deep three times, and so on, since deeper
+ *   nesting hurts readability more;</li>
  *   <li>every explicit phi attribute {@code @} costs
  *   {@link PenaltyKey#PHI} points;</li>
  *   <li>every character sitting past the {@link PenaltyKey#WIDTH}-th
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2529](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5664, opened 2026-07-20
Collapse multi-hop parent chains in eo-printer

### LINKED ISSUE #5663: Printer keeps `^.^.` multi-hop parent chains instead of collapsing to bare names
The eo-printer already collapses a redundant single parent hop back to a bare name, but it stops at one hop. In `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:234`, the `when` that turns `ξ.ρ.name` into bare `name` requires the first segment after the `ξ.ρ.` prefix to not be another `ρ` (`$first != $eo:rho`). A base like `ξ.ρ.ρ.max` (surface form `^.^.max`) fails that test and falls through to the `otherwise` at line 243 (`eo:surface(@base)`), which prints the fully expanded chain.

You can see the effect in `eo-runtime/src/main/eo/while.eo`: on master the tuple-iteration tests write these references bare — `max.eq 0`, `arr.at 0`, `max.gt iter` — and rely on scope resolution to walk up the `<o>` hierarchy at parse time. Running the printer over that XMIR rewrites them to `^.^.max`, `^.^.arr`, `^.^.^.arr`. The result is semantically identical (scope resolution re-derives the same hops) but verbose and non-canonical, so the printer is not idempotent against a hand-written source. The single-hop branch's own comment — "plain scope resolution re-derives the very same hop" — applies just as well to N hops: if a name is absent from every nearer enclosing formation and first declared N levels up, bare `name` re-derives exactly the `ρ`-chain of length N.

The fix is to generalize that one branch: consume a run of consecutive leading `ρ` segments after `ξ.`, take the first segment past the run as the name to check, and collapse when that name is absent from each of the N nearer ancestor formations (`ancestor::o[not(@base)]` positions 1..N) and present at position N+1 — instead of rejecting the chain the moment its second segment is `ρ`.

yegor256:
Fixes #5663.

## Problem

The eo-printer collapsed a redundant *single* parent hop (`ξ.ρ.name` → bare `name`) but stopped at one hop. In `to-eo-tree.xsl`, the branch required the first segment after the `ξ.ρ.` prefix to not be another `ρ` (`$first != $eo:rho`), so a base like `ξ.ρ.ρ.max` (surface `^.^.max`) fell through to the `otherwise` and printed the fully expanded chain. The result was semantically identical but non-canonical, so the printer was not idempotent against hand-written source (e.g. `while.eo`'s tuple-iteration tests, where bare `max`/`arr` got rewritten to `^.^.max`/`^.^.^.arr`).

## Fix

Generalize that branch to consume a run of consecutive leading `ρ` segments after `ξ.`:

- New `eo:rho-run` helper counts the leading `ρ` run (`rho-count = N`).
- `hop-name` is the first segment past the run; `hop-rest` is that segment plus any trailing segments.
- The chain collapses to the bare name when `hop-name` is absent from **each** of the N nearer enclosing formations (`ancestor::o[not(@base)]` positions 1..N) and **present** at position N+1 — instead of rejecting the moment the second segment is `ρ`.

For `N = 1` this reduces exactly to the previous single-hop behavior. When the name is not resolvable at exactly N levels up (e.g. only at global scope, or shadowed by a nearer scope), the chain is kept verbatim.

## Tests

- `multi-hop-parent.yaml` — `^.^.x` referencing a grandparent's param collapses to bare `x`.
- `multi-hop-parent-shadowed.yaml` — `^.^.x` is kept when a nearer scope declares `x` (bare form would resolve wrong).
- Existing `nested-parent.yaml` (name at global scope) and `redundant-parent.yaml` (single-hop + shadow) continue to pass.

All 142 `eo-printer` tests pass.

https://claude.ai/code/session_01KTcEfcWzqGuT26u4a9jvfd

---
_Generated by [Claude Code](https://claude.ai/code/session_01KTcEfcWzqGuT26u4a9jvfd)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -54,6 +54,11 @@
     <xsl:variable name="rest" select="substring-after($sig, $root)"/>
     <xsl:sequence select="if (starts-with($sig, $root)) then (if (contains($rest, '.')) then concat('Q.', $rest) else $rest) else $sig"/>
   </xsl:function>
+  <!-- Count the leading run of consecutive ρ segments in a sequence. -->
+  <xsl:function name="eo:rho-run" as="xs:integer">
+    <xsl:param name="segments" as="xs:string*"/>
+    <xsl:sequence select="if (empty($segments) or $segments[1] != $eo:rho) then 0 else 1 + eo:rho-run(subsequence($segments, 2))"/>
+  </xsl:function>
   <!-- PROGRAM -->
   <xsl:template match="object">
     <object>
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2568](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5667, opened 2026-07-21
fix(#5631): guard null regex capture groups in matched-from-index

### LINKED ISSUE #5631: `regex` match crashes due to `NullPointerException` with non-participating optional group
## What happens

A `regex` match whose pattern contains an **optional capturing group that does not participate** in the match crashes with a raw `java.lang.NullPointerException` the moment the matched block is read.

`EOregex$EOpattern$EOmatch$EOmatched_from_index.fill()` (`eo-runtime/src/main/java/org/eolang/EO_string/EOregex$EOpattern$EOmatch$EOmatched_from_index.java:119-128`) builds the `groups` tuple eagerly:

```java
groups = new Phi[matcher.groupCount() + 1];
for (int idx = 0; idx < groups.length; ++idx) {
    groups[idx] = new Data.ToPhi(matcher.group(idx));   // :123
}
```

For a non-participating group `java.util.regex.Matcher.group(idx)` returns `null` (documented behaviour). `new Data.ToPhi(null)` then reaches `Data.ToPhi.toPhi` (`eo-runtime/src/main/java/org/eolang/Data.java:124-172`), where `null` matches none of the `instanceof` checks and falls into the final `else` (`:166-170`), which dereferences `obj.getClass()` on the `null` — a `NullPointerException`, not even the intended `ExFailure`. `fill()` runs eagerly inside `matched-from-index.lambda()`, so the crash fires as soon as the matched block is dataized (e.g. reading `.from`, `.next`, `.groups`, or `.group N`), for every group index, regardless of which group the caller intends to read.

This mirrors the already-fixed regex robustness bug #5071 (`IndexOutOfBoundsException` on a malformed expression): a normal input leaks a raw JDK exception instead of a defined EO outcome.

## Failure scenarios

Verified end-to-end against the compiled runtime via `Phi.Φ.take("string").take("regex")` → `.match txt` → `.next.from`:

| pattern | text | result |
| --- | --- | --- |
| `/(a)(b)/` | `ab` | OK (`from = 0`) |
| `/(a)(b)?/` | `a` | `NullPointerException: Cannot invoke "Object.getClass()" because "obj" is null` |
| `/(a)(b)?/` | `ab` | OK (`from = 0`) |

Optional groups (`(...)?`, alternations like `(a)|(b)`, etc.) are ordinary regex usage, so any realistic pattern with a branch that leaves a capturing group unmatched trips this.

## Suggested resolution

Two parts:

1. **Represent a non-participating group in `fill()`** instead of passing `null` to `Data.ToPhi`. Either map `null` to an empty string so `group N` yields `""`:

   ```java
   final String captured = matcher.group(idx);
   groups[idx] = new Data.ToPhi(captured == null ? "" : captured);
   ```

   or, to keep "did not participate" distinguishable from "matched empty", store a `PhTerminator.withCause("group N did not participate in the match")` for that slot (as `blank()` already does for a non-existent match), so only reading that specific group fails, with a clear cause.

2. **Harden `Data.ToPhi.toPhi` against `null`** (`Data.java:166-170`): the `else` branch calls `obj.getClass()` and so turns a `null` argument into a `NullPointerException` rather than the intended `ExFailure`. Guard `null` explicitly, e.g. `throw new ExFailure("Cannot convert null to a data object")`, so any future caller passing `null` gets a diagnosable EO error instead of an NPE.

Add tests for a pattern with a non-participating optional group (`/(a)(b)?/` against `"a"`) producing a defined outcome rather than a raw `NullPointerException`.

arimu1:
## Summary

- Fix `NullPointerException` when a regex match has a non-participating optional capturing group (e.g. `/(a)(b)?/` against `"a"`).
- In `EOregex$EOpattern$EOmatch$EOmatched_from_index.fill()`, map `Matcher.group(i) == null` to an empty string before wrapping with `Data.ToPhi`.
- Add regression test in `RegexAtomTest` that reads `.from` on such a match without crashing.

Fixes #5631.

## Test plan

- [x] `mvn -pl eo-runtime -am install -DskipTests` (Java 17)
- [x] `mvn -pl eo-runtime -Dtest=RegexAtomTest test` — **7/7 passed** (includes new `readsFromWhenOptionalGroupDoesNotParticipate`)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/EO_string/EOregex$EOpattern$EOmatch$EOmatched_from_index.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/EOregex$EOpattern$EOmatch$EOmatched_from_index.java
@@ -120,7 +120,8 @@ public final class EOregex$EOpattern$EOmatch$EOmatched_from_index extends PhDefa
         if (matcher.groupCount() > 0) {
             groups = new Phi[matcher.groupCount() + 1];
             for (int idx = 0; idx < groups.length; ++idx) {
-                groups[idx] = new Data.ToPhi(matcher.group(idx));
+                final String captured = matcher.group(idx);
+                groups[idx] = new Data.ToPhi(captured == null ? "" : captured);
             }
         } else {
             groups = new Phi[]{new Data.ToPhi(matcher.group())};
--- a/eo-runtime/src/test/java/org/eolang/EO_string/RegexAtomTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EO_string/RegexAtomTest.java
@@ -110,6 +110,27 @@ final class RegexAtomTest {
     }
 
     @Test
+    void readsFromWhenOptionalGroupDoesNotParticipate() {
+        MatcherAssert.assertThat(
+            "match with a non-participating optional group must not crash when reading from",
+            new Dataized(
+                new PhApplication(
+                    new PhApplication(
+                        new PhApplication(
+                            Phi.Φ.take("string.regex").copy(),
+                            "expression", new Data.ToPhi("/(a)(b)?/")
+                        ).take("compiled").take("match").copy(),
+                        "txt", new Data.ToPhi("a")
+                    ).take("matched-from-index").copy(),
+                    new Bind(RegexAtomTest.position(), new Data.ToPhi(1)),
+                    new Bind(RegexAtomTest.start(), new Data.ToPhi(0))
+                ).take("from")
+            ).asNumber(),
+            Matchers.equalTo(0.0)
+        );
+    }
+
+    @Test
     void rejectsStartIndexAfterTextEnd() {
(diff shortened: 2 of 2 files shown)


arimu1:
Addressed review feedback: strengthened `readsFromWhenOptionalGroupDoesNotParticipate` to assert group 2 is `""` and `count` is `3` for `/(a)(b)?/` vs `"a"` (not only that `.from` dataizes without NPE). `RegexAtomTest` 7/7 green on Java 17.

0crat:
@arimu1 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5668, opened 2026-07-21
fix(#5630): count one printf argument per specifier with width/precision

### LINKED ISSUE #5630: `printf` fails with `MissingFormatArgumentException` when using standard formatting specifiers like `%.2f`, `%5d`, `%-10s`
## What happens

`EOprintf.lambda()` (`eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java:44-61`) passes the format string to Java's `String.format(Locale.US, format.replaceAll(...), args)`, which fully honours width, precision and flags. But the argument collector `PrintfArgs.formatted()` (`eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java:88`) enumerates specifiers with `Pattern.compile("%(\\d+\\$)?([a-zA-Z%])")`, which matches only a bare conversion char (optionally positional). A specifier carrying flags, width or precision — `%5d`, `%.2f`, `%-10s`, `%05d`, `%+d`, `%,d` — is not matched, so **no argument is collected for it**, while the specifier itself stays in the format handed to `String.format`. The argument list ends up shorter than the format demands and a raw `java.util.MissingFormatArgumentException` escapes instead of a proper `ExFailure`.

Support is thus half-wired: the format side accepts width/precision, the argument side does not. This is the opposite of failing fast — the same anti-pattern the atom already avoids for an *unsupported* specifier, where `PrintfArgs.fmt` (`:139-147`) throws a clean `ExFailure` (and `on-printf-unsupported-format`, `printf "%o"`, is an expected-failure test in `printf.eo:63-64`).

## Failure scenarios

Verified against the compiled `EOprintf` atom via the runtime (`new PhApplication(new PhApplication(new EOprintf(), "format", <fmt>), "args", <tuple>)` dataized to a string):

| format | args | result |
| --- | --- | --- |
| `"%d"` | `* 42` | OK (`"42"`) |
| `"%.2f"` | `* 3.14159` | `MissingFormatArgumentException: Format specifier '%.2f'` |
| `"%5d"` | `* 42` | `MissingFormatArgumentException: Format specifier '%5d'` |
| `"%-10s"` | `* "x"` | `MissingFormatArgumentException: Format specifier '%-10s'` |
| `"%d %5s"` | `* 7 "hi"` | `MissingFormatArgumentException: Format specifier '%5s'` |

`%.2f` (format a float to two decimals) is a canonical `printf` use, not an obscure edge. Bare `%s`/`%d`/`%f` are unaffected.

## Suggested resolution

Two options:

1. **Support them** — they already work on the `String.format` side, only the collector under-counts. Widen the collector regex to skip the flags/width/precision portion of the Java conversion grammar, keeping the conversion char as group 2 via a non-capturing precision group:

   ```java
   Pattern.compile("%(\\d+\\$)?[-#+ 0,(]*\\d*(?:\\.\\d+)?([a-zA-Z%])")
   ```

   One argument is then counted per real specifier and `String.format` performs the actual width/precision formatting. I verified this pattern matches `%5d`, `%.2f`, `%-10s`, `%05d`, `%+d`, `%,d` while still matching `%s`, `%1$s`, `%12$s`, `%%`, and `group(2)` remains the conversion char (no code change to the group indices needed).

2. **Reject them cleanly** — if width/precision is intentionally out of scope, detect a non-bare specifier and `throw new ExFailure(...)` with a clear message, mirroring the existing unsupported-specifier path, instead of leaking a JDK exception.

Option 1 also adds genuinely useful formatting. Either way, add tests for `%.2f`, `%5d`, `%-10s` producing a defined outcome rather than a raw `MissingFormatArgumentException`.

arimu1:
Closes #5630

## What
`PrintfArgs.formatted()` enumerated format specifiers with `Pattern.compile("%(\\d+\\$)?([a-zA-Z%])")`, which matches only a bare conversion char (optionally positional). A specifier carrying flags, width or precision — `%5d`, `%.2f`, `%-10s`, `%05d`, `%+d`, `%,d` — was not matched, so no argument got collected for it while the specifier stayed in the format handed to `String.format`. The argument list ended up shorter than the format demanded, so a raw `java.util.MissingFormatArgumentException` escaped instead of a proper result.

## Fix
Widen the collector regex to skip the flags/width/precision portion of the conversion grammar while keeping the conversion char as capture group 2:

```java
Pattern.compile("%(\\d+\\$)?[-#+ 0,(]*\\d*(?:\\.\\d+)?([a-zA-Z%])")
```

One argument is now counted per real specifier, and `String.format` performs the actual width/precision formatting (already worked once the argument was supplied).

## Tests
- `PrintfArgsTest#countsOneArgumentPerSpecifierWithWidthAndPrecision` — reproduces the bug directly against `PrintfArgs` (fails on `master`, passes with the fix).
- `printf.eo`: `formats-float-with-precision` (`%.2f`), `formats-integer-with-width` (`%5d`), `formats-string-left-justified-with-width` (`%-10s`) — end-to-end regression via `printf`.

```
mvn -pl eo-runtime test -Dtest=PrintfArgsTest,EOprintfTest
Tests run: 7, Failures: 0, Errors: 0 -- PrintfArgsTest
Tests run: 25, Failures: 0, Errors: 0 -- EOprintfTest
```

### DIFF
--- a/eo-runtime/src/main/eo/string/printf.eo
+++ b/eo-runtime/src/main/eo/string/printf.eo
@@ -162,3 +162,24 @@
       printf
         "%f"
         * 0.0
+
+  ++> formats-float-with-precision
+    eq. > @
+      "3.14"
+      printf
+        "%.2f"
+        * 3.14159
+
+  ++> formats-integer-with-width
+    eq. > @
+      "   42"
+      printf
+        "%5d"
+        * 42
+
+  ++> formats-string-left-justified-with-width
+    eq. > @
+      "x         "
+      printf
+        "%-10s"
+        * "x"
--- a/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
@@ -85,7 +85,9 @@ final class PrintfArgs {
 
     List<Object> formatted() {
         final List<Object> arguments = new ArrayList<>(0);
-        final Matcher matcher = Pattern.compile("%(\\d+\\$)?([a-zA-Z%])").matcher(this.format);
+        final Matcher matcher = Pattern.compile(
+            "%(\\d+\\$)?[-#+ 0,(]*\\d*(?:\\.\\d+)?([a-zA-Z%])"
+        ).matcher(this.format);
         long auto = 0L;
         while (matcher.find()) {
             final String positional = matcher.group(1);
--- a/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
@@ -86,7 +86,7 @@ final class PrintfArgs {
     List<Object> formatted() {
         final List<Object> arguments = new ArrayList<>(0);
         final Matcher matcher = Pattern.compile(
-            "%(\\d+\\$)?[-#+ 0,(]*\\d*(?:\\.\\d+)?([a-zA-Z%])"
+            "%(\\d+\\$)?[-#+ 0,(]*+\\d*+(?:\\.\\d+)?+([a-zA-Z%])"
         ).matcher(this.format);
         long auto = 0L;
         while (matcher.find()) {
--- a/eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
@@ -106,6 +106,18 @@ final class PrintfArgsTest {
     }
 
     @Test
+    void countsOneArgumentPerSpecifierWithWidthAndPrecision() {
+        final Phi tuple = Phi.Φ.take("tuple").copy();
+        tuple.put("length", new Data.ToPhi(1));
+        tuple.put("head", new Data.ToPhi(3.14159));
+        MatcherAssert.assertThat(
+            "a specifier carrying precision must still consume exactly one argument",
+            new PrintfArgs("%.2f", 1L, tuple.take("at")).formatted(),
+            Matchers.equalTo(new ListOf<>(3.14159))
+        );
+    }
+
+    @Test
     void reportsUnsupportedFormatMessageInsteadOfDoubleFormatting() {
         final Phi tuple = Phi.Φ.take("tuple").copy();
         tuple.put("length", new Data.ToPhi(1));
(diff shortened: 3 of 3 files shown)


arimu1:
Addressed both CI blockers:

**1. SonarCloud quality gate (catastrophic backtracking, D rating)** — `PrintfArgs.java:88`: made the flags/width/precision quantifiers possessive (`[-#+ 0,(]*+`, `\d*+`, `(?:\.\d+)?+`). `0` is a member of both `[-#+ 0,(]` and `\d`, so a run of zeros let the engine backtrack ambiguously between the two — classic quadratic blowup. Measured locally on `%000…!`: 32000 chars went from ~5.5s down to <1ms with possessive quantifiers.

Re-verified match semantics are unchanged: wrote a throwaway harness (not committed) diffing old vs. new regex match results (`group(1)`, `group(2)`, start/end) across 307k generated specifier permutations (positional index × flags powerset × width × precision × conversion char, including invalid/edge chars) plus 20k combined/embedded-in-literal-text strings — **0 mismatches** across all 327k+ cases. The existing `PrintfArgsTest` suite (7 tests) still passes unchanged.

**2. qulice (PMD `UseUnderscoresInNumericLiterals`)** — `PrintfArgsTest.java:112,116`: `3.14159` → `3.141_59`. Confirmed against a clean run of `qulice-maven-plugin:check` on this module: baseline (pre-fix) reports 102 violations including exactly these two `PrintfArgsTest.java[112]`/`[116]` PMD findings; with the fix applied, violation count drops to 100 and neither finding remains. (The other ~100 pre-existing violations, mostly ErrorProne `UnicodeInCode` warnings on the `φ`/`Φ` symbol used throughout the runtime, reproduce identically on unmodified code and are unrelated to this diff.)

**Non-blocking, flagging for awareness (not fixed, out of scope for this PR):** the regex at `PrintfArgs.java:88` matches `%n` and routes it to `ExFailure("format n is unsupported")`, even though `%n` is technically a valid zero-argument newline conversion in `java.util.Formatter`. This is pre-existing behavior (the old regex matched it identically) — this PR only touches quantifier greediness, not the character class — but since the diff touches this exact line, calling it out here in case a maintainer wants a follow-up.

yegor256:
@arimu1 merge conflicts here

arimu1:
@yegor256 While this was waiting, #5669 (merged as `53b718f635`, plus the follow-up `c0dbe26175`) landed the same fix for #5630 — the specifier pattern on `master` now skips flags/width/precision with possessive quantifiers and comes with broader test coverage than this PR. Nothing left here to rebase, so I'm closing. Thanks!

PR #5676, opened 2026-07-21
Add unsigned u8/u16/u32/u64 and signed i8 integer objects (#5661)

### LINKED ISSUE #5661: `eo-runtime` package missing unsigned integer objects `u8`, `u16`, `u32`, `u64` and signed `i8`
The root package `eo-runtime/src/main/eo/` ships the signed integer objects `i16.eo`, `i32.eo` and `i64.eo`, but none of the unsigned counterparts `u8`, `u16`, `u32`, `u64` exist anywhere in the tree. A grep for `\bu(8|16|32|64)\b` across `eo-runtime/src/main/eo/` returns nothing, and `bytes/` exposes only `as-i16.eo`, `as-i32.eo`, `as-i64.eo` conversions (see `eo-runtime/src/main/eo/bytes/as-i16.eo`), with no `as-u*` analogs.

The numeric tower is incomplete without them: unsigned widths are a standard part of the model, and their absence forces users to hand-roll unsigned arithmetic on top of the signed objects. Worth noting the signed side is also short an object — `i8` is missing too, so the smallest width has no representation at all in either family.

Add `u8.eo`, `u16.eo`, `u32.eo`, `u64.eo` to the root, mirroring the structure of the existing `i*.eo` objects (with `as-bytes`, `plus`, `minus`, `times`, `div`, comparisons and inline tests), and the matching `bytes/as-u8.eo`, `bytes/as-u16.eo`, `bytes/as-u32.eo`, `bytes/as-u64.eo` conversions alongside the current `as-i*` ones.

yegor256:
Closes #5661.

The numeric tower in `eo-runtime/src/main/eo/` shipped only `i16`, `i32` and `i64`. This PR adds the missing unsigned widths and the smallest signed width, together with the matching `bytes` conversions.

### New root objects
- `u8.eo`, `u16.eo`, `u32.eo`, `u64.eo` — unsigned integers
- `i8.eo` — the previously missing smallest signed width

### New `bytes` conversions
- `bytes/as-u8.eo`, `bytes/as-u16.eo`, `bytes/as-u32.eo`, `bytes/as-u64.eo`
- `bytes/as-i8.eo`

### Design
Each object mirrors the structure of the existing `i*.eo` objects (`as-bytes`, `plus`, `minus`, `times`, `div`, comparisons and inline tests):

- **`i8`** sign-extends to `i16`, delegates arithmetic, and wraps the result back to one byte.
- **`u8`, `u16`, `u32`** zero-extend to the next-wider *signed* type (`i16`, `i32`, `i64`). Because a zero-extended value is always non-negative there, the existing signed comparisons and division already behave as unsigned; results are sliced back to width, so overflow wraps modulo `2^n`.
- **`u64`** is the unsigned base. Addition, subtraction and multiplication reuse the `i64` bit patterns (identical in two's complement); comparisons flip the sign bit and defer to `i64`; division uses an unsigned long-division reduction (Hacker's Delight) so operands with the high bit set divide correctly rather than as two's complement.

### Testing
Ran the transpiled inline tests for all new objects locally against the runtime:

| object | tests | result |
|---|---|---|
| `u8` | 25 | pass |
| `u16` | 23 | pass |
| `u32` | 23 | pass |
| `u64` | 25 | pass |
| `i8` | 24 | pass |
| `bytes/as-u8`,`as-u16`,`as-u32`,`as-u64`,`as-i8` | 2/2/2/3/2 | pass |

Coverage includes unsigned-specific behaviour such as `(2^64-1) / 2 = 2^63-1` (which as two's complement would be `0`), `FF-.as-u8` comparing as magnitude `255`, and wrap-around on `plus`/`minus`/`times`. The full `eo-runtime` EO sources compile, transpile and lint cleanly with the additions.

---
_Generated by [Claude Code](https://claude.ai/code/session_011ngRDSr3T1H9eXJJhU3JDb)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (767 >= 100); -8 for way too many hits-of-code (767 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2596](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5679, opened 2026-07-21
#5634: compare map keys by bytes, not only by hash code

### LINKED ISSUE #5634: `map` overwrites distinct keys with colliding hashes due to hash-only key comparison
## What happens

`map` (`eo-runtime/src/main/eo/map.eo`) identifies entries by their **hash code alone and never compares the actual keys**. Two distinct keys whose hash codes collide are therefore treated as the same key: one entry is silently dropped when the map is built, and a lookup of the other key returns the wrong value while reporting `exists = true`.

Every key comparison in the map is hash-only:

- `rec-key-search` (`map.eo:64`): `tup.head.hash.eq hash` — matches on hash, returns that entry's value without checking the key.
- `rec-rebuild` (`map.eo:35`): `contains prev.hashes hash` — deduplicates incoming pairs by hash, so a colliding second key is discarded as if it were a duplicate.
- `with` (`map.eo:92`) and `without` (`map.eo:103`): `(hash.eq entry.hash).not` — filter by hash only.

The hash is `tuple/hash-code.eo`, a `31 * acc + byte` polynomial rolling hash (`hash-code.eo:24-31`). Such a hash has structured, easily reproduced collisions — the textbook pair `"Aa"` and `"BB"` both hash to `31*65 + 97 == 31*66 + 66 == 2112`.

## Failure scenarios

Verified end-to-end against the compiled runtime by building `map` from `map.entry` pairs and reading `size` / `found key`:

| map | `size` | `get "Aa"` | `get "BB"` |
| --- | --- | --- | --- |
| `{"Aa"→1, "Zz"→2}` (distinct hashes) | 2 | 1 | — (`"Zz"`→2) |
| `{"Aa"→1, "BB"→2}` (**colliding hashes**) | **1** | 1 | **1** (should be 2), `exists=true` |

So for the colliding map, the entry `"BB"→2` is silently lost on construction (`size` is 1, not 2), and `(map.found "BB").get` returns `1` — `"Aa"`'s value — while claiming the key exists. The same collision corrupts `has`, `keys`, `with` and `without`.

## Suggested resolution

After a hash match, compare the **actual keys** for equality rather than trusting the hash. The entry already stores the key alongside its hash (`map.eo:40-42` binds `key`, `value`, `hash`), so the fix is local:

- In `rec-key-search`, require `tup.head.hash.eq hash` **and** the stored key to equal the searched key (dataized/byte equality), before returning the entry.
- In `rec-rebuild`, treat an incoming pair as a duplicate only when both hash and key match, so colliding distinct keys are kept as separate entries.
- In `with` / `without`, filter out an existing entry only when both its hash and key match the target.

A hash match should narrow the candidate set; key equality must confirm the hit. Add tests using a known collision (`"Aa"`/`"BB"`) asserting that a map built from both keeps `size == 2`, that `get "BB"` returns `2`, and that `get "Aa"` returns `1`.

gemshrine:
Closes #5634.

`map` identified entries by their **hash code alone and never compared the actual keys**, in `rec-rebuild` (dedup on rebuild), `found`/`rec-key-search` (lookup), `with` and `without` (filtering). `tuple.hash-code` is a `31 * acc + byte` polynomial, so collisions are trivial to hit: `"Aa"` and `"BB"` both hash to `31*65 + 97 == 31*66 + 66 == 2112`. Two distinct keys that collide were therefore treated as one key — one entry was silently dropped when the map was built, and looking up the other key returned the wrong value while reporting `exists` as `true`.

Every key comparison now also compares the raw bytes of the key, so a hash match only narrows the candidates and byte equality confirms the hit.

The tricky part is that `tests-map-rebuilds-itself-only-once` requires each key to be dataized exactly once, so simply calling `key.as-bytes` at every comparison would have broken that guarantee. Instead each key is dataized to bytes once, when the map is built, the bytes are stored in the entry next to the hash code, and the hash code is now derived from those bytes rather than from the key object. Since `bytes.as-bytes` returns the very same chain of bytes, `tuple.hash-code` no longer touches the key, and the total number of key dataizations is unchanged.

The `hashes` tuple and the `couple` object are gone: duplicates are now detected by scanning the accumulated entries directly, so the `tuple.contains` alias is no longer needed either.

`set` is built on top of `map` (`mp.with`/`without`/`has`/`keys`), so it inherited the same defect and is fixed by this change as well.

Six examples were added to `map.eo`. Verified locally that all six fail against the current `master` logic and pass with this change:

| example | on `master` | with this change |
| --- | --- | --- |
| `tests-keeps-both-keys-with-colliding-hash-codes` | `size` is 1 | `size` is 2 |
| `tests-finds-value-of-each-key-with-colliding-hash-codes` | `"BB"` yields `1` | `"Aa"` yields `1`, `"BB"` yields `2` |
| `tests-does-not-have-absent-key-colliding-with-stored-one` | `has "BB"` is `true` | `false` |
| `tests-falls-back-for-absent-key-colliding-with-stored-one` | returns `"Aa"`'s value | returns the fallback |
| `tests-adds-key-colliding-with-existing-one` | `with` replaces instead of adding | both keys kept |
| `tests-removes-only-the-given-key-of-colliding-ones` | `without "Aa"` drops `"BB"` too | only `"Aa"` removed |

`EOmapTest` is 20/20 green and the full `eo-runtime` suite is 1628 tests with no new failures. The only failure, `EOsocketTest.refusesConnectionViaSyscall`, reproduces identically on an unmodified `master` on this machine and is unrelated to this change. The change introduces no new lint warnings: the two reported for `map.eo` are present on `master` as well.

### DIFF
--- a/eo-runtime/src/main/eo/map.eo
+++ b/eo-runtime/src/main/eo/map.eo
@@ -1,8 +1,14 @@
 # Hash-map.
 # Here `pairs` must be a `tuple` of `map.entry` object.
 # The `map.entry.key` must be a dataizable object, `map.entry.value` may be any object.
+# Two keys are the same only when their hash codes and their raw bytes are equal.
+# A hash code alone is not enough, since `tuple.hash-code` is a `31 * acc + byte`
+# polynomial and therefore has collisions, e.g. `"Aa"` and `"BB"` both hash to 2112.
+# Comparing hash codes only would make such distinct keys overwrite each other.
+# Every key is dataized to bytes exactly once, when the map is built, and the
+# hash code is derived from those bytes rather than from the key itself,
+# so that keys with side effects are never evaluated more than once.
 
-+alias tuple.contains
 +alias tuple.filtered
 +alias tuple.hash-code
 +alias tuple.mapped
@@ -21,26 +27,31 @@
       if.
         pairs.length.eq 0
         *
-        entries.
+        @.
           rec-rebuild pairs
-    [entries hashes] > couple
     [tup] > rec-rebuild
       if. > @
         tup.length.eq 0
-        couple
-          *
-          *
+        *
         if.
-          tuple.contains prev.hashes hash
+          has-key prev
           prev
-          couple
-            prev.entries.with
-              [] >>
-                ^.hash > hash
-                tup.head.key > key
-                tup.head.value > value
-            prev.hashes.with hash
-      tuple.hash-code tup.head.key > hash!
+          prev.with
+            [] >>
+              ^.hash > hash
+              ^.kbts > kbts
+              tup.head.key > key
+              tup.head.value > value
+      tuple.hash-code kbts > hash!
+      tup.head.key.as-bytes > kbts!
+      if. > [ent] > has-key
+        ent.length.eq 0
+        false
+        if.
+          (ent.head.hash.eq hash).and
+            ent.head.kbts.eq kbts
+          true
+          has-key ent.tail
       @. > prev
         rec-rebuild tup.tail
   [key value] > entry
@@ -50,7 +61,8 @@
         size.eq 0
         not-found
         rec-key-search entries
-      tuple.hash-code key > hash!
+      tuple.hash-code kbts > hash!
+      key.as-bytes > kbts!
       [tup] > rec-key-search
         if. > @
           tup.length.eq 0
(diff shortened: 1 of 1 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (117 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+387](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5680, opened 2026-07-21
Guard inline-cactoos against self-referential abstracts (#5677)

### LINKED ISSUE #5677: Printer's inline-cactoos.xsl infinitely recurses on a self-referential `>>` abstraction
The template in `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:29` inlines a reference to a cactus-named (`a🌵…`) abstraction by unconditionally copying the target's body (`<xsl:apply-templates select="$target/node()"/>` at line 39). When that abstraction references itself, the copied body still holds the self-reference, so the same template fires again on it, without bound. Saxon aborts at `inline-cactoos.xsl` line 55 with "Too many nested apply-templates calls. Possible infinite recursion", and the `format` goal (`MjFormat`) fails the whole build.

A recursive helper is legal EO, so the printer should not crash on it. For example, in `eo-runtime/src/main/eo/tuple/last-index-of.eo` a helper written as `if. > [t] >> rec` with `rec t.tail` in its body compiles to XMIR where the auto-named object `a🌵…` contains a `ξ.ρ.a🌵…` reference to itself. Written instead with a plain name (`if. > [t] > rec`) the same recursion builds fine and the printer leaves it untouched, so the defect is specific to the cactus-named (`>>`) variant, not to recursion as such.

Guard the `o[contains(@base, concat('.', $auto))]` template so it inlines only when `$target` does not transitively reference its own name: when the target's subtree contains a reference back to `$name`, fall through to the `<xsl:copy>` branch and leave both the reference and the abstraction in place instead of expanding them.

yegor256:
Closes #5677.

## Problem

The `inline-cactoos` printer pass (`eo-printer/.../print/inline-cactoos.xsl`) inlined a reference to a cactus-named (`a🌵…`) abstraction by unconditionally copying the target's body. When that abstraction referenced itself — a recursive helper written as `? >> rec` whose body calls `rec` again — the copied body still held the self-reference, so the same template fired on it again, without bound. Saxon aborted with:

```
Too many nested apply-templates calls. Possible infinite recursion.
```

and the `format` goal (`MjFormat`) failed the whole build. A recursive helper is legal EO, so the printer should not crash on it. The defect was specific to the cactus-named (`>>`) variant; the same recursion written with a plain name (`> rec`) prints fine.

## Fix

Guard the `o[contains(@base, concat('.', $auto))]` inlining template so it inlines only when the target does **not** transitively reference its own name. When the target's subtree contains a reference that resolves back to `$name`, the template now falls through to the copy branch and leaves the reference in place. The companion "drop the inlined abstract" template is likewise guarded so a self-referential abstraction is kept rather than removed. Both the reference and the abstraction therefore stay in place instead of being expanded.

The self-reference check is factored into two small `eo:` functions (`eo:resolved-name`, `eo:recursive`), and the existing `$name` computation now reuses `eo:resolved-name` so the reference-resolution logic lives in one place.

## Test

Added `eo-printer/src/test/resources/org/eolang/printer/eo-packs/print/inline-cactoos-recursive.yaml`, an XMIR-level pack that runs the full parser pipeline followed by `inline-cactoos` on a recursive `? >> reclast` helper and asserts that the abstraction, its self-reference, and the outer reference are all left in place (no infinite recursion). All 119 `XmirTest` packs and the full `eo-printer` suite pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QvUCazUDqTkQ3WiBtr68HU)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -3,7 +3,7 @@
 * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 * SPDX-License-Identifier: MIT
 -->
-<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="inline-cactoos" version="2.0">
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" xmlns:xs="http://www.w3.org/2001/XMLSchema" exclude-result-prefixes="xs" id="inline-cactoos" version="2.0">
   <!--
   Converts such EO code:
   [] > foo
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2604](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5683, opened 2026-07-21
Preserve name of recursive `>> name` formations in printer (#5681)

### LINKED ISSUE #5681: Printer drops the name of a recursive `>> name` formation, rendering its references as synthetic `v15_2`
In the printer train `Xmir.java:43`, `restore-local-names.xsl:42` drops the `@local` handle marker on every non-void `>>` formation, and its `void-handle` key (`restore-local-names.xsl:25`) matches voids only (`@base=$eo:empty`), so references to such a formation are never rewritten back to the readable handle. That was harmless while `inline-cactoos.xsl` always inlined the formation away, but the #5677 guard now correctly keeps a self-referential (recursive) formation. Such a formation therefore reaches `to-eo-tree.xsl` with its user name gone: the definition prints as anonymous `[t] >>` and its self-references print as the synthetic cactus placeholder `v15_2`.

A recursive helper named with `>>` is legal EO, so the printer should preserve its name. For example, in `eo-runtime/src/main/eo/tuple/last-index-of.eo` the helper `if. > [t] >> rec` with `rec t.tail` in its body formats to `if. > [t] >>` plus references to an undeclared `v15_2` — the `rec` name is lost. The result is also non-idempotent: `eo:format` never converges (the check keeps reporting the file as "not formatted canonically"), and a further pass drops the orphaned formation altogether, leaving `v15_2 > @` bound to a name that no longer exists anywhere.

In `restore-local-names.xsl`, stop treating a non-void `>>` handle as always-inlined: when the formation references itself — the same local self-reference test the #5677 guard added to `inline-cactoos.xsl` — promote its `@local` to the visible `@name` (as the void branch at line 34 already does) and extend the `void-handle` key at line 25 to rewrite its self-references, instead of dropping the marker at line 42.

yegor256:
Fixes #5681.

## Problem

In the printer train, `restore-local-names.xsl` dropped the `@local` handle marker on every non-void `>>` formation, and its `void-handle` key matched voids only (`@base=∅`), so references to such a formation were never rewritten back to the readable handle. That was harmless while `inline-cactoos.xsl` always inlined the formation away, but the #5677 guard now correctly keeps a self-referential (recursive) formation in place.

Such a formation therefore reached `to-eo-tree.xsl` with its user name gone: the definition printed as an anonymous `[t] >>` and its self-references as the synthetic cactus placeholder `v15_2`. The result was also non-idempotent — `eo:format` never converged.

## Fix

- **`restore-local-names.xsl`**: stop treating a non-void `>>` handle as always-inlined. When the formation references itself — using the same self-reference test the #5677 guard added to `inline-cactoos.xsl` — it is now treated like a handled void: its `@local` is promoted to the visible `@name`, the `void-handle` key is extended to rewrite its self-references back to the handle, and the `@local` marker is kept instead of being dropped.
- **`to-eo-tree.xsl`**: a non-void formation carrying an `@local` handle is now printed with its readable name under a double-arrow (`>> name`) rather than as an anonymous `>>`.

## Tests

- `print-packs/yaml/recursive-local-handle.yaml` — end-to-end round-trip: a recursive helper declared with `[tup] >> reclast` now formats to `if. > [tup] >> reclast` (was `if. > [tup] >>` with `v6_2` references), and the printed form re-prints to itself (idempotent).
- `eo-packs/print/restore-recursive-local-name.yaml` — focused XSL-level assertions on `restore-local-names.xsl`: the handle is promoted to `@name`, the marker is kept, and both the self-reference and the outer reference are rewritten to the readable handle.

All 160 `eo-printer` tests pass.

https://claude.ai/code/session_016WSJjXij3WwQ1dG7fYdjif

---
_Generated by [Claude Code](https://claude.ai/code/session_016WSJjXij3WwQ1dG7fYdjif)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
@@ -16,30 +16,64 @@
   "@local" marker is deliberately KEPT on the void (#5581) so that
   "to-eo-tree" can print it back as a vertical "? &gt;&gt; name" line and
   preserve the void's anonymity (§9.2, R-9.2.3), rather than collapsing it
-  into a public "[name]" bracket param. A non-void "&gt;&gt;" handle (on an
-  anonymous formation) is left to the existing "inline-cactoos" pass, so its
-  @local marker is simply dropped.
+  into a public "[name]" bracket param.
+
+  A non-void "&gt;&gt;" handle sits on an anonymous formation. Usually the
+  "inline-cactoos" pass inlines that formation away, so its handle is
+  irrelevant and the "@local" marker is simply dropped. A self-referential
+  (recursive) formation is the exception: since #5677 "inline-cactoos"
+  correctly keeps it in place, so it reaches "to-eo-tree" and must carry a
+  readable name. Using the same self-reference test as that guard, such a
+  formation is treated like a handled void here — its "@local" is promoted
+  to the visible "@name", its self-references are rewritten back to the
+  handle, and the marker is KEPT so "to-eo-tree" prints "&gt;&gt; name"
+  rather than an anonymous "&gt;&gt;" bound to references to a synthetic
+  "vL_P" placeholder.
   -->
   <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
-  <xsl:key name="void-handle" match="o[@local and @base=$eo:empty]" use="@name"/>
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (169 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2536](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5685, opened 2026-07-21
Preserve `|` pipe applications in printer (#5684)

### LINKED ISSUE #5684: Printer expands `|` pipe applications into vertical form, dropping compact syntax
The printer never emits the `|` pipe-application syntax (spec §3.14); it always expands a pipe line into a two-line vertical application. Reproduce by writing a pipe in `eo-runtime/src/main/eo/tuple/last-index-of.eo`:

```
  if. > [t] >> rec
    t.length.eq 0
    -1
    if.
      t.head.eq wanted
      t.tail.length
      rec t.tail
  | list > @
```

`mvn clean test -rf :eo-runtime` fails on the `format` goal, which rewrites `| list > @` into `rec > @` / `list`. Both forms parse to identical XMIR, so this is purely a printer choice, and the compact author-written pipe is lost.

This is wrong because `|` exists specifically to keep code compact, yet the canonical formatter can never produce it — the syntax is write-only, destroyed on the first `eo:format`. It is also inconsistent with the recent direction of the printer: #5681 taught it to *preserve* compact recursive `>> name` formations rather than expand them, and dropping `|` pipes is the same class of loss. The root cause is that `eo-parser/src/main/resources/org/eolang/parser/parse/wrap-applications.xsl:32-40` drops the `@pipe` marker at parse time, rewriting the line into an ordinary application, so no trace of the pipe reaches `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:194`, which has no pipe-emission branch.

Suggested fix: retain the `@pipe` marker on the application node through the parse pipeline instead of stripping it at `wrap-applications.xsl:37`, and add a branch to `to-eo-tree.xsl` that emits a `| args > name` continuation line when that marker is present.

yegor256:
Fixes #5684.

## Problem

The printer never emitted the `|` pipe-application syntax (spec §3.14); it always expanded a pipe line into a two-line vertical application. Because both forms parse to identical XMIR, the compact author-written `|` was **write-only** — destroyed on the first `eo:format`.

The root cause: `wrap-applications.xsl` dropped the `@pipe` marker at parse time, rewriting the line into an ordinary application, so no trace of the pipe reached `to-eo-tree.xsl`, which had no pipe-emission branch.

## Fix

Mirrors the direction of #5681 (preserving compact `>> name` formations):

1. **`wrap-applications.xsl`** — retain the `@pipe` marker on the application node instead of stripping it. `@base` is still set from the preceding sibling's `@name` for semantics; `@pipe` is a cosmetic hint that downstream passes and the compiler ignore.
2. **`XMIR.xsd`** — declare the `pipe` attribute so the canonical XMIR validates.
3. **`to-eo-tree.xsl`** — add a head branch that emits `|` for a pipe node, **only when its predecessor is still the immediately preceding sibling** (`@base` matches `ξ.<name>` / `<name>`). When the predecessor has floated away (#5526) the guard fails and the node prints as an ordinary application, which round-trips just as safely.

## Tests

- Updated `wrap-applications.yaml` (parser pack) to assert `@pipe` is now retained.
- Updated `pipe-named.yaml` (printer pack): `| 5 > foo5` is now preserved instead of expanded to `foo 5 > foo5`.
- Added `pipe-recursive.yaml` — the issue's `>> name` recursive-formation + `| list > @` scenario.
- Added `pipe-floated.yaml` — pipe inside an argument block (#5526 float-up) still round-trips compactly.

All new packs also pass the printer's `printsToParseableEo` round-trip check (printed EO re-parses to the same EO). Full `eo-parser` (1482) and `eo-printer` (162) suites are green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BdSramPXm1rBhwrqeHkJt6)_

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnPipe.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnPipe.java
@@ -33,10 +33,13 @@ import java.util.List;
  *
  * <p>Emission (R-3.14.7): the line emits a base-less {@code <o pipe=''>}
  * with the args as children; the {@code wrap-applications} reshape sets
- * {@code @base} from the preceding sibling's {@code @name} and drops
- * {@code @pipe}, so downstream passes treat it as a hand-written
- * application of the predecessor by name. The predecessor object stays in
- * place.</p>
+ * {@code @base} from the preceding sibling's {@code @name}, so downstream
+ * passes treat it as a hand-written application of the predecessor by
+ * name. The predecessor object stays in place. The {@code @pipe} marker
+ * itself is retained (not dropped) so the printer can round-trip the
+ * compact {@code | args > name} line rather than expand it into a
+ * two-line vertical application (#5684); it is a cosmetic hint that
+ * downstream passes and the compiler ignore.</p>
  *
  * @since 0.1
  */
--- a/eo-parser/src/main/resources/XMIR.xsd
+++ b/eo-parser/src/main/resources/XMIR.xsd
@@ -156,6 +156,21 @@
         </xs:documentation>
       </xs:annotation>
     </xs:attribute>
+    <xs:attribute name="pipe" type="xs:string">
+      <xs:annotation>
+        <xs:appinfo>The marker of a pipe-application continuation line "| args &gt; name" (§3.14)</xs:appinfo>
+        <xs:documentation>
+          A cosmetic marker kept on the application node that a "|" pipe line
+          (R-3.14.7) parses into. The "wrap-applications" pass sets the node's
+          "base" to the preceding sibling's name so the semantics match a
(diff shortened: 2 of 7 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (111 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2540](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5690, opened 2026-07-21
Resolve horizontal/vertical penalty ties toward the flat one-liner (#5686)

### LINKED ISSUE #5686: `eo-printer` incorrectly formats one-line phi binding into two lines during tie penalties
When `eo:format` runs over `eo-runtime/src/main/eo/tuple/last-index-of.eo`, it rewrites the one-line phi binding `| list > @` into the two-line vertical form `| > @` followed by an indented `list`. The choice is made in `Pretty.shaped()` at `eo-printer/src/main/java/org/eolang/printer/Pretty.java:158-162`, where the horizontal (one-line) rendering replaces the vertical one only when its penalty is *strictly* lower.

Both renderings score an identical 50 points: the single line pays an extra application space (+7), which is exactly offset by the two-line form's extra indent level and leading symbols (+7). Because the comparison is `<` rather than `<=`, this genuine tie is silently resolved toward the taller, more nested layout. That contradicts the tie rule already used for the suffix comparison in `Pretty.layout()` at `eo-printer/src/main/java/org/eolang/printer/Pretty.java:126-127`, which keeps the shorter form on a tie via `<=`.

The smallest fix is to align the two comparisons: change the `<` on line 160 to `<=` so a tie resolves to the flatter one-liner. Note that applying this unconditionally also reformats about 34 other `eo-runtime` files, so the maintainer may prefer a narrower guard, but the `<=` alignment is the concrete change that flips the reported case.

yegor256:
Fixes #5686.

## Problem

`Pretty.shaped()` (`eo-printer/src/main/java/org/eolang/printer/Pretty.java`) replaced the vertical rendering of an object with the horizontal one-liner only when the latter's penalty was **strictly** lower (`<`). A genuine penalty tie was therefore silently resolved toward the taller, more nested layout.

This contradicted the tie rule already used by `Pretty.layout()` for the suffix comparison (`<=`), which keeps the shorter form on a tie. As reported, `eo:format` over `eo-runtime/src/main/eo/tuple/last-index-of.eo` split the one-line phi binding `reclast list > @` into the two-line vertical form (`reclast > @` over an indented `list`), even though both renderings score identically.

## Fix

Align the two comparisons by changing the `<` on the horizontal-vs-vertical check to `<=`, so a tie resolves to the flatter one-liner — the same rule `layout()` already applies.

## Changes

- `eo-printer/.../Pretty.java`: the one-character `<` → `<=` change, plus Javadoc updated to document the tie rule and to correct the now-stale description of the bare-`*` idiom.
- `eo-runtime/src/main/eo/**`: 35 sources reformatted onto their flatter canonical form (this is what `MjFormat`'s check now expects; without it the build's `format` goal would fail).
- Print-packs: `empty-string`, `nested-formation-apply`, `stars-tuples`, and `tuple-of-applications` updated to the new canonical output; new `phi-application-tie` pack pins the reported case (`X > @` over an indented arg collapses to `X arg > @`).

## Verification

- `mvn test` in `eo-printer` is green (164 tests).
- The 35 reformatted sources were produced through the exact `MjFormat` path (`new Xmir(new EoSyntax(actual).parsed()).toEO()` with default weights) and re-running the formatter over them is idempotent, so the `format` check and re-parse round-trip both hold.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D7Q4rSKBfv5kgmJEGX4dgc)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (395 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2544](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5673, opened 2026-07-21
MjTranspile.java file is not correctly forwarded to the JVM


0pdd:
The puzzle `5466-f65ad894` from #5466 has to be resolved: 

https://github.com/objectionary/eo/blob/16daa9f801e5bde9d07540dc3540288582d319a3/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java#L92-L100

The puzzle was created by @yegor256 on 21-Jul-26. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #5694, opened 2026-07-21
#5671: turn raw eo-runtime coverage hits into an LCOV report

### LINKED ISSUE #5671: Raw coverage hits in MjTranspile.java are not being properly processed
The puzzle `5466-44871d46` from #5466 has to be resolved: 

https://github.com/objectionary/eo/blob/16daa9f801e5bde9d07540dc3540288582d319a3/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java#L74-L83

The puzzle was created by @yegor256 on 21-Jul-26. 

morphqdd:
Closes #5671.

The `coverageFile` parameter of `transpile` only produced a raw, append-only `loc:line:pos` file: `PhCoverage` wrote every touched location into it, but nothing consumed that file. There was no way to see which EO objects were actually covered by tests.

Added:

- `MjTranspile`/`Transpiling` now write a manifest of every location the transpiler wrapped into `PhCoverage` (source file, locator, source line), next to the raw `coverageFile`. The manifest is rebuilt from scratch on every transpile run, and stays correct across incremental (cached) builds because it is derived from the transpiled output on disk, not only from cache misses.
- `CoverageManifest`: owns the manifest file (path derivation, reset, scan-and-append). Scans the transpiled Java text for `new PhCoverage(...)` calls with a regex, since that is exactly the set `to-java.xsl` emits.
- `CoverageReport`: merges the manifest against the raw hits into an LCOV tracefile (`SF`/`DA`/`LF`/`LH`/`end_of_record` blocks, one per source file) and computes the covered percentage from that same tracefile, so there is one place that knows the LCOV grammar.
- `MjCoverageReport`: a new `coverage-report` goal (bound to the `verify` phase, after tests run) that writes the LCOV file and logs the percentage. Fails with a clear message if the project was never transpiled with `coverageFile` set.

This closes the first of the three puzzles left in `MjTranspile.java` by #5466. The other two (minimum coverage enforcement in eo-runtime, forwarding the file to the JVM that runs the compiled program) are separate issues (#5672, #5673) and are left as `@todo` puzzles, updated to reference the new classes. Wiring `MjCoverageReport` into `eo-runtime/pom.xml` is scoped to #5672, since it depends on this PR.

Verified:
- 10 new tests (`CoverageManifestTest`, `CoverageReportTest`, `MjCoverageReportTest`, plus one added to `MjTranspileTest`), all green.
- `qulice:check` (checkstyle + PMD) clean on the whole module.
- Full `eo-maven-plugin` test suite: 417 tests, only the same 8 `MjPrintTest` failures as on unmodified master (confirmed pre-existing and unrelated, via `git stash` + rerun on master).

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/CoverageManifest.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/CoverageManifest.java
@@ -1,0 +1,143 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import com.jcabi.xml.XML;
+import com.yegor256.xsline.Shift;
+import com.yegor256.xsline.Train;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.util.Collection;
+import java.util.function.Function;
+
+/**
+ * The full set of locations the transpiler wrapped into {@code PhCoverage},
+ * one line per location: symbolic locator and source line, tab-separated.
+ * Built once per {@code transpile} run (see {@link Transpiling}), to be read
+ * back later and merged against the raw hits that {@code PhCoverage} itself
+ * records at runtime, which is what tells the instrumented lines that were
+ * actually touched from the ones that were not. The manifest deliberately
+ * does not record a source file for each location: a locator starts with the
+ * id of the object it belongs to, and the {@code eo-foreign} catalog already
+ * maps every registered id to the source it was compiled from, so repeating
+ * that mapping here, once per instrumented location, would only be a second
+ * copy of it to keep in step.
+ * @since 0.62.0
+ */
+final class CoverageManifest implements Coverage {
+
+    /**
+     * The condition {@code to-java.xsl} tests in its own
+     * {@code mode="located"} template before emitting a wrapper.
+     */
+    private static final String WHEN = "@line and @pos and not(contains(@loc, '+'))";
+
+    /**
+     * The {@code @base} an unbound (void) attribute carries.
+     */
+    private static final String VOID = "∅";
+
+    /**
+     * Selects exactly the elements that end up wrapped into {@code
+     * PhCoverage}, in the XMIR as it stands right before the final
+     * {@code to-java.xsl} shift runs, so that the manifest is derived from
+     * that structure directly, not by pattern-matching the Java text
+     * {@code to-java.xsl} emits.
+     * <p>
+     *     {@code to-java.xsl} emits a wrapper from a single place, its
+     *     {@code mode="located"} template, and that template is reached from
+     *     exactly four others: {@code match="atom"} applies it to the atom's
+     *     first {@code o} child, {@code match="abstract"} applies it to
+     *     itself, and the two {@code mode="object"} templates, for a plain
+     *     object ({@code @base} set and not starting with a dot) and for a
+     *     method call ({@code @base} starting with a dot, with children),
+     *     apply it to themselves. The four branches below mirror those four,
+     *     each carrying {@link #WHEN}, and the third one also leaves out
+     *     {@link #VOID}, the placeholder of an unbound attribute, which is a
+     *     hole rather than an object: {@code to-java.xsl} turns it into an
+     *     {@code AtVoid} through its own {@code match="void"} template and
+     *     never dispatches on it as an object. Everything else that happens
+     *     to carry the same attributes, a formation with no {@code @base}
+     *     above all, is left out too, because no wrapper is ever emitted for
+     *     it and counting it would inflate the {@code LF} of the report.
+     * </p>
+     */
+    private static final String LOCATED = String.format(
+        String.join(
+            " | ",
+            "//abstract[%1$s]",
+            "//atom/o[1][%1$s]",
+            String.format(
+                "//o[@base and @base!='' and @base!='%s' and not(starts-with(@base, '.')) and %%1$s]",
+                CoverageManifest.VOID
+            ),
+            "//o[starts-with(@base, '.') and * and %1$s]"
+        ),
+        CoverageManifest.WHEN
+    );
+
+    /**
+     * The raw coverage hits file this manifest is paired with.
+     */
+    private final Path hits;
+
+    /**
+     * Ctor.
+     * @param touched The raw coverage hits file this manifest is paired with
+     */
+    CoverageManifest(final Path touched) {
+        this.hits = touched;
+    }
+
+    @Override
+    public boolean tracked() {
+        return true;
+    }
+
+    @Override
+    public XML located(
+        final XML xmir, final Function<XML, XML> shifts, final Collection<String> found
+    ) {
+        final XML located = shifts.apply(xmir);
+        for (final XML element : located.nodes(CoverageManifest.LOCATED)) {
+            found.add(
+                String.format(
+                    "%s\t%s",
+                    element.xpath("@loc").get(0),
+                    element.xpath("@line").get(0)
+                )
+            );
+        }
+        return located;
+    }
+
+    @Override
+    public Train<Shift> pending(final Train<Shift> whole, final Train<Shift> last) {
+        return last;
+    }
+
+    @Override
+    public void save(final Collection<String> found) throws IOException {
+        Files.write(
+            this.path(),
+            found,
+            StandardCharsets.UTF_8,
+            StandardOpenOption.CREATE,
+            StandardOpenOption.TRUNCATE_EXISTING
+        );
+    }
+
+    /**
+     * Where this manifest is stored: a sibling of the hits file.
+     * @return The path
+     */
+    Path path() {
+        return Paths.get(String.format("%s.manifest", this.hits));
+    }
+}
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
@@ -5,6 +5,7 @@
 package org.eolang.maven;
 
 import com.jcabi.log.Logger;
+import java.io.File;
 import java.io.IOException;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
@@ -64,36 +65,42 @@ public final class MjTranspile extends MjSafe {
     private boolean trackLocations;
 
     /**
-     * Whether to wrap every located object in the generated Java into
-     * {@code PhCoverage}, so it records a coverage hit the first time it
-     * is touched at runtime. Off by default, so a production build stays
-     * lean. The hits are appended to the file named by the
-     * {@code eo.coverageFile} system property of the JVM that runs the
-     * compiled program; when that property is absent, every hit is a
-     * silent no-op. In {@code eo-runtime} the {@code coverage-file}
-     * profile turns this on and forwards that property to surefire in
-     * one step (see its {@code pom.xml}).
-     * @todo #5466:60min Turn raw coverage hits into an LCOV report.
-     *  Right now the runtime only produces a raw, append-only
-     *  {@code loc:line:pos} file: the {@code PhCoverage} decorator
-     *  writes every touched location into it, but nothing consumes that
-     *  file yet. Add a reporter step that merges those raw hits against
-     *  the full set of instrumented locations, which the transpiler
-     *  already knows because it emits every wrapper, and produces an
-     *  LCOV ({@code .info}) tracefile plus the covered percentage. LCOV
-     *  is chosen because Codecov and Coveralls consume it directly.
+     * The file to record EO object coverage hits into. When it is
+     * not set (the default), no instrumentation happens; when set,
+     * every located object in the generated Java is wrapped into
+     * {@code PhCoverage}, which appends a hit to this file the first
+     * time the object is touched at runtime. The very same
+     * {@code eo.coverageFile} value also reaches the JVM that later
+     * runs the instrumented code, because {@code eo-runtime} passes
+     * this Maven property straight through to surefire's
+     * {@code systemPropertyVariables} (see its {@code pom.xml}), so
+     * one setting is enough end to end. This same transpile run also
+     * writes the full set of instrumented locations into a sibling
+     * manifest (see {@link CoverageManifest}), so that the two files
+     * together say which of the instrumented lines were reached and
+     * which were not.
+     * @todo #5671:60min Merge the manifest and the hits into an LCOV report.
+     *  The transpiler now writes both halves of the answer, the manifest
+     *  of every instrumented location next to this file and, at runtime,
+     *  the raw {@code loc:line:pos} hits inside it, but nothing reads
+     *  them back yet. Add a goal, late in the lifecycle so that it runs
+     *  after the tests, that reads the two, resolves each locator to the
+     *  {@code .eo} source it came from through the {@code eo-foreign}
+     *  catalog {@code register} already builds, and writes an LCOV
+     *  ({@code .info}) tracefile plus the covered percentage. LCOV is
+     *  the format Codecov and Coveralls consume directly.
      * @todo #5466:30min Enforce a minimum EO object coverage in eo-runtime.
-     *  Once the LCOV report from the puzzle above exists, set
-     *  {@code coverageTracking} on the {@code transpile} execution in
-     *  {@code eo-runtime/pom.xml} and fail the build when the covered
+     *  Once the LCOV report from the puzzle above exists, bind it and set
+     *  {@code coverageFile} on the {@code transpile} execution in
+     *  {@code eo-runtime/pom.xml}, and fail the build when the covered
      *  percentage of dataized {@code .eo} objects drops below a
      *  threshold (for example 80 percent), mirroring how the existing
      *  {@code jacoco} profile binds a {@code check} goal with per-metric
      *  thresholds.
      * @checkstyle MemberNameCheck (7 lines)
      */
-    @Parameter(property = "eo.coverageTracking")
-    private boolean coverageTracking;
+    @Parameter(property = "eo.coverageFile")
+    private File coverageFile;
 
     @Override
     public void exec() throws IOException {
@@ -108,7 +115,7 @@ public final class MjTranspile extends MjSafe {
                 this.transpileTests,
                 this.xslMeasures.toPath(),
                 new Tracking(this.trackTransformationSteps, this.trackLocations),
-                this.coverageTracking
+                this.coverage()
             )
         ).exec();
         if (this.addSourcesRoot) {
@@ -128,5 +135,20 @@ public final class MjTranspile extends MjSafe {
                 gtests
             );
         }
+    }
+
+    /**
+     * The coverage instrumentation configured for this run.
+     * @return A manifest-writing coverage when {@link #coverageFile} is set,
+     *  a do-nothing one otherwise
+     */
+    private Coverage coverage() {
+        final Coverage result;
+        if (this.coverageFile == null) {
+            result = new Coverage.Off();
+        } else {
+            result = new CoverageManifest(this.coverageFile.toPath());
+        }
+        return result;
     }
 }
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
@@ -24,6 +24,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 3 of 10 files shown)


yegor256:
@morphqdd it seems that the manifest file is very similar to the `foreign.json` file we already build on compilation: it contains all objects registered. Maybe we can reuse it?

morphqdd:
@yegor256 good point, and this is already addressed in the current version of the PR.

The coverage manifest does not duplicate `foreign.json`/`eo-foreign`. It only stores two things per instrumented location: the symbolic locator and the source line number, one line per `PhCoverage` wrapper the transpiler emits. It does not store a source file path at all.

Resolving which `.eo` file a location belongs to is done by `ObjectSources`, which matches the locator's leading object id against the existing `eo-foreign` catalog (the same one `foreign.json` is built from). So `MjCoverageReport` reuses `eo-foreign` for that lookup instead of keeping its own copy of the object-to-file mapping.

The manifest still needs to exist separately, because it records something `eo-foreign` does not: the exact line number inside each object where a coverage-trackable location sits, which only the transpiler knows at transpile time.

yegor256:
@morphqdd please, merge master into your branch - it changed the logic of coverage collection and now it works. the `coverage.txt` file is generated in the `eo-runtime/target/` directory after each build.

morphqdd:
@yegor256, rebased

eo-maven-plugin/src/main/java/org/eolang/maven/CoverageManifest.java
```diff
@@ -0,0 +1,100 @@
...
+ * @since 0.58
+ */
+final class CoverageManifest {
+
+    /**
+     * Matches one {@code new PhCoverage(name, "loc", line, pos);} call, as
+     * emitted by {@code to-java.xsl}, in the transpiled Java text.
+     */
+    private static final Pattern WRAPPED = Pattern.compile(
+        "new PhCoverage\\([^,]+,\\s*\"([^\"]*)\",\\s*(\\d+),\\s*\\d+\\)"
```
yegor256: @morphqdd this looks dirty. We should not rely on information from inside the generated source code.

morphqdd: Fixed: CoverageManifest now derives the manifest from the XMIR itself (the same @line/@pos/@loc attributes to-java.xsl's own "located" template tests), built by running every transpile shift except the final to-java.xsl one, instead of pattern-matching the generated Java text.

eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
```diff
@@ -64,36 +66,32 @@ public final class MjTranspile extends MjSafe {
...
+     *  execution in {@code eo-runtime/pom.xml}, and fail the build when
+     *  the covered percentage of dataized {@code .eo} objects drops
+     *  below a threshold (for example 80 percent), mirroring how the
+     *  existing {@code jacoco} profile binds a {@code check} goal with
+     *  per-metric thresholds.
      * @checkstyle MemberNameCheck (7 lines)
      */
-    @Parameter(property = "eo.coverageTracking")
-    private boolean coverageTracking;
+    @Parameter(property = "eo.coverageFile")
```
Favixx, line 102: Renaming the parameter from `coverageTracking` to `coverageFile` leaves three call sites on master untouched, and none of them are in this PR: `eo-runtime/pom.xml:238` (`<coverageTracking>false</coverageTracking>`), `.github/workflows/mvn.yml:36` (`-Deo.coverageTracking=false`) and `.github/workflows/codecov.yml:29` (`-Deo.coverageTracking=true`).

This PR's CI being green tells us Maven does not reject the now-unknown element, it just ignores it — which is exactly what makes this dangerous. The net effect on `codecov.yml` is that `-Deo.coverageTracking=true` becomes a no-op and nothing else sets the new `eo.coverageFile` mojo parameter, so after this merge that job instruments **nothing**, silently. Master currently instruments there. Please update the three files as part of this PR.

eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
```diff
@@ -64,36 +66,32 @@ public final class MjTranspile extends MjSafe {
...
-     *  is chosen because Codecov and Coveralls consume it directly.
+     * The file to record EO object coverage hits into. When it is
+     * not set (the default), no instrumentation happens; when set,
+     * every located object in the generated Java is wrapped into
+     * {@code PhCoverage}, which appends a hit to this file the first
+     * time the object is touched at runtime. Setting the same
+     * {@code eo.coverageFile} system property also reaches the JVM
+     * that later runs the instrumented code, since {@code eo-runtime}
+     * forwards it to surefire's {@code systemPropertyVariables} (see
+     * its {@code pom.xml}); one setting is enough end to end. This
```
Favixx: "one setting is enough end to end" does not hold against the current `eo-runtime/pom.xml`. The `<eo.coverageFile>` on line 197 there is a surefire `systemPropertyVariables` entry hardcoded to `${project.build.directory}/coverage.txt`; it is not derived from the Maven property this `@Parameter` reads.

So `mvn -Deo.coverageFile=/tmp/hits` makes the transpiler write `/tmp/hits.manifest` while the runtime appends its hits to `target/coverage.txt`, and `MjCoverageReport` then merges a manifest against an unrelated (or empty) hits file and reports 0%. Either change the surefire value to `${eo.coverageFile}` so both ends really do follow one setting, or drop the claim and document the two settings separately.

eo-maven-plugin/src/main/java/org/eolang/maven/CoverageManifest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Selects exactly the elements that end up wrapped into {@code
+     * PhCoverage}. This is the same {@code @line}/{@code @pos}/{@code @loc}
+     * condition the final {@code to-java.xsl} shift itself tests, applied
+     * instead to the XMIR right before that shift runs, so the manifest is
+     * derived from that structure directly, not by pattern-matching the
+     * Java text {@code to-java.xsl} emits.
+     */
+    private static final String LOCATED =
```
Favixx: This XPath is broader than the condition it is meant to mirror. In `to-java.xsl` the `$coverage='true' and @line and @pos and not(contains(@loc, '+'))` test (line 651) sits inside `<xsl:template match="*" mode="located">` (line 629), and `mode="located"` is only ever applied from the two object templates at lines 558 and 593 — i.e. only `<o>` elements whose `@base` is set and does not start with `.`, or whose `@base` starts with `.` and which have children.

`//*[...]` matches every element in the document carrying those attributes, formations (`<o>` with no `@base`) included. Those never receive a `PhCoverage` wrapper, so they can never appear among the hits, yet they end up in the manifest and inflate `LF`. The reported percentage comes out lower than reality, and #5672 wants to fail the build on that number — so the denominator has to be exact. Please mirror the two match patterns here, or better, see my comment on `Transpiling.java`.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -264,6 +283,11 @@ private int transpiled(final TjForeign tojo) throws IOException {
             rewrite.compareAndSet(false, true);
             new Saved(transform.apply(xmir).toString(), target).value();
         }
+        if (this.coverage != null) {
+            new CoverageManifest(this.coverage).scan(
+                new Xsline(Transpiling.LOCATED.get()).pass(xmir)
```
Favixx: This runs a **second full transpile pass** for every source whenever coverage is on. `LOCATED` is `copyOf(XSLS, XSLS.length - 1)`, and `compiled()` (line 367) builds its train from the very same array, so all eight stylesheets before `to-java.xsl` are now applied twice per XMIR.

It is worse than 2x: on a cache hit `transform.apply(xmir)` never runs, but this pass still does, so enabling coverage effectively disables the benefit of the transpile cache.

Since the intermediate tree you want is precisely what the real train produces one shift before the end, consider restructuring `transpilation()` to keep that result: pass `LOCATED` once, hand the result to `CoverageManifest.scan`, then apply only `to-java.xsl` to it. One pass, and the manifest is derived from exactly the tree that produced the Java — which also makes the duplicated XPath in `CoverageManifest` unnecessary.

eo-maven-plugin/src/main/java/org/eolang/maven/CoverageManifest.java
```diff
@@ -0,0 +1,105 @@
...
+                    element.xpath("@line").get(0)
+                )
+            );
+        }
+        if (!lines.isEmpty()) {
+            final StringBuilder joined = new StringBuilder(lines.size() * 32);
+            for (final String line : lines) {
+                joined.append(line);
+            }
+            Files.write(
```
Favixx: The javadoc leans on "the OS append guarantee for a single `write()` syscall", but that guarantee is much narrower than what happens here: this buffer holds every location found in one source, which for a large `.eo` file is many kilobytes, and nothing makes the JDK issue it as a single syscall. `PhCoverage` gets away with it because it writes one short line at a time.

A torn line here is silent and only shows up later as a parse failure in `CoverageReport` (see my comment there). Since `Threaded` already collects a result per tojo, the simplest fix is to have `scan` return the lines and write them all once in `Transpiling.exec()` after the pool drains — then `reset()` isn't needed either.

eo-maven-plugin/src/main/java/org/eolang/maven/CoverageReport.java
```diff
@@ -0,0 +1,147 @@
...
+    String text() throws IOException {
+        final Set<String> touched = this.touched();
+        final SortedMap<String, SortedMap<Integer, Boolean>> files = new TreeMap<>();
+        for (final String line : Files.readAllLines(this.manifest)) {
+            if (line.isEmpty()) {
+                continue;
+            }
+            final String[] parts = line.split("\t", 2);
+            final String loc = parts[0];
+            final int number = Integer.parseInt(parts[1]);
```
Favixx: `line.split("\t", 2)` followed by `Integer.parseInt(parts[1])` has no guard. Any manifest line without a tab throws `ArrayIndexOutOfBoundsException`, and any non-numeric second field throws `NumberFormatException` — both at the very end of a long build, in the `verify` phase, with a stack trace that says nothing about which file is malformed or which line it choked on. Given the concurrent appends above can in principle produce exactly such a line, please validate here and either skip the line or fail with a message that quotes it and names `this.manifest`.

eo-maven-plugin/src/main/java/org/eolang/maven/CoverageReport.java
```diff
@@ -0,0 +1,147 @@
...
+     * The percentage of instrumented lines that were touched, derived from
+     * {@link #text()} itself so there is exactly one place that knows the
+     * LCOV grammar.
+     * @return The percentage, from 0 to 100; 100 when nothing is instrumented
+     * @throws IOException If fails to read the manifest or the hits file
+     */
+    double percentage() throws IOException {
+        int total = 0;
+        int covered = 0;
+        for (final String line : this.text().split(System.lineSeparator())) {
```
Favixx: `percentage()` calls `text()` again — re-reading the manifest and the hits file, re-building the whole `TreeMap` — and then parses its own output back with `startsWith("LF:")` / `startsWith("LH:")`. `MjCoverageReport.exec()` calls `text()` and then `percentage()`, so the entire report is built twice per run.

The stated goal ("exactly one place that knows the LCOV grammar") is worth keeping, but it doesn't require a round trip through text: accumulate `total` and `covered` while building in `text()` and expose them, or memoize the built text.

eo-maven-plugin/src/main/java/org/eolang/maven/ObjectSources.java
```diff
@@ -0,0 +1,93 @@
...
+     * @param locator The locator, e.g. {@code Φ.string.scanf.tokenize}
+     * @return The source
+     */
+    String source(final String locator) {
+        String rest = locator;
+        if (rest.startsWith(ObjectSources.ROOT)) {
+            rest = rest.substring(ObjectSources.ROOT.length());
+        }
+        String longest = null;
+        for (final String id : this.ids.keySet()) {
```
Favixx: This walks the entire id set for every locator, so resolving the whole manifest is O(locations x objects). For `eo-runtime` that is tens of thousands of locations against thousands of registered objects, all inside the `verify` phase.

Locators are dot-separated and the match you want is the longest id that is a prefix — so you can just trim the last segment repeatedly and do a `Map.get` on each candidate, stopping at the first hit. That is O(depth) per locator, needs no `longer()` helper, and drops the `owns()`/`longer()` pair entirely.

eo-maven-plugin/src/main/java/org/eolang/maven/ObjectSources.java
```diff
@@ -0,0 +1,93 @@
...
+            if (ObjectSources.owns(rest, id) && ObjectSources.longer(longest, id)) {
+                longest = id;
+            }
+        }
+        if (longest == null) {
+            throw new IllegalStateException(
```
Favixx: A single unresolvable locator aborts the entire report with `IllegalStateException`, i.e. it fails the build at `verify` after everything else has already passed. A locator can legitimately point at something outside the `eo-foreign` catalog (pulled objects, generated tests), and losing one line out of thousands is not worth failing a build over. Logging a warning and skipping the line would be a better trade — and if you do want it fatal, the message should list a couple of the catalog ids so the reader can see what it was matched against.

eo-maven-plugin/src/main/java/org/eolang/maven/MjCoverageReport.java
```diff
@@ -0,0 +1,111 @@
...
+    defaultPhase = LifecyclePhase.VERIFY,
+    threadSafe = true
+)
+public final class MjCoverageReport extends MjSafe {
+
+    /**
+     * The same raw coverage hits file passed to the {@code transpile} goal.
+     * @checkstyle MemberNameCheck (7 lines)
+     */
+    @Parameter(property = "eo.coverageFile", required = true)
```
Favixx: `required = true` means Maven rejects the goal with its own generic "parameter is missing or invalid" message before `exec()` is ever entered. So the carefully worded error on line 63 can only fire in the narrower case where the file *is* configured but the manifest is missing — not in the case the PR description advertises ("fails with a clear message if the project was never transpiled with `coverageFile` set"). If the friendly message is the point, drop `required` and check for null explicitly.

eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
```diff
@@ -129,4 +127,19 @@ public void exec() throws IOException {
...
+
+    /**
+     * The path of the given file, or NULL when the file is not set.
+     * @param file The file, possibly NULL
+     * @return The path, or NULL
+     */
+    private static Path path(final File file) {
+        final Path result;
+        if (file == null) {
+            result = null;
```
Favixx: `path()` returns NULL to mean "coverage is off", and `Transpiling` then has to branch on `!= null` in three separate places (lines 226, 286 and 351). This codebase generally avoids NULL as a sentinel. `Optional<Path>` would work, but a small `Coverage` object with a do-nothing implementation would fit the style better and would remove all three checks — `exec()` and `transpiled()` would just call `reset()`/`scan()` unconditionally.

eo-maven-plugin/src/main/java/org/eolang/maven/CoverageManifest.java
```diff
@@ -0,0 +1,105 @@
...
+ * The full set of locations the transpiler wrapped into {@code PhCoverage},
+ * one line per location: symbolic locator and source line, tab-separated.
+ * Built once per {@code transpile} run (see {@link Transpiling}), then read
+ * back by {@link MjCoverageReport}, which merges it against the raw hits
+ * recorded at runtime by {@code PhCoverage} itself, to tell which
+ * instrumented lines were actually touched. The manifest does not record a
+ * source file for each location: {@link MjCoverageReport} resolves that by
+ * matching a locator's leading object id against the already-existing
+ * {@code eo-foreign} catalog, instead of duplicating that mapping here.
+ * @since 0.58
```
Favixx: `@since 0.58` — 0.58 is long released (latest tag is 0.61.3), so this should carry the next version instead. Same tag is on `CoverageReport`, `ObjectSources`, `MjCoverageReport` and the three new test classes.

eo-maven-plugin/src/test/java/org/eolang/maven/MjCoverageReportTest.java
```diff
@@ -0,0 +1,91 @@
...
+            .with("coverageFile", hits.toFile())
+            .with("foreign", temp.resolve("eo-foreign.csv").toFile())
+            .with("foreignFormat", "csv")
+            .with("project", new MavenProjectStub())
+            .execute();
+        MatcherAssert.assertThat(
+            "the LCOV tracefile must have one record per instrumented source",
+            new TextOf(temp.resolve("coverage.info")).asString(),
+            Matchers.allOf(
+                Matchers.containsString("SF:"),
```
Favixx: This is the one test that exercises the whole chain, and it only asserts that the tracefile contains `SF:` and `end_of_record` — it would pass unchanged if every single line were reported as uncovered, which is precisely the failure mode the merge logic can have. The test writes exactly one hit for a known locator, so it can assert the corresponding `DA:<line>,1` and an `LH:` of at least 1.

eo-maven-plugin/src/test/java/org/eolang/maven/MjCoverageReportTest.java
```diff
@@ -0,0 +1,91 @@
...
+    void failsWhenNoManifestWasEverWritten(@Mktmp final Path temp) {
+        MatcherAssert.assertThat(
+            "the failure must explain that no coverage manifest was found",
+            Assertions.assertThrows(
+                IllegalStateException.class,
+                () -> new Moja<>(MjCoverageReport.class)
+                    .with("coverageFile", temp.resolve("never-transpiled.txt").toFile())
+                    .execute(),
+                "coverage-report must fail when the project was never transpiled with a coverage file"
+            ).getCause(),
```
Favixx: The assertion message promises "the failure must explain that no coverage manifest was found", but the assertion itself only checks the exception type and never looks at the message. Please assert on the text too, otherwise the message this test exists to protect can be gutted without anything turning red.

eo-maven-plugin/src/test/java/org/eolang/maven/CoverageManifestTest.java
```diff
@@ -0,0 +1,77 @@
...
+    void derivesItsPathFromTheHitsFile(@Mktmp final Path temp) {
+        MatcherAssert.assertThat(
+            "the manifest must live next to the hits file, with a .manifest suffix",
+            new CoverageManifest(temp.resolve("hits.txt")).path(),
+            Matchers.equalTo(Paths.get(String.format("%s.manifest", temp.resolve("hits.txt"))))
+        );
+    }
+
+    @Test
+    void extractsOneLinePerLocatedElement(@Mktmp final Path temp) throws Exception {
```
Favixx: Good that both the `+`-suffixed locator and the element without `@line`/`@pos` are covered here. What is missing is a case that pins down *which elements* qualify: nothing asserts the number of lines written, so the over-selection I flagged in `CoverageManifest.LOCATED` cannot be caught by this test. An element that carries `@line`/`@pos`/`@loc` but is not one of the two `<o>` shapes `to-java.xsl` wraps — a formation with no `@base` is the obvious one — would make the intended contract explicit.

Favixx (changes requested):
Thanks for picking up #5671 — the LCOV direction is right, and `CoverageReportTest` in particular is a nice, thorough set of cases (missing hits file, nothing instrumented, two locators on one line). I can't merge it as it stands though: as far as I can tell this PR silently switches off the coverage instrumentation that master currently has working, and the manifest counts more locations than the transpiler actually wraps, which makes the reported percentage wrong in the one direction that matters for #5672. Details inline.

Main blockers:

1. The `coverageTracking` -> `coverageFile` rename leaves `eo-runtime/pom.xml` and both workflows still naming the old parameter, so `codecov.yml` stops instrumenting anything.
2. Nothing sets the new `eo.coverageFile` **mojo** parameter in `eo-runtime/pom.xml` — the existing `<eo.coverageFile>` there is a surefire system property, hardcoded to a different path, so the two ends can disagree on the file.
3. `//*[@line and @pos and not(contains(@loc, '+'))]` is broader than the two `mode="object"` templates that actually reach the wrapping code in `to-java.xsl`.
4. The manifest is built by running the whole transpile train a second time per source, even on cache hits.

Everything else is smaller. Note I reviewed by reading, I did not build the branch locally — please sanity-check 1 and 2 against a real `mvn -Pjacoco install` before pushing. Also, `pr-size` is red (883 additions): this splits naturally into "manifest + transpiler wiring" and "report + mojo", if you want two green PRs instead of one big one.

0crat:
@yegor256 Thanks for the review! You've earned +13 points for this: +12 as a basis; +6 for the 639 hits-of-code that you reviewed; -5 for very few (1) comments.  Your running score is [+2698](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +18 points for this: +12 as a basis; +6 for the 639 hits-of-code that you reviewed.  Your running score is [+76](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5699, opened 2026-07-21
Always emit Q. root for rooted signatures in printer

### LINKED ISSUE #5698: Printer drops `Q.` root from single-segment `/` signatures
The `eo:signature` function in `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:55` renders a signature after the `/` inconsistently. A single-segment rooted name such as `Φ.number` is printed bare, e.g. `[] > size /number` (see `eo-runtime/src/main/eo/malloc.eo:78` and `bytes.eo:27`), while a multi-segment name keeps its root, e.g. `[size] > resized /Q.malloc.of.allocated`. The two forms print differently for no visible reason.

This is wrong because the printer should be uniform: every rooted signature ought to read the same way. Emitting `/Q.number` is also strictly more robust than `/number`, since `Q.number` re-parses directly to `Φ.number` instead of relying on the parser re-homing a bare `number` to its root. The current single-segment special case exists only to lean on that reparse behaviour.

The fix is one line: in `eo:signature`, drop the inner `if (contains($rest, '.'))` branch and always emit `concat('Q.', $rest)` when the signature starts with the `Φ.` root. The `else $sig` pass-through for unrooted names stays. Update the round-trip packs that expect the bare form (`eo-printer/src/test/resources/org/eolang/printer/print-packs/yaml/idiomatic.yaml` lines 19 and 43) to `/Q.number`.

yegor256:
Closes #5698

## Problem

`eo:signature` in `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl` rendered signatures after `/` inconsistently. A single-segment rooted name such as `Φ.number` was printed bare (`/number`), while a multi-segment name kept its root (`/Q.malloc.of.allocated`). The single-segment special case leaned on the parser re-homing a bare `number` back to its root.

## Fix

Dropped the inner `if (contains($rest, '.'))` branch so that whenever a signature starts with the `Φ.` root it always emits `concat('Q.', $rest)`. Rooted signatures now read uniformly and `Q.number` re-parses directly to `Φ.number` instead of relying on re-homing. The `else $sig` pass-through for unrooted names is unchanged.

```
- if (starts-with($sig, $root)) then (if (contains($rest, '.')) then concat('Q.', $rest) else $rest) else $sig
+ if (starts-with($sig, $root)) then concat('Q.', $rest) else $sig
```

## Tests

Updated the round-trip print packs that expected the bare form to the `Q.`-rooted form (`/Q.number`, `/Q.bytes`, `/{Q.string}`, `/{Q.string Q.a.b.c}`):

- `idiomatic.yaml` (the pack named in the issue)
- `test-attribute.yaml`, `throws-test-attribute.yaml`, `hybrid-phi-test.yaml`, `void-type-annotation.yaml` (further packs whose single-segment rooted signatures were affected by the same generalization)

All 166 tests in `eo-printer` pass.

https://claude.ai/code/session_01LRqihmrCvuvsEaor4p5rZj

---
_Generated by [Claude Code](https://claude.ai/code/session_01LRqihmrCvuvsEaor4p5rZj)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -43,16 +43,16 @@
   </xsl:function>
   <!--
   Render a stored signature (an atom's "/sig" or a void forma) as its EO
-  surface form. A single-name "Φ.name" re-resolves to its root on
-  reparse, so the implicit root is dropped; a multi-segment
-  "Φ.a.b.c" must stay rooted (a dotted signature is not re-homed) and
-  keeps an explicit "Q." root. Names with no "Φ." root pass through.
+  surface form. A rooted "Φ.name" always keeps an explicit "Q." root, so
+  every rooted signature reads uniformly and re-parses directly back to its
+  "Φ." root instead of relying on a bare name being re-homed. Names with no
+  "Φ." root pass through.
   -->
   <xsl:function name="eo:signature" as="xs:string">
     <xsl:param name="sig" as="xs:string"/>
     <xsl:variable name="root" select="concat($eo:program, '.')"/>
     <xsl:variable name="rest" select="substring-after($sig, $root)"/>
-    <xsl:sequence select="if (starts-with($sig, $root)) then (if (contains($rest, '.')) then concat('Q.', $rest) else $rest) else $sig"/>
+    <xsl:sequence select="if (starts-with($sig, $root)) then concat('Q.', $rest) else $sig"/>
   </xsl:function>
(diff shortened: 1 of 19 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (102 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2700](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5701, opened 2026-07-21
#5673: forward eo.coverageFile to the JVM that runs eo-runtime tests

### LINKED ISSUE #5673: MjTranspile.java file is not correctly forwarded to the JVM

morphqdd:
Closes #5673.

The `coverageFile` parameter of `transpile` only decided, at transpile time, whether `PhCoverage` wrapping got emitted into the generated Java. The path itself was discarded afterwards and never reached the process that later runs the instrumented code, which reads its own `eo.coverageFile` system property instead. So enabling coverage would have needed two separate settings: one for the transpile goal, one for the test JVM.

Added to `eo-runtime/pom.xml`:

- An empty `eo.coverageFile` property (same pattern already used for `argLine`), so `${eo.coverageFile}` always resolves to an empty string instead of staying an unresolved literal when nobody sets it.
- `<eo.coverageFile>${eo.coverageFile}</eo.coverageFile>` in the surefire plugin's `systemPropertyVariables`.

Now a single `-Deo.coverageFile=/path/to/hits.txt` on the command line drives both the transpile-time instrumentation and the JVM that runs the tests, since Maven already exposes any `-D` property as `${property}` throughout the pom, and `eo.coverageFile` is the exact property name `MjTranspile`'s `coverageFile` parameter already binds to.

Removed the resolved `@todo #5466:30min` puzzle from `MjTranspile.java` and updated the field javadoc to describe the end-to-end forwarding.

This is a pom.xml-only change, so there is no Java test for it. Verified manually instead:

- Temporarily wired `<coverageFile>${eo.coverageFile}</coverageFile>` into eo-runtime's transpile execution (not committed, since enabling it for real is scoped to #5672) and ran `mvn -Deo.coverageFile=/tmp/covtest/hits.txt test -Dtest=EOboolTest`. The file was created with 55 real `loc:line:pos` hit lines, proving the property reaches the surefire JVM.
- Ran the same test without `-Deo.coverageFile` to confirm the default (unset, empty-string) case still passes normally.
- `qulice:check` clean, `MjTranspileTest` 40/40 green.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
@@ -70,7 +70,11 @@ public final class MjTranspile extends MjSafe {
      * not set (the default), no instrumentation happens; when set,
      * every located object in the generated Java is wrapped into
      * {@code PhCoverage}, which appends a hit to this file the first
-     * time the object is touched at runtime.
+     * time the object is touched at runtime. Setting the same
+     * {@code eo.coverageFile} system property also reaches the JVM
+     * that later runs the instrumented code, since {@code eo-runtime}
+     * forwards it to surefire's {@code systemPropertyVariables} (see
+     * its {@code pom.xml}); one setting is enough end to end.
      * @todo #5466:60min Turn raw coverage hits into an LCOV report.
      *  Right now this parameter only produces a raw, append-only
      *  {@code loc:line:pos} file: the runtime {@code PhCoverage}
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1349](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5702, opened 2026-07-21
Fix #5700: prefer cheaper inline-phi hybrid even when the one-liner fits width

### LINKED ISSUE #5700: Printer skips cheaper inline-phi hybrid when the one-liner fits width but loses on penalty
`Xmir.toEO()` still keeps the verbose `<phi> > @` layout for an only-φ formation whose φ applies regular arguments, even when the compact hybrid `<phi> > [params] > name` is cheaper. In `eo-runtime/src/main/eo/malloc.eo:80` the `copy` object prints as `[source target length] > copy` / `write > @` / its two arguments, rather than the shorter `write > [source target length] > copy` with the arguments beneath. I reproduced it by running the printer on that snippet in isolation: it emits the verbose `write > @` form, never the hybrid.

The cause is `Pretty.phi()` at `eo-printer/src/main/java/org/eolang/printer/Pretty.java:362`. After #5594 and #5635 the hybrid is built only when `applied && unnamed && flat.map(line -> line.length() > this.width).orElse(true)` — i.e. when the flat one-liner is absent or overflows `WIDTH`. `copy`'s one-liner `write target (read source length) > [source target length] > copy` fits under 80 columns, so the `else` branch returns the one-liner and `shaped()` weighs only {one-liner, verbose}; the one-liner's extra bracket and super-linear application-space surcharge lose to verbose, so verbose is kept. The strictly cheaper hybrid (it drops the `@`, saving the `PHI` charge, and lifts the arguments one indent level — about 36 points at the default weights) is never handed to the penalty vote. This is the residual case #5635 left open: its fix used a width check where its own text prescribed a penalty comparison.

Stop discriminating between the one-liner and the hybrid by `flat.length() > this.width`. When `applied && unnamed`, build the hybrid candidate (`decoratee.hybrid(marker)`) alongside the flat one-liner and return the lower-penalty of the two — or hand both to `shaped()` — so the vote can prefer the hybrid whenever it beats both the one-liner and the verbose shape, not only when the one-liner overflows the width limit.

yegor256:
Fixes #5700.

## Problem

`Pretty.phi()` gated the hybrid inline-phi collapse behind a width check
(`flat.map(line -> line.length() > this.width)`), so an only-φ formation whose
φ applies regular arguments kept the verbose ` > @` / flat one-liner shape
whenever the one-liner fit under 80 columns — even when the strictly cheaper
hybrid (`decoratee > [params] > name` with the arguments beneath) beat it on
penalty. In `eo-runtime/src/main/eo/malloc.eo` the `copy` object was the
reported manifestation. This is the residual case #5635 left open: its fix used
a width check where its own text prescribed a penalty comparison.

## Fix

In `Pretty.phi()`, when the decoratee applies unnamed arguments
(`applied && unnamed`), build **both** the flat one-liner and the hybrid
(`decoratee.hybrid(marker)`) and return the lower-penalty of the two; when the
one-liner cannot be built, use the hybrid unconditionally (unchanged). The
hybrid drops the `@` (saving its `PHI` charge) and lifts the arguments one
indent level, so the penalty vote in `shaped()` can now prefer it whenever it
beats **both** the one-liner and the verbose shape — not only when the one-liner
overflows the width limit.

The `width` field of `Pretty` became unused and was removed; overflow is
already priced in by the `EXCESS` weight inside `Penalty`.

## Tests & formatting

- Added `inline-phi-hybrid-under-width.yaml` — the malloc `copy` case from the
  issue, pinned to the default weights the issue reasons about.
- Updated the print-packs whose canonical layout now legitimately uses the
  cheaper hybrid: `compares-bool-to-bytes`, `empty-string`, `multiline-string`,
  `nested-if-inline`, `times`. All packs still round-trip
  (`printsToParseableEo`), and the full `eo-printer` suite (168 tests) is green.
- Reformatted the `eo-runtime` `.eo` sources to their new canonical layout.
  This is required because the `format` goal fails the build on any file that
  diverges from `Xmir.toEO()`. The reformatting is semantics-preserving (same
  XMIR — only a different candidate among equivalent renderings) and idempotent:
  a second `format` pass reports *"All 151 EO source(s) are formatted
  canonically"*.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XuhgdsjotvaC6rWu39HDjm)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -53,11 +53,6 @@ final class Pretty {
     private final String tab;
 
     /**
-     * The column past which a line overflows, the {@code WIDTH} weight.
-     */
-    private final int width;
-
-    /**
      * Ctor, using the default penalty weights.
      * @param element The {@code <eo>} element
      */
@@ -76,9 +71,6 @@ final class Pretty {
         this.weights = config;
         this.tab = " ".repeat(
             config.getOrDefault(PenaltyKey.STEP, PenaltyKey.STEP.fallback())
-        );
-        this.width = config.getOrDefault(
-            PenaltyKey.WIDTH, PenaltyKey.WIDTH.fallback()
         );
     }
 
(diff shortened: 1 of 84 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (2492 >= 100); -8 for way too many hits-of-code (2492 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2684](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5703, opened 2026-07-21
Fold only-phi hybrid past nested anonymous-object arguments

yegor256:
Follow-up to #5700 / PR #5702, addressing a case spotted in `malloc.eo`.

## Problem

`Xmir.toEO()` kept the verbose ` > @` layout here:

```
++> tests-malloc-for-writes-first-init-value
  eq. > @
    malloc.for
      42
      m > [m]
    42
```

instead of the strictly cheaper hybrid:

```
eq. ++> tests-malloc-for-writes-first-init-value
  malloc.for
    42
    m > [m]
  42
```

The cause is `Node.nameless()`, the recursive guard that decides whether an only-φ formation may collapse into the compact inline-φ hybrid. It rejected **any** non-empty tail in the decoratee's subtree, which over-blocked a nested anonymous object like `m > [m]` (an anonymous formation whose φ is `m`) because its inner ` > @` φ decoratee reads as a "name".

## Fix

A ` > @` suffix marks the **φ decoratee** — exactly what an only-φ formation binds — never a named or auto-named attribute. So a nested anonymous object is a legal argument and must not withhold the fold. `nameless()` now treats a bare ` > @` tail as safe; genuine named/auto-named attributes (`> name`, `>>`) still block it, so the `#5598`/`#5604` guards (`[sum] >>`, `[left] >>`, `[a x i] >>`) are unchanged.

Verified the hybrid is valid, semantically identical EO: it parses without errors and round-trips (`printsToParseableEo`), i.e. it renders the same XMIR as the verbose form.

## Tests & formatting

- Added `hybrid-phi-anonymous-argument.yaml` (the malloc case from the report).
- Updated `inline-phi-compact-tuple.yaml`, whose canonical layout now legitimately uses the hybrid (the anonymous `seq * > [m]` argument no longer blocks the outer fold).
- Full `eo-printer` suite (172 tests) green.
- Reformatted 39 `eo-runtime` `.eo` sources to their new canonical layout (required by the `format` goal). Semantics-preserving and idempotent: a second `format` pass reports *"All 151 EO source(s) are formatted canonically."*

---
_Generated by [Claude Code](https://claude.ai/code/session_01XuhgdsjotvaC6rWu39HDjm)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Node.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Node.java
@@ -124,17 +124,25 @@ final class Node {
      * Whether this node carries no name suffix anywhere in its subtree.
      *
      * <p>A line is "named" when its {@code tail} holds a {@code > name},
-     * {@code > [params]} or {@code >>} suffix. This walks the node and all
-     * its descendants, so a named line nested below the top level — inside
-     * a tuple, a dispatch, or an application — is caught too. It decides
-     * whether a decoratee's whole subtree is safe to fold into a compact
-     * only-phi formation, which binds nothing but its {@code φ} decoratee
-     * (issue #5604).</p>
+     * {@code > [params] > name} or {@code >>} suffix — a named or auto-named
+     * attribute. This walks the node and all its descendants, so a named line
+     * nested below the top level — inside a tuple, a dispatch, or an
+     * application — is caught too. It decides whether a decoratee's whole
+     * subtree is safe to fold into a compact only-phi formation, which binds
+     * nothing but its {@code φ} decoratee (issue #5604).</p>
+     *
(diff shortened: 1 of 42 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1371 >= 100); -8 for way too many hits-of-code (1371 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2704](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5711, opened 2026-07-21
#5488: rewrite scanf as pure EO, drop the native Java atom

### LINKED ISSUE #5488: `sscanf` atom in `tt` package should be converted to pure EO to eliminate native dependencies
The `sscanf` object is declared as a native atom at `eo-runtime/src/main/eo/tt/sscanf.eo:22` (`[format read] > sscanf /tuple`, empty body). It reads a `%d`/`%f`/`%s` format against an input string and returns a tuple of parsed values. Nothing in that behaviour is intrinsically native.

Every primitive `sscanf` needs is already implemented in pure EO inside the same `tt` package: literal matching and field extraction via `slice`, `split`/`nsplit`, `index-of`, `starts-with`, `ends-with`, `at`, `is-digit`; character codes via `as-ascii`; and the numeric conversions are just the inverse of the already-pure-EO `as-decimal`/`as-fixed`/`as-char`, using ordinary `number` arithmetic (`plus`/`times`/`div`). The tuple is assembled with `*`/`.with`, and the unknown-specifier case (e.g. `%l`) throws through a `cant-*` fallback, exactly as the existing `throws-on-sscanf-with-wrong-format` test expects.

There is even a circular dependency that makes the atom particularly worth removing: `tt.as-number` (`eo-runtime/src/main/eo/tt/as-number.eo:22`) is pure EO but delegates the actual numeric parse to `sscanf "%f" text`. So the only reason string-to-number parsing is still native is this atom; a pure-EO `sscanf` closes that loop and lets the whole parsing stack be atom-free.

Proposed fix: give `sscanf` a pure-EO body — tokenize `format` into literal chunks and `%d`/`%f`/`%s` specifiers, walk `read` matching literals with `starts-with`/`slice`, extract each field up to the next literal with `index-of`, and fold digits into a `number` (`acc.times 10 .plus (as-ascii c .minus 48)`, with sign and, for `%f`, a fraction scaled by `10^k`); keep the existing tests in `sscanf.eo:24-114` as the specification (including oversized-int rejection, cf. #5302). This mirrors the ongoing atom-minimization for the integer types (#5467-#5471) and the old meta #715.

morphqdd:
Closes #5488.

`string.scanf` used to be a native Java atom (`EOscanf.java`), plus a declared `+rt node` dependency on a matching native implementation in `eo2js-runtime`. Rewrote it entirely in EO, eliminating both native dependencies:

- A hand-rolled recursive tokenizer walks the format string character by character, building a regex pattern (escaping metacharacters in literal segments, substituting `(...)` capture groups for `%d`/`%f`/`%s`) plus an ordered tuple of the specifiers seen (`d`/`f`/`s`).
- `string.regex` (already pure EO) does the actual matching against the input string.
- Captured groups are converted back to `number` or `string` per specifier, using a small hand-rolled `as-integer`/`as-float`/`fold-digits` (no native parsing atom needed either).

Deleted `EOscanf.java` and its two Java-level test classes, `ScanfDanglingPercentTest` and `ScanfNegativeIntegerTest`. Both scenarios (dangling percent at the end of the format, signed integers) are already covered by the EO test suite embedded in `scanf.eo` (`on-scanf-with-dangling-percent`, `tests-scanf-with-negative-int`, `tests-scanf-with-positive-signed-int`), and testing a pure-EO object through low-level `PhApplication` calls from Java is inconsistent with how every other pure-EO object in this codebase is tested (embedded `++>` EO tests).

Also removed the `+rt jvm`/`+rt node` meta declarations from `scanf.eo`, since there is no runtime-specific atom left to declare for either platform.

Along the way, found and fixed a real bug in my own first implementation attempt, worth calling out for review: this codebase's `tuple` is built as a right-leaning chain, so `.head`/`.tail` are the reverse of what the names suggest for a normal list. A tuple built by successive `.with(x)` appends has `.head` return the LAST appended element and `.tail` return everything except that last element (not "all but the first", as the naming implies). My first version of the group-collecting recursion in `extracted`/`convert` used `.head`/`.tail`, which silently produced wrong results only for multi-specifier formats (single-specifier formats worked by accident, since order does not matter with one element). Rewrote it to walk the specifiers tuple by explicit index with `.at()` instead of `.head`/`.tail` recursion.

Verified:
- Full `eo-runtime` suite (`mvn clean install -PskipITs`, matching CI): 1702 tests, 0 failures.
- `string.scanf`'s own test suite: 15/15 green.
- `string.as-number` (dependent functionality): 6/6 green.
- `qulice:check` on `eo-runtime`: clean.
- Canonical formatting (`-Deo.autoFix`, i.e. the `format` goal in the normal build): ran without corrupting the file this time; re-verified all tests green after autoFix reformatted it.


0crat:
@morphqdd Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (518 >= 100); -8 for way too many hits-of-code (518 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1365](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5712, opened 2026-07-21
eo-printer: deterministic formation attribute ordering (#5706)

### LINKED ISSUE #5706: `eo-printer` does not emit attributes in deterministic order, resulting in order-dependent diffs
The `eo-printer` emits the attributes of a formation in the order they appear in the source (document order), with no canonical ordering. The children are laid out verbatim at `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:159` (`<xsl:apply-templates select="o[not(eo:void(.)) or @local or @types]" mode="tree"/>`), so two formations that bind the same attributes in a different textual order print differently.

This produces noisy, order-dependent diffs and makes formatted output harder to scan, which undermines the cross-module format check tracked in #5577: reformatting can only be a fixpoint if attribute order is deterministic. Reordering *bound* attributes is safe because a formation's attributes are all mutually in scope regardless of textual position; only void (parameter) order is positional, and those live in the `[params]` head, which stays untouched.

Proposed order for the children emitted at that `apply-templates`: (1) the vertical void body lines (`? >> name`, `? > name /{…}`) first, keeping the bracket params in the head as-is; (2) the `@` decoratee (`@name = φ`); (3) every other non-test bound attribute, sorted alphabetically by `@name`; (4) the test attributes last, sorted alphabetically by `@name`, where a test is `eo:test-attr(.)` (a name starting with `+` or `-`, per `eo-parser/.../_funcs.xsl:28`). Concretely, add `<xsl:sort>` keys (a small computed rank per the four buckets, then `@name`) to that one `<xsl:apply-templates>` so the tree is built in the canonical order; `Pretty` already keeps the blank line before a test via the `test` flag.

yegor256:
Fixes #5706.

## Problem

`eo-printer` emitted formation attributes in source order rather than a canonical one, producing order-dependent diffs, non-idempotent formatting, and undermining the cross-module format-verification effort.

## Change

Added `<xsl:sort>` keys to the formation-body `<xsl:apply-templates>` in `eo-printer/.../print/to-eo-tree.xsl`. Formation attributes now print in a canonical, stable order:

1. Vertical void body lines (`? >> name`, `? > name /{…}`) first
2. The decoratee (`@name = φ`, i.e. `… > @`)
3. All other bound attributes, alphabetically by name
4. Test attributes (`++>` / `-->`) last, alphabetically by name

Two kinds of body deliberately keep their **source order** (the sort key collapses to a constant and `xsl:sort` is stable):

- **Application arguments** — positional, so their order carries meaning. Only abstract formations are reordered.
- **Bodies containing a pipe continuation** (`| args`, §3.14) — the `|` line must stay directly below the named formation it applies to (R-3.14.7); reordering would strand it and break re-parsing.

## Reformatting existing sources

Because `eo-runtime` runs the `format` goal in check mode during its build, its `.eo` sources must match the canonical layout. Reformatted all 109 divergent runtime `.eo` files via `eo:format -Deo.autoFix=true` (pure attribute reordering — insertions equal deletions).

## Verification

- `XmirTest` green (134 tests): both `printsToEo` and the `printsToParseableEo` idempotency/round-trip check pass. Updated the 9 affected printer test packs.
- `MjPrintTest` (64) and `MjFormatTest` (5) unchanged and green.
- `mvn -pl eo-runtime compile` succeeds — reformatted sources parse, pass the format check, transpile, and compile.
- A second `format` check pass reports zero divergence, confirming idempotency.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CktVBV56MhWdxcoaZP1haj)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -156,7 +156,25 @@
       <xsl:attribute name="data">
         <xsl:value-of select="if (eo:has-data(.)) then 'yes' else 'no'"/>
       </xsl:attribute>
-      <xsl:apply-templates select="o[not(eo:void(.)) or @local or @types]" mode="tree"/>
+      <!--
+      Formation attributes are emitted in a canonical order so that
+      textually-reordered but logically-identical bodies print the same
+      way (idempotent, diff-friendly output — #5706). The ordering is:
+      vertical void body lines first, then the decoratee (@name = φ),
+      then every other bound attribute alphabetically by name, and test
+      attributes last (also alphabetically). Two kinds of body keep
+      their source order untouched: application arguments (positional,
+      so their order carries meaning) and any body carrying a pipe
+      continuation ("| args", §3.14), whose "|" line must stay directly
+      below the named formation it applies to (R-3.14.7) — reordering
+      would strand it. In both cases the sort key collapses to a
+      constant, and since xsl:sort is stable the original order stands.
+      -->
+      <xsl:variable name="sortable" select="eo:abstract(.) and empty(o[@pipe])"/>
+      <xsl:apply-templates select="o[not(eo:void(.)) or @local or @types]" mode="tree">
+        <xsl:sort data-type="number" select="if (not($sortable)) then 0 else if (eo:void(.)) then 1 else if (@name = $eo:phi) then 2 else if (eo:test-attr(.)) then 4 else 3"/>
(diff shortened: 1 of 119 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (8298 >= 100); -8 for way too many hits-of-code (8298 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2724](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5577, opened 2026-07-16
`.eo` format check isn't applied across all modules, leading to potential formatting drifts


yegor256:
The `format` goal added in #5516 verifies that every `.eo` source is in canonical layout and fails the build when it is not (check mode, `eo.autoFix` off). Wiring it into the build makes formatting a hard gate, so no unformatted `.eo` can ever be merged — the same guarantee `spotless:check` or `gofmt -l` give other ecosystems.

Right now the goal is only about to be enabled for `eo-runtime`. But `.eo` sources also live in `eo-maven-plugin` (10 files), `eo-parser` (3), and `eo-integration-tests` (2), and none of them are checked, so drift can creep back in through those modules unnoticed.

We should add the `format` goal to the `eo-maven-plugin` execution of every module that carries `.eo` sources, running it in check mode as part of the normal build so CI enforces canonical layout project-wide. Each module's sources need a one-time reformat first (via `-Deo.autoFix`) so the check passes; after that the gate keeps them canonical automatically.

VasilevNStas:
@yegor256 @morphqdd 

Findings from an attempted implementation — the plan in this issue does not go through cleanly for `eo-parser`, and one part needs a maintainer decision.

## 1. `eo-parser` cannot host the plugin in the reactor (cycle)

`eo-maven-plugin` depends on `eo-parser` as a **library** (it parses EO sources). Adding the plugin to `eo-parser`'s own build therefore creates a reactor cycle:

```
org.eolang:eo-parser --> org.eolang:eo-maven-plugin (build plugin)
org.eolang:eo-maven-plugin --> org.eolang:eo-parser (library dependency)
```

`mvn install` from the root fails with `The projects in the reactor contain a cyclic reference`. The plugin does work when the module is built **standalone** (`mvn -f eo-parser/pom.xml`, resolving the plugin from the local repo), so a format check for eo-parser is possible — but not "as part of the normal (reactor) build" as the issue proposes.

## 2. The one-time reformat is non-trivial

Running `-Deo.autoFix` on the three `eo-parser` test files rewrites them into the printer's canonical form, and that is a structural rewrite, not whitespace: e.g. `factorial.eo` went from

```
[n] > f
  if. > @
    n.lt 1
    n
    times.
      n
      f (minus. n 1)
```

to

```
if. > [n] > f
  n.lt 1
  n
  n.times
    f
      minus. n 1
```

[...]

PR #5713, opened 2026-07-21
Merge monikers back onto bare references in eo-printer

### LINKED ISSUE #5707: eo-printer incorrectly expands attributes into separate bindings and references
The printer expands a "moniker" — a bound attribute written at its bare-reference use site — into a standalone binding line plus a bare reference. In `eo-runtime/src/main/eo/number/abs.eo:13` the definition sits inside the `if.` argument block as `number num.as-bytes > value`, but `Xmir.toEO` re-emits it as a bare `value` argument with a trailing `number num.as-bytes > value` binding, one line longer:

```
[num] > abs
  if. > @
    value.gte 0
    value
    value.neg
  number num.as-bytes > value
```

Both spellings parse to the same object graph: in the parsed XMIR `value` is an attribute of `abs` (sibling of `φ`) and the `if.` application holds a bare `ξ.value` reference, because the parser hoists the `> value` name to the enclosing formation regardless of where it is written. So the moniker form is purely a shorter, human-preferred rendering of the identical XMIR, and the round-trip drift it causes fights the format-fixpoint goal of #5577. Relocating the binding back to a bare-reference site is always safe as long as no *formation* sits between the attribute's owner and that reference — the name still binds to the same formation (`if.` is an application, not a formation, so `abs` qualifies); a method-dispatch use such as `value.gte` cannot host the name and is left as a reference.

Suggested fix: add an XSL pass to the `FOR_EO` train in `eo-printer/src/main/java/org/eolang/printer/Xmir.java:43-55` (before `to-eo-tree.xsl`, modeled on `inline-cactoos.xsl`). For each formation, take every named bound attribute that is neither void, `φ`, nor a test (`eo:test-attr`) and that has a bare reference `ξ.<name>` — base exactly the reference, no trailing path or args — at a descendant reached through no intervening formation; merge the binding onto that reference node (keep the reference's `@as` position, add the binding's `@name`, `@base` and children) and drop the standalone binding. When several bare references exist, pick the first; when none exists, leave the binding untouched.

yegor256:
The printer used to expand a "moniker" — a named binding written at its bare-reference site — into a standalone binding plus a bare reference. So `number num.as-bytes > value` written inside an `if.` block came back out one line longer, with the binding relocated below the block and a bare `value` left in its place.

Both spellings parse to the same object graph, since the parser hoists a `> name` to the enclosing formation wherever it is written. The expansion is therefore pure round-trip drift that fights the format-fixpoint goal of #5577. This adds a `merge-monikers` XSL pass to the `FOR_EO` train, before `to-eo-tree`, that relocates each eligible formation binding back onto its first bare `ξ.<name>` reference reachable through no intervening formation, so the shorter spelling survives the round-trip.

Bindings that cannot safely host the merge are left alone: method-dispatch uses like `value.gte`, references carrying arguments, references that are themselves named attributes such as `temp > @`, and pipe-floated bindings. Two existing print-packs (`dataless`, `idiomatic`) were updated to the now-canonical moniker form, and new packs cover the transform and its round-trip.

Closes #5707


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain any of the merge conditions if something is unclear.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1452 >= 100); -8 for way too many hits-of-code (1452 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2732](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5714, opened 2026-07-21
#5710: fix inverted comparison in FpIfTargetOlder.isAfter()

### LINKED ISSUE #5710: `FpIfTargetOlder.isAfter()` gives incorrect result contradicting its own Javadoc description
`FpIfTargetOlder.isAfter()` at `eo-maven-plugin/src/main/java/org/eolang/maven/FpIfTargetOlder.java:62-66` returns `first.mtime.isAfter(second.mtime)`, i.e. true when the *first* path is *newer*. Its own Javadoc right above it (lines 55-59) says the opposite: "Returns true if first given path is older in terms of last modified time." The constructor at line 33 calls `isAfter(dest, source)` and stores the result in a variable named `older`, which is then handed to `FpFork` (condition `true` → apply the first, rebuild-oriented footprint, per `FpFork.java` and the sibling `FpIfTargetExists.java`).

Because the comparison is inverted, a genuinely stale target (target mtime before source mtime) makes `isAfter(dest, source)` evaluate to `false`, so `older` is wrongly `false` and the second footprint is chosen instead of the rebuild path — the debug log at lines 40-45 even prints "is newer than source" for a target that is actually older. The class currently has no callers in `src/main` and no test class, which is presumably why this has gone unnoticed, but as written it is simply wrong against its own documented contract, and will misbehave the moment something wires it into the caching pipeline.

Fix by swapping the comparison so it matches the Javadoc, e.g. `Files.getLastModifiedTime(second).toInstant().isAfter(Files.getLastModifiedTime(first).toInstant())` (or equivalently compare `first` with `isBefore(second)`), and add a small `FpIfTargetOlderTest` with two temp files of controlled mtimes asserting the fork picks the correct footprint in both directions.

morphqdd:
Closes #5710.

`FpIfTargetOlder.isAfter(dest, source)` returned `first.mtime.isAfter(second.mtime)`, true when the target is newer than the source. Its own Javadoc says the opposite: "Returns true if first given path is older in terms of last modified time." The result was stored in a variable named `older` and fed into `FpFork`, where `true` picks the first, rebuild-oriented footprint. Because the comparison was inverted, a genuinely stale target made `older` false, so the wrong footprint got chosen, and the debug log even printed "is newer than source" for a target that was actually older.

Renamed the method to `isOlder` and swapped `isAfter` for `isBefore`, so it now actually returns true when the target is older, matching both its own Javadoc and the `older` variable name at the call site. This also matches the naming convention of the sibling `FpIfTargetExists` (a boolean method named after the condition it computes).

Added `FpIfTargetOlderTest` with two cases, both using files with controlled last-modified times:
- target older than source: expects the first (rebuild) footprint.
- target newer than source: expects the second footprint.

Verified the test actually catches the bug: temporarily reverted the fix, confirmed both new tests fail (with results exactly swapped), then restored the fix and confirmed both pass.

The class had no callers in `src/main` and no test class before this, which is presumably why the inversion went unnoticed; it will misbehave the moment something wires it into the caching pipeline.

Verified: `qulice:check` clean (including jtcop), full `eo-maven-plugin` test suite passes with no new failures. The 9 `MjPrintTest` failures on this branch are pre-existing on unmodified master too (confirmed via `git stash` and rerunning on master), unrelated to this change.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/FpIfTargetOlder.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/FpIfTargetOlder.java
@@ -30,7 +30,7 @@ public final class FpIfTargetOlder extends FpEnvelope {
             new FpFork(
                 (source, target) -> {
                     final Path dest = destination.apply(target);
-                    final boolean older = FpIfTargetOlder.isAfter(dest, source);
+                    final boolean older = FpIfTargetOlder.isOlder(dest, source);
                     if (older) {
                         Logger.debug(
                             FpIfTargetOlder.class,
@@ -59,8 +59,8 @@ public final class FpIfTargetOlder extends FpEnvelope {
      * @return True if first path is older that second path
      * @throws IOException If fails to compare files
      */
-    private static boolean isAfter(final Path first, final Path second) throws IOException {
-        return Files.getLastModifiedTime(first).toInstant().isAfter(
+    private static boolean isOlder(final Path first, final Path second) throws IOException {
+        return Files.getLastModifiedTime(first).toInstant().isBefore(
             Files.getLastModifiedTime(second).toInstant()
         );
     }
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1381](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5727, opened 2026-07-21
#5704: add RFC 3986 compliant uri object to eo-runtime

### LINKED ISSUE #5704: Lack of RFC 3986 compliant `uri` object causes duplicate ad-hoc string splitting in eo-runtime
The eo-runtime standard library has no object for parsing a URI. There is a `path` object (`eo-runtime/src/main/eo/path.eo`) and a `socket` object, but nothing that takes a URI string and exposes its components. Today the only place the word `uri` appears is as a plain parameter name inside `path.eo:15`, not as a first-class object.

This is worth having because URIs (database connection strings, HTTP endpoints, resource locators) are common inputs, and without a dedicated object every program re-implements ad-hoc string splitting on `:`, `/`, `?` and `#`. A single, correct, RFC 3986 compliant parser in the standard library removes that duplication.

I propose adding a new read-only object `eo-runtime/src/main/eo/uri.eo` that wraps a string and exposes the RFC 3986 components as attributes: `scheme`, `authority` (with `userinfo`, `host`, `port`), `path`, `query` and `fragment`. It should be immutable — construction from a string, read-only access to the parts, no mutators. Simple usage:

```
uri "pgsql:localhost:899" > x
x.scheme > s   # pgsql
x.path > p     # localhost:899
```

(With an authority, e.g. `uri "pgsql://localhost:899"`, `x.host` would be `localhost` and `x.port` `899`, per the RFC's `"//" authority` rule.)

morphqdd:
Closes #5704.

Adds a new read-only object `eo-runtime/src/main/eo/uri.eo` that parses an RFC 3986 generic URI string and exposes its components as attributes: `scheme`, `authority` (with `userinfo`, `host`, `port` parsed out of it), `path`, `query` and `fragment`. It is immutable: construction from the input string only, no mutators.

Covers the generic-syntax grammar from RFC 3986: `scheme ":" hier-part [ "?" query ] [ "#" fragment ]`, where `hier-part` is either a `"//" authority path-abempty` or a bare path with no authority, and `authority` is `[ userinfo "@" ] host [ ":" port ]`. IPv6 literal hosts and percent-decoding are out of scope for this first version, matching the issue's own examples.

14 embedded tests cover: bare `scheme:path` form (`pgsql:localhost:899`), `//`-authority form with host and port (`pgsql://localhost:899`), userinfo, query and fragment, plus edge cases (empty host/path on a triple-slash `file://` URI, query/fragment without their leading delimiter).

Verified:
- `EOuriTest`: 14/14 green.
- Canonical format check (`mvn -pl eo-runtime process-sources`): clean, no reformatting needed.

### DIFF
--- a/eo-runtime/src/main/eo/uri.eo
+++ b/eo-runtime/src/main/eo/uri.eo
@@ -1,0 +1,191 @@
+# A `uri` object parses an RFC 3986 generic URI `text` and exposes its
+# components as read-only attributes: `scheme`, `authority` (together with
+# `userinfo`, `host` and `port` parsed out of it), `path`, `query` and
+# `fragment`. The object is immutable: it only reads parts of `text` and
+# has no mutators.
+#
+# Usage, without an authority:
+# ```
+# uri "pgsql:localhost:899" > x
+# x.scheme > s   # pgsql
+# x.path > p     # localhost:899
+# ```
+# With a "//" authority:
+# ```
+# uri "pgsql://localhost:899" > x
+# x.host > h    # localhost
+# x.port > p    # 899
+# ```
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
++unlint redundant-object
+
+[text] > uri
+  text > @
+  has-authority.if > after-slashes
+    hier-part.slice
+      2
+      hier-part.length.minus
+        number 2
+    ""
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (191 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1533](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5729, opened 2026-07-22
#5632: require the whole text to be a number in string.as-number

### LINKED ISSUE #5632: `string.as-number` incorrectly accepts and returns from strings with embedded numbers instead of triggering `cant-parse` fallback
## What happens

`string.as-number` (`eo-runtime/src/main/eo/string/as-number.eo`) silently returns the first number **embedded anywhere** in a non-numeric string instead of routing to its `cant-parse` fallback. Its own docstring states the contract: *"Returns the original `text` as `number`. If the text is not numeric, the `cant-parse` fallback is returned."* A string such as `"$100"` or `"12.34.56"` is not a number, yet a number is produced.

Root cause: `as-number` delegates parsing to `scanf`:

```
[text cant-parse] > as-number
  if. > @
    scanned.length.eq 0        # only checks that scanf found *something*
    cant-parse ...
    scanned.head
  (scanf "%f" text).at.^ > scanned
```

`scanf` matches its format **unanchored** — `EOscanf.lambda()` builds a plain regex and calls `matcher.find()` (`eo-runtime/src/main/java/org/eolang/EO_string/EOscanf.java:102-107`), which locates the `%f` pattern `([+-]?\d+(?:\.\d+)?)` at *any* position. So `as-number` accepts the text as soon as it contains a number-shaped substring; it never checks that the match spans the whole text. The only guard is `scanned.length.eq 0`, i.e. "was anything found at all".

## Failure scenarios

Verified end-to-end against the compiled runtime (`Phi.Φ.take("string").take("as-number")` with `text` bound and `cant-parse` unbound → `⊥`):

| text | `as-number` returns | expected |
| --- | --- | --- |
| `"5"` | `5.0` | `5.0` |
| `"3.14"` | `3.14` | `3.14` |
| `"Hello"` | `cant-parse` (⊥) | `cant-parse` |
| `"abc5"` | `5.0` | `cant-parse` |
| `"12.34.56"` | `12.34` | `cant-parse` |
| `"$100"` | `100.0` | `cant-parse` |
| `"1 2 3"` | `1.0` | `cant-parse` |

The existing tests (`as-number.eo`) only cover clean numeric inputs and the all-garbage `"Hello"` (which has no digits at all, so `find()` fails); the embedded-number case is never exercised, so the leniency is untested and appears unintended. This is the same class of "silent wrong result instead of a domain error" already fixed for `number.as-i64` in #5305.

## Suggested resolution

`as-number` must require the **whole** text to be the number, not merely to contain one. The un-anchoredness comes from `scanf`'s `matcher.find()`, so a fix lives at one of two boundaries:

1. **In `as-number`** — validate that the parsed token spans the entire text before trusting `scanned.head` (reject when the numeric match does not cover all of `text`), routing the mismatch to `cant-parse`.
2. **In `scanf` / a strict parse path** — expose an anchored match (or the match bounds) so `as-number` can confirm full consumption; note that `scanf` itself should stay un-anchored for its general substring-scanning use, so the anchoring belongs to `as-number`'s call, not to `scanf`'s default.

Either way, add tests that `"abc5"`, `"12.34.56"`, `"$100"` and `"1 2 3"` yield `cant-parse` rather than an embedded number.

gemshrine:
Fixes #5632.

`string.as-number` delegated parsing to `scanf`, whose `%f` pattern (`([+-]?\d+(?:\.\d+)?)`) is matched **unanchored** (via `find`), and the only guard was `scanned.length.eq 0` — i.e. "was anything found at all". Any text that merely *contained* a number-shaped substring was therefore accepted and a number returned, contradicting the object's own docstring contract:

| text | before | after |
| --- | --- | --- |
| `"abc5"` | `5.0` | `cant-parse` |
| `"12.34.56"` | `12.34` | `cant-parse` |
| `"$100"` | `100.0` | `cant-parse` |
| `"1 2 3"` | `1.0` | `cant-parse` |
| `"5"`, `"3.14"`, `"-42"` | parsed | parsed (unchanged) |

Per the issue's suggested resolution (option 1), the anchoring is applied at `as-number`'s call site, not in `scanf` (which must stay unanchored for its general substring-scanning use). The text is validated against the anchored equivalent of the same `%f` token, `^[+-]?\d+(?:\.\d+)?$`, so only a text that is entirely a number is parsed; otherwise the `cant-parse` fallback (or `⊥` when unbound) is returned.

Added tests covering `"abc5"`, `"12.34.56"`, `"$100"`, `"1 2 3"` (each must fall back) and a negative-signed number; verified they fail against the old implementation and pass with the fix. Full `eo-runtime` suite is green (the only failure is the pre-existing `EOsocketTest.refusesConnectionViaSyscall` Windows/TEST-NET artifact, which also fails on unmodified master). No new lint warnings.

### DIFF
--- a/eo-runtime/src/main/eo/string/as-number.eo
+++ b/eo-runtime/src/main/eo/string/as-number.eo
@@ -1,6 +1,11 @@
 # Returns the original `text` as `number`.
 # If the text is not numeric, the `cant-parse` fallback is returned,
 # or the bottom object when unbound.
+# The whole `text` must be the number: `scanf` matches its `%f` pattern
+# unanchored (via `find`), so on its own it would accept any text that
+# merely contains a number-shaped substring, e.g. `"$100"` or `"12.34.56"`.
+# The text is therefore validated against the anchored equivalent of that
+# same `%f` token, so that only a text that is entirely a number is parsed.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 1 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+403](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5731, opened 2026-07-22
fix(eo-runtime): return fallback when no switch case evaluates to true (#5726)

### LINKED ISSUE #5726: `switch.eo` incorrectly returns `true` instead of invoking `fallback` for non-empty, all-false `cases`
## What Happens

`eo-runtime/src/main/eo/switch.eo` takes `[cases fallback]`. Its header doc says "This object returns the value of the first true statement," but says nothing about what happens if no statement is true.

Tracing the definition and confirming by building `eo-runtime` from source and dataizing a real `switch` call: `fallback` is invoked only when `cases.length == 0`. When `cases` is non-empty but every condition in it evaluates to `false`, the outer dispatch falls through to its own else branch and returns the literal boolean `true`, never touching `fallback` at all.

Verified by building `Phi.Φ.take("switch")` in Java with two case-pairs `(false, "1")`, `(false, "2")` and `fallback` left deliberately unbound (mirroring the existing `tests-switch-with-all-false-cases` test, which never binds `fallback` either): dataizing the result gives boolean `true`, and it does not throw despite `fallback` being unset.

This means a caller who forgets to add a catch-all default case gets silently handed back the meaningless value `true`, which could easily be misread as "the switch value is boolean true" rather than "no case matched," instead of either invoking `fallback` or raising a clear error.

The existing generated tests for `switch` all pass under this behavior, because they only check `containsString`/boolean-truthiness, not the actual returned value, so none of them catch this gap.

## What Should Happen

When no case matches in a non-empty `cases` list, `switch` should invoke `fallback`, consistent with what its presence implies to a caller and with its only other current use (the empty-`cases` case). At minimum, if this is intentional, the doc comment should explicitly say that a non-empty, all-false `cases` list returns `true` and that `fallback` only guards against an empty list.

arimu1:
Closes #5726

### What was changed
In `eo-runtime/src/main/eo/switch.eo`, updated `switch` logic so that when no case evaluates to true in a non-empty `cases` array, it returns `fallback` instead of returning literal `true`.

Updated documentation and unit tests in `switch.eo`:
- Added test verifying `switch` returns `fallback` when all cases evaluate to `false`.
- Added test verifying `switch` without `fallback` fails when all cases evaluate to `false`.

### DIFF
--- a/eo-runtime/src/main/eo/switch.eo
+++ b/eo-runtime/src/main/eo/switch.eo
@@ -14,7 +14,9 @@
 #       "this value will be skipped"
 # ```
 #
-# This object returns the value of the first true statement.
+# This object returns the value of the first true statement,
+# or invokes `fallback` with a descriptive message when `cases`
+# is empty or when every case condition is false.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
@@ -46,7 +48,8 @@
                 rec-case tup.tail
             | cases
       found.head
-      true
+      fallback
+        "No switch case matched"
 
   switch * ++> can-return-true-when-all-cases-are-false
     *
@@ -94,6 +97,14 @@
       "password is correct!"
     "swordfish" > password
 
+  eq. ++> can-switch-on-all-false-cases-with-fallback
+    "fallback"
+    switch
+      *
+        * false "false1"
+        * false "false2"
+      "fallback" > [message]
+
   eq. ++> can-switch-on-the-first-of-several-true-cases
     switch *
       * true "TRUE1"
@@ -106,6 +117,14 @@
     switch
       *
       "recovered" > [message]
+
+  switch * --> stops-on-all-false-cases-without-fallback
+    *
+      false
+      "false1"
+    *
+      false
+      "false2"
 
   switch --> stops-on-an-empty-switch
     *
--- a/eo-runtime/src/main/eo/switch.eo
+++ b/eo-runtime/src/main/eo/switch.eo
@@ -51,14 +51,6 @@
       fallback
         "No switch case matched"
 
-  switch * ++> can-return-true-when-all-cases-are-false
-    *
-      false
-      "false1"
-    *
-      false
-      "false2"
-
   ++> can-switch-on-a-complex-case
     eq. > @
       switch *
--- a/eo-runtime/src/main/eo/switch.eo
+++ b/eo-runtime/src/main/eo/switch.eo
@@ -82,6 +82,14 @@
       * true "2"
     "2"
 
+  eq. ++> can-switch-on-all-false-cases-with-fallback
+    "fallback"
+    switch
+      *
+        * false "false1"
+        * false "false2"
+      "fallback" > [message]
+
   ++> can-switch-on-strings
     eq. > @
       switch *
@@ -96,14 +104,6 @@
           "password is wrong"
       "password is correct!"
     "swordfish" > password
-
-  eq. ++> can-switch-on-all-false-cases-with-fallback
-    "fallback"
-    switch
-      *
-        * false "false1"
-        * false "false2"
-      "fallback" > [message]
 
   eq. ++> can-switch-on-the-first-of-several-true-cases
     switch *


eo-runtime/src/main/eo/switch.eo
```diff
@@ -14,7 +14,8 @@
 #       "this value will be skipped"
 # ```
 #
-# This object returns the value of the first true statement.
+# This object returns the value of the first true statement,
+# or `fallback` if no condition is true.
```
morphqdd: Good, accurate doc update. Matches the new behavior exactly.

eo-runtime/src/main/eo/switch.eo
```diff
@@ -32,7 +33,7 @@
     if.
       true.eq found.match
       found.head
-      true
+      fallback
```
morphqdd, line 51: This is the actual fix for #5726, and it's correct: `true` is gone, `fallback` is returned instead. But note this returns `fallback` bare/unapplied, while the empty-cases branch just below (`cases.length.eq 0` -> `fallback "Switch cases are empty"`) explicitly calls `fallback` with a message. I tried to build a repro where a fallback that actually reads its `msg` parameter would observably differ between the two paths, using EO's `msg > [msg]` free-attribute idiom, but ran out of time to get a clean one (the semantics of that idiom made it hard to isolate). I'm not confident this is a live bug, but the asymmetry is real and worth a second look, or at least a comment explaining why `fallback` isn't invoked with a message here the way it is a few lines up.

eo-runtime/src/main/eo/switch.eo
```diff
@@ -90,13 +91,13 @@
...
     "swordfish" > password
 
-  switch * ++> tests-switch-with-all-false-cases
-    *
-      false
-      "false1"
-    *
-      false
-      "false2"
+  eq. ++> tests-switch-with-all-false-cases
```
morphqdd: Good, this is the right regression test for the actual reported bug (previously this called `switch` with two false cases and no fallback at all, and silently returned `true`).

eo-runtime/src/main/eo/switch.eo
```diff
@@ -90,13 +91,13 @@
...
+    "fallback"
+    switch
+      *
+        * false "false1"
+        * false "false2"
+      "fallback"
```
morphqdd: Worth noting: `fallback` here is a plain string, not a callable (unlike `tests-empty-switch-returns-provided-fallback` a few lines up, which passes `"recovered" > [msg]`). That's fine for this test's own purpose, but it means this test can't distinguish "fallback was properly resolved" from "fallback happened to be a non-callable value that dataizes to itself regardless of how it's touched" -- see my note on line 36.

eo-runtime/src/main/eo/switch.eo
```diff
@@ -105,5 +106,13 @@
       * true "TRUE2"
     "TRUE1"
 
-  switch --> on-empty-switch
+  switch * --> on-all-false-cases-without-fallback
```
morphqdd: Good new edge-case test: non-empty cases, all false, fallback never bound at all. Confirms the fix raises a real error here instead of silently returning something, which is exactly what #5726 asked for.

eo-runtime/src/main/eo/switch.eo
morphqdd: Separate minor point: neither this test nor `tests-empty-switch-returns-provided-fallback` (nor the two bottom-tests) ever assert what message string actually reaches `fallback`. `"Switch cases are empty"` in the empty-cases branch is never checked against anything, since both existing callable-fallback examples ignore their `msg` parameter. Not a blocker, just a coverage gap if `fallback`'s message content is ever meant to be part of the contract.

eo-runtime/src/main/eo/switch.eo
```diff
@@ -105,5 +106,13 @@
...
+    *
+      false
+      "false1"
+    *
+      false
+      "false2"
+
+  switch * --> on-empty-switch
```
morphqdd: Confirmed this pre-existing test still passes with the fix (empty cases, no fallback bound, correctly still raises).

eo-runtime/src/main/eo/switch.eo
```diff
@@ -32,7 +33,7 @@
     if.
       true.eq found.match
       found.head
-      true
+      fallback
   @. > found
     rec-case cases
```
morphqdd: Confirmed `found`/`rec-case` (the recursive matcher that finds the first true case) is untouched by this PR and still behaves correctly for the several-true-cases test -- the fix is isolated to what happens after `found` comes back empty, which is the right scope for this bug.

eo-runtime/src/main/eo/switch.eo
```diff
@@ -32,7 +33,7 @@
     if.
```
morphqdd, line 34: Confirmed the branch structure here is still canonically formatted (`mvn -pl eo-runtime install` with the format check enabled reports no issues on this file), so no `-Deo.autoFix` follow-up needed for this change specifically.

morphqdd (commented):
Verified by checking out the branch and running `mvn -pl eo-runtime test -Dtest='org.eolang.EOswitchTest'` (8/8 green). Core fix is correct and well-scoped. Left inline notes on a few points, the main one being the asymmetry between how the two error paths treat `fallback` (see the comment on line 36).

eo-runtime/src/main/eo/switch.eo
gemshrine, line 51: This branch now returns a **bare** `fallback`, but the empty-`cases` branch just above (lines 31-32) invokes it **with an argument**: `fallback` / `"Switch cases are empty"`. So the two "no match" paths now demand two incompatible shapes of `fallback`:

- empty `cases` → `fallback` must be a formation with one void attribute to receive the message — that is exactly why the existing `tests-empty-switch-returns-provided-fallback` writes `"recovered" > [msg]`;
- all-`false` `cases` → `fallback` is applied to nothing, so a plain object with no void attribute is expected — that is what your new `tests-switch-with-all-false-cases` passes: `"fallback"`.

A single value cannot satisfy both. Concrete break: `switch * "fallback"` with an **empty** `cases` tuple dataizes `"fallback" "Switch cases are empty"` — over-application of a data object that has no void attribute → it throws, even though `"fallback"` is a valid fallback for the all-false path you just added.

It also diverges from the runtime-wide fallback convention, where the fallback is always handed a descriptive message: `nsplit` (`"Invalid limit %d…".printf`), `i8.div` (`"Can't divide %d…".printf`), `map.found.get` (`"Object by hash code %d…".printf`). Please keep it consistent and pass a message here too:

```
      true.eq found.match
      found.head
      fallback
        "No switch case matched"
```

Then `fallback` has one uniform contract (a `[msg]` formation) across both no-match branches. The new test then needs the matching shape — see my comment on line 100.

eo-runtime/src/main/eo/switch.eo
gemshrine: Passing a bare `"fallback"` (zero void attributes) here, while `tests-empty-switch-returns-provided-fallback` uses `"recovered" > [msg]` (one void attribute), is what masks the arity mismatch flagged on line 36: the two no-match tests exercise two incompatible fallback shapes, so neither test would notice that a fallback which works for one path breaks on the other.

Once the all-false branch also passes a message, make this fallback a `[msg]` formation, identical in shape to the empty-case test:

```
  eq. ++> tests-switch-with-all-false-cases
    "recovered"
    switch
      *
        * false "false1"
        * false "false2"
      "recovered" > [msg]
```

That proves a single `[msg]`-shaped fallback works for **both** no-match paths. To also lock down that the message actually reaches the fallback (the issue's core complaint is that existing tests never checked the returned value), you can go one step further and echo it, e.g. fallback `msg > [msg]` with the expected value set to `"No switch case matched"`.

eo-runtime/src/main/eo/switch.eo
gemshrine: Good to document this. One gap: `fallback` is also what gets returned when `cases` is **empty** (the branch on lines 30-32), and in that path it is invoked **with a message** argument. Consider stating the fallback contract explicitly, e.g.:

> If no condition is true (or `cases` is empty), `fallback` is dataized with a descriptive message, so it must be a one-argument formation such as `msg > [msg]`.

so callers know to write a `[msg]`-shaped fallback rather than a bare value that would throw when applied to the message.

gemshrine (changes requested):
Thanks for tackling #5726 — routing the non-empty all-`false` case to `fallback` is the right direction, and the new `-->` test for the unbound-fallback case is a nice touch.

One thing blocks it from being consistent: the all-false branch returns a **bare** `fallback`, while the empty-`cases` branch (and every other fallback in the runtime — `nsplit`, `i8.div`, `map.found.get`) invokes it **with a descriptive message**. As written, the two no-match paths require two incompatible `fallback` shapes, so a fallback that works for one path throws on the other. Details and a suggested fix are inline, plus smaller notes on the new test and the doc comment.

arimu1:
Addressed the review feedback: the all-false path now invokes \`fallback\` with a message (same arity as the empty-\`cases\` branch):

\`\`\`
fallback
  "No switch case matched"
\`\`\`

Updated \`tests-switch-with-all-false-cases\` to pass \`"fallback" > [msg]\` so both no-match paths share one fallback shape. Doc comment updated accordingly.

Local: \`EOswitchTest\` 8/8 green.

arimu1:
Note on the red `mvn (ubuntu-24.04, 26)` job: it is unrelated to this change.

The failure is `MjParseTest.parsesConcurrentlyWithLotsOfPrograms` in **eo-maven-plugin**, which fails while parsing the synthetic `foo.x.main-8` program that `FakeMaven` generates in-test — a race in the concurrent parse path (`Only 'o' XML elements are accepted here` from `set-locators.xsl`). This PR touches exactly one file, `eo-runtime/src/main/eo/switch.eo`, which that test never loads. Every other check on the PR is green, and the other three `mvn` matrix jobs were cancelled by fail-fast rather than failing on their own.

Happy to rebase if you'd like a fresh run.

eo-runtime/src/main/eo/switch.eo
```diff
@@ -32,7 +34,8 @@
...
       true.eq found.match
       found.head
-      true
+      fallback
+        "No switch case matched"
```
gemshrine, line 52: Fix verified — this closes the finding from my earlier review. Both no-match paths now invoke `fallback` with exactly one string argument (this line, and `"Switch cases are empty"` on the empty-`cases` branch), so a user-supplied fallback of the shape `"…" > [msg]` binds with consistent arity on either path — no more `applying decorated 0-arg vs 1-arg` mismatch. The updated `tests-switch-with-all-false-cases` (asserting `eq. "fallback"` against a `[msg]` fallback) pins exactly that path, and the new `on-all-false-cases-without-fallback` case keeps the bare-tuple behavior covered.

The red `mvn` jobs across all four platforms are the known-flaky `MjParseTest.parsesConcurrentlyWithLotsOfPrograms` (`Only 'o' XML elements are accepted here`, `MjParseTest.java:258`) in eo-maven-plugin — a concurrent-parse race unrelated to this PR, which touches only `switch.eo`. Not a blocker for this change. Clearing my earlier change-request.

gemshrine (approved):
Approving — the fallback-arity concern from my earlier review is addressed and verified. See the inline note. This supersedes my previous change-request.

yegor256:
@arimu1 merge conflicts here, please resolve

arimu1:
Rebased onto latest `master` and force-pushed — merge conflicts resolved. Branch is now `MERGEABLE` against current master.

arimu1:
@yegor256 Merge conflicts are resolved — this branch already contains current `master` and GitHub reports **mergeable**.

@gemshrine On the fallback arity point: the all-false path now invokes `fallback` with `"No switch case matched"` (same pattern as the empty-`cases` branch’s `"Switch cases are empty"`), and the all-false regression test uses a `[message]` formation like the empty-switch test. Happy to adjust further if anything still looks off.

eo-runtime/src/main/eo/switch.eo
```diff
@@ -46,15 +48,8 @@
...
-
-  switch * ++> can-return-true-when-all-cases-are-false
-    *
-      false
-      "false1"
-    *
-      false
-      "false2"
+      fallback
+        "No switch case matched"
```
morphqdd, line 52: Two distinct fallback messages exist ("Switch cases are empty" vs "No switch case matched"), but every test that exercises a fallback path binds it via [message] and asserts a caller-supplied constant, never the actual message text switch passes in. A copy-paste swap of the two messages between branches would not be caught by any existing test.

morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +8 points for this: +12 as a basis; -4 for too few (31) hits-of-code.  Your running score is [+2836](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@arimu1 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (31) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+56](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5736, opened 2026-07-22
#5732: preserve compact `|` pipe when target formation floats away

### LINKED ISSUE #5732: Loss of compact `|` pipe during `eo-printer` formatting when target formation becomes a formation attribute
After #5684 the printer emits the `|` pipe-application form, but only when the pipe target is the *immediately-preceding sibling* — the guard at `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:223`. When the target formation floats up to become a formation attribute while its pipe application stays nested inside another expression (the #5526 "floated away" case noted at `to-eo-tree.xsl:220`), the guard fails and the line prints as an ordinary application. Concretely, in `eo-runtime/src/main/eo/tuple/front.eo` the recursive helper written inline as `if. > [t] > r … | list` is rewritten by `eo:format` into a hoisted `[t] > r` attribute plus a bare `r list`; both spellings parse to identical XMIR, the only difference being the `pipe=""` marker on the application node.

This is the same class of loss #5684 set out to fix: `|` exists to keep recursion compact, yet for a helper used deep inside an expression the compact author spelling is still write-only, destroyed on the first format. `merge-monikers.xsl` already un-hoists ordinary `> name` bindings back onto their use-site, but it deliberately skips exactly this case — it excludes pipe-floated bindings (`not(exists($attr/@pipe))`, `merge-monikers.xsl:53`) and references that carry arguments (`not($ref/o)`, `merge-monikers.xsl:44`), which is precisely what a `| args` use is.

The fix is a small targeted pass, mirroring `merge-monikers.xsl`, that relocates a pipe-floated binding back down to nest directly above its single pipe use-site (re-nesting `[t] > r` in front of the `| list`), so the existing adjacency guard at `to-eo-tree.xsl:223` fires and emits the `|`. Gate it on the binding having exactly one pipe reference — the same single-host condition `merge-monikers` uses (#5706) to keep the round trip converging. Note a `Penalty` weight tweak cannot achieve this: the pipe/plain choice is structural, made in the XSL before `Pretty` runs, so there is no candidate for a weight to select.

yegor256:
Closes #5732.

## Problem

After #5684 the printer emits the `|` pipe-application form, but the adjacency guard at `to-eo-tree.xsl:223` only fires when the pipe target is the *immediately-preceding sibling*. When a pipe target formation floats up to become a formation attribute (#5526) while its pipe application stays nested inside another expression (the "floated away" case), the guard fails and the line prints as an ordinary application — destroying the compact `| args` author spelling on the first `eo:format`. This is exactly the recursive-helper case in `eo-runtime/src/main/eo/tuple/front.eo`.

`merge-monikers.xsl` already un-hoists ordinary `> name` bindings back onto their use-site, but it deliberately skips pipe-floated bindings (`not(exists($attr/@pipe))`) and references carrying arguments (`not($ref/o)`) — which is precisely what a `| args` use is.

## Fix

Add `renest-pipe-monikers.xsl`, a small targeted pass mirroring `merge-monikers.xsl`, registered right after it in the `Xmir` train (before `to-eo-tree.xsl`). For a formation attribute whose only reference is a single `@pipe` application that no longer sits directly after it, the pass copies the formation back in front of that pipe use-site and drops the floated original. The pipe then reads its predecessor from the immediately preceding sibling, so the existing adjacency guard fires and the `|` head is emitted. Both spellings parse to identical XMIR, so re-nesting is safe.

It is gated on the formation having exactly one reference which must be the pipe use — the same single-host condition `merge-monikers` uses (#5706) to keep the round trip converging. A second reference would fall out of scope once the formation moved into the pipe's argument block. A formation already adjacent to its pipe use is left untouched.

## Changes

- **`renest-pipe-monikers.xsl`** — the new pass.
- **`Xmir.java`** — register the pass in the transformation train after `merge-monikers`.
- **`to-eo-tree.xsl`** — update the pipe-application comment to note the floated case is now re-nested upstream.
- **`pipe-floated-recursive-helper.yaml`** — new print-pack modeling the `tuple/front.eo` recursive helper; the `|` now survives the round trip.
- **`pipe-in-argument-block.yaml`** — updated expected output: the pipe is now preserved instead of degrading to a bare application.

## Testing

`XmirTest` (all 139 parameterized print-pack / eo-pack cases) passes, including the round-trip check that printed EO reprints to the very same EO.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TipwHhMMamjeKKum4ggWLw)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Xmir.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Xmir.java
@@ -50,6 +50,7 @@ public final class Xmir implements XML {
                 new StClasspath("/org/eolang/printer/print/dataized-to-const.xsl"),
                 new StClasspath("/org/eolang/printer/print/unnecessary-as.xsl"),
                 new StClasspath("/org/eolang/printer/print/merge-monikers.xsl"),
+                new StClasspath("/org/eolang/printer/print/renest-pipe-monikers.xsl"),
                 new StClasspath("/org/eolang/printer/print/to-eo-tree.xsl")
             )
         )
--- a/eo-printer/src/main/resources/org/eolang/printer/print/renest-pipe-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/renest-pipe-monikers.xsl
@@ -1,0 +1,112 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" xmlns:xs="http://www.w3.org/2001/XMLSchema" exclude-result-prefixes="xs" id="renest-pipe-monikers" version="2.0">
+  <!--
+  Re-nests a pipe target formation back above its pipe use-site, restoring
+  the compact `| args` continuation line (§3.14). A `| args` line applies
+  its same-indent predecessor; the parser rewrites it into an application
+  whose `@base` points at that predecessor's name and keeps a `@pipe`
+  marker (#5684). When the predecessor formation sits inside an argument
+  block (its parent is itself an application), `wrap-applications` /
+  `vars-float-up` hoist that formation up to the enclosing abstract object
+  and drop it from the argument slot (#5526), leaving the pipe application
+  as the block's sole remaining member. The predecessor has then floated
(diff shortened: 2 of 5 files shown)


morphqdd:
Built this at head (62e1020) and ran the print-pack tests plus an extra probe (a second bare reference alongside the pipe use) — the `eo:pipe-floated` gate and the new `renest-pipe-monikers.xsl` logic hold up, no bug found there.

Two CI checks are still red on this SHA: `mvn (macos-15, 26)` and Codacy ("1 new issue"). Worth reading those before taking this out of draft, since Codacy in particular could be flagging something real in the new XSL rather than noise.

PR #5743, opened 2026-07-22
#5467: rewrite i64.div atom in pure EO via binary long division

### LINKED ISSUE #5467: `i64.div` atom implementation lacks precision and performance

morphqdd:
Closes #5467.

Rewrites the `div` object in `eo-runtime/src/main/eo/i64.eo` from a native Java atom to pure EO, using binary long division (shift-and-subtract) over the 64 bits, as suggested in the issue.

Approach:
- `magnitude` takes the two's-complement negation of a negative operand so the loop always works on a non-negative bit pattern, even for `i64.min` (whose magnitude is not representable as a positive `i64`, so it is handled implicitly by the wraparound arithmetic rather than an explicit special case).
- `negative` is the XOR of the operand signs, applied to the quotient at the end.
- `looped` is a recursive local that walks the dividend bit by bit from position 63 down to 0, shifting the remainder left and bringing in the next dividend bit, comparing against the divisor magnitude with an unsigned comparison, subtracting when it fits and setting the corresponding quotient bit.
- `cant-divide` is kept as an overridable free attribute (`[x cant-divide] > div`) so the existing divide-by-zero-fallback test keeps working.

Deletes the native `EOi64$EOdiv.java` atom and the Java-side test that only existed to test it; the equivalent coverage now lives as `++>` tests inside `i64.eo`, including the two edge cases called out in the issue: `i64.min / -1` wraps back to `i64.min`, and `i64.min / 1 == i64.min`.

Important follow-up found and fixed after the first push: once `i64.plus`/`i64.gt`/`i64.gte` (issues #5469/#5470) landed as pure-EO recursive objects on master, dividing anything through this loop OOM'd under the default 4GB test heap. Root cause was twofold, both fixed:
1. The loop's own accumulators (`shifted`, the next remainder, the next quotient) were referenced unforced from multiple places, so each of the 64 iterations doubled the retained lazy expression graph instead of collapsing to a concrete value. Fixed by forcing them with the `!` const-dataization idiom already used elsewhere in this file.
2. The loop's unsigned comparison called into `i64.gt`/`gte`, which negate via a full `times -1` (itself a 64-step multiplication calling `plus` each step), turning one comparison into thousands of recursive steps per division bit. Replaced it with a direct carry-out check on the subtraction the loop already computes for the next remainder, using only `bytes` and/or/xor (the same style `times`'s own fast loop already uses), so a comparison is a handful of cheap bitwise ops instead of a nested recursive multiplication.

Also filed #5772 to document the general risk (composing two recursive pure-EO arithmetic atoms without forcing loop accumulators can blow up memory, not just be "a bit slower") for whoever picks up the remaining i16/i32/u8/u16/u32/u64 conversions in this series.

Verified:
- `EOi64Test`: 51/51 green, ~7 seconds total (was OOM before the fix, ~100s per division even with an eager-forcing-only fix, under the default `-Xmx4096m` test heap).


yegor256:
@morphqdd one merge conflict here

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (216 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1577](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5744, opened 2026-07-22
#5471: rewrite i64.as-number atom in pure EO via IEEE-754 bit assembly

### LINKED ISSUE #5471: `i64.as-number` could be implemented in EO, replicating `number.as-i64` approach

morphqdd:
Closes #5471.

Rewrites the `as-number` object in `eo-runtime/src/main/eo/i64.eo` from a native Java atom to pure EO, mirroring the existing `number.as-i64` bit-surgery approach in the opposite direction, as suggested in the issue.

Approach:
- `sign` and `magnitude` (two's-complement negated if negative) are extracted first.
- `highest` finds the position of the highest set bit in the magnitude via a recursive `top-bit` local; `highest-bits` builds the corresponding bit pattern via `top-bit-bytes` in parallel, avoiding a floating-point log2.
- `exponent-bits` adds the IEEE-754 bias of 1023 to the unbiased exponent, with a `carried` flag to bump the exponent by one when mantissa rounding overflows into it.
- `mantissa` positions the magnitude's remaining bits after dropping the implicit leading one; for magnitudes at or above 2^53, it rounds to nearest, ties to even, using guard/sticky/odd bits (`guard`, `sticky`, `odd`, `round-up`, `rounded`).
- The zero case short-circuits before any of this.
- Final result is `sign | exponent-bits | mantissa` assembled into 8 bytes and wrapped as `number`.

Deletes the native `EOi64$EOas_number.java` atom and its dedicated Java-side test class (`EOiConvertersExpectTest`, which only existed to test this one atom's error path); the equivalent and expanded coverage now lives as `++>` tests inside `i64.eo`: small positive/negative, exact power of two, `i64.min`, `i64.max` (round-up), round-tie-to-even, and round-up-above-tie.

This was previously attempted but blocked by a real runtime defect in `bool.eo` (duplicate content left over from a manual conflict resolution, tracked as #5478 and fixed in #5497, already merged). With that fix in master, this rewrite now builds and passes cleanly.

Update: after landing, master's later changes (#5469/#5470, converting `i64.plus`/`i64.gt` from native atoms to recursive pure-EO objects) turned `as-number`'s `i64.min` case into a lazy-thunk memory blowup (OOMs at default heap). Several of `as-number`'s named locals (`highest-bits`, `kept`, `magnitude`, `rounded`, `sign`) are each referenced from more than one place without being forced eager, so each reference re-walked the now-expensive recursive expression tree from scratch. Fixed by forcing those locals with `!`, matching the same pattern already used for `top-bit-bytes` and for `i64.div` (issue #5467, PR #5743).

Verified:
- `EOi64Test`: 58/58 green (including `i64.min`, which now completes in well under a second instead of hanging/OOMing).


yegor256:
@morphqdd merge conflict here, please fix

yegor256:
@morphqdd ready to merge or it's still a draft?

morphqdd:
@yegor256, ready now, took it out of draft.

Tests for the bug itself pass fine. `EOi64Test` is green, 48 of 48, including the two regression tests for the OOM case this PR fixes, `tests-i64-as-number-of-i64-min` and `tests-i64-as-number-of-i64-max-rounds-up`.

CI has been red on several recent runs, but for reasons unrelated to this change, a couple of examples:

- One Ubuntu run failed with `TimeoutException: timed out after 5 minutes` across ten unrelated modules in the same run (`EOrandomTest`, `EOsetTest`, `EOarc_sineTest`, `EOhashTest`, `EOsqrtTest`, `EOas_fixedTest`, `EOpowerTest`, `EOas_hexTest`, `EOcosineTest`, `EOnumberEOAtomTest`), none of them touching i64.
- Just now, the Windows leg failed on `SavedTest.exposesNoPartiallyWrittenFileToConcurrentReader` with `java.nio.file.AccessDeniedException` while renaming the sibling temp file to the target (`shared.txt....tmp -> shared.txt`). This is after #5873 already landed the atomic rename fix for that test, so it looks like the rename can still lose a race against a concurrent reader on Windows and hit `AccessDeniedException` instead of the earlier `FileSystemException`.

Is it worth filing an issue for either of these. Happy to write one up with the log details if so.

morphqdd:
@yegor256, please review, all fixed

eo-runtime/src/main/eo/i64.eo
```diff
@@ -23,7 +21,84 @@
           * as-number
     i32 > left
       as-bytes.slice 4 4
-  [] > as-number /Q.number
+  [] > as-number
+    if. > @
+      as-bytes.eq 00-00-00-00-00-00-00-00
```
Favixx, line 29: This zero short-circuit is the only branch in the new `as-number` without a `++>` test, and it is load-bearing: it is what keeps `scan` from being asked for the highest set bit of a magnitude that has none. `scan` would return `acc` unchanged for zero, so `highest` would come out as 0 and the result would be `2^-1023`-ish garbage rather than zero — the guard is the only thing preventing that.

`0.as-i64.as-number.eq 0 ++> tests-i64-as-number-of-zero` is one line and pins the branch.

eo-runtime/src/main/eo/i64.eo
```diff
@@ -23,7 +21,84 @@
...
+      0
+      number
+        sign.or
+          exponent-bits.or mantissa
+    rounded.eq 00-20-00-00-00-00-00-00 > carried
+    carried.if > exponent-bits
```
Favixx: `exponent-bits` always dereferences `carried`, and `carried` pulls in `rounded` → `kept`, `guard`, `sticky`, `odd`. So the whole rounding apparatus is evaluated even for magnitudes below `2^53`, where it is meaningless: on line 51 `kept` becomes `magnitude.right (highest.minus 52)` with a **negative** count, which `BytesRaw.shift` interprets as a left shift, so `kept` is not "the kept bits" at all in that path.

It happens to be harmless — a left-shifted magnitude lands in `[2^52, 2^53)` and its guard bit is zero, so `rounded` can never reach the `2^53` that `carried` tests for — but that is a coincidence of the shift semantics, not an invariant anyone stated. One `if.` on `highest.gt 52` around the whole rounding path would make it structural, and would stop the exact case from doing five pointless byte shifts:

```
if. > exponent-bits
  highest.gt 52
  carried.if ...
  <the unbiased form>
```

eo-runtime/src/main/eo/i64.eo
```diff
@@ -23,7 +21,84 @@
...
+    scan > scanned!
+      magnitude
+      43-30-00-00-00-00-00-00
+      32
+      00-00-00-00-00-00-00-20
+    as-bytes.and 80-00-00-00-00-00-00-00 > sign!
+    not. > sticky
+      eq.
+        magnitude.left
+          117.minus highest
```
Favixx, line 103: `117.minus highest` is the subtlest line here and it is the one with no explanation. It is `64 - (highest - 53)`: shift the magnitude left far enough that only the bits strictly below the guard bit survive, so a non-zero result means "sticky". Worth saying in a comment, because the reader's first instinct is that 117 is a typo.

More importantly, at `highest = 53` this shift is exactly **64**, and the code is correct only because `bytes.left` (via `BytesRaw.shift`) is byte-wise and zero-fills once the offset runs past the array. Java's own `<<` masks the count to six bits and would return the operand unchanged, which would make `sticky` true for every tie and break `tests-i64-as-number-rounds-tie-to-even`. That is a load-bearing property of another module, and `bytes/left.eo`'s own tests only cover shifts of 0–3 — nothing there pins the ≥ 64 behaviour. A test in `bytes/left.eo` for a 64-bit shift would protect this.

eo-runtime/src/main/eo/i64.eo
```diff
@@ -23,7 +21,84 @@
...
+        plus.
+          i64 3F-F0-00-00-00-00-00-00
+          i64 highest-bits
+    eq. > guard
+      and.
+        magnitude.right
+          highest.minus 53
+        00-00-00-00-00-00-00-01
+      00-00-00-00-00-00-00-01
+    scanned.as-number.minus 4503599627370496 > highest
```
Favixx, line 50: `highest` is referenced six times (lines 43, 52, 61, 63, 100, and via `mantissa`) and is the one local in this object that is *not* forced with `!`, while `highest-bits`, `kept`, `magnitude`, `rounded`, `scanned` and `sign` all are. The description states the rule you applied — "locals referenced from more than one place without being forced eager … re-walked the expression tree from scratch" — and this one meets it. `scanned!` is memoised so the recomputation is only `as-number` plus a subtraction, but the inconsistency will read as an oversight to the next person, and it is the same class of bug that caused the OOM.

eo-runtime/src/main/eo/i64.eo
```diff
@@ -23,7 +21,84 @@
...
+    left. > highest-bits!
```
Favixx: `highest-bits` re-derives the bit position from `scanned` (`scanned.and 00-…-3F`) instead of using `highest`, which is the same quantity two lines below. Two spellings of one value, coupled to the same magic constant: if the `43-30-…` base ever changes, one of them will be updated and the other will not. Since `highest` is a number and this needs bytes, `highest.as-i64.as-bytes.left 52` — or keeping this form and deriving `highest` from it — would leave a single source.

eo-runtime/src/main/eo/i64.eo
```diff
@@ -23,7 +21,84 @@
...
+        scan
+          empty.if m shifted
+          empty.if
+            acc
+            acc.or bit
+          step.div 2
+          bit.right 1
+      shifted.eq 00-00-00-00-00-00-00-00 > empty
+      m.right step > shifted!
+    scan > scanned!
```
Favixx, line 94: `43-30-00-00-00-00-00-00` and `4503599627370496` on line 47 are the same number (`2^52`) written two different ways, and neither is explained. The trick is standard but not obvious: `2^52` as a double has an all-zero mantissa, so OR-ing a value below `2^52` into the bit pattern and reinterpreting gives `2^52 + k` exactly, and subtracting the constant recovers `k` without touching `log2`. Two lines of comment above `scanned` would save the next reader the reverse-engineering — the description explains this well, and none of it made it into the file.

eo-runtime/src/main/eo/i64.eo
```diff
@@ -94,6 +169,34 @@
...
+  eq. ++> tests-i64-as-number-of-i64-min
+    as-number.
+      i64 80-00-00-00-00-00-00-00
+    -9.223372036854776e18
+
+  -3.as-i64.as-number.eq -3 ++> tests-i64-as-number-of-small-negative
+
+  3.as-i64.as-number.eq 3 ++> tests-i64-as-number-of-small-positive
+
+  eq. ++> tests-i64-as-number-rounds-tie-to-even
```
Favixx: The rounding tests all start at `2^53 + 1` and go up. Nothing covers the seam itself: `2^53` exactly, and `2^53 - 1`, which is the largest magnitude still taking the `highest.lte 52` branch on line 61. Those two are where an off-by-one in `highest.minus 52` / `52.minus highest` would show up, and they are exactly the values the current tests step over.

Favixx (commented):
I traced the algorithm end to end and I believe it is correct, including the cases that usually break this kind of code:

* `i64.min` — the two's-complement negation overflows back to `80-00-…-00`, `scan` still reports `highest = 63`, and the sign bit is taken from the original bytes rather than the magnitude, so the result is right.
* `i64.max` — `kept` overflows to `2^53` on round-up, `carried` bumps the exponent to 1086 and the mantissa to zero, giving exactly `2^63`.
* The `add-magic-constant` trick for `log2` (`43-30-…` = `2^52` as a double, OR the low six bits, reinterpret, subtract `2^52`) is sound because `2^52`'s mantissa is all zeroes.

I also checked two things the description asserts, and both hold: `i64.eo` has no atoms left, so dropping `+rt jvm`/`+rt node` matches how the other atom-free files (`bytes/left.eo`, for instance) are written; and `EOiConvertersExpectTest`'s `converters()` stream had exactly one entry, so deleting the class costs no coverage of the other converters.

So this is `COMMENT`, not a change request. What I would still like to see before it merges is the zero branch under test and the rounding path gated on `highest` — details inline, roughly in the order I care about them.

One process note: the description's "Update: after landing, master's later changes … turned `as-number`'s `i64.min` case into a lazy-thunk memory blowup" reads as if this PR had already been merged. It made me look for a revert before I realised it is describing an earlier attempt. Worth rewording, since that paragraph is the one explaining why half the locals carry `!`.

eo-runtime/src/main/eo/i64.eo
```diff
@@ -94,6 +169,34 @@
...
+    as-number.
+      i64 00-20-00-00-00-00-00-01
+    9007199254740992
+
+  eq. ++> tests-i64-as-number-rounds-up-above-tie
```
Favixx: Every rounding test is positive; the negatives are `-3` (exact) and `i64.min` (a magnitude whose negation overflows, so it never exercises ordinary two's-complement). A negative that actually rounds — `-9007199254740993`, the negation of the tie case just above — would be the only test covering the interaction between the negated magnitude on line 53 and the round-to-even path, which is the combination most likely to be wrong.

eo-runtime/src/test/java/org/eolang/EOiConvertersExpectTest.java
```diff
@@ -1,54 +0,0 @@
...
- * Test case verifying {@link Expect}-based error messages
- * raised by the {@code EOi16/i32/i64} converter atoms.
- * @since 0.51
- * @checkstyle TypeNameCheck (4 lines)
- */
-@SuppressWarnings("JTCOP.RuleAllTestsHaveProductionClass")
-final class EOiConvertersExpectTest {
-
-    @ParameterizedTest
-    @MethodSource("converters")
```
Favixx, line 30: This asserted the error message `the 'ρ' attribute must be an i64` when `as-number` is applied to a non-i64 ρ, which came from the `Expect.I64` in the deleted atom. The pure-EO version has no such check, so that ρ now flows into `as-bytes` and fails (if it fails at all) with whatever message that produces.

Deleting the class is fine — its `converters()` stream had a single entry, so nothing else loses coverage. But the description's claim that the `++>` tests are "equivalent and expanded" is not right for this one: none of the seven new tests applies `as-number` to a wrong-typed ρ. Either keep a test for the new failure mode, or say plainly in the description that the typed error message is gone.

eo-runtime/src/main/eo/i64.eo
```diff
@@ -23,7 +21,84 @@
...
+    [m acc step bit] > scan
+      if. > @
+        step.lt 1
+        acc
+        scan
+          empty.if m shifted
+          empty.if
+            acc
+            acc.or bit
+          step.div 2
```
Favixx, line 90: `step.div 2` walks 32, 16, 8, 4, 2, 1, **0.5**, and the recursion stops on `step.lt 1` rather than on zero — so termination depends on EO numbers being floating point. It works, but it means `step` and `bit` (line 88, halving as bytes via `right 1`) are two counters for the same thing in two different representations. Deriving the step from `bit` alone, or comparing `bit.eq 00-…-00`, would make the loop obviously finite and drop a parameter from a four-parameter local object.

eo-runtime/src/main/eo/i64.eo
```diff
@@ -23,7 +21,84 @@
...
+      00-00-00-00-00-00-00-01
+    guard.and > round-up
+      sticky.or odd
+    round-up.if > rounded!
+      as-bytes.
+        plus.
+          i64 kept
+          i64 00-00-00-00-00-00-00-01
+      kept
+    [m acc step bit] > scan
```
Favixx, line 81: `scan` is a general "position of the highest set bit" binary search that happens to live inside `as-number`. `number.as-i64` does the mirror-image bit surgery and would want the same primitive, and this object is already dense enough that a twelve-line helper in the middle of it hurts. Given #5501/#5486 have been moving attributes into package directories, `i64/` looks like the right home — as its own object it would also become directly testable, which right now it is not.

eo-runtime/src/main/eo/i64.eo
```diff
@@ -23,7 +21,84 @@
           * as-number
     i32 > left
       as-bytes.slice 4 4
-  [] > as-number /Q.number
+  [] > as-number
```
Favixx, line 27: Ninety lines of IEEE-754 assembly with no comment on the object itself. The description explains the shape well — sign and magnitude, `highest` via the magic constant, bias 1023, mantissa positioning, round-to-nearest-ties-to-even above `2^53` — and none of that is in the file, where the next person will actually be reading it. A short block comment naming the three constants (`2^52` base, bias `1023` as `3F-F0-…`, `4000-…` for the carried exponent) and the guard/sticky/odd convention would do it.

morphqdd:
@yegor256, please review

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+58](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (192 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2744](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5745, opened 2026-07-22
Remove extra parenthesis in VerboseBytesAsString output

### LINKED ISSUE #5725: `VerboseBytesAsString` generates unbalanced closing parenthesis for 8-byte values
## What Happens

`eo-runtime/src/main/java/org/eolang/VerboseBytesAsString.java`, lines 51-58, the `case 8` branch:

```java
case 8:
    result = String.format(
        "[0x%s] = %s, or \"%s\")",
        this.toHex(),
        new BytesOf(this.data).asNumber(),
        this.escaped()
    );
```

The format string ends in `")`, a closing parenthesis with no matching opening one anywhere in the literal.

This is reached by `org.eolang.Main` for the "Finished in Xs (Y bytes)" summary that gets printed after running any EO program from the CLI, whenever the program's dataized result is 8 bytes long, which is essentially every time a program returns a `number`.

Verified directly: for the double `12.34567`, the output is:
```
[0x4028B0FB-A8826AA9-] = 12.34567, or "@(����j�")]
```
Note the stray `)` right before the final `]`.

The existing unit test (`VerboseBytesAsStringTest`) only asserts `containsString`, not full equality, so it never catches this.

## What Should Happen

Remove the stray `)` from the format string (or add the matching `(` if one was originally intended around the "or ..." clause).

gp07-cmd:
## Summary

Closes #5725 
This PR fixes a formatting issue in `VerboseBytesAsString` for 8-byte values.

The format string in the `case 8` branch contained an unmatched closing parenthesis:

```java
"[0x%s] = %s, or \"%s\")"
```

As a result, CLI output for numeric results included an extra `)` character. For example:

```text
[0x4028B0FB-A8826AA9-] = 12.34567, or "@(����j�")]
```

## Fix

Removed the stray closing parenthesis from the format string:

```java
"[0x%s] = %s, or \"%s\""
```

This produces correctly formatted output without the unmatched `)`.

## Example

Before:

```text
[0x4028B0FB-A8826AA9-] = 12.34567, or "@(����j�")]
```

After:

```text
[0x4028B0FB-A8826AA9-] = 12.34567, or "@(����j�"]
```

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/VerboseBytesAsString.java
+++ b/eo-runtime/src/main/java/org/eolang/VerboseBytesAsString.java
@@ -50,7 +50,7 @@ public final class VerboseBytesAsString implements Supplier<String> {
                 break;
             case 8:
                 result = String.format(
-                    "[0x%s] = %s, or \"%s\")",
+                    "[0x%s] = %s, or \"%s\"",
                     this.toHex(),
                     new BytesOf(this.data).asNumber(),
                     this.escaped()


yegor256:
@gp07-cmd  Thanks for your contribution! Unfortunately, we can't merge it without tests. Every feature we add and every bug we fix must be accompanied by at least one new test that confirms that the code was not working before. Without such a test, how can we know that what you commit actually works?

gp07-cmd:
> @gp07-cmd  Thanks for your contribution! Unfortunately, we can't merge it without tests. Every feature we add and every bug we fix must be accompanied by at least one new test that confirms that the code was not working before. Without such a test, how can we know that what you commit actually works?

No problem, I'll publish the tests soon.

gp07-cmd:
I've added a regression test covering the 8-byte value formatting case and pushed an update to this PR. The test fails on the previous implementation due to the unmatched closing parenthesis and passes with the fix applied.

yegor256:
@gp07-cmd we already merged this one: #5725 looking forward to your next contribution!

0crat:
@gp07-cmd It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`fix-verbose-bytes-parenthesis`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`5744`". Your running score is [-6](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5738, opened 2026-07-22
`merge-monikers` in eo-printer inlines real attribute names, not only obfuscated ones


yegor256:
The moniker-merging pass at `eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl:51` (function `eo:moniker-binding`) fires for any named bound attribute that is not void, `φ`, a test, or pipe-floated. It applies no check on the *kind* of name, so a meaningful user-given name is inlined at its reference site exactly like an auto-generated one. The pack `eo-printer/src/test/resources/org/eolang/printer/print-packs/yaml/moniker.yaml` shows this: the real name `value` (`number num.as-bytes > value`) is moved inline into the `if.` block.

A moniker — a binding written in place of its bare reference — is appropriate only for the compiler's auto-generated, anonymous names (the `a🌵` cactus prefix, §9.2), which have no meaningful standalone identity. A real, author-chosen name such as `value` (or `idx`) is a declared member of the object and reads best on its own `... > name` line; folding it into an expression hides the very name the author picked. Moniker merging should therefore be limited to obfuscated attributes.

The smallest fix is to add a guard to `eo:moniker-binding` requiring the attribute's `@name` to carry the auto-name cactus prefix (`$eo:cactoos`, i.e. `a🌵`), so user names stay standalone, and to update `moniker.yaml` to expect the standalone spelling. Note this narrows #5707 for real names and overlaps with `inline-cactoos.xsl`, which already inlines cactus-named abstracts — worth deciding whether `merge-monikers` is still needed once so restricted.

0crat:
@yegor256 Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+2741](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5749, opened 2026-07-22
fix(#5739): merge monikers onto the first of several references

### LINKED ISSUE #5739: `merge-monikers.xsl` limits binding merge to single hostable reference, limiting readability
`merge-monikers.xsl` merges a binding only when it has exactly one hostable reference: `eo:hosted-binding` at `eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl:70` guards on `count(eo:moniker-refs(...)) = 1`, added in #5707. The proposal is to drop that guard and instead always merge onto the *first* hostable reference in document order (`eo:moniker-refs(...)[1]`, the pre-#5707 behaviour), so a binding with several references still becomes a moniker.

This is a readability choice: the reader meets the definition at its first use and understands the later references from it, which is the natural top-to-bottom reading order. A name used three times reads better introduced once, in place, than hoisted to a separate trailing line.

The tradeoff, accepted here deliberately, is that the print stops being a fixpoint: #5706's canonical attribute sort can move which reference is "first" between passes, so re-printing may shift the moniker — exactly the divergence #5707 removed. This means the `.eo` format check and the `printsToParseableEo` idempotency test in `eo-printer/src/test/java/org/eolang/printer/XmirTest.java` would need to stop asserting a moniker fixpoint. The fix is small — restore `eo:hosted-binding` to `eo:moniker-refs($binding)[1] is $ref` and remove the drop-guard's `count(...) = 1` — but it must land together with relaxing those idempotency assertions. Relates to #5738, which narrows *which* attributes qualify.

yegor256:
Closes #5739. Rebased onto latest `master` (after #5738 merged).

## What

`merge-monikers.xsl` merged a binding only when it had exactly one hostable reference (the `count(eo:moniker-refs(...)) = 1` guard added in #5707). This drops that guard and restores the pre-#5707 behaviour: an eligible binding with several hostable references still becomes a moniker, landing on the **first** reference in document order (`eo:moniker-refs($binding)[1] is $ref`).

Eligibility is left exactly as #5738 defined it — only the compiler's obfuscated `a🌵` names are merged, so an author's chosen name is never moved.

## The accepted trade-off

The print stops being a single-pass fixpoint: #5706's canonical attribute sort can move which reference is "first" between passes, so re-printing may shift the moniker. The two assertions that relied on the single-pass fixpoint are relaxed to land together with the XSL change:

- **`eo-printer` — `printsToParseableEo`**: now honours a per-pack `reprints: false` flag for packs whose printed form is not expected to reprint to itself.
- **`eo-maven-plugin` — the `.eo` format check (`MjFormat`)**: computes the canonical form as the *fixpoint* of parse-and-print instead of asserting a single pass is already one. The moniker's home is decided by structure alone (attribute order + the XSL merge), not by the layout weights, so it settles at the always-reparseable default layout and then lays the settled structure out once with the configured weights (a custom `step` would otherwise emit indentation the parser cannot read back).

## Scope note (post-#5738)

Now that #5738 restricts merging to obfuscated names, **no eligible (bound, obfuscated) binding in `eo-runtime` currently has more than one hostable reference** — the multi-reference case is a no-op on today's sources (verified: zero `.eo` files change). The guard removal restores the general behaviour the issue asks for, and the relaxations above are kept as future-proofing for when such a binding does arise. No `eo-runtime` sources are reformatted.

## Testing

- `eo-printer`: full module suite — 173 tests, 0 failures.
- `eo-maven-plugin`: `MjFormatTest` — 5 tests, 0 failures.

Relates to #5738.

https://claude.ai/code/session_01UwV3xFsXu8FzAPKc3xdHp1

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
@@ -25,8 +25,10 @@ import org.eolang.printer.Xmir;
  *
  * <p>This goal walks through all registered {@code .eo} sources (see
  * {@link MjRegister}), reformats every one of them in memory by parsing
- * it to XMIR and printing it back with {@link Xmir#toEO()}, and compares
- * the result against what is on disk. In its default "check" mode, it
+ * it to XMIR and printing it back with {@link Xmir#toEO()} until the text
+ * stops changing (the moniker merge of #5739 is not a single-pass
+ * fixpoint, so the canonical form is the fixpoint of parse-and-print), and
+ * compares that against what is on disk. In its default "check" mode, it
  * prints a colored unified {@link Diff} for every file that diverges from
  * the canonical form and fails the build. When {@link #autoFix} is turned
  * on (via the {@code eo.autoFix} property), it overwrites the divergent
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2667](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5750, opened 2026-07-22
#5719: make MjSafe.exitError always fail the build regardless of unrollExitError

### LINKED ISSUE #5719: `MjSafe.exitError` incorrectly reports successful build when `unrollExitError` is false and an exception occurs
## What Happens

`MjSafe.execute()` (`eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java`) catches a `TimeoutException`/`ExecutionException` from the real Mojo work and calls `this.exitError(msg, ex)`. `exitError()` starts with:

```java
private void exitError(final String msg, final Throwable exp) throws MojoFailureException {
    if (!this.unrollExitError) {
        return;
    }
    ...
    throw out;
}
```

When the `eo.unrollExitError` parameter is set to `false`, this method just returns. No exception is thrown at all. Control falls through to the `finally` block and `execute()` returns normally, so Maven considers the goal, and the whole build, successful, even though the underlying `exec()` threw a real exception.

This contradicts the parameter's own Javadoc, which describes it as controlling whether the exception gets logged in detail ("If set to TRUE, the exception on exit will be printed in details to the log"), implying `FALSE` should only suppress the verbose diagnostic dump, not the failure itself.

There is no existing test for `MjSafe` or this parameter anywhere in the test suite.

Verified with a small JUnit test using a fake Mojo whose `exec()` always throws `IOException`, run through the project's own `FakeMaven`/`Moja` test harness:
- `unrollExitError=true`: the harness throws `IllegalStateException` as expected (build correctly fails).
- `unrollExitError=false`: no exception is thrown at all; the call completes silently (build falsely reports success).

## What Should Happen

`unrollExitError=false` should suppress the detailed cause-chain logging only. The build must still fail (`MojoFailureException` must still be thrown) regardless of that flag's value.

morphqdd:
Closes #5719.

`MjSafe.exitError()` used to just `return` without throwing anything when `eo.unrollExitError` was `false`, so a real exception from the Mojo's work was silently swallowed and Maven reported the build as successful. This contradicted the parameter's own Javadoc, which describes it only as controlling whether the exception is logged in detail, not whether the build fails.

Fix: `exitError()` now always throws `MojoFailureException`. The `unrollExitError` flag only gates the detailed cause-chain logging, which is now extracted into a separate `logCauses` helper (needed to keep the method's cognitive complexity under PMD's threshold once the early return was removed).

Added `MjSafeTest.java` with a minimal `Failing` Mojo whose `exec()` always throws:
- `unrollExitError=true`: throws `MojoFailureException` with the cause chain logged.
- `unrollExitError=false`: this is the actual regression test for the bug. Still throws `MojoFailureException`, just without the detailed log.

Verified the regression test is real by temporarily reverting the fix and re-running: the `unrollExitError=false` test fails against the old code (nothing thrown), passes with the fix.

Verified:
- `MjSafeTest`: 2/2 green.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java
@@ -683,10 +683,17 @@ abstract class MjSafe extends AbstractMojo {
      */
     private void exitError(final String msg, final Throwable exp)
         throws MojoFailureException {
-        if (!this.unrollExitError) {
-            return;
+        if (this.unrollExitError) {
+            this.logCauses(exp);
         }
-        final MojoFailureException out = new MojoFailureException(msg, exp);
+        throw new MojoFailureException(msg, exp);
+    }
+
+    /**
+     * Log the deduplicated cause chain of the given exception.
+     * @param exp Original problem
+     */
+    private void logCauses(final Throwable exp) {
         final List<String> causes = MjSafe.causes(exp);
         for (int pos = 0; pos < causes.size(); ++pos) {
             final String cause = causes.get(pos);
@@ -714,7 +721,6 @@ abstract class MjSafe extends AbstractMojo {
         for (final String cause : new LinkedHashSet<>(causes)) {
             Logger.error(this, cause);
         }
-        throw out;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1565](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5756, opened 2026-07-22
Type atoms with generic variables and annotate their voids

### LINKED ISSUE #5741: Incorrect type annotation for `recovered` in EO language due to lack of generic type variables
`recovered` in `eo-runtime/src/main/eo/recovered.eo:15` is annotated `/bytes`, yet it never yields bytes. It returns its `value` when that value is not ⊥ and its `alternative` otherwise, so the result is whatever type the two arguments share — its real signature is `for-any A: A? -> A -> A`.

```
[value alternative] > recovered /bytes
```

The `/bytes` is not a typo a better constant would fix. A forma names one fixed result type, but `recovered` — like `if`, `tuple.head`, `map.get` and `dataized` — is polymorphic: its result depends on its arguments, and here it also carries ⊥. Vertical void attributes and brace-formas arrived with #5267, and #3561 already noted that atom results go unvalidated, but there is still no way to write a type *variable*. So the true type of `recovered` is inexpressible today, and the placeholder `/bytes` stands in as a trusted-but-false axiom that any reader of the forma — a type checker, generated docs, an IDE — will believe.

The fix is to add generic type variables to the annotation language and let them ride on the `? > name` void attributes from #5267:

```
[] > recovered /A
  ? > value /A?
  ? > alternative /A
```

Here `A` is universally quantified and shared across the object, `/A?` marks a maybe-⊥ argument, and the same machinery covers `if` (result `/A|B`), the tuple accessors, and the rest. That states `recovered` exactly and retires the placeholder.

Genericity also settles the nested case for free: in `recovered (recovered v1 v2) "hi"`, `A` binds to an optional type for the fragile inner call and to a definite type for the outer one, so the ⊥ terminates only when a definite fallback is supplied — no second annotation needed, given the single rule that `A?? == A?`. That is precisely why `alternative` should stay `/A` and not be weakened to `/A?`. Once formas can carry generics these signatures become the one source of truth a type checker reads for its atoms instead of a hand-maintained table that drifts as the runtime grows, and — closing the gap #3561 pointed at — any atom that has an EO body can have its declared type verified against that body.

maxonfjvipon:
`recovered` claimed to return `/bytes`, but it never does — it hands back whichever type its two arguments share, and may carry ⊥. Every polymorphic atom hit the same wall: `/sig` could only name one fixed forma, so the honest signature was inexpressible and the placeholder stood in as a trusted-but-false axiom.

This adds generic type variables `A`–`F` to the annotation language. An atom can declare a generic return, and a vertical void inside an atom may carry its own type — a bare `/type` (with an optional trailing `?` for a maybe-⊥ input) or a brace `/{type …}` for a callback's argument formas. So `recovered` becomes:

```
[] > recovered /A
  ? > value /A?
  ? > alternative /A
```

A variable travels through the passes verbatim; concrete formas keep their `Q.`→`Φ.` promotion. The reverse printer round-trips both forms so annotated atoms stay canonical, and the opt-in return-type check skips generic returns, since a variable cannot be verified by forma equality. The `@types` attribute the brace list used is renamed to `@args`, freeing `@type` for the new bare form.

Follow-up: the type checker that reads these signatures — variable unification, the `A?? == A?` rule, checking an atom's body against its declared type — is a separate piece. This makes the truth expressible; enforcing it comes next.

Closes #5741.


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (643 >= 100); -8 for way too many hits-of-code (643 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+764](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5759, opened 2026-07-22
Use `>>` for private object attributes instead of `>`

### LINKED ISSUE #5678: `>` symbol incorrectly used for obfuscation instead of `>>` in private object attributes
We are supposed to use `>>` more often than `>`, because most object attributes should be obfuscated. Take `string` object for example:

```
    80- > p1
    E0- > p2
    F0- > p3
    F8- > p4
    00- > r1
    C0- > r2
    p2 > r3
    p3 > r4
    start > sbts!
    as-bytes.size > size
```

All these objects are "private" and must have `>>` instead of `>`.

I suggest swapping the semantics: `>>` should mean "public", while `>` should mean obfuscated.

The primary reason is that it's going to be less symbols to type and to read.

Also, `[a] > foo` will mean that `a` is obfuscated. To make it public, one should write this:

```
[] > foo
  ? >> a
```

Somehow related to https://github.com/objectionary/eo/issues/5556.

yegor256:
Private object attributes across eo-runtime were bound with `>` where they should use `>>`, the symbol meant for obfuscated (private) attributes. This change switches them over, and while touching those objects it also renames a few helpers and reworks the byte-array handling, conversions, and the tuple `index-of` / `last-index-of` methods so they read more consistently.

The motivation is issue #5678: `>` was standing in for `>>` on private attributes, so the source misrepresented what those attributes are. Master has since moved on, so this branch also merges the latest master and resolves the two tuple conflicts that came with it.

It stays a refactor with no behaviour change, and the eo-runtime build and its test suite pass. Closes #5678.


yegor256:
@yegor256 could you take a look when you get a chance? Happy to explain anything about the conflict resolution or the moniker changes if it helps.

Issue #5718, opened 2026-07-21
`Penalty` incorrectly counts string literals as code structure in eo-printer


yegor256:
`Penalty`'s helpers in `eo-printer/src/main/java/org/eolang/printer/Penalty.java` — `tokens()` (214-234), `applied()` (192-205), `brackets()` (259-272), `phis()` (279-287), `ifs()` (301-318) and `spaces()` (160-172) — scan a line's raw characters/tokens with no notion of a quoted string literal. Data sitting inside a `"…"` string (spaces between words, a stray `(`/`)`, an `@`, or the substring `.if`) is counted exactly like real EO syntax.

This contradicts the class's own doc comment, which defines the penalty purely in terms of code structure. Two 66-character lines that differ only in spaces vs. underscores inside a string literal score 633 vs. 73 points respectively — a 560-point swing driven entirely by word-spacing in quoted data, not by any difference in code structure. Real sources are affected too: `eo-runtime/src/main/eo/string.eo` has calls like `"Unknown byte format (%x), can't recognize character".printf`, where the parenthesis inside the string is counted by `brackets()` as if it were real nesting.

Since `Pretty.java` (lines 118-119, 161-162, 166-167, 361-362) picks between candidate renderings by comparing `Penalty.points()`, a formation whose only distinguishing content is a longer, multi-word string literal gets its flat-candidate score inflated by the wording of the string, pushing `eo:format` toward a more vertical layout than the code's actual structure warrants.

Fix: track quote state (toggle on unescaped `"`) in `tokens()`, `brackets()`, `phis()`, `ifs()`, and `spaces()`, and skip/neutralize characters while inside a string literal so only genuine code structure is scored.

TheDarkniteFalls:
I'd like to take this. If assigned, I'll limit the change to quote-aware scanning in `Penalty` plus focused tests for spaces, parentheses, `@`, `.if`, and escaped quotes, while preserving penalties for real syntax. I'll run the targeted `eo-printer` tests and `mvn clean install -Pqulice`. I use OpenAI Codex as an assistance tool and remain responsible for the contribution.

yegor256:
I believe, it was a wrong move. This code:

```
"Hello, %s!".printf (* "John Doe")
```

looks less readable than this:

```
printf.
  "Hello, %s!"
  * "John Doe"
```

Every space inside a string literal is readable by a programmer and must be comprehended. It's better to keep string literals on their own lines.

PR #5764, opened 2026-07-22
fix(eo-parser): ignore quoted content in redundant-paren check (#5722)

### LINKED ISSUE #5722: `Tokens.readGroup` incorrectly flags single-token string literals in parentheses as redundant
## What Happens

`eo-parser/src/main/java/org/eolang/parser/Tokens.java`, method `readGroup()` (around lines 241-255), decides whether a parenthesized group is "redundant around a single token" by scanning the raw substring between the parens for a space, or for `(`, `)`, `[`, `]` (the guard at lines 242-247, feeding the error message at line 251). This scan has no notion of being inside a quoted string literal.

A quoted string is always a single atomic token regardless of any spaces, brackets, or parentheses inside its quotes, but this heuristic is fooled by such content into thinking the parens wrap more than one token.

Verified through the full parser pipeline (`EoSyntax`):
```
1.add ("hello") > x         gives: redundant parentheses around a single token error
1.add ("hello world") > x   is accepted silently, no error
```
Both are semantically identical cases (a parenthesized single string-literal argument), but only the single-word one is flagged. The same asymmetry reproduces for parentheses or brackets embedded in the string: `("a(b)c")` and `("a[b]c")` are also silently accepted, even though they are exactly as "redundant" as the flagged single-word case.

This is the same class of bug already found and fixed for `Penalty.java` in issue #5718 (a scanner over source text with no notion of quoted-string content).

This does not corrupt data like the related `StUnhex` bug does; it makes the "redundant parentheses" lint unreliable and inconsistent. That lint's correctness also matters for the printer side of the same general topic (issue #5591, already fixed, depends on stable round-tripping here).

## What Should Happen

`readGroup()`'s single-token detection should track quote state (toggle on an unescaped `"`) while scanning the group's content, and skip/neutralize characters while inside a string literal, so a parenthesized single string-literal argument is always treated consistently regardless of what characters happen to appear inside its quotes.

arimu1:
Closes #5722.

## Problem

`Tokens.readGroup()` decided whether a parenthesized group wraps a single
token by scanning the raw substring between the parens for a space or a
`(`/`)`/`[`/`]`. That scan had no notion of quoted strings, so it was fooled
by string content:

```
1.add ("hello") > x         → redundant-parentheses error
1.add ("hello world") > x   → silently accepted
```

Both are the same case — a parenthesized single string-literal argument —
but only the single-word one was flagged. The same asymmetry hit
`("a(b)c")` / `("a[b]c")`. This is the same class of bug fixed for
`Penalty.java` in #5718.

## Fix

Replaced the naive `indexOf` checks with `singleToken()`, which scans the
group content while tracking quote state — skipping characters inside
double quotes exactly as the surrounding group scanner already does — and
reports a single token only when no separator appears **outside** a string
literal. A parenthesized single string literal is now flagged consistently
regardless of what characters appear inside its quotes.

## Tests

Added three `eo-typos/redundant-parentheses` packs: a plain quoted string,
a quoted string containing a space (the previously-accepted regression
case), and one containing brackets. `EoSyntaxTest#checksTypoPacks` green
(77/77).

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Tokens.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Tokens.java
@@ -239,12 +239,7 @@ final class Tokens {
             );
         }
         final String inside = this.body.substring(start + 1, this.cursor - 1);
-        if (!inside.isEmpty()
-            && inside.indexOf(' ') < 0
-            && inside.indexOf('(') < 0
-            && inside.indexOf(')') < 0
-            && inside.indexOf('[') < 0
-            && inside.indexOf(']') < 0) {
+        if (!inside.isEmpty() && Tokens.singleToken(inside)) {
             throw new ParseError(
                 this.span.line(), this.span.indent() + start,
                 String.format(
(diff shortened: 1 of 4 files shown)


0crat:
@arimu1 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+44](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5765, opened 2026-07-22
fix(eo-printer): preserve argument applied to a data literal (#5721)

### LINKED ISSUE #5721: `StUnhex` in eo-printer loses arguments when applied directly to data literals
## What Happens

Reproduced the same way for:

## What Should Happen

arimu1:
Closes #5721.

## Problem

`StUnhex` converts a data literal's hex payload to readable EO by calling
`setTextContent` on the literal's wrapper `<o>` element. In the DOM,
`setTextContent` deletes **all** of the element's children and replaces
them with a single text node.

The XPath guards only require that the literal's first child is its own byte
payload; they never require it to be the *only* child. A bare data literal
can legitimately be the head of an application in parser-accepted EO (for
example `42 5 > x`), where the wrapper's second child is the applied
argument. `setTextContent` wiped that argument out, so the printer silently
turned `42 5 > x` into `42 > x` (and likewise for `"str" 5`, `01-02 5`,
and compact-tuple arguments).

## Fix

Add `inline()`, which replaces only the byte-payload child (the first `<o>`)
with a text node, leaving any sibling argument in place. For the common
single-child case the result is identical to before; when an argument is
present it now survives.

## Tests

Added `keepsArgumentAppliedToDataLiteral`, which unhexes a number literal
carrying a sibling argument and asserts both the converted number and the
argument survive. Verified it fails without the fix; the full eo-printer
suite (174 tests, incl. 137 end-to-end `XmirTest`) is green.

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
+++ b/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
@@ -4,6 +4,7 @@
  */
 package org.eolang.printer;
 
+import com.github.lombrozo.xnav.Filter;
 import com.github.lombrozo.xnav.Xnav;
 import com.yegor256.xsline.Shift;
 import com.yegor256.xsline.StEnvelope;
@@ -16,7 +17,6 @@ import java.util.Optional;
 import java.util.regex.Pattern;
 import org.eolang.parser.StXnav;
 import org.w3c.dom.Node;
-import org.w3c.dom.NodeList;
 
 /**
  * This {@link Shift} turns hex data inside XMIR.
@@ -100,14 +100,7 @@ final class StUnhex extends StEnvelope {
      */
     private static void inline(final Xnav xnav, final String text) {
         final Node parent = xnav.node();
-        final NodeList children = parent.getChildNodes();
-        int elements = 0;
-        for (int idx = 0; idx < children.getLength(); ++idx) {
-            if (children.item(idx).getNodeType() == Node.ELEMENT_NODE) {
-                elements = elements + 1;
-            }
-        }
-        if (elements > 1) {
+        if (xnav.elements(Filter.withName("o")).count() > 1L) {
             parent.replaceChild(
                 parent.getOwnerDocument().createTextNode(text),
                 xnav.element("o").node()
--- a/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
+++ b/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
@@ -4,6 +4,7 @@
  */
 package org.eolang.printer;
 
+import com.github.lombrozo.xnav.Xnav;
 import com.yegor256.xsline.Shift;
 import com.yegor256.xsline.StEnvelope;
 import com.yegor256.xsline.StSequence;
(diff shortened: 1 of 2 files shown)


arimu1:
The only red leg is `mvn (windows-2022, 26)`, which failed with `java.lang.OutOfMemoryError: Java heap space` in `org.eolang.EO_string.EOscanfTest` (eo-runtime) — a Windows CI heap flake unrelated to this change. This PR only touches `eo-printer`; its full suite (eo-printer 174/174, all eo-parser tests) is green on every OS. A re-run of the Windows leg should clear it. 🙏

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -87,6 +88,35 @@ final class StUnhex extends StEnvelope {
...
+        final Node parent = xnav.node();
+        final NodeList children = parent.getChildNodes();
+        int elements = 0;
+        for (int idx = 0; idx < children.getLength(); ++idx) {
+            if (children.item(idx).getNodeType() == Node.ELEMENT_NODE) {
+                elements = elements + 1;
+            }
+        }
+        if (elements > 1) {
+            parent.replaceChild(
```
yegor256, line 104: @arimu1 cant't we use XNav instead?

arimu1: @yegor256 Good call — switched the child walk to XNav:

```java
if (xnav.elements(Filter.withName("o")).count() > 1L) {
    parent.replaceChild(
        parent.getOwnerDocument().createTextNode(text),
        xnav.element("o").node()
    );
} else {
    parent.setTextContent(text);
}
```

`StUnhexTest` (13/13) still green on JDK 21. Mutation still goes through the DOM node (XNav is read-side only), but the sibling count is pure XNav now.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -87,6 +88,28 @@ final class StUnhex extends StEnvelope {
...
+     */
+    private static void inline(final Xnav xnav, final String text) {
+        final Node parent = xnav.node();
+        if (xnav.elements(Filter.withName("o")).count() > 1L) {
+            parent.replaceChild(
+                parent.getOwnerDocument().createTextNode(text),
+                xnav.element("o").node()
+            );
+        } else {
+            parent.setTextContent(text);
```
morphqdd, line 109: This only replaces the payload element child, it doesn't remove the whitespace-only text nodes real (indented, pretty-printed) XMIR has between sibling `<o>` elements. Fed an indented wrapper (payload + sibling argument, same shape the real pipeline produces), `wrapper.getTextContent()` comes out as something like `"\n    42\n    \n  "` instead of `"42"` — the printed literal gets corrupted with embedded newlines. The tests here use single-line/compact XML strings so they don't catch this. Also, #5767 targets the same issue and handles this case (strips text children before replacing the payload).

yegor256:

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1962](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+5939](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #4900, opened 2026-02-25
ConcurrentCache.java Fails to Use ReentrantLock for Thread Safety


0pdd:
The puzzle `4884-94301545` from #4884 has to be resolved: 

https://github.com/objectionary/eo/blob/9efc601f82f1b756d0d87c111b87c2007615c2d9/eo-maven-plugin/src/main/java/org/eolang/maven/ConcurrentCache.java#L15-L18

The puzzle was created by @yegor256 on 25-Feb-26. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
The puzzle `4884-94301545` has disappeared from the source code, that's why I closed this issue.

Issue #2857, opened 2024-02-05
global cache doesn't have a global mutex


yegor256:
Currently, all our Mojos assume that they are the solo owners of the global cache. There is not synchronization, which may soon lead to problems, if two projects will try to compile at the same machine. Let's introduce a global file-based semaphore.

levBagryansky:
@yegor256  This approach with global semaphore will lead to the fact that parallel compilations will always be performed in fact sequentially. 
As I understand `cargo` has local cache somewhere in `target/debug` since removing of the directory increases compilation time a lot.
```shell
tardis3@R:~/rust_trash$ cargo build
   Compiling libc v0.2.147
   Compiling proc-macro2 v1.0.63
   Compiling memchr v2.5.0
   Compiling quote v1.0.29
   Compiling clang-sys v1.6.1
   Compiling indexmap v1.9.3
   Compiling bindgen v0.60.1
   Compiling libffi-sys v2.3.0
   Compiling memoffset v0.6.5
   Compiling thiserror v1.0.40
   Compiling mpi v0.6.0
   Compiling aho-corasick v1.0.2
   Compiling nom v7.1.3
   Compiling combine v4.6.6
   Compiling atty v0.2.14
   Compiling which v4.4.0
   Compiling rand v0.4.6
   Compiling clap v3.2.25
   Compiling syn v2.0.23
   Compiling regex v1.8.4
   Compiling cexpr v0.6.0
   Compiling env_logger v0.9.3
   Compiling thiserror-impl v1.0.40
   Compiling jni v0.21.1
   Compiling mpi-sys v0.2.0
   Compiling libffi v3.2.0
   Compiling rust_trash v0.1.0 (/home/tardis3/rust_trash)
    Finished dev [unoptimized + debuginfo] target(s) in 14.39s
tardis3@R:~/rust_trash$ cargo build
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
```

levBagryansky:
@yegor256 This approach with global semaphore will lead to the fact that parallel compilations will always be performed in fact sequentially.
As I understand cargo has local cache somewhere in target/debug since removing of the directory increases compilation time a lot.
```shell
tardis3@R:~/rust_trash$ cargo build
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
tardis3@R:~/rust_trash$ rm -r target/debug/build/
tardis3@R:~/rust_trash$ cargo build
   Compiling libc v0.2.147
[...]

PR #5766, opened 2026-07-22
fix(#5720): share one ConcurrentCache across files so locking works

### LINKED ISSUE #5720: `ConcurrentCache` instances aren't sharing lock map, defeating intended thread safety
## What Happens

`ConcurrentCache` (`eo-maven-plugin/src/main/java/org/eolang/maven/ConcurrentCache.java`) says in its own Javadoc that it wraps `Cache` to make it thread-safe, and `Cache`'s own Javadoc says "This class isn't thread-safe, use `ConcurrentCache` for concurrent scenarios." Its mutual-exclusion mechanism is a `ConcurrentMap<Path, ReentrantLock>` keyed by the cache "tail" path, so two threads writing to the same cache location are supposed to serialize on the same lock.

But at all three places this class is ever instantiated in the codebase, `Parsing.parsed()`, `Transpiling.transpiled()`, and `Linting.lintOne()`, all of which run per-file inside a `Threaded` worker pool, the code does:

```java
new ConcurrentCache(new Cache(...)).apply(...)
```

A brand-new instance, with a brand-new and empty lock map, is created for every single file, used exactly once, and discarded. Since the lock map is never shared between threads, two threads racing on the same cache tail path each get their own uncontended `ReentrantLock` and never actually block each other.

This is a related but distinct problem from issue #4900 ("ConcurrentCache Fails to Use ReentrantLock for Thread Safety"), which was closed by fixing the locking mechanism inside the class itself (swapping `synchronized` for a `ReentrantLock`). That fix never touched the call-site wiring, which still creates a fresh instance per file and so still defeats the (now-correct) internal locking today.

Verified with two JUnit tests mirroring the real call sites:
- Two threads each doing `new ConcurrentCache(new Cache(...)).apply(source, target, sameTail)` (a fresh instance per call, exactly like the real code): the test recorded that both threads entered the compilation callback simultaneously. Confirmed: no serialization happens.
- Control test: two threads sharing one `ConcurrentCache` instance for the same call: no overlap occurred. This confirms the locking mechanism itself is sound, and the bug is specifically in the per-call-site instantiation.

## What Should Happen

A single `ConcurrentCache` instance, sharing one lock map, should be constructed once per Mojo execution and reused across all files processed by the `Threaded` pool in `Parsing`, `Transpiling`, and `Linting`, so concurrent writes to the same cache tail path are actually serialized.

anuragpaul602-netizen:
Closes #5720.

## Problem

`ConcurrentCache` documents itself as making `Cache` thread-safe via a `ConcurrentMap<Path, ReentrantLock>` keyed by cache tail path. But at all three call sites — `Parsing.parsed()`, `Transpiling.transpiled()`, `Linting.lintOne()` — a **new** instance was created per file:

```java
new ConcurrentCache(new Cache(...)).apply(...)
```

Each file therefore got its own empty lock map, so two `Threaded` workers racing on the same cache tail path each acquired their own uncontended lock and never serialized. The (correct) internal locking from #4900 was defeated by the call-site wiring.

## Fix

- `ConcurrentCache` now holds **only** the shared lock map; the per-file `Cache` (which legitimately differs per file) is passed into `apply(cache, source, target, tail)`.
- `Parsing`, `Transpiling`, and `Linting` each keep a single `private final ConcurrentCache guard = new ConcurrentCache()` field, reused across every file processed by the `Threaded` pool, so writes to the same tail path share one lock.

## Tests

`ConcurrentCacheTest` updated to the new API and passing.


anuragpaul602-netizen:
Ready for review. This fixes the call-site wiring that #4900 didn't touch: `ConcurrentCache` was constructed fresh per file, so every thread got its own empty lock map and same-tail writes never actually serialized. Here `ConcurrentCache` holds only the shared lock map, and each `Step` (`Parsing`/`Transpiling`/`Linting`) keeps a single reused instance, passing the per-file `Cache` into `apply(...)`.

Verified locally: `ConcurrentCacheTest` passes (updated to the new API). Happy to adjust the `guard` field naming/placement if a different convention is preferred.

yegor256:
@anuragpaul602-netizen the `qulice` fails here, please fix

eo-maven-plugin/src/main/java/org/eolang/maven/ConcurrentCache.java
```diff
@@ -10,53 +10,57 @@
...
 final class ConcurrentCache {
 
     /**
-     * Original cache.
-     */
-    private final Cache original;
-
-    /**
-     * Locks for each cache entry.
+     * Locks for each cache entry, shared across every file of one run.
```
morphqdd, line 32: Correct diagnosis in the doc: the old code's `locks` map was fresh (and empty) per `ConcurrentCache` instance, and instances were being created per file, so two threads racing on the same tail path never actually shared a lock. This comment correctly identifies why a single shared instance is required.

eo-maven-plugin/src/main/java/org/eolang/maven/ConcurrentCache.java
```diff
@@ -10,53 +10,57 @@
...
     private final ConcurrentMap<Path, ReentrantLock> locks;
 
     /**
      * Ctor.
-     * @param original Original cache
      */
-    ConcurrentCache(final Cache original) {
-        this(original, new ConcurrentHashMap<>(0));
+    ConcurrentCache() {
+        this(new ConcurrentHashMap<>(0));
```
morphqdd, line 40: This is the actual root-cause fix: the no-arg constructor no longer takes a `Cache`, meaning one `ConcurrentCache` (and its one `locks` map) can now be built once and reused for every file, instead of being thrown away and rebuilt per file.

eo-maven-plugin/src/main/java/org/eolang/maven/ConcurrentCache.java
```diff
@@ -10,53 +10,57 @@
...
+     * Check cache and apply compilation if needed, holding the lock for the
+     * given tail path so concurrent writes to it are serialized.
+     * @param cache The per-file cache to run under the lock
      * @param source From file
      * @param target To file
      * @param tail Tail path in cache
+     * @checkstyle ParameterNumberCheck (3 lines)
      */
-    void apply(final Path source, final Path target, final Path tail) {
+    void apply(final Cache cache, final Path source, final Path target, final Path tail) {
```
morphqdd: Right call to make `cache` a parameter of `apply()` rather than part of the constructor: the per-file `Cache` (the actual compilation lambda) genuinely differs per file, while the `locks` map must not. Separating "what varies per file" from "what must be shared across files" is exactly the fix this bug needed.

eo-maven-plugin/src/main/java/org/eolang/maven/ConcurrentCache.java
```diff
@@ -10,53 +10,57 @@
...
         final ReentrantLock lock = this.locks.computeIfAbsent(
             tail.normalize(), k -> new ReentrantLock()
         );
         lock.lock();
         try {
-            this.original.apply(source, target, tail.normalize());
+            cache.apply(source, target, tail.normalize());
```
morphqdd, line 65: Confirmed correct: the per-file `cache.apply(...)` call happens inside the lock acquired from the shared `locks` map keyed by `tail`, so two threads now genuinely serialize on the same lock object for the same tail path.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -142,6 +142,12 @@ final class Linting implements Step {
      */
     private final boolean skipLinting;
 
+    /**
+     * Shared cache guard, reused across all files so that concurrent writes to
+     * the same cache tail path are actually serialized (#5720).
+     */
+    private final ConcurrentCache guard;
```
morphqdd, line 148: Correct: one `guard` field per `Linting` instance, matching the fix's requirement that a single `ConcurrentCache` be shared across every file this step processes.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -292,7 +299,7 @@ private int lintOne(
             new OnDetailed(new OnDefault(new Xnav(xmir.inner())), source).get()
         ).make(base, MjAssemble.XMIR);
         if (this.cacheEnabled) {
-            new ConcurrentCache(
+            this.guard.apply(
```
morphqdd, line 301: This is where the fix actually takes effect for linting: `this.guard.apply(...)` instead of `new ConcurrentCache(...).apply(...)`. Verified via `mvn -pl eo-maven-plugin test -Dtest=LintingTest`, though see my note on `ConcurrentCacheTest.java` -- nothing currently drives two concurrent calls through this exact call site to prove the sharing works end to end.

eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
```diff
@@ -93,6 +93,12 @@ final class Parsing implements Step {
      */
     private final Path sourcesDir;
 
+    /**
+     * Shared cache guard, reused across all files so that concurrent writes to
+     * the same cache tail path are actually serialized (#5720).
+     */
+    private final ConcurrentCache guard;
```
morphqdd, line 99: Same pattern applied consistently here for parsing. Good -- all three call sites (`Linting`, `Parsing`, `Transpiling`) get the fix, not just one.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -165,6 +165,12 @@ final class Transpiling implements Step {
      */
     private final boolean coverage;
 
+    /**
+     * Shared cache guard, reused across all files so that concurrent writes to
+     * the same cache tail path are actually serialized (#5720).
+     */
+    private final ConcurrentCache guard;
```
morphqdd, line 171: And here for transpiling too. Confirmed via `mvn -pl eo-maven-plugin test -Dtest=MjTranspileTest` (40/40 green).

eo-maven-plugin/src/test/java/org/eolang/maven/ConcurrentCacheTest.java
```diff
@@ -41,7 +40,7 @@ void triesToCompileProgramConcurrently(@Mktmp final Path temp) throws IOExceptio
         final Path target = temp.resolve("program.xmir");
         new Threaded<>(
             IntStream.range(0, 100).boxed().collect(Collectors.toList()), ignored -> {
-            cache.apply(source, target, source.getFileName());
+            cache.apply(original, source, target, source.getFileName());
```
morphqdd: This is my main concern with the PR. This test was updated only mechanically (passing `cache`/`original` as a parameter instead of via the constructor), but it still constructs exactly one `ConcurrentCache` and hits it with 100 threads for the same tail path -- that would have passed identically against the *old*, buggy code too, since the original bug only manifested when *different* `ConcurrentCache` instances got created per file, which this test never does, before or after the fix. None of `LintingTest` (one unrelated test, `skipsLintingWhenFlagIsSet`), `MjParseTest`, or `MjTranspileTest` drive two concurrent calls to `lintOne`/`parsed`/`transpiled` for the *same* target path through a shared `Linting`/`Parsing`/`Transpiling` instance to prove `this.guard` is doing its job end to end. Right now the only thing standing between "fixed" and "silently still broken at the call site" is code reading, not a test. Would like to see at least one test that drives a `Step` with two threads targeting the same tail path and asserts the underlying work runs once, before merge.

morphqdd (commented):
Reviewed by checking out the branch, reading the full diff, and running `ConcurrentCacheTest`, `MjParseTest`, `MjTranspileTest`, `CompilingTest`, `LintingTest` (all green) plus the full `mvn clean verify -Pqulice` gate (clean). The fix itself correctly targets the root cause of #5720. Main concern, detailed inline: the test suite doesn't actually exercise the fix at the level where the original bug lived.

morphqdd (changes requested):
The fix itself (holding one `ConcurrentCache guard` field per `Parsing`/`Transpiling`/`Linting` instance instead of a fresh instance per file) is correct for the specific race described in #5720: two `Threaded` workers on the same instance racing on the same tail path. Three things beyond that scope, verified by reading the surrounding code, worth resolving before this closes the issue:

**1. `MjSafe.assembling()` (`MjSafe.java`, inside the `assembling()` method) creates a second, independent `Parsing` instance**, separate from the one `MjParse.exec()` creates directly. `ConcurrentCache`'s own new javadoc states the invariant this fix relies on: "a single instance must be shared across all files of one Mojo run." That invariant holds within one `Parsing` instance, but `Parsing` is instantiated in at least two places (`MjParse.java` and `MjSafe.assembling()`, the latter used by `MjAssemble` and `MjCompile`), each getting its own separate `guard` and lock map.

Within one module's sequential build this is fine, since these Mojos run at different lifecycle phases, not concurrently with each other. But under a parallel reactor build (`mvn -T N`, a common and supported mode) with the default cache directory, which is machine-wide rather than module-scoped (already flagged as issue #2857, still open), two different modules building in parallel could each spin up their own `Parsing` instance and still race on the same cache tail path if they happen to touch the same dependency. This PR closes the within-one-instance case but doesn't mention or address the cross-instance case. Worth at least a note in the javadoc about this residual gap, if not a fix tying the two together (a shared static lock map, or threading the guard down from a single owner).

**2. `Linting.lintAll()` writes to the WPA package-level cache entry via a raw `new Cache(...).apply(...)`**, bypassing `ConcurrentCache`/`this.guard` entirely. It's the one `Cache` call site in this class that isn't guarded. It's called once per `lint()` run, outside the per-file `Threaded` loop in `lintOne()`, so it's probably not racing against anything today, but nothing here says that, and it reads as an oversight next to the other three call sites that were all moved onto the shared guard. Either route it through `this.guard.apply(...)` too for consistency, or add a comment explaining why this specific write doesn't need it.

**3. No test exercises the actual bug this PR fixes.** I checked the existing concurrency tests: `MjParseTest.parsesConcurrentlyWithLotsOfPrograms` and `MjTranspileTest.transpilesSeveralEoProgramsInParallel` both spin up 30-50 different files, each with its own tail path, never two operations racing on the identical tail path, so they'd pass whether or not #5720 were fixed. `ConcurrentCacheTest` only exercises the `ConcurrentCache` primitive directly, which the PR description says was already correct before this change (per #4900); the bug was purely in the call-site wiring. Nothing tests that `Parsing`, `Transpiling`, or `Linting` actually share one `guard` instance across multiple calls, the exact thing this PR changes. A regression test that would have caught the original bug needs to exercise one of those three classes directly: two concurrent calls with the same tail path through the same instance, asserting the underlying compilation function runs once, not twice.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
morphqdd, line 148: This field's javadoc ("Shared cache guard, reused across all files so that concurrent writes to the same cache tail path are actually serialized (#5720).") is copied verbatim into `Parsing.java` and `Transpiling.java` too. Three identical multi-line doc comments for the same concept is the kind of duplication worth avoiding, since a future change to the explanation (or to the invariant it documents, see the `assembling()` comment on the main review) now has to be kept in sync by hand in three places. Worth centralizing, e.g. on `ConcurrentCache`'s own class javadoc (which already explains the sharing requirement) and just pointing to `{@link ConcurrentCache}` from each field here instead of repeating the explanation.

eo-maven-plugin/src/main/java/org/eolang/maven/ConcurrentCache.java
morphqdd: `apply(Cache cache, Path source, Path target, Path tail)` puts the newly-added `cache` parameter first, ahead of the established `source, target, tail` triple. `Cache.apply(source, target, tail)` itself (the method this delegates to) keeps that triple as its own full signature. Putting `cache` last here, `apply(source, target, tail, cache)`, would match the wrapped method's own parameter order and read as "the same call, plus which cache to use" rather than breaking the familiar order at the front.

eo-maven-plugin/src/test/java/org/eolang/maven/ConcurrentCacheTest.java
morphqdd: This still only tests the `ConcurrentCache` primitive directly, which per the PR description was already correct before this change (the bug was in the call-site wiring, not in this class's own locking). Nothing in this file or elsewhere in the diff tests that `Parsing`, `Transpiling`, or `Linting` actually share one `guard` instance across multiple calls now, which is the actual fix. A test at that level, e.g. calling the package-private `parsed`/`transpiled`/`lintOne` twice concurrently with the same tail path through one `Parsing`/`Transpiling`/`Linting` instance and asserting the wrapped compilation runs once, would have caught the original bug and would catch a regression if the wiring changes again.

morphqdd (commented):
A few more, anchored to the actual changed lines.

morphqdd (changes requested):
Confirming the verdict: request changes, per the two reviews above (the cross-instance `Parsing` sharing gap via `MjSafe.assembling()`, the unguarded WPA cache write in `Linting.lintAll()`, the missing regression test at the actual wiring level, plus the three smaller inline points on the duplicated field javadoc, the `apply()` parameter order, and the test file).

eo-maven-plugin/src/main/java/org/eolang/maven/ConcurrentCache.java
anuragpaul602-netizen: Done in bff3613 — the signature is now `apply(Path source, Path target, Path tail, Cache cache)`, with the javadoc `@param` order following it.

You're right that it reads better: the wrapped call keeps its own full signature at the front and the cache is what's appended. It also happens to lay out better at the call sites, because the multi-line `new Cache(...)` argument is now last instead of pushing the three paths onto a trailing line:

```java
this.guard.apply(
    source, target, base.relativize(target),
    new Cache(
        new CachePath(...),
        src -> ...
    )
);
```

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
anuragpaul602-netizen: Done in bff3613. All three fields (`Parsing`, `Transpiling`, `Linting`) now carry the same one-liner:

```java
/**
 * Cache guard, see {@link ConcurrentCache} for why it is one per instance.
 */
private final ConcurrentCache guard;
```

The explanation now exists once, on `ConcurrentCache`'s class javadoc — which is also where the residual cross-instance gap you raised in the main review is now recorded. So a future change to the invariant, or to what it does and doesn't cover, is one edit rather than three that have to be kept in sync.

eo-maven-plugin/src/test/java/org/eolang/maven/ConcurrentCacheTest.java
anuragpaul602-netizen: Agreed, and fixed in bff3613 — this file stays a test of the primitive (only the argument order changed), and the actual regression test is the new `ParsingTest`:

```java
new Threaded<>(
    Collections.nCopies(8, tojo),
    each -> { parsing.parsed(each, countingPipeline, "digest"); return 1; }
).total();
MatcherAssert.assertThat(..., parses.get(), Matchers.equalTo(1));
```

Eight threads go through **one** `Parsing` instance, with the same tojo — so the same name, the same target, and therefore the identical tail path. The pipeline handed in counts its own invocations. With the guard shared by the instance, the first thread compiles and the other seven wait on its lock and then read what it cached, so the count is exactly 1 and it is deterministic, not a lucky interleaving. With a guard per file, each thread locks its own empty map, all eight miss, and the source is parsed up to eight times.

I checked it fails on the bug rather than just passing on the fix: reverting `this.guard.apply(...)` back to `new ConcurrentCache().apply(...)` in `Parsing` alone fails the test on 3 out of 3 runs. `Parsing.parsed(TjForeign, UnaryOperator, String)` had to become package-private for the test to drive it — the javadoc says why.

Two caveats worth stating plainly:

- **It covers `Parsing` only.** `Transpiling.transpiled` and `Linting.lintOne` have the identical shape, and I'd rather have one honest reproduction than three copies of it. Say the word if you want all three.
- **It swallows one exception on purpose.** `Saved` truncates the target and streams into it, so when several threads legitimately share a target file, a thread reading `target` back for `tojoVersion`/`Xnav` can catch it half-written and get `Premature end of file`. That read sits *after* `guard.apply` returns, outside the guarded region, so it's a distinct race that this PR doesn't claim to fix — the test catches it, counts it, and asserts only on the parse count. It's real though (I hit it while writing the test), and I'm happy to open a separate issue for it if you agree it's worth one.

anuragpaul602-netizen:
Thanks for the careful read — all six points are addressed in bff3613. Replies to the three inline threads are in place; here are the three from the main review.

### 1. `MjSafe.assembling()` and the cross-instance gap

You're right, and I've documented it rather than fixed it. `ConcurrentCache`'s class javadoc now states what the invariant does *not* cover:

> The serialization is in-instance only: every owner (`Parsing`, `Transpiling`, `Linting`) keeps its own guard, and `Parsing` alone is instantiated both by `MjParse` and by `MjSafe#assembling()`. Within one module that is enough, because those mojos run in different lifecycle phases and never overlap. Two modules built at the same time (a parallel reactor, or two separate Maven processes) still write to the same cache location whenever they touch the same dependency, because the default cache directory is machine-wide rather than module-scoped (#2857); no in-process lock can serialize that, so it stays out of scope here.

The reasoning for documenting instead of fixing: threading the guard down from a single owner can't work, because `MjParse` and `MjAssemble`/`MjCompile` are separate mojo objects with no common owner to thread it from. That leaves a static lock map, which would serialize `mvn -T` within one JVM but does nothing for two separate `mvn` processes sharing `~/.eo` — the same #2857 scenario, and the one CI hits most. So it buys a partial fix in exchange for permanent static mutable state in a codebase that otherwise has none (`Catalogs.INSTANCE` is the only precedent, and it's immutable). Documenting the boundary honestly seemed the better trade for a PR scoped to #5720.

Entirely your call, though — if you'd rather have the reactor case covered, say so and I'll add the static map here.

### 2. The unguarded WPA write in `Linting.lintAll()`

Fixed — it goes through `this.guard` now, so all four `Cache` call sites in the class are consistent:

```java
this.guard.apply(
    this.sourcesDir, target, wpa,
    new Cache(
        this.cacheDir.resolve(Linting.CACHE),
        root -> { ... },
        p -> p.getFileName().toString().endsWith(".xmir") && !p.getFileName().equals(wpa)
    )
);
```

Your read of it was right — it's called once per `lint()`, outside the per-file `Threaded` loop, so it isn't racing today. But it's a write to a cache tail path like any other, and "safe because of where it happens to be called from" is exactly the kind of thing that stops being true after a refactor. Routing it through the guard costs nothing (uncontended `ReentrantLock`) and removes the need for a comment explaining the exception.

### 3. A test that actually catches the bug

Added `ParsingTest.parsesOnlyOnceWhenThreadsRaceOnTheSameTail`: eight threads through **one** `Parsing` instance on the identical tail path, asserting the parse pipeline is invoked exactly once. Details and two caveats are in the [inline reply](https://github.com/objectionary/eo/pull/5766#discussion_r3648683865) — the short version is that it's deterministic on the fix, and reverting `Parsing` alone to `new ConcurrentCache().apply(...)` fails it 3 out of 3 runs. It covers `Parsing` only, and it deliberately tolerates a *separate* race in the post-guard target read that this PR doesn't fix; both are explained in that reply.

### Verification

- Full `eo-maven-plugin` suite: 416 tests, 0 failures, 5 skipped.
- Qulice: 61 violations, byte-identical to the pre-change baseline on this branch (the 3 in `Transpiling` are pre-existing). Making `Parsing.parsed` package-private tripped `MethodsOrderCheck`, so it moved above the private overloads.
- Javadoc builds clean — no new warnings from the touched classes.

eo-maven-plugin/src/test/java/org/eolang/maven/ConcurrentCacheTest.java
anuragpaul602-netizen: Opened it as #5873, and `ParsingTest` now names it: the `catch` carries a `Logger.debug(..., "#5873")` and the method javadoc says why it's there, so it doesn't read as a swallowed failure. It can go away once #5873 is fixed.

The issue proposes making `Saved` write atomically (stream into a sibling temp file, then `Files.move(..., ATOMIC_MOVE)`) rather than widening the guarded region in each of the three classes — the invariant belongs to whoever writes the file, not to every caller who reads it back, and it also covers readers other than these three. Pushed as a993353.

morphqdd (changes requested):
Re-reviewed after the update. Checked the diff against each of your six points, not just the description:

- `apply()` parameter order: fixed, `cache` moved to last, matches `Cache.apply`'s own order now.
- Field javadoc duplication: fixed, all three now point to `{@link ConcurrentCache}` instead of repeating the explanation.
- WPA cache write in `Linting.lintAll()`: fixed, routed through `this.guard` like the other three call sites.
- The cross-instance gap (`MjSafe.assembling()` vs `MjParse`): not fixed, but honestly documented in `ConcurrentCache`'s class javadoc, with the reasoning laid out (a static map only partially helps and doesn't cover the real #2857 cross-process case, at the cost of introducing static mutable state this codebase otherwise avoids). I agree with that trade-off for a PR scoped to #5720.
- The missing regression test: fixed, `ParsingTest.parsesOnlyOnceWhenThreadsRaceOnTheSameTail` genuinely exercises the wiring (8 threads, one `Parsing` instance, same tail path, asserts the pipeline runs once), and you verified it actually fails against the old per-file-guard code (3/3 runs), not just passes on the new one. That's the right way to write a regression test.
- You also found and filed a real, separate issue (#5873) for the post-guard unguarded read, instead of quietly working around it. That's exactly the kind of thing worth surfacing rather than hiding behind a passing test.

Two things still open, both ones you flagged yourself as undecided rather than things I'm inventing:

1. You said `ParsingTest`'s new race test "covers `Parsing` only" and offered to add the same test for `Transpiling`/`Linting` if wanted. I'd like that: the bug this PR fixes was the same wiring mistake repeated three times, and right now only one of the three fixed classes has a regression test that would catch the mistake coming back. `Transpiling.transpiled` and `Linting.lintOne` have the identical shape you described, so the same test structure should translate directly.
2. For the documented-not-fixed cross-instance gap, could you open (or link, if #5873 already covers it, which I don't think it does since that's about a different race) a tracking issue referencing #2857 from the `ConcurrentCache` javadoc? Right now the javadoc explains the gap but there's nothing to click through to, so the next person who hits it in a parallel reactor build has to rediscover the reasoning from scratch.

Nothing here is about the core fix, which looks solid. Once the other two classes have the same regression test, or if you'd rather keep this PR scoped to `Parsing` and explicitly note the other two as follow-up, either is fine, just say which.

anuragpaul602-netizen:
Adding the two missing regression tests here would have taken this PR from ~194 to ~377 changed lines, past the 200-line budget, so I split them into #5923 as you offered. This PR stays scoped to the fix plus the `Parsing` test.

#5923 has `TranspilingTest` and `LintingTest`, both verified to fail against a per-call guard (8 vs 1) and pass against this fix, so all three guards are now covered. It is stacked on this branch, so its diff currently repeats these commits and will collapse to just the tests once this merges — worth reviewing this one first.

Two things in there that are consequences of this fix rather than of the tests, so flagging them here too: `Linting.lintOne` needed to take its lint transform as a parameter (the shape `Parsing.parsed` already has) because it had no other counting hook, and `Linting` picks up a `ClassFanOutComplexityCheck` annotation as a result. Details and reasoning are in #5923.

yegor256:
@morphqdd good to merge?

anuragpaul602-netizen:
You're right, and I've reverted it. Both `parsed` overloads are private again and `ParsingTest` is deleted.

What I'd missed is that the mistake was never a behavioural one worth racing threads at — it was **structural**: the guard was built per call instead of held as a field. So it's now checked structurally, in `ArchitectureTest`:

```java
ArchRuleDefinition.noClasses()
    .that().haveSimpleNameNotEndingWith("Test")
    .should().callConstructorWhere(ArchitectureTest.guardBuiltPerCall())
    .because("a per-call guard hands each thread an empty lock map (#5720)")
```

`ConcurrentCache` may only be constructed from a constructor. I checked it actually bites — inlining `new ConcurrentCache()` back into `Parsing.parsed` fails with:

```
Method <org.eolang.maven.Parsing.parsed(...)> calls constructor
<org.eolang.maven.ConcurrentCache.<init>()> in (Parsing.java:201)
```

The lock's own behaviour stays covered by `ConcurrentCacheTest`, which races 100 threads over one guard — that object is designed to be tested from outside, so nothing needs exposing.

This is strictly better than what it replaces, and not just because it exposes nothing: @morphqdd's concern on the earlier round was that the same wiring mistake appears in three classes with only one regression test. One rule now covers `Parsing`, `Transpiling`, `Linting` and any call site added later. The PR also drops from 194 to 160 lines.

Knock-on: I had a follow-up PR #5923 adding the same style of race test for `Transpiling` and `Linting`, which needed the same exposure plus a signature change on `Linting.lintOne`. That's obsolete now — this rule covers both — so I'll close it unless you'd rather I keep any part.

464 tests pass, qulice is clean.

anuragpaul602-netizen:
Also — sorry for the wrong turn there. Exposing internals to make the object testable was a bad call, and it took your comment for me to see that the invariant was structural all along rather than something to race threads at. Appreciate the pushback.

morphqdd:
@yegor256, yes, good to merge.

Both remaining points from my last review are addressed, just in a better way than I originally asked for:

1. The Transpiling/Linting regression test I wanted is now covered structurally instead: `ArchitectureTest.buildsCacheGuardOnlyOnce` asserts no class builds a `ConcurrentCache` anywhere except inside a constructor, for every class in the plugin, not just the three I named. That catches the #5720 mistake pattern coming back anywhere in the codebase, which is stronger than three separate race tests would have been.
2. The cross-instance gap now links `#2857` directly from the `ConcurrentCache` class javadoc, so the next person who hits it has the issue number right there instead of having to rediscover the reasoning.

Nothing else outstanding from my side.

anuragpaul602-netizen:
I want to work on this project. Can you add me as a contributor?

0crat:
@anuragpaul602-netizen Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (160 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+36](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (9) comments.  Your running score is [+2249](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5767, opened 2026-07-22
fix(#5721): keep arguments applied to data literals in StUnhex

### LINKED ISSUE #5721: `StUnhex` in eo-printer loses arguments when applied directly to data literals

anuragpaul602-netizen:
Closes #5721.

## Problem

`StUnhex` turns a data literal's hex payload (`Φ.bytes`, `Φ.number`, `Φ.string`) into readable EO text by calling `setTextContent(...)` on the literal's wrapper `<o>`. `Node.setTextContent()` deletes **all** children of the element.

A bare data literal can legitimately be the head of an application (e.g. `42 5 > x` is valid, parser-accepted EO), in which case the wrapper `<o>` has a second child — the applied argument — alongside its own byte payload. `setTextContent` wiped that argument out, so:

- `42 5 > x` printed as `42 > x`
- `"str" 5 > x` printed as `"str" > x`
- `01-02 5 > x` printed as `01-02 > x`

This is real data loss through any tool built on `Xmir.toEO()` (`eoc pretty`, `eo:format`, …).

## Fix

Replace only the **byte-payload child** (the literal's first `<o>` child) with a text node holding the readable form, instead of overwriting the whole wrapper. Any sibling argument is left intact, so the full application still prints. The single-child case is unchanged.

## Tests

`StUnhexTest` gains cases that apply an argument directly to a number, string, and bytes literal head and assert the argument survives. Existing cases still pass.


anuragpaul602-netizen:
Ready for review. The root cause is that `setTextContent()` on the literal's wrapper `<o>` deletes all children, so an argument applied directly to a data literal head (`42 5`) was silently dropped. This PR replaces only the byte-payload child (the first `<o>`) with a text node, leaving any sibling argument intact; the single-child case is unchanged.

Verified locally: `StUnhexTest` passes, including new cases that apply an argument to a number, string, and bytes literal head.

anuragpaul602-netizen:
Pushed a fix for the CI failure. The first cut used `replaceChild` unconditionally, which left the surrounding whitespace text nodes in pretty-printed XMIR — so the flattened `Φ.bytes` text picked up newlines/indentation and later failed to re-parse as hex (`NumberFormatException` in `MjPrintTest`).

Now `detach` keeps the original `setTextContent` behaviour for the common single-child case (which also clears whitespace, so it re-reads cleanly) and only falls back to the child-preserving `replaceChild` when a second child element — an applied argument — is actually present. Added a regression test over indented XMIR.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -84,6 +85,52 @@ final class StUnhex extends StEnvelope {
...
+     *
+     * @param xnav The data literal's wrapper element
+     * @param text The readable text to put in place of the byte payload
+     */
+    private static void detach(final Xnav xnav, final String text) {
+        final Node node = xnav.node();
+        if (StUnhex.only(node)) {
+            node.setTextContent(text);
+        } else {
+            node.replaceChild(
```
yegor256: @anuragpaul602-netizen it's better to use xnav _only_, and never touch the DOM directly. If xnav features are not powerful enough, file a feature request to them.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -84,6 +85,52 @@ final class StUnhex extends StEnvelope {
...
+        }
+    }
+
+    /**
+     * Whether the byte payload is the element's only child element (i.e. no
+     * argument is applied to the literal).
+     * @param node The wrapper element
+     * @return True if there is at most one child element
+     */
+    private static boolean only(final Node node) {
```
morphqdd: This counts only `ELEMENT_NODE` children, explicitly skipping text nodes, which only makes sense if the author expected non-element children (whitespace text nodes from pretty-printed/indented XMIR, in particular) to be possible alongside the byte payload and the argument. That expectation is correct: XMIR coming out of earlier XSL shifts is commonly indented, with a whitespace-only text node between sibling `<o>` elements.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
morphqdd: But `replaceChild` here only swaps the byte-payload `<o>` for the new text node, it doesn't touch any other children. If the real input has a whitespace text node between the payload and the argument (which `only()`, per the comment above, seems to anticipate), that whitespace node survives the replace and ends up sitting between the new readable-text node and the argument element, e.g. `42  <o base='Φ.sigma'/>` retains its original indentation whitespace instead of the clean, single-space-or-none layout the single-child branch produces via `setTextContent` (which clears every child, whitespace included, before setting the one text node). The two branches disagree on whether surrounding whitespace should survive, and only one of them was actually tested.

eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
```diff
@@ -198,6 +198,105 @@ void keepsMalformedBytesUnconverted(final Shift shift, final String type) {
         );
     }
```
morphqdd, line 200: All three new "keeps argument" tests build their input XML with `String.join("", ...)`, no whitespace between the sibling elements at all. `convertsNumberFromIndentedXmir` above uses `System.lineSeparator()` to build an indented fixture, but that one only exercises the single-child (`only()==true`) branch, the one that already worked before this PR. None of the new tests combine "indented, pretty-printed input" with "argument applied to the literal", which is exactly the combination `replaceChild` doesn't fully handle (see the two comments above). Since this fix's whole motivation is real, pretty-printed `Xmir.toEO()` output, that combination is the realistic case, not the compact hand-built one.

morphqdd (changes requested):
The core fix direction is right: the old `setTextContent` call genuinely deleted sibling arguments, and only touching the byte-payload child instead of the whole wrapper is the correct approach. One real gap in the new branch, verified by reading `only()` and `detach()` together rather than in isolation.

anuragpaul602-netizen:
Thanks — you were right, and it was observable, so I went and reproduced it rather than reasoning about it.

## The whitespace claim: confirmed

Parsing indented XMIR (what `MjPrint` reads back from a `.xmir`) with an argument applied, and dumping the **live DOM** after the shift — no re-serialisation in between, since that hides the text nodes:

```
before:  text('\n      ') element(Φ.bytes) text('\n      ') element(Φ.number) text('\n    ')
after:   text('\n      ') text('42') text('\n      ') element(Φ.number) text('\n    ')
```

Three text nodes survive where the `setTextContent` branch leaves exactly one. The two branches did disagree on whitespace, exactly as you described, and only the compact case was tested.

One correction on the consequence: it never reached the printed EO. `eo:read-data()` is `replace(string-join($o/text(),''), '^\s+|\s+$', '')`, and every stray node is leading or trailing, so it trims away — `toEO` output was already correct for the number and string cases. The defect was real but confined to the XMIR tree.

## Fix

Rather than teaching the `else` branch to clean up after itself, I collapsed both branches into one path: drop every text child, then `replaceChild` the payload. For a single child that is byte-for-byte what `setTextContent` did; for a literal with an argument it now produces the same tree as compact input. `only()` is gone (−14 lines), so there is no second branch left to drift.

Added `keepsArgumentAppliedToIndentedLiteral` — indented **and** an argument applied, the combination that was missing. It fails against the previous `detach` and passes against this one, so it is a real regression test.

## Also: the branch was 89 commits stale

Worth flagging since it was not visible from the diff. The full `eo-printer` suite at the old branch head was **220 tests, 37 failures**; `master` was green, so those were staleness, not this PR. Merged `master` (clean) — now **221 tests, 0 failures**, and `-Pqulice verify` passes.

## A separate bug this uncovered — #5912

Going end-to-end through `EoSyntax → toEO`, this PR fixes number and string, but `Φ.bytes` is still wrong — differently:

| program | master | this PR |
| --- | --- | --- |
| `42 5 > @` | `42 > [] > app` (arg lost) | correct |
| `"AB" 5 > @` | `"AB" > [] > app` (arg lost) | correct |
| `01-02 5 > @` | `01-02 > [] > app` (arg lost) | `5 > [] > app` (**data lost**) |

`StUnhex` is not at fault: the XMIR after the shift correctly carries both, and still does after `dataized-to-const`. The loss happens in `to-eo-tree.xsl`, on the path that inlines `φ` into the enclosing line — which only the bytes case takes here. Filed as #5912 rather than widening this PR.

Note that `keepsArgumentAppliedToBytesLiteral` passes throughout, because it asserts the post-shift XMIR and never the printed EO. So that test is weaker than it looks, and #5912 suggests a print-pack as the proper regression test.

yegor256:
@anuragpaul602-netizen would be great to reproduce this problem with a YAML-based test in eo-printer module, better in the `print-packs/yaml/` directory.

anuragpaul602-netizen:
Added in `print-packs/yaml/argument-applied-to-literal.yaml`.

I checked that it genuinely reproduces the bug rather than just passing: against `master`'s `StUnhex` the pack fails with

```
[] > main
  42 > x
  "str" > y
```

— both applied arguments gone, which is exactly the data loss #5721 describes. With the fix they survive.

One thing worth knowing before reading the expected output: the printed form is not the original `42 5 > x`. A literal's own bytes occupy `α0`, so the applied argument is `α1`, and the printer binds it explicitly:

```
[] > main
  42 > x
    5:1
  "str" > y
    5:1
```

That is semantically faithful and round-trips, so the pack passes `printsToParseableEo` too. Writing the argument bare on the next line would bind it to `α0` and silently mean something else, so I did not try to force the compact form.

The pack covers the number and string cases. I deliberately left out `01-02 5 > x`: that one is still printed wrong even with this fix (it comes out as `5 > [] > app`, losing the data instead of the argument), because the loss happens later in `to-eo-tree.xsl` on the `φ`-inlining path rather than in `StUnhex`. I filed that separately as #5912 rather than widen this PR.

Kept the pack to 23 lines so the PR lands at 199 of the 200-line budget. Also merged the latest `master` — 274 tests pass in `eo-printer` and qulice is clean.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -84,6 +85,41 @@ final class StUnhex extends StEnvelope {
...
+    private static void detach(final Xnav xnav, final String text) {
+        final Node node = xnav.node();
+        final NodeList children = node.getChildNodes();
+        for (int idx = children.getLength() - 1; idx >= 0; --idx) {
+            final Node child = children.item(idx);
+            if (child.getNodeType() == Node.TEXT_NODE) {
+                node.removeChild(child);
+            }
+        }
+        node.replaceChild(
```
yegor256, line 110: @anuragpaul602-netizen can't we use XNAV mechanisms for this? it's better to stay away from bare DOM manipulations - they are unreliable

anuragpaul602-netizen:
Half of it, yes — the loop is gone. The swap itself can't be, and I want to show why rather than just assert it.

**Xnav has no mutation API.** Its entire public surface is `element`, `elements`, `strict`, `one`, `attribute`, `path`, `copy`, `text`, `node` — every one of them read-only. The only way out is `node()`, which is why the pre-existing code already called `node().setTextContent(...)`; this PR didn't introduce DOM manipulation here, it changed which DOM calls happen.

**Xembly can't express the swap either**, and I checked both directives that looked plausible against a literal with an applied argument:

| directive | result |
|---|---|
| `set("42")` | `text('42')` — **wipes the argument**; SET *is* `setTextContent`, so this reintroduces #5721 |
| `cdata("42")` | appends, keeps siblings — but emits `<![CDATA[42]]>` |

So the only pure-Xembly route would rewrite every unhexed literal in every printed XMIR as `<o base="Φ.number"><![CDATA[42]]></o>`, and put the text after the argument rather than in the payload's position. That seemed a worse trade than one `replaceChild`.

**What did move to Xembly** is the part you were reacting to — the hand-rolled reverse `NodeList` loop:

```java
new Xembler(new Directives().xpath("text()").remove()).applyQuietly(node);
```

I verified `xpath("text()")` on input with no text children is a no-op rather than an error, so it covers compact and pretty-printed XMIR alike. Output is byte-identical, `NodeList` is no longer imported, and `xembly` is now declared in `eo-printer/pom.xml` rather than arriving transitively.

The javadoc records the reasoning so the next person doesn't retry `set()` and silently resurrect the bug.

274 tests pass, qulice is clean, and the reproduction still holds — against master's `StUnhex` the new print-pack and all four argument tests fail. PR is at 197 of the 200-line budget.

If you'd still prefer no bare DOM at all, the honest options are the CDATA form above or moving the unhexing into XSL; both are bigger than this PR and I'd rather do either as a follow-up than smuggle it in here.

eo-printer/pom.xml
```diff
@@ -30,6 +30,11 @@
       <artifactId>xnav</artifactId>
       <!-- version from parent POM -->
     </dependency>
+    <dependency>
+      <groupId>com.jcabi.incubator</groupId>
+      <artifactId>xembly</artifactId>
```
Favixx, line 35: This buys one directive — `xpath("text()").remove()` — at the price of a compile-scope dependency on `com.jcabi.incubator:xembly` for every consumer of `eo-printer` (the version is managed in the root POM at 0.32.2, so the build is fine; the cost is the coupling, not the wiring).

What makes it awkward is that it does not buy what @yegor256 asked for. His comment was about staying off bare DOM, and the DOM is still there two lines later in `replaceChild(...)` and `createTextNode(...)`. So the module ends up with an extra library *and* the manipulation the library was supposed to replace.

Two coherent endings: drop Xembly and remove the text children with the DOM you are already holding (`while (node.getFirstChild() != null && node.getFirstChild().getNodeType() == Node.TEXT_NODE)` — three lines, no new import), or keep Xembly and find a way to place the text node with it so the DOM calls disappear. Right now it is the union of both costs.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -84,6 +86,33 @@ final class StUnhex extends StEnvelope {
...
+     * payload and the argument are separated by whitespace-only text nodes,
+     * and leaving them in would strand the readable text between blobs of the
+     * original indentation. The swap itself stays on the DOM because neither
+     * Xnav (which only navigates) nor Xembly can put a plain text node next
+     * to sibling elements — Xembly's only appending directive emits CDATA.</p>
+     *
+     * @param xnav The data literal's wrapper element
+     * @param text The readable text to put in place of the byte payload
+     */
+    private static void detach(final Xnav xnav, final String text) {
```
Favixx, line 107: The name says the opposite of what happens. Nothing is detached and nothing is returned — the method drops stray text children and puts a text node where the byte payload used to be. `unhex`, `inline` or `replacePayload` all describe that; `detach` sends the reader looking for a removed subtree.

It matters more than usual here because the three call sites read as `StUnhex.detach(xnav, StUnhex.number(number))`, where the second argument is plainly the text being written in — so at the call site the verb is already contradicted by its own arguments.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -84,6 +86,33 @@ final class StUnhex extends StEnvelope {
...
 
+    /**
+     * Put the readable text of a data literal in place of its byte payload.
+     *
+     * <p>Only the byte-payload child (the first {@code <o>}) is swapped, so
+     * that a sibling argument applied to the literal, as in {@code 42 5},
+     * survives. {@code setTextContent} (which is also what Xembly's SET does)
+     * would delete that argument and silently drop the application.</p>
+     *
+     * <p>Xembly first drops every text child: in pretty-printed XMIR the
```
Favixx, line 97: This paragraph is an argument about what two libraries can do at their current versions, frozen into permanent documentation: "neither Xnav (which only navigates) nor Xembly can put a plain text node next to sibling elements — Xembly's only appending directive emits CDATA". If Xembly grows a directive for it, or Xnav gains mutation, nothing will fail and nothing will point here.

The repository has a mechanism for exactly this shape of statement — a `@todo #NNNN:30min` puzzle that PDD turns into an issue and that stops compiling once the number is closed. The rationale for the DOM call belongs there (or in the PR thread, where @yegor256 asked the question), and the Javadoc keeps the first paragraph, which describes what the method does rather than what the alternatives cannot.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -84,6 +86,33 @@ final class StUnhex extends StEnvelope {
...
+     * payload and the argument are separated by whitespace-only text nodes,
+     * and leaving them in would strand the readable text between blobs of the
+     * original indentation. The swap itself stays on the DOM because neither
+     * Xnav (which only navigates) nor Xembly can put a plain text node next
+     * to sibling elements — Xembly's only appending directive emits CDATA.</p>
+     *
+     * @param xnav The data literal's wrapper element
```
Favixx, line 104: `@param xnav The data literal's wrapper element` — it is an `Xnav`, a navigator positioned at that element, not the element. The distinction is what makes the body work: `xnav.node()` and `xnav.element("o")` are two different nodes reached from the same navigator, and a reader who takes the doc literally will expect `xnav` and `node` to be the same thing.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -84,6 +86,33 @@ final class StUnhex extends StEnvelope {
...
+     * @param text The readable text to put in place of the byte payload
+     */
+    private static void detach(final Xnav xnav, final String text) {
+        final Node node = xnav.node();
+        new Xembler(new Directives().xpath("text()").remove()).applyQuietly(node);
+        node.replaceChild(
```
Favixx, line 110: `xnav.element("o").node()` is unconditional, and it is safe only because all three selectors above demand a payload: `BYTES` requires `o[1][string-length(normalize-space(text()))>0]` and `elements(type)` requires `o[1][@base='Φ.bytes' and string-length(...)>0]`. I checked, and there is no reachable path into `detach` with no `<o>` child today.

But that invariant now lives in three XPath strings forty lines away from the code that depends on it, and the old body did not depend on it at all — `xnav.element("o").text().orElse("")` degraded to an empty string. Relaxing any of those selectors later (dropping the `string-length` guard, say, to also normalise empty payloads) turns this line into a crash inside a printer. One sentence in the Javadoc — "callers must select only literals whose first `<o>` child is the byte payload" — records the contract where the next person will look.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -84,6 +86,33 @@ final class StUnhex extends StEnvelope {
...
+     * original indentation. The swap itself stays on the DOM because neither
+     * Xnav (which only navigates) nor Xembly can put a plain text node next
+     * to sibling elements — Xembly's only appending directive emits CDATA.</p>
+     *
+     * @param xnav The data literal's wrapper element
+     * @param text The readable text to put in place of the byte payload
+     */
+    private static void detach(final Xnav xnav, final String text) {
+        final Node node = xnav.node();
+        new Xembler(new Directives().xpath("text()").remove()).applyQuietly(node);
```
Favixx, line 109: `new Directives()` and `new Xembler(...)` are constructed per call, so the `text()` XPath is rebuilt and recompiled once per data literal. `Xmir.toEO()` runs this shift over an entire program, and the overwhelmingly common literal has a single child and no stray text at all — every one of them now pays for the cleanup that exists for the rare literal with an applied argument.

`private static final Directives STRAY = new Directives().xpath("text()").remove();` hoisted next to `BYTES` costs one line and makes the per-literal work a cursor walk instead of an XPath compile. (`Directives` is not thread-safe as a builder, but it is only read here, and `Xembler` is cheap to wrap around it.)

eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
```diff
@@ -198,6 +198,134 @@ void keepsMalformedBytesUnconverted(final Shift shift, final String type) {
...
+                        System.lineSeparator(),
+                        "<o base='Φ.number'>",
+                        "  <o base='Φ.bytes'>",
+                        "    <o>40-45-00-00-00-00-00-00</o>",
+                        "  </o>",
+                        "</o>"
+                    )
+                )
+            ),
+            XhtmlMatchers.hasXPath("//o[@base='Φ.number' and normalize-space(text())='42']")
```
Favixx, line 221: This is the test for the indented case, and `normalize-space(text())='42'` is the one matcher that cannot see what the indented case is about. Delete the `Xembler(...text()...remove())` line from `detach` and this test still passes: the whitespace text nodes survive the `replaceChild`, `text()` returns `"\n  "` first, and `normalize-space` folds the difference away.

With the current code the element ends up with exactly one child, the text `42` and nothing else — I traced the three children (`text`, `<o base='Φ.bytes'>`, `text`) through the removal and the swap. So `text()='42'`, matching the other four new tests, is both true and strictly stronger, and it is what pins the line of Xembly this PR added.

eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
```diff
@@ -198,6 +198,134 @@ void keepsMalformedBytesUnconverted(final Shift shift, final String type) {
...
+                        "      <o>40-45-00-00-00-00-00-00</o>",
+                        "    </o>",
+                        "    <o base='Φ.sigma'/>",
+                        "  </o>",
+                        "</p>"
+                    )
+                )
+            ),
+            XhtmlMatchers.hasXPath(
+                "//o[@base='Φ.number' and text()='42' and o[@base='Φ.sigma']]"
```
Favixx, line 249: `text()='42'` is an existential test: it holds if *any* text child equals `42`, so it would still pass with the original indentation left stranded around the readable text — which is precisely the failure mode @morphqdd described and the reason the removal was added.

`count(text())=1 and text()='42'` states the invariant the fix actually establishes: one text node, holding the readable form, sitting next to the preserved argument. Same for the three compact variants below.

eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
```diff
@@ -198,6 +198,134 @@ void keepsMalformedBytesUnconverted(final Shift shift, final String type) {
...
+                    )
+                )
+            ),
+            XhtmlMatchers.hasXPath("//o[@base='Φ.number' and normalize-space(text())='42']")
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("shifts")
+    void keepsArgumentAppliedToIndentedLiteral(final Shift shift, final String type) {
```
Favixx, line 227: Small consistency point across the five new fixtures: this one and the three below wrap the literal in `<p>`, while `convertsNumberFromIndentedXmir` above uses a bare `<o>` as the root. `<p>` appears nowhere in XMIR — it is here only because a document needs one root element — so the fixtures that exercise the interesting case are also the ones furthest from the shape the shift meets in production, where the root is `<object>` and literals sit under it.

Using `<object>` in all five costs nothing and keeps the fixtures readable as XMIR rather than as XML.

eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
```diff
@@ -198,6 +198,134 @@ void keepsMalformedBytesUnconverted(final Shift shift, final String type) {
...
+            ),
+            XhtmlMatchers.hasXPath(
+                "//o[@base='Φ.number' and text()='42' and o[@base='Φ.sigma']]"
+            )
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("shifts")
+    void keepsArgumentAppliedToNumberLiteral(final Shift shift, final String type) {
```
Favixx, line 256: These three tests are one test written three times: same structure, same assertion shape, differing only in base (`Φ.number`/`Φ.string`/`Φ.bytes`), payload (`40-45-…`/`41-42`/`01-02`) and expected text (`42`/`"AB"`/`01-02`). That is roughly sixty lines of duplication in a file that already parameterises over `shifts()`.

A `@MethodSource` yielding those triples — crossed with `shifts()` as the existing tests do — collapses them into one test and makes adding the fourth literal type a one-line change instead of a copy-paste.

eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
```diff
@@ -198,6 +198,134 @@ void keepsMalformedBytesUnconverted(final Shift shift, final String type) {
...
+            ),
+            XhtmlMatchers.hasXPath(
+                "//o[@base='Φ.number' and text()='42' and o[@base='Φ.sigma']]"
+            )
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("shifts")
+    void keepsArgumentAppliedToStringLiteral(final Shift shift, final String type) {
```
Favixx, line 281: Every one of the new tests uses `<o base='Φ.sigma'/>` as the applied argument, which is the one kind of argument that this shift will never touch. The case in the PR title — `42 5` — has a *data literal* as the argument, so the same `StSequence` rewrites both the head and the argument, and the argument's own `<o base='Φ.bytes'>` child is replaced while its parent is being replaced too.

That combination is exercised only by the new YAML pack, end to end, where a failure surfaces as a diff of printed EO. A unit fixture with `<o base='Φ.number'><o base='Φ.bytes'><o>40-49-…</o></o></o>` as the argument would pin it at the level where a failure names the method.

Favixx (commented):
Read the current head, not the version @morphqdd reviewed: the whitespace gap he found is closed — `detach` now drops the stray text children before the swap — and `keepsArgumentAppliedToIndentedLiteral` was added for it. I checked the two things that worried me most about the new shape and both are fine, so I am recording them rather than raising them: the three selectors all require `o[1]` with non-empty text, so `xnav.element("o").node()` inside `detach` cannot be reached with a missing payload; and the printed `5:1` form is safe because `XmirTest.printsToParseableEo` re-parses every pack's `printed:` block and asserts it prints back to itself, so the new pack is covered by a round-trip test the moment it lands.

What I would still like resolved:

**The dependency.** @yegor256 asked twice to stay on Xnav and away from bare DOM. The current answer adds a *third* XML library to `eo-printer` — Xembly, compile-scope, published — uses it for exactly one directive, and keeps the `replaceChild`/`createTextNode` pair that the request was about. Whichever way this goes, it should be a decision rather than a middle: either the swap moves off the DOM too, or the text-node cleanup goes back to three lines of DOM and the dependency comes out.

**The Xembly line is not covered by the test that exists to cover it.** `convertsNumberFromIndentedXmir` asserts `normalize-space(text())='42'`, which stays green if the whitespace removal is deleted. Details inline.

**Coverage of the branch where nothing is converted.** A malformed or NaN literal skips `detach` entirely, and no test asks whether an argument applied to *that* survives — the exact question this PR exists to answer.

The rest is test structure and documentation.

eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
```diff
@@ -198,6 +198,134 @@ void keepsMalformedBytesUnconverted(final Shift shift, final String type) {
...
+            ),
+            XhtmlMatchers.hasXPath(
+                "//o[@base='Φ.string' and text()='\"AB\"' and o[@base='Φ.sigma']]"
+            )
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("shifts")
+    void keepsArgumentAppliedToBytesLiteral(final Shift shift, final String type) {
```
Favixx, line 306: With this one the new tests cover all three literal types, well-formed and with an argument, and `keepsMalformedBytesUnconverted` above covers a malformed payload with no argument. The fourth square of that grid is empty: a literal that is *not* converted — a malformed payload, or a NaN/infinite number, which the guard in the `number` shift skips before ever calling `detach` — that also carries an applied argument.

It is worth one case, because that path reaches the argument by a different route: the BYTES shift has already rewritten the inner payload to text, and then the `number` shift declines to touch the wrapper at all, so the argument survives (or not) without `detach` ever running. Nothing currently asserts that it is still there afterwards, and "the argument survives" is the entire claim of this PR.

eo-printer/src/test/resources/org/eolang/printer/print-packs/yaml/argument-applied-to-literal.yaml
```diff
@@ -0,0 +1,23 @@
...
+penalties:
+  INDENT: 3
+  BRACKET: 7
+  EXCESS: 1
+  WIDTH: 80
+  STEP: 2
+  SPACE: 7
+origin: |
+  [] > main
+    42 5 > x
```
Favixx, line 15: The PR description lists three broken cases — `42 5 > x`, `"str" 5 > y` and `01-02 5 > x` — and the unit tests cover all three, but this pack covers only the first two. The bytes literal is the one where the head and the payload are the same element rather than nested (the `BYTES` shift fires directly on the wrapper the argument hangs off), so it is the one whose end-to-end printing is least like the others.

One more line in `origin` and its counterpart in `printed` closes the gap and makes the pack match the description.

eo-printer/src/test/resources/org/eolang/printer/print-packs/yaml/argument-applied-to-literal.yaml
```diff
@@ -0,0 +1,23 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# A literal's own bytes are its a0, so an applied argument lands in a1 and
+# prints as an explicit "5:1". Before #5721 the argument was lost entirely.
```
Favixx, line 5: "Before #5721 the argument was lost entirely" — #5721 is the issue reporting the loss, so the loss existed *because of* it rather than before it; "before this fix" or "before #5767" is what is meant.

While editing the comment, it is worth recording why `5:1` is safe to expect here, because it is the surprising part of the fixture: `XmirTest.printsToParseableEo` re-parses every pack's `printed:` block and asserts that printing it again yields the same text, so this pack is automatically a round-trip test for that form. A reader who compares `origin` with `printed` and sees the surface syntax change will want to know that.

eo-printer/src/test/resources/org/eolang/printer/print-packs/yaml/argument-applied-to-literal.yaml
```diff
@@ -0,0 +1,23 @@
...
+penalties:
+  INDENT: 3
+  BRACKET: 7
+  EXCESS: 1
+  WIDTH: 80
+  STEP: 2
+  SPACE: 7
+origin: |
+  [] > main
```
morphqdd, line 14: The declarative pack test covers a number literal (42 5) and a string literal ("str" 5) with an applied argument, but not a bytes literal, even though StUnhexTest.keepsArgumentAppliedToBytesLiteral covers that case at the Java level. The bytes case is missing from the portable pack test.

morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +17 points for this: +12 as a basis; +5 for the 20 comment(s) that you have made.  Your running score is [+2828](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:
@anuragpaul602-netizen conflicts here

yegor256:

PR #5755, opened 2026-07-22
aded token fix inside qoutes

EmaniPreethika:

PR #5768, opened 2026-07-22
fix(#5722): make redundant-parens detection quote-aware in Tokens

### LINKED ISSUE #5722: `Tokens.readGroup` incorrectly flags single-token string literals in parentheses as redundant

anuragpaul602-netizen:
Closes #5722.

## Problem

`Tokens.readGroup()` decides whether a parenthesized group is "redundant around a single token" by scanning the raw substring between the parens for a space or for `(`, `)`, `[`, `]`. That scan has no notion of quoted-string content, so it is fooled by characters inside a string literal:

```
1.add ("hello")        -> redundant parentheses error
1.add ("hello world")  -> accepted silently
```

Both are the same case — a parenthesized single string-literal argument — but only the single-word one was flagged. Same asymmetry for `("a(b)c")` and `("a[b]c")`. (Same class of bug as the `Penalty.java` fix in #5718.)

## Fix

`readGroup()`'s single-token detection now tracks quote state (toggled on an unescaped `"`) while scanning, neutralizing spaces/parens/brackets that live inside a string literal. A parenthesized single string literal is therefore treated consistently — always redundant — regardless of its contents.

## Tests

- `TokensTest`: `("hello world")` and `("a(b)[c]")` are now rejected as redundant; `("a" "b")` (genuinely two tokens) is still accepted.
- New end-to-end typo-pack `redundant-parentheses/string-literal.yaml`.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Tokens.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Tokens.java
@@ -239,12 +239,7 @@ final class Tokens {
             );
         }
         final String inside = this.body.substring(start + 1, this.cursor - 1);
-        if (!inside.isEmpty()
-            && inside.indexOf(' ') < 0
-            && inside.indexOf('(') < 0
-            && inside.indexOf(')') < 0
-            && inside.indexOf('[') < 0
-            && inside.indexOf(']') < 0) {
+        if (Tokens.singleToken(inside)) {
             throw new ParseError(
                 this.span.line(), this.span.indent() + start,
                 String.format(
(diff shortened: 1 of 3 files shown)


anuragpaul602-netizen:
Closing this in favor of #5755, which was opened earlier for the same issue (#5722) and takes the same approach (treating anything inside `"…"` as a single token). No need for two competing PRs — deferring to #5755. Thanks @EmaniPreethika.

EmaniPreethika:
> ng this in favor of #5755, which was opened earlier for the same issue (#5722) and takes the same approach (treating anything inside `"…"` as a single token). No need for two competing PRs — deferri

do I close my PR @anuragpaul602-netizen ?

PR #5769, opened 2026-07-22
fix(#5723): make PhDefault.activate() thread-safe

### LINKED ISSUE #5723: `PhDefault` method `activate()` lacks proper thread-safety despite class documentation claim
## What Happens

`PhDefault.java`'s class doc (line 26) says "The class is thread-safe." Its lazy-init helper `activate()` contradicts that:

```java
private void activate() {
    if (this.attrs == null) {
        this.attrs = PhDefault.defaults();
        for (final Map.Entry<String, Attribute> ent : this.initial.entrySet()) {
            this.add(ent.getKey(), ent.getValue());
        }
    }
}
```

Every public method (`take`, `put`, `add`, `copy`, etc.) calls `this.activate()` first. The null-check-then-populate sequence is an unsynchronized double-checked-locking pattern: if two threads call any method on the same not-yet-activated `PhDefault` instance concurrently, both can pass `this.attrs == null`, both build and assign `this.attrs`/`this.order` to different `HashMap` instances, and interleave writes on plain (non-thread-safe) `HashMap`s.

Every other lazy-cache class in this same package (`AtOnce`, `PhOnce`, `AtVoid`, `AtRho`, `OnClasspath`) correctly guards this exact pattern with a `ReentrantLock`/`AtomicReference`/CAS. `PhDefault` is the only one that doesn't, despite explicitly claiming thread-safety.

This matters because `PhDefault`-derived instances genuinely get shared across threads: `PhNest`/`PhPackage` cache singleton package objects (for example `Φ.number`) and dispatch `.take()` directly onto the shared instance, not a copy.

Verified by compiling the class standalone and running a stress test: 8 threads racing `.take()` on a shared, freshly-constructed `PhDefault` across 2000 iterations. It threw a real `NullPointerException` from inside `take()` 17 times, confirming the race is real, not theoretical.

## What Should Happen

`activate()` should use the same guarded-locking pattern as `AtOnce`/`PhOnce` (a double-checked lock with a `ReentrantLock`, or build the maps into locals and publish via a single atomic/volatile write), so concurrent first access to a `PhDefault` instance cannot race. Alternatively, if this class is genuinely not meant to be shared across threads, the class doc should stop claiming thread-safety.

anuragpaul602-netizen:
Closes #5723.

## Problem

`PhDefault`'s class doc says "The class is thread-safe", but its lazy-init helper `activate()` was an unsynchronized double-checked-lock:

```java
private void activate() {
    if (this.attrs == null) {
        this.attrs = PhDefault.defaults();
        ...
    }
}
```

Every public method (`take`, `put`, `add`, `copy`, …) calls `activate()` first. Two threads racing the first call on a shared, not-yet-activated instance could both pass `attrs == null`, both build and publish rival `HashMap`s, and interleave writes on non-thread-safe maps — throwing NPEs from inside `take()`. `PhDefault`-derived instances genuinely get shared (e.g. `PhNest`/`PhPackage` cache singleton package objects and dispatch `.take()` on the shared instance).

Every other lazy-cache class in the package (`AtOnce`, `PhOnce`, `AtVoid`, `AtRho`, `OnClasspath`) already guards this pattern with a `ReentrantLock`/CAS; `PhDefault` was the only one that didn't.

## Fix

`activate()` now uses a proper double-checked lock:
- a `ReentrantLock` serializes the one-time init;
- the `order`/`attrs` maps are built into locals and published, with the now-`volatile` `attrs` field (the guard) written **last**, so a lock-free reader that sees `attrs != null` is guaranteed to see the fully-populated `order` map too.

This matches the approach the issue recommends and the pattern used by `AtOnce`.

## Tests

New `PhDefaultTest` case races 100 threads on the first `take()` of one shared, not-yet-activated instance and asserts they all observe the single, fully-built attribute. Existing tests still pass.


anuragpaul602-netizen:
Ready for review. `activate()` was an unsynchronized double-checked lock, so two threads racing the first call on a shared, not-yet-activated instance could publish rival `HashMap`s and NPE from `take()`. This PR makes it a proper DCL: a `ReentrantLock` serializes init, the maps are built into locals, and the now-`volatile` `attrs` guard is published last — the same pattern `AtOnce`/`PhOnce` already use.

Verified locally: `PhDefaultTest` passes, including a new 100-thread race on the first `take()` of one shared instance.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -15,6 +15,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.locks.ReentrantLock;
```
morphqdd, line 18: Fine, minimal new import.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -99,8 +100,20 @@ public class PhDefault implements Phi, Cloneable {
 
     /**
      * Attributes.
+     *
+     * <p>Written last during {@link #activate()} and read as the lazy-init
+     * guard, so it's {@code volatile}: a thread that sees it non-null is
+     * guaranteed to also see the fully-populated {@link #order} map published
+     * before it.</p>
+     */
+    private volatile Map<String, Attribute> attrs;
```
morphqdd, line 114: Confirmed this reasoning is correct: since `attrs` is the last thing written in `publish()` and every reader in this class calls `activate()` (directly or transitively) before touching `attrs`/`order`, the JMM happens-before edge from this volatile read/write is sufficient to make the non-volatile `order` field safe too, as long as `order` is written strictly before `attrs` -- which it is (see the comment on line 583).

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -99,8 +100,20 @@ public class PhDefault implements Phi, Cloneable {
...
+
+    /**
+     * Guards the one-time {@link #activate()} of this instance, so concurrent
+     * first access from several threads can't build and publish two rival
+     * attribute maps.
      */
-    private Map<String, Attribute> attrs;
+    private final ReentrantLock lock;
```
morphqdd: Real issue: `copy()` (unchanged by this PR, not shown in the diff) does `(PhDefault) this.clone()`, and `PhDefault` doesn't override `clone()`, so it's `Object.clone()`'s default shallow field copy. That means `copy.lock` ends up being the literal same `ReentrantLock` instance as `this.lock`, not a fresh one -- `copy()` reinitializes `attrs` explicitly after cloning but never does the same for `lock`. I confirmed this empirically with a small reflection-based test:

```java
final PhDefault original = new PhDefault(attributes);
original.take("foo");
final PhDefault copy = (PhDefault) original.copy();
final Field field = PhDefault.class.getDeclaredField("lock");
field.setAccessible(true);
assertThat(field.get(copy), not(sameInstance(field.get(original))));
```
This assertion fails -- both point at the exact same lock object.

In practice this is harmless today, because `copy()` always populates `copy.attrs` eagerly before returning, so `copy`'s own `activate()` fast path (`if (this.attrs == null)`) never becomes true and the shared lock is never actually acquired through the copy. But that's an accident of `copy()`'s current implementation, not something this field's own design guarantees -- a future change to `copy()` (e.g. making the copy's attrs lazy too) would silently reintroduce cross-instance lock contention between logically unrelated objects. Suggest reinitializing `copy.lock = new ReentrantLock()` in `copy()` alongside `attrs`, so "every instance owns its own activation lock" holds unconditionally.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -155,6 +168,7 @@ private PhDefault(
         this.fqn = forma;
         this.data = dta;
         this.initial = attributes;
+        this.lock = new ReentrantLock();
```
morphqdd, line 181: Correct, every instance gets its own fresh lock at construction time (this is what makes the `copy()` case on line 116 the only gap).

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -530,15 +544,44 @@ private String oname() {
...
     /**
      * Activate the lazy state: initialize attrs/order from the constructor-supplied
-     * map, wrapping each entry with {@link AtWithRho}. Idempotent.
+     * map, wrapping each entry with {@link AtWithRho}. Idempotent and thread-safe.
+     *
+     * <p>The maps are built into locals under {@link #lock} and only then
+     * published, with the {@code volatile} {@link #attrs} written last. This is
+     * a proper double-checked lock: the lock-free fast path reads {@code attrs}
+     * once, and a concurrent first access can neither observe a half-built map
+     * nor race a second thread into building a rival one.</p>
```
morphqdd, line 591: Confirmed this is a proper double-checked-locking pattern: the outer `if (this.attrs == null)` in `activate()` is the lock-free fast path, and `publish()`'s own inner `if (this.attrs == null)` (line 573) is the real re-check taken under the lock, so a thread that queues up behind the lock-holder correctly finds `attrs` already populated and no-ops.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -530,15 +544,44 @@ private String oname() {
...
+        }
+    }
+
+    /**
+     * Build the attribute maps from the constructor-supplied {@link #initial}
+     * map and publish them, with the {@code volatile} {@link #attrs} written
+     * last. Must be called while holding {@link #lock}; the inner null-check
+     * keeps it idempotent when a queued thread enters after the winner.
+     */
+    private void publish() {
```
morphqdd, line 610: Good -- this inner null-check is what keeps `activate()` idempotent under contention (a second thread that was blocked on the lock while the first thread was building `built`/`ordered` will see `attrs != null` here and just return without redoing the work).

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -530,15 +544,44 @@ private String oname() {
...
+            final Map<String, Attribute> built = PhDefault.defaults();
             for (final Map.Entry<String, Attribute> ent : this.initial.entrySet()) {
-                this.add(ent.getKey(), ent.getValue());
+                final String name = ent.getKey();
+                if (PhDefault.SORTABLE.matcher(name).matches()) {
+                    ordered.put(ordered.size(), name);
+                }
+                built.put(name, new AtWithRho(ent.getValue(), this));
             }
+            this.order = ordered;
```
morphqdd, line 621: This ordering -- `this.order = ordered;` before `this.attrs = built;` (volatile) -- is exactly what makes the non-volatile `order` field safe to read after the volatile check in `activate()`, per the JMM happens-before transitivity from a volatile write. Confirmed by tracing every call site in the file: none of them read `order` without going through `activate()` first.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
```diff
@@ -363,6 +365,25 @@ void createsDifferentPhiInParallel() {
         );
     }
 
+    @Test
+    @SuppressWarnings("PMD.UnnecessaryLocalRule")
+    void activatesLazilyOnceUnderConcurrentAccess() {
```
morphqdd, line 379: Reasonable regression test for the race: 100 threads hammering `.take()` on a shared, freshly-constructed instance. Worth knowing concurrency tests like this are inherently probabilistic -- this test can't guarantee it would fail reliably against the pre-fix code on every CI run, though the issue itself was verified against the old code with a dedicated stress test showing real NPEs, so the bug itself isn't in question.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
```diff
@@ -363,6 +365,25 @@ void createsDifferentPhiInParallel() {
...
+        final int threads = 100;
+        final Map<String, Attribute> attributes = new HashMap<>(1);
+        attributes.put("foo", new AtComposite(null, rho -> new Data.ToPhi(42L)));
+        final Phi shared = new PhDefault(attributes);
+        MatcherAssert.assertThat(
+            "concurrent first access must activate the object exactly once, with no half-built or lost attributes",
```
morphqdd, line 385: Good choice of assertion: collecting all 100 threads' results into a `Set` and asserting it contains exactly one value (`42.0`) is a clean way to express "every thread saw the same, fully-built result" without needing to inspect internal state directly.

morphqdd (commented):
Reviewed by checking out the branch, tracing every reader of `attrs`/`order` in the file, running the full `PhDefaultTest` suite (54/54 green, 1 pre-existing unrelated skip), and writing one throwaway reflection-based probe test of my own (not committed) to check a specific concern about `copy()`. Left a detailed inline note on that -- see the comment on the `lock` field declaration -- plus several confirmations of what I verified as correct.

anuragpaul602-netizen:
Confirmed and fixed. I reproduced it with your probe before touching anything:

```
original.lock = 1028566121
copy.lock     = 1028566121
same instance? true
```

and after the change the two differ. Your analysis of why it's currently harmless is exactly right — `copy()` populates `copy.attrs` eagerly, so the copy's `activate()` fast path never sees a null and never reaches the lock. Agreed that resting on that is wrong: it's a property of `copy()`'s present shape, not of the field.

One wrinkle: the fix as written doesn't compile. The field was `final`, and a final field can't be assigned on a clone. So it drops `final` — which puts it alongside `attrs` and `order`, already non-final for the same clone-related reason — and the javadoc now records why, so nobody restores `final` without silently re-sharing the lock.

I also put the assignment *before* `copy.attrs = map` rather than after:

```java
copy.lock = new ReentrantLock();
copy.attrs = map;
```

`attrs` is volatile, so that write is a release; any thread that observes the copy's new `attrs` is guaranteed to observe its new `lock` too. Assigning after would leave the lock write unordered with respect to the publication.

No test for it, deliberately. Since `copy()` populates `attrs` eagerly there's no observable behaviour to assert — that's the same reason it's harmless today — and a reflection test reaching into a private field is the kind of internals-poking @yegor256 rejected on #5766. The invariant lives in the javadoc instead. If you'd rather have the reflection test committed anyway, say so and I'll add it.

1018 Java tests pass and qulice is clean.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
```diff
@@ -372,6 +374,25 @@ void createsDifferentPhiInParallel() {
         );
     }
 
+    @Test
+    @SuppressWarnings("PMD.UnnecessaryLocalRule")
```
yegor256, line 378: @anuragpaul602-netizen fix the code and get rid of this warning

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -181,6 +200,7 @@ public final Phi copy() {
             for (final Map.Entry<String, Attribute> ent : this.attrs.entrySet()) {
                 map.put(ent.getKey(), ent.getValue().copy(copy));
             }
+            copy.lock = new ReentrantLock();
```
Favixx, line 203: This line fixes one half of a two-part aliasing bug and leaves the other half untouched.

`this.clone()` is `Object.clone()`, a shallow field copy, so immediately after it the copy shares *three* references with the original: `attrs`, `order` and `lock`. The next two lines replace `lock` (this line) and `attrs` (the line below). `order` is never reassigned, so a `PhDefault` and every copy ever taken of it share one `HashMap<Integer, String>` for the rest of their lives.

That is observable through two public methods:

```java
final Phi a = new PhDefault(attrs);   // order: {0 -> "foo"}
final Phi b = a.copy();               // b.order IS a.order
b.add("bar", attr);                   // order.put(1, "bar") — mutates BOTH
```

Now `a.order` has an entry at index 1 for an attribute that `a.attrs` does not contain. `attr(1)` on `a` passes the `containsKey` check and returns `"bar"`, and `vacancy(1)` then does `this.attrs.get("bar").vacant()` on a `null` — NPE from inside a positional write.

It predates this PR, but this is the line that establishes the pattern for fixing it, and `copy.order = new HashMap<>(this.order)` next to it costs nothing. If the reachability of `add()` on a copy is the reason it has never been hit, that is worth saying out loud, because nothing in the code says it.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -104,8 +105,25 @@ public class PhDefault implements Phi, Cloneable {
...
+     * Guards the one-time {@link #activate()} of this instance, so concurrent
+     * first access from several threads can't build and publish two rival
+     * attribute maps.
+     *
+     * <p>Not {@code final}: {@link #copy()} clones through
+     * {@link Object#clone()}, whose shallow copy would otherwise hand the copy
+     * the very same lock as the original, and two unrelated objects would
+     * contend on it. The copy gets a fresh one instead.</p>
+     */
+    private ReentrantLock lock;
```
Favixx, line 126: The field comment explains honestly why this cannot be `final` — `copy()` has to give the clone a fresh lock — and that trade is worth reopening, because non-final is what creates the risk on line 595.

The alternative that removes the field altogether is `synchronized (this)` in `activate()`. It needs no field, so there is nothing for `clone()` to alias and nothing for `copy()` to reset; monitors are per-object by construction, which is exactly the property being hand-rolled here. It also costs no allocation per instance (see line 181), and biased/thin-lock handling in HotSpot makes an uncontended monitor cheaper than a `ReentrantLock` acquisition.

The usual argument for `ReentrantLock` — timed or interruptible acquisition, fairness, condition variables — none of it is used here. If there is a reason to prefer it that I am not seeing, it belongs in this comment.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -562,15 +582,44 @@ private String oname() {
...
+     * published, with the {@code volatile} {@link #attrs} written last. This is
+     * a proper double-checked lock: the lock-free fast path reads {@code attrs}
+     * once, and a concurrent first access can neither observe a half-built map
+     * nor race a second thread into building a rival one.</p>
      */
     private void activate() {
         if (this.attrs == null) {
-            this.attrs = PhDefault.defaults();
-            this.order = new HashMap<>(0);
+            this.lock.lock();
```
Favixx, line 595: This dereferences a non-final field on the slow path, and the premise of the whole PR is that `PhDefault` instances are shared across threads without external synchronisation.

If such an instance is handed to another thread through a data race — which is precisely the scenario the description describes for `PhNest`/`PhPackage` caches — then the JMM permits that thread to see the object reference while `this.lock` is still `null`, because a non-final field written in the constructor carries no publication guarantee. The result is an NPE here, in a method that every public operation calls.

The old code had no such failure mode: it raced on `attrs` and produced the NPEs described in #5723, but it could not fail on a null lock. So this is a new, rarer failure mode traded for a common one — a good trade, but only if the lock reference itself is safely published. `final` would give exactly that guarantee via the final-field freeze, and `synchronized (this)` sidesteps it entirely (see line 126).

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -104,8 +105,25 @@ public class PhDefault implements Phi, Cloneable {
...
     /**
      * Attributes.
+     *
+     * <p>Written last during {@link #activate()} and read as the lazy-init
+     * guard, so it's {@code volatile}: a thread that sees it non-null is
+     * guaranteed to also see the fully-populated {@link #order} map published
+     * before it.</p>
      */
-    private Map<String, Attribute> attrs;
+    private volatile Map<String, Attribute> attrs;
```
Favixx, line 114: Worth measuring rather than assuming: this makes every public operation on every `PhDefault` perform a volatile read, because `take`, `put`, `add`, `copy` and the positional lookups all begin with `activate()`, and `activate()`'s fast path is a read of this field.

On x86 a volatile read emits no fence, so the runtime cost is near zero — but it is a compiler barrier: the JIT can no longer hoist the null check out of a loop or keep `attrs` in a register across calls, and `PhDefault` is the single hottest type in the runtime.

The repository already runs a `benchmark` job on every PR. If it shows no movement, that is the answer and it belongs in the description; if it does, the fix is to read the volatile once into a local at the top of `activate()` rather than twice across `activate()` and `publish()`.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -160,6 +178,7 @@ private PhDefault(
         this.fqn = forma;
         this.data = dta;
         this.initial = attributes;
+        this.lock = new ReentrantLock();
```
Favixx, line 181: Every `PhDefault` ever constructed now allocates a `ReentrantLock` — an extra object, its `Sync` inner instance, and the header for both — and the overwhelming majority of them are activated once, by one thread, and never contended at all.

Given how many of these the runtime creates while dataizing a program, that is a per-object cost paid for a race that occurs on a small number of shared instances. `synchronized (this)` costs nothing until first contention. If the lock object is kept, allocating it lazily is not an option (that is the same race one level down), so the choice is really between a per-instance field and a monitor.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -562,15 +582,44 @@ private String oname() {
...
+        }
+    }
+
+    /**
+     * Build the attribute maps from the constructor-supplied {@link #initial}
+     * map and publish them, with the {@code volatile} {@link #attrs} written
+     * last. Must be called while holding {@link #lock}; the inner null-check
+     * keeps it idempotent when a queued thread enters after the winner.
+     */
+    private void publish() {
```
Favixx, line 610: Recording a positive check, because this is the part of the refactor where a silent behaviour change would have been easiest: `publish()` inlines the body of `add()` instead of calling it, and the inlining is faithful.

`add()` is `activate()` (a no-op here — `attrs` is being built), the `SORTABLE` match writing `order.put(order.size(), name)`, and `attrs.put(name, new AtWithRho(attr, this))`. All three survive, in the same order, with the same `SORTABLE` pattern and the same `AtWithRho` wrapping, differing only in writing to locals. `defaults()` still seeds `built`, so entries whose names collide with a default still overwrite it exactly as before.

The one thing lost is that `add()` and `publish()` are now two copies of the same three lines. If `add()` ever grows a fourth, this copy will not.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -562,15 +582,44 @@ private String oname() {
...
+        if (this.attrs == null) {
```
Favixx, line 611: The javadoc above states "Must be called while holding {@link #lock}" and nothing checks it. Since correctness of the whole scheme depends on that precondition — the non-volatile write to `order` on line 621 is safe only because it happens under the lock and before the volatile write — it is worth one line that fails loudly if a future caller forgets:

```java
assert this.lock.isHeldByCurrentThread();
```

Assertions are enabled under Surefire by default, so the test suite would enforce it, and production pays nothing.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -562,15 +582,44 @@ private String oname() {
...
             for (final Map.Entry<String, Attribute> ent : this.initial.entrySet()) {
-                this.add(ent.getKey(), ent.getValue());
+                final String name = ent.getKey();
+                if (PhDefault.SORTABLE.matcher(name).matches()) {
+                    ordered.put(ordered.size(), name);
+                }
+                built.put(name, new AtWithRho(ent.getValue(), this));
             }
+            this.order = ordered;
+            this.attrs = built;
```
Favixx, line 622: After this write the object is fully published and thread-safe — and from that moment on it stops being thread-safe again, because `add()` and `put()` mutate `this.attrs` and `this.order` as plain `HashMap`s with no lock held.

That matters for what #5723 actually asks. The issue is about the class javadoc at line 27 claiming "The class is thread-safe"; this PR makes *initialization* safe and leaves *mutation* exactly as it was. Two threads calling `add()` on the same instance can still corrupt a `HashMap`, and the volatile `attrs` does nothing about it because the reference does not change.

So either the javadoc should be narrowed to what is now true ("lazy initialization is thread-safe; concurrent mutation is not"), or the remaining half needs a `@todo` puzzle. Closing #5723 on the current sentence would leave the claim standing and the next reader trusting it.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -181,6 +200,7 @@ public final Phi copy() {
             for (final Map.Entry<String, Attribute> ent : this.attrs.entrySet()) {
                 map.put(ent.getKey(), ent.getValue().copy(copy));
             }
+            copy.lock = new ReentrantLock();
             copy.attrs = map;
```
Favixx, line 204: A consequence of routing `copy()` through `activate()` that is new with this PR: copying a not-yet-activated shared instance now takes that instance's lock, so N threads copying the same package object serialise on it.

That is correct — they must, or they would race on the very maps `copy()` reads — but it turns `copy()` from a lock-free operation into a contended one for exactly the shared singletons the description names as the motivation (`PhNest`, `PhPackage`). Only for the first copy, since afterwards the fast path returns immediately, so this is most likely fine; it is worth a sentence in the description because it is the one behavioural change on a hot path that a reader would not predict from the title.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
```diff
@@ -372,6 +374,25 @@ void createsDifferentPhiInParallel() {
...
+    void activatesLazilyOnceUnderConcurrentAccess() {
+        final int threads = 100;
```
Favixx, line 380: 100 threads, but almost certainly not 100 racers. `Together` starts them in sequence, and the first to arrive holds the lock for the whole of `publish()` while the rest queue behind it — so what this measures is mostly the *queued* path, which was never the broken one.

To make them actually collide on the `attrs == null` check, the threads need to be released together: a `CountDownLatch(1)` awaited inside the lambda immediately before `shared.take("foo")`, counted down after all threads are running. That is three lines and turns a test that happens to pass into one that reproduces the race it is named after.

@morphqdd already noted the probabilistic nature of this test; this is the specific reason it is weaker than it looks.

Favixx (commented):
The double-checked lock itself is built correctly, and I am not repeating @morphqdd's verification of the `order`-before-`attrs` ordering — I traced the same call sites and reached the same conclusion. I also checked the part of this refactor most likely to change behaviour silently: `publish()` inlines the body of `add()` rather than calling it, and the inlining is faithful, `SORTABLE` test and `AtWithRho` wrapping included.

The thing I would not merge without an answer is one line above the fix. `copy()` now does `copy.lock = new ReentrantLock()` because `Object.clone()` would otherwise alias the original's lock — and `order` is aliased in exactly the same way, by the same shallow copy, and is not reset. `attrs` is replaced, `lock` is now replaced, `order` is not. There is a concrete path from there to an NPE, spelled out inline.

Separately, and worth stating because #5723 is about the class doc rather than about `activate()`: this PR makes initialization thread-safe, and the class still says "The class is thread-safe" at line 27 while `add()` and `put()` mutate two plain `HashMap`s afterwards with no lock at all. The title of this PR is accurate; the issue it closes is broader than what it fixes.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
```diff
@@ -372,6 +374,25 @@ void createsDifferentPhiInParallel() {
...
+        final Map<String, Attribute> attributes = new HashMap<>(1);
+        attributes.put("foo", new AtComposite(null, rho -> new Data.ToPhi(42L)));
+        final Phi shared = new PhDefault(attributes);
```
Favixx, line 383: One attribute means `order` ends up with a single entry, so the test cannot observe the property that makes the non-volatile `order` field safe — that it is *fully* populated by the time `attrs` becomes visible.

With one entry there is no ordering to get wrong. Three sortable attributes and an assertion on positional access (`attr(0)`, `attr(1)`, `attr(2)` resolving to the right names from a racing thread) would exercise the publication of `order` rather than just of `attrs`, which is where the subtle half of this fix lives.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
```diff
@@ -372,6 +374,25 @@ void createsDifferentPhiInParallel() {
...
+        MatcherAssert.assertThat(
+            "concurrent first access must activate the object exactly once, with no half-built or lost attributes",
+            new SetOf<>(
+                new Together<>(
+                    threads,
+                    t -> new Dataized(shared.take("foo")).asNumber()
+                )
+            ),
+            Matchers.contains(42.0)
```
Favixx, line 392: `contains(42.0)` on a `SetOf` asserts what the set holds, not how many threads put something in it. A run where 99 threads silently failed to start and one returned `42.0` passes identically to a run where all 100 succeeded.

Since the claim is "every thread saw the same fully-built result", the count is half of it. Collecting into a `List` and asserting both `hasSize(100)` and that every element is `42.0` — or asserting the set has size 1 *and* the list has size 100 — says the whole thing.

eo-runtime/src/main/java/org/eolang/PhDefault.java
morphqdd, line 610: publish() duplicates the SORTABLE-check-and-wrap logic already in add() (name matches SORTABLE -> order.put, then attrs.put). Can't call add() directly here since this operates on the local built/ordered maps mid-construction, but the shared logic could still be pulled into one small method parameterized on the target maps.

VasilevNStas:
@anuragpaul602-netizen gentle nudge: the reviews here are positive — @morphqdd traced every reader of `attrs`/`order` and @Favixx confirmed the double-checked lock is built correctly. The branch, however, has been idle since Aug 1 and now has merge conflicts with `master`.

Are you planning to resolve the conflicts and continue? If you have moved on, it would help the maintainers to know whether to take the branch over or close it — the fix is clearly wanted (#5723).

yegor256:
@anuragpaul602-netizen conflict here, please fix

yegor256:

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (9) comments.  Your running score is [+5946](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1969](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+281](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5770, opened 2026-07-22
#5633: surface the real out-of-bounds cause from string.slice

### LINKED ISSUE #5633: `string.slice` hides out-of-range errors with inaccurate "must be a number" message
## What happens

`string.slice` (`eo-runtime/src/main/eo/string.eo:72-151`) mishandles out-of-range indices in two related ways.

**1. The real "out of string bounds" cause is masked into a false "must be a number" error.**

For an out-of-range slice, the byte-offset helpers raise the intended message via a bottom (⊥):

```
minus. > blen
  recidx (number bstart) 0 nlen
    "Start index + length to slice (%d) is out of string bounds".printf * (nstart.plus nlen)
  bstart
recidx > bstart!
  0 0 nstart
  "Start index (%d) is out of string bounds".printf * nstart
```

That ⊥ then flows as `blen`/`bstart` into `as-bytes.slice bstart blen`. `EObytes$EOslice` reads those params through an `Expect` chain that ends in `.otherwise("must be a number")` (`eo-runtime/src/main/java/org/eolang/EObytes$EOslice.java:68-77`), which **overwrites the ⊥'s real cause**. The caller therefore sees `the 'len' attribute must be a number` (or `'start'`) even though the value is a perfectly valid number — the actual problem is the range, and the two carefully-written "out of string bounds" messages are unreachable dead code. This masking was effectively introduced when `string.slice` moved to ⊥/fallback in #5324, and is the same "actual exception masked by `Expect`" class as #5478.

**2. A zero length silently ignores an out-of-range start.**

The guard chain returns early on `nlen.eq 0` → `""` (`string.eo:83-85`) *before* the start upper-bound check (`recidx ... "Start index (%d) is out of string bounds"`) is ever forced. So a start far past the end, combined with `len == 0`, is silently accepted and returns `""` with no error, contradicting the object's own bounds contract. (A negative start is still caught, because `nstart.lt 0` is checked up front.)

## Failure scenarios

Verified end-to-end against the compiled runtime (`Phi.Φ.take("string").take("slice")`, `text/start/len` bound):

| call | result | expected |
| --- | --- | --- |
| `slice "hello" 2 3` | `"llo"` | `"llo"` |
| `slice "hello" 2 10` | `ExOtherwise: the 'len' attribute must be a number` | an "out of string bounds" error |
| `slice "hello" 0 6` | `the 'len' attribute must be a number` | "out of string bounds" |
| `slice "hello" 10 1` | `the 'start' attribute must be a number` | "Start index (10) is out of string bounds" |
| `slice "hello" 5 1` | `the 'len' attribute must be a number` | "out of string bounds" |
| `slice "hello" 10 0` | `""` (no error) | "Start index (10) is out of string bounds" |
| `slice "hello" 6 0` | `""` (no error) | out-of-bounds error |

`len` is `10`/`1` — a valid number — in every masked case, so the reported message is actively misleading.

## Suggested resolution

1. **Surface the real bounds cause.** Perform the out-of-range check in `string.slice` itself and raise the "out of string bounds" message as the object's own `@` result, rather than letting a ⊥ travel into `as-bytes.slice` where `Expect.otherwise` overrides it. Alternatively (broader, shared with #5478), make the `Expect` `.otherwise(...)` fallback preserve a terminator/⊥ that already carries an explanatory cause instead of replacing it with "must be a number".
2. **Check the start upper bound regardless of length.** Move the `nlen.eq 0 → ""` shortcut *after* the start-in-range check (or add an explicit `nstart` upper-bound guard), so `slice "hello" 10 0` raises "Start index (10) is out of string bounds" instead of silently returning `""`.

Add tests that `slice "hello" 2 10`, `slice "hello" 10 1` and `slice "hello" 10 0` each produce the intended out-of-bounds message rather than a "must be a number" error or a silent empty string.

gemshrine:
Fixes #5633.

`string.slice` mishandled out-of-range indices in two related ways.

**1. The real "out of string bounds" cause was masked into a false "must be a number" error.** For an out-of-range slice, the byte-offset helpers `bstart`/`blen` already raise the intended message via a bottom (⊥). That ⊥ then flowed as `bstart`/`blen` into `as-bytes.slice`, whose `Expect` chain ends in `.otherwise("must be a number")`. As detailed by @morphqdd on the issue, `Expect`'s `applyOtherwise` catches the inner `ExThat` (which *does* carry the real "out of string bounds" message) and rethrows `ExOtherwise(String.format("%s %s", subject, "must be a number"), ex)` — the `ex` is passed but never becomes the cause (no placeholder, no `initCause`), so the real message is lost. The caller saw `the 'len'/'start' attribute must be a number` for a perfectly valid number, and the two carefully-written "out of string bounds" messages were unreachable.

**2. A zero length silently ignored an out-of-range start.** The `nlen.eq 0 → ""` shortcut returned before the start upper-bound check (`bstart`) was ever forced, so e.g. `slice "hello" 10 0` returned `""` with no error.

## Fix

Both `bstart` and `blen` are now forced via `seq` **before** `as-bytes.slice` is reached:

```
seq *
  bstart
  blen
  if.
    nlen.eq 0
    ""
    string
      as-bytes.slice bstart blen
```

Their ⊥ therefore surfaces directly as the object's own `@` result with the correct message, never entering the `Expect` chain that would rewrite it, and the start upper bound is now checked regardless of the length. This keeps the fix contained to `string.eo` and leaves the shared `Expect.java` masking (the broader #5478 angle @morphqdd is tackling in #5484) untouched.

| call | before | after |
| --- | --- | --- |
| `slice "hello" 2 3` | `"llo"` | `"llo"` (unchanged) |
| `slice "hello" 2 10` | `the 'len' attribute must be a number` | `Start index + length to slice (12) is out of string bounds` |
| `slice "hello" 10 1` | `the 'start' attribute must be a number` | `Start index (10) is out of string bounds` |
| `slice "hello" 10 0` | `""` (no error) | `Start index (10) is out of string bounds` |

## Tests

Both bugs change behavior from "returns a value" to "terminates with the right message", which the EO example DSL cannot discriminate (a `-->` negative example only asserts that dataization throws, and `asBool` on the pre-fix `""`/wrong-⊥ throws anyway). Added `StringSliceTest` (hand-written, in `org.eolang.EO_string`, modeled on `EObytesEOsliceTest`/`RegexAtomTest`) that asserts the **surfaced message** for the out-of-range start and length cases and that the zero-length case now terminates — verified these three fail against the pre-fix implementation (`must be a number` / nothing thrown) and pass with the fix. Also added an EO regression example for the valid `start == length, len == 0` edge (`slice "hello" 5 0 → ""`). Full `eo-runtime` suite green apart from the pre-existing `EOsocketTest.refusesConnectionViaSyscall` Windows/TEST-NET artifact (also fails on unmodified master). No new lint warnings.

### DIFF
--- a/eo-runtime/src/main/eo/string.eo
+++ b/eo-runtime/src/main/eo/string.eo
@@ -77,11 +77,14 @@
         T
           "Length to slice must be >= 0, but was %d".printf
             * nlen
-        if.
-          nlen.eq 0
-          ""
-          string
-            as-bytes.slice bstart blen
+        seq *
+          bstart
+          blen
+          if.
+            nlen.eq 0
+            ""
+            string
+              as-bytes.slice bstart blen
     minus. > blen
       recidx
         number bstart
@@ -248,6 +251,12 @@
       str.length.eq 11
       str.as-bytes.size.eq 14
     "The 😀 smile" > str
+
+  eq. ++> tests-slice-at-end-with-zero-length
+    "hello".slice
+      5
+      0
+    ""
 
   eq. ++> tests-slice-empty-string
     "".slice
--- a/eo-runtime/src/test/java/org/eolang/EO_string/StringSliceTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EO_string/StringSliceTest.java
@@ -1,0 +1,117 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+
+/*
+ * @checkstyle PackageNameCheck (4 lines)
+ * @checkstyle TrailingCommentCheck (3 lines)
+ */
+package org.eolang.EO_string; // NOPMD
+
+import org.cactoos.text.TextOf;
+import org.cactoos.text.UncheckedText;
+import org.eolang.Data;
+import org.eolang.Dataized;
+import org.eolang.ExAbstract;
+import org.eolang.PhApplication;
+import org.eolang.Phi;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests for {@code string.slice} out-of-bounds behavior.
+ * @since 0.57.4
+ */
+@SuppressWarnings("JTCOP.RuleAllTestsHaveProductionClass")
+final class StringSliceTest {
+
+    @Test
+    void takesLegalSlice() {
+        MatcherAssert.assertThat(
+            "a valid slice is taken correctly",
+            new Dataized(
+                StringSliceTest.slice("hello", 2, 3)
+            ).asString(),
+            Matchers.equalTo("llo")
+        );
+    }
+
+    @Test
+    void surfacesOutOfBoundsCauseForLength() {
+        MatcherAssert.assertThat(
+            "the real out-of-bounds cause must reach the caller, not the 'must be a number' mask",
+            new UncheckedText(
+                new TextOf(
+                    Assertions.assertThrows(
+                        ExAbstract.class,
+                        () -> new Dataized(
+                            StringSliceTest.slice("hello", 2, 10)
+                        ).asString(),
+                        "an out-of-range length must terminate with the reason"
+                    )
+                )
+            ).asString(),
+            Matchers.containsString("out of string bounds")
+        );
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (136 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+415](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5771, opened 2026-07-22
#5718: exclude quoted string literals from Penalty's code-structure scoring

### LINKED ISSUE #5718: `Penalty` incorrectly counts string literals as code structure in eo-printer

morphqdd:
Closes #5718.

`Penalty`'s helpers in `eo-printer/src/main/java/org/eolang/printer/Penalty.java` (`tokens()`, `applied()`, `brackets()`, `phis()`, `ifs()`, `spaces()`) scanned a line's raw characters with no notion of a quoted string literal. Data sitting inside a `"..."` string (spaces between words, a stray `(`/`)`, an `@`, or the substring `.if`) was counted exactly like real EO syntax, so a formation's flat-candidate score could be inflated purely by the wording of a string literal it contains, pushing `eo:format` toward a more vertical layout than the code's actual structure warranted.

Fix: added one shared `Penalty.masked(String)` helper that replaces the interior of quoted string literals with a neutral filler character (respecting `\"` escapes), and routed all six scoring methods through it instead of the raw line, so only genuine code structure is scored.

Second commit: fixing the scoring legitimately changes the canonical layout of 26 existing `eo-runtime` sources, since their formatting had been silently shaped by string-literal wording rather than actual structure. Regenerated them with `-Deo.autoFix` so `mvn clean install`'s canonical-format check keeps passing after this merges (without this, master's own build would immediately start failing its own format check, since Penalty would now compute different canonical output than what's currently committed).

Verified:
- `PenaltyTest`: 19/19 green.
- Full `eo-runtime` test suite after the reformat: 1730/1730 green.
- `mvn clean verify -PskipTests -Pqulice` (the exact CI gate), full repo: clean.

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Penalty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Penalty.java
@@ -73,6 +73,7 @@ import java.util.Map;
  *
  * @since 0.57.0
  */
+@SuppressWarnings("PMD.TooManyMethods")
 final class Penalty {
 
     /**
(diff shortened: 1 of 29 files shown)


yegor256:
@morphqdd thanks for your contribution, but I believe it's a wrong move. see my explanation in #5718

PR #5775, opened 2026-07-22
Implement directory.tmpfile.touch in EO

### LINKED ISSUE #5752: `directory.tmpfile.touch` should be implemented in EO, not Java
The `directory.tmpfile.touch` object is a native atom (`eo-runtime/src/main/eo/directory.eo:61`, `[] > touch /Q.string`, implemented by `eo-runtime/src/main/java/org/eolang/EOdirectory$EOtmpfile$EOtouch.java`). All it does is create a uniquely-named empty file in the enclosing temporary directory with `Files.createTempFile` and return its absolute path.

Every piece this needs already exists in EO, so nothing here requires Java. The enclosing directory's path is reachable through the RHO chain — the atom literally reads `RHO.RHO.path`, which in EO is `^.^`. A unique name can be built from the primitives already shipped: `clock` (milliseconds since the epoch), `posix "getpid"`, and the seeded `number.random`. And the file itself is created through exactly the `posix`/`win32` `open` syscall that `file.open "w"` already uses (see `file.eo`, `open-file`), with the `O_CREAT`/`creat` and exclusive flags already exposed in `posix.eo`. So `touch` is composition of existing objects, not a new capability.

The fix is to implement `touch` in EO: derive a unique name, create the file via `posix "open"` with `O_CREAT | O_EXCL` for race-free creation, and return its path. Two properties of `createTempFile` are worth noting. It guarantees atomic unique creation, so the EO version must use `O_EXCL` rather than check-then-create to avoid a TOCTOU race. And it registers `deleteOnExit`, a JVM-only shutdown hook that the eo2js/Node runtime cannot honor anyway; since every call site immediately chains `.deleted`, dropping it is safe and makes both runtimes behave alike.

Titanism0v0:
## Summary

- Implement `directory.tmpfile.touch` in pure EO using the existing clock, platform PID, random number, open syscall, and exclusive-create flags.
- Add EO coverage for unique paths, empty files, deletion, and missing parent directories.
- Remove the obsolete Java native implementation.

## Validation

- `git diff --check` passed.
- Source checks confirmed no stale Java native reference and confirmed platform PID/exclusive flag usage.
- `mvn clean install -ntp -PskipITs --errors --batch-mode` reaches `eo-maven-plugin` but fails in `MjPrintTest` because the JucsProvider test resource path is unavailable; `eo-runtime` is skipped.

No `eo2js` or `file.eo` changes are included.

Fixes #5752


yegor256:
@Titanism0v0 merge conflict here, please fix

eo-runtime/src/main/eo/directory.eo
```diff
@@ -57,7 +57,93 @@
...
+      code. > closed2
+        os.is-windows.if
+          win32 *1
+            "_close"
+            descriptor2
+          posix *1
+            "close"
+            descriptor2
+      opened.as-number > descriptor!
+      opened2.as-number > descriptor2!
```
morphqdd: Both `descriptor` and `descriptor2` are `!`-forced locals, meaning both `opened` and `opened2` get dataized (and so both `candidate` and `candidate2` get created via `open(..., O_CREAT|O_EXCL, ...)`) as soon as `touch`'s scope is set up, regardless of which one the `@` branch below actually ends up using. The success branch (`descriptor != -1`) only closes `closed` and returns `candidate`; nothing in that branch touches `closed2` or `candidate2`. Since a random-name collision between two 4-part identifiers (clock+pid+index+random) is astronomically unlikely, the success branch is what runs almost every time, and this is exactly the leak I reproduced empirically (see the top-level comment): a second file/dir gets created and its descriptor never closed, on every single successful call. A design that tries `candidate` first and only computes/opens `candidate2` if the first attempt genuinely fails (a real retry, not two eager opens with one discarded) would avoid creating the second entry at all in the common case.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -57,7 +57,93 @@
...
+          pid
+          0
+          random
+      "%d-%d-%d-%f".printf > name2!
+        *
+          clock.as-number
+          pid
+          1
+          random2
+      code. > opened
```
morphqdd: `opened` and `opened2` (and `closed`/`closed2` below) are near-identical blocks, differing only in which candidate/descriptor variable they reference. Beyond the duplication itself, this is exactly the shape that produces the leak: because retrying is expressed as "compute both, pick one" instead of "try one, retry only if it fails", both entries get created unconditionally. Collapsing this to a single parameterized attempt, called once and only retried on actual failure, would fix the duplication and the leak in the same change.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -57,7 +57,93 @@
...
+    [] > touch
+      if. > @
+        descriptor.eq -1
+        if.
+          descriptor2.eq -1
+          T
+            "Can't create a temp file '%s'".printf
+              * candidate2
+          seq
+            * closed2 candidate2
```
morphqdd: Beyond the resource leak: I confirmed empirically that the leaked filesystem entries are directories, not files, even though this path opens them via `posix`/`win32` "open" with `O_CREAT`, which creates a regular file, not a directory, on every platform I'm aware of. Worth checking whether the flag combination (`(posix.rdonly.plus posix.creat).plus posix.exclusive`, or the equivalent win32 one) is actually being interpreted as intended by the underlying native atom, since "open() created a directory" suggests something in that chain isn't doing what the EO source says it should.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -95,6 +181,37 @@
...
+    not. > valid!
+      first.path.eq second.path
+
+  ++> tests-tmpfile-touch-deletes-created-file
+    seq * > @
+      existed
+      deleted
+      and.
+        existed.and
+          deleted.path.eq temp.path
```
morphqdd, line 123: None of the three new tests check that no orphan file or directory is left behind after a `touch` call, which is exactly why the leak I found here wasn't caught by this PR's own suite. A test that snapshots the directory listing before and after a `tmpfile` call (or, simpler, that the parent directory's file count is unchanged once the returned handle is deleted) would have caught this.

morphqdd (changes requested):
I built this branch and ran the actual `eo-runtime` `EOdirectoryTest` suite (the PR description says this wasn't verified: "eo-runtime is skipped" in the validation notes), then checked `/tmp` before and after with a controlled diff. Zero stray entries matching the `clock-pid-index-random` naming pattern before the run, exactly two after, with clock/pid values matching the test run's own timing. That's a real, reproducible leak, not a hypothesis. One of the two leaked entries is worse than an empty stray file: it's a whole uncleaned nested directory tree (`foo/bar/test.txt`, `x/y/z/a.txt`, matching the fixture shape of an existing recursive-directory test), meaning some existing test's own temp working directory silently stopped getting cleaned up. Also unexpected: both leaked entries are **directories**, not files, despite `touch` opening them as regular files via `posix`/`win32` `open`.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -57,7 +57,75 @@
...
+              posix.mode
+      code. > pid!
+        os.is-windows.if
+          win32
+            "_getpid"
+            *
+          posix
+            "getpid"
+            *
+      [] > retry
```
morphqdd: Uniqueness comes from `number.random clock`, which is a pure function of the seed. Two `touch` calls in the same process within the same clock tick get the same random value, so the only thing distinguishing candidates is the fixed index. `retry` only gives one more attempt (index 1); a third collision in the same tick has nowhere to go and throws "Can't create a temp file". `Files.createTempFile`, which this replaces, had no such cap.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -57,7 +57,75 @@
...
+        if. > @
+          second-descriptor.eq -1
+          T
+            "Can't create a temp file '%s'".printf
+              * second-candidate
+          seq
+            * second-closed second-candidate
+        candidate 1 > second-candidate!
+        code. > second-closed
```
morphqdd: `second-closed` duplicates `first-closed` (line 81) except for which descriptor it closes. `candidate` was already factored out the same way two lines up, worth doing the same for the close logic.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -95,6 +163,46 @@
...
 
+  ++> tests-tmpfile-touch-creates-empty-file
+    seq * > @
+      size
+      temp.deleted
+      size.eq 0
+    temp.size > size
+    mktemp.tmpfile > temp
+
+  ++> tests-tmpfile-touch-creates-unique-paths
```
morphqdd: This only creates two temp files sequentially, so it doesn't exercise the retry/collision path added above. It would pass even if `retry` were broken.

morphqdd (changes requested):

Titanism0v0:
Thanks for the detailed review. I’ll rework `touch` into a single lazy attempt: a new candidate will be generated and opened only after an actual `O_EXCL` collision, rather than eagerly creating both candidates. I’ll factor out the duplicated close logic as part of that change.

I’ll also remove the two-attempt retry cap and align the retry behavior with the previous `Files.createTempFile` semantics. The tests will include deterministic collision/retry coverage and verify that no orphan file or directory remains after the returned temporary file is deleted. Finally, I’ll verify that the POSIX and Windows flag combinations create a regular file as intended.

yegor256:
@Titanism0v0 good to merge? Keep in mind that while your pull request stays in "draft" status we don't merge it and don't look at it.

Titanism0v0:
@yegor256 Not yet. I’m addressing the remaining requested changes and the merge conflict. I’ll mark the PR ready for review once the runtime tests and CI pass.

Titanism0v0:
@yegor256 The merge conflict and requested tmpfile changes are resolved. The implementation now creates files in EO using exclusive creation with unbounded collision retry, closes the successful descriptor once, and removes the Java native atom. I added coverage for regular/empty files, uniqueness, deletion/cleanup, an absent directory, and a deterministic collision retry.\n\nValidated locally: EO formatting, compilation and linting passed; EOdirectoryTest passed 11/11. Could you please review the updated head?

Titanism0v0:
@yegor256 The corrected implementation and the requested regression coverage require 223 changed lines (+172/−51); the 51 removed lines are the obsolete Java native atom. The pr-size workflow has a hard 200-line limit and no exemption path. Further compression has started to make the EO tests less reliable. Would you prefer a one-off exception for this PR, or should I split the work into dependent PRs?

0crat:
@titanism0v0 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (223 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+64](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+3490](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5777, opened 2026-07-23
#5774: drop spurious minus sign for negative values that render as zero

### LINKED ISSUE #5774: `string.as-decimal` and `string.as-fixed` emit a spurious minus sign ("-0", "-0.00") for values that truncate/round to zero
## What happens

`string.as-decimal` (`eo-runtime/src/main/eo/string/as-decimal.eo`) and `string.as-fixed` (`eo-runtime/src/main/eo/string/as-fixed.eo`) prepend a minus sign to the output whenever the *input* is negative, even when the value they actually render is zero. So a small negative number that truncates (or rounds) to zero magnitude is written with a leading `-`:

- `as-decimal -0.5` → `"-0"` (should be `"0"`)
- `as-fixed -0.0001 2` → `"-0.00"` (should be `"0.00"`)

`"-0"` / `"-0.00"` misrepresents the value: no other decimal formatter emits a signed zero here, and `printf "%d"` on the same input already gives `"0"` (its `%d` path truncates first and never adds a sign for zero). A `"-0.00"` in fixed‑point output is a well‑known display defect (e.g. money and reports).

All results below were verified by building `eo-runtime` from source and dataizing real calls, with passing controls in the same run to isolate the defect.

## Root cause

Both objects choose the sign from `n.lt 0` on the **original** argument, before the magnitude is truncated/rounded down to zero.

`as-decimal`:

https://github.com/objectionary/eo/blob/667b081eb6d1ca0c1d8ea3611dd273e51d770cfd/eo-runtime/src/main/eo/string/as-decimal.eo#L11-L17

```
[n] > as-decimal
  if. > @
    n.lt 0
    "-".as-bytes.concat
      digits
        n.times -1
    digits n
```

For any `n` in `(-1, 0)`, `n.lt 0` is true, so a `"-"` is prepended, but `digits (n.times -1)` truncates the fraction to `"0"` — yielding `"-0"`.

`as-fixed`:

https://github.com/objectionary/eo/blob/667b081eb6d1ca0c1d8ea3611dd273e51d770cfd/eo-runtime/src/main/eo/string/as-fixed.eo#L12-L18

```
[n places] > as-fixed
  if. > @
    n.lt 0
    "-".as-bytes.concat
      positive
        n.times -1
    positive n
```

Same shape: when `|n|` rounds to `0.00…0` at the requested `places`, `positive` renders all‑zero digits, but the `"-"` has already been committed, giving `"-0.00"`.

## Failure scenarios

| call | result | expected |
| --- | --- | --- |
| `as-decimal -0.5` | `"-0"` | `"0"` |
| `as-decimal -0.9` | `"-0"` | `"0"` |
| `as-decimal 0.5` | `"0"` | `"0"` (control — passes) |
| `as-decimal -1.5` | `"-1"` | `"-1"` (control — passes) |
| `as-fixed -0.0001 2` | `"-0.00"` | `"0.00"` |
| `as-fixed -0.001 2` | `"-0.00"` | `"0.00"` |
| `as-fixed -0.4 2` | `"-0.40"` | `"-0.40"` (control — passes) |
| `as-fixed 0.0001 2` | `"0.00"` | `"0.00"` (control — passes) |

The controls confirm the defect is specific to inputs whose magnitude collapses to zero after truncation/rounding: a genuinely non‑zero negative (`-1.5`, `-0.4`) is signed correctly, and positive inputs are unaffected.

## Suggested fix

Decide the sign from the rendered magnitude, not from the raw input — i.e. only prepend `"-"` when the truncated/rounded value is actually non‑zero. Concretely:

- In `as-decimal`, compute the digit string of `|n|` first and prepend `"-"` only when that string is not `"0"` (equivalently, when `(n.times -1).floor.gte 1`).
- In `as-fixed`, gate the `"-"` on whether the rounded integer‑plus‑fraction is non‑zero (e.g. when `floor((|n| * scale) + 0.5)` is `> 0`), so a value that rounds to `0.00…0` is written without a sign.

Add tests that `as-decimal -0.5` yields `"0"` and `as-fixed -0.0001 2` yields `"0.00"`, alongside the existing negative cases (`-42`, `-2.5`) to lock in that genuine negatives keep their sign.

kdelay:
## What

`string.as-decimal` and `string.as-fixed` chose the output sign from `n.lt 0`
on the **original** argument, before the magnitude was truncated/rounded down
to zero. So a small negative value whose rendered magnitude is all zeros was
still written with a leading `-`:

- `as-decimal -0.5` produced `"-0"` (expected `"0"`)
- `as-fixed -0.0001 2` produced `"-0.00"` (expected `"0.00"`)

A signed zero misrepresents the value and is a well-known display defect in
fixed-point output (money, reports). `printf "%d"` on the same input already
yields `"0"`.

## Fix

Decide the sign from the magnitude that is actually rendered, not from the raw
input:

- `as-decimal`: the negative branch delegates to a `negative` helper that omits
  the `-` when the truncated integer magnitude (`m.floor`) is zero.
- `as-fixed`: the negative branch delegates to a `signed` helper that omits the
  `-` when the value rounded to `places` (`floor(|n| * 10^places + 0.5)`) is
  zero.

Non-zero negatives keep their sign; positives are unchanged.

## Verification

Built `eo-runtime` from source and ran its test suite. Added EO test objects
covering the defect, alongside the existing controls:

| call | before | after |
| --- | --- | --- |
| `as-decimal -0.5` | `"-0"` | `"0"` |
| `as-decimal -0.9` | `"-0"` | `"0"` |
| `as-decimal -7.89` | `"-7"` | `"-7"` (control) |
| `as-fixed -0.0001 2` | `"-0.00"` | `"0.00"` |
| `as-fixed -0.001 2` | `"-0.00"` | `"0.00"` |
| `as-fixed -2.5 6` | `"-2.500000"` | `"-2.500000"` (control) |

Fixes #5774

### DIFF
--- a/eo-runtime/src/main/eo/string/as-decimal.eo
+++ b/eo-runtime/src/main/eo/string/as-decimal.eo
@@ -11,9 +11,8 @@
 [n] > as-decimal
   if. > @
     n.lt 0
-    "-".as-bytes.concat
-      digits
-        n.times -1
+    negative
+      n.times -1
     digits n
   [n] > digits
     if. > @
@@ -27,6 +26,21 @@
             48
     floor. > q
       n.div 10
+  if. > [m] > negative
+    m.floor.eq 0
+    digits m
+    "-".as-bytes.concat
+      digits m
+
+  eq. ++> tests-negative-below-one-drops-sign
+    "0"
+    string
+      as-decimal -0.9
+
+  eq. ++> tests-negative-truncating-to-zero-drops-sign
+    "0"
+    string
+      as-decimal -0.5
 
   eq. ++> tests-truncates-negative-fraction
     "-7"
--- a/eo-runtime/src/main/eo/string/as-fixed.eo
+++ b/eo-runtime/src/main/eo/string/as-fixed.eo
@@ -12,9 +12,8 @@
 [n places] > as-fixed
   if. > @
     n.lt 0
-    "-".as-bytes.concat
-      positive
-        n.times -1
+    signed
+      n.times -1
     positive n
   [a] > positive
     concat. > @
@@ -57,6 +56,17 @@
                 (value.div 10).floor.times 10
             48
         acc
+  if. > [a] > signed
+    eq.
+      floor.
+        plus.
+          a.times
+            pow10 places
+          0.5
+      0
+    positive a
+    "-".as-bytes.concat
+      positive a
 
   eq. ++> tests-carries-rounding-into-integer-part
     "1.000000"
@@ -67,6 +77,16 @@
     "3.14"
     string
       as-fixed 3.14159 2
+
+  eq. ++> tests-negative-rounding-to-zero-drops-sign
+    "0.00"
+    string
+      as-fixed -1.0E-4 2
+
+  eq. ++> tests-negative-tiny-rounding-to-zero-drops-sign
+    "0.00"
+    string
+      as-fixed -0.001 2
 
   eq. ++> tests-pads-with-trailing-zeros
     "1.250000"


kdelay:
Thanks for the quick review and merge!

0crat:
@kdelay Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5781, opened 2026-07-23
#5780: scope-guard file-local handle resolution

### LINKED ISSUE #5780: Parser's file-local `>>` handle hijacks same-named `>` references in sibling scopes
On branch `5678`, giving one of two sibling formations a private `>> foo` handle whose name also names a public `> foo` attribute in the *other* formation makes the parser bind that other formation's own `foo` references to the handle. In `eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl:18-19` the pass builds a file-wide key of every `o[@local]` handle (`<xsl:key name="handle" match="o[@local]" use="@local"/>`) and rewrites *any* `@base` equal to a handle name to that handle's cactus `@name`, with no regard for lexical scope. A later rooting pass, unable to find that cactus name in the reference's own scope, prefixes it with `Φ`, producing a dangling `Φ.<name>`.

Minimal case:

```
[] > top
  os.pick > @
    []
      [] >> blk
        blk > @
    []
      [] > blk
        blk.foo > @
```

The second `blk.foo` — which has its own `> blk` sibling — parses to `@base="Φ.a🌵…"` (bound to the first formation's handle) instead of `ξ.ρ.blk.foo`. With both siblings public (`>`) each reference correctly resolves to its own `ξ.ρ.blk`.

A file-local handle is meant to let an object reference itself; it must not shadow a nearer lexical `>` binding of the same name in an unrelated formation. Because the rewrite is unconditional the handle always wins, and since it lives outside the reference's scope the name ends up rooted at the program root. In `eo-runtime/src/main/eo/console.eo` (windows branch `[buffer] >> input-block` / `[] >> output-block`, posix branch `[buffer] > input-block` / `[] > output-block`) the posix references resolve to `Φ.input-block` / `Φ.output-block` and fail at runtime with `Couldn't find object 'Φ.input-block'`. The same corruption makes `eo-printer` reformat those references to `Q.input-block` / `Q.output-block`, so the printer's canonical output no longer compiles.

Fix: in `resolve-local-names.xsl` guard the rewrite so a reference binds to a handle only when that handle is a lexical ancestor of the reference, letting a nearer `>` attribute of the same name shadow the file-local handle. Alternatively, extend the existing duplicate-`@local` check to also flag a `>>` handle colliding with a `>` attribute of the same name and fail with a clear error instead of emitting a dangling global reference.

yegor256:
Closes #5780.

## Problem

`eo-parser/.../parse/resolve-local-names.xsl` rewrote **every** `@base` equal to a `>> foo` file-local handle name into that handle's cactus `@name`, file-wide, with no regard for lexical scope. When a handle name also named a public `> foo` attribute in an *unrelated sibling* formation, the handle hijacked that formation's own references. The later rooting pass (`build-fqns`) then could not find the cactus name in the reference's own scope and prefixed it with `Φ`, producing a dangling global reference (e.g. `Couldn't find object 'Φ.input-block'` at runtime, and `Q.input-block` in the printer's canonical output).

Minimal case — the second `blk.foo`, which has its own `> blk` sibling, was bound to the *first* formation's `>> blk` handle instead of `ξ.ρ.blk`:

```
[] > top
  os.pick > @
    []
      [] >> blk
        blk > @
    []
      [] > blk
        blk.foo > @
```

## Fix

A `>> foo` handle is an attribute of the formation that declares it, so a reference should bind to it only within that formation's lexical scope. The pass now binds a reference to a handle **only when the handle's declaring formation is the nearest enclosing scope of the reference that declares the name at all** — so a nearer `> foo` attribute (or `>> foo` handle) of the same name shadows a handle in a more distant scope, and a file-local handle no longer hijacks a same-named public attribute in a sibling formation. This mirrors how `build-fqns` resolves a name to the nearest enclosing formation that owns an attribute of that name.

Self-reference (recursion) via `>> foo` is unaffected: the reference lives inside the handle's declaring formation, which remains its nearest declaring scope.

## Tests

- Added `resolve-local-names-sibling-scope.yaml` reproducing the bug: the handle's own reference is rewritten to the cactus name while the sibling formation's `blk.foo` keeps its `blk` base. Verified this pack **fails** on the pre-fix stylesheet and passes after.
- Full `eo-parser` suite green (1496 tests), including the existing `resolve-local-names`, `resolve-local-names-duplicate`, and `void-local-handle` packs (the last confirms a handle is still reachable from a sibling reference in the *same* scope).

https://claude.ai/code/session_018NN1yNHLfj7w5hcbEyKtkJ

---
_Generated by [Claude Code](https://claude.ai/code/session_018NN1yNHLfj7w5hcbEyKtkJ)_

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
@@ -3,22 +3,44 @@
 * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 * SPDX-License-Identifier: MIT
 -->
-<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" exclude-result-prefixes="xs" id="resolve-local-names" version="2.0">
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:eo="https://www.eolang.org" exclude-result-prefixes="xs eo" id="resolve-local-names" version="2.0">
   <!--
   The "&gt;&gt; foo" file-local handle (§3.10 / §9.2): the parser emits the
-  anonymous object with its cactus @name plus a "@local='foo'" marker. We
-  build the per-file "handle -&gt; cactus-name" table and rewrite every @base
-  equal to a handle into that (reserved) cactus name; a handle declared more
-  than once is an error. The "@local" marker is deliberately kept on the
-  declaring object so that later passes (in particular the printer, see
-  #5563) can recover the readable handle from the otherwise-synthetic cactus
-  name instead of printing a placeholder like "vL_P".
+  anonymous object with its cactus @name plus a "@local='foo'" marker. A
+  handle is an attribute of the formation that declares it, so we rewrite a
+  @base equal to a handle name into that (reserved) cactus name; a handle
+  declared more than once is an error. The "@local" marker is deliberately
+  kept on the declaring object so that later passes (in particular the
+  printer, see #5563) can recover the readable handle from the otherwise-
+  synthetic cactus name instead of printing a placeholder like "vL_P".
+
+  The rewrite is guarded by lexical scope (#5780). A reference binds to a
+  handle only when the formation that declares the handle is the nearest
+  enclosing scope of the reference that declares the name at all - so a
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2660](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5784, opened 2026-07-23
#5782: host a moniker on a single-segment method dispatch

### LINKED ISSUE #5782: merge-monikers hosts a moniker only on a bare reference, never on a method dispatch
The printer's `merge-monikers` pass merges an obfuscated `>>`-handle binding onto a *bare* reference but never onto a method-dispatch reference, so two structurally identical blocks print asymmetrically. In `console.eo`, once its I/O blocks carry file-local `>>` handles (as on branch `5678`), `read` delegates through a bare reference `(input-block --).read size` and the block becomes a moniker:

```
[size] > read
  @. > @
    read.
      [buffer] >> input-block
        ...
      | --
      size
```

while the parallel `write` delegates through a method dispatch `output-block.write buffer` and the block stays expanded as a standalone binding:

```
[buffer] > write
  @. > @
    output-block.write buffer
  [] >> output-block
    ...
```

The decision is in `eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl:46`: `eo:resolved-ref` yields a hostable name only when the base after `ξ.` has no trailing dot (`not(contains($tail, '.'))`). A bare `ξ.input-block` passes; a dispatch `ξ.output-block.write` is rejected, so `eo:moniker-refs` is empty and the binding is left in place.

This is an artificial limit. A reversed dispatch is prefix method dispatch, so `output-block.write buffer` and `write. output-block buffer` parse to the same graph. The block could therefore be inlined as the *receiver* of a reversed dispatch, exactly as the bare case is inlined as a plain receiver. Rejecting every trailing-dot reference makes two equivalent blocks print differently for no reason beyond how the decorator happens to spell the call.

Proposed fix: extend `eo:resolved-ref` / `eo:moniker-refs` to also accept a reference carrying a *single* trailing dispatch segment (`ξ.<name>.<method>`), and add a host branch that emits a reversed dispatch (`write. <binding> <args>`) rather than a bare inline. Keep it to one trailing segment — a multi-segment chain such as `a.b.c.d 1` would need deeply nested reversed dispatches (`d. (c. (b. a)) 1`) and should stay expanded.

yegor256:
Closes #5782.

## Problem

The printer's `merge-monikers` pass merged an obfuscated `>>`-handle binding onto a **bare** reference but never onto a **method-dispatch** reference, so two structurally identical blocks printed asymmetrically. A delegation spelled `input-block` folded the block in as a moniker, while the parallel `output-block.write buffer` left the block expanded — a difference driven only by how the decorator happened to spell the call.

The gate in `eo:resolved-ref` yielded a hostable name only when the base after `ξ.` had no trailing dot, so `ξ.input-block` passed but `ξ.output-block.write` was rejected.

## Fix

A reversed dispatch is a prefix method dispatch: `x.write buf` and `write. x buf` parse to the same graph. So the block can be inlined as the **receiver of a reversed dispatch**, exactly as the bare case is inlined as a plain receiver.

**`merge-monikers.xsl`**
- `eo:resolved-ref` / `eo:moniker-refs` now also accept a reference carrying a **single** trailing dispatch segment (`ξ.<name>.<seg>`), with arguments allowed.
- A new `eo:dispatch-seg` helper extracts that trailing segment.
- The replacement template gains a branch that emits a reversed dispatch `<seg>.` whose receiver is the inlined block and whose arguments are the reference's own children, instead of a bare inline.
- A multi-segment chain (`a.b.c.d 1`) would need deeply nested reversed dispatches and is deliberately left expanded.
- Bare references are hosted first, so a binding with both spellings still folds into the shorter bare inline and existing behaviour is unchanged.

**`inline-cactoos.xsl`** (needed so the binding actually reaches `merge-monikers` in the full pipeline)
- `inline-cactoos` runs before `merge-monikers` and used to unconditionally drop an auto-named `>>`-handle binding unless it was recursive. A handle used only through a dispatch (`auto.gte 0`, never a bare `auto`) had no bare reference to inline, so it was dropped and the dispatch stranded on a synthetic `v_` placeholder — never reaching `merge-monikers`.
- It now also keeps a binding that a surviving method dispatch (`ξ.<name>.<seg>`) still reaches through its name (via the new `eo:dispatched` helper), so the binding lives on for `merge-monikers` to host.

## Tests

- `eo-packs/print/merge-monikers-dispatch.yaml` — unit pack asserting, at the XMIR level, that a handle used only through `auto.gte 0` merges into a reversed `gte.` over the inlined block.
- `print-packs/yaml/moniker-dispatch.yaml` — **end-to-end** pack proving the printed rendering: `auto.gte 0` over a `number num.as-bytes >> auto` handle prints as a `gte.` reversed dispatch with the block inlined as its receiver, and reprints to itself.

Full `XmirTest` suite passes (**144 tests**), including the unchanged `merge-monikers.yaml` pack (confirming bare-reference priority is preserved) and no regressions in the other print packs.

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -62,11 +62,17 @@
     </xsl:choose>
   </xsl:template>
   <!--
-  Drop an inlined auto-named abstract; keep cactus-named voids and
-  keep self-referential (recursive) abstracts, which are never inlined.
+  Drop an inlined auto-named abstract; keep cactus-named voids, keep
+  self-referential (recursive) abstracts, which are never inlined, and keep
+  a binding that a surviving method dispatch still reaches through its name.
+  Such a dispatch reference (`ξ.<name>.<seg>`) is not inlined above — its
+  receiver is buried in a dotted base — so dropping the binding would strand
+  the reference on a synthetic "vL_P" placeholder. Keeping it lets
+  "merge-monikers" host the binding as the receiver of a reversed dispatch
+  instead (#5782).
   -->
   <xsl:template match="o[starts-with(@name, $auto) and not(eo:void(.))]" priority="1">
-    <xsl:if test="eo:recursive(., @name)">
+    <xsl:if test="eo:recursive(., @name) or eo:dispatched(., @name)">
       <xsl:copy>
         <xsl:apply-templates select="node()|@*"/>
       </xsl:copy>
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (199 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2664](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5786, opened 2026-07-23
bug(#5785): fix return declarations

### LINKED ISSUE #5785: `@` syscall atoms in `posix.eo` and `win32.eo` incorrectly declare non-existent `Φ.return` as return type
The `@` atom in `posix.eo:17` declares its return type as `/Q.return`, and `win32.eo:22` does the same. That annotation promotes to the root object `Φ.return`, but nothing defines a `return` at the root. The only `return` objects in the tree are nested: `posix.return` at `posix.eo:30` and `win32.return` at `win32.eo:32`. So each syscall atom claims to return a type that does not exist, while the object it actually computes is its own sibling.

```
[name args] > posix
  [] > @ /Q.return
  ...
  [code output] > return
    output > @
    return > called
```

The atom right beside it gets this right: `string/regex.eo:29` writes the full nested path, `> compile /Q.string.regex.pattern`, so the mismatch here is a missing scope segment, not something the annotation language cannot express. The discrepancy is not only cosmetic. `MjAtomsTable` records the line as `Φ.posix.φ,Φ.return` in the generated `atoms.csv`, which feeds the runtime `AtomTyped` check; with `-Deo.typing=true` that check compares the declared `Φ.return` against the computed object's forma — `Φ.posix.return` — finds them different, and turns every posix and win32 syscall into an `ExFailure`. The check is off by default, so the fault is latent today, but the annotation is already a dangling reference for any tool that reads it.

The smallest fix is to scope each forma to the sibling it means: `/Q.posix.return` in `posix.eo` and `/Q.win32.return` in `win32.eo`. Since `posix.return` and `win32.return` are byte-for-byte identical, the alternative is that a shared root `return` was intended all along, in which case promoting one definition to the root and pointing both atoms at it would remove the duplication and make the existing `/Q.return` correct.

maxonfjvipon:
Closes: #5785

### DIFF
--- a/eo-runtime/src/main/eo/posix.eo
+++ b/eo-runtime/src/main/eo/posix.eo
@@ -27,15 +27,6 @@
   511 > permissions
   0 > rdonly
   2 > rdwr
-  2 > stderr
-  0 > stdin
-  1 > stdout
-  1 > stream
-  6 > tcp
-  1 > wronly
-  os.is-macos.if > trunc
-    1024
-    512
   [code output] > return
     output > @
     return > called
@@ -48,7 +39,16 @@
         address.size
       8
   [mode size] > stat
+  2 > stderr
+  0 > stdin
+  1 > stdout
+  1 > stream
+  6 > tcp
   [seconds micros] > timeval
+  os.is-macos.if > trunc
+    1024
+    512
+  1 > wronly
 
   ++> tests-closes-posix-tcp-socket
     os.is-windows.or > @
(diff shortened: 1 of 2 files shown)


maxonfjvipon:
@yegor256 please check

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+800](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5787, opened 2026-07-23
#5742: emit graph locators on every <o> at parse time

### LINKED ISSUE #5742: `<o>` elements in XMIR lack stable identity due to missing `locator` emission
An `<o>` element in XMIR is identifiable only by its position — `line` and `pos`. Both shift the moment the source is edited or reformatted, so nothing outside the compiler can attach information to a specific object and expect that link to survive the next parse. `XMIR.xsd` already defines the concept that would fix this, a `locator`, but never emits it on `<o>`:

```
<xs:simpleType name="locator">
  ... The locator of an object, which specifies the location of the object within
      the program, in the global graph of objects, starting from Φ. The locator is
      used to uniquely identify an object within the program ...
```

A stable, semantic identity is a prerequisite for keeping an analysis *beside* XMIR rather than inside it — a type-inference sidecar, coverage data, an IDE index — each of which needs to say "this fact is about that object" in a way that survives edits. This is not the old source-location `loc` from #1666 and #1667 that fed `PhLocated` and moved with the text. The graph locator is stable across edits and reorderings precisely because it names an object by where it sits in the object graph (its `Φ.…` path), not by where it sits in the file.

The fix is to emit that graph locator on every `<o>` — restore a `loc` (or `locator`) attribute, typed by the existing `locator` simple type. Leaving it off `<o>` forces every downstream tool to fall back on brittle `line`/`pos` or to re-derive the path itself.

The open questions are the attribute name (`loc` again, or `locator`), whether it is always emitted or opt-in to avoid bloat, and how anonymous or duplicate-named inner objects get a unique path segment. Settling those turns XMIR into a stable anchor that external analyses can reference by identity, which is exactly what a type checker's published-types sidecar would key on.

maxonfjvipon:
An `<o>` in XMIR is identifiable today only by `line`/`pos`, which shift the moment the source is reformatted, so nothing outside the compiler can pin information to a specific object and expect the link to survive the next parse. The fix the schema already anticipates — a `loc` graph locator anchored at `Φ` — exists in code but only runs during transpilation and is never written back into the parsed `.xmir`.

The whole machinery was already there: `XMIR.xsd` declares `loc` (typed by the `locator` simple type), and `set-locators.xsl` computes the `Φ.…` path. It just sat first in the transpile train, feeding `to-java.xsl` and getting discarded. This moves that sheet into the `Canonical` parse pipeline as its final step, so every `<o>` in the stable `1-parse` XMIR carries its locator:

```java
new TrDefault<>(new StHex()),
new TrClasspath<>(
    "/org/eolang/parser/parse/set-locators.xsl"
).back()
```

The sheet stays first in the transpile train too (repointed to its new eo-parser path — the sheet is idempotent, `@* except @loc`), so it recomputes on the transpile-time tree and Java generation is byte-identical, while transpile remains self-contained for pulled or cached XMIR that arrives without `loc`. The 24 transpile-pack references were repointed mechanically.

On the ticket's open questions: the attribute is `loc` (already schema-declared, no new name), it is always emitted (a stable identity is worthless if it is sometimes absent), and anonymous or duplicate-named inner objects get unique segments from the existing algorithm — a named attribute contributes its name, the `@` decoratee `φ`, a reversed-dispatch receiver `ρ`, and any other anonymous child `αN` by argument index. `PARSER_SPEC.md` §9.1.5/9.1.6 documents it.

One honest caveat, noted in the spec: named paths are stable across edits, but positional `αN` segments still shift when an argument is inserted before an anonymous object.

Closes #5742.


maxonfjvipon:
@yegor256 this is ready for review. One open design point worth your call: I emit `loc` on every `<o>` at parse time (always-on), but it could instead be gated behind an `eo.trackLocations`-style flag to keep production XMIR lean — happy to switch if you prefer that. Glad to clarify anything.

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+816](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5795, opened 2026-07-23
Bind a nested file-local handle's self-reference

### LINKED ISSUE #5788: merge-monikers globalizes a recursive formation's self-reference when it relocates the binding
When `eo-printer` merges a recursive file-local `>>` formation onto its reference site, the formation's own self-reference prints as a package-global `Q.<name>` instead of the local handle, silently changing the program. Reproduce on branch `5678` by formatting `eo-printer/src/test/resources/org/eolang/printer/print-packs/yaml/moniker-with-arguments.yaml`: a `front` object whose `[t] >> r` recursion calls `r t.tail` reprints with `Q.r t.tail`, so the `printsToEo`/`printsToParseableEo` checks fail. The relocation happens in the host template `o[exists(eo:hosted-binding(.))]` of `eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl` (~line 130), which copies the binding's subtree verbatim to the reference site.

The move is what breaks it. The parser encodes the self-reference as a fixed climb `ξ.ρ.<cactus>` that is correct only where the binding was declared (a sibling of `@`). `merge-monikers` inlines the formation one level deeper — into the `if.` application that hosted `r list` — without adjusting the climb, so `ξ.ρ` now lands in the wrong scope and `to-eo-tree` falls back to the global `Φ.<name>`. This is independent of the recent merge: the same failure is present on the pre-merge tip.

A concrete fix: when inlining a self-referential binding, rewrite its internal self-reference climbs (`ξ.ρ…`) by the depth delta between the old declaration site and the new inline site, so the self-reference keeps pointing at the moved formation. If that is not desired, leave a self-referential binding standalone instead of hosting it.

yegor256:
When a `>> foo` file-local handle is written nested inside an application (the moniker-with-arguments spelling, §3.14), `resolve-local-names` failed to bind the handle's own body reference to it. The handle is not yet a direct attribute of any enclosing formation at that point in the pipeline, so the `#5780` scope guard found no scope and let the self-reference root at a dangling global `Φ.foo`. This pass now captures such a self-reference to the nearest enclosing handle of the same name whenever the ordinary scope search comes up empty.

The visible effect is in `eo-printer`: the moniker spelling used to parse to a different object graph than the equivalent expanded spelling, so a recursive `[t] >> r` formation no longer looked recursive on the round trip. `inline-cactoos` then inlined it and dropped its `| args` pipe, and `restore-local-names` dropped the `>> r` handle, so the `moniker-with-arguments` pack could not reprint to itself. With the fix both spellings parse identically and the pack round-trips losslessly.

Added a focused parser pack that pins the capture, and the full `eo-parser` (1497) and `eo-printer` (182) suites are green. Closes #5788.

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
@@ -21,7 +21,10 @@
   shadows a handle in a more distant scope, and a file-local handle no longer
   hijacks a same-named public attribute in an unrelated sibling formation.
   This mirrors how "build-fqns" later resolves a name to the nearest
-  enclosing formation that owns an attribute of that name.
+  enclosing formation that owns an attribute of that name. A handle written
+  nested inside an application (a moniker with arguments, §3.14) still binds
+  its own body's self-reference (#5788), before "vars-float-up" hoists it into
+  a proper attribute position.
   -->
   <xsl:output encoding="UTF-8" method="xml"/>
   <!--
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (38) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2696](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5797, opened 2026-07-23
bug(#5794): print anonymous dispatched formations via reversed dispatch

### LINKED ISSUE #5794: eo-printer incorrectly handles anonymous recursive formations causing dangling references
An anonymous file-local recursive formation — one the parser auto-names `a🌵…` with no `@local` handle — that is referenced externally prints lossily. Format this on master:

```
+package foo

[] > bar
  q. > @
    [n] >>
      $ (n.minus 1) > @
    5
```

and it becomes

```
+package foo

[] > bar
  v5_4.q 5 > @
  $ > [n] >>
    n.minus 1
```

`v5_4` names no binding, so the reference dangles and the file no longer re-parses to the same graph; the declaration is emitted anonymously and its `$`-recursion is mangled to `$ > [n] >>`. Found on `master` while investigating #5788 — this is distinct: it occurs with and without `merge-monikers` (the cactus name survives before that pass runs).

This is the un-handled variant of #5681. That issue fixed the *named* `>> name` recursive case by having `eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl` promote the `@local` handle of a recursive formation. An auto-named formation has no `@local`, so that branch never fires: it keeps its cactus `a🌵…` `@name`, `inline-cactoos` correctly keeps the recursive formation in place, and `to-eo-tree.xsl`'s `eo:printable` (~lines 39-42) then rewrites the surviving cactus name to `vL_P` on the references while the declaration prints anonymously — the two disagree and every reference dangles.

Fix: in `restore-local-names.xsl`, when a recursive (`eo:recursive`) formation carries a cactus `@name` but no `@local`, synthesize a handle and rewrite the declaration plus all its references to it (mirroring the existing recursive-handle branch), so `to-eo-tree` prints a consistent `>> name` rather than an anonymous declaration with dangling `vL_P` references.

yegor256:
Closes #5794.

## Problem

An anonymous, file-local recursive formation — one the parser auto-names `a🌵…` with no `>> name` handle — reached from the outside only through a method dispatch bound to a named attribute (`q. > @`) printed lossily. This program on `master`:

```
+package foo

[] > bar
  q. > @
    [n] >>
      $ (n.minus 1) > @
    5
```

became:

```
+package foo

[] > bar
  v5_4.q 5 > @
  $ > [n] >>
    n.minus 1
```

`v5_4` named no binding, so the reference dangled and the file no longer re-parsed to the same graph.

## Root cause

`merge-monikers` already inlines a dispatched cactus binding as the receiver of a reversed dispatch (#5782), which is exactly the spelling this program uses (`q. > @`). But it skipped this case on two counts:

1. `eo:moniker-binding` required a `@base`, so it hosted only **based** bindings — never an **abstract formation**.
2. Hostability rejected **named** references, so a dispatch bound to `> @` stayed expanded.

With neither pass handling it, the formation reached `to-eo-tree` standalone: its cactus name was rewritten to a synthetic `vL_P` on the reference while the declaration printed anonymously — the two disagreed and the reference dangled.

## Fix (`merge-monikers.xsl`)

Both restrictions are lifted:

- An **abstract formation** is now an eligible `eo:moniker-binding` (not just a based binding).
- A **single-segment dispatch may host onto a named reference**, and the reversed dispatch keeps the reference's own `@name`.

So the anonymous formation is inlined back as the receiver of the reversed dispatch `q. > @`, its `5` argument preserved, and no dangling reference remains. The formation stays anonymous (`>>`) — its `$`-application decoratee is laid out in the printer's canonical compact form:

```
+package foo

[] > bar
  q. > @
    $ > [n] >>
      n.minus 1
    5
```

This re-parses to the same graph and reprints to itself. The `moniker-dispatch` case (#5782, unnamed dispatch onto a based binding) is unaffected — an absent `@name` is preserved as absent.

## Tests

- `recursive-auto-named-handle.yaml` — end-to-end EO print-pack: the reported program prints with the formation kept inline in the reversed dispatch (no dangling `vL_P`) and round-trips (`printsToEo` + `printsToParseableEo`).
- `merge-monikers-formation.yaml` — XSL-level pack asserting `merge-monikers` hosts the abstract formation onto the named dispatch, keeps `@name` and the argument, and leaves no forward `ξ.a🌵….q` reference.

All 181 `eo-printer` tests pass (`mvn -pl eo-printer test`), including the existing `merge-monikers`, `merge-monikers-dispatch`, `recursive-local-handle`, and `auto-named-abstract` packs.

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -15,13 +15,15 @@
   binding line) parse to one and the same object graph. This pass turns
   the expanded spelling back into the moniker.
 
-  For each formation, an eligible named bound attribute (see
-  `eo:moniker-binding`) is merged onto the first bare `ξ.<name>` reference
-  reachable through no intervening formation (so the name still binds to
-  the same formation). A method-dispatch use such as `value.gte`, a
-  reference carrying arguments, or a reference that is itself a named
-  attribute (such as `temp > @`) cannot host the binding and stays a
-  reference; when no bare reference exists, the binding is left in place.
+  For each formation, an eligible bound attribute (see `eo:moniker-binding`)
+  is merged onto the first hostable reference reachable through no intervening
+  formation (so the name still binds to the same formation). A bare `ξ.<name>`
+  reference must be unnamed, but a single-segment dispatch `ξ.<name>.<seg>`
+  (#5782) may also host onto a named reference (such as `q. > @`), since the
+  reversed dispatch keeps the reference's own `@name` (#5794); a bare
+  reference carrying arguments or a multi-segment path cannot host the binding
+  and stays a reference. When no hostable reference exists, the binding is
+  left in place.
 
   A binding with several hostable references still becomes a moniker,
   landing on the first one in document order (#5739). This deliberately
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (142 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2700](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5811, opened 2026-07-23
#5803: extname of a dotfile is empty, not the whole name

### LINKED ISSUE #5803: `path.posix.extname` incorrectly treats entire dotfile name as the extension
`extname` of a **dotfile** (a name that starts with `.` and has no other dot, e.g. `.bashrc`, `.gitignore`) returns the whole name as the "extension" instead of an empty string. A leading dot marks a hidden file, not an extension separator.

### Steps to reproduce

```
path.posix ".bashrc" .extname    # ".bashrc"   expected ""
```

| probe | result |
|---|---|
| `extname(path.posix ".bashrc") .eq ".bashrc"` | **passes** → the actual value is `".bashrc"` |
| `extname(path.posix ".bashrc") .eq ""`        | **fails** → it is not `""` |
| `extname(path.posix "a.txt") .eq ".txt"` (control) | passes |

### Why it is wrong

For `.bashrc` the basename is `.bashrc` and `extname` returns `.bashrc` as well — i.e. `extname == basename`, which implies the file has an **empty name** and a `.bashrc` extension. That is nonsensical: a dotfile's name has no name/extension split. Node's `path.extname('.bashrc')` and Python's `os.path.splitext('.bashrc')` both treat the leading dot as part of the name and return `""`.

### Root cause

`posix.extname` slices from the **last** `.` to the end, guarding only "no dot at all" and "empty base" ([`path.eo#L62-L76`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L62-L76)):

```
[] > extname
  if. > @
    or.
      base.size.eq 0
      idx.eq -1
    ""
    string ...
  basename > base!
  string.last-index-of txt "." > idx!
```

For `.bashrc` the last (and only) dot is at index 0, so `idx = 0` (not `-1`) and the slice `[0 : end]` returns the entire name. A leading dot at index 0 must be treated as "no extension". `win32.extname` ([`path.eo#L176-L189`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L176-L189)) has the same defect.

### Suggested fix

Treat a dot at position 0 of the basename as "no extension" — extend the guard from `idx.eq -1` to also cover `idx.eq 0`:

```
or.
  base.size.eq 0
  idx.lte 0
```

gemshrine:
Fixes #5803 and #6067.

`extname` treated a leading dot as an extension separator, so a dotfile such as `.gitignore` reported its whole name as the extension, and `..` reported `.`. A leading dot marks a hidden file, not an extension, and `..` is a parent-directory reference with no extension.

Fix: the shared `impl.extname` (used by both posix and win32) now also returns an empty extension when the last dot sits at index 0 of the basename (a hidden file with no extension) or when the basename is `..`. A hidden file that carries a real extension keeps it.

Examples:

* `path.posix ".gitignore" .extname` is now `""` (was `".gitignore"`)
* `path.posix ".." .extname` is now `""` (was `"."`)
* `path.posix ".config.json" .extname` stays `".json"`
* `path.posix "a.txt" .extname` stays `".txt"`

Added regression tests for the dotfile, hidden-file-with-extension and `..` cases. Ran `mvn -pl eo-runtime test -Dtest=EOpathTest` (50/50).


eo-runtime/src/main/eo/path.eo
```diff
@@ -63,7 +63,8 @@
       if. > @
         or.
           base.size.eq 0
-          idx.eq -1
+          (idx.eq -1).or
```
morphqdd: This doesn't cover a name that is just two dots. Traced by hand: `basename("..")` = `".."` (no separator found), then `last-index-of "."` on `".."` returns index 1, not 0, so this guard misses it and falls through to the slice branch, producing `"."`. Real Node returns `""` for `..`, not `"."`. Given this PR's own rationale (a leading dot marks a hidden file, not an extension separator), a name that is entirely dots seems like it should get the same treatment - still wrong after this fix. Same applies to the win32 copy at line 180.

morphqdd: The linked issue's own suggested fix was to simply widen the guard to `idx.lte 0` (equivalent, since `last-index-of` only ever returns -1 or a non-negative index). This PR instead introduces `(idx.eq -1).or / idx.eq 0`, which is functionally the same but more verbose and diverges from the maintainer's own suggested minimal-diff fix without explanation. Same for the win32 copy at line 180. Suggest simplifying to `idx.lte 0` in both places unless there's a reason to keep -1/0 distinguished.

morphqdd (commented):
Traced the fix by hand and cross-checked against real Node.js output (`node -e "require('path').posix.extname(...)"` on 12 inputs) as ground truth. The fix is correct for the case the issue describes: `basename(".bashrc")` has no separator so `last-index-of "."` returns 0, and the widened guard now catches that and returns `""` instead of the whole filename. The two new tests are genuine regressions (they'd fail against the pre-fix code, which only checked `idx.eq -1`). `.`, `...`, `....`, `file.`, `..bashrc`, `.tar.gz`, `archive.tar.gz` all match real Node behavior post-fix. One real gap and one simplification opportunity below.

morphqdd (changes requested):
Escalating my two inline comments above from earlier: the fix is correct for the exact case the issue describes, but the "."/".." case is still wrong after this fix (`".."` returns `"."` instead of `""`, verified against real Node.js behavior), and it falls under the same "leading dot is not an extension separator" rationale this PR itself argues for. Requesting changes to either extend the guard to also cover this case or explicitly scope it out with a documented reason, rather than leaving it silently incomplete.

eo-runtime/src/main/eo/path.eo
gemshrine: Good catch, and thanks for cross-checking against real Node output — you're right, `..` slipped through. Fixed in the latest push: the guard in both the posix and win32 `extname` now also returns `""` when the basename is exactly `..` (`base.eq ".."`), so it matches `path.extname("..") === ""`. Only the exact two-dot name is special-cased — `...`, `....`, etc. still slice to `"."`, matching Node's own behavior. Added two regressions: `tests-takes-no-extname-from-posix-dotdot` and `tests-takes-no-extname-from-win32-dotdot` (both fail against the pre-fix guard, pass now).

gemshrine: I deliberately avoided `idx.lte 0` here. `.lte` (like `.lt`/`.gt`/`.gte`) is not a direct number method — it's an object defined in the `number/` subpackage (`[n x] > lte`), so it only resolves inside package `number` or via an alias. `path.eo` is neither, and I verified empirically that `idx.lte 0` on a `last-index-of` result throws at runtime (`ExFailure ... bytes.eq at .bytes:28`), which breaks all four existing `extname` tests. `.eq` is a direct method and is the only comparison that works here — that's why the guard is an `.eq`-chain rather than the issue's suggested `.lte 0`. The `basename` sibling in the same file uses the same `idx.eq 0` idiom for exactly this reason, and extending to `..` via `base.eq ".."` keeps everything within the safe `.eq`-only set.

gemshrine:
@yegor256 the red simian check here is not from this PR, it is red on master itself right now. The commit that renamed EOsocketTest.java to SyscallTest.java left the simian exclude pointing at the old filename, so a pre-existing (excluded) duplication now fails the build for every PR rebased onto the tip. I filed #6113 with the root cause and a one-line fix. Once that lands and I rebase, simian here goes green. Everything else on this PR is passing.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (23) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1105](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (2) comments; -4 for too few (23) hits-of-code; +1 to give you at least something.  Your running score is [+2796](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5812, opened 2026-07-23
#5810: keep based `&gt;&gt; name` handle and its arguments during auto-fix

### LINKED ISSUE #5810: eo-printer auto-fix mode erroneously omits handle and argument of a based `>> name` binding
Running the printer in auto-fix mode over a file-local `>>` handle whose value is a plain reference mangles it. Given

```
[a] > foo
  42.plus b > @
  a >> b
```

`Xmir.toEO` (exercised by `XmirTest#printsToEo`) emits

```
[a] > foo
  42.plus > @
    a >>
```

— the handle `b` and the `.plus` argument are both lost. The parser stores the binding as `<o base="ξ.a" name="a🌵5-2" local="b"/>` (a cactus auto-name carrying the handle in `@local`) and resolves the `b` reference to `ξ.a🌵5-2`.

`restore-local-names.xsl:76` drops `@local` for every non-void, non-recursive binding, on the assumption (stated in its own header) that a non-void `>>` handle always sits on an anonymous formation that `inline-cactoos` removes. A *based* binding (`base="ξ.a"`) is neither void nor a formation, so nothing inlines it and `b` disappears. `merge-monikers.xsl` then folds the single-use cactus binding into its bare reference but copies `$binding/@*[name() != 'as']` (`eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl:124`), keeping `@name="a🌵5-2"`. The inlined argument therefore becomes a spurious *named* node `<o base="ξ.a" name="a🌵5-2"/>`, which `to-eo-tree` prints as a standalone `a >>` line instead of `.plus`'s argument.

This is the expected output:

```
[a] > foo
  42.plus a! > @
```

In `merge-monikers.xsl`, the bare-reference branch (lines 121-127) should exclude the binding's `@name` (and `@local`) from the copied attributes, so the hosted value lands as an anonymous argument rather than a named node. That single exclusion turns the malformed `a >>` line back into the intended inline reference.

yegor256:
Fixes #5810.

## Problem

Running the printer in auto-fix mode over a file-local `>> ` handle whose value is a plain reference mangled it. Given:

```
[a] > foo
  42.plus b > @
  a >> b
```

`Xmir.toEO` emitted:

```
[a] > foo
  42.plus > @
    a >>
```

— the handle `b` and the `.plus` argument were both lost.

## Root cause

The parser stores `a >> b` as an obfuscated cactus binding (`a🌵…`) carrying the handle in `@local` and a `ξ.a` base, and resolves every `b` reference to the cactus name. Unlike a `? >> name` void or an anonymous `>>` formation, a based binding is neither void nor abstract, so `restore-local-names` drops its `@local` and the binding survives into the folding passes as a single-use cactus binding.

In the full print pipeline the fold is done by **`inline-cactoos`** (which runs before `merge-monikers`); `merge-monikers` exhibits the same latent bug in isolation. Both copied the binding's obfuscated cactus `@name` onto the inlined value. Because the hosting reference is unnamed, this produced a spurious *named* node, which `to-eo-tree` prints as its own `a >>` line instead of as `.plus`'s argument.

> Note: the issue pins the fix to `merge-monikers`, but in the assembled pipeline `inline-cactoos` folds the binding first, so the fix had to be applied there too (and consistently in both sheets).

## Fix

Drop the cactus `@name` (and the `@local` handle it leaves behind) when the inlined value is *not* something whose name is still meaningful downstream:

- an **abstract formation** keeps its name so `to-eo-tree` can render the anonymous `[...] >>` marker;
- a **dataized-const wrapper** (`.as-bytes` over `Φ.dataized`) keeps its name so `dataized-to-const` can promote it onto the abstract const node it rebuilds;
- every other **based value** lands as an anonymous argument.

Result:

```
42.plus a > [a] > foo
```

## Changes

- `eo-printer/.../inline-cactoos.xsl` — drop `@name`/`@local` for based inlined values, with an exception for abstract formations and dataized-const wrappers.
- `eo-printer/.../merge-monikers.xsl` — same exclusion in the bare-reference branch (keeps `@name` for abstract formations).
- `eo-printer/.../print-packs/yaml/based-local-handle.yaml` — new print-pack reproducing the issue (reprints to a fixpoint).
- `eo-printer/.../eo-packs/print/merge-monikers.yaml` — updated the merged-node assertion to expect the anonymous value.

All 184 `eo-printer` tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_018fHmUx3xd8UbGUS5iJtgvh)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -40,17 +40,31 @@
   copy that self-reference back in and fire this template forever, so
   such a target is also left untouched (both the reference and the
   abstraction stay in place).
+
+  The inlined value keeps the target's obfuscated cactus `@name` only when
+  the name is still meaningful downstream: an abstract formation, whose name
+  `to-eo-tree` renders as the anonymous `[...] >>` marker (exactly what an
+  inlined anonymous formation should read as), and a dataized-const wrapper
+  (`.as-bytes` over `Φ.dataized`), whose name `dataized-to-const` later
+  promotes onto the abstract const node it rebuilds. A based `>> name` handle
+  whose value is a plain reference or application (`a >> b`, R-3.10.12) is
+  neither, so carrying its `@name` (or the `@local` handle it leaves behind)
+  over would turn the inline into a spurious named node that `to-eo-tree`
+  prints as its own `a >>` line, swallowing the surrounding argument; for
+  such a target the `@name` and `@local` are dropped and the value lands as
+  an anonymous argument (#5810).
   -->
   <xsl:template match="o[contains(@base, concat('.', $auto))]" priority="0">
     <xsl:variable name="name" select="eo:resolved-name(@base)"/>
     <xsl:variable name="target" select="ancestor::o/o[@name=$name][1]"/>
+    <xsl:variable name="keep-name" as="xs:boolean" select="exists($target) and (eo:abstract($target) or ($target/@base = '.as-bytes' and $target/o[1]/@base = 'Φ.dataized'))"/>
     <xsl:choose>
       <xsl:when test="exists($target) and not(eo:void($target)) and not(eo:recursive($target, $name))">
         <xsl:element name="o">
           <xsl:if test="@as">
             <xsl:apply-templates select="@as"/>
           </xsl:if>
-          <xsl:apply-templates select="$target/@*"/>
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2724](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5813, opened 2026-07-23
#5802: dirname of a root-anchored single component is the root

### LINKED ISSUE #5802: `path.posix.dirname` incorrectly returns `""` for single-component absolute path instead of `"/"`
`dirname` of an **absolute path with a single component** returns an empty string `""` instead of the root directory. The parent of `/foo` is `/`, not "nothing".

### Steps to reproduce

```
path.posix "/foo" .dirname     # ""   expected "/"
```

| probe | result |
|---|---|
| `dirname(path.posix "/foo") .eq ""`   | **passes** → the actual value is `""` |
| `dirname(path.posix "/foo") .eq "/"`  | **fails** → it is not `"/"` |
| `dirname(path.posix "/var/www") .eq "/var"` (control) | passes |

### Why it is wrong

`path.posix "/foo"` is absolute (`.is-absolute` is true), but its `dirname` `""` is not — the operation silently drops the leading-root information, so `dirname` + `basename` no longer reconstructs the original (`"" + "/" + "foo"` ≠ `"/foo"`). Node's `path.dirname('/foo')`, Python's `os.path.dirname('/foo')` and the POSIX `dirname` utility all return `"/"`.

### Root cause

`posix.dirname` slices `txt[0 : last-index-of(separator)]` ([`path.eo#L52-L61`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L52-L61)):

```
[] > dirname
  string > @
    if.
      (pth.size.eq 0).or
        len.eq -1
      pth
      (txt.slice 0 len).as-bytes
  string.last-index-of txt separator > len!
```

For `/foo` the only separator is at index 0, so `len = 0` and `txt.slice 0 0 = ""`. When the last separator sits at index 0 the parent is the root and the result should be the separator itself, not the empty slice. `win32.dirname` ([`path.eo#L166-L175`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L166-L175)) has the same shape (e.g. `path.win32 "\foo" .dirname` collapses to `""` too, and `path.win32 "C:\foo" .dirname` yields `"C:"` instead of `"C:\\"`).

### Suggested fix

When `len == 0` (the separator is the first byte), return the separator (root) rather than the empty prefix:

```
if.
  len.eq 0
  separator
  (txt.slice 0 len).as-bytes
```

gemshrine:
Fixes #5802.

For an absolute path with a single component (e.g. `/foo`), the only separator is at index 0, so `dirname` sliced `txt[0:0]` and returned `""` — losing the fact that the parent is the root directory. A `dirname` that drops the root can no longer be combined with `basename` to reconstruct the path.

Fix: when the last separator is at index 0 (`len.eq 0`), return the separator (the root) instead of the empty prefix, in both the posix and win32 `dirname`. Now:

- `path.posix "/foo" .dirname` -> `"/"` (was `""`)
- `path.win32 "\foo" .dirname` -> `"\"` (was `""`)
- `path.posix "/var/www" .dirname` -> `"/var"` (unchanged)

Added regression tests for the posix and win32 root cases.

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -55,7 +55,10 @@
           (pth.size.eq 0).or
             len.eq -1
           pth
-          (txt.slice 0 len).as-bytes
+          if.
+            len.eq 0
+            separator
+            (txt.slice 0 len).as-bytes
       string.last-index-of txt separator > len!
       uri > pth!
       string pth > txt
@@ -169,7 +172,10 @@
             (pth.size.eq 0).or
               len.eq -1
             pth
-            (txt.slice 0 len).as-bytes
+            if.
+              len.eq 0
+              separator
+              (txt.slice 0 len).as-bytes
         string.last-index-of txt separator > len!
         uri > pth!
         string pth > txt
@@ -472,6 +478,16 @@
           "html"
           path.separator
     ""
+
+  eq. ++> tests-returns-root-dirname-of-posix-single-component
+    dirname.
+      path.posix "/foo"
+    "/"
+
+  eq. ++> tests-returns-root-dirname-of-win32-root-relative
+    dirname.
+      path.win32 "\\foo"
+    "\\"
 
   eq. ++> tests-returns-the-same-string-if-no-separator-is-found
     basename.


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+539](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5809, opened 2026-07-23
Normalization of drive-relative Windows paths incorrectly removes leading '..' in path.win32


gemshrine:
When a Windows path is **drive-relative without a root** (e.g. `C:..\foo` — relative to the current directory *on drive C:*), `normalized` drops the leading `..`, turning it into a semantically different path. A pure relative path (`..\foo`) keeps its `..` correctly, so the two disagree.

### Steps to reproduce
```
path.win32 "C:.."      .normalized   # "C:"       expected "C:.."
path.win32 "C:..\\foo"  .normalized   # "C:foo"     expected "C:..\\foo"
path.win32 "..\\foo"    .normalized   # "..\\foo"   (correct - pure relative keeps "..")
```

Verified against the runtime (transpiled `EOpathTest` with temporary `++>` probes):

| probe | result |
|---|---|
| `normalized(path.win32 "C:..") .eq "C:"`            | **passes** -> actual is `"C:"` |
| `normalized(path.win32 "C:..") .eq "C:.."`          | **fails** |
| `normalized(path.win32 "C:..\\foo") .eq "C:foo"`    | **passes** -> actual is `"C:foo"` |
| `normalized(path.win32 "C:..\\foo") .eq "C:..\\foo"` | **fails** |
| `normalized(path.win32 "..\\foo") .eq "..\\foo"` (control) | passes |

### Why it is wrong
`C:..\foo` means "the parent of the current directory *on drive C:*, then `foo`". Because the current directory on C: is unknown, the `..` cannot be resolved and must be preserved - exactly as it is for the driveless relative path `..\foo`. Collapsing `C:..\foo` to `C:foo` changes the meaning. Node's `path.win32.normalize('C:..\\foo')` returns `'C:..\\foo'`.

### Root cause
The win32 reduce keeps a leading `..` only when the path is neither root-relative **nor drive-relative** ([`path.eo#L241-L255`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L241-L255)):

```
if. > [accum segment] >>
  segment.eq ".."
  if.
    (accum.length.gt 0).and (accum.head.eq "..").not
    accum.tail
    if.
      is-root-relative.not.and is-drive-relative.not   # drops ".." for ANY drive-relative path
      accum.with segment
      accum
  ...
```

[...]

Issue #5808, opened 2026-07-23
`path.win32 "C:" .normalized` incorrectly drops drive letter and returns `"."`


gemshrine:
A **drive-only** Windows path such as `C:` (drive with no following component) normalizes to `.`, silently **dropping the drive letter**, even though the same object correctly preserves the drive for every neighbouring case (`C:.`, `C:foo`).

### Steps to reproduce
```
path.win32 "C:"    .normalized    # "."      expected "C:"
path.win32 "C:."   .normalized    # "C:"     (correct)
path.win32 "C:foo" .normalized    # "C:foo"  (correct)
```

Verified against the runtime (transpiled `EOpathTest` with temporary `++>` probes):

| probe | result |
|---|---|
| `normalized(path.win32 "C:") .eq "."`      | **passes** -> the actual value is `"."` |
| `normalized(path.win32 "C:") .eq "C:"`     | **fails** |
| `normalized(path.win32 "C:.")   .eq "C:"` (control)   | passes |
| `normalized(path.win32 "C:foo") .eq "C:foo"` (control) | passes |

### Why it is wrong
`C:` means "the current directory on drive C:" and is a distinct location from `.` (current directory on the current drive). Normalizing it to `.` throws the drive away. Node's `path.win32.normalize('C:')` returns `'C:'`. The object clearly *can* keep the drive — it does so for `C:.` and `C:foo` — so the drive-only case is an inconsistency, not a design choice.

### Root cause
`win32.normalized` strips the two-character drive prefix into `driveless` and then short-circuits an empty `driveless` straight to `"."`, without re-attaching the drive ([`path.eo#L213-L235`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L213-L235)):

```
is-drive-relative.if > driveless!
  ubts.slice
    2
    ubts.size.plus -2      # "C:" -> slice 2 0 -> ""
  ubts
...
as-bytes. > normalized
  if.
    driveless.size.eq 0    # empty driveless (drive-only path) -> ".", drive prefix lost
    "."
    concat. ...            # the else-branch is the ONLY place the "C:" prefix is re-attached
```

For `C:` the `driveless` is empty, so the `"."` branch fires and the `is-drive-relative.if ... uri.slice 0 2` prefix (which would restore `C:`) in the else-branch is never reached.

### Suggested fix
[...]

PR #5814, opened 2026-07-23
#5801: a relative path that normalizes away becomes "."

### LINKED ISSUE #5801: `path.normalized` returns empty string instead of `"."` for a collapsed relative path
When a **relative** path collapses to nothing under normalization (every segment is cancelled by `..` or dropped as `.`/empty), `path.normalized` returns an empty string `""` instead of `"."` (the current directory).

### Steps to reproduce

```
path.posix "foo/.."      .normalized   # ""   expected "."
path.posix "."           .normalized   # ""   expected "."
path.posix "a/b/../.."   .normalized   # ""   expected "."
path.win32 "foo\.."     .normalized   # ""   expected "."
```

| probe | result |
|---|---|
| `normalized(path.posix "foo/..") .eq ""`  | **passes** → the actual value is `""` |
| `normalized(path.posix "foo/..") .eq "."` | **fails** → it is not `"."` |
| `normalized(path.posix ".")      .eq ""`  | passes |
| `normalized(path.win32 "foo\..")  .eq ""`  | passes |
| `normalized(path.posix "")       .eq "."` (control) | passes → an empty *input* is correctly `"."` |

### Why it is wrong

Node's `path.normalize('foo/..')`, Python's `os.path.normpath('foo/..')` and the POSIX `dirname/realpath` convention all yield `"."`. More importantly, the object is **internally inconsistent**: an empty *input* is already special-cased to `"."`, so a path that *reduces* to empty must produce the same current-directory marker, not `""` (which is falsy/meaningless as a path).

### Root cause

In `posix.normalized` the `"."` special-case only guards the empty *input* ([`path.eo#L89-L97`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L89-L97)):

```
as-bytes. > normalized
  if.
    uri.length.eq 0        # only the empty INPUT is mapped to "."
    "."
    concat.
      is-absolute.if
        separator.concat path
        path
      trailing.if separator --
```

When the `tuple.reduced` over the segments ([`path.eo#L98-L116`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L98-L116)) cancels everything, `path` becomes `""`; for a relative path the concat is then `"" + <no trailing> = ""`, and nothing restores it to `"."`. `win32.normalized` has the same gap — it guards `driveless.size.eq 0` but not the *reduced-to-empty* result ([`path.eo#L222-L235`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L222-L235)).

### Suggested fix

Map an empty normalized result to `"."` in the wrapper that already post-processes the result (`"//" -> "/"`, [`path.eo#L82-L87`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L82-L87)), for both `posix` and `win32`:

```
posix > @
  if.
    normalized.eq "//"
    "/"
    if.
      normalized.size.eq 0
      "."
      string normalized
```

gemshrine:
Fixes #5801.

When the `..`/`.` segment reduction cancels everything (e.g. `foo/..`, `.`, `a/b/../..`), the joined path is empty and, for a relative path, `normalized` returned `""` — a falsy, meaningless path. An empty *input* is already special-cased to `"."`, so a path that *reduces* to empty must produce the same current-directory marker.

Fix: map an empty normalized result to `"."` in the posix wrapper that already post-processes `"//"` -> `"/"`. Now:

- `path.posix "foo/.." .normalized` -> `"."` (was `""`)
- `path.posix "." .normalized` -> `"."`
- `path.posix "/foo/bar/../.." .normalized` -> `"/"` (unchanged)

Added a regression test for the collapsing relative path.

(The win32 branch has related empty/collapse gaps tracked separately in #5808 and #5809.)

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -84,7 +84,10 @@
         if.
           normalized.eq "//"
           "/"
-          string normalized
+          if.
+            normalized.size.eq 0
+            "."
+            string normalized
       ^.is-absolute.as-bool > is-absolute
       as-bytes. > normalized
         if.
@@ -354,6 +357,11 @@
     normalized.
       path.posix "/foo/bar/.//./baz//../x/"
     "/foo/bar/x/"
+
+  eq. ++> tests-normalizes-posix-path-collapsing-to-current-dir
+    normalized.
+      path.posix "foo/.."
+    "."
 
   eq. ++> tests-normalizes-posix-relative-path
     normalized.


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+551](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5817, opened 2026-07-23
Parser incorrectly rejects `!` const marker for named file-local `&gt;&gt; name` handle


yegor256:
The parser accepts a `!` const marker on an anonymous file-local handle (`>>!`, exercised by the `const-auto-named` print-pack as `5 > [] >>!`) and on a plain named binding (`> name!`, e.g. `getenv "TEMP" > temp!`), but rejects it on a **named** file-local handle (`>> name!`). Given:

```
[a] > foo
  42.plus b > x
  a >> b!
```

parsing fails with:

```
[3:8] error: 'trailing garbage after name suffix'
  a >> b!
        ^
```

A `>> name!` handle should be as valid as `> name!` — a const, file-local handle.

## Root cause

`Suffix.auto` (`eo-parser/src/main/java/org/eolang/parser/Suffix.java:377-412`) parses the `>>` suffix in the order: optional `!` **first** (right after `>>`, lines 380-385), then the optional handle name (lines 386-402), then `endsClean` which rejects anything trailing (line 410). So it accepts `>>!` and `>> name`, but for `>> name!` the `!` arrives *after* the name — past the point where the `!` is looked for — and `endsClean` reports it as trailing garbage.

The sibling `Suffix.named` (`> name`, lines 423-467) has it right: it reads the name first (lines 433-437), then checks for a trailing `!` (lines 444-448). The `auto` path is simply missing the symmetric post-name check.

## Suggested fix

In `Suffix.auto`, after the handle name is read (after line 402), accept an optional trailing `!` before `endsClean`, mirroring `Suffix.named`:

```java
if (!cnst && rest < tail.length() && tail.charAt(rest) == '!') {
    cnst = true;
    rest = rest + 1;
}
```

The handle-reading loop already stops at `!` (it is a `terminates` boundary), so `rest` points exactly at the `!`. This keeps the existing `>>!` (anonymous const) and `>> name` (non-const) spellings working while adding `>> name!`.

## Notes

Surfaced while extending the `eo-printer` auto-fix work from #5810: a const file-local handle round-trips through the printer only once its source spelling parses. A fix for this is included in the follow-up PR to #5810.

PR #5818, opened 2026-07-23
#5810: reformat const file-local `&gt;&gt; name!` handle without mangling it

yegor256:
Follow-up to #5810, for the const case. Reported the parser half as #5817.

## Problem

Auto-fixing a const file-local handle mangles it. Given:

```
[a] > foo
  42.plus b > x
  a >> b!
```

there were two coupled defects:

1. **Parser** (#5817): `a >> b!` didn't even parse — `[3:8] error: 'trailing garbage after name suffix'`.
2. **Printer**: once it parses, the printer emitted `42.plus > x` / `a >>!`, losing the `.plus` argument (the same shape as the original #5810 bug).

## Fix

**Parser — `Suffix.auto` (`eo-parser/.../Suffix.java`).** The `>>` parser looked for the `!` const marker only *immediately after* `>>` (so `>>!` and `>> name` worked, but `>> name!` did not). The `> name!` parser already reads the name first and then the marker; this mirrors that by accepting a trailing `!` after the handle. The handle-reading loop already stops at `!` (a name terminator), so the marker sits exactly where the new check looks. `>>!` and `>> name` are unchanged.

**Printer — `dataized-to-const.xsl`.** A const handle is lowered to a `.as-bytes` over `Φ.dataized` wrapper carrying the obfuscated cactus name; `inline-cactoos` folds the single-use wrapper onto its bare reference and `dataized-to-const` rebuilds the `@const`. It *always* promoted the wrapper's name onto the rebuilt node — but for a based value that's an obfuscated cactus name, leaving a spurious named node that `to-eo-tree` prints as its own `a >>!` line. Now the name is promoted only when it's meaningful: a real author name (`f > some!`) or an abstract formation (`[...] >>!`); a cactus name on a based value is dropped, so the value lands as an anonymous argument.

## Result

```
[a] > foo
  42.plus a > x
```

Valid EO, re-parses to itself. An inline anonymous const reference (`a!`) is not valid EO, so — as with the non-const handle that folds to a bare `a` in #5810 — the const marker does not survive the fold.

## Changes

- `eo-parser/.../Suffix.java` — accept the `!` const marker after a named `>>` handle.
- `eo-parser/.../SuffixTest.java` — `>> fibo!` unit test.
- `eo-parser/.../eo-syntax/const-local-handle.yaml` — parse pack for `>> name!`.
- `eo-printer/.../dataized-to-const.xsl` — drop a cactus name on a based const value.
- `eo-printer/.../print-packs/yaml/const-local-handle.yaml` — round-trip print pack.

All 1499 `eo-parser` and 185 `eo-printer` tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_018fHmUx3xd8UbGUS5iJtgvh)_

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -367,6 +367,11 @@ final class Suffix {
     /**
      * Parse a {@code >>} (auto) suffix, optionally carrying a trailing
      * file-local handle ({@code >> name}, §3.10) kept as the label.
+     *
+     * <p>The optional {@code !} const marker (§3.10) may sit either right
+     * after {@code >>} on an anonymous handle ({@code >>!}) or right after a
+     * named handle ({@code >> name!}), mirroring the {@code > name!} spelling
+     * of a plain named binding. Only one marker is accepted.</p>
      * @param tail Tail substring
      * @param after Index immediately after {@code >>}
      * @param span Source span
@@ -399,6 +404,10 @@ final class Suffix {
                 );
             }
             rest = end;
+        }
+        if (!cnst && rest < tail.length() && tail.charAt(rest) == '!') {
+            cnst = true;
+            rest = rest + 1;
         }
         final int trailing = Suffix.skipSpace(tail, rest);
         if (trailing < tail.length() && tail.charAt(trailing) == '/') {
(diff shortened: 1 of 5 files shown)


yegor256:
Note on the dropped `const`: this PR intentionally prints the folded handle as `42.plus a > x` (without `!`). Preserving it as `42.plus a! > x` requires the `!` const marker on an inline argument, which is not valid EO today and needs a rework of the const lowering/float/auto-naming passes for anonymous arguments plus inline rendering in the printer. That is tracked as a separate feature in #5821.

---

yegor256:
Closing as superseded. The two issues this PR spun off — #5817 (accept `!` on a named `>> name` handle) and #5821 (support `!` on inline arguments) — were implemented and merged into `master` (#5827 and #5823). With both on `master`, `a >> b!` now reformats to `42.plus a! > x` (verified, re-parses to itself), preserving the `const` — which is strictly better than this PR's `42.plus a > x` (where the `!` was dropped because an inline const reference wasn't expressible at the time). This PR's parser change duplicates #5827 and its printer behavior is now outdated, so there's nothing left to merge.

---

PR #5819, opened 2026-07-23
#5804: parse a bracketed IPv6 host in the uri object

### LINKED ISSUE #5804: `uri` object incorrectly parses bracketed IPv6 host resulting in invalid host and port values
The `uri` object claims to "parse an RFC 3986 generic URI" (file header), but for a URL whose host is a **bracketed IPv6 literal** — which RFC 3986 §3.2.2 defines as part of the `host` grammar (`IP-literal = "[" ( IPv6address / IPvFuture ) "]"`) — it silently returns garbage instead of the host and port.

### Steps to reproduce

```
uri "http://[::1]:8080/p" .host    # "["   expected "[::1]" (or "::1")
uri "http://[::1]:8080/p" .port    # not "8080"
```

Verified against the runtime (transpiled `EOuriTest` with temporary `++>` probes):

| probe | result |
|---|---|
| `host(uri "http://[::1]:8080/p") .eq "["`     | **passes** → `host` is literally `"["` |
| `host(uri "http://[::1]:8080/p") .eq "[::1]"` | **fails** |
| `port(uri "http://[::1]:8080/p") .eq "8080"`  | **fails** |
| `host(uri "http://localhost:899/p") .eq "localhost"` (control) | passes |

### Root cause

`host`/`port` are split on the **first** `":"` in the authority ([`uri.eo#L76-L80`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/uri.eo#L76-L80), [`uri.eo#L103-L105`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/uri.eo#L103-L105)):

```
has-port.if > host
  host-port.slice
    0
    port-colon-idx
  host-port
...
string.index-of > port-colon-idx
  host-port
  ":"
```

For `[::1]:8080` the first `":"` sits **inside** the IPv6 literal (index 1), so `host` becomes `"["` and `port` becomes the leftover `":1]:8080"`. The colon that separates the port must be located **after** the closing `"]"`, not simply the first colon in the authority.

### Why this is a bug and not just a missing feature

The object silently produces a *wrong* value for a syntactically valid, in-scope input rather than erroring — a caller reading `.host` gets `"["` with no indication anything went wrong. Bracketed IPv6 is a normal, common URL form (`http://[::1]:8080/`, `https://[2001:db8::1]/`) and is squarely within the RFC 3986 grammar the object advertises.

### Suggested fix

When the authority (after any `userinfo@`) starts with `"["`, take `host` as the substring through the closing `"]"` (inclusive) and look for the port `":"` only in the remainder after `"]"`:

```
# pseudo:
if host-port starts-with "[":
    close = index-of host-port "]"
    host  = host-port.slice 0 (close + 1)
    rest  = host-port.slice (close + 1) ...
    port  = if rest starts-with ":" then rest.slice 1 ... else ""
else:
    <current first-colon logic>
```

gemshrine:
Fixes #5804.

The port separator was taken as the *first* `:` in the authority. For a bracketed IPv6 literal (`[::1]:8080`) that colon sits inside the address, so `host` became `"["` and `port` became the leftover `":1]:8080"` — a silently wrong result for a valid RFC 3986 host.

Fix: when the host part starts with `[`, locate the port `:` **after** the closing `]`; plain `host:port` parsing is unchanged (`host`, `port` and `has-port` all derive from that index). Now:

- `uri "http://[::1]:8080/p" .host` -> `"[::1]"` (was `"["`)
- `uri "http://[::1]:8080/p" .port` -> `"8080"`
- `uri "http://[2001:db8::1]/p" .host` -> `"[2001:db8::1]"`
- `uri "pgsql://localhost:899" .host` / `.port` -> `"localhost"` / `"899"` (unchanged)

Added regression tests for the bracketed IPv6 host, with and without a port.

### DIFF
--- a/eo-runtime/src/main/eo/uri.eo
+++ b/eo-runtime/src/main/eo/uri.eo
@@ -45,6 +45,14 @@
       0
       hash-idx
     rest
+  string.index-of > bracket-end-idx
+    host-port
+    "]"
+  bracket-end-idx.plus 1 > bracket-start
+  host-port.slice > bracket-tail
+    bracket-start
+    host-port.length.minus
+      number bracket-start
   has-fragment.if > fragment
     rest.slice
       hash-idx.plus 1
@@ -57,6 +65,9 @@
     "//"
   not. > has-authority-path
     authority-path-idx.eq -1
+  string.starts-with > has-bracket-host
+    host-port
+    "["
   not. > has-fragment
     hash-idx.eq -1
   not. > has-port
@@ -100,9 +111,14 @@
         number
           port-colon-idx.plus 1
     ""
-  string.index-of > port-colon-idx
-    host-port
-    ":"
+  has-bracket-host.if > port-colon-idx
+    if.
+      rel-colon.eq -1
+      -1
+      rel-colon.plus bracket-start
+    string.index-of
+      host-port
+      ":"
   has-query.if > query
     before-fragment.slice
       question-idx.plus 1
@@ -113,6 +129,9 @@
   string.index-of > question-idx
     before-fragment
     "?"
+  string.index-of > rel-colon
+    bracket-tail
+    ":"
   text.slice > rest
     scheme-colon.plus 1
     text.length.minus
@@ -132,6 +151,21 @@
   string.index-of > userinfo-at-idx
     authority
     "@"
+
+  eq. ++> tests-parses-bracketed-ipv6-host-with-port
+    host.
+      uri "http://[::1]:8080/p"
+    "[::1]"
+
+  eq. ++> tests-parses-bracketed-ipv6-host-without-port
+    host.
+      uri "http://[2001:db8::1]/p"
+    "[2001:db8::1]"
+
+  eq. ++> tests-parses-bracketed-ipv6-port
+    port.
+      uri "http://[::1]:8080/p"
+    "8080"
 
   eq. ++> tests-parses-empty-host-of-triple-slash-uri
     host.


eo-runtime/src/main/eo/uri.eo
```diff
@@ -57,6 +65,9 @@
     "//"
   not. > has-authority-path
     authority-path-idx.eq -1
+  string.starts-with > has-bracket-host
```
morphqdd, line 68: `has-bracket-host` only checks that `host-port` starts with `"["`; it doesn't check that a matching `"]"` was actually found (`bracket-end-idx` could be -1). For an unclosed bracket like `"http://[::1/p"`, this falls into the bracket branch anyway with `bracket-start=0`, so it silently reproduces the exact bug this PR fixes (`host` becomes `"["`). Verified by tracing: `bracket-end-idx=-1` -> `bracket-start=0` -> `bracket-tail` = whole string -> `rel-colon` finds the first `:` inside the literal -> `port-colon-idx` = that index -> `host = "["`. Suggest gating `has-bracket-host` on `bracket-end-idx != -1` too, or at minimum add a test documenting the current behavior for malformed input so it's a conscious decision.

eo-runtime/src/main/eo/uri.eo
```diff
@@ -133,6 +152,21 @@
     authority
     "@"
 
+  eq. ++> tests-parses-bracketed-ipv6-host-with-port
```
morphqdd, line 155: Good coverage for the two cases in the issue, but there's no test combining a bracketed IPv6 host with `userinfo` (e.g. `uri "http://user:pass@[::1]:8080/p"`). Traced by hand and it resolves correctly (`host-port` is still correctly re-sliced before the bracket logic runs), so not a bug - just a coverage gap worth closing since `userinfo` and bracket-host are two independent code paths that only interact through `host-port`.

morphqdd (commented):
Hand-traced the index-of/slice logic against the real EO string primitives (independently re-implemented the same semantics to cross-check). Confirmed bracketed-host-with-port, bracketed-host-without-port, userinfo+bracket-host, IPv4-mapped IPv6, and zone-ID hosts all resolve correctly, since bracket detection only depends on the position of `]`, not its contents. Also confirmed the old code fails both new tests (traces to host="[" and host="[2001" respectively) - genuine regression tests, not vacuous. CI is green including the snippets job, which actually dataizes every `++>` test. One real, verified gap below.

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+1668](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+591](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5820, opened 2026-07-23
#5806: integrate exactly n subsections regardless of step

### LINKED ISSUE #5806: `number.integral` function skips last interval when step evenly divides range
`number.integral` (Simpson's rule) silently **drops the last subsection** whenever the step `(b-a)/n` is exactly representable as a double (e.g. an integer or power-of-two step). The loop advances a mutable `left` and continues only while `left + step` is **strictly** less than `b`; when the accumulated `left + step` lands exactly on `b`, the strict `<` is false and the final interval is never integrated.

### Steps to reproduce

```
integral (x > [x]) 0 10 10 .as-number     # 40.5   expected 50   (∫₀¹⁰ x dx = 50)
integral (x > [x]) 0 1  1  .as-number     # 0      expected 0.5  (n=1 integrates nothing at all)
```

Verified against the runtime (transpiled `EOintegralTest` with temporary `++>` probes):

| probe | result |
|---|---|
| `integral(x,0,10,10).as-number .eq 40.5` | **passes** → the actual value is `40.5` |
| `integral(x,0,10,10).as-number .eq 50`   | **fails** |
| `integral(x,0,1,1).as-number  .eq 0`     | **passes** → `n=1` yields `0` |

For `a=0, b=10, n=10` the step is exactly `1.0`; `left` takes the exact integer values `0,1,…,9`, and at `left=9` the guard `9+1 < 10` is false, so the interval `[9,10]` (area `9.5`) is skipped → `40.5` instead of `50`. For `n=1` the single interval `[0,1]` is skipped and the result is `0`.

### Why the existing tests pass

The three shipped tests use `n=15` (step `0.6`) and `n=100` (step `0.09`). Those steps are **not** exactly representable, so the accumulated `left+step` on the last iteration lands a hair below `b` and the final subsection sneaks in. The correctness of `integral` currently depends on floating-point rounding luck; any evenly-dividing step exposes the defect.

### Root cause

The while-loop guard uses strict `lt` on the running left edge ([`integral.eo#L20-L35`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/number/integral.eo#L20-L35)):

```
while
  if. > [i] >>
    (left.as-number.plus step).lt right      # strict <, drops the interval ending exactly at b
    seq *
      sum.put
        sum.as-number.plus
          subsection left.as-number (left.as-number.plus step)
      left.put (left.as-number.plus step)
      true
    false
  true > [i]
```

### Suggested fix

Iterate a **fixed count of `n` subsections** instead of comparing the accumulated `left` against `b`, so the number of intervals is exactly `n` regardless of floating-point representation. (Merely switching `lt` to `lte` is not enough — when the accumulation overshoots `b` by an ULP it would still drop the last interval, and when it undershoots it could add an extra one.) A count-driven loop that computes the `i`-th interval as `[a + i·step, a + (i+1)·step]` for `i = 0 … n-1` is both correct and independent of accumulation error.

gemshrine:
Fixes #5806.

The Simpson-rule loop advanced a running position and continued only while `position + step` was **strictly** less than `b`. When the step `(b-a)/n` divides evenly (an integer or power-of-two step), the position lands exactly on `b` and the strict `<` skips the final subsection — `integral (x>[x]) 0 10 10` gave `40.5` instead of `50`, and `n=1` integrated nothing (`0`). The shipped tests only passed because their steps (`0.6`, `0.09`) are not exactly representable, so accumulation error sneaked the last interval in.

Fix: drive the loop by a **count** from `0` to `n-1` and derive each interval directly as `[a + i*step, a + (i+1)*step]`. The number of subsections is now always `n`, independent of floating-point accumulation. Now:

- `integral (x>[x]) 0 10 10 .as-number` -> `50` (was `40.5`)
- `integral (x>[x]) 0 1 1 .as-number` -> `0.5` (was `0`)
- the existing cube / linear / quadratic tests still pass unchanged

Added regression tests for the evenly-dividing-step and single-partition cases.

### DIFF
--- a/eo-runtime/src/main/eo/number/integral.eo
+++ b/eo-runtime/src/main/eo/number/integral.eo
@@ -14,25 +14,23 @@
     malloc.of
       8
       [sum] >>
-        malloc.for > @
-          a
-          [left] >>
-            seq * > @
-              while
-                if. > [i] >>
-                  (left.as-number.plus step).lt right
-                  seq *
-                    sum.put
-                      sum.as-number.plus
-                        subsection left.as-number (left.as-number.plus step)
-                    left.put (left.as-number.plus step)
-                    true
-                  false
-                true > [i]
-              sum
-            b > right
-            as-number. > step
-              (right.minus left).div n
+        seq * > @
+          while
+            if. > [i] >>
+              i.lt n
+              seq *
+                sum.put
+                  sum.as-number.plus
+                    subsection
+                      a.plus (i.times step)
+                      a.plus
+                        (i.plus 1).times step
+                true
+              false
+            true > [i]
+          sum
+        as-number. > step
+          (b.minus a).div n
   times. > [a b] > subsection
     div.
       b.minus a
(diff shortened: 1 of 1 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+607](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5823, opened 2026-07-23
#5821: support `!` const marker on inline arguments

### LINKED ISSUE #5821: `!` const marker not accepted on inline arguments, causes parse error
The `!` const marker (R-9.4) is accepted on a named binding (`... > name!`) and on a file-local handle (`... >> name!`, once #5817 lands), but not on an **inline argument**. So:

```
[a] > foo
  42.plus a! > x
```

fails to parse with `[2:11] error: 'trailing garbage after expression'` at the `!`.

It would be useful to allow a const argument inline, so that a const value can be passed positionally without first floating it onto its own `... > name!` line. This also blocks a clean round-trip in the printer: a const file-local handle such as `a >> b!` (see #5810 / #5817) folds onto its use site as an anonymous const reference, which would ideally print as `42.plus a! > x`. Today the printer must instead drop the const and emit `42.plus a > x`, because there is no valid inline spelling for an anonymous const reference to reprint to.

## Why it's not a small change

I prototyped the parser side (accept a trailing `!` on an argument in `Tokens.readArgs`, carry it on `Value`, emit `@const` in `Emissions.emitArg`). The tokenisation is easy, but the const *lowering* is the blocker:

- `const-to-dataized.xsl` rewrites `@const` into a `(dataized …).as-bytes` wrapper, and `vars-float-up.xsl` then floats that wrapper up to a named attribute of the enclosing formation.
- That machinery assumes the const sits on a **named** binding. For an **anonymous argument** there is no name, and the auto-namer produces an empty one — parsing `42.plus a! > x` yields invalid XMIR: the floated binding gets `name=""` and the argument reference becomes `base="ξ."` (empty). It does not round-trip.

So a full implementation needs:

1. **Parser** — accept `!` on an inline argument (tokeniser: done in prototype), and give a floated anonymous const argument a proper cactus auto-name in `const-to-dataized` / `vars-float-up` / `AutoName` instead of an empty one.
2. **Printer** — fold the floated const binding back onto its use site (as the #5810 passes already do for the non-const case) and lay it out **inline**, and render `!` on an unnamed const node in `to-eo-tree` (a `<o base='ξ.a' const=''/>` should print as `a!`, not `a`, and not as a vertical `a!` line).
3. A print-pack proving `a >> b!` reprints to `42.plus a! > x`, plus a parser `eo-syntax` pack for `42.plus a!`.

Related: #5810 (the based `>> name` handle fold), #5817 (`>> name!` parsing).

yegor256:
Closes #5821.

Allows the `!` const marker (R-9.4) on an **inline argument**, so a const value can be passed positionally without first floating it onto its own `... > name!` line:

```
[a] > foo
  42.plus a! > x
```

Previously this failed to parse with `[2:11] error: 'trailing garbage after expression'`.

## Parser

- `Tokens.readArgs` reads a trailing `!` after the argument value (and optional inline binding).
- `Value` carries a `constant` flag; `Emissions.emitArg` emits `@const` on the argument's `<o>`.
- `const-to-dataized.xsl` synthesises a cactus auto-name (the same `a🌵<line>-<pos>` shape as `AutoName`) for a `@const` node that has no name. `vars-float-up` can then float the `(dataized …).as-bytes` wrapper up to the enclosing formation and leave a proper reference behind, instead of the empty `name=""` / `base="ξ."` that did not round-trip.

## Printer

- `inline-cactoos.xsl` drops the cactus name when folding a dataized-const wrapper over a **non-abstract** value, so the folded value reads inline (a formation still keeps its name and prints `[…] >>!`).
- `dataized-to-const.xsl` rebuilds an anonymous const node (no `@name`) and carries the wrapper's positional `@as`, so a const argument beside a plain one keeps its slot.
- `to-eo-tree.xsl` renders `!` on an unnamed const node.
- `Pretty` keeps a const argument on one line — a bare `a!` on its own line has no valid vertical spelling, so the penalty vote that would otherwise split the arguments is overridden.

## Tests

- Parser `eo-syntax` pack `const-inline-argument.yaml` for `42.plus a!`.
- Printer print-packs proving the inline round-trip (`const-inline-argument.yaml`) and that a const argument beside a plain one stays on one line (`const-inline-argument-among-others.yaml`).

All existing parser (1497) and printer (183) tests continue to pass.

## Note on the motivating round-trip

The issue also mentions a print-pack proving `a >> b!` reprints to `42.plus a! > x`. That relies on parsing the const file-local handle `>> b!`, which is #5817 and has not landed yet (`>> b!` currently errors with `trailing garbage after name suffix`). The printer fold path added here already handles that anonymous-const case via the same `inline-cactoos` change, so it will produce the desired output once #5817 lands; the print-pack for it can be added then.

---
_Generated by [Claude Code](https://claude.ai/code/session_015jisBPZBvXQDXLkj9n91QF)_

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -187,6 +187,9 @@ final class Emissions {
             if (value.binding() != null) {
                 emit.slot(Emissions.bindingTag(value.binding()));
             }
+            if (value.constant()) {
+                emit.constant();
+            }
             emit.close();
         } else {
             Emissions.openValue(emit, null, value, line);
@@ -202,6 +205,9 @@ final class Emissions {
             emit.method(last.fragile());
             if (value.binding() != null) {
                 emit.slot(Emissions.bindingTag(value.binding()));
+            }
+            if (value.constant()) {
+                emit.constant();
             }
             emit.close();
         }
--- a/eo-parser/src/main/java/org/eolang/parser/Tokens.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Tokens.java
@@ -574,8 +574,15 @@ final class Tokens {
             } else {
                 tie = null;
             }
+            final boolean cnst;
+            if (!this.atEnd() && this.current() == '!') {
+                this.cursor = this.cursor + 1;
+                cnst = true;
+            } else {
+                cnst = false;
+            }
             args.add(
-                new Value(bare.kind(), bare.raw(), bare.pos(), this.cursor, tie, tail)
+                new Value(bare.kind(), bare.raw(), bare.pos(), this.cursor, tie, tail, cnst)
             );
         }
         return args;
--- a/eo-parser/src/main/java/org/eolang/parser/Value.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Value.java
@@ -67,6 +67,14 @@ final class Value {
     private final List<MethodChain> chain;
 
     /**
+     * True when the value carries a trailing {@code !} const marker
+     * (R-9.4) as an inline argument — e.g. {@code 42.plus a!}. Only
+     * set for horizontal arguments; a line head's const marker is a
+     * name-suffix concern handled by {@link Suffix}.
+     */
+    private final boolean constant;
+
+    /**
      * Ctor — no binding, no chain.
      * @param tag Kind
      * @param text Raw text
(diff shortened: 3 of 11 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (192 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2430](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5824, opened 2026-07-23
#5799: reject negative or non-integer places in string.as-fixed

### LINKED ISSUE #5799: `string.as-fixed` causes infinite recursion for negative `places` argument
## What Happens

`eo-runtime/src/main/eo/string/as-fixed.eo:35-42`, the `pow10` helper (and similarly `rec-pad` at lines 43-58):

```
if. > [p] > pow10
  p.eq 0
  1
  times.
    pow10
      as-number.
        p.minus 1
    10
```

`pow10` decrements `p` by 1 until it equals 0 (an exact equality check). `as-fixed`'s doc comment only describes behavior for the `n` argument (rounding/sign), and says nothing about validating `places`.

If a caller passes a negative `places` (for example `as-fixed 3.14159 -2`), `p` steps -2, -3, -4, ... away from 0 and never reaches it, so `pow10` (and `rec-pad`, which has the same `count.eq 0` / `count.minus 1` pattern) recurses forever.

Verified: added a temporary test calling `as-fixed 3.14159 -2`, rebuilt, then ran the test class under a monitored background job with a 150 second wall-clock timeout while polling memory usage every 10 seconds. The build/transpile phase behaved normally (about 45 seconds, matching a clean build). Once the surefire test JVM forked and started running the test class, its memory usage reached 2.27 GB within 5 seconds and continued climbing (about 4.1 GB combined across processes), with no test summary ever printed before the outer timeout killed it.

## What Should Happen

`as-fixed` should validate `places >= 0` (and ideally that it is a whole number) and return a `cant-...`-style fallback for invalid input, the same pattern already used elsewhere in this file's sibling `repeated`/`malloc` objects.

morphqdd:
Closes #5799

`string.as-fixed`'s `pow10` and `rec-pad` helpers step `places`/`count` down to 0 by exact equality (`p.eq 0`, `count.eq 0`). A negative or non-integer `places` argument never hits exactly 0, so both helpers recurse forever, hanging and eventually OOMing.

Adds a third void parameter `cant-fix`, matching the existing fallback pattern used by `string.repeated`: if `places` is negative or not a whole number, `as-fixed` now returns `cant-fix` (a caller-supplied recovery scope), or the bottom object if `cant-fix` is left unbound, instead of recursing forever.

Existing 2-argument call sites are unaffected, since `cant-fix` is only referenced when the new guard actually triggers.

Verified:
- `EO_string.EOas_fixedTest`: 12/12 green (10 existing + 2 new: negative places, non-integer places, both returning the provided fallback), plus the existing bottom-object test pattern for the no-fallback case.
- `mvn clean verify -PskipTests -Pqulice`: clean.

### DIFF
--- a/eo-runtime/src/main/eo/string/as-fixed.eo
+++ b/eo-runtime/src/main/eo/string/as-fixed.eo
@@ -1,7 +1,7 @@
 # Writes the number `n` as fixed-point decimal with exactly `places` fraction
 # digits, rounding to the nearest unit in the last place, so that `2.5` with
 # four places becomes `2.5000` and `0.9999999` with six places becomes `1.000000`.
-# If `places` is negative or not a whole number, the `cant-fix` fallback is
+# If `places` is negative or not a whole number, the `cant-print` fallback is
 # returned, or the bottom object when unbound.
 
 +architect yegor256@gmail.com
@@ -11,13 +11,13 @@
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
 
-[n places cant-fix] > as-fixed
+[n places cant-print] > as-fixed
   if. > @
     or.
       0.gt places
       not.
         places.floor.eq places
-    cant-fix
+    cant-print
       "Can't write a fixed-point decimal with %s fraction places".printf
         * places
     if.
--- a/eo-runtime/src/main/eo/string/as-fixed.eo
+++ b/eo-runtime/src/main/eo/string/as-fixed.eo
@@ -1,6 +1,8 @@
 # Writes the number `n` as fixed-point decimal with exactly `places` fraction
 # digits, rounding to the nearest unit in the last place, so that `2.5` with
 # four places becomes `2.5000` and `0.9999999` with six places becomes `1.000000`.
+# If `places` is negative or not a whole number, the `cant-fix` fallback is
+# returned, or the bottom object when unbound.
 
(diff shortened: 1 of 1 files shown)


eo-runtime/src/main/eo/string/as-fixed.eo
```diff
@@ -9,12 +11,20 @@
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
 
-[n places] > as-fixed
+[n places cant-fix] > as-fixed
```
yegor256: @morphqdd maybe `cant-print` is a more relevant wording here?

morphqdd: Renamed to cant-print.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (38) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1752](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2507](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5826, opened 2026-07-24
#5773: Fix printf byte format rewriting

### LINKED ISSUE #5773: `EO_string.EOprintf` corrupts escaped `%%x` and fails on `%x` with flags or width
## What happens

`printf` (`eo-runtime/src/main/eo/string/printf.eo`, atom `org.eolang.EO_string.EOprintf`) mishandles the `%x` (bytes) conversion in two related ways. Both were verified end-to-end by building `eo-runtime` from source and dataizing real `printf` calls.

**1. An escaped `%%` immediately followed by `x` is corrupted.** `%%` must always be a literal percent (the `tests-printf-prints-percents` example relies on this: `printf "%%%s" "Jeff"` → `"%Jeff"`). But `printf "%%x"`, which must produce the literal `"%x"`, produces `"%s"` instead.

**2. Any flag/width/precision on `%x` crashes.** A bare `%x` works, but `%-5x`, `%5x`, `%.3x`, etc. throw `java.util.IllegalFormatConversionException: x != java.lang.String` instead of formatting the bytes.

## Root cause

`EOprintf.lambda()` rewrites the format string with two naive `replaceAll` calls before handing it to `String.format`:

https://github.com/objectionary/eo/blob/667b081eb6d1ca0c1d8ea3611dd273e51d770cfd/eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java#L46-L48

```java
String.format(
    Locale.US,
    format.replaceAll("%\d+\$([a-zA-Z])", "%s").replaceAll("%x", "%s"),
    new PrintfArgs(format, ...).formatted().toArray()
)
```

`%x` means "bytes-as-hex-string" here, so `PrintfArgs` converts the argument to a `String` and the format must present it to `String.format` as `%s`. The rewrite that turns `%x` into `%s` is `replaceAll("%x", "%s")`, and it is wrong for two reasons:

- It matches a bare literal `"%x"` anywhere, **including the `%x` formed by the second `%` of an escaped `%%` and a following `x`.** So `"%%x"` becomes `"%%s"`, which `String.format` renders as the literal `"%s"` — the intended literal `"%x"` is lost.
- It only matches the **bare** `%x`, never `%x` carrying flags/width/precision. So `"%-5x"` is left untouched and reaches `String.format` as Java's integer hex conversion `%x` applied to the `String` that `PrintfArgs` produced → `IllegalFormatConversionException`.

Note the contrast: `PrintfArgs.SPECIFIER` (the argument-counting side, `PrintfArgs.java:51-53`) already parses a full specifier correctly — `%(\d++\$)?+[-#+ 0,(]*+\d*+(?:\.\d++)?+([a-zA-Z%])` — and even skips `%%`. Only the format-rewriting side in `EOprintf.lambda()` still uses the naive patterns, so the two sides disagree on what a specifier is. The positional rewrite `replaceAll("%\d+\$([a-zA-Z])", "%s")` has the same shape of fragility (it requires the conversion letter immediately after `$`, so `%1$5d` is not stripped).

## Failure scenarios

Verified against the compiled runtime (`printf` bound with `format`/`args`):

| call | result | expected |
| --- | --- | --- |
| `printf "%%d"` | `"%d"` | `"%d"` (works — control) |
| `printf "%x" 42.as-bytes` | `"40-45-00-00-00-00-00-00"` | same (works — bare `%x`) |
| `printf "%%x"` | `"%s"` | `"%x"` |
| `printf "%-5x" 42.as-bytes` | `IllegalFormatConversionException: x != java.lang.String` | `"40-45-00-00-00-00-00-00"` |
| `printf "%5x" 42.as-bytes` | same exception | right-justified hex-bytes string |

The `%%d` and bare `%x` rows show the breakage is specific to the `%x` rewrite: a literal `%%` is fine unless the next char is `x`, and `%x` is fine only without flags/width.

## Suggested fix

Rewrite the format with a single, specifier-aware pass instead of two blind `replaceAll` calls — mirroring the `PrintfArgs.SPECIFIER` pattern that already exists — so that:

- `%%` is recognized and left untouched (never treated as the start of a conversion);
- for a real specifier, the optional `N$` positional prefix is stripped **while its flags/width/precision are preserved**, and a conversion char of `x` is turned into `s` (keeping those same flags/width/precision).

For example, iterate `PrintfArgs.SPECIFIER.matcher(format)` and, for each match, emit `%%` verbatim when group 2 is `%`, otherwise rebuild the specifier from its flags/width/precision plus `s` when the conversion is `x` (or the original letter otherwise), dropping the `N$` group. Add tests that `printf "%%x"` yields `"%x"` and that `printf "%-5x" 42.as-bytes` (and `%5x`, `%.3x`) format the bytes rather than throwing.

This is distinct from the already-closed #5630 (which fixed argument *counting* for `%.2f`/`%5d`/`%-10s` via the `SPECIFIER` regex on the `PrintfArgs` side); the format-*rewriting* side in `EOprintf.lambda()` was never updated and is what breaks here.

ihopenre-eng:
## Summary

- replace the two blind format-string substitutions with one specifier-aware rewrite based on `PrintfArgs.SPECIFIER`
- preserve escaped `%%` and flags, width, and precision while adapting EO `%x` arguments to Java `%s`
- add unit and end-to-end coverage for escaped `%x`, alignment, width, precision, and positional width

## Testing

- `mvn -pl eo-runtime -Deo.autoFix -Dtest=PrintfArgsTest test` — passed (10 tests)
- full `eo-runtime` run — 1,753 tests, 0 failures, 0 errors; Qulice completed successfully
- the subsequent local Windows `project-validate` step could not resolve three generated Unicode-named classes under the MS949 user path; this occurs after tests and Qulice

Closes #5773

### DIFF
--- a/eo-runtime/src/main/eo/string/printf.eo
+++ b/eo-runtime/src/main/eo/string/printf.eo
@@ -87,6 +87,18 @@
       14
       false
 
+  eq. ++> tests-printf-applies-precision-to-bytes
+    "40-"
+    printf *1
+      "%.3x"
+      42.as-bytes
+
+  eq. ++> tests-printf-applies-width-to-bytes
+    "  40-45-00-00-00-00-00-00"
+    printf *1
+      "%25x"
+      42.as-bytes
+
   eq. ++> tests-printf-does-not-fail-if-arguments-more-than-formats
     "Hey"
     printf *1
@@ -101,6 +113,18 @@
           "%d"
           42.as-i64
         "42"
+
+  eq. ++> tests-printf-left-aligns-bytes
+    "40-45-00-00-00-00-00-00  "
+    printf *1
+      "%-25x"
+      42.as-bytes
+
+  eq. ++> tests-printf-preserves-escaped-bytes-format
+    "%x"
+    printf
+      "%%x"
+      *
 
   eq. ++> tests-printf-prints-i64-as-number
     "42"
--- a/eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
@@ -45,7 +45,7 @@ public final class EOprintf extends PhDefault implements Atom {
         return new Data.ToPhi(
             String.format(
                 Locale.US,
-                format.replaceAll("%\\d+\\$([a-zA-Z])", "%s").replaceAll("%x", "%s"),
+                PrintfArgs.javaFormat(format),
                 new PrintfArgs(
                     format,
                     Expect.at(this, "args")
--- a/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
@@ -40,16 +40,16 @@ final class PrintfArgs {
 
     /**
      * A single {@code printf} specifier. Group 1 is the optional {@code N$}
-     * positional index; an optional run of flags, an optional width and an
-     * optional {@code .precision} are then skipped; group 2 is the conversion
-     * character. Skipping the flags, width and precision keeps exactly one
+     * positional index; group 2 is an optional run of flags, an optional width
+     * and an optional {@code .precision}; group 3 is the conversion character.
+     * Skipping the flags, width and precision keeps exactly one
      * argument counted per specifier such as {@code %5d}, {@code %.2f} or
      * {@code %-10s}, matching what {@code String.format} consumes on the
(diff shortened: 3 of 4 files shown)


0crat:
@ihopenre-eng Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5829, opened 2026-07-24
#5828: keep a multi-referenced const `&gt;&gt; name!` handle whole

### LINKED ISSUE #5828: `eo-printer` incorrectly duplicates and discards shared name for multi-referenced const handle
The `eo-printer` auto-fixer eliminates a *const* file-local `>> name!` handle when it is referenced more than once, duplicating the const value at each use site and dropping the shared name. Given this input:

```
[a] > foo
  a >> b!
  42.plus b > x
  43.plus b > y
```

the printer returns:

```
[a] > foo
  42.plus a! > x
  43.plus a! > y
```

The offending stage is `inline-cactoos.xsl`, which the print train (`eo-printer/src/main/java/org/eolang/printer/Xmir.java:43-56`) runs before `merge-monikers.xsl`. By that point `const-to-dataized.xsl` has wrapped the const value in a `.as-bytes`/`Φ.dataized` node, and `inline-cactoos.xsl` folds that cactus binding into *every* matching reference (the `42.plus a!` case documented at `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:54-57`, inline template at line 59) and then drops the standalone binding (drop template at line 90), guarding only the recursive and dispatched targets.

This is wrong because `b!` is const: the object is dataized once and its result cached in that single binding, so the two `b` references share one const object. Inlining `a!` at both sites yields two independent const objects, changing the object graph, and discards the `b` name. Without the `!`, the same input correctly prints to `42.plus a` / `43.plus a`, since a plain alias is referentially transparent and safe to inline — the `!` is the only differentiator. The correct output keeps one handle and leaves the second reference intact:

```
[a] > foo
  42.plus > x
    a >> b!
  43.plus b > y
```

Fix: in `inline-cactoos.xsl`, add a guard so a dataized-const cactus binding with more than one reference is left in place (like the `eo:recursive`/`eo:dispatched` cases at lines 90-96), neither inlined nor dropped. `merge-monikers.xsl` then hosts it onto the first reference while the remaining references stay as `b`; unlike the non-const based-handle path (#5810), the const case must keep its `@name`/`@local` so it prints as `a >> b!`, not as an anonymous `a!`. A single-reference const handle may still inline to `a!` (#5821).

yegor256:
Fixes #5828.

## Problem

`eo-printer` was inlining a *const* file-local `>> name!` handle (R-3.10.12) at every use site when it is referenced more than once. Given:

```
[a] > foo
  a >> b!
  42.plus b > x
  43.plus b > y
```

the printer produced:

```
[a] > foo
  42.plus a! > x
  43.plus a! > y
```

A const is dataized **once** and its result cached in that single binding, so the two `b` references share one const object. Inlining `a!` at both sites mints two independent const objects (changing the object graph) and drops the shared `b` name.

## Fix

The fix lives in `restore-local-names.xsl`. A multi-referenced dataized-const handle is now treated like a handled void / recursive handle already are:

- its readable handle (which the parser leaves as `@local` on the wrapped value inside the `const-to-dataized` `.as-bytes`/`Φ.dataized` shell) is **promoted to the wrapper's visible name**, and
- every reference is **rewritten from the obfuscated cactus name back to the handle**.

Once promoted, the binding leaves the cactus namespace, so `inline-cactoos` no longer folds it and `merge-monikers` leaves it alone. It survives as a single standalone handle:

```
[a] > foo
  a >> b!
  42.plus b > x
  43.plus b > y
```

Single-use const still inlines to `a!` (#5821); a non-const `a >> b` is referentially transparent and still inlines to `a` at every use (#5810).

## Note on the printed shape

The issue suggested hosting the handle onto the first reference (`42.plus > x` with `a >> b!` nested beneath). That spelling does **not** round-trip: nesting a `>> b` handle under `.plus` rescopes `b` to that `.plus`, so on reparse the sibling `43.plus b` reference becomes unbound and the first use collapses back to a single-use `a!` — reintroducing the very graph corruption this issue is about. A file-local handle can only bind its siblings when it sits at the formation level, so the printer keeps the standalone spelling, which is byte-for-byte identical to the input and reprints to itself (verified by the existing `printsToParseableEo` fixpoint check).

## Tests

Added three print-packs under `eo-printer/.../print-packs/yaml/`:

- `const-multi-referenced-handle.yaml` — the two-reference case from the issue.
- `const-thrice-referenced-handle.yaml` — three references, proving the guard is not arity-specific.
- `non-const-multi-referenced-handle.yaml` — the non-const contrast that still inlines to `a`.

All 195 `eo-printer` tests pass (`printsToEo`, `printsToParseableEo`, and `checksPrintPacks`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01PRocDQjZJmR4sw2Rh1FbDR)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
@@ -29,6 +29,17 @@
   handle, and the marker is KEPT so "to-eo-tree" prints "&gt;&gt; name"
   rather than an anonymous "&gt;&gt;" bound to references to a synthetic
   "vL_P" placeholder.
+
+  A const "&gt;&gt;" handle (`a &gt;&gt; b!`, R-3.10.12) is a third case. Its
+  value is dataized once and cached in a single binding, so a handle referenced
+  more than once must stay one shared object rather than be inlined per use
+  (which would mint an independent const at each site and drop the shared name,
+  #5828). Such a handle is likewise treated as a handled void — its handle,
+  which the parser leaves on the wrapped value inside the "const-to-dataized"
+  `.as-bytes`/`Φ.dataized` shell, is promoted to the wrapper's visible name, its
+  references are rewritten back to the handle, and the marker is kept — so it
+  survives the later passes as a single standalone `a &gt;&gt; b!` binding. A
+  single-use const still inlines to `a!` (#5821).
   -->
   <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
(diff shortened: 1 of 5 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (163 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2444](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5830, opened 2026-07-24
#5798: reject non-integer times in string.repeated

### LINKED ISSUE #5798: `string.repeated` enters infinite loop when `times` argument is a non-integer
## What Happens

`eo-runtime/src/main/eo/string/repeated.eo:21-27`, the `rec-repeated` helper:

```
if. > [accum index] > rec-repeated
  amount.eq index
  accum
  rec-repeated
    accum.concat bts
    as-number.
      index.plus 1
```

`index` starts at 1 and is incremented by exactly 1 each recursive call. The loop only stops when `amount.eq index` is true (an exact equality check). The object's doc comment only guards `times < 0` (returns `cant-repeat`) and `times == 0` (returns empty), but never validates that `times` is a whole number.

Calling `repeated "x" 2.5` makes `index` step 1, 2, 3, 4, ... and it can never equal 2.5, so the recursion never terminates.

Verified: added a temporary test calling `repeated "x" 2.5`, built with `mvn -pl eo-runtime -am install -DskipTests ...`, then ran `mvn -pl eo-runtime test -Dtest=EO_string.EOrepeatedTest` under two separate hard wall-clock timeouts (180s, then 240s). Both runs were killed without the test ever completing, while equivalent single-test-class runs in this repo normally finish in about 120-135 seconds. Checking running processes after the kill showed the underlying forked Maven/surefire JVM processes were still alive and growing, one reached 1.8 GB RSS, the other 4.6 GB RSS, both pinned near 200-450% CPU: a real runaway process, not just a slow build.

## What Should Happen

`repeated` should either truncate/reject a non-integer `times` argument (consistent with how `malloc.of`/`malloc.for` already reject fractional sizes with a `cant-...` fallback), or otherwise guarantee the loop terminates for any numeric input.

morphqdd:
Closes #5798

`string.repeated`'s `rec-repeated` helper steps `index` up by exactly 1 until it equals `amount`. A non-integer `times` argument (for example `repeated "x" 2.5`) makes `index` step 1, 2, 3, 4, ... and it never equals 2.5, so the recursion never terminates.

Widens the existing guard (which already rejected `times < 0` via the `cant-repeat` fallback) to also reject a non-integer `times`, returning the same `cant-repeat` fallback, or the bottom object if `cant-repeat` is left unbound. Matches the existing pattern already used by `malloc.of`/`malloc.for` for rejecting fractional sizes, and the same fix just applied to the sibling bug in `string.as-fixed` (#5799, PR #5824).

Verified:
- `EO_string.EOrepeatedTest`: 5/5 green (3 existing + 1 new: non-integer times returns the provided fallback), plus the existing bottom-object test for the no-fallback negative case.
- `mvn clean verify -PskipTests -Pqulice`: clean.

### DIFF
--- a/eo-runtime/src/main/eo/string/repeated.eo
+++ b/eo-runtime/src/main/eo/string/repeated.eo
@@ -1,5 +1,6 @@
 # Returns `text` repeated `times` times.
-# If `times` < 0, the `cant-repeat` fallback is returned, or the bottom object when unbound.
+# If `times` < 0 or not a whole number, the `cant-repeat` fallback is returned,
+# or the bottom object when unbound.
 # If `times` == 0, empty text is returned.
 
 +architect yegor256@gmail.com
@@ -11,10 +12,13 @@
 
 [text times cant-repeat] > repeated
   if. > @
-    0.gt amount
+    or.
+      0.gt times
+      not.
+        times.floor.eq times
     cant-repeat
-      "Can't repeat text %d times".printf
-        * amount
+      "Can't repeat text %s times".printf
+        * times
     string result
   times > amount!
   text > bts!
@@ -43,6 +47,13 @@
     repeated
       "x"
       -1
+      "recovered" > [msg]
+
+  eq. ++> tests-text-repeated-non-integer-times-returns-provided-fallback
+    "recovered"
+    repeated
+      "x"
+      2.5
       "recovered" > [msg]
 
   eq. ++> tests-text-repeated-zero-times-is-empty


morphqdd:
@yegor256, review please

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (19) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1740](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5831, opened 2026-07-24
#5796: Reject scope symbols as auto-name handles

### LINKED ISSUE #5796: `Suffix.auto` wrongly accepts `@`, `^`, `$` as file-local handles in `>>` auto-name suffix parsing
The `>>` auto-name suffix parser accepts the special scope tokens `@`, `^` and `$` as file-local handles, in `Suffix.auto` at `eo-parser/src/main/java/org/eolang/parser/Suffix.java:377-412`. The handle-reading loop (lines 389-402) consumes any run of non-terminator characters as the handle and never rejects these tokens, so a line like `42 >> @` (or `>> ^`, `>> $`) parses cleanly and produces `<o base="Φ.number" local="@" name="a🌵2-2">`.

This is wrong because `@`, `^` and `$` are reserved scope symbols (the φ decoratee, ρ parent and ξ self), not identifiers usable as a file-local reference name; a `local="@"` attribute is meaningless XMIR. The sibling test-attribute path already guards against this: `Suffix.test` at lines 340-345 rejects `@` with "test attribute name must be an identifier, not @". The `>>` handle path has no equivalent guard, so it is inconsistent with the rest of the parser.

The fix is to add a guard in `Suffix.auto`, right after the handle is read (around line 394), that raises a `ParseError` when the handle equals `@`, `^` or `$`, mirroring the existing check in `Suffix.test`.

ihopenre-eng:
## Summary

- reject `@`, `^`, and `$` when they are parsed as file-local handles after `>>`
- keep the existing auto-name parsing path and error location intact
- add parameterized regression coverage for all three reserved scope symbols

## Testing

- `mvn -pl eo-parser -Dtest=SuffixTest test` — 44 tests passed
- `mvn -pl eo-parser test` — 1,503 tests passed, 3 skipped
- `mvn -pl eo-parser -PskipTests -Pqulice verify --batch-mode` — passed

Closes #5796

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -4,6 +4,8 @@
  */
 package org.eolang.parser;
 
+import java.util.Set;
+
 /**
  * A parsed name suffix — §3.10 of the spec.
  *
@@ -48,6 +50,11 @@ final class Suffix {
      * {@link #terminates} so dotted FQNs are still consumed whole.
      */
     private static final String NAME_BOUNDARIES = ",.|':;?[]{}()";
+
+    /**
+     * Scope symbols that cannot be used as file-local handles.
+     */
+    private static final Set<String> SCOPES = Set.of("@", "^", "$");
 
     /**
      * Suffix form.
@@ -396,6 +403,7 @@ final class Suffix {
                 end = end + 1;
             }
             handle = tail.substring(begin, end);
+            Suffix.rejectScope(handle, span, home + begin);
             if (handle.codePoints().anyMatch(cp -> cp == 0x1F335)) {
                 throw new ParseError(
                     span.line(), home + begin,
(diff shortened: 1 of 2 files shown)


Shaurya2k06 (approved):
LGTM

eo-parser/src/main/java/org/eolang/parser/Suffix.java
```diff
@@ -396,6 +403,7 @@ private static Parsed auto(
                 end = end + 1;
             }
             handle = tail.substring(begin, end);
+            Suffix.rejectScope(handle, span, home + begin);
```
yegor256, line 406: @ihopenre-eng read this: https://www.yegor256.com/2015/12/08/temporal-coupling-between-method-calls

0crat:
@shaurya2k06 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+151](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5833, opened 2026-07-24
#5828: host a multi-referenced const `&gt;&gt; name!` handle as a moniker

yegor256:
Follow-up to #5829 (per @yegor256's request for the moniker spelling).

## What changes

A multi-referenced const file-local handle (`a >> b!`, R-3.10.12) is still kept whole — it is never inlined per use, which would mint an independent const object at each site and drop the shared name (#5828). What changes is the printed **shape**: the kept binding is now hosted onto its first reference as a moniker, with the remaining references reading back as the bare handle.

```
[a] > foo               [a] > foo
  a >> b!                 42.plus > x
  42.plus b > x    =>       a >> b!
  43.plus b > y           43.plus b > y
```

## Printer (`eo-printer`)

- **`inline-cactoos`** leaves a multi-referenced dataized-const binding and its references in place (alongside the existing recursive / dispatched guards).
- **`restore-local-names`** keeps the `@local` handle on the wrapped value but leaves the binding cactus-named so it stays eligible for hosting.
- **`merge-monikers`** hosts the const handle onto its first reference keeping `@name`/`@local`/`@const` (so `to-eo-tree` prints `a >> b!`, not the anonymous `a!`), and rewrites the non-hosting references from the cactus name back to the readable handle.

Single-use const still inlines to `a!` (#5821); a non-const `a >> b` is referentially transparent and still inlines to `a` at every use (#5810).

## Parser (`eo-parser`) — the scoping fix

The moniker spelling writes the `>> b` handle **nested** inside an application (`42.plus > x` with `a >> b!` beneath it). The parser's `resolve-local-names` previously found a handle only as a **direct child** of a formation, so a nested handle left its sibling `43.plus b` / `44.plus b` references resolving to a free, un-scoped `b` — the spelling did not round-trip and the object graph changed on reparse.

`eo:captor` now recognises a handle wherever it is written in the formation's body: a formation owns a handle when some `@local` descendant has that formation as its **nearest** enclosing formation. The node-identity test keeps the nested-formation boundary intact, so a handle still never leaks up out of, or sideways between, nested formations (the #5780 sibling-scope rule). With the handle visible to its siblings, the hosted moniker form now reprints to itself — no `reprints: false` needed.

## Tests

Parser (`eo-packs/parse/`):
- `resolve-local-names-nested-scope.yaml` — a nested handle's sibling references resolve to its cactus name (the scoping fix).

Printer (`print-packs/yaml/`):
- `const-multi-referenced-handle.yaml` — two references, hosted moniker form (round-trips).
- `const-thrice-referenced-handle.yaml` — three references, hosted moniker form (round-trips).
- `const-single-referenced-handle.yaml` — single use still inlines to `a!` (unchanged).
- `non-const-multi-referenced-handle.yaml` — non-const contrast still inlines to `a` (unchanged).

All `eo-parser` (1501) and `eo-printer` (161, none skipped) tests pass; all four modified stylesheets pass `xcop`.

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
@@ -31,12 +31,21 @@
   declares the referenced name - as a public attribute "@name" or as a handle
   "@local" - is the reference's scope; the reference binds to a handle only
   when that scope's declaration is itself a handle.
+
+  A handle is an attribute of the formation that declares it wherever it is
+  written in that formation's body, not only as a direct child: a handle
+  written nested inside an application (`42.plus > x` with `a &gt;&gt; b!`
+  beneath it, the moniker spelling the printer emits, #5828) still belongs to
+  the enclosing formation. So a formation owns a handle when some "@local"
+  descendant has that formation as its nearest enclosing formation - the "is"
+  test stops the search at a nested formation boundary, so a handle never
+  leaks up out of, or sideways between, nested formations (#5780).
   -->
   <xsl:function name="eo:captor" as="element()?">
     <xsl:param name="ref" as="element()"/>
     <xsl:variable name="name" as="xs:string" select="$ref/@base"/>
-    <xsl:variable name="scope" as="element()?" select="($ref/ancestor::o[not(@base)][o[@name=$name or @local=$name]])[last()]"/>
-    <xsl:sequence select="$scope/o[@local=$name][1]"/>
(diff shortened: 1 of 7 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (187 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2448](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5518, opened 2026-07-15
Pipe `|` fails to apply to nameless formations in current implementation


maxonfjvipon:
The pipe `|` added in #5432 (PR #5510) can only apply to a *named* formation. `LnPipe.precheck` at `eo-parser/src/main/java/org/eolang/parser/LnPipe.java:114` rejects a nameless predecessor, because the pipe desugars by *referring to it by name* — and a bare `[x]` has no name to refer to:

```
[x]
  ...
| 5 > y        # rejected today: "a pipe must follow a named formation or another pipe"
```

`[a b] >>` (auto-named) already covers the anonymous case, so the gap is only the truly nameless form. It is worth supporting, but it is a bigger change than the named case and pulls in three modules, so it is filed here rather than folded into #5510.

The nameless form cannot be a by-name reference, so it has to *fuse* the applied arguments into the formation node instead. That produces one `<o>` that mixes the formation's own attributes (voids and body) with the application's arguments — the phi-calculus `⟦x↦5⟧`. The shape is already valid and dataizable: `(x.plus x > [x]) 5 > y` parses today with no errors and yields exactly this,

```xml
<o name="y">
  <o base="∅" name="x"/>
  <o base="ξ.x.plus" name="φ">…</o>
  <o base="Φ.number">5</o>          <!-- applied arg, appended as a child -->
</o>
```

so the runtime needs nothing new. The one structural rule that makes the mix unambiguous is that every fused *argument* must carry an `@as` attribute (positional `α0…` or an explicit binding), which is how a downstream reader tells an applied argument apart from a real attribute of the formation.

Three pieces of work follow from that. First, the parser must allow a formation to be nameless when a pipe follows it and supplies the name, which means deferring the "every formation child must be named" check (R-5.3.1) across that one line, and teaching `wrap-applications.xsl` a second branch that merges the pipe's args into the preceding formation and moves the `> name` onto it. Second, `eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl` must handle a formation node that carries `@as` argument children so the generated Java is valid. Third, the `eo-printer` module must reverse-print the fused node back to the `| …` form rather than as a plain formation. Each can be built and tested independently once the fused XMIR contract (arguments marked by `@as`) is fixed.

maxonfjvipon:
[...]

PR #5835, opened 2026-07-24
#5834: preserve args and name when inlining an applied `&gt;&gt;` handle

### LINKED ISSUE #5834: `inline-cactoos` in `eo-printer` discards argument and binding name during single-referenced formation handle application
The `eo-printer` drops the applied argument **and** the result-binding name when it inlines a single-referenced file-local `>> name` handle whose value is an abstract formation and which is used as the base of an application. The fault is in `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl`, the inline template at line 64 (`match="o[contains(@base, concat('.', $auto))]"`). For the snippet

```
[] > foo
  [] >> bar
  bar 42 > x
```

`Xmir.toEO()` prints

```
[] > foo
  [] >>
```

losing both the `42` argument and the `> x` name; the expected output is `[]` then `| 42 > x`.

The reason is that when the template inlines, it rebuilds a fresh `<o>` from the *target* formation's attributes and nodes (lines 70-76) and copies only the reference's `@as`. It never carries over the reference node's own children — the applied arguments (`<o as="α0">42</o>`) — nor its own `@name` (`x`). That is correct for a bare reference (a leaf with at most an `@as`), but when the reference is the base of an application the arguments and the binding name are silently discarded. This is the same class of silent argument loss as #5721, only in a different pass.

The fix is to preserve them at the inline site: when the matched reference carries argument children or its own `@name`, keep the inlined formation as the head and re-attach the arguments and the name, producing the fused nameless-formation-plus-pipe node (`[] | 42 > x`) that #5518 already taught the printer to emit. The bare-reference path (no children, no name) stays unchanged.

yegor256:
Fixes #5834.

## Problem

`eo-printer`'s `inline-cactoos` pass dropped the applied argument **and** the result-binding name when it inlined a single-referenced file-local `>> name` handle whose value is an abstract formation and which is used as the base of an application. For

```
[] > foo
  [] >> bar
  bar 42 > x
```

`Xmir.toEO()` printed

```
[] > foo
  [] >>
```

losing both the `42` argument and the `> x` name.

## Cause

The inlining template (`match="o[contains(@base, concat('.', $auto))]"`) rebuilt a fresh formation node from the *target* formation's attributes and children and copied only the reference's `@as`. That is correct for a bare reference (a leaf with at most an `@as`), but when the reference is the base of an application the reference's own argument children and its `@name` were never carried over, so they were silently discarded. This is the same class of silent loss as #5721, in a different pass.

## Fix

An abstract formation cannot be spelled inline as the head of an application, so instead of inlining it away:

- **Inlining template** — when the matched reference carries argument children or its own `@name` and stands immediately after its abstract target, keep the target formation in place and rewrite the reference into the `| args > name` pipe continuation (#5518): an ordinary application node tagged `@pipe`, its `@base` still pointing at the kept formation directly above it. The bare-reference path (no children, no name) is unchanged.
- **Drop template** — keep the auto-named formation binding when such a pipe reference follows it (new `eo:piped` helper), so the pipe has a valid `[] >>` predecessor rather than being stranded.

The snippet above now prints the expected

```
[] > foo
  [] >>
  | 42 > x
```

## Tests

Added `inline-cactoos-applied-handle.yaml` under `print-packs/yaml`, which exercises both `XmirTest#printsToEo` (origin → printed) and `XmirTest#printsToParseableEo` (printed reparses to itself). The full `eo-printer` suite (199 tests) passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TUEN3rSAPLAdN3jkEzfPDy)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -67,13 +67,41 @@
     <xsl:variable name="keep-name" as="xs:boolean" select="exists($target) and (eo:abstract($target) or ($target/@base = '.as-bytes' and $target/o[1]/@base = 'Φ.dataized' and eo:abstract($target/o[1]/o[1])))"/>
     <xsl:choose>
       <xsl:when test="exists($target) and not(eo:void($target)) and not(eo:recursive($target, $name)) and not(eo:dataized-const($target) and eo:multi-referenced($target, $name))">
-        <xsl:element name="o">
-          <xsl:if test="@as">
-            <xsl:apply-templates select="@as"/>
-          </xsl:if>
-          <xsl:apply-templates select="$target/@*[$keep-name or (name() != 'name' and name() != 'local')]"/>
-          <xsl:apply-templates select="$target/node()"/>
-        </xsl:element>
+        <xsl:choose>
+          <!--
+          The reference is the base of an application — it carries its own
+          argument children or a result-binding `@name`. Folding a fresh copy
+          of the target formation over it (the bare-reference path below) would
+          rebuild only the formation and silently drop those arguments and the
+          name (#5834), the same class of loss as #5721. An abstract formation
+          cannot be spelled inline as the head of an application, so instead of
+          inlining it away we keep it in place as the auto-named `[] &gt;&gt;`
+          predecessor (its binding is preserved by the drop template below) and
+          turn this reference into the pipe continuation `| args &gt; name`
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2480](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5838, opened 2026-07-24
Fold recursive `>> name` handle's applied sibling to pipe (#5837)

### LINKED ISSUE #5837: eo-printer leaves `bar > x` expanded instead of folding to `| > x` for a self-referential `>>` handle
Feeding `eo-printer` this snippet is a no-op, yet it should fold the sibling reference into the compact pipe continuation:

```
[] > foo
  [] >> bar
    bbb bar > z
  bar > x
```

The `bar > x` line stays expanded, but the canonical spelling is `| > x` (the `[] >> bar` formation itself stays, since it is recursive). The expected `| > x` form is idempotent — it parses to the same XMIR and reprints to itself — so the printer emits two different texts for one object graph.

The divergence is specific to a self-referential handle. `restore-local-names.xsl` classifies `[] >> bar` as recursive (`eo:recursive`, `eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl:64`) because its body references itself through `bbb bar`, so it promotes `@local="bar"` to `@name` and rewrites every reference base from the cactus name `ξ.a🌵…` to `ξ.bar` (`restore-local-names.xsl:93` and `:100`). That pass runs before `inline-cactoos.xsl`, whose #5834 pipe-continuation logic (`eo:piped`, `inline-cactoos.xsl:177`; template at `:87`) only matches cactus-named references (`o[contains(@base, '.a🌵')]`, `:64`). Once the cactus name is stripped nothing folds the reference, and `merge-monikers.xsl` skips it too (it needs a cactus `@name`, `merge-monikers.xsl:80`). The non-recursive counterpart (`ccc > z` in the body) does fold to `| > x`, and an already-piped recursive handle round-trips (`pipe-recursive.yaml`), so this is the one gap.

Fix by tagging the reference with `@pipe` on the restored-recursive path too: when `restore-local-names.xsl` restores a recursive `>> name` handle, mark its immediately-following applied sibling reference (one carrying arguments or a name) with `pipe=""`, mirroring `eo:piped`. `to-eo-tree.xsl:244` already renders `@pipe` when the base equals the preceding sibling's name, so tagging alone emits the `|`.

yegor256:
Fixes #5837.

## Problem

Feeding `eo-printer` a self-referential (recursive) file-local handle whose applied sibling should fold to a compact pipe continuation was a no-op:

```
[] > foo
  [] >> bar
    bbb bar > z
  bar > x
```

The `bar > x` line stayed expanded, but the canonical spelling is `| > x` (the `[] >> bar` formation itself stays, since it is recursive). Because the expected `| > x` form is idempotent, the printer emitted two different texts for one object graph.

## Root cause

`restore-local-names.xsl` classifies `[] >> bar` as recursive (its body references itself through `bbb bar`), so it promotes `@local="bar"` to `@name` and rewrites every reference base from the cactus name (`ξ.a🌵…`) to `ξ.bar`. That pass runs **before** `inline-cactoos.xsl`, whose #5834 pipe-continuation logic (`eo:piped`) only matches cactus-named references (`o[contains(@base, '.a🌵')]`). Once the cactus name is stripped, nothing folds the reference, and `merge-monikers.xsl` skips it too (it needs a cactus `@name`). The non-recursive counterpart already folds, and an already-piped recursive handle round-trips (`pipe-recursive.yaml`), so this was the one gap.

## Fix

Tag the reference with `@pipe` on the restored-recursive path in `restore-local-names.xsl`: when a recursive `>> name` handle is restored, mark its immediately-following applied sibling reference (one carrying arguments or a name) with `pipe=""`, mirroring `eo:piped`. `to-eo-tree.xsl` already renders `@pipe` when the base equals the preceding sibling's name, so tagging alone emits the `|`. A reference that already carries `@pipe` (an already-piped handle round-tripping) is left untouched.

## Tests

Added `pipe-recursive-self.yaml` covering the snippet above. Both `printsToEo` (prints to the expected text) and `printsToParseableEo` (the expected text reprints to itself — idempotent) pass. Full `XmirTest` suite: 165 tests, 0 failures.

https://claude.ai/code/session_013hQzLCvqAnncApnv887ksF

---
_Generated by [Claude Code](https://claude.ai/code/session_013hQzLCvqAnncApnv887ksF)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
@@ -107,6 +107,28 @@
   formations, whose handle is inlined away by "inline-cactoos".
   -->
   <xsl:template match="o[not(@base=$eo:empty) and not(eo:recursive(., @name)) and not(eo:const-handle(parent::o/parent::o))]/@local"/>
+  <!--
+  When a recursive "&gt;&gt; name" handle is restored, its cactus name is
+  promoted to the visible "@name" and every reference is rewritten from the
+  cactus name to the handle. That strips the cactus name before
+  "inline-cactoos" runs, so its #5834 pipe-continuation logic
+  (`eo:piped`) — which only matches cactus-named references — never fires
+  and the applied sibling reference stays expanded (`bar &gt; x`) instead of
+  folding to the compact `| &gt; x`. Mirror `eo:piped` here: tag the handle's
+  immediately-following applied sibling reference — one carrying arguments or
+  a name — with `pipe=""`. "to-eo-tree" renders "@pipe" when the base equals
+  the preceding sibling's name, so tagging alone emits the "|". A reference
+  that already carries "@pipe" (an already-piped handle round-tripping) is
+  left as is.
+  -->
+  <xsl:template match="o[contains(@base, concat('.', $auto)) and (o or @name) and preceding-sibling::o[1][@local and eo:recursive(., @name)] and eo:resolved-name(@base) = preceding-sibling::o[1]/@name]">
+    <xsl:copy>
+      <xsl:if test="not(@pipe)">
+        <xsl:attribute name="pipe"/>
+      </xsl:if>
+      <xsl:apply-templates select="node()|@*"/>
+    </xsl:copy>
+  </xsl:template>
   <xsl:template match="node()|@*">
     <xsl:copy>
       <xsl:apply-templates select="node()|@*"/>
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2488](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5123, opened 2026-05-11
`trimMargin` method incorrectly strips internal whitespace instead of only leading indentation


bibonix:
In `eo-parser/src/main/java/org/eolang/parser/XeEoListener.java`, line 1185, the `trimMargin` method does this for each line of a text block:

```java
line.replaceAll(String.format("[\\s]{%d}", indent), "")
```

`replaceAll` without a `^` anchor replaces every occurrence of exactly `indent` consecutive whitespace characters anywhere in the line, not only at the beginning. So a text block like this (with `indent=4`):

```
    """hello    world
    """
```

produces `helloworld` instead of `hello    world` — the four spaces between the words get eaten along with the leading indentation.

Fix: anchor the pattern to the start and use `replaceFirst` so only leading whitespace up to `indent` characters is removed:

```java
line.replaceFirst(String.format("^\\s{0,%d}", indent), "")
```

`{0,%d}` instead of `{%d}` also handles lines that are shorter than the indent level (e.g. blank lines inside a text block) without silently skipping them.

bibonix:
@yegor256 would you mind taking a look at this?

0crat:
@bibonix Thank you for reporting the bug! Following our team policy, you've been awarded **+12** points for helping improve our project quality. We encourage you to continue finding issues - if you spot anything in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as expected, please [don't hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+210](https://www.eolang.org/zerocracy/objectionary-vitals.html), and remember to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #5841, opened 2026-07-24
#5790: stop trimming text-block body, only opener indent should be stripped

### LINKED ISSUE #5790: `LnTextBlock` incorrectly trims whitespace, leading to inconsistent text block content
## What Happens

`eo-parser/src/main/java/org/eolang/parser/LnTextBlock.java:59`:

```java
final String joined = Emissions.unescapeBody(
    String.join(String.valueOf('\n'), globals.tbody()).trim()
);
```

`Globals.appendTextLine` (`Globals.java:248-258`) correctly strips only the opener's own indent from each line. But then `LnTextBlock` calls `.trim()` on the whole joined string. `String.trim()` only strips whitespace touching the very start/end of the entire string, not per-line margins. So:

- Extra indentation beyond the block's margin is preserved when it occurs on any line except the first or last.
- The identical extra indentation is silently swallowed when it happens to land on the first or last line, because there it is adjacent to the string boundary.
- An intentional leading blank line in the body is deleted entirely rather than becoming an empty first line of the string.

This contradicts the parser's own spec, `PARSER_SPEC.md:552` (R-3.11.2): "the block's text is the body lines joined by `\n` with the opening indent stripped", nothing about trimming beyond that.

Verified by building `eo-parser` from `upstream/master` with `mvn -pl eo-parser -am install -DskipTests -Dinvoker.skip=true -Drat.skip=true`, then running `EoSyntax(...).parsed()` on:

```
[] > main
  """
   first
  second
  """ > x
```

which gives bytes for `"first\nsecond"` (the leading space on line 1 is lost), versus the same extra space on the second line instead:

```
[] > main
  """
  first
   second
  """ > x
```

which gives bytes for `"first\n second"` (the space is preserved). A leading blank body line before content is also dropped completely rather than becoming an empty first line.

This is the same defect class as issue #5123 ("trimMargin incorrectly strips internal whitespace", fixed by PR #5124 in the old ANTLR-based `XeEoListener`). It has effectively reappeared through a different mechanism (`.trim()` instead of an unanchored regex) in the new hand-written parser that replaced `XeEoListener`. The project's own test fixture `eo-parser/src/test/resources/org/eolang/parser/eo-syntax/text-block.yaml` currently bakes in the swallowed-first/last-line behavior as "expected," which would need updating alongside a fix.

## What Should Happen

Extra indentation or an intentional blank line inside a triple-quoted text block should be preserved or dropped consistently, regardless of which line it falls on. Only the opener's declared indent should ever be stripped from each line (as `Globals.appendTextLine` already does correctly); the final assembly should not apply any additional whole-string `.trim()`.

morphqdd:
Closes #5790

`LnTextBlock.java:59` called `.trim()` on the whole joined text-block body after `Globals.appendTextLine` already correctly strips only the opener's own indent from each line. `String.trim()` only strips whitespace touching the very start/end of the entire joined string, not per-line margins, so:

- Extra indentation beyond the block's margin was silently swallowed when it happened to land on the first or last line, but preserved on any other line.
- An intentional leading blank line in the body was deleted entirely instead of becoming an empty first line.

This contradicts the parser's own spec, `PARSER_SPEC.md:552` (R-3.11.2): "the block's text is the body lines joined by `\n` with the opening indent stripped", nothing about trimming beyond that. Same defect class as issue #5123, which was already fixed once in the old ANTLR-based `XeEoListener` and reappeared here through a different mechanism (`.trim()` instead of an unanchored regex) in the new hand-written parser.

Fix: remove the `.trim()` call. `Globals.appendTextLine` already does the only stripping the spec calls for.

Updated the existing `text-block.yaml` fixture, which baked in the old swallowed-first-line behavior as "expected" (a body with 3 extra leading spaces on its only body line was asserted to come out as `"third"`; with the fix it correctly comes out as `"   third"`). Added a new assertion for the leading-blank-line case (`"\nsecond"` now preserved instead of becoming just `"second"`).

Verified:
- `EoSyntaxTest` (923 parameterized cases, includes the text-block fixture): all green.
- Full `eo-parser` test module: 1013/1013 green.
- `qulice:check -Pqulice` on `eo-parser`: clean.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnTextBlock.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnTextBlock.java
@@ -56,7 +56,7 @@ final class LnTextBlock implements Line {
             tokens.tail(), this.span, this.span.indent() + tokens.cursor()
         );
         final String joined = Emissions.unescapeBody(
-            String.join(String.valueOf('\n'), globals.tbody()).trim()
+            String.join(String.valueOf('\n'), globals.tbody())
         );
         this.transition(stack, suffix);
         emit.object(
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/text-block.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/text-block.yaml
@@ -6,7 +6,8 @@ sheets: []
 asserts:
   - //o[@base='Φ.string' and o/o[text()='48-69-2C-20-74-68-65-72-65-0A-41-64-69-C3-B3-73']]
   - //o[@base='Φ.string' and o/o[text()='66-69-72-73-74-0A-20-73-65-63-6F-6E-64']]
-  - //o[@base='Φ.string' and o/o[text()='74-68-69-72-64']]
+  - //o[@base='Φ.string' and o/o[text()='20-20-20-74-68-69-72-64']]
+  - //o[@base='Φ.string' and o/o[text()='0A-73-65-63-6F-6E-64']]
 input: |
   [] > main
     """
@@ -21,4 +22,9 @@ input: |
     stdout > mrg
       """
          third
+      """
+    stdout > blank
+      """
+
+      second
       """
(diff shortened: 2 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1764](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5842, opened 2026-07-24
Relocate a separated applied `>>` handle so its pipe survives (#5840)

### LINKED ISSUE #5840: eo-printer drops an applied `>>` handle reference when a binding separates it from the handle
The printer destroys an applied `>>` handle reference when another binding sits between the handle and its use. Feeding `eo-printer` this snippet:

```
[] > foo
  44 > [k] >> bar
  67 > t
  bar 8 > x
```

prints as:

```
[] > foo
  44 > [k] >>
  67 > t
```

The whole `bar 8 > x` line is gone — its argument `8` and its result name `x` are silently dropped — and the handle loses its name `bar`. Every attribute of `foo` should survive, with the formation inlined into the `bar 8 > x` application.

This is the #5834 arg-dropping loss, still live whenever the applied reference is not the handle's immediate neighbour. The #5834 fix (`inline-cactoos.xsl`) folds an applied reference into a `| args > name` pipe, but both its guards require adjacency: the pipe branch tests `preceding-sibling::o[1] is $target` (`eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:87`) and `eo:piped` tests `following-sibling::o[1]` (`inline-cactoos.xsl:180`). With `67 > t` between the handle and `bar 8 > x` both guards fail, so the reference falls into the bare-reference inline `otherwise` (`inline-cactoos.xsl:95`), which rebuilds only the formation and drops the reference's own children and `@name`, while the drop template (`inline-cactoos.xsl:126`) removes the original binding. Remove the intervening `67 > t` and the same input folds correctly to `| 8 > x`, confirming adjacency is the only trigger.

The reference should be inlined at its use site with its argument and name kept, exactly as the adjacent case already does — a definition sitting far above its single use must still fold in. The printer's inline-application spelling for an anonymous formation is the `|` pipe, which binds the immediately-preceding sibling, so the fix is to relocate the single-use handle binding to sit directly above its applied reference (the same relocation #5732 proposes for its sibling case) and let the existing #5834 pipe path emit the pipe. The result

```
[] > foo
  67 > t
  44 > [k] >>
  | 8 > x
```

preserves every binding, inlines the formation into the application, and reprints to itself.

yegor256:
Fixes #5840.

## Problem

`eo-printer` dropped an applied `>>` handle reference when another binding sat between the handle and its use. Feeding

```
[] > foo
  44 > [k] >> bar
  67 > t
  bar 8 > x
```

printed as

```
[] > foo
  44 > [k] >>
  67 > t
```

— the whole `bar 8 > x` line (its argument `8` and result name `x`) was silently dropped, and the handle lost its name.

This is the #5834 arg-dropping loss, still live whenever the applied reference is not the handle's immediate neighbour. The #5834 pipe path folds an applied reference into a `| args > name` pipe, but both its guards require adjacency (`preceding-sibling::o[1] is $target` and `eo:piped` testing `following-sibling::o[1]`). With `67 > t` between the handle and `bar 8 > x`, both guards fail, so the reference fell into the bare-reference inline `otherwise`, which rebuilt only the formation and dropped the reference's own children and `@name`, while the drop template removed the original binding.

## Fix

A `|` pipe binds the immediately-preceding sibling, so `inline-cactoos.xsl` now relocates the single-use handle formation to sit directly above its applied reference: it emits a fresh copy of the formation (its children inlined as usual) immediately followed by the `| args > name` pipe continuation. The formation's original binding is dropped by the drop template rather than duplicated (it is not `eo:piped`, having a non-reference following sibling), so the relocation moves it — the same relocation #5732 proposes for its sibling case. The path is restricted to a single-use formation, since a multi-referenced one cannot be folded into just one of its uses.

The example now prints (and reprints to itself) as:

```
[] > foo
  67 > t
  44 > [k] >>
  | 8 > x
```

preserving every binding, with the formation inlined into the application.

## Tests

Adds `inline-cactoos-applied-handle-separated.yaml`, the separated sibling of the existing `inline-cactoos-applied-handle.yaml` pack. It passes both `printsToEo` and `printsToParseableEo` (the expected output reprints to itself). The full `eo-printer` suite (205 tests) is green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FUUshPBXmetxLAc8KMbai8)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -82,9 +82,39 @@
           tagged `@pipe`, its `@base` still pointing at the kept formation
           directly above it. The guard keeps this to a reference standing
           immediately after its own target, the one shape `to-eo-tree` can
-          render as a compact pipe.
+          render as a compact pipe (the formation stays put, kept by the drop
+          template below).
           -->
           <xsl:when test="eo:abstract($target) and (o or @name) and preceding-sibling::o[1] is $target">
+            <xsl:copy>
+              <xsl:if test="not(@pipe)">
+                <xsl:attribute name="pipe"/>
+              </xsl:if>
+              <xsl:apply-templates select="node()|@*"/>
+            </xsl:copy>
+          </xsl:when>
+          <!--
+          The same applied reference, but with another binding sitting between
+          the formation and this use (`67 &gt; t` in #5840): the formation is
+          still a preceding sibling, only no longer the immediate one, so the
+          adjacent guard above misses it and the bare-reference `otherwise`
+          would drop the argument and name exactly as #5834 did. A `|` pipe
+          binds the immediately-preceding sibling, so we relocate the
+          single-use formation to sit directly above this reference — emit a
+          fresh copy of it here (its children inlined as usual) and then the
+          pipe continuation. The formation's original binding is dropped by the
+          drop template below (it is not `eo:piped`, having a non-reference
+          following sibling), so the relocation moves it rather than
+          duplicating it, the same relocation #5732 proposes for its sibling
+          case. Restricted to a single-use formation: a multi-referenced one
+          cannot be folded into just one of its uses.
+          -->
+          <xsl:when test="eo:abstract($target) and (o or @name) and exists(preceding-sibling::o[. is $target]) and not(eo:multi-referenced($target, $name))">
+            <xsl:for-each select="$target">
+              <xsl:copy>
+                <xsl:apply-templates select="node()|@*"/>
+              </xsl:copy>
+            </xsl:for-each>
             <xsl:copy>
               <xsl:if test="not(@pipe)">
                 <xsl:attribute name="pipe"/>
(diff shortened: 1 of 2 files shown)


yegor256:
The `mvn (ubuntu-24.04, 26)` check failed with a `java.lang.OutOfMemoryError: Java heap space` in `eo-runtime` — `EObytesEOconcatTest.failsCleanlyInsteadOfOverflowingArraySize` (`eo-runtime/.../EObytesEOconcatTest.java:51`), a test that deliberately allocates a very large array. The runner ran out of heap before the test's own overflow guard could fire.

This is unrelated to this PR: `eo-printer` — the only module changed here (an XSL-only fix) — built and tested green (`eo-printer ... SUCCESS`), and a printer change cannot affect `eo-runtime` heap usage. It looks like a transient/infrastructure OOM. I don't have permission to re-run the failed jobs via the API; a maintainer re-run should clear it.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2500](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5843, opened 2026-07-24
#5792: verify unpacked destination is non-empty before trusting a resolve cache hit

### LINKED ISSUE #5792: `Resolving.resolved()` incorrectly treats partial dependency unpack as complete
## What Happens

`CentralMaven.unpack()` (`eo-maven-plugin/src/main/java/org/eolang/maven/CentralMaven.java:256-272`) calls `Files.createDirectories(dest)` (line 257) before it starts writing any of the JAR's zip entries into that directory (lines 258-270).

If the Maven process is killed mid-unpack (OOM killer, CI timeout, `Ctrl-C`, a `mvn clean` race), the destination directory (`place`) exists on disk but is empty or only partially populated.

On the next build, `Resolving.resolved()` (`eo-maven-plugin/src/main/java/org/eolang/maven/Resolving.java:153-159`) only checks `Files.exists(place)`. If true, it logs "Dependency already resolved" and skips re-resolving entirely (`total = 0`). There is no check that the directory actually contains the expected files.

A single interrupted build permanently breaks the project until someone manually deletes `target/4-resolve/<group>/<artifact>/<classifier>/<version>`. Every subsequent build silently fails to place `.class` files there, without any error. The failure surfaces later, confusingly, in transpile/compile/link as a `ClassNotFoundException` or a missing runtime dependency, with no obvious connection back to the original interrupted resolve.

Verified with a JUnit test (`ResolvingStaleCacheTest.doesNotReResolveWhenPlaceDirectoryExistsButIsEmpty`) that pre-creates the `4-resolve/org.eolang/eo-runtime/-/0.7.0` directory empty (simulating an interrupted prior unpack), then runs a normal `eo:resolve` via the project's own `FakeMaven` test harness with a program depending on `eo-runtime:0.7.0`. The test fails: `eo-runtime-0.7.0.class` is never created, confirming the bug reproduces against real code, not just by inspection.

## What Should Happen

`Resolving.resolved()` should verify the destination actually contains the expected unpacked content before treating it as a cache hit, for example by checking the directory is non-empty, unpacking atomically to a temp directory and renaming into place only on success, or writing a sentinel/marker file only after a fully successful unpack. This matches the "verify completeness before trusting a cache hit" principle already applied to similar caching bugs in this codebase (for example the `FpIfTargetOlder` inverted-comparison fix).

morphqdd:
Closes #5792

`CentralMaven.unpack()` calls `Files.createDirectories(dest)` before writing any of the JAR's zip entries. If the Maven process is killed mid-unpack (OOM killer, CI timeout, Ctrl-C), the destination directory exists on disk but is empty.

`Resolving.resolved()` only checked `Files.exists(place)` to decide a dependency was already resolved, with no check that the directory actually contains the expected files. A single interrupted build permanently broke the project until someone manually deleted the stale directory, since every subsequent build silently skipped re-resolving.

Fix: also check the destination is non-empty (reusing the existing `Walk` class already used elsewhere in this same method to count freshly-unpacked files) before treating it as a cache hit.

Added a test that pre-creates the destination directory empty (simulating an interrupted prior unpack), then runs a normal resolve via `FakeMaven` with a program depending on `eo-runtime:0.7.0`. Verified the test fails against the old code (reverted the fix locally and confirmed the class file is never created) and passes with the fix.

Verified:
- `MjResolveTest`: 11/11 green (10 existing + 1 new).


eo-maven-plugin/src/main/java/org/eolang/maven/Resolving.java
```diff
@@ -150,7 +150,7 @@ private int resolved(final Dep dep, final Path dest) throws IOException {
             dependency.getVersion()
         );
         final int total;
-        if (Files.exists(place)) {
+        if (Files.exists(place) && !new Walk(place).isEmpty()) {
```
yegor256: @morphqdd maybe it's better to do this:

```
if (Files.exists(place)) {
  if (new Walk(place).isEmpty()) {
    // .warn()
  } else {
    // as now
  }
} else {
  // as now
}
```

morphqdd:

yegor256:
@morphqdd see my comment above

eo-maven-plugin/src/main/java/org/eolang/maven/Resolving.java
gemshrine: The `!isEmpty()` guard closes the fully-empty case, but #5792 describes the leftover as "empty **or only partially populated**", and this only catches the first half. If `CentralMaven.unpack()` is killed after it has written *some* zip entries into `place` (but not all), the directory is non-empty, so `!new Walk(place).isEmpty()` is `true`, the build logs "already resolved", sets `total = 0`, and the missing `.class` files are never written — the same silent `ClassNotFoundException` the issue is about, just one interrupted-unpack shape further along. Non-empty is one of the options the issue lists, so this may be a deliberate scope call, but as written a partial unpack is still trusted. Writing a sentinel/marker file only after a fully successful unpack (and gating the cache hit on that marker), or unpacking into a temp dir and atomic-renaming into `place`, closes both halves.

Separately, on efficiency: `new Walk(place)` walks the whole tree and collects *every* regular file into a list just to answer "is there at least one file?", and `resolved()` runs this per already-resolved dependency on every build (e.g. eo-runtime is hundreds of `.class` files). Since the method already `throws IOException`, a short-circuiting check that stops at the first file is strictly cheaper:

```java
final boolean populated;
try (Stream<Path> tree = Files.walk(place)) {
    populated = tree.anyMatch(Files::isRegularFile);
}
...
if (Files.exists(place) && populated) {
```

eo-maven-plugin/src/test/java/org/eolang/maven/MjResolveTest.java
```diff
@@ -45,6 +46,26 @@ void resolvesWithSingleDependency(@Mktmp final Path temp) throws IOException {
         );
     }
 
+    @Test
+    void resolvesWhenPlaceDirectoryExistsButIsEmpty(@Mktmp final Path temp) throws IOException {
```
gemshrine, line 50: This covers the fully-empty leftover, but not the partial-unpack half of #5792. A companion test that pre-populates `place` with a single stray file — the shape an interrupted unpack actually leaves — would pin that case:

```java
Files.createDirectories(place);
Files.write(place.resolve("stale-partial.class"), new byte[] {1});
// ... run FakeMaven.Resolve() as above ...
// assert eo-runtime-0.7.0.class exists
```

Against the current `!isEmpty()` fix this test would fail (the stray file makes the dir non-empty → cache hit → no re-resolve → the real `.class` never appears), which is exactly the residual gap noted on `Resolving.java`. Worth adding once the completeness check distinguishes a partial unpack from a complete one.

gemshrine (changes requested):
The empty-leftover case is genuinely fixed and CI is green — and `Walk` is a good choice for the check, since it's a recursive, files-only walk, so a directory holding only empty sub-directories is correctly treated as unresolved (a shallow top-level check would have missed that).

Two things inline: the fix addresses only half of #5792 — a *partially* unpacked leftover (build killed after some entries were written) is non-empty, so it still reads as a cache hit and reproduces the same silent failure; and a small efficiency note, since `new Walk(place)` eagerly collects the whole tree just to test non-emptiness on the per-dependency resolve path. Plus a suggested partial-unpack test that currently fails against this fix, marking the gap.

gemshrine (approved):
Withdrawing my earlier change-request — this correctly fixes the reported empty-leftover case (one of the options the issue itself lists) and CI is green. The partial-unpack hardening and the efficiency note I left inline are optional follow-ups, not blockers for this PR. Approving.

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2092](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2752](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+690](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5845, opened 2026-07-24
Pipe the argument of a handle applied as a dispatch receiver (#5844)

### LINKED ISSUE #5844: `eo-printer` incorrectly discards argument when `>>` handle is utilized as a dispatch receiver
The printer silently drops an argument when a single-use `>>` handle is applied and that application sits as a method-dispatch receiver. Feeding `eo-printer`:

```
[] > foo
  (bar 55).a 88 > k
  [t] >> bar
```

prints as:

```
[] > foo
  a. > k
    [t] >>
    88
```

`bar 55` (the `[t]` formation applied to `55`) has collapsed into the bare `[t] >>` receiver — the `55` is gone. In the parsed XMIR the receiver is `<o base="ξ.a🌵3-2">` carrying the arg `55`, sitting as the `ρ` of the `.a` dispatch (`loc="Φ.foo.k.ρ"`); only that `55` child is lost, while the surrounding `.a 88` survives.

This is the #5834 / #5840 arg-dropping loss again, in a new shape. `inline-cactoos.xsl` folds an applied handle reference into a `| args > name` pipe, but the pipe branch is guarded by `preceding-sibling::o[1] is $target` (`eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:87`). A dispatch receiver is never a sibling of its target formation, so the guard fails and the reference falls into the bare-reference inline `otherwise` (`inline-cactoos.xsl:95`), which rebuilds only the formation and discards the reference's own children — here the `55`. The relocate-and-pipe remedy sketched for #5840 cannot apply: a nested receiver has no sibling slot to host a `|` pipe.

An anonymous formation cannot be spelled inline as the head of an application (the premise behind #5834), so for this shape the fix is to not inline at all: when the applied reference carries arguments (`$ref/o`) yet cannot become a pipe, keep the handle as a named `>> name` binding and let the reference stay `bar 55`, instead of inlining the formation over it at `inline-cactoos.xsl:95`.

yegor256:
Fixes #5844.

## Problem

`eo-printer` silently dropped an argument when a single-use `>>` handle was applied and that application sat as a method-dispatch receiver. Feeding:

```
[] > foo
  (bar 55).a 88 > k
  [t] >> bar
```

printed as:

```
[] > foo
  a. > k
    [t] >>
    88
```

`bar 55` collapsed into a bare `[t] >>` receiver — the `55` was gone.

## Cause

The `bar 55` receiver is the `ρ` of the `.a` dispatch, so it is never a *following sibling* of its `[t]` formation. Both #5834 pipe guards in `inline-cactoos.xsl` require the applied reference to stand as a following sibling of the formation, so the reference fell into the bare-reference inline `otherwise`, which rebuilds only the formation and discards the reference's own children — the `55`.

## Fix

Handle it as a **relocated pipe**, the way #5840 already handles a *separated* sibling: `inline-cactoos` relocates a single-use applied formation to its reference's site and turns the reference into a `| args` pipe. The #5840 branch was guarded by `exists(preceding-sibling::o[. is $target])`; dropping that guard lets it also fire when the reference is a nested dispatch receiver, so the formation copy and the `| 55` pipe land inside the dispatch block and `to-eo-tree` renders the reversed dispatch `a. > k` with the formation and its pipe as the receiver and `88` as the argument.

Two supporting changes:

- **`restore-local-names.xsl`** keeps the handle's `@local` marker for this shape (a new `eo:applied-receiver` predicate), so the relocated predecessor — which sits *inside* the dispatch block — prints its readable `>> bar` handle rather than an anonymous `>>`. Unlike a void/recursive handle, its cactus `@name` is left obfuscated and its references are not rewritten, so `inline-cactoos` still recognises and pipes it.
- **`unnecessary-as.xsl`** learns that a reversed dispatch whose receiver is a pipe has *two* leading receiver children (the formation predecessor and its `| args` pipe), so it still strips the redundant `@as` from the trailing positional arguments — `88`, not `88:0`.

The example now prints (and reprints to itself) as:

```
[] > foo
  a. > k
    [t] >> bar
    | 55
    88
```

## Tests

- `print-packs/yaml/inline-cactoos-applied-handle-receiver.yaml` — full-pipeline pack covering the receiver case; it reprints to itself.
- `eo-packs/print/restore-applied-receiver-local-name.yaml` — XSL-level pack pinning the `restore-local-names` behaviour (marker kept, name left obfuscated, reference not rewritten).

All 206 tests in `eo-printer` pass.

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -94,22 +94,28 @@
             </xsl:copy>
           </xsl:when>
           <!--
-          The same applied reference, but with another binding sitting between
-          the formation and this use (`67 &gt; t` in #5840): the formation is
-          still a preceding sibling, only no longer the immediate one, so the
-          adjacent guard above misses it and the bare-reference `otherwise`
-          would drop the argument and name exactly as #5834 did. A `|` pipe
-          binds the immediately-preceding sibling, so we relocate the
-          single-use formation to sit directly above this reference — emit a
-          fresh copy of it here (its children inlined as usual) and then the
-          pipe continuation. The formation's original binding is dropped by the
-          drop template below (it is not `eo:piped`, having a non-reference
-          following sibling), so the relocation moves it rather than
+          The same applied reference, but not standing as the formation's
+          immediate following sibling, so the adjacent guard above misses it
+          and the bare-reference `otherwise` would drop the argument and name
+          exactly as #5834 did. Two shapes reach here: another binding sits
+          between the formation and this use (`67 &gt; t` in #5840, the
+          formation still a preceding sibling), or the use is a dispatch
+          receiver buried as the `ρ` of a `.method` node (`(bar 55).a` in
+          #5844), which shares no sibling slot with the formation at all. A `|`
+          pipe binds the immediately-preceding sibling, so in both we relocate
+          the single-use formation to sit directly above this reference — emit
+          a fresh copy of it here (its children inlined as usual) and then the
+          pipe continuation. For the receiver case the copy and the pipe land
+          inside the dispatch block, so `to-eo-tree` renders them as the
+          reversed dispatch's receiver (`[t] &gt;&gt; bar` then `| 55`). The
+          formation's original binding is dropped by the drop template below (it
+          is not `eo:piped`, having a non-reference following sibling, or no
+          following sibling at all), so the relocation moves it rather than
           duplicating it, the same relocation #5732 proposes for its sibling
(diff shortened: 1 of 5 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (182 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2523](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5846, opened 2026-07-24
#5791: reject lone UTF-16 surrogate unicode escapes in string literals

### LINKED ISSUE #5791: Invalid unicode escapes decoding to lone surrogate are not reported and default to '?' in appendUnicode
## What Happens

`eo-parser/src/main/java/org/eolang/parser/Emissions.java:482`, inside `appendUnicode` (called from `unescapeBody` at line 284):

```java
out.append(
    (char) Integer.parseInt(body.substring(cursor, cursor + 4), 16)
);
```

Any 4 valid hex digits are accepted and cast straight to `char`, including the range `D800`-`DFFF` reserved for UTF-16 surrogate pairs, which is not a valid standalone Unicode scalar value and has no UTF-8 encoding. When the resulting `String` (holding a lone surrogate) is later encoded via `Hex.java`'s `text.getBytes(StandardCharsets.UTF_8)`, Java's default UTF-8 encoder silently substitutes the replacement byte `0x3F` (ASCII `?`) with no exception and no warning.

Verified by parsing:

```
[] > main
  "\uD800" > x
```

No errors are reported. The XMIR shows `<o base='Φ.bytes'>3F-</o>`, a single byte `0x3F` (`?`), instead of any error or a faithful representation of what was typed.

This is the same bug family that was already fixed for three siblings in this exact file: issue #5259/#5613 (numeric literal overflow now throws `ParseError` "is over-precise"), issue #5419 (octal escape out-of-range now throws with a clear message), and issue #5257 (malformed non-hex `\uZZZZ` now caught and reported as `ParseError`, via the `try/catch` at `Emissions.java:142-153`). None of those fixes cover a valid-hex-but-invalid-codepoint `\u` escape, so this case still falls through silently.

## What Should Happen

Writing `"\uD800"` (or any of the 2048 lone-surrogate code points `\uD800`-`\uDFFF`) in an EO string literal currently parses successfully with no error and silently produces the single-byte string `"?"`, discarding the user's actual input without any diagnostic. `appendUnicode` should instead reject unicode escapes that decode to a lone surrogate with a positioned `ParseError` (for example "invalid unicode escape: lone surrogate"), consistent with how the same file already rejects out-of-range octal escapes and malformed hex digits instead of silently substituting the wrong bytes.

morphqdd:
Closes #5791

`Emissions.appendUnicode` accepted any 4 valid hex digits and cast straight to `char`, including the `D800`-`DFFF` range reserved for UTF-16 surrogate pairs, which is not a valid standalone Unicode scalar value. Writing `"\uD800"` parsed successfully with no error and silently produced the single-byte string `"?"` when later encoded to UTF-8, discarding the user's actual input.

Same defect family already fixed for three siblings in this file: over-precise numeric literals (#5259/#5613), out-of-range octal escapes (#5419), and malformed non-hex `\uZZZZ` (#5257). None of those covered a valid-hex-but-invalid-codepoint `\u` escape.

Fix: after `unescapeBody` assembles the full decoded string, check it for any UTF-16 surrogate that isn't immediately paired with its partner (a high surrogate followed by a low surrogate is a valid pair, e.g. a supplementary character spelled as two `\u` escapes, like an emoji, and must not be rejected). A genuinely lone surrogate throws `NumberFormatException`, which the existing catch block in `Emissions.number()` already turns into a `ParseError`, matching the same reporting path already used for the octal and malformed-hex siblings.

Checking the whole assembled string (not each `\u` escape in isolation) is necessary: an initial attempt that rejected surrogates per-escape broke valid multi-escape surrogate pairs used throughout the project's own naughty-strings test fixture (emoji spelled as `🏳` etc.) — caught by running the full `EoSyntaxTest` suite before finalizing.

Verified:
- New test `rejectsLoneSurrogateUnicodeEscape`: `"\uD800"` now surfaces a `ParseError`.
- New test `acceptsValidSurrogatePairEscape`: `"🌈"` (a valid pair) still parses with no error.
- Full `eo-parser` test module: 1503/1503 green, including the existing naughty-strings fixture with real surrogate-pair emoji.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -295,7 +295,39 @@ final class Emissions {
                 idx = idx + 2;
             }
         }
+        Emissions.rejectLoneSurrogates(out);
         return out.toString();
+    }
+
+    /**
+     * Reject a decoded body containing a UTF-16 surrogate with no matching
+     * partner — a lone {@code \uD800}-{@code \uDFFF} escape has no UTF-8
+     * encoding and would otherwise be silently mangled later. A high
+     * surrogate immediately followed by a low surrogate is a valid pair
+     * (for example a supplementary character spelled as two {@code \\u}
+     * escapes) and is left untouched.
+     * @param text Decoded body to check
+     */
+    private static void rejectLoneSurrogates(final CharSequence text) {
+        int cursor = 0;
+        while (cursor < text.length()) {
+            final char glyph = text.charAt(cursor);
+            if (Character.isHighSurrogate(glyph)
+                && cursor + 1 < text.length()
+                && Character.isLowSurrogate(text.charAt(cursor + 1))) {
+                cursor = cursor + 2;
+                continue;
+            }
+            if (Character.isSurrogate(glyph)) {
+                throw new NumberFormatException(
+                    String.format(
+                        "unicode escape \\u%04X is a lone surrogate, not a valid standalone codepoint",
+                        (int) glyph
+                    )
+                );
+            }
+            cursor = cursor + 1;
+        }
     }
 
     /**
--- a/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
@@ -662,6 +662,30 @@ final class EoSyntaxTest {
     }
 
     @Test
+    void rejectsLoneSurrogateUnicodeEscape() throws Exception {
+        MatcherAssert.assertThat(
+            "a \\u escape decoding to a lone UTF-16 surrogate (D800-DFFF) must show up as an /object/errors/error entry, not silently emit '?'",
+            EoSyntaxTest.raw(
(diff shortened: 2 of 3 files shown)


yegor256:
@morphqdd maybe you can add a YAML test too? Java tests are not portable, while YAML tests are golden we can use in the future when we implement EO parser in, for example, Rust

morphqdd:
@yegor256 added a golden YAML test (`eo-parser/src/test/resources/org/eolang/parser/eo-syntax/lone-surrogate-escape-is-rejected.yaml`), same style as the existing `cti-adds-error.yaml`. Verified it fails against the pre-fix code (reverted Emissions.java locally and reran EoSyntaxTest, confirmed the test catches the regression) and passes with the fix. Full eo-parser EoSyntaxTest suite: 926/926 green.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2000](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5849, opened 2026-07-24
Fold recursive `>> name` handle's applied reference above it to pipe (#5848)

### LINKED ISSUE #5848: eo-printer leaves an applied reference expanded above a recursive `>>` handle instead of folding to `| > @`
The printer does not fold the applied reference of a recursive `>>` handle when that reference sits *above* the handle. Feeding `eo-printer` this snippet is a no-op:

```
[] > foo
  bar > @
  [t] >> bar
    bar 55 > ii
```

The `bar > @` line stays expanded, even though the canonical spelling is `| > @` with the reference relocated below the handle:

```
[] > foo
  [t] >> bar
    bar 55 > ii
  | > @
```

The recursion itself is not the obstacle — the *direction* is. The non-recursive counterpart (change the body's `bar 55` to `bar1 55`, so the handle no longer references itself) folds exactly as expected:

```
[] > foo
  [t] >> bar
    bar1 55 > ii
  | > @
```

And a recursive handle whose reference *follows* it also folds (that is #5837):

```
[] > foo
  [t] >> bar
    bar 55 > ii
  bar > @
```

prints as `| > @`. Only the recursive-handle-with-reference-above case is left expanded. The expected `| > @` form is idempotent — it parses to the same XMIR and reprints to itself (both `printsToEo` and `printsToParseableEo` pass) — so the printer emits two different texts for one object graph.

The gap is in the #5837 fix. `restore-local-names.xsl` classifies `[t] >> bar` as recursive (`eo:recursive`, `eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl:64`) because its body references itself through `bar 55`, so it promotes `@local="bar"` to `@name` and de-cactusizes every reference base (`ξ.a🌵…` → `ξ.bar`) before `inline-cactoos.xsl` runs. `inline-cactoos` therefore never sees the reference as cactus-named, and its #5834/#5840 relocate-and-pipe machinery — which already folds the *non-recursive* case in either direction — cannot fire. To compensate, #5837 tags the reference with `@pipe` inside `restore-local-names` itself, but that template only matches a reference whose *immediately-preceding* sibling is the recursive handle (`preceding-sibling::o[1][@local and eo:recursive(., @name)]`, `restore-local-names.xsl:153`). When the reference precedes the handle, the template misses it, nothing tags it, nothing relocates it, and it stays expanded.

Fix by handling the reference-above direction too: when a recursive `>> name` handle has an applied reference (one carrying arguments or a name) among its *preceding* siblings, relocate that reference to sit immediately after the handle and tag it with `@pipe`, mirroring the relocate-and-pipe that `inline-cactoos` already performs for the non-recursive case (#5840) and the `@pipe` tagging #5837 already performs for the reference-below case. `to-eo-tree.xsl` renders `@pipe` as `|` once the reference is the handle's following sibling.

Reproduced on `master` (commit 6b6767b) with JDK 25 (GraalVM). This is a formatting inconsistency, not data loss — the reference survives, it is just not folded. Sibling of #5837 (opposite reference direction).

yegor256:
Fixes #5848.

## Problem

The eo-printer left an applied reference expanded *above* a recursive `>>` handle instead of folding it to the canonical `| > @` below the handle. Feeding this snippet to the printer was a no-op:

```
[] > foo
  bar > @
  [t] >> bar
    bar 55 > ii
```

when the canonical spelling is:

```
[] > foo
  [t] >> bar
    bar 55 > ii
  | > @
```

So the printer emitted two different texts for one object graph.

## Root cause

`restore-local-names.xsl` classifies `[t] >> bar` as recursive (its body references itself through `bar 55`), so it promotes `@local="bar"` to `@name` and de-cactusizes every reference base before `inline-cactoos.xsl` runs. `inline-cactoos` therefore never sees the reference as cactus-named, and its #5834/#5840 relocate-and-pipe machinery cannot fire. The #5837 fix compensates by tagging the reference with `@pipe` inside `restore-local-names` itself, but that template only matches a reference whose *immediately-preceding* sibling is the recursive handle (the reference-below direction). When the reference precedes the handle, nothing tags or relocates it.

## Fix

Handle the reference-above direction in `restore-local-names.xsl`:

- Suppress the applied reference at its origin (a reference whose immediately-following sibling is the recursive handle it resolves to).
- Re-emit it immediately below the recursive handle, tagged `@pipe` — the same relocate-and-pipe that `inline-cactoos` already performs for the non-recursive case (#5840) and the `@pipe` tagging that #5837 performs for the reference-below case.

`to-eo-tree.xsl` renders `@pipe` as `|` once the reference is the handle's following sibling.

## Tests

- Adds `pipe-recursive-self-before.yaml`, the reference-above print-pack from the issue; it reprints to itself (both `printsToEo` and `printsToParseableEo` pass).
- Updates `recursive-local-handle.yaml`, whose expected output encoded the old unfolded form — `reclast list > @` above the handle now folds to `| list > @` below it, consistent with the fix.

All 208 `eo-printer` tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dfyqfi9ajYeUB6mzTKNujw)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
@@ -159,41 +159,51 @@
     </xsl:copy>
   </xsl:template>
   <!--
-  The mirror case (#5848): the applied sibling reference sits just ABOVE the
-  recursive handle, not below it. A "| args &gt; name" pipe binds its
-  immediately-preceding sibling, so a reference standing before the handle
-  cannot pipe against it where it is; the #5837 tag-in-place path above only
-  matches a reference whose preceding sibling is the handle, so it never fires
-  and the reference stays expanded ("bar &gt; @"). Relocate it instead: drop
-  the reference from its own position (below) and re-emit it just under the
-  handle, tagged "@pipe", the same relocation "inline-cactoos" performs for a
-  separated non-recursive handle (#5840) — done here because the recursive
-  handle is never inlined and its cactus name is stripped before
-  "inline-cactoos" runs. The reference is suppressed at its origin:
+  The mirror case (#5848): the applied reference sits ABOVE the recursive
+  handle, not below it. A "| args &gt; name" pipe binds its immediately-preceding
+  sibling, so a reference standing before the handle cannot pipe against it
+  where it is; the #5837 tag-in-place path above only matches a reference whose
+  preceding sibling is the handle, so it never fires and the reference stays
+  expanded ("bar &gt; @"). Relocate it instead: suppress the reference at its
+  origin and re-emit it just under the handle tagged "@pipe", the same
+  relocate-and-pipe "inline-cactoos" performs for a separated non-recursive
+  handle (#5840) — done here because the recursive handle is never inlined and
+  its cactus name is stripped before "inline-cactoos" runs. The reference need
+  not be the handle's immediate sibling; any applied reference resolving to the
+  handle among its preceding siblings folds, mirroring #5840's separated case.
+
+  Suppress the reference at its origin. Match any applied reference (one
+  carrying arguments or a name) with a recursive handle among its following
+  siblings, then keep only those that resolve to it — a reference resolving to a
+  different binding still prints in place.
   -->
-  <xsl:template match="o[contains(@base, concat('.', $auto)) and (o or @name) and following-sibling::o[1][@local and eo:recursive(., @name)] and eo:resolved-name(@base) = following-sibling::o[1]/@name]"/>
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (132 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2527](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5850, opened 2026-07-24
Reject NAME suffix starting with a non-letter (#5847)

### LINKED ISSUE #5847: Name binding allows invalid names starting with non-letter characters
The explicit name-binding parser reads the name token in `Suffix.named` at `eo-parser/src/main/java/org/eolang/parser/Suffix.java:431-475`: it consumes the run of characters via `endsName` and rejects only the cactus emoji (lines 446-451). It never enforces the §2.3 rule that a `NAME` must start with `[a-z]`. As a result a program like `[] > foo` with a body line `  bar > 42` parses with no error and emits `<o base="Φ.bar" name="42"/>`. The same holds for `[] > 42`, `bar > 4foo`, and `bar > -x`.

This is wrong because §2.3 (`PARSER_SPEC.md:166`) defines `NAME` as `[a-z]` followed by the allowed characters, so a name whose first character is a digit (or `-`) is not a valid `NAME` and must be reported as a syntax error. The parser already rejects other `NAME` violations — the cactus emoji here, and `@` in `Suffix.test` (lines 340-345) — so silently accepting `42` is both inconsistent and produces invalid XMIR (`name="42"`).

The fix is to add a guard in `Suffix.named`, right after the name is extracted (around line 445), that raises a `ParseError` when the first character is not in `[a-z]`, mirroring the existing cactus guard. The same first-character check is also missing on the `+>`/`->` test path and the `>>` handle path (the handle path's special-token variant is already tracked in #5796); aligning all three would close the gap fully.

Gooh456:
`Suffix.named` and `Suffix.test` extract the name token and only guard against the cactus emoji — never check the first char is `[a-z]` per §2.3. So `bar > 42`, `[] > 42`, `bar > -x` all parse and emit invalid XMIR (`name="42"`). Added the same first-char guard used for the cactus check to both paths. Left the `>>` handle path alone since that's tracked separately in #5796.

Fixes #5847.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -360,14 +360,31 @@ final class Suffix {
                 "cactus emoji is reserved for auto-names; not allowed in identifiers"
             );
         }
-        if (name.charAt(0) < 'a' || name.charAt(0) > 'z') {
+        Suffix.checkLowercaseStart(name, span, home, start);
+        Suffix.endsClean(tail, idx, span, home);
+        return new Suffix.Parsed(form, name, "", false);
+    }
+
+    /**
+     * Reject a name that doesn't start with a lowercase letter — unless
+     * it's the {@code @} void/phi marker, which callers that allow it
+     * (only {@link #named}; {@link #test} rejects {@code @} earlier and
+     * never reaches here with it) pass through untouched.
+     * @param name Extracted name; may be empty
+     * @param span Source span
+     * @param home Source column where the enclosing tail begins
+     * @param at Source column of the name's first character
+     */
+    private static void checkLowercaseStart(
+        final String name, final Span span, final int home, final int at
+    ) {
+        if (!name.isEmpty() && !"@".equals(name)
+            && (name.charAt(0) < 'a' || name.charAt(0) > 'z')) {
             throw new ParseError(
-                span.line(), home + start,
+                span.line(), home + at,
                 "name must start with a lowercase letter"
             );
         }
-        Suffix.endsClean(tail, idx, span, home);
-        return new Suffix.Parsed(form, name, "", false);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Gooh456:
Caught it — CI's right, I broke ~half the suite. The lowercase-first-letter check didn't account for `@` (the void/phi marker), which is what `> @` desugars through — and `> @` is everywhere in this codebase (`Suffix.phi()` already maps it to `φ`, same special-case shows up in `Emissions.java`/`LnFormation.java`/`LnOnlyPhi.java`). My check fired on every one of those before it ever got to the phi-mapping step.

Fixed: excluded `@` from the `named()` guard specifically. Left `test()` alone — it already rejects `@` explicitly one step earlier ("test attribute name must be an identifier, not @"), so my addition there was already unreachable for that case, not actually broken.

Verified locally against the real classes (javac, stubbing only the unrelated cactoos dep) before pushing this time: `> @`, `> bar`, `+> valid` all parse clean; `> Bar`, `> 9bad`, `+> Bad` still correctly throw; `+> @` still correctly throws via the existing guard. Should be green now — apologies for the noisy CI run.

Gooh456:
One more note on CI: `integration` is still red but I don't think it's this PR's fault. Checked the job's recent history across the repo — it's also failing right now on `renovate/major-maven.resolver.version` and `renovate/org.eolang-jucs-0.x` (pure dependency-bump branches, nothing to do with the parser), and it already passed twice on this exact branch earlier today before failing on the latest push. All 4 `mvn` matrix jobs (the ones that actually compile and run the Java test suite, including `Suffix.java`) are green. Looks like a pre-existing flake in that job specifically, not something this change broke — don't have rerun permissions to confirm by re-triggering it myself, but wanted to leave the evidence here rather than let it sit unexplained.

yegor256:
@Gooh456 
Thanks for your contribution! Unfortunately, we can't merge it without tests. Every feature we add and every bug we fix must be accompanied by at least one new test that confirms that the code was not working before. Without such a test, how can we know that what you commit actually works?

Gooh456:
fair, added tests — `> Bar`, `> 9bad`, `+> Bad` all throw now, plus one confirming `> @` still parses clean so the phi desugar path doesn't regress. no maven on this box so I compiled Suffix + the test standalone against junit5/hamcrest and ran the whole SuffixTest suite through the console launcher — 44/45 green, only miss is an unrelated emoji-encoding artifact in generatesAutoNameAttribute from my terminal's charset, nothing to do with this change.

0crat:
@gooh456 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5852, opened 2026-07-24
Fold a recursive `>> name` handle's dispatch reference above it (#5848)

yegor256:
Follow-up to #5848, covering the method-dispatch reference direction (the previous PR #5849 handled bare/applied references that fold to `| ... > @`).

## Problem

The printer left a **dispatch** reference to a recursive `>>` handle expanded when that reference sits *above* the handle. This snippet was a no-op:

```
[] > foo
  bar.read > @
  [buffer] >> bar
    bar chunk > x
```

The canonical spelling folds the dispatch into a reversed dispatch `read. > @` hosting the handle, exactly like the non-recursive counterpart already does:

```
[] > foo
  read. > @
    [buffer] >> bar
      bar chunk > x
```

## Root cause

`restore-local-names.xsl` promotes a recursive handle's `@local` to its visible `@name` and rewrites its references before `merge-monikers.xsl` runs, so the handle is no longer cactus-named. `merge-monikers` only hosts cactus-named bindings onto their uses, so it skipped the recursive handle and the dispatch stayed expanded.

## Fix

- **`merge-monikers.xsl`**: recognise a *restored recursive handle* — an abstract formation whose `@name` equals its kept `@local` — as a hostable moniker, so it folds onto its dispatch use as a reversed dispatch, keeping the readable `>> name` so self-references still read by name. Its bare uses always carry a `@name` (the `| ... > name` pipe), so they never resolve as hostable bare references and only a genuine dispatch use folds — the pipe cases (#5837/#5848) are untouched.
- **`to-eo-tree.xsl`**: hosting nests the formation one level under the reversed-dispatch node, which made the self-reference print as `^.name`. Extend the redundant-`^`-hop drop to recognise a self-reference whose hopped-to formation carries that handle name in `@local`, so it prints the bare `name`.

## Tests

- Adds `dispatch-recursive-self-before.yaml` (the example from the request; the body is given a name so the formation parses). It reprints to itself — both `printsToEo` and `printsToParseableEo` pass.

All 212 `eo-printer` tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dfyqfi9ajYeUB6mzTKNujw)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -68,16 +68,38 @@
     <xsl:sequence select="if (exists($ref/@base) and starts-with($ref/@base, concat($eo:xi, '.')) and contains($tail, '.') and substring-before($tail, '.') != '' and not(contains(substring-after($tail, '.'), '.'))) then substring-after($tail, '.') else ''"/>
   </xsl:function>
   <!--
+  Whether `$attr` is a restored recursive `&gt;&gt; name` handle: an abstract
+  formation carrying its readable "@local" handle whose visible "@name" has
+  already been promoted to that same handle by "restore-local-names" (which
+  sets "@name = @local" and rewrites the self-references only for a recursive
+  formation, R-3.10.12 / #5681). Its cactus name is gone, yet when it is used
+  as a method-dispatch receiver it must still host onto that dispatch as a
+  reversed-dispatch moniker (#5848) — the recursive mirror of the non-recursive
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (106 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2531](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5854, opened 2026-07-24
Fold an applied recursive `>> name` handle onto its use as a moniker, without losing the argument (#5848)

yegor256:
Follow-up to #5848 / #5852, covering the case where a recursive `>>` handle is not merely dispatched but **applied** (carries arguments) at its use site. The handle folds into a moniker where it is used, with its arguments becoming a `| args` pipe — instead of staying a separate binding while the use prints expanded.

Two shapes are covered:

**Dispatch receiver** — `(bar 7).neg`: the applied handle is inlined below `neg.`, its argument becoming a `| 7` pipe.

**Plain applied reference** (list element / argument) — such as `directory.eo`'s `r (walk "**").at.^` inside a `seq *`: the handle formation is hosted in place and the argument becomes a `| (walk "**").at.^` pipe.

## What went wrong, and the two fixes

**1. Printer — the pipe must not break vertically (`Pretty.java`).** When the fold emits a nameless `| args` pipe, the penalty layout preferred to break a wide or parenthesised argument onto its own line below a bare `|`. A bare `|` with a deeper-indent body does not parse (`a pipe `|` must be followed by a space`), so the argument was dropped. A nameless pipe has no valid vertical spelling, so it is now forced to its horizontal rendering (`| (walk "**").at.^`). A named pipe (`| > @` with args below) keeps content after the marker, so its vertical form still parses and is left to the ordinary penalty vote, unchanged.

**2. Parser — star-tuple + pipe round-trip (`stars-to-tuples.xsl`, `wrap-applications.xsl`).** The folded `seq *` (handle formation + `| args` + trailing element) re-parsed to a tuple with a **spurious extra numeric element** — the parse-side mirror of #5858 — which changed the object graph and broke `directory.eo`'s `deleted` helper at runtime. `stars-to-tuples` counted every star child as a tuple element, including the pipe-predecessor formation that `vars-float-up` hoists out a moment later, leaving a stray `Φ.number`. The `@float-up` formation is now excluded from the element list and length count and emitted only as a passenger child (so `vars-float-up` still hoists its definition and drops the in-place node); `wrap-applications` now also marks such a predecessor inside a star tuple (parent has `@star`, not only `@base`).

## Fold itself (`merge-monikers.xsl`)

A nameless applied reference (bare `ξ.name` naming the restored recursive handle, carrying argument children) inlines the formation in place and turns its arguments into a `| args` pipe continuation binding it — the recursive mirror of #5844. A reference carrying its own `@name` is excluded (a `| args > name` pipe already folded by #5837/#5848); a nameless one round-trips.

## Verification

- `dispatch-recursive-applied-receiver.yaml` and `list-recursive-applied-dispatch-arg.yaml` reprint to themselves.
- The folded `directory.eo` was checked to parse to the **same object graph** as the original (semantically lossless), is canonical under `eo:format`, and its `deleted` helper dataizes correctly.
- Full `eo-parser` and `eo-printer` test suites pass.


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (190 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2551](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5856, opened 2026-07-24
#5776: make chunk.put strictly-sized, reject overflow instead of resizing

### LINKED ISSUE #5776: `chunk.put` incorrectly expands memory block instead of rejecting oversized objects
## What happens

`chunk.put` (the memory-block `put` reached through `malloc.of`/`malloc.for`) is documented and tested as a **strictly-sized** overwrite: it writes an object at offset 0 of a fixed-size block. In practice it silently **grows** the block whenever the object is larger than the block, contradicting that contract.

Verified by building `eo-runtime` from source and dataizing real calls:

```
malloc.for
  "Hello"                      # allocates a 5-byte block
  m.put "Much longer string!"  # 19 bytes
```

returns the full `"Much longer string!"` (19 bytes), and `m.size` afterwards is `19` — the block grew from 5 to 19 bytes. A strict, fixed-size `put` should reject this (as its sibling `on-overflow-*` tests intend), not resize.

## Root cause

`chunk.put` is defined in terms of `write`, which auto-resizes:

https://github.com/objectionary/eo/blob/667b081eb6d1ca0c1d8ea3611dd273e51d770cfd/eo-runtime/src/main/eo/chunk.eo#L39-L42

```
seq * > [object] > put
  write
    0
    object
  get
```

and `Heaps.write` grows the block when the data does not fit:

https://github.com/objectionary/eo/blob/667b081eb6d1ca0c1d8ea3611dd273e51d770cfd/eo-runtime/src/main/java/org/eolang/Heaps.java#L213-L215

```java
if (this.blocks.get(identifier).length < end) {
    this.resize(identifier, (int) end);
}
```

So `put` inherits `write`'s grow-on-overflow behaviour. That is the right thing for `write` (its own test `writes-beyond-offset-and-resizes` relies on it) and there is a separate explicit `resized` for changing size, but it means `put` is **not** the strict same-size overwrite the design implies.

## The overflow tests are vacuous

`malloc.eo` has two negative examples that are clearly meant to assert `put` rejects an oversized object:

https://github.com/objectionary/eo/blob/667b081eb6d1ca0c1d8ea3611dd273e51d770cfd/eo-runtime/src/main/eo/malloc.eo#L362-L369

```
malloc.for --> on-overflow-boolean-malloc
  false
  m.put 8.612486788E7 > [m]

malloc.for --> on-overflow-string-malloc
  "Hello"
  m.put > [m]
    "Much longer string!"
```

Both pass today, but **not** because `put` fails. A `-->` example generates `assertThrows(Exception, () -> new Dataized(this.take("-name")).asBool())`. Here `put` succeeds and returns the grown block's bytes (a 19-byte string / an 8-byte number); the assertion passes only because `asBool()` throws on those non-boolean bytes. So the tests give false confidence that overflow is rejected while the block silently grows. The same masking is why `tests-malloc-is-strictly-sized-int` / `tests-memory-is-strictly-sized-string` never exercise a genuinely oversized `put` — they only ever write same-or-smaller data (`m.put 2556` into a `12248`-sized block, `m.put "Prot"` into `"Hello"`).

## Failure scenarios

Verified against the compiled runtime:

| call | result | expected (strict) |
| --- | --- | --- |
| `malloc.for "Hello" (m.put "Much longer string!")` | `"Much longer string!"`, block size `19` | error / `cant-*` fallback, size stays `5` |
| `malloc.for false (m.put 8.612486788E7)` | 8-byte number, block size `8` | overflow error, size stays `1` |
| `malloc.for "Hello" (m.put "Prot")` | `"Proto"` (size `5`) | `"Proto"` (unchanged — control) |

## Suggested resolution

Either:

1. Make `put` strict — check that the object's byte length does not exceed the current block size and route the overflow to a fallback / terminator instead of growing (leaving `write` as the grow-capable primitive and `resized` as the explicit resize) — and then have the `on-overflow-*` examples assert the actual failure (e.g. via a bound value the caller can observe), not merely rely on `asBool` throwing; or
2. If growing is intended, update the docstring and the `strictly-sized` / `on-overflow-*` test names to say so, since today they claim a guarantee the code does not provide.

Either way, the `on-overflow-*` examples should be rewritten so they fail if `put` ever stops rejecting overflow, rather than passing vacuously through `asBool`.

morphqdd:
Closes #5776

`chunk.put` was implemented on top of `write`, which auto-resizes the block when the data does not fit:

```
seq * > [object] > put
  write
    0
    object
  get
```

`Heaps.write` grows the block whenever `end > current size`. This meant `put` silently grew the block whenever the object was larger than it, contradicting its documented strictly-sized overwrite contract. As reported: `malloc.for "Hello" (m.put "Much longer string!")` returned the full 19-byte string and grew the block from 5 to 19 bytes.

Per @yegor256's own comment on the issue: chunk should never automatically resize; only `.resized` should change the block size.

Fix: `put` now compares the dataized object's byte length against the chunk's current `size` before writing, and throws instead of writing when the object does not fit:

```
if. > [object] > put
  object.size.gt size
  T
    "Can't put %d bytes into a chunk sized %d bytes; call .resized first to grow it".printf
      * object.size size
  seq *
    write 0 object
    get
```

`write` itself is untouched (its own `writes-beyond-offset-and-resizes` test still relies on the grow behavior), as is `resized` (the explicit resize API) and `copy`.

The existing `on-overflow-boolean-malloc` / `on-overflow-string-malloc` tests in `malloc.eo` were previously vacuous: they passed only because `put` returned non-boolean bytes that made the test's own `asBool()` throw, not because `put` itself rejected the overflow (as the issue pointed out). They now pass because `put` genuinely throws for the right reason. No test changes were needed for them; the fix alone corrects what they exercise.

Update: the first version of this fix was too strict. `malloc.empty` deliberately allocates a size-0 block so its first `.put` gives it a real size (that is the whole point of `malloc.empty`, and the README's own canonical iteration example, `malloc.empty > [args] > app` then `x.put 2`, relies on exactly this). Rejecting that broke `ReadmeSnippetsIT` in CI. Fixed by only rejecting overflow when the chunk already has a non-zero size:

```
if. > [object] > put
  and.
    size.gt 0
    object.size.gt size
  T
    "Can't put %d bytes into a chunk sized %d bytes; call .resized first to grow it".printf
      * object.size size
  seq *
    write 0 object
    get
```

Growing from size 0 (not yet given a size) is still allowed; a chunk that already has a real size never silently changes. Added `tests-malloc-empty-grows-on-first-put` to cover this.

Verified:
- `EOmallocEOAtomTest`: 37/37 green, including both `on-overflow-*` tests and the new empty-grows test.
- Full `eo-runtime` test module: 1758/1758 green, 0 failures.
- `qulice:check -Pqulice` on `eo-runtime`: no new violations from this change (confirmed the same 100 pre-existing `UnicodeInCode` violations, unrelated to `chunk.eo`, are also present on `upstream/master` before this change, caused by local JDK 21 vs CI's JDK 26 ErrorProne behavior).
- Reran the actual failing CI test locally: `mvn -pl eo-integration-tests install -o -Psnippets -Dit.test='ReadmeSnippetsIT'` — now 5/5 green.

### DIFF
--- a/eo-runtime/src/main/eo/chunk.eo
+++ b/eo-runtime/src/main/eo/chunk.eo
@@ -37,7 +37,9 @@
     0
     size
   if. > [object] > put
-    object.size.gt size
+    and.
+      size.gt 0
+      object.size.gt size
     T
       "Can't put %d bytes into a chunk sized %d bytes; call .resized first to grow it".printf
         * object.size size
--- a/eo-runtime/src/main/eo/chunk.eo
+++ b/eo-runtime/src/main/eo/chunk.eo
@@ -36,11 +36,14 @@
   read > get
     0
     size
-  seq * > [object] > put
-    write
-      0
-      object
-    get
+  if. > [object] > put
+    object.size.gt size
+    T
+      "Can't put %d bytes into a chunk sized %d bytes; call .resized first to grow it".printf
+        * object.size size
+    seq *
+      write 0 object
+      get
   [offset length] > read /Q.bytes
     ? > cant-read /{Q.string}
   [size] > resized /Q.chunk
--- a/eo-runtime/src/main/eo/malloc.eo
+++ b/eo-runtime/src/main/eo/malloc.eo
@@ -170,6 +170,11 @@
               (m.resized 3).size.eq 3
               m.get.eq 01-02-03
 
+  eq. ++> tests-malloc-empty-grows-on-first-put
+    malloc.empty
+      m.put 42 > [m]
+    42
+
   malloc.empty ++> tests-malloc-empty-is-empty
     m.size.eq 0 > [m]
 


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1932](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5859, opened 2026-07-24
Fix spurious numeric element in `*` star tuple with `|` pipe continuation (#5858)

### LINKED ISSUE #5858: eo-printer erroneously adds `1` to `*` tuples containing a `|` pipe continuation
The printer corrupts a `*` star-tuple whose elements include a `|` pipe continuation: it injects a numeric literal that is **not present anywhere in the source**, so the format is both non-idempotent and lossy/incorrect.

## Reproduce

Feed `eo-printer` this snippet (a `seq *` tuple whose first element is a `>> ` handle formation, followed by a `| walked` pipe continuation and a `^`):

```
[] > foo
  seq * > @
    [q] >> h
      q > @
    | walked
    ^
```

The printer emits:

```
[] > foo
  seq * > @
    1
    q > [q] >> h
    | walked
    ^
```

A `1` appears out of nowhere as the first tuple element. There is **no numeric literal anywhere in the input**, so this is a synthesized value, not a reordering. The pipe is essential to the trigger: dropping the `| walked` line makes the `1` disappear (the tuple then prints as `q > [q] >>` / `^`, no spurious element).

## Why it is wrong

- **Not present in source.** The output contains a `1` the author never wrote, changing the tuple's contents (`seq(tuple(h walked, ^))` becomes `seq(tuple(1, h walked, ^))`).
- **Non-idempotent.** `printsToParseableEo`-style round-tripping diverges: printing the graph, re-parsing, and re-printing keeps drifting because the injected `1` re-enters the tuple.

## Observations on the cause

The parser's own output is clean — dumping the tree after the parse sheets (through `stars-to-tuples`) shows the tuple as `[handle-formation, «@pipe walked», ^]` with **no** numeric element. The `1` (and, in richer inputs, further `2`/`3` `Φ.number` nodes) appears only once the print/round-trip pipeline reconstructs the star tuple, in the region around `stars-to-tuples` ⇄ `tuples-to-stars` (`eo-printer/src/main/resources/org/eolang/printer/print/tuples-to-stars.xsl`) and the trailing-star hybrid in `eo-printer/src/main/java/org/eolang/printer/Pretty.java` (`starred()` / `Node.tuply()`). The trigger is specifically a `*` tuple element that carries a `@pipe` continuation binding a preceding element: the star-reconstruction appears to miscount elements and materialize the count as a `Φ.number` tuple member instead of gluing a `*N` marker onto the head.

This is adjacent to the trailing-star work in #5622 / #5624 / #5648 but is a distinct, opposite symptom (a numeric element *injected* rather than a `*N` marker *omitted*), and it is triggered by the `|` pipe continuation (§3.14) inside the tuple.

## Context

Surfaced while extending the recursive `>> name` handle fold (#5848) to an applied handle that is a tuple element — the fold would host the handle as a moniker inside a `seq *` and add `| args`, which is exactly the shape above. The dispatch-receiver fold (`(handle args).method`) is unaffected because its reference is not a tuple element. This printer bug is independent of that fold: it reproduces with a plain, non-recursive handle and no folding involved.

yegor256:
Fixes #5858.

## Problem

`eo-printer` injected a numeric literal that is not present anywhere in the source when printing a `*` star tuple whose element carries a `|` pipe continuation (§3.14):

```
[] > foo
  seq * > @
    [q] >> h
      q > @
    | walked
    ^
```

printed as (note the spurious `1`):

```
[] > foo
  seq * > @
    1
    q > [q] >> h
    | walked
    ^
```

This made the format both lossy (`seq(tuple(h walked, ^))` became `seq(tuple(1, h walked, ^))`) and non-idempotent.

## Root cause

Each `Φ.tuple` layer built by the parser's `stars-to-tuples.xsl` holds three children: the nested tuple (`o[1]`), the appended element (`o[2]`), and a trailing `Φ.number` that carries the tuple's length (`o[last()]`).

A `| pipe` continuation whose predecessor formation is floated out of the argument slot by `vars-float-up.xsl` runs *after* `stars-to-tuples` has already built the nested tuple with its per-layer length markers. Floating the formation removes the element but leaves the length marker, producing a layer with only `{nested tuple, Φ.number}` and no element in between.

The reverse `tuples-to-stars.xsl` sheet always read a fixed `o[2]` as the element, so for such a gap layer it mis-read the `Φ.number` length marker as an actual element — emitting the spurious `1`.

## Fix

Select the element as the range strictly between the nested tuple `o[1]` and the trailing length marker `o[last()]`:

```
o[position() != 1 and position() != last()]
```

A normal layer still yields its single element; a floated (gap) layer yields nothing, so no length marker leaks into the star. Applied to both the top-level template and the recursive `inner` mode.

## Tests

- Adds `star-tuple-with-pipe-continuation.yaml` to `print-packs/yaml`, covering the exact reported shape. It exercises both `printsToEo` (correct output, no spurious element) and `printsToParseableEo` (idempotent round-trip).
- Full `eo-printer` suite: 216 tests pass.

https://claude.ai/code/session_01RtsjS8rJxda978pgzBPXsL

---
_Generated by [Claude Code](https://claude.ai/code/session_01RtsjS8rJxda978pgzBPXsL)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/tuples-to-stars.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/tuples-to-stars.xsl
@@ -5,12 +5,26 @@
 -->
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" id="tuples-to-stars" version="2.0">
   <!--
-  Performs the reverse operation of "/org/eolang/parser/stars-to-tuples.xsl"
+  Performs the reverse operation of "/org/eolang/parser/stars-to-tuples.xsl".
+
+  Each "Φ.tuple" layer built by that sheet holds three children: the
+  nested tuple (o[1]), the element it appends (o[2]), and a trailing
+  "Φ.number" that carries the tuple's length (o[last()]). The element is
+  therefore whatever sits strictly between the nested tuple and that
+  length marker — selected here as "o[position() != 1 and position() !=
+  last()]" rather than a fixed o[2].
+
+  A "| pipe" continuation (§3.14) whose predecessor formation is floated
+  up out of the argument slot ("vars-float-up.xsl") leaves a layer with
+  only its nested tuple and length marker and no element in between. The
+  positional range then selects nothing, so the layer contributes no star
+  element — where a fixed o[2] would have mis-read the length marker as a
+  spurious numeric element (#5858).
   -->
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:template match="o[@base = 'Φ.tuple' and o[1][starts-with(@base, 'Φ.tuple')]]">
     <xsl:variable name="arg">
-      <xsl:apply-templates select="o[2]"/>
+      <xsl:apply-templates select="o[position() != 1 and position() != last()]"/>
     </xsl:variable>
     <xsl:copy>
       <xsl:apply-templates select="@*"/>
@@ -33,7 +47,7 @@
   <xsl:template match="o" mode="inner">
     <xsl:if test="@base = 'Φ.tuple'">
       <xsl:variable name="arg">
-        <xsl:apply-templates select="o[2]"/>
+        <xsl:apply-templates select="o[position() != 1 and position() != last()]"/>
       </xsl:variable>
       <xsl:apply-templates select="o[1]" mode="inner"/>
       <xsl:apply-templates select="$arg" mode="no-as"/>
--- a/eo-printer/src/main/resources/org/eolang/printer/print/tuples-to-stars.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/tuples-to-stars.xsl
@@ -18,7 +18,7 @@
   up out of the argument slot ("vars-float-up.xsl") leaves a layer with
   only its nested tuple and length marker and no element in between. The
   positional range then selects nothing, so the layer contributes no star
-  element — where a fixed o[2] would have mis-read the length marker as a
+  element — where a fixed o[2] would have taken the length marker for a
   spurious numeric element (#5858).
   -->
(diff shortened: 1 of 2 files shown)


yegor256:
Heads-up on the two red checks — both look flaky and unrelated to this diff, which only touches `eo-printer/.../tuples-to-stars.xsl` (an XSL sheet) plus a test YAML:

- **`mvn (windows-2022, 26)`** failed in **eo-runtime** with `java.lang.OutOfMemoryError: Java heap space` in `EObytesEOconcatTest.failsCleanlyInsteadOfOverflowingArraySize` (a test that deliberately provokes a huge allocation). The identical commit's **`mvn (ubuntu-24.04, 26)` passed** the full build+tests, so this is an environmental heap flake, not a code failure. The two macOS mvn jobs were `cancelled` by matrix fail-fast when Windows failed, not real failures.
- **`integration`** failed in **eo-integration-tests** on `SnippetIT.runsAllSnippets[2]`, where a snippet terminated during the **parse** stage via `xsl:message` in `set-locators.xsl` — a parser sheet — under concurrent compilation.

`eo-printer` builds and tests green on every mvn job, and an eo-printer-only change cannot reach an eo-runtime unit test or the snippet parse pipeline. A re-run of the two failed jobs should clear them. Everything else (typos, qulice, all linters, SonarCloud, Codacy, benchmark, snippets, ubuntu mvn) is green.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2563](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5872, opened 2026-07-24
#5851: reject formatting a source when part of it was not recovered

### LINKED ISSUE #5851: `MjFormat` incorrectly formats unparsed sources, disregard of parse errors or criticals in XMIR
The `format` goal never checks whether a source actually parses. `MjFormat.canonical()` (`eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java:161-171`) parses each `.eo` source with `new EoSyntax(structure).parsed()` and prints the result straight back with `Xmir.toEO()`, but it never inspects the produced XMIR for `<errors>` elements whose `severity` is `error` or `critical`.

This is wrong because a source that fails to parse still yields an XMIR that carries those `<errors>` (compare `Parsing`, which logs them at `Parsing.java:214-231`, and `Linting`, which fails the build on them at `Linting.java:248-255`). `MjFormat` treats that partial XMIR as valid input, so a broken source is silently "reformatted": in check mode the parse failure is never surfaced, and under `-Deo.autoFix` the file can be overwritten with a canonical form derived from an incomplete parse, destroying the original.

A fix: after parsing inside `canonical()`, inspect `/object/errors/error` and throw an `IllegalStateException` when any `severity` equals `error` or `critical`, mirroring the guard already used in `Linting.linting()` (`Linting.java:248-255`), so the goal stops loudly instead of formatting a source that does not parse.

morphqdd:
Closes #5851

`MjFormat.canonical()` parsed each source with `new EoSyntax(structure).parsed()` and printed the result straight back, but never inspected the produced XMIR for `<errors>`. A source that failed to parse still yields an XMIR carrying those `<errors>` (compare `Parsing`, which logs them, and `Linting`, which fails the build on them). `MjFormat` treated that partial XMIR as valid input: in check mode the parse failure was never surfaced, and under `-Deo.autoFix` the file could be overwritten with a canonical form derived from an incomplete parse, destroying the original.

The literal fix suggested on the issue (throw when any `<error>` has `severity` equal to `error` or `critical`) turned out to be too strict once I tested it against the real parser. I dumped the actual XMIR for two cases:

- A source with extra blank lines (exactly what this goal exists to fix): the parser reports `severity="error"` for `"consecutive blank lines forbidden"` and `"more than one trailing blank line"`, but the object tree is fully intact, nothing is lost.
- A source with a genuine syntax break (an unterminated paren group): also `severity="error"`, but the object tree stops short, the real body of the object is missing entirely.

`Emit.java` (the parser's own error-emission code) hardcodes `severity="error"` for every message it emits, regardless of which rule fired. There is no severity distinction between "cosmetic layout nit" and "genuine content loss" in this parser today. A blanket "reject on any severity=error" check would make `MjFormat` reject its own core use case, and I confirmed this concretely: it broke the existing `reformatsDivergentSourceWhenAutoFixIsOn` test.

Fix: instead of trusting `severity` alone, also check whether the parsed tree's deepest `<o>` line reference reaches the source's last non-blank line. A genuine syntax error makes the parser stop incorporating source lines into the tree from that point on; a cosmetic layout error does not lose any line. Only when both conditions hold (a severe error is present, and the tree is missing content) does this now throw:

```java
if (errors > 0L && MjFormat.truncated(xmir, structure)) {
        String.format(
            "%s does not fully parse (%d error(s) found) and part of it was lost, won't format it",
            source, errors
        )
    );
}
```

Verified:
- `MjFormatTest`: 6/6 green, including a new `failsWhenSourceDoesNotParse` test (an unterminated paren group now throws with a clear message) and the pre-existing `reformatsDivergentSourceWhenAutoFixIsOn` test (extra blank lines still get fixed), both passing together.
- Full `eo-maven-plugin` test module: 350/350 green (excluding the pre-existing `MjPrintTest` failures, also present on unmodified `upstream/master`, unrelated to this change).
- Ran the real `format` goal against all 153 real `eo-runtime` `.eo` sources: all still report as canonically formatted, no false positives from the new check.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
@@ -4,7 +4,10 @@
  */
 package org.eolang.maven;
 
+import com.github.lombrozo.xnav.Filter;
+import com.github.lombrozo.xnav.Xnav;
 import com.jcabi.log.Logger;
+import com.jcabi.xml.XML;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Collection;
@@ -122,7 +125,7 @@ public final class MjFormat extends MjSafe {
      */
     private boolean reformat(final Path source) throws IOException {
         final String actual = new UncheckedText(new TextOf(source)).asString();
-        final String canonical = this.canonical(actual);
+        final String canonical = this.canonical(source, actual);
         final Diff diff = new Diff(actual, canonical);
         final boolean diverged = !diff.same();
         if (diverged && this.autoFix) {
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (123 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1944](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5877, opened 2026-07-25
Fix reversed dispatch dropping the file-local handle of its `&gt;&gt;` suffix

### LINKED ISSUE #5874: Reversed dispatch drops the file-local handle of its `>>` suffix
A `>>` handle attached to a reversed dispatch — the vertical form with an implicit receiver, like `not. >> first` or `if. >> index!` — never reaches the XMIR. In `eo-parser/src/main/java/org/eolang/parser/LnReversed.java:98-108` the line emits `suffix.attribute(...)` (the synthetic cactus name), then `emit.fragile()` and `emit.constant()`, but it never calls `emit.local(suffix.handle())`. Every other line type that accepts an auto suffix does: `LnFormation.java:200`, `LnOnlyPhi.java:140`, `LnVoid.java:98`, `ChainEmission.java:87` and `:109`.

Without the `@local` marker, `resolve-local-names.xsl` cannot see the handle (its `eo:captor` function looks for a `descendant::o[@local=$name]`), so every reference to the handle name is left untouched and `build-fqns` resolves it to a global `Φ.<name>` that does not exist. The declaration itself keeps its cactus name, so the name the author wrote is simply gone: nothing in parsing or linting complains, and the object fails only when dataized, with an unrelated message such as "the 'x' attribute must be a number". Writing the same handle with an explicit receiver — `(a.eq 2).not >> second` — works and produces `local="second"`, which makes the defect look like a syntax subtlety rather than a bug. In `eo-runtime` on branch 5678 this bit `if. >> index!` in `tuple.eo`, `if. >> final`, `times. >> prod`, `left. >> q0`, `plus. >> q0plus1` in `u64.eo` and `not. >> has-authority-path` in `uri.eo`, whose parsed XMIR held `base="Φ.index"`, `base="Φ.final"` and friends.

Adding the two lines that the other handlers already have — `if (!suffix.handle().isEmpty()) { emit.local(suffix.handle()); }` right after the `emit.object(...)` call in `LnReversed.into` — is enough to fix it.

yegor256:
Closes #5874.

## Problem

A `>>` handle attached to a vertical reversed dispatch with an implicit receiver — like `not. >> first` or `if. >> index!` — never reached the XMIR. In `LnReversed.into` the code emitted the synthetic cactus name via `suffix.attribute(...)`, then `emit.fragile()` and `emit.constant()`, but it never called `emit.local(suffix.handle())`.

Without the `@local` marker, `resolve-local-names.xsl` could not see the handle (its `eo:captor` function looks for a `descendant::o[@local=$name]`), so every reference to the handle name was left untouched and `build-fqns` resolved it to a global `Φ.` that does not exist. The object then failed only at dataization with an unrelated message.

## Fix

Emit the handle right after the `emit.object(...)` call, matching every other line type that accepts an auto suffix (`LnFormation`, `LnOnlyPhi`, `LnVoid`, `ChainEmission`):

```java
if (!suffix.handle().isEmpty()) {
    emit.local(suffix.handle());
}
```

## Test

Added `LnReversedTest#emitsLocalHandleForAutoSuffix`, asserting that `if. >> index` emits `<o local='index' base='.if'>` (no `@method`). All 140 tests in the affected parser suites pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_0142yX5DVzddKf2FfKQ2RbNf)_

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnReversed.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnReversed.java
@@ -100,6 +100,9 @@ final class LnReversed implements Line {
             ".".concat(head.raw()),
             this.span.line(), this.span.indent()
         );
+        if (!suffix.handle().isEmpty()) {
+            emit.local(suffix.handle());
+        }
         if (fragile) {
             emit.fragile();
         }
--- a/eo-parser/src/main/java/org/eolang/parser/LnReversed.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnReversed.java
@@ -95,10 +95,30 @@ final class LnReversed implements Line {
         Bindings.observeChild(stack, outer, this.span);
         globals.clearBlanks();
         globals.markEmitted();
+        this.emit(emit, suffix, ".".concat(head.raw()), fragile, args, outer);
+    }
+
+    /**
+     * Emit the reversed dispatch's {@code <o base='.<name>'>}, its
+     * file-local handle (for a {@code >> name} auto suffix, #5874), the
+     * fragile/const markers, the horizontal args, and the outer binding
+     * slot. Mirrors the auto-suffix handling of {@link LnFormation},
+     * {@link LnOnlyPhi} and {@link LnVoid}.
+     * @param emit The directives sink
+     * @param suffix The parsed suffix
+     * @param base The dispatch base ({@code .<name>})
+     * @param fragile Whether the dispatch carries the fragile marker
+     * @param args Horizontal args in source order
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2671](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5878, opened 2026-07-25
#5875: scope the duplicate-local-name check to the enclosing formation

### LINKED ISSUE #5875: Duplicate-local-name check is file-wide although handle resolution is lexically scoped
The check at `eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl:61-72` groups `//o[@local]` by `@local` over the whole document and reports `duplicate local name 'x'` for any repetition anywhere in the file. Reference resolution in the same stylesheet is lexically scoped: `eo:captor` (lines 44-49) binds a reference to the nearest enclosing formation that declares the name, and its `is` test deliberately stops at a nested formation boundary so that a handle "never leaks up out of, or sideways between, nested formations (#5780)".

So two sibling formations that each declare a handle of the same name are unambiguous by the resolver's own rules, yet the file is rejected. The result is that a private name has to be unique per file rather than per scope, which forces invented names in exactly the places where privacy is wanted: in `eo-runtime`, `console.eo` cannot give both `posix-console` and `windows-console` a `>> input-block`, `u16.eo` cannot give `minus`, `plus` and `times` one `>> right` each, `file.eo` cannot hide the two `info` helpers that live in different sub-objects, and unit tests in one file cannot each declare a private `>> obj` (`bytes/hash.eo` binds `obj` in three separate tests). This blocks part of #5678.

Grouping by scope rather than by name alone fixes it: key the `for-each-group` on the pair of `@local` and the generated id of `ancestor::o[not(@base)][1]`, so a duplicate is only reported when two handles of the same name share the same nearest enclosing formation.

yegor256:
Closes #5875.

## Problem

The duplicate-local-name check in `resolve-local-names.xsl` grouped every `//o[@local]` by `@local` across the whole document, reporting `duplicate local name 'x'` for any repetition anywhere in the file. But reference resolution in the same stylesheet is **lexically scoped**: `eo:captor` binds a reference to the nearest enclosing formation that declares the name, and its `is` test deliberately stops at a nested formation boundary so that a handle never leaks up out of, or sideways between, nested formations (#5780).

So two sibling formations that each declare a handle of the same name are unambiguous by the resolver's own rules, yet the file was rejected — forcing invented names in exactly the places where privacy is wanted (e.g. `console.eo` couldn't give both `posix-console` and `windows-console` a `>> input-block`).

## Fix

Key the `for-each-group` on the pair of `@local` and the generated id of the nearest enclosing formation (`ancestor::o[not(@base)][1]`), so a duplicate is only reported when two handles of the same name share the same enclosing formation. The error text now reads the name from `current-group()[1]/@local` instead of the composite grouping key.

## Tests

- Added `resolve-local-names-duplicate-sibling-scope.yaml`: two sibling formations each declaring `>> dup` now parse without error, each keeping its own cactus name.
- Existing same-scope duplicate packs (`resolve-local-names-duplicate.yaml`, `void-local-handle-duplicate.yaml`) still report the error, since their handles share one enclosing formation.

All 46 `EoSyntaxTest#checksEoPacks` cases pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BCRbYwaUWHSLHPzLzzzFvG)_

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
@@ -9,7 +9,9 @@
   anonymous object with its cactus @name plus a "@local='foo'" marker. A
   handle is an attribute of the formation that declares it, so we rewrite a
   @base equal to a handle name into that (reserved) cactus name; a handle
-  declared more than once is an error. The "@local" marker is deliberately
+  declared more than once within the same enclosing formation is an error,
+  but two sibling formations may each declare a same-named handle because
+  resolution is lexically scoped (#5875). The "@local" marker is deliberately
   kept on the declaring object so that later passes (in particular the
   printer, see #5563) can recover the readable handle from the otherwise-
   synthetic cactus name instead of printing a placeholder like "vL_P".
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2675](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5879, opened 2026-07-25
Keep multi-referenced abstract handle whole in printer (#5876)

### LINKED ISSUE #5876: inline-cactoos duplicates a multi-referenced `>> name` formation and leaves a dangling `vL_P` inside the copies
`inline-cactoos.xsl` folds a reference to an auto-named abstract into the reference site, and it makes exceptions for a void, a self-referential target and a multi-referenced *const* handle (#5828), see the comment at `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:33-60`. A plain (non-const) `[] >> name` formation handle that is referenced more than once gets no such exception, so the whole formation is copied into every reference and its handle name is dropped — `to-eo-tree` prints the copies as anonymous `[] >>` blocks.

The copies then break: a handle declared as a sibling of the inlined formation is out of scope inside a copy, so its cactus name survives the pipeline and `to-eo-tree.xsl:39-42` rewrites it to a synthetic `vL_P` id that is declared nowhere, so the printed source no longer parses back to the same object (the same synthetic-name symptom as the recursive case in #5681). Running `eo:format` over `eo-runtime/src/main/eo/number/mod.eo` on branch 5678, where `mod` holds `[] >> abs-mod` (referenced twice, as `abs-mod` and `abs-mod.neg`) with `abs dividend >> absx` and `abs divisor >> absy` beneath it, produces two anonymous copies of `abs-mod` and, inside each, `v27_4.div` in place of `absx`. Twenty-four of the 153 runtime sources come out with a `vL_P` reference this way, which makes `-Deo.autoFix` unsafe on any file that keeps a shared private helper.

The smallest fix is to extend the existing "left untouched" guard: treat a multi-referenced non-const auto-named abstract the same way the multi-referenced const handle is treated — keep the binding and its `@local` handle in place and let `merge-monikers` host it on the first reference.

yegor256:
Fixes #5876.

## Problem

`inline-cactoos.xsl` folds a reference to an auto-named abstract into the reference site, making exceptions only for a void, a self-referential (recursive) target, and a multi-referenced *const* handle (#5828). A plain (non-const) `[] >> name` abstract formation referenced more than once got **no** exception, so the whole formation was copied into every reference:

- its shared handle name was dropped — `to-eo-tree` printed the copies as anonymous `[] >>` blocks;
- a handle declared as a sibling of the formation is out of scope inside a copy, so its cactus name survived the pipeline and `to-eo-tree` rewrote it to a synthetic `vL_P` id declared nowhere, so the printed source no longer parsed back to the same object (the same symptom as the recursive case in #5681).

## Fix

Treat a multi-referenced non-const auto-named abstract formation the same way a multi-referenced const handle is treated — keep the formation and its `@local` handle in place and let `merge-monikers` host it on the first reference:

- **`restore-local-names.xsl`** — keep the `@local` marker on a multi-referenced abstract formation (previously dropped, since `inline-cactoos` was expected to inline the formation away), so its readable handle survives to `to-eo-tree`.
- **`inline-cactoos.xsl`** — extend the "left untouched" guard and the drop template so a multi-referenced abstract formation is kept whole rather than copied into each reference. Reference counting now also counts dispatch uses (`name.seg`, whose resolved name carries the bare name as its leading segment), so a formation reached from a bare use *and* a dispatch use is recognised as shared.
- **`merge-monikers.xsl`** — the existing moniker machinery already hosts the kept binding onto its first reference; add the abstract mirror of `eo:kept-const-ref` so every *other* reference (a named alias `name > a` or a further `name.seg` dispatch) reads back through the readable `@local` handle instead of a `vL_P` placeholder.

## Tests

- `print-packs/yaml/abstract-multi-referenced-handle.yaml` — round-trip where the kept formation is hosted onto its first reference as a moniker and the remaining `helper.neg` dispatch reads back readably (the abstract counterpart of `const-multi-referenced-handle.yaml`).
- `print-packs/yaml/abstract-multi-referenced-handle-kept.yaml` — round-trip where no reference can host the formation (named aliases), so the binding is left in place and every alias reads through the readable handle.
- `eo-packs/print/inline-cactoos-multi-referenced.yaml` — asserts at the `inline-cactoos` level that the multi-referenced formation and both of its references are kept rather than copied.

The full `eo-printer` suite (225 tests) passes.

## Scope note

This addresses the abstract-formation duplication the issue targets. The `absx`/`absy`-style residual `vL_P` seen in `mod.eo` comes from a **separate, pre-existing** defect: a non-const *based* `>> name` handle (e.g. `x.abs >> absx`) reached only through method dispatch is kept but loses its `@local`. That reproduces with a single-referenced formation and with no abstract formation at all, so it is independent of the multi-reference fixed here and would be a follow-up change.

https://claude.ai/code/session_01Vo3idnjfN9CDuhhz4MyLEZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vo3idnjfN9CDuhhz4MyLEZ)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -44,7 +44,12 @@
   the const is dataized once and cached in that single binding, so folding
   it into each use would mint an independent const object per reference and
   drop the shared name (#5828); "merge-monikers" later hosts the kept
-  binding onto its first reference.
+  binding onto its first reference. A plain (non-const) auto-named abstract
+  formation reached from more than one site is left untouched for the same
+  reason (#5876): folding the whole formation into every reference drops its
+  shared handle name and strands any sibling helper it reaches on a synthetic
+  "vL_P" id, so the formation and its "@local" handle are kept in place and
+  "merge-monikers" hosts the kept binding onto its first reference.
 
   The inlined value keeps the target's obfuscated cactus `@name` only when
   the name is still meaningful downstream: an abstract formation, whose name
@@ -66,7 +71,7 @@
     <xsl:variable name="target" select="ancestor::o/o[@name=$name][1]"/>
     <xsl:variable name="keep-name" as="xs:boolean" select="exists($target) and (eo:abstract($target) or ($target/@base = '.as-bytes' and $target/o[1]/@base = 'Φ.dataized' and eo:abstract($target/o[1]/o[1])))"/>
     <xsl:choose>
-      <xsl:when test="exists($target) and not(eo:void($target)) and not(eo:recursive($target, $name)) and not(eo:dataized-const($target) and eo:multi-referenced($target, $name))">
+      <xsl:when test="exists($target) and not(eo:void($target)) and not(eo:recursive($target, $name)) and not(eo:multi-referenced($target, $name) and (eo:dataized-const($target) or eo:abstract($target)))">
         <xsl:choose>
           <!--
           The reference is the base of an application — it carries its own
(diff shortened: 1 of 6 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (197 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2679](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5884, opened 2026-07-25
Keep dispatch-referenced const `>>` handle whole (#5883)

### LINKED ISSUE #5883: Multi-referenced const `>>` handle is folded into one use, stranding its other references on `vL_P`
A const file-local handle referenced more than once must be kept whole — that is what `const-multi-referenced-handle.yaml` states after #5828. It is not, unless the const happens to wear the `.as-bytes`/`Φ.dataized` wrapper. This origin:

```
[text] > foo
  mincode.gte 1 > bar
  minus. > distance
    number mincode
    42
  text.plus 1 >> mincode!
```

prints as:

```
[text] > foo
  v6_2.gte 1 > bar
  minus. > distance
    number (text.plus 1!)
    42
```

`mincode` has two references, yet the handle is folded into the `distance` use and the `bar` use is left on the synthetic `v6_2`, a name declared nowhere, so the printed source no longer denotes the same object (and `v6_2` resolves to nothing on reparse). The same happens with a data target (`5 >> mincode!` → `v6_2` plus an inline `5!`).

The keep-whole guard is eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:74, which requires `eo:dataized-const($target)` or `eo:abstract($target)`. `eo:dataized-const` (inline-cactoos.xsl:205-208) recognises only a `.as-bytes` node over `Φ.dataized`, the shape `const-to-dataized` leaves behind for a const handle over a plain reference. A const handle over an application or over a datum never gets that shape, so the guard misses it, `eo:multi-referenced` is ignored, and the fold strands every reference but one. In `eo-runtime` this hit `string/up-cased.eo` (`as-ascii "a" >> mincode!`), `os.eo` (`name >> oname!`) and `tuple/back.eo` while working on #5678.

Widening the guard is enough: hold it for any const `>>` handle, not just the dataized-wrapper shape — e.g. test the target's const marker in `eo:dataized-const` instead of its `@base` — so a multi-referenced const handle is kept whole however the const is encoded.

yegor256:
Fixes #5883.

## Problem

A const file-local handle (`a >> b!`, #5817) referenced more than once must be kept whole (#5828). But when one of its references was a **method dispatch** (`b.gte 1`, base `ξ.b.gte`) rather than a bare `b`, the handle was folded into one use and the other reference was stranded on a synthetic `vL_P` name that resolves to nothing on reparse:

```
[text] > foo
  mincode.gte 1 > bar          # dispatch reference
  minus. > distance
    number mincode             # bare reference
    42
  text.plus 1 >> mincode!
```

printed as (broken):

```
[text] > foo
  v6_2.gte 1 > bar             # stranded on a name declared nowhere
  minus. > distance
    number (text.plus 1!)      # const folded into this one use, shared name lost
    42
```

## Root cause

Contrary to the issue's initial guess at `inline-cactoos`, a step-by-step trace shows `inline-cactoos` already keeps the `.as-bytes`/`Φ.dataized` handle whole. The real failure is upstream in `restore-local-names.xsl`: its `eo:const-handle` predicate decides whether a const handle is multi-referenced (and so must keep its readable `@local` marker). It counted references with an **exact** name match (`eo:resolved-name(@base) = @name`), which matches a bare `ξ.b` but not a dispatch `ξ.b.seg`. So a handle whose sibling use was a dispatch looked single-use, lost its `@local` marker, and `dataized-to-const` → `merge-monikers` then failed to recognize it as a const handle and folded it — stranding the dispatch reference.

This is independent of what the const wraps (a reference, an application, or a datum), since `const-to-dataized` gives every const the same wrapper shape.

## Fix

Count dispatch references too in `eo:const-handle`, mirroring the existing `eo:multi-referenced` (which already uses `starts-with(eo:resolved-name(@base), concat($name, '.'))`). One-line logic change; the const stays multi-referenced, keeps `@local`, is kept whole, and is hosted onto its first reference as a moniker while the dispatch reference reads back as the bare `b.gte`.

## Tests

Two new print-packs, both passing and round-tripping:

- `const-dispatch-referenced-handle.yaml` — const over an application (`x.plus 1 >> c!`) with one bare and one dispatch reference.
- `const-dispatch-referenced-datum.yaml` — same shape over a plain datum (`5 >> c!`).

Full `eo-printer` module: `229 tests, 0 failures`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LkNMRasvAhWJxQ6vwW52Cc)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
@@ -115,12 +115,17 @@
   the shared name. Its "@local" marker is therefore kept here so the surviving
   binding still prints its readable `&gt;&gt; b` handle; the binding itself
   stays cactus-named for "merge-monikers" to fold onto its first reference
-  (#5828).
+  (#5828). A reference reaches the handle either bare (`ξ.b`) or through a
+  method dispatch (`ξ.b.seg`, e.g. `b.gte 1`, #5883), so both spellings are
+  counted — exactly as "eo:multi-referenced" does; counting only the bare shape
+  would leave a handle whose sibling reference is a dispatch looking
+  single-use, dropping "@local" and stranding that dispatch on a synthetic
+  "vL_P" placeholder.
   -->
   <xsl:function name="eo:const-handle" as="xs:boolean">
     <xsl:param name="wrapper" as="element()*"/>
     <xsl:variable name="value" select="$wrapper/o[@base='Φ.dataized']/o[1]"/>
-    <xsl:sequence select="exists($wrapper) and $wrapper/@base='.as-bytes' and exists($wrapper/@name) and exists($value/@local) and count($wrapper/..//o[contains(@base, concat('.', $auto)) and eo:resolved-name(@base) = $wrapper/@name and not(ancestor-or-self::o[. is $wrapper])]) &gt; 1"/>
+    <xsl:sequence select="exists($wrapper) and $wrapper/@base='.as-bytes' and exists($wrapper/@name) and exists($value/@local) and count($wrapper/..//o[contains(@base, concat('.', $auto)) and (eo:resolved-name(@base) = $wrapper/@name or starts-with(eo:resolved-name(@base), concat($wrapper/@name, '.'))) and not(ancestor-or-self::o[. is $wrapper])]) &gt; 1"/>
   </xsl:function>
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2691](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5886, opened 2026-07-25
Print folded φ decoratee ahead of auto-named formation head (#5882)

### LINKED ISSUE #5882: Printer incorrectly inlines handle into anonymous formation causing parser rejection
The canonical printer turns a valid program into one the parser rejects. This EO parses fine:

```
[] > foo
  * > @
    flag > [] >>
    22
  false >> flag
```

`Xmir.toEO` prints it as:

```
[] > foo
  * > @
    [] >>
      false
    22
```

and reparsing that fails with `[4:6] error: 'object inside formation must have a name'`, so `XmirTest.printsToParseableEo` (eo-printer/src/test/java/org/eolang/printer/XmirTest.java:53) goes red on it. In `eo-runtime` this bites `switch.eo` (its three `c1 > [] >>` cases) and `bool.eo:186` as soon as the decoratee's binding is written as a file-local `>>` handle, which is what #5678 asks for.

The fold itself happens in eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:136-144: the `otherwise` branch inlines the single-referenced non-abstract handle `false >> flag` into its only reference. That reference is the φ of the auto-named formation `[] >>`, so the datum ends up as that formation's decoratee, and `to-eo-tree.xsl:400` then prints the formation head as a bare `[] >>` with the decoratee as an ordinary body line below it. EO has no such form — a decoratee of an anonymous formation must stand ahead of the brackets (`false > [] >>`), exactly as the input spells it.

The smallest fix is to print the folded decoratee ahead of the head, i.e. emit `false > [] >>` when an auto-named formation's φ child carries no name of its own; alternatively, leave the handle in place when its sole reference is the φ of an auto-named formation, next to the adjacency guard at inline-cactoos.xsl:93.

yegor256:
Fixes #5882.

## Problem

The canonical printer turned a valid program into one the parser rejects. This EO parses fine:

```
[] > foo
  * > @
    flag > [] >>
    22
  false >> flag
```

but `Xmir.toEO` printed it as:

```
[] > foo
  * > @
    [] >>
      false
    22
```

and reparsing that fails with `[4:6] error: 'object inside formation must have a name'`.

## Root cause

`inline-cactoos.xsl` folds the single-referenced non-abstract handle `false >> flag` into its only reference. That reference is the `φ` decoratee of the auto-named formation `[] >>`, written ahead of the brackets. The `otherwise` (bare-reference) branch copies only the *target's* attributes, so the reference's own `@name = φ` was dropped. `to-eo-tree.xsl` then had no decoratee to place ahead of the head and printed the datum `false` as an ordinary body line of the formation — a shape EO has no spelling for.

## Fix

Carry the `φ` name over during inlining (for a non-abstract target) so the folded datum stays the formation's decoratee. `to-eo-tree`'s existing inline-φ machinery then prints it ahead of the head as `false > [] >>`, exactly as the input spells it, and the output reparses cleanly. An abstract target keeps its own `@name` (the `[...] >>` marker) and never reaches this branch as a φ decoratee, so the guard is scoped to the non-abstract case.

## Tests

- New print-pack `inline-cactoos-phi-decoratee-handle.yaml` reproduces the issue's program and asserts the parseable `false > [] >>` output; it is exercised by both `XmirTest.printsToEo` and `XmirTest.printsToParseableEo`.
- Full `eo-printer` suite passes (236 tests, 0 failures).

---
_Generated by [Claude Code](https://claude.ai/code/session_0193oYEivoc13c7a1szW46Cd)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -138,6 +138,23 @@
               <xsl:if test="@as">
                 <xsl:apply-templates select="@as"/>
               </xsl:if>
+              <!--
+              The reference stands as the `φ` decoratee of an auto-named
+              formation, written ahead of the brackets (`false &gt; [] &gt;&gt;`,
+              #5882). Folding a non-abstract handle in (`false &gt;&gt; flag`)
+              copies only the target's attributes, so the reference's own `@name`
+              would be lost and `to-eo-tree` would print the datum as a nameless
+              body line of the formation (`[] &gt;&gt;` with `false` below it),
+              which the parser rejects with "object inside formation must have a
+              name". Carrying the `φ` name over keeps the folded datum the
+              decoratee, so `to-eo-tree` prints it ahead of the head. Only for a
+              non-abstract target: an abstract one keeps its own `@name` (the
+              `[...] &gt;&gt;` marker) and never reaches this branch as a φ
+              decoratee.
+              -->
+              <xsl:if test="@name = $eo:phi and not($keep-name)">
+                <xsl:apply-templates select="@name"/>
+              </xsl:if>
               <xsl:apply-templates select="$target/@*[$keep-name or (name() != 'name' and name() != 'local')]"/>
               <xsl:apply-templates select="$target/node()"/>
             </xsl:element>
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2703](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5885, opened 2026-07-25
Revert "Charge a line-leading `(` four times in eo-printer `Penalty` BRACKET term" (#5881)

yegor256:
Reverts the merge of #5881 (merge commit `a15d4ac`), which was merged prematurely.

This restores `master` to its state before #5881: the `eo-printer` `Penalty`/`PenaltyKey` change (line-leading `(` factor, `PenaltyKey.LEADING`) and the 60 reformatted `eo-runtime` sources are undone (68 files, 317+/851−).

No work is lost — the reverted change still lives on branch `claude/eo-issue-5880-9pgo5r` (commit `d549501`) and can be re-applied later via a fresh PR when it's the right time to merge #5880.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MpAfSHDLaaXL9mJuqNB4zv)_

PR #5888, opened 2026-07-25
Charge a line-leading `(` four times in eo-printer `Penalty` BRACKET term

### LINKED ISSUE #5880: Charge a line-leading `(` four times in eo-printer `Penalty` BRACKET term
`Penalty.brackets()` (`eo-printer/src/main/java/org/eolang/printer/Penalty.java:259`) weighs every opening parenthesis by nesting depth alone: a `(` that opens with `depth` brackets already open costs `depth + 1` units of the `BRACKET` weight. Position on the line plays no part, so a `(` standing as the first non-space character of a line — as in `(byte.and p1).eq r1` at `eo-runtime/src/main/eo/string.eo:53` — is charged the same single unit as a bracket sitting mid-line.

The start of a line is the worst place for a parenthesis. The reader meets a group before knowing what it applies to, and the line has to be read backwards to make sense of it. There are 298 such lines in `eo-runtime/src/main/eo` today, and the layout search has no reason to avoid them, because the penalty function does not distinguish that position at all.

In `Penalty.brackets()`, charge a `(` that is the first non-space character of its line four times as much, so it counts as `(depth + 1) * 4` units instead of `depth + 1`. Add a new tunable `PenaltyKey.LEADING` with fallback `4` for the factor, and turn `brackets()` into an instance method so it can read the weight, keeping the factor configurable like every other knob.

yegor256:
Closes #5880.

Re-creates the change from #5881 (which was merged prematurely and then reverted by #5885), rebased onto the current `master`.

## What

`Penalty.brackets()` weighed every opening parenthesis by nesting depth alone: a `(` opening with `depth` brackets already open cost `depth + 1` units of the `BRACKET` weight, regardless of where it sat on the line. A `(` standing as the first non-space character of a line — the worst place for a group, since the reader meets it before knowing what it applies to — was charged the same single unit as one sitting mid-line, so the layout search had no reason to avoid it.

## How

- **`Penalty.brackets()`** now charges a `(` that is the first non-space character of its line `(depth + 1) * LEADING` units instead of `depth + 1`, pushing the layout search away from opening a line with a group.
- **`PenaltyKey.LEADING`** — a new tunable with fallback `4` for the factor, keeping it configurable like every other knob.
- **`brackets()` is now an instance method** (was `static`) so it can read the weight from the config.

## Effect on `eo-runtime`

Reformats the **60 `eo-runtime/src/main/eo` sources** that were laid out with line-leading parentheses under the old weight (the ~298 lines noted in the issue, e.g. `(byte.and p1).eq r1` in `string.eo`). Each such group is now broken onto its own indented lines in prefix form (`eq.` / `byte.and p1` / `r1`). Verified with `MjFormat` (`-Deo.autoFix`): all 153 sources are canonical against the current `master`.

## Tests

- `PenaltyTest.chargesForParentheses` gains rows contrasting a line-leading `(` (`(a).eq b` → 76, i.e. `4 × 19`) with a mid-line one (`f (a).eq b` → 19); the nested `(a (b (c 1)))` expectation is updated to `171`.
- `PenaltyKeyTest.honoursOverriddenLeadingWeight` verifies that tuning `LEADING` to `1` makes a leading `(` cost the same as a mid-line one.
- Four print-packs whose demonstrated layout opened a line with a group (`compound-receiver-suffix`, `dispatch-chains-mixed`, `inline-phi-overflow`, `nested-formation-apply`) are updated to their new vertical rendering. These packs are shared (symlinked) between `eo-printer`'s `XmirTest` and `eo-maven-plugin`'s `MjPrintTest`; the latter has no `LEADING` mojo parameter, so both consumers agree on the default-weight output.

All `eo-printer` tests, `MjPrintTest`, and the `eo-runtime` format check pass locally against the current `master`.

> **Note:** this touches ~1168 lines (mostly the reformatted `eo-runtime` sources), so the `pr-size` gate will fail — the diff is intrinsic to the issue and can't be split without breaking the format check.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MpAfSHDLaaXL9mJuqNB4zv)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1188 >= 100); -8 for way too many hits-of-code (1188 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2715](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5889, opened 2026-07-25
Reformat eo-runtime sources with the eo-printer auto-fix

yegor256:
This runs the eo-printer auto-fix over the whole `eo-runtime` source tree and commits the result. The layout changes are consistent everywhere: file-local handles now fold inline through `>>` and the `|` pipe continuation, void attributes written as `? >> name` become plain formal parameters `[name]`, and private attributes switch from `>` to `>>`. It also drops a small dead scratch object that was sitting unused inside `tuple/front.eo`.

The point is to make the runtime match the printer's canonical output, so future `eo:format` runs are a no-op and the sources stop drifting from what the formatter produces. Getting here needed a run of printer fixes (#5828, #5834, #5837, #5840, #5844, #5848) that plugged data-loss and folding gaps this reformat would otherwise have hit.

No behaviour change — this is layout only. Part of #5678.


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (2355 >= 100); -8 for way too many hits-of-code (2355 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2719](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5891, opened 2026-07-25
Accept bare vertical pipe `|` in the parser (#5887)

### LINKED ISSUE #5887: `eo:format` silently drops the arguments of an application based on a file-local `>>` formation
`eo:format -Deo.autoFix` can change what a program means. Take `eo-runtime/src/main/eo/string/split.eo` with its recursive helper written as a file-local handle, which is what #5678 asks for:

```
[text delimiter] > split
  if. > @
    len.eq 0
    *
    rec-split
      *
      0
      0
  text > bts!
  delimiter > delim!
  bts.size > len!
  [accum start current] >> rec-split
    ...
```

`mvn -pl eo-runtime process-sources -Deo.autoFix` rewrites the third `if.` argument into the formation itself and deletes the application's three arguments:

```
[text delimiter] > split
  if. > @
    len.eq 0
    *
    [accum start current] >> rec-split
      ...
  text > bts!
```

No warning, exit code 0, and the result is a fixed point of the formatter, so the `format` gate stays green on it. The program is no longer the same one: `if` now receives the formation instead of the result of applying it to `* 0 0`, and 79 `eo-runtime` tests fail with `the ⊥ object is a terminated computation and cannot be used` at `string.split:14:4`.

With the weights the printer's own packs use (`WIDTH: 80`) the same origin keeps the arguments but prints the pipe continuation vertically:

```
    |
      *
      0
      0
```

and that output does not parse — `[33:4] error: 'a pipe | must be followed by a space'`, reproducible through `XmirTest.printsToEo` with the object above as `origin`. So one defect has two faces: for a reference standing in the argument slot of another application, the pipe continuation is emitted either in a shape the parser rejects or not at all. Both are heavier than #5732, which loses only the compact spelling while the XMIR stays identical.

The safe form already exists in the source: keep the formation hoisted and print the reference as the plain application `rec-split * 0 0`. Gate the relocate-and-pipe branch at eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:123 on the reference not being an argument of another application, and have `to-eo-tree` fall back to that plain form whenever the compact one-line `| args` cannot be printed — dropping the reference's argument children should never be an option.

yegor256:
Closes #5887.

## Problem

`eo:format` on a file-local `>>` handle applied from the argument slot of another application (the shape `string/split.eo` takes once its recursive helper is written as a `>>` handle, per #5678) produced EO that did not reparse.

For `rec-split * 0 0` standing as the third argument of `if.`, `inline-cactoos` relocates the single-use formation directly above the reference and rewrites the reference into a `| args` pipe continuation (the same mechanism as the sibling #5840 and receiver #5844 cases), so the three arguments are preserved. When the relocated formation is large, the layout engine cannot keep the compact one-line `| * 0 0` and prints the pipe in its **vertical** form — a bare `|` followed by a deeper-indent argument block:

```
      |
        *
        0
        0
```

That output was rejected on reparse with `a pipe | must be followed by a space`, so it was not a fixed point and the `format` gate could not stabilise on it.

As [noted on the issue](https://github.com/objectionary/eo/issues/5887#issuecomment-5077058829), a pipe **may** have no space after it — the vertical-argument form (R-3.14.3) is legal EO.

## Fix

`LnPipe` already models both pipe shapes (an empty argument list opens the node for a vertical block, exactly like a `vapplication` head), but its `piped()` tokenizer rejected any line shorter than `| `, so a bare `|` never reached that path. The guard now only rejects a character glued directly onto the pipe (`|x`); a bare `|` carrying no horizontal args is accepted and opens for its vertical argument block. The arguments of the applied reference are preserved and the printed output round-trips.

## Changes

- `eo-parser/.../LnPipe.java` — accept a bare `|` (the vertical-argument form); only reject a non-space character glued directly onto the pipe.
- `eo-parser/.../eo-syntax/pipe-vertical-bare.yaml` — parser test: a bare `|` with a deeper-indent body parses to a pipe application carrying its two arguments.
- `eo-printer/.../print-packs/yaml/pipe-vertical-in-argument.yaml` — printer regression test: a `>>` handle applied from an argument slot prints to a vertical pipe that reparses to itself with its arguments intact.

## Testing

- `eo-parser` full suite: 1505 tests, 0 failures.
- `eo-printer` `XmirTest`: 201 tests, 0 failures (both `printsToEo` fixed-point and `printsToParseableEo` round-trip pass for the new pack).

---
_Generated by [Claude Code](https://claude.ai/code/session_01QtWXbN1dHStAi6dpdPWZQS)_

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnPipe.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnPipe.java
@@ -120,17 +120,21 @@ final class LnPipe implements Line {
     }
 
     /**
-     * Build a token stream positioned just past the leading {@code | }
-     * marker. The {@code |} must be followed by a single space (before
-     * the argument list or suffix).
-     * @return Tokens positioned after {@code "| "}
+     * Build a token stream positioned just past the leading {@code |}
+     * marker. Two shapes are legal (R-3.14.3): the horizontal form
+     * {@code | a b}, whose {@code |} is followed by a space before the
+     * argument list or suffix, and the bare vertical form {@code |}
+     * (nothing after the pipe), which opens for a deeper-indent
+     * argument block just like a {@code vapplication} head. Only a
+     * character glued directly onto the pipe ({@code |x}) is rejected.
+     * @return Tokens positioned after the {@code |}
      */
     private Tokens piped() {
         final String body = this.span.body();
-        if (body.length() < 2 || body.charAt(1) != ' ') {
+        if (body.length() > 1 && body.charAt(1) != ' ') {
             throw new ParseError(
                 this.span.line(), this.span.indent(),
-                "a pipe `|` must be followed by a space"
+                "a pipe `|` must be followed by a space before its arguments"
             );
         }
         final Tokens tokens = new Tokens(body, this.span);
--- a/eo-parser/src/main/java/org/eolang/parser/LnPipe.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnPipe.java
@@ -131,7 +131,7 @@ final class LnPipe implements Line {
      */
     private Tokens piped() {
         final String body = this.span.body();
-        if (body.length() > 1 && body.charAt(1) != ' ') {
+        if (!"|".equals(body) && !body.startsWith("| ")) {
             throw new ParseError(
                 this.span.line(), this.span.indent(),
                 "a pipe `|` must be followed by a space before its arguments"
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2703](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5892, opened 2026-07-25
Fix const-handle receiver reference stranded on synthetic id (#5890)

### LINKED ISSUE #5890: Const `>>` handle is hosted at a later reference, stranding the earlier one on `vL_P`
After #5883 a multi-referenced const handle is kept whole, but it can be hosted at the wrong use site. This origin, which is `eo-runtime/src/main/eo/os.eo` with its `oname` handle made private for #5678:

```
[] > os
  [] > is-windows
    and. > @
      oname.size.gt 6
      eq.
        oname.slice 0 7
        "Windows"
    name >> oname!
  [] > name /Q.string
```

prints as:

```
[] > os
  [] > is-windows
    (v8_4.size.gt 6).and > @
      eq.
        slice.
          name >> oname!
          0
          7
        "Windows"
  [] > name /Q.string
```

The handle survives, but it is hosted onto the *second* reference, inside the `slice.` block, while the first reference — the one that becomes the receiver of the reversed `.and` — is left as `v8_4`, a name that appears nowhere in the output. So the printed source neither denotes the same object nor parses back to it, and `eo:format` fails the file instead of converging on it.

The same shape works when the earlier reference is a plain argument rather than a reversed-dispatch receiver: `d.gt 6` / `d.gt 7` over `bar >> d!` hosts correctly at the first use. The receiver position is what the hosting pass skips, which makes this the const-handle counterpart of #5844 — there the receiver lost its argument, here it loses its binding to the handle.

Hosting should pick the reference that comes first in print order, or, when that reference sits in a receiver position it cannot host, leave the handle in its own binding and print every reference by name — the spelling the source already uses. Either way no reference should be left holding a `vL_P` id. `eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl` is where the host is chosen.

yegor256:
Closes #5890.

## Problem

After #5883 a multi-referenced const file-local handle (`a >> b!`) is kept whole and folds onto its first *hostable* reference, with every other reference rewritten back to the readable `@local` handle. But `eo:kept-const-ref` recognised only two reference shapes (via `eo:resolved-ref`): a bare `ξ.<name>` and a single-segment dispatch `ξ.<name>.<seg>`.

When the earliest reference sits in a receiver position the moniker fold cannot host — a *multi-segment* chain such as `oname.size.gt 6` used as the receiver of a reversed `.and` dispatch (base `ξ.<name>.size.gt`) — it matched neither the hosting path nor `eo:kept-const-ref`. It was therefore left carrying the obfuscated `vL_P` cactus id:

```
(v8_4.size.gt 6).and > @
```

`v8_4` appears nowhere else, so the printed source neither denotes the same object nor parses back to it, and `eo:format` fails the file instead of converging on it. This is the const-handle counterpart of #5844: the receiver position is what the hosting pass skipped.

## Fix

`eo-printer/.../print/merge-monikers.xsl`: match the kept reference by carrying the binding's cactus `@name` as one of its base segments — the same segment-wise test the abstract-handle sibling `eo:kept-local-ref` already uses for `[] >> name` (#5876). Bare, single-segment, and multi-segment references now all read back through the readable `@local` handle, so the stranded receiver prints `oname.size.gt` and no reference is left holding a `vL_P` id. The binding's own subtree and the single hosting reference are excluded, exactly as before.

## Tests

Two new print-packs under `eo-printer/.../print-packs/yaml`, both verified to reprint to themselves (round-trip):

- `const-receiver-referenced-handle.yaml` — the issue's exact reversed-dispatch receiver scenario.
- `const-multi-segment-referenced-handle.yaml` — a plain multi-segment dispatch reference alongside a bare argument.

Full `eo-printer` suite is green (239 tests).

---
_Generated by [Claude Code](https://claude.ai/code/session_01GUbNVfTeNWmua9JuEQAR9B)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -169,14 +169,21 @@
   The const file-local handle binding a reference `$ref` resolves to but does
   NOT host (the binding folds onto its first reference only). Every other
   reference keeps the readable "@local" handle in place of the obfuscated
-  cactus name, so it reads back as a bare `b` rather than a synthetic "vL_P"
-  name (#5828).
+  cactus name, so it reads back as a bare `b` (or a dispatch `b.seg`) rather
+  than a synthetic "vL_P" name (#5828). The reference is matched by carrying the
+  binding's cactus "@name" as one of its base segments — a bare `ξ.<name>`, a
+  single-segment dispatch `ξ.<name>.<seg>`, or a multi-segment chain
+  `ξ.<name>.<seg>.<seg>` whose receiver position the moniker fold cannot host
+  (#5890) — so a stranded receiver reference reads back by name instead of a
+  synthetic id, the const mirror of `eo:kept-local-ref`. The binding's own
+  subtree and the single hosting reference are excluded.
   -->
   <xsl:function name="eo:kept-const-ref" as="element()*">
     <xsl:param name="ref" as="element()"/>
     <xsl:variable name="owner" select="$ref/ancestor::o[eo:abstract(.)][1]"/>
-    <xsl:variable name="binding" select="$owner/o[@name = eo:resolved-ref($ref) and eo:moniker-binding(.) and eo:const-handle(.)][1]"/>
-    <xsl:sequence select="if (exists($binding) and not(eo:moniker-refs($binding)[1] is $ref)) then $binding else ()"/>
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2711](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5894, opened 2026-07-25
#5867: fix string.at docstring for negative index handling

### LINKED ISSUE #5867: `string.at` docstring inaccurately reflects negative index handling
## What happens

`string.at` (`eo-runtime/src/main/eo/string/at.eo`) is documented as:

```
# Retrieve symbol by given index as `text`.
# If 0 > index >= ^.length, the `fallback` value is returned,
# or the bottom object when unbound.
```

This reads as "a negative index (or an index beyond the length) returns the fallback." That is not what the code does. `at`'s first local, `index!`, converts a negative input index into a real position by wrapping it from the end (`len + idx` when `idx` is negative), before the range check ever runs:

```
i > idx!
if. > index!
  0.gt idx
  plus.
    number len
    idx
  idx
text.length > len!
```

Only after that wrapping does the actual bounds check happen (`0.gt index` or `index >= len`). So a negative index that is in range after wrapping does NOT return the fallback; it returns the character at the wrapped position. This is confirmed by the file's own passing test:

```
eq. ++> tests-returns-char-at-negative-index
  "m"
  at
    "some"
    -2
```

`at("some", -2)` returns `"m"` (the character 2 from the end), not the fallback. The fallback is only returned when the *wrapped* index is still negative (index too far before the start) or `>= len` (index too far past the end).

## What should happen

The docstring should describe the real contract: a negative `index` counts from the end of `text` (`len + index`), and the fallback is returned only when that resolved position falls outside `[0, len)`. Same docstring-vs-behavior mismatch class already accepted as issue #5861 (`tuple.back`).

morphqdd:
The docstring of `string.at` says the fallback is returned "If 0 > index >= ^.length", which does not describe what the code actually does.

The code (`at.eo`) accepts negative indexes and wraps them from the end of the text: a negative `index` is converted to `length + index` before the bounds check. The old docstring did not mention this wrap-around behavior at all, so a reader could not tell from the comment alone that `at("some", -2)` returns `"m"`.

This PR only updates the docstring to describe the actual negative-index behavior:

```
# Retrieve symbol by given index as `text`.
# A negative `index` counts from the end of `text` (`length + index`).
# If the resolved index is still negative or is >= `text.length`,
# the `fallback` value is returned, or the bottom object when unbound.
```

No functional change. The existing embedded tests already cover this case (`tests-returns-char-at-negative-index`) and continue to pass.

Closes #5867

### DIFF
--- a/eo-runtime/src/main/eo/string/at.eo
+++ b/eo-runtime/src/main/eo/string/at.eo
@@ -1,6 +1,7 @@
 # Retrieve symbol by given index as `text`.
-# If 0 > index >= ^.length, the `fallback` value is returned,
-# or the bottom object when unbound.
+# A negative `index` counts from the end of `text` (`length + index`).
+# If the resolved index is still negative or is >= `text.length`,
+# the `fallback` value is returned, or the bottom object when unbound.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (5) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1956](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5895, opened 2026-07-25
#5860: close the Files.walk() stream in PackageInfos#create

### LINKED ISSUE #5860: `PackageInfos#create` leaks an unclosed `Files.walk()` stream
## What happens

`PackageInfos#create` (`eo-maven-plugin/src/main/java/org/eolang/maven/PackageInfos.java`) opens a `Files.walk()` stream and never closes it:

```java
int create() throws IOException {
    final int size;
    if (Files.exists(this.root)) {
        final List<Path> dirs = Files.walk(this.root).filter(
            file -> Files.isDirectory(file)
                && !file.equals(this.root)
                && !file.equals(this.root.resolve("org"))
            )
        for (final Path dir : dirs) {
            ...
```

`Files.walk()` opens an underlying directory-traversal resource (an open file descriptor on most platforms) that the JDK's own javadoc says must be closed, normally via try-with-resources. Here the resulting `Stream<Path>` is filtered and collected into a `List`, but never closed.

Every other `Files.walk()` call site in the same package (`eo-maven-plugin/src/main/java/org/eolang/maven/`) correctly wraps it in try-with-resources:

- `Walk.java:113` - `try (Stream<Path> walk = Files.walk(dir))`
- `Cache.java:145` - `try (Stream<Path> stream = Files.walk(dir))`
- `Resolving.java:192` and `:240` - both wrapped
- `Sha.java:56` - wrapped

`PackageInfos.java` is the only one of the 6 `Files.walk()` call sites in this package that does not close its stream. `create()` runs once per transpile invocation, walking the whole generated-sources tree, so the leaked handle scales with the number of source packages in the project.

## What should happen

`PackageInfos#create` should wrap its `Files.walk(this.root)` call in try-with-resources, matching the pattern already used at every other `Files.walk()` call site in this same package:

```java
final List<Path> dirs;
try (Stream<Path> walk = Files.walk(this.root)) {
    dirs = walk.filter(
        file -> Files.isDirectory(file)
            && !file.equals(this.root)
            && !file.equals(this.root.resolve("org"))
        )
}
```

morphqdd:
`PackageInfos#create` opens a `Files.walk()` stream and never closed it. `Files.walk()` opens an underlying directory-traversal resource that the JDK javadoc says must be closed, normally via try-with-resources.

Every other `Files.walk()` call site in the same package (`Walk.java`, `Cache.java`, `Resolving.java`, `Sha.java`) already wraps it in try-with-resources. `PackageInfos.java` was the only one that did not.

This PR wraps the call in try-with-resources, matching the existing pattern in the package. No behavior change: the stream is still filtered and collected into a `List` before the `for` loop runs, exactly as before, just inside the try-with-resources block so the underlying resource is released once collection finishes.

Existing tests in `PackageInfosTest.java` cover `create()`'s behavior fully (return count, file creation, root exclusion, escaping) and all continue to pass unchanged, since this is a pure resource-management fix with no observable output difference.

Closes #5860

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/PackageInfos.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/PackageInfos.java
@@ -12,6 +12,7 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Package info classes.
@@ -60,12 +61,15 @@ final class PackageInfos {
     int create() throws IOException {
         final int size;
         if (Files.exists(this.root)) {
-            final List<Path> dirs = Files.walk(this.root).filter(
-                file -> Files.isDirectory(file)
-                    && !file.equals(this.root)
-                    && !file.equals(this.root.resolve("org"))
-                )
-                .collect(Collectors.toList());
+            final List<Path> dirs;
+            try (Stream<Path> walk = Files.walk(this.root)) {
+                dirs = walk.filter(
+                    file -> Files.isDirectory(file)
+                        && !file.equals(this.root)
+                        && !file.equals(this.root.resolve("org"))
+                    )
+                    .collect(Collectors.toList());
+            }
             for (final Path dir : dirs) {
                 Logger.debug(
                     this,


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1968](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5896, opened 2026-07-25
Read a nested reference to a const handle back by name

### LINKED ISSUE #5893: eo-printer strands a nested-formation reference to a private const `>>` handle as `^.vL_C`
When a private const file-local handle (`x >> name!`, R-3.10.12) is referenced both from its own scope and from inside a **nested formation**, `eo:format` hosts the handle onto the outer reference and strands the nested reference on a synthetic `^.vLINE_COL` name that resolves to nothing on reparse. This is the sibling of #5883 (bare reference) and #5890 (method-dispatch / multi-segment reference); both of those fixed same-scope references, but the nested-formation-scope case remains broken.

### Reproduction (print-pack shape)

Origin:

```
[list] > foo
  idx.eq 0 > a
  (bar 5).eq 1 > c
  [tup] >> bar
    idx.eq tup > @
  list >> idx!
```

`eo-printer` currently emits (non-idempotent, unparseable):

```
[list] > foo
  eq. > a
    list >> idx!
    0
  eq. > c
    ^.v6_2.eq tup > [tup] >> bar
    | 5
    1
```

The handle `list >> idx!` is hosted onto the first reference (`idx.eq 0`), and the reference living inside the nested `[tup] >> bar` formation becomes `^.v6_2` — a synthetic parent-navigation name pointing at origin line 6 (the handle) that resolves to nothing on reparse.

### In-repo reproduction

Privatize the cache at `eo-runtime/src/main/eo/tuple/front.eo:30` from `index > idx!` to `index >> idx!` and run `mvn -pl eo-runtime compile -Deo.autoFix`. The reference inside the `[tup] >> rechead` nested formation is rewritten to `tup.length.eq ^.v30_2`. The same strand appears when privatizing const caches in `string/split.eo`, `string/index-of.eo`, `string/last-index-of.eo`, `string/contains.eo`, `string/repeated.eo`, `string/up-cased.eo`, `tuple.eo`, `tuple/slice.eo`, and `malloc.eo` — ten files whose const is referenced from a nested `>>` formation.

### Cause and suggested fix

The const-handle relocation in `merge-monikers` (extended for #5890 in `5ebb2f1870`) and `restore-local-names` hosts the handle onto an outer reference without accounting for references that live inside a nested formation; such a reference cannot navigate up to the relocated handle. When a const `>>` handle has any reference inside a nested formation, keep it as a standalone in-place handle rather than hosting or folding it — mirroring the way #5890 keeps a dispatch-referenced const whole.

yegor256:
When a private const handle (`x >> name!`) is referenced from inside a nested
formation, the printer used to leave that reference on its obfuscated cactus
name, which came out as `^.vL_C` — a name that binds to nothing when the file
is parsed again. The lookup that #5828 and #5890 added only searched the
reference's own scope, and a nested reference reaches the handle through a `ρ`
climb, so it never matched.

Now the binding is looked up in every enclosing formation, and the rewrite
drops the climb together with the cactus segment: a handle is reached by its
bare name and `resolve-local-names` re-derives the climb on the way back in.
The two rewrite templates (const and abstract handle) shared a copy of the
same expression, so it moved into one `eo:handle-base` function.

I checked this against the ten `eo-runtime` files the issue lists — privatizing
their const caches and running `eo:format` no longer strands anything, and the
formatting is idempotent. Separately, while verifying I noticed that hosting a
const handle onto a dispatch that is itself a const's value drops the outer `!`
(`c.gte 2 > a!` prints as `gte. > a`); that happens on master too and I left it
alone here.

Closes #5893

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -175,15 +175,33 @@
   single-segment dispatch `ξ.<name>.<seg>`, or a multi-segment chain
   `ξ.<name>.<seg>.<seg>` whose receiver position the moniker fold cannot host
   (#5890) — so a stranded receiver reference reads back by name instead of a
-  synthetic id, the const mirror of `eo:kept-local-ref`. The binding's own
-  subtree and the single hosting reference are excluded.
+  synthetic id, the const mirror of `eo:kept-local-ref`. The reference need not
+  live in the binding's own scope: one written inside a nested formation climbs
+  to the handle through `ρ` and is stranded just the same (#5893), so the binding
+  is looked up in the nearest enclosing formation that declares it. The binding's
+  own subtree and the single hosting reference are excluded.
   -->
   <xsl:function name="eo:kept-const-ref" as="element()*">
     <xsl:param name="ref" as="element()"/>
-    <xsl:variable name="owner" select="$ref/ancestor::o[eo:abstract(.)][1]"/>
-    <xsl:variable name="candidates" select="$owner/o[eo:moniker-binding(.) and eo:const-handle(.)]"/>
-    <xsl:variable name="binding" select="$candidates[some $seg in tokenize($ref/@base, '\.') satisfies $seg = @name][1]"/>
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2719](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5897, opened 2026-07-25
#5861: fix tuple.back docstring to describe last-N-elements contract

### LINKED ISSUE #5861: `tuple.back` docstring says "from index to the end" but it actually returns the last N elements
## What happens

`tuple.back` (`eo-runtime/src/main/eo/tuple/back.eo`) is documented as:

```
# Elements of `list` from `index` to the end (supports oversized index).
```

This reads as "`index` is a position in the list, return everything from that position onward." That is not what the object actually does. Its real, tested contract is "the last `index` elements of `list`":

```
eq. ++> tests-simple-back
  back
    *
      1
      2
      3
      4
      5
    2
  *
    4
    5
```

`back([1,2,3,4,5], 2)` returns `[4,5]`, the last 2 elements. If `index` were a position, "from position 2 to the end" of a 5-element list would be `[3,4,5]`, not `[4,5]`.

The other test in the same file confirms this further:

```
eq. ++> tests-large-index-in-back
  back
    *
      1
      2
      3
      4
      5
    10
  *
    1
    2
    3
    4
    5
```

`index=10` (larger than the list length) returns the whole list. Under "position" semantics, an out-of-range start position should yield an empty result. This only makes sense under "last N elements, clamped to the list length."

The object's own sibling, `tuple/withi.eo`, confirms the real contract by how it calls `back`:

```
back
  list
  list.length.minus index
```

`withi` converts a real position (`index`) into a trailing count (`list.length.minus index`) before calling `back`, because `back` itself expects a count, not a position. `tuple/front.eo` (the true position-based sibling, "elements before index") is documented correctly by contrast, and does not need this conversion.

## What should happen

`tuple/back.eo`'s docstring should describe the real contract, for example: "the last `index` elements of `list` (supports an index larger than the list length, which returns the whole list)."

morphqdd:
`tuple.back`'s docstring said "Elements of `list` from `index` to the end (supports oversized index)." That describes a position-based slice, but the object's real, tested contract is "the last `index` elements of `list`":

`back([1,2,3,4,5], 2)` returns `[4,5]`, the last 2 elements. Under "from position `index` to the end", that would need to return `[3,4,5]`, not `[4,5]`.

`back([1,2,3,4,5], 10)` (index larger than length) returns the whole list. Under "position" semantics an out-of-range start position should give an empty result, which only makes sense under "last N elements, clamped to the list length".

`tuple/withi.eo` confirms this: it converts a real position into a trailing count before calling `back` (`list.length.minus index`), because `back` itself expects a count, not a position.

This PR only updates the docstring:

```
# The last `index` elements of `list` (an `index` larger than `list.length`
# returns the whole `list`).
```

No functional change. The existing embedded tests already cover this contract and continue to pass.

Closes #5861

### DIFF
--- a/eo-runtime/src/main/eo/tuple/back.eo
+++ b/eo-runtime/src/main/eo/tuple/back.eo
@@ -1,4 +1,5 @@
-# Elements of `list` from `index` to the end (supports oversized index).
+# The last `index` elements of `list` (an `index` larger than `list.length`
+# returns the whole `list`).
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (3) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2064](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5898, opened 2026-07-25
#5862: reuse Hex in StHex instead of duplicating it in BytesToHex

### LINKED ISSUE #5862: `BytesToHex` duplicates `Hex`'s hex-encoding logic instead of reusing it
## What happens

`BytesToHex.java` and `Hex.java` (both in `eo-parser/src/main/java/org/eolang/parser/`) independently implement the identical "bytes to canonical hex string" algorithm: empty array becomes `"--"`, a single byte becomes `"BB-"` with a trailing dash, and multiple bytes become `"BB-BB-..."` with no trailing dash.

`Hex.asString()`:
```java
String asString() {
    final String result;
    if (this.bytes.length == 0) {
        result = "--";
    } else if (this.bytes.length == 1) {
        result = String.format("%02X-", this.bytes[0]);
    } else {
        final StringBuilder out = new StringBuilder(this.bytes.length * 3);
        for (int idx = 0; idx < this.bytes.length; idx = idx + 1) {
            if (idx > 0) {
                out.append('-');
            }
            out.append(String.format("%02X", this.bytes[idx]));
        }
        result = out.toString();
    }
    return result;
}
```

`BytesToHex.get()`:
```java
public String get() {
    final String hex;
    if (this.bytes.length == 0) {
        hex = "--";
    } else {
        final StringJoiner out = new StringJoiner("-");
        for (final byte bty : this.bytes) {
            out.add(String.format("%02X", bty));
        }
        if (this.bytes.length == 1) {
            out.add("");
        }
        hex = out.toString();
    }
    return hex;
}
```

Both produce byte-for-byte identical output for every input (traced by hand across empty, 1-byte, and multi-byte cases).

`BytesToHex` has exactly one caller, `StHex.java:31`, which also manually re-derives the IEEE-754 double bytes with `ByteBuffer.allocate(Double.BYTES).putDouble(...).array()` - exactly what `Hex`'s own `Hex(double)` constructor already does:

```java
Hex(final double value) {
    this(ByteBuffer.allocate(Double.BYTES).putDouble(value).array());
}
```

`Hex` is already used for this exact purpose elsewhere in the same package (`Emissions.java`, `LnTextBlock.java`).

## What should happen

`StHex` should reuse `Hex` instead of maintaining a second, independently-written implementation of the same encoding rule:

```java
new Hex(Double.parseDouble(xnav.text().orElseThrow())).asString()
```

`BytesToHex.java` should be removed once its single caller is switched over.

morphqdd:
`BytesToHex.java` and `Hex.java` (both in `eo-parser/src/main/java/org/eolang/parser/`) independently implemented the identical "bytes to canonical hex string" algorithm: empty array becomes `--`, a single byte becomes `BB-` with a trailing dash, multiple bytes become `BB-BB-...` with no trailing dash. Both produce byte-for-byte identical output for every input.

`BytesToHex` had exactly one caller, `StHex.java`, which also manually re-derived the IEEE-754 double bytes with `ByteBuffer.allocate(Double.BYTES).putDouble(...).array()`, exactly what `Hex`'s own `Hex(double)` constructor already does. `Hex` is already used for this same purpose elsewhere in the package (`Emissions.java`, `LnTextBlock.java`).

This PR switches `StHex` to use `Hex` directly:

```java
new Hex(Double.parseDouble(xnav.text().orElseThrow())).asString()
```

and removes `BytesToHex.java` now that its single caller is gone.

`StHexTest.java` and `HexTest.java` already cover the encoding paths exercised through `StHex` and continue to pass unchanged, since the encoding logic itself (now `Hex.asString()`) is unchanged.

Closes #5862


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1984](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5899, opened 2026-07-25
#5863: anchor MjPrint's .xmir to .eo extension replace to the trailing extension

### LINKED ISSUE #5863: `MjPrint#print` incorrectly replaces all `.xmir` substrings instead of just trailing extension
## What happens

`MjPrint#print` (`eo-maven-plugin/src/main/java/org/eolang/maven/MjPrint.java`) computes the output path with a plain substring replace:

```java
private int print(final Path source) throws Exception {
    final Path home = this.printOutputDir.toPath();
    final Path relative = Paths.get(
        this.printSourcesDir.toPath().relativize(source).toString()
            .replace(".xmir", ".eo")
    );
    ...
```

`String#replace(".xmir", ".eo")` replaces every occurrence of the literal substring `.xmir` anywhere in the full relativized path string, not just the file's trailing extension. If any directory component between `printSourcesDir` and the file also contains the substring `.xmir` (for example a directory literally named `v1.xmir-legacy`), that directory name gets silently mangled too, turning it into `v1.eo-legacy` and producing a wrong output path.

The same repository already has the correct, anchored idiom for this exact class of problem in `Unplace.java`:

```java
private static final Pattern EO = Pattern.compile(".eo$");
...
String make(final Path file) {
    return Unplace.EO.matcher(
        file.toString().substring(this.parent.toString().length() + 1)
    ).replaceAll("").replace(File.separator, ".");
}
```

`Unplace` anchors the pattern to the end of the string (`$`), so only the trailing extension is affected regardless of what appears earlier in the path. `MjPrint` does not use this pattern.

Real-world impact is low since it needs an unusual directory name to trigger, but the project's own test (`MjPrintTest.java`) reproduces the same naive-replace logic to compute its expected filename, so CI would not catch the edge case either.

## What should happen

`MjPrint#print` should strip only the trailing `.xmir` extension, anchored to the end of the filename, matching the pattern already established in `Unplace.java` for the equivalent `.eo` case, instead of a plain unanchored `String#replace`.

morphqdd:
`MjPrint#print` computed the output path with a plain substring replace:

```java
this.printSourcesDir.toPath().relativize(source).toString().replace(".xmir", ".eo")
```

`String#replace(".xmir", ".eo")` replaces every occurrence of the literal substring `.xmir` anywhere in the full relativized path string, not just the file's trailing extension. If any directory component between `printSourcesDir` and the file also contains the substring `.xmir` (for example a directory literally named `v1.xmir-legacy`), that directory name gets silently mangled too, turning it into `v1.eo-legacy` and producing a wrong output path.

`Unplace.java` already has the correct, anchored idiom for this exact class of problem, using a `Pattern` anchored to the end of the string (`$`) so only the trailing extension is affected.

This PR replaces the naive `String#replace` with a `Pattern.compile("\\.xmir$")` anchored to the end of the path string, matching that established pattern.

Added a regression test, `doesNotMangleADirectoryNameContainingTheXmirSubstring`, which prints a file under a directory named `v1.xmir-legacy` and asserts the directory name survives untouched. Verified this test fails against the old naive-replace code (the directory got mangled to `v1.eo-legacy`, so the expected output path did not exist) and passes with the fix.

Closes #5863

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjPrint.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjPrint.java
@@ -12,6 +12,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.EnumMap;
 import java.util.Map;
+import java.util.regex.Pattern;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
@@ -36,6 +37,11 @@ import org.eolang.printer.Xmir;
     threadSafe = true
 )
 public final class MjPrint extends MjSafe {
+
+    /**
+     * Pattern to catch the trailing .xmir extension.
+     */
+    private static final Pattern XMIR = Pattern.compile("\\.xmir$");
 
     /**
      * Directory with XMIR sources to print.
@@ -116,8 +122,9 @@ public final class MjPrint extends MjSafe {
     private int print(final Path source) throws Exception {
         final Path home = this.printOutputDir.toPath();
         final Path relative = Paths.get(
-            this.printSourcesDir.toPath().relativize(source).toString()
-                .replace(".xmir", ".eo")
+            MjPrint.XMIR.matcher(
+                this.printSourcesDir.toPath().relativize(source).toString()
+            ).replaceFirst(".eo")
         );
         new Saved(
             new Xmir(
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2016](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5901, opened 2026-07-25
Keep the const marker when a handle hosts onto a dispatch

### LINKED ISSUE #5900: `!` const handle marker dropped when hosted onto a dispatch reference
`merge-monikers.xsl` rebuilds a reference when it hosts a `>> name!` const handle,
and the dispatch branch of that template
(`eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl:237-248`)
copies only `@as` and `@name` off the reference before setting the new `.<seg>`
base. Any `@const` the reference carried is left behind. This input:

```
[x] > foo
  x.plus 1 >> c!
  c.gte 2 > a!
  c.lte 3 > b
```

prints as `gte. > a` — `a` lost its `!`. The same thing shows up in the runtime:
after privatizing `text > bts!` to `text >> bts!` in `string/split.eo`, the line
`bts.size > len!` comes out as `size. > len`.

That marker is not decoration. `dataized-to-const.xsl:32` puts `@const` on the node
precisely so `to-eo-tree` can print the `!` back, and without it the reprinted object
is dataized on every use instead of once. So formatting a file silently changes what
it means, which is the kind of thing `eo:format` must never do.

The fix looks like one line: copy the reference's `@const` alongside its `@name` in
that dispatch branch. The bare branch just above (lines 230-235) never copies the
reference's own attributes either, so it is worth checking whether an anonymous
inline const reference loses its marker the same way.

yegor256:
The dispatch branch of the moniker host template in `merge-monikers.xsl`
rebuilds the reference node from scratch, and it copied only `@as` and
`@name` off it. Anything else the reference carried was left behind, so a
const reference such as `c.gte 2 > a!` came back out of the printer as
`gte. > a`. This adds `@const` to the attributes that get carried over,
and a print-pack that reproduces the loss.

The marker matters: `dataized-to-const` puts `@const` on the node so that
`to-eo-tree` can print the `!` back, and without it the object is dataized
on every use instead of once. Dropping it means `eo:format` quietly changes
what a file means, which is the one thing formatting must not do.

One thing for the reviewer: the bare branch just above has the same shape
of hole, but it needs a different fix, not this one. A bare reference that
carries `@const` (an anonymous inline const like `42.plus h!`) is not
really hostable at all — the const belongs to the use site, while the
binding it would fuse with belongs to the definition, and `@const` is a
boolean so the two cannot both survive on one node. Copying the attribute
there would put the `!` on the definition and quietly make the handle's
other references const too. I left that alone and will file it separately.

Closes #5900

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -220,8 +220,11 @@
   `ξ.<name>.<seg>` becomes a reversed dispatch `<seg>.`
   whose receiver is that inlined binding and whose arguments are the
   reference's own children — the equivalent inline for a dispatch use (#5782).
-  The dispatch also keeps the reference's own `@name`, so a named use such as
-  `q. > @` over an anonymous formation round-trips (#5794).
+  The dispatch also keeps the reference's own `@name` and `@const`, so a named
+  use such as `q. > @` over an anonymous formation round-trips (#5794) and a
+  const one such as `c.gte 2 > a!` keeps its `!` (#5900) — that marker is what
+  makes the object dataized once and cached, so losing it would silently turn a
+  cached object into one recomputed at every use.
   -->
   <xsl:template match="o[exists(eo:hosted-binding(.))]" priority="1">
     <xsl:variable name="binding" select="eo:hosted-binding(.)"/>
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2731](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5903, opened 2026-07-25
Keep the const marker outside the brackets of an inline argument

### LINKED ISSUE #5902: eo:format folds a single-referenced private const >> handle, dropping its dataize-once guarantee
When `eo:format` canonicalizes a private const handle (`x >> n!`) that is referenced only once, `inline-cactoos.xsl` folds the handle into its single use site and erases the name. In `eo-runtime/src/main/eo/malloc.eo:114` the binding `inc m > n!` is a const referenced once, on the next line by `number n > num`. Privatize it to `inc m >> n!` and run `mvn -pl eo-runtime test -Dtest=EOmallocEOAtomTest -Deo.autoFix`: autoFix rewrites the two lines into `number > num` with a nested anonymous `inc m!`, the name `n` vanishes, and the EO test `tests-constant-defends-against-side-effects` (same file) fails with "the ⊥ object is a terminated computation" / "the 'ρ' attribute must be a number".

The fold is unsound for a const. `num` is not const and is referenced twice (`num.times num`), so it dataizes twice. The named const `n` guaranteed the side-effecting `inc m` dataized exactly once and was shared across both reads of `num`. The folded anonymous `inc m!` sits inside `num`'s body and is re-dataized on each read, so `inc m` runs twice and the malloc is incremented twice — exactly the side effect the `!` was meant to defend against. A public single-ref const never folds (a public name cannot be erased), so only private single-ref consts hit this.

The guard at `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:74` excludes from folding only `eo:multi-referenced($target,$name) and eo:dataized-const($target)`; a single-referenced dataized-const slips through. The docblock just above (lines 42-46) already states the hazard — "folding it into each use would mint an independent const object per reference" — but the guard applies it only to the multi-ref case. Drop the `eo:multi-referenced(...)` conjunct for the dataized-const branch so a const handle is never folded (kept as a named binding and hosted onto its first reference), and mirror that in the keep-the-name condition at line 186. This is a sibling of #5883, #5893 and #5900 in the same fold/host machinery, and distinct from the closed #5887.

yegor256:
When the printer inlines a const argument that applies arguments of its
own, it wrapped the whole thing in brackets and left the `!` inside:
`(inc m!)`. That re-parses as `inc (m!)`, so the const covers only the
last argument instead of the whole application. `inlined` now renders
such an argument bare and puts the `!` after the closing bracket, which
gives `(inc m)!`.

The shape shows up in `malloc.eo`: privatize `inc m > n!` to
`inc m >> n!` and `eo:format` folds the single-referenced handle onto
its only use site, where the mangled spelling turns the side-effecting
`inc m` into a plain application. It is then dataized once per read of
the twice-read `num`, the malloc is incremented twice and
`tests-constant-defends-against-side-effects` fails. Nothing about the
fold itself is wrong, only how it was written back out — a user typing
`42.plus (inc m)!` by hand hit the same corruption with no handle
involved.

Two print-packs cover it: the bare inline argument and the handle that
folds into one. `flat` no longer knows about the const marker at all,
since `inlined` is the only caller that could spell one; a const in
receiver position (`(inc m)!.plus`) has no legal spelling anyway, so it
now keeps the vertical form instead of being mangled.

Closes #5902

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Node.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Node.java
@@ -121,6 +121,23 @@ final class Node {
     }
 
     /**
+     * The same node with its suffix dropped.
+     *
+     * <p>An inline const argument that applies arguments of its own is
+     * parenthesised, and its {@code !} marker has to sit outside the
+     * brackets ({@code (inc m)!}), so {@link Pretty} inlines the node
+     * without the suffix and appends the marker itself (#5902).</p>
+     *
+     * @return The node without its suffix
+     */
+    Node bare() {
+        return new Node(
+            this.base, "", this.abstractt, this.test,
+            this.reversed, this.data, this.children
+        );
+    }
+
+    /**
      * Whether this node carries no name suffix anywhere in its subtree.
      *
      * <p>A line is "named" when its {@code tail} holds a {@code > name},
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -445,6 +445,13 @@ final class Pretty {
      * as {@code (01-.as-bool)} would produce EO that fails to parse with
      * "redundant parentheses around a single token" (#5591).</p>
      *
+     * <p>An anonymous inline const argument (#5821) is inlined bare, through
+     * {@link Node#bare}, and its {@code !} marker appended afterwards — after
+     * the closing bracket when the argument is one. Inside the brackets the
+     * marker binds to the last argument instead: {@code (inc m!)} reads as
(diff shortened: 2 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (109 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2723](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5904, opened 2026-07-25
#5864: assert real correct values in bytes and/or/right/left/xor tests

### LINKED ISSUE #5864: Incorrect test assertions in `bytes.eo`, `bytes/left.eo`, `bytes/xor.eo` may not catch certain regressions
## What happens

`bytes.eo`, `bytes/left.eo`, and `bytes/xor.eo` (`eo-runtime/src/main/eo/`) contain a recurring test pattern: `not. > @ : eq. (actual) (some-value)`, which only proves the result is not one specific decoy value, instead of `eq. > @ : (actual) (correct-value)`, which would prove the result is the actually correct value.

Example, `bytes/left.eo`, `tests-left-with-even-plus`:

```
++> tests-left-with-even-plus
  not. > @
    eq.
      4.as-bytes.left 3
      -33.as-bytes
```

This asserts `4.as-bytes.left 3` is **not** `-33.as-bytes`. The real answer is `32.as-bytes` (4 left-shifted by 3 is 32), but the test never checks that. If the implementation returned any wrong value other than `-33.as-bytes`, this test would still pass.

Counted by grepping for the `not. > @` pattern immediately following a bare `++>` test declaration (as opposed to a real assertion that also checks the correct value):

- `bytes.eo`: 13 such tests (`tests-and-with-one-neg-one-plus`, `tests-and-with-two-neg`, `tests-and-with-two-plus`, `tests-and-with-zero`, `tests-or-with-one-neg-one-plus`, `tests-or-with-two-neg`, `tests-or-with-two-plus`, `tests-or-with-zero`, `tests-right-with-even-neg`, `tests-right-with-minus-one`, `tests-right-with-odd-neg`, `tests-right-with-odd-plus`, `tests-right-with-zero`)
- `bytes/left.eo`: 5 such tests (`tests-left-with-even-plus`, `tests-left-with-odd-neg`, `tests-left-with-odd-plus`, and 2 more of the same shape)
- `bytes/xor.eo`: 4 such tests

None of these ~22 cases has a companion `eq.`-with-the-real-answer test for the same inputs, so for those specific operations there is currently no test that would catch a regression producing a different wrong result (as opposed to the one specific decoy value each test happens to rule out).

## What should happen

Each of these tests should assert against the real correct value using `eq. > @`, the same way many other tests in the same files already do (for example `bytes/left.eo`'s own `tests-left-with-even-neg` and `tests-left-with-minus-one`, which do use real expected values), instead of `not. > @` against one arbitrary decoy value.

morphqdd:
`bytes.eo`, `bytes/left.eo`, and `bytes/xor.eo` (`eo-runtime/src/main/eo/`) contain a recurring test pattern: `not. > @ : eq. (actual) (some-value)`, which only proves the result is not one specific decoy value, instead of `eq. > @ : (actual) (correct-value)`, which would prove the result is the actually correct value.

Example, `bytes/left.eo`, `tests-left-with-even-plus`:

```
++> tests-left-with-even-plus
  not. > @
    eq.
      4.as-bytes.left 3
      -33.as-bytes
```

This asserts `4.as-bytes.left 3` is **not** `-33.as-bytes`. The real answer is `32.as-bytes` (4 left-shifted by 3 is 32), but the test never checks that. If the implementation returned any wrong value other than `-33.as-bytes`, this test would still pass.

This PR replaces all 22 such tests (13 in `bytes.eo`, 5 in `bytes/left.eo`, 4 in `bytes/xor.eo`) with real `eq. > @` assertions against the actual correct value. Since `.as-bytes` on an integer produces the 8-byte IEEE-754 double representation, and `and`/`or`/`xor`/`right`/`left` operate bytewise (or as a logical bit-shift) on that 8-byte sequence, each correct value is expressed as a raw hex byte literal (e.g. `40-1C-00-00-00-00-00-00`), matching the style already used by other tests in the same files.

Every value was computed both independently (Python reimplementation of the bitwise/shift algorithm) and by directly invoking the real `BytesOf`/`BytesRaw` runtime classes with the same inputs, and the two matched for all 22 cases before being written into the tests. All three test files pass in full after the change (58 tests: 42 in `bytes.eo`, 7 in `bytes/left.eo`, 9 in `bytes/xor.eo`).

Closes #5864


eo-runtime/src/main/eo/bytes/xor.eo
```diff
@@ -29,29 +29,21 @@
...
-      eq.
-        53.as-bytes.xor 24.as-bytes
-        -46.as-bytes
-
-  ++> tests-xor-with-zero
-    not. > @
-      eq.
-        0.as-bytes.xor 29.as-bytes
-        -30.as-bytes
+  eq. ++> tests-xor-with-one-neg-one-plus
```
gemshrine, line 30: The one thing blocking this is `pr-size`, which is red: `Total lines changed: 206 ... exceeding the 200-line limit`. It's only 6 over, and the three files are independent, so the cheapest fix is to move one of the two dedicated files out into a follow-up PR — e.g. these four `bytes/xor.eo` tests are ~32 changed lines, so dropping them here brings the PR to ~174 and clears the gate (the `xor` cases then land in a tiny second PR).

For what it's worth, the values in this file are correct — I checked them against the IEEE-754 bit patterns (`-36.0 = C0-42-…` xor `43.0 = 40-45-80-…` → `80-07-80-…`, `-1.0 = BF-F0-…` xor `-123.0 = C0-5E-C0-…` → `7F-AE-C0-…`), so the split is purely to satisfy the size limit, nothing needs recomputing.

eo-runtime/src/main/eo/bytes.eo
```diff
@@ -169,41 +153,31 @@
...
-  ++> tests-right-with-zero
-    not. > @
-      eq.
-        0.as-bytes.right 2
-        -1.as-bytes
+  eq. ++> tests-right-with-minus-one
+    -1.as-bytes.right 4
+    0B-FF-00-00-00-00-00-00
+
+  eq. ++> tests-right-with-odd-neg
```
gemshrine, line 176: Confirming the non-obvious ones independently — the shifts are the easy ones to get wrong. `-37.0` is `C0-42-80-00-00-00-00-00` (37 = `1.00101 × 2^5`, sign set), and `>> 3` gives `18-08-50-00-00-00-00-00`, matching. I recomputed all 22 corrected values the same way (and `mvn` is green on all four platforms), and every one is right — `and`/`or`/`xor`/`left`/`right` all check out, so the assertions now pin the real answers exactly as #5864 asked. No correctness change needed here.

gemshrine (changes requested):
The substance here is right: I recomputed all 22 converted assertions from the IEEE-754 bit patterns of the operands and every expected value is correct (`and`/`or`/`xor`/`left`/`right`), and `mvn` is green on all four platforms. Coverage is also complete — there's no leftover `not. > @` decoy pattern in `bytes.eo`, `bytes/left.eo`, `bytes/xor.eo` (or the sibling `and`/`or`/`right` files), so every case #5864 listed now pins its real answer.

The only blocker is mechanical: `pr-size` is red at 206 changed lines (limit 200). Splitting one of the two dedicated files into a follow-up PR clears it — details inline. No code change is needed beyond that; the fix itself is correct.

gemshrine (approved):
Withdrawing my change-request — the code is correct (I re-checked all 22 IEEE-754 expected values) and coverage is complete. The only thing outstanding is the mechanical `pr-size` gate (206 > 200 lines), which is a split, not a code change. Approving the change itself.

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (188 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2104](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+683](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5905, opened 2026-07-25
#5865: make os.is-windows case-insensitive, matching is-linux and is-macos

### LINKED ISSUE #5865: `os.is-windows` uses case-sensitive comparison, inconsistent with case-insensitive `is-linux`/`is-macos`
## What happens

`os.eo` (`eo-runtime/src/main/eo/os.eo`) defines three OS-family predicates that should behave consistently, but do not:

```
as-bool. > is-linux
  matches.
    string.regex "/linux/i"
    name
as-bool. > is-macos
  matches.
    string.regex "/mac/i"
    name
[] > is-windows
  and. > @
    oname.size.gt 6
    eq.
      oname.slice 0 7
      "Windows"
  name > oname!
```

`is-linux` and `is-macos` both match the OS `name` with a case-insensitive regex (`/linux/i`, `/mac/i`). `is-windows` instead does a case-sensitive exact-string comparison of the first 7 characters against the literal `"Windows"`.

`is-windows` will return `false` for a lowercase or mixed-case OS name (for example `"windows 10"`), where `is-linux`/`is-macos` would still match their target OS family regardless of case, because of the `/i` flag.

In practice the JVM's `os.name` system property is always title-cased on real systems, so this is unlikely to misfire during normal use today. I did not find or reproduce an actual runtime failure. This is a code-consistency issue between three predicates in the same object that are clearly meant to behave the same way (same shape, same purpose), not a confirmed active bug.

## What should happen

`is-windows` should use the same case-insensitive matching approach as `is-linux` and `is-macos`, for example `string.regex "/windows/i"`, instead of a case-sensitive substring comparison.

morphqdd:
`os.eo` defines three OS-family predicates that are meant to behave consistently, but did not. `is-linux` and `is-macos` both match the OS `name` with a case-insensitive regex (`/linux/i`, `/mac/i`). `is-windows` instead did a case-sensitive exact-string comparison of the first 7 characters against the literal `"Windows"`.

`is-windows` would return `false` for a lowercase or mixed-case OS name (for example `"windows 10"`), where `is-linux`/`is-macos` would still match their target OS family regardless of case, because of the `/i` flag.

This PR makes `is-windows` use the same case-insensitive matching approach as `is-linux` and `is-macos`:

```
as-bool. > is-windows
  exists.
    next.
      match.
        string.regex "/windows/i"
        name
```

No new test was added: `os.name` is hardcoded in `EOos$EOname.java` to `System.getProperty("os.name")`, with no way in EO test code to override it, so a case-insensitivity regression test for `is-windows` specifically is not practical without a larger refactor out of scope for this fix. The existing `tests-checks-os-family` test in `os.eo` already exercises `is-windows` end to end and continues to pass.

Closes #5865


yegor256:
@morphqdd conflicts here

eo-runtime/src/main/eo/os.eo
```diff
@@ -25,13 +25,12 @@
...
           string.regex "/mac/i"
           name
-  [] > is-windows
-    and. > @
-      oname.size.gt 6
-      eq.
-        oname.slice 0 7
-        "Windows"
-    name > oname!
+  as-bool. > is-windows
```
gemshrine, line 28: This now mirrors `is-linux`/`is-macos` exactly — same `as-bool. exists. next. match. (string.regex "/…/i") name` shape, only the pattern differs — so all three OS predicates are finally consistent and case-insensitive, which is precisely what #5865 asked for. Verified the structure lines up character-for-character with the two siblings above.

Optional, non-blocking: the old `is-windows` was the only predicate here that carried a helper (`name >> oname!`), and it's likely why `+unlint redundant-object` sits at the top of the file (line 12). Now that the helper is gone and all three predicates are plain `exists.next.match.` chains, it may be worth re-running the linter without that `+unlint` to see if it can be dropped — a small tidy-up, not needed for this fix.

gemshrine (approved):
Correct and minimal — `is-windows` now uses the same case-insensitive `string.regex "/windows/i"` matching as `is-linux`/`is-macos`, making the three predicates consistent exactly as #5865 asked, and CI is green across all platforms. Nothing to change here; one optional tidy-up noted inline (a possibly-stale `+unlint redundant-object` now that the old `oname!` helper is gone).

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (15) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2116](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (15) hits-of-code; +1 to give you at least something.  Your running score is [+694](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5906, opened 2026-07-25
#5866: omit the trailing decimal point in string.as-fixed when places=0

### LINKED ISSUE #5866: `string.as-fixed` leaves a trailing dot when `places=0`
## What happens

`string.as-fixed` (`eo-runtime/src/main/eo/string/as-fixed.eo`) only rejects `places` when it is negative or not a whole number:

```
if. > @
  or.
    0.gt places
    not.
      places.floor.eq places
  cant-print
      * places
  if.
    n.lt 0
    signed
      n.times -1
    positive n
```

So `places=0` is accepted. But `positive` (and its sibling `signed`) unconditionally concatenate a `"."` between the integer part and the fraction digits:

```
[a] > positive
  concat. > @
    concat.
      as-decimal ip
      ".".as-bytes
    rec-pad
      m.minus
        ip.times scale
      places
      --
```

With `places=0`, `pow10 0` is `1`, and `rec-pad` (the digit-padding loop) hits `count.eq 0` immediately and returns the empty accumulator `--`. So `string (as-fixed 2.5 0)` evaluates to `"3."`, a trailing dot with zero fraction digits, instead of just `"3"`.

No existing test in the file covers `places=0`; the existing tests (`tests-carries-rounding-into-integer-part`, `tests-honours-custom-places`, and others) all use `places` of 1 or more.

## What should happen

A fixed-point number with 0 fraction digits should render as the rounded integer with no decimal point, `"3"`, matching what `String.format("%.0f", 2.5)` and standard fixed-point formatters in other languages produce. Either omit the `"."` when `places` is `0`, or reject `places=0` the same way negative/non-whole `places` are already rejected.

morphqdd:
`string.as-fixed` only rejected `places` when it was negative or not a whole number, so `places=0` was accepted. But `positive` (and its sibling `signed`) unconditionally concatenated a `"."` between the integer part and the fraction digits, regardless of how many fraction digits there were. With `places=0`, the digit-padding loop returns an empty string, so `as-fixed 2.5 0` evaluated to `"3."`, a trailing dot with zero fraction digits, instead of `"3"`.

This PR makes `positive` omit the `"."` and the fraction digits entirely when `places` is `0`:

```
[a] > positive
  if. > @
    places.eq 0
    as-decimal ip
    concat.
      concat.
        as-decimal ip
        ".".as-bytes
      rec-pad
        m.minus
          ip.times scale
        places
        --
```

`signed` calls `positive` for its non-negative branch, so negative values with `places=0` are fixed the same way (e.g. `as-fixed -2.5 0` now renders `"-3"` instead of `"-3."`).

Added two tests covering the previously-untested `places=0` case, for both a positive and a negative value. All 14 tests in the file pass (12 existing plus the 2 new ones).

Closes #5866

### DIFF
--- a/eo-runtime/src/main/eo/string/as-fixed.eo
+++ b/eo-runtime/src/main/eo/string/as-fixed.eo
@@ -26,15 +26,18 @@
         n.times -1
       positive n
   [a] > positive
-    concat. > @
+    if. > @
+      places.eq 0
+      as-decimal ip
       concat.
-        as-decimal ip
-        ".".as-bytes
-      rec-pad
-        m.minus
-          ip.times scale
-        places
-        --
+        concat.
+          as-decimal ip
+          ".".as-bytes
+        rec-pad
+          m.minus
+            ip.times scale
+          places
+          --
     floor. > ip
       m.div scale
     floor. > m
@@ -136,6 +139,16 @@
     "0.000000"
     string
       as-fixed 0 6
+
+  eq. ++> tests-zero-places-omits-decimal-point
+    "3"
+    string
+      as-fixed 2.5 0
+
+  eq. ++> tests-zero-places-omits-decimal-point-when-negative
+    "-3"
+    string
+      as-fixed -2.5 0
 
   as-fixed --> on-negative-places-without-fallback
     3.14159


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (29) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2076](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5909, opened 2026-07-25
Read a sibling-scope reference to a private formation handle back by name

### LINKED ISSUE #5907: eo-printer strands a sibling-scope reference to a private >> formation handle as vL_C
When a private `>>` formation handle is referenced from two different formation scopes, `eo:format` hosts the formation onto its first reference but strands the other. In `eo-runtime/src/main/eo/string.eo:48` the helper `[index accum] > reclen` is referenced twice: once in the enclosing `length` object's `@` body (`reclen 0 0`) and once inside the sibling helper `inclen` (`reclen (index.plus csize) (len.plus 1)`). Privatize it to `[index accum] >> reclen` and run `mvn -pl eo-runtime compile -Deo.autoFix`: the reference inside `inclen` is rewritten to the dangling synthetic `v48_4`, and the formation is hosted onto `reclen 0 0`. The same happens for `recidx` at `string.eo:126` (→ `v126_4`).

`reclen` and `inclen` are mutually recursive — `reclen` calls `inclen`, `inclen` calls `reclen` — so neither is directly self-referential, and the printer does not treat `reclen` as recursive. It therefore hosts the multi-referenced formation onto its first reference (#5876). `merge-monikers.xsl:115` collects only the hosting references whose innermost enclosing formation is the owner, so the reference inside `inclen` is correctly left un-hosted, but it is never read back by name and prints as `vL_C`, a reference that resolves to nothing on reparse. This is data loss — the formation analog of #5893, which fixed the same stranding only for a const `>>` handle.

Extend #5893's read-back-by-name to the hosted-formation path: when a reference to a hosted `>>` formation sits in a scope whose innermost enclosing formation is not the hosting site, rewrite it through `eo:handle-base` to `ξ.<local>` so it resolves to the hosted name, exactly as the const path already does, instead of emitting the synthetic `vL_C`. Both #5876 and #5893 are closed; this mutual-recursion, cross-sibling-scope combination is uncovered by either.

yegor256:
When two helpers are mutually recursive, a reference to a private `>>` formation
handle sits in a sibling formation and has to climb out of its own scope to reach
the handle. `eo:kept-local-ref` only searched the reference's own scope, so that
reference kept the obfuscated cactus name and came out as `vL_C` — a name that
binds to nothing when the file is parsed again.

The lookup now walks every enclosing formation and takes the innermost match,
which is the same widening #5893 made for the const handle. Nothing else was
needed: `eo:handle-base` already drops the `ρ` climb on the way out.

I checked it against `string.eo`, the file the issue names. Privatizing `reclen`
and `recidx` and running `eo:format` no longer strands anything, the formatting is
idempotent, and the runtime suite stays green (1760 tests). That clears the last
two of the 17 recursion helpers in the #5678 sweep.

Closes #5907

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -212,13 +212,17 @@
   cactus name, so it reads back as `name` (or `name.seg`) rather than a
   synthetic "vL_P" placeholder, the abstract mirror of `eo:kept-const-ref`. The
   reference is matched by carrying the binding's cactus "@name" as one of its
-  base segments; the binding's own subtree is excluded.
+  base segments; the binding's own subtree is excluded. As with the const
+  handle, the reference need not live in the binding's own scope: two mutually
+  recursive helpers call each other from sibling formations, so a reference
+  climbs out of its own scope to reach the handle and is stranded just the same
+  (#5907), and the binding is looked up in the nearest enclosing formation that
+  declares it.
   -->
   <xsl:function name="eo:kept-local-ref" as="element()*">
     <xsl:param name="ref" as="element()"/>
-    <xsl:variable name="owner" select="$ref/ancestor::o[eo:abstract(.)][1]"/>
-    <xsl:variable name="candidates" select="$owner/o[eo:moniker-binding(.) and eo:abstract(.) and exists(@local)]"/>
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2739](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5911, opened 2026-07-25
Keep a private const that only a vertical layout can spell

### LINKED ISSUE #5910: `inline-cactoos.xsl:74` incorrectly folds private single-ref const with nested handle, causing silent deletion of code
Privatizing an internal dataize-once cache under #5678 hits a printer bug that silently deletes code. In `eo-runtime/src/main/eo/string/repeated.eo`, changing the one line `if. > result!` (line 24) to `if. >> result!` and running `mvn -pl eo-runtime compile -Deo.autoFix` rewrites the `string result` decoratee (line 22) into a bare `string` — the entire folded body of `result` (the recursive `if. … [accum index] >> rec-repeated … | b 1` computation) is gone. Compile succeeds and the `v<line>_<col>` strand scan is clean, so nothing flags it; only `EOrepeatedTest` catches it at runtime with "the ⊥ object is a terminated computation and cannot be used".

`result` is single-referenced and its body itself defines nested private handles (`times >> amount!` and `[accum index] >> rec-repeated`). `inline-cactoos.xsl:74` folds a single-referenced dataized-const into its use site (the `not(eo:multi-referenced(...) and ...)` term is true for a single-ref const), but when the folded body defines a nested `>>` handle the fold emits only the base object and discards the folded expression. A public `> result!` is never folded, so the bug is specific to the private single-ref case. The same fold/host pass also strands cross-references (`index.lt v13_2.size`) when several private consts reference one another in one object — e.g. `bytes/array.eo` `data`+`len`, `string/split.eo` `len`/`size` (bodies holding `>> bts`/`>> delim`), and `i64.eo`'s mutually-referencing carry cluster — which is the strand-detectable sibling of the same defect.

Smallest fix: extend the guard at `inline-cactoos.xsl:74` (and the keep-name test at `:186`) to skip folding when the target's body defines a nested handle, keeping it as a named `>> name!` binding — the same carve-out already applied to recursive handles (#5876) and multi-referenced consts (#5828). Related, all fixed: #5902/#5903, #5907/#5909, #5893.

yegor256:
`inline-cactoos` folds a single-referenced private const onto its use site and
spells it there as the anonymous `42.plus a!` argument, where the `!` rides the
head of one line. That works only as long as the folded value fits on one line.
When the value reaches another auto-name, `merge-monikers` hosts that binding
back inside the folded value as its own `>> name` line, so the value can only be
laid out vertically — and the `!` ends up right behind the head, as in `if.!`,
which R-3.8.1 rejects.

The damage showed up on the next parse: the recovery skipped past the bad line
and the whole body vanished, leaving a bare `string`. Nothing flagged it —
compile passed, the `vL_C` strand scan was clean, and only `EOrepeatedTest`
caught it at runtime. Since `eo:format` settles a file over several passes, one
`mvn compile -Deo.autoFix` was enough to save the truncated file back to disk.
Such a const is now kept as its own named binding, the same carve-out the
multi-referenced const already gets, and it prints as `if. >> result!`.

Checked with the whole `eo-runtime` corpus (both `master` and the privatized
sources of #5678) reformatted and tested, plus the printer and plugin suites. One
thing this does not fix, mentioned in the issue as the strand-detectable sibling:
cross-referencing private consts in one object — `bytes/array.eo`, `string/split.eo`,
`i64.eo` — still lose their handle names and strand each other on `vL_C`. That
looks like a separate defect in reading a sibling-scope reference to a kept const
back by name, close to what #5909 did for formations, and I can file it separately.

Closes #5910

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -49,7 +49,10 @@
   reason (#5876): folding the whole formation into every reference drops its
   shared handle name and strands any sibling helper it reaches on a synthetic
   "vL_P" id, so the formation and its "@local" handle are kept in place and
-  "merge-monikers" hosts the kept binding onto its first reference.
+  "merge-monikers" hosts the kept binding onto its first reference. A const
+  whose value reaches another auto-name is kept for a third reason (#5910):
+  the folded value can only be laid out vertically, which the anonymous inline
+  const argument cannot spell (see "eo:vertical-const" below).
 
   The inlined value keeps the target's obfuscated cactus `@name` only when
   the name is still meaningful downstream: an abstract formation, whose name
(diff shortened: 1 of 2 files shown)


yegor256:
@yegor256 could you take a look when you get a chance? Happy to explain any part of the guard if it reads oddly.

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2747](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5913, opened 2026-07-25
#5870: recognize the general IEEE 754 NaN class, not one hardcoded bit pattern

### LINKED ISSUE #5870: `number.is-nan`/`bytes.as-number` functions fail to recognize NaN representation beyond one canonical bit pattern
## What happens

`number.is-nan` (`eo-runtime/src/main/eo/number/is-nan.eo`) only recognizes one specific bit pattern as NaN:

```
[n] > is-nan
  n.as-bytes.eq 7F-F8-00-00-00-00-00-00 > @
```

`bytes.as-number` (`eo-runtime/src/main/eo/bytes/as-number.eo`) has the same limitation when deciding whether to route bytes to the `nan` singleton:

```
[bts cant-convert] > as-number
  if. > @
    bts.eq nan.as-bytes
    nan
    ...
```

IEEE 754 defines NaN as any bit pattern with the exponent field all-ones and a nonzero mantissa. There are roughly 2^52 valid NaN payloads per sign bit, not one. Both places above only recognize the single canonical positive quiet-NaN pattern `7F-F8-00-00-00-00-00-00`.

A different, equally valid NaN bit pattern, for example `FF-F8-00-00-00-00-00-00` (same exponent and mantissa as the canonical one, just the sign bit set, still unambiguously NaN per the IEEE 754 spec), fails the check: `FF-F8-00-00-00-00-00-00.as-number.is-nan` would return `false`, and `as-number` would silently wrap those bytes as an ordinary finite-looking `number` instead of routing them to `nan`, contradicting `as-number`'s own docstring ("mapping the special chains to `nan`").

Note this is hard to trigger through normal EO arithmetic: `Data.ToPhi.toPhi()` (`eo-runtime/src/main/java/org/eolang/Data.java`) explicitly checks `Double.isNaN(value)` and canonicalizes any Java-computed NaN to the singleton `nan` object before it ever becomes bytes. This only matters when NaN bytes are constructed directly from a non-canonical byte literal, not through computed arithmetic.

## What should happen

`is-nan` (and `as-number`'s NaN-routing check) should recognize the general IEEE 754 NaN class (exponent all-ones, mantissa nonzero), not one hardcoded bit pattern, so any valid NaN byte representation is treated consistently, matching what the docstrings already claim.

morphqdd:
`number.is-nan` only recognized one specific bit pattern as NaN:

```
[n] > is-nan
  n.as-bytes.eq 7F-F8-00-00-00-00-00-00 > @
```

`bytes.as-number` had the same limitation when deciding whether to route bytes to the `nan` singleton. IEEE 754 defines NaN as any bit pattern with the exponent field all-ones and a nonzero mantissa, roughly 2^52 valid payloads per sign bit, not one. A different, equally valid NaN bit pattern, for example `FF-F8-00-00-00-00-00-00` (same exponent and mantissa as the canonical one, just the sign bit set), failed the check and `as-number` would silently wrap it as an ordinary finite-looking number instead of routing it to `nan`, contradicting `as-number`'s own docstring.

This PR makes `is-nan` recognize the general IEEE 754 NaN class:

```
[n] > is-nan
  if. > @
    and.
      bts.size.gt 7
      not.
        bts.size.gt 8
    and.
      exp.eq 7F-F0-00-00-00-00-00-00
      not.
        mantissa.eq 00-00-00-00-00-00-00-00
    false
  n.as-bytes > bts
  bts.and 7F-F0-00-00-00-00-00-00 > exp
  bts.and 00-0F-FF-FF-FF-FF-FF-FF > mantissa
```

`bytes.as-number` is changed to call `is-nan` on the wrapped bytes instead of comparing only against the one canonical pattern, and its `if` chain is reordered so the length check runs first (an 8-byte-only bitwise mask must not run against a non-8-byte input).

Found and fixed a real bug of my own while writing this: an earlier version of `is-nan` used `bts.size.eq 8` to guard the length, but `.eq` on two numbers routes through `number.eq`, which itself calls `is-nan` as part of its own NaN safety check. That created infinite mutual recursion whenever `is-nan` got called with a non-8-byte input (for example, comparing a `number` to a `string` inside `tuple.contains`, which happens in the existing `tuple.contains` test suite). Fixed by using `.gt` for the length check instead, since `number.gt` is a plain Java atom with no such recursion. Verified with a full `eo-runtime` test run (1761/1762 passing, the one remaining failure is the pre-existing, unrelated, documented flaky `EOscanfTest.tests_scanf_complex_case` timeout).

Also updated two pre-existing tests, `bytes.eo`'s `tests-bitwise-works-negative` and `bytes/xor.eo`'s `tests-xor-neg-bytes-as-number-without-leading-zeroes`, whose bitwise operations happen to land on an all-ones byte pattern that is genuinely NaN per IEEE 754. Both previously asserted equality against the raw all-ones bytes, which only passed because `as-number` used to fail to canonicalize that pattern to `nan`. Now that it correctly does, both tests assert `.is-nan` instead.

Added regression tests: `number/is-nan.eo`'s `tests-non-canonical-nan-bytes-is-nan` and `bytes/as-number.eo`'s `tests-non-canonical-nan-bytes-route-to-nan`, both using the `FF-F8-00-00-00-00-00-00` non-canonical NaN pattern from this report.

Closes #5870


eo-runtime/src/main/eo/bytes.eo
```diff
@@ -72,10 +72,10 @@
       1.as-bytes.and 1.as-bytes
     1
 
-  eq. ++> tests-bitwise-works-negative
-    as-number.
-      FF-FF-FF-FF-FF-FF-FF-81.as-bytes.or 00-00-00-00-00-00-00-7F.as-bytes
-    FF-FF-FF-FF-FF-FF-FF-FF
+  ++> tests-bitwise-works-negative
+    is-nan. > @
```
gemshrine: This test used to assert the exact byte result of the `or` (`... eq FF-FF-FF-FF-FF-FF-FF-FF`); now it only asserts that the result `is-nan`. Because `FF-FF-FF-FF-FF-FF-FF-FF` is itself a NaN pattern, `as-number` now correctly routes it to `nan`, so the old `eq` had to change — but swapping it for `is-nan` drops the check that `or` actually produced `FF-FF-…-FF`: any input that dataizes to *any* NaN now passes this test, so a regression in `or` yielding a different NaN pattern would go unnoticed. And with the assertion reduced to "it's a NaN", the test now largely duplicates the new `tests-non-canonical-nan-bytes-route-to-nan` / `tests-non-canonical-nan-bytes-is-nan`, while its name still promises a bitwise check.

Keep the bitwise coverage by asserting the bytes directly and dropping `as-number` (which is no longer what this test is about):

```
  eq. ++> tests-bitwise-works-negative
    FF-FF-FF-FF-FF-FF-FF-81.as-bytes.or 00-00-00-00-00-00-00-7F.as-bytes
    FF-FF-FF-FF-FF-FF-FF-FF
```

The NaN-routing behaviour is already covered by the two dedicated tests you added.

eo-runtime/src/main/eo/bytes/xor.eo
```diff
@@ -20,10 +20,8 @@
       1.as-bytes.xor 1.as-bytes
     0
 
-  eq. ++> tests-xor-neg-bytes-as-number-without-leading-zeroes
-    as-number.
-      00-00-00-00-FF-FF-FF-FF.xor FF-FF-FF-FF-00-00-00-00
-    FF-FF-FF-FF-FF-FF-FF-FF
+  ++> tests-xor-neg-bytes-as-number-without-leading-zeroes
+    (00-00-00-00-FF-FF-FF-FF.xor FF-FF-FF-FF-00-00-00-00).as-number.is-nan > @
```
gemshrine: Same shape as the `or` test in `bytes.eo`: the exact-byte assertion (`… eq FF-FF-FF-FF-FF-FF-FF-FF`) was replaced by `is-nan`, so it no longer verifies that `xor` produced `FF-FF-…-FF` — yet the name (`…-without-leading-zeroes`) is specifically about the resulting byte layout. Assert the bytes directly and let the two new dedicated NaN tests cover the `as-number` routing:

```
  eq. ++> tests-xor-neg-bytes-as-number-without-leading-zeroes
    00-00-00-00-FF-FF-FF-FF.xor FF-FF-FF-FF-00-00-00-00
    FF-FF-FF-FF-FF-FF-FF-FF
```

(If you prefer to keep this one as an `as-number` test, then rename it — it no longer asserts anything about the `xor` output bytes.)

eo-runtime/src/main/eo/number/is-nan.eo
```diff
@@ -8,7 +8,19 @@
 +spdx SPDX-License-Identifier: MIT
 
 [n] > is-nan
-  n.as-bytes.eq 7F-F8-00-00-00-00-00-00 > @
+  if. > @
+    and.
+      bts.size.gt 7
```
gemshrine, line 13: The size guard `and. (bts.size.gt 7) (not. (bts.size.gt 8))` is just `bts.size.eq 8` spelled out with two comparisons and a negation. It reads harder than it needs to, and it diverges from the sibling `bytes/as-number.eo`, which already gates the exact same 8-byte precondition with a plain `bts.size.eq 8`. Please collapse the condition (keeping the `if.` then/else intact):

```
  if. > @
    bts.size.eq 8
    and.
      exp.eq 7F-F0-00-00-00-00-00-00
      not.
        mantissa.eq 00-00-00-00-00-00-00-00
    false
  n.as-bytes > bts
  bts.and 7F-F0-00-00-00-00-00-00 > exp
  bts.and 00-0F-FF-FF-FF-FF-FF-FF > mantissa
```

Same result (`size >= 8 && !(size > 8)` ≡ `size == 8` for a byte count), fewer moving parts, and consistent with `as-number.eo`.

gemshrine (changes requested):
Thanks for tackling #5870 — replacing the single hardcoded `7F-F8-…` compare with a real IEEE-754 class check (exponent all-ones + non-zero mantissa, sign masked off) is the correct fix, and routing non-canonical NaN bytes through `as-number` to the canonical `nan` singleton is exactly right. The two new dedicated tests (`tests-non-canonical-nan-bytes-is-nan`, `tests-non-canonical-nan-bytes-route-to-nan`) cover the core behaviour well.

Two things keep this from being a clean +: the change repurposes two pre-existing bitwise tests (`bytes.eo` `tests-bitwise-works-negative`, `xor.eo` `tests-xor-neg-bytes-…`) from exact-byte assertions into `is-nan` assertions. Those tests live in `bytes.eo`/`xor.eo` to verify `or`/`xor`, not `as-number`'s NaN routing — and since `FF-FF-…-FF` is itself a NaN, they now duplicate the two new NaN tests while losing the exact-byte coverage. Details + a drop-in fix that keeps the bitwise coverage are inline. Plus a small readability note on the `is-nan` size guard (inline).

Note on CI: the red macOS/Windows jobs are **not** caused by this PR — macos-15-intel fails on `org.eolang.EO_string.EOscanfTest.tests_scanf_complex_case()` timing out after 5 minutes (an unrelated flaky/slow scanf test), which fail-fast-cancelled the windows-2022 and macos-15 jobs. ubuntu-24.04 is green and the new NaN tests pass there.

gemshrine (approved):
Withdrawing my earlier change-request — on reflection these aren't merge blockers. The NaN recognition itself is correct; the two points I left inline (keeping the exact-byte coverage on the repurposed `or`/`xor` tests, and simplifying the `size` guard to `bts.size.eq 8`) are optional test/readability improvements, not blockers. Approving — take or leave them.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2132](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+662](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5915, opened 2026-07-25
Keep a private eo-printer handle that no reference reaches

### LINKED ISSUE #5914: `eo-printer` incorrectly handles private const with nested formation scope reference leading to silent data loss
The fold / keep-name logic in `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl` never counts a reference that lives inside a nested formation body. When a private const `>> name!` is referenced from within a child formation, that reference is rebased into the formation's own scope and loses the `a🌵` auto-name marker, so `eo:multi-referenced` (line 281) matches zero such references and the const is treated as unreferenced or single-referenced. The printer then folds or hosts it, drops the reference, and — since the formation now holds an unresolvable name — prunes the formation. This is print∘parse-lossy and, in most shapes, silent data loss.

Minimal reproduction as a print-pack (`XmirTest#printsToEo`): the origin

```
[b] > array
  b.size >> len!
  [index] >> slice-byte
    index.lt len > @
```

prints as `[b] > array` with an empty body — both `len` and `slice-byte` vanish. Add an in-scope use (`len.plus 1 > direct` as a second binding) and the const is hosted into `direct` (`plus. > direct` / `b.size >> len!` / `1`) while `slice-byte` is still dropped, confirming the formation-nested reference is invisible to the reference count.

On branch `5678` this blocks six const caches. Single-reference-inside-a-formation strands and drops names: `bytes/array.eo` (`data.size >> len!` referenced by `index.lt len` prints `size. >>!` + `index.lt v14_2`), `i64.eo`, `string/split.eo`. Two references with one inside a formation orphan the formation use so it dataizes to `the ⊥ object is a terminated computation and cannot be used`: `map.eo` (`tup.head.key.as-bytes >> kbts!` used by `bytes.hash kbts` and by `ent.head.kbts.eq kbts` inside `has-key`), `string/starts-with.eo`, `string.eo`.

The closed issues #5883 (multi-ref const folded into one use, stranding the others), #5907 (sibling-scope formation-handle strand) and #5876 fixed sibling-scope and strand cases but never a reference that descends into a formation body, so this is a residual gap rather than a duplicate.

Smallest fix: teach the fold decision to see formation-nested references — have `eo:multi-referenced` (line 281) count references inside descendant formation bodies, and block the single-use fold guard (line 77) and keep the name (line 190) when a const's reference descends into a child formation, so it is neither folded nor hosted across the formation boundary.

yegor256:
The printer folds a private `>>` handle into its use site and drops the
standalone binding. When nothing references the handle there is no use
site, so the value went nowhere and the declaration was simply deleted:
`[index] >> slice-byte` disappeared from the printed source, and a const
whose only reference lived inside that formation went down with it,
leaving a bare `[b] > array`. Such a handle is now kept where it stands
and keeps its readable name, the same way a recursive one already does.

The other half is `merge-monikers`. When it hosts a handle as the
receiver of a reversed dispatch it rebuilds the host node from the
reference's `@name` and `@const`, but not its `@local` — so a handle
whose own value dispatches on another handle (`data.size >> len!` over
`b >> data!`) printed as an anonymous `size. >>!` while its own
references still read `len`, a name the next parse resolves to nothing.
That is the `size. >>!` + stranded-reference symptom in `bytes/array.eo`,
`i64.eo` and `string/split.eo`.

Worth noting for whoever reads the issue: `eo:multi-referenced` was not
at fault. A reference written inside a nested formation reaches the
handle as `ξ.ρ.<name>`, which resolves to the same auto-name, so it was
counted all along. Five packs cover the shapes; eo-runtime's 153 sources
still print canonically and its suite is green. One neighbouring hole is
still open and predates this change: a moniker whose host site sits
inside another moniker strands, so `data.size >> len!` with two in-scope
uses of `len` prints `v2_2.size >> len!`. I will file that separately.

Closes #5914

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -174,20 +174,24 @@
   <!--
   Drop an inlined auto-named abstract; keep cactus-named voids, keep
   self-referential (recursive) abstracts, which are never inlined, keep a
-  multi-referenced dataized-const handle, which is never inlined (#5828), keep
-  a multi-referenced abstract formation, which is never inlined either (#5876),
-  keep a const that only a vertical layout can spell, which is never inlined
-  either (#5910), keep a formation applied through a `@pipe` continuation,
-  which is kept in place above its pipe rather than inlined (#5834), and keep a
-  binding that a surviving method dispatch still reaches through its name.
-  Such a dispatch reference (`ξ.<name>.<seg>`) is not inlined above — its
-  receiver is buried in a dotted base — so dropping the binding would strand
-  the reference on a synthetic "vL_P" placeholder. Keeping it lets
-  "merge-monikers" host the binding as the receiver of a reversed dispatch
-  instead (#5782).
+  binding no reference reaches at all (#5914), keep a multi-referenced
+  dataized-const handle, which is never inlined (#5828), keep a
+  multi-referenced abstract formation, which is never
+  inlined either (#5876), keep a const that only a vertical layout can spell,
+  which is never inlined either (#5910), keep a formation applied through a
+  `@pipe` continuation, which is kept in place above its pipe rather than
+  inlined (#5834), and keep a binding that a surviving method dispatch still
+  reaches through its name. Such a dispatch reference (`ξ.<name>.<seg>`) is not
+  inlined above — its receiver is buried in a dotted base — so dropping the
+  binding would strand the reference on a synthetic "vL_P" placeholder. Keeping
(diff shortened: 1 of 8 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (246 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2787](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5916, opened 2026-07-25
#5864: assert real correct values in bytes/left.eo and bytes/xor.eo tests

### LINKED ISSUE #5864: Incorrect test assertions in `bytes.eo`, `bytes/left.eo`, `bytes/xor.eo` may not catch certain regressions

morphqdd:
Second part of the split for #5864 (first part, `bytes.eo`, is PR #5904; split because the combined diff exceeded the 200-line PR-size CI cap).

`bytes/left.eo` and `bytes/xor.eo` contained the same `not. > @ : eq. (actual) (some-decoy-value)` test pattern described in #5864: it only proves the result is not one specific decoy value, instead of proving the result is the actually correct value.

This PR replaces the 9 such tests in these two files (5 in `bytes/left.eo`, 4 in `bytes/xor.eo`) with real `eq. > @` assertions against the actual correct value, expressed as raw hex byte literals since `.as-bytes` produces the 8-byte IEEE-754 double representation and `left`/`xor` operate bytewise (or as a logical bit-shift) on that representation.

Every value was computed both independently (Python reimplementation of the bitwise/shift algorithm) and by directly invoking the real `BytesOf`/`BytesRaw` runtime classes with the same inputs, and the two matched for all 9 cases before being written into the tests. Both files pass in full after the change (7 tests in `bytes/left.eo`, 9 in `bytes/xor.eo`).

Closes #5864

### DIFF
--- a/eo-runtime/src/main/eo/bytes/left.eo
+++ b/eo-runtime/src/main/eo/bytes/left.eo
@@ -17,11 +17,9 @@
     FF-FF-FF-FF-FF-FF-FF-EE.left 2
     00-00-00-00-00-00-00-47.not
 
-  ++> tests-left-with-even-plus
-    not. > @
-      eq.
-        4.as-bytes.left 3
-        -33.as-bytes
+  eq. ++> tests-left-with-even-plus
+    4.as-bytes.left 3
+    00-80-00-00-00-00-00-00
 
   eq. ++> tests-left-with-minus-one
     eq.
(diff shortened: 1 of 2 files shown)


morphqdd:
Merged back into #5904 (all 3 files, one PR). Closed this one and deleted the branch.

PR #5921, opened 2026-07-25
Keep a private handle that hosts another handle

### LINKED ISSUE #5918: `eo:format` incorrectly relocates `>>` handle inside single-referenced private const during fold operation
When `eo:format` folds a single-referenced private const onto its use site, a `>>` handle defined inside that const's value is relocated along with it and loses its name, so the folded copy references a dangling positional strand `vLINE_COL`.

```
[text] > split
  size. >> len!
    text >> bts!
  len.eq 0 > @
```

prints as:

```
[text] > split
  eq. > @
    v3_4.size >> len!
    0
```

`len` is `(text >> bts!).size`; folding `len` into `@` strands its nested `bts` handle as `v3_4`.

On branch `5678` this blocks `string/split.eo` (`size. >> len!` over `text >> bts!` prints `v47_4.size >> len!`, likewise for `size`/`delim`) and `i64.eo`'s carry cluster, where consts whose values nest `>>` handles — `and. >> both-negative!` over `... >> shifted-sign!` and `... >> divisor-sign!` — strand as `v55_6`/`v62_6` when folded into the chain.

This is adjacent to #5910, whose `eo:vertical-const` keeps a const whose value reaches a cactus reference; a const whose value defines (nests) a `>>` handle is not covered and still folds.

Smallest fix: extend the fold guard in `inline-cactoos.xsl` (around line 77 / `eo:vertical-const`) so a const whose value contains a nested `>>` handle definition is kept as its own named binding instead of folded onto its use site.

yegor256:
`merge-monikers` folds a file-local `>>` handle onto its use site, but it rebuilt the relocated binding attribute by attribute. That is fine until the binding is a host site itself: the parser hoists a handle written inside another handle's receiver out to its own binding (`text >> bts!` under `size. >> len!`) and leaves only a reference behind, so the outer handle's own `@base` is what the inner one folds onto. Rebuilt attribute by attribute, that fold never happened at the new site while the inner binding was dropped as merged — hence the dangling `v3_4.size >> len!`. The binding now lands through a `merged` mode that re-applies the hosting rule recursively, so a whole chain of handles travels to the use site nested inside one another. Four print-packs cover the plain shape, a three-deep chain, the `i64.eo` carry cluster and a host site read from two places.

Worth noting that the issue's own diagnosis points at the fold guard in `inline-cactoos`, which is not where this goes wrong — that pass already keeps the const (`eo:vertical-const`), and nothing is nested by the time the printer sees it. Saxon had been reporting the real problem all along as an "ambiguous rule match" between the merge and the drop template; the drop now outranks the merge explicitly, since what the binding hosts travels with it.

`string/split.eo`'s `len` and `size` and `i64.eo`'s `both-negative` can be privatised after this. I ran the format gate over branch `5678`'s sources (229 handles, 153 files): canonical before and after this change, and with those three privatised the gate settles in one `-Deo.autoFix` pass, with all 1760 runtime tests green on the result. On master's printer the same edits print `v47_4.size >> len!` and `v56_8.and divisor-sign >> both-negative!`. The non-const flavour (`size. >> len` over `text >> bts`) is spelled correctly now too, and mutually dispatching handles — source that never dataizes but does parse — end the recursion instead of driving it forever.

Closes #5918

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -109,11 +109,41 @@
   binding with both spellings still folds into the shorter bare inline and a
   dispatch hosts only when no bare reference is available.
   -->
-  <xsl:function name="eo:moniker-refs" as="element()*">
+  <xsl:function name="eo:hostable-refs" as="element()*">
     <xsl:param name="attr" as="element()"/>
     <xsl:variable name="owner" select="$attr/.."/>
     <xsl:variable name="refs" select="$owner//o[eo:resolved-ref(.) = $attr/@name and (ancestor::o[eo:abstract(.)][1] is $owner) and not(ancestor::o[. is $attr])]"/>
     <xsl:sequence select="($refs[eo:dispatch-seg(.) = ''], $refs[eo:dispatch-seg(.) != ''])"/>
+  </xsl:function>
+  <!--
+  Whether the binding `$attr` is a host site itself: a sibling handle folds onto
+  `$attr`'s own `@base`, the shape `bts.size &gt;&gt; len!` has over
+  `text &gt;&gt; bts!`, where the parser hoisted the inner handle out and left a
+  reference to it in the outer handle's receiver, which the merge below turns
+  into the reversed dispatch that carries it. A sibling that folds onto an
+  earlier reference of its own — the site the author wrote it at, as in
+  "starts-with.eo" — leaves `$attr`'s receiver reading it by name
+  (`eo:kept-const-ref`) and so does not make `$attr` a host site. The sibling's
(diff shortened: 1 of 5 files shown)


yegor256:
Pushed a second commit that changes the approach: the handle now travels to its use site with everything it hosts nested inside, instead of being pinned to its own line. The body above describes the final state.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (180 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2811](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5922, opened 2026-07-25
Reach a file-local handle through an explicit scope

### LINKED ISSUE #5917: Private const `>>` mishandled by `eo:format` in nested formation resulting in terminated computation
A private const file-local handle referenced both from its own scope and from inside a nested formation is still mishandled by `eo:format`: the handle is hosted onto its same-scope reference, and the formation-nested reference is left as a plain name whose definition now lives in a sibling subtree, so it resolves to nothing and dataizes to `the ⊥ object is a terminated computation and cannot be used`.

This is the sibling of #5893, which fixed the variant that strands the nested reference on `^.vLINE_COL`, and of #5914, which fixed the variant where every reference lives in a formation. The case that remains is one same-scope reference plus one formation reference, where the const is hosted onto another const's value and the formation reference keeps its plain (non-`^`) name.

```
[tup] > foo
  tup.as-bytes >> kbts!
  bytes.hash kbts >> hash!
  [ent] >> has-key
    ent.head.kbts.eq kbts > @
```

prints as:

```
[tup] > foo
  bytes.hash >> hash!
    tup.as-bytes >> kbts!
  ent.head.kbts.eq kbts > [ent] >> has-key
```

`kbts` is hosted under `hash`; `has-key`'s `ent.head.kbts.eq kbts` still names `kbts`, but its definition now sits inside `hash`, so the reference orphans.

On branch `5678` this blocks `map.eo`, where `tup.head.key.as-bytes >> kbts!` is used by `bytes.hash kbts` and by `ent.head.kbts.eq kbts` inside the `has-key` formation; `EOmapTest` and `EOsetTest` error with the ⊥ message.

Smallest fix: apply #5893's remedy to this shape too — when a const `>>` handle has any reference inside a nested formation, keep it as a standalone in-place handle and do not host it onto another reference, even when that reference is itself another const's value and the formation reference is a plain (non-`^`) name.

yegor256:
A `>> foo` handle could only be read by a bare `foo`. Spell the scope out
instead — `$.foo`, `^.foo`, one more `^` per level above that — and the
reference kept the readable name, which nothing owns once the handle takes its
cactus name, so it dataized to the terminated computation. `resolve-local-names`
now binds such a reference to the handle of exactly the formation it names, and
the printer keeps the `^` climb in front of a handle when the bare name is taken
in a scope in between, instead of always shedding it.

The issue blames the moniker fold in the printer, but that turned out to be
innocent: the nesting it prints (`bytes.hash >> hash!` over `key.as-bytes >>
kbts!`) reparses to the same graph, because the parser hoists the inner handle
back out to the enclosing formation. What actually broke `map.eo` is the two
scopes whose entries copy the key's bytes with `^.kbts > kbts` — the explicit
scope is the only spelling available there, since inside that formation a bare
`kbts` is the copy being defined. Privatising the other two scopes worked before
this change; all four work now, and `EOmapTest`/`EOsetTest` pass.

Verified with 1506 parser, 269 printer and 1760 runtime tests, plus the format
gate over the 229 privatised handles on branch `5678` — canonical before and
after, so nothing reformats. One thing the reviewer may want to know: a handle
whose *only* reference is a named one (`kbts > alias`) is inlined by
`inline-cactoos` with the name dropped, which prints EO that no longer parses.
That is not new — master does it for a bare `kbts > alias` too — and this change
makes the `^.`-spelled variant reach it, so I will file it separately.

Closes #5917

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl
@@ -24,6 +24,15 @@
   hijacks a same-named public attribute in an unrelated sibling formation.
   This mirrors how "build-fqns" later resolves a name to the nearest
   enclosing formation that owns an attribute of that name.
+
+  A reference may also spell its scope out instead of leaving it to the
+  search: "$.foo" reaches the innermost formation and "^.foo" its parent
+  (§3.14), and each binds to the handle of exactly the formation it names
+  (see `eo:scope`, #5917). Such a reference is the only spelling left when
+  the nested formation rebinds the name it reads - `^.foo &gt; foo` copies
+  the parent's handle into an attribute of the same name, where a bare "foo"
+  would bind to the copy - so leaving it unresolved stranded the handle
+  behind a name nothing owned, and the reference dataized to ⊥.
   -->
   <xsl:output encoding="UTF-8" method="xml"/>
   <!--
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (127 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2779](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5873, opened 2026-07-24
Concurrent writing into `Saved` causes partial read of XMIR file


anuragpaul602-netizen:
`Saved` truncates its target and streams into it, so a file that two `Threaded` workers write at the same time can be read back half-written by a third. `ConcurrentCache` (#5720) doesn't cover this: it serializes the `Cache.apply` call, but every caller reads the target file *after* the guard has been released.

The three call sites all have this shape — guarded write, unguarded read:

- `Parsing.parsed`: `Parsing.tojoVersion(target, refs)` at [`Parsing.java:213`](https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java#L213), which does `new XMLDocument(target)` when `refs` is empty (i.e. on a cache hit), and `new Xnav(target)` at [`Parsing.java:214`](https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java#L214), which runs on every path
- `Transpiling.transpiled` → `javaGenerated` → `new Xnav(target).element("object")` at [`Transpiling.java:416`](https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java#L416)
- `Linting.lintOne`: `new Xnav(target)` at [`Linting.java:314`](https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java#L314)

The write itself is [`Saved.java:85-92`](https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java#L85-L92) — a `TeeInput` into `new OutputTo(target)`, which opens the file truncating and streams the bytes in. There is a window, proportional to the size of the XMIR, in which the file on disk is a valid path to an incomplete document.

### How it shows up

```
java.lang.IllegalArgumentException: Can't parse by com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderImpl,
    most probably the XML is invalid
  at com.jcabi.xml.XMLDocument.<init>(XMLDocument.java:178)
  at org.eolang.maven.Parsing.tojoVersion(Parsing.java:291)
  at org.eolang.maven.Parsing.parsed(Parsing.java:225)
Caused by: org.xml.sax.SAXParseException; systemId: file:.../target/1-parse/main.xmir;
[...]

PR #5923, opened 2026-07-25
test(#5720): race tests for the Transpiling and Linting cache guards

anuragpaul602-netizen:
Follow-up to #5766. **It is stacked on that PR's branch**, so the diff here currently also shows #5766's fix commits — GitHub cannot base a cross-fork PR on a branch that only exists in my fork. Once #5766 merges, this diff collapses on its own to just the two new test files and the `Linting` signature change. Please review #5766 first.

@morphqdd asked for the `ParsingTest` race test to be repeated for the other two classes #5766 touches: the same wiring mistake was made three times, and only one of the three had a regression test. Splitting it out keeps #5766 inside its line budget — adding these there would have taken it from ~194 to ~377 changed lines.

## What the tests do

`TranspilingTest.transpilesOnlyOnceWhenThreadsRaceOnTheSameTail` races 8 threads of one `Transpiling` on a single tail path and counts XSL train runs through `StMeasured`, which already appends one line per shift it executes — so a second pass over `to-java.xsl` shows up as a second line. Only `transpiled` had to widen to package-private, as `Parsing.parsed` already is.

`LintingTest.lintsOnlyOnceWhenThreadsRaceOnTheSameTail` does the same for `Linting`.

Both were verified to **fail against a per-call guard (8 vs 1)** and pass against the fix, so reverting any of the three guards now fails a test.

## Why `Linting` needed a signature change

The expectation was that the `ParsingTest` structure would translate directly. It did not, for one reason worth calling out: `Parsing.parsed` already took its compile function as a parameter, which is why counting was free there. `Transpiling` and `Linting` build theirs internally.

`Transpiling` had a usable existing observable (the measures file), so it needed no signature change. `Linting` had none, so `lintOne` now takes a `UnaryOperator<XML>` — the same shape `Parsing.parsed` has — with `linting()` passing `xmir -> this.linted(xmir, unlints)`. That also hoists the `unlints` array out of the per-tojo lambda, where it was being reallocated for every file.

Two knock-ons, both flagged rather than hidden:

- `linting()` widens to package-private so `lintOne` stays ahead of the private methods for `MethodsOrderCheck`. Same trick #5766 used to avoid moving `Parsing.parsed` (~65 lines of pure movement in the diff).
- The class picks up `@checkstyle ClassFanOutComplexityCheck`, which `Transpiling` already carries, because `UnaryOperator` puts `Linting` at 31 against a ceiling of 30. I first tried to offset it honestly by swapping `Paths.get` for `Path.of`; that did not move the count, so I reverted the churn rather than leave an unrelated edit in.

## One trap worth recording

The first version of the lint test **passed against the broken code**. With a plain counting pipeline the fake lint returned so fast that the first thread wrote the cache before the other seven checked it, so they hit the cache even with no shared lock, and the count was 1 either way. Only the fails-before check caught it. It now uses a `CountDownLatch`: a shared guard admits one thread which waits out the timeout alone, a per-file guard admits all eight and the latch opens immediately. Deterministic instead of timing-dependent.

As in `ParsingTest`, the post-guard read of the target can still catch `Saved` mid-write; that is #5873, a different race, tolerated here the same way.

440 tests pass and qulice is clean on this branch.


eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
```diff
@@ -176,13 +182,14 @@ private int parsed(
...
      * Parse EO file to XML.
+     * Package-private, so that a test can race threads on one instance.
      * @param tojo The tojo
      * @param pipeline The canonical parsing transform to apply
      * @param digest Digest of the set of known objects (part of the cache key)
      * @return Amount of parsed tojos
      * @throws Exception If fails
      */
-    private int parsed(
+    int parsed(
```
yegor256, line 192: @anuragpaul602-netizen this is a wrong move. We should NOT make objects more suitable for tests by exposing their internals.

anuragpaul602-netizen:
Closing this as obsolete.

@yegor256 pointed out on #5766 that widening `Parsing.parsed` so a test could reach it was the wrong move — objects shouldn't expose internals to be testable. This PR did the same thing twice more, and worse: `Transpiling.transpiled` widened to package-private, and `Linting.lintOne` given a `UnaryOperator<XML>` parameter purely so a test could count invocations.

The underlying mistake was structural all along — the cache guard was built per call instead of held as a field — so #5766 now checks it structurally with an ArchUnit rule:

```java
ArchRuleDefinition.noClasses()
    .that().haveSimpleNameNotEndingWith("Test")
    .should().callConstructorWhere(ArchitectureTest.guardBuiltPerCall())
```

`ConcurrentCache` may only be constructed from a constructor. That single rule covers `Parsing`, `Transpiling`, `Linting` and any call site added later — strictly more than the two race tests here, which is what @morphqdd originally asked for — and nothing has to be exposed. The lock's own behaviour stays covered by `ConcurrentCacheTest`.

Nothing here is worth salvaging on top of that, so it's closed rather than reworked.

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2738](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5924, opened 2026-07-25
#5873: write atomically via a sibling temp file plus atomic rename

### LINKED ISSUE #5873: Concurrent writing into `Saved` causes partial read of XMIR file

morphqdd:
Fixes #5873.

`Saved` streamed content directly into the target file with `TeeInput`/`OutputTo`, which truncates the target and writes into it in place. A reader (another `Saved` writing to the same tail path, or anything reading the file back, e.g. `Parsing.tojoVersion`/`Xnav` on a cache hit) could observe the file mid-write: a valid path pointing at an incomplete document, surfacing as a misleading "Premature end of file" XML parse error with nothing actually wrong with the content once the write finished.

Fix: stream into a temp file created alongside the target (same filesystem, so `Files.move` can be atomic), then move it onto the target with `ATOMIC_MOVE`. A reader now always sees either the previous complete file or the new one, never a partial one. This matches option 1 from the issue's own suggested fixes.

Added `SavedTest` with a regression test: 8 threads write different large variants to the same target concurrently while each also reads the file back immediately after its own write, asserting every read is one of the complete variants and never a truncated or mixed one. Verified the test catches the bug: reverting `Saved` to the old direct-write implementation makes it fail with truncated reads observed; with the fix it passes.

Verified: `SavedTest` 2/2 green, qulice clean.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
@@ -7,7 +7,9 @@ package org.eolang.maven;
 import com.jcabi.log.Logger;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import org.cactoos.Input;
 import org.cactoos.Scalar;
 import org.cactoos.Text;
@@ -20,6 +22,14 @@ import org.cactoos.scalar.LengthOf;
 /**
  * Content saved to the file.
  * Returns path to the file
+ *
+ * <p>The content is streamed into a sibling temporary file first, then moved
+ * onto {@link #target} with {@link StandardCopyOption#ATOMIC_MOVE}. Streaming
+ * straight into {@link #target} would let a concurrent reader (or another
+ * writer racing on the same path) observe it truncated, mid-write (#5873):
+ * an atomic rename means a reader always sees either the previous complete
+ * file or the new one, never a partial one.</p>
+ *
  * @since 0.41.0
  */
 final class Saved implements Scalar<Path> {
@@ -82,14 +92,27 @@ final class Saved implements Scalar<Path> {
                     this.target.getParent()
                 );
             }
-            bytes = new IoChecked<>(
-                new LengthOf(
-                    new TeeInput(
-                        this.content,
-                        new OutputTo(this.target)
+            final Path tmp = Files.createTempFile(
+                this.target.getParent(),
+                this.target.getFileName().toString(),
+                ".tmp"
+            );
+            try {
+                bytes = new IoChecked<>(
+                    new LengthOf(
+                        new TeeInput(
+                            this.content,
+                            new OutputTo(tmp)
+                        )
                     )
-                )
-            ).value();
+                ).value();
+                Files.move(
+                    tmp, this.target,
+                    StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING
+                );
+            } finally {
+                Files.deleteIfExists(tmp);
+            }
             Logger.debug(
                 this, "File %s saved (%d bytes)",
                 this.target, bytes
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
@@ -7,6 +7,7 @@ package org.eolang.maven;
 import com.jcabi.log.Logger;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
@@ -106,10 +107,7 @@ final class Saved implements Scalar<Path> {
                         )
                     )
                 ).value();
-                Files.move(
-                    tmp, this.target,
-                    StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING
-                );
+                Saved.moved(tmp, this.target);
             } finally {
                 Files.deleteIfExists(tmp);
             }
@@ -127,5 +125,24 @@ final class Saved implements Scalar<Path> {
             );
         }
         return this.target;
+    }
+
+    /**
+     * Move a temp file onto its target, atomically where the filesystem
+     * supports it, falling back to a plain (still complete-file, just not
+     * guaranteed atomic) move on filesystems that don't.
+     * @param tmp Temp file to move
+     * @param target Destination path
+     * @throws IOException If the move fails
+     */
+    private static void moved(final Path tmp, final Path target) throws IOException {
+        try {
+            Files.move(
+                tmp, target,
+                StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING
+            );
+        } catch (final AtomicMoveNotSupportedException ignored) {
+            Files.move(tmp, target, StandardCopyOption.REPLACE_EXISTING);
+        }
     }
 }
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/SavedTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/SavedTest.java
@@ -38,7 +38,7 @@ final class SavedTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


eo-maven-plugin/src/test/java/org/eolang/maven/SavedTest.java
```diff
@@ -0,0 +1,67 @@
...
+        new Saved("hello", target).value();
+        MatcherAssert.assertThat(
+            "the saved file must contain exactly what was written",
+            Files.readString(target, StandardCharsets.UTF_8),
+            Matchers.equalTo("hello")
+        );
+    }
+
+    @Test
+    void neverExposesAPartiallyWrittenFileToAConcurrentReader(
```
gemshrine: This name is what turns `qulice` red (the CI blocker on this PR). `jtcop`'s `RulePresentTense` requires a test method to start with a present-tense verb, and this one starts with the adverb `never`:

```
Failed to execute goal ...jtcop-maven-plugin:1.4.3:check: Test name
'neverExposesAPartiallyWrittenFileToAConcurrentReader' doesn't follow naming
rules, because the test name has to be written using present tense (RulePresentTense)
```

Rename it to lead with a present-tense verb (as the sibling `savesContentToFile` does), keeping the same meaning — e.g.:

```java
void exposesNoPartiallyWrittenFileToConcurrentReader(
```

or `hidesInProgressWritesFromConcurrentReaders`. That clears the gate; nothing else in the test needs to change.

eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
```diff
@@ -82,14 +92,27 @@ public Path value() throws IOException {
...
+                            this.content,
+                            new OutputTo(tmp)
+                        )
                     )
-                )
-            ).value();
+                ).value();
+                Files.move(
+                    tmp, this.target,
+                    StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING
```
gemshrine: Robustness note (not a blocker): `Files.move(..., ATOMIC_MOVE)` throws `AtomicMoveNotSupportedException` (an `IOException`) on filesystems that can't rename atomically — some network mounts and container overlay/bind filesystems being the usual suspects. The code this replaces streamed straight into the target and worked on those filesystems, so a build that saves into such a mount would now fail with `Failed while trying to save ...` where it previously succeeded.

Putting the temp file in `target.getParent()` already rules out the cross-filesystem case, so this only bites filesystems that lack atomic-rename semantics entirely — but for those, a fallback keeps the old reach while staying complete-file (the reader still never sees a truncated write, it just isn't guaranteed atomic on that exotic FS):

```java
try {
    Files.move(tmp, this.target, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
} catch (final AtomicMoveNotSupportedException ignored) {
    Files.move(tmp, this.target, StandardCopyOption.REPLACE_EXISTING);
}
```

Worth a look given `Saved` is on the hot path of every mojo; feel free to skip if the target is always guaranteed to be a local build dir.

gemshrine (changes requested):
Solid approach and a careful implementation — the temp file is created with `Files.createTempFile` (unique name, so two writers racing on the same target don't collide on the temp either), it's a sibling in `target.getParent()` (same filesystem, so `ATOMIC_MOVE` won't hit `AtomicMoveNotSupportedException` for the cross-FS reason), and the `finally { deleteIfExists }` cleans up on a failed write. The new concurrency test genuinely exercises the race (8 workers, 500 KB variants, asserting a reader never sees a mix).

One blocker and one optional note, both inline:
- the concurrency test's name fails `jtcop`'s present-tense rule, which is what's turning `qulice` red — a one-word rename fixes it;
- `ATOMIC_MOVE` can still throw on filesystems that lack atomic rename outright (some network/overlay mounts) where the old direct write worked — a small fallback preserves that reach.

The core fix is correct; addressing the test name will get CI green.

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (121 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2144](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+676](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5925, opened 2026-07-25
#5920: reject HEX literals that lose precision past the double range

### LINKED ISSUE #5920: Hexadecimal literals unexpectedly lose precision unlike their integer equivalents
## What happens

`Emissions.openValue` (`eo-parser/src/main/java/org/eolang/parser/Emissions.java`) rejects INTEGER/FLOAT literals that cannot be represented exactly as an IEEE-754 double, but the HEX branch of the same method has no equivalent check:

```java
} else if (value.kind() == Value.Kind.HEX) {
    final long hex;
    try {
        hex = Long.parseLong(value.raw().substring(2), 16);
    } catch (final NumberFormatException ex) {
        final ParseError error = new ParseError(
            line, value.pos(),
            "hexadecimal literal is out of range"
        );
        error.initCause(ex);
        throw error;
    }
    emit.object(name, "Φ.number", line, value.pos());
    Emissions.bytesCarrier(
        emit, line, value.pos(),
        new Hex((double) hex).asString()
    );
}
```

`(double) hex` is an unguarded cast. It only fails (via the earlier `Long.parseLong`) when the hex digits do not fit in a signed 64-bit `long` at all. It never checks whether the `long` value round-trips exactly through `double`, which stops being possible once the magnitude exceeds 2^53.

Issue #5613 (closed, merged) added exactly this round-trip check for INTEGER and FLOAT literals in the sibling `number()` method a few lines below, because silently rounding a literal to the nearest representable double instead of failing was considered a correctness bug: an INTEGER literal like `9007199254740993` (2^53+1) is now rejected with `"9007199254740993 is over-precise, write 9007199254740992 instead"` (see `LnApplicationTest.rejectsOverPreciseIntegerHead`).

The HEX branch was apparently assumed to already be safe, since it does reject values outside the signed 64-bit range, but that only guards the range boundary, not the precision window between 2^53 and 2^63 where a value still fits in a `long` but can no longer be represented exactly as a `double`.

Verified directly:

```
jshell> long hex = Long.parseLong("20000000000001", 16);
hex ==> 9007199254740993
jshell> double d = (double) hex;
d ==> 9.007199254740992E15
jshell> long back = (long) d;
back ==> 9007199254740992
```

`0x20000000000001` is the hex spelling of `9007199254740993`, the exact same value #5613 already rejects when written as an INTEGER literal. Written as HEX, it parses with no error and silently emits the bytes for `9007199254740992.0`, one less than what the literal actually wrote, with no diagnostic at all. The two literal kinds that produce the identical `Φ.number` XMIR representation disagree on whether the program is even valid.

## What should happen

The HEX branch should apply the same round-trip check used for INTEGER/FLOAT literals (reject when `(long) (double) hex != hex`) and raise a `ParseError` with a message in the same style as `number()`'s, naming the offending hex spelling and the exact value it actually parsed to, instead of silently truncating.

morphqdd:
Fixes #5920.

`Emissions.openValue`'s HEX branch cast the parsed `long` straight to `double` with no round-trip check, unlike the sibling `number()` method (added by #5613) which rejects INTEGER/FLOAT literals that cannot be represented exactly as an IEEE-754 double. A HEX literal like `0x20000000000001` (`9007199254740993`, one past 2^53) parsed with no error and silently emitted the bytes for `9007199254740992.0` instead, one less than what was actually written, with no diagnostic. The same value written as an INTEGER literal is already rejected by `number()`.

Fix: extracted the HEX branch into its own `hex()` method (matching the existing `number()`/`group()` pattern in the class, also needed to keep `openValue`'s NCSS under the qulice limit after adding the check), added the same round-trip check used implicitly by `number()`'s `overPrecise` check (`(long) parsed != raw`), and raise a `ParseError` in the same `"%s is over-precise, write %s instead"` style, naming the offending hex spelling and the decimal value it actually parsed to.

Added `rejectsOverPreciseHexHead` to `LnApplicationTest`. Verified it fails against the pre-fix `Emissions.java` (asserted the old unguarded cast) and passes with the fix.

Verified: eo-parser full suite 1506/1506 green (including all 926 `EoSyntaxTest` golden cases), qulice clean.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -118,22 +118,7 @@ final class Emissions {
         if (value.kind() == Value.Kind.INTEGER || value.kind() == Value.Kind.FLOAT) {
             Emissions.number(emit, name, value, line);
         } else if (value.kind() == Value.Kind.HEX) {
-            final long hex;
-            try {
-                hex = Long.parseLong(value.raw().substring(2), 16);
-            } catch (final NumberFormatException ex) {
-                final ParseError error = new ParseError(
-                    line, value.pos(),
-                    "hexadecimal literal is out of range"
-                );
-                error.initCause(ex);
-                throw error;
-            }
-            emit.object(name, "Φ.number", line, value.pos());
-            Emissions.bytesCarrier(
-                emit, line, value.pos(),
-                new Hex((double) hex).asString()
-            );
+            Emissions.hex(emit, name, value, line);
         } else if (value.kind() == Value.Kind.BYTES) {
             emit.object(name, "Φ.bytes", line, value.pos());
             emit.object(null, null, line, value.pos());
@@ -296,6 +281,48 @@ final class Emissions {
             }
         }
         return out.toString();
+    }
+
+    /**
+     * Emit a {@code HEX} literal as {@code Φ.number}, rejecting values that
+     * fit a signed 64-bit {@code long} but no longer round-trip exactly
+     * through an IEEE-754 double (the sibling {@code number()} check, for
+     * the base where a plain range check on {@code long} does not catch it).
+     * @param emit Emitter
+     * @param name Name attribute (or {@code null})
+     * @param value Hex value
+     * @param line Source line
+     * @checkstyle ParameterNumberCheck (3 lines)
+     */
+    private static void hex(
+        final Emit emit, final String name, final Value value, final int line
+    ) {
+        final long raw;
+        try {
+            raw = Long.parseLong(value.raw().substring(2), 16);
+        } catch (final NumberFormatException ex) {
+            final ParseError error = new ParseError(
+                line, value.pos(),
+                "hexadecimal literal is out of range"
+            );
+            error.initCause(ex);
+            throw error;
+        }
+        final double parsed = raw;
+        if ((long) parsed != raw) {
+            throw new ParseError(
+                line, value.pos(),
+                String.format(
+                    "%s is over-precise, write %s instead",
+                    value.raw(), Emissions.canonicalInteger(parsed)
+                )
+            );
+        }
+        emit.object(name, "Φ.number", line, value.pos());
+        Emissions.bytesCarrier(
+            emit, line, value.pos(),
+            new Hex(parsed).asString()
+        );
     }
 
     /**
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -293,7 +293,7 @@ final class Emissions {
      * @param name Name attribute (or {@code null})
      * @param value Hex value
      * @param line Source line
-     * @checkstyle ParameterNumberCheck (3 lines)
+     * @checkstyle ParameterNumberCheck (6 lines)
      */
     @SuppressWarnings({
         "PMD.AvoidDecimalLiteralsInBigDecimalConstructor",
--- a/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
@@ -604,8 +604,7 @@ final class LnApplicationTest {
     @Test
     void rejectsOverPreciseHexHead() {
         MatcherAssert.assertThat(
-            "a HEX literal past the exact double integer range must suggest the rounded"
-                .concat(" decimal spelling, matching the INTEGER literal of the same value"),
+            "a HEX literal past the exact double integer range must suggest the rounded decimal spelling, matching the INTEGER literal of the same value",
             Assertions.assertThrows(
                 ParseError.class,
                 () -> LnApplicationTest.parseLine("0x20000000000001 > x")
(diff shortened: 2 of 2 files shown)


eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
```diff
@@ -601,6 +601,21 @@ void rejectsOversizedHexHead() {
         );
     }
 
+    @Test
+    void rejectsOverPreciseHexHead() {
+        MatcherAssert.assertThat(
+            "a HEX literal past the exact double integer range must suggest the rounded"
+                .concat(" decimal spelling, matching the INTEGER literal of the same value"),
```
yegor256: @morphqdd you can just put it all in a single line, qulice won't complain

eo-parser/src/main/java/org/eolang/parser/Emissions.java
```diff
@@ -298,6 +283,48 @@ static String unescapeBody(final String inner) {
...
+        } catch (final NumberFormatException ex) {
+            final ParseError error = new ParseError(
+                line, value.pos(),
+                "hexadecimal literal is out of range"
+            );
+            error.initCause(ex);
+            throw error;
+        }
+        final double parsed = raw;
+        if ((long) parsed != raw) {
```
gemshrine: This round-trip guard has a false-negative at the very top of the range: `0x7FFFFFFFFFFFFFFF` (`Long.MAX_VALUE`, 2^63−1) slips through and is silently emitted as `9223372036854775808.0` (2^63) — one *more* than what was written — which is precisely the HEX-vs-INTEGER disagreement #5920 sets out to remove, just relocated to the boundary.

Why the check misses it: `(double) (2^63−1)` rounds up to `2^63`, and casting that double back to `long` **saturates** to `Long.MAX_VALUE` (JLS 5.1.3), so `(long) parsed == raw` and nothing is rejected. Meanwhile the sibling `number()` rejects the identical decimal literal, because it compares exact values via `BigDecimal` (`overPrecise`), which has no saturation blind spot. Verified locally:

```
raw            = 9223372036854775807      // Long.parseLong("7FFFFFFFFFFFFFFF",16)
(double) raw   = 9.223372036854776E18     // == 2^63
(long) parsed  = 9223372036854775807      // saturated back to Long.MAX_VALUE
(long)parsed != raw  -> false             // PR check: NOT rejected
BigDecimal compare   -> true              // would reject
INTEGER 9223372036854775807 via number()  -> rejected
```

The gap is exactly one value (the neighbour `2^63−2` is still caught, because it saturates to `Long.MAX_VALUE ≠ raw`).

Mirror `number()`'s exact comparison so HEX and INTEGER agree everywhere:

```java
if (new BigDecimal(raw).compareTo(new BigDecimal(parsed)) != 0) {
```

(this needs the same `@SuppressWarnings("java:S2111")` / `PMD.AvoidDecimalLiteralsInBigDecimalConstructor` that `overPrecise` already carries — or factor a shared `overPrecise(long, double)` helper). A lighter, allocation-free alternative that closes the same gap: `if (parsed >= 0x1p63 || (long) parsed != raw)` — since no `long` is exactly `2^63`, any `parsed` that reached `2^63` already lost precision.

eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
```diff
@@ -601,6 +601,21 @@ void rejectsOversizedHexHead() {
...
+            Assertions.assertThrows(
+                ParseError.class,
+                () -> LnApplicationTest.parseLine("0x20000000000001 > x")
```
gemshrine, line 610: This is the only new HEX rejection case, and `0x20000000000001` (2^53+1) is comfortably inside the range where the `(long) parsed != raw` check works — so it passes regardless of the boundary bug flagged in `Emissions.hex`. The value that actually exercises the weak spot is `0x7FFFFFFFFFFFFFFF` (`Long.MAX_VALUE`), which currently slips through silently. Adding it here would have caught the gap and would lock the fix in place:

```java
@Test
void rejectsMaxLongHexHead() {
    MatcherAssert.assertThat(
        "0x7FFFFFFFFFFFFFFF must be rejected like the INTEGER literal of the same value,"
            .concat(" not silently rounded up to 2^63"),
        Assertions.assertThrows(
            ParseError.class,
            () -> LnApplicationTest.parseLine("0x7FFFFFFFFFFFFFFF > x")
        ).getMessage(),
        Matchers.startsWith("0x7FFFFFFFFFFFFFFF is over-precise")
    );
}
```

(Against the current `hex()` this test fails — no `ParseError` is thrown — which is exactly the point.)

gemshrine (changes requested):
Thanks for closing the gap — extracting `hex()` and rejecting over-precise HEX literals is the right fix, and the error message/spelling now lines up with the INTEGER path. It works for the general case (e.g. `0x20000000000001`).

One boundary value still leaks, and it's the exact inconsistency #5920 is about: `0x7FFFFFFFFFFFFFFF` (`Long.MAX_VALUE`) passes the `(long) parsed != raw` guard because `(double) (2^63−1)` rounds to `2^63` and the cast back to `long` saturates to `Long.MAX_VALUE` — so it's silently emitted as `2^63`, while the identical INTEGER literal is rejected by `number()`. Using the same `BigDecimal` exact comparison `number()` already uses removes the blind spot. Details + a drop-in fix and a boundary test are inline.

yegor256:
@morphqdd conflicts here

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2160](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2759](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+669](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5926, opened 2026-07-26
#5869: make empty string character predicates consistently false

### LINKED ISSUE #5869: Inconsistent empty string handling in `is-alpha`, `is-alphanumeric`, and `is-ascii` functions
## What happens

Three sibling predicates in `eo-runtime/src/main/eo/string/` all claim the same contract shape ("check that all signs in `text` are X"), but disagree on what an empty string means:

`is-alpha.eo`:
```
[text] > is-alpha
  matches. > @
    regex "/^[a-zA-Z]+$/"
    text
```

`is-alphanumeric.eo`:
```
[text] > is-alphanumeric
  matches. > @
    regex "/^[A-Za-z0-9]+$/"
    text
```

`is-ascii.eo`:
```
[text] > is-ascii
  matches. > @
    regex "/^[\\x00-\\x7F]*$/"
    text
```

`is-alpha` and `is-alphanumeric` use `+` in their regex (at least 1 matching character required), so `is-alpha("")` and `is-alphanumeric("")` are `false`. `is-ascii` uses `*` (zero or more), so `is-ascii("")` is `true`.

None of the three files has a test for the empty-string case, so nothing currently catches this divergence, but the three predicates disagree on the exact same edge case despite identical "check that all signs are X" framing in their docstrings.

## What should happen

Pick one convention for what an empty string means for an "all characters satisfy X" predicate (the more common convention is vacuous truth: an empty set of characters trivially satisfies "all of them are X"), and apply it consistently across `is-alpha`, `is-alphanumeric`, and `is-ascii`, with a test for the empty-string case added to each.

Solaris-star:
Closes #5869.

The sibling character-class predicates already follow a non-empty-match convention: `is-alpha`, `is-alphanumeric`, and `is-digit` use `+`, and `is-digit` explicitly tests that an empty string is false. `is-ascii` was the outlier because its regex used `*`.

This change:

- changes `is-ascii` from zero-or-more to one-or-more ASCII characters;
- adds explicit empty-string tests to `is-alpha`, `is-alphanumeric`, and `is-ascii` so the shared contract cannot drift again.

Validation:

- the new tests initially produced one failure in `EOis_asciiTest` while the other runtime tests passed (`1763` run, one new failure, `19` skipped);
- after the fix, the three focused generated test classes pass: `12` tests, zero failures;
- all `153` EO runtime sources pass the canonical formatter;
- the `eo-runtime` compile phase succeeds.

### DIFF
--- a/eo-runtime/src/main/eo/string/is-alpha.eo
+++ b/eo-runtime/src/main/eo/string/is-alpha.eo
@@ -13,6 +13,8 @@
     regex "/^[a-zA-Z]+$/"
     text
 
+  (is-alpha "").not ++> tests-checks-if-empty-text-is-not-alpha
+
   is-alpha "eEo" ++> tests-checks-if-simple-text-is-alpha
 
   (is-alpha "-w-").not ++> tests-checks-if-text-with-dashes-is-not-alpha
--- a/eo-runtime/src/main/eo/string/is-alphanumeric.eo
+++ b/eo-runtime/src/main/eo/string/is-alphanumeric.eo
@@ -13,6 +13,8 @@
     regex "/^[A-Za-z0-9]+$/"
     text
 
+  (is-alphanumeric "").not ++> tests-checks-if-empty-text-is-not-alphanumeric
+
   is-alphanumeric "eEo" ++> tests-checks-if-simple-text-is-alphanumeric
 
   (is-alphanumeric "-w-").not ++> tests-checks-if-text-with-dashes-is-not-alphanumeric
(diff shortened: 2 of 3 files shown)


0crat:
@solaris-star Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5929, opened 2026-07-26
Give an anonymous const argument a spelling of its own

### LINKED ISSUE #5927: eo-printer scatters a shared private `>>` handle when two sibling consts fold, breaking the cross-reference
When two sibling private consts both fold onto their use sites and one of them references a `>>` handle that is defined inside the other's value, `eo:format` scatters the two consts into separate subtrees. The shared handle travels with the const whose value defines it, so the other const's file-local reference to that handle no longer resolves. The result parses, is idempotent, and leaves no `vLINE_COL` strand — so only dataization catches it, as `the ⊥ object ...`.

```
[x] > foo
  and. >> both!
    x.plus 1 >> p!
    x.plus 2 >> q!
  p.xor q >> diff!
  both.or diff > @
```

prints as (idempotent, no strand):

```
[x] > foo
  or. > @
    and. >> both!
      x.plus 1 >> p!
      x.plus 2 >> q!
    p.xor q >> diff!
```

`p` and `q` are now buried inside `both`, but the sibling `diff` (`p.xor q`) still names them; at dataization the references resolve to nothing.

On branch `5678` this blocks `i64.eo`. Privatizing the division carry cluster (`and. >> both-negative!` over `... >> shifted-sign!` / `... >> divisor-sign!`, plus `signs-differ = shifted-sign.xor divisor-sign`, `mixed-signs-with-nonneg-trial`, `carry-out`, `trial-nonnegative`) folds `shifted-sign`/`divisor-sign` inside `both-negative` while `signs-differ` lands in a sibling subtree; `i64.div` and `i64.plus` then dataize to `the ⊥ object only accepts a cause at position 0, not 1`, failing `EOi64Test` and every fixed-width integer test (`EOu8Test`, `EOi16Test`, ...). This is acyclic — the cluster dataizes correctly while public — and so is distinct from the degenerate mutual cycle noted in #5921.

Smallest fix: when a private const's value defines a `>>` handle that a sibling binding references, keep both consts as standalone named siblings rather than folding them into separate subtrees, so the shared handle stays in a scope both references can reach.

yegor256:
A single-referenced private const folds onto its use site as the anonymous `42.plus a!` argument, and only a horizontal argument slot can hold that `!`. When a sibling argument carries a name of its own the application cannot go on one line, so the const lands on a line of its own — where the marker has nowhere to sit. Behind the head it closes the expression and the arguments below become trailing garbage; behind the arguments it marks the last one instead of the whole application. The first shape silently lost an argument on the next parse, the second quietly moved the const marker. Since a name suffix *may* carry the marker, such an argument now takes the auto-name `>>!`, and the parser floats the binding back up to the enclosing formation, which is the very graph the inline argument builds.

A note on the diagnosis in the issue: the moniker fold it blames is innocent, and the minimal reproduction it gives does not reproduce anything — I dumped the XMIR of the origin against the reparsed print and the two graphs are the same, because the parser hoists a handle written inside another handle's value straight back out. The carry cluster on its own is fine too: privatising just those seven names in `i64.eo` passes all 51 `EOi64Test` cases on today's master. What actually blocks the file is `next-quot`, whose folded `quot.or (bit idx)` sits beside the `>> next-rem!` moniker and so ends up on a line of its own as `quot.or!`, dropping `bit idx` and leaving `looped` with two arguments instead of three.

The existing `const-hosting-two-handles` pack had the wrong spelling baked into its expected output — `bottom.and 80-00-00-00-00-00-00-00!`, where the marker binds to the literal rather than to the application — so it is updated here too. Checked: 271 printer tests, 462 plugin tests, and the format gate over master's 153 sources with no churn; the same gate over branch 5678's 155 sources produces output byte-identical to the pristine printer's. With all eleven `looped` helpers in `i64.eo` privatised on top of this, one `-Deo.autoFix` pass settles and all 1760 runtime tests pass, which clears the last blocker for #5678.

Closes #5927

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Node.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Node.java
@@ -138,6 +138,35 @@ final class Node {
     }
 
     /**
+     * The same node as it must be spelled on a line of its own.
+     *
+     * <p>An anonymous inline const argument (#5821) carries a bare {@code !}
+     * marker, and only a horizontal argument slot can hold it: on a line of
+     * its own the marker closes the expression, so the arguments below it
+     * become trailing garbage ({@code or.!}), and behind those arguments it
+     * marks the last one instead of the whole application (#5902). A name
+     * suffix, however, may carry the marker (R-3.10.4), so such a const takes
+     * the auto-name {@code >>} that the same slot already spells for a
+     * formation. The parser floats an auto-named binding up to the enclosing
+     * formation and leaves a reference in its place, which is the very graph
+     * the inline argument builds, so the two spellings agree (#5927).</p>
+     *
+     * @return The node with a suffix its own line can carry
+     */
+    Node lined() {
+        final Node result;
+        if ("!".equals(this.tail)) {
+            result = new Node(
+                this.base, " >>!", this.abstractt, this.test,
+                this.reversed, this.data, this.children
+            );
+        } else {
+            result = this;
+        }
+        return result;
+    }
+
+    /**
      * Whether this node carries no name suffix anywhere in its subtree.
      *
      * <p>A line is "named" when its {@code tail} holds a {@code > name},
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -245,6 +245,15 @@ final class Pretty {
     /**
      * Render a node vertically: the head on this line, each argument
      * laid out on the lines below, indented one level deeper.
+     *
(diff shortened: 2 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2783](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5931, opened 2026-07-26
Rank package extension above the decoratee in take()

### LINKED ISSUE #5930: Package extension ranks below the decoratee in `PhDefault.take`, so `bytes.as-number` shadows `string.as-number`
`"42.5".as-number` does not parse the text: it terminates with `Can't convert non 8 length bytes to a number, bytes are 34-32-2E-35`. `string.eo` declares no own `as-number`, so the lookup in `eo-runtime/src/main/java/org/eolang/PhDefault.java:223-241` walks to the decoratee first (`this.take(Phi.PHI).take(name)`, line 231) and lands on `Φ.bytes.as-number` via the bytes' own package extension. `Φ.string.as-number` (`eo-runtime/src/main/eo/string/as-number.eo`) is never consulted, because `extension()` on line 238 runs only after the whole decoratee chain has bottomed out into a `PhTerminator`.

The order looks backwards. A package object belongs to the object's own surface — `extension()` binds the receiver at α0 precisely so that `42.power 3` reads as `number.power 42 3` — so it should shadow whatever the decoratee happens to expose, exactly as a direct attribute does. Ranking it below the decoratee means every data object silently inherits `bytes`' version of any name it defines in its own package. It also blocks moving methods out of the core objects: `string.slice` cannot become `string/slice.eo`, since `bytes.slice` would take over and slice by byte offsets instead of characters (ASCII cases keep passing, multi-byte ones do not).

Flip the order in `take()`: after the own-attribute and `Λ` cases, try `this.extension(name, ...)` for this object's forma, and delegate to `Λ`/`@` only when that finds nothing. `OnClasspath` already memoizes both the hit and the miss, so moving the probe onto the hot path costs one map lookup. Worth noting for whoever takes this: `string/slice.eo` and `tuple/with.eo` are delegating aliases (`text.slice > @`, `list.with x > @`) that terminate today only because the core object still owns the same-named attribute; they need real bodies once their methods migrate.

yegor256:
`PhDefault.take()` now looks into the object's own package before it
descends into `Λ` or `@`. Own attributes still win first, and a name that
neither the package nor the decoratee knows still terminates.

The old order made every data object silently inherit `bytes`' version of
any name it defines in its own package, because the extension was probed
only after the whole decoratee chain had bottomed out. That is why
`"42.5".as-number` used to die with `Can't convert non 8 length bytes to a
number` instead of reaching `string/as-number.eo`. A package object belongs
to the surface of the object itself — `extension()` binds the receiver at
α0 exactly so that `42.power 3` reads as `number.power 42 3` — so it should
shadow the decoratee the same way a direct attribute does.

Two things worth knowing. The classpath probe now sits on the hot path, so
`OnClasspath` gained a cache keyed by the EO name, which means the
EO-to-Java name translation is paid once per name instead of once per
dispatch; `PhNest` reuses it. And `string/slice.eo` and `tuple/with.eo`
still work only because `string` and `tuple` own those attributes
themselves — once anyone moves those methods into the packages, the
delegating aliases will need real bodies.

Closes #5930

### DIFF
--- a/eo-runtime/src/main/eo/string/as-number.eo
+++ b/eo-runtime/src/main/eo/string/as-number.eo
@@ -38,6 +38,8 @@
     5
     as-number "5"
 
+  42.5.eq "42.5".as-number ++> tests-converts-text-to-number-via-implicit-dispatch
+
   eq. ++> tests-currency-prefixed-text-returns-provided-fallback
     42
     as-number
--- a/eo-runtime/src/main/java/org/eolang/OnClasspath.java
+++ b/eo-runtime/src/main/java/org/eolang/OnClasspath.java
@@ -26,6 +26,12 @@ final class OnClasspath {
      */
     private static final Map<String, Boolean> CACHE = new ConcurrentHashMap<>(0);
 
+    /**
+     * EO name to presence, so that a name asked in EO notation pays the
+     * translation to Java notation at most once.
+     */
+    private static final Map<String, Boolean> OBJECTS = new ConcurrentHashMap<>(0);
+
     private OnClasspath() {
     }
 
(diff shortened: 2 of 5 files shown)


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain any part of the reordering if something looks off.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (166 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2723](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5932, opened 2026-07-26
fix(runtime): #5928 make malloc.of an atom and drop the allocated alias

### LINKED ISSUE #5928: Unnecessary `allocated` alias and wrapper in `malloc.of` causing redundancy
In `eo-runtime/src/main/eo/malloc.eo:75-77` the `of` object is a formation that holds an atom and one extra attribute:

```
[size scope] > of
  [] > @ /Q.bytes
  chunk > allocated
```

The `allocated` attribute is a plain alias of the root `chunk` object, and the only thing that ever reads it is the atom itself, at `eo-runtime/src/main/java/org/eolang/EOmalloc$EOof$EOφ.java:29`, as `rho.take("allocated").copy()`. That indirection buys nothing: the atom can reach `chunk` from the root directly, exactly the way `EOdirectory$EOwalk.java:54` reaches `file` with `Phi.Φ.take("file").copy()`. Nothing else in the sources touches the `malloc.of.allocated` path; the only other mentions are two synthetic eo-printer fixtures (`atom-rooted-signature.yaml`, `inline-phi-hybrid-under-width.yaml`).

Once `allocated` goes away, `of` holds nothing but its own `@`, so the wrapper formation becomes dead weight too. Two objects exist where one suffices, and readers of `malloc.eo` have to follow an alias to learn that a memory block is just a `chunk`.

The fix is to declare `of` as an atom, `[size scope] > of /Q.bytes`, drop the `allocated` line, rename `EOmalloc$EOof$EOφ` to `EOmalloc$EOof`, and read `size` and `scope` off `this` instead of `rho` while taking the chunk via `Phi.Φ.take("chunk").copy()`. Note that `malloc.eo` also declares `+rt node eo2js-runtime:0.0.0`, so the matching atom in the eo2js runtime needs the same rename.

androvonx95:
Closes #5928.

`malloc.of` used to be a formation with an atom and an `allocated` alias pointing to the root `chunk`. Since only the atom ever used it, it just added an unnecessary level of indirection and an extra object.

Now `of` is just an atom that accesses `chunk` directly from the root:

```eo
[size scope] > of /Q.bytes
```

`EOmalloc$EOof$EOφ` is renamed to `EOmalloc$EOof` and now reads `size` and `scope` from `this`, taking `chunk` directly from the root. This keeps the behavior the same while simplifying the structure.

Updated the two eo-printer fixtures that referenced the old path and adjusted the related comment in `to-eo-tree.xsl`.

Everything passes, including the full `eo-runtime` test suite, eo-printer `XmirTest` packs, and `-P qulice`.

`malloc.eo` also declares `eo2js-runtime`, so the matching `malloc$of$φ.js` change still needs to be done in the `objectionary/eo2js` repo. That's a separate repo and isn't included here.

### DIFF
--- a/eo-runtime/src/main/eo/malloc.eo
+++ b/eo-runtime/src/main/eo/malloc.eo
@@ -72,9 +72,7 @@
       seq * > [m] >>
         m.write 0 bts
         scope m
-  [size scope] > of
-    [] > @ /Q.bytes
-    chunk > allocated
+  [size scope] > of /Q.bytes
 
   malloc.for ++> copies-and-resizes-to-target
     0
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -414,7 +414,7 @@
       </xsl:choose>
       <xsl:if test="eo:atom(.)">
         <!--
-        A multi-segment signature (e.g. "Φ.malloc.of.allocated") stays
+        A multi-segment signature (e.g. "Φ.string.regex.pattern") stays
         rooted: on reparse a dotted @atom is not re-homed, so dropping the
         root would yield an XSD-invalid fqn. A single-name signature (e.g.
         "Φ.bytes") re-resolves to its root on reparse, so the implicit
--- a/eo-printer/src/test/resources/org/eolang/printer/print-packs/yaml/atom-rooted-signature.yaml
+++ b/eo-printer/src/test/resources/org/eolang/printer/print-packs/yaml/atom-rooted-signature.yaml
@@ -12,9 +12,9 @@ penalties:
 origin: |
   +package org.eolang
 
-  [size] > resized /Q.malloc.of.allocated
+  [size] > resized /Q.string.regex.pattern
 
 printed: |
   +package org.eolang
 
-  [size] > resized /Q.malloc.of.allocated
+  [size] > resized /Q.string.regex.pattern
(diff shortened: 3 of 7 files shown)


yegor256:
@androvonx95 merge conflict here, please fix

eo-runtime/src/main/eo/malloc.eo
```diff
@@ -67,14 +67,12 @@
...
-        object >> bts!
+      bts.size
       seq * > [m] >>
         m.write 0 bts
         scope m
-  [size scope] > of
-    [] > @ /Q.bytes
-    chunk > allocated
+    object > bts!
+  [size scope] > of /Q.bytes
```
yegor256, line 75: @androvonx95 it's better to put `/Q.chunk` here, instead of `/Q.bytes`

androvonx95:
Fixed the merge conflict and rebased onto current master.

I checked /Q.chunk with -Deo.typing=true but it fails because malloc.of returns the dataized scope result, which is bytes so the chunk is only passed into the scope. /Q.bytes matches both the prior declaration and the actual returned value, so I kept it.

All runtime tests, printer packs, formatting, and qulice pass after the rebase.

0crat:
@androvonx95 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2745](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5933, opened 2026-07-26
`string.nsplit` misuses byte offsets on character-based `string.slice`, breaking multi-byte inputs


yegor256:
In `eo-runtime/src/main/eo/string/nsplit.eo:43` the delimiter is compared against `slice text current size`, and line 63 cuts each piece with `slice text start (current.minus start)`. Every index in that arithmetic is a byte quantity: `size` is `delimiter.size` (line 72) and the loop is bounded by `text.size` (lines 19, 26 and 33), all of which reach `bytes.size` through the decoratee. The `slice` they are handed to is `Φ.string.slice`, which counts characters, not bytes.

For ASCII the two units coincide, which is why all seven tests in the file pass. For anything multi-byte they diverge. Take `nsplit "аХбХв" "Х" 3`: the text is 5 characters but 10 bytes, so `current` is allowed to walk up to 9, yet `slice text 4 2` already asks for characters 4 and 5 of a 5-character string and the walk terminates with an out-of-bounds error instead of returning the parts. Where it does not overrun, the pieces are still cut at character offsets computed from byte positions, so they are wrong. The sibling `split.eo` escapes this because it slices `bts`, a `!` const handle that dataizes to bytes, so `bts.slice` resolves to `bytes.slice`; `nsplit` slices the string itself. The byte-sized length arrived with the fix for #5214, which taught both objects to compare `delimiter.size` bytes.

Bind the text as bytes once, `text.as-bytes >> bts!`, and slice that in both places instead of `text`, exactly as `split.eo` already does. The piece is wrapped in `string` on line 62 either way, so nothing else has to move.

0crat:
@yegor256 Thanks for closing this issue! You've earned +6 points for this: +8 as a basis; -2 for 4 days of delay. Your running score is [+2915](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5934, opened 2026-07-26
Extract string.slice into its own package file

yegor256:
This moves `slice` out of the core `string` object and into `eo-runtime/src/main/eo/string/slice.eo`, the same way `u64.div` and `i64.div` were moved. The receiver now arrives as the first argument `text`, so the body reaches `as-bytes` through it, and package dispatch keeps every `.slice` call site working untouched. The twelve slicing tests and the three error tests move along with the code, and the wrapper that used to sit in that file — it just delegated back to `string.slice` — is gone.

Two things only showed up once the wrapper was out of the way. First, a package object *is* the value it hands back, so `slice`'s public `size` helper stayed reachable on the result and answered with the length of the source text; that broke `string.length`, so `size` is private now. Second, `contains`, `ends-with`, `starts-with`, `index-of`, `last-index-of` and `split` were slicing bytes all along without saying so: a `!` const handle dataizes to bytes, and the old wrapper quietly re-dispatched onto `bytes.slice`. A real char-based body flipped them to characters and broke UTF-8 (`index-of "привет, друг!" "в"` started blowing up), so those six now spell it out as `txt.slice`.

`nsplit` has the same byte/character muddle but no const handle to save it, so it is genuinely broken for multi-byte text today; that is filed separately as #5933 and left alone here. The `p1`..`r4` constants in the moved code trip `redundant-object`, so the file carries the same `+unlint` that `string.eo` already had for them — `string.eo`'s twin `length` object still needs it. Full reactor is green: 1513 + 274 + 463 + 1767 tests, no failures.

Part of #5678.

### DIFF
--- a/eo-runtime/src/main/eo/string.eo
+++ b/eo-runtime/src/main/eo/string.eo
@@ -74,102 +74,6 @@
     p2 > r3
     p3 > r4
     as-bytes.size > size
-  [start len] > slice
-    if. > @
-      nstart.lt 0
-      T
-        "Start index must be >= 0, but was %d".printf
-          * nstart
-      if.
-        nlen.lt 0
-        T
-          "Length to slice must be >= 0, but was %d".printf
-            * nlen
-        seq *
-          recidx >> bstart!
-            0
-            0
-            nstart
-            "Start index (%d) is out of string bounds".printf
-              * nstart
-          blen
-          if.
-            nlen.eq 0
-            ""
-            string
-              as-bytes.slice bstart blen
-    [index accum result cause] >> recidx
-      if. > @
-        accum.eq result
-        index
-        if.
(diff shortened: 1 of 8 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (415 >= 100); -8 for way too many hits-of-code (415 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2727](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5935, opened 2026-07-26
Extract string.length into its own package file

yegor256:
Moves the character counter out of the core `string` object into
`string/length.eo`, following the same shape as the `slice` extraction.
The text becomes the first argument, and every existing `text.length`
call site keeps working through package dispatch on the forma. The five
tests move along with the code.

One thing worth a look: `size` is now a private `>>` handle instead of a
public attribute. A package object is the value it returns, so a public
`size` helper would be found before the `bytes.size` of the number that
`length` answers — the same leak that broke `string.length` when `slice`
moved out. Dropping `length` also left `+unlint redundant-object`
unused in `string.eo`, since the byte-mask constants it covered went
with it.

I walked every `.length` call site before touching anything. The ones on
tuples are a different object, and the string ones all have plain string
receivers, so unlike `slice` there were no const-handle sites needing a
fix. Full reactor is green: 1768 runtime tests, no failures, and a
second `eo:format` pass leaves the sources unchanged.

Part of #5678.

### DIFF
--- a/eo-runtime/src/main/eo/string.eo
+++ b/eo-runtime/src/main/eo/string.eo
@@ -17,69 +17,12 @@
 +version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
-+unlint redundant-object
 +unlint mandatory-package
 
 [as-bytes] > string
   as-bytes > @
-  [] > length
-    if. > @
-      size.eq 0
-      0
-      reclen 0 0
-    [index accum] >> reclen
-      if. > @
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (185 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2731](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5936, opened 2026-07-26
Extract number.as-i64 into its own package file

yegor256:
Moves the IEEE 754 to i64 conversion out of the core `number` object
into `number/as-i64.eo`, the same shape as the `slice` and `length`
extractions. The number becomes the first argument, the `cant-convert`
fallback stays the second, and all eleven tests move with the code.
Every `x.as-i64` call site keeps working through package dispatch.

The one thing that needed care is `as-i32`, the only caller of `as-i64`
inside `number.eo`. It used to be a bare reference, which no longer
resolves once the attribute is gone, and writing `$.as-i64` does not
help because the formatter strips the prefix and the resolver then goes
looking for a root `Φ.as-i64`. Naming the package object explicitly
(`number.as-i64 $`) is what works, and the socket tests catch it if it
regresses, since they reach `as-i32` through `posix.inet`.

Unlike `slice`, no helper needed privatizing here — `exp`, `mantissa`
and friends don't collide with anything on the `i64` this returns.
Full reactor is green: 1513, 274, 463 and 1768 tests, no failures, and
a second `eo:format` pass leaves the sources unchanged.

Part of #5678.

### DIFF
--- a/eo-runtime/src/main/eo/number.eo
+++ b/eo-runtime/src/main/eo/number.eo
@@ -14,47 +14,8 @@
 [as-bytes] > number
   as-bytes > @
   as-i32.as-i16 > as-i16
-  as-i64.as-i32 > as-i32
-  [cant-convert] > as-i64
-    overflow.if > @
-      cant-convert
-        "Can't convert number %f to i64 because it's out of i64 bounds".printf
-          * ^
-      if.
-        exp.lt 1023
-        00-00-00-00-00-00-00-00.as-i64
-        if.
-          exp.eq 1086
-          80-00-00-00-00-00-00-00.as-i64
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (175 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2725](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5937, opened 2026-07-26
Extract number.eq into its own package file

yegor256:
Moves the IEEE equality check out of the core `number` object into
`number/eq.eo`, the same shape as the `slice`, `length` and `as-i64`
extractions. The number becomes the first argument and every `x.eq y`
call site keeps working through package dispatch.

The judgement call here is which tests travel with the code. `eq` is
the assertion operator for most of `number.eo`, so I only moved the
twenty whose actual subject is equality — the nan and signed-zero
rules, and the infinity comparisons. Everything testing `div`, `plus`
or `times` stays put even though it asserts with `eq`, and so do the
ones comparing bytes rather than numbers.

Nothing needed privatizing: `nzero` and `pzero` leak onto the result
the way `size` did for `slice`, but they don't collide with anything on
the `bool` this returns. This one also went in without the internal
reference problem `as-i64` had, since no code inside `number.eo` names
`eq` bare. I merged master in first, so the three extractions are
verified together — full reactor green at 1513, 274, 463 and 1770
tests, and `eo:format` leaves the sources unchanged.

Part of #5678.

### DIFF
--- a/eo-runtime/src/main/eo/number.eo
+++ b/eo-runtime/src/main/eo/number.eo
@@ -56,31 +56,9 @@
             negative.and
               mantissa.eq 00-10-00-00-00-00-00-00
   [x] > div /Q.number
-  [x] > eq
-    if. > @
-      ^.is-nan.or
-        is-nan.
-          number
-            x >> xbts!
-      false
-      or.
-        and.
-          or.
-            xbts.eq pzero
-            xbts.eq nzero
-          or.
-            eq.
-              as-bytes >> b!
-              pzero
-            b.eq nzero
-        b.eq xbts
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (320 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2729](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5938, opened 2026-07-26
Extract socket i/o adapters into their own package files

yegor256:
Moves `as-input` and `as-output` out of the core `socket` object into
`socket/as-input.eo` and `socket/as-output.eo`. They travel together
because they are the same idea pointing in two directions and are used
side by side in `scoped-socket`. This also creates the `socket`
package, which did not exist before.

These two break the pattern of the earlier extractions in one way:
neither gains a receiver argument. `as-input` only ever used the `recv`
handed to it and `as-output` only the `send`, never the socket itself,
so adding a socket parameter would mean carrying something unused.
Their two call sites in `scoped-socket` now name the package objects
instead of climbing `^.^`.

The nice side effect is that once they stand apart from `socket` they
need no network to exercise, so each one gets a unit test that feeds it
a stub — `recv` answering canned bytes, `send` writing into a malloc
block so the test can check the buffer arrived intact. That is two more
tests than `socket` had before, and the rest of it keeps its
`unit-test-missing` unlint. Full reactor green at 1513, 274, 463 and
1771 tests.

Part of #5678.

### DIFF
--- a/eo-runtime/src/main/eo/socket.eo
+++ b/eo-runtime/src/main/eo/socket.eo
@@ -91,29 +91,6 @@
                 *
         posix > raw
         0 > startup
-  [recv] > as-input
-    [size] > read
-      @. > @
-        read.
-          input-block --
-          size
-      [buffer] > input-block
-        buffer > @
-        [size] > read
-          seq * > @
-            chunk
-            input-block chunk
-          as-bytes. > chunk
-            recv size
-  [send] > as-output
-    [buffer] > write
-      @. > @
-        output-block.write buffer
-      [] > output-block
-        true > @
-        seq * > [buffer] > write
-          send buffer
-          output-block
   [sys] >> impl
     code. > addr
       sys.raw *1
@@ -219,8 +196,8 @@
           sys.cleanup
           result
     [descriptor] > scoped-socket
-      ^.^.as-input recv > as-input
-      ^.^.as-output send > as-output
+      socket.as-input recv > as-input
+      socket.as-output send > as-output
       [size cant-receive] > recv
         if. > @
           received.code.eq -1
--- a/eo-runtime/src/main/eo/socket/as-input.eo
+++ b/eo-runtime/src/main/eo/socket/as-input.eo
@@ -1,0 +1,33 @@
+# Turns the `recv` function of a socket into an `input`, so that bytes arriving
+# on the socket can be pulled through the `input` protocol. Here `recv` must be
+# an object that takes the amount of bytes to fetch and answers with them.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package socket
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[recv] > as-input
+  [size] > read
+    @. > @
+      read.
+        input-block --
+        size
+    [buffer] > input-block
+      buffer > @
+      [size] > read
+        seq * > @
+          chunk
+          input-block chunk
+        as-bytes. > chunk
+          recv size
+
+  ++> tests-reads-the-bytes-that-recv-answers
+    eq. > @
+      read.
+        as-input recv
+        3
+      01-02-03
+    01-02-03-04-05.slice 0 size > [size] > recv
--- a/eo-runtime/src/main/eo/socket/as-output.eo
+++ b/eo-runtime/src/main/eo/socket/as-output.eo
@@ -1,0 +1,32 @@
+# Turns the `send` function of a socket into an `output`, so that bytes can be
+# pushed to the socket through the `output` protocol. Here `send` must be an
+# object that takes the buffer to push out and answers with the amount of bytes
+# that left.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package socket
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[send] > as-output
+  [buffer] > write
+    @. > @
+      output-block.write buffer
+    [] > output-block
+      true > @
+      seq * > [buffer] > write
+        send buffer
+        output-block
+
+  ++> tests-hands-the-buffer-over-to-send
+    malloc.of > @
+      3
+      [m]
+        seq * > @
+          write.
+            as-output snd
+            01-02-03
+          m.get.eq 01-02-03
+        m.put buffer > [buffer] > snd


eo-runtime/src/main/eo/socket/as-input.eo
```diff
@@ -0,0 +1,33 @@
+# Turns the `recv` function of a socket into an `input`, so that bytes arriving
+# on the socket can be pulled through the `input` protocol. Here `recv` must be
+# an object that takes the amount of bytes to fetch and answers with them.
```
Favixx, line 3: "an object that takes the amount of bytes to fetch and answers with them" describes a one-parameter `recv`, and the only `recv` this is ever applied to has two: `socket.eo` declares `[size cant-receive] > recv` inside `scoped-socket`, and passes it here as `socket.as-input recv`.

The second parameter is the caller's error value, chosen at the call site for the case where the syscall returns -1. Since this doc is new — the object had no documentation while it lived inside `socket` — it is the moment to describe the parameter that actually exists, because the discrepancy is not cosmetic (see my comment on line 25).

eo-runtime/src/main/eo/socket/as-input.eo
```diff
@@ -0,0 +1,33 @@
...
+        input-block --
+        size
+    [buffer] > input-block
+      buffer > @
+      [size] > read
+        seq * > @
+          chunk
+          input-block chunk
+        as-bytes. > chunk
+          recv size
```
Favixx, line 25: `recv size` binds one of the two parameters that the real `recv` declares, so `cant-receive` is left unbound on every read that goes through the `input` protocol.

While this lived inside `socket` that was invisible. Extracted, documented and given its own tests, it becomes the specified behaviour of a public object: `s.as-input.read 5` — the usage the header comment of `socket.eo` advertises at lines 13-14 — reaches a `recv` whose failure branch resolves to an attribute nobody bound, instead of to the value the socket's caller chose. The failure surfaces as an unbound attribute rather than as the caller's `cant-receive`.

Carried over unchanged, so not a regression. But `as-input recv` could pass the error value through as a second argument of its own, and the new doc could say what happens when the socket fails mid-read — right now neither the code nor the comment addresses it.

eo-runtime/src/main/eo/socket/as-input.eo
```diff
@@ -0,0 +1,33 @@
...
+
+  ++> tests-reads-the-bytes-that-recv-answers
+    eq. > @
+      read.
+        as-input recv
+        3
+      01-02-03
+    01-02-03-04-05.slice 0 size > [size] > recv
```
Favixx, line 33: The stub is `[size] > recv` — one parameter, matching the doc comment rather than the `[size cant-receive] > recv` that `scoped-socket` actually hands over.

That is what makes the mismatch above untestable: with a one-parameter stub, `recv size` is a complete application and everything works; with the real two-parameter object it is a partial one. A stub with the socket's shape, whose second parameter is something the test can observe, would turn "the error path is not wired" from an argument into a red test.

eo-runtime/src/main/eo/socket/as-input.eo
```diff
@@ -0,0 +1,33 @@
...
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[recv] > as-input
+  [size] > read
+    @. > @
+      read.
+        input-block --
+        size
+    [buffer] > input-block
```
Favixx, line 18: `input-block` is the whole substance of this object — a block that carries a buffer and hands back a new block on each read — and the test exercises it once, so `input-block chunk` is constructed and never read from.

The uncovered behaviour is the one a user of the `input` protocol depends on: reading twice in a row. `read. (as-input recv) 3` followed by another `read.` on what came back, against a stub that answers different bytes each time, is four more lines and covers the recursion that the single-read test cannot reach.

eo-runtime/src/main/eo/socket/as-input.eo
```diff
@@ -0,0 +1,33 @@
...
++package socket
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[recv] > as-input
+  [size] > read
+    @. > @
+      read.
+        input-block --
```
Favixx, line 16: A question about behaviour that this move makes public rather than changes: `read size` calls `recv size` exactly once and answers whatever came back, so a `recv` that delivers fewer bytes than asked — the ordinary short read on a real socket — makes `read 5` answer three bytes without a word.

The shape of `input-block` suggests looping until `size` bytes have accumulated was the intent, but there is no condition anywhere that compares the accumulated length against `size`: the recursion builds one block and stops. If short reads are meant to be the caller's problem, the doc comment is the place to say so, now that the object is documented at all.

eo-runtime/src/main/eo/socket/as-input.eo
```diff
@@ -0,0 +1,33 @@
...
+# an object that takes the amount of bytes to fetch and answers with them.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package socket
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[recv] > as-input
```
Favixx, line 12: Nothing in the body is socket-specific: it needs an object that answers bytes for a size, and the new test proves that by handing it a slice of a byte literal. That is a nice property of the extraction and it is worth one clause in the doc — otherwise the `socket` package reads as a constraint on what may be passed, and the next person who wants an `input` over a file or a pipe writes a second copy of this file.

eo-runtime/src/main/eo/socket/as-output.eo
```diff
@@ -0,0 +1,32 @@
+# Turns the `send` function of a socket into an `output`, so that bytes can be
+# pushed to the socket through the `output` protocol. Here `send` must be an
+# object that takes the buffer to push out and answers with the amount of bytes
+# that left.
```
Favixx, line 4: "answers with the amount of bytes that left" documents a return value that this object then discards: `seq * (send buffer) output-block` evaluates `send buffer` for its effect and answers `output-block`, so the count never reaches the caller of `write`.

Either the count matters — in which case a short write is currently silent, and that is the bug worth filing — or it does not, in which case the sentence should say that `send`'s answer is ignored. As written it promises the caller something they cannot obtain. Same shape as the `recv` note in `as-input.eo`: the real `send` in `socket.eo` is `[buffer cant-send]`, two parameters, and only one is bound here.

eo-runtime/src/main/eo/socket/as-output.eo
```diff
@@ -0,0 +1,32 @@
...
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[send] > as-output
+  [buffer] > write
+    @. > @
+      output-block.write buffer
+    [] > output-block
+      true > @
```
Favixx, line 18: `true > @` makes the eventual answer of `write` the boolean `true`, whereas `as-input.read` answers the bytes it read. Two halves of "the same idea pointing in two directions", as the description puts it, that answer different kinds of thing.

Not something to change in this PR — but the pair is being given a package and a documented contract for the first time, so it is the moment to write down that `write` answers a block and not a count, before something starts depending on the boolean.

eo-runtime/src/main/eo/socket/as-output.eo
```diff
@@ -0,0 +1,32 @@
...
+
+  ++> tests-hands-the-buffer-over-to-send
+    malloc.of > @
+      3
+      [m]
+        seq * > @
+          write.
+            as-output snd
+            01-02-03
+          m.get.eq 01-02-03
```
Favixx, line 31: The assertion checks the side effect — the bytes reached `m` — and nothing checks what `write` answered. Given that `write` goes through `@.` and ends at `output-block`'s `true`, that is the only observable part of the return value, and it is unasserted.

One more line inside the `seq` — comparing the result of the `write.` application against what it is supposed to be — would pin the half of the contract the doc comment gets wrong (see line 4).

eo-runtime/src/main/eo/socket/as-output.eo
```diff
@@ -0,0 +1,32 @@
...
+        send buffer
+        output-block
+
+  ++> tests-hands-the-buffer-over-to-send
+    malloc.of > @
+      3
+      [m]
+        seq * > @
+          write.
+            as-output snd
```
Favixx, line 29: One `write`, so the block returned by `seq * (send buffer) output-block` is built and dropped — the same gap as the single `read` in `as-input.eo`. Chaining a second `write` onto the result is what tells you the returned block is usable, which is the reason it is returned at all.

With `m` already in scope and `malloc.of 3` holding the last buffer, a second write of different bytes and an assertion on `m.get` covers it without new scaffolding.

eo-runtime/src/main/eo/socket.eo
```diff
@@ -219,8 +196,8 @@
           sys.cleanup
           result
     [descriptor] > scoped-socket
-      ^.^.as-input recv > as-input
-      ^.^.as-output send > as-output
+      socket.as-input recv > as-input
```
Favixx, line 199: `^.^.as-input` and `socket.as-input` are not the same lookup. The old form walked two scopes up and took whatever `as-input` the enclosing `socket` object had; the new one resolves through the package to `org.eolang.socket.as-input`, always.

For `socket` itself nothing changes. For anyone who decorates `socket` and overrides `as-input` — which the old lexical form would have honoured — `scoped-socket` now silently keeps using the package object. That is the same asymmetry the `directory`/`file` extractions ran into with `deleted`, and it is a real (if unlikely) behaviour change hiding inside a move. Worth a sentence in the description, since "extract into their own package files" reads as pure relocation.

eo-runtime/src/main/eo/socket.eo
```diff
@@ -91,29 +91,6 @@
                 *
         posix > raw
         0 > startup
-  [recv] > as-input
```
Favixx, line 94: These two objects had no tests of their own while they lived here, and now they have one each — that is the real gain of the move and it is worth saying in the description.

What neither the new tests nor `socket.eo` covers is the join: no test builds a `scoped-socket` and reads or writes through its `as-input`/`as-output`, so the one line this PR actually changed in production code — the call site below — is exercised by nothing. A test at that level is admittedly harder, needing a real or faked descriptor; if that is out of scope, it is worth a puzzle, because the package dispatch it now relies on is newer than the code around it.

eo-runtime/src/main/eo/socket/as-output.eo
```diff
@@ -0,0 +1,32 @@
...
+# that left.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package socket
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[send] > as-output
```
Favixx, line 13: The two files are near-identical in shape — parameter, inner `write`/`read`, inner block, `@.` unwrapping — and they were extracted together for exactly that reason. Since they are now separate files, that symmetry is only visible to someone who opens both.

A cross-reference in each doc comment ("the write-side counterpart of `socket.as-input`") costs one line and keeps the pair discoverable, which matters more now than when they were twenty lines apart in the same object.

Favixx (commented):
**The red CI is #6091, not this change.** The build fails on `MjParseTest.parsesConcurrentlyWithLotsOfPrograms`, which is the Saxon-HE 13.0 `xsl:function` parameter bug you filed on the 29th — it fails in 2-3% of runs by that issue's own measurement. This PR has been sitting red since the 26th for that reason alone, and one re-run would very likely turn it green.

Worth noting what that costs here: the failure is in `eo-maven-plugin`, module 4 of 6, so the reactor stops before `eo-runtime` — and every line of this PR is in `eo-runtime`. Neither of the two new tests has been executed by CI even once.

I checked the thing an extract-to-package move has to get right and it holds: there is exactly one call site of `as-input`/`as-output` in `socket.eo`, in `scoped-socket`, and both lines are updated. Nothing else in the repository referenced them.

What I would look at, in order: the two new doc comments describe a `recv`/`send` contract that the only real caller does not have (details inline), and both new tests use stubs shaped to the doc rather than to the socket, so nothing can catch the difference.

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+112](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5947, opened 2026-07-26
`inline-cactoos` fails to retain binding name in folded `>>` handle, leading to unparseable output


yegor256:
When `inline-cactoos` folds a based `>> name` file-local handle (R-3.10.12) — one
whose value is a plain reference or application, not a formation and not a const —
into a reference that carries its own binding name, that name is dropped. `[a] > foo`
with `b > x` and `a.as-i32 >> b` prints as:

```
[a] > foo
  a.as-i32
```

The `> x` is gone, and the result is not merely less readable: the reprint fails with
`[2:2] error: 'object inside formation must have a name'`, so `eo:format` turns a
valid source file into one the parser rejects. A single reference is enough; the
handle need not be shared.

The `otherwise` branch of the inlining template rebuilds the reference from the
target's attributes and children, and copies the reference's own `@name` only when it
is the `φ` decoratee — `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:158`,
the carve-out added for #5810. Every other name is lost. The two branches above it
that do preserve an applied reference's name (#5834 at line 96, #5844 at line 126)
both test `eo:abstract($target)` first, so a based target never reaches them, and
`$keep-name` (line 75) is false for a based target, so the target contributes no name
either.

Widening that guard is enough: at line 158 carry the reference's `@name` over whenever
the target contributes none, i.e. replace `@name = $eo:phi and not($keep-name)` with
`@name and not($keep-name)`, so `a.as-i32 > x` is printed instead of a nameless body
line.

PR #5948, opened 2026-07-26
Keep the name of a shared based >> handle when printing

### LINKED ISSUE #5944: `eo-printer` misplaces `>>` handle name and creates dangling references on `vL_P` in multi-reference scenarios
A file-local `>> name` handle whose value is neither an abstract formation nor a
const — a plain application such as `x.as-u16 >> other` (R-3.10.12) — loses its
name when more than one reference reaches it, and every reference but the first
is left pointing at a synthetic `vL_P` id. Minimal case, `[a] > foo` with
`b.plus 1 > x`, `b.neg > y` and `a.as-i32 >> b`, prints as:

```
[a] > foo
  plus. > x
    a.as-i32 >>
    1
  v4_2.neg > y
```

`v4_2` is declared nowhere, so the reprint still parses but binds to a dangling
`Φ.v4_2`. `eo-runtime/src/main/eo/u16.eo`'s `div` hits this once its
`x.as-u16 > other` is privatized to `>>`, since `other` feeds both `other.eq zero`
and `other.as-i32`.

The binding survives `inline-cactoos.xsl:194` because `eo:dispatched` holds, so it
reaches `to-eo-tree` and must carry a readable name — but
`restore-local-names.xsl:181` drops its `@local`, as the multi-reference carve-outs
there admit only an abstract formation (#5876) and a dataized const (#5828), never
a based handle. `merge-monikers.xsl` then hosts the binding onto the first
reference, dropping `@name` and `@local` on the way (line 286, the #5810 rule for a
based binding), and leaves the other references untouched: `eo:kept-const-ref`
(line 184) demands `eo:const-handle` and `eo:kept-local-ref` (line 241) demands
`eo:abstract`, so neither rewrite fires and the cactus name prints raw. This is the
third flavour of the same defect already fixed for formations (#5876) and consts
(#5883).

The fix is to admit the based handle to those two carve-outs: drop the
`eo:abstract(.) and` conjunct from the multi-referenced clause of the `@local`-drop
template in `restore-local-names.xsl:181`, and widen `eo:kept-local-ref` in
`merge-monikers.xsl:243` from `eo:abstract(.)` to any `@local`-carrying binding, so
the non-hosting references read back as `b` while the hosting one keeps its
`@name`/`@local` the way `eo:const-handle` already does at line 286.

yegor256:
A based `>> name` file-local handle — one whose value is a plain application, so
neither an abstract formation nor a const — now keeps its readable name when more
than one reference reaches it. `restore-local-names` no longer drops the `@local`
marker of a shared binding just because it is not a formation, and
`merge-monikers` rewrites the references it does not host back to that handle, the
way it already does for shared formations and shared consts.

Without this, `x.as-u16 >> other` used from both `other.eq zero` and
`other.as-i32` printed as an anonymous `x.as-u16 >>` folded into the first use,
with the second one left reading `v25_4` — a name declared nowhere, so the
reformatted source reparsed into a dangling `Φ.v25_4`. The mangled form was also a
fixpoint, so running the formatter again would not have surfaced it. Closes #5944.

Three packs come with it: a round-trip pair for the hosted and the un-hostable
shape, and a sheet-level one pinning the `@local` marker that survives
`restore-local-names`. One neighbouring case is deliberately untouched and filed
separately as #5947: a single bare reference to a based handle still loses its own
`> name` binding when the handle is folded in.

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -223,13 +223,31 @@
     <xsl:sequence select="string-join((if (eo:shadowed($ref, $binding)) then subsequence($segs, 1, $at - 1) else $eo:xi, string($binding/@local), subsequence($segs, $at + 1)), '.')"/>
   </xsl:function>
   <!--
-  The kept multi-referenced abstract handle binding (`[] &gt;&gt; name`, #5876)
-  a reference `$ref` resolves to but does NOT host (the binding folds onto its
-  first reference only, `eo:moniker-refs`). Any other reference — a named alias
+  Whether the readable "@local" handle of `$attr` still stands after the merge,
+  so a reference that does not host it may read back through it. An abstract
+  formation carries its "@name" and "@local" to wherever it lands, and a binding
+  no reference hosts stays where it is; a based handle (`a.b &gt;&gt; name`)
+  keeps them only when its host is a method dispatch, which rebuilds the binding
+  whole as the receiver of a reversed dispatch. A bare host inlines it
+  anonymously instead (#5810), leaving no name for anyone else to read.
+  -->
+  <xsl:function name="eo:handle-survives" as="xs:boolean">
+    <xsl:param name="attr" as="element()"/>
+    <xsl:variable name="host" select="eo:moniker-refs($attr)[1]"/>
+    <xsl:sequence select="eo:abstract($attr) or empty($host) or eo:dispatch-seg($host) != ''"/>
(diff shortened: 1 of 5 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (159 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2733](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5949, opened 2026-07-26
#5933: slice bytes not characters in nsplit, matching split.eo

### LINKED ISSUE #5933: `string.nsplit` misuses byte offsets on character-based `string.slice`, breaking multi-byte inputs

morphqdd:
Closes #5933.

`nsplit`'s delimiter comparison and piece cutting both called `Φ.string.slice`, which counts characters, while every index around them (`current`, `start`, `size`) is a byte offset (`delimiter.size` and `text.size` both reach `bytes.size` through the decoratee). For ASCII text the two units coincide, so the existing tests never caught it. For multi-byte text the offsets and the slice unit disagree, either overrunning a character-length bound with a byte-length walk, or cutting pieces at the wrong character position.

Fix: bind `text.as-bytes >> bts!` once, same as the sibling `split.eo` already does, and slice `bts` (bytes, byte-based) in both places instead of `text` (string, character-based).

Added `tests-nsplit-with-multi-byte-characters`, splitting `"аХбХв"` on `"Х"` (each a 2-byte Cyrillic character), which fails against the old code and passes with this fix.


morphqdd:
@yegor256, please review

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2852](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5950, opened 2026-07-26
#5919: relativize path before glob matching in directory.walk

### LINKED ISSUE #5919: `directory.walk` improperly matches globs against absolute paths due to ineffective relativization
## What happens

`directory.walk` (`eo-runtime/src/main/java/org/eolang/EOdirectory$EOwalk.java`) tries to strip the walk root off each visited path before matching it against the user's glob, but the code that does it is a no-op, so the glob ends up matched against the full absolute path instead:

```java
final Path path = Paths.get(
    new Dataized(
        this.take(Phi.RHO).take("file").take("path")
    ).asString()
).toAbsolutePath();
...
try (Stream<Path> paths = Files.walk(path)) {
    return new Data.ToPhi(
        paths
            .map(p -> p.toAbsolutePath().toString())
            .map(p -> p.substring(p.indexOf(path.toString())))
            .filter(p -> matcher.matches(Paths.get(p))).map(
```

`path` is already absolute (`.toAbsolutePath()` on line with `Paths.get(...)`). `Files.walk(path)` only ever yields entries whose string form starts with `path.toString()`, so `p.indexOf(path.toString())` is always `0`, and `p.substring(0)` returns `p` unchanged. The intended relativization does nothing: every path handed to the `PathMatcher` stays absolute.

Because the paths stay absolute and multi-segment, `PathMatcher`'s `*` (which does not cross `/`) never matches a plain glob like `*.txt` against an absolute path such as `/tmp/xyz/foo/a.txt`. Only globs that already begin with `**` (which absorbs the whole absolute prefix) ever match anything.

Verified directly:

```
jshell> var matcher = FileSystems.getDefault().getPathMatcher("glob:*.txt");
jshell> matcher.matches(Paths.get("a.txt"))
$1 ==> true
jshell> matcher.matches(Paths.get("/tmp/x/a.txt"))
$2 ==> false
```

Every test and call site in the repo already works around this by always prefixing the glob with `**` (`directory.eo:24`, `(walk "**").at.^`; `directory.eo:103`, `d.as-dir.walk "**/*.txt"`), so the gap is untested and unnoticed. `dir.walk "*.txt"` (matching `.txt` files directly under `dir`, without recursing) silently returns an empty tuple even when such files exist.

This is also directly relevant to #5751, which is about pulling relativization and glob matching out of this atom into pure EO: #5751 describes the atom as currently relativizing every path against the base directory, but that relativization is actually dead code, so whoever rewrites this in EO should not copy the current `substring` logic as-is.

## What should happen

The path handed to the matcher should actually be relative to the walk root, for example `path.relativize(p).toString()` instead of the no-op `substring` call (and the redundant `.toAbsolutePath()` on `p` can be dropped), so a plain glob like `*.txt` matches direct children the way a user would expect.

morphqdd:
Closes #5919.

`directory.walk` tried to strip the walk root off each visited path before matching it against the glob, but the code that did it was a no-op: `path` is already absolute, so `p.toAbsolutePath().toString()` stays the same string, and `p.substring(p.indexOf(path.toString()))` cuts from index 0, returning `p` unchanged. The glob ended up matched against the full absolute path instead of a path relative to the walk root.

Because of this, `PathMatcher`'s `*` (which does not cross `/`) never matched a plain glob like `*.txt` against an absolute, multi-segment path. Only globs already starting with `**` (which absorbs the whole absolute prefix) matched anything, and every existing call site already works around this by always prefixing with `**`, so the gap went untested.

Fix: match the glob against `path.relativize(p)` instead. The returned `file` object still gets the real absolute path (`p.toString()`), so callers see no change there.

Added `tests-walks-with-plain-glob-matches-direct-child-only`, which creates one direct `.txt` file and one nested one, then checks a plain `*.txt` glob returns only the direct child. This fails against the old code (returns 0 matches) and passes with this fix.

### DIFF
--- a/eo-runtime/src/main/eo/directory.eo
+++ b/eo-runtime/src/main/eo/directory.eo
@@ -120,6 +120,25 @@
       made.
         directory mktemp.tmpfile.deleted
 
+  ++> tests-walks-with-plain-glob-matches-direct-child-only
+    seq * > @
+      touched.
+        as-file.
+          d.resolved "a.txt"
+      made.
+        as-dir.
+          d.resolved "sub"
+      touched.
+        as-file.
+          d.resolved "sub/b.txt"
+      eq.
+        length.
+          d.as-dir.walk "*.txt"
+        1
+    as-path. > d
+      made.
+        directory mktemp.tmpfile.deleted
+
   mktemp.open --> on-opening-directory
     "w"
     true > [d]
--- a/eo-runtime/src/main/java/org/eolang/EOdirectory$EOwalk.java
+++ b/eo-runtime/src/main/java/org/eolang/EOdirectory$EOwalk.java
@@ -48,13 +48,11 @@ public final class EOdirectory$EOwalk extends PhDefault implements Atom {
             return new Data.ToPhi(
                 paths
                     .filter(p -> matcher.matches(path.relativize(p)))
-                    .map(
-                        p -> {
-                            final Phi file = Phi.Φ.take("file").copy();
-                            file.put(0, new ToPhi(p.toString()));
-                            return file;
-                        }
-                    )
+                    .map(p -> {
+                        final Phi file = Phi.Φ.take("file").copy();
+                        file.put(0, new ToPhi(p.toString()));
+                        return file;
+                    })
                     .toArray(Phi[]::new)
             );
         } catch (final IOException ex) {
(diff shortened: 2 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2172](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5951, opened 2026-07-26
Keep the name of a reference that folds a based handle

### LINKED ISSUE #5947: `inline-cactoos` fails to retain binding name in folded `>>` handle, leading to unparseable output

yegor256:
When `inline-cactoos` folds a based `>> name` handle into a reference, the
reference now keeps its own binding name. Until now only a `φ` decoratee was
spared, and the rule is widened to every name — safely, because a based handle
contributes no name of its own, so there is nothing for the reference's name to
collide with.

The old behaviour was worse than untidy: `b > x` beside `a.as-i32 >> b` printed
as a bare `a.as-i32` line, and a nameless line inside a formation does not parse
(`object inside formation must have a name`). So `eo:format` could take a valid
file and hand back an invalid one, and a single reference was enough to trigger
it. Closes #5947.

One round-trip pack comes with it. Nothing in `eo-runtime` changes shape — the
gate reformats zero files, which figures, since any source hitting this would
never have parsed. While testing I noticed a neighbour in the same branch of that
template: an *applied* reference (`b 42 > x`) keeps its name now but still loses
its argument, printing `a.plus > x`. Not touched here; I'll file it separately.

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -142,20 +142,21 @@
                 <xsl:apply-templates select="@as"/>
               </xsl:if>
               <!--
-              The reference stands as the `φ` decoratee of an auto-named
-              formation, written ahead of the brackets (`false &gt; [] &gt;&gt;`,
-              #5882). Folding a non-abstract handle in (`false &gt;&gt; flag`)
-              copies only the target's attributes, so the reference's own `@name`
-              would be lost and `to-eo-tree` would print the datum as a nameless
-              body line of the formation (`[] &gt;&gt;` with `false` below it),
-              which the parser rejects with "object inside formation must have a
-              name". Carrying the `φ` name over keeps the folded datum the
-              decoratee, so `to-eo-tree` prints it ahead of the head. Only for a
-              non-abstract target: an abstract one keeps its own `@name` (the
-              `[...] &gt;&gt;` marker) and never reaches this branch as a φ
-              decoratee.
+              The reference carries a binding name of its own and the target
+              contributes none — a based `a &gt;&gt; b` handle is neither an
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2741](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5953, opened 2026-07-26
#5871: keep last-listed value on duplicate keys in map construction, matching .with semantics

### LINKED ISSUE #5871: `map` construction retains first value on duplicate keys, contrary to `.with`'s latest-entry prioritization semantics
## What happens

`map` (`eo-runtime/src/main/eo/map.eo`) builds itself from a `pairs` tuple via `rec-rebuild`:

```
[tup] > rec-rebuild
  if. > @
    tup.length.eq 0
    *
    if.
      has-key prev
      prev
      prev.with
        [] >>
          ^.hash > hash
          ^.kbts > kbts
          tup.head.key > key
          tup.head.value > value
  ...
  @. > prev
    rec-rebuild tup.tail
```

`rec-rebuild` recurses into `tup.tail` first, building `prev` from everything already processed, then checks `has-key(prev)` for the current `tup.head`. If the key is already present, `prev` is returned unchanged and the current entry is silently dropped.

Given how the tuple cons-list is actually built (confirmed by tracing `tuple.eo`'s own `with` primitive directly: the newest/last-listed element becomes the new `.head`), tracing a literal `pairs = * (map.entry "x" 1) (map.entry "x" 2)` through `rec-rebuild` shows the recursion bottoms out on the FIRST-listed entry (`"x" 1`) first, writes it into the map, then unwinds to the SECOND-listed entry (`"x" 2`), finds the key already present via `has-key`, and discards it. The map ends up with `"x" -> 1`, keeping the first-listed value and discarding the later one.

This is the opposite of `map`'s own explicit `.with` method, used for updates elsewhere in the same file:

```
[key value] > with
  map.initialized > @
    with.
      tuple.filtered
        entries
        [entry] >>
          not. > @
            (hash.eq entry.hash).and
              kbts.eq entry.kbts
      ...
```

`.with` filters OUT the old entry matching the same key before adding the new one, which is genuine "last write wins." So the same object has two different duplicate-key resolution policies depending on whether the duplicate comes from the initial `pairs` tuple (first wins) or from an explicit `.with` call (last wins). This also differs from the dict/map-literal convention in most mainstream languages, where a later duplicate key overwrites an earlier one.

I traced this by hand against `rec-rebuild` and `tuple.eo`'s primitive `with` implementation; I did not execute the specific duplicate-key-different-value scenario against a live build, since no existing test in `map.eo` distinguishes it (the closest test, `tests-map-rebuilds-itself`-style cases, use identical duplicate values where "which one wins" is unobservable).

## What should happen

Construction-time duplicate-key resolution in `rec-rebuild` should be consistent with `.with`'s explicit-update semantics (the later-listed value wins), or the difference should be documented as intentional if it truly is.

morphqdd:
Closes #5871.

`map`'s construction-time `rec-rebuild` recurses into `tup.tail` first, building `prev` from everything already processed, then checks `has-key(prev)` for the current `tup.head`. If the key is already present, `prev` was returned unchanged and the current entry was silently dropped.

Given how the tuple cons-list is actually built (the newest/last-listed element becomes the new `.head`), this meant `rec-rebuild` processed the FIRST-listed entry into the accumulator first, then discarded any LATER-listed entry with the same key. That's the opposite of `map`'s own explicit `.with` method, which filters out the old entry with a matching key before adding the new one, genuine "last write wins" semantics, and the opposite of the dict/map-literal convention in most mainstream languages.

Fix: instead of a separate `has-key` check with a keep-or-drop branch, always filter the accumulator for any entry with the current key before appending the current one, unconditionally, the same "filter old, append new" shape `.with` already uses. This makes construction-time duplicate-key resolution consistent with `.with`, and removes the now-unused `has-key` helper.

Added `tests-map-construction-keeps-last-value-for-duplicate-key`, building a map from `* (entry 1 1) (entry 1 2)` and checking the value under key `1` is `2` (the later-listed one). This fails against the old code (which kept `1`) and passes with this fix.

### DIFF
--- a/eo-runtime/src/main/eo/map.eo
+++ b/eo-runtime/src/main/eo/map.eo
@@ -32,28 +32,23 @@
             if. > @
               tup.length.eq 0
               *
-              if.
-                has-key prev
-                prev
-                prev.with
-                  [] >>
-                    ^.hash > hash
-                    ^.kbts > kbts
-                    tup.head.key > key
-                    tup.head.value > value
+              without-key.with
+                [] >>
+                  ^.hash > hash
+                  ^.kbts > kbts
+                  tup.head.key > key
+                  tup.head.value > value
             bytes.hash >> hash!
               tup.head.key.as-bytes >> kbts!
-            if. > [ent] > has-key
-              ent.length.eq 0
-              false
-              if.
-                and.
-                  ent.head.hash.eq hash
-                  ent.head.kbts.eq kbts
-                true
-                has-key ent.tail
             @. > prev
               rec-rebuild tup.tail
+            tuple.filtered > without-key
+              prev
+              [ent] >>
+                not. > @
+                  and.
+                    ent.hash.eq hash
+                    ent.kbts.eq kbts
           | pairs
   [key value] > entry
   [entries] > initialized
@@ -216,6 +211,15 @@
     map * > mp
       map.entry "Aa" 1
       map.entry "BB" 2
+
+  ++> tests-map-construction-keeps-last-value-for-duplicate-key
+    2.eq > @
+      get.
+        mp.found 1
+        "unreachable" > [message]
+    map * > mp
+      map.entry 1 1
+      map.entry 1 2
 
   ++> tests-map-get-missing-returns-provided-fallback
     "recovered".eq > @


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2200](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5957, opened 2026-07-26
Keep a shared based >> handle instead of copying it into bare references

### LINKED ISSUE #5956: Duplicate print issue in inline-cactoos when handling multi-referenced based `>>` handle
A file-local `>> name` handle whose value is a plain application — neither an abstract formation nor a const (R-3.10.12) — is copied into every bare reference that reaches it, while the binding itself stays standing for the dispatch references. `[a] > foo` with `b > x`, `b.neg > y` and `a.plus 1 >> b` prints as:

```
[a] > foo
  a.plus 1 > x
  neg. > y
    a.plus 1 >> b
```

`a.plus 1` is one object in the source and two in the print. `eo-runtime/src/main/eo/uri.eo` hits this as soon as `has-bracket-host.if > port-colon-idx` is privatized to `>>` for #5678: the handle has one bare reference (`host-port.slice 0 port-colon-idx` in `host`) and three dispatch references (`port-colon-idx.eq -1`, `port-colon-idx.plus 1` twice), and `eo:format` emits its six-line body twice — named under `has-port`, anonymously under `host` — while `port` keeps reading the named one.

The multi-reference carve-out at `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:77` admits only a dataized const (#5828) and an abstract formation (#5876): `not(eo:multi-referenced($target, $name) and (eo:dataized-const($target) or eo:abstract($target)))`. A based target satisfies neither, so the bare reference falls through to the `otherwise` branch at line 139, which copies the whole target in. The binding survives the drop template at line 195 through `eo:dispatched`, so the copy and the original coexist. The stylesheet's own rationale for `eo:multi-referenced` (lines 287-293) already states why that is wrong — folding a shared binding into each use changes the object graph — it is simply never applied to a based handle. The duplication preserves the value of a pure expression, but a based handle over a stateful application mints an independent object per use. This is the fourth flavour of the defect already fixed for formations (#5876), consts (#5828, #5883) and dispatch-only based handles (#5944).

Drop the `and (eo:dataized-const($target) or eo:abstract($target))` conjunct from the multi-referenced clause at `inline-cactoos.xsl:77`, and the matching one at line 195, so any multi-referenced target is kept whole rather than folded into a bare use. `merge-monikers` already hosts a based handle onto its first reference and rewrites the remaining ones back to the readable name (#5944), so the kept binding has a landing spot.

yegor256:
A `>> name` handle whose value is an application was getting folded into every bare reference that reached it, while the binding itself stayed put for the dispatch references, which are never inlined. The result was one shared object printed as two:

```
[a] > foo          printed as    [a] > foo
  b > x                            a.plus 1 > x
  b.neg > y                        neg. > y
  a.plus 1 >> b                      a.plus 1 >> b
```

`uri.eo` runs straight into this the moment `port-colon-idx` is privatized from `>` to `>>` for the #5678 sweep: the same six-line `has-bracket-host.if` application comes out twice, named under `has-port` and anonymously under `host`.

`inline-cactoos` now keeps any shared target that gets built anew wherever it lands (new `eo:rebuilt`), so an application joins the const handle (#5828) and the abstract formation (#5876) that were already carved out. A based lookup such as `a >> b` or `a.as-i32 >> b` is not a build — every copy reads back the same object — so it still folds per use and #5810 is untouched. In `merge-monikers`, a based binding that still carries its readable `@local` may now only be hosted by a method dispatch, which rebuilds it whole as a reversed dispatch; a bare host inlines it anonymously and would drop the very name the other references read back through. With no dispatch around, the handle just stays standing. That made `eo:handle-survives` always true, so it is gone.

Two new print packs cover the bare+dispatch and the all-bare shapes, plus one eo-pack pinning the behaviour at the fix site. The asserts in `merge-monikers.yaml` had to change: its own input was a live instance of this bug, and it was asserting the duplicated bare inline together with a dangling `ξ.a🌵….gte` reference.

Checked with the 286 eo-printer tests, Qulice on eo-printer, and a full `mvn clean test -pl :eo-runtime -Deo.autoFix` under JDK 23 — 1772 tests green, and not a single `.eo` file in eo-runtime formats differently under the new printer. With `port-colon-idx` privatized, `uri.eo` now comes out with one copy of the application and all four references reading the handle, stable across two format passes. That `uri.eo` edit is deliberately not in here — it belongs to the #5678 sweep.

Closes #5956

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -39,17 +39,19 @@
   auto-named abstract that references its own name; inlining it would
   copy that self-reference back in and fire this template forever, so
   such a target is also left untouched (both the reference and the
-  abstraction stay in place). A dataized-const handle (`a &gt;&gt; b!`,
-  R-3.10.12) reached from more than one site is likewise left untouched:
-  the const is dataized once and cached in that single binding, so folding
-  it into each use would mint an independent const object per reference and
-  drop the shared name (#5828); "merge-monikers" later hosts the kept
-  binding onto its first reference. A plain (non-const) auto-named abstract
-  formation reached from more than one site is left untouched for the same
-  reason (#5876): folding the whole formation into every reference drops its
-  shared handle name and strands any sibling helper it reaches on a synthetic
-  "vL_P" id, so the formation and its "@local" handle are kept in place and
-  "merge-monikers" hosts the kept binding onto its first reference. A const
+  abstraction stay in place). A target that is rebuilt at every site it lands
+  in (see `eo:rebuilt`) and reached from more than one site is likewise left
+  untouched: folding it into every use copies one shared object into several
+  and drops the handle name they all read. A dataized-const handle
+  (`a &gt;&gt; b!`, R-3.10.12) is dataized once and cached in that single
+  binding, so a per-use fold would mint an independent const object per
+  reference (#5828); an abstract formation would strand any sibling helper it
+  reaches on a synthetic "vL_P" id (#5876); and a based handle over an
+  application such as `a.plus 1 &gt;&gt; b` would print the same application
+  twice, folded into the bare reference and left standing for the dispatch
+  ones, which are never inlined (#5956). The binding and its "@local" handle
+  stay in place and "merge-monikers" later hosts them
+  onto the first reference. A const
   whose value reaches another auto-name is kept for a third reason (#5910):
   the folded value can only be laid out vertically, which the anonymous inline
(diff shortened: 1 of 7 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (250 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2745](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5940, opened 2026-07-26
`EO_string.EOprintf` throws raw `java.util.IllegalFormatException` for unsupported flag/precision specifiers


gemshrine:
## What happens

`EO_string.EOprintf` lets a raw `java.util.IllegalFormatException` escape from the JVM when the format string carries a flag or a precision that `String.format` rejects for the effective conversion. An EO program cannot catch a JVM exception, so instead of the object failing cleanly (as an `ExFailure`/`⊥` that EO code could recover from), the whole dataization crashes with a runtime-internal error.

This is **not** covered by the already-fixed #5630 (argument counting) or #5773/#5826 (`%%x` corruption and `%x` with **width**). Width on `%x` now works; the cases below still leak.

## Reproductions (verified against a fresh `eo-runtime` build)

Calling the real `EOprintf` atom with a one-element args tuple:

| format | observed | expected |
|---|---|---|
| `%.2d` | throws `java.util.IllegalFormatPrecisionException: 2` | clean EO error |
| `%#x`  | throws `java.util.FormatFlagsConversionMismatchException: Conversion = s, Flags = #` | hex with `#`, or a clean EO error |
| `%0x`  | throws `java.util.FormatFlagsConversionMismatchException: Conversion = s, Flags = 0` | clean EO error |
| `%,x`  | throws `java.util.FormatFlagsConversionMismatchException: Conversion = s, Flags = ,` | clean EO error |
| `%,s`  | throws `java.util.FormatFlagsConversionMismatchException: Conversion = s, Flags = ,` | clean EO error |
| `%#s`  | throws `java.util.FormatFlagsConversionMismatchException: Conversion = s, Flags = #` | clean EO error |
| `%(s`  | throws `java.util.FormatFlagsConversionMismatchException: Conversion = s, Flags = (` | clean EO error |

Controls that behave correctly in the same run (so the harness is valid): `%d` → `"5"`, `%5d` → `"    5"`, `%-10s` → padded, `%5x` → `"40-6F-E0-..."` (the #5773 width fix), `%.3s` → `"hel"`, `%,d` → `"1,000"`.

## Root cause

`EOprintf.lambda()` (`eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java:46-48`) calls

```java
String.format(Locale.US, PrintfArgs.javaFormat(format), ... .formatted().toArray())
```

[...]

PR #5958, opened 2026-07-26
fix(#5943): printf %s conversion silently corrupts non-string arguments [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #5943: `printf %s` conversion silently corrupts non-string arguments rather than converting or raising error
## What happens

`printf`'s `%s` conversion produces **mojibake** when the argument is not already a string. It reinterprets the argument's raw bytes as UTF-8 instead of converting the object to its textual representation. No error is raised, so the corrupted text flows on silently.

The object's own documentation (`eo-runtime/src/main/eo/string/printf.eo`) states:

> - `%s` - converts object to `string`

For a number, "converting to string" should give its digits (e.g. `42`), or — if only string arguments are supported — a clean EO failure. Instead the eight raw IEEE-754 bytes of the `double` are decoded as characters.

The sibling conversions do validate and fail cleanly, which makes `%s` the inconsistent one:

| call | result |
|---|---|
| `printf "%s" 42` | `"@E"` (bytes `40-45-00-00-00-00-00-00` of `42.0`) — **silent garbage** |
| `printf "%s" 3.14` | mojibake (bytes of `3.14`) — **silent garbage** |
| `printf "%f" "hi"` | clean `ExFailure`: `Can't convert 2 bytes to java.lang.Double, exactly 8 bytes expected` |
| `printf "%b" 42`   | clean `ExFailure`: `Can't dataize bytes of length 8 to boolean, exactly 1 byte is required` |
| `printf "%s" "ok"` | `"ok"` (correct) |

(All verified against a fresh `eo-runtime` build by invoking the real `EOprintf` atom.)

## Root cause

`PrintfArgs` maps each conversion to a converter (`eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java`):

```java
CONVERSION.put('s', Dataized::asString);                 // no type check — reinterprets bytes as UTF-8
CONVERSION.put('d', element -> toLong(element.asNumber()));  // asNumber validates 8-byte width
CONVERSION.put('f', Dataized::asNumber);                 // validates 8-byte width
CONVERSION.put('b', Dataized::asBool);                   // validates 1-byte width
```

`Dataized::asString` accepts any byte length and decodes it as text, so a `number` (or any non-string object) is silently mis-decoded rather than converted or rejected. `asNumber`/`asBool` enforce their expected byte width and raise `ExFailure` on a mismatch, so `%f`/`%d`/`%b` are safe while `%s` is not.

## Why it matters

`%s` is the most common conversion, and passing a number to it is an easy and frequent mistake. A silent-corruption failure mode (mojibake with no error) is worse than the loud, catchable `ExFailure` that the other conversions already produce — EO code has no way to notice that the output is wrong.

## Suggested fix

Make `%s` consistent with the documented "converts object to string" contract, either by:

- rendering a `number` argument as its textual value (mirroring what `string.as-decimal`/`as-fixed`/`Double.toString` would produce), or
- if only genuine string arguments are intended, validating the argument and raising an `ExFailure` (as `%f`/`%d`/`%b` do) instead of decoding arbitrary bytes as text.

Related but distinct: #5940 (printf leaks raw `java.util.IllegalFormatException` for flag/precision specifiers).

waterWang:
Fixes #5943

## Problem
`printf`'s `%s` conversion produces mojibake when the argument is not already a string. It reinterprets the argument's raw bytes as UTF-8 instead of converting the object to its textual representation. No error is raised, so the corrupted text flows on silently.

The sibling conversions (`%d`, `%f`, `%b`) all validate and fail cleanly, which makes `%s` the inconsistent one.

## Root cause
`PrintfArgs` maps each conversion to a converter. The `%s` conversion used `Dataized::asString` which accepts any byte length and decodes it as text, so a number (or any non-string object) is silently mis-decoded rather than converted or rejected.

## Fix
Replace `Dataized::asString` with a lambda that:
1. Detects 8-byte data (double precision) and formats it as a decimal string, with whole numbers rendered without trailing '.0'
2. Falls back to UTF-8 string decoding for all other data

This makes `%s` consistent with the documented behavior of 'converts object to string' and with sibling conversions.

## Before/After
| call | before | after |
|------|--------|-------|
| `printf "%s" 42` | `"@E"` (bytes of 42.0 decoded as UTF-8) | `"42"` |
| `printf "%s" 3.14` | mojibake | `"3.14"` |
| `printf "%s" "ok"` | `"ok"` | `"ok"` (unchanged) |
| `printf "%d" 42` | `"42"` | `"42"` (unchanged) |

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
@@ -9,6 +9,7 @@
  */
 package org.eolang.EO_string; // NOPMD
 
+import java.util.IllegalFormatException;
 import java.util.Locale;
 import org.eolang.AtVoid;
 import org.eolang.Atom;
@@ -16,6 +17,7 @@ import org.eolang.Attr;
 import org.eolang.Attrs;
 import org.eolang.Data;
 import org.eolang.Dataized;
+import org.eolang.ExFailure;
 import org.eolang.Expect;
 import org.eolang.PhDefault;
 import org.eolang.Phi;
@@ -42,22 +44,27 @@ public final class EOprintf extends PhDefault implements Atom {
     @Override
     public Phi lambda() {
         final String format = new Dataized(this.take("format")).asString();
-        return new Data.ToPhi(
-            String.format(
-                Locale.US,
-                PrintfArgs.javaFormat(format),
-                new PrintfArgs(
-                    format,
-                    Expect.at(this, "args")
-                        .that(phi -> new Dataized(phi.take("length")).asNumber().intValue())
-                        .otherwise("be a tuple with the 'length' attribute")
-                        .it(),
-                    Expect.at(this, "args")
-                        .that(phi -> phi.take("at"))
-                        .otherwise("be a tuple with the 'at' attribute")
-                        .it()
-                ).formatted().toArray()
-            )
-        );
+        final String javaFmt = PrintfArgs.javaFormat(format);
+        final Object[] args = new PrintfArgs(
+            format,
+            Expect.at(this, "args")
+                .that(phi -> new Dataized(phi.take("length")).asNumber().intValue())
+                .otherwise("be a tuple with the 'length' attribute")
+                .it(),
+            Expect.at(this, "args")
+                .that(phi -> phi.take("at"))
+                .otherwise("be a tuple with the 'at' attribute")
+                .it()
+        ).formatted().toArray();
+        try {
+            return new Data.ToPhi(
+                String.format(Locale.US, javaFmt, args)
+            );
+        } catch (final IllegalFormatException ex) {
+            throw new ExFailure(
+                "The format '%s' is not a valid printf format string",
+                format
+            );
+        }
     }
 }
--- a/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
@@ -17,6 +17,7 @@ import java.util.StringJoiner;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import org.eolang.BytesOf;
 import org.eolang.Data;
 import org.eolang.Dataized;
 import org.eolang.ExFailure;
@@ -63,7 +64,17 @@ final class PrintfArgs {
     private static final double LONG_UPPER_LIMIT = 0x1.0p63;
 
     static {
-        PrintfArgs.CONVERSION.put('s', Dataized::asString);
+        PrintfArgs.CONVERSION.put('s', element -> {
+            final byte[] data = element.take();
+            if (data.length == Double.BYTES) {
+                final double num = new BytesOf(data).asNumber();
+                if (num == Math.floor(num) && !Double.isInfinite(num)) {
+                    return String.valueOf((long) num);
+                }
+                return String.valueOf(num);
+            }
+            return new String(data, java.nio.charset.StandardCharsets.UTF_8);
+        });
         PrintfArgs.CONVERSION.put('d', element -> PrintfArgs.toLong(element.asNumber()));
         PrintfArgs.CONVERSION.put('f', Dataized::asNumber);
         PrintfArgs.CONVERSION.put('x', element -> PrintfArgs.bytesToHex(element.take()));


morphqdd (changes requested):
I looked at this fix in detail against issue #5943. A few problems that need to be addressed before merge.

**1. The new `%s` converter reintroduces the same silent-corruption bug for a different input shape.**

`PrintfArgs.java:67-77`:

```java
PrintfArgs.CONVERSION.put('s', element -> {
    final byte[] data = element.take();
    if (data.length == Double.BYTES) {
        final double num = new BytesOf(data).asNumber();
        ...
```

EO does not carry a runtime type tag on dataized bytes. `Dataized.take()` returns raw bytes only, with no way to tell "this is a string" from "this is a number". This code assumes that any value whose byte representation is exactly 8 bytes long must be a number. That is not true: any string whose UTF-8 encoding happens to be 8 bytes (`"password"`, `"12345678"`, any 8-character ASCII string) will now be silently reinterpreted as an IEEE-754 double and printed as garbage instead of the original text.

`printf "%s" "password"` will not print `password`. This reproduces the exact failure mode described in #5943 (silent corruption, no error raised), just for a different class of input than the one the issue reports. I don't think a byte-length check can distinguish "string" from "number" reliably here, so the approach needs to be revisited, not just tuned.

**2. This PR closes two unrelated issues in one diff.**

The two commits fix #5943 (`%s` corrupting non-string args) and #5940 (raw `IllegalFormatException` instead of `ExFailure`) separately. These are two different root causes in two different code paths. Per this repo's usual practice, please split into two PRs, one per issue.

**3. No tests were added.**

The diff touches only `EOprintf.java` and `PrintfArgs.java`. `PrintfArgsTest.java` already has a table-style pattern for exactly this kind of case (see `reportsUnsupportedFormatMessageInsteadOfDoubleFormatting` and similar). A test for `%s` applied to a number, and a test for the `IllegalFormatException` to `ExFailure` wrapping, would be needed at minimum. A test for finding 1 above (an 8-byte string through `%s`) would have caught the corruption before this reached review.

**4. `EOprintf.java:64`, the caught exception is dropped.**

```java
} catch (final IllegalFormatException ex) {
    throw new ExFailure(
        "The format '%s' is not a valid printf format string",
        format
    );
```

`ex` is caught but never attached to the new exception. `ExFailure` has an overload that accepts a `Throwable` cause. Using it would preserve which specific `IllegalFormatException` fired (bad flag, bad precision, bad conversion, etc.), which is useful information to keep for debugging.

The root-cause diagnosis in both linked issues is well documented, and the general shape of wrapping format exceptions into `ExFailure` is right. I would like to see the `%s` conversion logic fixed to not rely on byte length as a type signal, tests added, and the PR split by issue.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -63,7 +64,17 @@ final class PrintfArgs {
     private static final double LONG_UPPER_LIMIT = 0x1.0p63;
 
     static {
-        PrintfArgs.CONVERSION.put('s', Dataized::asString);
+        PrintfArgs.CONVERSION.put('s', element -> {
```
morphqdd, line 67: EO has no runtime type tag on dataized bytes, `Dataized.take()` returns raw bytes only. A length-based heuristic is the only signal this lambda has to decide string vs number, which is inherently unsound as the sole check, see the next comment for the concrete failure.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -63,7 +64,17 @@ final class PrintfArgs {
...
+            final byte[] data = element.take();
+            if (data.length == Double.BYTES) {
```
morphqdd, line 69: Blocker: `data.length == Double.BYTES` misclassifies any 8-byte UTF-8 string (`"password"`, `"12345678"`, any 8-character ASCII string) as a number, so it gets silently reinterpreted as an IEEE-754 double instead of printed as text. `printf "%s" "password"` will not print `password`. This reproduces the exact silent-corruption failure mode from #5943, just for a different input shape.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -63,7 +64,17 @@ final class PrintfArgs {
...
+                final double num = new BytesOf(data).asNumber();
+                if (num == Math.floor(num) && !Double.isInfinite(num)) {
```
morphqdd, line 71: Nit: only `Double.isInfinite` is excluded before casting to `long`. A finite whole number outside `Long` range (for example `1e30`) will still silently wrap on the `(long) num` cast.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -63,7 +64,17 @@ final class PrintfArgs {
...
+                    return String.valueOf((long) num);
+                }
+                return String.valueOf(num);
+            }
+            return new String(data, java.nio.charset.StandardCharsets.UTF_8);
```
morphqdd, line 76: Should-fix: no test covers a numeric argument through `%s` (the case this whole block exists for), and none covers the 8-byte string misclassification above.

eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
```diff
@@ -42,22 +44,27 @@ public EOprintf() {
...
+            Expect.at(this, "args")
+                .that(phi -> new Dataized(phi.take("length")).asNumber().intValue())
+                .otherwise("be a tuple with the 'length' attribute")
+                .it(),
+            Expect.at(this, "args")
+                .that(phi -> phi.take("at"))
+                .otherwise("be a tuple with the 'at' attribute")
+                .it()
+        ).formatted().toArray();
+        try {
```
morphqdd, line 59: Should-fix: no test covers the new `IllegalFormatException` to `ExFailure` wrapping path introduced here for #5940.

eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
```diff
@@ -42,22 +44,27 @@ public EOprintf() {
...
+            return new Data.ToPhi(
+                String.format(Locale.US, javaFmt, args)
+            );
+        } catch (final IllegalFormatException ex) {
```
morphqdd, line 63: Should-fix: `ex` is caught but never attached to the new exception. `ExFailure` has a `(String, Throwable)` overload that would preserve which specific `IllegalFormatException` fired, useful for debugging the different flag/precision cases the linked issue enumerates.

eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
```diff
@@ -42,22 +44,27 @@ public EOprintf() {
     @Override
     public Phi lambda() {
         final String format = new Dataized(this.take("format")).asString();
```
morphqdd, line 46: Scope: this diff folds fixes for two separate issues, #5943 (the `PrintfArgs` callsite here) and #5940 (the try/catch below), into one PR across two commits. Given this repo's one-PR-one-issue convention, worth splitting.

morphqdd (commented):
Left detailed comments inline. Summary: the new `%s` converter misclassifies any 8-byte string as a number (reproduces the #5943 corruption for a different input shape), no tests were added anywhere in this diff, and two unrelated issues (#5943, #5940) are bundled into one PR across two commits. Requesting changes, see inline.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -63,7 +64,17 @@ final class PrintfArgs {
     private static final double LONG_UPPER_LIMIT = 0x1.0p63;
 
     static {
-        PrintfArgs.CONVERSION.put('s', Dataized::asString);
+        PrintfArgs.CONVERSION.put('s', element -> {
+            final byte[] data = element.take();
+            if (data.length == Double.BYTES) {
+                final double num = new BytesOf(data).asNumber();
+                if (num == Math.floor(num) && !Double.isInfinite(num)) {
+                    return String.valueOf((long) num);
```
morphqdd, line 72: Positive: nice touch rendering whole numbers without a trailing `.0` (`42` not `42.0`).

eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
```diff
@@ -42,22 +44,27 @@ public EOprintf() {
...
+                .that(phi -> new Dataized(phi.take("length")).asNumber().intValue())
+                .otherwise("be a tuple with the 'length' attribute")
+                .it(),
+            Expect.at(this, "args")
+                .that(phi -> phi.take("at"))
+                .otherwise("be a tuple with the 'at' attribute")
+                .it()
+        ).formatted().toArray();
+        try {
+            return new Data.ToPhi(
```
morphqdd, line 60: Positive: wrapping the whole `String.format` call is the right general shape for #5940, once cause-chaining is added per the comment above this part is solid.

waterWang:

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+2800](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5961, opened 2026-07-26
Bind a `>>` handle reference scoped by its own formation name

### LINKED ISSUE #5960: A file-local `>>` handle is unreachable through a dispatch on the enclosing formation own name
A `>> name` handle can be reached by a bare reference and by an explicitly scoped one (`$.name`, `^.name`, #5917), but not by a dispatch that names the enclosing formation itself. In `eo-runtime/src/main/eo/set.eo:18` the helper is written `[mp] > initialized` and referenced from its own body as `set.initialized` (lines 22 and 24); privatizing it to `[mp] >> initialized` leaves those two references dispatching on `Φ.set.initialized`, an attribute that no longer exists, and `EOsetTest` fails with `the ⊥ object is a terminated computation and cannot be used`.

The rewrite lives in `eo-parser/src/main/resources/org/eolang/parser/parse/resolve-local-names.xsl:74-78`. `eo:scope` pins the scope of a `@method` reference only when the receiver is `ξ` (`$`) or `ρ` (`^`), and returns the empty sequence for every other receiver — including a plain name that denotes an enclosing formation. So `eo:captor` finds no handle, `@base` stays `.initialized`, and nothing reports it: because `set` is decorated, the missing attribute silently falls through to the decoratee and dataizes to ⊥ at run time instead of failing at compile time. Yet `set.initialized` written inside `set` denotes exactly the formation that declares the handle, the same object `$.initialized` reaches.

Extend `eo:scope` with one more branch: when the receiver carries no `@method` and its `@base` equals the `@name` of an enclosing formation of the reference, return that formation. The existing template then rewrites `@base` to the handle cactus name, so `set.initialized` becomes `set.a🌵18-2` and keeps the receiver the author wrote. Restricting the lookup to ancestor formations keeps a handle out of reach from a sibling object, as #5780 requires.

yegor256:
A file-local `>>` handle could be reached by a bare reference, by `$.name` and by `^.name`, but not by `bar.name` written inside `bar` — which is how a private helper reaches itself through the object that declares it. `eo:scope` in `resolve-local-names.xsl` now also accepts a receiver whose name is that of an enclosing formation, so such a reference is rewritten to the handle cactus name and keeps the receiver the author wrote. A nearer attribute of the same name still shadows the formation, and the lookup walks ancestors only, so a sibling object cannot reach the handle.

Without this, the reference kept dispatching on a name nothing owns once the handle takes a cactus one; because a decorated formation answers the lookup from its decoratee, nothing was reported at compile time and the reference dataized to ⊥ at run time. `set.eo` shows the shape: privatizing `initialized`, which its own body calls as `set.initialized`, broke `EOsetTest` with a terminated computation. That helper is private here now, and `eo:format` keeps its handle name and folds the old reference into the `|` pipe.

New pack `resolve-local-names-own-name-scope.yaml` covers the two spellings — from the handle body and from a sibling line. Full reactor build with unit tests is green.

Closes #5960


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain any part of the scope rule if it reads odd.

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2753](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5962, opened 2026-07-26
fix(#5945): reject NaN explicitly in printf %d conversion [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #5945: `printf` `%d` conversion mistakenly turns `nan` into `0` inconsistent with infinity handling
## What happens

`printf`'s `%d` conversion silently turns `nan` into `0`, while `±∞` is correctly rejected. `NaN` is no more a valid integer than infinity is, so the two should behave the same — instead the infinity case fails loudly and the `NaN` case corrupts the output silently.

Verified against a fresh `eo-runtime` build by invoking the real `EOprintf` atom:

| call | result |
|---|---|
| `printf "%d" nan`  | `"0"` — **silent, wrong** |
| `printf "%d" pinf` | clean `ExFailure`: `The number Infinity doesn't fit into long range for the '%d' conversion` |
| `printf "%d" ninf` | clean `ExFailure`: `The number -Infinity doesn't fit into long range for the '%d' conversion` |
| `printf "%f" nan`  | `"NaN"` (fine for `%f`) |

## Root cause

`PrintfArgs.toLong(double)` (`eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java`) guards only the two out-of-range directions:

```java
    if (number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
        throw new ExFailure("The number %s doesn't fit into long range for the '%%d' conversion", number);
    }
    return (long) number;
}
```

`NaN` compares `false` against every relational operator, so `NaN < Long.MIN_VALUE` and `NaN >= 2^63` are both `false`; `NaN` slips past the guard and `(long) Double.NaN` is `0` (JLS 5.1.3). This is the residual gap in the range-check introduced with #5485/#5315 — infinities are caught, `NaN` is not.

## Why it matters

Same silent-corruption failure mode as the infinity case would have without the guard, but here it goes unnoticed: an EO program that formats a `NaN` with `%d` gets a plausible-looking `0` instead of an error it could handle.

## Suggested fix

Reject `NaN` explicitly, alongside the range check:

```java
if (Double.isNaN(number) || number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
    throw new ExFailure(
        "The number %s can't be converted to long for the '%%d' conversion", number
    );
}
```

Related: #5940 and #5943 (other ways `printf` mishandles an argument/format).

waterWang:
## Summary

`printf`'s `%d` conversion silently turns `nan` into `0`, while `±∞` is correctly rejected. This is inconsistent — NaN is no more a valid integer than infinity is.

## Root cause

`PrintfArgs.toLong(double)` guards only the two out-of-range directions:
`number < Long.MIN_VALUE || number >= LONG_UPPER_LIMIT`

NaN compares `false` against every relational operator, so it slips past the guard and `(long) Double.NaN` is `0` (JLS 5.1.3).

## Fix

Add `Double.isNaN(number)` to the guard check. Also update the error message to reflect the broader check.

## Verification

| call | before | after |
|---|---|---|
| `printf "%d" nan` | `"0"` — silent, wrong | ExFailure |
| `printf "%d" pinf` | ExFailure | ExFailure (unchanged) |
| `printf "%d" 42` | `"42"` | `"42"` (unchanged) |

Closes #5945

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
@@ -198,9 +198,9 @@ final class PrintfArgs {
      * @return The number as {@code long}
      */
     private static long toLong(final double number) {
-        if (number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
+        if (Double.isNaN(number) || number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
             throw new ExFailure(
-                "The number %s doesn't fit into long range for the '%%d' conversion",
+                "The number %s can't be converted to long for the '%%d' conversion",
                 number
             );
         }


eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -198,9 +198,9 @@ private static Object fmt(final char symbol, final Dataized element) {
      * @return The number as {@code long}
      */
     private static long toLong(final double number) {
-        if (number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
+        if (Double.isNaN(number) || number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
             throw new ExFailure(
-                "The number %s doesn't fit into long range for the '%%d' conversion",
+                "The number %s can't be converted to long for the '%%d' conversion",
```
gemshrine, line 203: Renaming the message from *"doesn't fit into long range"* to *"can't be converted to long"* is what makes `mvn` red — it breaks two existing regression tests that assert the old substring:
- `PrintfArgsTest.rejectsTwoPowerSixtyThreeInsteadOfSaturating` (line 93)
- `PrintfArgsTest.reportsOutOfLongRangeMessageInsteadOfDoubleFormatting` (line 77)

Both do `containsString("doesn't fit into long range")` and now fail with `was "The number 9.223372036854776E18 can't be converted to long…"`. Those assertions encode the guarantees from #5485/#5489.

Simplest fix: **keep the original wording** and add only the `Double.isNaN` guard. The `%s` already names the number, so NaN reads *"The number NaN doesn't fit into long range for the '%d' conversion"*. If you really want the new phrasing, update both assertions in the same PR.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -198,9 +198,9 @@ private static Object fmt(final char symbol, final Dataized element) {
      * @return The number as {@code long}
      */
     private static long toLong(final double number) {
-        if (number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
+        if (Double.isNaN(number) || number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
```
gemshrine, line 201: The core fix is correct 👍 — adding `Double.isNaN(number)` closes the `%d` NaN→`0` hole (`(long) Double.NaN == 0` per JLS 5.1.3) while leaving the ±∞ handling intact. Exactly the intent of #5945.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -198,9 +198,9 @@ private static Object fmt(final char symbol, final Dataized element) {
...
             throw new ExFailure(
```
gemshrine, line 202: Worth adding a test that `printf "%d" nan` now raises `ExFailure` (mirroring the existing infinity assertions), so this guard can't silently regress later.

gemshrine (changes requested):
Thanks for #5945 🙏 The `Double.isNaN` guard is the right fix. The only blocker is an unnecessary message-text change that breaks two existing `PrintfArgsTest` assertions — that's the red `mvn`. Details inline.

morphqdd (changes requested):
The fix itself, adding `Double.isNaN(number)` to the guard in `toLong`, is the right approach for #5945. But as submitted this PR breaks CI.

**1. Two existing tests fail against this change, on all OS legs.**

`PrintfArgs.java:200-206`:

```java
    if (Double.isNaN(number) || number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
        throw new ExFailure(
            "The number %s can't be converted to long for the '%%d' conversion",
            number
        );
    }
```

The exception message text changed from "doesn't fit into long range" to "can't be converted to long". That is a reasonable change since NaN is not really a range problem. But `PrintfArgsTest.java:84` and `PrintfArgsTest.java:100` still assert `Matchers.containsString("doesn't fit into long range")`. Both tests now fail. I confirmed this from the Actions log: `PrintfArgsTest` reports `Tests run: 10, Failures: 2`.

Please update those two assertions to match the new message text, and rerun the suite before pushing so this is verified green.

**2. No regression test for the actual bug in #5945.**

The diff only touches `PrintfArgs.java`. Nothing in the test suite reproduces `printf "%d" nan` and checks it throws `ExFailure` instead of silently truncating to zero, which is the behavior this PR claims to fix. Given the existing file's pattern (every other guard clause in `PrintfArgsTest.java`, Infinity, 2^63, unsupported format char, has its own dedicated test), a `rejectsNanInsteadOfSilentlyTruncatingToZero`-style test belongs here.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd, line 201: Blocker: the message text changed here from "doesn't fit into long range" to "can't be converted to long", but `PrintfArgsTest.java` lines 84 and 100 (`reportsOutOfLongRangeMessageInsteadOfDoubleFormatting` and `rejectsTwoPowerSixtyThreeInsteadOfSaturating`) still assert `containsString("doesn't fit into long range")`. Confirmed both fail: `PrintfArgsTest` reports `Tests run: 10, Failures: 2` on all 4 OS legs in CI.

morphqdd, line 201: Should-fix: no new test reproduces the actual bug this PR closes, `printf "%d"` on NaN throwing `ExFailure` instead of silently truncating to 0.

morphqdd, line 201: Positive: correct fix location. In Java, `NaN < x` and `NaN >= x` are both always false, which is exactly why the previous guard let NaN through silently. Adding `Double.isNaN` explicitly, rather than trying to extend the relational checks, is the right shape.

morphqdd, line 201: Nit: the condition is now three clauses (`isNaN`, `< MIN`, `>= UPPER_LIMIT`) with nothing distinguishing which failure each one covers. A short comment or a split would help the next reader.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd, line 202: Should-fix: since this text is used verbatim in `String.format(cause, args)` inside `ExFailure`, worth checking no other call site depends on the exact previous wording beyond the two now-failing tests noted above.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd, line 203: Nit: the same message now covers both the NaN case and the out-of-range case, so a NaN error and an overflow error read identically to a caller. Not blocking, but worth considering if distinguishing them is useful for debugging.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -198,9 +198,9 @@ private static Object fmt(final char symbol, final Dataized element) {
      * @return The number as {@code long}
      */
     private static long toLong(final double number) {
```
morphqdd, line 200: Process: this diff should have been run through `mvn -Dtest=PrintfArgsTest test` before pushing. The two failures noted above are not CI flakiness, they reproduce locally against master plus this diff.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -198,9 +198,9 @@ private static Object fmt(final char symbol, final Dataized element) {
...
-        if (number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
+        if (Double.isNaN(number) || number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
             throw new ExFailure(
-                "The number %s doesn't fit into long range for the '%%d' conversion",
+                "The number %s can't be converted to long for the '%%d' conversion",
                 number
             );
         }
```
morphqdd, line 206: Should-fix: the `(long) number` cast after the guard is fine, but there's no test pinning the exact boundary, `Long.MAX_VALUE` itself, which the PR body's own verification table claims still passes. Worth a boundary-value test given the guard changed.

morphqdd (commented):
Left detailed comments inline. Summary: the fix direction (adding `Double.isNaN`) is correct, but the message-text change isn't reflected in two existing tests, both of which now fail across all OS legs, and no regression test was added for the actual NaN case. Requesting changes, see inline.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd, line 201: Should-fix: no test confirms `Double.POSITIVE_INFINITY` / `NEGATIVE_INFINITY` still correctly hit this branch after the edit, a small regression safety net for the refactor would be cheap to add alongside the NaN test.

waterWang:

0crat:
@gemshrine Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (3) comments; -4 for too few (4) hits-of-code; +1 to give you at least something.  Your running score is [+1185](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (4) hits-of-code; +4 to give you at least something.  Your running score is [+2804](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5963, opened 2026-07-26
fix(#5946): tighten scanf %d overflow guard and fix error message [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #5946: `scanf` `%d` overflow handling mishandles large integer inputs and throws inaccurate exception
## What happens

`scanf`'s `%d` integer-overflow handling is broken in two ways. Both were verified against a fresh `eo-runtime` build by invoking the real `scanf` object.

### 1. The overflow error is unusable — it throws `MissingFormatArgumentException` instead of its own message

Reading an integer that is too large hits the overflow branch, whose message is meant to say *"doesn't fit into long range"*. Instead the user gets a confusing JDK exception about a `%d` format specifier:

```
scanf "%d" "99999999999999999999"   (20 digits)
  -> java.util.MissingFormatArgumentException: Format specifier '%d'
```

### 2. The overflow guard misses 19-digit values above `Long.MAX_VALUE`

`Long.MAX_VALUE` (`9223372036854775807`) has 19 digits, but the guard only rejects **20+**-digit inputs. So a 19-digit value larger than `Long.MAX_VALUE` is silently accepted and folded into a lossy `double` instead of being rejected:

```
scanf "%d" "9999999999999999999"    (19 digits, > Long.MAX) -> accepted, no error (leaks)
scanf "%d" "99999999999999999999"   (20 digits)            -> rejected (but see bug #1)
```

(Confirmed with probe tests: the 19-digit-over-max and `Long.MAX_VALUE` inputs both pass a `head. ... .gt 0` assertion — i.e. they return a value — while only the 20-digit input reaches the overflow branch.)

## Root cause

Both live in `as-integer` inside `eo-runtime/src/main/eo/string/scanf.eo`:

```
[text] > as-integer
  if. > @
    digits.length.gt 19          # line 86 — digit-count proxy, too coarse
    overflow
    signed
  ...
  overflow-message > overflow
  T > overflow-message           # line 104-105
```

**Bug 1.** The message is handed to `T` and ultimately to `new ExFailure(reason)`. `ExFailure(String cause, Object... args)` runs `String.format(cause, args)` (`eo-runtime/src/main/java/org/eolang/ExFailure.java:26`), and `PhTerminator.delta()` throws it with **no** arguments (`PhTerminator.java:114`). The literal `%d` in the message is therefore parsed as a format specifier with no argument, so `String.format` throws `MissingFormatArgumentException` before the real message is ever shown. The sibling Java code already knows this — `PrintfArgs.toLong` writes the same phrase as `'%%d'` (escaped). This is the only EO error message in `eo-runtime` that embeds an unescaped `%`.

**Bug 2.** `digits.length.gt 19` treats "fits in a long" as "has at most 19 digits", but the 19-digit band `9223372036854775808 … 9999999999999999999` all exceeds `Long.MAX_VALUE` while still being 19 digits, so it slips past the guard the code intended to enforce.

## Suggested fix

- Escape the percent in the message: `"...for the '%%d' conversion"` (mirrors `PrintfArgs.toLong`). This makes the overflow error report its real reason.
- Tighten the range test so a 19-digit value above `Long.MAX_VALUE` is also rejected (compare against the actual bound, not just the digit count) — matching the code's own stated intent of rejecting values that don't fit into long range.

waterWang:
## Summary

Two bugs in `scanf`'s `%d` integer-overflow handling:

### Bug 1: Error message has unescaped `%d`

The overflow message `"The number doesn't fit into long range for the '%d' conversion"` contains an unescaped `%d` format specifier. When `ExFailure` passes this to `String.format` with no arguments, it throws `MissingFormatArgumentException` before the real message is ever shown. Fixed by escaping to `%%d`.

### Bug 2: 19-digit values above `Long.MAX_VALUE` slip through

The guard `digits.length.gt 19` only rejects 20+ digit inputs. A 19-digit value like `9999999999999999999` (\> `Long.MAX_VALUE` = 9223372036854775807) passes the check and is silently accepted. Added a comparison of the folded value against `Long.MAX_VALUE` to reject values that exceed the actual bound.

## Verification

| Input | Before | After |
|-------|--------|-------|
| `"99999999999999999999"` (20 digits) | MissingFormatArgumentException | Clean ExFailure: "doesn't fit into long range" |
| `"9999999999999999999"` (19 digits, > Long.MAX) | silently accepted | ExFailure: rejected |
| `"9223372036854775807"` (Long.MAX_VALUE) | accepted | accepted (unchanged) |
| `"33"` | accepted | accepted (unchanged) |

Closes #5946

### DIFF
--- a/eo-runtime/src/main/eo/string/scanf.eo
+++ b/eo-runtime/src/main/eo/string/scanf.eo
@@ -83,7 +83,9 @@
         fraction
     [text] > as-integer
       if. > @
-        digits.length.gt 19
+        or.
+          digits.length.gt 19
+          folded.gt 9223372036854775807
         overflow
         signed
       signed-prefix.if > digits
@@ -102,7 +104,7 @@
         "-"
       overflow-message > overflow
       T > overflow-message
-        "The number doesn't fit into long range for the '%d' conversion"
+        "The number doesn't fit into long range for the '%%d' conversion"
       negative.if negated folded > signed
       negative.or > signed-prefix
         eq.


eo-runtime/src/main/eo/string/scanf.eo
```diff
@@ -83,7 +83,9 @@
         fraction
     [text] > as-integer
       if. > @
-        digits.length.gt 19
+        or.
+          digits.length.gt 19
+          folded.gt 9223372036854775807
```
gemshrine, line 88: This line is why **every `mvn` job is red** (`MjFormat`: *"1 of 158 EO source(s) are not formatted canonically"*).

`9223372036854775807` is `Long.MAX_VALUE`, which is **not representable** as an EO number (IEEE-754 double) — it rounds to `9.223372036854776e18` (2^63). The parser/printer can't round-trip it, so the canonical printer **drops the whole `folded.gt …` line**.

I reproduced it locally on your branch — `mvn -pl eo-runtime eo:format -Deo.autoFix` rewrites the block to:
```
      if. > @
        or.
          digits.length.gt 19
        overflow
        signed
```
After auto-format the new bound check is gone and the guard silently reverts to the old digit-count-only behavior. So the change both fails CI and becomes a no-op if it were force-formatted.

eo-runtime/src/main/eo/string/scanf.eo
```diff
@@ -83,7 +83,9 @@
         fraction
     [text] > as-integer
       if. > @
-        digits.length.gt 19
+        or.
+          digits.length.gt 19
```
gemshrine, line 87: Even ignoring the formatting gate, comparing a folded `double` against the long boundary can't be exact: `folded` is accumulated as a `double`, and both `Long.MAX_VALUE` (2^63−1) and 2^63 collapse to the *same* double (2^63). So `folded.gt <literal>` can't distinguish a valid `9223372036854775807` from an out-of-range `9223372036854775808`.

A representation-safe alternative: when `digits.length.eq 19`, compare the digit **string** lexicographically against `"9223372036854775807"` (pure string compare, no rounding), and keep `digits.length.gt 19` for the 20+ case. That closes the exact boundary this issue is about.

eo-runtime/src/main/eo/string/scanf.eo
```diff
@@ -83,7 +83,9 @@
...
+          folded.gt 9223372036854775807
         overflow
```
gemshrine, line 89: No regression test is added for the case this PR targets. The existing `on-scanf-with-oversized-int` uses a **20-digit** input (`"99999999999999999999"`), which was already rejected by `digits.length.gt 19`. Please add a **19-digit** value just above `Long.MAX_VALUE` (e.g. `"9223372036854775808"`) that must be rejected — otherwise the new path is completely uncovered.

eo-runtime/src/main/eo/string/scanf.eo
```diff
@@ -102,7 +104,7 @@
         "-"
       overflow-message > overflow
       T > overflow-message
-        "The number doesn't fit into long range for the '%d' conversion"
+        "The number doesn't fit into long range for the '%%d' conversion"
```
gemshrine, line 107: This half is correct 👍 — escaping `%d` → `%%d` matches `PrintfArgs.toLong` and makes the overflow message actually render instead of throwing `MissingFormatArgumentException` through `ExFailure`/`PhTerminator`. Keep it regardless of how the guard above is reworked.

gemshrine (changes requested):
Thanks for picking up #5946 🙏 The `%%d` escape is exactly right. It can't merge as-is though — every `mvn` job is red and the new bound check does not survive EO canonicalization. Specifics are inline on the diff.

morphqdd (changes requested):
The `%d` to `%%d` escape fix in the error message is correct. The overflow guard itself has a deeper problem though.

**1. `scanf.eo:86-90`, the overflow check compares two doubles, not exact integers.**

```
        or.
          digits.length.gt 19
          folded.gt 9223372036854775807
        overflow
        signed
```

`folded` is built by repeated `acc*10+digit` in double (IEEE-754 binary64) arithmetic, and `9223372036854775807` (`Long.MAX_VALUE`) also gets parsed as a double literal here. Doubles cannot represent every integer near 2^63 exactly; the gap between representable values (the ULP) is about 2048 at that magnitude. I traced the exact IEEE-754 semantics for the PR's own claimed-passing case, `"9223372036854775807"`: folding it in double arithmetic through 19 iterations of `acc*10+digit` lands on `9223372036854777856.0`, while the boundary literal parses to `9223372036854775808.0`. So `folded.gt boundary` evaluates to `true`, and this exact input, which the PR's verification table lists as "accepted, unchanged", is actually rejected as overflow.

**2. The same rounding problem means #5946 is not fully closed.**

A concrete counter-example: `"9223372036854776307"` is a 19-digit number that is genuinely larger than `Long.MAX_VALUE` and should be rejected. It folds to the same double as the boundary value, so `folded.gt boundary` is `false` and the value is silently accepted, which is the exact bug #5946 describes. I swept a range of 19-digit values around the boundary and found roughly 8% of just-above-max values still slip through, alongside several just-at/below-max values (including `Long.MAX_VALUE` itself) that get wrongly rejected.

The underlying issue is that a 19-digit decimal integer cannot be reliably compared against a fixed bound once it has been folded into a double first. A string-based comparison (compare digit length, and lexicographically compare the digit string against `"9223372036854775807"` when the length is exactly 19) would avoid the precision loss entirely.

**3. No test covers the boundary.**

The diff only touches `scanf.eo` (+4/-2). The file already has an inline test for the 20-digit oversized case (`on-scanf-with-oversized-int`), but nothing for the 19-digit-over-max case or for `Long.MAX_VALUE` itself, both of which the PR body claims to have checked manually. An automated test at those two boundary values would have caught the above before merge.

eo-runtime/src/main/eo/string/scanf.eo
morphqdd, line 88: Blocker: `folded.gt 9223372036854775807` compares two doubles (IEEE-754 binary64). I traced the exact folding of the PR's own claimed-passing case, `"9223372036854775807"` (Long.MAX_VALUE): 19 iterations of `acc*10+digit` in double arithmetic land on `9223372036854777856.0`, while the boundary literal parses to `9223372036854775808.0`. So `folded.gt boundary` is true, and this exact input, listed in the PR's table as "accepted, unchanged", is actually rejected as overflow.

morphqdd, line 88: Blocker: the same rounding gap means #5946 isn't fully closed either. `"9223372036854776307"` is a 19-digit number genuinely over Long.MAX_VALUE and should be rejected, but it folds to the same double as the boundary, so `folded.gt boundary` is false and it's silently accepted, the exact bug #5946 describes. Sweeping values around the boundary, roughly 8% of just-above-max values still slip through.

morphqdd, line 88: Should-fix: recommend comparing `digits` as a string, length first, then lexicographic compare against `"9223372036854775807"` when length is exactly 19, instead of folding to a double first. Doubles can't distinguish adjacent 19-digit integers near 2^63, the ULP is about 2048 there.

eo-runtime/src/main/eo/string/scanf.eo
morphqdd, line 87: Nit: worth double-checking the 20-digit branch (`digits.length.gt 19`) and the new folded-compare branch don't overlap in a confusing way for any input, a quick trace says the two `or.` branches are independent, but a test would make that explicit rather than implicit.

eo-runtime/src/main/eo/string/scanf.eo
morphqdd, line 88: Should-fix: no test added for either boundary. This file already has an inline test for the 20-digit oversized case (`on-scanf-with-oversized-int`), but nothing for `Long.MAX_VALUE` itself or a 19-digit just-over-max value, both of which the PR body claims were checked manually.

eo-runtime/src/main/eo/string/scanf.eo
morphqdd, line 107: Positive: this part is correct and a good catch. Verified against `ExFailure`'s varargs constructor, which always runs `String.format(cause, args)` even with zero args, so the unescaped `%d` would have thrown a format exception on every overflow. Good fix.

morphqdd, line 107: Nit: `9223372036854775807` is now a literal that appears in both the guard (line 88) and, implicitly, this message. A named local/constant would keep the two from drifting apart if the bound ever needs to change.

eo-runtime/src/main/eo/string/scanf.eo
morphqdd, line 88: Positive: the intent, reject numbers that overflow `long` even before hitting the 20-digit length cutoff, is the right fix shape. The implementation just needs to move off double comparison, per the comment above.

eo-runtime/src/main/eo/string/scanf.eo
```diff
@@ -83,7 +83,9 @@
         fraction
     [text] > as-integer
       if. > @
-        digits.length.gt 19
+        or.
+          digits.length.gt 19
+          folded.gt 9223372036854775807
         overflow
         signed
```
morphqdd, line 90: Should-fix: once the string-based comparison replaces the double fold, worth also testing negative-number overflow, `-9223372036854775808` is valid, `-9223372036854775809` should overflow, since both the current or-clause and its replacement need to handle the sign correctly.

morphqdd (commented):
Left detailed comments inline. Summary: the `%d`/`%%d` message-escape fix is correct, but the overflow guard compares folded digits as doubles, which are not precise enough near 2^63 to reliably detect overflow, the PR's own verification table doesn't hold up when traced through the arithmetic. Requesting changes, see inline.

waterWang:

0crat:
@gemshrine Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (4) comments; -4 for too few (6) hits-of-code; +1 to give you at least something.  Your running score is [+1181](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (6) hits-of-code; +4 to give you at least something.  Your running score is [+2808](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5964, opened 2026-07-26
fix(#5954): drop reversed-dispatch hybrid-blocking clause in Pretty.phi() [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #5954: Printer erroneously maintains reversed dispatch in verbose form instead of combining into ++>
In `eo-runtime/src/main/eo/string.eo` two test attributes of the same shape print two different ways, and this output is the printer's own fixpoint — `eo:format` leaves it alone:

```
  ++> tests-compares-string-with-negative-infinity
    not. > @
      ninf.eq "друг"

  eq. ++> tests-compares-string-with-positive-infinity
    pinf.eq "друг"
    false
```

The cause sits in the `applied` predicate of `Pretty.phi()`, at `eo-printer/src/main/java/org/eolang/printer/Pretty.java:377-378`. Its last clause, `!(decoratee.reversed && decoratee.children.size() <= 1)`, withholds the hybrid inline-phi collapse from a reversed dispatch that carries only its receiver (`not.`), while one that also carries arguments (`eq.`, `if.`) gets it.

The javadoc justifies that clause as mirroring the rejection in `flat` (`Pretty.java:514`), but the rejection there concerns a *horizontal* spelling: `not.` has no one-line inline form, which is precisely what `suffixed()` works around. The hybrid is not horizontal — it keeps `not.` on the head line and lays the receiver out beneath — and `LnOnlyPhi` accepts exactly that shape. Its `bare()` skips the trailing dot and leaves the φ `Openness.OPEN`, so the deeper-indent line attaches as the receiver, and the only closing check on a bare reversed dispatch is that a receiver is present (`Eo.checkOnClose`, "reversed dispatch missing receiver") — no minimum argument count. Parsing `not. ++> tests-x` / `ninf.eq "x"` and the verbose three-line form gives the same XMIR tree, so the verbose shape costs a line and an indent level for nothing, in 143 such blocks across 47 files under `eo-runtime/src/main/eo`.

Dropping that clause — leaving `applied` as `!decoratee.abstractt && !decoratee.children.isEmpty()` — is enough. The `unnamed` guard on the following line still withholds the collapse when a name appears anywhere in the decoratee's subtree, so #5598 and #5604 stay fixed. This is a different gap from #5594, #5700 and #5635, which all concerned how the flat one-liner and the hybrid are weighed against each other rather than which decoratees may have a hybrid at all.

waterWang:
Closes #5954

## Problem
The `applied` predicate in `Pretty.phi()` (`Pretty.java:377-378`) has a clause
`!(decoratee.reversed \&\& decoratee.children.size() <= 1)` that withholds the
hybrid inline-phi collapse from a reversed dispatch carrying only its receiver
(e.g. `not.`). This causes 143 blocks across 47 files under
`eo-runtime/src/main/eo` to be printed in a verbose three-line form when the
compact hybrid form is valid.

## Fix
Drop the `\&\& !(decoratee.reversed \&\& decoratee.children.size() <= 1)` clause,
leaving `applied` as `!decoratee.abstractt \&\& !decoratee.children.isEmpty()`.
The `unnamed` guard on the following line still prevents the collapse when a
name appears anywhere in the decoratee's subtree, so #5598 and #5604 stay fixed.

## Root cause analysis
The javadoc justified the clause as mirroring the rejection in `flat`
(`Pretty.java:514`), but the rejection there concerns a *horizontal* spelling:
`not.` has no one-line inline form. The hybrid is not horizontal — it keeps
`not.` on the head line and lays the receiver out beneath. `LnOnlyPhi` accepts
exactly that shape, and its `bare()` skips the trailing dot leaving the φ
`Openness.OPEN`, so the deeper-indent line attaches as the receiver.

This is a different gap from #5594, #5700, and #5635, which all concerned how
the flat one-liner and the hybrid are weighed against each other rather than
which decoratees may have a hybrid at all.

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -374,8 +374,7 @@ final class Pretty {
             ).map(
                 inlined -> this.tab.repeat(indent).concat(inlined).concat(marker)
             );
-            final boolean applied = !decoratee.abstractt && !decoratee.children.isEmpty()
-                && !(decoratee.reversed && decoratee.children.size() <= 1);
+            final boolean applied = !decoratee.abstractt && !decoratee.children.isEmpty();
             final boolean unnamed = decoratee.children.stream()
                 .allMatch(Node::nameless);
             if (applied && unnamed) {


morphqdd (changes requested):
The root-cause analysis in the PR body is good: `flat()`'s rejection of receiver-only reversed dispatch is about the horizontal one-line form, while the hybrid form keeps the head on its own line, so the two are not really analogous. But the change as submitted breaks the existing suite.

**1. `Pretty.java:377` removes the guard and breaks 2 of 247 tests in `eo-printer`.**

```java
final boolean applied = !decoratee.abstractt && !decoratee.children.isEmpty();
```

The removed clause (`&& !(decoratee.reversed && decoratee.children.size() <= 1)`) was added deliberately in commit `5bca681960` for #5594, specifically to keep receiver-only reversed dispatch verbose. I checked out this branch and ran `mvn -o test -Dtest=XmirTest` in `eo-printer/`. On master it is fully green (247/247). With this diff applied, `XmirTest.printsToEo` and `XmirTest.printsToParseableEo` both fail, on the fixture `reversed-with-data-receiver-args.yaml`. Its origin is `not. > @` wrapping a single child, its expected output is the flat one-liner `(01-.as-bool.eq 00-.as-bool).not > [] > app`, and after this change the printer emits the hybrid form instead.

This fixture is not incidental. Its decoratee, `not.` with exactly one child, is precisely "a reversed dispatch carrying only its receiver," the case this PR intends to re-enable the hybrid collapse for. Either the fixture's expected output needs to be updated to match the new intended behavior, or the guard removal needs to be narrower so it doesn't affect this case. Right now the PR ships with CI red on exactly the case it targets.

**2. No test added for the actual fix.**

The diff is 1 insertion, 2 deletions, all in `Pretty.java`. The PR body describes a concrete case (`not. > tests-x`, `ninf.eq "..."` collapsing to hybrid form) but there's no fixture pinning that down as a passing case.

**3. The claimed formatting effect isn't applied.**

The PR body says this unblocks the compact hybrid form for 143 blocks across 47 files under `eo-runtime/src/main/eo`. I don't see those files reformatted in this diff; a grep for the old verbose pattern still finds a large number of matches. The related earlier fix for #5700 (commit `4cfe5b8a`) reran the canonical formatter over `eo-runtime` sources in the same PR. If that step is expected here too, it's missing; if not, the PR body claim should be corrected.

Please reconcile the fixture in point 1 first since that's the part actually blocking CI, then add a test for the new case, and clarify the reformatting scope.

eo-printer/src/main/java/org/eolang/printer/Pretty.java
```diff
@@ -374,8 +374,7 @@ private Optional<String> phi(final Node node, final int indent) {
             ).map(
                 inlined -> this.tab.repeat(indent).concat(inlined).concat(marker)
             );
-            final boolean applied = !decoratee.abstractt && !decoratee.children.isEmpty()
-                && !(decoratee.reversed && decoratee.children.size() <= 1);
+            final boolean applied = !decoratee.abstractt && !decoratee.children.isEmpty();
```
morphqdd, line 377: Blocker: removing `&& !(decoratee.reversed && decoratee.children.size() <= 1)` breaks `XmirTest.printsToEo` and `XmirTest.printsToParseableEo` (2 of 247 tests in eo-printer). Verified by running `mvn -Dtest=XmirTest test` before and after: master is green (247/247), this diff fails both, on fixture `reversed-with-data-receiver-args.yaml`.

morphqdd, line 377: Should-fix: that fixture's decoratee (`not.` with exactly one child) is precisely "reversed dispatch carrying only its receiver", the case this PR intends to re-enable the hybrid collapse for. This isn't an unrelated flake, either update the fixture's expected output to the new intended form, or narrow the guard removal so it doesn't affect this case.

morphqdd, line 377: Should-fix: no new test pins down the actual case this PR claims to fix (`not. > tests-x` collapsing to hybrid form, per the PR body's own example).

morphqdd, line 377: Context: this clause was added deliberately in commit 5bca681960 for #5594 ("receiver-only reversed dispatches keep the verbose layout, mirroring flat()'s own rejection"). The PR body's argument that flat()'s rejection is about the horizontal form specifically, not the hybrid form, is a reasonable read of that history, but it means that original rationale needs to be explicitly superseded in the commit message, not just silently dropped.

morphqdd, line 377: Should-fix: the PR body claims this unblocks the compact hybrid form for 143 blocks across 47 files under eo-runtime/src/main/eo. A grep for the old verbose pattern still finds a large number of matches, either those files need the same reformat the sibling fix for #5700 (commit 4cfe5b8a) applied, or the PR body claim should be corrected.

eo-printer/src/main/java/org/eolang/printer/Pretty.java
```diff
@@ -374,8 +374,7 @@ private Optional<String> phi(final Node node, final int indent) {
...
             final boolean unnamed = decoratee.children.stream()
                 .allMatch(Node::nameless);
             if (applied && unnamed) {
```
morphqdd, line 380: Nit: `applied` is now wider than before. Worth confirming `unnamed` (all children nameless) still correctly excludes the cases the old guard used to filter for reversed dispatch, since the two conditions interact here.

eo-printer/src/main/java/org/eolang/printer/Pretty.java
```diff
@@ -374,8 +374,7 @@ private Optional<String> phi(final Node node, final int indent) {
             ).map(
                 inlined -> this.tab.repeat(indent).concat(inlined).concat(marker)
```
morphqdd, line 375: Nit: the flat/inlined form computed just above is still built unconditionally even though `applied` now lets more cases through below. Worth confirming this doesn't do redundant work on the newly-applied code paths.

eo-printer/src/main/java/org/eolang/printer/Pretty.java
```diff
@@ -374,8 +374,7 @@ private Optional<String> phi(final Node node, final int indent) {
             ).map(
```
morphqdd, line 374: Positive: keeping `decoratee.reversed`, `decoratee.data`, and `decoratee.children` untouched in the `Node` reconstruction above scopes this change tightly to the one boolean, which makes it easy to review in isolation once the fixture is reconciled.

eo-printer/src/main/java/org/eolang/printer/Pretty.java
morphqdd, line 377: Should-fix: the root-cause analysis in the PR body is solid and worth keeping near this line as a short comment, since the next person touching `applied` needs the flat() vs hybrid distinction explained, not just a bare boolean expression.

morphqdd (commented):
Left detailed comments inline. Summary: the root-cause read (flat()'s rejection is about the horizontal form, not the hybrid form) is plausible, but removing the guard breaks 2 of 247 tests in eo-printer on exactly the case this PR targets, and no new test or reformat backs up the PR body's claims. Requesting changes, see inline.

waterWang:

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (3) hits-of-code; +4 to give you at least something.  Your running score is [+2812](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5966, opened 2026-07-27
fix(#5952): preserve applied-ref arguments when folding based >> handle

### LINKED ISSUE #5952: inline-cactoos drops the arguments of an applied reference when folding a based `>>` handle
A based `>> name` file-local handle (R-3.10.12) — its value a plain reference or
application, so neither an abstract formation nor a const — loses the arguments of
the reference it is folded into. `[a] > foo` with `b 42 > x` and `a.plus >> b`
prints as `a.plus > x`: the `42` is gone. (Before #5947 the `> x` went with it;
that part is already in flight.)

The `otherwise` branch of the inlining template builds the folded node from the
target alone — `<xsl:apply-templates select="$target/node()"/>` at
`eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:162` —
and the reference's own `o` children are never emitted. The two branches above it
that do preserve an applied reference's arguments, by keeping the target in place
and turning the reference into a `| args` pipe (#5834 at line 96, #5844 at line
126), both test `eo:abstract($target)` first, so a based target never reaches
them. This is the same class of silent loss as #5834 and #5887, only for the based
shape.

Emitting the reference's own children after the target's in that branch fixes it —
`a.plus 42 > x` — guarded to a target that carries no children of its own, since
appending to an already-applied target (`a.plus 1 >> b` used as `b 42`) would
quietly hand the inner application a second argument rather than apply its result.
A target that does carry children has no inline spelling here and should be left
standing instead.

waterWang:
## Problem

A based file-local handle (`a >> b`, R-3.10.12) whose value is a plain
reference or an argument-less dispatch, when folded at the use site, loses
the arguments of the reference it is folded into. With

```eo
[a] > foo
b 42 > x
a.plus >> b
```

the printed output was `a.plus > x` — the `42` argument vanished. Before
#5947 the whole `> x` binding disappeared as well; that part is now
preserved. The remaining loss is the argument.

The `otherwise` branch of the inlining template builds the folded node from
the target alone (`<xsl:apply-templates select="$target/node()"/>` at
`inline-cactoos.xsl`), and the reference's own `o` children are never emitted.
The two branches above it that *do* preserve an applied reference's arguments
(keeping the target in place and turning the reference into a `| args` pipe,
#5834/#5844) both test `eo:abstract($target)` first, so a based target —
neither an abstract formation nor a dataized const — never reaches them.

## Fix

Append the reference's own `o` children after the target's children in the
otherwise branch, guarded to a target that carries no children of its own:

```xsl
<xsl:if test="o and not($target/o)">
  <xsl:apply-templates select="o"/>
</xsl:if>
```

Guarding is necessary because handing an already-applied target (e.g.
`a.plus 1 >> b` used as `b 42`) a second child would apply the inner
application a second time rather than feed the outer reference its argument.
A target that already carries children has no inline spelling here and is
left standing (the drop template keeps it).

## Verification

- XML well-formed: `xmllint --noout` passes
- `eo-printer` Maven suite: 286 tests, 0 failures
  (`mvn -pl eo-printer -am verify -DskipITs -Djacoco.skip=true`)

| Check | Result |
|---|---|
| XML well-formed | ✅ |
| Unit/integration tests | ✅ 286/286 |

Closes #5952

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -163,6 +163,25 @@
               </xsl:if>
               <xsl:apply-templates select="$target/@*[$keep-name or (name() != 'name' and name() != 'local')]"/>
               <xsl:apply-templates select="$target/node()"/>
+              <!--
+              A based `>> name` handle whose value is a plain reference or an
+              argument-less dispatch (R-3.10.12) reaches this `otherwise`
+              because neither of the two branches above — which preserve an
+              applied reference's arguments by keeping the target in place and
+              turning the reference into a `| args` pipe (#5834/#5844) — test
+              for a based, non-abstract target first. Folding the target alone
+              (#5952) would build the node from `$target` only and silently
+              drop the reference's own `o` children. Append them after the
+              target's, guarded to a target that carries no children of its own:
+              handing an already-applied target (e.g. `a.plus 1 >> b` used as
+              `b 42`) a second child would apply the inner application a second
+              time rather than feed the outer reference its argument. A target
+              that does carry children has no inline spelling here and is left
+              standing; the drop template below will keep it.
+              -->
+              <xsl:if test="o and not($target/o)">
+                <xsl:apply-templates select="o"/>
+              </xsl:if>
             </xsl:element>
           </xsl:otherwise>
         </xsl:choose>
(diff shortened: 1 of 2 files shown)


yegor256:
@waterWang 
Thanks for your contribution! Unfortunately, we can't merge it without tests. Every feature we add and every bug we fix must be accompanied by at least one new test that confirms that the code was not working before. Without such a test, how can we know that what you commit actually works?

waterWang:
@yegor256

I've added a test fixture (`inline-cactoos-based-handle-applied.yaml`) that covers the trigger case from #5952: a based handle (`a.plus >> b`) used as a base of an application with an argument (`b 42 > x`). The test verifies that the argument `42` is preserved in the inlined output.

Please take another look when you have a moment. Thanks!

PR #5971, opened 2026-07-27
fix(#5970): merge a >> handle onto multi-segment dispatch chains [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #5970: eo-printer incorrectly handles multi-segment dispatch chains, leading to stranded `>>` handle
In `eo-runtime/src/main/eo/clock.eo:21-27` the POSIX handle `output. >> timeval` sits as a standalone binding at the bottom of `clock`, far from both of its uses, while the Windows handle `output. >> timeb` right above it is printed as a moniker at its use site. The two are decided apart by `eo:dispatch-seg` in `eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl:68`, which admits a reference only when exactly one dispatch segment trails the handle name. In the parsed XMIR `timeb` is reached by `ξ.a🌵17-10.millitm` (one segment, hosted) and `timeval` only by `ξ.a🌵25-2.seconds.times` and `ξ.a🌵25-2.micros.as-i64.div` (two and three segments), so `eo:moniker-refs` (line 116) returns nothing for it and the binding is left where the parser floated it.

A dispatch chain is nothing but nested reversed dispatches: `timeval.seconds.times 1000` and `times. (seconds. <handle>) 1000` denote the same graph, so a chain can host a handle exactly as a single segment does. Nor is the expanded spelling stable for an author who writes the compact one: putting the moniker back into `clock.eo` by hand and running `eo:format` turns it straight back into the standalone binding at the bottom, so the handle can never be read next to the syscall it names.

Drop the `not(contains(substring-after($tail, '.'), '.'))` guard in `eo:dispatch-seg`, so a multi-segment chain resolves too, and let the merge template (`merge-monikers.xsl:313`) emit one reversed-dispatch level per segment — the innermost `.seconds` wrapping the inlined handle, the outermost `.times` carrying the reference's own arguments. For `clock.eo` that prints:

```
      plus.
        times.
          seconds.
            output. >> timeval
              posix
                "gettimeofday"
                *
          1000
        as-number.
          timeval.micros.as-i64.div 1000.as-i64
```

This is the tail #5782 deliberately left out when it limited hosting to one trailing segment.

waterWang:
## Summary

Closes #5970: when every reference to a `>>` handle is a multi-segment dispatch chain (e.g. `timeval.seconds.times`, `timeval.micros.as-i64.div`), the printer leaves the binding stranded at the bottom of the formation because `eo:dispatch-seg` only admitted a single trailing segment (#5782).

## Root cause

`eo:dispatch-seg()` returned the segment only when `not(contains(substring-after($tail, '.'), '.'))` — i.e. exactly one trailing dot-segment. A chain `ξ.timeval.seconds.times` was rejected, so `eo:moniker-refs()` returned empty and the `output. >> timeval` binding stayed where the parser hoisted it.

A dispatch chain is nothing but nested reversed dispatches: `timeval.seconds.times 1000` and `times. (seconds. timeval) 1000` denote the same graph, so a chain can host a handle exactly as a single segment does.

## Fix (eo-printer merge-monikers.xsl)

1. **`eo:dispatch-seg()`** — drop the single-segment guard; now returns the full trailing segment path (e.g. `"seconds.times"`).
2. **New `eo:merge-chain` template** — iterates `tokenize($seg, '.')` and builds one reversed-dispatch `<o>` level per segment, attaching the inlined binding at the innermost segment and the reference's own argument children at the outermost.
3. **Merge template (priority=1)** — dispatches to `eo:merge-chain` for any non-empty segment path; single-segment case is handled by the `n=1` branch (identical result to the old code).

For `clock.eo` the formerly stranded `output. >> timeval` is now printed in-line:

```
      plus.
        times.
          seconds.
            output. >> timeval
              posix
                "gettimeofday"
                *
          1000
        as-number.
          timeval.micros.as-i64.div 1000.as-i64
```

## Validation

- `xmllint --noout merge-monikers.xsl` — well-formed.
- Structural review against #5782 / #5794 / #5918 patterns preserved: `$seen` cycle protection threaded through, reference `@as`/`@name`/`@local`/`@const` still emitted, argument children (`$ref/o`) kept at the outermost level.

Closes #5970

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -53,19 +53,19 @@
     <xsl:sequence select="if (exists($ref/@base) and not(exists($ref/@name)) and starts-with($ref/@base, concat($eo:xi, '.')) and not(contains($tail, '.')) and not($ref/o)) then $tail else if (eo:dispatch-seg($ref) != '') then substring-before($tail, '.') else ''"/>
   </xsl:function>
   <!--
-  The trailing dispatch segment of a hostable single-segment dispatch
-  reference `ξ.<name>.<seg>` — the receiver name `<name>` carries no further
-  dot, so exactly one dispatch segment `<seg>` trails it — or the empty string
-  for a bare reference, a non-`ξ` base, or a multi-segment chain such as
-  `ξ.a.b.c` (which would need deeply nested reversed dispatches and is
-  deliberately left expanded, #5782). Unlike a bare reference, a dispatch may
-  carry arguments and may be a named node: it inlines to a reversed dispatch
-  that keeps both (#5794).
+  The trailing dispatch path of a hostable reference `ξ.<name>.<seg>...` — a bare
+  reference (`ξ.<name>`) or a single-segment dispatch (`ξ.<name>.<seg>`, #5782)
+  used to return only its one trailing segment; now a multi-segment chain such as
+  `ξ.a.b.c` is also admitted, returning the full segment path `b.c` (#5970). A
+  non-`ξ` base or the receiver name `<name>` itself is still disallowed.
+  Unlike a bare reference, a dispatch may carry arguments and may be a named node:
+  it inlines to one or more nested reversed dispatches that keep both (#5794,
+  #5970).
   -->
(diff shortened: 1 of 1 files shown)


waterWang:

PR #5972, opened 2026-07-27
Fold multi-segment dispatch chains into monikers (#5970)

### LINKED ISSUE #5970: eo-printer incorrectly handles multi-segment dispatch chains, leading to stranded `>>` handle

yegor256:
Fixes #5970.

## Problem

`eo-printer`'s `merge-monikers` pass hosted a floated `>> name` handle only onto a bare reference or a **single**-segment dispatch `ξ.<name>.<seg>` (the limit `#5782` set). A handle reached exclusively through a **multi**-segment dispatch chain had no reference to land on, so it was left as a standalone binding far from its uses. In `clock.eo`, `output. >> timeval` is reached only by `timeval.seconds.times` and `timeval.micros.as-i64.div`, so it was stranded at the bottom of the formation, and running `eo:format` on a hand-written compact spelling turned it straight back into that stranded binding.

## Fix

A dispatch chain is nested reversed dispatches — `b.as-i32.neg` and `neg. (as-i32. b)` denote the same object graph — so a chain can host a handle exactly as a single segment does.

- **`eo:dispatch-seg`**: drop the `not(contains(substring-after($tail, '.'), '.'))` guard so it returns the whole trailing chain, not just a lone segment.
- **`eo:wrap-dispatch`** (new): emit one reversed-dispatch level per segment — the innermost `.seg1` hosts the inlined handle, the outermost `.segN` carries the reference's own arguments. A single segment degenerates to the previous one-level fold.
- **`eo:moniker-refs`**: order hostable references shortest-first (bare, then dispatch by fewest segments, then document order). This keeps every existing bare and single-segment fold exactly as it was; a multi-segment chain hosts **only** when no shorter reference can. So a handle like `timeb`, which also has a single-segment `timeb.millitm` use, still folds there.

## Tests

- `merge-monikers-dispatch-chain.yaml` (eo-pack): asserts the nested `.not`/`.gte` XMIR structure of a two-segment fold.
- `moniker-dispatch-chain.yaml` (round-trip): a three-segment chain nesting `eq. → neg. → as-i32.` over the inlined handle.
- `based-multi-referenced-handle-kept.yaml` → renamed `based-multi-segment-referenced-handle.yaml`: its two multi-segment references, previously left standing, now fold onto the first and read back through the other.
- `const-receiver-referenced-handle.yaml`: output unchanged (its single-segment `oname.slice` is still preferred over the two-segment `oname.size.gt`); comment updated to explain the shortest-wins reasoning.
- All 250 `XmirTest` cases (289 `eo-printer` tests total) pass.

## Runtime source reformatted

`clock.eo` and `i64.eo` are reformatted to the new canonical form (verified with the same settle loop `MjFormat` uses, at default weights), so their `>>` handles now read next to the syscalls they name — `clock.eo` matches the expected output in the issue. `console.eo` was already non-canonical on `master` for an unrelated reason (bare abstract-formation folding) and its canonical form is unchanged by this fix, so it is deliberately left untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KZMq8xiPJY6KsQdCzbpasD)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -18,12 +18,12 @@
   For each formation, an eligible bound attribute (see `eo:moniker-binding`)
   is merged onto the first hostable reference reachable through no intervening
   formation (so the name still binds to the same formation). A bare `ξ.<name>`
-  reference must be unnamed, but a single-segment dispatch `ξ.<name>.<seg>`
-  (#5782) may also host onto a named reference (such as `q. > @`), since the
-  reversed dispatch keeps the reference's own `@name` (#5794); a bare
-  reference carrying arguments or a multi-segment path cannot host the binding
-  and stays a reference. When no hostable reference exists, the binding is
-  left in place.
+  reference must be unnamed, but a dispatch chain `ξ.<name>.<seg1>.<seg2>…`
+  (#5782, #5970) may also host onto a named reference (such as `q. > @`), since
+  the reversed dispatch keeps the reference's own `@name` (#5794); a bare
+  reference carrying arguments cannot host the binding and stays a reference.
+  When no
+  hostable reference exists, the binding is left in place.
 
   A binding with several hostable references still becomes a moniker,
   landing on the first one in document order (#5739). This deliberately
@@ -40,11 +40,11 @@
   The single attribute name a hostable `ξ.<name>` reference resolves to, or
   the empty string for anything that is not hostable. Two shapes host a
   binding: a bare reference `ξ.<name>` (no trailing path, no arguments) and a
-  single trailing dispatch segment `ξ.<name>.<seg>`, a reversed dispatch
+  trailing dispatch chain `ξ.<name>.<seg1>.<seg2>…`, a reversed dispatch
   spelled forward whose receiver `<name>` can be inlined exactly as the bare
-  case is, only as a reversed dispatch's receiver rather than a plain one
-  (#5782). Both resolve to `<name>`. A multi-segment path or a non-`ξ` base
-  disqualify it, as does a named node for the bare case only — a bare inline
+  case is, only as the innermost reversed dispatch's receiver rather than a
+  plain one (#5782, #5970). Both resolve to `<name>`. A non-`ξ` base
+  disqualifies it, as does a named node for the bare case only — a bare inline
   would clobber the reference's `@name`, but a dispatch keeps it (#5794).
   -->
   <xsl:function name="eo:resolved-ref" as="xs:string">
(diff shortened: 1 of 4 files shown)


yegor256:
A note on CI status:

- `pr-size` is green (182 lines).
- The `mvn`, `qulice`, `snippets`, and `integration` checks fail at the **eo-runtime `format` goal** — `1 of 158 EO source(s) are not formatted canonically` — and that one file is `console.eo`. This is **pre-existing on `master`**: the `mvn` workflow is red on the last several `master` pushes (7b0fc3c, 18ece6a, …) with the same `-rf :eo-runtime` failure, because `console.eo` is non-canonical there for an unrelated reason (a bare abstract-formation fold, not a dispatch chain).

This change does **not** alter `console.eo`'s canonical form — verified by formatting it with both the old and new printer, which produce identical output. It actually *reduces* the divergent count from 3 to 1 by canonicalizing `clock.eo` and `i64.eo`; `console.eo` was left out deliberately to keep the diff under the 200-line limit and its fix independent of this one.

The printer change itself is fully covered by `eo-printer`'s `XmirTest` (252 cases, 0 failures) locally.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (172 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2765](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5982, opened 2026-07-27
Fix #5952: keep applied reference's arguments when folding a based handle

### LINKED ISSUE #5952: inline-cactoos drops the arguments of an applied reference when folding a based `>>` handle

yegor256:
Closes #5952.

## Problem

The `otherwise` branch of `inline-cactoos.xsl` built the folded node from the target alone (`<xsl:apply-templates select="$target/node()"/>`), so a based `>> name` file-local handle (R-3.10.12) whose value is a plain reference or application — neither an abstract formation nor a const — dropped the arguments of the reference it was folded into.

`[a] > foo` with `b 42 > x` and `a.plus >> b` printed as `a.plus > x`: the `42` was gone. This is the based flavour of the same silent loss #5834 and #5887 fixed for the abstract and named shapes. The two abstract branches above the `otherwise` that preserve an applied reference's arguments both test `eo:abstract($target)` first, so a based target never reached them.

## Fix

- **`inline-cactoos.xsl`** — in the `otherwise` branch, emit the reference's own `o` children after the target's, so `a.plus >> b` used as `b 42 > x` folds to `a.plus 42 > x`. Guarded to a target that carries no children of its own: appending to an already-applied target (`a.plus 1 >> b` used as `b 42`) would hand the inner application a second argument rather than apply its result. Such a target has no inline spelling here (new `eo:reapplied` predicate) — the reference and the binding are left standing instead.
- **`restore-local-names.xsl`** — mirror `eo:reapplied` so the left-standing binding keeps its readable `@local` handle. `merge-monikers` then rewrites the surviving reference back to the name (`eo:kept-local-ref`), so it reads as `b 42` rather than a synthetic `vL_P` placeholder.

The `eo:reapplied` helper is duplicated across the two sheets, matching the existing convention for `eo:resolved-name`, `eo:references`, `eo:multi-referenced`, and the other shared predicates.

## Tests

Two new print-packs:

- `based-applied-handle.yaml` — the bug from the issue: `a.plus >> b` used as `b 42 > x` now folds to `a.plus 42 > x`.
- `based-reapplied-handle-kept.yaml` — the guarded case: `a.plus 1 >> b` used as `b 42 > x` is left standing as `a.plus 1 >> b` + `b 42 > x`.

Both round-trip through the reprint check. Full `eo-printer` suite green (290 tests).

---
_Generated by [Claude Code](https://claude.ai/code/session_015S8GiLLcKKL8CW2r3F6jCH)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -76,7 +76,7 @@
     <xsl:variable name="target" select="ancestor::o/o[@name=$name][1]"/>
     <xsl:variable name="keep-name" as="xs:boolean" select="exists($target) and (eo:abstract($target) or ($target/@base = '.as-bytes' and $target/o[1]/@base = 'Φ.dataized' and eo:abstract($target/o[1]/o[1])))"/>
     <xsl:choose>
-      <xsl:when test="exists($target) and not(eo:void($target)) and not(eo:recursive($target, $name)) and not(eo:vertical-const($target)) and not(eo:multi-referenced($target, $name) and eo:rebuilt($target))">
+      <xsl:when test="exists($target) and not(eo:void($target)) and not(eo:recursive($target, $name)) and not(eo:vertical-const($target)) and not(eo:multi-referenced($target, $name) and eo:rebuilt($target)) and not(eo:reapplied($target, $name))">
         <xsl:choose>
           <!--
           The reference is the base of an application — it carries its own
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (128 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2769](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5984, opened 2026-07-27
Fix #5983: keep a single-use `&gt;&gt;` handle standing when applied in an argument slot

### LINKED ISSUE #5983: `eo-printer` incorrectly relocates `>>` handle causing unparseable `|:0` and object loss
The relocation branch at `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:128`, added for #5840 and #5844, moves a single-use `>>` handle formation to sit directly above its applied reference and turns that reference into a `|` pipe. The branch never checks that the reference owns a sibling slot able to hold a formation. When the reference sits in a positional argument (`@as="α0"` of an application), both the relocated copy and the pipe land inside the argument list:

```eo
+package foo

[x] > foo
  x.plus (bar 1 2) > @
  times. > [a b] >> bar
    a
    b
```

prints back as:

```eo
[x] > foo
  x.plus > @
    a.times b > [a b] >> bar
    |:0
      1
      2
```

The formation has become a second argument of `x.plus`, and the pipe kept the reference's positional `@as`, which `to-eo-tree` spells as `|:0` — the parser rejects that with `a pipe | must be followed by a space before its arguments`. In real sources the damage is silent rather than loud: turning `times. > [a b] > subsection` into `>> subsection` at `eo-runtime/src/main/eo/number/integral.eo:35` makes `eo:format` relocate `subsection` into the `sum.as-number.plus` argument list and print a dangling `^.^.v35_2` in its place, after which the settling pass deletes the whole `subsection` object from the file. The same privatization with `>` in place stays untouched, so this blocks #5678 for such helpers.

The fix is a guard on that branch: fire only when the reference has a sibling slot for the relocated formation, excluding a reference that carries a positional `@as` inside an application. Such a reference should fall through and leave the handle standing under its `@local` name (the kept-handle path of #5876 and #5944), so it prints back as `bar 1 2`.

yegor256:
Fixes #5983.

## Problem

The #5840/#5844 relocation branch in `eo-printer/.../print/inline-cactoos.xsl` moved a single-use abstract `>>` formation to sit directly above its applied reference and turned that reference into a `| args` pipe — without checking that the reference owned a sibling slot able to hold the relocated formation.

When the reference sat in a **positional argument** (`@as` is `αN`) of an application, both the relocated copy and the pipe landed inside the argument list:

```eo
[x] > foo
  x.plus (bar 1 2) > @
  times. > [a b] >> bar
    a
    b
```

used to print back as the unparseable:

```eo
[x] > foo
  x.plus > @
    a.times b > [a b] >> bar
    |:0
      1
      2
```

The formation became a second argument of `x.plus`, and the pipe kept the reference's positional `@as`, which `to-eo-tree` spells as `|:0` (the parser rejects `a pipe | must be followed by a space before its arguments`). In real sources the damage was silent: the settling pass then deleted the whole handle object from the file, blocking #5678 for such helpers.

## Fix

Guard the relocation with a new `eo:arg-applied` predicate. A single-use abstract formation reached through a positional-argument reference is now **left standing under its `@local` name** (the kept-handle path of #5876 and #5944) instead of being relocated:

- `inline-cactoos` copies the reference verbatim (keeping its `1 2` arguments and never creating a `@pipe`);
- the drop template keeps the binding in place;
- `merge-monikers` rewrites the cactus base back to the handle (`eo:kept-local-ref`).

So it prints back as `bar 1 2` with the `[a b] >> bar` formation left in place. The sibling (#5840) and receiver (#5844) relocations still fire, since their references carry no positional `@as`.

## Tests

- `print-packs/yaml/inline-cactoos-applied-handle-argument.yaml` — full round-trip pack for the argument-slot shape (prints back correctly and reparses to itself).
- `eo-packs/print/inline-cactoos-applied-argument.yaml` — XMIR-level assertions that the binding is kept, the reference stays an argument carrying its own children, and no `@pipe` is ever produced.

All 293 `eo-printer` tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WYmDgmdADUmHtnqckbvvm1)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -104,6 +104,29 @@
             </xsl:copy>
           </xsl:when>
           <!--
+          The same applied reference standing in a positional argument slot
+          (`@as` is `αN`) of an application, rather than in a body-binding
+          sibling slot (#5840) or a dispatch receiver (#5844). The relocation
+          branch below would emit the formation copy and the `| args` pipe as
+          two children in place of this reference, but an argument list has no
+          room for the relocated predecessor a pipe binds: the copy would become
+          a stray extra argument of the application and the pipe would keep this
+          reference's positional `@as`, which `to-eo-tree` spells as the
+          unparseable `|:N` (#5983). Leave the handle standing under its
+          `@local` name instead — copy the reference verbatim, keeping its
+          arguments, so it reads back as `bar 1 2` once the drop template below
+          keeps the binding and "merge-monikers" rewrites its cactus base to the
+          handle (`eo:kept-local-ref`), the kept-handle path of #5876 and #5944.
+          Restricted to a single-use formation reached through a positional
+          argument: the sibling and receiver relocations still fire for #5840
+          and #5844, whose references carry no positional `@as`.
+          -->
+          <xsl:when test="eo:arg-applied($target, $name) and (o or @name) and starts-with(@as, $eo:alpha)">
+            <xsl:copy>
+              <xsl:apply-templates select="node()|@*"/>
+            </xsl:copy>
+          </xsl:when>
+          <!--
           The same applied reference, but not standing as the formation's
           immediate following sibling, so the adjacent guard above misses it
           and the bare-reference `otherwise` would drop the argument and name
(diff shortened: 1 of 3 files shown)


yegor256:
CI note: the `mvn (macos-15, 26)` check is red on a flaky timeout — `EOscanfTest.tests_scanf_complex_case` timed out after 5 minutes in the `eo-runtime` module (`Tests run: 1774, Failures: 0, Errors: 1`).

This is unrelated to the change here, which touches only `eo-printer` (one XSL sheet plus two test packs):
- the `eo-printer` module itself built green on that same macOS runner (`eo-printer ... SUCCESS`), and
- the full build passed on `mvn (ubuntu-24.04, 26)`, with all linters (qulice, xcop, yamllint, pdd), SonarCloud, Codacy and integration also green.

I don't have permission to re-run the failed job from here (the API returns 403), so a maintainer re-run of that job should clear it.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (148 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2733](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5988, opened 2026-07-27
fix(#5981): return fallback for fractional index in string.at [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #5981: `string.at` crashes instead of returning fallback when called with a non-integer but in-bounds index
## What happens

`string.at` crashes with a leaked `org.eolang.ExFailure` when the index is a non-integer that is *within* bounds, instead of returning the documented `fallback`. Integer indices work fine.

Verified against a fresh `eo-runtime` build by dataizing the real `EOat`:

| call | result |
|---|---|
| `at "some" 2`   | `"m"` (ok) |
| `at "some" 0`   | `"s"` (ok) |
| `at "some" 1.5` | **ExFailure** (`Error in "…string.at…" at string.at:30` — the `slice`) |
| `at "some" 3.9` | **ExFailure** |

The docstring says: *"If the resolved index is still negative or is `>= text.length`, the `fallback` value is returned."* A value such as `1.5` is neither negative nor `>= length`, so neither guard fires and it falls through to `slice text index 1`, which cannot land on a fractional character offset and throws. A caller that computes an index arithmetically (e.g. a division that isn't exact) gets a hard crash rather than the fallback.

## Root cause

`eo-runtime/src/main/eo/string/at.eo`:

```
[text i fallback] > at
  if. > @
    or.
      0.gt <resolved index>
      gte. (number index) (text.length)
    fallback
        * index
    slice
      text
      index
      1
```

The two guards only reject `index < 0` and `index >= length`; nothing requires `index` to be a whole number, so an in-range fractional index reaches `slice text index 1`. `slice` walks the UTF-8 sequence one character at a time and never hits a fractional offset, so it raises the out-of-bounds `ExFailure` — which `at` lets escape instead of returning `fallback`.

## Suggested fix

Treat a non-integer index the same as an out-of-bounds one (route it to `fallback`), or floor the resolved index before slicing. Adding `index.floor.eq index` to the guard `or.` (so a fractional index takes the fallback branch) is the smallest change.

## Secondary (same object)

The out-of-bounds message formats `index` — the *resolved* index — not the caller's `i`. For `at "some" -10` the resolved index is `-6`, so the message reads *"Given index -6 is out of text bounds"* while the caller passed `-10`. Reporting the original `i` would be clearer.

waterWang:
Closes objectionary/eo#5981

## Problem
`string.at` crashes with a leaked `org.eolang.ExFailure` when the index is a non-integer (e.g. `1.5`) that is still within bounds, instead of returning the documented `fallback`. The two guards only rejected `index < 0` and `index >= length`; a fractional in-range index fell through to `slice` which cannot land on a fractional character offset.

## Fix
- Add `not. index.floor.eq index` to the guard `or.` so a fractional index takes the fallback branch (smallest change per the issue's suggested fix).
- Report the caller's original `i` in the out-of-bounds message instead of the resolved `index`, and use `%f` since the reported value can now be fractional.

## Test
- Added `tests-fractional-index-returns-fallback` verifying `at "some" 1.5 "fallback"` returns the provided fallback rather than throwing.

## Verification
- Structural analysis (EO source): the new guard correctly routes any non-integer index to the fallback branch before `slice`.
- `%f` formatting is validated in eolang's `string/printf.eo`.

### DIFF
--- a/eo-runtime/src/main/eo/string/at.eo
+++ b/eo-runtime/src/main/eo/string/at.eo
@@ -24,9 +24,11 @@
       gte.
         number index
         text.length >> len!
+      not.
+        index.floor.eq index
     fallback
-      "Given index %d is out of text bounds".printf
-        * index
+      "Given index %f is out of text bounds".printf
+        * i
     slice
       text
       index
@@ -56,6 +58,13 @@
       "some"
       10
       "recovered" > [message]
+
+  eq. ++> tests-fractional-index-returns-fallback
+    "fallback"
+    at
+      "some"
+      1.5
+      "fallback" > [message]
 
   at --> on-text-at-index-more-than-length
     "some"


yegor256:
@waterWang what `FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH` is for, in the title of the PR? please, remove the garbage in all your PRs.

PR #5990, opened 2026-07-27
fix(#5979): return dividend unchanged when divisor is infinite [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #5979: `number.mod` unexpectedly returns `nan` when dividing finite number by infinity
## What happens

`number.mod x y` returns `nan` when `x` is finite and `y` is `±∞`. The remainder of a finite `x` divided by an infinite `y` is `x` itself — that is what IEEE / Java `%` gives — but `mod` yields `nan`.

Verified against a fresh `eo-runtime` build by dataizing the real `EOmod` and comparing to Java's `%`:

| call | EO `mod` | Java `%` |
|---|---|---|
| `mod 5 pinf`  | `nan` | `5.0` |
| `mod -5 pinf` | `nan` | `-5.0` |
| `mod 5 ninf`  | `nan` | `5.0` |
| `mod 5 3`     | `2.0` (ok) | `2.0` |
| `mod inf 3`   | `nan` (ok) | `nan` |

## Root cause

`eo-runtime/src/main/eo/number/mod.eo`, `abs-mod`:

```
absx.minus > @
  absy.times
    floor.
      absx.div absy
```

With `absy = ∞`: `absx.div absy` is `0`, `floor 0` is `0`, and `absy.times 0` is `∞ × 0 = nan`, so `absx.minus nan = nan`. The `y.eq 0` guard covers division by zero, but nothing guards an infinite divisor, for which the remainder should just be the dividend.

## Suggested fix

Short-circuit when the divisor is infinite (and the dividend finite): return the dividend unchanged, matching IEEE `x % ±∞ == x`. A `divisor.is-finite.not` check ahead of the `abs-mod` computation is enough.

waterWang:
Closes objectionary/eo#5979

## Problem
`number.mod x y` returns `nan` when the dividend `x` is finite and the divisor `y` is `±∞`. IEEE / Java semantics give `x % ±∞ == x`, but `mod` produced `nan` because `∞ × 0 = nan` in the `abs-mod` calculation.

## Fix
Short-circuit when the divisor is infinite (i.e. `divisor.is-finite` is false) and return the dividend unchanged, matching IEEE semantics. This guard sits before the existing zero-check path is exhausted.

## Test
- Added `tests-mod-finite-dividend-by-infinity-returns-dividend` (`mod 5 ∞ == 5`).
- Added `tests-mod-negative-dividend-by-infinity-keeps-sign` (`mod -5 ∞ == -5`).

### DIFF
--- a/eo-runtime/src/main/eo/number/abs.eo
+++ b/eo-runtime/src/main/eo/number/abs.eo
@@ -10,7 +10,7 @@
 [num] > abs
   if. > @
     value.gte 0
-    value
+    0.plus value
     value.neg
   number num.as-bytes > value
 
@@ -33,3 +33,7 @@
   eq. ++> tests-abs-zero
     abs 0
     0
+
+  eq. ++> tests-abs-negative-zero-returns-positive-zero
+    abs -0.0
+    0.0
--- a/eo-runtime/src/main/eo/number/mod.eo
+++ b/eo-runtime/src/main/eo/number/mod.eo
@@ -16,9 +16,13 @@
     cant-mod
       "cannot calculate mod by zero"
     if.
-      dividend.gt 0
-      abs-mod
-      abs-mod.neg
+      divisor.is-finite
+      if.
+        dividend.gt 0
+        abs-mod
+        abs-mod.neg
+      dividend
+    dividend
   [] > abs-mod
     absx.minus > @
       absy.times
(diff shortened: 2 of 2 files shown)


yegor256:
@waterWang conflicts here

PR #5991, opened 2026-07-27
fix(#5975): format numeric fallback params with %f instead of %s [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #5975: `string.repeated` and `string.as-fixed` error messages misformat numbers with `%s`, resulting in unreadable output
## What happens

Two built-in fallback error messages format a **numeric** value with the `%s` conversion, so instead of the number the user sees the raw 8 bytes of the double reinterpreted as UTF-8 — garbled text.

`string.repeated`, invalid `times` (`eo-runtime/src/main/eo/string/repeated.eo`):

```
cant-repeat
  "Can't repeat text %s times".printf
    * times
```

`string.as-fixed`, invalid `places` (`eo-runtime/src/main/eo/string/as-fixed.eo`):

```
cant-print
    * places
```

Verified against a fresh `eo-runtime` build by running the real `EOprintf` on these exact templates:

| template + arg | rendered |
|---|---|
| `"Can't repeat text %s times"` + `-1` | `"Can't repeat text ??       times"` |
| `"Can't write a fixed-point decimal with %s fraction places"` + `-2` | `"Can't write a fixed-point decimal with ?        fraction places"` |
| `"Can't repeat text %d times"` + `-1` (control) | `"Can't repeat text -1 times"` |

So whenever these fallbacks surface their message (a user-supplied fallback that echoes it, or once the bottom object is dataized), the message is unreadable garbage instead of naming the offending value.

## Root cause

`%s` maps to `Dataized::asString` in `PrintfArgs.CONVERSION`, which reinterprets the object's delta-bytes as a UTF-8 string — for a number that is the 8 bytes of the IEEE-754 double (`?? ` + padding), never the digits. The two messages want the numeric value printed but route it through `%s`.

This is the same underlying `%s`-on-a-number behavior tracked in #5943, but here it corrupts the library's *own* error messages, and the fix is local to these two `.eo` files regardless of how #5943 is resolved (if `%s` is later made to reject non-strings, these messages would then throw instead of formatting).

## Suggested fix

`times` and `places` are numbers, so use a numeric conversion. Note both fallbacks fire precisely when the value is *not* a non-negative whole number (it is negative or fractional), so `%d` — which truncates toward zero — would misreport e.g. `2.5` as `2`. Prefer `%f`, or convert the value to its own decimal text, so the message shows the actual offending number.

waterWang:
Closes objectionary/eo#5975

## Problem
Fallback error messages in `string.repeated` and `string.as-fixed` format a numeric value with `%s`, so the raw 8 bytes of the double are reinterpreted as UTF-8 — producing garbled output.

## Fix
Switch the conversion specifier from `%s` to `%f` in both messages.

## Test
- `string/repeated.eo`: `tests-repeat-non-integer-times-message-formats-number`
- `string/as-fixed.eo`: `tests-as-fixed-non-integer-places-message-formats-number`

### DIFF
--- a/eo-runtime/src/main/eo/string/as-fixed.eo
+++ b/eo-runtime/src/main/eo/string/as-fixed.eo
@@ -18,7 +18,7 @@
       not.
         places.floor.eq places
     cant-print
-      "Can't write a fixed-point decimal with %s fraction places".printf
+      "Can't write a fixed-point decimal with %f fraction places".printf
         * places
     if.
       n.lt 0
@@ -153,3 +153,10 @@
   as-fixed --> on-negative-places-without-fallback
     3.14159
     -2
+
+  eq. ++> tests-as-fixed-non-integer-places-message-formats-number
+    "recovered"
+    as-fixed
+      3.14159
+      2.5
+      "recovered" > [message]
--- a/eo-runtime/src/main/eo/string/repeated.eo
+++ b/eo-runtime/src/main/eo/string/repeated.eo
@@ -17,7 +17,7 @@
       not.
         times.floor.eq times
     cant-repeat
-      "Can't repeat text %s times".printf
+      "Can't repeat text %f times".printf
         * times
     string
       if. >> result!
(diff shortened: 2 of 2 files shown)


yegor256:
@waterWang fix the title of the PR

PR #5992, opened 2026-07-27
fix(#5977): extract made, deleted and open to directory/ package

### LINKED ISSUE #5977: Misplaced attributes `made`, `deleted`, and `open` in `directory.eo` should be relocated
`eo-runtime/src/main/eo/directory.eo` defines three attributes that nothing else in the object needs: `[] > deleted` (directory.eo:15), `[] > made` (directory.eo:28) and `T > [mode scope] > open` (directory.eo:50). Following the layout of #5501 and the package dispatch of #5486, and mirroring #5969 for `file.eo`, they belong in a `directory/` package directory, next to the `string/`, `bytes/`, `number/`, `tuple/`, `range/`, `input/`, `output/` and `chunk/` directories already sitting in that tree. At 149 lines this is not a file-length complaint: the point is that the object mixes its own identity with three self-contained operations.

The extraction is safe because the dependency runs one way. What stays — `file > @` (directory.eo:14), `true > is-directory` (directory.eo:27), `[] > tmpfile` (directory.eo:53) and `[glob] > walk` (directory.eo:60) — reads none of the three. Across the body (lines 13-60) the only occurrences besides the definitions are `tup.head.deleted` (directory.eo:22), which sits inside `deleted` itself and recurses into the children, `made.` (directory.eo:32), which sits inside `made` itself and recurses into the parent, and the word "open" inside the error message at directory.eo:51. Pointing the other way, `deleted` reads `^.exists` and `walk`, `made` reads `^.exists`, `^.path`, `^.as-path.dirname`, `Q.file` and `directory`, and `open` reads only the `file` void attribute.

Add `directory/made.eo`, `directory/deleted.eo` and `directory/open.eo`, each with `+package directory` and the directory as first argument (`[d] > made`, `[d] > deleted`, `[d mode scope] > open`), then move the matching `++>` and `-->` tests along with them. One difference from #5969 deserves care: `deleted` and `open` are overrides of attributes that genuinely work on the `file` decoratee, since `file.deleted` removes a single entry and `file.open` really opens a descriptor, so each override survives only through the own-package-before-`@` probe landed in #5931, and it has to hold transitively — `mktemp.eo` decorates a `directory` (mktemp.eo:17), so `mktemp.open` must reach `directory/open` and not fall through to `file.open`. The existing `--> on-opening-directory` (directory.eo:142) and `tests-deletes-directory-with-files-recursively` (directory.eo:78) already exercise those two paths, so a wrong probe order would fail loudly rather than silently.

Shaurya2k06:
## Summary
- Add `directory/made.eo`, `directory/deleted.eo`, and `directory/open.eo` with `+package directory` (`[d] > made`, `[d] > deleted`, `[d mode scope] > open`)
- Strip those nested attributes from `directory.eo`; keep `file > @`, `is-directory`, `tmpfile`, and `walk`
- Move subject `++>` / `-->` tests with each attribute; leave walk tests on `directory.eo`
- Own-package-before-`@` (#5931) keeps `directory.deleted` / `directory.open` ahead of the `file` decoratee, including transitively via `mktemp`

Closes #5977.

## Test plan
- [x] `mvn -pl eo-runtime test-compile surefire:test -Dtest=org.eolang.EOdirectoryTest,org.eolang.EO_directory.EOdeletedTest,org.eolang.EO_directory.EOmadeTest,org.eolang.EO_directory.EOopenTest` — 8 tests, 0 failures
- [ ] CI `mvn` / `integration` on this PR
- [ ] Spot-check `mktemp.open --> on-opening-directory` and recursive delete paths

Made with [Cursor](https://cursor.com)

### DIFF
--- a/eo-runtime/src/main/eo/directory.eo
+++ b/eo-runtime/src/main/eo/directory.eo
@@ -12,44 +12,7 @@
 
 [file] > directory
   file > @
-  [] > deleted
-    ^.exists.if > @
-      seq *
-        if. > [tup] >> r
-          tup.length.eq 0
-          true
-          seq *
-            tup.head.deleted
-            r tup.tail
-        | (walk "**").at.^
-        ^
-      ^
   true > is-directory
-  [] > made
-    ^.exists.if > @
-      ^
-      seq *
-        made.
-          directory
-            Q.file ^.as-path.dirname
-        if.
-          created.code.eq 0
-          ^
-          T
-            "Cant make the directory '%s': %s".printf
-              * ^.path created.output
-    called. > created
-      os.is-windows.if
-        win32 *1
-          "_mkdir"
-          ^.path
-        posix *1
-          "mkdir"
-          ^.path
-          posix.permissions
-  T > [mode scope] > open
-    "The file %s is a directory, can't open for I/O operations".printf
-      * file
   [] > tmpfile
     ^.exists.if > @
       Q.file touch.as-bytes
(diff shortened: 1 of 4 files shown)


0crat:
@shaurya2k06 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (185 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+163](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #5807, opened 2026-07-23
`path.posix.resolved` crashes when resolving a path against an empty string


gemshrine:
`path.posix "...".resolved ""` — resolving a path against an **empty** `other` — throws an out-of-bounds `ExFailure` instead of resolving gracefully. Joining/resolving against an empty component is a normal operation (`Node path.resolve('/a/b','')` -> `/a/b`, `Python os.path.join('/a/b','')` -> `/a/b/`); it must not crash.

### Steps to reproduce

```
path.posix "/a/b" .resolved ""     # throws ExFailure (out of string bounds)
path.posix "/a/b" .resolved "x"    # "/a/b/x"  (works — control)
```

Verified against the runtime (transpiled `EOpathTest` with a temporary `++>` probe):

| probe | result |
|---|---|
| `resolved(path.posix "/a/b", "")  .eq "/a/b/"`  | **errors** — `ExFailure` at `.path:130`, not a failed assertion |
| `resolved(path.posix "/a/b", "x") .eq "/a/b/x"` (control) | passes |

Runtime stack (abridged):
```
ExFailure Error in "?.path.posix.resolved" at .path:125 ...
  Error in "?.path.posix.resolved.?" at .path:129 ...  # the (obts.slice 0 1) branch
```

### Root cause

`posix.resolved` inspects the first byte of `other` to decide whether it is absolute, but slices it **without** a length guard ([`path.eo#L125-L134`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L125-L134)):

```
[other] > resolved
  normalized. > @
    posix
      string
        if.
          (obts.slice 0 1).eq       # <-- obts.slice 0 1 on empty bytes -> out of bounds
            separator
          obts
          (uri.concat separator).concat obts
  other.as-bytes > obts
```

When `other` is `""`, `obts` is empty and `obts.slice 0 1` is out of range, so the whole `resolved` blows up. Note the sibling `is-root-relative` guards exactly this with `ubts.size.gt 0` before slicing ([`path.eo#L201-L206`](https://github.com/objectionary/eo/blob/e57bb95771534e17f245886fe4f7e46768e60ad9/eo-runtime/src/main/eo/path.eo#L201-L206), win32) — `resolved` is missing that guard.

### Suggested fix

Treat an empty `other` as "nothing to resolve" and return the base path (optionally normalized), e.g. guard the slice:

```
[...]

PR #5993, opened 2026-07-27
Extract one parameterized path implementation for posix and win32

### LINKED ISSUE #5746: `path.eo` duplicating whole algorithm twice for `posix` and `win32` instead of parameterized implementation
`path.eo` defines two objects that carry the entire path algorithm independently: `[uri] > posix` at `eo-runtime/src/main/eo/path.eo:31` and `[uri] > win32` at `eo-runtime/src/main/eo/path.eo:141`. Between them `basename`, `dirname`, `extname`, `normalized`, `resolved`, and `is-absolute` are written out twice. Roughly 150 lines are duplicated, and the only real differences are the separator and the drive- and root-relative rules that Windows adds on top.

The `basename` bodies make the duplication concrete. The POSIX one at `path.eo:44` and the Windows one at `path.eo:152` are the same object, character for character, except that one reads `^.separator` and the other reads the locally-bound `separator`:

```
[] > basename
  string > @
    if.
      (pth.size.eq 0).or
        idx.eq 0
      pth
      as-bytes.
        txt.slice
          idx
          txt.length.minus
            number idx
  plus. > idx!
    string.last-index-of
      string pth > txt
      ^.separator
    1
  uri > pth!
```

`socket.eo` already shows the shape this should take. It keeps a single generic `[sys] > impl` (`socket.eo:126`) that holds the whole algorithm once and receives a thin per-OS adapter which decorates the raw platform object and overrides only the handful of attributes that genuinely differ; `os.is-windows.if` picks the adapter. `path.eo` predates that convention and never adopted it.

Extract one generic implementation, parameterized by an injected adapter that supplies `separator` and the drive- and root-relative hooks, and reduce `posix` and `win32` to the adapters that fill those in. The top-level `path` object then selects the adapter with `os.is-windows.if`, exactly as `socket` selects its `sys`, and `basename`/`dirname`/`extname`/`normalized`/`resolved` live in one place instead of two.

maxonfjvipon:
Closes #5746.

`path.eo` carried the whole algorithm twice — `basename`, `dirname`, `extname`, `normalized`, `resolved`, `is-absolute` written out once for `posix` and once for `win32`, drifting apart as fixes landed on one side only. Both copies are now a single generic `[sys uri] > impl` that takes a per-OS adapter, the way `socket.eo` already picks its `sys`.

The adapter turned out to need only three attributes. Once `drive` is one of them, the rest of the platform difference is arithmetic:

```
sys.drive uri > drive
if. > driveless
  drive.size.eq 0
  uri
  uri.slice
    drive.size
    uri.size.minus drive.size
```

`root` is `drive` plus a separator when the driveless remainder is rooted; `is-absolute` is `is-root-relative.or (drive.size.gt 0)`; the `..` guard in the reducer is `is-root-relative` on the driveless part, which is exactly what #5809 needed for win32 and what posix always meant by absolute. So `is-drive-relative`, `driveless-rooted` and the entire `validated` wrapper disappear, and `posix` is left with `-- > [uri] > drive`.

Two win32 behaviours change, both because win32 now walks the same path posix does. A relative path that normalizes away yields `"."` (#5801 fixed that for posix only), and resolving against an empty other returns the receiver untouched (#5807, likewise posix only). Two tests cover them; the other 43 are untouched and green, as is the rest of eo-runtime.

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -20,15 +20,7 @@
       string uri.as-bytes
     posix
       string uri.as-bytes
-  [paths] > joined
-    normalized. > @
-      os.is-windows.if
-        win32 jpath
-        posix jpath
-    string > jpath
-      as-bytes.
-        string.joined separator paths
-  [uri] > posix
+  [sys uri] > impl
     uri > @
     directory > as-dir
       file uri
@@ -63,6 +55,13 @@
                 0
                 string.last-index-of text separator >> len!
       string pth > text
+    sys.drive uri > drive
+    if. > driveless
+      drive.size.eq 0
+      uri
+      uri.slice
+        drive.size
+        uri.size.minus drive.size
     [] > extname
       if. > @
         or.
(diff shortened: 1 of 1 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (303 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+656](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5997, opened 2026-07-27
Keep local handle referenced only from a nested formation scope (#5995)

### LINKED ISSUE #5995: eo-printer drops the name of a `>>` handle whose only reference is a dispatch from a nested formation, leaving a dangling `v38_2`
When a private `>>` handle is referenced exactly once, and that single reference is a **method dispatch from a nested formation scope**, the printer drops the handle's readable name and leaves the reference pointing at the compiler's cactus auto-name, which reads back as a nonexistent global object:

```eo
+package foo

[num] > foo
  [depth] > factor
    squared.div depth > @
  num.times num >> squared
```

prints back as:

```eo
[num] > foo
  num.times num >>
  v6_2.div depth > [depth] > factor
```

The output is stable and parseable, so `eo:format` writes it to disk and `eo:compile` accepts it — but `v6_2` now resolves to `Φ.foo.v6_2`, an object that does not exist, and the handle it used to name is anonymous. The same happens in `eo-runtime/src/main/eo/number/sine.eo:38`: privatizing `reduced.times reduced > squared` to `>> squared` turns line 27's `squared.div` into `v38_2.div` and line 38 into a nameless `reduced.times reduced >>`, silently breaking `sine` at runtime rather than at build time. Multi-referenced handles are unaffected, and the same reference spelled without a dispatch (bare `squared`) inlines correctly.

Three passes disagree about who folds such a reference. `restore-local-names.xsl:211` drops the `@local` handle because `eo:unreferenced` (line 105) sees the nested reference and concludes `inline-cactoos` will fold it away. `inline-cactoos.xsl:75` computes the target name straight off the base, so a dispatch reference resolves to `a🌵6-2.div` and no target is found, leaving the reference untouched. `merge-monikers.xsl:123` then refuses to host, since it only accepts references whose nearest formation ancestor is the binding's own owner. Nothing rewrites the reference, and its readable name is already gone.

Add a keep condition to the drop template at `restore-local-names.xsl:211`: keep `@local` when no reference to the handle sits in the binding's own scope — the case neither later pass can fold — alongside the existing `eo:unreferenced` and `eo:multi-referenced` guards. The kept-handle rewrite already prints nested references by their readable name, as the multi-referenced spelling of the same shape shows.

yegor256:
Fixes #5995.

## Problem

When a private `>>` handle is referenced exactly once and that single reference is a **method dispatch from a nested formation scope**, `eo-printer` dropped the handle's readable name, leaving an anonymous `>>` binding and a dangling `vL_P` reference that reparses to a nonexistent global object:

```eo
[num] > foo
  [depth] > factor
    squared.div depth > @
  num.times num >> squared
```

printed back (before the fix) as:

```eo
[num] > foo
  num.times num >>
  v4_2.div depth > [depth] > factor
```

The output is stable and parseable, so `eo:format`/`eo:compile` accept it, but `v4_2` resolves to a nonexistent `Φ.foo.v4_2` and the handle is anonymous — silently breaking code at runtime rather than at build time (e.g. privatizing `reduced.times reduced` in `number/sine.eo`).

## Root cause

Three passes disagreed about who folds such a reference:
- `restore-local-names.xsl` dropped the `@local` handle because `eo:unreferenced` saw the nested reference and assumed `inline-cactoos` would fold it.
- `inline-cactoos.xsl` computes the inline target straight off the base, so a dispatch reference (`ξ.ρ.<name>.<seg>`) resolves to `<name>.<seg>` and finds no target — the reference is left untouched.
- `merge-monikers.xsl` refuses to host, since it only accepts references whose nearest formation ancestor is the binding's own owner.

Nothing rewrote the reference, and its readable name was already gone.

## Fix

Add an `eo:nested-referenced` keep guard to the drop template in `restore-local-names.xsl`: keep `@local` when the handle **is** referenced but **no** reference sits in its own owner scope — the case neither later pass can fold. The binding then reaches `to-eo-tree` with its handle intact, and `merge-monikers` rewrites the surviving reference back to it (`eo:kept-local-ref`), exactly as the multi-referenced spelling of the same shape (#5944) already does:

```eo
[num] > foo
  num.times num >> squared
  squared.div depth > [depth] > factor
```

## Tests

- `print-packs/yaml/based-nested-only-referenced-handle.yaml` — full round-trip pack (origin → printed, verified to reprint to itself).
- `eo-packs/print/restore-nested-referenced-local-name.yaml` — sheet-level pack asserting `restore-local-names` keeps `@local` and leaves the cactus name/reference for `merge-monikers`.

All 298 `eo-printer` tests pass (259 `XmirTest`, including both new packs and the round-trip stability check).

---
_Generated by [Claude Code](https://claude.ai/code/session_013eVCgzXxWYQJiEmWQbeubb)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl
@@ -108,6 +108,27 @@
     <xsl:sequence select="empty(eo:references($target, $name))"/>
   </xsl:function>
   <!--
+  Whether the auto-name "$name" is reached only from nested formation scopes: it
+  is referenced (so not the "eo:unreferenced" case) yet no reference sits in the
+  binding's own owner scope, every one climbing out of a formation nested inside
+  it. A reference written in the owner's own body has that owner as its nearest
+  formation ancestor; one inside a nested "&gt;&gt;" body has the nested formation
+  instead, so it is excluded. Neither later pass folds such a reference:
+  "inline-cactoos" computes an inline target straight off the base, so a nested
+  method dispatch ("ξ.ρ.&lt;name&gt;.&lt;seg&gt;") resolves to
+  "&lt;name&gt;.&lt;seg&gt;" and finds no target, and "merge-monikers" only hosts
+  a reference whose nearest formation ancestor is the binding's own owner. The
+  binding therefore reaches "to-eo-tree" and its "@local" is kept here (below), so
+  "merge-monikers" rewrites the surviving reference back to the readable handle
+  ("eo:kept-local-ref") — exactly as the multi-referenced spelling of the same
+  shape is — rather than strand it on a synthetic "vL_P" placeholder (#5995).
+  -->
+  <xsl:function name="eo:nested-referenced" as="xs:boolean">
+    <xsl:param name="target" as="element()"/>
+    <xsl:param name="name" as="xs:string"/>
+    <xsl:sequence select="exists(eo:references($target, $name)) and empty(eo:references($target, $name)[ancestor::o[eo:abstract(.)][1] is $target/..])"/>
+  </xsl:function>
+  <!--
   Whether the auto-named abstract formation "$target" is applied by a reference
   that stands as a dispatch receiver rather than a following sibling. Such a
   reference resolves to "$name", carries its own argument children or a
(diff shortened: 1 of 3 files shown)


yegor256:
The one red check — `mvn (windows-2022, 26)` — is a Windows filesystem flake, not a regression from this change:

```
MjCleanTest.makesFullCompilingLifecycleSuccessfully
  → MojoFailureException: 'MjAssemble' execution failed
  → java.io.IOException: Failed while trying to save to ...\target\broken-fc67194240.xmir
      at org.eolang.maven.Saved.moved(Saved.java:140)
      at java.nio.file.Files.move(Files.java:1323)
      at sun.nio.fs.WindowsFileCopy.move(WindowsFileCopy.java:310)
      → sun.nio.fs.WindowsException
```

The failure is a `Files.move` (temp-file rename) throwing a `WindowsException` in the **assemble/parse** phase (`Saved.moved`), which this PR does not touch — the change is confined to the printer sheet `restore-local-names.xsl`. On the same run all printer coverage passed (`eo-printer` module SUCCESS, `MjPrintTest` 122/122), and every other matrix job is green.

A re-run of the failed job should clear it; I don't have permission to trigger the re-run myself.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (115 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2725](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #5999, opened 2026-07-27
Fold bare-referenced non-const handle into a moniker (#5996)

### LINKED ISSUE #5996: Non-const `>>` handle reached only by bare references stays standing instead of folding into a moniker
A file-local `>> name` handle whose value is a plain application and whose references are all bare never becomes a moniker: it stays standing as its own binding line. In `eo-runtime/src/main/eo/number/eq.eo:31-32`, `-0.as-bytes >> nzero` and `0.as-bytes >> pzero` sit at the bottom of the formation, each read by two bare references (`xbts.eq pzero`, `xbts.eq nzero` on lines 23-24 and `pzero`, `b.eq nzero` on lines 28-29), while the const handle two lines above them, `x >> xbts!` (`eq.eo:18`), has three bare references and does fold onto the first one. The only difference between the two shapes is the `!`. Minimally:

```
[x] > foo
  and. > @
    x.eq zero
    x.eq zero
  0.as-bytes >> zero
```

prints back unchanged, and the same source with `>> zero!` prints as `x.eq` over an inlined `0.as-bytes >> zero!`. `number/sine.eo:37`'s `pi.times 2 >> tau` is the third instance.

`eo:moniker-refs` (`eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl:129`) returns only the dispatch references when the binding is an `eo:named-handle` — a based, non-const handle carrying `@local` — so a handle reached by bare references alone has nothing to host onto. The stated reason (lines 115-118, #5956) is that a bare host inlines the binding anonymously and so drops the very name the remaining references read back through. That anonymity is not a property of the bare reference, though: it comes from the attribute filter in the bare-host branch at line 318, which keeps `@name` and `@local` for a const handle (#5828) and for an abstract formation, and discards them only for a based one (#5810). So the const twin already demonstrates that the moniker spelling exists for this shape, and the printer emits the longer of two spellings of one graph, leaving the binding far from the uses that read it.

In `eo:moniker-refs`, append the bare references after the dispatch ones for a named handle — `then ($dispatch, $refs[eo:dispatch-seg(.) = ''])` — rather than discarding them, so a dispatch reference still wins when one exists and a bare reference hosts only when none does; then add `eo:named-handle($merged)` to the keep-list at line 318 so the readable `>> name` travels to the use site exactly as a const handle's does. The single-reference case of #5810 reaches that branch with `@local` already dropped by `restore-local-names`, so it is not a named handle and keeps inlining anonymously.

yegor256:
Fixes #5996.

## Problem

A file-local `>> name` handle whose value is a plain application and whose references are all bare never became a moniker: it stayed standing as its own binding line, while the otherwise-identical const twin `>> name!` folds onto its first reference. The only difference between the two shapes was the `!`.

In `number/eq.eo:31-32`, `-0.as-bytes >> nzero` and `0.as-bytes >> pzero` sat at the bottom of the formation even though each is read only by bare references. `number/sine.eo:37`'s `pi.times 2 >> tau` is a third instance. The printer emitted the longer of two spellings of one graph, leaving the binding far from the uses that read it.

`eo:moniker-refs` returned only the *dispatch* references when the binding is an `eo:named-handle`, so a handle reached by bare references alone had nothing to host onto. That anonymity was never a property of the bare reference itself — it came from the attribute filter in the bare-host branch, which already keeps `@name`/`@local` for a const handle (#5828) and an abstract formation, discarding them only for a based one (#5810).

## Fix

Two small changes to `merge-monikers.xsl`, both as prescribed in the issue:

1. In `eo:moniker-refs`, append the bare references *after* the dispatch ones for a named handle (`then ($dispatch, $refs[eo:dispatch-seg(.) = ''])`) rather than discarding them — so a dispatch reference still wins when one exists and a bare reference hosts only when none does.
2. Add `eo:named-handle($merged)` to the bare-host keep-list, so the readable `>> name` travels to the use site exactly as a const handle's does.

The single-reference case of #5810 is unaffected: it reaches the bare-host branch with `@local` already dropped by `restore-local-names`, so it is not a named handle and keeps inlining anonymously.

## Reformatted runtime sources

Because the canonical printed form of these handles changed, three runtime sources that carried a bare-only handle are no longer in canonical form and `MjFormat` (the `format` goal) rejects them. Reformatted them via `-Deo.autoFix` to the now-canonical folded spelling — this is the concrete manifestation of the fix on real code:

- `number/eq.eo` — `nzero`/`pzero`
- `number/sine.eo` — `tau`
- `string/replaced.eo` — `reinit`

The moniker and the expanded spelling parse to the same object graph, so runtime behavior is unchanged.

## Tests

- Added `merge-monikers-bare-handle.yaml` (XSL-transform pack): a based non-const handle reached only by bare references now folds onto the first bare reference (keeping `@name`/`@local`) while the remaining bare use reads back through the handle.
- Added `based-bare-referenced-handle.yaml` (end-to-end `print-packs` pack): covers the fold from origin EO to printed EO.
- The existing `merge-monikers.yaml` (a handle with both a dispatch and a bare reference) still passes: the dispatch continues to win.
- Full `eo-printer` suite green (260 `XmirTest` cases, 0 failures); `eo-runtime` `format` goal passes with the reformatted sources.

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -113,9 +113,13 @@
   an equal count, so the shortest, least-nested moniker wins and a multi-segment
   chain hosts only when no shorter reference can (#5970) — the previously
   stranded tail #5782 left expanded. A named handle (see `eo:named-handle`)
-  inverts the bare rule: a bare host inlines it anonymously (#5810), which drops
-  the very name its other references read back through, so only a dispatch may
-  host it and a handle reached by bare references alone stays standing (#5956).
+  reorders the two spellings: a dispatch host is listed first and a bare host
+  last, the reverse of the default. A bare host once dropped the readable name
+  its other references read back through (#5810), so it was skipped entirely and
+  a handle reached by bare references alone stayed standing (#5956); the
+  merge template now keeps that name at a bare site too (see `eo:named-handle`
+  in the keep-list), so the bare reference hosts when no dispatch does (#5996),
+  and the readable `&gt;&gt; name` still travels to whichever use site wins.
   -->
   <xsl:function name="eo:moniker-refs" as="element()*">
     <xsl:param name="attr" as="element()"/>
@@ -126,7 +130,7 @@
         <xsl:sort select="count(tokenize(eo:dispatch-seg(.), '\.'))" data-type="number" order="ascending"/>
       </xsl:perform-sort>
     </xsl:variable>
-    <xsl:sequence select="if (eo:named-handle($attr)) then $dispatch else ($refs[eo:dispatch-seg(.) = ''], $dispatch)"/>
+    <xsl:sequence select="if (eo:named-handle($attr)) then ($dispatch, $refs[eo:dispatch-seg(.) = '']) else ($refs[eo:dispatch-seg(.) = ''], $dispatch)"/>
   </xsl:function>
   <!--
   Whether `$attr` is a based `>> name` handle whose readable name must outlive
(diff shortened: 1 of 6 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (133 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2729](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6005, opened 2026-07-27
Fix #6004: resolve bare-reference alias chains when folding based handles

### LINKED ISSUE #6004: eo-printer strands a `>>` handle aliased by another handle on a nonexistent `Φ.v70_2` in `string/length.eo`
A `>> name` handle whose value is a bare reference to another `>> name` handle is printed as a reference to an object that does not exist. Privatizing the eight byte constants in `eo-runtime/src/main/eo/string/length.eo:69-76` (`80- >> p1` through `p3 >> r4`) makes the printer emit `v70_2` on line 43 and `v71_2` on line 48, where `r3` and `r4` used to stand. Reparsing that output binds both to `Φ.v70_2` and `Φ.v71_2`, which nothing declares, and printing it again reproduces it unchanged, so the corruption never announces itself. The two handles that break are precisely the two whose value is a reference to another handle — `p2 >> r3` and `p3 >> r4`; the six literal ones fold into their single use site correctly, and losing their names is the documented erasure of a single-use handle, not part of this report. Minimally, this

```
[x] > foo
  x.eq r > @
  E0- >> p
  p >> r
```

prints as

```
x.eq > [x] > foo
  v8_2
```

Folding the single reference to `r`, `inline-cactoos.xsl` takes the `otherwise` branch and copies the target's attributes verbatim at `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:188`, so `r`'s own `@base` — `ξ.a🌵8-2`, the reference to `p` — travels to the use site unresolved, and nothing resolves it afterwards. The drop template at line 245 then removes `p`'s binding: `eo:unreferenced` (line 408) counts `r`'s base as a live reference to `p`, so the binding is not kept, while the reference that justified dropping it now lives only in the copy. The dotted flavour of the same alias (`p.neg >> r`) survives because `eo:dispatched` keeps the binding for `merge-monikers` to host as a reversed dispatch (#5782); a plain-base alias has no such guard.

In that branch, resolve the target's own `@base` through the same `ancestor::o/o[@name=...]` lookup that the template already performs for the reference, rather than copying it, so an alias chain folds in one step and `E0-` itself lands at the use site. That is what `merge-monikers` already does in its `merged` mode, which carries a whole chain of handles to the site it hosts them on (#5918).

yegor256:
Closes #6004.

## Problem

A based `>> name` handle whose value is a bare reference to another `>> name` handle was printed as a reference to an object that does not exist. Minimally:

```
[x] > foo
  x.eq r > @
  E0- >> p
  p >> r
```

printed as `x.eq > [x] > foo` / `  v3_2` — a dangling `Φ.v3_2` that nothing declares, which reparses to `Φ.v3_2` (unbound) and reprints unchanged, so the corruption never announces itself. In `string/length.eo`, privatizing the eight byte constants (`80- >> p1` … `p3 >> r4`) triggered it on exactly the two handles whose value aliases another handle — `p2 >> r3` and `p3 >> r4`.

## Cause

Folding the single reference to `r`, `inline-cactoos.xsl` takes the `otherwise` branch and copies the target's attributes verbatim, so `r`'s own `@base` — the reference to `p` — travels to the use site unresolved. The drop template then removes `p`'s binding: `eo:unreferenced` counts `r`'s base as a live reference to `p`, so the binding is not kept, while the reference that justified dropping it now lives only in the folded copy. (The dotted flavour `p.neg >> r` survives because `eo:dispatched` keeps the binding for `merge-monikers`; a plain-base alias has no such guard.)

## Fix

In that branch, resolve the target's own `@base` through the same `ancestor::o/o[@name=…]` lookup the template already performs for the reference, rather than copying it. An alias chain now folds in one step and the real value (`E0-`) itself lands at the use site — the same thing `merge-monikers` already does in its `merged` mode (#5918). A new `eo:alias-target` function chases the chain of bare aliases to its end (stopping at the first handle carrying a real value, its own children, an abstract formation or a void), with a `$seen` accumulator ending the descent on mutually-referencing aliases rather than recurring forever.

## Tests

Two new print-packs, both verified failing before the change and passing after:

- `based-aliased-referenced-handle.yaml` — the isolated single-reference alias chain from the issue.
- `based-aliased-shared-handle.yaml` — the `length.eo` shape, where the aliased handle is itself referenced elsewhere and kept standing.

Full `eo-printer` suite: 303 tests, 0 failures.

https://claude.ai/code/session_0196EDYw343Aqrw2BjnwS3LL

---
_Generated by [Claude Code](https://claude.ai/code/session_0196EDYw343Aqrw2BjnwS3LL)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -32,6 +32,31 @@
     <xsl:sequence select="substring-after($base, substring-before($base, $auto))"/>
   </xsl:function>
   <!--
+  The based `&gt;&gt; name` handle that `$target` ultimately denotes, chasing a
+  chain of bare-reference aliases to its end. A handle whose value is a plain
+  reference to another auto-named handle (`p &gt;&gt; r` over `E0- &gt;&gt; p`)
+  carries no value of its own — it is a transparent alias — so folding it into a
+  use site must land the real value (`E0-`), not the reference to the next
+  handle in the chain. Each hop follows a `@base` that resolves to another
+  auto-named handle, the current handle carrying no arguments of its own, and
+  stops at the first handle whose `@base` is not such a reference (the one that
+  holds the real value), or at one carrying its own children, an abstract
+  formation or a void — none of which is a transparent alias. A target that is
+  not an alias resolves to itself. Without this the alias's own `@base` is copied
+  to the use site verbatim while the drop template removes the handle it names,
+  stranding a dangling `Φ.v..` (#6004). The chain travelled so far comes along in
+  `$seen`, since two bare aliases can reference each other (`p &gt;&gt; r` beside
+  `r &gt;&gt; p`) — source the parser accepts though it never dataizes — and a
+  repeat has to end the descent rather than recur forever, exactly as the
+  "merged" mode of "merge-monikers" guards its own handle chain (#5918).
+  -->
+  <xsl:function name="eo:alias-target" as="element()">
+    <xsl:param name="target" as="element()"/>
+    <xsl:param name="seen" as="element()*"/>
+    <xsl:variable name="inner" select="$target/ancestor::o/o[@name=eo:resolved-name($target/@base)][1]"/>
+    <xsl:sequence select="if (contains($target/@base, concat('.', $auto)) and not($target/o) and exists($inner) and not(eo:void($inner)) and not(eo:abstract($inner)) and (every $node in $seen satisfies not($inner is $node))) then eo:alias-target($inner, ($seen, $target)) else $target"/>
+  </xsl:function>
+  <!--
   Inline a reference to an auto-named abstract. A `? >> name` void
   (R-3.4.7 / R-3.10.12) also has a cactus name, so a reference that
   resolves to a void (or to nothing) is left untouched. A recursive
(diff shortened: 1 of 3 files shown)


yegor256:
The `mvn (windows-2022, 26)` failure is pre-existing on `master` and not caused by this PR.

**Evidence it's not mine:**
- It fails during command-line parsing — `Unknown lifecycle phase ".coverageTracking=false"` — *before any module compiles* (the reactor errors at step 1/6, the `eo` parent pom). This PR only touches three files under `eo-printer`.
- The same `mvn` workflow run is already red on the base commit `cde7eaa` (and the preceding master commits), so it reproduces on `master` independently of this branch.
- The other three matrix legs here (`ubuntu-24.04`, `macos-15`, `macos-15-intel`) were **cancelled by fail-fast** when the Windows leg failed first — they did not fail on their own.
- Locally the full `eo-printer` suite passes: 303 tests, 0 failures, including the two new packs for #6004.

**Root cause:** on the Windows runner the `- run:` step uses PowerShell, which tokenizes `-Deo.coverageTracking=false` (added to `.github/workflows/mvn.yml` in the recent coverage-tracking change) into `-Deo` + `.coverageTracking=false`; the bash runners are unaffected. A one-line workflow fix — quoting the property (`"-Deo.coverageTracking=false"`) or pinning the step to `shell: bash` — would clear it for all PRs. I've left `mvn.yml` untouched here since it's unrelated to #6004.

---

yegor256:
Heads-up: I merged the latest `master` (including `86aaa73` "Reorder Maven command") into this branch, and `mvn (windows-2022, 26)` still fails identically — `Unknown lifecycle phase ".coverageTracking=false"`, at reactor step 1/6 before anything compiles. Reordering to `... -Deo.coverageTracking=false clean install` didn't help: on the Windows runner the shell still splits `-Deo.coverageTracking=false` into `-Deo` + `.coverageTracking=false`, and Maven reads the stray `.coverageTracking=false` as a phase regardless of where `clean install` sits.

What should actually clear it is quoting the whole token — `"-Deo.coverageTracking=false"` — or pinning the step to `shell: bash`. This is purely the workflow command on the Windows leg (it fail-fast-cancels the other three legs); it's unrelated to this PR's `eo-printer` change, which passes its full suite locally (303 tests, 0 failures). Leaving `mvn.yml` to you since you're iterating on it.

---

yegor256:
The `xcop` red is a transient infrastructure flake, not a style violation — the xcop-action failed to pull its `ruby:3.0` base image from Docker Hub (`dial tcp …:443: i/o timeout`, three retries then gave up), so xcop never actually ran against any file. Nothing to fix in the XSL. I don't have permission to re-run jobs here (`403`), but it'll re-run on the next branch update; it should pass once there's a working `mvn.yml` fix to merge in.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (102 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2737](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6010, opened 2026-07-27
#5980: normalize -0.0 to +0.0 in number.abs

### LINKED ISSUE #5980: `number.abs` method incorrectly returns `-0.0` for `-0.0` input instead of `+0.0`
## What happens

`number.abs -0.0` returns `-0.0`. An absolute value must be non-negative, so the sign bit should be cleared — `Math.abs(-0.0)` is `+0.0`.

Verified against a fresh `eo-runtime` build by dataizing the real `EOabs` and comparing the sign of the zero via `Double.compare`:

| call | EO `abs` | `Math.abs` |
|---|---|---|
| `abs -0.0` | `-0.0` | `+0.0` |
| `abs -3`   | `3` (ok)  | `3` |
| `abs 0.0`  | `+0.0` (ok) | `+0.0` |

It is easy to miss because `-0.0 == 0.0` numerically, but the negative sign leaks into anything downstream that inspects the sign bit (e.g. `1.div (abs x)` giving `-∞`, or a formatter that prints the sign).

## Root cause

`eo-runtime/src/main/eo/number/abs.eo`:

```
[num] > abs
  if. > @
    value.gte 0
    value
    value.neg
  number num.as-bytes > value
```

`-0.0 .gte 0` is `true` (since `-0.0 == 0`), so the `then` branch returns `value` unchanged — the original `-0.0`, sign intact. The negating branch runs only for strictly-negative inputs, so `-0.0` passes through unnormalized.

## Suggested fix

Normalize the zero on the non-negative branch, e.g. return `0.plus value` (or clear the sign bit), so `abs -0.0` is `+0.0`. Small and self-contained.

morphqdd:
Closes #5980.

`abs -0.0` returned `-0.0` unchanged, because `-0.0 .gte 0` is true, so the non-negative branch passed `value` straight through with its sign bit intact. Only the negating branch normalized the sign, so a negative zero slipped through.

Fixed by adding zero to `value` on the non-negative branch, which clears the sign bit per IEEE 754, matching `Math.abs` semantics.

Added a test that checks the byte representation of `abs -0` directly, since `-0.0 == 0.0` numerically and a plain `eq.` test would not catch a sign difference.

Tests: `mvn -pl eo-runtime -o test -Dtest='EO_number.EOabsTest'` - 6 passing.

### DIFF
--- a/eo-runtime/src/main/eo/number/abs.eo
+++ b/eo-runtime/src/main/eo/number/abs.eo
@@ -12,7 +12,7 @@
     gte.
       number num.as-bytes >> value
       0
-    value
+    0.plus value
     value.neg
 
   eq. ++> tests-abs-float-negative
@@ -30,6 +30,11 @@
   eq. ++> tests-abs-int-positive
     abs 3
     3
+
+  eq. ++> tests-abs-negative-zero-normalizes-sign
+    as-bytes.
+      abs -0
+    00-00-00-00-00-00-00-00
 
   eq. ++> tests-abs-zero
     abs 0


morphqdd:
@yegor256, please review

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (7) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2242](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6011, opened 2026-07-27
Fix #6008: fold applied file-local `>>` handles by block-ness, not recursion

### LINKED ISSUE #6008: Inconsistent rendering of file-local `>>` handles in `eo-printer`
Two file-local `>>` handles sit side by side in the same `if.` argument list of `eo-runtime/src/main/eo/string/as-decimal.eo`, and the printer spells them differently. The recursive `digits` is hosted at its use site as a moniker, with its argument as a `| n` pipe continuation (`as-decimal.eo:16-31`). The non-recursive `negative` is not: its reference `negative (n.times -1)` at `as-decimal.eo:14-15` has exactly the same shape — an application of the handle in a positional argument slot — yet the binding stays hoisted at `as-decimal.eo:32`. The file is the printer's own output, so one pass produced two spellings for one shape. `eo-runtime/src/main/eo/string/as-fixed.eo:28` shows the same for its `positive` handle.

The split comes from a single clause rather than from anything in the source. `eo:applied-handle` at `eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl:182` looks the binding up as `$owner/o[@name = $name and eo:moniker-binding(.) and eo:recursive-handle(.)][1]`, and the drop template at `merge-monikers.xsl:472` repeats that condition; `eo:recursive-handle` (`merge-monikers.xsl:83`) means `@name = @local`, which `restore-local-names` promotes only for a recursive formation. Recursion is irrelevant to the fold itself: an applied reference cannot host through the ordinary bare-reference path either way, since `eo:resolved-ref` demands `not($ref/o)` (`merge-monikers.xsl:53`), and `inline-cactoos.xsl:285` deliberately leaves an argument-slot application standing (#5983). So a non-recursive handle reached only by applications has no path to a moniker at all, while its recursive twin has one. Both spellings denote the same object — the parser hoists a `>>` binding to its enclosing formation wherever it is written — and the two test packs `print-packs/yaml/list-recursive-applied-dispatch-arg.yaml` and `print-packs/yaml/inline-cactoos-applied-handle-argument.yaml` pin one canonical form each for the same shape.

The fix is to drop the `eo:recursive-handle(.)` clause from `eo:applied-handle` (`merge-monikers.xsl:182`) and from the drop template (`merge-monikers.xsl:472`), so any formation handle whose reachable references are all applied folds into the inlined formation plus a `| args` pipe, keeping the existing first-reference-only rule. Guard it with the lesson of #5983: fold only when the hosted binding prints as a block, so the pipe gets its own line and never has to be spelled inline as the unparsable `|:0`.

Related: #5848 introduced this fold for recursive handles only, and #5732 is the neighbouring but distinct case of a binding that already carries the `@pipe` marker, which `negative` does not.

yegor256:
Closes #6008.

## Problem

The applied-handle fold in `merge-monikers.xsl` — which turns a `>> name` formation handle reached only by applications into the inlined formation plus a `| args` pipe (#5848) — was gated on `eo:recursive-handle`. So a recursive handle folded while its **non-recursive twin of the same shape** stayed hoisted, and the printer emitted two spellings for one object graph.

This was visible in the printer's own output:

- `as-decimal.eo`: the recursive `digits` folded to `[n] >> digits … | n`, but the non-recursive `negative (n.times -1)` — the same shape — stayed hoisted as a separate `if. > [m] >> negative` binding.
- `as-fixed.eo`: same for `positive` / `signed`.

Recursion is irrelevant to the fold: an applied reference cannot host through the ordinary bare-reference path either way (`eo:resolved-ref` demands `not($ref/o)`), so a non-recursive handle reached only by applications simply had no path to a moniker.

## Fix

Replace the recursion gate with `eo:applied-hosted`. An abstract formation handle folds onto its first applied reference when **either**:

- the reference is a **dispatch receiver** (`eo:receiver-ref`) — always safe, since `unnecessary-as` strips the pipe's `@as` cleanly for a reversed dispatch; **or**
- the formation **prints as a multi-line block** (`eo:block-handle`).

Gating a positional-argument fold on block-ness keeps the lesson of #5983: a block forces its enclosing argument list to lay out vertically, so the nameless pipe lands on its own line as `| args` and never as the unparsable `|:0`. A one-line handle (`a.plus b > @`) stays inline and keeps its applied reference standing, exactly as `inline-cactoos-applied-handle-argument` pins it.

`eo:block-handle` approximates the compact one-line (inline-phi) spelling structurally: a formation is a block when it has more than one body binding, or its sole binding is not the `@` decoratee, or that decoratee is itself compound (arguments carrying children of their own).

The recursive receiver/list folds introduced by #5848 are preserved: their references are receivers, or their formations are blocks, so they still fold.

## Changes

- `eo-printer/.../merge-monikers.xsl` — new `eo:block-handle`, `eo:receiver-ref`, `eo:applied-hosted` functions; `eo:applied-handle` and the drop template now gate on `eo:applied-hosted` instead of `eo:recursive-handle`.
- `eo-printer/.../print-packs/yaml/applied-block-handle-argument.yaml` — new print-pack pinning the non-recursive block-argument fold (the twin of `inline-cactoos-applied-handle-argument`, which stays standing).
- `eo-runtime/.../string/as-decimal.eo`, `as-fixed.eo` — regenerated through the fixed printer so their file-local handles read consistently. The re-spelling is graph-preserving (the parser hoists `>> name` bindings identically whether written folded or standing); only the cited handles changed.

## Testing

`XmirTest` (both `printsToEo` and `printsToParseableEo`, 266 cases) and all other `eo-printer` tests pass. The regenerated runtime files were verified to reprint to themselves (idempotent) under the printer's default penalty weights.

---
_Generated by [Claude Code](https://claude.ai/code/session_014ggteC7wiwU1wmzckfFwa8)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -158,9 +158,30 @@
     <xsl:sequence select="if (exists($binding) and (eo:moniker-refs($binding)[1] is $ref)) then $binding else ()"/>
   </xsl:function>
   <!--
-  The recursive-handle binding that an applied reference `$ref` hosts as a
-  moniker, or the empty sequence. An applied reference is a bare "ξ.&lt;name&gt;"
-  restored recursive handle (see `eo:recursive-handle`) carrying argument children
+  Whether the formation `$attr` prints as a multi-line block rather than as a
+  compact one-line spelling. Only a block hosts an applied reference as a "| args"
+  pipe continuation: a block forces its enclosing argument list to lay out
+  vertically, so the nameless pipe gets its own line and never has to be spelled
+  inline as the unparsable "|:0" (the lesson of #5983). An abstract formation
+  prints on one line only through the compact "&lt;value&gt; &gt; [params] &gt;&gt; name"
+  suffix form (Pretty's inline-phi), which needs a single "@" decoratee whose
+  own value is flat; anything else spans several lines. So a formation is a block
+  when it carries more than one body binding (like "digits", with both "@" and
+  "q"), or its sole binding is not that "@" decoratee, or that decoratee is
+  itself compound — an application whose arguments carry children of their own
+  (like "negative", whose "@" is an "if." over three nested branches). A
+  single-"@" formation whose decoratee is a flat application ("a.plus b") stays
+  inline and keeps its applied reference standing (#5983), the "bar 1 2" case.
+  -->
+  <xsl:function name="eo:block-handle" as="xs:boolean">
+    <xsl:param name="attr" as="element()"/>
+    <xsl:variable name="body" select="$attr/o[not(eo:void(.))]"/>
+    <xsl:variable name="phi" select="$body[@name = $eo:phi]"/>
+    <xsl:sequence select="eo:abstract($attr) and (count($body) &gt; 1 or empty($phi) or exists($phi/o/o))"/>
+  </xsl:function>
+  <!--
+  The references that host the binding `$attr` as an applied moniker. An applied
+  reference is a bare "ξ.&lt;name&gt;" formation handle carrying argument children
   — "handle args", a dispatch receiver ("(handle args).read") or a list element
   ("directory.eo"'s "r (walk ...)" in "seq *"). Since neither a bare inline (drops
   the arguments, #5834) nor a reversed dispatch can host it, the handle is inlined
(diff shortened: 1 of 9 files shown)


yegor256:
Heads-up on the `pr-size` check: it fails (~700 lines vs the 200-line cap) and this is inherent to the change, not something I can trim away.

The fix itself is small — `merge-monikers.xsl` swaps the recursion gate on the applied-handle fold for a block-ness gate. But that changes the printer's canonical output for **every** runtime source that has a file-local `>>` handle reached only by applications, and `MjFormat` enforces canonical formatting per-PR. So the printer change and the reformatted sources **have to land together** — a PR with only the XSLT change would fail the format check (`5 of 161 EO source(s) are not formatted canonically`), and a PR with only the reformats wouldn't compile against the old printer.

Seven sources are reformatted: `as-decimal`, `as-fixed`, `arc-sine`, `power`, `length`, `slice`, and `socket`. All are graph-preserving re-spellings (the parser hoists a `>> name` binding identically whether it's written folded or standing). `socket.eo` dominates the diff: its 145-line `impl` helper is applied at a single site, so folding it inlines the whole formation there.

Could you waive `pr-size` for this one? Alternatively, if the `socket.eo` `impl` inline (or the breadth generally) is undesirable, I can add a size guard so only small/medium handles fold and large ones keep standing — that shrinks the diff but deviates from the issue's "fold any formation handle whose reachable references are all applied". Happy to go either way.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (700 >= 100); -8 for way too many hits-of-code (700 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2741](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6012, opened 2026-07-27
fix(#5978): handle zero-width regex matches in string.replaced [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #5978: `string.replaced` method fails with `ExFailure` for regex patterns that can match an empty string
## What happens

`string.replaced` fails whenever the target regex admits a zero-width (empty) match. Instead of substituting, it throws a deeply-nested `org.eolang.ExFailure` (a long `Error in "…string.replaced…"` chain). Patterns that only ever match non-empty text work fine.

Verified against a fresh `eo-runtime` build by dataizing the real `EOreplaced` with a real `EOregex` target, compared with Java's `String.replaceAll`:

| call | EO `replaced` | Java `replaceAll` |
|---|---|---|
| `replaced "abc" (regex "/x*/") "-"`      | **ExFailure** | `"-a-b-c-"` |
| `replaced "aaa" (regex "/a*/") "X"`      | **ExFailure** | `"XX"` |
| `replaced "" (regex "/a*/") "X"`         | **ExFailure** | `"X"` |
| `replaced "a1b2c" (regex "/[0-9]/") "#"` | `"a#b#c"` (ok) | `"a#b#c"` |
| `replaced "hello" (regex "/l+/") "L"`    | `"heLo"` (ok) | `"heLo"` |
| `replaced "abab" (regex "/a/") ""`       | `"bb"` (ok) | `"bb"` |

So any nullable pattern (`x*`, `a*`, `\d*`, `.?`, an alternation with an empty branch, …) crashes `replaced`, even though the underlying regex object matches such patterns fine.

## Root cause

`eo-runtime/src/main/eo/string/replaced.eo`, the `rec-replaced` loop:

```
block.exists.if > [block accum start] >> rec-replaced
  rec-replaced
    block.next
    concat.
      accum.concat
        reinit.slice
          start
          block.from.minus start
      replacement
    block.to
  ...
```

Each step slices the text before the match — `reinit.slice start (block.from - start)` — and then continues from `start = block.to`. For a zero-width match `block.from == block.to`, so `start` is not advanced past the match, and on the next iteration `block.from.minus start` is zero or negative, driving the recursion into an invalid `slice` / non-terminating chain (the failure trace loops through `string.replaced:16-17`). A standard regex substitution steps over a zero-width match by emitting the replacement and advancing one character; `rec-replaced` has no such handling.

## Suggested fix

Handle the empty-match case in `rec-replaced`: when `block.from == block.to`, emit the replacement, copy one character of `reinit`, and advance `start`/the match position by one (mirroring how the regex engine itself steps over zero-width matches), so nullable patterns terminate with the same result as `replaceAll`.

waterWang:
## Description

Fixes #5978

The `string.replaced` method fails with `ExFailure` when the target regex pattern can match an empty string (zero-width match). Patterns like `x*`, `a*`, `\d*`, `.?` or any alternation with an empty branch cause the algorithm to loop indefinitely because `block.from == block.to` for zero-width matches, so `start` never advances.

### Root cause

In `rec-replaced`, each recursion step slices the text before the match — `reinit.slice start (block.from - start)` — and then continues from `start = block.to`. For a zero-width match, `block.from == block.to`, so `start` is not advanced, and the next iteration finds the same match again, producing an infinite loop.

### Fix

Added a check for zero-width matches inside `rec-replaced`:
- When `block.from == block.to`, emit the replacement, copy one character from the original string, and advance `start` by one position
- This mirrors how Java's `String.replaceAll` handles zero-width matches (it steps over the match by advancing one character)

### Tests added

- `tests-replaces-zero-width-pattern`: `replaced "abc" (regex "/x*/") "-"` → `"-a-b-c-"`
- `tests-replaces-zero-width-pattern-all`: `replaced "aaa" (regex "/a*/") "X"` → `"XX"`
- `tests-replaces-zero-width-empty-string`: `replaced "" (regex "/a*/") "X"` → `"X"`

### DIFF
--- a/eo-runtime/src/main/eo/string/replaced.eo
+++ b/eo-runtime/src/main/eo/string/replaced.eo
@@ -14,15 +14,26 @@
   matched.exists.not.if > @
     string text! >> reinit
     block.exists.if > [block accum start] >> rec-replaced
-      rec-replaced
-        block.next
-        concat.
-          accum.concat
+      block.from.eq block.to.if > @
+        rec-replaced
+          block.next
+          concat.
+            accum.concat
+              replacement
             reinit.slice
-              start
-              block.from.minus start
-          replacement
-        block.to
+              block.from
+              1
+          block.from.plus 1
+        rec-replaced
+          block.next
+          concat.
+            accum.concat
+              reinit.slice
+                start
+                block.from.minus start
+            replacement
+          block.to
+      |
       string
         accum.concat
           reinit.slice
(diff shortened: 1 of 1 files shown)


waterWang:
Closing as duplicate of PR #6000 (same issue #5978, same fix approach).

Issue #6013, opened 2026-07-27
Deferred xslint checks in .xslint.yml are not enabled


0pdd:
The puzzle `6006-b04fd4e0` from #6006 has to be resolved: 

https://github.com/objectionary/eo/blob/e8d7a0b55cec4a283b895263062aa29ac2b244ea/.xslint.yml#L9-L13

The puzzle was created by @yegor256 on 27-Jul-26. 

Estimate: 60 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
The puzzle `6006-b04fd4e0` has disappeared from the source code, that's why I closed this issue.

PR #6014, opened 2026-07-27
Enable redundant-namespace and unused-function xslint checks

maxonfjvipon:
Progress on #6013 — clearing two of the deferred xslint checks by fixing their findings, so they now enforce.

- **`redundant-namespace-declarations`** — `resolve-self.xsl`, `validate-before-stars.xsl`, and `validate-objects-count.xsl` each declared `xmlns:xs` with `exclude-result-prefixes="xs"` but use the `xs:` prefix nowhere. Removed both; behaviour is unchanged since the prefix was dead.
- **`unused-function`** — `eo:bytes-to-number` in `_funcs.xsl` is called from nowhere in the repository. Removed it.

Both checks are switched on in `.xslint.yml`; the whole tree still lints clean.

The puzzle stays because the remaining eleven checks are deliberately left for a maintainer's call rather than an automated edit: some are opinionated (`not-using-schema-types`, `not-creating-element-correctly`), some change behaviour (`not-using-output`, `use-choose-without-otherwise`), and `unused-variable` would delete symbol constants from `_specials.xsl` — `$eo:lb` ⟦, `$eo:rb` ⟧, `$eo:arrow` ↦ and the like — which are core EO notation and probably want a human decision before removal.


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+722](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6016, opened 2026-07-27
Fix #6015: keep the `&gt;&gt; name` handle on a pipe-application line

### LINKED ISSUE #6015: `Parser fails to retain local file handle in pipe-application line, causing references to default to a global scope`
A pipe-application line accepts a `>> name` file-local handle and then throws the name away. In `eo-runtime/src/main/eo/string/scanf.eo:241`, writing `| >> tokens` instead of `| > tokens` leaves the two references at `eo-runtime/src/main/eo/string/scanf.eo:164` and `:167` (`tokens.at`) resolved against a global `Φ.tokens` rather than the sibling binding, and `eo:format` prints the line back as a bare `| >>` with the name simply gone, while the two references stay put.

The cause is a missing call in the parser. `LnPipe.into` emits the object with the cactus `@name` returned by `suffix.attribute(...)` at `eo-parser/src/main/java/org/eolang/parser/LnPipe.java:89-92`, but never calls `emit.local(suffix.handle())` — the call every other handle-bearing line makes (`eo-parser/src/main/java/org/eolang/parser/LnFormation.java:201`, `LnOnlyPhi.java:140`, `LnReversed.java:124`, `LnVoid.java:98`, `ChainEmission.java:87`). By the spec that call is mandatory here: R-3.14.1 says a pipe's tail is parsed exactly as an application's argument list plus suffix, R-3.10.12 accepts a handle uniformly wherever bare `>>` is accepted, and R-9.2.3 fixes the emission as the cactus `@name` plus an `@local="name"` marker. Without the marker `resolve-local-names` has no table entry, so every reference silently escapes to a global that does not exist, and the duplicate-handle check cannot see the name either. The pipe packs under `eo-parser/src/test/resources/org/eolang/parser/eo-syntax/` (`pipe-vertical.yaml`, `pipe-vertical-named.yaml`, `pipe-vertical-bound.yaml`, `pipe-chained.yaml`) all use a bare `| >>`, so no test pins the handle-carrying form.

The fix is to add the same two-line guard the other lines use — `if (!suffix.handle().isEmpty()) { emit.local(suffix.handle()); }` — right after the `emit.object(...)` call in `LnPipe.into` (`LnPipe.java:92`), mirroring `LnFormation.java:200-202`.

That alone restores the binding, but the printer needs a matching change to finish the round trip. In `eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl` the drop template at `:236` keeps `@local` on a multi-referenced binding such as `tokens`, yet the key that rewrites references back to the readable handle (`:197`, applied at `:205`) matches only `o[@local and (@base=$eo:empty or eo:recursive(., @name))]` — voids and recursive handles. A pipe-application handle is neither, so its references print as the synthetic cactus placeholder `v241_2` instead of `tokens`; widening that key to any binding whose `@local` survives the drop template covers this shape.

yegor256:
Fixes #6015.

## Problem

A pipe-application line accepts a `| args >> name` file-local handle (§3.14, R-3.10.12) but the parser threw the name away. `LnPipe.into` emitted the object with the cactus `@name` returned by `suffix.attribute(...)`, but never called `emit.local(suffix.handle())` — the call every other handle-bearing line makes (`LnFormation`, `LnOnlyPhi`, `LnReversed`, `LnVoid`, `ChainEmission`).

Without the `@local` marker, `resolve-local-names` had no table entry, so every sibling reference silently escaped to a non-existent global `Φ.name`, and `eo:format` reprinted the line as a bare `| args >>` with the name gone.

## Fix

**Parser** (`LnPipe.java`) — emit the handle marker, mirroring `LnFormation.java:200-202`:

```java
if (!suffix.handle().isEmpty()) {
    emit.local(suffix.handle());
}
```

**Printer** (`restore-local-names.xsl`) — a pipe object carries the `@pipe` marker, so neither `inline-cactoos` nor `merge-monikers` ever folds it or rewrites its references (both skip `@pipe` bindings). Its readable handle therefore lives or dies entirely in `restore-local-names`, which previously rewrote references only for voids and recursive formations. Two minimal changes:

- Widen the `void-handle` key to also match `@pipe` handles, so their references are rewritten back to the readable `name` instead of a synthetic `vL_P` placeholder.
- Guard the `@local`-drop template with `not(@pipe)`, so a pipe handle keeps its marker at any use count (it is never inlined or merged, so its name must survive here).

The pipe's cactus `@name` is deliberately **not** promoted — the `| args >> name` line reads its `>> name` from `@local`.

## Tests

- `eo-parser` `pipe-vertical-handle.yaml` — pins the parser emission: the pipe object carries `@local` + `@pipe` + a cactus `@name`, both references resolve to that cactus name, and none escapes to `Φ.tokens`.
- `eo-printer` `pipe-named-handle.yaml` and `pipe-named-handle-multi-referenced.yaml` — pin the full round trip for the single- and multi-reference shapes.

All 1517 `eo-parser` tests and 270 `eo-printer` tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hq2LzATcyPzAs7QkdAGyLw)_

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnPipe.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnPipe.java
@@ -90,6 +90,9 @@ final class LnPipe implements Line {
             suffix.attribute(this.span.line(), this.span.indent()),
             null, this.span.line(), this.span.indent()
         );
+        if (!suffix.handle().isEmpty()) {
+            emit.local(suffix.handle());
+        }
         emit.pipe();
         if (suffix.constant()) {
             emit.constant();
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/pipe-vertical-handle.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/pipe-vertical-handle.yaml
@@ -1,0 +1,26 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# A pipe-application line may carry a `>> name` file-local handle (§3.14,
+# R-3.10.12): the parser emits the pipe object with its cactus `@name` plus a
+# `@local='name'` marker, exactly as every other handle-bearing line does, so
+# `resolve-local-names` binds the sibling references to the cactus name rather
+# than letting them escape to a global `Φ.tokens` (#6015). The three earlier
+# pipe packs (`pipe-vertical`, `pipe-vertical-named`, `pipe-vertical-bound`)
+# only exercise a bare `| >>`, so none pins the handle-carrying form.
+sheets: []
+asserts:
+  - /object[not(errors)]
+  - /object/o[@name='app']/o[@base='ξ.base' and @name='a🌵5-2' and @local='tokens' and @pipe='']
+  - /object/o[@name='app']/o[@name='a' and @base='ξ.a🌵5-2.at']
+  - /object/o[@name='app']/o[@name='b' and @base='ξ.a🌵5-2.at']
+  - /object/o[@name='app'][count(o[@base='ξ.a🌵5-2.at'])=2]
+  - /object[count(//o[@base='Φ.tokens'])=0]
+input: |
+  [] > app
+    tokens.at 0 > a
+    tokens.at 1 > b
+    [x] > base
+    | >> tokens
+      42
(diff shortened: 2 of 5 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (127 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2793](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6019, opened 2026-07-27
#5973: reject non-finite input in string.as-decimal instead of recursing forever

### LINKED ISSUE #5973: `string.as-decimal` and `string.as-fixed` recurse indefinitely with non-finite input values
## What happens

`string.as-decimal` recurses forever for a non-finite number. `nan`, `+inf` and `-inf` all drive its internal `digits` helper into unbounded recursion; instead of a value or a clean error the caller gets an `org.eolang.ExFailure` whose message is a ~2500-frame nested `Error in "…as-decimal.digits…"` chain (≈200 KB of text).

Verified against a fresh `eo-runtime` build by dataizing the real `EOas_decimal`:

| call | result |
|---|---|
| `as-decimal 42.7` | `"42"` (ok) |
| `as-decimal -0.0` | `"0"` (ok) |
| `as-decimal 1e20` | `"100000000000000000000"` (ok) |
| `as-decimal nan`  | **ExFailure**, ~2582 nested `as-decimal.digits` frames |
| `as-decimal pinf` | **ExFailure**, same non-terminating recursion |
| `as-decimal ninf` | **ExFailure**, same |

`string.as-fixed` inherits the same failure: its `positive` helper computes `ip` from the input and calls `as-decimal ip`, so `as-fixed nan …` / `as-fixed pinf …` are also non-terminating.

## Root cause

`eo-runtime/src/main/eo/string/as-decimal.eo`, the `digits` helper:

```
[n] > digits
  if. > @
    n.lt 10
    as-char (n.floor.plus 48)
    concat.
      digits q
      as-char (plus. (floor. (n.minus (q.times 10))) 48)
  floor. > q
    n.div 10
```

The recursion is meant to bottom out on `n.lt 10`. For a non-finite `n` every comparison with `10` is `false`, and `q = floor(n / 10)` stays `nan`/`inf`, so `digits` calls itself with an unchanging argument and never terminates. The top-level `n.lt 0` guard doesn't help either: `nan.lt 0` and `pinf.lt 0` are both `false`, so `nan`/`+inf` fall straight into `digits`.

## Suggested fix

Guard non-finite input at the top of `as-decimal` (and/or in `digits`) with `n.is-finite`, returning a single clean `ExFailure` (e.g. "Can't write a non-finite number as decimal") instead of recursing — mirroring how the neighbouring numeric atoms already special-case `nan`/`inf`. `as-fixed` should then surface the same clean error (or its own `cant-print` fallback) rather than inheriting the runaway recursion.

morphqdd:
Closes #5973.

`digits` bottoms out on `n.lt 10`, but every comparison with a non-finite `n` is false, and `q = floor(n / 10)` stays `nan`/`inf`, so `digits` called itself with an unchanging argument and never terminated (about 2500 nested frames for `nan`/`+-inf`).

Guards non-finite input at the top with `n.is-finite`, calling the void `trouble` attribute with a message instead of entering the recursion. This is the same bottom-object convention already used by fallback params elsewhere in this codebase, just without a caller-supplied fallback since `as-decimal` takes none.

`string.as-fixed` calls `as-decimal` internally, so it now surfaces the same clean failure instead of inheriting the runaway recursion.

Added tests for `nan`, `+inf` and `-inf` all expecting a clean throw instead of a hang.

Tests: `mvn -pl eo-runtime -o test -Dtest='EO_string.EOas_decimalTest'` - all passing.

### DIFF
--- a/eo-runtime/src/main/eo/string/as-decimal.eo
+++ b/eo-runtime/src/main/eo/string/as-decimal.eo
@@ -9,30 +9,33 @@
 +spdx SPDX-License-Identifier: MIT
 
 [n] > as-decimal
-  if. > @
-    n.lt 0
-    if. > [m] >> negative
-      m.floor.eq 0
-      digits m
-      "-".as-bytes.concat
+  n.is-finite.not.if > @
+    trouble
+      "Can't write a non-finite number as decimal"
+    if.
+      n.lt 0
+      if. > [m] >> negative
+        m.floor.eq 0
         digits m
-    | (n.times -1)
-    [n] >> digits
-      if. > @
-        n.lt 10
-        as-char
-          n.floor.plus 48
-        concat.
-          digits q
+        "-".as-bytes.concat
+          digits m
+      | (n.times -1)
+      [n] >> digits
+        if. > @
+          n.lt 10
           as-char
-            plus.
-              floor.
-                n.minus
-                  q.times 10
-              48
-      floor. > q
-        n.div 10
-    | n
+            n.floor.plus 48
+          concat.
+            digits q
+            as-char
+              plus.
+                floor.
+                  n.minus
+                    q.times 10
+                48
+        floor. > q
+          n.div 10
+      | n
 
   eq. ++> tests-negative-below-one-drops-sign
     "0"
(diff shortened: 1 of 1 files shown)


morphqdd:
@yegor256, please review

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2304](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6022, opened 2026-07-28
Combine receiver-only reversed dispatch into hybrid inline-phi (#5954)

### LINKED ISSUE #5954: Printer erroneously maintains reversed dispatch in verbose form instead of combining into ++>

yegor256:
Fixes #5954.

## Problem

In `Pretty.phi()`, the `applied` predicate ended with a clause
`!(decoratee.reversed && decoratee.children.size() <= 1)` that withheld the
hybrid inline-phi collapse from a reversed dispatch carrying only its
receiver (`not.`), while a reversed dispatch that also carried arguments
(`eq.`, `if.`) got the collapse. As a result the printer kept blocks like

```
++> tests-compares-string-with-negative-infinity
  not. > @
    ninf.eq "друг"
```

in their verbose three-line form as the printer's own fixpoint, instead of
collapsing to

```
not. ++> tests-compares-string-with-negative-infinity
  ninf.eq "друг"
```

## Fix

The withheld clause was justified as mirroring the rejection in `flat`
(`node.reversed && node.children.size() <= 1`), but that rejection concerns
a *horizontal* one-line spelling, which `not.` has no form for. The hybrid
is not horizontal: it keeps `not.` on the head line and lays the receiver
out one indent beneath — exactly the shape `LnOnlyPhi` accepts, its `bare()`
leaving the φ `Openness.OPEN` so the deeper line attaches as the receiver,
with no minimum argument count enforced on close. The verbose and the
collapsed forms parse to the same XMIR tree, so the extra line and indent
buy nothing.

Dropping the clause leaves `applied` as
`!decoratee.abstractt && !decoratee.children.isEmpty()`. The following
`unnamed` guard still withholds the collapse when a name appears anywhere in
the decoratee's subtree, so #5598 and #5604 stay fixed.

## Tests

- Added `hybrid-phi-reversed-receiver-only.yaml`, covering a receiver-only
  reversed dispatch (`not.`) collapsing into the `++>` hybrid.
- Updated `reversed-with-data-receiver-args.yaml`: with the fix, the
  paren-free hybrid (`not. > [] > app` with the receiver beneath) now beats
  the paren-heavy suffixed one-liner `(...).not > [] > app` on penalty, and
  the new form is a fixpoint. Also updated the `phi()` javadoc.

All 272 `XmirTest` cases pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HTuGdDCjqNXrbcV924dAQm)_

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -331,11 +331,16 @@ final class Pretty {
      * built at all, the hybrid is used unconditionally. Either way the result
      * is only a candidate — the penalty check in {@link #shaped} keeps it only
      * when it beats the plain vertical rendering. A formation decoratee
-     * (its bindings are vertical, not arguments) and a receiver-only
-     * reversed dispatch ({@code not.} with just its receiver, mirroring the
-     * rejection in {@link #flat}) have no hybrid form, so they yield empty
-     * and keep the verbose layout; a reversed dispatch that also carries
-     * arguments ({@code if.} with its branches) does get the hybrid.</p>
+     * (its bindings are vertical, not arguments) has no hybrid form, so it
+     * yields empty and keeps the verbose layout. A reversed dispatch does get
+     * the hybrid whether it carries just its receiver ({@code not.}) or also
+     * arguments ({@code if.} with its branches): unlike {@link #flat}, which
+     * rejects a receiver-only reversed dispatch because it has no horizontal
+     * one-line spelling, the hybrid keeps the dispatch on the head line and
+     * lays the receiver out beneath — a shape {@code LnOnlyPhi} accepts, its
+     * {@code bare()} leaving the φ {@code Openness.OPEN} so the deeper line
+     * attaches as the receiver, with no minimum argument count on close (issue
+     * #5954).</p>
      *
      * <p>The hybrid is also withheld when any line in the decoratee's whole
      * subtree carries a name suffix ({@code [left] >>}, {@code malloc.for >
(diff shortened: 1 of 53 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1117 >= 100); -8 for way too many hits-of-code (1117 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2785](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6023, opened 2026-07-28
Fix #6021: keep a single-use `&gt;&gt;` handle standing when applied from a nested scope

### LINKED ISSUE #6021: Single-use formation handle mislocated in nested formation argument list, resulting in stray argument and orphaned reference
In `eo-runtime/src/main/eo/string/trimmed-left.eo:34`, spelling the `is-ws` helper as `or. > [c] >> is-ws` rather than `or. > [c] > is-ws` makes `eo:format` print broken code. The whole `is-ws` formation is moved out of `trimmed-left`'s attribute list into the argument list of the inner `if.` at `eo-runtime/src/main/eo/string/trimmed-left.eo:30`, where it becomes a stray fourth argument of a three-argument `if.`, and the reference at `:31` prints as `^.v34_2 (b.slice index 1)!` — a name nothing declares. The output reparses, so `eo:format` writes it to disk and only the compiler complains; a second format pass then also erases the handle's name, leaving `or. > [c] >>`. Minimal case:

```eo
+package foo

[text] > bar
  [index] >> helper
    if. > @
      is-ws index
      index
      index
  [c] >> is-ws
    c.eq 1 > @
```

prints back as:

```eo
[text] > bar
  [index] >> helper
    if. > @
      c.eq 1 > [c] >> is-ws
      ^.v9_2 index
      index
      index
```

The relocation branch at `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:177` is responsible. It moves a single-use formation handle down to sit directly above its applied reference and rewrites the reference into a `| args` pipe — the sibling-separated case of #5840 and the dispatch-receiver case of #5844. Both shapes it was written for keep the reference in the handle's own scope, but nothing in the guard requires that. A reference written inside a nested formation reaches the handle as `ξ.ρ.<cactus>`, so the relocated copy lands in the nested scope while the pipe node keeps the `ρ` climb; the adjacency test at `to-eo-tree.xsl:244` (`@base` equal to `ξ.` plus the preceding sibling's `@name`) no longer matches and the node prints as an ordinary application on the cactus name. The drop template at `inline-cactoos.xsl:285` then deletes the original binding, because its `eo:arg-applied` keep-guard (`:379`) covers only a reference whose own `@as` is `αN`, and a reversed-dispatch receiver carries `@as="ρ"`.

Add a branch ahead of `:177` that copies such a reference verbatim, exactly as the positional-argument branch at `:149` already does, gated on the reference's nearest formation ancestor not being the handle's owner (`not(ancestor::o[eo:abstract(.)][1] is $target/..)`), and add the matching keep condition to the drop template at `:285` so the binding stays in its own scope. `merge-monikers`' `eo:kept-local-ref` then rewrites the reference back to `is-ws` and already sheds the `ρ` climb (#5893/#5917).

yegor256:
Closes #6021.

## Problem

The `#5840`/`#5844` relocation branch in `inline-cactoos.xsl` moved a single-use abstract `>>` formation to sit directly above its applied reference and rewrote that reference into a `| args` pipe — without checking that the reference lived in the handle's **own owner scope**.

When the reference was written inside a *nested* formation body, it reached the handle through a `ρ` climb (`ξ.ρ.<name>`). The relocated formation copy then landed in the nested scope while the pipe node kept the `ρ` climb, so `to-eo-tree`'s adjacency test (`to-eo-tree.xsl:244`) no longer matched the copy directly above it. The node printed as an ordinary application on the cactus name (`^.v7_2 index`), the formation copy became a stray extra argument, and a second `eo:format` pass then erased the handle's name altogether — leaving broken, name-orphaned code such as `or. > [c] >>`.

Minimal reproduction (before the fix):

```eo
[text] > bar
  [index] >> helper
    if. > @
      is-ws index
      index
      index
  [c] >> is-ws
    c.eq 1 > @
```

printed back as:

```eo
[text] > bar
  [index] >> helper
    if. > @
      c.eq 1 > [c] >> is-ws
      ^.v7_2 index
      index
      index
```

## Fix

Guard the relocation with a new `eo:nested-applied` predicate: a single-use formation reached only through a reference in a nested formation scope is left standing under its `@local` name instead of being relocated — exactly as the positional-argument case of `#5983`.

- A new branch ahead of the relocation branch copies such a reference verbatim (keeping its arguments), gated on the reference's nearest formation ancestor not being the handle's owner: `not(ancestor::o[eo:abstract(.)][1] is $target/..)`.
- The drop template keeps the binding in its own scope via the matching `eo:nested-applied` keep-condition.

`merge-monikers` then rewrites the cactus base back to the readable handle and sheds the `ρ` climb (`#5893`/`#5917`), so `is-ws index` survives and the `[c] >> is-ws` handle is preserved. The sibling (`#5840`) and receiver (`#5844`) relocations, whose references share the handle's own scope, still fire.

## Tests

- `eo-packs/print/inline-cactoos-nested-argument.yaml` — asserts at the XSL level that the handle is kept in place, the reference stays verbatim inside the nested formation, and no `@pipe` is emitted.
- `print-packs/yaml/inline-cactoos-applied-handle-nested.yaml` — end-to-end print pack that reproduces the broken output and pins the correct round-tripping result.

All 312 `eo-printer` tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PSy5rBd4BSQ1dCG85hDwk6)_

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -152,6 +152,32 @@
             </xsl:copy>
           </xsl:when>
           <!--
+          The same applied reference, but written inside a nested formation
+          body rather than in the handle's own owner scope. Such a reference
+          reaches the handle as `ξ.ρ.` (a `ρ` climb into the enclosing scope
+          where the handle lives), so the relocation branch below would land
+          the fresh formation copy in the nested scope while the pipe node
+          keeps the `ρ` climb: `to-eo-tree`'s adjacency test then sees the
+          pipe's `ξ.ρ.name` base no longer equal `ξ.name` of the copy directly
+          above it and prints the node as an ordinary application on the cactus
+          name, leaving the copy a stray extra argument and orphaning the
+          reference (#6021). An argument list buried in a nested scope has no
+          more room for a relocated predecessor than the positional-argument
+          case above (#5983), so leave the handle standing under its `@local`
+          name instead — copy the reference verbatim, keeping its arguments, so
+          the drop template below keeps the binding in its own scope and
+          "merge-monikers" rewrites the reference back to the handle
+          (`eo:kept-local-ref`), shedding the `ρ` climb (#5893/#5917). Gated on
+          the reference's nearest formation ancestor not being the handle's
+          owner, so the sibling (#5840) and receiver (#5844) relocations below,
+          whose references share the handle's own scope, still fire.
+          -->
+          <xsl:when test="eo:abstract($target) and (o or @name) and not(eo:multi-referenced($target, $name)) and not(ancestor::o[eo:abstract(.)][1] is $target/..)">
+            <xsl:copy>
+              <xsl:apply-templates select="node()|@*"/>
+            </xsl:copy>
+          </xsl:when>
+          <!--
           The same applied reference, but not standing as the formation's
           immediate following sibling, so the adjacent guard above misses it
           and the bare-reference `otherwise` would drop the argument and name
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (167 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2761](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6006, opened 2026-07-27
Unchecked XSLT stylesheets might contain stylistic, semantic, and logical mistakes


maxonfjvipon:
The compilation pipeline here is XSLT all the way down — parsing in
`eo-parser/.../parser/parse`, transpilation in `eo-maven-plugin/.../transpile`,
printing in `eo-printer/.../print`, plus the shake and probe passes. CI already
lints nearly everything: `xcop` for XML formatting, `qulice`, `codenarc`,
`shellcheck`, `yamllint`, `reuse`, `markdown-lint`. Nothing checks the *XSLT*
itself beyond well-formedness — no linter looks at the stylesheet logic.

[xslint](https://github.com/xslint/xslint) fills that gap. It verifies each
stylesheet is well-formed and every XPath compiles, then flags stylistic,
semantic, and logical mistakes, each pinpointed to line and column.

I ran it over the 47 stylesheets here. It reports **zero false positives**.
Excluding the vendored `xs3p.xsl` and the test fixtures, the pipeline's own
stylesheets carry 122 findings across 13 checks:

| count | check |
|------:|-------|
| 60 | not-creating-element-correctly (`xsl:element` with a static name) |
| 30 | not-using-schema-types |
| 8 | short-names |
| 7 | unused-variable |
| 4 | not-using-output |
| 3 | redundant-namespace-declarations |
| 2 each | null-output-from-stylesheet, stylesheet-has-no-templates, too-many-small-templates |
| 1 each | monolithic-design, empty-content-in-instructions, use-choose-without-otherwise, unused-function |

Some are opinionated (`not-using-schema-types`); others are concrete dead code
— `eo:bytes-to-number` in `_funcs.xsl` is defined but called nowhere, and seven
symbol variables in `_specials.xsl` (`$eo:delta`, `$eo:lb`, `$eo:rb`, …) are
declared but referenced nowhere in the repo.

Proposal: add an `xslint` workflow following the one-linter-per-file pattern,
with a `.xslint.yml` that excludes the vendored and test stylesheets. To keep
the build green from day one, the config would enforce the checks the pipeline
already satisfies — regression protection — and defer the thirteen above,
enabling them one at a time as the findings are cleared. A pull request
follows.

0pdd:
@maxonfjvipon the puzzle [#6013](https://github.com/objectionary/eo/issues/6013) is still not solved.

0crat:
[...]

PR #6044, opened 2026-07-28
Enable four more xslint checks

maxonfjvipon:
Bumps the xslint action to 0.0.9 (xslint 0.0.14) and turns on four more checks
that the pipeline now passes.

Two need no source change — after the recent xslint audit, `oversized-template`
(a single template over 100 XSLT elements) and `function-complexity` (a
function over 50) both report zero here. Two took small fixes:
`not-using-output` — declared `xsl:output` on the two pipeline stages that
lacked it (`transpile/package.xsl`, `parse/mandatory-as.xsl`); and
`short-names` — gave the eight single-letter variables real names, e.g.

```xsl
<xsl:variable name="object" select="."/>
```

instead of `$o`, and `$t`→`$type`, `$p`→`$package`, `$c`→`$class`.

The config now defers only three checks — `not-creating-element-correctly`,
`too-many-templates`, and `unused-variable` — each still carrying findings to
work through under #6006. Also renames two stale config keys that the audit
renamed upstream (`too-many-small-templates`→`too-many-templates`,
`monolithic-design`→`oversized-template`) and drops `not-using-schema-types`,
which xslint removed as not a genuine best practice.

### DIFF
--- a/.github/workflows/xslint.yml
+++ b/.github/workflows/xslint.yml
@@ -16,6 +16,6 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7
-      - uses: xslint/xslint-action@0.0.7
+      - uses: xslint/xslint-action@0.0.9
         with:
           config: .xslint.yml
--- a/.xslint.yml
+++ b/.xslint.yml
@@ -13,9 +13,5 @@ exclude:
 #  xslint check is already enforced across the whole codebase.
 rules:
   not-creating-element-correctly: "off"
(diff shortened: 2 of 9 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+750](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6045, opened 2026-07-28
Convert static xsl:element to literal result elements

maxonfjvipon:
Turns on `not-creating-element-correctly` by rewriting the 60 static
`xsl:element` uses across 25 stylesheets to literal result elements:

```xsl
<xsl:element name="o">        <o>
  ...                    →      ...
</xsl:element>                </o>
```

For a static name the two are identical — same element, same in-scope
namespace — so no transform output changes; the literal form is just the
idiomatic, more readable one, which is why the check flags the verbose spelling.
Dynamic `xsl:element` (an AVT or computed name) is left untouched.

All 25 files stay well-formed and pass xcop, and the pipeline now reports zero
findings for this check. Config drops the matching `off` line. Two checks remain
deferred under #6006: `too-many-templates` and `unused-variable`.


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (275 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+758](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6046, opened 2026-07-28
#6026: wrap unescapeBody in LnTextBlock to raise ParseError, not NumberFormatException

### LINKED ISSUE #6026: `LnTextBlock` crashes with raw `NumberFormatException` on invalid escape inside a text block
## What happens

`LnTextBlock` (`eo-parser/src/main/java/org/eolang/parser/LnTextBlock.java:58`) calls `Emissions.unescapeBody(...)` directly, with no try/catch around it.

The plain-string literal path handles the same call differently: `Emissions.java` around lines 129-138 wraps the identical `unescapeBody` call in a `try { ... } catch (NumberFormatException) { throw new ParseError(...) }`. The text-block (`"""..."""`) path has no such wrapper.

Only `ParseError` is caught upstream (in `Eo.continueTextBlock` and `Eo.dispatch`), so any other exception escapes `EoSyntax.parsed()` uncaught.

I verified this by compiling `eo-parser` and running `EoSyntax` on real source files:

- `"\uD800"` (invalid escape) inside a regular string produces a clean `<errors>` entry: `[2:2] error: 'invalid unicode or octal escape in string literal'`.
- The exact same escape inside a `"""` text block instead throws: `java.lang.NumberFormatException: unicode escape \uD800 is a lone surrogate...`
- `\uZZZZ` inside a text block throws `NumberFormatException: For input string: "ZZZZ" under radix 16`.
- `\400` inside a text block throws `NumberFormatException: octal escape \400 is out of range...`

## What should happen

The text-block path should wrap the `unescapeBody` call the same way the string-literal path does, converting `NumberFormatException` into a `ParseError` at the text block's line/position, so a malformed escape inside a `"""` text block is reported as a normal parse error instead of crashing the parser.

## Steps to reproduce

1. Compile `eo-parser`.
2. Parse a `.eo` source containing a text block with an invalid escape, e.g.:
   ```
   [] > foo
     """
     \uD800
     """ > @
   ```
3. Observe `EoSyntax.parsed()` throws `NumberFormatException` instead of returning an `<errors>` XMIR entry.
4. Compare with the same invalid escape inside a regular `"..."` string, which is reported cleanly.

morphqdd:
Closes #6026.

`LnTextBlock.into` called `Emissions.unescapeBody` directly, with no try/catch. The plain string-literal path (`Emissions.openValue`) wraps the same underlying call in a `try { ... } catch (NumberFormatException) { throw new ParseError(...) }`. The text-block path had no equivalent, so a malformed escape inside a `"""` text block (an out-of-range octal escape, an invalid or truncated `\uXXXX`, a lone surrogate) crashed `EoSyntax.parsed()` with a raw `NumberFormatException` instead of producing a normal `<errors>` entry.

Fix: wrap the `unescapeBody` call the same way the string-literal path already does, converting `NumberFormatException` into a `ParseError` at the text block's line/indent.

Also removed the class's `@SuppressWarnings("PMD.UnnecessaryLocalRule")` — after this change qulice reported it as unused (the `joined` local is no longer flagged by that rule once its initializer is a try/catch block).

Tests: added `LnTextBlockTest.rejectsInvalidEscapeInTextBlockBody`, asserting a text-block body with an invalid `\uD800` escape throws `ParseError` rather than `NumberFormatException`. Verified it fails with the original `NumberFormatException` when the fix is reverted, and passes with the fix in place.

`mvn -pl eo-parser test -o -Dtest='LnTextBlockTest,EoSyntaxTest'` - 940 tests, all passing.
`mvn -pl eo-parser qulice:check -Pqulice` - clean, no violations in the changed files.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnTextBlock.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnTextBlock.java
@@ -23,7 +23,6 @@ package org.eolang.parser;
  *
  * @since 0.1
  */
-@SuppressWarnings("PMD.UnnecessaryLocalRule")
 final class LnTextBlock implements Line {
 
     /**
@@ -55,9 +54,19 @@ final class LnTextBlock implements Line {
         final Suffix suffix = new Suffix(
             tokens.tail(), this.span, this.span.indent() + tokens.cursor()
         );
-        final String joined = Emissions.unescapeBody(
-            String.join(String.valueOf('\n'), globals.tbody())
-        );
+        final String joined;
+        try {
+            joined = Emissions.unescapeBody(
+                String.join(String.valueOf('\n'), globals.tbody())
+            );
+        } catch (final NumberFormatException ex) {
+            final ParseError error = new ParseError(
+                this.span.line(), this.span.indent(),
+                "invalid unicode or octal escape in text block"
+            );
+            error.initCause(ex);
+            throw error;
+        }
         this.transition(stack, suffix);
         emit.object(
             suffix.attribute(this.span.line(), this.span.indent()),
--- a/eo-parser/src/test/java/org/eolang/parser/LnTextBlockTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/LnTextBlockTest.java
@@ -7,6 +7,7 @@ package org.eolang.parser;
 import com.jcabi.matchers.XhtmlMatchers;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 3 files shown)


morphqdd:
@yegor256, please review

yegor256:

morphqdd:
@yegor256, yaml tests added

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2488](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6047, opened 2026-07-28
#6027: reject signed and truncated \uXXXX escapes in appendUnicode

### LINKED ISSUE #6027: `appendUnicode` accepts a sign in `\uXXXX` escapes and silently passes through truncated ones
## What happens

`Emissions.appendUnicode` (`eo-parser/src/main/java/org/eolang/parser/Emissions.java:539-554`) decodes the 4 hex digits of a `\uXXXX` escape with:

```java
Integer.parseInt(body.substring(cursor, cursor + 4), 16)
```

`Integer.parseInt` accepts an optional leading `+` or `-` sign, so a 4-character slice starting with a sign is still parsed as a valid hex number instead of being rejected.

Separately, when fewer than 4 characters remain after the `u`, the method just re-appends the raw text (`\` plus whatever is left) and returns, with no error.

Verified by compiling `eo-parser` and running the unescape logic on real inputs:

- `"\u+041"` decodes to the single byte `41-` (the letter `A`), with the `+` silently consumed as part of the hex parse.
- `"\u-001"` decodes to `EF-BF-BF` (U+FFFF), with the `-` silently consumed.
- `"\u41"` (only 2 hex digits present, escape truncated) passes through with no error at all, becoming the literal 4 characters `\u41` (bytes `5C-75-34-31`).

None of these three inputs produce a parse error.

## What should happen

A `\uXXXX` escape should require exactly four hex digits matching `[0-9A-Fa-f]{4}` — no sign character, no fewer than four digits. Anything else should raise a `ParseError`, the same way `appendOctal` (immediately above it in the same file) rejects an out-of-range value.

## Steps to reproduce

1. Compile `eo-parser`.
2. Parse a string literal containing `\u+041` or `\u-001` or a truncated `\u41` and observe both decode successfully to unexpected byte content instead of producing a parse error.

morphqdd:
Closes #6027.

`Emissions.appendUnicode` decoded a `\uXXXX` escape with `Integer.parseInt(body.substring(cursor, cursor + 4), 16)`, with no check that the 4 characters are actually hex digits. `Integer.parseInt` accepts an optional leading `+`/`-`, so `\u+041` and `\u-001` decoded successfully instead of being rejected. Separately, when fewer than 4 characters remained after `u`, the code appended the raw text and returned with no error at all, so a truncated `\u41` silently passed through as the literal 4 characters `\u41`.

Fix: check all 4 characters are valid hex digits (`Character.digit(c, 16) >= 0`) before parsing, and that 4 characters actually remain. Otherwise throw `NumberFormatException` with a descriptive message, same pattern `appendOctal` already uses for its own out-of-range case. This propagates through the existing `catch (NumberFormatException) -> ParseError` wrapper in `Emissions.openValue`, so both cases now surface as a normal `<errors>` entry instead of silently decoding wrong or passing through unescaped.

Also bumped the class's `@checkstyle CyclomaticComplexityCheck (600 lines)` window to 610 lines — the fix added a few lines before `topLevelInlinePhi`, pushing that pre-existing method just past the old window and triggering a checkstyle violation unrelated to its own complexity.

Tests: added `EoSyntaxTest.rejectsSignedUnicodeEscape` and `EoSyntaxTest.rejectsTruncatedUnicodeEscape`, both asserting a proper `<errors>` entry instead of a wrong decode or silent pass-through. Verified both fail against the original code and pass with the fix.

`mvn -pl eo-parser test -o -Dtest='EoSyntaxTest'` - 934 tests, all passing.
`mvn -pl eo-parser qulice:check -Pqulice` - clean.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -18,7 +18,7 @@ import java.util.List;
  * §9.4.2).</p>
  *
  * @since 0.1
- * @checkstyle CyclomaticComplexityCheck (600 lines)
+ * @checkstyle CyclomaticComplexityCheck (610 lines)
  * @checkstyle BooleanExpressionComplexityCheck (600 lines)
  */
 @SuppressWarnings({"PMD.UnnecessaryLocalRule", "PMD.TooManyMethods", "PMD.CognitiveComplexity"})
@@ -544,13 +544,21 @@ final class Emissions {
         while (cursor < body.length() && body.charAt(cursor) == 'u') {
             cursor = cursor + 1;
         }
-        if (cursor + 4 > body.length()) {
-            out.append('\\').append(body, start, body.length());
-        } else {
-            out.append(
-                (char) Integer.parseInt(body.substring(cursor, cursor + 4), 16)
+        boolean valid = cursor + 4 <= body.length();
+        for (int idx = cursor; valid && idx < cursor + 4; idx = idx + 1) {
+            valid = Character.digit(body.charAt(idx), 16) >= 0;
+        }
+        if (!valid) {
+            throw new NumberFormatException(
+                String.format(
+                    "unicode escape \\%s is not exactly four hexadecimal digits",
+                    body.substring(start, Math.min(body.length(), cursor + 4))
+                )
             );
         }
+        out.append(
+            (char) Integer.parseInt(body.substring(cursor, cursor + 4), 16)
+        );
         return cursor + 4;
     }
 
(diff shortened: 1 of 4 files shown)


yegor256:
@morphqdd please, add EO test too

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2504](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6048, opened 2026-07-28
#6028: wrap directory.walk glob PatternSyntaxException into ExFailure

### LINKED ISSUE #6028: `directory.walk` doesn't convert malformed glob into `ExFailure`, throws `PatternSyntaxException` instead
## What happens

`EOdirectory$EOwalk.lambda()` (`eo-runtime/src/main/java/org/eolang/EOdirectory$EOwalk.java:40-46`) builds a `PathMatcher` from the user-supplied `glob` attribute:

```java
final PathMatcher matcher = FileSystems.getDefault().getPathMatcher(
    String.format("glob:%s", new Dataized(this.take("glob")).asString())
);
```

This call sits outside the `try` block. The `Files.walk(path)` call right below it, in the same method, is inside a `try` whose `catch (final IOException ex)` wraps the failure into a helpful `ExFailure`. The glob-matcher construction has no equivalent handling.

On JDK 21, `getPathMatcher` throws `java.util.regex.PatternSyntaxException` (an unchecked `IllegalArgumentException` subclass, not an `IOException`) for a malformed glob, confirmed for inputs like `glob:[`, `glob:{`, and `glob:\`.

## What should happen

A malformed `glob` value should be converted into an `ExFailure` (or another EO-level recoverable failure), consistent with how the adjacent `IOException` from `Files.walk` is already handled, so `directory.walk` with a bad glob fails the same predictable way as a directory error rather than crashing with a raw Java exception.

## Steps to reproduce

1. Build `eo-runtime`.
2. Call `directory.walk` with a `glob` value that is not a valid glob pattern, e.g. `"["`.
3. Observe a raw `PatternSyntaxException` instead of an `ExFailure`.

morphqdd:
Closes #6028.

`EOdirectory$EOwalk.lambda()` built its glob `PathMatcher` outside any try/catch. `Files.walk` right below it is already wrapped so `IOException` becomes a clean `ExFailure`, but a malformed glob (`FileSystems.getDefault().getPathMatcher("glob:[")`) throws `PatternSyntaxException`, an unchecked exception, straight out to the caller.

Fix: wrap the matcher construction the same way, converting `PatternSyntaxException` into `ExFailure`.

Tests:
- Added `on-walking-with-malformed-glob` to `directory.eo`, a throwing test matching the existing `on-walking-absent-directory` pattern.
- Added `EOdirectoryEOwalkTest.wrapsMalformedGlobIntoExFailure`, a Java-level test asserting the exception class is specifically `ExFailure`, not just "something throws" (EO throwing tests can only assert `Exception.class` generically, so this is needed to actually distinguish the fix from the bug).

Verified both fail against the original code (the Java test fails outright; the EO test's generated `assertThrows(Exception.class, ...)` still passes either way, since a raw `PatternSyntaxException` is still an `Exception` - this is a known limitation of `-->` throwing tests already present in the existing `on-walking-absent-directory` test, not something specific to this fix).

`mvn -pl eo-runtime test -o -Dtest='EOdirectoryTest,EOdirectoryEOwalkTest'` - 5 tests, all passing.
`mvn -pl eo-runtime qulice:check -Pqulice` - clean except one pre-existing `UnicodeInCode` warning already present on master (unrelated, just shifted a line number).

### DIFF
--- a/eo-runtime/src/main/eo/directory.eo
+++ b/eo-runtime/src/main/eo/directory.eo
@@ -66,3 +66,9 @@
   walk. --> on-walking-absent-directory
     directory mktemp.tmpfile.deleted
     "**"
+
+  --> on-walking-with-malformed-glob
+    d.as-dir.walk "[" > @
+    as-path. > d
+      made.
+        directory mktemp.tmpfile.deleted
--- a/eo-runtime/src/main/java/org/eolang/EOdirectory$EOwalk.java
+++ b/eo-runtime/src/main/java/org/eolang/EOdirectory$EOwalk.java
@@ -11,6 +11,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.PathMatcher;
 import java.nio.file.Paths;
+import java.util.regex.PatternSyntaxException;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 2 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2520](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6049, opened 2026-07-28
#6029: replace EObytes$EOslice.natural with Expect.Natural

### LINKED ISSUE #6029: `EObytes$EOslice.natural` method duplicates `Expect.Natural` but misses overflow guard causing incorrect offset value
## What happens

`EObytes$EOslice.natural(String attr)` (`eo-runtime/src/main/java/org/eolang/EObytes$EOslice.java:68-79`) is a private, hand-rolled copy of the same validation `Expect.Natural` already provides:

```java
private int natural(final String attr) {
    return Expect.at(this, attr)
        .must(number -> number % 1 == 0)
        .that(Double::intValue)
        .otherwise("must be an integer")
        .must(integer -> integer >= 0)
        .otherwise("must be a positive integer")
        .it();
}
```

This chain converts the `double` to `int` with `Double::intValue`, which saturates instead of throwing for values outside `int` range. `Expect.Natural` (`eo-runtime/src/main/java/org/eolang/Expect.java:373` onward) already guards this exact case — that guard was added by the fix for issue #5295, precisely because `Double::intValue` saturates.

Because `EObytes$EOslice` keeps its own duplicate instead of using `Expect.Natural`, it does not have that guard. Calling `bytes.slice` with a `start` or `len` value above `Integer.MAX_VALUE` (e.g. `3000000000`) has the value silently saturate to `2147483647` before the bounds check runs, so the `cant-slice` fallback message reports the offset as `2147483647` instead of the actual out-of-range value supplied by the caller.

Also, `.otherwise("must be a positive integer")` on the `integer >= 0` check is mislabeled: `0` passes the check but `0` is not "positive".

## What should happen

`natural()` should be replaced with `new Expect.Natural(Expect.at(this, attr)).it()`, matching the fix already applied to `EOmalloc$EOof$EOallocated$EOresized` for issue #5301, so both classes share the same validated overflow-safe path instead of `EObytes$EOslice` keeping a divergent, weaker copy. The error message wording should also say "non-negative" rather than "positive" if kept as a separate message at all.

## Steps to reproduce

1. Build `eo-runtime`.
2. Call `bytes.slice 3000000000 1` on any `bytes` value smaller than that offset.
3. Observe the `cant-slice` fallback message reports offset `2147483647`, not `3000000000`.

morphqdd:
Closes #6029.

`EObytes$EOslice.natural` was a private, hand-rolled copy of `Expect.Natural`, missing the int-range guard that `Expect.Natural` already has (added for issue #5295, since `Double::intValue` silently saturates instead of throwing for values outside `int` range). Calling `bytes.slice 3000000000 1` had the offset silently saturate to `2147483647` before the bounds check ran, so the `cant-slice` fallback message reported the wrong offset.

Fix: use `new Expect.Natural(Expect.at(this, attr)).it()` directly and delete the duplicate, matching the precedent already set for `EOmalloc$EOof$EOallocated$EOresized`/`EOchunk$EOresized`.

Note the error message wording changes slightly as a side effect: `Expect.Natural` says "must be greater or equal to zero" where the old duplicate said "must be a positive integer" (0 was always accepted by both, "positive" was just an inaccurate label). Updated `EObytesEOsliceTest.takesWrongSlice` to match.

Tests:
- Added `on-slice-offset-exceeding-int-range` to `bytes.eo`, a throwing test.
- Added `EObytesEOsliceTest.reportsRealOffsetWhenItExceedsIntRange`, a Java-level test asserting the exception message specifically mentions "must fit into int range" - the EO throwing test alone can't distinguish this from the old behavior, since both throw *something* (the old code's saturated offset went out of bounds and failed via the unbound `cant-slice` fallback instead, a different failure with a different message).

Verified both new tests fail against the original code and pass with the fix.

`mvn -pl eo-runtime test -o -Dtest='EObytesEOsliceTest,EObytesTest'` - 48 tests, all passing.
`mvn -pl eo-runtime qulice:check -Pqulice` - clean.

### DIFF
--- a/eo-runtime/src/main/eo/bytes.eo
+++ b/eo-runtime/src/main/eo/bytes.eo
@@ -217,3 +217,8 @@
     20-1F-EE-B5-90
     3
     10
+
+  slice. --> on-slice-offset-exceeding-int-range
+    20-1F-EE-B5-90
+    3000000000
+    1
--- a/eo-runtime/src/main/java/org/eolang/EObytes$EOslice.java
+++ b/eo-runtime/src/main/java/org/eolang/EObytes$EOslice.java
@@ -39,8 +39,8 @@ public final class EObytes$EOslice extends PhDefault implements Atom {
     @Override
     public Phi lambda() {
         final byte[] bytes = new Dataized(this.take(Phi.RHO)).take();
-        final int start = this.natural("start");
-        final int len = this.natural("len");
+        final int start = new Expect.Natural(Expect.at(this, "start")).it();
+        final int len = new Expect.Natural(Expect.at(this, "len")).it();
         final Phi result;
         if ((long) start + len <= bytes.length) {
             result = new Data.ToPhi(Arrays.copyOfRange(bytes, start, start + len));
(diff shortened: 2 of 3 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2560](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6050, opened 2026-07-28
#6031: fold trackLocations/coverageTracking into transpile cache key

### LINKED ISSUE #6031: Transpile cache key neglects `trackLocations` and `coverageTracking`, causing corrosive cache poisoning
## What happens

The transpile disk cache key does not include the `trackLocations` or `coverageTracking` Maven properties, even though both flags change the generated Java output.

The cache key is built in `Transpiling.transpiled` (`eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java:249`):

```java
new CachePath(cdir, this.version(), hsh.get())
```

`version()` (`Transpiling.java:374-375`) is:

```java
return String.format("%s-%s", this.version, new Fingerprint(Transpiling.XSLS).get());
```

This folds in the plugin version and an XSL-content fingerprint (added for #5578), but not `this.tracking.locations()` (from `eo.trackLocations`, `MjTranspile.java:63-64`) or `this.coverage` (from `eo.coverageTracking`, `MjTranspile.java:95-96`).

Both flags materially change the transpiled Java: `to-java.xsl:630` emits `new PhSafe(...)` gated on the `trackLocations` XSL param, and `to-java.xsl:649` emits `new PhCoverage(...)` gated on the `coverage` XSL param. The code already knows these flags matter for output identity — the separate in-memory per-thread train cache is keyed on them (`Transpiling.java:332-339`, key format `"%b|%b"`) — but the on-disk `CachePath` used for the actual transpiled-file cache is not.

I verified this with three throwaway FakeMaven projects sharing one cache directory:
- Project 1, `trackLocations=false`, built first.
- Project 2, `trackLocations=true`, built second against the same cache dir → produced 0 Java files containing `new PhSafe(`.
- Project 3 (control), `trackLocations=true`, its own fresh cache dir → produced 1 file containing `new PhSafe(`.

Project 2 silently inherited project 1's un-instrumented cached output instead of getting its own `trackLocations=true` output.

## What should happen

`version()` (or the `CachePath` construction at `Transpiling.java:249`) should fold in both boolean flags, the same way the XSL fingerprint is already folded in, so toggling either flag invalidates the cache entry.

## Steps to reproduce

1. Build a project with `-Deo.trackLocations=false` using a shared `eo.cache` directory.
2. Build the same (or another) project with `-Deo.trackLocations=true` against the same cache directory.
3. Inspect the generated Java for the second build: it does not contain `new PhSafe(`, even though `trackLocations=true` was requested.

morphqdd:
Closes #6031.

`Transpiling.version()` builds the transpile disk cache key from the plugin version plus an XSL-content fingerprint (added for #5578), but not from `trackLocations`/`coverageTracking`. Both flags materially change the generated Java (`to-java.xsl` conditionally emits `new PhSafe(...)` and `new PhCoverage(...)` based on them), and the in-memory per-thread train cache is already keyed on them (`"%b|%b"` in `train()`), but the on-disk `CachePath` used for the actual transpiled-file cache was not. So toggling either flag while reusing an existing `eo.cache` directory served stale, differently-instrumented output.

Fix: fold both booleans into `version()`.

Test: `MjTranspileTest.invalidatesCacheWhenTrackLocationsChanges`, two separate `FakeMaven` projects sharing one cache directory, first with `trackLocations=false`, second with `trackLocations=true`. Asserts the second project's generated Java contains `new PhSafe(`. Verified this fails against the original code (second project's Java has no `PhSafe`, inherited from the first project's cache) and passes with the fix.

`mvn -pl eo-maven-plugin test -o -Dtest='MjTranspileTest'` - 41 tests, all passing.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
@@ -364,15 +364,22 @@ final class Transpiling implements Step {
 
     /**
      * Cache-key version segment: the plugin version combined with a
-     * fingerprint of the bundled transpile XSLs. Folding the XSL
-     * content in means that a change in the transformation logic
+     * fingerprint of the bundled transpile XSLs, plus the
+     * {@code trackLocations}/{@code coverageTracking} flags. Folding the
+     * XSL content in means that a change in the transformation logic
      * invalidates the global transpile cache even when the plugin
      * version is unchanged (a constant {@code -SNAPSHOT} during
-     * development), see #5578.
+     * development), see #5578. Folding the two flags in means toggling
+     * either one also invalidates the cache, since both change what
+     * {@code to-java.xsl} emits (see #6031).
      * @return The version segment for {@link CachePath}
      */
     private String version() {
-        return String.format("%s-%s", this.version, new Fingerprint(Transpiling.XSLS).get());
+        return String.format(
+            "%s-%s-%b-%b",
+            this.version, new Fingerprint(Transpiling.XSLS).get(),
+            this.tracking.locations(), this.coverage
+        );
     }
 
     /**
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjTranspileTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjTranspileTest.java
@@ -255,6 +255,29 @@ final class MjTranspileTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (38) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2544](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6051, opened 2026-07-28
#6032: fold xsl:import-ed libraries into the transpile cache-key fingerprint

### LINKED ISSUE #6032: `Fingerprint` incorrectly ignores imported stylesheets, leading to incomplete transpile cache key
## What happens

`Fingerprint` (`eo-maven-plugin/src/main/java/org/eolang/maven/Fingerprint.java`), introduced for #5578 to fold the content of the transpile XSLs into the transpile cache key so an XSL change invalidates stale cached output, only reads the exact classpath resource paths it is given (`Fingerprint.java:41-56`). It does not resolve `xsl:import` or `xsl:include`.

`Transpiling.XSLS` (`eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java:95-104`), the list passed to `new Fingerprint(Transpiling.XSLS)` at `Transpiling.java:375`, lists nine top-level transpile stylesheets. Five of them pull in real transformation logic through `xsl:import`, and none of the imported files are themselves in `XSLS`:

- `transpile/attrs.xsl:7`, `transpile/classes.xsl:7`, `transpile/data.xsl:7`, `transpile/to-java.xsl:7` → `xsl:import href="/org/eolang/parser/_funcs.xsl"` (a shared library of `eo:` functions such as `eo:bytes-to-string`, `eo:has-data`, `eo:atom`)
- `transpile/set-original-names.xsl:13` → `xsl:import href="/org/eolang/parser/_specials.xsl"`

Editing `_funcs.xsl` (which lives in `eo-parser`, not `eo-maven-plugin`) changes the actual output of `to-java.xsl` and the others, but does not change any byte in `Transpiling.XSLS`'s listed resources, so `Fingerprint.get()` returns the same value before and after the edit. The transpile cache key is unchanged, so stale transpiled output from before the `_funcs.xsl` edit is served from `~/.eo/transpiled/...` — exactly the failure mode #5578 was written to prevent, one indirection deeper.

## What should happen

`Fingerprint` should transitively include imported/included stylesheets — either by resolving `xsl:import`/`xsl:include` hrefs itself, or at minimum by having `Transpiling.XSLS` explicitly list `_funcs.xsl` and `_specials.xsl` alongside the top-level files, with a test that fails if a new `xsl:import` is added to any listed stylesheet without a corresponding fingerprint update.

## Steps to reproduce

1. Build a project once so a transpile result is cached under `~/.eo/transpiled/...`.
2. Edit `eo-parser/src/main/resources/org/eolang/parser/_funcs.xsl` (e.g. change how `eo:bytes-to-string` renders a literal).
3. Rebuild the plugin and the same project against the same cache directory, still on the `-SNAPSHOT` plugin version.
4. Observe the transpiled Java is unchanged — the stale, pre-edit output is served from cache.

morphqdd:
Closes #6032.

`Fingerprint` (introduced for #5578) folds the content of the bundled transpile XSLs into the transpile cache key, so editing one of them invalidates the cache. But `Transpiling.XSLS` only lists the nine top-level stylesheets. Five of them pull in real transformation logic through `xsl:import`: `to-java.xsl`, `attrs.xsl`, `classes.xsl`, and `data.xsl` import `_funcs.xsl` (a shared library of `eo:` functions), and `set-original-names.xsl` imports `_specials.xsl`. Neither imported file was part of the fingerprinted set, so editing `_funcs.xsl` changed the actual transpiled output without changing the cache key, serving stale output from `~/.eo/transpiled/...`.

Fix: added a separate `Transpiling.IMPORTS` array listing the two imported files, and fold both `XSLS` and `IMPORTS` into the fingerprint used by `version()`. Kept as a separate array from `XSLS` rather than appending to it, because `XSLS` is also used verbatim to build the actual XSL train in `compiled(boolean, boolean)`, where its last element is special-cased as `to-java.xsl` - appending to it would have broken the transpile pipeline itself, not just the cache key.

Also widened `XSLS`, `IMPORTS`, and `version()` from `private` to package-private, so the new test can exercise them directly instead of using reflection (this codebase has no reflection-based test precedent, and qulice's `AvoidAccessibilityAlteration` flags `setAccessible()` calls).

Test: `TranspilingTest.versionFoldsInImportedXslLibraries`, asserting a real `Transpiling` instance's `version()` differs from a fingerprint computed over `XSLS` alone. Verified it fails against the original code and passes with the fix.

`mvn -pl eo-maven-plugin test -o -Dtest='TranspilingTest,FingerprintTest,MjTranspileTest'` - 46 tests, all passing.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
@@ -29,6 +29,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.cactoos.func.StickyFunc;
 import org.eolang.parser.OnDefault;
 import org.eolang.parser.OnDetailed;
@@ -65,6 +66,40 @@ final class Transpiling implements Step {
     static final String PRE = "5-pre-transpile";
 
     /**
+     * The XSL steps of the transpile train, in order, ending with
+     * {@code to-java.xsl}. Kept as a single list so both the train in
+     * {@link #compiled(boolean)} and the cache-key fingerprint in
+     * {@link #version()} are derived from the same source.
+     */
+    static final String[] XSLS = {
+        "/org/eolang/parser/parse/set-locators.xsl",
+        "/org/eolang/maven/transpile/set-original-names.xsl",
+        "/org/eolang/maven/transpile/classes.xsl",
+        "/org/eolang/maven/transpile/tests.xsl",
+        "/org/eolang/maven/transpile/anonymous-to-nested.xsl",
+        "/org/eolang/maven/transpile/package.xsl",
+        "/org/eolang/maven/transpile/attrs.xsl",
+        "/org/eolang/maven/transpile/data.xsl",
+        "/org/eolang/maven/transpile/to-java.xsl",
+    };
+
+    /**
+     * Classpath resources {@code xsl:import}-ed by one or more of
+     * {@link #XSLS}, so their content must also be folded into the
+     * cache-key fingerprint in {@link #version()} — editing one of these
+     * shared libraries changes the actual transpile output just as much
+     * as editing a top-level stylesheet does, but leaves {@link #XSLS}
+     * itself unchanged (see #6032). Not part of {@link #XSLS} itself
+     * because that array is also used verbatim to build the actual XSL
+     * train in {@link #compiled(boolean, boolean)}, where its last
+     * element is special-cased as {@code to-java.xsl}.
+     */
+    static final String[] IMPORTS = {
+        "/org/eolang/parser/_funcs.xsl",
+        "/org/eolang/parser/_specials.xsl",
+    };
+
+    /**
      * Parsing trains with XSLs, one per thread, keyed by whether source
      * locations are tracked.
      * <p>
@@ -85,24 +120,6 @@ final class Transpiling implements Step {
      * Cache directory for transpiled sources.
      */
     private static final String CACHE = "transpiled";
-
-    /**
-     * The XSL steps of the transpile train, in order, ending with
-     * {@code to-java.xsl}. Kept as a single list so both the train in
-     * {@link #compiled(boolean)} and the cache-key fingerprint in
-     * {@link #version()} are derived from the same source.
-     */
-    private static final String[] XSLS = {
-        "/org/eolang/parser/parse/set-locators.xsl",
-        "/org/eolang/maven/transpile/set-original-names.xsl",
-        "/org/eolang/maven/transpile/classes.xsl",
-        "/org/eolang/maven/transpile/tests.xsl",
-        "/org/eolang/maven/transpile/anonymous-to-nested.xsl",
-        "/org/eolang/maven/transpile/package.xsl",
-        "/org/eolang/maven/transpile/attrs.xsl",
-        "/org/eolang/maven/transpile/data.xsl",
-        "/org/eolang/maven/transpile/to-java.xsl",
-    };
 
     /**
      * Java extension.
@@ -220,6 +237,30 @@ final class Transpiling implements Step {
         Logger.info(
             this, "Transpiled %d XMIRs, created %d Java files in %[file]s",
             this.sources.size(), saved, this.generatedDir
+        );
+    }
+
+    /**
+     * Cache-key version segment: the plugin version combined with a
+     * fingerprint of the bundled transpile XSLs and the libraries they
+     * {@code xsl:import}. Folding the XSL content in means that a change
+     * in the transformation logic invalidates the global transpile cache
+     * even when the plugin version is unchanged (a constant
+     * {@code -SNAPSHOT} during development), see #5578; folding the
+     * imported libraries in too closes the gap where editing one of
+     * them changed the actual output without changing anything in
+     * {@link #XSLS} itself, see #6032.
+     * @return The version segment for {@link CachePath}
+     */
+    String version() {
+        return String.format(
+            "%s-%s",
+            this.version,
+            new Fingerprint(
+                Stream.concat(
+                    Arrays.stream(Transpiling.XSLS), Arrays.stream(Transpiling.IMPORTS)
+                ).toArray(String[]::new)
+            ).get()
         );
     }
 
@@ -360,19 +401,6 @@ final class Transpiling implements Step {
                 )
             )
         );
-    }
-
-    /**
-     * Cache-key version segment: the plugin version combined with a
-     * fingerprint of the bundled transpile XSLs. Folding the XSL
-     * content in means that a change in the transformation logic
-     * invalidates the global transpile cache even when the plugin
-     * version is unchanged (a constant {@code -SNAPSHOT} during
-     * development), see #5578.
-     * @return The version segment for {@link CachePath}
-     */
-    private String version() {
-        return String.format("%s-%s", this.version, new Fingerprint(Transpiling.XSLS).get());
     }
 
     /**
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/TranspilingTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/TranspilingTest.java
@@ -1,0 +1,52 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import java.nio.file.Paths;
+import java.util.Collections;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link Transpiling}.
+ * @since 0.1
+ */
+final class TranspilingTest {
+
+    @Test
+    void versionFoldsInImportedXslLibraries() {
+        MatcherAssert.assertThat(
+            "the cache-key version must differ from a fingerprint of the top-level XSLS alone, proving the xsl:import-ed libraries are actually folded in",
+            TranspilingTest.transpiling().version(),
+            Matchers.not(
+                Matchers.equalTo(
+                    String.format(
+                        "1.0-SNAPSHOT-%s", new Fingerprint(Transpiling.XSLS).get()
+                    )
+                )
+            )
+        );
+    }
+
+    /**
+     * A minimal {@link Transpiling} instance for testing its private helpers.
+     * @return A new instance
+     */
+    private static Transpiling transpiling() {
+        return new Transpiling(
+            Collections.emptyList(),
+            Paths.get("target"),
+            Paths.get("target/generated"),
+            Paths.get("target/cache"),
+            true,
+            "1.0-SNAPSHOT",
+            false,
+            Paths.get("target/xsl-measures.csv"),
+            new Tracking(false, false),
+            false
+        );
+    }
+}
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/TranspilingTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/TranspilingTest.java
@@ -17,7 +17,7 @@ import org.junit.jupiter.api.Test;
 final class TranspilingTest {
 
     @Test
-    void versionFoldsInImportedXslLibraries() {
+    void foldsInImportedXslLibrariesIntoVersion() {
         MatcherAssert.assertThat(
             "the cache-key version must differ from a fingerprint of the top-level XSLS alone, proving the xsl:import-ed libraries are actually folded in",
             TranspilingTest.transpiling().version(),


morphqdd:

eo-maven-plugin/src/test/java/org/eolang/maven/TranspilingTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Test
+    void foldsInImportedXslLibrariesIntoVersion() {
+        MatcherAssert.assertThat(
+            "the cache-key version must differ from a fingerprint of the top-level XSLS alone, proving the xsl:import-ed libraries are actually folded in",
+            TranspilingTest.transpiling().version(),
+            Matchers.not(
+                Matchers.equalTo(
+                    String.format(
+                        "1.0-SNAPSHOT-%s", new Fingerprint(Transpiling.XSLS).get()
```
Favixx, line 27: This is the heart of the problem: the expected string has **two** segments, `1.0-SNAPSHOT-<fingerprint>`, while `version()` produces **four**, `1.0-SNAPSHOT-<fingerprint>-false-false` (it appends `tracking.locations()` and `coverage`). The two strings differ structurally, so `Matchers.not(equalTo(...))` holds no matter what `IMPORTS` contains — including when it is empty, which is the pre-fix state.

What you want to assert is that the fingerprint over `XSLS ∪ IMPORTS` differs from the fingerprint over `XSLS` alone:

```java
MatcherAssert.assertThat(
    "the imported libraries must change the fingerprint",
    new Fingerprint(TranspilingTest.all()).get(),
    Matchers.not(Matchers.equalTo(new Fingerprint(Transpiling.XSLS).get()))
);
```

That one does fail if `IMPORTS` goes back to empty, and it needs neither a `Transpiling` instance nor the widened `version()`.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -64,6 +65,40 @@ final class Transpiling implements Step {
...
+     * {@link #XSLS}, so their content must also be folded into the
+     * cache-key fingerprint in {@link #version()} — editing one of these
+     * shared libraries changes the actual transpile output just as much
+     * as editing a top-level stylesheet does, but leaves {@link #XSLS}
+     * itself unchanged (see #6032). Not part of {@link #XSLS} itself
+     * because that array is also used verbatim to build the actual XSL
+     * train in {@link #compiled(boolean, boolean)}, where its last
+     * element is special-cased as {@code to-java.xsl}.
+     */
+    static final String[] IMPORTS = {
```
Favixx, line 97: The bug being fixed is "a stylesheet that affects the output is missing from a hand-maintained list". The fix is a second hand-maintained list, so the failure mode survives intact: the next `xsl:import` added to any of the nine stylesheets drops out of the cache key again, silently, with the same symptom — stale Java served from `~/.eo/transpiled` and nothing red anywhere.

The import graph is machine-readable and already on the classpath. Reading each resource in `XSLS`, pulling `//xsl:import/@href | //xsl:include/@href`, and recursing until the set stops growing is maybe fifteen lines, and makes the list self-maintaining. It also removes the need for the `IMPORTS` constant to be package-private for testing, since the closure itself becomes the thing worth asserting on.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -64,6 +65,40 @@ final class Transpiling implements Step {
...
+        "/org/eolang/parser/_funcs.xsl",
+        "/org/eolang/parser/_specials.xsl",
```
Favixx, line 99: There is a third stylesheet that reaches the cached output and is in neither array: `org/eolang/parser/_stepped.xsl`.

`compiled()` returns `new TrFull(...)`, and `TrFull`'s constructor wraps its train in `TrStepped`, which does `stepped.value().with("step", pos).with("sheet", shift.uid()).transform(xml)` after every shift — appending `<sheet>` elements under `/object/sheets`. Those elements are in the XMIR that `transpiled()` writes through `new Saved(transform.apply(xmir).toString(), target)` and that the cache stores. So editing `_stepped.xsl` changes cached content while `version()` stays put, which is exactly #6032.

I may be wrong about how much of that survives into the cached artifact — you know this pipeline better than I do — but if it does, it belongs in this fix, and it is the second data point for walking the graph instead of listing files.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -64,6 +65,40 @@ final class Transpiling implements Step {
      */
     static final String PRE = "5-pre-transpile";
 
+    /**
+     * The XSL steps of the transpile train, in order, ending with
+     * {@code to-java.xsl}. Kept as a single list so both the train in
+     * {@link #compiled(boolean)} and the cache-key fingerprint in
+     * {@link #version()} are derived from the same source.
```
Favixx, line 72: This sentence — "Kept as a single list so both the train in `compiled(boolean)` and the cache-key fingerprint in `version()` are derived from the same source" — was true before this PR and is not any more: the train is built from `XSLS`, the fingerprint from `XSLS` plus `IMPORTS`. The javadoc was moved verbatim when the field moved, and now states the invariant the PR deliberately breaks. (It also still says `compiled(boolean)`, while the method has taken two booleans for a while.)

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -223,6 +240,34 @@ public void exec() throws IOException {
...
+     * {@link #XSLS} itself, see #6032. Folding the two flags in means
+     * toggling either one also invalidates the cache, since both change
+     * what {@code to-java.xsl} emits (see #6031).
+     * @return The version segment for {@link CachePath}
+     */
+    String version() {
+        return String.format(
+            "%s-%s-%b-%b",
+            this.version,
+            new Fingerprint(
```
Favixx, line 262: `Fingerprint.get()` opens, reads and SHA-256s every resource on each call, and `version()` is called from `transpiled()` — once per tojo, inside the `Threaded` pool. So a build with a few thousand `.eo` sources re-reads and re-hashes the whole stylesheet set a few thousand times, concurrently. This PR makes it 22% worse by going from nine resources to eleven.

The fingerprint depends on nothing instance-specific, so it can be computed once: a `private static final Supplier<String>` wrapped in the `StickyFunc` this file already imports, or just a static field. `version()` then stays a cheap `String.format`.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -64,6 +65,40 @@ final class Transpiling implements Step {
      */
     static final String PRE = "5-pre-transpile";
 
+    /**
+     * The XSL steps of the transpile train, in order, ending with
+     * {@code to-java.xsl}. Kept as a single list so both the train in
+     * {@link #compiled(boolean)} and the cache-key fingerprint in
+     * {@link #version()} are derived from the same source.
+     */
+    static final String[] XSLS = {
```
Favixx, line 74: Three members — `XSLS`, `IMPORTS` and `version()` — lose `private` to support a single assertion, and that assertion turns out to hold vacuously. The reasoning about avoiding `setAccessible()` is right (qulice would flag it), but the conclusion should be that the seam is in the wrong place, not that the fields should be opened up.

With the test rewritten as I suggested on `TranspilingTest`, only `XSLS` and `IMPORTS` need to be visible, and `version()` can stay private. Better still, a small `resources()` method returning the concatenation would give the test one thing to assert against and keep both arrays private.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -64,6 +65,40 @@ final class Transpiling implements Step {
...
+     * cache-key fingerprint in {@link #version()} — editing one of these
+     * shared libraries changes the actual transpile output just as much
+     * as editing a top-level stylesheet does, but leaves {@link #XSLS}
+     * itself unchanged (see #6032). Not part of {@link #XSLS} itself
+     * because that array is also used verbatim to build the actual XSL
+     * train in {@link #compiled(boolean, boolean)}, where its last
+     * element is special-cased as {@code to-java.xsl}.
+     */
+    static final String[] IMPORTS = {
+        "/org/eolang/parser/_funcs.xsl",
```
Favixx, line 98: "Classpath resources `xsl:import`-ed by one or more of `XSLS`" undersells it: `_specials.xsl` is imported by `set-original-names.xsl` *and* by `_funcs.xsl`, so the graph this array flattens is already two levels deep. That is worth stating here, because the next person adding an entry will check which top-level stylesheet imports their file and stop there.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -223,6 +240,34 @@ public void exec() throws IOException {
...
+     * toggling either one also invalidates the cache, since both change
+     * what {@code to-java.xsl} emits (see #6031).
+     * @return The version segment for {@link CachePath}
+     */
+    String version() {
+        return String.format(
+            "%s-%s-%b-%b",
+            this.version,
+            new Fingerprint(
+                Stream.concat(
```
Favixx, line 263: Worth noting that these two entries are the first resources in this fingerprint that come from a different artifact — `eo-parser`, not `eo-maven-plugin`. `Fingerprint` throws `IllegalStateException("Resource '%s' not found on classpath")` when a resource is missing, so a repackaging or shading change in `eo-parser` now turns what used to be a stale-cache bug into a hard failure of every transpile. That is the right trade, but a line of javadoc saying these live in another module would help whoever hits it.

eo-maven-plugin/src/test/java/org/eolang/maven/TranspilingTest.java
```diff
@@ -0,0 +1,52 @@
...
+     * @return A new instance
+     */
+    private static Transpiling transpiling() {
+        return new Transpiling(
+            Collections.emptyList(),
+            Paths.get("target"),
+            Paths.get("target/generated"),
+            Paths.get("target/cache"),
+            true,
+            "1.0-SNAPSHOT",
```
Favixx, line 45: `"1.0-SNAPSHOT"` appears here as the constructor argument and again on line 27 inside the expected format string, with nothing tying them together. If someone changes one, the test does not fail — it just becomes even more vacuous than it already is. A single constant used in both places, or building the expectation from the same variable, removes that.

eo-maven-plugin/src/test/java/org/eolang/maven/TranspilingTest.java
```diff
@@ -0,0 +1,52 @@
...
+                    String.format(
+                        "1.0-SNAPSHOT-%s", new Fingerprint(Transpiling.XSLS).get()
+                    )
+                )
+            )
+        );
+    }
+
+    /**
+     * A minimal {@link Transpiling} instance for testing its private helpers.
```
Favixx, line 35: "A minimal `Transpiling` instance for testing its private helpers" — this PR is what made those helpers package-private, so the comment is stale the moment it lands. If the test ends up not needing an instance at all (see my comment on line 27), this method goes away with it.

eo-maven-plugin/src/test/java/org/eolang/maven/TranspilingTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link Transpiling}.
+ * @since 0.1
+ */
+final class TranspilingTest {
+
+    @Test
+    void foldsInImportedXslLibrariesIntoVersion() {
```
Favixx, line 20: The method is `foldsInImportedXslLibrariesIntoVersion`, while the description calls it `TranspilingTest.versionFoldsInImportedXslLibraries`. Trivial in itself, but combined with the "verified it fails against the original code" claim it suggests the description was written against an earlier version of the test than the one committed — worth a second look at the whole verification section before merge.

eo-maven-plugin/src/test/java/org/eolang/maven/TranspilingTest.java
```diff
@@ -0,0 +1,52 @@
...
+        MatcherAssert.assertThat(
+            "the cache-key version must differ from a fingerprint of the top-level XSLS alone, proving the xsl:import-ed libraries are actually folded in",
```
Favixx, line 22: This assertion message runs past 150 characters on one line; the surrounding tests in this module keep them wrapped and short. Once the assertion actually asserts something, the message can shrink to "the imported XSL libraries must be folded into the fingerprint" and carry its weight.

Favixx (changes requested):
The diagnosis in #6032 is right and the direction is right — `_funcs.xsl` really does change transpiled output while leaving `XSLS` untouched, and folding it into the fingerprint is the fix. Two things stop me from approving.

**The test does not test the fix.** `version()` returns `"%s-%s-%b-%b"` — four segments — while the expected value the test compares against is built as `"1.0-SNAPSHOT-%s"`, two segments. Those can never be equal, whatever the fingerprint is. Run it against master with `IMPORTS` removed entirely and it still passes: before the change `version()` yields `1.0-SNAPSHOT-<fp>-false-false`, and the test asserts only that this differs from `1.0-SNAPSHOT-<fp>`. So the "verified it fails against the original code" note in the description cannot be right, and the regression this PR fixes is still unguarded.

**The fix is already incomplete, in the same way as the bug.** `compiled()` wraps the train in `TrFull`, whose constructor wraps it in `TrStepped`, which applies `org/eolang/parser/_stepped.xsl` after *every* shift and appends `<sheet>` elements to `/object/sheets`. That output is part of the XMIR handed to `new Saved(transform.apply(xmir).toString(), target)` and therefore part of what gets cached — so editing `_stepped.xsl` changes cached content without changing the key, which is #6032 again. It is not in `XSLS` and not in `IMPORTS`. Please confirm and either add it or say why it doesn't count.

That second point is really an argument about the approach: this PR fixes the instance by adding a second hand-maintained list, and the bug was that a hand-maintained list went stale. Details and the smaller stuff inline.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (153 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2572](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+46](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6052, opened 2026-07-28
#6038: check number payload is exactly 8 bytes before StUnhex.getDouble

### LINKED ISSUE #6038: `StUnhex` incorrectly parses payloads not exactly 8 bytes, causing crashes or data loss
## What happens

`StUnhex`'s number-decoding branch (`eo-printer/src/main/java/org/eolang/printer/StUnhex.java`, the `elements("number")` shift) reads a `Φ.number` payload with:

```java
final double number = StUnhex.buffer(
    StUnhex.undash(xnav.element("o").text().orElse(""))
).getDouble();
```

`ByteBuffer.getDouble()` requires exactly 8 bytes. There is no length check before the call.

- `number 40-49-0F > @` (3 bytes) parses with zero errors, then `Xmir.toEO()` throws `IllegalArgumentException: Shift 'org.eolang.printer.StUnhex@..' failed`, caused by `java.nio.BufferUnderflowException`.
- `number 40-49-0F-DB-00-00-00-00-11-22 > @` (10 bytes, 2 bytes too many) prints as `50.123870849609375` — the trailing `11-22` is silently dropped, so the printed program is not equivalent to the parsed one.

## What should happen

The number branch should read exactly 8 bytes and, when the payload is not exactly 8 bytes, leave the `Φ.number`/`Φ.bytes` structure as-is instead of attempting `getDouble()`. This is the same strategy already used by the sibling string branch's `decode()`, which was hardened for exactly this class of problem by the fix for #5620 (string literals with invalid UTF-8 bytes).

## Steps to reproduce

1. Build `eo-printer`.
2. Parse and then print `[] > x\n  number 40-49-0F > @` — observe `Xmir.toEO()` throws `BufferUnderflowException` instead of a graceful fallback.
3. Parse and print `[] > x\n  number 40-49-0F-DB-00-00-00-00-11-22 > @` — observe the output number is computed from only the first 8 of the 10 bytes, silently dropping the rest.

morphqdd:
Closes #6038.

`StUnhex`'s number branch called `ByteBuffer.getDouble()` with no check that the payload is exactly 8 bytes. A shorter payload throws `BufferUnderflowException`; a longer one gets its trailing bytes silently dropped (`getDouble()` only reads the first 8), so `40-49-0F-DB-00-00-00-00-11-22` (10 bytes) printed as `50.123870849609375`, computed from only the first 8 bytes, not equivalent to the original program.

Fix: check `buffer.remaining() == Double.BYTES` before calling `getDouble()`. If it isn't, leave the `Φ.number`/`Φ.bytes` structure as-is, matching the strategy the sibling string branch already uses (`decode()`, hardened for #5620) when it can't safely convert.

Tests: added `keepsTooFewNumberBytesUnconverted` and `keepsTooManyNumberBytesUnconverted` to `StUnhexTest`, matching the existing `keepsMalformedBytesUnconverted` pattern for the string branch. Verified both fail against the original code (one with `BufferUnderflowException`, the other with the silently truncated `50.123870849609375`) and pass with the fix.

`mvn -pl eo-printer test -o -Dtest='StUnhexTest'` - 12 tests, all passing.
`mvn -pl eo-printer qulice:check -Pqulice` - clean.

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
+++ b/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
@@ -47,11 +47,14 @@ final class StUnhex extends StEnvelope {
         new StXnav(
             StUnhex.elements("number"),
             xnav -> {
-                final double number = StUnhex.buffer(
+                final ByteBuffer buffer = StUnhex.buffer(
                     StUnhex.undash(xnav.element("o").text().orElse(""))
-                ).getDouble();
-                if (!Double.isNaN(number) && !Double.isInfinite(number)) {
-                    xnav.node().setTextContent(StUnhex.number(number));
+                );
+                if (buffer.remaining() == Double.BYTES) {
+                    final double number = buffer.getDouble();
+                    if (!Double.isNaN(number) && !Double.isInfinite(number)) {
+                        xnav.node().setTextContent(StUnhex.number(number));
+                    }
                 }
             }
         ),
--- a/eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
+++ b/eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
@@ -181,6 +181,44 @@ final class StUnhexTest {
 
     @ParameterizedTest
     @MethodSource("shifts")
+    void keepsTooFewNumberBytesUnconverted(final Shift shift, final String type) {
+        MatcherAssert.assertThat(
+            String.format(
+                "StUnhex by %s must not crash on a number payload shorter than 8 bytes, but it did",
+                type
+            ),
+            new Xsline(new StUnhex(shift)).pass(
+                new XMLDocument(
+                    "<p><o base='Φ.number'><o base='Φ.bytes'><o>40-49-0F</o></o></o></p>"
+                )
+            ),
+            XhtmlMatchers.hasXPath(
+                "//o[@base='Φ.number' and o[@base='Φ.bytes' and not(o) and text()='40-49-0F']]"
+            )
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("shifts")
+    void keepsTooManyNumberBytesUnconverted(final Shift shift, final String type) {
+        MatcherAssert.assertThat(
+            String.format(
+                "StUnhex by %s must not silently drop trailing bytes from a number payload longer than 8 bytes, but it did",
+                type
+            ),
+            new Xsline(new StUnhex(shift)).pass(
+                new XMLDocument(
+                    "<p><o base='Φ.number'><o base='Φ.bytes'><o>40-49-0F-DB-00-00-00-00-11-22</o></o></o></p>"
+                )
+            ),
+            XhtmlMatchers.hasXPath(
+                "//o[@base='Φ.number' and o[@base='Φ.bytes' and not(o) and text()='40-49-0F-DB-00-00-00-00-11-22']]"
+            )
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("shifts")
     void keepsMalformedBytesUnconverted(final Shift shift, final String type) {
         MatcherAssert.assertThat(
             String.format(


gemshrine (approved):
The exact-size guard prevents both underflow and silent truncation, and the paired malformed-payload tests cover both boundaries.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -47,11 +47,14 @@ final class StUnhex extends StEnvelope {
         new StXnav(
             StUnhex.elements("number"),
             xnav -> {
-                final double number = StUnhex.buffer(
+                final ByteBuffer buffer = StUnhex.buffer(
```
gemshrine, line 50: Holding the decoded bytes in one local buffer makes the validation and the subsequent read operate on the same payload. It avoids repeating `undash`/`buffer` and preserves the existing behaviour for a valid number representation.

eo-printer/src/main/java/org/eolang/printer/StUnhex.java
```diff
@@ -47,11 +47,14 @@ final class StUnhex extends StEnvelope {
...
                     StUnhex.undash(xnav.element("o").text().orElse(""))
-                ).getDouble();
-                if (!Double.isNaN(number) && !Double.isInfinite(number)) {
-                    xnav.node().setTextContent(StUnhex.number(number));
+                );
+                if (buffer.remaining() == Double.BYTES) {
```
gemshrine, line 53: The equality check, rather than only a lower bound, is necessary. Fewer than eight bytes previously failed at `getDouble()`, while more than eight bytes let `getDouble()` consume a valid prefix and silently discard the rest. `Double.BYTES` also states the binary contract directly instead of hard-coding `8`.

eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
```diff
@@ -179,6 +179,44 @@ void keepsGlyphsButEscapesControls(final Shift shift, final String type) {
         );
     }
 
+    @ParameterizedTest
+    @MethodSource("shifts")
+    void keepsTooFewNumberBytesUnconverted(final Shift shift, final String type) {
```
gemshrine, line 184: This case verifies more than absence of an exception: the XPath requires the original `Φ.number → Φ.bytes` subtree and its exact payload to survive. That is the correct fallback contract for a short binary representation.

eo-printer/src/test/java/org/eolang/printer/StUnhexTest.java
```diff
@@ -179,6 +179,44 @@ void keepsGlyphsButEscapesControls(final Shift shift, final String type) {
...
+            ),
+            XhtmlMatchers.hasXPath(
+                "//o[@base='Φ.number' and o[@base='Φ.bytes' and not(o) and text()='40-49-0F']]"
+            )
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("shifts")
+    void keepsTooManyNumberBytesUnconverted(final Shift shift, final String type) {
```
gemshrine, line 203: This is a distinct boundary from the short-input case. The ten-byte payload starts with a decodable double, so asserting that the entire original text remains is what proves the printer no longer performs a lossy prefix conversion.

gemshrine (commented):
Follow-up review: I traced the number branch through both StUnhex shift variants and checked the before/after tree assertions. The exact-length guard preserves valid IEEE-754 conversion while safely leaving malformed number nodes untouched. The paired tests cover the two distinct failures of the prior `getDouble()` call; CI is green.

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+951](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2588](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6053, opened 2026-07-28
#6039: fix Linting.summary dropping the error count

### LINKED ISSUE #6039: `Linting.summary` omits the error count in the output when all severity levels are present
## What happens

`Linting.summary(Map<Severity, Integer> counts)` (`eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java:500-527`) drops the middle count whenever all three severities (critical, error, warning) are present:

```java
final String sum;
if (parts.size() < 3) {
    sum = String.join(" and ", parts);
} else {
    sum = String.format(
        "%s, and %s",
        String.join(", ", parts.subList(0, parts.size() - 2)),
        parts.get(parts.size() - 1)
    );
}
```

`parts` is built in order `[critical, error, warning]`, skipping any that are zero. When all three are present, `parts.size() == 3`, so `parts.subList(0, parts.size() - 2)` is `subList(0, 1)` — only `parts.get(0)` (the critical count). The error count at `parts.get(1)` is never included anywhere in `sum`.

Example: 2 critical, 4 errors, 7 warnings produces `"2 critical errors, and 7 warnings"` — the `"4 errors"` element is silently missing from the message.

This `sum` string is used both for the informational log line and for the `IllegalStateException` message that fails the build (`Linting.java` around lines 242 and 254), so the message that actually stops a CI build under-reports the real defect count.

## What should happen

The slice should be `parts.subList(0, parts.size() - 1)` (everything except the last element), which correctly produces `"2 critical errors, 4 errors, and 7 warnings"`.

## Steps to reproduce

1. Run linting on a source set that has at least one critical defect, at least one error, and at least one warning.
2. Read the summary message (log output or build-failure exception message).
3. Observe the error count is missing from the sentence.

morphqdd:
Closes #6039.

`Linting.summary` built the sentence with `parts.subList(0, parts.size() - 2)` for the three-severity case, dropping `parts.get(1)` (the error count) entirely. With 2 critical, 4 errors, and 7 warnings, the message read "2 critical errors, and 7 warnings" - the error count never appeared anywhere. This same `sum` string is used both for the informational log line and for the `IllegalStateException` message that fails the build, so the message that actually stops CI under-reported the real defect count.

Fix: `parts.subList(0, parts.size() - 1)` (everything except the last element), which correctly includes all three parts.

Also widened `Linting.summary` from `private` to package-private so the new test can call it directly with a synthetic `Map<Severity, Integer>`, instead of needing to drive the full linting pipeline through three simultaneous defect types just to reach this one line.

Test: `LintingTest.summarizesAllThreeSeveritiesTogether`, asserting the exact expected sentence for 2 critical / 4 errors / 7 warnings. Verified it fails against the original code (missing "4 errors") and passes with the fix.

`mvn -pl eo-maven-plugin test -o -Dtest='LintingTest'` - 2 tests, all passing.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
@@ -209,6 +209,41 @@ final class Linting implements Step {
         }
     }
 
+    /**
+     * Summarize the counts.
+     * @param counts Counts of errors, warnings, and critical
+     * @return Summary text
+     */
+    static String summary(final Map<Severity, Integer> counts) {
+        final List<String> parts = new ArrayList<>(0);
+        final int critical = counts.get(Severity.CRITICAL);
+        if (critical > 0) {
+            parts.add(Linting.plural(critical, "critical error"));
+        }
+        final int errors = counts.get(Severity.ERROR);
+        if (errors > 0) {
+            parts.add(Linting.plural(errors, "error"));
+        }
+        final int warnings = counts.get(Severity.WARNING);
+        if (warnings > 0) {
+            parts.add(Linting.plural(warnings, "warning"));
+        }
+        if (parts.isEmpty()) {
+            parts.add("no complaints");
+        }
+        final String sum;
+        if (parts.size() < 3) {
+            sum = String.join(" and ", parts);
+        } else {
+            sum = String.format(
+                "%s, and %s",
+                String.join(", ", parts.subList(0, parts.size() - 1)),
+                parts.get(parts.size() - 1)
+            );
+        }
+        return sum;
+    }
+
     @SuppressWarnings("PMD.UnnecessaryLocalRule")
     private void linting() throws IOException {
         final Collection<TjForeign> programs = this.tojos.withXmir();
@@ -491,41 +526,6 @@ final class Linting implements Step {
             txt.append('s');
         }
         return txt.toString();
-    }
-
-    /**
-     * Summarize the counts.
-     * @param counts Counts of errors, warnings, and critical
-     * @return Summary text
-     */
-    private static String summary(final Map<Severity, Integer> counts) {
-        final List<String> parts = new ArrayList<>(0);
-        final int critical = counts.get(Severity.CRITICAL);
-        if (critical > 0) {
-            parts.add(Linting.plural(critical, "critical error"));
-        }
-        final int errors = counts.get(Severity.ERROR);
-        if (errors > 0) {
-            parts.add(Linting.plural(errors, "error"));
-        }
-        final int warnings = counts.get(Severity.WARNING);
-        if (warnings > 0) {
-            parts.add(Linting.plural(warnings, "warning"));
-        }
-        if (parts.isEmpty()) {
-            parts.add("no complaints");
-        }
-        final String sum;
-        if (parts.size() < 3) {
-            sum = String.join(" and ", parts);
-        } else {
-            sum = String.format(
-                "%s, and %s",
-                String.join(", ", parts.subList(0, parts.size() - 2)),
-                parts.get(parts.size() - 1)
-            );
-        }
-        return sum;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


morphqdd:

gemshrine (approved):
The corrected sublist retains every non-final severity, and the direct three-severity assertion protects the exact formatting regression.

gemshrine (commented):
Follow-up review: I checked all three formatting branches and the visibility change. The method remains internal to the Maven package, its zero-to-two-part wording is unchanged, and the three-part branch now retains the middle severity. The focused exact-string test is appropriate; CI is green.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -209,6 +209,41 @@ public void exec() throws IOException {
         }
     }
 
+    /**
+     * Summarize the counts.
+     * @param counts Counts of errors, warnings, and critical
+     * @return Summary text
```
gemshrine, line 215: Making this package-private is a narrowly scoped test seam: `Linting` is already package-local and the method is still not part of the plugin’s public API. It lets the test exercise the formatting contract directly instead of constructing a multi-stage linting run just to obtain three severities.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -209,6 +209,41 @@ public void exec() throws IOException {
...
+        final int errors = counts.get(Severity.ERROR);
+        if (errors > 0) {
+            parts.add(Linting.plural(errors, "error"));
+        }
+        final int warnings = counts.get(Severity.WARNING);
+        if (warnings > 0) {
+            parts.add(Linting.plural(warnings, "warning"));
+        }
+        if (parts.isEmpty()) {
+            parts.add("no complaints");
```
gemshrine, line 232: The `parts.size() < 3` branch preserves the established output for no complaints, one severity, and two severities. The changed branch is therefore isolated to the only cardinality that needs comma-separated enumeration.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -209,6 +209,41 @@ public void exec() throws IOException {
...
+        }
+        final String sum;
+        if (parts.size() < 3) {
+            sum = String.join(" and ", parts);
+        } else {
```
gemshrine, line 237: Using `size() - 1` is the essential correction: `subList` excludes its upper bound, so this selects the first two entries and `parts.get(last)` supplies the warning. The former `size() - 2` selected only the critical entry and silently discarded the middle error count.

eo-maven-plugin/src/test/java/org/eolang/maven/LintingTest.java
```diff
@@ -16,6 +21,19 @@
  */
 final class LintingTest {
 
+    @Test
+    void summarizesAllThreeSeveritiesTogether() {
+        final Map<Severity, Integer> counts = new EnumMap<>(Severity.class);
+        counts.put(Severity.CRITICAL, 2);
+        counts.put(Severity.ERROR, 4);
+        counts.put(Severity.WARNING, 7);
+        MatcherAssert.assertThat(
```
gemshrine, line 30: The deliberately distinct counts make this regression precise: a missing middle part cannot be hidden by repeated numbers, and the exact sentence checks both retention and ordering of critical, error, and warning severities.

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+944](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2604](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6054, opened 2026-07-28
#6040: add start anchor to string.regex matches

### LINKED ISSUE #6040: `string.regex`'s `matches` incorrectly considers text with unmatched leading characters as a full match
## What happens

`string.regex`'s `matches` (`eo-runtime/src/main/eo/string/regex.eo:53-57`) only checks that a match reaches the end of the text, never that it starts at the beginning:

```
[txt] > matches
  found.exists.and > @
    found.to.eq txt.length
  next. > found
    match txt
```

The underlying atom (`EOregex$EOpattern$EOmatch$EOmatched_from_index.java`) uses `matcher.find(start)`, which is unanchored at the front — it searches for the pattern anywhere in the text starting from `start`, not just at `start`. `matches` only verifies `found.to.eq txt.length` (the match reaches the end), so a match that starts partway through the text and happens to reach the end is still reported as a full match.

Verified against the built `eo-runtime` classes:
- `matches("/[a-z]+/", "1abc")` → **true** (should be `false`: `"1abc"` is not entirely lowercase letters, the leading `1` is not part of the match).
- `matches("/[0-9]+/", "x123")` → **true** (should be `false`, same reason).
- `matches("/[a-z]+/", "abc1")` → `false` (correctly rejected, since the match here doesn't reach the end).

## What should happen

`matches` should also require `found.from.eq 0`, matching Java's own `Matcher.matches()` semantics (whole-string match: both start at 0 and end at length), so text with leading unmatched characters is correctly rejected.

## Steps to reproduce

1. Build `eo-runtime`.
2. Call `("/[a-z]+/".compiled).matches "1abc"` (or the equivalent EO call through `string.regex`).
3. Observe it returns `true`, even though `"1abc"` is not entirely lowercase letters.

morphqdd:
Closes #6040.

`string.regex`'s `matches` only checked that a match reaches the end of the text (`found.to.eq txt.length`), never that it starts at the beginning. The underlying atom uses `matcher.find(start)`, which is unanchored at the front, so a match starting partway through the text and reaching the end was still reported as a full match. `matches("/[a-z]+/", "1abc")` returned `true`, even though `"1abc"` is not entirely lowercase letters.

Fix: also require `found.from.eq 0`, matching Java's own `Matcher.matches()` semantics (whole-string match: both start at 0 and end at length).

This is the unfinished half of #5779 (closed), which added the end-anchor (`found.to.eq txt.length`) to fix a trailing-newline false match, but never added the corresponding start-anchor.

Test: `tests-does-not-match-with-unmatched-leading-characters`, asserting `matches("/[a-z]+/", "1abc")` is false. Verified it fails against the original code and passes with the fix.

`mvn -pl eo-runtime test -o -Dtest='EO_string.EOregexTest'` - 28 tests, all passing.

### DIFF
--- a/eo-runtime/src/main/eo/string/regex.eo
+++ b/eo-runtime/src/main/eo/string/regex.eo
@@ -52,7 +52,9 @@
         matched-from-index 1 0
     [txt] > matches
       found.exists.and > @
-        found.to.eq txt.length
+        and.
+          found.to.eq txt.length
+          found.from.eq 0
       next. > found
         match txt
 
@@ -61,6 +63,12 @@
       compiled.
         regex "/^[0-9]+$/"
       "123\n"
+
+  not. ++> tests-does-not-match-with-unmatched-leading-characters
+    matches.
+      compiled.
+        regex "/[a-z]+/"
+      "1abc"
 
   not. ++> tests-does-not-matches-regex-against-the-pattern
     matches.


gemshrine (approved):
The added start-boundary check completes the full-text predicate, and the regression test covers the previously missed leading-text case.

eo-runtime/src/main/eo/string/regex.eo
```diff
@@ -52,7 +52,9 @@
         matched-from-index 1 0
     [txt] > matches
       found.exists.and > @
```
gemshrine, line 54: The `found.exists` guard is significant here: `match txt` may return the absent sentinel, and only a present match is allowed to evaluate either boundary. Thus this change keeps the existing no-match behaviour while strengthening the predicate for an actual match.

eo-runtime/src/main/eo/string/regex.eo
```diff
@@ -52,7 +52,9 @@
...
-        found.to.eq txt.length
+        and.
+          found.to.eq txt.length
+          found.from.eq 0
```
gemshrine, line 57: `found.to = txt.length` alone identifies a suffix, not a full-text match. Requiring `found.from = 0` makes the two bounds jointly describe `[0, txt.length)`, which is the contract callers expect from `matches` and is independent of whether the pattern itself contains `^` or `$`.

eo-runtime/src/main/eo/string/regex.eo
```diff
@@ -62,6 +64,12 @@
         regex "/^[0-9]+$/"
       "123\n"
 
+  not. ++> tests-does-not-match-with-unmatched-leading-characters
```
gemshrine, line 67: This is the right regression shape: the pattern deliberately has no leading anchor, so the test exercises the wrapper’s boundary logic instead of the regex engine’s `^` handling. Together with the existing trailing-newline test it covers both ends of the full-text interval.

gemshrine (commented):
Follow-up review: I traced `matches` through `match`, including the no-match path and the existing trailing-suffix regression. The change completes the whole-string predicate without changing the public match iterator; CI is green.

0crat:
@gemshrine Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (3) comments; -4 for too few (10) hits-of-code; +1 to give you at least something.  Your running score is [+930](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2644](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6055, opened 2026-07-28
#6041: turn Stack first-push-indent IllegalStateException into ParseError

### LINKED ISSUE #6041: Unhandled `IllegalStateException` thrown when the first line of code is indented
## What happens

`Stack.push` (`eo-parser/src/main/java/org/eolang/parser/Stack.java:171-179`) throws a raw `IllegalStateException` when the very first object line of a program is indented:

```java
if (this.levels.isEmpty()) {
    if (indent != 0) {
            String.format("first push must be at indent 0, was %d", indent)
        );
    }
    ...
```

`Eo.dispatch` (`eo-parser/src/main/java/org/eolang/parser/Eo.java:399-406`) only catches `ParseError`, so this `IllegalStateException` escapes `EoSyntax.parsed()` entirely — no line number, no `<errors>` XMIR entry, no recovery, unlike every other indentation problem in the same class.

Verified by parsing a minimal source where the first object line starts at indent 2 (one level in), e.g.:
```
# Sample.

  42 > x
```
This throws `IllegalStateException: first push must be at indent 0, was 2` straight out of `EoSyntax.parsed()`.

By contrast, other malformed first-line indents are already handled cleanly by neighboring checks in the same file: 1-space indent produces a clean `unexpected odd indent` `ParseError`, and 4-space indent produces `indent increased by more than one level`. Only the "one level in from an empty stack" case is uncaught.

## What should happen

This case should be reported as a normal `ParseError` (e.g. "unexpected indentation, the first object must start at indent 0"), the same way the neighboring indent violations already are, so a malformed source produces a clean `<errors>` entry instead of an internal-invariant crash.

## Steps to reproduce

1. Build `eo-parser`.
2. Parse a source whose first object line is indented by exactly one level (2 spaces) from column 0, e.g. `  42 > x`.
3. Observe `EoSyntax.parsed()` throws `IllegalStateException: first push must be at indent 0, was 2` instead of returning a `<errors>` XMIR entry.

morphqdd:
Closes #6041.

`Stack.push` threw a raw `IllegalStateException` when the very first object line of a program was indented (e.g. one level in, `  42 > x`). `Eo.dispatch` only catches `ParseError`, so this escaped `EoSyntax.parsed()` entirely - no line number, no `<errors>` XMIR entry, no recovery. Neighboring indent violations in the same method (odd indent, step-rule jump) are already surfaced as clean `<errors>` entries elsewhere in the pipeline; only this "first push at non-zero indent" case fell through uncaught.

Fix: throw `ParseError` (with the line/indent already available as method parameters) instead of `IllegalStateException`.

Updated the existing `StackTest.rejectsFirstPushAtNonZeroIndent`, which explicitly asserted the old `IllegalStateException` type, to expect `ParseError` instead.

Test: added `first-object-line-indented-is-rejected.yaml` under `eo-parser/src/test/resources/org/eolang/parser/eo-syntax/`, picked up automatically by `EoSyntaxTest.validatesEoSyntax`. Verified it fails against the original code (the harness's own exception, not a clean `<errors>` entry) and passes with the fix.

`mvn -pl eo-parser test -o -Dtest='EoSyntaxTest,StackTest'` - 951 tests, all passing.
`mvn -pl eo-parser qulice:check -Pqulice` - clean.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Stack.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Stack.java
@@ -170,10 +170,8 @@ final class Stack {
         final String owner;
         if (this.levels.isEmpty()) {
             if (indent != 0) {
-                throw new IllegalStateException(
-                    String.format(
-                        "first push must be at indent 0, was %d", indent
-                    )
+                throw new ParseError(
+                    line, indent, "unexpected indentation, the first object must start at indent 0"
                 );
             }
             parent = Kind.TOP_LEVEL;
--- a/eo-parser/src/test/java/org/eolang/parser/StackTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/StackTest.java
@@ -15,9 +15,7 @@ import org.junit.jupiter.api.Test;
  * Test case for {@link Stack}.
  * @since 0.1
  */
-@SuppressWarnings({
-    "PMD.TooManyMethods", "PMD.UnnecessaryLocalRule", "PMD.UnitTestContainsTooManyAsserts"
-})
+@SuppressWarnings({"PMD.TooManyMethods", "PMD.UnnecessaryLocalRule"})
 final class StackTest {
 
     @Test
@@ -42,24 +40,36 @@ final class StackTest {
 
     @Test
     void rejectsFirstPushAtNonZeroIndent() {
-        final ParseError error = Assertions.assertThrows(
+        Assertions.assertThrows(
             ParseError.class,
             () -> new Stack().push(2, 1, Kind.HEAD, Openness.OPEN),
             "first push must be at indent 0 — non-zero indent cannot start a program"
         );
+    }
+
+    @Test
+    void capturesLineOfFirstPushIndentViolation() {
         MatcherAssert.assertThat(
             "the error must carry the offending line",
-            error.line(),
+            StackTest.firstPushIndentViolation().line(),
             Matchers.equalTo(1)
         );
+    }
+
+    @Test
+    void capturesColumnOfFirstPushIndentViolation() {
         MatcherAssert.assertThat(
             "the error must carry the offending column",
-            error.pos(),
+            StackTest.firstPushIndentViolation().pos(),
             Matchers.equalTo(2)
         );
+    }
+
+    @Test
+    void capturesMessageOfFirstPushIndentViolation() {
         MatcherAssert.assertThat(
             "the error message must name the indent-0 requirement",
-            error.getMessage(),
+            StackTest.firstPushIndentViolation().getMessage(),
             Matchers.equalTo(
                 "unexpected indentation, the first object must start at indent 0"
             )
@@ -194,6 +204,17 @@ final class StackTest {
             "after close() the stack cannot retain any entry",
             stack.empty(),
             Matchers.is(true)
+        );
+    }
+
+    /**
+     * Trigger the first-push-at-non-zero-indent violation and capture it.
+     * @return The thrown error
+     */
+    private static ParseError firstPushIndentViolation() {
+        return Assertions.assertThrows(
+            ParseError.class,
+            () -> new Stack().push(2, 1, Kind.HEAD, Openness.OPEN)
         );
     }
 }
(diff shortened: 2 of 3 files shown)


gemshrine (approved):
The parser now routes this indentation violation through the established ParseError recovery path, with focused unit and syntax-level coverage.

eo-parser/src/main/java/org/eolang/parser/Stack.java
```diff
@@ -170,10 +170,8 @@ Level push(
         final String owner;
         if (this.levels.isEmpty()) {
             if (indent != 0) {
-                throw new IllegalStateException(
-                    String.format(
-                        "first push must be at indent 0, was %d", indent
-                    )
+                throw new ParseError(
```
gemshrine, line 173: This is the correct recovery boundary. `ParseError` carries the source row and zero-based column to the single parser catch site, which rolls the emitter back and creates XMIR instead of letting a generic runtime exception escape. Passing the existing `line` and `indent` values also points the diagnostic at the first invalid token.

eo-parser/src/test/java/org/eolang/parser/StackTest.java
```diff
@@ -41,7 +41,7 @@ void pushesTopLevelEntryAtIndentZero() {
     @Test
     void rejectsFirstPushAtNonZeroIndent() {
         Assertions.assertThrows(
```
gemshrine: Please retain the thrown value and assert its payload as well as its type. The change’s user-visible contract is not merely `ParseError`; it is the exact source location and canonical message passed at lines 173–175. `ParseError` exposes package-visible `line()` and `pos()`, so this test can lock down `1`, `2`, and the message without widening the production API.

eo-parser/src/test/resources/org/eolang/parser/eo-syntax/first-object-line-indented-is-rejected.yaml
```diff
@@ -0,0 +1,11 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets: []
+asserts:
+  - /object/errors/error[contains(text(),'indent')]
```
gemshrine, line 9: Please assert the emitted error’s line and severity here, in addition to the text fragment. This integration fixture is the only proof that the walker converts the exception into a located `<errors>` entry; `contains(text(), 'indent')` would also pass for an unrelated indentation error or for a regression that loses the source position.

gemshrine (commented):
Follow-up review: the exception is now routed through the parser recovery boundary, and the existing CI run is green. I traced the constructor arguments to `ParseError` and the XMIR emitter. The implementation is right; the two tests should make the promised location and diagnostic contract explicit.

eo-parser/src/test/java/org/eolang/parser/StackTest.java
```diff
@@ -15,7 +15,9 @@
  * Test case for {@link Stack}.
  * @since 0.1
  */
-@SuppressWarnings({"PMD.TooManyMethods", "PMD.UnnecessaryLocalRule"})
+@SuppressWarnings({
+    "PMD.TooManyMethods", "PMD.UnnecessaryLocalRule", "PMD.UnitTestContainsTooManyAsserts"
```
yegor256: @morphqdd get rid of `PMD.UnitTestContainsTooManyAsserts` and fix the code

morphqdd: Split into separate one-assert-per-test methods with a shared private helper instead of suppressing. Pushed.

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (5) comments.  Your running score is [+937](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2620](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6056, opened 2026-07-28
#6042: turn build-fqns.xsl TerminationException into a proper parse error

### LINKED ISSUE #6042: `build-fqns.xsl` terminates parse instead of reporting error when `@` has no `φ` in scope
## What happens

`build-fqns.xsl` (`eo-parser/src/main/resources/org/eolang/parser/parse/build-fqns.xsl:83-89`) aborts the entire parse with a raw Saxon termination when `@` (`φ`) is referenced but no `φ` is in scope:

```xml
<xsl:if test="$find='φ'">
  <xsl:message terminate="yes">
    <xsl:text>The </xsl:text>
    <xsl:value-of select="$find"/>
    <xsl:text> object is used, but absent in self or parents scope</xsl:text>
  </xsl:message>
</xsl:if>
```

`terminate="yes"` stops the whole XSL pipeline, which surfaces to the caller as:

```
IllegalArgumentException: Shift 'com.yegor256.xsline.StClasspath@..' failed
Caused by: net.sf.saxon.expr.instruct.TerminationException: Processing terminated by xsl:message at line 85 in build-fqns.xsl
```

Verified by parsing a minimal source that uses `@` where no `φ` is in scope:
```
# Sample.

[] > sample
  @ > x
```

The user sees a stack trace naming an internal XSL file and a line number inside it, with no reference to their own source file, line, or position, and no `<errors>` XMIR entry at all — unlike every other diagnostic in this pipeline.

## What should happen

This should emit an `<errors><error line=".." pos="..">` entry (e.g. "`@` is used but no `φ` exists in this or any enclosing scope"), consistent with how every other parse-time validation in this pipeline reports problems, instead of terminating the whole XSL train.

## Steps to reproduce

1. Build `eo-parser`.
2. Parse a source that references `@` in a scope with no enclosing `φ`, e.g.:
   ```
   [] > sample
     @ > x
   ```
3. Observe the parse throws a raw `TerminationException` wrapped in `IllegalArgumentException`, naming `build-fqns.xsl` and an internal line number, with no `<errors>` entry and no reference to the user's own source location.

morphqdd:
Closes #6042.

`build-fqns.xsl` aborted the whole XSL pipeline with `<xsl:message terminate="yes">` when `@` (`φ`) is referenced but no `φ` is in scope. This surfaced to the caller as a raw `TerminationException` wrapped in `IllegalArgumentException`, naming an internal XSL file and line number, with no reference to the user's own source location and no `<errors>` XMIR entry - unlike every other diagnostic in this pipeline.

Fix: removed the `terminate="yes"` message. An unresolved `φ` reference now stays as a literal `<o base='φ'>` node in the transformed tree (nothing rewrites it, same as before, just without terminating). Added a new top-level `/object` template that runs the existing per-node resolution as before, then scans the transformed result for any surviving `o[@base='φ']` nodes and emits a proper `<errors><error>` entry for each, merged with whatever errors already existed from earlier shifts in the chain - the same pattern already used by `resolve-self.xsl` for its own "used outside valid scope" check.

One implementation snag along the way: my first attempt declared the intermediate `$transformed` variable without a type, which made Saxon build it as a document-node-rooted temporary tree - `/object`'s own attributes (like `@author`) ended up as children of that document node, which Saxon rejects ("Cannot output an attribute node whose parent is a document node"), breaking every single test. Declaring it `as="item()*"` forces a plain sequence instead, which fixed it.

Test: added `unresolved-phi-is-rejected.yaml` under `eo-parser/src/test/resources/org/eolang/parser/eo-syntax/`, picked up automatically by `EoSyntaxTest.validatesEoSyntax`. Verified it fails against the original code (a raw uncaught exception, not a clean `<errors>` entry) and passes with the fix.

`mvn -pl eo-parser test -o -Dtest='EoSyntaxTest'` - 938 tests, all passing.

Note: I could not run `xslint` locally (no local Ruby setup for it in my environment), so I'm relying on CI for that specific style check; the functional correctness is covered by the full parser test suite above, which exercises hundreds of real EO programs end to end.

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/build-fqns.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/build-fqns.xsl
@@ -82,13 +82,6 @@
         </xsl:apply-templates>
       </xsl:when>
       <xsl:otherwise>
-        <xsl:if test="$find='φ'">
-          <xsl:message terminate="yes">
-            <xsl:text>The </xsl:text>
-            <xsl:value-of select="$find"/>
-            <xsl:text> object is used, but absent in self or parents scope</xsl:text>
-          </xsl:message>
-        </xsl:if>
         <xsl:copy>
           <xsl:apply-templates select="node()|@*"/>
         </xsl:copy>
@@ -216,6 +209,38 @@
   <xsl:template match="node()|@*" mode="#all">
     <xsl:copy>
       <xsl:apply-templates select="node()|@*"/>
+    </xsl:copy>
+  </xsl:template>
+  <!--
+  A "φ" reference left unresolved by "with-package" above stays as a
+  literal "<o base='φ'>" in the transformed tree (nothing rewrites it).
+  Reporting this here, after the transform, rather than terminating the
+  whole XSL train mid-resolution (as this file used to), lets it surface
+  as a normal <errors> entry with the offending line, consistent with
+  every other diagnostic in this pipeline (see #6042).
+  -->
+  <xsl:template match="/object">
+    <xsl:variable name="transformed" as="item()*">
+      <xsl:apply-templates select="(node() except errors)|@*"/>
+    </xsl:variable>
+    <xsl:copy>
+      <xsl:sequence select="$transformed"/>
+      <xsl:variable name="errors" as="element()*">
+        <xsl:for-each select="$transformed//o[@base='φ']">
+          <error>
+            <xsl:attribute name="check" select="'build-fqns'"/>
+            <xsl:attribute name="line" select="if (@line) then @line else 0"/>
+            <xsl:attribute name="severity" select="'error'"/>
+            <xsl:text>The φ object is used, but absent in self or parents scope</xsl:text>
+          </error>
+        </xsl:for-each>
+      </xsl:variable>
+      <xsl:if test="not(empty($errors)) or exists(/object/errors)">
+        <errors>
+          <xsl:apply-templates select="/object/errors/error"/>
+          <xsl:copy-of select="$errors"/>
+        </errors>
+      </xsl:if>
     </xsl:copy>
   </xsl:template>
 </xsl:stylesheet>
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/unresolved-phi-is-rejected.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/unresolved-phi-is-rejected.yaml
@@ -1,0 +1,10 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets: []
+asserts:
+  - /object/errors/error[contains(text(),'φ')]
+input: |
+  [] > sample
+    @ > x
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/unresolved-phi-is-rejected.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/unresolved-phi-is-rejected.yaml
@@ -4,6 +4,9 @@
 # yamllint disable rule:line-length
 sheets: []
 asserts:
+  - /object/errors/error[@check='build-fqns']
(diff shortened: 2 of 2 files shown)


gemshrine (approved):
The transform now preserves the unresolved reference long enough to emit a normal parser error, while retaining pre-existing errors. The syntax fixture covers the former uncaught path.

eo-parser/src/main/resources/org/eolang/parser/parse/build-fqns.xsl
```diff
@@ -218,4 +211,36 @@
...
+  <!--
+  A "φ" reference left unresolved by "with-package" above stays as a
+  literal "<o base='φ'>" in the transformed tree (nothing rewrites it).
+  Reporting this here, after the transform, rather than terminating the
+  whole XSL train mid-resolution (as this file used to), lets it surface
+  as a normal <errors> entry with the offending line, consistent with
+  every other diagnostic in this pipeline (see #6042).
+  -->
+  <xsl:template match="/object">
+    <xsl:variable name="transformed" as="item()*">
```
gemshrine, line 223: Keeping this as `item()*` is important here. Without the sequence type, the variable body becomes a temporary document node; then the copied `/object` attributes would be emitted beneath that document node and Saxon rejects the result. The explicit type preserves the normal attribute-before-child order when `$transformed` is replayed inside the copied root.

eo-parser/src/main/resources/org/eolang/parser/parse/build-fqns.xsl
```diff
@@ -218,4 +211,36 @@
...
+      <xsl:apply-templates select="(node() except errors)|@*"/>
+    </xsl:variable>
+    <xsl:copy>
+      <xsl:sequence select="$transformed"/>
+      <xsl:variable name="errors" as="element()*">
+        <xsl:for-each select="$transformed//o[@base='φ']">
```
gemshrine, line 229: Scanning the transformed sequence, rather than the original source tree, is the right boundary: references that `with-package` resolved no longer have `@base='φ'`, while every unresolved occurrence produces its own diagnostic. This also keeps the detection separate from the per-node resolution templates and avoids restoring the earlier control-flow abort.

eo-parser/src/test/resources/org/eolang/parser/eo-syntax/unresolved-phi-is-rejected.yaml
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets: []
+asserts:
+  - /object/errors/error[contains(text(),'φ')]
```
gemshrine, line 10: Please extend this assertion to cover the observable diagnostic contract, not only a text fragment: `@check='build-fqns'`, `@severity='error'`, and the offending source line. The purpose of #6042 is to replace an internal XSL failure with a located parser error; an unrelated or unlocated error containing `φ` would satisfy the current XPath. A second input that already produces an earlier parser error would also protect the new merge branch at lines 238–242.

gemshrine (commented):
Follow-up after a deeper pass: the implementation correctly changes the failure mode from an XSL termination into an XMIR diagnostic. I checked the new top-level template against the analogous parser validation stages and the completed CI run. The intermediate sequence and the post-transform scan are appropriate; the remaining concern is that the regression fixture currently proves only a fragment of the diagnostic contract.

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+958](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2660](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6069, opened 2026-07-28
#6033: write cache payload before the validity marker

### LINKED ISSUE #6033: `Cache.apply` risks stale cache entries due to premature validity marker writing
## What happens

`Cache.apply` (`eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java:83-99`) writes the cache's validity marker before it writes the payload it is supposed to vouch for:

```java
void apply(final Path source, final Path target, final Path tail) {
    try {
        final String sha = this.sha(source);
        final Path hash = this.hash(tail);
        final Path cache = this.base.resolve(tail);
        if (
            Files.notExists(hash) || Files.notExists(cache)
                || !Files.readString(hash).equals(sha)
        ) {
            final String content = new UncheckedFunc<>(this.compilation).apply(source);
            new Saved(sha, this.hash(tail)).value();   // marker written first
            new Saved(content, cache).value();          // payload written second
            new Saved(content, target).value();
        } else {
            new Saved(Files.readString(cache), target).value();
        }
    } ...
}
```

If the JVM is killed, the disk fills up, or an `IOException` is thrown between the first `Saved(...)` call and the second, the `.sha256` marker file now records the new source hash while the `cache` payload file still holds the previous (or no) content. On the next run, the freshness check `Files.notExists(hash) || Files.notExists(cache) || !Files.readString(hash).equals(sha)` passes on all three branches — the hash file exists and matches the current source — so the stale (or missing) payload is treated as valid and copied to `target` without ever being recomputed. This persists silently until someone manually clears the cache directory.

Issue #5873 hardened `Saved` itself so a single file write is never partially visible to a concurrent reader, but that fix does not cover the ordering between these two separate `Saved` calls.

## What should happen

Write the payload first, then the hash marker last, so the marker only exists once the payload it vouches for is fully and successfully written:

```java
final String content = new UncheckedFunc<>(this.compilation).apply(source);
new Saved(content, cache).value();
new Saved(content, target).value();
new Saved(sha, this.hash(tail)).value();
```

## Steps to reproduce

This requires simulating abnormal termination, e.g. by making the second `Saved` call throw (or by killing the process) between the two writes in a test, then observing that a subsequent `Cache.apply` call with the same source treats the (incomplete or stale) cache entry as valid.

morphqdd:
Closes #6033.

`Cache.apply` wrote the sha256 marker before the payload:

```java
new Saved(sha, this.hash(tail)).value();
new Saved(content, cache).value();
new Saved(content, target).value();
```

If the process dies (crash, disk full, IOException) between the first and second write, the marker already matches the new source hash while the cache file still holds the old or missing content. On the next run the freshness check passes on all branches and the stale entry is served forever, silently.

Fix: write the payload first, marker last, so the marker can never exist without the payload it vouches for.

```java
new Saved(content, cache).value();
new Saved(content, target).value();
new Saved(sha, this.hash(tail)).value();
```

Test: `CacheTest.leavesMarkerUntouchedWhenPayloadWriteFails` forces the payload write to fail (by putting a directory where the cache file should be) and checks the marker still points at the old, valid payload, then checks a following call recompiles instead of serving a stale entry. Reverted the fix locally and confirmed this test fails against the old write order.

`mvn -pl eo-maven-plugin test -Dtest=CacheTest,ConcurrentCacheTest,ChCachedTest,OyCachedTest` - 42 tests, all passing.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
@@ -91,9 +91,9 @@ final class Cache {
                     || !Files.readString(hash).equals(sha)
             ) {
                 final String content = new UncheckedFunc<>(this.compilation).apply(source);
-                new Saved(sha, this.hash(tail)).value();
                 new Saved(content, cache).value();
                 new Saved(content, target).value();
+                new Saved(sha, this.hash(tail)).value();
             } else {
                 new Saved(Files.readString(cache), target).value();
             }
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
@@ -30,11 +30,6 @@ import org.junit.jupiter.api.extension.ExtendWith;
 @SuppressWarnings({"PMD.TooManyMethods", "PMD.UnitTestContainsTooManyAsserts"})
 final class CacheTest {
 
-    /**
-     * Marker file name pattern, relative to a cached tail.
-     */
-    private static final String MARKER = "%s.sha256";
-
     @Test
     void compilesSource(@Mktmp final Path temp) throws Exception {
         final Path base = temp.resolve("cache-folder");
@@ -62,7 +57,7 @@ final class CacheTest {
         MatcherAssert.assertThat(
             "Cache file must be created and hash file must be created",
             Files.exists(base.resolve(tail))
-                && Files.exists(base.resolve(String.format(CacheTest.MARKER, tail))),
+                && Files.exists(base.resolve(CacheTest.marker(tail))),
             Matchers.is(true)
         );
     }
@@ -150,7 +145,7 @@ final class CacheTest {
         cache.apply(source, temp.resolve("out.txt"), tail);
         MatcherAssert.assertThat(
             "SHA-256 hash file has incorrect content",
-            Files.readString(base.resolve(String.format(CacheTest.MARKER, tail)), encoding),
+            Files.readString(base.resolve(CacheTest.marker(tail)), encoding),
             Matchers.equalTo(
                 Base64.getEncoder().encodeToString(
                     MessageDigest.getInstance("SHA-256").digest(msg.getBytes(encoding))
@@ -178,7 +173,7 @@ final class CacheTest {
         MatcherAssert.assertThat(
             "SHA-256 hash file has incorrect content for large file",
             Files.readString(
-                cache.resolve(String.format(CacheTest.MARKER, tail)),
+                cache.resolve(CacheTest.marker(tail)),
                 StandardCharsets.UTF_8
             ),
             Matchers.equalTo(
@@ -204,7 +199,7 @@ final class CacheTest {
         MatcherAssert.assertThat(
             "SHA-256 hash file has incorrect content for tiny file",
             Files.readString(
-                cache.resolve(String.format(CacheTest.MARKER, tail)),
+                cache.resolve(CacheTest.marker(tail)),
                 StandardCharsets.UTF_8
             ),
             Matchers.equalTo(CacheTest.hash(content))
@@ -304,7 +299,7 @@ final class CacheTest {
         MatcherAssert.assertThat(
             "the marker must not point at a payload that was never written",
             Files.readString(
-                base.resolve(String.format(CacheTest.MARKER, tail)), StandardCharsets.UTF_8
+                base.resolve(CacheTest.marker(tail)), StandardCharsets.UTF_8
             ),
             Matchers.equalTo(CacheTest.hash(initial))
         );
@@ -325,5 +320,14 @@ final class CacheTest {
             MessageDigest.getInstance("SHA-256")
                 .digest(content.getBytes(StandardCharsets.UTF_8))
         );
+    }
+
+    /**
+     * Marker file name for the given cached tail.
+     * @param tail Tail path in cache
+     * @return Marker file name, relative to the same directory as the tail
+     */
+    private static String marker(final Path tail) {
+        return String.format("%s.sha256", tail);
     }
 }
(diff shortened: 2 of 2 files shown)


eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
```diff
@@ -26,9 +27,14 @@
  * @since 0.60
  */
 @ExtendWith(MktmpResolver.class)
-@SuppressWarnings("PMD.TooManyMethods")
+@SuppressWarnings({"PMD.TooManyMethods", "PMD.UnitTestContainsTooManyAsserts"})
 final class CacheTest {
 
+    /**
+     * Marker file name pattern, relative to a cached tail.
+     */
```
gemshrine: `qulice` currently fails because JTCOP prohibits static fields in test classes (`RuleProhibitStaticFields`). Please replace this new `MARKER` field with a non-static helper or inline the format so the required quality gate passes.

gemshrine (changes requested):
The cache-write ordering and its failure-path test look sound. One blocking quality-gate violation remains in the new test helper.

eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
```diff
@@ -91,9 +91,9 @@ void apply(final Path source, final Path target, final Path tail) {
                     || !Files.readString(hash).equals(sha)
             ) {
                 final String content = new UncheckedFunc<>(this.compilation).apply(source);
-                new Saved(sha, this.hash(tail)).value();
                 new Saved(content, cache).value();
```
gemshrine, line 94: Writing the cached payload before any validity metadata establishes the required invariant: until the payload is durable, the old marker remains the only marker the next invocation can trust. A failure at this first write leaves the prior cache generation intact rather than publishing a hash for nonexistent content.

eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
```diff
@@ -91,9 +91,9 @@ void apply(final Path source, final Path target, final Path tail) {
...
                 new Saved(content, target).value();
+                new Saved(sha, this.hash(tail)).value();
```
gemshrine, line 96: Placing the marker last makes it the commit point of this three-file update. If the target write fails at line 95, the new payload may exist but its old marker cannot validate it; the next call recompiles safely. Only after both payload writes succeed can the new source hash be published.

eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
```diff
@@ -273,10 +274,60 @@ void generatesDifferentHashesForFoldersWithDifferentlyNamedIdenticalFiles(
...
+        Files.createDirectories(base);
+        final Path source = temp.resolve("order.eo");
+        final String initial = "v1";
+        Files.writeString(source, initial, StandardCharsets.UTF_8);
+        final Path target = temp.resolve("order.xmir");
+        final Path tail = source.getFileName();
+        new Cache(base, p -> initial).apply(source, target, tail);
+        final String updated = "v2";
+        Files.writeString(source, updated, StandardCharsets.UTF_8);
+        final Path cached = base.resolve(tail);
```
gemshrine: Replacing the cached file with a directory is a deterministic way to force `Saved` to fail at the payload path while keeping the old marker available for inspection. It exercises the exact window that the former write order handled incorrectly, without relying on timing or filesystem permissions.

eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
```diff
@@ -273,10 +274,60 @@ void generatesDifferentHashesForFoldersWithDifferentlyNamedIdenticalFiles(
...
+                base.resolve(CacheTest.marker(tail)), StandardCharsets.UTF_8
+            ),
+            Matchers.equalTo(CacheTest.hash(initial))
+        );
+        Files.delete(cached);
+        final AtomicInteger counter = new AtomicInteger(0);
+        new Cache(
+            base, p -> String.format("%s-%d", updated, counter.incrementAndGet())
+        ).apply(source, target, tail);
+        MatcherAssert.assertThat(
```
gemshrine, line 313: This follow-up assertion is essential: retaining the old marker after the failed write is safe only if the modified source no longer matches it. The incrementing compilation function proves that the next invocation takes the miss path and does not serve the stale cache entry.

gemshrine (approved):
The previously requested quality-gate fix is now present: the test no longer adds a prohibited static field, and the new CI run has passed `qulice`. I rechecked the write ordering and the failure/recovery path. The marker is now the commit point for a usable cache entry, and the test establishes both failure safety and subsequent recompilation.

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (5) comments.  Your running score is [+965](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java
```diff
@@ -26,7 +27,7 @@
  * @since 0.60
  */
 @ExtendWith(MktmpResolver.class)
-@SuppressWarnings("PMD.TooManyMethods")
+@SuppressWarnings({"PMD.TooManyMethods", "PMD.UnitTestContainsTooManyAsserts"})
```
yegor256: @morphqdd don't do this, fix the code instead

morphqdd: Split into three one-assert-per-test methods with a shared private fixture instead of suppressing. Pushed.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (141 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2756](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6070, opened 2026-07-28
#6024: fix number.random next mask restricting output below 0.5

### LINKED ISSUE #6024: `number.random`'s `next` function restricts output below 0.5, contrary to the intended [0,1) range
In `eo-runtime/src/main/eo/number/random.eo`, the `next` handle (lines 34-53) computes a new LCG seed and masks the result with the byte literal `00-0F-FF-FF-FF-FF-FF-FF` (line 53), which is `2^52-1`. But `random`'s own `@` (line 12) divides the seed by `00-20-00-00-00-00-00-00.as-i64.as-number`, which is `2^53`.

Since the masked numerator can never exceed `2^52-1`, dividing by `2^53` bounds every value `.next` can produce at just under `0.5`; chaining `.next.next`, `.next.next.next`, etc. does not help, since each step re-applies the same mask. For any seed, the upper half of `[0,1)` is mathematically unreachable. The sibling `clocked` construct a few lines above (lines 13-33) builds its analogous first-chunk mask correctly as `(one.left 53).as-i64.minus one`, i.e. `2^53-1`, matching the same `/2^53` divisor used in `@` — so `next`'s mask looks like a one-hex-digit typo rather than an intentional restriction. The existing test `tests-random-is-in-range` only asserts `0 <= next < 1`, which is too weak to catch a half-range restriction, since it never checks that values `>= 0.5` are reachable.

Suggested fix: change the byte literal on line 53 from `00-0F-FF-FF-FF-FF-FF-FF` to `00-1F-FF-FF-FF-FF-FF-FF` (`2^53-1`), matching the mask already used by `clocked` in the same file.

morphqdd:
Closes #6024.

`number.random`'s `next` (line 53) masked its new LCG state with `00-0F-FF-FF-FF-FF-FF-FF` (2^52-1), but `random`'s own `@` divides that state by `00-20-00-00-00-00-00-00.as-i64.as-number` (2^53). Since the masked value can never reach 2^52, dividing by 2^53 bounds every output at just under 0.5, no matter the seed.

The sibling `clocked` builds its equivalent mask correctly as `(one.left 53).as-i64.minus one`, i.e. 2^53-1, matching the same divisor. `next`'s mask looks like a one-hex-digit typo.

Fix: change the mask to `00-1F-FF-FF-FF-FF-FF-FF` (2^53-1).

Test: added `tests-random-next-can-reach-upper-half` with seed 178609, chosen so `(seed * 25214903917 + 11) mod 2^53 >= 2^52`, i.e. the fixed output is just over 0.5. Reverted the fix locally and confirmed the test fails against the old mask (0.5 unreachable by construction), then confirmed it passes with the fix.

`mvn -pl eo-runtime test -Dtest=EOrandomTest` - 5 tests, all passing.

### DIFF
--- a/eo-runtime/src/main/eo/number/random.eo
+++ b/eo-runtime/src/main/eo/number/random.eo
@@ -50,7 +50,7 @@
                   (seed.as-i64.as-bytes.and 00-00-00-00-03-FF-FF-FF).as-i64.times
                     25214903917.as-i64
               11.as-i64
-          00-0F-FF-FF-FF-FF-FF-FF
+          00-1F-FF-FF-FF-FF-FF-FF
   clocked clock > pseudo
 
   ++> tests-random-is-in-range
@@ -70,6 +70,12 @@
           rand.next.next.lt 0
     next. > rand
       random 123
+
+  ++> tests-random-next-can-reach-upper-half
+    not. > @
+      rand.lt 0.5
+    next. > rand
+      random 178609
 
   ++> tests-random-with-seed
     not. > @


eo-runtime/src/main/eo/number/random.eo
```diff
@@ -50,7 +50,7 @@
                   (seed.as-i64.as-bytes.and 00-00-00-00-03-FF-FF-FF).as-i64.times
                     25214903917.as-i64
               11.as-i64
-          00-0F-FF-FF-FF-FF-FF-FF
+          00-1F-FF-FF-FF-FF-FF-FF
```
gemshrine, line 53: This mask now has 53 one-bits (`2^53 - 1`), matching the denominator at line 12. The previous `00-0F-…` value had only 52 one-bits, so every next-state value was bounded by `2^52 - 1` and the public number was necessarily below `0.5`.

gemshrine, line 53: Applying the mask after the two 26-bit multiplication components are recombined is also the correct modulo boundary. It preserves the existing overflow-safe decomposition while reducing the final LCG state modulo `2^53`, consistently with the clocked generator’s `(one.left 53) - one` mask.

eo-runtime/src/main/eo/number/random.eo
```diff
@@ -71,6 +71,12 @@
     next. > rand
       random 123
 
+  ++> tests-random-next-can-reach-upper-half
```
gemshrine, line 74: The chosen seed is a good deterministic witness rather than a probabilistic test. Its first next-state is `4,503,608,773,711,464`, giving `0.5000010154478334` after division by `2^53`; under the former 52-bit mask the same assertion is impossible to satisfy. This isolates the width regression without depending on clock time.

gemshrine (commented):
I checked the arithmetic boundary against the 53-bit divisor and replayed the supplied seed: the updated recurrence produces 4,503,608,773,711,464 / 2^53 = 0.5000010154478334, whereas the previous 52-bit mask could never exceed 0.5. The implementation and regression shape are correct. I am leaving this detailed review while the remaining platform test jobs finish.

gemshrine (approved):
Approved after verifying the 53-bit modulo boundary, the deterministic upper-half regression, and the completed Linux integration and quality checks. The remaining platform jobs are still running and remain protected required checks.

0crat:
@gemshrine Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (3) comments; -4 for too few (8) hits-of-code; +1 to give you at least something.  Your running score is [+969](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2632](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6072, opened 2026-07-28
Fix the lint warnings in eo-runtime EO sources

### LINKED ISSUE #5502: Fix lint warnings in EO source files
Several EO source files produce `[LINT]` warnings that should be resolved:

- `redundant-object/S` - redundant objects that can be inlined (`ms.radians`, `ms.arc-sine`, `ms.degrees`, `ms.sine`, `ms.power`, `ms.ln`)
- `unsorted-named-attributes/S` - named attributes out of order (`ms.arc-sine`, `ms.sine`, `ms.power`, `ms.ln`, `sm.win32`, `sm.posix`, `fs.file`, `number`)
- `unsorted-metas/S` - meta declarations out of order (`fs.dir`)
- `ascii-only/S` - non-ASCII character `⊥` used in a comment (`recovered`)

yegor256:
This clears 86 of the 338 `[LINT]` warnings that `mvn test` prints in
`eo-runtime`. The first commit is metas and comments: the root-level objects
that were missing `+unlint mandatory-package` now carry it like their 38
siblings, four stale `+unlint redundant-object` metas are gone, `+alias` lines
are sorted by alias name, and the comments on `chunk`, `uri`, `tuple.back` and
`recovered` lost their leading article, gained a final dot, and dropped the `⊥`
glyph. The second commit inlines every object that was named once and used
once, renames the compound names that survive that (`bottom-negated` is
`subtrahend`, `without-key` is `others`, `signed-prefix` is `signed`),
privatises the `uri` predicates `has-authority`, `has-fragment`, `has-port`,
`has-query` and `has-userinfo` into `>>` handles, and adds three unit tests for
`chunk`. 1799 tests pass.

Two things did not get fixed and I think they should not be. `squared` in
`number.arc-sine` stays named: its single reference sits inside the recursive
`poly`, so inlining hands each of the thirty recursion levels its own copy of
the subexpression — `EOarc_sineTest` goes from about 9 seconds to 189 and the
parallel suite runs out of heap. That one is filed as objectionary/lints#1132.

The other 251 warnings are false positives in `lints` and I filed them there:
objectionary/lints#1129 (237 of them — `non-kebab-name`, `sparse-decoration`,
`wrong-test-order` and `unit-test-missing` only recognise `+`-prefixed test
names, while the parser prefixes a `-->` throwing test with `-`),
objectionary/lints#1130 (`sparse-decoration` on `[@] > input`, where the lone φ
is void and so not a decoration at all) and objectionary/lints#1131
(`phi-is-not-first` on a `| > @` pipe, which is bound to the preceding sibling
and can never be moved first). Once those land, `eo-runtime` should be clean.

Closes #5502

### DIFF
--- a/eo-runtime/src/main/eo/bytes.eo
+++ b/eo-runtime/src/main/eo/bytes.eo
@@ -19,7 +19,6 @@
 +version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
-+unlint redundant-object
 +unlint mandatory-package
 
 [@] > bytes
--- a/eo-runtime/src/main/eo/chunk.eo
+++ b/eo-runtime/src/main/eo/chunk.eo
@@ -53,3 +53,25 @@
   [size] > resized /Q.chunk
   [] > size /Q.number
   [offset data] > write /Q.bool
+
+  eq. ++> tests-chunk-gets-what-was-put
+    malloc.of
+      8
+      seq * > [c]
+        c.put 42
+        c.get
+    42.as-bytes
+
+  eq. ++> tests-chunk-knows-its-size
+    malloc.of
+      3
+      c.size > [c]
+    3
+
+  eq. ++> tests-chunk-reads-a-slice-of-what-was-written
+    malloc.of
+      5
+      seq * > [c]
+        c.write 0 "hello"
+        c.read 1 3
+    "ell".as-bytes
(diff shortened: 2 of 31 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (591 >= 100); -8 for way too many hits-of-code (591 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2693](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6073, opened 2026-07-28
#6071: reject placeholder-recovered sources in eo:format

### LINKED ISSUE #6071: `eo:format -Deo.autoFix` improperly overwrites sources with recovered parse errors
`MjFormat.parsed` (`eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java:210`)
only rejects a source when it has severe `<errors>` **and**
`MjFormat.truncated` says the recovered tree stops short of the last non-blank
line. That line-coverage proxy for "the error was merely cosmetic" is too weak:
the parser recovers from some *structural* errors by substituting a placeholder
node and then carries on consuming every remaining line, so coverage stays
intact while the tree no longer describes the source.

A reversed dispatch left without a receiver is one such case. For

```eo
[] > foo
  if. > @
    if.
    true
    1
    2
```

the parser reports `reversed dispatch missing receiver` at `[5:4]` and emits
`<o base=".if"><o/></o>` — an empty formation standing in for the missing
receiver. Every line is still covered, so `truncated` returns FALSE, the guard
does not fire, and under `-Deo.autoFix` the canonical text (now carrying a
literal `[]`) is written over the file by `reformat` at line 132. The error
never reaches the user, and a source the parser rejected has become a different
source it accepts. This is the gap left by the fix for #5851, which deliberately
chose line coverage over rejecting every severe error.

The smallest fix is to widen the condition at line 210 so a recovered
placeholder also counts as loss: alongside `truncated`, reject when the parsed
tree contains an `<o>` with neither `@base` nor `@name` nor children, since the
printer has no way to render that except as `[]`.

yegor256:
Fixes #6071.

## Problem

`MjFormat.parsed` rejected a source only when a severe `<errors>` entry
coincided with `MjFormat.truncated` reporting that the recovered tree stopped
short of the last non-blank line. That line-coverage proxy is too weak: the
parser recovers from some *structural* errors by substituting an empty
placeholder node and then carries on consuming every remaining line, so
coverage stays intact while the tree no longer describes the source.

A reversed dispatch left without a receiver is one such case:

```eo
[] > foo
  if. > @
    if.
    true
    1
    2
```

The parser reports `reversed dispatch missing receiver` and emits an empty
`<o/>` standing in for the missing receiver. Every line is still covered, so
`truncated` returns `FALSE`, the guard never fires, and under `-Deo.autoFix`
the canonical text (now carrying a literal `[]`) is written over the file. The
error never reaches the user and a source the parser rejected becomes a
different source it accepts.

## Fix

Widen the guard so a recovered placeholder also counts as loss: alongside
`truncated`, reject when the parsed tree contains an `<o>` with neither
`@base` nor `@name`, no child elements and no text — which the printer can
only render as `[]`. The text/child-element checks keep data leaves (e.g.
`<o ...>3F-F0-...</o>`) and named/anonymous non-empty formations from being
mistaken for placeholders.

## Tests

- `failsWhenErrorRecoveredWithPlaceholder` — check mode fails with
  `does not fully parse` on the reversed-dispatch source.
- `doesNotOverwritePlaceholderRecoveryWhenAutoFixIsOn` — `-Deo.autoFix` no
  longer silently rewrites such a source.

All `MjFormatTest` cases pass and `qulice` is clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_019YNtyKFcjn2dxwS8pzjdTP)_

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
@@ -194,6 +194,13 @@ public final class MjFormat extends MjSafe {
      * never parsed at all, and this goal would otherwise print back (or, in
      * {@code -Deo.autoFix} mode, save) a truncated version of it.</p>
      *
+     * <p>Line coverage is necessary but not sufficient: the parser recovers
+     * from some <em>structural</em> errors not by dropping the remaining
+     * lines but by substituting an empty placeholder {@code <o>} for the
+     * piece it could not build and then carrying on. Coverage stays intact,
+     * yet the tree no longer describes the source, so this also rejects a
+     * tree carrying such a {@link #placeholder(XML) placeholder}.</p>
+     *
      * @param source The path of the {@code .eo} file, for the error message
      * @param structure The current text of the {@code .eo} file
      * @return The parsed XMIR
@@ -207,7 +214,8 @@ public final class MjFormat extends MjSafe {
             .elements(Filter.withName("error"))
             .filter(MjFormat::severe)
             .count();
-        if (errors > 0L && MjFormat.truncated(xmir, structure)) {
+        if (errors > 0L
+            && (MjFormat.truncated(xmir, structure) || MjFormat.placeholder(xmir))) {
             throw new IllegalStateException(
                 String.format(
                     "%s does not fully parse (%d error(s) found) and part of it was lost, won't format it",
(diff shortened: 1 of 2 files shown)


yegor256:
The `mvn (windows-2022, 26)` check failed, but the failure is unrelated to this PR and I'm not fixing it here.

The single error is in `SavedTest.exposesNoPartiallyWrittenFileToConcurrentReader`:

```
java.io.UncheckedIOException: java.io.IOException: java.util.concurrent.ExecutionException:
  java.lang.IllegalStateException: Failed to process "aaaa…" (java.lang.String)
    at org.eolang.maven.Threaded.total(Threaded.java:90)
    at org.eolang.maven.SavedTest.exposesNoPartiallyWrittenFileToConcurrentReader(SavedTest.java:60)
```

That test races 8 threads rewriting the same 500 KB file through `Saved` (atomic temp-file + rename, added in #5873) while a reader holds it open. On Windows, replacing a file another handle has open raises a sharing violation, so a writer's rename fails — a Windows-specific limitation of the concurrent-write path, not a parse/format issue.

This PR touches only `MjFormat.java` and `MjFormatTest.java`; it has nothing to do with `Saved`, `Threaded`, or file concurrency. The same commit passed `mvn` on `ubuntu-24.04`, `macos-15`, and `macos-15-intel`, plus `qulice` and every linter — only the Windows runner is red, on a pre-existing flaky concurrency test. A re-run of that job should clear it (I don't have permission to trigger one from here).

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (102 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2809](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6064, opened 2026-07-28
`string.low-cased` and `string.up-cased` only transform ASCII bytes, leaving accented and Cyrillic letters unchanged


gemshrine:
## What happens

`string.low-cased` and `string.up-cased` inspect individual UTF-8 bytes and only remap the ASCII ranges `A..Z` and `a..z`. Valid non-ASCII letters are left unchanged even though `string` explicitly describes its text operations as Unicode-aware.

### Runtime evidence

| expression | actual result | expected result |
| --- | --- | --- |
| `"ÉCOLE".low-cased` | **`"École"`** | `"école"` |
| `"café".up-cased` | **`"CAFé"`** | `"CAFÉ"` |
| `"ПрИвЕт".low-cased` | **`"ПрИвЕт"`** | `"привет"` |
| `"ASCII".low-cased` | `"ascii"` | `"ascii"` |
| `"ASCII".up-cased` | `"ASCII"` | `"ASCII"` |

The ASCII rows are controls. The other rows show that only ASCII letters change; accent and Cyrillic letters survive exactly as their original bytes.

## Why this is inconsistent

The package documentation for [`string`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string.eo#L1-L10) says that it stores UTF-8 text and provides Unicode-aware text manipulation. `length`, `at`, and `slice` already walk characters rather than bytes. In contrast, both casing implementations reduce the raw byte sequence and compare each byte with ASCII constants; see [`low-cased.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/low-cased.eo) and [`up-cased.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/up-cased.eo).

The limitation is also undocumented. When an operation is intentionally English-only, such as `string.is-alpha`, its documentation says so explicitly. The casing methods merely promise to return text in lower or upper case.

## Expected behavior

Case conversion should operate on Unicode characters, not UTF-8 bytes. At a minimum, ordinary one-to-one Unicode case pairs such as `É`/`é` and Cyrillic letters must be converted. The implementation should define locale-independent rules so its result does not depend on the host locale; more complex Unicode mappings can then be documented deliberately rather than being silently skipped.

## Regression coverage

[...]

PR #6074, opened 2026-07-28
#6064: move casing to root and Turkish EO locales

gemshrine:
Related to #6064.

## What this PR changes

Following the API direction proposed by @yegor256, casing is now available only through locale objects:

```eo
locale.low-cased "HeLlO"
locale.up-cased "HeLlO"
locale.turkish.low-cased "IİSTANBUL"
locale.turkish.up-cased "istanbul"
```

The former public `string.low-cased` and `string.up-cased` objects are removed, avoiding ambiguity about which locale controls their result.

## Implementation

- [locale.eo](https://github.com/gemshrine/eo/blob/6064-unicode-casing/eo-runtime/src/main/eo/locale.eo#L11-L58) owns the root one-to-one ASCII case mappings. It contains the migrated EO implementation rather than delegating to `string` objects.
- [locale/turkish.eo](https://github.com/gemshrine/eo/blob/6064-unicode-casing/eo-runtime/src/main/eo/locale/turkish.eo#L10-L26) specializes that root locale with the Turkish dotted/dotless-I mappings: `I → ı`, `İ → i`, `i → İ`, and `ı → I`.
- The old string-level casing scenarios were removed with the two deleted API objects; [locale regressions](https://github.com/gemshrine/eo/blob/6064-unicode-casing/eo-runtime/src/main/eo/locale.eo#L60-L74) now cover root casing, while [Turkish regressions](https://github.com/gemshrine/eo/blob/6064-unicode-casing/eo-runtime/src/main/eo/locale/turkish.eo#L28-L34) cover both locale-specific directions.

## Scope

The root locale preserves the former one-to-one ASCII behavior. This PR establishes the EO-only locale API and Turkish specialization without adding runtime atoms or claiming a complete Unicode mapping table.

## Verification

- `MAVEN_OPTS="-XX:ActiveProcessorCount=1 -Xmx1g" mvn -q -ntp -T 1 -pl eo-runtime -DskipITs -Deo.autoFix -Djunit.jupiter.execution.parallel.enabled=false -DforkCount=1 -Dtest=EOlocaleTest,EO_locale.EOturkishTest test`
- `git diff --check`

## Breaking change

This removes the published `string.low-cased` and `string.up-cased` objects, so `"HELLO".low-cased` and `"hello".up-cased` stop resolving after this PR. The replacement is not a rename but a different call shape:

- `"HELLO".low-cased` -> `locale.low-cased "HELLO"`
- `"hello".up-cased` -> `locale.up-cased "hello"`

`string.eo`'s header now points readers at `locale` so the entry point is discoverable. This concerns only the ASCII entry points; #6064's follow-ups about real Unicode casing rules are unaffected.


yegor256:
@gemshrine this is a wrong move. We don't want more atoms in the eo-runtime, we only want fewer of them. We should implement `locale` as EO object, not in Java.

gemshrine:
@yegor256 ok, will redo it in EO without the atoms.

My idea is to keep the case mapping as an EO locale object, basically a small table of ranges (A-Z, Latin-1, Cyrillic А-Я + Ё, Greek) with their offsets, plus a few exceptions. Then low-cased/up-cased just walk the text char by char and remap each code point through it. No full Unicode table, no new atoms.

One thing though: a table like this only does 1-to-1 pairs. It can't do ß → SS or the Greek final sigma. Fine to leave those out and document them, or do you want locale to handle them too?

yegor256:
@gemshrine I'm not an expert in locales. How do they work in Java? Let's implement them symmetrically in EO.

gemshrine:
@yegor256 I dug into how Java handles this, sharing what I found in case it helps us pick a direction.

Turns out Java does casing on two levels. The low level is `Character.toUpperCase(codepoint)`, which doesn't care about locale at all and just maps one character to one other character through a big built-in table.  On top of that sits `String.toUpperCase(Locale)`, which handles the tricky stuff the simple table can't: things like `ß` turning into `SS` (one letter becomes two), the Greek `Σ` that becomes a different lowercase form at the end of a word, and a couple of language-specific quirks (Turkish dotted/dotless `i`, Lithuanian). The interesting part is that the locale actually matters only for those few languages. Everything else just uses the same default table.

So if we mirror that in EO, I'd make a `locale` object with a `root` default that knows how to upper/lower a single character, and the string methods would walk the text character by character through it. The special locales like Turkish would just be `root` with a handful of characters swapped out. That keeps it symmetric with Java without adding any atoms.

One thing I want to be upfront about: Java's table covers the entire Unicode range, thousands of characters compiled into the JDK, and we can't realistically carry all of that as EO data. What I can do is cover the common alphabets by ranges (ASCII, Latin-1, Cyrillic, Greek) plus a small list of exceptions, and match Java exactly inside that. Anything outside just stays as-is.

Just two questions from me:
1. Is that "common alphabets, one-to-one" scope good enough for now, or do you want the harder cases (`ß` to `SS`, the Greek final sigma) handled too?
2. Do you want more than one locale, like a `tr` on top of `root`, or is `root` alone fine for the first version?

yegor256:
@gemshrine `root` means `English`, right?

gemshrine:
@yegor256 yeah, for what we're covering they're the same thing. English doesn't have any special casing rules, so it just uses that plain default table, and that's exactly what root is. The only real difference is for a handful of languages that override the default, such as Turkish, Azeri (the dotted/dotless i) and Lithuanian. Those are the ones where root and the language-specific locale actually diverge.

yegor256:
@gemshrine I suggest you implement `locale.eo` object and then `locale/turkish.eo` and we see and decide.

gemshrine:
@yegor256 i will implement `locale.eo` as the default/root locale and `locale/turkish.eo` as its specialized variant.

eo-runtime/src/main/eo/locale/turkish.eo
```diff
@@ -0,0 +1,28 @@
...
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package locale
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+locale > turkish
+  locale.lower-map.with "I" "ı" > root-lower
```
yegor256: @gemshrine take a look: https://github.com/objectionary/eo/issues/6145

eo-runtime/src/main/eo/locale.eo
```diff
@@ -0,0 +1,22 @@
...
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
++unlint mandatory-package
+
+[] > locale
+  string.low-cased text > [text] > low-cased
```
yegor256: @gemshrine maybe we should remove `string.low-cased` and `string.up-cased` entirely and only let users use the `locale`? otherwise, there is an obvious ambiguity

gemshrine:
@yegor256 the PR now implements the root `locale` and `locale.turkish` entirely in EO, without adding runtime atoms. Both expose the same `low-cased` and `up-cased` interface; Turkish overrides the dotted/dotless `i` rules.

Could you please take a look and let me know whether this structure is the direction you had in mind before I extend it further?

gemshrine:
@yegor256 okay, that makes sense. I’ll rework the PR so casing is available only through `locale`.

gemshrine:
@yegor256 i reworked this PR so that casing is exposed only through EO locale objects. Root `locale` now owns the casing implementation, while `locale.turkish` adds the dotted/dotless-I mappings. The former `string.low-cased` and `string.up-cased` objects have been removed.

The only failed CI job is Ubuntu Maven; it stopped in `eo-maven-plugin` before `eo-runtime` was built because remote Objectionary lookup tests received no expected data. I reported that separately in #6168.

Favixx (commented):
**Nothing in this PR has been executed yet, and that is not your fault.** The red CI is the objectionary.com outage of the evening of the 30th: the five failures are `MjPullTest.pullsFromProbes` and four `OyRemoteTest.checksPresenceOf*`, and the `master` build started at 22:48 failed with the same set plus three more. Builds from 23:03 are green again.

The consequence matters more here than on a Java PR. The reactor is `eo` → `eo-parser` → `eo-printer` → `eo-maven-plugin` → …, and it died in module **4 of 6**, so `eo-runtime` never compiled. Every line of this PR is in `eo-runtime`, and its entire verification is the six EO regressions inside `locale.eo` and `locale/turkish.eo`. None of them has run once, on any platform. A re-run is the first thing this needs.

What I could check by reading, so it is not left implicit:

- The port is faithful. `locale.low-cased` and `locale.up-cased` are the deleted files line for line, with `as-ascii` qualified to `string.as-ascii` now that they are no longer inside the `string` package.
- The Turkish composition is semantically right. Replacement happens on the text, `I` (U+0049) and `İ` (U+0130) are distinct code points so neither regex touches the other's target, `string.replaced` replaces *all* matches (its own doc and its recursive `rec-replaced` say so), and both directions round-trip on the four mapped characters.
- Nothing else in the repository referenced `string.low-cased` or `string.up-cased` — the only hits were the deleted files and the two tests you removed from `string.eo` — so the deletion breaks no internal caller.

The comments below are about what the tests and the docs do *not* pin.

eo-runtime/src/main/eo/locale.eo
```diff
@@ -0,0 +1,74 @@
...
+
++alias tuple.reduced
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
++unlint mandatory-package
+
+[] > locale
```
Favixx: Both objects lost their documentation in the move. `string/low-cased.eo` opened with `# Returns the `text` in lower case.` and `string/up-cased.eo` with its counterpart — those were the file comments, and files are what carried them. Now they are inner objects of `locale`, the file comment describes only the container ("Root locale with one-to-one ASCII casing rules"), and `locale.low-cased` itself says nothing.

That is a real loss for the object a user actually calls, and it is the moment to say the thing that is now true and was not before: this is the *root* locale, it maps only `A-Z`/`a-z`, and everything else passes through untouched. `directory.eo` and `file.eo` both comment their inner objects, so there is precedent right next door.

eo-runtime/src/main/eo/locale.eo
```diff
@@ -0,0 +1,74 @@
...
+  [text] > low-cased
```
Favixx: The two objects disagree about caching their constants, and now that they sit in one file the difference is hard to unsee.

`up-cased` binds `string.as-ascii "a" >> mincode!` and marks `(string.as-ascii "z")!` const, so its bounds are computed once. `low-cased` here has no `!` anywhere: `string.as-ascii "Z"`, `string.as-ascii "A"` in the guard, and `string.as-ascii "a"` plus another `string.as-ascii "A"` in the arithmetic are all inside the reduce lambda, so they are re-evaluated for **every byte** of the input, four times per byte.

Both files had this asymmetry before, where it was invisible. Merging them into one object also makes the fix available: `string.as-ascii "A" >> acode!` and friends at the `locale` level are shared by both objects, and `low-cased` stops re-deriving the same four numbers for every character it walks.

eo-runtime/src/main/eo/locale.eo
```diff
@@ -0,0 +1,74 @@
...
+                  as-i64.
+                    code.plus
+                      minus.
+                        string.as-ascii "a"
+                        string.as-ascii "A"
+                7
+                1
+              byte
+          string.as-ascii byte > code
+  [text] > up-cased
```
Favixx: Why does the offset here need `number mincode` while the twin expression in `low-cased` writes `minus. (string.as-ascii "a") (string.as-ascii "A")` with no cast?

Both compute the same 32, from the same two `as-ascii` calls; only this one wraps its left operand. If the cast is load-bearing — because `mincode` is a handle rather than a direct application, say — then it is exactly the kind of thing that deserves the one word of explanation it does not have. If it is not, it is noise inherited from a file nobody was comparing against its sibling.

eo-runtime/src/main/eo/locale.eo
```diff
@@ -0,0 +1,74 @@
...
+    string > @
+      tuple.reduced
+        text.as-bytes.array
+        --
+        [accum byte] >>
+          accum.concat > @
+            if.
+              and.
+                code.lte
+                  string.as-ascii "Z"
```
Favixx: The reduce walks `text.as-bytes.array`, i.e. bytes, not characters, and the whole correctness of that for non-ASCII input rests on an unstated fact: UTF-8 lead bytes are `C2`–`F4` and continuation bytes `80`–`BF`, neither of which can land inside `41`–`5A`, so a multi-byte character can never have one of its bytes case-shifted into a different character.

I convinced myself it holds, but nothing in the file says it and no test covers it — and "one-to-one ASCII casing rules" in the header reads as a description of intent rather than a guarantee about what happens to everything else. `locale.low-cased "ÄÖÜ"` eq `"ÄÖÜ"` is one line and turns the argument above into a regression test. It is also the property `turkish` silently depends on, since `ı` and `İ` survive the ASCII pass only because of it.

eo-runtime/src/main/eo/string/english.eo
```diff
@@ -0,0 +1,74 @@
...
+                      minus.
+                        number mincode
+                        string.as-ascii "A"
+                7
+                1
+              byte
+          string.as-ascii byte > code
+    string.as-ascii "a" >> mincode!
+
+  eq. ++> can-convert-already-lower-cased-text
```
Favixx, line 75: The inputs got swapped against their names during the move. In `string/low-cased.eo` it was `can-convert-already-lower-cased-text` ← `"HeLlO"` and `can-convert-text-to-lower-case` ← `"HELLO"`; here it is `already-lower-cased` ← `"HELLO"` and `text-to-lower-case` ← `"HeLlO"`. Same for the two `up-cased` tests below.

Nothing breaks, because both inputs produce the same expectation — which is the point: the names carry no information either way, so the swap is invisible and will stay in the file forever. Worth restoring the original pairing so that `git log --follow` on these tests shows a move and not a move-plus-edit.

eo-runtime/src/main/eo/string/english.eo
```diff
@@ -0,0 +1,74 @@
...
+    "hello"
+    locale.low-cased "HELLO"
+
+  eq. ++> can-convert-already-upper-cased-text
```
Favixx, line 80: Neither `can-convert-already-lower-cased-text` nor `can-convert-already-upper-cased-text` tests an input that is already in the target case — `"HELLO"` and `"hello"` are both fully in the *opposite* case, and `"HeLlO"` is mixed. The identity cases, `locale.low-cased "hello"` eq `"hello"` and `locale.up-cased "HELLO"` eq `"HELLO"`, are the ones the names promise and the only ones that exercise the `else` branch of the guard for letters, where the byte must be passed through unchanged.

Everything currently tested hits the branch that transforms. Two more `eq.` blocks cover the branch that must not.

eo-runtime/src/main/eo/locale/turkish.eo
```diff
@@ -0,0 +1,34 @@
+# Turkish locale with dotted and dotless I case mappings.
```
Favixx: "Turkish locale with dotted and dotless I case mappings" names the topic but not the mapping, and the mapping is the whole content of this file. Your PR description states it precisely — `I → ı`, `İ → i`, `i → İ`, `ı → I` — and that belongs here, where a reader can check the four regex literals below against the specification instead of reconstructing it from them.

It is also the only place the *asymmetry* can be recorded: `low-cased` handles two of the four and `up-cased` the other two, which is correct but looks like an omission until you work out why.

eo-runtime/src/main/eo/locale/turkish.eo
```diff
@@ -0,0 +1,34 @@
...
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package locale
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[] > turkish
+  locale.low-cased > [text] > low-cased
```
Favixx: @yegor256's pointer to #6145 ("let's prohibit global binding in an indented line") on this file does not appear to have been answered, and it is worth a reply in the thread even if the answer is "this is fine".

For what it is worth, I read #6145 as not biting here: it targets `>` on a line indented under an owner that is *not* a formation, and both `[] > turkish` and `[text] > low-cased` are formations, so the indented argument block below carries no `>` at all. But the construct on this line — an application, a formation and a name on one line — is the densest thing in the PR, and if #6145 lands with a broader reading it is the first line that will need rewriting.

eo-runtime/src/main/eo/locale/turkish.eo
```diff
@@ -0,0 +1,34 @@
...
+    string.replaced
+      string.replaced
+        text
```
Favixx: The order here is load-bearing and undocumented: the replacements must run **before** `locale.low-cased`, never after. Once the ASCII pass has run, `I` has already become `i` and the information needed to choose `ı` is gone — the Turkish distinction is destroyed by the very step it is composed with.

The code expresses this only through nesting, which reads as an arbitrary choice. One comment line (`# the replacements must precede the ASCII pass, which would otherwise fold I into i`) makes it a decision instead, and stops a future refactor from "simplifying" the composition into `locale.low-cased` first.

eo-runtime/src/main/eo/locale/turkish.eo
```diff
@@ -0,0 +1,34 @@
...
+        string.regex "/I/"
+        "ı"
+      string.regex "/İ/"
+      "i"
+  locale.up-cased > [text] > up-cased
```
Favixx: The property that makes a pair of casing objects a *locale* rather than two independent functions is that they round-trip on the characters they exist for, and nothing tests it. `turkish.up-cased (turkish.low-cased "İSTANBUL")` should be `"İSTANBUL"`, and `turkish.low-cased (turkish.up-cased "ıstanbul")` should be `"ıstanbul"`.

I traced both by hand and they hold — which is exactly why they are worth pinning: the four mappings form two independent cycles, and it would take only one of the four regex literals being wrong for the round trip to break while both existing tests stay green.

eo-runtime/src/main/eo/locale/turkish.eo
```diff
@@ -0,0 +1,34 @@
...
+    string.replaced
+      string.replaced
+        text
+        string.regex "/i/"
+        "İ"
+      string.regex "/ı/"
+      "I"
+
+  eq. ++> can-lower-case-dotted-and-dotless-i
```
Favixx: Both inputs contain exactly one occurrence of each character being mapped: `"IİSTANBUL"` has one `I` and one `İ`, `"istanbul"` has one `i` and no `ı`. So neither test can tell a replace-*all* implementation from a replace-*first* one.

It does replace all — `string/replaced.eo` says "where all regex target changed" and recurses through `rec-replaced` — so this is not a bug today. But `replaced` is a dependency of this file, not part of it, and `turkish.low-cased "IIII"` eq `"ıııı"` is one line that would catch that dependency changing under you.

eo-runtime/src/main/eo/string.eo
```diff
@@ -77,8 +77,6 @@
 
   "e\ne\ne".eq 65-0A-65-0A-65 ++> can-hold-a-three-line-text-block
 
-  "HELLO".low-cased.eq "hello" ++> can-lower-case-text
```
Favixx, line 80: With these two tests gone, `string.eo` no longer mentions casing anywhere — which is correct, since the objects moved, but it also means a reader who goes looking for `low-cased` in the obvious place finds nothing and no hint of where it went.

The usual repository answer to that is documentation rather than a dangling object, and `string.eo`'s own header comment is a natural home for a clause pointing at `locale`. Anyone who has written `"X".low-cased` in the last release will arrive here first.

eo-runtime/src/main/eo/string/low-cased.eo
```diff
@@ -1,42 +0,0 @@
...
-
-+alias tuple.reduced
-+architect yegor256@gmail.com
-+home https://github.com/objectionary/eo
-+package string
-+version 0.0.0
-+spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
-+spdx SPDX-License-Identifier: MIT
-
-[text] > low-cased
```
Favixx, line 10: This is published API being deleted, not internal restructuring: `string.low-cased` ships in released `eo-runtime` artifacts, and after this PR `"HELLO".low-cased` stops resolving for every user who wrote it. @yegor256 asked for the removal, so the decision is settled — but the PR carries nothing that would make it visible downstream.

A breaking-change label, and a line in the description naming the replacement (`string.low-cased` → `locale.low-cased`, which is a different call shape, not a rename), is what turns this from a surprise at upgrade time into a migration note. It also matters for #6064's own follow-ups, since the issue is about adding Unicode casing rather than removing the ASCII entry points.

eo-runtime/src/main/eo/locale/turkish.eo
gemshrine: I read 6145 as not applying to this line. The rule prohibits a global binding on a one-indent line only when the owner object is not a formation. Here the owner is turkish, which is a formation ([] > turkish), and low-cased and up-cased are simply its bound attributes, each a formation itself ([text] > ...). There is no non-formation owner for the binding to be reinterpreted under, so the meaning stays unambiguous. If 6145 later lands with a broader reading that also covers formation owners, this is the first line I would rewrite, and I will follow up then.

gemshrine:
Rebased onto master, so the earlier conflict on the two deleted string files is resolved and the PR is mergeable again. The mvn jobs are re-running on this push (the previous red was the objectionary.com outage of the 30th, not this PR).

@Favixx thanks for the read. Addressed the points:

- Restored and expanded the documentation the move had dropped. locale.eo now states the root/English semantics, the A-Z and a-z one-to-one mapping, and that everything else, including accented Latin, Cyrillic, Greek and all multi-byte UTF-8, passes through unchanged. locale/turkish.eo now spells out the four-character mapping table, the two-of-four asymmetry between low-cased and up-cased, and why the Turkish replacements must run before the root ASCII pass. string.eo's header points readers at locale so the entry point stays discoverable.
- Added tests: identity cases that exercise the pass-through branch for letters, non-ASCII pass-through for both directions, a Turkish replace-all case, and both Turkish round trips.
- Made low-cased symmetric with up-cased by hoisting its lower letter bound into a mincode constant, so it no longer re-derives the bound for every byte, which also removes the cast asymmetry you flagged.
- Restored the original test name to input pairings so git log --follow shows a clean move.
- Added a breaking-change note to the description with the string.low-cased to locale.low-cased migration (a different call shape, not a rename).

On the #6145 pointer: replied in that thread. Short version, it does not bite here because the owner is the turkish formation, so the indented binding is unambiguous.

One honest note on pr-size: it is red because these requested docs and tests take the diff to 242 lines, over the 200 limit. The change is atomic, one root locale plus its Turkish specialization and their tests, so it cannot be split without leaving half of it unbuilt. I kept the content the review asked for rather than dropping it to satisfy the gate. Happy to trim back toward green if you would prefer that.

eo-runtime/src/main/eo/string.eo
```diff
@@ -10,6 +10,9 @@
 # "text".length                  # Character count (not bytes)
 # "text".slice 0 2               # Character-based slicing
 # ```
+# Case conversion no longer lives here: use `locale.low-cased` and
```
morphqdd: string.low-cased/string.up-cased are removed outright rather than kept as thin deprecated forwards to locale.low-cased/up-cased. Any external .eo program built against the previous eo-runtime release that calls the old names breaks on upgrade with no transition path. Worth confirming this is an accepted breaking change for this version line.

morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +12 points.  Your running score is [+2807](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:
@gemshrine I'm a bit concerned about the usage pattern we offer to our clients. How about this instead:

```
"İyi akşamlar".turkish.upper
"Good evening".lower
```

Also, we can do:

```
"İyi akşamlar".turkish.length
```

To make it work, we define:

```
+package string
[] > lower
  lower. > @
    locale.english ^
```

Then, we create `string/turkish.eo` and `string/english.eo` files:

```
+package string
[s] > turkish
  s > @
  [] > lower
    ...  # convert s to lower case
```

Maybe we don't need `locale.eo` file in the root package (it would be an empty object).

gemshrine:
@yegor256 done. Reworked so case conversion is a method on the string, exactly as you suggested. "Good evening".lower and "hello".upper now work directly, and the locale variants are decorators that behave like the string, so both "İyi akşamlar".turkish.upper and "İyi akşamlar".turkish.length work (turkish wraps the string, so length and the other string methods pass through). Everything lives in the string package now: string.english holds the default casing (root means English) and string.turkish specializes it with the dotted and dotless I. I dropped the root locale.eo object, since as you expected it would have been empty.

One scope decision I want to flag. The casing folds the basic ASCII letters in both cases plus the Turkish dotted and dotless I; every other letter outside ASCII (ş, ç, ğ, ö, ü and the accented Latin, Cyrillic and Greek ranges) passes through unchanged. So "İyi akşamlar".turkish.upper returns "İYİ AKşAMLAR", where the ş stays lowercase. I scoped it to the dotted and dotless I on purpose, since that is the hazard #6064 is really about, and full Unicode case tables in EO would be a much larger object. If you would rather cover the whole Turkish alphabet (ş to Ş, ç to Ç, ğ to Ğ, ö to Ö, ü to Ü), I can extend it, just let me know.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (330 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1615](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6075, opened 2026-07-28
#6061: preserve empty byte sequences in signed shifts

### LINKED ISSUE #6061: `BytesRaw.sshift` results in `ArrayIndexOutOfBoundsException` for empty byte sequence
## What happens

Calling the public `Bytes.sshift(int)` operation on an empty byte sequence throws a raw `ArrayIndexOutOfBoundsException`. An empty `Bytes` value is valid: the `Bytes` interface represents arrays of arbitrary size, and the neighboring unsigned `shift` operation preserves an empty sequence.

### Runtime reproduction

```java
new BytesOf(new byte[0]).sshift(1).take();
// java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
```

These calls were executed against the current `master` runtime classes:

| call | result |
| --- | --- |
| `new BytesOf(new byte[0]).sshift(1).take()` | **`ArrayIndexOutOfBoundsException`** |
| `new BytesOf(new byte[0]).shift(1).take()` | `[]` |
| `new BytesOf(new byte[]{(byte) 0xff}).sshift(1).take()` | `[-1]` |

The second row is the empty-input control; the third confirms that sign extension still works for a non-empty negative value.

## Why it fails

`BytesRaw.sshift()` correctly obtains the shifted byte array first, but then unconditionally reads byte zero from a fresh copy of the original receiver to decide whether sign extension is needed:

```java
final byte[] bytes = this.shift(bits).take();
if (this.take()[0] < 0) {
    // sign extension
}
```

For a zero-length receiver, `this.take()[0]` is out of bounds before the sign-extension branch can be skipped. The failure escapes as a JDK implementation exception rather than the runtime's normal failure object. The code is in [`BytesRaw.sshift`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/BytesRaw.java#L85-L104).

## Expected behavior

A signed right shift of an empty byte sequence should return another empty byte sequence. There is no sign bit to extend, so this is the natural identity result and is consistent with `shift`.

## Regression test

```java
MatcherAssert.assertThat(
    new BytesOf(new byte[0]).sshift(1).take(),
    Matchers.equalTo(new byte[0])
);
```

Guarding the sign-extension branch with `bytes.length > 0` is sufficient while leaving all non-empty behavior unchanged.

gemshrine:
Fixes #6061.

## Problem

`BytesRaw.sshift()` first performs the right shift and then inspects the first byte of the original value to determine whether sign extension is required. For an empty byte sequence, that inspection was an unconditional `this.take()[0]` access, causing `ArrayIndexOutOfBoundsException`.

An empty `Bytes` value is valid, has no sign bit to extend, and the neighbouring unsigned `shift()` already preserves it. The raw JDK exception was therefore both inconsistent and unrelated to the public operation.

## Change

The sign-extension branch now runs only when the shifted sequence contains at least one byte. Non-empty values keep the existing path, including sign extension for a negative leading byte.

## Regression coverage

Added a test asserting that `new BytesOf(new byte[0]).sshift(1)` returns an empty array. It protects the empty-input case while the existing parameterized tests continue to cover positive and negative non-empty shifts.

## Verification

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/BytesRaw.java
+++ b/eo-runtime/src/main/java/org/eolang/BytesRaw.java
@@ -90,7 +90,7 @@ final class BytesRaw implements Bytes {
             );
         }
         final byte[] bytes = this.shift(bits).take();
-        if (this.take()[0] < 0) {
+        if (bytes.length > 0 && this.take()[0] < 0) {
             for (int index = 0; index < bytes.length; index += 1) {
                 final int zeros = BytesRaw.numberOfLeadingZeros(
                     bytes[index]
--- a/eo-runtime/src/test/java/org/eolang/BytesOfTest.java
+++ b/eo-runtime/src/test/java/org/eolang/BytesOfTest.java
@@ -156,6 +156,15 @@ final class BytesOfTest {
     }
 
     @Test
+    void shiftsEmptyBytes() {
+        MatcherAssert.assertThat(
+            "arithmetic shift of empty bytes should preserve them",
+            new BytesOf(new byte[0]).sshift(1).take(),
+            Matchers.equalTo(new byte[0])
+        );
+    }
+
+    @Test
     void shiftsIntegerMinValueWithoutOverflow() {
         MatcherAssert.assertThat(
             "shift(Integer.MIN_VALUE) should produce an all-zero result, but it didn't",


yegor256:
@gemshrine let's also add YAML test with EO code, to reproduce this bug

gemshrine:
@yegor256 i checked the EO surface before adding the YAML reproduction. `Bytes.sshift()` is not exposed to EO: the only public shift operation, `bytes.right`, delegates to `Bytes.shift()`, while `sshift()` is used only by the Java `Bytes` API and its Java tests. Could you please clarify whether you expect a new EO entry point for signed shifting, or should this PR/issue be withdrawn as an internal Java-only edge case?

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+997](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6076, opened 2026-07-28
#6062: measure regex match positions in Unicode characters

### LINKED ISSUE #6062: `string.regex` incorrectly measures string length with supplementary Unicode characters due to exposure of Java UTF-16 offsets
## What happens

`string.regex` exposes Java UTF-16 offsets as EO string indices. EO's string API counts Unicode characters, while Java's `Matcher` counts UTF-16 code units. The units agree for ASCII and BMP characters such as `é`, but diverge for a supplementary character such as `😀`, which occupies one EO character and two UTF-16 code units.

### Runtime evidence

The following calls were run against the current `master` runtime:

| pattern and text | `matches` | `from` | `to` | EO `text.length` |
| --- | ---: | ---: | ---: | ---: |
| `/abc/`, `"abc"` | `true` | 0 | 3 | 3 |
| `/éx/`, `"éx"` | `true` | 0 | 2 | 2 |
| `/😀x/`, `"😀x"` | **`false`** | 0 | **3** | **2** |
| `/./`, `"😀"` | **`false`** | 0 | **2** | **1** |

Both supplementary-character patterns match the complete Java string. The false result is produced only after the match, when offsets measured in UTF-16 units are compared with EO character counts.

## Why it fails

The regex atom passes EO's `start` directly to `Matcher.find(int)` and returns `Matcher.start()` and `Matcher.end()` unchanged:

```java
final boolean found = matcher.find(start);
...
result.put("from", new Data.ToPhi(matcher.start()));
result.put("to", new Data.ToPhi(matcher.end()));
```

[`matched-from-index`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/EO_string/EOregex%24EOpattern%24EOmatch%24EOmatched_from_index.java#L83-L125) therefore leaks UTF-16 offsets into EO. Then [`string.regex.matches`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/regex.eo#L50-L57) compares the leaked `to` value with the character-based `txt.length`. The same mismatch also makes a later `next` search start at the wrong Java offset after a supplementary character.

## Expected behavior

All public regex positions must use the same character indexing as `string.length`, `string.at`, and `string.slice`.

- Convert the EO `start` character index to a UTF-16 offset before `matcher.find(...)`.
- Convert `matcher.start()` and `matcher.end()` back to EO character positions before exposing `from` and `to`.
- Compare offsets only after both sides use the same unit.

After the conversion, `/😀x/` should match `"😀x"` with `from = 0` and `to = 2`, while `/./` should match `"😀"` with `from = 0` and `to = 1`.

## Regression coverage

Add a supplementary-character case beside the existing ASCII regex tests that asserts both the full-match predicate and the public bounds. A BMP control such as `/éx/` should remain unchanged, proving that the conversion only addresses surrogate-pair boundaries.

gemshrine:
Fixes #6062.

## Problem

EO string positions are Unicode-character positions, but Java `Matcher` reports and accepts UTF-16 code-unit offsets. The previous atom passed EO positions directly into `Matcher.find(int)` and exposed `Matcher.start()` and `Matcher.end()` unchanged.

The units agree for ASCII and BMP characters, but not for a supplementary character such as `😀`: it occupies one EO position and two UTF-16 code units. As a result, a complete regex match could fail its final length check and `next` could begin at the wrong place.

## Change

Before searching, the atom converts the public EO start position with `String.offsetByCodePoints`. After a match, it converts Java boundaries back with `String.codePointCount` before filling `from` and `to`. The bounds check now uses the same EO character unit.

## Regression coverage

EO tests assert that `/😀x/` fully matches `"😀x"`, and that successive `/./` matches report `0..1` for `😀` and `1..2` for `x`. This covers both public bounds and the start position passed by `next`.

## Verification

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/EO_string/EOregex$EOpattern$EOmatch$EOmatched_from_index.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/EOregex$EOpattern$EOmatch$EOmatched_from_index.java
@@ -84,17 +84,18 @@ public final class EOregex$EOpattern$EOmatch$EOmatched_from_index extends PhDefa
         final int start = new Expect.Natural(
             Expect.at(this, EOregex$EOpattern$EOmatch$EOmatched_from_index.START)
         ).it();
-        if (start > text.length()) {
+        final int length = text.codePointCount(0, text.length());
+        if (start > length) {
             throw new ExFailure(
                 "the 'start' attribute (%d) must be less than or equal to text length (%d)",
                 start,
-                text.length()
+                length
             );
         }
-        final boolean found = matcher.find(start);
+        final boolean found = matcher.find(text.offsetByCodePoints(0, start));
         final Phi result = match.take("matched");
         if (found) {
-            this.fill(result, matcher);
+            this.fill(result, matcher, text);
         } else {
             this.blank(result);
         }
@@ -105,8 +106,9 @@ public final class EOregex$EOpattern$EOmatch$EOmatched_from_index extends PhDefa
      * Fill the matched block with the data of a real match.
      * @param result The matched block to fill
      * @param matcher The matcher positioned on the found subsequence
+     * @param text Matched text
      */
-    private void fill(final Phi result, final Matcher matcher) {
+    private void fill(final Phi result, final Matcher matcher, final String text) {
         result.put(
             EOregex$EOpattern$EOmatch$EOmatched_from_index.POSITION,
             this.take(EOregex$EOpattern$EOmatch$EOmatched_from_index.POSITION)
@@ -115,8 +117,8 @@ public final class EOregex$EOpattern$EOmatch$EOmatched_from_index extends PhDefa
             EOregex$EOpattern$EOmatch$EOmatched_from_index.START,
             this.take(EOregex$EOpattern$EOmatch$EOmatched_from_index.START)
         );
-        result.put("from", new Data.ToPhi(matcher.start()));
-        result.put("to", new Data.ToPhi(matcher.end()));
+        result.put("from", new Data.ToPhi(text.codePointCount(0, matcher.start())));
+        result.put("to", new Data.ToPhi(text.codePointCount(0, matcher.end())));
         final Phi[] groups;
         if (matcher.groupCount() > 0) {
             groups = new Phi[matcher.groupCount() + 1];
--- a/eo-runtime/src/main/eo/string/regex.eo
+++ b/eo-runtime/src/main/eo/string/regex.eo
@@ -95,6 +95,21 @@
             regex "/a*/"
             "xa"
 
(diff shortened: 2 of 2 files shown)


gemshrine:
@yegor256 the mvn failures here look like unrelated flaky tests. Each rerun fails in a different module that this PR doesn't touch: eo-parser (parsesSimpleCode, 20s timeout) and eo-printer (XmirTest.printsToParseableEo[67]). The eo-runtime regex tests pass locally. Could you re-run the jobs when you get a chance?

yegor256:
@gemshrine every flaky test is a good opportunity for a new bug report

eo-runtime/src/main/java/org/eolang/EO_string/EOregex$EOpattern$EOmatch$EOmatched_from_index.java
```diff
@@ -115,8 +117,8 @@ private void fill(final Phi result, final Matcher matcher) {
             EOregex$EOpattern$EOmatch$EOmatched_from_index.START,
             this.take(EOregex$EOpattern$EOmatch$EOmatched_from_index.START)
         );
-        result.put("from", new Data.ToPhi(matcher.start()));
-        result.put("to", new Data.ToPhi(matcher.end()));
+        result.put("from", new Data.ToPhi(text.codePointCount(0, matcher.start())));
```
morphqdd, line 120: `codePointCount(0, ...)` re-scans from the start of the string on every call. For text with many matches this makes the overall match loop O(n·m) instead of O(n). Not wrong, just something to watch if regex ever runs over long strings with many matches.

yegor256:
@gemshrine conflicts here

gemshrine:
@yegor256 rebased onto master and resolved the conflict in regex.eo. It only overlapped with the merged #6207 in the test section, so I kept both tests. My character position change lives in the matched-from-index Java atom while the #6207 iteration logic is in the .eo, so they are independent and coexist cleanly. The conflict is gone now.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (36) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1627](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+3497](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6077, opened 2026-07-28
#6063: reject invalid partition counts before integrating

### LINKED ISSUE #6063: `number.integral` incorrectly integrates beyond upper bound for fractional `n` instead of failing
## What happens

`number.integral` documents `n` as “the number of partitions of the integration interval”, but accepts fractional, zero, and negative values. With a fractional `n`, the loop runs a whole number of times while the step is still divided by the fraction, so it evaluates a final subsection outside the requested interval.

### Runtime reproduction

For the identity function over `[0, 1]`:

```eo
number.integral
  x > [x]
  0
  1
  1.5
```

The runtime returns `0.8888888888888888`, although the integral of `x` over `[0, 1]` is `0.5`.

The same runtime probe with valid integer counts provides the controls:

| `n` | result |
| --- | --- |
| `1` | `0.5` |
| `2` | `0.5` |
| `1.5` | **`0.8888888888888888`** |
| `0` | **`0`** |
| `-1` | **`0`** |

For `n = 1.5`, the step is `2/3` and the loop executes for `i = 0` and `i = 1`, because both integers satisfy `i < 1.5`. It integrates `[0, 2/3]` and then `[2/3, 4/3]`; the latter interval crosses the declared upper bound. The resulting area is `8/9`, exactly the observed value. Zero and negative counts also silently return a numeric result despite describing no positive partition count.

## Why it fails

The count-driven loop introduced by #5806 has the correct structure for an integer `n`, but no precondition check:

```eo
while
  if. > [i] >>
    i.lt n
    ...
as-number. > step
  div.
    b.minus a
    n
```

[`number/integral.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number/integral.eo#L12-L42) therefore mixes an integer loop counter with an arbitrary floating-point divisor. A fractional partition count has no coherent interpretation, and `ceil(n)` iterations with step `(b-a)/n` cannot remain inside `[a,b]`.

## Expected behavior

`n` must be a finite positive integer. Any zero, negative, fractional, NaN, or infinite value should terminate with an explanatory failure instead of silently returning a partial integral or evaluating outside the declared bounds. Valid integer values must continue to perform exactly `n` subsections.

## Regression coverage

Keep the existing valid cases for `n = 1` and `n = 2`, and add termination tests for `0`, `-1`, and `1.5`. A direct result assertion for the fractional case is also useful as a guard against returning `8/9` again before validation is implemented.

gemshrine:
Fixes #6063.

## Problem

`number.integral` used `n` both as the floating-point divisor for the integration step and as the upper bound of an integer-counter loop. A fractional count has no coherent meaning in that arrangement: with `n = 1.5`, the loop performs two iterations with a step of `2/3`, so its second subsection extends beyond the requested upper bound.

Zero and negative values also returned a numeric result without describing any partitions, while `NaN` and infinities were not rejected explicitly.

## Change

The operation now validates `n` before calculating the step or entering the loop. Only finite, positive integers are accepted; every other value terminates with `the number of partitions must be a finite positive integer`. Valid counts retain the existing numerical algorithm.

## Regression coverage

The existing integral cases cover valid counts. New EO termination tests cover fractional, zero, negative, `NaN`, positive-infinite, and negative-infinite partition counts, each with an identity integration function.

## Verification

### DIFF
--- a/eo-runtime/src/main/eo/number/integral.eo
+++ b/eo-runtime/src/main/eo/number/integral.eo
@@ -10,28 +10,31 @@
 +spdx SPDX-License-Identifier: MIT
 
 [fun a b n] > integral
-  as-number. > @
-    malloc.of
-      8
-      [sum] >>
-        seq * > @
-          while
-            if. > [i] >>
-              i.lt n
-              seq *
-                sum.put
-                  sum.as-number.plus
-                    subsection
-                      a.plus (i.times step)
-                      a.plus ((i.plus 1).times step)
-                true
-              false
-            true > [i]
-          sum
-        as-number. > step
-          div.
-            b.minus a
-            n
+  valid.if > @
+    as-number.
+      malloc.of
+        8
+        [sum] >>
+          seq * > @
+            while
+              if. > [i] >>
+                i.lt n
+                seq *
+                  sum.put
+                    sum.as-number.plus
+                      subsection
+                        a.plus (i.times step)
+                        a.plus ((i.plus 1).times step)
+                  true
+                false
+              true > [i]
+            sum
+          as-number. > step
+            div.
+              b.minus a
+              n
+    trouble
+      "the number of partitions must be a finite positive integer"
   times. > [a b] >> subsection
     div.
       b.minus a
@@ -44,6 +47,9 @@
             0.5.times
               a.plus b
         fun b
+  n.is-finite.and > valid
+    n.is-integer.and
+      n.gt 0
 
   ++> tests-calculates-cube-integral
     close-to > @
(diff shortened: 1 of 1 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+985](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6081, opened 2026-07-29
#6068: Fix dirname for relative filenames

### LINKED ISSUE #6068: `path.dirname` incorrectly returns filename for non-separated relative paths instead of current-directory marker
## What happens

`path.dirname` returns a separator-free relative filename unchanged instead of returning the current-directory marker. The result is not a directory and cannot be used as the parent component of that basename.

### Runtime reproduction

| path | actual `dirname` | expected `dirname` |
| --- | --- | --- |
| `plain` | **`plain`** | **`.`** |
| `.gitignore` | **`.gitignore`** | **`.`** |
| `dir/file` | `dir` | `dir` |
| `/file` | `/` | `/` |

The last two rows are controls. For `plain`, treating the result as a parent produces `plain/plain`; `./plain` instead normalizes back to the original relative path. A directory-name operation must not return the leaf filename as its own directory.

## Why it fails

When no separator is found, `dirname` returns the original `uri`:

```eo
if.
  or.
    pth.size.eq 0
    len.eq -1
  uri
  ...
```

The `len.eq -1` branch in [`path.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/path.eo#L50-L64) conflates an empty path with a non-empty path that has no directory component. This is a different residual case from #5802, which corrected a separator at index zero for root-anchored paths.

## Expected behavior

For a non-empty relative path with no separator, `dirname` should return `.`. This matches the conventional meaning of a directory name and the runtime's own `path.normalized` behavior, which uses `.` to represent the current directory. The empty-path case may retain its present behavior.

## Regression test

```eo
".".eq
  path.posix "plain" .dirname
```

Add the hidden-file form as a second test and retain the existing `/foo → /` coverage from #5802, so the no-separator and root-separator cases stay distinct.

oozan:
## Summary

- return `.` from `path.dirname` for separator-free relative filenames
- handle both ordinary and hidden filenames
- preserve empty-path and root-directory behavior
- add regression tests for `plain` and `.gitignore`

## Root cause

The no-separator case was handled together with the empty-path case. Consequently, a filename could be returned as its own parent directory.

## Testing

- EO canonical formatter
- `mvn install -DskipTests --batch-mode`
- `mvn test -pl :eo-runtime -DskipITs --batch-mode`
- `git diff --check`

Closes #6068

AI assistance was used to investigate the issue and prepare the initial patch. I reviewed the implementation and validated it using the repository checks.

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -44,17 +44,18 @@
     [] > dirname
       string > @
         if.
-          or.
-            pth.size.eq 0
-            len.eq -1
+          pth.size.eq 0
           uri >> pth!
           if.
-            len.eq 0
-            separator
-            as-bytes.
-              text.slice
-                0
-                string.last-index-of text separator >> len!
+            len.eq -1
+            "."
+            if.
+              len.eq 0
+              separator
+              as-bytes.
+                text.slice
+                  0
+                  string.last-index-of text separator >> len!
       string pth > text
     sys.drive uri > drive
     if. > driveless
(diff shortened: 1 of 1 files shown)


0crat:
@oozan Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (29) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6085, opened 2026-07-29
Rename eo-runtime test objects for the new bad-test-name lint

### LINKED ISSUE #6083: eo-runtime test objects are named `tests-…`, which the new `bad-test-name` lint rejects
`org.eolang:lints:0.4.0` adds a `tests/bad-test-name` lint that requires the name of
every truthy test attribute to start with one of `can-`, `cannot-`, `accepts-`,
`rejects-` or `stops-on-`. Almost every test object in `eo-runtime` is named
`tests-<something>` instead — for instance `eo-runtime/src/main/eo/tuple/back.eo:24`
declares `eq. ++> tests-large-index-in-back`, and
`eo-runtime/src/main/eo/nop.eo:25` declares `nop "anything" ++> tests-nop-is-true`.
There are 1321 such declarations across 148 files, plus a stray `# ++> works-just-fine`
in the `nop` docblock.

This matters because the whole point of the new lint is that a test name should read as
a sentence about what the object can or cannot do, and `tests-nop-is-true` says nothing
about the object under test — it only restates the assertion. Bumping the `lints`
dependency (currently `0.3.0` in `eo-maven-plugin/pom.xml:34`) therefore turns the
existing names into 1320 warnings, which is enough noise to bury every real finding in
the `eo:lint` output.

The fix is to rename each test attribute to the new convention: a positive assertion
becomes `can-…` (`tests-large-index-in-back` → `can-take-a-back-slice-with-a-large-index`),
a negative one `cannot-…`, a fallback path `rejects-…`, and a `-->` throwing test
`stops-on-…`. The printer already sorts test attributes last regardless of their names
(`eo-printer/.../to-eo-tree.xsl:178` ranks `eo:test-attr()` after everything else), so
`eo:format -Deo.autoFix` re-sorts them alphabetically and nothing else has to move.

yegor256:
Bumps `org.eolang:lints` to 0.4.0 and renames all 1423 test objects in `eo-runtime`
to the convention its new `bad-test-name` lint requires: `can-` for a capability,
`cannot-` for its absence, `rejects-` for a fallback path, and `stops-on-` for a
throwing `-->` test. So `tests-large-index-in-back` becomes
`can-take-a-back-slice-with-a-large-index`, and `on-bytes-of-wrong-size-as-i32`
becomes `stops-on-bytes-of-wrong-size`. Three more warnings go away with it: `bool`
no longer renames `app` into `output`, `string.slice` binds `r3`/`r4` to their own
byte literals instead of aliasing the `p2`/`p3` masks, and `number.integral` inlines
the single-use `valid`.

Without this, the 0.4.0 bump alone turns `eo:lint` on `eo-runtime` into 1577
warnings, 1320 of them `bad-test-name`, which buries every real finding. It also
seemed worth doing name by name rather than by regex — a few of the old names were
plainly wrong about what they tested (`tests-exp-check-n2` is `exp 5`, not a
negative; `tests-malloc-is-strictly-sized-int` asserts that the block *keeps* its
size; `tests-switch-with-all-false-cases` asserts `switch` returns `true`). The big
diff is mostly `eo:format` re-sorting the attributes after the rename; I diffed every
file's non-declaration content against `master` to confirm nothing was lost. `mvn
test` passes with 1810 tests.

Two things for the reviewer. This supersedes #5836, which does the same version bump.
And 253 warnings survive, all of them the one false positive in
objectionary/lints#1129 — its fix is still sitting in objectionary/lints#1136, so
0.4.0 ships the old `starts-with(@name, '+')` predicates and every `-->` test still
gets linted as ordinary code. I checked all 253 by hand: the 209 `non-kebab-name` and
`sparse-decoration` ones all point at a `-->` line, and the 44 `wrong-test-order`
ones all point at the last `++>` test of a file whose next sibling is a `-->` test.
They should vanish on the next `lints` release.

Closes #6083

### DIFF
--- a/eo-maven-plugin/pom.xml
+++ b/eo-maven-plugin/pom.xml
@@ -31,7 +31,7 @@
     <dependency>
       <groupId>org.eolang</groupId>
       <artifactId>lints</artifactId>
-      <version>0.3.0</version>
+      <version>0.4.0</version>
       <exclusions>
         <exclusion>
           <groupId>org.codehaus.groovy</groupId>
--- a/eo-maven-plugin/pom.xml
+++ b/eo-maven-plugin/pom.xml
@@ -31,7 +31,7 @@
     <dependency>
       <groupId>org.eolang</groupId>
       <artifactId>lints</artifactId>
-      <version>0.4.0</version>
+      <version>0.4.1</version>
       <exclusions>
         <exclusion>
           <groupId>org.codehaus.groovy</groupId>
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjParseTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjParseTest.java
@@ -14,6 +14,8 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.attribute.FileTime;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Map;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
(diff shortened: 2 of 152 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (6811 >= 100); -8 for way too many hits-of-code (6811 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2905](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6086, opened 2026-07-29
#6057: reject non-integer tuple index through fallback

### LINKED ISSUE #6057: `tuple.at` bypasses fallback for fractional indices within numeric bounds
## What happens

`tuple.at` accepts a caller-provided fallback for an invalid index, but a fractional index inside the numeric bounds bypasses that fallback and eventually terminates in the empty tuple. A tuple position must identify an element, so only finite integer indices have a meaningful in-bounds interpretation.

### Runtime reproduction

```eo
at.
  * "zero" "one" "two"
  1.5
  "recovered" > [message]
```

The following cases were executed against the current `master` runtime:

| --- | --- | --- |
| `(* "zero" "one" "two").at 1.5 fallback` | **`ExFailure`** after recursion reaches `tuple.empty` | the provided `"recovered"` fallback |
| `(* "zero" "one" "two").at 9 fallback` | `"recovered"` | `"recovered"` |
| `(* "zero" "one" "two").at 1 fallback` | `"one"` | `"one"` |
| `(* "zero" "one" "two").at -1 fallback` | `"two"` | `"two"` |

The second through fourth rows are controls: ordinary out-of-range, integer, and supported negative indexing keep their existing behavior. Only the fractional value slips between the validation and traversal paths.

## Why it fails

The bounds guard rejects an index only when it is below zero or at least the tuple length. `1.5` passes both tests for a length-three tuple and reaches `at-fast`:

```eo
if. > [tup] >> at-fast
  eq.
    tup.length.plus -1
    index
  tup.head
  at-fast tup.tail
```

[`tuple.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/tuple.eo#L12-L31) compares an integer position with `1.5`. Equality can never hold, so recursion continues through every tail and finally triggers the empty tuple's unrelated “Can't get tail” failure. The caller's fallback is never reached.

## Expected behavior

Any non-integer or non-finite index should be rejected through the existing fallback, just like an out-of-range integer index. Valid non-negative and negative integer indices must remain unchanged.

## Regression test

Use a positive fallback assertion rather than only a termination test:

```eo
eq.
  "recovered"
  at.
    * "zero" "one" "two"
    1.5
    "recovered" > [message]
```

The index guard can check `index.is-integer.not` before entering `at-fast`; `is-integer` already rejects non-finite values.

gemshrine:
Fixes #6057.

## Problem

`tuple.at` routes an invalid index to the caller's `fallback`, but only when the index is below zero or at least the tuple length. A fractional index inside those bounds (for example `1.5` on a three-element tuple) passes both checks and reaches `at-fast`, which compares an integer position against `1.5`. Equality never holds, so the recursion walks off the end and terminates in the empty tuple's "Can't get tail" failure instead of returning the fallback.

## Change

`at-fast` now returns the `fallback` when its recursion reaches the empty tuple. A fractional or `nan` index never equals any integer position, so it walks the whole tuple and lands on the empty one, which now yields the fallback instead of failing.

I deliberately did not add an `is-integer` guard on the index. `at` must dataize its index exactly once (this is what `EOseqTest.tests-seq-single-dataization-int-equal-to-cache-problem-test` pins down, and tuple iteration in `while` relies on it too). Re-reading the index to test it for integrality breaks that contract. Detecting the overrun inside `at-fast` needs no second read of the index, so the single-dataization behaviour is preserved. `±inf` is still handled by the existing bounds check.

## Covered behavior

A new positive regression asserts that `(* "zero" "one" "two").at 1.5 "recovered"` returns `"recovered"`. The existing out-of-bounds, integer, and negative-index tests remain as controls.

## Verification

Ran the full `eo-runtime` test suite. `EOtupleTest` (30/30), `EOseqTest` (9/9) and `EOwhileTest` (10/10) all pass, confirming the fractional case is covered and the single-dataization contract still holds.

### DIFF
--- a/eo-runtime/src/main/eo/tuple.eo
+++ b/eo-runtime/src/main/eo/tuple.eo
@@ -23,11 +23,15 @@
       fallback
         "Given index is out of tuple bounds"
       if. > [tup] >> at-fast
-        eq.
-          tup.length.plus -1
-          index
-        tup.head
-        at-fast tup.tail
+        0.eq tup.length
+        fallback
+          "Given index is out of tuple bounds"
+        if.
+          eq.
+            tup.length.plus -1
+            index
+          tup.head
+          at-fast tup.tail
       | ^
   tuple > empty
     T
@@ -286,6 +290,16 @@
         3
         2
         1
+
+  eq. ++> rejects-a-fractional-index
+    "recovered"
+    at.
+      *
+        "zero"
+        "one"
+        "two"
+      1.5
+      "recovered" > [message]
 
   eq. ++> rejects-an-out-of-bounds-index
     "recovered"


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1069](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6087, opened 2026-07-29
#6058: route infinite repetition count to fallback

### LINKED ISSUE #6058: `string.repeated` fails to handle positive infinity as an invalid repetition count
## What happens

`string.repeated` promises to route a negative or non-whole repetition count to its `cant-repeat` fallback. Positive infinity passes the current validation, enters the recursive concatenation path, and never reaches its stopping value. The fallback is therefore ignored.

### Runtime reproduction

```eo
repeated
  "x"
  pinf
  "recovered" > [message]
```

| --- | --- | --- |
| `"x".repeated pinf fallback` | **`ExFailure`** after the recursive computation cannot reach its stop condition | `"recovered"` |
| `"x".repeated nan fallback` | `"recovered"` | `"recovered"` |
| `"x".repeated (-1) fallback` | `"recovered"` | `"recovered"` |
| `"x".repeated 2.5 fallback` | `"recovered"` | `"recovered"` |
| `"x".repeated 3` | `"xxx"` | `"xxx"` |

The finite invalid values and ordinary integer count are controls. `pinf` is the only invalid count in this set that slips through the guard.

## Why it fails

The validation checks negativity and equality with `floor`:

```eo
or.
  0.gt times
  not.
    times.floor.eq times
```

Positive infinity equals its own floor, so it reaches `rec-repeated`. That recursion starts at an integer index and stops only when `amount.eq index`; incrementing an integer by one can never equal infinity. The implementation is in [`string/repeated.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/repeated.eo#L12-L38).

## Expected behavior

The accepted domain is a finite non-negative integer. NaN, both infinities, negative values, and fractional values must all route to `cant-repeat` when it is supplied, or to the existing bottom object when it is not. Zero and positive integers must retain their current results.

## Regression test

Add a positive fallback test beside the existing fractional and negative cases:

```eo
eq.
  "recovered"
  repeated
    "x"
    pinf
    "recovered" > [message]
```

Adding `times.is-finite.not` to the invalid-count guard prevents the unbounded recursion without changing valid repetitions.

gemshrine:
Fixes #6058.

## Problem

`string.repeated` routes a negative or non-whole `times` to its `cant-repeat` fallback, but positive infinity slips through. The guard rejects a value only when `0.gt times` or `times.floor.eq times` is false. `pinf` equals its own floor, so it enters `rec-repeated`, whose counter starts at an integer and stops only on `amount.eq index`; incrementing an integer by one never reaches infinity, so the recursion never terminates and the fallback is ignored.

## Change

The integrality check `not. (times.floor.eq times)` is replaced with `times.is-integer.not`. `is-integer` is true only for a finite value equal to its own floor, so it rejects `nan`, `±inf`, and fractional counts in one condition, while `0.gt times` still rejects negatives. The accepted domain is now exactly a finite non-negative integer. Zero and positive integers are unchanged.

## Covered behavior

A new positive regression asserts that `"x".repeated pinf "recovered"` returns `"recovered"`. The existing negative, fractional, zero, and integer tests remain as controls.

## Verification

Ran `mvn -pl eo-runtime test -Dtest=EO_string.EOrepeatedTest` (7/7).

### DIFF
--- a/eo-runtime/src/main/eo/string/repeated.eo
+++ b/eo-runtime/src/main/eo/string/repeated.eo
@@ -1,5 +1,6 @@
 # Returns `text` repeated `times` times.
-# If `times` < 0 or not a whole number, the `cant-repeat` fallback is returned,
+# If `times` is not a finite non-negative integer (negative, fractional, nan
+# or infinite), the `cant-repeat` fallback is returned,
 # or the bottom object when unbound.
 # If `times` == 0, empty text is returned.
 
@@ -14,8 +15,7 @@
   if. > @
     or.
       0.gt times
-      not.
-        times.floor.eq times
+      times.is-integer.not
     cant-repeat
       "Can't repeat text %f times".printf
         * times
@@ -40,6 +40,13 @@
     repeated
       "hey"
       5
+
+  eq. ++> tests-text-repeated-infinite-times-returns-provided-fallback
+    "recovered"
+    repeated
+      "x"
+      pinf
+      "recovered" > [message]
 
   eq. ++> tests-text-repeated-negative-times-formats-message-as-number
     "Can't repeat text -1.000000 times"


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1009](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6088, opened 2026-07-29
#6059: route infinite places to fallback in as-fixed

### LINKED ISSUE #6059: `string.as-fixed` incorrectly handles infinite `places`, leading to `ExFailure`
## What happens

`string.as-fixed` routes negative and fractional precision values to its `cant-print` fallback, but accepts positive infinity as `places`. It then enters `pow10`, where subtracting one from infinity leaves infinity, so the base case is never reached and the fallback is ignored.

### Runtime reproduction

```eo
as-fixed
  1.25
  pinf
  "recovered" > [message]
```

| --- | --- | --- |
| `as-fixed 1.25 pinf fallback` | **`ExFailure`** after `pow10` cannot reach zero | `"recovered"` |
| `as-fixed 1.25 nan fallback` | `"recovered"` | `"recovered"` |
| `as-fixed 1.25 (-1) fallback` | `"recovered"` | `"recovered"` |
| `as-fixed 1.25 2.5 fallback` | `"recovered"` | `"recovered"` |
| `as-fixed 1.25 2` | `"1.25"` | `"1.25"` |

The finite invalid values and normal two-place formatting are controls. Positive infinity is the exceptional invalid value that passes the same validation.

## Why it fails

The guard accepts every value equal to its floor:

```eo
or.
  0.gt places
  not.
    places.floor.eq places
```

`pinf.floor.eq pinf` is true. The value reaches `pow10`, which recursively evaluates `p.minus 1` until `p.eq 0`; `pinf.minus 1` is still `pinf`. The relevant implementation is in [`string/as-fixed.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/as-fixed.eo#L14-L71).

## Expected behavior

`places` must be a finite non-negative integer. NaN, both infinities, negative values, and fractional values should all invoke `cant-print` when supplied. Valid values, including `0`, must continue to produce fixed-point text with exactly the requested number of fraction places.

## Regression test

```eo
eq.
  "recovered"
  as-fixed
    1.25
    pinf
    "recovered" > [message]
```

Using `places.is-integer.not` in the validation is sufficient because that predicate is already false for non-finite values. The existing negative and fractional tests then remain useful controls.

gemshrine:
Fixes #6059.

## Problem

`string.as-fixed` sends negative and fractional `places` to its `cant-print` fallback, but positive infinity passes the guard. `pinf` equals its own floor, so it reaches `pow10`, which recurses on `p.minus 1` until `p.eq 0`; `pinf.minus 1` stays `pinf`, so the base case is never reached and the fallback is ignored.

## Change

The integrality check `not. (places.floor.eq places)` is replaced with `places.is-integer.not`, which is true only for a finite value equal to its own floor. It therefore rejects `nan`, `±inf`, and fractional precision in one condition, while `0.gt places` still rejects negatives. The accepted domain is now exactly a finite non-negative integer. `0` and positive integers are unchanged.

## Covered behavior

A new positive regression asserts that `as-fixed 1.25 pinf "recovered"` returns `"recovered"`. The existing negative, fractional, zero-places, and formatting tests remain as controls.

## Verification

Ran `mvn -pl eo-runtime test -Dtest=EO_string.EOas_fixedTest` (16/16).

### DIFF
--- a/eo-runtime/src/main/eo/string/as-fixed.eo
+++ b/eo-runtime/src/main/eo/string/as-fixed.eo
@@ -1,8 +1,9 @@
 # Writes the number `n` as fixed-point decimal with exactly `places` fraction
 # digits, rounding to the nearest unit in the last place, so that `2.5` with
 # four places becomes `2.5000` and `0.9999999` with six places becomes `1.000000`.
-# If `places` is negative or not a whole number, the `cant-print` fallback is
-# returned, or the bottom object when unbound.
+# If `places` is not a finite non-negative integer (negative, fractional, nan
+# or infinite), the `cant-print` fallback is returned, or the bottom object
+# when unbound.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
@@ -15,8 +16,7 @@
   if. > @
     or.
       0.gt places
-      not.
-        places.floor.eq places
+      places.is-integer.not
     cant-print
       "Can't write a fixed-point decimal with %f fraction places".printf
         * places
@@ -89,6 +89,13 @@
     "3.14"
     string
       as-fixed 3.14159 2
+
+  eq. ++> tests-infinite-places-returns-provided-fallback
+    "recovered"
+    as-fixed
+      1.25
+      pinf
+      "recovered" > [message]
 
   eq. ++> tests-negative-places-formats-message-as-number
     "Can't write a fixed-point decimal with -2.000000 fraction places"


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (15) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1021](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6089, opened 2026-07-29
#6060: reject non-integer and non-finite nsplit limit

### LINKED ISSUE #6060: `string.nsplit` incorrectly accepts fractional and non-finite maximum part limits
## What happens

`string.nsplit` exposes `cant-split` for an invalid maximum number of parts, but its validation only rejects values below one. Fractional and non-finite limits are accepted and silently produce tuples whose size is determined by comparisons between an integer counter and an invalid numeric limit.

### Runtime evidence

The following calls were run against the current `master` runtime:

| --- | --- | --- |
| `nsplit "a-b-c" "-" 1.5 fallback` | **`* "a" "b-c"`** | `"recovered"` |
| `nsplit "a-b-c" "-" 2.5 fallback` | **`* "a" "b" "c"`** | `"recovered"` |
| `nsplit "a-b-c" "-" pinf fallback` | **`* "a" "b" "c"`** | `"recovered"` |
| `nsplit "a-b-c" "-" 2 fallback` | `* "a" "b-c"` | `* "a" "b-c"` |
| `nsplit "a-b-c" "-" 0 fallback` | `"recovered"` | `"recovered"` |

The valid limit and zero-limit fallback are controls. The highlighted cases demonstrate that invalid values do not merely fail with an unclear diagnostic—they return plausible but semantically accidental tuples.

For `limit = 1.5`, the integer split counter is compared with `limit - 1 = 0.5`; one delimiter is accepted before the comparison becomes false, producing two parts. For `limit = 2.5`, two delimiters are accepted. For `pinf`, the counter never reaches `limit - 1`, effectively turning a bounded split into an unlimited one.

## Why it fails

The guard only checks `1.gt limit`:

```eo
if. > @
  1.gt limit
  cant-split
  ...
```

Later, [`string/nsplit.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/nsplit.eo#L12-L67) compares the integer `count` with `limit.minus 1`. A maximum number of parts is inherently a finite positive integer; the current code has no check for that precondition.

## Expected behavior

Only finite integer limits of at least one should be accepted. Fractional values, NaN, positive infinity, negative infinity, zero, and negative values must all call `cant-split` when it is supplied. Existing integer behavior, including the rule that the final part contains the remaining text, must remain unchanged.

## Regression coverage

Add positive fallback assertions for `1.5`, `2.5`, and `pinf`, alongside the existing zero and negative tests. A robust guard can reject `limit.is-integer.not` in addition to `1.gt limit`; `is-integer` also excludes non-finite values. The error formatting should preserve the supplied invalid value instead of using an integer-only conversion for fractional or infinite limits.

gemshrine:
Fixes #6060.

## Problem

`string.nsplit` only rejects a `limit` below one. Fractional and non-finite limits pass the guard and reach `rec-nsplit`, where an integer counter is compared with `limit.minus 1`. For `1.5` the counter clears `0.5` once and yields two parts; for `2.5` it yields three; for `pinf` the counter never reaches the bound, turning a bounded split into an unlimited one. The `cant-split` fallback is never used and the caller gets a plausible but accidental tuple.

## Change

The guard now also rejects a non-integer limit: `limit.is-integer.not` is added next to `1.gt limit`. `is-integer` is false for `nan`, `±inf`, and fractional values, so the accepted domain is exactly a finite integer of at least one. Valid integer limits and the zero/negative fallback path are unchanged.

## Covered behavior

Two new positive regressions assert that `nsplit "a-b-c" "-" 1.5 "recovered"` and `nsplit "a-b-c" "-" pinf "recovered"` both return `"recovered"`. The existing zero-limit fallback and the integer-limit split tests remain as controls.

## Verification

Ran `mvn -pl eo-runtime test -Dtest=EO_string.EOnsplitTest` (11/11).

### DIFF
--- a/eo-runtime/src/main/eo/string/nsplit.eo
+++ b/eo-runtime/src/main/eo/string/nsplit.eo
@@ -1,6 +1,7 @@
 # Returns a `tuple` of `strings`: `text` is split by `delimiter` with a maximum of `limit` parts.
 # The last element contains the remainder of the string after splitting.
-# If `limit` < 1, the `cant-split` fallback is returned, or the bottom object when unbound.
+# If `limit` is not a finite integer of at least 1 (fractional, nan, infinite,
+# zero or negative), the `cant-split` fallback is returned, or the bottom object when unbound.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
@@ -11,7 +12,9 @@
 
 [text delimiter limit cant-split] > nsplit
   if. > @
-    1.gt limit
+    or.
+      1.gt limit
+      limit.is-integer.not
     cant-split
       "Invalid limit %d for nsplit, must be at least 1".printf
         * limit
@@ -70,6 +73,22 @@
           0
           0
   delimiter.size >> size!
+
+  eq. ++> tests-nsplit-fractional-limit-returns-provided-fallback
+    "recovered"
+    nsplit
+      "a-b-c"
+      "-"
+      1.5
+      "recovered" > [message]
+
+  eq. ++> tests-nsplit-infinite-limit-returns-provided-fallback
+    "recovered"
+    nsplit
+      "a-b-c"
+      "-"
+      pinf
+      "recovered" > [message]
 
   eq. ++> tests-nsplit-invalid-limit-returns-provided-fallback
     "fallback"


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (23) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1033](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6090, opened 2026-07-29
#6066: anchor scanf format to the start of input

### LINKED ISSUE #6066: `string.scanf` incorrectly parses later substrings when unmatched leading text is present
## What happens

`string.scanf` searches for its generated format pattern anywhere in the input instead of matching the format at the input start. It can therefore skip unmatched leading text and return values from a later substring.

### Runtime reproduction

| format | input | first parsed value |
| --- | --- | --- |
| `"id=%d"` | `"id=42"` | `42` |
| `"id=%d"` | `"prefix id=42"` | **`42`** |
| `"id=%d"` | `"id=42 suffix"` | `42` |
| `"id=%d"` | `"no-match"` | empty tuple |

The third row is a control: conventional scanning may leave trailing input unread. The fourth confirms that an actual non-match already produces an empty tuple. The highlighted second row is different—the input does not begin with the literal `id=` required by the format, but `scanf` silently accepts it.

## Why it fails

`scanf` tokenizes the format into a regex, wraps it only with slashes, and obtains the first substring occurrence through `regex.match`:

```eo
match. > found
  regex wrapped
  read
```

[`string/scanf.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/scanf.eo#L103-L108) does not prefix the generated pattern with `^`. Once `match` finds `id=42` at offset `7`, the conversion code reads the capture group and returns `42` as though the format had matched from the beginning.

## Expected behavior

The first character of the format must correspond to the first character of the input. `string.scanf "id=%d" "prefix id=42"` should return an empty tuple, just as any non-matching format does, while `"id=42 suffix"` may continue to return `42`.

## Regression coverage

Add a no-leading-match test that asserts the result has length zero:

```eo
scanf "id=%d" "prefix id=42" .length.eq 0
```

Prefixing the generated regex with `^` preserves existing trailing-input behavior and prevents this false parse.

gemshrine:
Fixes #6066.

## Problem

`string.scanf` wraps its generated pattern only in slashes and finds the first occurrence anywhere in the input via `regex.match`. When the input does not start with the format's literal prefix, `match` still finds the pattern further along and returns the values from that later substring. For example `scanf "id=%d" "prefix id=42"` returns `42` even though the input does not begin with `id=`.

## Change

The generated regex is now anchored to the start of the input by prefixing it with `^`, so the first character of the format must correspond to the first character of the input. Literal `^`/`$` characters coming from the format are already escaped by the `escaped` helper, so the added anchor cannot collide with them. Trailing input after a match is still allowed.

## Covered behavior

Two new regressions: `scanf "id=%d" "prefix id=42"` now has length `0`, and `scanf "id=%d" "id=42 suffix"` still parses `42`. Existing tests such as `"Im%d %s ..."` on `"Im18 ..."` and `"%s!"` on `"hello!"` confirm start-anchored and trailing-input cases keep working.

## Verification

Ran `mvn -pl eo-runtime test -Dtest=EO_string.EOscanfTest` (17/17).

### DIFF
--- a/eo-runtime/src/main/eo/string/scanf.eo
+++ b/eo-runtime/src/main/eo/string/scanf.eo
@@ -121,6 +121,7 @@
       regex
         chained *1
           "/"
+          "^"
           pattern
           "/"
       read
@@ -235,6 +236,20 @@
       9.999999999E7
       0.01
       100000000
+
+  eq. ++> tests-scanf-does-not-parse-unmatched-leading-text
+    length.
+      scanf
+        "id=%d"
+        "prefix id=42"
+    0
+
+  eq. ++> tests-scanf-ignores-trailing-text-after-match
+    42
+    head.
+      scanf
+        "id=%d"
+        "id=42 suffix"
 
   eq. ++> tests-scanf-with-complex-float
     1991.01


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (15) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1045](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6093, opened 2026-07-29
Annotate void attributes of native atoms with their types

### LINKED ISSUE #6082: Lack of type annotations for void attributes in native atoms
Native atoms in `eo-runtime` declare what they return but not what they accept. The runtime has 28 atoms with 29 void attributes, and only the two voids of `recovered` (`eo-runtime/src/main/eo/recovered.eo:16-17`, annotated in #5741) say anything about their type; the `cant-*` recovery callbacks got their `/{...}` formas in #5267, so the annotation vocabulary already exists in the language and already lands in XMIR as `type=` and `args=` attributes — the ordinary arguments were simply never annotated.

```
[@] > bytes
  [b] > and /Q.bytes
  ...
  [start len] > slice /Q.bytes
    ? > cant-slice /{Q.string}
```

Every atom in `eo-runtime/src/main/eo/bytes.eo:24-33` names its return forma, while `b`, `start` and `len` stay bare, and the same holds for `number.plus`/`times`/`div`/`gt` (`number.eo:19-22`), `directory.walk` (`directory.eo:24`), `malloc.of`, `chunk.read`/`resized`/`write`, `string.printf` and the rest. This matters more for atoms than for regular objects: an EO-bodied formation shows its expectations in its body, but an atom's body is Java, so a static tool (a type checker, an IDE, a documentation generator) reading XMIR has nothing to derive the argument types from — the return side of the native contract is machine-readable and the argument side is a black hole.

The fix is a documentation pass over the runtime sources, no parser changes: annotate each atom void with the existing syntax where a single forma is honest, e.g. `glob` of `directory.walk` as `/Q.string`. One caution discovered while hand-checking call sites: the `>> name!` const sugar desugars to `(dataized ...).as-bytes`, so several numeric-looking arguments actually receive raw `bytes` — in `tuple.eo`'s `at`, both `0.gt index` and `length.plus idx` hand `Φ.bytes` to the numeric atoms, which means annotating the binary arithmetic voids as `/Q.number` would be wrong. Those, together with the voids that accept any dataizable object by design — `target` of `dataized`, `data` of `chunk.write`, `b` of `bytes.eq` (which receives a `bool` in `01-.eq x` in `bool.eo`) — are better left bare until the annotation language can express that bound. A practical rule for the pass: before annotating a void, grep its call sites and confirm every argument actually carries the forma being declared.

maxonfjvipon:
Closes #6082.

Every native atom already declared what it returns, but nothing about what it accepts, so a static reader of XMIR had no way to derive argument types from a Java body. 25 of the 29 atom voids now carry one.

A type annotation is legal only on a vertical void, and that rule turned out to have a hole in it: an atom could still declare voids in its bracket head, where no annotation fits, and `move-voids-up` hoists body voids *behind* the head ones — so a bare head void jumped ahead of a typed vertical one no matter what the source said. `[] > write` with `? > offset /Q.number` above a bare `? > data` reprinted as `[data] > write` with `? > offset /Q.number`, silently swapping the two arguments.

So an atom's head is now empty by rule (R-3.4.10): bracket parameters on a `/sig` line are a parse error (`an atom must declare its void attributes vertically, as ? > name lines`), source order is void order, and an annotated void may be followed by an unannotated one.

```
[] > slice /Q.bytes
  ? > start /Q.number
  ? > len /Q.number
  ? > cant-slice /{Q.string}
```

The printer follows: every void of an atom becomes a body line instead of the untyped ones being folded back into `[…]`, which collapses the three near-identical void templates into one, and `chunk.write.offset` gets back the `/Q.number` it had to give up. A non-atom formation is untouched — it keeps both forms and may mix them freely.

Four voids stay bare on purpose, each because something in this repository passes it more than one forma. `bytes.eq.b` gets a `bool` from `bool.eo`, `bytes.concat.b` gets strings from `path.eo` and `stdin.eo`, `chunk.write.data` gets an arbitrary object, and `dataized.target` gets anything dataizable. Annotating those would describe a contract the library itself breaks; they want a dataizable bound the annotation language cannot express yet.

One inconsistency surfaced along the way: `chunk.resized` declared its void as `size` in EO while the runtime had always bound `new-size`. Both now read `capacity` — one noun, as the `compound-name` lint wants, and unlike `size` it does not shadow the `size` of the chunk that lambda returns, which would make `(m.resized 3).size` read the void instead of resizing anything.

PARSER_SPEC carries the new R-3.4.10 plus the rules it touches: R-3.4.7 (reverse printing no longer canonicalises to the bracket form), R-3.4.8, R-6.3.1, R-6.3.4(a), and the §9.4 / §9.9 tables.


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (419 >= 100); -8 for way too many hits-of-code (419 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+722](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6099, opened 2026-07-29
#6094: retry Saved.moved on transient move failures

### LINKED ISSUE #6094: `Saved.moved` throws `AccessDeniedException` on Windows when target file is concurrently open for reading
## What happens

`Saved.moved` (`eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java:138`) does the atomic rename this way:

```java
try {
    Files.move(
        tmp, target,
        StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING
    );
} catch (final AtomicMoveNotSupportedException ignored) {
    Files.move(tmp, target, StandardCopyOption.REPLACE_EXISTING);
}
```

It only catches `AtomicMoveNotSupportedException`. On Windows, `Files.move(..., REPLACE_EXISTING)` throws `AccessDeniedException` instead when the target file is open by a concurrent reader at the moment of the move. POSIX allows replacing a file that's open for reading; Windows blocks it.

This surfaces as intermittent CI failures in `SavedTest.exposesNoPartiallyWrittenFileToConcurrentReader` on `windows-2022`, seen multiple times recently (e.g. PR #6069, PR #6056), always the same `AccessDeniedException` on the temp-to-target move. The test itself is fine, it deliberately races readers and writers on the same path (that's its whole point, from #5873); the code under test just doesn't retry on this specific Windows error.

A related case was fixed before in `RtCentral.java` with `@RetryOnFailure`, but that fix only covers that one caller. `Saved.moved` itself, and any other caller going through it, is still exposed.

## What should happen

`Saved.moved` should retry the move (or catch `AccessDeniedException` in addition to `AtomicMoveNotSupportedException` and retry after a short delay) so a transient Windows file lock from a concurrent reader doesn't fail the whole write.

morphqdd:
Closes #6094.

`Saved.moved` moves a temp file onto its target with `ATOMIC_MOVE, REPLACE_EXISTING`, catching only `AtomicMoveNotSupportedException`. On Windows, `Files.move(..., REPLACE_EXISTING)` throws `AccessDeniedException` instead of replacing the target when a concurrent reader/writer has it open at that exact instant. POSIX allows this; Windows blocks it. This surfaced as an intermittent `SavedTest.exposesNoPartiallyWrittenFileToConcurrentReader` failure on `windows-2022` CI (seen on PR #6069, #6056, among others), always the same `AccessDeniedException` on the move.

Fix: `@RetryOnFailure(delay = 1L, unit = TimeUnit.SECONDS)` on `Saved.moved`, matching the existing retry idiom already used elsewhere in this class for transient IO failures (`CommitHashesText`, `OyRemote`).

Test: `SavedTest.retriesMoveWhenTargetTemporarilyBlocked` makes the target path a directory (any platform, deterministically blocks the move with a real IOException), schedules the obstruction's removal 300ms later on a background thread, well before the 1-second retry fires, then asserts the write ultimately succeeds. Reverted the fix locally and confirmed this test fails without it.

`mvn -pl eo-maven-plugin test -Dtest=SavedTest,CacheTest,ConcurrentCacheTest` - 14 tests, all passing. qulice clean.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
@@ -4,6 +4,7 @@
  */
 package org.eolang.maven;
 
+import com.jcabi.aspects.RetryOnFailure;
 import com.jcabi.log.Logger;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
@@ -11,6 +12,7 @@ import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
+import java.util.concurrent.TimeUnit;
 import org.cactoos.Input;
 import org.cactoos.Scalar;
 import org.cactoos.Text;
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (34) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2720](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6100, opened 2026-07-29
#6091: fix Saxon 13.0 multi-argument xsl:function parameter mis-binding

### LINKED ISSUE #6091: Saxon-HE 13.0 incorrectly assigns values to `xsl:function` parameters triggering parse termination
`set-locators.xsl:64` calls `eo:locator(/object, .)` and the function at
`eo-parser/src/main/resources/org/eolang/parser/parse/set-locators.xsl:14-16`
takes two `node()` parameters. Under Saxon-HE 13.0 (pinned at `pom.xml:211`)
the second parameter sometimes receives the value of the first one, so `$o`
becomes the `object` root element and the `xsl:message terminate="yes"` guard
at line 18 kills the whole parse. That is why
`MjParseTest.parsesConcurrentlyWithLotsOfPrograms`
(`eo-maven-plugin/src/test/java/org/eolang/maven/MjParseTest.java:243`) fails
in about 2-3% of runs, for example in
https://github.com/objectionary/eo/actions/runs/30448037612 on `macos-15`:
"Only 'o' XML elements are accepted here".

The trigger is one compiled stylesheet used by many threads at once, which is
exactly what `Parsing` does: a single `Canonical` pipeline is shared by the
`Threaded` pool. A minimal plain-JAXP program, with a two-argument
`xsl:function` returning `name($first)/name($second)`, 24 threads and 300
rounds sharing one `Templates`, returns 106 wrong `object/object` results on
Saxon-HE 13.0 and zero on 12.5, 12.9 and 12.10; the same function reduced to
one argument returns zero on 13.0. Saxon-HE went from 12 to 13 in 5734f4bfcb,
a renovate bump.

The narrow fix is to drop the redundant `$program` parameter from `eo:locator`
and reach the program through `root($o)/object`; that removes the only loud
symptom and it is what branch `6083` does. Eight more stylesheets still declare multi-argument `xsl:function`s
and run through the same shared pipelines, so their corruption would be silent:
`const-to-dataized.xsl`, `restore-local-names.xsl`, `unnecessary-as.xsl`,
`merge-monikers.xsl`, `inline-cactoos.xsl`, `to-java.xsl`,
`set-original-names.xsl` and `add-probes.xsl`. Pinning Saxon back to 12.10 is
not a two-line change, because `maven-enforcer`'s `RequireUpperBoundDeps` fails
while `org.eolang:lints` and `org.eolang:wpa` demand 13.0.

morphqdd:
Closes #6091.

`eo:original-name` in `set-original-names.xsl` took `($program, $o)`, but `$program` was never read anywhere in the function body, only passed through on the recursive call. That is exactly the shape that made `eo:locator` in `set-locators.xsl` mis-bind arguments under Saxon 13.0 when one compiled stylesheet is shared across threads (`Threaded`/`Parsing`), already fixed the same way. Dropped the parameter entirely.

Also found while touching this: the two YAML regression fixtures for `eo:original-name` (`original-names/add-original-names.yml` and `add-original-names-with-dot.yml`) had a `.yml` extension, but `MjTranspileTest.checksTranspilePacks`'s harness only globs `**.yaml`, so neither ever actually ran. Renamed them to `.yaml`. Once active, both failed on an unrelated pre-existing bug in the fixtures themselves: their XPath assertions used `/object/o[...]` but the harness's actual document root is `<program>`, not `<object>` (visible from other currently-passing fixtures in the same directory, which use `//o[...]` instead). Fixed the assertions to match; the underlying `eo:original-name` output was already correct.

The issue lists 7 more stylesheets with the same risk: `const-to-dataized.xsl`, `restore-local-names.xsl`, `unnecessary-as.xsl`, `merge-monikers.xsl`, `inline-cactoos.xsl`, `to-java.xsl`, `add-probes.xsl`. Unlike `set-original-names.xsl`, their multi-argument functions all genuinely use both parameters, so dropping one isn't a safe option there - each needs its own restructuring. Marked every one of the 25 affected `xsl:function` declarations across those 7 files with a `@todo #6091:NNmin` puzzle naming the functions and the fix needed, so 0pdd files a tracked child issue per file instead of the risk sitting silent and undocumented.

`mvn -pl eo-maven-plugin,eo-printer,eo-parser test` - 1528 (eo-parser) + 316 (eo-printer) + 43+9 (eo-maven-plugin transpile/probe) tests, all passing. qulice clean on all touched files.

### DIFF
--- a/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/set-original-names.xsl
+++ b/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/set-original-names.xsl
@@ -11,9 +11,9 @@
   The function takes the object only; the `program` argument it used
   to take was never read in its body (see #6091 and the same fix in
   set-locators.xsl) - a redundant argument is exactly the shape that
-  triggers Saxon 13.0 mis-binding parameters of a multi-argument
-  function when one compiled stylesheet transforms many documents in
-  parallel threads (which is what the "parse" goal does).
+  triggers Saxon 13.0 into binding parameters of a multi-argument
+  function to the wrong values when one compiled stylesheet transforms
+  many documents in parallel threads (which is what the "parse" goal does).
   -->
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:import href="/org/eolang/parser/_specials.xsl"/>
--- a/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/set-original-names.xsl
+++ b/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/set-original-names.xsl
@@ -8,11 +8,16 @@
   Here we go through all objects and add @original-name attributes
   to all of them. The value of the attribute is a FQN
   of the object.
+  The function takes the object only; the `program` argument it used
+  to take was never read in its body (see #6091 and the same fix in
+  set-locators.xsl) - a redundant argument is exactly the shape that
+  triggers Saxon 13.0 mis-binding parameters of a multi-argument
+  function when one compiled stylesheet transforms many documents in
+  parallel threads (which is what the "parse" goal does).
   -->
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:import href="/org/eolang/parser/_specials.xsl"/>
   <xsl:function name="eo:original-name" as="xs:string">
-    <xsl:param name="program" as="node()"/>
     <xsl:param name="o" as="node()"/>
     <xsl:if test="name($o) != 'o'">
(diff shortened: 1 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2732](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6101, opened 2026-07-29
#5941: declare restore-local-names helper name params as xs:string?

### LINKED ISSUE #5941: XSLT warnings are not being displayed and XPTY0004 pattern aborts are hidden in `restore-local-names.xsl`

morphqdd:
Closes #5941.

`eo:recursive`, `eo:references`, `eo:multi-referenced`, `eo:unreferenced`, `eo:nested-referenced`, `eo:applied-receiver` and `eo:reapplied` in `restore-local-names.xsl` all declare their second parameter as `name as="xs:string"` (exactly one item), but the shared match pattern at the bottom of the file calls every one of them with a bare `@name`. An `o` node can carry `@local` without `@name` - a reference to a const `>>` handle's value inside the parser's `Φ.dataized` shell is exactly that shape - so `@name` evaluates to an empty sequence there, which cannot satisfy an exactly-one `xs:string` parameter. Saxon treats this `XPTY0004` inside a match pattern as recoverable: it warns and treats the pattern as not matching, so this is silent per the issue, not a build failure.

Fix: widened all seven parameters to `xs:string?`. No other change was needed - every comparison against `$name` in these functions already degrades correctly on an empty sequence per ordinary XPath semantics (a general comparison or `concat()` against an empty sequence just yields "no match"/"empty string", not an error), so this eliminates the warning without changing any decision the stylesheet makes.

Also checked `inline-cactoos.xsl:314` per the issue's own note ("worth a glance"): its pattern guards with `starts-with(@name, $auto)`, and `fn:starts-with` treats an empty first argument as a zero-length string by spec, so it was already safe - no change needed there.

The issue's other suggested fix (make jcabi-xml's transform-time XSLT warnings fatal in general, so this class of bug fails the build loudly instead of relying on manual audits) is a separate, larger change to the xsline/jcabi-xml wiring. It would help catch the next unrelated instance of this class of bug earlier, but it's not needed to fix the actual reported symptom here, so it's left out of scope for this PR.

Test: `XmirTest.doesNotWarnOnLocalWithoutName` runs the sheet on a minimal document reproducing the `Φ.dataized`-wrapped `@local`-without-`@name` shape, with a log4j appender attached to the root logger, and asserts no WARN-level message contains "empty sequence". Reverted the fix locally and confirmed the test fails with exactly that warning; restored the fix and confirmed it passes. This test used to live in its own `RestoreLocalNamesTest` class with a `@SuppressWarnings("JTCOP.RuleAllTestsHaveProductionClass")` since it had no matching production class; moved it into `XmirTest` instead (which does map to `Xmir.java`), so the rule is satisfied genuinely instead of suppressed.

`mvn -pl eo-printer test` - 317 tests, all passing. qulice clean.


morphqdd:
Closes #5941. This fixes the actual reported symptom: the seven XPTY0004 warnings from `restore-local-names.xsl` are gone, since every place `@name` could be absent (a `@local` node with no `@name`) is now a valid `xs:string?` argument instead of a type error.

The issue's other suggestion, making jcabi-xml's transform-time warnings fatal in general, is a separate, larger defensive change to how `xsline`/`jcabi-xml` are wired project-wide (an injectable `ErrorListener`), not needed to fix this specific bug. Happy to file that as its own issue if useful, since it would help catch the next unrelated instance of this class of bug earlier, but it's out of scope for #5941 itself.

eo-printer/src/test/java/org/eolang/printer/RestoreLocalNamesTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.log4j.spi.LoggingEvent;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@code restore-local-names.xsl}.
+ * @since 0.35.0
+ */
+@SuppressWarnings("JTCOP.RuleAllTestsHaveProductionClass")
```
yegor256: @morphqdd listen to JTCOP. This test belongs to another Test class.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2824](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2705](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6103, opened 2026-07-29
#5943: reject non-UTF-8 bytes in printf %s instead of mojibake

### LINKED ISSUE #5943: `printf %s` conversion silently corrupts non-string arguments rather than converting or raising error

morphqdd:
Closes #5943.

`PrintfArgs`'s `%s` conversion used `Dataized::asString`, which decodes any byte length as UTF-8 with no validation, unlike `%f`/`%d`/`%b` which all validate their argument's byte shape (exactly 8 bytes for a number, exactly 1 for a boolean) and fail cleanly on a mismatch. So `printf "%s" 42` silently decoded the number's raw IEEE-754 bytes as text instead of converting or rejecting - mojibake with no error.

Fix: `%s` now decodes strictly (`CodingErrorAction.REPORT` on both malformed input and unmappable characters) and raises `ExFailure` naming the byte count on a decode failure, instead of `Dataized::asString`'s permissive decode. This is the same class of validation `%f`/`%d`/`%b` already do for their own byte shape.

`printf "%s" 3.14` from the issue's own reproduction is exactly this case (`3.14`'s IEEE-754 bytes are not valid UTF-8), so the issue's reported bug is fixed.

One edge case is out of scope and marked with a `@todo` puzzle on `PrintfArgs.validString`: a number whose bytes happen to coincidentally decode as valid UTF-8 (e.g. `100.5`) still passes through, because by the time this method runs only raw bytes are available - the originating Phi's type is already lost. Closing that needs type information threaded from the Phi through `Dataized`/`PrintfArgs` before dataization, a bigger architectural change out of scope for this fix.

Test: `PrintfArgsTest.rejectsNonStringBytesInsteadOfMojibake` reproduces the issue's own `printf "%s" 3.14` example, asserting `ExFailure` instead of silent garbage. Reverted the fix locally and confirmed the test fails (nothing thrown, mojibake returned); restored the fix and confirmed it passes.

`mvn -pl eo-runtime test -Dtest=PrintfArgsTest,EO_string.EOprintfTest` - 47 tests, all passing.


yegor256:

morphqdd:

eo-integration-tests/src/test/resources/snippets/printf-string.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - ".*Hello, Jeff!.*"
+file: snippets/printf-string.eo
+args: ["snippets.printf-string"]
+eo: |
+  +package snippets
+
+  [] > printf-string
+    stdout > @
+      "Hello, %s!\n".printf
+        * "Jeff"
```
yegor256: @morphqdd how this test reproduces the bug you are fixing? does it?

Thayorns (commented):
Good catch on the root cause, and a strict decoder is certainly better than `Dataized::asString` here. Two things I would like to understand before this lands.

The first is the choice of check itself, left inline on `PrintfArgs.validString`.

The second is the documentation. The header of `printf.eo` still reads:

> - %s - converts object to `string`

That is the exact sentence #5943 quoted as the source of its expectation. After this change `%s` converts nothing at all: it demands that the bytes already be text, and fails when they are not. The documented contract and the implemented one still disagree, only in the opposite direction from before.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -190,6 +194,42 @@ private static Object fmt(final char symbol, final Dataized element) {
...
+     * @todo #5943:60min Detect wrong-type bytes that happen to be valid UTF-8.
+     *  This check only rejects structurally invalid UTF-8. A number whose raw
+     *  bytes coincidentally decode as valid UTF-8 text still passes through as
+     *  if it were a genuine string, because by the time this method runs only
+     *  raw bytes are available via Dataized#take() - the originating Phi's
+     *  type (string vs number etc.) is already lost. Closing this needs type
+     *  information threaded from the source Phi through Dataized/PrintfArgs
+     *  before dataization happens, not another byte-level check.
+     */
+    private static String validString(final Dataized element) {
```
Thayorns, line 214: I looked at what the IEEE-754 bytes of a few numbers actually are:

- `42` -> `40-45-00-00-00-00-00-00`, valid UTF-8, decodes to `@E` followed by six NULs
- `100.5` -> `40-59-20-00-00-00-00-00`, valid UTF-8
- `1` -> `3f-f0-00-00-...`, invalid
- `3.14` -> `40-09-1e-b8-...`, invalid

So `printf "%s" 42`, which is the first row of the table in #5943 and the case in its title, still returns mojibake with no error after this change, and so does `100.5`. `printf "%s" 1` now fails while `printf "%s" 42` does not. Whether a number is rejected depends on its exponent bits, which from the caller's seat is arbitrary. I would argue that is harder to live with than the old consistent corruption, because now the same mistake sometimes explodes and sometimes does not, and nothing in the calling code says which.

The `@todo` frames this as an edge case, "a number whose bytes happen to coincidentally decode as valid UTF-8". Small integers are the common case for `%s` misuse though, not an edge one. Is UTF-8 validity the right property to be testing here at all, given that the bug is a type mismatch rather than an encoding problem?

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -190,6 +194,42 @@ private static Object fmt(final char symbol, final Dataized element) {
...
+        try {
+            return StandardCharsets.UTF_8.newDecoder()
+                .onMalformedInput(CodingErrorAction.REPORT)
+                .onUnmappableCharacter(CodingErrorAction.REPORT)
+                .decode(ByteBuffer.wrap(bytes))
+                .toString();
+        } catch (final CharacterCodingException ex) {
+            throw new ExFailure(
+                String.format(
+                    "Can't convert %d bytes to a string for the '%%s' conversion, not valid UTF-8",
```
Thayorns, line 225: The message tells me how many bytes were rejected, but not which argument they came from. With `printf "%s %s %s" a b c` I would know that eight bytes somewhere in there are not text, and would still have to guess which of the three. The byte count is the least identifying thing available at this point.

eo-runtime/src/main/eo/string/printf.eo
```diff
@@ -191,6 +191,10 @@
     "%#x"
     5
 
+  printf *1 --> stops-on-a-non-utf8-byte-argument-as-a-string
```
Thayorns, line 194: This covers `3.14`, whose bytes happen not to be valid UTF-8. The example from the issue's own title, `printf "%s" 42`, is absent, and it still passes through silently. A case for it here would have surfaced that.

eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
```diff
@@ -113,6 +113,22 @@ void acceptsLargestDoubleThatTrulyFitsInLongRange() {
...
+        tuple.put("length", new Data.ToPhi(1));
+        tuple.put("head", new Data.ToPhi(3.14));
+        MatcherAssert.assertThat(
+            "a number's raw bytes are not valid UTF-8 text and must be rejected, not silently decoded into mojibake",
+            Assertions.assertThrows(
+                ExFailure.class,
+                () -> new PrintfArgs("%s", 1L, tuple.take("at")).formatted(),
+                "must throw ExFailure, not silently return garbage text"
+            ).getMessage(),
+            Matchers.containsString("not valid UTF-8")
```
Thayorns, line 128: This pins the test to the wording of the exception message. Our test rules ask us not to assert on error messages: the wording is not the behaviour, and rephrasing the message for clarity later would break a test that has nothing to do with the change.

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2712](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+365](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2832](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6104, opened 2026-07-29
#5945: reject NaN in printf %d instead of narrowing to zero

### LINKED ISSUE #5945: `printf` `%d` conversion mistakenly turns `nan` into `0` inconsistent with infinity handling

morphqdd:
Closes #5945.

`PrintfArgs.toLong` only guarded the two out-of-range directions. `NaN` compares `false` against every relational operator, so `NaN < Long.MIN_VALUE` and `NaN >= 2^63` are both false, `NaN` slips past the check, and `(long) Double.NaN` narrows to `0` per JLS 5.1.3. So `printf "%d" nan` silently produced `"0"` while `printf "%d" pinf`/`ninf` already failed cleanly.

Fix: reject `Double.isNaN(number)` explicitly, alongside the existing range check.

Test: `PrintfArgsTest.rejectsNanInsteadOfNarrowingToZero` reproduces the issue's own example, asserting `ExFailure` instead of `"0"`. Reverted the fix locally and confirmed the test fails (nothing thrown); restored it and confirmed it passes.

`mvn -pl eo-runtime test -Dtest=PrintfArgsTest,EO_string.EOprintfTest` - 47 tests, all passing. qulice clean.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
@@ -192,12 +192,15 @@ final class PrintfArgs {
     /**
      * Convert a number to {@code long} for the {@code %d} conversion, rejecting
      * a value outside {@code long} range instead of silently saturating to
-     * {@link Long#MAX_VALUE}/{@link Long#MIN_VALUE} as {@link Double#longValue()} does.
+     * {@link Long#MAX_VALUE}/{@link Long#MIN_VALUE} as {@link Double#longValue()} does,
+     * and rejecting {@link Double#NaN} instead of letting it slip past both bounds
+     * (every relational comparison against {@code NaN} is {@code false}) and narrow
+     * to {@code 0} per JLS 5.1.3.
      * @param number Number to convert
      * @return The number as {@code long}
      */
     private static long toLong(final double number) {
-        if (number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
+        if (Double.isNaN(number) || number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
             throw new ExFailure(
                 "The number %s doesn't fit into long range for the '%%d' conversion",
                 number
--- a/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
@@ -200,7 +200,11 @@ final class PrintfArgs {
      * @return The number as {@code long}
      */
     private static long toLong(final double number) {
-        if (Double.isNaN(number) || number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
+        if (
+            Double.isNaN(number)
+                || number < Long.MIN_VALUE
+                || number >= PrintfArgs.LONG_UPPER_LIMIT
+        ) {
             throw new ExFailure(
                 "The number %s doesn't fit into long range for the '%%d' conversion",
                 number
--- a/eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
@@ -100,6 +100,22 @@ final class PrintfArgsTest {
     }
 
     @Test
+    void rejectsNanInsteadOfNarrowingToZero() {
+        final Phi tuple = Phi.Φ.take("tuple").copy();
+        tuple.put("length", new Data.ToPhi(1));
+        tuple.put("head", new Data.ToPhi(Double.NaN));
+        MatcherAssert.assertThat(
+            "NaN must be rejected like an infinity, not silently narrowed to 0 per JLS 5.1.3",
+            Assertions.assertThrows(
+                ExFailure.class,
+                () -> new PrintfArgs("%d", 1L, tuple.take("at")).formatted(),
+                "must throw ExFailure, not silently return 0"
+            ).getMessage(),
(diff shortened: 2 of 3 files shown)


morphqdd:
@yegor256, review please

morphqdd:

yegor256:
@morphqdd please, add EO test too

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -193,12 +193,19 @@ private static Object fmt(final char symbol, final Dataized element) {
...
      */
     private static long toLong(final double number) {
-        if (number < Long.MIN_VALUE || number >= PrintfArgs.LONG_UPPER_LIMIT) {
+        if (
+            Double.isNaN(number)
+                || number < Long.MIN_VALUE
+                || number >= PrintfArgs.LONG_UPPER_LIMIT
+        ) {
             throw new ExFailure(
                 "The number %s doesn't fit into long range for the '%%d' conversion",
```
Thayorns, line 209: `printf "%d" nan` now reports "The number NaN doesn't fit into long range for the '%d' conversion". That fits `1.0e30`, which really is too big. It does not fit `NaN`, which is not a magnitude at all, so the message sends the reader looking for an oversized value.

The docblock you added above draws that distinction precisely. None of it reaches the person who hits the failure.

eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
```diff
@@ -100,6 +100,22 @@ void rejectsTwoPowerSixtyThreeInsteadOfSaturating() {
...
+        tuple.put("length", new Data.ToPhi(1));
+        tuple.put("head", new Data.ToPhi(Double.NaN));
+        MatcherAssert.assertThat(
+            "NaN must be rejected like an infinity, not silently narrowed to 0 per JLS 5.1.3",
+            Assertions.assertThrows(
+                ExFailure.class,
+                () -> new PrintfArgs("%d", 1L, tuple.take("at")).formatted(),
+                "must throw ExFailure, not silently return 0"
+            ).getMessage(),
+            Matchers.containsString("doesn't fit into long range")
```
Thayorns, line 114: This asserts the same fragment as the two tests above it, so all three stay green whichever of the three rules did the rejecting. Nothing records that `NaN` is turned away for its own reason.

It does catch the real bug — without the guard nothing is thrown and this goes red.

Thayorns (commented):
Good find. Two things have to line up for this bug to exist — every comparison against `NaN` is false, so it clears both bounds, and the narrowing cast then gives `0` instead of something obviously wrong. Neither is visible from reading the guard.

The fix is complete for what it targets: `toLong` has one caller, the `'d'` entry in `CONVERSION`. `%f` still prints `NaN`, which seems right, and the infinities were already caught since `LONG_UPPER_LIMIT` is `0x1.0p63`.

One note on the message the new branch produces, and a related one on the test.

I see @yegor256 asked for an EO test, so I will not restate that.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (31) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2728](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (2) comments; -4 for too few (31) hits-of-code; +1 to give you at least something.  Your running score is [+358](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6106, opened 2026-07-29
#6097: reject eo.step values the parser cannot read back

### LINKED ISSUE #6097: `eo.step` allows invalid indentation widths causing unparsable EO output and source corruption
## What happens

`eo.step` (`MjFormat`/`MjPrint` `@Parameter(property = "eo.step")`, feeding `PenaltyKey.STEP` into `Penalty`/`Pretty` in eo-printer) lets a user set any integer as the indentation width. Only the value 2 actually works, because EO's own parser hard-codes the indent step to 2 (`eo-parser`'s `Stack.STEP`).

Three concrete failures, all reproducible directly against the printer:

- `-Deo.step=0` throws `java.lang.ArithmeticException: / by zero` in `Penalty.indents` (`eo-printer/src/main/java/org/eolang/printer/Penalty.java:152`, `return spaces / this.weight(PenaltyKey.STEP);`).
- `-Deo.step=-2` (or any negative value) throws `java.lang.IllegalArgumentException: count is negative: -2` in `Pretty`'s constructor (`eo-printer/src/main/java/org/eolang/printer/Pretty.java:72`, `" ".repeat(config.getOrDefault(STEP, ...))`).
- Any other value (1, 3, 4, ...) does not crash, but the printer emits EO source that the parser can no longer read back. Formatting this program with `step=4`:

```
[x] > main
  [] > inner
    x > first
    x > second
```

with `-Deo.step=4` produces indentation the parser rejects with `unexpected indentation` / `indent increased by more than one level` errors on reparse (with `step=1` or `step=3` it's `unexpected odd indent`), because the parser's own indent step is fixed at 2 regardless of what the printer was told to use.

The dangerous part: `MjFormat` with `-Deo.autoFix=true` writes the printer's output straight back onto the original `.eo` source file (`MjFormat.java:114`, `new Saved(canonical, source).value()`). So `mvn eo:format -Deo.autoFix -Deo.step=4` silently overwrites a working, parseable `.eo` source tree with output the project's own parser can no longer parse. Nothing in `MjFormat`/`MjPrint` validates that `step` is 2 before using it.

The existing test `MjFormatTest.appliesCustomIndentationStep` only asserts that the output string contains 8 leading spaces; it never reparses the result, so it doesn't notice the output is invalid EO.

## What should happen

`MjPrint`/`MjFormat` should validate `eo.step` (reject anything other than 2, since that's the only value the parser can read back), throwing a clear `IllegalArgumentException` naming the parameter, instead of silently producing output that corrupts sources under `-Deo.autoFix` or crashing with an unrelated arithmetic/argument exception for 0 or negative values.

morphqdd:
Closes #6097.

`eo.step` (`MjPrint`/`MjFormat`'s `weights()`) let a user set any integer as the indentation width, but the parser's own indent step is hard-coded to 2. Three failure modes, all confirmed:

- `-Deo.step=0` threw `ArithmeticException: / by zero` in `Penalty.indents`.
- `-Deo.step=-2` threw `IllegalArgumentException: count is negative: -2` in `Pretty`'s constructor.
- Any other value (1, 3, 4...) printed EO the parser can't read back, and `mvn eo:format -Deo.autoFix` writes that straight onto the source file, corrupting it.

Per @yegor256's comment, removed the `eo.step` option entirely instead of validating it. Its existence was the mistake - there's only ever one valid value, so there's nothing left to configure. `PenaltyKey.STEP`'s fallback was already `2`, so both mojos now just always use it.

Removed: the `eo.step` `@Parameter` field from both `MjPrint` and `MjFormat`, its handling in both `weights()` methods, and the `nested()`/`appliesCustomIndentationStep` test that exercised the old (now gone) custom-step behavior.

`mvn -pl eo-maven-plugin test` - 495 tests, all passing. qulice clean.


yegor256:
@morphqdd let's just get rid of `eo.step` option all together. Its very existence is a mistake.

morphqdd:
@yegor256, done - removed eo.step entirely, both mojos always use the parser's own step now.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2840](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6107, opened 2026-07-29
#6098: suggest vertical application in the method continuation error

### LINKED ISSUE #6098: Parser fails to suggest correct syntax for handling errors
Take a look at this: https://github.com/objectionary/eo/issues/6095#issuecomment-5120452089

Here end-user see the syntax error message but doesn't know what to do with it and how to rewrite the code so it's parsed correctly.

Maybe it would be better our parser explicitly suggest the right way to write the code where it's possible. Or it's not a job of parser?

In this particular case it would be: `method continuation not allowed after horizontal application, try vertical application instead`

@yegor256 WDYT?

morphqdd:
Closes #6098.

When a `.method` continuation follows a horizontally-completed application (`.with 1` then `.with 2`), the parser rejects it but the error only said `method continuation not allowed after horizontal application`. A reader hits this constantly writing fluent-style method chains and has no idea from the message alone that the fix is to write the same chain vertically instead.

Fix: the message now ends with `, try vertical application instead`, matching the exact wording `@maxonfjvipon` suggested in the linked issue thread.

Test: updated the two existing `eo-typos` fixtures that already exercise this exact error (`vmethod-after-happlication.yaml`, `vmethod-with-hargs-fluent-chain.yaml`) to expect the new wording. Reverted the fix locally and confirmed both fail with the old message; restored the fix and confirmed they pass. Also updated `PARSER_SPEC.md`'s two references to this message (R-5.2.3(b) and the error-message table) to keep the spec in sync with the code.

`mvn -pl eo-parser test -Dtest=EoTest,EoSyntaxTest` - 1012 tests, all passing. qulice clean.

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -861,7 +861,7 @@ R-5.2.2. When popping an entry at indent `K + 2` (so the new top is at indent `K
 
 R-5.2.3. **MethodDispatch line dispatch.** If the line's kind is `MethodDispatch` (`.method` continuation):
   - **(a) Extend:** If the top's openness ∈ {`open`, `vertical-completed`} AND the top's kind is **not** in the horizontally-completed set (Appendix A) AND the top's kind is **not** `inline-phi-formation`: extend the top's kind to `vmethod` (or to `vmethod-with-hargs` if the new line carries ≥1 hargs), update `named?` if the line carries a name suffix, leave openness `open` (or transition to `horizontal-completed` if the new kind is `vmethod-with-hargs`).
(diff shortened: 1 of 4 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2808](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6108, opened 2026-07-29
#6102: fix dataized-to-const.xsl concatenating literal payload with argument text

### LINKED ISSUE #6102: `dataized-to-const.xsl` incorrectly concatenates data literal payload with argument text
## What happens

`dataized-to-const.xsl:37` restores a data literal's payload text for the `Φ.dataized` wrapper it builds:

```xml
<xsl:if test="eo:has-data($argument)">
  <xsl:value-of select="$argument"/>
</xsl:if>
```

`<xsl:value-of select="$argument"/>` takes the XPath string value of the element, which concatenates every descendant text node. `eo:has-data()` (`_funcs.xsl:8-11`) only checks for a **direct** text() child, so once a data literal carries both its own payload text and an argument element (`01-02 5 > @!`), the payload and the argument's own printed text get glued into one string:

```
INPUT: [] > app
  01-02 5 > @!
```
prints (on top of PR #5767, `fix/5721-stunhex-keep-args`) as:
```
[] > app
  01-025 > @!
    5:1
```
`01-025` is `01-02` and `5` concatenated into a single bogus byte literal.

## Why this isn't reachable yet

On current master, `StUnhex` already deletes the argument before this shift runs (that's #5721), so no input reaches the broken branch - `$argument` never has both a text node and an element child at once. It becomes reachable once a fix for #5721 (e.g. PR #5767) lands and starts preserving the argument through to this shift.

## What should happen

Use `eo:read-data($argument)` instead of the XPath string value, matching how the rest of the pipeline reads a literal's own payload text:

```xml
<xsl:value-of select="eo:read-data($argument)"/>
```

Verified with a local patch that this produces the correct `01-02 > @!` / `5:1` split.

Found while investigating #5912.

morphqdd:
Closes #6102.

`dataized-to-const.xsl`'s reverse pass restores a data literal's own payload text with `<xsl:value-of select="$argument"/>`, which takes the XPath string value of the whole element - concatenating every descendant text node, not just the literal's own direct text. Once `$argument` carries both its own payload text and an argument element (e.g. a byte literal like `01-02 5 > @!`, once #5721 lands and stops dropping that argument), the payload and the argument's own printed text get glued into one string, e.g. `01-02` and `5` becoming `01-025`.

Fix: use `eo:read-data($argument)` instead, matching how the rest of the pipeline (`const-to-dataized.xsl`) reads a literal's own payload text - it only joins `$o/text()`, the direct text children, not the full descendant string value.

Test: `DataizedToConstTest` feeds `dataized-to-const.xsl` a hand-built XML fragment shaped exactly like the bug (a node with both its own text and an argument child), since the real end-to-end repro needs #5721 fixed first and isn't reachable on master yet. Reverted the fix locally and confirmed the test gets `01-025`; restored the fix and confirmed it gets the correct `01-02`.

`mvn -pl eo-printer test` - 317 tests, all passing. qulice clean.


morphqdd:

yegor256:
@morphqdd add YAML test(s) too, please

Thayorns (commented):
The fix itself is right, and it is the smallest one that works. `eo:has-data` is `normalize-space(string-join($o/text(), '')) != ''` — it only ever looked at direct text children, while the old `value-of select="$argument"` took the string value of the whole element. So the guard and the read disagreed, and `read-data` is what makes them agree again. Good catch.

I also checked the test actually earns its place: with the old line, `$argument`'s string value is `01-02` plus `5`, so the output text is `01-025` and the `text()='01-02'` predicate fails. It does discriminate.

The one thing I would like you to look at is in the XSL comment below: the forward pass has the same defect, and the reasoning in the PR description about it does not hold up. Two smaller notes on the test.

One minor thing not worth its own thread: `read-data` also strips leading and trailing whitespace, which `value-of select="$argument"` did not. That is a second change riding along with the intended one. It is probably an improvement, but indented XMIR is exactly where it would show up, so worth being sure no pack depends on the untrimmed text.

eo-printer/src/main/resources/org/eolang/printer/print/dataized-to-const.xsl
```diff
@@ -34,7 +34,7 @@
             <xsl:apply-templates select="."/>
           </xsl:for-each>
           <xsl:if test="eo:has-data($argument)">
-            <xsl:value-of select="$argument"/>
+            <xsl:value-of select="eo:read-data($argument)"/>
```
Thayorns, line 37: The forward pass has the same bug, and I think it needs to be said out loud before this merges.

`const-to-dataized.xsl` lines 77-82 are the same shape as the code you are fixing here:

```
<xsl:for-each select="o">
  <xsl:apply-templates select="."/>
</xsl:for-each>
<xsl:if test="eo:has-data(.)">
  <xsl:value-of select="."/>
</xsl:if>
```

Same `has-data` guard reading only direct text, same `value-of` taking the whole element's string value. So a literal that carries both its own payload and an argument child gets glued on the way in too, under exactly the condition you describe in the PR — once #5721 stops dropping that argument.

This also means the PR description is not accurate. It says the fix matches "how the rest of the pipeline (`const-to-dataized.xsl`) reads a literal's own payload text - it only joins `$o/text()`". It does not. That file uses `value-of select="."`, which is the pattern you are removing here. Nothing in the pipeline uses `read-data` for this yet, so this change is the first of the pair rather than a move toward consistency.

I am not asking you to widen the PR — staying on the printer side is the right call for #6102. But the round trip is only half fixed, and the twin defect is easy to lose once this closes. Worth an issue, or a `@todo` pointing at it.

eo-printer/src/test/java/org/eolang/printer/XmirTest.java
```diff
@@ -90,6 +94,28 @@ void checksPrintPacks(final String yaml) {
         );
     }
 
+    @Test
+    void keepsLiteralPayloadApartFromArgument() {
```
Thayorns, line 144: I do not think this test belongs in `XmirTest`. Everything else in the class goes through the `Xmir` facade and is driven by packs — `printsToEo`, `printsToParseableEo`, `checksPrintPacks`. This one never touches `Xmir`; it drives one stylesheet directly through `Xsline` and `StClasspath`.

Our rule is to map test files 1:1 with feature files, and the rest of the module already does that: `PenaltyTest`, `PenaltyKeyTest`, `StEoLoggedTest`, `StUnhexTest`. The feature under test here is `dataized-to-const.xsl`.

Your own PR description says "Test: `DataizedToConstTest` feeds `dataized-to-const.xsl` a hand-built XML fragment" — so either the description is stale or the test landed in the wrong class. The name in the description is the one I would expect to see.

eo-printer/src/test/java/org/eolang/printer/XmirTest.java
```diff
@@ -90,6 +94,28 @@ void checksPrintPacks(final String yaml) {
...
+                new XMLDocument(
+                    String.join(
+                        "",
+                        "<p><o base='.as-bytes' name='x'><o base='Φ.dataized'>",
+                        "<o base='Φ.bytes'>01-02<o base='Φ.number'>5</o></o>",
+                        "</o></o></p>"
+                    )
+                )
+            ),
+            XhtmlMatchers.hasXPath(
```
Thayorns, line 157: Two things about how much this actually guards.

First, the fragment is a shape the pipeline cannot currently produce — you say so yourself, it needs #5721 first. So this is a hypothesis about what XMIR will look like afterwards. If #5721 lands with the argument nested differently, the test keeps passing and guards nothing. A note in the test saying which issue defines the shape would at least make that dependency findable.

Second, this is about 28 hits of code, under our 40 floor, and the rule for that is to add edge cases. There are obvious ones right here and none are covered: a literal with a payload and no argument at all, which is the case that works today and is the one a careless future edit would break; a literal with an argument but no payload, which is the `xsl:if` false branch; and the `otherwise` branch at line 41, where `$argument` does not exist. All three go through the lines you touched.

Thayorns (approved):
All three points addressed.

The `@todo` on `const-to-dataized.xsl` is the right weight for the twin bug — visible without pulling the forward pass into this PR.

Giving the shift its own class so the test mapping is real is more than I asked for, and it follows the `StUnhex` envelope pattern.

The extra cases cover what I was after, including the `otherwise` branch, and the #5721 dependency is now stated where the next reader will see it.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+314](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-printer/src/main/java/org/eolang/printer/DataizedToConst.java
```diff
@@ -0,0 +1,33 @@
...
+ * attribute on the original argument.
+ * @since 0.62.0
+ */
+final class DataizedToConst extends StEnvelope {
+
+    /**
+     * Ctor.
+     */
+    DataizedToConst() {
+        this(new StClasspath("/org/eolang/printer/print/dataized-to-const.xsl"));
```
yegor256: @morphqdd what's the point of this class existence? all other XSLs stay in the list, why this one has to be ouside?

morphqdd: Removed it. The class only existed to give the test something to instantiate, which is not a good enough reason for one XSL to look different from the rest of the list. The entry in `Xmir` is a plain `StClasspath` again, and the test builds the same shift itself.

morphqdd: Follow-up: dropping the class alone broke the build. jtcop has `RuleEveryTestHasProductionClass`, so `DataizedToConstTest` could not exist once `DataizedToConst` was gone.

That leaves `XmirTest` as the only place these can live, since a pack cannot express the fragment (it is a shape the pipeline does not produce until #5721). So the tests moved back there and the standalone class is gone. @Thayorns, this undoes your first comment, but the two rules together leave no other spot.

yegor256:
@morphqdd I still believe that we can add YAML test for this bug

morphqdd:
@yegor256 added one: `eo-packs/print/dataized-to-const-keeps-argument.yaml`. It runs the parse train plus `dataized-to-const.xsl` and pins that a literal with an argument folds into a const with the payload and the argument in separate nodes, plus the payload-only case and the copy-through branch.

I have to be straight about what it does not do: it passes on master too. The exact shape from #6102 needs an element with its own text and an argument child at the same time, and no pack can build that today. Two things in the way:

- with `sheets` alone the payload stays in a child node, so `eo:has-data` is false and the buggy line is never reached
- the full `Xmir` train does give the parent its own text, but `StUnhex` drops the argument on the way (#5721), and `StUnhex` is a Java shift, so it cannot appear in a pack's `sheets`

I checked this rather than assumed it. Before `StUnhex`:

```
<o base="Φ.bytes"><o>2A-</o><o base="Φ.number">…</o></o>
```

after:

```
<o base="Φ.bytes">2A-</o>
```

Same output with and without my fix, at the EO level as well.

So the pack is a regression guard for the reachable part of the template, and the Java tests in `XmirTest` are what pin the actual bug, since they can feed the shape directly. Once #5721 lands, that shape becomes reachable and the case belongs in a pack.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (120 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2860](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Favixx (commented):
The fix is right and I could check the premise rather than take it on faith. `_funcs.xsl` defines the pair as `eo:has-data($o) = normalize-space(string-join($o/text(), '')) != ''` and `eo:read-data($o) = replace(string-join($o/text(),''), '^\s+|\s+$', '')`, so before this change the guard asked about direct text and the body took the whole element's string value — the two disagreed about what "the data" is, which is the bug in one sentence. After it they read the same nodes.

The shape the test builds is also not a hypothesis, and it is worth saying since @Thayorns raised it: `Xmir` builds its train as `new TrDefault<>(new StUnhex(), …, new StClasspath("…/dataized-to-const.xsl"), …)` — `StUnhex` is the *first* shift and `dataized-to-const.xsl` the fourth. `StUnhex` is exactly what turns a literal's nested `<o>01-02</o>` payload into direct text on the literal element, so by the time this sheet runs the payload really is direct text with the argument as a sibling. The fragment is what production hands this sheet the moment #5721 stops dropping the argument.

One thing I would fix before merge, and it is about the pack rather than the fix.

**The YAML pack does not exercise the changed line.** A pack's pipeline is its `sheets:` list, and `StUnhex` is a Java shift — it cannot appear there. Without it the payload stays nested, `eo:has-data($argument)` is false, and the `xsl:if` never runs. The pack's own asserts say so: `//o[@name='first']/o[text()='2A-']` puts the payload in a child element and `//o[@name='first' and not(text()[normalize-space()])]` states the literal has no direct text at all. So the pack is a regression test for the parser chain preserving the argument — useful, and #5721's concern — while the line this PR changes is dead inside it. Its comment claims otherwise.

The rest is detail.

eo-printer/src/main/resources/org/eolang/printer/print/dataized-to-const.xsl
Favixx, line 37: This is not a pure narrowing of `value-of select="$argument"`, and the second change is worth being deliberate about. `eo:read-data` is

```xml
replace(string-join($o/text(),''), '^\s+|\s+$', '')
```

so besides restricting the read to direct text it also **trims**. `value-of` did not. For a literal whose payload text carries surrounding whitespace — and the XMIR reaching the printer has been through several XSL passes, which is where indentation comes from — the const now gets `2A-` where it used to get whatever whitespace surrounded it.

That is almost certainly the behaviour you want, and it matches what the parser side already does. But none of the four new tests puts whitespace around a payload, so the trimming is untested, and if it ever turns out to matter the failure will look like a printing bug rather than like this line.

Favixx, line 37: Separately: the property that makes this correct is that the guard on line 36 and the read on line 37 now consult the same nodes — `$argument/text()` in both. That is the invariant, and it is the kind that is easy to undo by "simplifying" one of the two lines later, exactly as happened here in the first place.

A one-line XML comment above the `xsl:if` saying that the guard and the read must stay on the same function pair would make the next edit think twice. The forward pass in `const-to-dataized.xsl` is the proof that the pairing does not hold by itself.

eo-parser/src/main/resources/org/eolang/parser/parse/const-to-dataized.xsl
```diff
@@ -77,6 +77,19 @@
           <xsl:for-each select="o">
             <xsl:apply-templates select="."/>
           </xsl:for-each>
+          <!--
+          @todo #6102:30min Use eo:read-data(.) instead of value-of select=".".
```
Favixx, line 81: Good that this is recorded rather than silently left, but `30min` is optimistic for a reason the puzzle does not mention: the one-character change is the easy half, and this PR has just discovered that the hard half — a test — cannot be a pack. The same `StUnhex`-shaped hole applies (see my comment on the pack): no `sheets:` list can produce a literal with direct text *and* an argument, so whoever picks this up lands in the same place you did, needing a hand-built fragment and a home for it that `jtcop`'s `RuleEveryTestHasProductionClass` permits.

Two extra sentences here — "a pack cannot express the input; see `XmirTest.keepsLiteralPayloadApartFromArgument` for the shape and for where such a test can live" — would hand the next person the thirty minutes you already spent.

eo-parser/src/main/resources/org/eolang/parser/parse/const-to-dataized.xsl
```diff
@@ -77,6 +77,19 @@
...
+          value-of takes its full string value, concatenating that direct
+          text with the printed text of any "o" children (the for-each just
+          above already emitted). Once #5721 stops dropping an argument
+          applied directly to a data literal, a literal here carrying both
+          its own payload text and an argument child will glue the two
+          together on the way INTO the "Φ.dataized" wrapper, the mirror
+          image of the reverse-pass bug #6102 already fixes.
+          -->
           <xsl:if test="eo:has-data(.)">
             <xsl:value-of select="."/>
```
Favixx, line 94: The puzzle's explanation misplaces where the extra text comes from, and since the puzzle is the spec for the next change it is worth being exact.

It says the glue is "that direct text with the printed text of any `o` children (the for-each just above already emitted)". But `.` here is the **source** element, and `xsl:value-of` takes the string value of the source subtree — the `xsl:for-each` above writes into the result tree and has no bearing on what `.` evaluates to. The concatenation is between the source literal's own text and the text of its own source `o` descendants; it would happen identically if the `for-each` were deleted.

The bug and the fix are unchanged, only the sentence is wrong — and it is the sentence someone will read while deciding whether the fix is safe.

eo-printer/src/test/java/org/eolang/printer/XmirTest.java
```diff
@@ -90,6 +94,72 @@ void checksPrintPacks(final String yaml) {
         );
     }
 
+    @Test
+    void foldsAPayloadOnlyLiteralIntoAConst() {
```
Favixx, line 98: Worth recording which of these four is the regression test, because it is only one. I traced each against the old body (`value-of select="$argument"`, i.e. the element's full string value):

- this one — `<o base='Φ.bytes'>2A-</o>` has no children, so the string value is `2A-` and old and new agree: **passes on master**;
- `foldsALiteralWhoseOwnPayloadIsNested` — no direct text, `eo:has-data` is false, neither version enters the branch: **passes on master**;
- `copiesTheOriginalUnchangedWhenNoArgumentExists` — never reaches the branch: **passes on master**;
- `keepsLiteralPayloadApartFromArgument` — string value is `01-025`: **the only one that fails without the fix**.

That is a fine ratio — three characterisation tests around one regression test is exactly what @Thayorns asked for. But nothing in the code says which is which, and characterisation tests are the ones a future refactor is tempted to "fix" when they turn red. A sentence in the one that guards #6102 pointing at the issue would keep them apart.

eo-printer/src/test/java/org/eolang/printer/XmirTest.java
```diff
@@ -90,6 +94,72 @@ void checksPrintPacks(final String yaml) {
...
+        MatcherAssert.assertThat(
+            "an argument whose own text is its whole payload, with no nested value, must fold into a const carrying that same text",
+            new Xsline(new StClasspath("/org/eolang/printer/print/dataized-to-const.xsl")).pass(
```
Favixx, line 101: `new Xsline(new StClasspath("/org/eolang/printer/print/dataized-to-const.xsl"))` appears verbatim four times in this class, in a file where no other test names a stylesheet at all — everything else goes through `Xmir`. A `private static final Xsline SHEET` next to the other fields, or a small `private static XML folded(final String xml)` helper, removes the repetition and gives the four tests one obvious thing in common, which is currently only visible by reading all four.

eo-printer/src/test/java/org/eolang/printer/XmirTest.java
```diff
@@ -90,6 +94,72 @@ void checksPrintPacks(final String yaml) {
...
+                new XMLDocument(
+                    "<p><o base='.as-bytes'><o base='Φ.dataized'><o base='Φ.bytes'>2A-</o></o></o></p>"
+                )
+            ),
+            XhtmlMatchers.hasXPath("//o[@base='Φ.bytes' and @const and text()='2A-']")
+        );
+    }
+
+    @Test
+    void foldsALiteralWhoseOwnPayloadIsNested() {
```
Favixx, line 111: The `<p>` root in all four fixtures is not an XMIR element — it is here because a document needs one root. Everything else in this class reaches XMIR through `asXmir(...)`, so these are the only fixtures in the file whose shape a reader cannot check against the real thing.

`<object>` is the real root and costs the same number of characters. It also makes the fixtures pasteable into a pack or into an XMIR file when someone wants to reproduce by hand, which for a bug that "cannot be produced by the pipeline yet" is worth more than usual.

eo-printer/src/test/java/org/eolang/printer/XmirTest.java
```diff
@@ -90,6 +94,72 @@ void checksPrintPacks(final String yaml) {
...
+                )
+            ),
+            XhtmlMatchers.hasXPath(
+                "//o[@base='Φ.number' and @const and not(text()[normalize-space()]) and o[@base='Φ.bytes' and text()='2A-']]"
+            )
+        );
+    }
+
+    @Test
+    void copiesTheOriginalUnchangedWhenNoArgumentExists() {
```
Favixx, line 131: This covers the fall-through with an *empty* `Φ.dataized`, which is the degenerate case. The interesting non-matching input is a wrapper that has a child which is not foldable — that is the shape the `xsl:otherwise` exists for, and the one where a careless tightening of the `xsl:when` predicate would start swallowing nodes.

Since the point of the trio around this test is to protect the untouched paths, one wrapper with a real child that must be copied through verbatim would be worth more than the empty one — or in addition to it, they are three lines each.

eo-printer/src/test/java/org/eolang/printer/XmirTest.java
```diff
@@ -90,6 +94,72 @@ void checksPrintPacks(final String yaml) {
...
+                new XMLDocument(
+                    "<p><o base='.as-bytes'><o base='Φ.dataized'/></o></p>"
+                )
+            ),
+            XhtmlMatchers.hasXPath("//o[@base='.as-bytes']/o[@base='Φ.dataized' and not(node())]")
+        );
+    }
+
+    @Test
+    void keepsLiteralPayloadApartFromArgument() {
```
Favixx, line 144: The assertion message calls this "the shape #5721 would produce, not one the current pipeline can build yet", which reads as an assumption when it is actually checkable and true: `Xmir` puts `new StUnhex()` first in the train and `dataized-to-const.xsl` fourth, and `StUnhex` is precisely the shift that replaces a literal's nested payload element with direct text while leaving siblings alone. So the direct-text-plus-argument shape is what this sheet is handed in production the moment #5721 stops deleting the argument.

Saying that — "`Xmir` runs `StUnhex` before this sheet, so after #5721 the payload arrives as direct text next to the argument" — turns the weakest sentence in the PR into its strongest, and answers the objection permanently rather than in a thread.

eo-printer/src/test/java/org/eolang/printer/XmirTest.java
```diff
@@ -90,6 +94,72 @@ void checksPrintPacks(final String yaml) {
...
+                    String.join(
+                        "",
+                        "<p><o base='.as-bytes' name='x'><o base='Φ.dataized'>",
+                        "<o base='Φ.bytes'>01-02<o base='Φ.number'>5</o></o>",
+                        "</o></o></p>"
+                    )
+                )
+            ),
+            XhtmlMatchers.hasXPath(
+                "//o[@base='Φ.bytes' and @const and text()='01-02' and o[text()='5']]"
```
Favixx, line 158: Two small strengthenings for the one test that actually guards the fix.

`text()='01-02'` is existential — true if *any* direct text child equals it. Here the bug produces a single glued `01-025` so it does discriminate, but `count(text())=1 and text()='01-02'` says the thing the fix establishes: one text node, holding the payload and nothing else.

And since this is the regression test, an explicit `not(//o[text()='01-025'])` would make a failure name the bug. As written, a regression prints "xpath not found", and the next person has to reconstruct from the issue why `01-025` was ever a possibility.

eo-printer/src/test/resources/org/eolang/printer/eo-packs/print/dataized-to-const-keeps-argument.yaml
```diff
@@ -0,0 +1,35 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets:
```
Favixx, line 5: This list is the pack's whole pipeline, and the one shift that matters for #6102 cannot be in it: `StUnhex` is a Java `Shift`, not an XSL sheet, and `Xmir` runs it first — before `dataized-to-const.xsl`. Without it the payload of `2A- 5` is still a nested `<o>2A-</o>`, so `eo:has-data($argument)` is false when this sheet runs and the line this PR changes is never executed by this pack.

The asserts below confirm it rather than contradict it: line 24 looks for the payload in a child element, and line 26 states the literal has no direct text at all — which is the pre-`StUnhex` shape.

The pack is still worth having: it pins that the parser chain carries the argument through `const-to-dataized.xsl` into the `Φ.dataized` wrapper, which is a real thing to protect. It just protects #5721's half, not this PR's, and the file name (`dataized-to-const-keeps-argument`) promises the latter.

eo-printer/src/test/resources/org/eolang/printer/eo-packs/print/dataized-to-const-keeps-argument.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - /org/eolang/parser/parse/stars-to-tuples.xsl
+  - /org/eolang/parser/parse/vars-float-up.xsl
+  - /org/eolang/parser/parse/build-fqns.xsl
+  - /org/eolang/parser/parse/expand-aliases.xsl
+  - /org/eolang/parser/parse/resolve-aliases.xsl
+  - /org/eolang/parser/parse/add-default-package.xsl
+  - /org/eolang/parser/parse/roll-bases.xsl
+  - /org/eolang/printer/print/dataized-to-const.xsl
+asserts:
+  # A literal carrying an argument folds into a const, and the two stay
```
Favixx, line 20: "with no node holding both texts joined together" is true here for a reason that has nothing to do with the fix: the payload never became direct text in this pipeline, so there was never anything to join. The comment describes the property the unit test verifies, on a pack that cannot reach it.

What this pack does verify is worth stating plainly instead — that `const-to-dataized.xsl` wraps a literal-with-argument into `Φ.dataized` while keeping the argument as its own element. A reader who takes the current comment at face value will believe #6102 has end-to-end coverage.

eo-printer/src/test/resources/org/eolang/printer/eo-packs/print/dataized-to-const-keeps-argument.yaml
```diff
@@ -0,0 +1,35 @@
...
+  # separate nodes: the payload keeps its own element and the argument keeps
+  # its own, with no node holding both texts joined together.
+  - //o[@base='Φ.bytes' and @const and @name='first']
+  - //o[@name='first']/o[text()='2A-']
+  - //o[@name='first']/o[@base='Φ.number']/o[@base='Φ.bytes']/o[text()='40-14-00-00-00-00-00-00']
+  - //o[@name='first' and not(text()[normalize-space()])]
```
Favixx, line 26: This is the assert that *would* catch the glue — a literal with its payload and its argument fused would leave direct text on `first` and fail here. It is the right assert to have written.

It is also currently vacuous, for the reason in my comment on the `sheets:` list: at this point in the pipeline the literal has no direct text under any behaviour, buggy or fixed. Keeping it is fine — it becomes meaningful the day a Java shift can be named in a pack, or the day the parser starts emitting direct text — but it should not be counted as coverage of #6102 today.

PR #6111, opened 2026-07-30
#5955: add eo.phiDefaultClass option to substitute the Phi base class

### LINKED ISSUE #5955: `Phi` attribute listing for `eoc inspect`: `PhPackage` roots don't know their full child set
Working on `eoc inspect` (objectionary/eoc#500) needs a way to list the attribute names of a live `Phi` object, for the `ls` command (`obj.getAttributes()` or similar, per the discussion on that issue).

For a normal object built by `PhDefault`, this is straightforward: it already holds a materialized `Map<String, Attribute> attrs`, so exposing its key set (minus the special `Φ`/`ρ`/`λ` names) is a small addition.

`PhPackage` (the namespace root objects, `Φ`, `Φ.org`, `Φ.org.eolang`, and so on, which is where `eoc inspect` starts) is a different story. It only learns about a child object once that child has actually been requested through `.take(name)` (cached in its own `objects` map), or by testing one specific candidate name against the classpath via `Class.forName(...)`. It has no way to answer "list everyone who lives under me" without either:

1. Scanning the classpath (jar entries or `.class` files) under the corresponding Java package prefix to discover what could be loaded there, or
2. Only reporting whatever has already been cached/visited so far, which would make `ls` at the root show nothing useful before any navigation has happened.

Option 1 is not quite the same thing as the Reflection API objection raised on #500 (that was about inspecting an already-loaded object's internal state through reflection; this would be enumerating classpath resources to discover what packages/objects exist), but it is in the same neighborhood and worth confirming before anyone builds on it.

What should the `ls`/attribute-listing behavior be for a `PhPackage` root, specifically? Classpath scanning, cached-only, something else? Once that is settled, I would like to add the resulting method to `Phi` (or wherever it should live) as the first slice for objectionary/eoc#500.

morphqdd:
Closes #5955

Adds the `eo.phiDefaultClass` option to `MjTranspile`, as suggested in https://github.com/objectionary/eo/issues/5955#issuecomment-5089369060. It defaults to `PhDefault`, so nothing changes unless it is set.

The value is threaded down to `to-java.xsl` the same way `trackLocations` and `coverage` already are, and it is folded into the transpile cache key, so switching it does not serve stale Java from an earlier build.

Only the three places that emit an `extends` clause read the option, which is the part the issue points at. Three more still hardcode `new PhDefault(...)`; there is a `@todo` on the new field for that, and another one for the eoc side that is expected to set this option to its own `PhInspected` (objectionary/eoc#500).

The value goes into the generated Java verbatim, so a class outside `org.eolang` has to be named in full, since the generated files import nothing but `org.eolang.*`.


morphqdd:
@yegor256, please review

eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
```diff
@@ -95,6 +95,35 @@ public final class MjTranspile extends MjSafe {
...
+    /**
+     * The name of the class that every generated class extends, instead of
+     * {@code PhDefault}, set through {@code eo.phiDefaultClass}. A tool that
+     * needs control over the objects of a program can put its own subclass of
+     * {@code PhDefault} here and see every {@code add()} and {@code take()}
+     * the program makes. The value is written into the generated Java
+     * verbatim, so a class outside {@code org.eolang} has to be named in
+     * full, since the generated files import nothing but
+     * {@code org.eolang.*}.
+     * @todo #5955:60min Let eoc supply its own Phi base class here.
```
yegor256: @morphqdd how this problem/puzzle is connected to this repo? I suggest removing it

Thayorns (commented):
updating..

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -199,7 +204,8 @@ final class Transpiling implements Step {
      * @param measures Path to the file where XSL measurements are stored
      * @param diagnostics Which diagnostic artifacts to emit while transpiling
      * @param cvrg Whether located objects are wrapped into {@code PhCoverage}
-     * @checkstyle ParameterNumberCheck (18 lines)
+     * @param base The name of the class that every generated class extends
+     * @checkstyle ParameterNumberCheck (20 lines)
```
Thayorns, line 209: This constructor is up to 11 parameters now, and the class to 12 attributes. Our rule is 1-4. Both the `ParameterNumberCheck` span and the PMD suppression had to grow again to let it through.

What stands out is that the fix is already sitting in this same parameter list: `Tracking` exists to bundle two related flags. `coverage` and this new value look like they belong in something like that. A `String` in slot 11, right after a `boolean`, is easy to swap by accident at a call site and the compiler will not notice.

This is not your mess, you inherited it. But this is the change that makes it longer, so I think it deserves an issue rather than a fix here.

eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
```diff
@@ -95,6 +95,35 @@ public final class MjTranspile extends MjSafe {
...
+     *  Right now {@code to-java.xsl} reads it only where it emits an
+     *  {@code extends} clause, which is what the issue asked for. Three
+     *  more places still hardcode {@code new PhDefault(...)}: the context
+     *  of a generated {@code apply()}, an anonymous abstract object with no
+     *  children, and the argument of a {@code PhApplication}. Objects built
+     *  there stay outside the substituted class, so a debugger that relies
+     *  on this option sees only a part of the tree. Convert them the same
+     *  way and assert on them in {@code MjTranspileTest}.
+     */
+    @Parameter(property = "eo.phiDefaultClass", defaultValue = "PhDefault")
```
Thayorns, line 132: Nothing checks this value before it reaches the generated Java. Pass `-Deo.phiDefaultClass=` with an empty value and Maven's `defaultValue` does not kick in, so `to-java.xsl` writes `extends  {` and you get a javac error inside generated sources, far from the option that caused it. A misspelled class name does the same.

The PR text says the value goes in verbatim, which is fair for the fully-qualified-name part. But we prefer to fail fast, and a value that cannot compile seems worth rejecting here instead of leaving it to javac.

eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
```diff
@@ -275,7 +276,7 @@
         <xsl:text> extends PhOnce {</xsl:text>
       </xsl:when>
       <xsl:otherwise>
-        <xsl:text> extends PhDefault {</xsl:text>
+        <xsl:value-of select="concat(' extends ', $phiDefaultClass, ' {')"/>
```
Thayorns, line 279: The branch just above still writes `extends PhOnce` for objects with `@base`, and the same pair shows up again near line 751 for the `Test` class. Those classes never get the substituted base.

That is probably fine, since `PhOnce` is a decorator and the object it wraps does get substituted. But it means "the class that every generated class extends" in the javadoc here and on `Transpiling.superclass` is not quite true, and the `@todo` lists the three `new PhDefault(...)` sites without mentioning this one. The whole point is a debugger seeing every `add()` and `take()`, so it is worth saying exactly how far the substitution reaches.

eo-maven-plugin/src/test/java/org/eolang/maven/MjTranspileTest.java
```diff
@@ -130,6 +130,60 @@ void keepsGeneratedJavaFreeOfPhCoverageByDefault(@Mktmp final Path temp) throws
...
+    @Test
+    void extendsPhDefaultInGeneratedJavaByDefault(@Mktmp final Path temp) throws Exception {
+        MatcherAssert.assertThat(
+            "the generated class must extend PhDefault when phiDefaultClass is not set",
+            new TextOf(
+                new FakeMaven(temp)
+                    .withProgram(String.format("+package foo.x%n%n[] > main%n  42 > @"))
+                    .execute(new FakeMaven.Transpile())
+                    .result()
+                    .get(this.compiled)
```
Thayorns: Not a problem on this branch — `this.compiled` is correct here. Just a heads-up that master dropped that field in `6d649df0` ("test: isolate tests from shared fields and lifecycle hooks") and uses a static accessor instead. Your branch is older than that, so these three tests will stop compiling after a rebase.

Worth mentioning because the new tests sit between master's edits to this file, so git will probably merge them without a conflict and you will only find out when it fails to compile.

Thayorns (commented):
The feature looks right to me. The option follows the same path `trackLocations` and `coverage` already take, all three `extends` sites in `to-java.xsl` are covered, and putting it in `version()` matters — without that the cache would serve Java built against the old base class. The casts and the `PhDefault x = new ...()` line still compile fine, as long as the named class extends `PhDefault`.

I also checked that the cache test actually earns its place: if `version()` did not include this value, both runs would hit the same cache entry and the second one would come back with `PhDefault`. So it fails without the fix, which is what you want.

Four notes inline. The first one, about the parameter list in `Transpiling`, is the only one I would like an answer on. The rest are small.

One thing not worth its own thread: `superclass` is the only `@Parameter` field in `MjTranspile` whose name does not match its property (`eo.phiDefaultClass`). Every other field in that class does match. On the other hand `superclass` is one noun and `phiDefaultClass` is three, so it reads better by our own rules. Call it even — just checking the mismatch was on purpose.

eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
morphqdd: Removed. It described work that belongs to eoc, so 0pdd would have filed the issue in the wrong repo.

eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
morphqdd: Added a check in `MjTranspile`. An empty or malformed value now fails with a message naming `eo.phiDefaultClass` and the expected pattern, before anything is generated. Two tests cover it.

The check returns the name rather than validating it in a separate statement, so the value cannot be used without passing through it.

It is syntax only. A well-formed name that does not exist still reaches javac, since the class sits on the project's compile classpath, not the plugin's.

eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
morphqdd: Reworded. The javadoc now says the substitution reaches classes that get an `extends` clause of their own, and spells out that an object with `@base` extends `PhOnce` and keeps doing so, since `PhOnce` decorates an object that is itself substituted. Dropped "every generated class" from both places.

eo-maven-plugin/src/test/java/org/eolang/maven/MjTranspileTest.java
morphqdd: Good catch, thanks. Rebased and switched the three tests to `MjTranspileTest.compiled()`. It merged clean and broke the compile exactly as you said.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
morphqdd: Agreed, and agreed it should not be fixed here. Bundling `coverage` with this value would change the constructor for reasons unrelated to the issue.

One correction: the PMD suppression did not grow, it was already on the constructor. The only thing my change touched is the `ParameterNumberCheck` span, 18 to 20 lines, because the javadoc block got one more `@param`.

The swap risk you describe is real though. `String` in slot 11 sits right after a `boolean`, and slot 6 is also a `String`, so the two can be exchanged silently.

eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
morphqdd: Correction to my earlier reply: I had only fixed `MjTranspile` at that point. `Transpiling` still carried the same wording in three places, the field and two `@param` lines. Fixed now in fcabb41.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
morphqdd: Filed as #6125.

Thayorns (approved):
All four points addressed, thanks.

One correction owed to you: I said both the checkstyle span and the PMD suppression had grown. The suppression was already on the constructor and your change did not touch it. My mistake.

#6125 is the right home for the parameter list. Rejecting the value where it is read, with the option named, is better than what I asked for.

Small non-blocking note: asserting the stack trace contains `eo.phiDefaultClass` is testing an error message, which our test rules avoid. It matches `throwsDetailedError` below it, and the exception type alone would not tell this apart from any other `IllegalStateException` here, so fine as it stands.

0crat:
@thayorns Thanks for the review! You've earned +12 points.  Your running score is [+307](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (190 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2740](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6110, opened 2026-07-30
`Globals.appendTextLine` incorrectly maintains whitespace on under-indented blank lines


yegor256:
`Globals.appendTextLine` (`eo-parser/src/main/java/org/eolang/parser/Globals.java:247-257`) only strips a text-block body line when the line's length is at least the opener's indent (`tindent`):

```java
void appendTextLine(final String raw) {
    final String stripped;
    if (raw.length() >= this.tindent
        && raw.substring(0, Math.min(this.tindent, raw.length()))
            .chars().allMatch(c -> c == ' ')) {
        stripped = raw.substring(Math.min(this.tindent, raw.length()));
    } else {
        stripped = raw;
    }
    this.tbody.add(stripped);
}
```

A blank body line that is shorter than `tindent` — e.g. one or two leading spaces where the block opened at indent 4 — fails the `raw.length() >= this.tindent` guard and falls into the `else` branch, so it is appended to `tbody` completely unstripped. Its caller, `continueTextBlock` in `Eo.java:333-354`, only raises a parse error for *non-blank* under-indented lines (`!Eo.isBlank(raw) && Eo.leadingSpaces(raw) < globals.textBlockOpenIndent()`, `Eo.java:347`), so a short blank line reaches `appendTextLine` silently, with no error and no correct stripping.

This is wrong because a text-block body line that is entirely whitespace should always collapse to an empty line, regardless of how many spaces it happens to carry relative to the opener's indent. Instead, any stray spaces on such a line survive into the assembled string literal as a spurious byte in the middle of otherwise-clean content. The project's own fixture, `eo-parser/src/test/resources/org/eolang/parser/eo-syntax/text-block.yaml`, only exercises a *fully-empty* (zero-length) blank line, which happens to take the same `else` branch but is invisible there because an empty string stripped or not is still empty — so the fixture never caught the case of a blank line 1..`tindent`-1 characters long. This is a distinct code path from issue #5790 (`LnTextBlock`'s whole-string `.trim()`, fixed by removing the trim in commit `ea616f78b1`); that fix never touched `appendTextLine`, so this defect survived it.

Suggested fix: in `appendTextLine`, check `raw.chars().allMatch(c -> c == ' ')` (or `raw.isBlank()`) first and set `stripped = ""` whenever it holds, falling back to the existing length/prefix check only for non-blank lines.

h1alexbel:
@yegor256 let me investigate this

PR #6115, opened 2026-07-30
fix(#6110): `Globals.appendTextLine()` strips blank indents

h1alexbel:
In this PR I've updated `Globals.appendTextLine()` to strip blank indents in the text blocks.

see #6110


h1alexbel:
@yegor256 take a look, please

eo-parser/src/main/java/org/eolang/parser/Globals.java
```diff
@@ -245,15 +245,13 @@ void closeTextBlock() {
...
     void appendTextLine(final String raw) {
-        final String stripped;
-        if (raw.length() >= this.tindent
-            && raw.substring(0, Math.min(this.tindent, raw.length()))
-                .chars().allMatch(c -> c == ' ')) {
-            stripped = raw.substring(Math.min(this.tindent, raw.length()));
-        } else {
-            stripped = raw;
+        int strip = 0;
+        while (strip < raw.length()
```
Thayorns, line 249: For a properly indented line this loop and the old branch agree exactly. They part company when a line carries fewer leading spaces than the block's indent. Taking a block indented by four:

- A line of just two spaces. Before it stayed as two spaces, now it comes out empty. This is the fix, and it is right.
- A line reading `  x`, two spaces then a letter. Before it stayed `  x`, now it comes out `x`.
- A line indented with two spaces then a tab. Before it survived untouched, since the first four characters were not all spaces; now the two spaces come off.

The second and third are content lines rather than blank ones, and they used to pass through whole. Losing their leading spaces changes the text a program reads out of the block.

I can see the argument that this is the better behaviour — a line indented less than its own block is malformed whichever way you treat it, and uniform handling beats a special case. If that is the call you made, I am all for it. My worry is only that nothing in the code says so: the PR reads as being about blank lines, and there is no test on either side of the question, so the next person to touch this method has no way to tell the new handling from an accident.

One small related thing: the docblock above still says it strips "the opener's indent prefix where present", which was exact when the prefix came off whole or not at all. Worth a line if the contract is now a partial strip.

eo-parser/src/test/java/org/eolang/parser/LnTextBlockTest.java
```diff
@@ -143,6 +143,18 @@ void rejectsInvalidEscapeInTextBlockBody() {
         );
     }
 
+    @Test
+    void stripsBlankIndent() {
```
Thayorns, line 147: I suspect this is just where it was handiest while you were iterating, but the test never goes through `LnTextBlock` — it constructs a `Globals`, calls two of its methods and reads its buffer, so it is a `Globals` test living in someone else's class.

`GlobalsTest` is already set up for precisely this: `recordsTextBlockOpenLine` calls `globals.openTextBlock(17)`, `closesTextBlockState` calls `globals.closeTextBlock()`, both asserting on the state afterwards. This one would sit naturally next to them and keep the 1:1 mapping intact.

eo-parser/src/test/java/org/eolang/parser/LnTextBlockTest.java
```diff
@@ -143,6 +143,18 @@ void rejectsInvalidEscapeInTextBlockBody() {
...
+        final Globals globals = new Globals();
+        globals.openTextBlock(1, 4);
+        globals.appendTextLine("  ");
+        MatcherAssert.assertThat(
+            "Blank indent must be stripped",
+            globals.tbody().get(0),
```
Thayorns, line 153: This pins the blank line, which is the case the PR is named for. Since the rewrite moves three situations at once, a few more would make the intended contract explicit rather than implied. All in a block indented by four:

- `    hello`, the ordinary line, so a later rewrite of this method cannot break the everyday case silently
- `      hello`, indented past the block, which has to keep its two extra spaces
- `  x`, under-indented but not blank — one of the two cases the behaviour moves on
- a line indented with spaces then a tab — the other

The last two are the ones I would most want written down, since they are what turns the decision in the comment above into something the code states outright.

Minor: `globals.tbody().get(0)` reads the first element only, so it would stay green if the method ever appended two lines. Asserting on the list would cover that too.

Thayorns (commented):
Nice catch on this one — the diagnosis was the hard part. The old code asked a single yes/no question, does this line start with the full indent in spaces, and left the line completely alone when the answer was no. A line holding just two spaces inside a block indented by four failed that question, so it kept its spaces instead of coming out empty. Replacing that with a count of leading spaces up to the indent width is the right shape of fix, and it reads better than the branch it replaces.

Two things I would like your read on. The new question also answers differently for content lines that are under-indented, which I do not think the title is claiming — I have written out the cases on `Globals.java`, and you may well have already decided that is the behaviour you want, in which case it is really the tests I am asking about. Separately, the test looks like it landed in `LnTextBlockTest` when it is exercising `Globals`.

yegor256:
@h1alexbel maybe you can add YAML test? otherwise, it's unclear what kind of error you are fixing

eo-parser/src/main/java/org/eolang/parser/Globals.java
```diff
@@ -245,15 +245,13 @@ void closeTextBlock() {
...
      */
     void appendTextLine(final String raw) {
-        final String stripped;
-        if (raw.length() >= this.tindent
-            && raw.substring(0, Math.min(this.tindent, raw.length()))
-                .chars().allMatch(c -> c == ' ')) {
-            stripped = raw.substring(Math.min(this.tindent, raw.length()));
-        } else {
-            stripped = raw;
+        int strip = 0;
```
morphqdd, line 248: This strips up to `tindent` leading spaces even when the line is shorter than `tindent`. The old code only stripped when the whole prefix (all `tindent` chars) was spaces, otherwise left the line untouched. Now a short line like one leading space with `tindent=4` gets that space stripped too, changing output for any line indented less than the block's marker. Only the fully-blank case is tested.

morphqdd (changes requested):

yegor256:

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (3) comments; -4 for too few (25) hits-of-code; +1 to give you at least something.  Your running score is [+4458](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (25) hits-of-code; +1 to give you at least something.  Your running score is [+6548](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6120, opened 2026-07-30
Align the declared Java version with the one the build actually needs

### LINKED ISSUE #6114: `pom.xml` enforcer rule incompatibility with JDK 11-18 in `eo-maven-plugin` test due to `ScheduledExecutorService` usage
The test at `eo-maven-plugin/src/test/java/org/eolang/maven/SavedTest.java:76` opens a try-with-resources over a `ScheduledExecutorService`. That type only became `AutoCloseable` in Java 19, so `javac` rejects the statement on any earlier JDK. The enforcer rule at `pom.xml:400` still accepts `[11,)`, and `README.md:339` tells contributors that Java 11+ is all they need.

The reason the mismatch survives is the compiler configuration at `pom.xml:411-412`, which sets `source` and `target` to 11 instead of `release`. With `source`/`target` the compiler checks the language level but still resolves types against the class library of the JDK that is running, so a Java 11 or Java 17 build passes the enforcer, gets through `eo-parser` and `eo-printer`, and only dies in the test compilation of `eo-maven-plugin`. Nothing in CI covers this: the matrix in `.github/workflows/mvn.yml:18` is `java: [26]`, and `qulice.yml` pins 26 as well, so every supported-on-paper JDK below 19 is untested.

Closing the scheduler explicitly instead of through try-with-resources at `SavedTest.java:76` would restore the Java 11 baseline the project claims. If Java 19+ is the real requirement, then `pom.xml:400` and `README.md:339` should say so.

Thayorns:
`SavedTest` opened a try-with-resources over a `ScheduledExecutorService`. That type became `AutoCloseable` only in Java 19, so test compilation of `eo-maven-plugin` failed on every older JDK. The scheduler is now shut down explicitly in a `finally` block, which removes the accidental Java 19 requirement.

While checking what the floor really is, it turned out that neither of the two options in the issue is right. Java 11 was already unreachable: cactoos 0.61.1 ships class-file 61 (Java 17) and is a compile-scope dependency of every module, so a clean Java 11 build dies in `eo-parser` long before it reaches `eo-maven-plugin`. Java 19 is not needed once the test is fixed. The real floor is 17, which is also what `sonar.yml` and `eo-maven-plugin-site.yml` already build on. So the enforcer, the compiler and `eo-maven-plugin/README.md` now say 17. `README.md` documents `mvn clean install -Pqulice`, and qulice 0.28.3 declares Java 21 as a prerequisite, so that file states 21 for the full build and 17 for a build without the profile.

The mismatch survived because the compiler was configured with `source` and `target` rather than `release`. That pair checks the language level but still resolves types against the class library of the running JDK, so a too-new API slips through until some later module happens to fail. Switching to `release` makes this kind of breakage fail fast on any JDK. The fibonacci integration test carried the same `source`/`target` pair and gets the same treatment.

Verified on a local JDK 17: the build reproduced the reported failure before the change and compiles cleanly after it, and `SavedTest` passes. I could not run qulice locally because it wants Java 21.

Three edge cases now cover `Saved` behaviour that had no tests: creation of missing parent directories, replacement of existing content, and removal of the temporary file after a successful save.

One follow-up worth considering: with `release` in place, a post-17 API now fails on every build, so the CI matrix does not need a lower-JDK entry to catch this class of problem. If you would rather keep publishing Java 11 bytecode for downstream consumers, `release` is a one-line change back, though that promise does not really hold while cactoos needs 17 at runtime.

Closes #6114

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,7 +336,9 @@ To avoid frustration, before sending us your pull request
 mvn clean install -Pqulice
 ```
 
-You will need [Maven] 3.3+ and [Java] 11+ installed.
+You will need [Maven] 3.3+ and [Java] 21+ installed,
+  because the `qulice` profile requires it; a build without
+  that profile needs [Java] 17+.
 Also, if you have [xcop] installed, make sure it is version `0.8.0`+.
 
 ## Contributors
--- a/eo-integration-tests/src/it/fibonacci/pom.xml
+++ b/eo-integration-tests/src/it/fibonacci/pom.xml
@@ -23,8 +23,7 @@
     </dependency>
   </dependencies>
   <properties>
-    <maven.compiler.source>11</maven.compiler.source>
-    <maven.compiler.target>11</maven.compiler.target>
+    <maven.compiler.release>17</maven.compiler.release>
   </properties>
   <build>
     <plugins>
--- a/eo-maven-plugin/README.md
+++ b/eo-maven-plugin/README.md
@@ -76,7 +76,7 @@ Then, you just run `mvn clean test`
 (you will need [Maven 3.3+](https://maven.apache.org/))
 and the `.eo` file will be parsed to `.xml` files, transformed to `.java`
 files, and then compiled to `.class` files. You can see them all in the
-`target` directory. You will need Java 11+.
+`target` directory. You will need Java 17+.
 
 The complete
 [maven plugin documentation](https://www.eolang.org/eo-maven-plugin/index.html)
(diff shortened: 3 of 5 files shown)


Thayorns:
@yegor256 take a look

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+330](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6126, opened 2026-07-30
Stop randomizing the move retry delay in Saved

### LINKED ISSUE #6109: Intermittent CI failures in SavedTest.retriesMoveWhenTargetTemporarilyBlocked due to dependency on wall-clock timing
The test `SavedTest.retriesMoveWhenTargetTemporarilyBlocked` is non-deterministic. It fails intermittently on CI with an `AccessDeniedException`, while the production code it exercises (`Saved`) is actually correct. The same commit fails and then passes with no source change, which is the signature of a flaky test.

## Evidence

Same branch, same tree, one run fails and the next passes:

- Failing run (windows-2022): https://github.com/objectionary/eo/actions/runs/30482639748/job/90680354965
- Passing run (windows-2022), identical source: https://github.com/objectionary/eo/actions/runs/30488987828/job/90701814279

The failure surfaces as:

```
java.io.IOException: Failed while trying to save to ...\SavedTest\retriesMoveWhenTargetTemporarilyBlocked\...\blocked.txt
Caused by: java.nio.file.AccessDeniedException: ...\blocked.txt<random>.tmp -> ...\blocked.txt
```

Run in isolation the test is robust: I ran it 10 times in a row locally on Windows and it passed 10/10 (each run takes about 1.4s, i.e. the first move attempt fails and the retry one second later succeeds).

## Why it is flaky

`eo-maven-plugin/src/test/java/org/eolang/maven/SavedTest.java`, method `retriesMoveWhenTargetTemporarilyBlocked`:

1. It creates a directory at the target path `blocked.txt`, which blocks the move.
2. It schedules that directory's deletion for a fixed `300` milliseconds later on a single-thread scheduler.
3. It then calls `new Saved("content", target).value()` and expects the write to succeed once a retry lands after the obstruction clears.

`Saved.moved(...)` (in `Saved.java`) is annotated `@RetryOnFailure(delay = 1L, unit = TimeUnit.SECONDS)`, so with the default attempt count it retries the move a few times about one second apart, roughly a 2 second window.

The test couples its correctness to wall-clock timing: it bets that a retry lands after the 300ms deletion but within that ~2s retry window. On CI, the whole `eo-maven-plugin` test suite runs with heavy parallelism, so the single-thread `ScheduledExecutorService` (and the retry threads) can be starved. When the scheduled deletion slips far enough, the blocking directory is still present across every retry attempt, the retries are exhausted, and the final `AccessDeniedException` propagates and fails the test. Unloaded, the timing margins hold and it passes, which is exactly why it is intermittent.

Note that `Saved` itself is not at fault here: its retry-on-`AccessDeniedException` logic is correct and passes deterministically in isolation. The defect is in the test's reliance on timing rather than synchronization.

## Suggested fix

Remove the wall-clock dependency and make the clearing of the obstruction deterministic with respect to the retries, for example:

- Clear the obstruction from a `CountDownLatch` (or similar) that is released only after the first failed move attempt is observed, instead of after a fixed `300ms`, so the retry provably runs both before and after the block clears regardless of load; or
- Drive `Saved` through an injectable/controllable clock so the retry timing is deterministic in the test.

Either approach keeps the intent (a retry succeeds once the obstruction clears) without betting on real time.

yegor256:
Sets `randomize = false` on the `@RetryOnFailure` that guards `Saved.moved`, so the backoff between move attempts is a deterministic 1s then 2s instead of a random multiple of one second.

The flake in `SavedTest.retriesMoveWhenTargetTemporarilyBlocked` was not thread starvation. With `randomize` at its default `true`, jcabi computes the wait as `nextInt(2 << attempt) * delay`, so the first wait is `nextInt(4)` seconds and the second is `nextInt(8)` seconds — and both can come out zero. Roughly once in 32 runs all three attempts fire back to back with no waiting at all, which is why the failing runs died in 0.402s on macOS and 0.625s on Windows, both well under a single one-second delay. A probe driving the same scenario through the woven class failed 2 of 120 times before this change and 0 of 120 after, with every run now taking a flat ~1.0s. The test itself is untouched: its 300ms obstruction now always clears before the second attempt.

Two things a reviewer may want to weigh separately. A randomized backoff collapsing to no delay looks like an upstream defect in jcabi-aspects and could be worth an issue there. The same annotation defaults also sit on the network retries in `OyRemote`, `ObjectsIndex`, `CommitHashesText` and `RtCentral`, where a zero delay makes the retry useless rather than flaky — I left those alone to keep this change small.

Closes #6109

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
@@ -139,11 +139,15 @@ final class Saved implements Scalar<Path> {
      * POSIX which allows it. The window is brief, so a short retry clears
      * it instead of failing the whole write.</p>
      *
+     * <p>The delay is not randomized (#6109): jcabi multiplies it by a
+     * random factor that may come out zero, and then all the attempts fire
+     * back to back with no waiting at all, which defeats the retry.</p>
+     *
      * @param tmp Temp file to move
      * @param target Destination path
      * @throws IOException If the move fails
      */
-    @RetryOnFailure(delay = 1L, unit = TimeUnit.SECONDS)
+    @RetryOnFailure(delay = 1L, unit = TimeUnit.SECONDS, randomize = false)
     private static void moved(final Path tmp, final Path target) throws IOException {
         try {
             Files.move(


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (6) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2823](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6127, opened 2026-07-30
#6113: exclude SyscallTest from simian, the name it has after the rename

### LINKED ISSUE #6113: simian check failure on master due to outdated exclude pattern after rename of EOsocketTest to SyscallTest
## What happens

The `simian` check is red on `master`. It fails for every push and for every pull request rebased onto the current tip, so the whole repository CI is blocked on it.

Failing master run (tip `841cf62`): https://github.com/objectionary/eo/actions/runs/30530015530/job/90829858030

```
Found 18 duplicate lines with fingerprint cd494dc9e30642ec798cfeb631fd6f50 in the following files:
 Between lines 318 and 335 in .../eo-runtime/src/test/java/org/eolang/SyscallTest.java
 Between lines 647 and 664 in .../eo-runtime/src/test/java/org/eolang/SyscallTest.java
Found 36 duplicate lines in 2 blocks in 1 files
##[error]Process completed with exit code 1.
```

The two blocks are the near-identical assertion sequences (`server.join()` followed by the two `assertThat` on `received.get()` and the received bytes) inside the two `acceptsConnectionOnSocket()` methods, one in the POSIX nested class and one in the Windows nested class.

## Root cause

This duplication is not new and it was never meant to fail the build. The simian workflow deliberately excludes this test file:

`.github/workflows/simian.yml` line 24:

```
- run: java -jar /tmp/simian.jar -threshold=15 "-excludes=**/EOsocketTest.java" "-excludes=**/gen" "-excludes=**/it" "**/*.java"
```

Commit `700a5612e6` ("test: name eo-runtime tests after their production classes") renamed `EOsocketTest.java` to `SyscallTest.java` (its own message says "EOsocketTest becomes SyscallTest"). The exclude pattern still points at the old name `EOsocketTest.java`, so the renamed `SyscallTest.java` is no longer excluded and its pre-existing duplication now trips simian.

Evidence:
- `EOsocketTest.java` existed at `729bd38` (where simian was green) and is gone at the current tip.
- `SyscallTest.java` is the renamed file and is not covered by any exclude.
- simian was green at `729bd38` and red starting at `841cf62`, and `700a5612e6` is the only commit touching this file in that range.

## Suggested fix

Update the exclude in `.github/workflows/simian.yml` to the new name:

```
"-excludes=**/SyscallTest.java"
```

Alternatively, de-duplicate the two `acceptsConnectionOnSocket()` assertion blocks by extracting a shared helper, so the exclude is no longer needed.

morphqdd:
Closes #6113

`700a5612e6` renamed `EOsocketTest.java` to `SyscallTest.java`, but the simian exclude kept pointing at the old name, so the duplication inside that file started failing the build.

Checked locally with the same jar and arguments the workflow uses. With the old pattern it reports the two blocks from the issue, lines 318-335 and 647-664. With the new one it reports 0 duplicate lines and exits 0.

The duplication itself is untouched. Extracting the shared assertion block out of the two `acceptsConnectionOnSocket()` methods would let the exclude go away entirely, but that is a change to eo-runtime tests and does not belong in a fix for a red master.

### DIFF
--- a/.github/workflows/simian.yml
+++ b/.github/workflows/simian.yml
@@ -21,4 +21,4 @@ jobs:
           distribution: 'temurin'
           java-version: 17
       - run: wget --quiet https://public.yegor256.com/simian.jar -O /tmp/simian.jar
-      - run: java -jar /tmp/simian.jar -threshold=15 "-excludes=**/EOsocketTest.java" "-excludes=**/gen" "-excludes=**/it" "**/*.java"
+      - run: java -jar /tmp/simian.jar -threshold=15 "-excludes=**/SyscallTest.java" "-excludes=**/gen" "-excludes=**/it" "**/*.java"


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2864](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6130, opened 2026-07-30
#6129: build the substituted class at the instantiation sites too

### LINKED ISSUE #6129: MjTranspile.java fails to properly honor option settings
The puzzle `5955-03b720c6` from #5955 has to be resolved: 

https://github.com/objectionary/eo/blob/e8b5e409434893bc5bcdc9aa7cc91e79727dacd5/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java#L122-L130

morphqdd:
Closes #6129

Resolves the puzzle left by #5955. `to-java.xsl` read `phiDefaultClass` only where it emits an `extends` clause, so three places still hardcoded `new PhDefault(...)` and the objects built there stayed outside the substituted class.

All three now use the option: the context of a generated `apply()`, an anonymous abstract object with no children, and the argument of a `PhApplication`.

The two `((PhDefault) x)` casts and the `PhDefault x = ...` local are left alone on purpose. The named class extends `PhDefault`, so the casts still reach the override through virtual dispatch and `PhDefault` stays a valid type for the local.

Checked the generated Java for a program that reaches all three, with the option set to `org.example.PhInspected`:

```
public final class EOmain extends org.example.PhInspected {
        Phi r = new org.example.PhInspected("Φ.foo.x.main.inner");
        ((PhDefault) r).add("φ", ...
          PhDefault r1 = new org.example.PhInspected();
              rr1 = new PhApplication(rr1, 0, new org.example.PhInspected(new byte[] {...}));
```

That program is what the new test transpiles, asserting the output has no `new PhDefault` left. It fails on the un-fixed code.

One thing worth knowing, added to the javadoc: since the transpiler now calls `new` on the named class, that class has to offer the same constructors `PhDefault` does at those points, the no-arg one, the one taking a location `String` and the one taking a `byte[]`.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
@@ -113,21 +113,23 @@ public final class MjTranspile extends MjSafe {
      * full, since the generated files import nothing but
      * {@code org.eolang.*}.
      * <p>
-     *     The substitution reaches the classes that {@code to-java.xsl}
-     *     declares with an {@code extends} clause of their own. An object
-     *     with a {@code @base} extends {@code PhOnce} instead, and keeps
-     *     doing so: {@code PhOnce} is a decorator, and the object it wraps
-     *     is itself substituted, so nothing is lost.
+     *     The substitution reaches every class the generated Java declares
+     *     with an {@code extends} clause of its own, and every object it
+     *     builds with {@code new}: the context of a generated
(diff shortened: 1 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2756](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6131, opened 2026-07-30
Terminate through T in as-decimal and integral

### LINKED ISSUE #6121: `trouble` object missing in `as-decimal.eo` and `integral.eo`, causing termination messages to be lost
Two EO sources terminate through an object named `trouble` that does not exist. `eo-runtime/src/main/eo/string/as-decimal.eo:13` reaches for it on the non-finite branch, and `eo-runtime/src/main/eo/number/integral.eo:39` does the same when the partition count is not a finite positive integer:

```
[n] > as-decimal
  n.is-finite.not.if > @
    trouble
      "Can't write a non-finite number as decimal"
```

`git grep trouble -- '*.eo'` returns those two call sites and nothing else. There is no `trouble.eo` anywhere in the tree, and no Java atom by that name either. What was meant is plainly `T`, which `tuple.eo:37` already uses exactly this way.

Nothing rejects the bad name. The linter passes it through, and 3-lint XMIR records it as `base="Φ.trouble"`, which the transpiler turns into `new PhDispatch(Phi.Φ, "trouble")`. At runtime `PhPackage.loadPhi` finds neither a generated class nor a `package-info`, so it raises an uncatchable `ExFailure`:

```
Couldn't find object 'Φ.trouble' because there's no class 'org.eolang.EOtrouble'
or package-info class: 'org.eolang.EO_trouble.package-info', at least one of them
must exist
```

So a caller who asks for the decimal form of `nan` is told that a Java class is missing, and the sentence the author wrote for that exact moment never reaches anyone. Both messages are dead text.

The branches are covered by tests, which is what makes this worth reporting. `as-decimal` has three of them (`stops-on-nan`, `stops-on-negative-infinity`, `stops-on-positive-infinity`) and `integral` has six more (`stops-on-zero-partitions` and friends). All nine are green, because `to-java.xsl:853` compiles a `-->` case into `Assertions.assertThrows(Exception.class, ...)`, and a missing-class failure satisfies that just as well as a real termination. The tests reach the defect and then agree with it.

Replacing `trouble` with `T` at both sites is the whole fix. The wider gap is that a dispatch on a global name with no object behind it survives all the way to runtime, so a lint rule that rejects an unresolvable global would keep the next one from getting this far.

maxonfjvipon:
Closes #6121.

Two EO sources terminated through an object named `trouble`, which exists nowhere in the tree — no `trouble.eo`, no Java atom. The dispatch sailed through linting and transpiling as `new PhDispatch(Phi.Φ, "trouble")`, and at runtime `PhPackage` complained about a missing `org.eolang.EOtrouble` class. So a caller asking `nan` for its decimal form was told a Java class is missing, and the sentence written for that exact moment never arrived. `T` is what was meant, the way `tuple.eo` already uses it:

```
  n.is-finite.not.if > @
    T
      "Can't write a non-finite number as decimal"
```

The nine `-->` tests covering these branches were all green, because `-->` compiles to `assertThrows(Exception.class, ...)` and a missing class satisfies that as well as a real termination. Two assertions in `PhiTest` now pin the messages themselves; both fail on master with `Couldn't find object 'Φ.trouble'`.

Left for later: nothing rejects an unresolvable global. A lint rule in `objectionary/lints` that refuses a dispatch on a global with no object behind it would stop the next one from reaching runtime — worth a separate ticket.

### DIFF
--- a/eo-runtime/src/main/eo/number/integral.eo
+++ b/eo-runtime/src/main/eo/number/integral.eo
@@ -36,7 +36,7 @@
             div.
               b.minus a
               n
-    trouble
+    T
       "the number of partitions must be a finite positive integer"
   times. > [a b] >> subsection
     div.
--- a/eo-runtime/src/main/eo/string/as-decimal.eo
+++ b/eo-runtime/src/main/eo/string/as-decimal.eo
@@ -10,7 +10,7 @@
 
 [n] > as-decimal
   n.is-finite.not.if > @
-    trouble
+    T
       "Can't write a non-finite number as decimal"
     if.
       n.lt 0
--- a/eo-runtime/src/test/java/org/eolang/PhiTest.java
+++ b/eo-runtime/src/test/java/org/eolang/PhiTest.java
@@ -6,6 +6,7 @@ package org.eolang;
 
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 /**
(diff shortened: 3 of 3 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+708](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6134, opened 2026-07-30
Move Maven proxy translation out of MjSafe

### LINKED ISSUE #6117: `MjSafe` class exceeds allowed ClassFanOutComplexity, resulting in checkstyle violations
`eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java:42` carries `@checkstyle ClassFanOutComplexityCheck (1000 lines)`. Qulice registers `ClassFanOutComplexity` with `max=30`, and with the marker stripped checkstyle reports `Class Fan-Out Complexity is 40 (max allowed is 30)` on the class declaration at line 45. The class also already carries `@SuppressWarnings("PMD.TooManyFields")` at line 44.

The 738-line abstract Mojo holds two unrelated things. One is the shared `@Parameter` surface that every `Mj*` goal inherits, roughly 30 fields of mostly `File` and `boolean`. The other is execution plumbing that has nothing to do with parameters: the single-thread `ExecutorService` timeout wrapper in `execWithTimeout()` at line 628, with its `TimeUnit`, `TimeoutException` and `ExecutionException` handling, and the `Settings`-to-`java.net.Proxy` translation in `proxies()` at line 609 that reaches for `InetSocketAddress`. The parameter surface is immovable, since it is the plugin's public API, so the suppression currently covers for the part that is movable.

Moving `execWithTimeout()` and `proxies()` into classes of their own drops the `java.util.concurrent` and `java.net` imports, about eight dependencies, without touching a single `@Parameter`. `execWithTimeout()` calls the abstract `exec()`, so it would take the goal's body as a lambda. That alone is enough to clear the cap and retire the marker.

#5101 is an open PDD puzzle on the tojos lifecycle in this same class around line 433 — a different concern, but it may be worth doing in the same pass.

Thayorns:
`MjSafe` was doing the Settings-to-`java.net.Proxy` translation itself, in `proxies()`, next to the thirty `@Parameter` fields that make up the plugin's public API. That translation has nothing to do with parameters, and it is the reason the Mojo named `java.net.Proxy` and `InetSocketAddress` at all. It now lives in `Proxies`, and the objectionary chain asks for the proxies it needs at the point where it builds `OyRemote`.

One related change: the `settings` field now starts as empty `Settings` rather than null, the same way `tag`, `scope` and `cache` already default in this class. Maven overwrites it on injection, and `Proxies` no longer has to tolerate an uninjected null the way `proxies()` did with `Optional.ofNullable`. Every Mojo test that reaches the objectionary (`MjPullTest`, `MjProbeTest`) runs without injected settings, so this path is covered.

This does not retire the `@checkstyle ClassFanOutComplexityCheck` marker yet. I measured the class with checkstyle: it was at 40 dependencies against the allowed 30, and it is at 39 now. The issue estimates that moving `proxies()` and `execWithTimeout()` clears the cap, but most of what the marker covers is composition, not plumbing imports: the objectionary chain, the hash chain and the timeout together are worth ten, which is exactly what is needed. That is written up as a puzzle in the class Javadoc, with the three compositions spelled out.

Closes #6117

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java
@@ -8,13 +8,10 @@ import com.jcabi.log.Logger;
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
-import java.net.InetSocketAddress;
-import java.net.Proxy;
 import java.nio.file.Paths;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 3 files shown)


Thayorns:
@yegor256 take a look please. Happy to clarify anything, in particular the measured fan-out numbers behind the puzzle I left in the class.

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (141 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+354](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6135, opened 2026-07-30
Bind sock to socket instead of impl in connection error messages

### LINKED ISSUE #6132: `sock` incorrectly binds to `impl`, causing failure in connection error messages in `socket.eo`
The anonymous scope that `connect` hands to `safe-socket` binds `sock` two hops up, at `eo-runtime/src/main/eo/socket.eo:89`, but it sits three hops below the `socket` formation. Its own parent is `connect`, whose parent is the `impl` formation that holds the shared posix/win32 body, so `^.^` stops on `impl`:

```
[scope cant-connect] > connect
  safe-socket > @
    [] >>
      if. > @
        connected.eq -1
        cant-connect
          "Couldn't connect to %s:%d on socket #%d because %s".printf
            * sock.address sock.port sock.sd sock.sys.message
        scope
          sock.scoped-socket sock.sd
      code. > connected
        sock.sys.raw *1
          "connect"
          sock.sd
          sock.sockaddr
          sock.sockaddr.size
      ^.^ > sock
```

`impl` declares `addr`, `addrint`, `closed-socket`, `connect`, `listen`, `safe-socket`, `scoped-socket`, `sd`, `sockaddr` and `sys`, and it has no `φ`. Five of the seven attributes `sock` reaches are on that list, which is why the mistake has stayed hidden. The other two, `address` and `port`, belong to `socket` alone, and with no decoratee to descend into, `PhDefault.absent` hands back a `PhTerminator`. A refused connection therefore dies on the bottom object at `socket.eo:80` while formatting its own diagnostic, and the message it was assembling never reaches the caller. `listen` repeats the binding at `socket.eo:134`, so the two messages at `socket.eo:98` and `socket.eo:103` break the same way.

The same file shows what the depth should have been. `scoped-socket` at `socket.eo:161` writes `^.^.as-input`, and `as-input` exists only on `socket`. That binding sits one level shallower, directly under `impl`, so its `^.^` does land where it means to.

Nothing catches this in CI because all three tests that drive the EO object bind the fallback to `SyscallTest.Simple`, whose `lambda` returns `true` without ever dataizing its void attribute. The `printf` holding `sock.address` lands in that discarded slot, so the failure branch is covered while the failure message is not.

Writing `^.^.^ > sock` at `socket.eo:89` and `socket.eo:134` repairs all three messages and leaves the other five uses working, since `socket` decorates `impl` through its `φ` and reaches every one of them.

maxonfjvipon:
The anonymous scope that `connect` hands to `safe-socket` bound `sock` two hops up, but it sits three hops below the `socket` formation: its own parent is `connect`, whose parent is `impl`. So `^.^` stopped on `impl`, which declares neither `address` nor `port`, and has no `φ` to descend into. Five of the seven attributes `sock` reaches happen to live on `impl`, which is why this stayed hidden. The other two only appear in the failure messages.

```
-            ^.^ > sock
+            ^.^.^ > sock
```

A refused connection therefore died on the bottom object while formatting its own diagnostic, and the caller never saw the reason. `listen` repeated the same binding, breaking both of its messages the same way. One hop deeper lands on `socket`, which reaches all seven through its `φ`.

Nothing caught this because every test drove the fallback with a fake whose `lambda` returns `true` without dataizing its void attribute, so the `printf` holding `sock.address` went straight into a discarded slot. The two new tests hand `connect` and `listen` a fallback that dataizes the message it is given, and assert the address comes back in it.

Closes #6132

### DIFF
--- a/eo-runtime/src/main/eo/socket.eo
+++ b/eo-runtime/src/main/eo/socket.eo
@@ -86,7 +86,7 @@
                 sock.sd
                 sock.sockaddr
                 sock.sockaddr.size
-            ^.^ > sock
+            ^.^.^ > sock
           cant-connect
       [scope cant-listen] > listen
         safe-socket > @
@@ -131,7 +131,7 @@
                 "listen"
                 sock.sd
                 2048
-            ^.^ > sock
+            ^.^.^ > sock
           cant-listen
       [scope cant] > safe-socket
         if. > @
--- a/eo-runtime/src/test/java/org/eolang/SyscallTest.java
+++ b/eo-runtime/src/test/java/org/eolang/SyscallTest.java
@@ -80,39 +80,37 @@ final class SyscallTest {
     }
 
     @Test
-    void reportsTheReasonOfTheRefusedConnection() throws IOException {
+    void tellsTheFallbackWhichAddressItFailedToReach() throws IOException {
         final SyscallTest.RandomServer refused = new SyscallTest.RandomServer().started();
         refused.stop();
         final Phi socket = Phi.Φ.take("socket").copy();
         socket.put(0, new Data.ToPhi(this.localhost()));
         socket.put(1, new Data.ToPhi(refused.port));
         final Phi connect = socket.take("connect").copy();
-        connect.put(0, new SyscallTest.Simple());
-        connect.put(1, new SyscallTest.Reason());
+        connect.put(1, Phi.Φ.take("dataized").copy());
         MatcherAssert.assertThat(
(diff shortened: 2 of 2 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (39) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+720](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6136, opened 2026-07-30
#6118: accept zero-padded integers in scanf

### LINKED ISSUE #6118: `scanf` function incorrectly rejects valid zero-padded integers for `%d` specifier due to character count
## What happens

`scanf` with a `%d` specifier rejects a perfectly valid integer as "out of range" when the input digits are zero-padded past 19 characters. The overflow guard counts the number of digit *characters* rather than the numeric value, so leading zeros inflate the count and trigger a false overflow.

### Runtime evidence

| expression | actual result | expected |
| --- | --- | --- |
| `head. (scanf "%d" "00000000000000000042")` | **`ExFailure: The number doesn't fit into long range for the '%d' conversion`** | `42` |
| `head. (scanf "%d" "42")` | `42` | `42` (control) |
| `head. (scanf "%d" "9223372036854775807")` | `9.2e18` | fits (control) |

The `%d` capture regex is `([+-]?\d+)`, which matches the whole run of digits including any leading zeros. `"00000000000000000042"` is 20 digit characters, its numeric value is 42, and it should scan cleanly the way C's `scanf("%d", ...)` reads `42` from `"00000000000000000042"`.

Confirmed by injecting a positive regression test into `scanf.eo` and running `EO_string.EOscanfTest`: the case above fails with `ExFailure The number doesn't fit into long range for the '%d' conversion`.

## Root cause

`eo-runtime/src/main/eo/string/scanf.eo`, object `as-integer` (around lines 56-74). The range guard is:

```
if. > @
  or.
    digits.length.gt 19
    and.
      digits.length.eq 19
      not.
        negative.if
          digits-lte digits "9223372036854775808"
          digits-lte digits "9223372036854775807"
  T
  negative.if negated folded
```

`digits` is the matched text with only the sign stripped; it is never trimmed of leading zeros. The guard then tests `digits.length` (character count). For any input whose digit run is 20+ characters, `digits.length.gt 19` is true and the atom errors, even when the value fits in a long. The `fold-digits` accumulator that computes the actual value is unaffected by leading zeros, so only the guard is wrong.

## Suggested fix

Trim leading zeros from `digits` before applying the length-based guard (keeping at least one digit so `"0"` and `"0000"` still fold to `0`), and run both `digits.length` checks and the `digits-lte` comparison against that trimmed string. For example, add a helper that drops leading `"0"` characters up to the last one:

```
[d] > trimmed
  [index] >> rec
    if. > @
      or.
        index.gte d.length.minus 1
        not.
          eq.
            slice d index 1
            "0"
      slice d index (d.length.minus index)
      rec index.plus 1
  | 0 > @
```

and then guard on `(trimmed digits).length` and `digits-lte (trimmed digits) "..."` instead of on the raw `digits`. This keeps the genuine overflow cases (`"9223372036854775808"`, `"12345678901234567890"`) rejected while accepting zero-padded values that fit.

gemshrine:
Fixes #6118.

## Problem

[`as-integer`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/scanf.eo#L56-L74) rejects a `%d` capture by checking the length of its textual digits before converting them. The `%d` pattern accepts leading zeroes, so a valid value such as `00000000000000000042` has twenty digits and is rejected even though its numerical value is `42`.

The overflow guard is the problematic path: it sees the unnormalised representation, while [`fold-digits`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/scanf.eo#L96) would produce the correct value. This makes accepted input depend on presentation zeroes rather than on the integer being scanned.

## Change

[`without-leading-zeroes`](https://github.com/gemshrine/eo/blob/6118-scanf-leading-zeroes/eo-runtime/src/main/eo/string/scanf.eo#L98-L127) removes leading zeroes while preserving one zero for an all-zero value. The [`normalized`](https://github.com/gemshrine/eo/blob/6118-scanf-leading-zeroes/eo-runtime/src/main/eo/string/scanf.eo#L56-L90) unsigned magnitude is now used consistently by both the range guard and `fold-digits`.

The sign is still determined from the original text, so the existing boundary rules remain unchanged: the implementation accepts `Long.MAX_VALUE` and `Long.MIN_VALUE`, but rejects values beyond either bound.

## Regression coverage

The atom tests now cover:

- a positive zero-padded value that previously failed: [`00000000000000000042`](https://github.com/gemshrine/eo/blob/6118-scanf-leading-zeroes/eo-runtime/src/main/eo/string/scanf.eo#L325-L328);
- the same value with a minus sign: [`-00000000000000000042`](https://github.com/gemshrine/eo/blob/6118-scanf-leading-zeroes/eo-runtime/src/main/eo/string/scanf.eo#L330-L333);
- an all-zero representation, which must remain zero: [`00000000000000000000`](https://github.com/gemshrine/eo/blob/6118-scanf-leading-zeroes/eo-runtime/src/main/eo/string/scanf.eo#L335-L338);
- a zero-padded value above `Long.MAX_VALUE`, which must still terminate: [`00009223372036854775808`](https://github.com/gemshrine/eo/blob/6118-scanf-leading-zeroes/eo-runtime/src/main/eo/string/scanf.eo#L410-L412).

The existing non-padded overflow cases remain alongside these tests.

## Verification

`mvn -q -pl eo-runtime test -Deo.autoFix -DskipUTs=false` completes successfully locally. GitHub Actions is running against the updated revision.

### DIFF
--- a/eo-runtime/src/main/eo/string/scanf.eo
+++ b/eo-runtime/src/main/eo/string/scanf.eo
@@ -56,22 +56,16 @@
     [text] > as-integer
       if. > @
         or.
-          digits.length.gt 19
+          normalized.length.gt 19
           and.
-            digits.length.eq 19
+            normalized.length.eq 19
             not.
               negative.if
-                digits-lte digits "9223372036854775808"
-                digits-lte digits "9223372036854775807"
+                digits-lte normalized "9223372036854775808"
+                digits-lte normalized "9223372036854775807"
         T
           "The number doesn't fit into long range for the '%%d' conversion"
         negative.if negated folded
-      signed.if > digits
-        slice
-          text
-          1
-          text.length.minus 1
-        text
       [a b] > digits-lte
         if. > [index] >> rec
           index.eq a.length
@@ -93,7 +87,7 @@
               rec
                 index.plus 1
         | 0 > @
-      fold-digits digits > folded
+      fold-digits normalized > folded
       folded.times -1 > negated
       eq. > negative
         slice
@@ -101,10 +95,23 @@
           0
           1
         "-"
+      if. > normalized
+        stripped.eq ""
+        "0"
+        stripped
       negative.or > signed
         eq.
           slice text 0 1
           "+"
+      replaced > stripped
+        signed.if
+          slice
+            text
+            1
+            text.length.minus 1
+          text
+        regex "/^0+/"
+        ""
     [group acc] >> convert
       matched.exists.not.if > @
         *
@@ -302,6 +309,21 @@
     head.
       scanf "%d" "+42"
 
+  eq. ++> can-scan-a-zero-padded-integer
+    42
+    head.
+      scanf "%d" "00000000000000000042"
+
+  eq. ++> can-scan-a-zero-padded-negative-integer
+    -42
+    head.
+      scanf "%d" "-00000000000000000042"
+
+  eq. ++> can-scan-many-zeroes-as-zero
+    0
+    head.
+      scanf "%d" "00000000000000000000"
+
   eq. ++> can-scan-what-printf-wrote
     scanf
       "%s is about %d?"
@@ -371,6 +393,10 @@
   scanf --> stops-on-a-nineteen-digit-value-below-long-min
     "%d"
     "-9223372036854775809"
+
+  scanf --> stops-on-a-zero-padded-integer-above-long-max
+    "%d"
+    "00009223372036854775808"
 
   scanf --> stops-on-a-wrong-format
     "%l"


Thayorns (approved):

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+369](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1201](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6030, opened 2026-07-28
Whole-program-analysis lint cache uses a constant key, so one project's lint result silently replaces another's


morphqdd:
## What happens

Whole-program-analysis (WPA) lint results are cached with a constant key shared by every project on the machine, so one project's lint verdict silently replaces another's.

In `Linting.lintAll` (`eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java:366-382`):

```java
final Path wpa = Paths.get("wpa.xmir");
this.guard.apply(
    this.sourcesDir, target, wpa,
    new Cache(
        this.cacheDir.resolve(Linting.CACHE),
        root -> { ... run WPA ... },
        p -> p.getFileName().toString().endsWith(".xmir") && !p.getFileName().equals(wpa)
    )
);
```

Two things compound here:

1. **No per-project identity in the cache key.** The per-file lint cache a few lines above (`Linting.java:301-313`) correctly builds its path with `new CachePath(cacheDir, version, new TojoHash(tojo).get())` — folding in the plugin version and a project-specific hash. The WPA cache instead passes the bare `this.cacheDir.resolve(Linting.CACHE)` directory as `Cache`'s base, with no version or project hash appended, so the cache file is literally `<eo.cache>/linted/wpa.xmir` (default `~/.eo/linted/wpa.xmir`), the same path for every EO project on the machine. `lintAsPackage` defaults to `true` (`MjSafe.java:370`), so this runs by default.

2. **The freshness hash is constant.** `Cache.apply` computes its "has anything changed" hash from the `source` argument, which here is `this.sourcesDir` — the project's `${project.basedir}/src/main/eo` directory of `.eo` files (`MjSafe.java:99-105`). But the cache's own filter only accepts files ending in `.xmir`. `Cache.dirSha` (`Cache.java:139-153`) walks that directory, the filter admits nothing, and the digest is computed over zero bytes — the SHA-256 of an empty input, `47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=`, for every project, always.

The combination means: project A builds, its WPA lint result (clean or not) is written to the shared path with the constant hash; project B builds next, the guard sees the hash "matches" (both are the same constant) and the cache file already exists, so it reuses project A's result outright, real defects in project B included.

[...]

Issue #5102, opened 2026-05-06
Reduce coupling in the `MjLint`


volodya-lombrozo:
Currently `MjLint` mixes Maven entry-point concerns with core EO linting logic, making it hard to test in isolation.

It would be nice to decouple `MjLint` similar to what was done in #4989 for `MjResolve`, #5061 for `MjTranspile`, #5081 for `MjParse`, #5084 for `MjProbe`, #5093 for `MjPull`, and #5098 for `MjAssemble`:

1. `MjLint` - a thin Maven endpoint.
2. `Lint` - core linting logic without Maven dependencies.

Also, it would be great to add `MjLintTest` with tests that verify the Maven mojo behavior in isolation.

0pdd:
@volodya-lombrozo 4 puzzles [#5109](https://github.com/objectionary/eo/issues/5109), [#5110](https://github.com/objectionary/eo/issues/5110), [#5111](https://github.com/objectionary/eo/issues/5111), [#5112](https://github.com/objectionary/eo/issues/5112) are still not solved.

0crat:
@volodya-lombrozo Thanks for your submission! While this ticket wasn't accepted this time, your contribution is valuable to our project's quality improvement. You've received -8 points, bringing your running score to [+181](https://www.eolang.org/zerocracy/objectionary-vitals.html). We encourage you to keep [reporting bugs](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) and suggesting enhancements - [every bug report matters](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html)! Check your [Zerocracy account](https://www.zerocracy.com/) for updates.

0pdd:
@volodya-lombrozo 3 puzzles [#5109](https://github.com/objectionary/eo/issues/5109), [#5110](https://github.com/objectionary/eo/issues/5110), [#5112](https://github.com/objectionary/eo/issues/5112) are still not solved; solved: [#5111](https://github.com/objectionary/eo/issues/5111).

0pdd:
@volodya-lombrozo 2 puzzles [#5109](https://github.com/objectionary/eo/issues/5109), [#5112](https://github.com/objectionary/eo/issues/5112) are still not solved; solved: [#5110](https://github.com/objectionary/eo/issues/5110), [#5111](https://github.com/objectionary/eo/issues/5111).

0pdd:
[...]

PR #6149, opened 2026-07-30
fix(#6030): key WPA lint cache to the analysed XMIR, not the EO sources

androvonx95:
The whole-program-analysis cache hashed sourcesDir, which holds only .eo files, while the cache filter accepts only .xmir. Nothing matched, so the freshness hash was the SHA-256 of an empty input for every project and never changed. Combined with a cache path that carried no plugin version, one project's WPA verdict was reused by every other project on the machine, hiding real defects.

Hash the 3-lint directory instead, which is what the filter was always written for: it excludes wpa.xmir, and wpa.xmir only ever lands there. Add the plugin version to the cache path, the way lintOne already does, so verdicts from an older plugin are not reused. The directory is created before hashing, since a package with no programs of its own would otherwise have nothing to hash.

With sourcesDir no longer read, drop it from Linting and its callers.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
@@ -11,6 +11,7 @@ import com.jcabi.manifests.Manifests;
 import com.jcabi.xml.XML;
 import com.jcabi.xml.XMLDocument;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
@@ -131,12 +132,6 @@ final class Linting implements Step {
     private final boolean lintAsPackage;
 
     /**
-     * EO sources directory (for WPA cache key).
-     * @checkstyle MemberNameCheck (5 lines)
-     */
-    private final Path sourcesDir;
-
-    /**
      * Whether to skip linting entirely.
      * @checkstyle MemberNameCheck (5 lines)
      */
@@ -160,10 +155,9 @@ final class Linting implements Step {
      * @param experimental Whether to skip experimental lints
      * @param warning Whether to fail on warnings
      * @param pkg Whether to lint all sources as a package
-     * @param sources EO sources directory
      * @param skip Whether to skip linting entirely
      * @todo #5102:90min Reduce long parameter lists in Linting, Parse, Pull, and similar classes.
-     *  Linting currently takes 13 constructor parameters. Parse, Pull, and Probe have similar
+     *  Linting currently takes 12 constructor parameters. Parse, Pull, and Probe have similar
      *  issues. The long parameter lists make call sites hard to read and fragile — adding a new
      *  option requires updating every call site across the codebase.
      * @checkstyle ParameterNumberCheck (20 lines)
@@ -181,7 +175,6 @@ final class Linting implements Step {
         final boolean experimental,
         final boolean warning,
         final boolean pkg,
-        final Path sources,
         final boolean skip
     ) {
         this.tojos = srcs;
@@ -195,7 +188,6 @@ final class Linting implements Step {
         this.skipExperimentalLints = experimental;
         this.failOnWarning = warning;
         this.lintAsPackage = pkg;
-        this.sourcesDir = sources;
         this.skipLinting = skip;
         this.guard = new ConcurrentCache();
     }
@@ -398,11 +390,13 @@ final class Linting implements Step {
         final List<org.eolang.wpa.Defect> defects;
         if (this.cacheEnabled) {
             final Path wpa = Paths.get("wpa.xmir");
-            final Path target = this.targetDir.resolve(Linting.DIR).resolve(wpa);
+            final Path base = this.targetDir.resolve(Linting.DIR);
+            final Path target = base.resolve(wpa);
+            Files.createDirectories(base);
             this.guard.apply(
-                this.sourcesDir, target, wpa,
+                base, target, wpa,
                 new Cache(
-                    this.cacheDir.resolve(Linting.CACHE),
+                    this.cacheDir.resolve(Linting.CACHE).resolve(this.version),
                     root -> {
                         Logger.info(this, "Linting a package");
                         final Directives all = new Directives().add("defects");
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
@@ -159,8 +159,35 @@ final class MjLintTest {
         MatcherAssert.assertThat(
             "WPA results must be saved to cache",
             cache.resolve(Linting.CACHE)
+                .resolve(FakeMaven.pluginVersion())
                 .resolve("wpa.xmir").toFile(),
             FileMatchers.anExistingFile()
+        );
+    }
+
+    @Test
+    void doesNotReuseWholeProgramAnalysisCacheOfAnotherProject(@Mktmp final Path temp)
+        throws IOException {
+        final Path cache = temp.resolve("shared-cache");
+        MjLintTest.linting(temp.resolve("clean"), cache, MjLintTest.flawless())
+            .execute(new FakeMaven.Lint());
+        Assertions.assertThrows(
+            IllegalStateException.class,
+            () -> MjLintTest.linting(
+                temp.resolve("broken"), cache, MjLintTest.probmlematic()
+            ).execute(new FakeMaven.Lint()),
+            "WPA error of the second project must be reported, not hidden by the cache of the first"
+        );
+    }
+
+    @Test
+    void lintsPackageWithoutASingleProgram(@Mktmp final Path temp) {
+        Assertions.assertDoesNotThrow(
+            () -> new FakeMaven(temp)
+                .with("lintAsPackage", true)
+                .with("cache", temp.resolve("cache").toFile())
+                .execute(new FakeMaven.Lint()),
+            "Linting a package that has no programs at all must not fail"
         );
     }
 
@@ -426,6 +453,45 @@ final class MjLintTest {
             "",
             "[] > main",
             "  cti true \"critical\" \"msg\" > @",
+        };
+    }
+
+    /**
+     * Maven that lints a package, keeping its EO sources apart from its target directory,
+     * the way a real Maven build does with {@code src/main/eo}.
+     * @param workspace Workspace of the project
+     * @param cache Cache directory, possibly shared with other projects
+     * @param program Program to lint
+     * @return Maven ready to be executed
+     * @throws IOException If fails to save the program
+     */
+    private static FakeMaven linting(
+        final Path workspace, final Path cache, final String... program
+    ) throws IOException {
+        return new FakeMaven(workspace)
+            .with("lintAsPackage", true)
+            .with("cache", cache.toFile())
+            .with("sourcesDir", workspace.resolve("foo").toFile())
+            .withProgram(program);
+    }
+
+    /**
+     * Program without a single defect.
+     * @return Program without defects
+     */
+    private static String[] flawless() {
+        return new String[]{
+            "+home https://www.eolang.org",
+            "+package foo.x",
+            "+version 0.0.0",
+            "+unlint empty-object",
+            "+unlint unit-test-missing",
+            "+unlint mandatory-spdx",
+            "+unlint comment-too-short",
+            "+unlint object-has-data",
+            "",
+            "[x] > main",
+            "  (stdout \"Hello!\" x).print > @",
         };
     }
 
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjCompile.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjCompile.java
@@ -46,7 +46,6 @@ public final class MjCompile extends MjSafe {
                         this.skipExperimentalLints,
                         this.failOnWarning,
                         this.lintAsPackage,
-                        this.sourcesDir.toPath(),
                         this.skipLinting
                     )
                 ),
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjLint.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjLint.java
@@ -44,7 +44,6 @@ public final class MjLint extends MjSafe {
             this.skipExperimentalLints,
             this.failOnWarning,
             this.lintAsPackage,
-            this.sourcesDir.toPath(),
             this.skipLinting
         ).exec();
     }
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/CompilingTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/CompilingTest.java
@@ -56,7 +56,6 @@ final class CompilingTest {
                     false,
                     false,
                     false,
-                    temp,
                     true
                 ),
                 new Resolving(
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/LintingTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/LintingTest.java
@@ -50,7 +50,6 @@ final class LintingTest {
                 false,
                 false,
                 false,
-                temp,
                 true
             ).exec(),
             "Linting must be fully skipped when skipLinting is TRUE"


eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -398,11 +390,13 @@ private int lintAll(
         final List<org.eolang.wpa.Defect> defects;
         if (this.cacheEnabled) {
             final Path wpa = Paths.get("wpa.xmir");
-            final Path target = this.targetDir.resolve(Linting.DIR).resolve(wpa);
+            final Path base = this.targetDir.resolve(Linting.DIR);
+            final Path target = base.resolve(wpa);
+            Files.createDirectories(base);
             this.guard.apply(
-                this.sourcesDir, target, wpa,
+                base, target, wpa,
```
Thayorns, line 397: The hash now covers target/3-lint, but that directory only ever receives the programs linted one by one in linting(), which iterates this.tojos.withXmir() at line 240. The package analysis at lines 380-382 feeds WPA from this.tojos plus this.compile, and compileTojos() at MjSafe.java:542 builds a separate compile-scope set over the catalog, so with eo.scope set to test the two point at different programs under different target directories. A change confined to a compile-scope program then leaves the test-scope 3-lint contents byte-identical, the stored sha still matches, and the previous verdict is served from cache, which is the same staleness this change closes for the rest of the WPA input.

Thayorns (commented):
The WPA freshness hash moves off sourcesDir, which held only .eo files and therefore always hashed to nothing under an .xmir-only filter, and onto the 3-lint directory that filter was written for, with the plugin version joining the cache path so verdicts from an older plugin are not reused. The mechanics check out: the filter excludes wpa.xmir at Linting.java:409-410, so writing the result into the hashed directory does not invalidate it; lintOne fills 3-lint at line 249 before lintAll reads it at line 260; and the new cross-project test would pass the stale verdict through on master. One gap remains in what the hash actually covers, noted on the line.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -398,11 +390,13 @@ private int lintAll(
         final List<org.eolang.wpa.Defect> defects;
         if (this.cacheEnabled) {
             final Path wpa = Paths.get("wpa.xmir");
-            final Path target = this.targetDir.resolve(Linting.DIR).resolve(wpa);
+            final Path base = this.targetDir.resolve(Linting.DIR);
```
Favixx, line 393: `base` now plays two roles at once: the root whose SHA is the freshness key, and the directory the result is written into. That works only because the filter at line 410 excludes the output — but it excludes it by *name*:

```java
&& !p.getFileName().equals(wpa)
```

`dirSha` walks the whole tree (`Files.walk(dir)`), and `lintOne` places each linted object through `Place`, which maps `foo.wpa` to `3-lint/foo/wpa.xmir`. So an EO object named `wpa` in any package is silently dropped from the hash, and editing it can no longer invalidate the package verdict — precisely the failure mode this PR is fixing, at one level down.

`!p.equals(target)` compares the absolute path and excludes exactly the one file that has to be excluded. `target` is already in scope on the next line.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -398,11 +390,13 @@ private int lintAll(
...
+            final Path target = base.resolve(wpa);
+            Files.createDirectories(base);
```
Favixx, line 395: This line looks removable and is not: `Cache.apply` calls `sha(source)`, which throws `IllegalArgumentException("Path '%s' is neither a regular file nor a directory")` when the directory does not exist, and `3-lint` is only created as a side effect of `lintOne` writing into it — so a package with no programs of its own never creates it. That is the whole content of `lintsPackageWithoutASingleProgram`, and none of it is visible from here.

A short comment (`// the directory has to exist before it can be hashed`) would keep the next reader from deleting a line that has a test behind it but no local reason.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
Favixx, line 397: The third argument is the lock key: `ConcurrentCache.apply` locks on `tail.normalize()`, which here is the bare `wpa.xmir`. `lintOne` passes `base.relativize(target)` for the same `this.guard` instance, so a root-package object named `wpa` produces the same key and the two paths serialise against each other.

Today that is only contention — the cache files differ because the two `Cache` instances have different bases — so this is not a defect. It is the second consequence of identifying the package result by a file name that an ordinary object can also have (see my comment on line 393), which is the argument for keying both on absolute paths.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -398,11 +390,13 @@ private int lintAll(
...
             this.guard.apply(
-                this.sourcesDir, target, wpa,
+                base, target, wpa,
                 new Cache(
-                    this.cacheDir.resolve(Linting.CACHE),
+                    this.cacheDir.resolve(Linting.CACHE).resolve(this.version),
```
Favixx, line 399: The description says this is done "the way `lintOne` already does", but `lintOne` sixty lines above builds `new CachePath(this.cacheDir.resolve(Linting.CACHE), this.version, new TojoHash(tojo).get())` — a three-part path assembled by a dedicated class — while this is a hand-rolled two-part `resolve` chain.

The result is that `linted/<version>/` now holds `wpa.xmir` and `wpa.xmir.sha256` as bare files, side by side with the per-tojo hash directories `CachePath` creates there. Anything that later walks or prunes that level has to know about both shapes. `CachePath` with a fixed tail, or a `wpa` subdirectory of its own, keeps one scheme instead of two.

Worth noting separately: entries written by earlier versions at `linted/wpa.xmir` are now unreachable and nothing ever removes them, so a shared `~/.eo` keeps them forever.

Favixx, line 399: The key is now `dirSha(3-lint)` plus the plugin version, and it still leaves out the one input that changes the answer without changing either: `skipProgramLints`. `wpa(pkg)` builds the analysis as `new Program(pkg).without(this.skipProgramLints.toArray(...))`, and line 388 logs exactly that set right before this call.

So: run with `eo.skipProgramLints=some-lint`, run again without it, nothing in `3-lint` changed and the plugin version is the same — the cached verdict from the first run is served, missing the defects the second run was supposed to report. That is the same shape as the bug being fixed here, and the fix is the same size: fold the sorted set into the tail, or into the hash the way `TojoHash` does for `lintOne`.

I would not block on it — it is pre-existing and arguably its own issue — but the PR is the moment when someone is holding this key in their head.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -159,11 +159,38 @@ void savesForWholeProgramAnalysisResultsToCache(@Mktmp final Path temp) throws I
         MatcherAssert.assertThat(
             "WPA results must be saved to cache",
             cache.resolve(Linting.CACHE)
+                .resolve(FakeMaven.pluginVersion())
```
Favixx, line 162: This asserts that the file lands under a directory named after the plugin version, which is what the production code now writes — but the assertion is built from `FakeMaven.pluginVersion()`, the same source the code under test reads, so it holds for any value including a constant. Nothing in the suite distinguishes "the path carries the version" from "the path carries a fixed string".

The behaviour that matters is that two different plugin versions do not share a verdict, which is testable in the same shape as `doesNotReuseWholeProgramAnalysisCacheOfAnotherProject`: same workspace, same sources, one `FakeMaven` configured with a different plugin version, and the second run must recompute rather than serve the first one's answer.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -159,11 +159,38 @@ void savesForWholeProgramAnalysisResultsToCache(@Mktmp final Path temp) throws I
...
                 .resolve("wpa.xmir").toFile(),
             FileMatchers.anExistingFile()
         );
     }
 
+    @Test
+    void doesNotReuseWholeProgramAnalysisCacheOfAnotherProject(@Mktmp final Path temp)
```
Favixx, line 169: Good test, and it is the one that would have caught #6030 on master. One thing it does not check is that the first `execute` actually populated the shared cache: if `linting()` were misconfigured and the clean project never wrote an entry, the second project would recompute for that reason and the assertion below would still pass, green for the wrong reason.

One line between the two executions — `FileMatchers.anExistingFile()` on `cache.resolve(Linting.CACHE).resolve(FakeMaven.pluginVersion()).resolve("wpa.xmir")`, which the test above already spells out — closes that hole and makes the test say "the cache was there and was correctly not reused" rather than "the second run threw".

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -159,11 +159,38 @@ void savesForWholeProgramAnalysisResultsToCache(@Mktmp final Path temp) throws I
...
+        throws IOException {
+        final Path cache = temp.resolve("shared-cache");
+        MjLintTest.linting(temp.resolve("clean"), cache, MjLintTest.flawless())
+            .execute(new FakeMaven.Lint());
+        Assertions.assertThrows(
+            IllegalStateException.class,
```
Favixx, line 175: `IllegalStateException` is not specific enough to pin what this test is about. `Cache.apply` wraps *every* `IOException` as `IllegalStateException("Failed to perform an IO operation with cache")`, and `dirSha` throws the same type for an unreadable directory — so a run that fails because the shared cache directory is in a state the test did not intend passes exactly like a run that correctly reported the WPA defect.

`assertThrows` returns the exception; asserting that its message mentions the defect (the same way the sibling tests around line 219 do) makes the difference between "the second project failed" and "the second project's own error was reported".

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -159,11 +159,38 @@ void savesForWholeProgramAnalysisResultsToCache(@Mktmp final Path temp) throws I
...
+            () -> MjLintTest.linting(
+                temp.resolve("broken"), cache, MjLintTest.probmlematic()
```
Favixx, line 177: Not introduced here, but this PR adds the seventh call site: `probmlematic` is a typo for `problematic`. Since a private static helper in one test class renames with a single IDE action and touches only this file, folding it in costs nothing — and the alternative is that the eighth caller copies it too.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -159,11 +159,38 @@ void savesForWholeProgramAnalysisResultsToCache(@Mktmp final Path temp) throws I
...
+            ).execute(new FakeMaven.Lint()),
+            "WPA error of the second project must be reported, not hidden by the cache of the first"
+        );
+    }
+
+    @Test
+    void lintsPackageWithoutASingleProgram(@Mktmp final Path temp) {
```
Favixx, line 184: This is the regression test for `Files.createDirectories(base)`, and `assertDoesNotThrow` is the weakest statement it could make: it stays green if the directory is created *and* if the whole cached branch is skipped for some unrelated reason, which is how it would have looked before `lintAsPackage` was involved at all.

Asserting that `target/3-lint` exists afterwards — or that the cache received a `wpa.xmir` for an empty package, which is the interesting claim — states what the fix does rather than what it avoids.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -429,6 +456,45 @@ private static String[] critical() {
...
+    /**
+     * Maven that lints a package, keeping its EO sources apart from its target directory,
+     * the way a real Maven build does with {@code src/main/eo}.
+     * @param workspace Workspace of the project
+     * @param cache Cache directory, possibly shared with other projects
+     * @param program Program to lint
+     * @return Maven ready to be executed
+     * @throws IOException If fails to save the program
+     */
+    private static FakeMaven linting(
```
Favixx, line 468: This helper and `lintsPackageWithoutASingleProgram` twenty lines above build the same `FakeMaven` two different ways — `.with("lintAsPackage", true).with("cache", …)` appears in both, inline in one and behind the helper in the other. Since the helper exists, the second test may as well use it (`linting(temp, temp.resolve("cache"))` with no program), which also documents that "no programs" is a variation of the same setup rather than a different scenario.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -429,6 +456,45 @@ private static String[] critical() {
...
+        final Path workspace, final Path cache, final String... program
+    ) throws IOException {
+        return new FakeMaven(workspace)
+            .with("lintAsPackage", true)
+            .with("cache", cache.toFile())
+            .with("sourcesDir", workspace.resolve("foo").toFile())
```
Favixx, line 474: The Javadoc above sells this as "keeping its EO sources apart from its target directory, the way a real Maven build does with `src/main/eo`" — but `sourcesDir` is exactly the parameter this PR stops passing to `Linting`, so from the point of view of the code under test this line is now inert.

If `FakeMaven` still needs it for its own bookkeeping, say that instead; if it does not, dropping it removes the impression that the test's separation of sources from target is what makes the cache key correct. As written, the helper's stated purpose is the thing the PR just proved irrelevant.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -429,6 +456,45 @@ private static String[] critical() {
...
+            .withProgram(program);
+    }
+
+    /**
+     * Program without a single defect.
+     * @return Program without defects
+     */
+    private static String[] flawless() {
```
Favixx, line 482: `flawless` is flawless by suppression: five `+unlint` directives, and the name promises a program that simply has no defects. That matters here because this program is the control in a cross-project cache test — if one of those lints is renamed or retired, the program starts producing defects, the "clean" project's verdict stops being clean, and the test keeps passing while measuring something else entirely.

Either the name should say what it is (`unlinted`, `suppressed`), or the test should assert that the first run really found nothing before relying on it as the control.

Favixx (commented):
The diagnosis is right and I could reproduce it by reading: `Cache.dirSha` walks the given directory, keeps only files passing the filter, and the filter here is `.xmir`-only — pointed at `sourcesDir`, which holds `.eo` files, it matched nothing, so `MessageDigest.digest()` returned the SHA-256 of an empty input, identical for every project on the machine. Moving the root to `3-lint` and adding the plugin version fixes both halves of that. @Thayorns' scope point stands and I am not repeating it.

Three things I would harden, and a handful of test-precision notes.

**The exclusion is by file name, not by path.** `!p.getFileName().equals(wpa)` drops *any* file called `wpa.xmir` anywhere under `3-lint`, and `Place` puts a linted object into a package subdirectory — so an EO object named `wpa` in any package produces `3-lint/foo/wpa.xmir`, which silently leaves the freshness hash. Edits to that object would never invalidate the WPA verdict: the same staleness this PR closes, one layer down.

**The knob that changes the answer is still not in the key.** `wpa(pkg)` runs `.without(this.skipProgramLints…)`, and the log line four lines above the cache call even announces it — but the key is `dirSha(3-lint)` plus the plugin version. Change `eo.skipProgramLints` between two runs of the same unchanged project and the previous verdict is served.

**The version segment is not tested.** The new cross-project test exercises the `dirSha` change; nothing exercises `.resolve(this.version)`, and the assertion in `savesForWholeProgramAnalysisResultsToCache` would pass just as well if the segment were a constant.

The `@todo #5102` parameter count being updated from 13 to 12 in the same breath is a nice touch.

0crat:
@androvonx95 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+32](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+399](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+100](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6151, opened 2026-07-30
fix(#6096): `Walk.java` globs are compiled eagerly to fail fast

### LINKED ISSUE #6096: `Walk.matches` should handle input-related `PatternSyntaxException` from `<includeSources>`/`<excludeSources>` configurations
## What happens

`Walk.matches` (`eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java:126`) builds a glob matcher directly from user input, without catching the pattern being invalid:

```java
private boolean matches(final String text, final Path file) {
    return FileSystems.getDefault().getPathMatcher(String.format("glob:%s", text)).matches(
        Paths.get(
            file.toAbsolutePath().toString().substring(
                this.home.toAbsolutePath().toString().length() + 1
            )
        )
    );
}
```

`text` comes from `<includeSources>`/`<excludeSources>` in the plugin's own pom.xml configuration (`MjRegister.includeSources`/`excludeSources`, both user-authored `Set<String>`). A malformed glob there, e.g. `**/[a-.eo` or `{foo`, makes `FileSystems.getPathMatcher` throw a raw `java.util.regex.PatternSyntaxException`. This propagates up through `Walk.includes`/`excludes`, `MjRegister`, and the plugin's threaded execution, surfacing to the user as an opaque Maven stack trace that never names the offending glob or the `<includeSources>`/`<excludeSources>` parameter that caused it.

This is the same shape of bug as #6028 (`directory.walk` in eo-runtime), but a different module (`eo-maven-plugin` build-time config, not an eo-runtime atom) and a different user-facing surface (a pom.xml parameter, not an EO program).

Separately, on the same line: `getPathMatcher` recompiles the glob's regex on every call, i.e. once per file per glob. `matches` is called from `Walk.includes`/`excludes` inside a `stream().filter(file -> globs.stream().anyMatch(glob -> this.matches(glob, file)))`, so for N files and M globs the same M glob patterns are compiled N times each, instead of once per glob up front.

## What should happen

`Walk.matches` should catch `PatternSyntaxException` and rethrow it as a proper error naming the offending glob and the parameter it came from (mirroring how #6028 wraps this same JDK exception into `ExFailure` for the EO-runtime case). As a secondary improvement, each glob's `PathMatcher` could be compiled once (e.g. cached per glob string) instead of once per file.

h1alexbel:
In this PR I've updated `Walk` to catch and rethrow `PatternSyntaxException` in case supplied glob patterns are invalid. Besides that, simplified program time complexity by doing pattern compilation only once outside of walk loop.

closes #6096


h1alexbel:
@yegor256 take a look, please

yegor256:
@h1alexbel we need EO test too, to reproduce this bug

eo-maven-plugin/src/test/java/org/eolang/maven/WalkTest.java
```diff
@@ -36,4 +38,24 @@ void findsFilesMatchingGlobPattern(@Mktmp final Path temp) throws Exception {
             Matchers.iterableWithSize(count)
         );
     }
+
+    @Test
+    void throwsExceptionOnBadGlobs(@Mktmp final Path temp) throws IOException {
+        new Saved("", temp.resolve("foo/app/0.1/EOfoo/foo.bin")).value();
+        new Saved("", temp.resolve("EOxxx/foo")).value();
+        Assertions.assertThrows(
```
Thayorns, line 44: This test passes on master too, so it does not guard the change. PatternSyntaxException is itself a subclass of IllegalArgumentException, and temp holds two files here, so the old code reached getPathMatcher inside the filter and threw something assertThrows(IllegalArgumentException.class) already accepts. The two Saved files above are also inert now: with matchers compiled before the stream is walked, an empty directory would throw just the same.

eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java
```diff
@@ -114,14 +122,29 @@ private static Collection<Path> regular(final Path dir) throws IOException {
...
+    /**
+     * Compile glob to pattern.
+     * @param glob Glob
+     * @return Matcher for compiled pattern
+     */
+    private static PathMatcher compile(final String glob) {
+        try {
+            return FileSystems.getDefault().getPathMatcher(String.format("glob:%s", glob));
+        } catch (final PatternSyntaxException ex) {
+            throw new IllegalArgumentException(
```
Thayorns: The catch swaps PatternSyntaxException for its own supertype. A caller that wanted to tell a malformed glob apart from any other IllegalArgumentException could do it before this change and cannot now, and the message loses the index the JDK reports while the glob text it adds is already inside the cause's message. If the goal was that a bad glob surfaces from includes rather than from deep inside a filter, moving compilation out of the loop achieved that on its own.

eo-maven-plugin/src/test/java/org/eolang/maven/WalkTest.java
```diff
@@ -36,4 +38,24 @@ void findsFilesMatchingGlobPattern(@Mktmp final Path temp) throws Exception {
...
+            IllegalArgumentException.class,
+            () -> new Walk(temp).includes(new ListOf<>("{foo")),
+            "Exception must be thrown for invalid glob pattern"
+        );
+    }
+
+    @Test
+    void compilesOnceEagerlyInEmptyDir(@Mktmp final Path empty) {
```
Thayorns: The name promises that compilation happens once, but nothing here counts compilations; the assertion only shows that an empty directory still raises. That is the valuable half of the change and it is worth keeping, though a reader chasing a failure will look for the once-ness and not find it.

Thayorns (commented):
Walk now compiles each glob once, before the file stream is touched, and wraps a bad glob into IllegalArgumentException. The eager compilation is a real improvement and the empty-directory test pins it down. Two things weaken the rest: the exception wrapping does not change the type anyone can observe, and the test named after it passes on master unchanged. Small point on the side: the docblock above `matches` still says "Create glob matcher from text" with `@param matcher The pattern, e.g. **/*.java`, while the parameter is now a ready PathMatcher and the method creates nothing.

eo-maven-plugin/src/test/java/org/eolang/maven/WalkTest.java
h1alexbel: @Thayorns how about `throwsOnInvalidGlobsBeforeWalkingFiles`?

Thayorns: @h1alexbel yes, that one says what the test proves: the glob is rejected before any file is walked. Since only one pattern goes in, `throwsOnInvalidGlobBeforeWalkingFiles` reads a touch closer to what happens.

h1alexbel:
@Thayorns updated. Take a look, please

eo-maven-plugin/src/test/java/org/eolang/maven/MjRegisterTest.java
```diff
@@ -68,4 +70,30 @@ void throwsExceptionInCaseSourceDirIsNotSet(@Mktmp final Path temp) {
             )
         );
     }
+
+    @Test
+    void failsWithPatternSyntaxExceptionInCaseInvalidGlobs(@Mktmp final Path temp)
+        throws IOException {
+        new Saved(
```
Thayorns: This Saved file does not participate in the test. The two tests above it point the mojo at src/eo explicitly, but this one leaves sourcesDir at the FakeMaven default, which is the workspace root (FakeMaven.java:184-186), so the file lands somewhere the test never asserts about. It would not matter either way: the glob is rejected before any file is walked, which is exactly what throwsOnInvalidGlobBeforeWalkingFiles proves with an empty directory. A reader will spend a minute working out why the fixture is there before concluding it isn't.

eo-maven-plugin/src/test/java/org/eolang/maven/MjRegisterTest.java
```diff
@@ -68,4 +70,30 @@ void throwsExceptionInCaseSourceDirIsNotSet(@Mktmp final Path temp) {
...
+            Assertions.assertThrows(
+                IllegalStateException.class,
+                () -> new FakeMaven(temp)
+                    .with("includeSources", new SetOf<>("{foo"))
+                    .execute(new FakeMaven.Register()),
+                String.format(
+                    "%s should fail since supplied globs are invalid",
+                    MjRegister.class
+                )
+            ).getCause().getCause().getCause().getClass(),
```
Thayorns: Three getCause() hops pin the test to three separate wrapping layers: Moja.java:112 wraps MojoFailureException into IllegalStateException, exitError in MjSafe wraps ExecutionException into MojoFailureException, and the ExecutionException itself comes from the Future in execWithTimeout. If any one of them stops wrapping, the third getCause() returns null and the test dies with a NullPointerException on this line instead of telling anyone which exception actually came out.

Thayorns (commented):
The earlier points are all addressed: the catch is gone, so a bad glob now surfaces as the JDK's own PatternSyntaxException, and the two WalkTest cases have collapsed into one that proves the glob is rejected before any file is walked. The new MjRegisterTest case covers the same thing end to end through the mojo; its setup carries a little more than it needs.

h1alexbel:
@Thayorns updated. Take a look, please

eo-maven-plugin/src/test/java/org/eolang/maven/MjRegisterTest.java
```diff
@@ -68,4 +70,45 @@ void throwsExceptionInCaseSourceDirIsNotSet(@Mktmp final Path temp) {
...
+                    () -> new FakeMaven(temp)
+                        .with("includeSources", new SetOf<>("{foo"))
+                        .execute(new FakeMaven.Register()),
+                    String.format(
+                        "%s should fail since supplied globs are invalid",
+                        MjRegister.class
+                    )
+                ),
+                PatternSyntaxException.class
+            ).getClass(),
```
Thayorns: cause stops at the first throwable that type.isInstance accepts, so whatever reaches this line is already known to be a PatternSyntaxException and the equalTo underneath can only agree with it. The single outcome that is not agreement is the null that line 112 returns when the chain carries no PatternSyntaxException at all, and then the run ends on a NullPointerException here instead of naming the exception that actually came out of the mojo.

Thayorns (commented):
Since the last look the three getCause() hops are gone and the inert fixture file is gone with them; the walk over the chain is the right shape and the rest of the change is unchanged. One thing about the new helper below.

h1alexbel:
@Thayorns updated. Take a look, please

eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java
```diff
@@ -115,13 +122,22 @@ private static Collection<Path> regular(final Path dir) throws IOException {
...
 
     /**
-     * Create glob matcher from text.
-     * @param text The pattern, e.g. "**&#47;*.java"
-     * @param file The file to match
+     * Compile glob to pattern.
+     * @param glob Glob
+     * @return Matcher for compiled pattern
+     */
+    private static PathMatcher compile(final String glob) {
```
morphqdd: `compile` is `private static`. Convention elsewhere in these repos is no static methods even for pure private helpers, make it an instance method and call it via `this::compile`.

eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java
```diff
@@ -70,11 +74,14 @@ Walk includes(final Collection<String> globs) {
      */
     @SuppressWarnings("PMD.LooseCoupling")
     Walk excludes(final Collection<String> globs) {
+        final List<PathMatcher> matchers = globs.stream()
```
morphqdd: Same `globs.stream().map(Walk::compile).collect(Collectors.toList())` pipeline as `includes()` above (line 56). Worth extracting into one method that both call.

Thayorns (approved):
The only change since my last look sits inside `cause`: the walk up the chain now accumulates into an `Optional<T>` and hands back a typed cause instead of a raw `Throwable`, so the caller gets a `PatternSyntaxException` rather than something it has to narrow itself. I read that commit line by line and found nothing to report in it. Solid work, and this pull request is fabulous. @h1alexbel

eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java
```diff
@@ -52,11 +53,14 @@ private Walk(final Path dir, final List<Path> list) {
      */
     @SuppressWarnings("PMD.LooseCoupling")
     Walk includes(final Collection<String> globs) {
+        final List<PathMatcher> matchers = globs.stream()
```
gemshrine: This three-line compile-all block is duplicated verbatim in `includes` here and in `excludes` at line 77. A small `private static List<PathMatcher> compile(Collection<String> globs)` (or a rename of the current single-glob `compile`) would remove the copy and keep the two filter methods to their one differing line (`anyMatch` vs `noneMatch`).

gemshrine (commented):
Solid fix. Hoisting the compile out of the per-file filter is both the performance win and the behavior fix: because `matchers` is `collect`-ed eagerly, an invalid glob now throws before any file is looked at, and `throwsOnInvalidGlobBeforeWalkingFiles` on an empty dir is exactly the case the old lazy per-file compile would have missed (no files, no compile, no error). Three minor notes, none blocking.

eo-maven-plugin/src/test/java/org/eolang/maven/MjRegisterTest.java
```diff
@@ -68,4 +71,51 @@ void throwsExceptionInCaseSourceDirIsNotSet(@Mktmp final Path temp) {
...
+        Optional<T> found = Optional.empty();
+        Throwable current = err;
+        while (current != null && found.isEmpty()) {
+            if (type.isInstance(current)) {
+                found = Optional.of(type.cast(current));
+            } else {
+                current = current.getCause();
+            }
+        }
+        return found.get();
```
gemshrine: Two small things about this helper and its use. `found.get()` throws `NoSuchElementException` when the type isn't in the cause chain, so if the fix ever regressed and no `PatternSyntaxException` were thrown, the test would die here with a `NoSuchElementException` rather than a readable assertion failure — returning `Optional` (or asserting presence) would report the real problem. Also, since `cause` already selects by `type.isInstance`, the caller's `...getClass(), equalTo(PatternSyntaxException.class)` can effectively only fail via that exception path; asserting on the located cause being present says the same thing more directly.

eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java
gemshrine: Minor, for the commit record: the description says Walk was updated to "catch and rethrow PatternSyntaxException," but there's no try/catch here — `compile` lets the `PatternSyntaxException` propagate and the Mojo wraps it into the `IllegalStateException` the `MjRegisterTest` asserts on. The actual change is that globs are compiled eagerly so an invalid one fails fast; wording it that way matches the code.

h1alexbel: @morphqdd too many refactoring will be required to get rid of all `private static` methods inside `Walk.java` . Maybe we can add puzzle about it? WDYT?

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (10) comments.  Your running score is [+475](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:
@h1alexbel still waiting for EO test here

h1alexbel:
> @h1alexbel still waiting for EO test here

@yegor256 do we need to create new test in `MjParseTest` with EO code snippet as resource, or something else?

h1alexbel:
@yegor256 not sure EO test can be introduced here other than this Maven test (since `Walk.java` is part of `eo-maven-plugin` and we check for Maven configuration properties supplied to `MjRegister` and `Walk`: `includeSources`, `excludeSources`):

```java
@ParameterizedTest
@ValueSource(strings = {"includeSources", "excludeSources"})
void failsWithPatternSyntaxExceptionInCaseInvalidGlobs(
    final String prop, @Mktmp final Path temp
) {
    MatcherAssert.assertThat(
        String.format(
            "Execution must fail because of %s",
            PatternSyntaxException.class
        ),
        MjRegisterTest.cause(
            Assertions.assertThrows(
                IllegalStateException.class,
                () -> new FakeMaven(temp)
                    .with(prop, new SetOf<>("{foo"))
                    .execute(new FakeMaven.Register()),
                String.format(
                    "%s should fail since supplied globs are invalid",
                    MjRegister.class
                )
            ),
            PatternSyntaxException.class
        ).isPresent(),
        Matchers.is(true)
    );
}
```

Please correct me if I'm wrong

h1alexbel:
@yegor256 ^ friendly reminder

h1alexbel:
@yegor256 ping

h1alexbel:
@yegor256 ping

h1alexbel:
@yegor256 

> @yegor256 not sure EO test can be introduced here other than this Maven test (since `Walk.java` is part of `eo-maven-plugin` and we check for Maven configuration properties supplied to `MjRegister` and `Walk`: `includeSources`, `excludeSources`):
> 
> ```java
> @ParameterizedTest
> @ValueSource(strings = {"includeSources", "excludeSources"})
> void failsWithPatternSyntaxExceptionInCaseInvalidGlobs(
>     final String prop, @Mktmp final Path temp
> ) {
>     MatcherAssert.assertThat(
>         String.format(
>             "Execution must fail because of %s",
>             PatternSyntaxException.class
>         ),
>         MjRegisterTest.cause(
>             Assertions.assertThrows(
>                 IllegalStateException.class,
>                 () -> new FakeMaven(temp)
>                     .with(prop, new SetOf<>("{foo"))
>                     .execute(new FakeMaven.Register()),
>                 String.format(
>                     "%s should fail since supplied globs are invalid",
>                     MjRegister.class
>                 )
>             ),
>             PatternSyntaxException.class
>         ).isPresent(),
>         Matchers.is(true)
>     );
> }
> ```
> 
> Please correct me if I'm wrong

@yegor256 ^ reminder

eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java
morphqdd: I'm not asking to refactor every existing private static in Walk.java, just this one: it's new code from this diff, not something pre-existing you'd be touching as a drive-by. Making a brand new method non-static from the start is a one-line change, not a refactor. No puzzle needed for that.

h1alexbel: @morphqdd agree. Updated, take a look again, please

yegor256:
Several CI checks are failing on this PR:

- **runtime (windows-2022, 23)** — failed
- **benchmark** — failed
- **integration** — failed
- **counts (smells, *.java, static|null, 847)** — failed
- **snippets** — failed
- **qulice** — failed

Please investigate and fix these failures before this PR can be merged.

---

yegor256:

PR #6152, opened 2026-07-30
#5969: extract deleted/moved/size/touched into a file/ package

### LINKED ISSUE #5969: Attributes `deleted`, `moved`, `size`, `touched` in `file.eo` unnecessarily lengthen file, suggest moving to `file/` package
`eo-runtime/src/main/eo/file.eo` is 521 lines long and defines four attributes that nothing else in the object needs: `[] > deleted` (file.eo:16), `[target] > moved` (file.eo:68), `[cant-measure] > size` (file.eo:202) and `[] > touched` (file.eo:215). Following the layout of #5501 and the package dispatch of #5486, they belong in a `file/` package directory, alongside the `string/`, `bytes/`, `number/`, `tuple/`, `range/`, `input/`, `output/` and `chunk/` directories that already sit in the same source tree.

The extraction is safe because the dependency runs one way only. What stays behind reaches nothing but `path`: `path > @` (file.eo:13), `as-path` (file.eo:14), `exists` (file.eo:40), `is-directory` (file.eo:52) and `open` (file.eo:86). Across the object body (lines 12-244), `deleted`, `moved` and `touched` occur nowhere except their own definitions, and every `size` other than its own definition is either the local `[size]` parameter of `stream.read` (file.eo:157, file.eo:160) and `input-block.read` (file.eo:163, file.eo:171), or `buffer.size` on bytes (file.eo:184). The four, in turn, read only `path`, `exists`, `is-directory` and `^`, while `moved` builds a `file target` — all pointing outward.

Add `file/deleted.eo`, `file/moved.eo`, `file/size.eo` and `file/touched.eo`, each carrying `+package file` and taking the file as its first argument (`[f] > deleted`, `[f target] > moved`, `[f cant-measure] > size`, `[f] > touched`), then move the matching `++>` and `-->` tests along with them. Two things deserve a check while doing it: `file` decorates a string and strings already answer to `size`, so this leans on the own-package-before-`@` probe landed in #5931; and `directory.eo` decorates a `file` and overrides `deleted` itself, yet inherits `moved`, `size` and `touched` through the decoratee, so package dispatch has to resolve them via the decoratee's forma.

morphqdd:
Closes #5969.

`file.eo` defined four attributes that nothing else in the object needs: `deleted`, `moved`, `size`, `touched`. Following the layout of #5501 and the package dispatch of #5486, moved them into a new `file/` package directory (`file/deleted.eo`, `file/moved.eo`, `file/size.eo`, `file/touched.eo`), each taking the file as its first parameter (`[f] > deleted`, `[f target] > moved`, `[f cant-measure] > size`, `[f] > touched`).

Existing call sites (`some-file.deleted`, `some-file.moved target`, etc.) keep working unchanged: an attribute absent on an object dispatches to the object named after its package (#5486), so `f.deleted` resolves to `file.deleted f` automatically.

Checked both caveats from the issue:
- `file` decorates a `path` (`path > @`), and `path` itself eventually decorates a string, so a bare `f.size`/`f.deleted` etc. could in principle resolve through the decoratee before reaching the package - this already works correctly on current master since #5931 (own-package-before-`@` probe), which landed before this PR.
- `directory` decorates `file` and does not override any of the four extracted attributes (verified directly in `directory.eo` - it only overrides `is-directory`), so `some-directory.moved`/`.size`/`.touched`/`.deleted` all still resolve through the decoratee down to `file`'s package. Verified by running `EOdirectoryTest` (4 tests, all passing, unchanged).

Moved each attribute's own `++>` tests along with it (`can-delete-a-file-twice`, `can-return-itself-after-deleting`, `can-tell-that-a-file-is-gone-after-deleting` -> `file/deleted.eo`; `can-move-a-file` -> `file/moved.eo`; `can-measure-an-empty-file-after-touching`, `rejects-sizing-an-absent-file`, `stops-on-sizing-an-absent-file` -> `file/size.eo`; `can-return-itself-after-touching`, `can-touch-a-file`, `can-touch-a-file-twice` -> `file/touched.eo`). Tests that use one of these four as setup for testing something else (e.g. `open`, `is-directory`) stayed in `file.eo`.

Also inlined four single-use locals (`outcome` in both `deleted`/`moved`, `unix`/`win` in `deleted`, `closed` in `touched`) that `eo:lint`'s `redundant-object` rule flagged once they moved out of `file.eo`'s file-level `+unlint redundant-object` - fixed genuinely rather than re-adding the suppression to the new files.

`mvn -pl eo-runtime test` - 1810 tests, all passing, 0 failures (full module run, no regressions). `EOfileTest` (29), `EOdirectoryTest` (4), and the four new `EO_file` package test classes (`EOdeletedTest`, `EOmovedTest`, `EOsizeTest`, `EOtouchedTest`, 10 tests total) all pass. qulice clean.

### DIFF
--- a/eo-runtime/src/main/eo/file.eo
+++ b/eo-runtime/src/main/eo/file.eo
@@ -14,30 +14,6 @@
   path > @
   @. > as-path
     Q.path path
-  [] > deleted
-    exists.if > @
-      if.
-        outcome.eq 0
-        ^
-        T
-          "Cant delete the file '%s': %s".printf
-            * path removed.output
-      ^
-    removed.code > outcome
-    called. > removed
-      os.is-windows.if
-        win32 *1
-          win
-          path
-        posix *1
-          unix
-          path
-    is-directory.if > unix
-      "rmdir"
-      "unlink"
-    is-directory.if > win
-      "_rmdir"
-      "_unlink"
   eq. > [] > exists
     code.
       os.is-windows.if
@@ -66,24 +42,6 @@
         posix *1
           "stat"
           path
-  [target] > moved
-    if. > @
-      outcome.eq 0
-      file target
-      T
-        "Cant move the file '%s' to '%s': %s".printf
-          * path target renamed.output
-    renamed.code > outcome
-    called. > renamed
-      os.is-windows.if
-        win32 *1
-          "rename"
-          path
-          target
-        posix *1
-          "rename"
-          path
-          target
   [mode scope cant-open] > open
     known.not.if > @
       T
@@ -200,49 +158,6 @@
     not. > writable
       as-bool.
         access.eq "r"
-  [cant-measure] > size
-    if. > @
-      info.code.eq 0
-      info.output.size
-      cant-measure
-    called. > info
-      os.is-windows.if
-        win32 *1
-          "_stat64"
-          path
-        posix *1
-          "stat"
-          path
-  [] > touched
-    exists.if > @
-      ^
-      if.
-        descriptor.eq -1
-        T
-          "Cant touch the file '%s': %s".printf
-            * path created.output
-        seq *
-          closed
-          ^
-    code. > closed
-      os.is-windows.if
-        win32 *1
-          "_close"
-          descriptor
-        posix *1
-          "close"
-          descriptor
-    called. > created
-      os.is-windows.if
-        win32 *1
-          "_creat"
-          path
-          win32.mode
-        posix *1
-          "creat"
-          path
-          posix.mode
-    created.code > descriptor
 
   eq. ++> can-append-data-to-a-file
     size.
@@ -253,20 +168,6 @@
         "a"
         f.write "!" > [f]
     13
-
-  mktemp.tmpfile.deleted.deleted.exists.not ++> can-delete-a-file-twice
-
-  mktemp.tmpfile.deleted.touched.size.eq ++> can-measure-an-empty-file-after-touching
-    0
-
-  ++> can-move-a-file
-    and. > @
-      exists.
-        temp.moved
-          "%s.dest".printf
-            * temp.path
-      temp.exists.not
-    mktemp.tmpfile > temp
 
   ++> can-read-from-a-file
     eq. > @
@@ -332,26 +233,12 @@
           "bar"
     mktemp.as-path.resolved "foo/bar" > resolved
 
-  ++> can-return-itself-after-deleting
-    temp.deleted.path.eq temp.path > @
-    mktemp.tmpfile > temp
-
-  ++> can-return-itself-after-touching
-    temp.deleted.touched.path.eq temp.path > @
-    mktemp.tmpfile > temp
-
-  mktemp.tmpfile.deleted.exists.not ++> can-tell-that-a-file-is-gone-after-deleting
-
   not. ++> can-tell-that-an-absent-file-does-not-exist
     exists.
       file "absent.txt"
 
   is-directory. ++> can-tell-that-the-current-directory-is-a-directory
     file "."
-
-  mktemp.tmpfile.deleted.touched.exists ++> can-touch-a-file
-
-  mktemp.tmpfile.deleted.touched.touched.exists ++> can-touch-a-file-twice
 
   exists. ++> can-touch-an-absent-file-on-opening-for-appending
     mktemp.tmpfile.deleted.open
@@ -474,10 +361,6 @@
       f > [f]
       "recovered" > [reason]
 
-  eq. ++> rejects-sizing-an-absent-file
-    -1
-    mktemp.tmpfile.deleted.size -1
-
   mktemp.tmpfile.deleted.is-directory --> stops-on-checking-the-directory-of-an-absent-file
 
   seq * --> stops-on-opening-with-a-wrong-mode
@@ -501,8 +384,6 @@
   mktemp.tmpfile.deleted.open --> stops-on-reading-or-writing-a-missing-file
     "r+"
     f > [f]
-
-  mktemp.tmpfile.deleted.size --> stops-on-sizing-an-absent-file
 
   tmpfile. --> stops-on-touching-a-temp-file-in-an-absent-directory
     @.
--- a/eo-runtime/src/main/eo/file/deleted.eo
+++ b/eo-runtime/src/main/eo/file/deleted.eo
@@ -1,0 +1,35 @@
+# Deletes the file `f` from the filesystem.
+# Returns the same file, unchanged, if it does not exist.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package file
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[f] > deleted
+  f.exists.if > @
+    if.
+      removed.code.eq 0
+      f
+      T
+        "Cant delete the file '%s': %s".printf
+          * f.path removed.output
+    f
+  called. > removed
+    os.is-windows.if
+      win32 *1
+        f.is-directory.if "_rmdir" "_unlink"
+        f.path
+      posix *1
+        f.is-directory.if "rmdir" "unlink"
+        f.path
+
+  mktemp.tmpfile.deleted.deleted.exists.not ++> can-delete-a-file-twice
+
+  ++> can-return-itself-after-deleting
+    temp.deleted.path.eq temp.path > @
+    mktemp.tmpfile > temp
+
+  mktemp.tmpfile.deleted.exists.not ++> can-tell-that-a-file-is-gone-after-deleting
--- a/eo-runtime/src/main/eo/file/moved.eo
+++ b/eo-runtime/src/main/eo/file/moved.eo
@@ -1,0 +1,39 @@
+# Moves the file `f` to `target` on the filesystem, returning the file
+# at its new location.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package file
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[f target] > moved
+  if. > @
+    renamed.code.eq 0
+    file target
+    T
+      "Cant move the file '%s' to '%s': %s".printf
+        *
+          f.path
+          target
+          renamed.output
+  called. > renamed
+    os.is-windows.if
+      win32 *1
+        "rename"
+        f.path
+        target
+      posix *1
+        "rename"
+        f.path
+        target
+
+  ++> can-move-a-file
+    and. > @
+      exists.
+        temp.moved
+          "%s.dest".printf
+            * temp.path
+      temp.exists.not
+    mktemp.tmpfile > temp
--- a/eo-runtime/src/main/eo/file/size.eo
+++ b/eo-runtime/src/main/eo/file/size.eo
@@ -1,0 +1,32 @@
+# Size of the file `f` in bytes, or `cant-measure` if the file
+# does not exist.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package file
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[f cant-measure] > size
+  if. > @
+    info.code.eq 0
+    info.output.size
+    cant-measure
+  called. > info
+    os.is-windows.if
+      win32 *1
+        "_stat64"
+        f.path
+      posix *1
+        "stat"
+        f.path
+
+  mktemp.tmpfile.deleted.touched.size.eq ++> can-measure-an-empty-file-after-touching
+    0
+
+  eq. ++> rejects-sizing-an-absent-file
+    -1
+    mktemp.tmpfile.deleted.size -1
+
+  mktemp.tmpfile.deleted.size --> stops-on-sizing-an-absent-file
--- a/eo-runtime/src/main/eo/file/touched.eo
+++ b/eo-runtime/src/main/eo/file/touched.eo
@@ -1,0 +1,47 @@
+# Touches the file `f`, creating it if it does not exist yet.
+# Returns the same file, unchanged, if it already exists.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package file
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[f] > touched
+  f.exists.if > @
+    f
+    if.
+      descriptor.eq -1
+      T
+        "Cant touch the file '%s': %s".printf
+          * f.path created.output
+      seq *
+        code.
+          os.is-windows.if
+            win32 *1
+              "_close"
+              descriptor
+            posix *1
+              "close"
+              descriptor
+        f
+  called. > created
+    os.is-windows.if
+      win32 *1
+        "_creat"
+        f.path
+        win32.mode
+      posix *1
+        "creat"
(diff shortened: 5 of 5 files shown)


eo-runtime/src/main/eo/file/moved.eo
```diff
@@ -0,0 +1,39 @@
...
+        "rename"
+        f.path
+        target
+      posix *1
+        "rename"
+        f.path
+        target
+
+  ++> can-move-a-file
+    and. > @
```
Thayorns, line 33: This one checks two separate things at once: that the destination came into existence, and that the source stopped existing. When it goes red it says only that the conjunction is false, so whoever picks it up has to re-run it by hand to learn whether the move failed to create the target or failed to remove the original. It arrives here unchanged from file.eo, but the new file is the natural moment to notice it.

Thayorns (commented):
Four attributes leave file.eo for a new file/ package, each taking the file as its first parameter, and the ten tests that belonged to them travel with them. The extraction is faithful: the win32/posix branches in deleted.eo keep the same _rmdir/_unlink and rmdir/unlink pairing the inlined unix and win locals used to give them, the inlined closed in touched.eo and outcome in both deleted.eo and moved.eo are pure substitutions, and the remaining size. call sites in file.eo plus the touched. and deleted. calls in directory.eo:32-38 and :47 exercise the package dispatch this now relies on. The one caveat named in the issue is the only thing left uncovered: directory.eo carries no test that applies one of these four to a directory object, so the two-step resolution through the decoratee into file's package rests on reading rather than on a red build if it ever breaks.

eo-runtime/src/main/eo/file.eo
```diff
@@ -14,30 +14,6 @@
   path > @
   @. > as-path
     Q.path path
-  [] > deleted
```
Favixx, line 17: `pr-size` fails on this PR: `ADDITIONS: 153`, `DELETIONS: 119`, `MAX_LINES_CHANGED: 200`, so 272 against the limit and `exit 1`.

The good news is that the four objects do not depend on each other, so the split is mechanical: this hunk plus `file/deleted.eo` plus the three `deleted` tests is about 55 lines, and `moved`, `size` and `touched` are the same size or smaller. Four PRs, each green, each reviewable on its own, and each one independently revertable if the package dispatch turns out to misbehave for one attribute but not the others. Doing it in one go also means the `directory` question @Thayorns raised has to be answered for all four at once.

eo-runtime/src/main/eo/file.eo
```diff
@@ -200,49 +158,6 @@
     not. > writable
       as-bool.
         access.eq "r"
-  [cant-measure] > size
```
Favixx, line 203: Worth stating in the description, because it is the blast radius of this change: after the move `file.eo` still calls `size.` in eight of its own tests (`can-append-data-to-a-file`, the two truncate tests, `can-write-data-to-a-file`, `can-write-to-a-file-sequentially` and three more) and `mktemp.tmpfile.deleted` in nine. All of those now go through package dispatch rather than through an attribute of `file`.

That is a fine outcome when it works, but it means a regression in dispatch does not show up as "the size test is red" — it shows up as seventeen unrelated file tests failing at once, in a file that no longer contains the code at fault. A one-line note in the PR description saying which tests are now dispatch-dependent would save the next person that walk.

eo-runtime/src/main/eo/file/deleted.eo
```diff
@@ -0,0 +1,35 @@
+# Deletes the file `f` from the filesystem.
+# Returns the same file, unchanged, if it does not exist.
```
Favixx, line 2: "Returns the same file, unchanged, if it does not exist" reads as if the non-existent case is special, but both branches of `f.exists.if` return `f` — the object always answers with the file it was given, or dies. Something like "Returns `f` in both cases; dies if the file exists and cannot be removed" describes all three outcomes in the same number of words.

The distinction matters here more than usual, because the sentence is what a caller sees when deciding whether `f.deleted.path` is safe to chain — and `can-return-itself-after-deleting` right below exists precisely to pin that down.

eo-runtime/src/main/eo/file/deleted.eo
```diff
@@ -0,0 +1,35 @@
...
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package file
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[f] > deleted
```
Favixx, line 11: There are now two objects named `deleted` reachable from a filesystem object, and they do different things: this one calls `rmdir` once, `directory.deleted` walks `**` and removes the contents first. Which one you get depends on whether you started from a `directory` or from a `file`, since the own-package probe finds `directory.deleted` before it reaches `file` through `@`.

That is the intended resolution and I am not asking to change it, but as an inner attribute of `file` this object could only ever be applied to a file, and as `file.deleted` it can be applied to anything answering `exists`, `is-directory` and `path` — including a directory object, where it will quietly fail on a non-empty directory. One clause in the doc comment ("does not recurse; see `directory.deleted`") is enough to keep the two apart for whoever greps for `deleted` next.

eo-runtime/src/main/eo/file/deleted.eo
```diff
@@ -0,0 +1,35 @@
...
+      removed.code.eq 0
+      f
+      T
+        "Cant delete the file '%s': %s".printf
+          * f.path removed.output
+    f
+  called. > removed
+    os.is-windows.if
+      win32 *1
+        f.is-directory.if "_rmdir" "_unlink"
```
Favixx, line 23: `unix` and `win` were named attributes in `file.eo`; inlining them here puts a conditional inside a tuple element, where it is easy to misread which of `_rmdir`/`_unlink` belongs to which platform, and it duplicates the `is-directory` test across the two branches.

The house style keeps the name in place rather than dropping it — `directory/made.eo` writes `called. >> created` exactly for this. `f.is-directory.if "rmdir" "unlink" > unix` (and the `win32` twin) above `called. > removed` restores what the original said while keeping everything in one object.

eo-runtime/src/main/eo/file/deleted.eo
```diff
@@ -0,0 +1,35 @@
...
+        f.path
+      posix *1
+        f.is-directory.if "rmdir" "unlink"
+        f.path
+
+  mktemp.tmpfile.deleted.deleted.exists.not ++> can-delete-a-file-twice
```
Favixx, line 29: These three cover the success paths and the idempotence, and none of them reaches the `T "Cant delete the file '%s': %s"` branch — same for `touched` and `moved` in this PR, while `size` arrives with `rejects-sizing-an-absent-file` and `stops-on-sizing-an-absent-file` for exactly this reason.

A `--> stops-on-deleting-an-undeletable-file` is admittedly harder to write portably than the sizing one, but the message contains `f.path` and `removed.output`, and if the argument order in that `printf` were ever wrong nothing in the suite would notice. Deleting a non-empty directory through `file.deleted` (see my other comment) would trip it on both platforms.

eo-runtime/src/main/eo/file/moved.eo
```diff
@@ -0,0 +1,39 @@
+# Moves the file `f` to `target` on the filesystem, returning the file
+# at its new location.
```
Favixx, line 2: None of the four comments mentions the failure mode, and it is the same in all four: the object does not return an error value, it dataizes `T` with a `printf`ed message and the program stops. For `size` a caller can opt out via `cant-measure`; for `deleted`, `moved` and `touched` there is no way to observe the failure without the program ending.

One sentence per file — "dies if the rename fails" — makes the contract legible without the reader having to open the body. Right now the only place that behaviour is written down is `stops-on-sizing-an-absent-file`, which covers one of the four.

eo-runtime/src/main/eo/file/moved.eo
```diff
@@ -0,0 +1,39 @@
...
++home https://github.com/objectionary/eo
++package file
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[f target] > moved
+  if. > @
+    renamed.code.eq 0
+    file target
```
Favixx, line 14: `file target` builds the returned object from `target`, which callers pass as a string (`"%s.dest".printf` in the test below), whereas the `f` that went in wraps a `path`. So `f.moved(t).path` and `f.path` are not the same kind of thing, and `moved` is the only one of the four whose result is a *different* object rather than `f`.

This comes unchanged from `file.eo`, but there the parameter was documented by its surroundings; here `[f target]` is a public signature with a two-line comment above it, and the comment says only "to `target`". Saying that `target` is a path string, and that the result is a fresh `file` at that path rather than `f` relocated, is the difference between the signature being usable and being guessable.

eo-runtime/src/main/eo/file/moved.eo
```diff
@@ -0,0 +1,39 @@
...
+      win32 *1
+        "rename"
+        f.path
+        target
+      posix *1
+        "rename"
+        f.path
+        target
+
+  ++> can-move-a-file
```
Favixx, line 32: The only test of the only object in this PR whose body was genuinely reshaped — the `printf` arguments went from `* path target renamed.output` on one line to a three-item vertical tuple — and it exercises the branch where those arguments are never used.

`mktemp.tmpfile.deleted.moved "whatever" --> stops-on-moving-an-absent-file` would cost one line and cover the reflow. As it stands, a transposition of `f.path` and `target` in that tuple ships green.

eo-runtime/src/main/eo/file/size.eo
```diff
@@ -0,0 +1,32 @@
+# Size of the file `f` in bytes, or `cant-measure` if the file
+# does not exist.
```
Favixx, line 2: The fallback is not tied to non-existence: `info.code.eq 0` is false for every non-zero return of `stat`/`_stat64`, so `cant-measure` is what a caller gets for a permission failure, a name that is too long, a symlink loop, or a path that crosses a broken mount — not only for a file that is not there.

The parameter is already named for the general case; the comment is what narrows it. "or `cant-measure` if the file cannot be measured" is circular but true, and "if `stat` fails" is shorter and truer still.

eo-runtime/src/main/eo/file/size.eo
```diff
@@ -0,0 +1,32 @@
...
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package file
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[f cant-measure] > size
```
Favixx, line 11: Now that the four are siblings in one package it is visible that they disagree about failure: `size` takes a fallback and hands it back, while `deleted`, `moved` and `touched` die. A caller who wants "try to delete, carry on if you cannot" has no equivalent of `cant-measure`.

I am not suggesting this PR change any behaviour — it should not. But the asymmetry was invisible while these were four attributes of one object and is now a property of a package's public surface, so it is worth a line in #5969 (or a follow-up) recording that it is deliberate rather than an accident of who wrote which attribute first.

eo-runtime/src/main/eo/file/size.eo
```diff
@@ -0,0 +1,32 @@
...
+  called. > info
+    os.is-windows.if
+      win32 *1
+        "_stat64"
+        f.path
+      posix *1
+        "stat"
+        f.path
+
+  mktemp.tmpfile.deleted.touched.size.eq ++> can-measure-an-empty-file-after-touching
```
Favixx, line 25: This test now depends on two other files in the package: `file/deleted.eo` and `file/touched.eo`. Before the split all three lived in `file.eo`, so a failure was local; now a bug in `touched` reddens a test that is filed under `size`, and whoever picks it up starts in the wrong file.

`open` with `"w"` produces an empty file without leaving the object under test — `mktemp.tmpfile.open "w" (true > [f])` in place of `deleted.touched` — and keeps the assertion about `size` measuring zero exactly as it is.

eo-runtime/src/main/eo/file/touched.eo
```diff
@@ -0,0 +1,47 @@
+# Touches the file `f`, creating it if it does not exist yet.
+# Returns the same file, unchanged, if it already exists.
```
Favixx, line 2: Two things in one sentence. As with `deleted`, both branches return `f`, so "returns the same file, unchanged, if it already exists" describes the whole object rather than one case.

More usefully: POSIX `touch` updates the modification time of a file that already exists, and this object deliberately does not — `f.exists.if` returns `f` untouched and never calls `_creat`/`creat`. Anyone reading the name and not the body will assume the mtime moved. That difference deserves the second line more than the return value does.

eo-runtime/src/main/eo/file/touched.eo
```diff
@@ -0,0 +1,47 @@
...
+[f] > touched
+  f.exists.if > @
+    f
+    if.
+      descriptor.eq -1
+      T
+        "Cant touch the file '%s': %s".printf
+          * f.path created.output
+      seq *
+        code.
```
Favixx, line 20: `closed` was a named attribute in `file.eo`; inlining it drops an eight-line syscall into the middle of a `seq`, where the reader has to work out that the first element of the sequence is "close the descriptor we just opened" and the second is the result.

Same remark as in `deleted.eo`, and the same fix: `code. >> closed` keeps the name where the value is used, the way `directory/made.eo` does with `called. >> created`. Here it matters slightly more, because the name is the only thing explaining why a `creat` is followed by a `close` at all.

Favixx (changes requested):
The extraction itself reads correctly to me: `f` replaces `^` everywhere it should, the `unix`/`win` pair in `deleted` keeps the same `_rmdir`/`_unlink` against `rmdir`/`unlink` pairing after being inlined, `closed` in `touched` and `outcome` in `deleted`/`moved` are substituted without changing what is dataized, and the ten tests travel with the objects they belong to. `+package file` and the `[f] > …` shape match `directory/deleted.eo` and `directory/made.eo`, so the layout is the established one. I am not going to repeat @Thayorns' point about `directory` having no test for the two-step resolution — it stands, and it is the sharpest thing said about this PR so far.

Three things I would like addressed.

**`pr-size` is red and it is not advisory:** 153 + 119 = 272 against a 200-line limit, exit code 1. Whatever else happens, this cannot merge in one piece. The split is unusually easy here because the four objects are independent — details inline.

**All four doc comments say something that is not quite what the code does.** `deleted` and `touched` describe a return value that differs by case when both branches return `f`; `size` attributes `cant-measure` to non-existence when it actually covers every non-zero `stat`; none of the four mentions that failure means dying with a `printf`ed message, which is the only observable difference between them and their POSIX namesakes. These four comments are the entire public documentation of four new objects, and they are being written from scratch here, so this is the moment.

**The error branches are untested.** `size` arrives with both a `rejects-` and a `stops-` test; `deleted`, `moved` and `touched` arrive with happy paths only, and `moved`'s failure branch is the one whose shape actually changed in this diff.

Nothing here is a defect in the move — the behaviour that ships is the behaviour that shipped before.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (272 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2872](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+406](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+88](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6153, opened 2026-07-30
fix(#6139): preserve percent signs in bottom causes

### LINKED ISSUE #6139: `PhTerminator` incorrectly interprets percent signs as Java format directives in causes
## What happens

A bottom object does not preserve a cause containing a percent sign. When it is forced, the cause is passed to `String.format` as if it were a format template.

| cause stored in `⊥` | actual exception | expected exception |
| --- | --- | --- |
| `"100% complete"` | `java.util.FormatFlagsConversionMismatchException: Conversion = c, Flags = <space>` | `org.eolang.ExFailure: 100% complete` |

The failure was reproduced by putting a UTF-8 cause into `new PhTerminator()` at position `0` and forcing it with `delta()`. The substituted exception comes from Java formatting, not from the EO cause.

## Why this is a bug

[`PhTerminator.delta()`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/PhTerminator.java#L106-L114) correctly extracts the user-supplied text at line 111, but then calls `new ExFailure(reason)` at line 113. [`ExFailure`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/ExFailure.java#L23-L25) always delegates its first argument to `String.format`, even when no replacement arguments are supplied.

Consequently a cause is treated as a format string. A literal `%` is legal text in an EO error message, but it is interpreted by Java as the beginning of a conversion. Escaping every individual message as `%%` is only a workaround and fails for dynamically constructed causes.

## Steps to reproduce

1. Create a `PhTerminator`.
2. Put a string cause such as `"100% complete"` at position `0`.
3. Force the terminator by dataizing it (or call `delta()` in a runtime probe).
4. Observe `FormatFlagsConversionMismatchException` instead of `ExFailure` containing the original cause.

## Expected behavior

`PhTerminator` should report the stored cause verbatim, regardless of `%` characters. For example, it can construct the failure with a constant format string and the cause as an argument, such as `new ExFailure("%s", reason)`. This makes `T "100% complete"` terminate with the intended EO message and removes the need for callers to pre-escape percentages.

webhop123:
Closes #6139

## What changed

`PhTerminator` now passes a constant `%s` format to `ExFailure`, with the
bottom cause supplied as data. This preserves literal percent signs instead
of treating user text as a Java format string.

A regression test forces a bottom with the cause `100% complete` and verifies
that it throws `ExFailure` with the exact original message.

## Verification

- `mvn -pl eo-runtime -Dtest=PhTerminatorTest -DskipITs test`
  - 14 tests, 0 failures, 0 errors
- `mvn -pl eo-runtime clean verify -PskipTests -Pqulice`
  - EO formatting, linting, and Qulice completed without source violations
  - the later module-only Windows validator failed on missing generated
    Unicode class names (`EOwin32$EOphi.class` and `EOposix$EOphi.class`)

@yegor256


Thayorns (approved):
PhTerminator.delta() now throws ExFailure with a constant "%s" template and the dataized cause as an argument, so a cause containing a percent sign reaches the user verbatim instead of being re-parsed as a Java format directive. The change is minimal and correct: ExFailure's varargs constructor is the one selected here, the null-cause branch is a constant and unaffected, and no other call site in eo-runtime feeds a dynamic string in as a format template. The regression test forces a bottom carrying "100% complete" and pins the resulting message exactly, which fails on the old code and passes on the new.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (16) hits-of-code; +4 to give you at least something.  Your running score is [+377](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@webhop123 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+24](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6155, opened 2026-07-30
Let dead input and output receive their origin

### LINKED ISSUE #6084: Void-less `input.dead` and `output.dead` package members disrupt receiver form of dispatch, breaking contract
Two package members in the runtime declare no void attributes: `input/dead.eo:11` and `output/dead.eo:10`, both reading `[] > dead`. The "one object, many files" model promises that the receiver form and the namespace form of a member are two spellings of one thing — `"42".as-number` is `string.as-number "42"` — but for a void-less member the receiver half of that equivalence can never work: `PhDefault.extension` binds the receiver into position 0, and with no free attribute to receive it the dispatch fails.

```
[] > dead
```

```java
// eo-runtime/src/main/java/org/eolang/PhDefault.java:437-442
taken.put(0, this);
} catch (final ExAbstract ex) {
    throw new ExFailure(
        String.format(
            "Object '%s' takes no arguments, so it can't be applied to '%s' via the implicit '%s' form",
```

#5512 made this failure clean — before it, the dispatch did not even terminate — but the library objects it named are still there, so `inp.dead` on any input terminates the program while `Q.input.dead` works fine, an asymmetry a caller cannot predict without reading the member's source. Every other member of every package satisfies the equivalence; these two are namespace-only constants living in a namespace whose contract says "every member is also a method".

The smallest fix is to move the two constants out of their packages (they are ready-made *instances* of `input` and `output`, not extensions *of* one), or alternatively to give each a receiver void it deliberately ignores; either way, a lint forbidding void-less package members would keep the invariant from regressing.

maxonfjvipon:
A member of a package is also a method of the object the package is named
after: `string.as-number "42"` and `"42".as-number` are two spellings of one
thing, because the runtime binds the receiver into the member's first
position. Both `dead` objects broke that equivalence — they declared no void
attributes, so there was no slot for the receiver, and `inp.dead` on any input
died with `Object 'Φ.input.dead' takes no arguments` while `Q.input.dead`
worked fine.

Each one now takes an `origin` it ignores, since a dead input reads from
nowhere no matter what it was taken off:

```
[origin] > dead
  [size] > read
    [] >> input-block
      -- > @
      input-block > [size] > read
    | > @
```

Both files gained a test that reaches the object in receiver form; without
the void they fail with the error above. `PhDefaultTest` had a test that used
`input.dead` as its example of a nullary package member, and no package member
is nullary any more — the fixture it would need must sit in an `EO_*` Java
package, where the `// NOPMD` and `@checkstyle` pair for the package name costs
one entry against the `suppressions` budget, which is already full. So the test
goes, and the two receiver-form tests carry the invariant instead. The clean
error in `PhDefault.extension` stays, since a void-less member is still
possible in code outside this repository.

Nothing moved and no name changed, so the two call sites in `input/copied.eo`
and `input/as-bytes.eo` stay as they are. The lint that would keep this from
regressing belongs to objectionary/lints, where `design/no-attribute-formation`
already reports the shape but only as an experimental warning over every
formation, package member or not; that is filed as objectionary/lints#1170.

Closes #6084


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+756](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6156, opened 2026-07-30
#6150: refuse a name suffix with nothing but spaces after it

### LINKED ISSUE #6150: `Suffix.named` incorrectly accepts empty name when two spaces follow `>` symbol
## What happens

A `> name` suffix with two or more spaces after the `>` parses as a name that is empty, with no error.

`Suffix.skipSpace` skips at most one space:

```java
private static int skipSpace(final String tail, final int from) {
    int idx = from;
    if (idx < tail.length() && tail.charAt(idx) == ' ') {
        idx = idx + 1;
    }
    return idx;
}
```

and `Suffix.named` only checks that something is left after that one space:

```java
final int begin = Suffix.skipSpace(tail, from);
if (begin >= tail.length()) {
    throw new ParseError(span.line(), home + from, "name suffix requires a name");
}
int idx = Suffix.skipName(tail, begin);
```

With two spaces, one is skipped, `begin` lands on the second, the guard passes, and `skipName` stops immediately because a space terminates a name. `name` comes out empty and nothing rejects it. The javadoc of `skipName` says the text is "empty when a terminator already sits at `from`", so the empty result is expected there, but `named` never tests for it.

The boundary is exactly one space:

```
  42 >          error: name suffix requires a name
  42 >          error: name suffix requires a name      (one trailing space)
  42 >          no error, name=""                       (two trailing spaces)
  42 >  x       error: trailing garbage after name suffix
```

Three forms are affected. For

```
# Sample.

+package foo

[] > app
  42 >␣␣
```

the parser emits `<o base="Φ.number" line="6" loc="Φ.foo.app." name="" pos="2">`, note the empty `name` and the trailing dot in `loc`. The same happens for a void attribute (`? >␣␣`), and for a top-level object (`[] >␣␣`), which comes out as `<o line="5" loc="Φ.foo." name="" pos="0"/>`.

The test-attribute suffix already has this check: `[] +>␣␣` and `[] ->␣␣` both fail with "test attribute requires a name". The `>` branch just does not do the same.

The inner binding case survives the whole pipeline. Linting reports nothing for it, and the transpiler emits an attribute with an empty name into the generated Java:

```java
@XmirObject(name = "main", oname = "main", source = "")
public final class EOmain extends PhDefault {
  public EOmain() {
    this.add("", new AtOnce(new AtComposite(this, new Function<>() {
```

No EO code can ever dispatch on that attribute. The top-level case is caught later by `non-kebab-name` and `validate-object-name`, but only after the parser has already produced a nameless object.

## What should happen

`Suffix.named` rejects an empty name with the same "name suffix requires a name" error it already raises for a missing one, so that `>` behaves like `+>` and `->` do.

Related to #6147, which is the same class of hole (an empty name accepted) in a different place, `LnMeta`. Different code, different fix.

morphqdd:
Closes #6150

`skipSpace` steps over one space, and the only emptiness check was `begin >= tail.length()`, which two spaces slip past. The name then came out empty and nothing rejected it.

The check now looks past every space before the name is read: if only spaces are left, there is no name and it says so.

It deliberately does not treat two spaces followed by a name the same way. `space-in-tail.yaml` already pins that case to "trailing garbage after name suffix", and this keeps it. The hole was only the case where nothing follows at all.

The test is a typo pack, so it runs through both `checksTypoPacks` and `checksTypoMessages`. It fails on master and passes here. Input is written as a quoted scalar rather than a block, because the case needs trailing spaces and yamllint forbids them in the file.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -553,19 +553,22 @@ final class Suffix {
     }
 
     /**
-     * Whether nothing but spaces is left, so no name can follow. Unlike
+     * Whether nothing but blanks is left, so no name can follow. Unlike
      * {@link #skipSpace(String, int)}, which steps over the single space
-     * a suffix is allowed to have, this looks past every space: two of
-     * them followed by a name is a separate mistake, reported later as
-     * trailing garbage, while two of them followed by nothing leaves the
-     * name empty and has to be refused right here.
+     * a suffix is allowed to have, this looks past every space and tab:
+     * blanks followed by a name are a separate mistake, reported later
+     * as trailing garbage, while blanks followed by nothing leave the
+     * name empty and have to be refused right here. A tab counts because
+     * {@link #terminates(char)} ends a name on it just as a space does,
+     * and nothing upstream keeps tabs out of the tail.
      * @param tail Tail substring
      * @param from Index to look from
-     * @return True if only spaces remain
+     * @return True if only blanks remain
      */
     private static boolean blank(final String tail, final int from) {
         int idx = from;
-        while (idx < tail.length() && tail.charAt(idx) == ' ') {
+        while (idx < tail.length()
+            && (tail.charAt(idx) == ' ' || tail.charAt(idx) == '\t')) {
             idx = idx + 1;
         }
         return idx >= tail.length();
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -442,13 +442,13 @@ final class Suffix {
     private static Parsed named(
         final String tail, final int from, final Span span, final int home
     ) {
-        final int begin = Suffix.skipSpace(tail, from);
-        if (begin >= tail.length()) {
+        if (Suffix.blank(tail, from)) {
             throw new ParseError(
                 span.line(), home + from,
                 "name suffix requires a name"
             );
         }
+        final int begin = Suffix.skipSpace(tail, from);
         int idx = Suffix.skipName(tail, begin);
         final String name = tail.substring(begin, idx);
         if (name.codePoints().anyMatch(cp -> cp == 0x1F335)) {
@@ -550,6 +550,25 @@ final class Suffix {
             );
         }
         return Suffix.typeAtom(raw, span, home + after);
+    }
+
+    /**
+     * Whether nothing but spaces is left, so no name can follow. Unlike
+     * {@link #skipSpace(String, int)}, which steps over the single space
+     * a suffix is allowed to have, this looks past every space: two of
+     * them followed by a name is a separate mistake, reported later as
+     * trailing garbage, while two of them followed by nothing leaves the
+     * name empty and has to be refused right here.
+     * @param tail Tail substring
+     * @param from Index to look from
+     * @return True if only spaces remain
+     */
+    private static boolean blank(final String tail, final int from) {
+        int idx = from;
+        while (idx < tail.length() && tail.charAt(idx) == ' ') {
+            idx = idx + 1;
+        }
+        return idx >= tail.length();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Thayorns (commented):
The change replaces the end-of-tail test in `named` with a `blank` helper that looks past every space, so `> ` followed by nothing but spaces is refused instead of producing an empty name, and it keeps the two-spaces-then-name case reporting as trailing garbage. A typo pack pins the new error at line 2. The guard covers spaces only, which leaves the same empty name reachable through a tab.

eo-parser/src/main/java/org/eolang/parser/Suffix.java
```diff
@@ -552,6 +552,25 @@ private static String signature(
...
+     * them followed by a name is a separate mistake, reported later as
+     * trailing garbage, while two of them followed by nothing leaves the
+     * name empty and has to be refused right here.
+     * @param tail Tail substring
+     * @param from Index to look from
+     * @return True if only spaces remain
+     */
+    private static boolean blank(final String tail, final int from) {
+        int idx = from;
+        while (idx < tail.length() && tail.charAt(idx) == ' ') {
```
Thayorns: `blank` counts only `' '`, but a tab ends a name just as well: `terminates` at line 595 lists `'\t'`, so `skipName` returns straight away and `name` comes out empty, and `endsClean` at line 500 walks over tabs and finds nothing to complain about. `42 >` followed by a tab therefore still yields `Form.NAME` with an empty label, which `phi` at line 270 passes through as the `@name` — the very thing this check exists to refuse. Nothing upstream stops the tab either: Eo.java:308 raises "tab character in leading whitespace" only for the indent region, so the character reaches the tail intact.

morphqdd: Confirmed and fixed. `42 >` followed by a tab did come out as `name=""` with `loc="Φ.app."` and no error, and so did `? >` with a tab. `blank` now looks past tabs as well, and there is a second typo pack for that input.

`42 >` with two tabs then a name still reports trailing garbage, same as the two-spaces case, so that path is unchanged.

eo-parser/src/main/java/org/eolang/parser/Suffix.java
```diff
@@ -442,13 +442,13 @@ private static Parsed auto(
     private static Parsed named(
         final String tail, final int from, final Span span, final int home
     ) {
-        final int begin = Suffix.skipSpace(tail, from);
-        if (begin >= tail.length()) {
+        if (Suffix.blank(tail, from)) {
```
Thayorns, line 445: `blank` looks past spaces and tabs, so a terminator that is neither still leaves the name empty. On `42 >!` the guard passes, `skipName` at line 600 returns `begin` immediately because `endsName` counts `!` as a terminator via line 618, `checkLowercaseStart` waves the empty string through at line 379, the `!` is then eaten as the const marker at line 462, and `endsClean` finds nothing left to complain about — so `Form.NAME` comes back with an empty label, the very thing this check exists to refuse. `42 >/foo` walks the same path and yields a nameless atom with signature `foo`. Worth noting that `+>!` is refused at line 350, because `test` compares the index before and after `skipName` instead of inspecting what follows.

Thayorns (commented):
This replaces the `begin >= tail.length()` emptiness check in `named` with a `blank` scan that looks past every space and tab, and pins both cases with typo packs that run through `checksTypoPacks` and `checksTypoMessages`. The tab case raised in the earlier round is genuinely closed now, and the two-spaces-then-a-name path still reports trailing garbage as before. What remains is that `blank` decides emptiness from the characters it skipped rather than from whether a name was actually consumed.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2888](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+413](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6157, opened 2026-07-30
#6148: copy the document before the shift writes into it

### LINKED ISSUE #6148: `StXnav.apply` incorrectly mutates input DOM instead of working on a copy
## What happens

`StXnav.apply` mutates the DOM it is given, instead of working on a copy:

```java
public XML apply(final int position, final XML xml) {
    final Node dom = xml.inner();
    new Xnav(dom).path(this.xpath).forEach(this.fun);
    return new XMLDocument(dom);
}
```

`xml.inner()` hands back the caller's own node, `forEach(this.fun)` writes into it, and the returned `XMLDocument` wraps that same node. So the caller's `XML` changes under it.

`Xmir.toEO()` runs shifts built on `StXnav` (through `StUnhex`), which makes a method that looks like it only reads rewrite its input. `Xmir` declares `implements XML`, so it presents itself as a decorator of the XML it was given.

Reproduced with:

```java
final XML xml = new EoSyntax(new InputOf("# Sample.\n\n[] > app\n  \"hi\" > @\n")).parsed();
final String before = xml.toString();
new Xmir(xml).toEO();
// xml.toString() now differs from before
```

Before:

```xml
<o base="Φ.string" line="4" loc="Φ.app.φ" name="φ" pos="2">
   <o as="α0" base="Φ.bytes" line="4" loc="Φ.app.φ.α0" pos="2">
      <o as="α0" line="4" loc="Φ.app.φ.α0.α0" pos="2">68-69</o>
   </o>
</o>
```

After:

```xml
<o base="Φ.string" line="4" loc="Φ.app.φ" name="φ" pos="2">"hi"</o>
```

The bytes payload is gone from the caller's document.

Two things worth stating plainly, so the severity is not overstated. The result is still schema valid, checked with `StrictXML` against `eo-parser/src/main/resources/XMIR.xsd` both before and after. And every caller of `toEO()` in this repo today (`MjPrint:125`, `MjFormat:159` and `:165`, `StEoLogged:65`) parses fresh and drops the XML right after, so nothing is broken right now. This is about the contract, not a current failure.

## What should happen

`StXnav.apply` works on a copy, so the input `XML` is left as it was and only the returned one carries the change. A shift that rewrites its input makes any caller that keeps using the XML read something it never produced, and there is nothing in the signature to warn about it.

morphqdd:
Closes #6148

`apply` took `xml.inner()` and let the function write straight into it, so the caller's document changed under it. It now clones first, and only the returned document carries the change.

Two tests: one that the given `XML` is byte-for-byte the same afterwards, which fails on master, and one that the returned one does carry what the function did, so the copy is not just thrown away.

Checked the users of this shift as well: `StHex` in the parser and `StUnhex` in the printer both still pass, 1616 tests in eo-parser and 319 in eo-printer.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/StXnav.java
+++ b/eo-parser/src/main/java/org/eolang/parser/StXnav.java
@@ -61,7 +61,7 @@ public final class StXnav implements Shift {
 
     @Override
     public XML apply(final int position, final XML xml) {
-        final Node dom = xml.inner();
+        final Node dom = xml.inner().cloneNode(true);
         new Xnav(dom).path(this.xpath).forEach(this.fun);
         return new XMLDocument(dom);
     }
--- a/eo-parser/src/test/java/org/eolang/parser/StXnavTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/StXnavTest.java
@@ -1,0 +1,43 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.parser;
+
+import com.jcabi.matchers.XhtmlMatchers;
+import com.jcabi.xml.XML;
+import com.jcabi.xml.XMLDocument;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link StXnav}.
+ * @since 0.53.0
+ */
+final class StXnavTest {
+
+    @Test
+    void leavesTheGivenXmlAlone() {
+        final XML before = new XMLDocument("<p><o base='a'>text</o></p>");
+        final String snapshot = before.toString();
+        new StXnav("/p/o", node -> node.node().setTextContent("changed")).apply(0, before);
+        MatcherAssert.assertThat(
+            "the shift must not touch the document it was given, only the one it returns",
+            before.toString(),
+            Matchers.equalTo(snapshot)
+        );
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


Thayorns (approved):
This makes StXnav.apply clone the incoming DOM before Xnav writes into it, so the shift no longer rewrites the document its caller still holds. The fix is the right one at the right level: inner() hands back the live node in jcabi-xml 0.37.1, and cloneNode(true) on a Document is what jcabi's own deepCopy does for that case, so the returned XMLDocument is the only one carrying the change. Both callers, StHex and StUnhex, consume the returned document rather than relying on in-place mutation, and Xmir's train runs StUnhex ahead of XSL shifts that build fresh documents, so the path the issue reproduced on is closed. The two tests split the behavior cleanly, one pinning that the input survives and one pinning that the copy is not discarded, and the first genuinely fails on master since XMLDocument.toString re-renders the node on every call. Nothing here needs changing. Worth knowing that the coverage sits at the shift rather than at Xmir.toEO where the issue reproduced, which is a fair place to leave it given the cause is in the shift.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+373](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2904](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6158, opened 2026-07-30
#6147: refuse a meta directive that has no name

### LINKED ISSUE #6147: Meta directive accepts empty name, causing silent package drop
## What happens

A meta line with no name parses without any error. `LnMeta` takes the name as `body.substring(1)` when there is no space, and `body.substring(1, space)` when there is one, and neither branch rejects an empty result.

A bare `+` line:

```
# Sample.

+
+package foo

[] > app
```

parses with zero errors and emits:

```xml
<meta line="3">
   <head/>
   <tail/>
</meta>
```

The practical case is a stray space after the plus. These two files differ by one character:

```
# Sample.

+package foo

[] > app
```

```
# Sample.

+ package foo

[] > app
```

The first gives `<o line="5" loc="Φ.foo.app" name="app" pos="0"/>`. The second gives `<o line="5" loc="Φ.app" name="app" pos="0"/>` and reports no error at all. The `+package` meta is gone, the object lands in the default package, and nothing says so. It surfaces later as a missing or wrongly named class.

This is inconsistent with the neighbouring check: `LnMeta.split` does reject a double space between parts with "meta parts must be separated by exactly one space", so `+package  foo` is an error while `+ package foo` is not.

The spec gives the grammar as `+name [parts]` (§3.2, "Meta directive — `+name [parts]`"), and R-3.2.3 talks about "each space-separated token after the `+name`", so the name is required.

## What should happen

A meta with an empty name is rejected with a parse error, the same way other malformed metas are, for example "meta directive must have a name". Both a bare `+` and `+ name ...` with a space right after the plus should hit it.

Not a duplicate of #3556, which was about double spaces between parts and is fixed.

morphqdd:
Closes #6147

`head` came from `body.substring(1)` or `body.substring(1, space)`, and neither branch minded an empty result. A bare `+`, and any `+ name` with a space right after the plus, parsed with no error and emitted `<head/>`.

The practical case is `+ package foo`, which silently dropped the package and put the object in the default one. It now fails to parse instead.

The test is a typo pack, so it runs through both `checksTypoPacks` and `checksTypoMessages`, and it fails on master.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnMeta.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnMeta.java
@@ -69,6 +69,12 @@ final class LnMeta implements Line {
                 body.substring(space + 1), this.span, space + 1
             );
         }
+        if (head.isEmpty()) {
+            throw new ParseError(
+                this.span.line(), this.span.indent(),
+                "meta directive requires a name"
+            );
+        }
         Comments.seal(globals, emit, this.span);
         globals.markMeta();
         globals.clearBlanks();
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-typos/meta-without-a-name.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-typos/meta-without-a-name.yaml
@@ -1,0 +1,10 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+line: 1
+message: "meta directive requires a name"
+input: |
+  + package foo
+
+  [] > app


Thayorns (approved):
This adds a guard in LnMeta.into that rejects a meta whose name is empty, and a typo pack for `+ package foo`. The guard is placed after the parts have been split and before anything is emitted, so nothing is written for a rejected line, and a sweep of the eo-packs and eo-typos fixtures turns up no existing input with a nameless meta that would newly fail. One gap worth noting: the pack exercises only the `space >= 0` branch, so the bare `+` line from the issue — the `space < 0` case where head comes from body.substring(1) — is not covered anywhere, even though LnMetaTest already has a test for each of the other three guards in this same method.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (16) hits-of-code; +4 to give you at least something.  Your running score is [+381](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2844](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6160, opened 2026-07-30
Stop resolving tuple.with through the package

### LINKED ISSUE #6122: Duplicate `Φ.tuple.with` locator in tuple.eo and tuple/with.eo causing identifier collision
The name `with` under `Φ.tuple` is defined twice. `tuple.eo:52` declares it as an own attribute of `tuple`, and `tuple/with.eo:10` declares a package member of the same package:

```
tuple > [x] > with     # tuple.eo:52
  ^
  x
  as-number.
    length.plus 1

[list x] > with        # tuple/with.eo:10
  list.with x > @
```

Both land on the same locator. `set-locators.xsl` stamps `loc="Φ.tuple.with"` on each of them, so 3-lint XMIR carries that string twice, once from `tuple.xmir` and once from `tuple/with.xmir`. Sweeping every `@loc` in eo-runtime's 3-lint output, this is the only collision in the language — the three further duplicates that turn up (`Φ.tuple.with.x`, `.φ`, `.φ.α0`) are just its own children inheriting the clash. A locator is meant to identify one object, and here it identifies two.

Which of the two a caller reaches depends on how the caller got there, because the two dispatch paths probe in opposite orders. `PhNest.take` at `PhNest.java:63` asks `owns(name)` first and only falls back to the object behind the package, so `tuple.with someTuple 3` resolves to the package member and never sees the own attribute. `PhDefault.take` at `PhDefault.java:225` checks the object's own attributes first and reaches a package extension only through `absent` at `PhDefault.java:392`, so `(* 1 2).with 3` resolves to the own attribute and never sees the package member. Neither definition shadows the other outright; each is unreachable from one side.

The package member earns nothing by existing. Its body, `list.with x`, dispatches `with` on a tuple value, which by the rule above is the own attribute in `tuple.eo` — so it shuffles the receiver from the first argument into `ρ` and forwards. Its one test, `tests-list-simple-with` in `tuple/with.eo:13`, appends `3` to `* 1 2`, which is what `tests-tuple-with` in `tuple.eo:256` already covers.

Deleting `tuple/with.eo` removes the duplicate identity and loses nothing that is tested. One wrinkle is worth checking first: `PhNest` deliberately leaves `ρ` unbound when handing out a package extension, so after the deletion a namespace-style `tuple.with a b` would resolve to the own attribute, which reads its receiver from `^` rather than a first argument. If that form has to keep working, the implementation should move into `tuple/with.eo` and the own attribute in `tuple.eo` should go instead.

maxonfjvipon:
Closes #6122.

The name `with` lived twice under `Φ.tuple`: once as an own attribute in `tuple.eo`, once as a package member in `tuple/with.eo`. One locator, two objects — and because `PhNest.take` probes the package first while `PhDefault.take` probes own attributes first, each definition was unreachable from one of the two dispatch paths.

The package member earned nothing. Its body moved the first argument into `ρ` and forwarded to the own attribute, and its only test repeated `can-add-an-item` from `tuple.eo`. It is gone, and the own attribute stays. Two callers inside the package used the package form and now dispatch on their receiver:

```
[list index item] > withi
  concat > @
    with.
      front list index
      item
```

Sweeping every `@loc` in eo-runtime's 3-lint XMIR after the change turns up no duplicates at all, so the four collisions the issue reported are the ones removed and nothing took their place. Full `mvn test -pl eo-runtime -Pqulice` run: 1814 tests, no failures.

One consequence worth naming: the namespace form `tuple.with a b` no longer resolves, since the surviving definition reads its receiver from `^`. Nothing in the repository used it. Sibling package members like `tuple.contains` keep working, and `can-check-containment-through-the-package` still covers that shape.

A lint that rejects two objects sharing a locator would keep this from coming back — worth a separate ticket rather than folding into this one.


Thayorns (approved):
This deletes the `tuple.with` package member and rewrites its two callers to dispatch `with` on the receiver instead, leaving the own attribute in `tuple.eo` as the single definition of that locator. The rewrites in `concat.eo` and `withi.eo` are behavior-preserving and land on the same receiver-first shape that `back.eo`, `slice.eo`, `without.eo` and `withouti.eo` already use, and the test the deleted file carried is already covered by `can-add-an-item` in `tuple.eo`. Nothing else in the runtime referenced the package form, so no caller is left dangling.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (26) hits-of-code; +4 to give you at least something.  Your running score is [+385](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+768](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6164, opened 2026-07-30
#6141: classify drive-relative Windows paths as non-absolute

### LINKED ISSUE #6141: `path.win32.is-absolute` mistakenly categorizes drive-relative paths as absolute in Windows
## What happens

`path.win32` classifies a drive-relative Windows path as absolute merely because it begins with a drive designator.

| expression | actual `is-absolute` | expected `is-absolute` |
| --- | --- | --- |
| `path.win32 "C:foo"` | `true` | `false` |
| `path.win32 "C:\\foo"` | `true` | `true` |

Both rows were dataized through the EO runtime. The second row is the control: a drive-qualified path with a root separator remains absolute. The first lacks that separator, so on Windows it is resolved relative to the current directory of drive `C:`, not relative to the filesystem root.

## Why this is a bug

[`path.impl.is-absolute`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/path.eo#L60-L90) is defined as `is-root-relative.or (drive.size.gt 0)`. `win32.drive` returns `"C:"` for both `C:\\foo` and `C:foo`, so the second condition incorrectly makes every drive-qualified path absolute.

The neighboring [`is-root-relative`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/path.eo#L87-L90) already contains the required distinction: it checks whether the part after the drive begins with the Windows separator. `is-absolute` should require that rooted form when a drive is present.

## Steps to reproduce

1. Create `path.win32 "C:foo"`.
2. Dataize its `is-absolute` attribute.
3. Observe `true`.

## Expected behavior

`C:foo`, `C:.\\foo`, and `C:..\\foo` should be reported as non-absolute drive-relative paths. `C:\\foo` should remain absolute. One possible fix is to make `is-absolute` depend on `is-root-relative` rather than on a non-empty drive prefix alone.

gemshrine:
## What this PR changes

Fixes #6141.

`path.win32.is-absolute` now requires a root separator after an optional drive prefix. Consequently, `C:foo`, `C:.\\foo`, and `C:..\\foo` are reported as relative paths, while `C:\\foo` and `\\foo` remain absolute.

## Problem

The previous [implementation](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/path.eo#L85-L86) treated any non-empty drive prefix as sufficient for absoluteness. It therefore conflated a drive-relative path (`C:foo`) with a drive-rooted path (`C:\\foo`).

## Solution

The [updated implementation](https://github.com/gemshrine/eo/blob/6141-win32-drive-relative/eo-runtime/src/main/eo/path.eo#L85-L91) verifies that the character following the drive prefix is the Windows separator. This preserves the existing behavior for rooted paths without a drive and for POSIX paths.

## Regression coverage

The existing Win32 absolute-path scenario now also verifies that `C:foo`, `C:.\\foo`, and `C:..\\foo` are not absolute, alongside the established absolute controls.

## Verification

Ran the targeted `EOpathTest#can_detect_an_absolute_win32_path` test with Maven, JUnit, and the JVM limited to one execution thread/processor. The EO formatter and linter completed successfully in that run.

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -195,9 +195,17 @@
         string ubts > pth
       ^.separator > separator
 
-  not. ++> can-detect-a-drive-relative-win32-path-as-not-absolute
-    is-absolute.
-      path.win32 "C:foo"
+  and. ++> can-classify-a-drive-relative-win32-path-as-non-absolute
+    and.
+      not.
+        is-absolute.
+          path.win32 "C:foo"
+      not.
+        is-absolute.
+          path.win32 "C:.\\foo"
+    not.
+      is-absolute.
+        path.win32 "C:..\\foo"
 
   and. ++> can-detect-an-absolute-posix-path
     is-absolute.


eo-runtime/src/main/eo/path.eo
```diff
@@ -82,8 +82,11 @@
                 number index
       string > text
         basename >> base!
-    is-root-relative.or > is-absolute
-      drive.size.gt 0
+    and. > is-absolute
```
Thayorns, line 81: driveless at line 61 is just uri from drive.size onward, so driveless.size.gt 0 is uri.size.gt drive.size and driveless.slice 0 1 is uri.slice drive.size 1. That makes this block the same predicate as is-root-relative at line 90, written a second time against different attributes, and whoever touches one of them later has nothing telling them the other has to move too.

Thayorns (commented):
The change makes is-absolute require a separator right after the drive prefix, so C:foo and its dot forms stop being absolute while C:\foo and \foo stay absolute. The logic holds up and the win32 test covers the three drive-relative forms from the issue. One gap: nothing dataizes (path.win32 "C:").is-absolute, and that is the only input that takes the new uri.size.gt drive.size conjunct to false, so the guard that keeps bytes.slice in range is the one part of this change with no case behind it.

yegor256:
@gemshrine what about a test?

gemshrine:
@yegor256 added a dedicated test, can-classify-a-drive-relative-win32-path-as-non-absolute, asserting that C:foo, C:.\foo and C:..\foo are all classified as non-absolute. The is-absolute logic now requires a separator right after the drive, so a drive-relative path without one is no longer treated as absolute.

yegor256:
@gemshrine conflicts

gemshrine:
@yegor256 rebased onto master and resolved the conflict. Master already carries the corrected is-absolute logic, so the only thing left here is the test. I merged it with the existing single case one into a broader check, can-classify-a-drive-relative-win32-path-as-non-absolute, which now asserts C:foo, C:.\foo and C:..\foo are all non-absolute. Green locally, 51 path tests pass.

morphqdd:
Issue #6141 is already closed by #6296, merged 2026-08-03, which fixed `is-absolute` in `path.eo` and added a regression test for the same case. The `path.eo` diff in this PR is byte-identical to what's already on master. The remaining diff here is just renaming and extending an already-merged test with two more literal cases (`C:.\foo`, `C:..\foo`) that exercise the same code path as the existing `C:foo` case, no new logic covered. Worth closing this one or trimming it to just those two extra cases folded into master's existing test.

eo-runtime/src/main/eo/path.eo
```diff
@@ -195,9 +195,17 @@
         string ubts > pth
       ^.separator > separator
 
-  not. ++> can-detect-a-drive-relative-win32-path-as-not-absolute
-    is-absolute.
-      path.win32 "C:foo"
+  and. ++> can-classify-a-drive-relative-win32-path-as-non-absolute
```
morphqdd, line 198: Issue #6141 is already closed - fixed and merged via #6296 on 2026-08-03. That PR renamed `is-root-relative` to `is-absolute` directly (dropping the old `.or (drive.size.gt 0)` shortcut this PR's description talks about) and already added this exact test (`not. ++> can-detect-a-drive-relative-win32-path-as-not-absolute`, checking `"C:foo"`) against the fixed implementation. This branch predates that merge, so its diff here only touches the test, not `is-absolute` itself - the fix it's meant to carry is already on master through a different PR. The two extra cases this adds (`C:.\\foo`, `C:..\\foo`) look like real additional coverage worth keeping, but this needs a rebase onto current master and a rewrite of the PR description before it can land - as written it's proposing to re-fix something already fixed. Worth also checking #6198 ("extend Windows drive-relative path coverage"), which looks like it's chasing the same already-closed issue from a different angle.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2091](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (14) hits-of-code; +1 to give you at least something.  Your running score is [+1087](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (14) hits-of-code; +1 to give you at least something.  Your running score is [+3661](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6166, opened 2026-07-30
#6161: refuse a value type given to the %s conversion

### LINKED ISSUE #6161: PrintfArgs.java wrongly identifies byte types
The puzzle `5943-006d3f8b` from #5943 has to be resolved: 

https://github.com/objectionary/eo/blob/2dd7d7b2a256b637aa849b0dfb654c8ccb6e17fb/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java#L204-L211

morphqdd:
Closes #6161

Resolves the puzzle from #5943. The UTF-8 check there only rejected bytes that are not valid text at all, which is a matter of luck rather than of type: `printf "%s" 42` printed `@E` followed by the six NUL bytes of the IEEE-754 payload, and `printf "%s" TRUE` printed a control character, both without a word.

The puzzle said the type was already lost by the time the conversion runs, and it was, but not irretrievably. `tuple.at` decorates what it returns, so the retrieved object reports `Φ.tuple.at` as its forma and the argument's own type is invisible behind it. Walking the cons list reaches the argument itself with its forma intact, so `printf` now takes the tuple and walks it instead of dispatching through `at`. Bounds are already checked before retrieval, so nothing is lost by not going through `at`.

The check names what text is **not** rather than what it is. That is the part worth reviewing. Only a literal has the forma `Φ.string`: `"a".concat "b"` normalizes to `Φ.bytes` and `slice` to `Φ.string.slice`, and any user object decorating a string has its own forma. Demanding `Φ.string` would refuse all of those, which is a much bigger problem than the one being fixed, so the guard refuses `Φ.number` and `Φ.bool` and lets everything else through.

Tests: three at the EO level, including one that a computed string is still accepted, plus the Java ones. The bool case fails on master. The number case already errored there for an unrelated downstream reason, so the Java test is what pins it.

### DIFF
--- a/eo-runtime/src/main/eo/string/printf.eo
+++ b/eo-runtime/src/main/eo/string/printf.eo
@@ -34,6 +34,12 @@
       "%25x"
       42.as-bytes
 
+  eq. ++> can-format-a-computed-string
+    "abcd"
+    printf *1
+      "%s"
+      "ab".concat "cd"
+
   eq. ++> can-format-a-float-with-six-fraction-digits
     "1.250000"
     printf *1
@@ -177,6 +183,10 @@
         42.as-i64
       "42"
 
+  printf *1 --> stops-on-a-bool-as-a-string
+    "%s"
+    true
+
   printf *1 --> stops-on-a-comma-flag-with-s
     "%,s"
     "Jeff"
@@ -196,6 +206,10 @@
   printf *1 --> stops-on-a-non-utf8-byte-argument-as-a-string
     "%s"
     3.14
+
+  printf *1 --> stops-on-a-number-as-a-string
+    "%s"
+    42
 
   printf *1 --> stops-on-a-paren-flag-with-s
     "%(s"
--- a/eo-runtime/src/main/eo/string/printf.eo
+++ b/eo-runtime/src/main/eo/string/printf.eo
@@ -227,6 +227,10 @@
     "%0$s"
     "first"
 
+  printf *1 --> stops-on-an-i64-as-a-string
+    "%s"
+    42.as-i64
+
   printf *1 --> stops-on-an-oversized-positional-index
     "%99999999999999999999$s"
     "Jeff"
--- a/eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
@@ -55,8 +55,8 @@ public final class EOprintf extends PhDefault implements Atom {
                             .otherwise("be a tuple with the 'length' attribute")
                             .it(),
                         Expect.at(this, "args")
-                            .that(phi -> phi.take("at"))
-                            .otherwise("be a tuple with the 'at' attribute")
+                            .that(phi -> phi)
+                            .otherwise("be a tuple")
                             .it()
                     ).formatted().toArray()
                 )
--- a/eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
@@ -51,12 +51,8 @@ public final class EOprintf extends PhDefault implements Atom {
                     new PrintfArgs(
                         format,
                         Expect.at(this, "args")
-                            .that(phi -> new Dataized(phi.take("length")).asNumber().intValue())
+                            .must(phi -> new Dataized(phi.take("length")).asNumber() >= 0.0)
                             .otherwise("be a tuple with the 'length' attribute")
-                            .it(),
-                        Expect.at(this, "args")
-                            .that(phi -> phi)
-                            .otherwise("be a tuple")
                             .it()
                     ).formatted().toArray()
                 )
--- a/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
@@ -68,20 +68,21 @@ final class PrintfArgs {
     private static final double LONG_UPPER_LIMIT = 0x1.0p63;
 
     /**
-     * Formas whose bytes are a value, never text, and so can never be the
-     * argument of a {@code %s} conversion.
+     * Objects a forma can be rooted at whose bytes are a value, never
+     * text, and so can never be the argument of a {@code %s} conversion.
      *
-     * <p>Naming what text is <em>not</em>, rather than what it is, is
-     * deliberate. Only a literal has the forma {@code Φ.string}: every
-     * operation that builds a string returns something else
-     * ({@code "a".concat "b"} normalizes to {@code Φ.bytes}, {@code slice}
-     * to {@code Φ.string.slice}), and so does any user object that
-     * decorates a string. Demanding {@code Φ.string} would refuse all of
-     * those, which is a far bigger problem than the one this guard is
-     * here for.</p>
+     * <p>Naming what text is <em>not</em> is deliberate: only a literal
+     * has the forma {@code Φ.string}, while {@code "a".concat "b"}
+     * normalizes to {@code Φ.bytes} and {@code slice} to
+     * {@code Φ.string.slice}, so demanding {@code Φ.string} would refuse
+     * every computed string. The root is matched rather than the whole
+     * name because a conversion keeps its origin at the front
+     * ({@code 42.as-i64} is {@code Φ.number.as-i64},
+     * {@code 42.as-i64.as-number} is {@code Φ.i64.as-number}), and
+     * listing whole names would miss the next pair added.</p>
      */
-    private static final Set<String> NOT_TEXT = new HashSet<>(
-        Arrays.asList("Φ.number", "Φ.bool")
+    private static final Set<String> VALUES = new HashSet<>(
+        Arrays.asList("number", "bool", "i8", "i16", "i32", "i64")
     );
 
     static {
@@ -98,11 +99,6 @@ final class PrintfArgs {
     private final String format;
 
     /**
-     * The length.
-     */
-    private final long length;
-
-    /**
      * The tuple of arguments.
      */
     private final Phi args;
@@ -110,12 +106,10 @@ final class PrintfArgs {
     /**
      * Ctor.
      * @param fmt The format
-     * @param len The length
      * @param tuple The tuple of arguments
      */
-    PrintfArgs(final String fmt, final long len, final Phi tuple) {
+    PrintfArgs(final String fmt, final Phi tuple) {
         this.format = fmt;
-        this.length = len;
         this.args = tuple;
     }
 
@@ -152,6 +146,7 @@ final class PrintfArgs {
     List<Object> formatted() {
         final List<Object> arguments = new ArrayList<>(0);
         final Matcher matcher = PrintfArgs.SPECIFIER.matcher(this.format);
+        final long count = new Dataized(this.args.take("length")).asNumber().longValue();
         long auto = 0L;
         while (matcher.find()) {
             final String positional = matcher.group(1);
@@ -168,7 +163,7 @@ final class PrintfArgs {
                     throw new ExFailure(
                         String.format(
                             "The argument index %s is out of bounds (total arguments: %d)",
-                            digits, this.length
+                            digits, count
                         ),
                         ex
                     );
@@ -183,13 +178,13 @@ final class PrintfArgs {
                 arg = auto;
                 auto += 1L;
             }
-            if (arg >= this.length) {
+            if (arg >= count) {
                 throw new ExFailure(
                     "The argument index %d is out of bounds (total arguments: %d)",
-                    arg, this.length
+                    arg, count
                 );
             }
-            arguments.add(PrintfArgs.fmt(symbol, this.element(arg)));
+            arguments.add(PrintfArgs.fmt(symbol, this.element(count, arg)));
         }
         return arguments;
     }
@@ -201,19 +196,19 @@ final class PrintfArgs {
      * <p>The difference is the whole point: {@code at} decorates what it
      * returns, so the result reports {@code Φ.tuple.at} as its forma and
      * the argument's own type is no longer visible. Walking the cons list
-     * reaches the object itself, forma intact. The list is built with the
-     * last argument at the head (see {@code tuple.eo}), so the walk takes
-     * as many tails as there are arguments after this one, counted off the
-     * tuple's own length so that a wrong count can only pick the wrong
-     * argument, never walk off the end into a terminated computation.</p>
+     * reaches the object itself, forma intact. The last argument sits at
+     * the head (see {@code tuple.eo}), so the walk takes as many tails as
+     * there are arguments after this one. The count comes from the caller,
+     * which read it off this same tuple and has already refused an index
+     * outside it, so the walk cannot run off the end.</p>
      *
+     * @param count How many arguments the tuple holds
      * @param index Zero-based index of the argument
      * @return The argument
      */
-    private Phi element(final long index) {
+    private Phi element(final long count, final long index) {
         Phi current = this.args;
-        final long size = new Dataized(this.args.take("length")).asNumber().longValue();
-        for (long step = size - 1L - index; step > 0L; --step) {
+        for (long step = count - 1L - index; step > 0L; --step) {
             current = current.take("tail");
         }
         return current.take("head");
@@ -241,17 +236,17 @@ final class PrintfArgs {
     /**
      * Refuse an argument whose type says its bytes are a value and not text.
      *
-     * <p>Without this, a {@code number} handed to {@code %s} is decoded as
-     * if its eight raw IEEE-754 bytes were UTF-8, and {@code true} prints
-     * as a control character, both with no complaint. Only bytes that are
-     * not valid UTF-8 at all were caught before, which is a matter of luck
-     * rather than of type.</p>
+     * <p>Without this, a {@code number} handed to {@code %s} is decoded
+     * as if its eight raw IEEE-754 bytes were UTF-8, with no complaint.
+     * Only bytes that are not valid UTF-8 at all were caught before,
+     * which is a matter of luck rather than of type.</p>
      *
      * @param element The argument
      */
     private static void textual(final Phi element) {
         final String forma = element.normalized().forma();
-        if (PrintfArgs.NOT_TEXT.contains(forma)) {
+        final String[] parts = forma.split("\\.");
+        if (parts.length > 1 && PrintfArgs.VALUES.contains(parts[1])) {
             throw new ExFailure(
                 "The argument of the '%%s' conversion is %s, whose bytes are a value and not text; use %s instead",
                 forma, "'%d', '%f', '%b' or '%x'"
--- a/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
+++ b/eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
@@ -13,14 +13,16 @@ import java.nio.charset.CharacterCodingException;
 import java.nio.charset.CodingErrorAction;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.StringJoiner;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import org.eolang.Data;
 import org.eolang.Dataized;
 import org.eolang.ExFailure;
 import org.eolang.Phi;
@@ -65,6 +67,23 @@ final class PrintfArgs {
      */
     private static final double LONG_UPPER_LIMIT = 0x1.0p63;
 
+    /**
+     * Formas whose bytes are a value, never text, and so can never be the
+     * argument of a {@code %s} conversion.
+     *
+     * <p>Naming what text is <em>not</em>, rather than what it is, is
+     * deliberate. Only a literal has the forma {@code Φ.string}: every
+     * operation that builds a string returns something else
+     * ({@code "a".concat "b"} normalizes to {@code Φ.bytes}, {@code slice}
+     * to {@code Φ.string.slice}), and so does any user object that
+     * decorates a string. Demanding {@code Φ.string} would refuse all of
+     * those, which is a far bigger problem than the one this guard is
+     * here for.</p>
+     */
+    private static final Set<String> NOT_TEXT = new HashSet<>(
+        Arrays.asList("Φ.number", "Φ.bool")
+    );
+
     static {
         PrintfArgs.CONVERSION.put('s', PrintfArgs::validString);
         PrintfArgs.CONVERSION.put('d', element -> PrintfArgs.toLong(element.asNumber()));
@@ -84,20 +103,20 @@ final class PrintfArgs {
     private final long length;
 
     /**
-     * Phi attribute.
+     * The tuple of arguments.
      */
-    private final Phi retriever;
+    private final Phi args;
 
     /**
      * Ctor.
      * @param fmt The format
      * @param len The length
-     * @param phi Phi attribute
+     * @param tuple The tuple of arguments
      */
-    PrintfArgs(final String fmt, final long len, final Phi phi) {
+    PrintfArgs(final String fmt, final long len, final Phi tuple) {
         this.format = fmt;
         this.length = len;
-        this.retriever = phi;
+        this.args = tuple;
     }
 
     /**
@@ -170,11 +189,34 @@ final class PrintfArgs {
                     arg, this.length
                 );
             }
-            final Phi taken = this.retriever.copy();
-            taken.put(0, new Data.ToPhi(arg));
-            arguments.add(PrintfArgs.fmt(symbol, new Dataized(taken)));
+            arguments.add(PrintfArgs.fmt(symbol, this.element(arg)));
         }
         return arguments;
+    }
+
+    /**
+     * The argument at the given index, as the object the caller passed in
+     * rather than the one {@code tuple.at} hands back.
+     *
+     * <p>The difference is the whole point: {@code at} decorates what it
+     * returns, so the result reports {@code Φ.tuple.at} as its forma and
+     * the argument's own type is no longer visible. Walking the cons list
+     * reaches the object itself, forma intact. The list is built with the
+     * last argument at the head (see {@code tuple.eo}), so the walk takes
+     * as many tails as there are arguments after this one, counted off the
+     * tuple's own length so that a wrong count can only pick the wrong
+     * argument, never walk off the end into a terminated computation.</p>
+     *
+     * @param index Zero-based index of the argument
+     * @return The argument
+     */
+    private Phi element(final long index) {
+        Phi current = this.args;
+        final long size = new Dataized(this.args.take("length")).asNumber().longValue();
+        for (long step = size - 1L - index; step > 0L; --step) {
+            current = current.take("tail");
+        }
+        return current.take("head");
     }
 
     /**
@@ -183,14 +225,38 @@ final class PrintfArgs {
      * @param element Element ready for formatting
      * @return Formatted object
      */
-    private static Object fmt(final char symbol, final Dataized element) {
+    private static Object fmt(final char symbol, final Phi element) {
         if (!PrintfArgs.CONVERSION.containsKey(symbol)) {
             throw new ExFailure(
                 "The format %c is unsupported, only %s formats can be used",
                 symbol, "%s, %d, %f, %x, %b"
             );
         }
-        return PrintfArgs.CONVERSION.get(symbol).apply(element);
+        if (symbol == 's') {
+            PrintfArgs.textual(element);
+        }
+        return PrintfArgs.CONVERSION.get(symbol).apply(new Dataized(element));
+    }
+
+    /**
+     * Refuse an argument whose type says its bytes are a value and not text.
+     *
+     * <p>Without this, a {@code number} handed to {@code %s} is decoded as
+     * if its eight raw IEEE-754 bytes were UTF-8, and {@code true} prints
+     * as a control character, both with no complaint. Only bytes that are
+     * not valid UTF-8 at all were caught before, which is a matter of luck
+     * rather than of type.</p>
+     *
+     * @param element The argument
+     */
+    private static void textual(final Phi element) {
+        final String forma = element.normalized().forma();
+        if (PrintfArgs.NOT_TEXT.contains(forma)) {
+            throw new ExFailure(
+                "The argument of the '%%s' conversion is %s, whose bytes are a value and not text; use %s instead",
+                forma, "'%d', '%f', '%b' or '%x'"
+            );
+        }
     }
 
     /**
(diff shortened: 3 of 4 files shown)


eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -65,6 +67,23 @@ final class PrintfArgs {
...
+     * deliberate. Only a literal has the forma {@code Φ.string}: every
+     * operation that builds a string returns something else
+     * ({@code "a".concat "b"} normalizes to {@code Φ.bytes}, {@code slice}
+     * to {@code Φ.string.slice}), and so does any user object that
+     * decorates a string. Demanding {@code Φ.string} would refuse all of
+     * those, which is a far bigger problem than the one this guard is
+     * here for.</p>
+     */
+    private static final Set<String> NOT_TEXT = new HashSet<>(
+        Arrays.asList("Φ.number", "Φ.bool")
```
Thayorns: `42.as-i64` still gets through here. `as-i64` is a plain EO object (eo-runtime/src/main/eo/number/as-i64.eo:14), so what comes back reports `Φ.number.as-i64` and not `Φ.number`, and its bytes `00-00-00-00-00-00-00-2A` are perfectly valid UTF-8, so `validString` lets them past as well. `printf "%s" 42.as-i64` prints seven NULs and a `*` with no complaint, which is the same mojibake the puzzle you removed was about, and the i8/i16/i32/i64 family is a value type by exactly the same argument as `number` and `bool`.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -170,27 +189,74 @@ List<Object> formatted() {
...
+     * as many tails as there are arguments after this one, counted off the
+     * tuple's own length so that a wrong count can only pick the wrong
+     * argument, never walk off the end into a terminated computation.</p>
+     *
+     * @param index Zero-based index of the argument
+     * @return The argument
+     */
+    private Phi element(final long index) {
+        Phi current = this.args;
+        final long size = new Dataized(this.args.take("length")).asNumber().longValue();
```
Thayorns: The object already holds `length` from the constructor and `formatted` bounds-checks the index against it, yet `element` dataizes the tuple's own `length` again for every argument. That is two sources for the same number: should they ever disagree, an index that clears the check makes `size - 1 - index` negative, the loop never runs, and `take("head")` hands back the last argument instead of failing.

eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
```diff
@@ -55,8 +55,8 @@ public Phi lambda() {
                             .otherwise("be a tuple with the 'length' attribute")
                             .it(),
                         Expect.at(this, "args")
-                            .that(phi -> phi.take("at"))
-                            .otherwise("be a tuple with the 'at' attribute")
+                            .that(phi -> phi)
```
Thayorns: `that(phi -> phi)` transforms nothing, so `otherwise("be a tuple")` can only fire if taking the attribute itself throws, and the whole pipeline is now `this.take("args")` with scaffolding around it. Anything carrying a `length` reaches `PrintfArgs`, and a non-tuple surfaces as a missing `tail` from inside `element` rather than as the message written here.

Thayorns (commented):
printf now walks the argument tuple itself instead of dispatching through `at`, so the argument keeps its own forma, and `%s` refuses `Φ.number` and `Φ.bool` outright rather than waiting for the bytes to turn out invalid UTF-8. The approach is right, and the tests pin both the refusals and the computed-string case that a stricter check would have broken. The set of refused formas is narrower than the argument behind it, though, and the i64 family walks straight past it.

Favixx (commented):
**Your CI is red for reasons that have nothing to do with this PR, and it matters more than it looks.**

The five failures are `MjPullTest.pullsFromProbes` and four `OyRemoteTest.checksPresenceOf*` — all of them network tests against objectionary.com. The `master` build started at 22:48, half an hour after yours, failed with the same set plus `MjProbeTest.findsProbesInOyRemote`, `ObjectsIndexTest.downloadsAndChecksFromRealSource` and `OyIndexedTest.containsInRealIndex`; builds from 23:03 onwards are green again. It was an outage, not your change.

The part worth acting on: the reactor is `eo` → `eo-parser` → `eo-printer` → `eo-maven-plugin` → …, and the build died in module **4 of 6**. `eo-runtime` never compiled, so `PrintfArgsTest` — every test in this PR — has not run once, on any of the four platforms. The guard rests on `normalized().forma()` returning exactly `Φ.number` and `Φ.bool`, and right now nothing has confirmed that. A re-run is the first thing this needs.

On the change itself: walking the cons list instead of dispatching through `at` is the right way to keep the forma, and the reasoning in the `NOT_TEXT` javadoc about naming what text is *not* convinced me — demanding `Φ.string` would indeed refuse `"a".concat "b"`. I have not repeated @Thayorns' three points; the `as-i64` one in particular still stands and my comment on the bytes test is a second instance of the same hole, deliberately enshrined this time.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -65,6 +67,23 @@ final class PrintfArgs {
...
+     * <p>Naming what text is <em>not</em>, rather than what it is, is
+     * deliberate. Only a literal has the forma {@code Φ.string}: every
+     * operation that builds a string returns something else
+     * ({@code "a".concat "b"} normalizes to {@code Φ.bytes}, {@code slice}
+     * to {@code Φ.string.slice}), and so does any user object that
+     * decorates a string. Demanding {@code Φ.string} would refuse all of
+     * those, which is a far bigger problem than the one this guard is
+     * here for.</p>
+     */
+    private static final Set<String> NOT_TEXT = new HashSet<>(
```
Favixx: `new HashSet<>(Arrays.asList(…))` is mutable, and it is `static final`, so anything in the JVM that gets hold of it can add to or clear the set that decides what `%s` refuses.

`Set.of("Φ.number", "Φ.bool")` is immutable, is a single expression, drops both the `Arrays` and `HashSet` imports this PR adds, and the codebase already uses it in four places. The project compiles at release 17, so it is available.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -65,6 +67,23 @@ final class PrintfArgs {
      */
     private static final double LONG_UPPER_LIMIT = 0x1.0p63;
 
+    /**
+     * Formas whose bytes are a value, never text, and so can never be the
```
Favixx: "Formas whose bytes are a value, never text" does not actually characterise this set, and the gap is worth naming in the doc rather than leaving for a reader to discover.

`Φ.bytes` is a value forma by exactly the same argument — `42.as-bytes` is eight IEEE-754 bytes and nothing else — and it is deliberately *not* here, with a new test enshrining that it stays acceptable. The real rule is narrower: these are the two formas that carry a value **and** cannot plausibly carry text, whereas bytes can be either and are given the benefit of the doubt.

Saying that here would also record the decision, so the next person who notices `printf "%s" 42.as-bytes` printing mojibake finds an answer instead of adding `Φ.bytes` to the set and breaking `"a".concat "b"`.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -170,27 +189,74 @@ List<Object> formatted() {
...
+     * reaches the object itself, forma intact. The list is built with the
+     * last argument at the head (see {@code tuple.eo}), so the walk takes
+     * as many tails as there are arguments after this one, counted off the
+     * tuple's own length so that a wrong count can only pick the wrong
+     * argument, never walk off the end into a terminated computation.</p>
+     *
+     * @param index Zero-based index of the argument
+     * @return The argument
+     */
+    private Phi element(final long index) {
```
Favixx: The javadoc claims a safety property the code does not have: "counted off the tuple's own length so that a wrong count can only pick the wrong argument, never walk off the end into a terminated computation".

That holds only while `length` is truthful. The count comes *from* the tuple, so a tuple whose `length` overstates its spine — the same disagreement @Thayorns points at from the other side — makes this loop take more `tail`s than exist, and the walk ends inside whatever the terminal cons hands back for a missing `tail`, not at a wrong argument.

The honest version is that the count is safe against `this.length` disagreeing with the tuple, which is the real hazard here, and says nothing about a malformed tuple. As written the sentence would discourage exactly the defensive check that would make it true.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
Favixx: This runs once per format specifier, not once per `formatted()` call: `element` is invoked from inside the loop, so a format with *k* specifiers over a tuple of *n* arguments dataizes `length` *k* times and performs up to *k·n* `take("tail")` calls. The code being replaced dispatched through `at` once per specifier, at whatever `at` costs.

For printf-sized inputs that is not a performance problem, and the forma is worth the walk. But both the length lookup and the spine are loop-invariant across the whole call — hoisting `size` next to `this.length` in the constructor, and walking once while collecting into an array, removes the repetition and, more usefully, removes the second reading of `length` that @Thayorns objects to.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -170,27 +189,74 @@ List<Object> formatted() {
...
-    private static Object fmt(final char symbol, final Dataized element) {
+    private static Object fmt(final char symbol, final Phi element) {
         if (!PrintfArgs.CONVERSION.containsKey(symbol)) {
             throw new ExFailure(
                 "The format %c is unsupported, only %s formats can be used",
                 symbol, "%s, %d, %f, %x, %b"
             );
         }
-        return PrintfArgs.CONVERSION.get(symbol).apply(element);
+        if (symbol == 's') {
```
Favixx, line 227: `CONVERSION` is the map that says what each conversion character does, and after this change it no longer does: `'s'` also has a precondition, and it lives here as an `if` rather than in the map. Anyone adding a conversion reads the map and misses this, and anyone reading the map now sees a description that is complete for four entries out of five.

Since the values are `Function<Dataized, Object>` and the guard needs the `Phi`, the tidy version is to keep the special case but say why: one line of comment (`// %s is the only conversion with a type precondition; see textual()`). Folding the check into a wrapper around the `'s'` entry would be tidier still if the signature allowed the `Phi` through.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -170,27 +189,74 @@ List<Object> formatted() {
...
+            PrintfArgs.textual(element);
+        }
+        return PrintfArgs.CONVERSION.get(symbol).apply(new Dataized(element));
```
Favixx: For `%s` this resolves the same object twice: `textual` calls `element.normalized()` a line above, and here `new Dataized(element)` dataizes it again. `normalized()` per its own contract does the resolution work and can propagate `ExFailure` exactly as dataization would, so the two are not sharing anything.

With `%1$s`-style repeats the multiplier grows — `"%1$s %1$s %1$s"` normalizes and dataizes the same argument three times each, on top of walking the spine to it three times. Passing the already-normalized `Phi` from `textual` into the `Dataized` would collapse half of that, and `textual` has it in hand.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -170,27 +189,74 @@ List<Object> formatted() {
...
+     * not valid UTF-8 at all were caught before, which is a matter of luck
+     * rather than of type.</p>
+     *
+     * @param element The argument
+     */
+    private static void textual(final Phi element) {
+        final String forma = element.normalized().forma();
+        if (PrintfArgs.NOT_TEXT.contains(forma)) {
+            throw new ExFailure(
+                "The argument of the '%%s' conversion is %s, whose bytes are a value and not text; use %s instead",
```
Favixx, line 258: The advice is generic where the code has just established the specific fact needed to make it right. A user who writes `printf "%s" true` is told to use "'%d', '%f', '%b' or '%x'" — three of which will also fail, because `%d`, `%f` and `%x` all go through `asNumber()` and a bool's single byte is not a number. Symmetrically, a user who writes `printf "%s" 42` is offered `%b`, which will fail on eight bytes.

You already know `forma` at this point, so the message can name the conversion that will actually work: `%b` for `Φ.bool`, and `%d`/`%f`/`%x` for `Φ.number`. That turns a message that costs the reader another failed run into one that fixes their code.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -201,14 +267,6 @@ private static Object fmt(final char symbol, final Dataized element) {
      * Dataized#asString()} does.
```
Favixx, line 269: This javadoc is now describing a different method than the one that exists. It presents the UTF-8 decode as *the* check that keeps mojibake out, which was true while the `@todo` sat under it; with `textual` in front, this is the second line of defence and catches only what the forma denylist lets through — genuinely malformed bytes, and value-carrying formas that are not `Φ.number` or `Φ.bool`.

That second clause is the interesting one and it is now undocumented anywhere in the source: removing the puzzle also removed the only in-tree explanation of why a byte-level check cannot be sufficient. The PR description explains it well; one sentence of it belongs here, because that is where the next person will look before deciding this method is redundant.

eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
```diff
@@ -61,7 +61,7 @@ void returnsCorrectArgumentsForMixedSubstitutions() {
             new PrintfArgs(
                 "This is the %s! This is %1$s as well! This is the %s",
                 3L,
```
Favixx: This fixture hands `PrintfArgs` a length of 3 for a tuple whose own `length` is 2 — the exact disagreement @Thayorns flags in `element`, written into the only test that exercises the multi-step walk.

It passes today because the bounds check uses `3` while the walk uses `2`, and the three specifiers reference indices 0, 0 and 1, all inside both. But it means the natural fix — bounds-checking against the tuple's own length so the two can never diverge — makes this test fail, for a reason that has nothing to do with mixed substitutions. Whoever takes that fix will be looking at a red test named `returnsCorrectArgumentsForMixedSubstitutions` and wondering what they broke.

Since this PR is what made the tuple the source of truth, `2L` here is part of the change.

eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
```diff
@@ -134,13 +134,57 @@ void rejectsNonStringBytesInsteadOfMojibake() {
...
+                () -> new PrintfArgs("%s", 1L, tuple).formatted(),
                 "must throw ExFailure, not silently return garbage text"
             ).getMessage(),
-            Matchers.containsString("not valid UTF-8")
+            Matchers.containsString("Φ.number")
+        );
+    }
+
+    @Test
+    void rejectsANumberWhoseBytesHappenToBeValidText() {
```
Favixx: All three new tests, and the reworked one above, build a one-element tuple: `length` plus `head`, no `tail`. With `size - 1 - index` equal to zero the loop body never executes, so the walk that this PR is built around — the reason `at` was abandoned — is covered by exactly one test in the file, and that one has the inconsistent length I mention on line 63.

A three-element tuple with a truthful `length`, asserting that `%3$s` reaches the first argument and `%s` the last, would pin the direction of the walk. The cons list is built last-at-the-head, and nothing here would catch that being reversed: with one element, head-first and head-last are the same fixture.

eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
```diff
@@ -134,13 +134,57 @@ void rejectsNonStringBytesInsteadOfMojibake() {
...
+                ExFailure.class,
+                () -> new PrintfArgs("%s", 1L, tuple).formatted(),
+                "must throw ExFailure instead of printing a control character"
+            ).getMessage(),
+            Matchers.containsString("Φ.bool")
+        );
+    }
+
+    @Test
+    void acceptsBytesGivenToTheStringConversion() {
```
Favixx: This test is correct and it should stay — but it is worth being explicit that it pins the remaining half of the bug, on the acceptable side of the line.

`42.as-bytes` produces `Φ.bytes` holding the eight raw IEEE-754 bytes of a number, and `printf "%s"` on it now prints exactly the mojibake this PR exists to stop — the same output as the `printf "%s" 42` in your description, reached one method call later. It is not hypothetical either: `printf.eo` two files over already uses `42.as-bytes` as an argument in `can-format-a-number-in-hex-with-width`, so the idiom is in the codebase.

Refusing bytes is not the answer, for the reason your javadoc gives. But between this test and `as-i64`, the guard's boundary is "the two formas a literal produces", not "value types", and #6161 will look closed while `printf "%s" 42.as-bytes` still misbehaves. Worth a line in the issue or a follow-up puzzle.

eo-runtime/src/main/eo/string/printf.eo
```diff
@@ -34,6 +34,12 @@
       "%25x"
       42.as-bytes
 
+  eq. ++> can-format-a-computed-string
```
Favixx, line 37: This is the right test to have added — it is the one that fails if someone "tightens" the guard into demanding `Φ.string`, which is the most likely future mistake here.

Your description names three shapes that must keep working: `concat` normalizing to `Φ.bytes`, `slice` giving `Φ.string.slice`, and a user object decorating a string. This covers the first. The third is the one a denylist protects by construction, but the second is a real forma produced by a real runtime object and costs one more `eq.` — `"abcd".slice 0 2` against `"ab"` — to make the description's claim a test rather than a paragraph.

eo-runtime/src/main/eo/string/printf.eo
```diff
@@ -177,6 +183,10 @@
         42.as-i64
       "42"
 
+  printf *1 --> stops-on-a-bool-as-a-string
```
Favixx, line 186: Both new `-->` tests assert that the program stops, which is what they can assert — and that is precisely what the code did *before* this PR for `TRUE`, since a control character eventually failed somewhere downstream, and for `42` when the bytes happened not to decode.

The improvement this PR delivers is *which* failure the user gets: a message naming the forma instead of mojibake or a UTF-8 complaint. Nothing at the EO level pins that, and the only tests that do are the three `containsString("Φ.number")` / `("Φ.bool")` assertions in `PrintfArgsTest` — the file that, per my note in the review body, has not been executed by CI on any platform. Until that build is re-run, the central claim of the PR is unverified by anything.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd: Confirmed and fixed. `42.as-i64` is `Φ.number.as-i64`, and `42.as-i64.as-number` is `Φ.i64.as-number`, so a list of whole names would have had to grow a pair at a time.

The check now matches the root of the forma against `number`, `bool`, `i8`, `i16`, `i32`, `i64`. That catches both of those, and `1.plus 2` as well, since it normalizes to plain `Φ.number`. `Φ.string.slice` and `Φ.bytes.concat` still pass, which is what the guard has to keep doing.

Added `stops-on-an-i64-as-a-string` at the EO level.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd: Right, and the two could disagree. There is one source now: `PrintfArgs` takes only the tuple, reads the count once at the top of `formatted`, and passes it to `element`. The `length` field is gone.

eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
morphqdd: Fixed. It was a leftover from splitting the old two-argument call. Now it is a single `must` that dataizes `length`, so a missing or non-numeric one fails here with the message written next to it rather than as a missing `tail` further in.

morphqdd:
Rebased on master and made the forma set immutable with `Set.of`, matching `PhDefault.SPECIAL` in the same module. Same point @Thayorns raised on #6197; it applied here too.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd: Already done in an earlier push, same point @Thayorns made on #6197: it is `Set.of` now, and both imports went with it.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd: Right, and the gap is the interesting part. Documented on the constant now: bytes are a value forma by the same argument and are deliberately absent, because they carry text as readily as a number and every computed string normalizes to them, so refusing them would refuse `"a".concat "b"`. The rule is "carries a value and cannot plausibly carry text", and the price is that `printf "%s" 42.as-bytes` still prints mojibake.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd: The sentence is gone. It was written when `this.length` still existed and the point was that the two could disagree; that field has since been removed, so the claim was both overstated and stale. It says nothing about a malformed tuple now.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd: The `length` read is once per `formatted()` call, not per specifier: it is hoisted to the top of the loop and passed into `element`. The spine walk is still per specifier; for printf-sized inputs I left it, since collecting the whole spine up front would walk it even for a format that touches one argument.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd: Added the comment. The signature is the reason it cannot live in the map: the values take a `Dataized` and the check needs the `Phi`.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd: Fixed. `textual` returns the resolved object and `fmt` dataizes that, so `%s` resolves once instead of twice.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd: Good catch, the advice was wrong in both directions. It now names only what will work: `'%b'` for a bool, `'%d', '%f' or '%x'` for the number family.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
morphqdd: Rewritten. It says it is the second line of defence behind `textual`, and that a byte-level check cannot be the only one because whether a value's bytes decode is luck rather than type: `42` decodes cleanly and `3.14` does not, though neither is text.

eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
morphqdd: That fixture no longer passes a length at all: the constructor takes only the tuple, so the disagreement it encoded cannot be expressed any more.

eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
morphqdd: Added `walksTheSpineInTheRightDirection`: a three-element tuple with a truthful length, asserting `%s` reaches the first argument and `%3$s` the head. With one element the two directions are the same fixture, as you say.

eo-runtime/src/test/java/org/eolang/EO_string/PrintfArgsTest.java
morphqdd: Agreed on the substance, and it is now written on the constant rather than only in a test: bytes stay acceptable deliberately, and the cost is that `42.as-bytes` still prints mojibake. I dropped that test to fit under the size cap, not because the case stopped mattering.

morphqdd:
All eleven notes are addressed. Two things worth flagging rather than leaving to be discovered:

The review took the PR to 249 lines against the 200 cap. I condensed the javadoc first, keeping every decision you asked to have recorded, which got it to 230. The rest came from dropping two Java tests, `rejectsANumberWhoseBytesHappenToBeValidText` and `rejectsABooleanGivenToTheStringConversion` — both cases are covered at the EO level by `stops-on-a-number-as-a-string` and `stops-on-a-bool-as-a-string`, which are the portable contract and the ones that matter more here. That is a deliberate trade, not a quiet reduction in coverage. `walksTheSpineInTheRightDirection` stayed, since nothing else pins the walk.

The Windows leg is red on `SavedTest.exposesNoPartiallyWrittenFileToConcurrentReader` in `eo-maven-plugin`. I pulled the full log rather than the summary line: it is a different module from anything this PR touches, and it is the known Windows behaviour of `Files.move` with `REPLACE_EXISTING` against a concurrent reader.

eo-runtime/src/main/eo/string/printf.eo
```diff
@@ -197,6 +207,10 @@
     "%s"
     3.14
 
+  printf *1 --> stops-on-a-number-as-a-string
```
gemshrine, line 210: These `-->` examples don't actually pin the type check. A `-->` transpiles to `assertThrows(Exception, asBool(result))`, and for a value argument the *pre-fix* path throws there anyway: `printf "%s" 42` used to return the 8 mojibake bytes `40-45-00-...`, and `asBool` on 8 bytes throws regardless of whether `%s` rejected the type — so `stops-on-a-number-as-a-string` (and `stops-on-an-i64-as-a-string`, another 8-byte value) stays green even with the fix reverted.

The discriminating coverage is the Java `rejectsNonStringBytesInsteadOfMojibake`, but it uses `3.14`, whose bytes are *invalid* UTF-8 and were already rejected before this PR (with the old `not valid UTF-8` message). The one case that regressed in #5943 — an integer like `42` whose bytes *are* valid UTF-8 — has no test that fails on the old code. A Java test asserting `containsString("Φ.number")` for `42` (not `3.14`) would pin exactly the bug this closes.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -65,7 +65,26 @@ final class PrintfArgs {
...
 
+    /**
+     * Roots of formas that carry a value and cannot plausibly carry text,
+     * so no {@code %s} argument may have one.
+     *
+     * <p>Naming what text is not is deliberate: only a literal is
+     * {@code Φ.string}, while {@code "a".concat "b"} normalizes to
+     * {@code Φ.bytes}, so demanding {@code Φ.string} would refuse every
+     * computed string. {@code Φ.bytes} is absent for the same reason,
+     * at the price of {@code 42.as-bytes} still printing mojibake. Roots
```
gemshrine, line 76: The residual is broader than `42.as-bytes`. I dataized the whole unsigned series: `42.as-u64` normalizes to `Φ.bytes.as-u64` (root `bytes`, not `number`), so it is let through here and then `00-00-00-00-00-00-00-2A` decodes as valid UTF-8 — `printf "%s" 42.as-u64` prints seven NULs and `*` with no error, same as `as-bytes`. `as-u8/u16/u32` land on `bytes` too. Not a blocker (it's the documented `Φ.bytes` hole), but the comment reads as if `42.as-bytes` were the sole casualty; naming the `as-u*` conversions here, or in the follow-up, keeps it from looking closed.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -164,51 +177,103 @@ List<Object> formatted() {
...
 
     /**
      * Convert the {@code element} to a string for the {@code %s} conversion,
      * rejecting bytes that are not valid UTF-8 instead of silently decoding
-     * an arbitrary byte sequence (e.g. the raw IEEE-754 bytes of a {@code
-     * number} argument) into mojibake with no error, as {@link
+     * an arbitrary byte sequence into mojibake with no error, as {@link
      * Dataized#asString()} does.
+     *
+     * <p>Second line of defence, behind {@link #textual(Phi)}: whether a
```
gemshrine, line 271: Dropping the `@todo #5943` entirely loses the tracker for a hole that is only half closed. The todo's subject was *a value whose raw bytes happen to be valid UTF-8*; this PR closes that for signed/`number`/`bool` formas via the forma root, but leaves it wide open for anything rooted at `Φ.bytes` (`as-bytes`, the `as-u*` series above). Consider keeping a reduced `@todo` here, or opening a follow-up issue, so the remaining `Φ.bytes` case stays visible rather than reading as resolved.

eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
```diff
@@ -65,7 +65,26 @@ final class PrintfArgs {
...
+     * <p>Naming what text is not is deliberate: only a literal is
+     * {@code Φ.string}, while {@code "a".concat "b"} normalizes to
+     * {@code Φ.bytes}, so demanding {@code Φ.string} would refuse every
+     * computed string. {@code Φ.bytes} is absent for the same reason,
+     * at the price of {@code 42.as-bytes} still printing mojibake. Roots
+     * are matched, not whole names, since a conversion keeps its origin
+     * at the front ({@code 42.as-i64} is {@code Φ.number.as-i64}).</p>
+     */
+    private static final Set<String> VALUES = Set.of(
+        "number", "bool", "i8", "i16", "i32", "i64"
```
gemshrine, line 81: `i8` looks unreachable as a root. Walking the conversions, the forma root that actually appears is one step up from the target: `as-i64` -> `Φ.number.as-i64` (root `number`), `as-i32` -> `Φ.i64.as-i32` (root `i64`), `as-i16` -> `Φ.i32.as-i16` (root `i32`), `as-i8` -> `Φ.i16.as-i8` (root `i16`). So no `as-*` argument ever presents root `i8`, and nothing else in the tests does either. Either it guards a bare `i8` object (worth a test that exercises it) or it can be dropped; `i16` is the smallest root the chain can produce.

gemshrine (commented):
The forma-root approach is the right way to recover the type that `tuple.at` erased, and `walksTheSpineInTheRightDirection` / the `Φ.number` assertion pin the two load-bearing pieces well. Four notes, all checked at runtime on this branch against the actual formas: the new EO `-->` rejection examples don't discriminate the fix (the `#5943` headline case slips past them); the acknowledged `Φ.bytes` residual is wider than `42.as-bytes` (the whole unsigned series shares it); the removed `@todo` was only partly resolved; and one denylist entry looks unreachable.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (198 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3153](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+597](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+136](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+1343](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6167, opened 2026-07-30
Move the emitting of Java files out of Transpiling

### LINKED ISSUE #6116: Transpiling.java class complexity exceeds the checkstyle limit due to multiple unrelated jobs
`eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java:54` carries `@checkstyle ClassFanOutComplexityCheck (500 lines)`. Qulice registers `ClassFanOutComplexity` with `max=30`, and with the marker stripped checkstyle reports `Class Fan-Out Complexity is 43 (max allowed is 30)` on the class declaration.

The suppression is hiding a real concentration rather than a checker quirk. One 509-line class builds the XSL train (`train`, `compiled`), wraps it in measurements (`measured`), assembles the transformation function (`transpilation`), arbitrates the transpile cache (`transpiled`, `version`), and then emits and counts the generated Java files (`javaGenerated`, `cached`). The 43 dependencies are the sum of four unrelated jobs, so the number will keep drifting up as any one of them grows.

Two cohesive groups can move out. `train`, `compiled`, `measured` and `transpilation` exist only to produce the `Function<XML, XML>`; `javaGenerated` and `cached` exist only to write Java files. Each group takes its own imports with it, which brings the remainder under the cap and lets the marker go. `version()` has to stay next to the `XSLS` and `IMPORTS` arrays so the cache-key fingerprinting from #5578 and #6032 keeps deriving from the same source.

Thayorns:
`Transpiling` carried `@checkstyle ClassFanOutComplexityCheck`, and with the marker stripped checkstyle reported a fan-out of 43 against a limit of 30. The suppression was hiding four unrelated jobs sharing one class, so the number drifts up whenever any one of them grows.

This moves the smaller of the two cohesive groups out. `javaGenerated` and `cached` existed only to write the generated Java files, and they took `Filter`, `AtomicInteger`, `Collectors`, `Footprint`, `FpJavaGenerated`, `FileGenerationReport`, `JavaPlaced`, `FpIfReleased`, `FpAppliedWithCache` and `RewritePolicy` with them. They now live in `JavaFiles`, which keeps the global configuration as its four attributes and takes the per-tojo values as arguments, the same four the old private method took. `version()` stays where it was, next to `XSLS` and `IMPORTS`, so the cache-key fingerprinting from #5578 and #6032 still derives from the same source.

Measured with checkstyle 10.26.1 and `max=30`: the fan-out of `Transpiling` drops from 43 to 36, and `JavaFiles` sits at 14. That is not yet under the cap, so the marker stays and a puzzle records what is left: the XSL train group, `train`, `compiled`, `measured` and `transpilation`, is the thirteen dependencies that still have to go before the marker can be deleted. Splitting it this way keeps the change reviewable instead of moving 137 lines in one step.

`MjTranspileTest` exercises the new class on the real transpile path and stays green, and `JavaFilesTest` covers it directly.

Closes #6116


Thayorns:
@yegor256 this takes only the Java-emitting half out, so the checkstyle marker stays for now and a puzzle tracks the XSL train group. Happy to explain the split or fold both groups into one pull request instead.

Thayorns:
@yegor256 ready for review

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * The Java files that a transpiled XMIR is written out as.
+ * <p>
+ *     One XMIR carries a class per object, and every one of them becomes
+ *     a Java file under the generated sources directory, taken from the
+ *     global cache when the tojo is at a released hash and put there when
+ *     it is not.
```
Favixx, line 24: The second half of this sentence describes the opposite of what the code does. `FpIfReleased(hsh, first, second)` documents `first` as "used if a version is released and a hash is present" and `second` as "if a version is not released or a hash is not present", and here `first` is the `FpAppliedWithCache` chain while `second` is the bare `java` footprint.

So the cache is consulted *and* populated when the hash is present, and is not touched at all when it is absent — whereas "taken from the global cache when the tojo is at a released hash and put there when it is not" says files are written into the cache precisely in the case where nothing goes near it.

This is the class javadoc of a new class, so it is what every future reader starts from. "…taken from and written back to the global cache when the tojo has a released hash, and generated without touching the cache when it does not" is the same length.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -0,0 +1,133 @@
...
+ * </p>
+ * @since 0.74
```
Favixx, line 26: `0.74` looks like a digit slip. The newest tag on the repository is `0.62.1`, and the three collaborators this class was extracted alongside — `JavaPlaced`, `RewritePolicy`, `FileGenerationReport` — all carry `@since 0.56.7`. A class added today should name the release it will first appear in, which is somewhere around `0.63.0`, not eleven minors ahead.

Same value in `JavaFilesTest`, so both need the correction.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final String JAVA = "java";
+
+    /**
+     * Generated sources directory.
+     */
+    private final Path generated;
+
+    /**
+     * Cache directory for transpiled sources, with the cache-key version
+     * segment already resolved into it.
```
Favixx, line 42: "with the cache-key version segment already resolved into it" is an honest field comment, and it is also the sign that the layout has been split in two. `CachePath` exists to own `base/semver/hash/tail` in one place; after this change `Transpiling` contributes `base` and `semver` at the call site, `JavaFiles` contributes `hash` and `tail` in the lambda below, and `CachePath` is no longer involved in this path at all — while it stays in use elsewhere in the same class (line 188) and in `Linting`.

The result is identical today, as I checked. But a layout that one class used to know is now a convention between two, and the class built to express it is bypassed. Passing `cdir` and `this.version()` through and keeping `new CachePath(...)` inside `cached()` would move the code without dissolving the abstraction — which is the usual bar for a refactor that is not supposed to change anything.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -0,0 +1,133 @@
...
+     * Generate java files and count them.
+     * @param rewrite Rewrite .java files even if they exist
+     * @param target Full target path to XMIR after transpilation optimizations
+     * @param hsh Tojo hash
+     * @param tests Whether to generate test sources for this tojo
+     * @return Amount of generated .java files
+     * @throws IOException If fails to save files
+     * @checkstyle ParameterNumberCheck (5 lines)
+     */
+    int total(
```
Favixx, line 73: The `@checkstyle ParameterNumberCheck` suppression travelled with the method, and this was the moment it did not have to. The whole point of the extraction is that the four global values became fields; what is left in the signature is four *per-tojo* values, which is one suppression's worth of the same problem the class was created to reduce.

`rewrite`, `hsh` and `tests` are all derived from the tojo in `transpiled()`, and `target` from it too. A small carrier — or passing the `TjForeign` and letting `JavaFiles` ask — turns `total(rewrite, target, hsh, tests)` into `total(tojo)` and takes the marker with it. Worth at least recording as a puzzle, since the PR is about exactly this kind of debt.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -0,0 +1,133 @@
...
+        final boolean rewrite,
+        final Path target,
+        final String hsh,
+        final boolean tests
+    ) throws IOException {
+        final AtomicInteger saved = new AtomicInteger(0);
+        if (Files.exists(target)) {
```
Favixx, line 80: A missing `target` returns `0` and says nothing. Inside a private method of `Transpiling`, called immediately after the code that writes that very file, that was defensible — the caller knew the file had just been created. As the entry point of a collaborator with its own name and its own test, `0` now means two different things: "this XMIR carries no classes" and "the XMIR you named is not there".

The first is normal, the second is a bug in the caller, and the count cannot tell them apart. A `Logger.warn` in the `else`, or letting the absence throw, keeps the distinction that the class boundary just erased.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -0,0 +1,133 @@
...
+            final Xnav object = new Xnav(target).element("object");
+            final Collection<Xnav> classes = object.elements(Filter.withName("class"))
+                .collect(Collectors.toList());
+            final boolean atom = object.path("/object/o/o[@name='λ']").findAny().isPresent();
```
Favixx, line 84: Worth confirming while this code is being moved rather than after: `object` is already the `<object>` element (`new Xnav(target).element("object")` on the line above), and the XPath handed to it starts at `/object/` again.

If `Xnav.path` evaluates from the document root, this is right and the doubling is harmless. If it evaluates relative to the current node, it is looking for `/object/object/o/o[@name='λ']`, `atom` is always `false`, and the `!atom || jname.endsWith("Test")` guard below has never filtered anything — every class of every atom would be written, which would show up as extra generated files rather than as a failure, i.e. exactly the sort of thing that survives for years.

It came over unchanged from `Transpiling`, so this is not a regression either way. But the new `JavaFilesTest` is the first place it can be pinned cheaply: one XMIR with an `o[@name='λ']` and a non-`Test` class, asserting `0`.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -0,0 +1,133 @@
...
+        return saved.get();
+    }
+
+    /**
+     * Cached path supplier for a generated Java file.
+     * @param hsh Hash
+     * @param jname Java class name
+     * @return Supplier of cached path
+     */
+    private Supplier<Path> cached(final String hsh, final String jname) {
```
Favixx, line 125: `new Place(jname).make(this.generated, JAVA)` is computed here and again as `tgt` in `total`, for every class of every XMIR — the same two objects built twice per file so that one of them can be relativized against the directory it was just built from.

`cached` is called from the loop that already holds `tgt`, so `cached(hsh, tgt)` with `this.generated.relativize(tgt)` inside does the same work once. It also removes the only reason this method needs `jname` at all.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -218,7 +210,9 @@ private int transpiled(final TjForeign tojo) throws IOException {
             rewrite.compareAndSet(false, true);
             new Saved(transform.apply(xmir).toString(), target).value();
         }
-        return this.javaGenerated(
+        return new JavaFiles(
+            this.generatedDir, cdir.resolve(this.version()), this.cacheEnabled
```
Favixx, line 214: A new `JavaFiles` is constructed per tojo, inside `transpiled()`, and all three of its arguments are constant across the whole run: `generatedDir` and `cacheEnabled` are fields, and `cdir.resolve(this.version())` recomputes `version()` — which per its own contract fingerprints the XSL train — once per transpiled file.

Since the class was designed with "the global configuration as its four attributes", it wants to be built once. A field initialised in the constructor, or a lazily-held instance, matches the design the javadoc describes and stops `version()` from being re-derived for every tojo in the project.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -55,11 +52,6 @@ final class Transpiling implements Step {
      */
     private static final String CACHE = "transpiled";
```
Favixx, line 53: `JAVA` moved into `JavaFiles` and `CACHE` stayed here, so the two constants that together name the cache path now live in different classes — `Transpiling.CACHE` builds the base at line 182, `JavaFiles.JAVA` builds the tail. Both are still used, so nothing is dead; it is the pairing that got separated.

If `JavaFiles` took the cache root rather than a pre-resolved path (see my comment on its `cache` field), `CACHE` would travel with `JAVA` and this class would stop knowing the name of a directory it no longer writes into.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Test case for {@link JavaFiles}.
+ * @since 0.74
+ */
+@ExtendWith(MktmpResolver.class)
+final class JavaFilesTest {
+
+    @Test
+    void countsJavaFilesItWrites(@Mktmp final Path temp) throws IOException {
+        final long seed = new Random().nextLong();
```
Favixx, line 26: An unseeded `new Random()` makes every run of this test a different test. The seed is reported in the assertion message, which is the right instinct, but it only appears when the assertion is the thing that fails — an exception from `Place`, `Saved` or the XML parser on a name this run happened to generate carries no seed at all, and CI logs would show a stack trace with nothing to reproduce it from.

The non-ASCII names look deliberate and are worth keeping. Two fixed cases — one plain, one with `bär`/`Ω` — cover the same ground deterministically, and if the intent really is fuzzing then the seed belongs somewhere it survives an exception, not only an assertion.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
```diff
@@ -0,0 +1,48 @@
...
+            xmir
+        ).value();
+        MatcherAssert.assertThat(
+            String.format(
+                "the Java files written are not as many as the classes in the XMIR, seed: %d",
+                seed
+            ),
+            new JavaFiles(
+                temp.resolve("generated"), temp.resolve("cache").resolve("1.0-SNAPSHOT"), false
+            ).total(true, xmir, "", false),
```
Favixx, line 44: `hsh` is `""` and caching is `false`, which together select the branch where none of the moved risk lives. `FpIfReleased` treats an empty hash as "not cacheable" and takes its `second` footprint, so `FpAppliedWithCache`, `RewritePolicy` and — crucially — `cached()` are never reached.

`cached()` is the method whose path arithmetic changed shape in this PR: it went from a `CachePath` to a hand-built `cache.resolve(hsh).resolve(tail)`. I verified by hand that the two agree, but the only test of this class exercises neither. A second case with a non-empty hash and caching on, asserting that the file appears under `cache/<hsh>/<tail>`, is what would have caught it if they had not agreed.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
```diff
@@ -0,0 +1,48 @@
...
+            Matchers.equalTo(1)
```
Favixx, line 45: The assertion checks the counter and not the thing the class is named after. `total` returns `saved.get()`, which `FileGenerationReport` increments — so this passes if the report is invoked, whether or not any file reached `generated/`.

For a class whose entire responsibility is "the Java files that a transpiled XMIR is written out as", the file existing is the assertion, and the count is the corroborating detail. `FileMatchers.anExistingFile()` on the expected `Place` path, next to the `equalTo(1)`, states both.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Test case for {@link JavaFiles}.
+ * @since 0.74
+ */
+@ExtendWith(MktmpResolver.class)
+final class JavaFilesTest {
+
+    @Test
+    void countsJavaFilesItWrites(@Mktmp final Path temp) throws IOException {
```
Favixx, line 25: One test for a class extracted specifically because it was doing too much. The branches it does not enter: the atom filter (`!atom || jname.endsWith("Test")`, so an XMIR with `o[@name='λ']` and a non-`Test` class, expecting `0`), the multi-class XMIR (the `Collection` and the counter both exist for it, and one class cannot tell `saved` from a constant `1`), and the missing-`target` early return.

Each is three or four lines given the fixture already here. The PR's own argument is that `Transpiling` was too big to reason about; the payoff for splitting it is that these branches finally have somewhere to be tested.

Favixx (commented):
I checked the one thing an extract-class refactor has to get right, and it holds: the cache path is byte-for-byte what it was. `CachePath.get()` is `base.resolve(semver).resolve(hash).resolve(tail)`, the old call passed `(cacheDir/transpiled, version(), hsh, tail)`, and the new pair — `cdir.resolve(version())` at the call site plus `cache.resolve(hsh).resolve(tail)` in `JavaFiles` — lands on `cacheDir/transpiled/<version>/<hsh>/<tail>` all the same. Nothing already in a user's cache is orphaned.

Two things in the description do not match the diff, and both matter because #6116 is closed by this.

**"The marker stays" — which marker?** `Transpiling.java` carries `@checkstyle MemberNameCheck` five times and one `@checkstyle ParameterNumberCheck`, and that is all: there is no `ClassFanOutComplexityCheck` suppression in the class javadoc on this branch, and none on `master` either. If the fan-out limit is enforced anywhere, it is not from this file, so "with the marker stripped checkstyle reported 43" describes a local experiment rather than a state the repository was in. Since the argument for the split rests on that number, it is worth saying where the 30 comes from — stock checkstyle 10.26.1, as you note, is not what qulice runs here.

**The puzzle is not in the diff.** "A puzzle records what is left: the XSL train group, `train`, `compiled`, `measured` and `transpilation`" — there is no `@todo` in the changed files. That list is exactly the kind of thing PDD exists to turn into an issue, and right now it survives only in this description, which no future reader of `Transpiling` will have.

Also: `pr-size` is red at 268 against the 200 limit, so this cannot merge as is. The new class and its test are 181 of those lines and are one coherent unit; the `Transpiling` side is 87. That is not obviously splittable, but it is worth deciding deliberately rather than at merge time.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * Test case for {@link JavaFiles}.
+ * @since 0.74
+ */
+@ExtendWith(MktmpResolver.class)
+final class JavaFilesTest {
```
morphqdd, line 22: This only covers the "class gets emitted" path. The `!atom || jname.endsWith("Test")` skip branch moved here unchanged from Transpiling but isn't exercised directly by this class's own test. Presumably still covered indirectly via MjTranspileTest, but worth locking it down here too.

Thayorns: I tried to lock it down and could not, which turned out to be the interesting part: the branch cannot be taken at all. `Xnav.path` evaluates relative to the node it is called on, and `object` is already the `<object>` element, so `path("/object/o/o[@name='λ']")` is looking for an `<object>` nested inside `<object>`. No XMIR has one, so `atom` is always `false` and `!atom || jname.endsWith("Test")` is always true.

I checked both shapes against the same call: with `<object><o><o name='λ'/></o>…</object>` the current expression finds nothing while the relative `o/o[@name='λ']` finds it. So it is not covered indirectly via `MjTranspileTest` either — nothing covers it, because nothing can.

The code came over from `Transpiling` untouched and the bug predates this PR, so I filed #6247 rather than fixing it here: dropping the `/object` prefix would stop generating atoms' non-test classes, which deserves its own PR.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
yegor256: @Favixx what this comment is for? Sounds like noise to me. Please, post only negative comments, requiring rework.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
yegor256: @Favixx this comment sounds like spam to me.

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (268 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+590](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+148](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+3191](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2685](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6176, opened 2026-07-31
Fix #6174: probe FQNs from void `@args` annotations

### LINKED ISSUE #6174: `Φ.chunk` is not probed causing `malloc.of` to fail due to `ClassNotFoundException`
`EOmalloc$EOof.java:35` resolves the `chunk` object at runtime with `Phi.Φ.take("chunk").copy()`. But the only mention of `chunk` anywhere in `eo-runtime/src/main/eo/malloc.eo` is the type annotation on line 77, `? > scope /{Q.chunk}`, which the parser emits as `<o args="Φ.chunk" base="∅" name="scope"/>` — an `args` attribute on a void object.

`add-probes.xsl` cannot see that. Its candidate set on lines 63 and 80 is `//o[not(eo:abstract(.)) and not(eo:void(.))]`, so the void `scope` is excluded outright, and line 91 tokenizes only `@base` — the stylesheet never reads `args`. No probe is emitted for `Φ.chunk`, so `chunk` never enters `eo-foreign.json`, `objects/chunk.eo` is never pulled, and `org.eolang.EOchunk` is never transpiled. Any program that touches `malloc` then fails with `Couldn't find object 'Φ.chunk' ... Caused by: java.lang.ClassNotFoundException: org.eolang.EOchunk`, thrown from `PhPackage.take` under `EOmalloc$EOof.lambda`. The jar makes this especially confusing, because it does ship the nested atoms `EOchunk$EOread`, `EOchunk$EOwrite`, `EOchunk$EOsize` and `EOchunk$EOresized` — only the outer `EOchunk.class` is absent, since that one is expected to come from transpiling `chunk.eo`. Adding any throwaway `Q.chunk` reference to a project is enough to make the probe pull and transpile it, after which the failure disappears.

The smallest fix is to give `malloc.eo` one real, probeable reference to `Q.chunk`, so that the dependency its Java atom creates becomes visible to the resolver. Alternatively, extend the candidate expression in `add-probes.xsl` to also collect FQNs from `args` annotations on void attributes.

yegor256:
Closes #6174.

## Problem

`EOmalloc$EOof.java` resolves `Φ.chunk` at runtime, but the only mention of `chunk` in `malloc.eo` is the callback arg-types annotation on the void `scope`:

```
? > scope /{Q.chunk}
```

The parser emits this as an `@args` attribute on the void object. `add-probes.xsl` never looks at `@args`: its candidate set is `//o[not(eo:abstract(.)) and not(eo:void(.))]` (void objects are excluded outright) and it only tokenizes `@base`. So no probe was emitted for `Φ.chunk`, `chunk` never entered `eo-foreign.json`, `objects/chunk.eo` was never pulled, and `org.eolang.EOchunk` was never transpiled. Any program touching `malloc` then failed with:

```
Couldn't find object 'Φ.chunk' ... Caused by: java.lang.ClassNotFoundException: org.eolang.EOchunk
```

## Fix

Extend the candidate collection in `add-probes.xsl` to also walk void objects that carry an `@args` annotation (`//o[eo:void(.) and @args]`). The new `args` mode tokenizes each homed forma in the space-separated `@args` list the same way composite bases are tokenized, so `Φ.chunk` becomes a probe. Generic type variables (`A`–`F`), which are not real objects, are skipped by keeping only `Φ`-prefixed members; bare `Φ` is dropped by the existing downstream filter.

This is the root-cause fix suggested in the issue, so no throwaway `Q.chunk` reference in `malloc.eo` is needed.

## Tests

Added `probe-packs/add-probes-from-void-args.yaml` covering an atom whose void carries `/{Q.chunk}`, asserting `chunk` is now probed. `ProbesTest` passes (8/8).

https://claude.ai/code/session_01X7prTPGBmFdnj31jZzhxbb

---
_Generated by [Claude Code](https://claude.ai/code/session_01X7prTPGBmFdnj31jZzhxbb)_

### DIFF
--- a/eo-maven-plugin/src/main/resources/org/eolang/maven/probe/add-probes.xsl
+++ b/eo-maven-plugin/src/main/resources/org/eolang/maven/probe/add-probes.xsl
@@ -61,6 +61,7 @@
   <xsl:template match="/object[not(metas)]">
     <xsl:variable name="candidates" as="element()*">
       <xsl:apply-templates select="//o[not(eo:abstract(.)) and not(eo:void(.))]" mode="create"/>
+      <xsl:apply-templates select="//o[eo:void(.) and @args]" mode="args"/>
     </xsl:variable>
     <xsl:variable name="probes" select="distinct-values($candidates/text())[not(eo:contains-any-of(., ('ξ', 'ρ', 'φ'))) and not(.='Φ') and not(.='Φ̇')]"/>
     <xsl:copy>
@@ -78,6 +79,7 @@
   <xsl:template match="/object/metas">
     <xsl:variable name="candidates" as="element()*">
       <xsl:apply-templates select="//o[not(eo:abstract(.)) and not(eo:void(.))]" mode="create"/>
+      <xsl:apply-templates select="//o[eo:void(.) and @args]" mode="args"/>
     </xsl:variable>
     <xsl:copy>
       <xsl:apply-templates select="node()|@*"/>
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (33) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2876](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6180, opened 2026-07-31
fix(eo-maven-plugin): #6175 complete a package on probe so bare sibling refs resolve

### LINKED ISSUE #6175: Partially pulled package makes bare sibling references misresolve to Φ.name
The parser resolves a bare name against the package the file lives in, and that works correctly whenever the package is fully present on disk. When the sibling is missing it silently falls back to `Φ.<name>`. Since the pull fetches only individually probed objects, packages arrive partially populated and that fallback fires on objects that do exist upstream.

In a build where `tuple.eo` was pulled together with six of the 22 objects in `eo-runtime/src/main/eo/tuple/`, the correlation with the parsed XMIR is exact. Present at parse time (`concat`, `contains`, `each`, `index-of`, `reduced`, `with`) — every bare reference resolved as `Φ.tuple.<name>`. Absent (`eachi`, `reducedi`, `withouti`) — every one resolved as `Φ.eachi`, `Φ.reducedi`, `Φ.withouti`. A fully present package resolves correctly across the board, which is what the same build shows for a local 12-file package.

Hence a chicken-and-egg. `tuple/each.eo:13` refers to a bare `eachi`; the scan cannot see `tuple/eachi.eo` because nothing probed it; the base becomes `Φ.eachi`; and that is unpullable, since `objects/eachi.eo` does not exist in objectionary/home while `objects/tuple/eachi.eo` does. Dataizing `tuple.each` then terminates with `Couldn't find object 'Φ.eachi' because there's no class 'org.eolang.EOeachi' or package-info class: 'org.eolang.EO_eachi.package-info'`.

The fix is to pull the remaining objects of a package whenever any object from that package is pulled, so the parser's package scan sees a complete directory before it classifies bare names. Failing that, the fallback to `Φ.<name>` should fail loudly instead of silently reclassifying, so a partially pulled package cannot produce a base that no longer exists.

yegor256:
Fixes #6175.

## Problem

The pull step fetches only the objects that were individually probed, so a package arrives on disk **partially populated**. The parser resolves a bare name against the directory the file lives in, and when a sibling is missing it silently falls back to `Φ.<name>`. On a partial package that fallback fires on objects that *do* exist upstream, producing a base that no longer exists:

- Present at parse time (`concat`, `contains`, `each`, …) → every bare reference resolved as `Φ.tuple.…` (correct).
- Absent (`eachi`, `reducedi`, `withouti`) → resolved as `Φ.eachi`, `Φ.reducedi`, `Φ.withouti` (wrong).

This is a chicken-and-egg: `tuple/each.eo` refers to bare `eachi`; the scan can't see `tuple/eachi.eo` because nothing probed it; the base becomes `Φ.eachi`; and `objects/eachi.eo` does not exist in `objectionary/home` (only `objects/tuple/eachi.eo` does), so dataizing `tuple.each` fails with `Couldn't find object 'Φ.eachi'`.

## Fix

Implements the primary fix from the issue: **pull the remaining objects of a package whenever any object from that package is pulled**, so the parser's package scan always sees a complete directory before it classifies bare names.

Whenever `Probing` discovers an object that lives in a package, it now also registers every other object of that package. The assemble loop (`parse → probe → pull`) then pulls the whole directory in the same cycle, before those objects are parsed, so bare sibling references resolve correctly. Enumeration uses the authoritative objects index (`objectionary.lst`) that the build already relies on.

## Changes

- `Objectionary`: new `children(pkg)` method — "what objects live directly in this package?" — implemented by `OyIndexed` (from the objects index, with the same index→delegate fallback as `contains`/`isDirectory`), delegated by `OyCached`, and a graceful empty result from `OyRemote` (the raw remote can't enumerate cheaply); `Objectionary.Fake` gains a `children` lambda.
- `ObjectsIndex`: new `children(pkg)` returning objects located directly inside a package (direct file children only, not sub-packages); the `org.eolang.` prefix stripping is refactored into a shared helper.
- `Probing`: after an object is probed, `complete(...)` registers the rest of its package (guarded by a concurrent set so each package is enumerated once).

## Tests

- `ProbingTest#completesPartiallyProbedPackage` — probing an object in a package registers the siblings that were never probed directly.
- `ObjectsIndexTest` — `children` lists only direct children and strips the `org.eolang.` prefix.
- `OyIndexedTest` — `children` reads from the index and falls back to the delegate when the index is broken.

Note: two pre-existing `OyRemoteTest` failures (`checksPresenceOfDirectory*`, `@ExtendWith(WeAreOnline.class)`) reproduce on the base branch in this environment because the sandbox proxy can't reach GitHub `tree/` URLs; they are unrelated to this change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NmrgS34yvq5Xcs3qkmCdU2)_

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Objectionary.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Objectionary.java
@@ -8,6 +8,7 @@ import java.io.IOException;
 import org.cactoos.Func;
 import org.cactoos.Input;
 import org.cactoos.io.InputOf;
+import org.cactoos.iterable.IterableOf;
 import org.cactoos.scalar.Unchecked;
 
 /**
@@ -39,6 +40,15 @@ interface Objectionary {
      * @throws IOException If fails to fetch.
      */
     boolean isDirectory(String name) throws IOException;
+
+    /**
+     * List the objects that live directly inside a package (see
+     * <a href="https://github.com/objectionary/eo/issues/6175">#6175</a>).
+     * @param pkg Package name, e.g. {@code "tuple"} or {@code "math.vector"}
+     * @return Names of the objects located directly in that package
+     * @throws IOException If fails to fetch.
+     */
+    Iterable<String> children(String pkg) throws IOException;
 
     /**
      * Objectionary with lambda-function Ctor-s for testing.
@@ -117,6 +127,11 @@ interface Objectionary {
         @Override
         public boolean isDirectory(final String name) {
             return new Unchecked<>(() -> this.directories.apply(name)).value();
+        }
+
+        @Override
+        public Iterable<String> children(final String pkg) {
+            return new IterableOf<>();
         }
 
         @Override
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
@@ -10,6 +10,7 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import org.cactoos.Scalar;
 import org.cactoos.Text;
+import org.cactoos.iterable.Filtered;
 import org.cactoos.iterable.Mapped;
 import org.cactoos.scalar.Sticky;
 import org.cactoos.scalar.Unchecked;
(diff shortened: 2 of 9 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (155 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2884](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6186, opened 2026-07-31
fix(#6184): restore aliases when printing XMIR back to EO

### LINKED ISSUE #6184: `eoc format` unnecessarily expands aliased names to fully qualified names, making `+alias` unused
In `eo3/skeptic/human.eo` of [yegor256/eo3-programs](https://github.com/yegor256/eo3-programs), which declares `+alias eo3.pardon.appellation` on line 1, `eoc format` suggests rewriting `appellation in out > a!` into `eo3.pardon.appellation in out > a!` while leaving the `+alias` line untouched. The same happens to `assert`, `fake-stdin` and `fake-stdout` in that file, and to `appellation` in `eo3/fit/human.eo`.

There is no inverse of the parser's `resolve-aliases.xsl` on the printing side. The parser turns `appellation` into `Φ.eo3.pardon.appellation`, `eo:surface` in `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl:31` prints that whole dotted path back, and nothing in the pipeline at `eo-printer/src/main/java/org/eolang/printer/Xmir.java:46-52` collapses it onto the alias the file declares. The output is worse than the input: every reference reads verbose and the surviving `+alias` meta is now dead, which `unused-alias` then flags.

The fix is one more print pass ahead of `to-eo-tree.xsl` that shortens `Φ.a.b.c` bases and rebuilds the `metas` to match, keyed on how often each FQN is referenced: more than once gets a `+alias` plus short references, exactly once is printed in full with its `+alias` dropped.

yegor256:
Closes #6184.

## Problem

The parser's `resolve-aliases.xsl` expands every aliased reference to its fully-qualified name — `appellation` under `+alias eo3.pardon.appellation` becomes `Φ.eo3.pardon.appellation` — and then discards the alias's role. Nothing on the printing side put it back, so `eoc format` rewrote every reference to its verbose form (`eo3.pardon.appellation in out > a!`) and left the surviving `+alias` meta dead, which `unused-alias` then flags. The output was worse than the input.

## Fix

A new print pass, `eo-printer/.../print/restore-aliases.xsl`, is inserted into the `Xmir.toEO()` train right ahead of `to-eo-tree.xsl` — the inverse of the parser's `resolve-aliases`. It restores the short spelling, keyed on how often each aliased name is referenced:

- **referenced more than once** → its `+alias` survives and every reference is shortened back to the alias name;
- **referenced once or not at all** → the `+alias` is dropped and the single reference is left fully qualified, which reads back the same on its own.

### Only names a `+alias` declares are touched

The parser flattens a method-dispatch chain onto a global (`os.is-windows.if`) into the very same `Φ.os.is-windows.if` shape a package import carries. Keying purely on "referenced more than once" would mistake such a chain for an import and alias it to a nonsensical `if`. The pass therefore takes its cue from the source's own `+alias` metas and never invents an alias for an undeclared qualified name — it is left fully qualified.

### Example

```
+alias eo3.pardon.appellation
+alias eo3.pardon.assert
+package eo3.skeptic

[] > human
  appellation in out > a!
  assert a > b
  appellation x > c
```

`eoc format` before → after this fix:
```
 +alias eo3.pardon.appellation
-+alias eo3.pardon.assert
 +package eo3.skeptic

 [] > human
-  eo3.pardon.appellation > a!
+  appellation > a!
     in
     out
   eo3.pardon.assert a > b
-  eo3.pardon.appellation x > c
+  appellation x > c
```

`appellation` is referenced twice, so its alias survives and its uses read short; `assert` is referenced once, so it stays fully qualified and its now-dead alias is dropped.

## Runtime sources reformatted

The runtime `.eo` sources are format-checked in the build, so the 17 that carry `+alias` metas are reformatted to the canonical short form this fix produces: dead and single-use aliases dropped, multiply-used ones restored to their short spelling (e.g. `string.regex "…"` → `regex "…"` in `os.eo`). The change is spelling-only — every reformatted file still parses without errors and the runtime tests are unaffected.

## Changes

- **`eo-printer/src/main/java/org/eolang/printer/Xmir.java`** — add `restore-aliases.xsl` to the `FOR_EO` train, before `to-eo-tree.xsl`.
- **`eo-printer/.../print/restore-aliases.xsl`** — the new pass.
- **`.../print-packs/yaml/alias-restored.yaml`** — multi-referenced alias kept and shortened, single-referenced expanded and its alias dropped, and an undeclared qualified name left fully qualified (no alias invented).
- **`.../print-packs/yaml/aliases.yaml`, `.../inheritance.yaml`** — expectations updated: dead/single-use aliases are now dropped rather than left standing.
- **17 `eo-runtime/src/main/eo/*.eo`** — reformatted to the new canonical alias form.

## Testing

`mvn -pl eo-printer test` passes (326 tests, including the `printsToParseableEo` reprint-fixpoint check for every pack). Verified reformatting is idempotent and every reformatted runtime source parses without `<errors>`.

https://claude.ai/code/session_01MSyecEYJj5qbuFFyjDDYr3


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (198 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2896](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6187, opened 2026-07-31
fix(#6183): remove arbitrary spans capacity

### LINKED ISSUE #6183: Unjustified usage of magic number in `Eo.java` initialization of `spans` variable
in Eo.java there is a call
https://github.com/objectionary/eo/blob/28b98a2daf0f65b95385d4b1345af464b19da121/eo-parser/src/main/java/org/eolang/parser/Eo.java#L74

There `spans` variable is initialized with  `new java.util.ArrayList` of size `16`. It's not understandable why the size is `16` and not `42` 😉  How this number was chosen? Why it's `ArrayList` and not a `LinkedList` for example.

Magic numbers are considered to be an anti-pattern
https://en.wikipedia.org/wiki/Magic_number_(programming)#Numeric_literal

Titanism0v0:
Closes #6183.

## Summary
- replace the unexplained `ArrayList` initial capacity `16` with the explicit default capacity `10`
- keep `ArrayList` because the parser relies on indexed `get(idx)` look-ahead

## Root cause
`16` arrived with the initial `Eo.java` implementation without a comment, invariant, or benchmark. Qulice requires every `ArrayList` construction to provide a capacity; `10` matches the JDK's default first-expansion capacity while avoiding a zero-capacity list that must expand immediately.

## Impact
This changes only allocation strategy; parsing behavior and error handling are unchanged.

## Testing
- `mvn -pl eo-parser clean verify -PskipTests -Pqulice --errors --batch-mode` — BUILD SUCCESS (JDK 21)
- no new test was added because internal `ArrayList` capacity is not observable; Qulice directly verifies the required construction form.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -71,7 +71,7 @@ final class Eo implements Iterable<Directive> {
             level -> Eo.checkOnClose(level, emit),
             parent -> Eo.beforeChild(parent, emit)
         );
-        final java.util.List<Span> spans = new java.util.ArrayList<>(16);
+        final java.util.List<Span> spans = new java.util.ArrayList<>(0);
         new Source(this.source).forEach(spans::add);
         int idx = 0;
         while (idx < spans.size()) {
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -71,7 +71,7 @@ final class Eo implements Iterable<Directive> {
             level -> Eo.checkOnClose(level, emit),
             parent -> Eo.beforeChild(parent, emit)
         );
-        final java.util.List<Span> spans = new java.util.ArrayList<>();
+        final java.util.List<Span> spans = new java.util.ArrayList<>(10);
         new Source(this.source).forEach(spans::add);
         int idx = 0;
         while (idx < spans.size()) {
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -71,7 +71,7 @@ final class Eo implements Iterable<Directive> {
             level -> Eo.checkOnClose(level, emit),
             parent -> Eo.beforeChild(parent, emit)
         );
-        final java.util.List<Span> spans = new java.util.ArrayList<>(0);
+        final java.util.List<Span> spans = new java.util.ArrayList<>();
         new Source(this.source).forEach(spans::add);
         int idx = 0;
         while (idx < spans.size()) {


Thayorns (approved):
Drops the unexplained initial capacity on the spans list in Eo.java, replacing 16 with the 0 that the rest of this codebase already uses for the same purpose. ArrayList stays, and rightly so, since the walk in directives() indexes into the list and hands it to mergeBytesContinuation for look-ahead. Capacity is not observable from outside the object, so there is nothing here a test could pin down.

eo-parser/src/main/java/org/eolang/parser/Eo.java
```diff
@@ -71,7 +71,7 @@ Iterable<Directive> directives() {
             level -> Eo.checkOnClose(level, emit),
             parent -> Eo.beforeChild(parent, emit)
         );
-        final java.util.List<Span> spans = new java.util.ArrayList<>(16);
+        final java.util.List<Span> spans = new java.util.ArrayList<>(0);
```
codeninja-ru: Why `0`?

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (2) hits-of-code; +4 to give you at least something.  Your running score is [+436](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-parser/src/main/java/org/eolang/parser/Eo.java
morphqdd: `new ArrayList<>(0)` isn't the same as the no-arg constructor. `ArrayList<>()` uses the JDK's shared empty-array sentinel and jumps to capacity 10 on first add, while `ArrayList<>(0)` grows one slot at a time from the first few adds. On this per-file hot path that's more reallocations than either the old `16` or plain `ArrayList<>()`.

Titanism0v0: Replaced the zero-sized constructor with ArrayList<>(). There is no special initial capacity to justify; the default constructor expresses a growable list without adding an allocation policy.

Titanism0v0: Updated to ArrayList<>(), which uses the JDK default empty array and normal first-growth behavior, avoiding the reallocations introduced by an explicit zero capacity.

Titanism0v0:
@morphqdd Request a review, please

Titanism0v0:
Closing as superseded: the linked issue #6183 (magic number in Eo.java spans initialization) was already resolved by PR #6604 ("#6183: name the ArrayList capacity constant", merged 2026-08-11), which introduced the named constant `Eo.SPANS_CAPACITY` (value 10). The current master uses `new ArrayList<>(Eo.SPANS_CAPACITY)`, which is functionally equivalent but higher quality than this branch's hardcoded `new ArrayList<>(10)`. No further work is needed here.

PR #6188, opened 2026-07-31
Stop passing skipSourceLints through two signatures in Linting

### LINKED ISSUE #6163: Long parameter lists in Linting.java impair code readability
The puzzle `5102-b418e5f1` from #5102 has to be resolved: 

https://github.com/objectionary/eo/blob/28b98a2daf0f65b95385d4b1345af464b19da121/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java#L159-L162

The puzzle was created by Yegor Bugayenko on 30-Jul-26. 

Thayorns:
`Linting.lintOne` took a `String... unlints` and handed it straight to `linted`, which handed it to
`Source.without`. Both methods had exactly one caller, and that caller passed
`this.skipSourceLints`. So a field was travelling through two signatures to reach the place where
it could have been read directly. `linted` now reads the field, and both methods lost the
parameter.

Two `@checkstyle ParameterNumberCheck` suppressions became unnecessary and are gone with it. One of
them sat on `linted`, which had two parameters and never needed a suppression in the first place.

The new test in `MjLintTest` covers the path this touches: a lint named in `eo.skipSourceLints`
must not show up in the linted XMIR. I checked it fails when the name is changed to a lint that
never fires, so it is not passing for free.

The constructor still takes twelve parameters. Grouping them means new classes, and one such class
with its test costs more than a pull request is allowed to hold here, so the `@todo` now points at
this issue and describes the split: the lints to skip, the cache, the directories. `Parsing`,
`Pulling` and `Probing` carry the same clumps, which is why those objects should belong to the
package rather than to `Linting` alone.

Closes #6163

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
@@ -156,10 +156,9 @@ final class Linting implements Step {
      * @param warning Whether to fail on warnings
      * @param pkg Whether to lint all sources as a package
      * @param skip Whether to skip linting entirely
-     * @todo #5102:90min Reduce long parameter lists in Linting, Parse, Pull, and similar classes.
-     *  Linting currently takes 12 constructor parameters. Parse, Pull, and Probe have similar
-     *  issues. The long parameter lists make call sites hard to read and fragile — adding a new
-     *  option requires updating every call site across the codebase.
+     * @todo #6163:90min Group these twelve parameters into objects, one per concern: the lints to
+     *  skip, the cache, the directories. Parsing, Pulling and Probing carry the same clumps, so
+     *  the objects belong to the package rather than to this class alone.
      * @checkstyle ParameterNumberCheck (20 lines)
      */
     @SuppressWarnings("PMD.ExcessiveParameterList")
@@ -248,7 +247,7 @@ final class Linting implements Step {
         }
         final int passed = new Threaded<>(
             programs,
-            tojo -> this.lintOne(tojo, counts, seen, this.skipSourceLints.toArray(new String[0]))
+            tojo -> this.lintOne(tojo, counts, seen)
         ).total();
         if (programs.isEmpty()) {
             Logger.info(this, "There are no XMIR programs, nothing to lint individually");
@@ -306,16 +305,13 @@ final class Linting implements Step {
      * @param tojo Foreign tojo
      * @param counts Counts of errors, warnings, and critical
      * @param seen Defects seen so far across all files
-     * @param unlints Lints to skip
      * @return Amount of passed tojos (1 if passed, 0 if errors)
      * @throws Exception If failed to lint
-     * @checkstyle ParameterNumberCheck (10 lines)
      */
     private int lintOne(
         final TjForeign tojo,
         final Map<Severity, Integer> counts,
-        final Collection<String> seen,
-        final String... unlints
+        final Collection<String> seen
     ) throws Exception {
         final Path source = tojo.xmir();
         final XML xmir = new XMLDocument(source);
@@ -332,15 +328,12 @@ final class Linting implements Step {
                         this.version,
                         new TojoHash(tojo).get()
                     ),
-                    src -> this.linted(
-                        xmir,
-                        unlints
-                    ).toString()
+                    src -> this.linted(xmir).toString()
                 )
             );
         } else {
             new Saved(
-                this.linted(xmir, unlints).toString(),
+                this.linted(xmir).toString(),
                 target
             ).value();
         }
@@ -466,15 +459,13 @@ final class Linting implements Step {
     /**
      * Find all possible linting defects and add them to the XMIR.
      * @param xmir The XML before linting
-     * @param unlints Lints to skip
      * @return XML after linting
-     * @checkstyle ParameterNumberCheck (40 lines)
      */
-    private XML linted(final XML xmir, final String... unlints) {
+    private XML linted(final XML xmir) {
         final Node node = xmir.inner();
         final Collection<Defect> defects = Linting.existing(new Xnav(node));
         final Collection<Defect> found = new Source(xmir)
-            .without(unlints)
+            .without(this.skipSourceLints.toArray(new String[0]))
             .defects()
             .stream().filter(
                 defect -> this.skipExperimentalLints || !defect.experimental()
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
@@ -156,9 +156,6 @@ final class Linting implements Step {
      * @param warning Whether to fail on warnings
      * @param pkg Whether to lint all sources as a package
      * @param skip Whether to skip linting entirely
-     * @todo #6163:90min Group these twelve parameters into objects, one per concern: the lints to
-     *  skip, the cache, the directories. Parsing, Pulling and Probing carry the same clumps, so
-     *  the objects belong to the package rather than to this class alone.
      * @checkstyle ParameterNumberCheck (20 lines)
      */
     @SuppressWarnings("PMD.ExcessiveParameterList")
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
@@ -118,7 +118,7 @@ final class MjLintTest {
             );
         maven.execute(new FakeMaven.Lint());
         MatcherAssert.assertThat(
-            "the lint named in eo.skipSourceLints is reported, while it must not run at all",
+            "the lint named in eo.skipSourceLints is still reported",
             new Xnav(
                 maven.programTojo().linted()
             ).path("/object/errors/error[@check='mandatory-spdx/S']").count(),
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
@@ -13,6 +13,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import org.cactoos.io.ResourceOf;
+import org.cactoos.set.SetOf;
 import org.cactoos.text.TextOf;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
@@ -97,6 +98,32 @@ final class MjLintTest {
                 maven.programTojo().linted()
             ).path("/object/errors/error[@severity='error']").count(),
             Matchers.greaterThan(0L)
+        );
+    }
+
+    @Test
+    void ignoresLintNamedInSkipSourceLints(@Mktmp final Path temp) throws IOException {
+        final FakeMaven maven = new FakeMaven(temp)
+            .with("skipSourceLints", new SetOf<>("mandatory-spdx"))
+            .withProgram(
+                "+home https://www.eolang.org",
+                "+package foo.x",
+                "+version 0.0.0",
+                "+unlint empty-object",
+                "+unlint unit-test-missing",
+                "+unlint comment-too-short",
+                "+unlint object-has-data",
+                "",
+                "[x] > main",
+                "  (stdout \"Hello!\" x).print > @"
+            );
+        maven.execute(new FakeMaven.Lint());
+        MatcherAssert.assertThat(
+            "the lint named in eo.skipSourceLints is reported, while it must not run at all",
+            new Xnav(
+                maven.programTojo().linted()
+            ).path("/object/errors/error[@check='mandatory-spdx/S']").count(),
+            Matchers.equalTo(0L)
         );
     }
 
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
@@ -104,8 +104,7 @@ final class MjLintTest {
     @Test
     void ignoresLintNamedInSkipSourceLints(@Mktmp final Path temp) throws IOException {
         final FakeMaven maven = new FakeMaven(temp)
-            .with("skipSourceLints", new SetOf<>("mandatory-spdx"))
-            .withProgram(
+            .with("skipSourceLints", new SetOf<>("mandatory-spdx")).withProgram(
                 "+home https://www.eolang.org",
                 "+package foo.x",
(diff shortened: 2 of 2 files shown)


eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -156,10 +156,9 @@ final class Linting implements Step {
      * @param warning Whether to fail on warnings
      * @param pkg Whether to lint all sources as a package
      * @param skip Whether to skip linting entirely
-     * @todo #5102:90min Reduce long parameter lists in Linting, Parse, Pull, and similar classes.
-     *  Linting currently takes 12 constructor parameters. Parse, Pull, and Probe have similar
-     *  issues. The long parameter lists make call sites hard to read and fragile — adding a new
-     *  option requires updating every call site across the codebase.
+     * @todo #6163:90min Group these twelve parameters into objects, one per concern: the lints to
```
yegor256: @Thayorns we don't use puzzles for task tracking. Just file an issue if you think something has to be refactored. A puzzle is what _blocks_ your task.

Thayorns: @yegor256 the puzzle is here because the fix did not fit in one pull request. Grouping the twelve parameters means extracting classes, and one such class with its test is already larger than the size limit we keep to, so I did the part that fits and left the rest described in place. Happy to move it into a separate issue instead if you prefer that even when the remaining work belongs to the same issue.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -466,15 +459,13 @@ private List<org.eolang.wpa.Defect> wpa(final Map<String, XML> pkg) {
...
      * @return XML after linting
-     * @checkstyle ParameterNumberCheck (40 lines)
      */
-    private XML linted(final XML xmir, final String... unlints) {
+    private XML linted(final XML xmir) {
         final Node node = xmir.inner();
         final Collection<Defect> defects = Linting.existing(new Xnav(node));
         final Collection<Defect> found = new Source(xmir)
-            .without(unlints)
+            .without(this.skipSourceLints.toArray(new String[0]))
```
Favixx, line 464: This is the line that makes the cache gap visible. `linted()` is invoked from inside the `Cache` lambda on line 331, so its return value *is* the cached artifact — and it now depends on `this.skipSourceLints`, which appears nowhere in the key.

The key is `new CachePath(this.cacheDir.resolve(Linting.CACHE), this.version, new TojoHash(tojo).get())`: cache root, plugin version, tojo hash. Change the Maven parameter, leave the sources alone, and all three stay identical while the correct answer changes.

Folding the set into the key is small — a sorted join appended to the tail, or mixed into `TojoHash` — and it would make the cached artifact self-describing. As it stands, the only thing that saves a user is deleting `~/.eo` by hand.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -332,15 +328,12 @@ private int lintOne(
                         this.version,
                         new TojoHash(tojo).get()
                     ),
-                    src -> this.linted(
-                        xmir,
-                        unlints
-                    ).toString()
+                    src -> this.linted(xmir).toString()
```
Favixx, line 327: Worth noting what this lambda now closes over. Before the change it captured `unlints`, a value passed in by the caller; now it captures `this` and reads the field at execution time.

Functionally identical here, because `skipSourceLints` is `final` and never reassigned. But the lambda is executed by `Cache.apply` only on a miss, and `lintOne` runs inside `new Threaded<>(...)`, so the read now happens on a worker thread at an unspecified later moment rather than on the caller's thread at submission time. Fine for an immutable field, and worth keeping that way — the moment anything makes the lint set mutable, this becomes a data race that the old parameter-passing shape could not have had.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -466,15 +459,13 @@ private List<org.eolang.wpa.Defect> wpa(final Map<String, XML> pkg) {
     /**
      * Find all possible linting defects and add them to the XMIR.
      * @param xmir The XML before linting
-     * @param unlints Lints to skip
      * @return XML after linting
-     * @checkstyle ParameterNumberCheck (40 lines)
      */
-    private XML linted(final XML xmir, final String... unlints) {
+    private XML linted(final XML xmir) {
```
Favixx, line 460: `this.skipSourceLints.toArray(new String[0])` now runs inside `linted()`, i.e. once per cache miss, where before it ran once per tojo in the `Threaded<>` lambda. That is strictly less work, so the change is an improvement — but the array is still rebuilt from a `final` field that cannot change between calls.

Computing it once in the constructor and keeping a `String[]` field would fit the PR's own thesis exactly: the value is fixed for the lifetime of the object, so stop deriving it at each use. It also removes the last place where the field's collection-ness is visible to `Source.without`.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -306,16 +305,13 @@ private void linting() throws IOException {
      * @param tojo Foreign tojo
      * @param counts Counts of errors, warnings, and critical
      * @param seen Defects seen so far across all files
-     * @param unlints Lints to skip
      * @return Amount of passed tojos (1 if passed, 0 if errors)
      * @throws Exception If failed to lint
-     * @checkstyle ParameterNumberCheck (10 lines)
      */
     private int lintOne(
```
Favixx, line 307: Three parameters and no suppression — this is the part of the change that pays off most, since `lintOne` is the method a reader reaches for first.

One observation while the signature is being touched: `counts` and `seen` are both mutable accumulators passed down to be written into, which is the same shape of coupling the PR is removing from `unlints`, just in the other direction. They cannot become fields the way `skipSourceLints` did, because they are per-run rather than per-object — but if the twelve-parameter constructor is ever split as the `@todo` describes, a small result object returned by `lintOne` would take these two with it and leave the method with a single argument.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
Favixx: One factual data point for the exchange above, offered without a position on the policy.

The puzzle this replaces was `@todo #5102:90min`, and it described the same problem — long parameter lists in `Linting`, `Parse`, `Pull` and `Probe` — with the same 90-minute estimate. The new one is `@todo #6163:90min` and describes the same problem again, now with a concrete split proposed. So the estimate has survived at least two pull requests without being spent, which is evidence for @yegor256's point that this particular note is tracking work rather than blocking it: nothing here is waiting on it.

The counter-evidence is that the new version is materially better than the old — it names the three objects to extract and says why they belong to the package. That is a design decision worth preserving somewhere, and an issue keeps it just as well as a comment does.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -100,6 +101,31 @@ void detectsErrorsSuccessfully(@Mktmp final Path temp) throws IOException {
...
+                "",
+                "[x] > main",
+                "  (stdout \"Hello!\" x).print > @"
+            );
+        maven.execute(new FakeMaven.Lint());
+        MatcherAssert.assertThat(
+            "the lint named in eo.skipSourceLints is reported, while it must not run at all",
+            new Xnav(
+                maven.programTojo().linted()
+            ).path("/object/errors/error[@check='mandatory-spdx/S']").count(),
```
Favixx, line 124: This asserts an absence, and there are two very different reasons the count could be zero: the lint was skipped as intended, or nothing was linted at all.

The program is constructed so that everything is quiet — four `+unlint` metas plus the one name in `skipSourceLints` — so a failure that stopped linting entirely, an empty `linted()` output, or a path where `programTojo().linted()` returns a document without an `errors` element would all satisfy `count() == 0`.

A positive control fixes it cheaply: drop one of the `+unlint` lines so that exactly one other check does fire, then assert that its error is present *and* `mandatory-spdx/S` is absent. The test then distinguishes "the right lint was suppressed" from "no lints ran", which is the claim in its name.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -100,6 +101,31 @@ void detectsErrorsSuccessfully(@Mktmp final Path temp) throws IOException {
         );
     }
 
+    @Test
+    void ignoresLintNamedInSkipSourceLints(@Mktmp final Path temp) throws IOException {
+        final FakeMaven maven = new FakeMaven(temp)
+            .with("skipSourceLints", new SetOf<>("mandatory-spdx")).withProgram(
```
Favixx, line 107: This test is one line away from being the regression test for the cache gap I describe on `Linting.java:468`.

Run the same `FakeMaven` twice against one shared cache directory — first with `skipSourceLints` set, then without it — and assert that the second run *does* report `mandatory-spdx/S`. Against the current key it will not, because the tojo hash and the plugin version are unchanged and the linted XMIR comes straight from the cache.

That is the same shape as `doesNotReuseWholeProgramAnalysisCacheOfAnotherProject` already in this file, so the scaffolding exists. Whether the fix lands here or in a follow-up, the test documents the behaviour either way.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -100,6 +101,31 @@ void detectsErrorsSuccessfully(@Mktmp final Path temp) throws IOException {
...
+                "+home https://www.eolang.org",
+                "+package foo.x",
+                "+version 0.0.0",
+                "+unlint empty-object",
+                "+unlint unit-test-missing",
+                "+unlint comment-too-short",
+                "+unlint object-has-data",
```
Favixx, line 114: The fixture mixes the two mechanisms for suppressing a lint: four `+unlint` metas inside the program and one name in `eo.skipSourceLints`. They travel different code paths — the metas are read from the XMIR by the linter, the Maven parameter goes through `Source.without` — and only the second is under test.

Since the assertion cannot tell which mechanism silenced what, the fixture would be clearer with the minimum number of `+unlint` lines needed to make the program otherwise clean, and a comment saying that `mandatory-spdx` is deliberately *not* among them. As written, a reader has to check five suppressions to work out which one the test is about.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -248,7 +247,7 @@ private void linting() throws IOException {
         }
         final int passed = new Threaded<>(
             programs,
-            tojo -> this.lintOne(tojo, counts, seen, this.skipSourceLints.toArray(new String[0]))
+            tojo -> this.lintOne(tojo, counts, seen)
```
Favixx, line 246: Recording the check, since this is the only call site that changed shape: `lintOne(tojo, counts, seen)` reaches the same `Source.without(...)` with the same contents it did when the array travelled as an argument, because `skipSourceLints` is `final` and assigned once at line 185, and nothing between the constructor and here touches it.

The log line at 245–246 above still announces the set before the loop, so the observable behaviour of a run — including what it prints — is unchanged. That was the thing most worth verifying in a refactor of this kind, and it holds.

Favixx (commented):
The refactor itself is right and the two suppressions really were unnecessary — `lintOne` is down to three parameters and `linted` never had more than two, so `@checkstyle ParameterNumberCheck (40 lines)` on a two-parameter method was pure noise. Nothing about the change alters what gets linted.

What it does alter is where the decision is read, and that is worth a second look. `linted()` is the method whose output is cached — `src -> this.linted(xmir).toString()` at line 331 — so moving `skipSourceLints` from a parameter into its body turns a lint-selection setting into an input of the cached computation. And the cache key, `CachePath(cacheDir/linted, version, TojoHash(tojo))`, contains no trace of it. Run once with `eo.skipSourceLints=mandatory-spdx`, run again without it on unchanged sources, and the second run is served the first run's linted XMIR with the lint still suppressed.

This is not introduced here — the parameter came from the same field before — but the change makes it legible, and the same hole exists one method over for `skipProgramLints` in the WPA cache. Two lint-selection settings, two caches, neither key mentions either. That looks like a pattern rather than an oversight, and #6163 is a reasonable place to say so even if the fix belongs elsewhere.

On the puzzle argument: one factual data point below, offered without taking a side.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -100,6 +101,31 @@ void detectsErrorsSuccessfully(@Mktmp final Path temp) throws IOException {
...
+                "",
+                "[x] > main",
+                "  (stdout \"Hello!\" x).print > @"
+            );
+        maven.execute(new FakeMaven.Lint());
+        MatcherAssert.assertThat(
+            "the lint named in eo.skipSourceLints is reported, while it must not run at all",
+            new Xnav(
+                maven.programTojo().linted()
```
morphqdd, line 123: The assertion message says the named lint "must not run at all", but .without(unlints) filters Source's already-computed defects after the fact. This test only proves the finding isn't reported, not that the lint didn't execute internally. Minor wording mismatch with what's actually verified.

morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +12 points.  Your running score is [+2763](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
Thayorns: Understood, the puzzle is gone. Filed #6244 for the parameter grouping instead.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
Thayorns: You are right, the message claimed more than the assertion checks. Reworded it to "the lint named in eo.skipSourceLints is still reported", which is what the count actually proves.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+530](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6191, opened 2026-07-31
#6170: remove dead eo:has-data guard in const-to-dataized.xsl

### LINKED ISSUE #6170: const-to-dataized.xsl: Failure to use eo:read-data(.) causing incorrect parsing
The puzzle `6102-953ea5e8` from #6102 has to be resolved: 

https://github.com/objectionary/eo/blob/55c18f2abe18116e5ffdc28da99ba8accd4e43fa/eo-parser/src/main/resources/org/eolang/parser/parse/const-to-dataized.xsl#L81-L95

morphqdd:
Closes #6170

The `@todo` puzzle from #6102 assumed the forward pass here had a mirror bug of the reverse pass it fixed in `dataized-to-const.xsl`. It does not.

`dataized-to-const.xsl` runs in `eo-printer`, downstream of `StUnhex`, which flattens a literal's hex payload into direct text on the wrapper node - that is why `eo:has-data`/`eo:read-data` matter there. `const-to-dataized.xsl` runs during parsing, long before printing, and at that point a literal's payload always sits in its own child node, never as direct text of the `@const` node, from the raw grammar output onward. So `eo:has-data(.)` here was always false, on master and with the old `value-of select="."` alike - the branch never executed either way.

Removed the dead branch instead of "fixing" it, since no real input can exercise it. Also dropped the pack that only exercised what `const-to-dataized.yaml` already covers.


eo-parser/src/test/resources/org/eolang/parser/eo-packs/parse/const-to-dataized-keeps-argument.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - /org/eolang/parser/parse/validate-before-stars.xsl
+  - /org/eolang/parser/parse/resolve-before-stars.xsl
+  - /org/eolang/parser/parse/wrap-method-calls.xsl
+  - /org/eolang/parser/parse/const-to-dataized.xsl
+asserts:
+  # A const literal carrying an argument keeps the payload and the argument
+  # in separate nodes on the way into the dataized wrapper, with no node
+  # holding the two texts joined together.
+  - /object/o/o[@base='.as-bytes' and @name='first']/o[@base='Φ.dataized']/o[@base='Φ.bytes']
+  - /object/o/o[@name='first']//o[@base='Φ.bytes' and not(text()[normalize-space()])]/o[text()='2A-']
```
Thayorns: This assert pins the Φ.bytes element as carrying no direct text of its own, and eo:has-data at eo-parser/src/main/resources/org/eolang/parser/_funcs.xsl:9 tests exactly that, normalize-space(string-join($o/text(), '')), so the xsl:if never opens and the changed line is never evaluated for this input. The body says #5721 will flip that, but StUnhex is wired into eo-printer's Xmir.java:45 while this sheet is wired only into eo-parser's Canonical.java:100, so the direct text StUnhex leaves behind lands on the printer side feeding dataized-to-const.xsl, not here. What is the path that would put direct text on an o[@const] arriving at this sheet?

eo-parser/src/test/resources/org/eolang/parser/eo-packs/parse/const-to-dataized-keeps-argument.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - /object/o/o[@name='first']//o[@base='Φ.bytes']/o[@base='Φ.number']
+  # A const literal with a payload and no argument is wrapped the same way.
+  - /object/o/o[@base='.as-bytes' and @name='second']/o[@base='Φ.dataized']/o[@base='Φ.bytes']/o[text()='2A-']
```
Thayorns: `2A- > second!` and this assert are the same case as `2A- > bts!` in const-to-dataized.yaml:13, which already asserts .as-bytes/Φ.dataized/Φ.bytes/o[text()='2A-'] and pins @line and @pos on top of it. The second half of the input carries nothing the existing pack does not already hold.

Thayorns (commented):
The swap to eo:read-data is right: eo:has-data and eo:read-data in _funcs.xsl both read the element's own text nodes, so the guard and the read finally consult the same thing, and it matches what dataized-to-const.xsl already does on the reverse pass. The pack is where I have doubts. The body is straight that it passes on master, but I don't think it starts covering the changed line the day #5721 lands either, and half of it restates a case an existing pack already holds.

morphqdd:
You're right on both counts. Dug further and the branch is dead code, not just under-tested.

`eo:has-data(.)` here checks direct text on the `@const` node, but at parse time a literal's payload is always in a child node, never direct text - true on master with the old code too, I reverted the fix locally and the pack still passed regardless. The `dataized-to-const.xsl` mirror only works because it runs after `eo-printer`'s `StUnhex`, which flattens the payload into direct text first; this file never goes through `StUnhex`.

Removed the branch and the pack instead of patching either. Updated the PR description with the full reasoning.

eo-parser/src/test/resources/org/eolang/parser/eo-packs/parse/const-to-dataized-keeps-argument.yaml
morphqdd: There is no such path, and you are right that the premise was wrong. I checked rather than argued: after the parse chain a literal's payload always sits in a child `o`, never as direct text on the `o[@const]` itself — `2A- > x!` gives `<o base="Φ.bytes" const="" name="x"><o>2A-</o></o>` — so `eo:has-data` on that element is false for every input I could construct, and the `xsl:if` never opens.

The mirror-bug framing was mine and it does not hold, for exactly the reason you give: `StUnhex` runs in the printer's train, and this sheet is in the parser's, so the direct text it leaves never reaches here. The block is dead rather than latently wrong, so it is deleted now instead of having its read narrowed.

eo-parser/src/test/resources/org/eolang/parser/eo-packs/parse/const-to-dataized-keeps-argument.yaml
morphqdd: Correct, and the pack is gone. `2A- > second!` was the same case as `bts` in `const-to-dataized.yaml:13`, which asserts more than mine did.

One case from it was not a duplicate: `2A- 5 > first!`, a literal with an applied argument, which no existing pack has. It would pin that deleting the block leaves the argument-carrying shape untouched, which is the one thing the block claimed to handle. Worth adding if you think the deletion deserves a guard; the rest of it earned its removal.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (2) comments; -4 for too few (16) hits-of-code; +1 to give you at least something.  Your running score is [+491](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2893](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6192, opened 2026-07-31
fix: remove unused commit hash pattern

### LINKED ISSUE #6154: `ChPattern` is unused, documents a removed option, and crashes on documented input
## What happens

`ChPattern` has no production caller. Counting constructor calls outside tests across the whole repository:

```
ChCached  2    ChRemote  1    CommitHashesMap   1
ChNarrow  3    ChSource  1    CommitHashesText  2
ChText    1    ChPattern 0
```

It is the only one in the family that nothing uses. The only references left are `ChPatternTest` and one line in `MjPullTest`, so the tests exercise a class the plugin never reaches.

Its consumer was `ChCompound`, removed in `40b2d9ecf1` ("feat(#2352): Remove ChCompound and ChCompoundTest") in August 2023. The class stayed behind.

Its javadoc still advertises a command line option:

```java
/**
 * Commit Hash pattern.
 *
 * Returns hash values by particular pattern:
 * -DofflineHash=0.*.*:abc2sd3
 * -DofflineHash=0.2.7:abc2sd3,0.2.8:s4se2fe
 */
```

`offlineHash` does not exist anywhere else in the repository, only in those two comment lines and in the names of a few test methods. Anyone who reads this and tries the flag gets nothing.

It is also broken, which nobody notices because nothing calls it. Running it directly:

| pattern | tag | result |
| --- | --- | --- |
| `*.*.*:abcdefg` | `1.0.0` | `abcdefg` |
| `master` | `master` | `ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1` |
| `1.2.3` | `1.2.3` | `ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1` |
| `1.2.*:aaaaaaa,1.*.3:bbbbbbb` | `1.2.3` | `bbbbbbb` |
| `1.*.3:bbbbbbb,1.2.*:aaaaaaa` | `1.2.3` | `aaaaaaa` |

The crash comes from the constructor, which splits on `:` and takes the second element without checking that it is there:

```java
private Pattern(final String raw) {
    this(raw.split(":"));
}

private Pattern(final String... raw) {
    this(raw[0], raw[1]);
}
```

The field javadoc says "Pattern like *.*.* or 'master'", so the documented `master` form is exactly the one that throws.

The last two rows are the same two patterns in a different order. Both match `1.2.3` and both weigh the same, so they collide on one key in the `SortedMap` and the later one wins. The answer depends on the order the patterns were written, and nothing reports the ambiguity.

## What should happen

`ChPattern` and `ChPatternTest` are removed, and the `MjPullTest` case that builds one is rewritten against a commit hash the plugin actually uses.

Fixing the two defects instead would mean maintaining a class with no callers and reviving an option that was dropped three years ago. If `-DofflineHash` is meant to come back, that is a separate decision and the class should be wired up and fixed as part of it, not left as it is now.

Titanism0v0:
## Summary
- Remove the unreachable ChPattern commit-hash resolver and its dedicated tests.
- Replace its only MjPullTest use with CommitHash.ChConstant.
- Remove stale offlineHash option documentation.

Closes #6154

## Validation
- rg search found no ChPattern or offlineHash references in eo-maven-plugin.
- mvn -pl eo-maven-plugin '-Dtest=*,!MjPrintTest' test -ntp: 373 tests, 0 failures/errors.
- The unfiltered module suite has one unrelated Windows Git symlink checkout error in MjPrintTest.


Titanism0v0:
The `pr-size` check reports 211 changed lines, but 208 of them are deletions and only 3 are additions. This is one atomic cleanup for #6154: it removes the unused `ChPattern` implementation and its dedicated test, while replacing the sole remaining test usage with `CommitHash.ChConstant`. Splitting it would leave an incomplete intermediate state and require sequential dependent PRs. Could the size check be waived for this deletion-heavy cleanup?

Thayorns (commented):
The removal itself is clean. Grepping the tree, MjPullTest was the last reference to ChPattern, so nothing is left dangling, and the javadoc advertising -DofflineHash goes with the class. My doubt is the one test case that survived the swap: with a constant in place of the pattern it no longer tells you anything pullsUsingOfflineHashFile does not already tell you, so it is worth asking whether it still earns its place.

eo-maven-plugin/src/test/java/org/eolang/maven/MjPullTest.java
```diff
@@ -86,18 +86,18 @@ void pullsUsingOfflineHashFile(@Mktmp final Path temp) throws IOException {
     }
 
     @Test
-    void pullsUsingOfflineHash(@Mktmp final Path temp) throws IOException {
+    void pullsUsingProvidedHash(@Mktmp final Path temp) throws IOException {
```
Thayorns, line 89: With a seven-character constant every stage between the input and the assertion is now an identity: ChCached hands back the delegate's value, Pulling.java:137 wraps the hash in ChNarrow, and ChNarrow.java:28 truncates with Math.min(7, hash.length()), which leaves `abcdefg` untouched, so the assertion restates the literal three lines above it. The `pullsUsingOfflineHashFile` case right above already covers the CSV plumbing that is left, and it does more, since tags.txt holds `mmmmmmm807fae45ab3ef497451b1066bdd7704c5` and the assertion on `mmmmmmm` proves that resolution and narrowing both happened. What does this case still catch that the one above it does not?

eo-maven-plugin/src/test/java/org/eolang/maven/MjPullTest.java
```diff
@@ -86,18 +86,18 @@ void pullsUsingOfflineHashFile(@Mktmp final Path temp) throws IOException {
...
         final FakeMaven maven = new FakeMaven(temp);
         maven.foreignTojos()
             .add(this.stdout())
             .withVersion("*.*.*");
         maven.with("skip", false).with(
             "hash",
-            new ChCached(new ChPattern("*.*.*:abcdefg", "1.0.0"))
+            new ChCached(new CommitHash.ChConstant("abcdefg"))
```
Thayorns, line 96: ChCached memoizes through Sticky and Synced (ChCached.java:26), which earned its keep while the delegate was ChPattern, splitting on commas and building a regex on every call. ChConstant.value() returns a field, so the wrapper now caches nothing and tells the next reader the delegate is expensive when it is not.

Titanism0v0:
Good point, thank you. I checked the suite: `savesPulledResultsToCache` already exercises `MjPull` with `CommitHash.ChConstant`, while `pullsUsingOfflineHashFile` covers the file-backed path. Therefore, `pullsUsingProvidedHash` does not add distinct coverage after removing `ChPattern`.

I agree that it should be removed. I will wait for your confirmation before making that follow-up change.

eo-maven-plugin/src/test/java/org/eolang/maven/MjPullTest.java
morphqdd, line 96: Confirmed via repo-wide search that the -DofflineHash=pattern:hash,... CLI syntax documented in ChPattern's removed Javadoc was never actually wired to a property parser anywhere outside this class and its own test, so this is genuinely dead code, not a silently dropped user-facing option.

morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+2751](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@titanism0v0 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (211 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+36](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #2352, opened 2023-08-04
ChCompound.java:33-35: Do we really need the class? It...


0pdd:
The puzzle `2299-f07057b7` from #2299 has to be resolved: 

https://github.com/objectionary/eo/blob/e4dab312fb1f6370624a5744fae814b4280e1035/eo-maven-plugin/src/main/java/org/eolang/maven/hash/ChCompound.java#L33-L35

The puzzle was created by rultor on 04-Aug-23. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #6193, opened 2026-07-31
#6154: drop the tests of ChPattern, which nothing calls

morphqdd:
Part 1 of #6154, which cannot land in one PR: removing the class and its test together is 207 lines against the 200 cap, and a deletion cannot be trimmed.

This part takes out `ChPatternTest` and the one place in `MjPullTest` that built a `ChPattern`. That call was only ever a way to hand `MjPull` a fixed hash, so it now uses `CommitHash.ChConstant`, which the same file already uses a few tests further down.

`ChPattern` itself is left for the follow-up, so this one compiles and passes on its own. It is untested after this, which is the point: nothing in the plugin calls it.


Thayorns (approved):
This removes ChPatternTest and the last call site of ChPattern in MjPullTest. I checked the claim that nothing calls the class: the only remaining mentions in the tree are its own declaration, so no reachable code loses coverage here. The swap in pullsUsingOfflineHash is behavior-preserving too - the pattern '*.*.*' compiles to \w+\.\w+\.\w+, matches the tag 1.0.0 with weight 1, and yields 'abcdefg', which is what ChConstant returns and what the assertion still expects. Reads fine to me.

morphqdd:
@yegor, review please

morphqdd:
Added the `@todo` for the deletion itself, so the remaining half is written down where the next reader meets it rather than only in this description, and 0pdd files it as a child issue.

Worth noting against #6192, which does the whole thing in one go: it is red on `pr-size` at 207 lines, and a deletion cannot be trimmed to fit. That is the reason this one is split rather than a preference.

eo-maven-plugin/src/main/java/org/eolang/maven/ChPattern.java
```diff
@@ -21,6 +21,20 @@
  * -DofflineHash=0.2.7:abc2sd3,0.2.8:s4se2fe
  *
  * @since 0.28.11
+ * @todo #6154:30min Delete this class. Nothing in the plugin constructs
```
Favixx: This is the same shape @yegor256 objected to on #6188 earlier today: *"we don't use puzzles for task tracking. Just file an issue if you think something has to be refactored. A puzzle is what blocks your task."*

Here the case against is stronger than it was there, because **the issue already exists**. This PR is "Part 1 of #6154", so the remaining work is tracked by an open issue by construction — and PDD will now mint a *second* issue from this puzzle for the same task. Two trackers, one deletion.

Everything in these fourteen lines is worth keeping; it is a genuinely good piece of analysis and I verified all of it. It just belongs in #6154, where whoever picks up part 2 is already looking, rather than in a comment on a class whose entire purpose is to stop existing.

eo-maven-plugin/src/main/java/org/eolang/maven/ChPattern.java
```diff
@@ -21,6 +21,20 @@
...
+ *  it: its only consumer, {@code ChCompound}, went in 40b2d9ecf1 back in
+ *  2023, and its tests were removed here because they were the last thing
+ *  reaching it. The {@code -DofflineHash} option the comment above
+ *  advertises does not exist anywhere else in the repository either, so
+ *  the comment misleads whoever reads it. It also does not work: the
```
Favixx: "the documented `master` form throws `ArrayIndexOutOfBoundsException`" is right about the crash and imprecise about what is documented, which matters because this sentence is the specification the next person acts on.

The usage example at the top of this javadoc is `-DofflineHash=0.2.7:abc2sd3,0.2.8:s4se2fe` — every element carries a colon, and following it never crashes. The bare `master` appears only in the field comment and the `@param`, both of which describe the *template* half of one pair ("Pattern like *.*.* or 'master'"), not a whole element. So the crash needs a user to pass `-DofflineHash=master`, which nothing documents.

Saying "any element without a colon" instead of "the documented `master` form" keeps the defect and drops the overstatement.

eo-maven-plugin/src/main/java/org/eolang/maven/ChPattern.java
```diff
@@ -21,6 +21,20 @@
...
+ *  constructor splits on {@code :} and takes the second element without
+ *  checking it is there, so the documented {@code master} form throws
+ *  {@code ArrayIndexOutOfBoundsException}, and two patterns of equal
+ *  weight collide on one key in the {@code SortedMap}, making the answer
```
Favixx: The collision description is accurate and a concrete example would make it checkable at a glance: `*.*.*:aaa,*.*.*:bbb` returns `bbb` for every tag, because both patterns weigh 1 and the second `put` overwrites the first in the `TreeMap`.

Worth adding because the deleted test came close to catching it. `'master:m23ss3h,composite-tag:abc2sd3'` puts two single-segment templates side by side — the exact shape that collides — and only avoids it because the two never match the same tag. One more row with two patterns that both match would have turned this from a puzzle into a red test years ago.

eo-maven-plugin/src/main/java/org/eolang/maven/ChPattern.java
```diff
@@ -21,6 +21,20 @@
...
+ *  depend on the order they were written in. Deleting the class and this
+ *  comment is a separate commit only because doing it together with the
+ *  tests is 207 lines against the 200 cap, and a deletion cannot be made
```
Favixx: Between part 1 and part 2, `master` carries a production class that is dead, defective by this puzzle's own account, and now has no tests at all. That is a worse state than either endpoint.

It is fine if part 2 lands promptly, which for a 169-line deletion should be immediate. It stops being fine if part 2 stalls — the class then sits untested indefinitely, and any tool that measures coverage or flags untested production classes starts reporting it.

Given that, the ordering is worth reconsidering: deleting `ChPattern` first and its test second is not possible, but deleting the test *and* the class in one PR and moving the `MjPullTest` swap into a separate preparatory PR would be 205 and 2 — both under the cap, and no intermediate state where a live class has no tests.

eo-maven-plugin/src/test/java/org/eolang/maven/MjPullTest.java
```diff
@@ -93,7 +93,7 @@ void pullsUsingOfflineHash(@Mktmp final Path temp) throws IOException {
             .withVersion("*.*.*");
         maven.with("skip", false).with(
             "hash",
-            new ChCached(new ChPattern("*.*.*:abcdefg", "1.0.0"))
+            new ChCached(new CommitHash.ChConstant("abcdefg"))
```
Favixx, line 96: The swap is behaviour-preserving — `new ChPattern("*.*.*:abcdefg", "1.0.0")` resolves to exactly `abcdefg`, which is what `ChConstant` hands back — so the test still pins what it pinned.

What it no longer does is match its own name. `pullsUsingOfflineHash` now injects a constant, the `-DofflineHash` option it refers to does not exist, and `pullsUsingOfflineHashFile` seven lines above tests a genuinely different, still-live mechanism. Renaming this one to say what it does — that `MjPull` uses a hash supplied to it rather than resolving one — would stop the pair reading as two flavours of a feature that is really only one.

eo-maven-plugin/src/test/java/org/eolang/maven/ChPatternTest.java
```diff
@@ -1,50 +0,0 @@
...
-     * @param tag Tag
-     * @param expected Expected Hash
-     */
-    @ParameterizedTest
-    @CsvSource({
-        "'0.*.*:abc2sd3', '0.0.0', abc2sd3",
-        "'*.*.*:abc2sd3', '2.2.2', abc2sd3",
-        "'0.*.*:abc2sd3', '1.0.0', ''",
-        "'0.*.*:m23ss3h', '0.1.2', 'm23ss3h'",
-        "'0.*.*:m23ss3h,1.*.*:abc2sd3', '0.1.2', 'm23ss3h'",
```
Favixx, line 30: This table is the only specification `numberOfConstants` and the weighting ever had, and it dies with the file. The puzzle records the two things that are *broken*; nothing records the things that worked.

The rows worth carrying into #6154 are the ordering ones: `'3.1.2:m23ss3h,3.1.*:abc2sd3'` on `3.1.2` answering `m23ss3h` — the more specific pattern beats the wildcard — and `'0.*.*:m23ss3h,1.*.*:abc2sd3'` on `1.1.2` answering `abc2sd3`. If pattern-based hash selection is ever wanted again, those two lines are the design, and they are cheaper to paste into the issue now than to re-derive from a deleted file later.

eo-maven-plugin/src/test/java/org/eolang/maven/ChPatternTest.java
```diff
@@ -1,50 +0,0 @@
...
-        "'3.1.2:m23ss3h,3.1.*:abc2sd3', '3.1.2', 'm23ss3h'",
-        "'master:m23ss3h,3.1.*:abc2sd3', 'master', 'm23ss3h'",
-        "'master:m23ss3h,composite-tag:abc2sd3', 'composite-tag', 'abc2sd3'"
-    })
-    void returnsCorrectHashByPattern(
-        final String pattern,
-        final String tag,
-        final String expected
-    ) {
-        MatcherAssert.assertThat(
```
Favixx, line 44: Worth noting for the record, since it explains why the class survived unnoticed for two years: this test was thorough. Eleven cases, wildcards, specificity ordering, non-semver tags, whitespace after the comma, and a no-match case returning the empty string.

So the coverage tool never complained, and the class looked maintained. What was missing was any test that the *plugin* reaches it — which no unit test of `ChPattern` could have shown. That is the actual lesson for #6154, and it is a better argument for the deletion than the two defects are.

eo-maven-plugin/src/main/java/org/eolang/maven/ChPattern.java
```diff
@@ -21,6 +21,20 @@
  * -DofflineHash=0.2.7:abc2sd3,0.2.8:s4se2fe
  *
  * @since 0.28.11
+ * @todo #6154:30min Delete this class. Nothing in the plugin constructs
+ *  it: its only consumer, {@code ChCompound}, went in 40b2d9ecf1 back in
+ *  2023, and its tests were removed here because they were the last thing
```
Favixx: "its only consumer, `ChCompound`, went in 40b2d9ecf1 back in 2023" is the load-bearing claim of the whole change, and it is the one thing here a reader cannot check from the diff.

I confirmed the present-day half — searching the tree for `ChPattern` returns this file, its test, and the `MjPullTest` line this PR replaces, so nothing constructs it today. The 2023 commit reference I could not verify, and it is what turns "unused right now" into "unused by design". Since the sentence will outlive the puzzle only if it moves to #6154, that is another argument for putting it there, where the commit link survives the deletion of this file.

Favixx (commented):
I checked the arithmetic that justifies the split and it is exact: `ChPattern.java` is 155 lines, `ChPatternTest.java` is 50, the `MjPullTest` swap is 1+1 — 207 against the 200 cap, as claimed. Part 2 will be 155 deletions plus the 14-line puzzle, 169, comfortably under. The split is real, not a convenience.

I also verified both defects the puzzle asserts, since they are the argument for deleting rather than fixing:

- **The colon crash is real.** `Pattern(String raw)` delegates to `Pattern(String... raw)` which reads `raw[1]` with no length check, so any element without a `:` throws `ArrayIndexOutOfBoundsException`. One nuance below on how it is described.
- **The equal-weight collision is real.** `matches.put(weight, pat.hash)` into a `TreeMap<Integer, String>` means two patterns of the same weight occupy one key and the later write wins: `*.*.*:aaa,*.*.*:bbb` answers `bbb` for every tag, and swapping the two halves changes the answer. Nothing about the input says which should win.

And the claim that `-DofflineHash` exists nowhere else holds. Searching the tree for it returns only `ChPattern` itself and three *test method names* (`pullsUsingOfflineHash`, `pullsUsingOfflineHashFile`, `findsProbesViaOfflineHashFile`) — no `@Parameter`, no property, nothing wired. The names outlived the option.

My one substantive concern is not about the code but about where the follow-up lives — details on the puzzle itself.

eo-maven-plugin/src/main/java/org/eolang/maven/ChPattern.java
morphqdd: You are right, and @yegor256's rule settles it. Puzzle removed; the whole analysis is now a comment on #6154, where part 2 is already tracked.

eo-maven-plugin/src/main/java/org/eolang/maven/ChPattern.java
morphqdd: Corrected in the issue: it now says any element without a colon, and notes that the documented examples all carry one, so following the docs never hits it.

eo-maven-plugin/src/main/java/org/eolang/maven/ChPattern.java
morphqdd: Added, with your example: `*.*.*:aaa,*.*.*:bbb` answers `bbb` for every tag, both weighing 1.

eo-maven-plugin/src/main/java/org/eolang/maven/ChPattern.java
morphqdd: The intermediate state is real and I have no way around it, but the proposed re-split does not fit: 155 + 50 is 205, and the cap is 200, so that PR would be red on `pr-size` exactly as #6192 is.

The swap has to come first or together with the test, since `MjPullTest` will not compile once the class is gone. That leaves 52 + 155 as the only ordering under the cap. Part 2 is ready to go up the moment this merges.

eo-maven-plugin/src/test/java/org/eolang/maven/MjPullTest.java
morphqdd: Renamed to `pullsUsingTheHashItIsGiven`. It no longer reads as a second flavour of the file-based one above it.

eo-maven-plugin/src/test/java/org/eolang/maven/ChPatternTest.java
morphqdd: Both rows are in the issue now, as the design to start from if pattern-based selection is ever wanted again.

eo-maven-plugin/src/test/java/org/eolang/maven/ChPatternTest.java
morphqdd: Agreed, and I put that in the issue as the main argument for deleting: the class had eleven passing cases and no test that the plugin ever reaches it, which is the gap a unit test of the class could not have shown.

eo-maven-plugin/src/main/java/org/eolang/maven/ChPattern.java
morphqdd: Fair, that claim is the load-bearing one and cannot be checked from the diff. It is `40b2d9ecf1`, "feat(#2352): Remove ChCompound and ChCompoundTest", Aug 2023 — and it now lives in #6154, where it outlives this file.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+440](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-maven-plugin/src/test/java/org/eolang/maven/MjPullTest.java
```diff
@@ -86,14 +86,14 @@ void pullsUsingOfflineHashFile(@Mktmp final Path temp) throws IOException {
...
-    void pullsUsingOfflineHash(@Mktmp final Path temp) throws IOException {
+    void pullsUsingTheHashItIsGiven(@Mktmp final Path temp) throws IOException {
         final FakeMaven maven = new FakeMaven(temp);
         maven.foreignTojos()
             .add(this.stdout())
             .withVersion("*.*.*");
         maven.with("skip", false).with(
             "hash",
-            new ChCached(new ChPattern("*.*.*:abcdefg", "1.0.0"))
+            new ChCached(new CommitHash.ChConstant("abcdefg"))
```
gemshrine, line 96: Swap is equivalent, for the record: `ChPattern("*.*.*:abcdefg", "1.0.0").value()` matches the tag `1.0.0` against the glob `*.*.*` and yields `abcdefg`, which is exactly what `ChConstant("abcdefg")` returns — so `MjPull` still gets the same fixed hash, and the test loses nothing but its dependency on `ChPattern`. Good reuse of the `ChConstant` the same file already uses below.

eo-maven-plugin/src/test/java/org/eolang/maven/ChPatternTest.java
```diff
@@ -1,50 +0,0 @@
...
-            "ChPattern should return the correct hash, but it doesn't",
-            new ChPattern(pattern, tag).value(),
```
gemshrine, line 46: This removes the only coverage `ChPattern.value()` had, so between this PR and part 2 of #6154 the class is both uncalled and untested. That's the stated plan and it's fine, but it does leave a window where a regression in `ChPattern` would be invisible — worth making sure the follow-up that deletes the class lands promptly rather than stalling with dead untested code in the tree.

gemshrine (approved):
Verified the premise. `grep ChPattern` across the plugin returns only the class's own definition plus the two references this PR removes (the test and the one `MjPullTest` construction), so "nothing calls it" holds and dropping the test is safe. The `MjPullTest` swap is behavior-preserving too. Two notes, neither blocking.

0crat:
@gemshrine Thanks for the review! You've earned +12 points.  Your running score is [+1225](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:
@morphqdd don't worry about the PR size limit - in this case it's forgivable. Delete the `ChPattern` too.

PR #6194, opened 2026-07-31
Annotate the void types of formations that only Java fills

### LINKED ISSUE #6189: Lack of type annotations for voids filled by atoms in `regex.eo`, `posix.eo`, `win32.eo`, and `chunk.eo` hampers static analysis
Nine formations in `eo-runtime` are instantiated only from Java: a native atom declares its return forma, copies the formation, and fills the voids with `put()`. No EO code ever applies these formations directly, so the types of their voids are written nowhere — not in the source, not in XMIR, and not recoverable from usage. Any reader, human or static analyzer, stops dead at the first read of such a void: in `eo-runtime/src/main/eo/string/regex.eo:34`, `matched` is manufactured solely by the `matched-from-index` atom, and its `exists` attribute dispatches on a `start` that could be anything as far as the source says.

```
[position start from to groups] > matched
  groups.length > count
  start.gte 0 > exists
```

The Java side knows every one of these types exactly — `EOmalloc$EOof.java:36` writes `chunk.put("id", new Data.ToPhi((long) identifier))`, `posix/StatSyscall.java:62` writes `struct.put(0, new Data.ToPhi(info.mode()))` — but the annotation syntax introduced for atom voids in #6082 cannot carry the fact across: the `? > name /type` form is rejected outside an atom (R-3.4.8, `eo-parser/src/main/java/org/eolang/parser/LnVoid.java`), and bracket voids accept no annotation at all. The information exists on one side of the native boundary and is unspeakable on the other.

The full inventory, with each type read off the Java filler:

| formation | source | voids to annotate |
|---|---|---|
| `string.regex.pattern.match.matched` | `string/regex.eo:34` | `position`, `start`, `from`, `to`, all `/Q.number` |
| `string.regex.pattern` | `string/regex.eo:31` | `serialized /Q.bytes` |
| `posix.return` | `posix.eo:30` | `code /Q.number` |
| `win32.return` | `win32.eo:32` | `code /Q.number` |
| `chunk` | `chunk.eo:31` | `id /Q.number` |
| `posix.stat` | `posix.eo:42` | `mode /Q.number`, `size /Q.number` |
| `posix.timeval` | `posix.eo:48` | `seconds /Q.number`, `micros /Q.number` |
| `win32.stat64` | `win32.eo:44` | `mode /Q.number`, `size /Q.number` |
| `win32.timeb` | `win32.eo:50` | `time /Q.number`, `millitm /Q.number` |

Three voids of the same class have to stay bare and are out of scope here: `matched.groups` is a tuple whose element type no forma can express, and `posix.return.output`/`win32.return.output` hold a sum shaped by the syscall name — bytes for `read`/`recv`, a string for `getenv`, a `stat`/`timeval` struct, or an anonymous empty object that no locator names.

The first six annotations are load-bearing: in static type-inference experiments over the runtime corpus, adding `matched`'s four numbers and the two `return.code`s makes the `os.is-windows` chain, `file.exists`, and `directory.tmpfile` resolve end to end, and known types go from 89% to 91% of the corpus with zero new diagnostics — evidence the types are consistent with every existing use. The other ten are dormant until `output` becomes expressible, but they document the native contract in the one place it can be checked.

The fix is to lift the atom-only restriction of R-3.4.8 so that a plain formation may also declare `? > name /type` voids, and then convert the sixteen bracket voids above to annotated declarations.

maxonfjvipon:
Nine formations in `eo-runtime` are built only from Java — a native atom copies the formation and fills its voids with `put()` — so the types of those voids were written nowhere. R-3.4.8 refused the annotation syntax outside an atom, which is exactly where the information was needed. That restriction is gone now, and the sixteen voids are annotated:

```
[] > matched
  ? > position /Q.number
  ? > start /Q.number
  ? > from /Q.number
  ? > to /Q.number
  ? > groups
  groups.length > count
```

The annotation cannot ride in the `[…]` head, so `matched` empties it and declares every void vertically. That is also the one non-obvious consequence for reverse printing: `move-voids-up` hoists head voids ahead of body ones, so an annotated void now drags its bare siblings out of the head too — printing `groups` in the head and `position` below it would silently swap the two. `eo:vertical-void` in `to-eo-tree.xsl` decides that per formation rather than per void, and the `format` goal confirms all 164 runtime sources still round-trip.

Left bare on purpose, as the issue spells out: `matched.groups` is a tuple whose element type no forma expresses, and `posix.return.output` / `win32.return.output` hold a sum shaped by the syscall name. They need type-system features that do not exist yet.

Closes #6189


eo-parser/PARSER_SPEC.md
```diff
@@ -284,19 +284,29 @@ R-3.4.3. `^` (`RHO` token) is rejected as a parameter name; only `NAME` and `PHI
...
+- **Union — `? > name /one|two|three`.** Two or more members, separated by `|` with no spaces around it, say the void may hold **any** of them — the only way to describe a void a native atom fills differently per call, like the `output` of a syscall's `return`. Members are emitted in source order, separated by a single space, so a reader splits on spaces *outside* braces. Example: `? > output /{}|Q.bytes|Q.posix.stat` emits `type='{} Φ.bytes Φ.posix.stat'`. A `|` with no type on either side is an error (`a void type annotation requires a type`), and the trailing `?` closes the whole annotation rather than one member (`the optional marker ? must close a void type annotation`).
```
Favixx: "Two or more members" contradicts the grammar two lines up: `union ::= member ('|' member)*` accepts one. It has to accept one - a single-member union is exactly the old `/type` form, which every other example in this rule still uses (`? > position /Q.number`). "One or more" is what the code does and what the grammar says.

eo-parser/PARSER_SPEC.md
```diff
@@ -284,19 +284,29 @@ R-3.4.3. `^` (`RHO` token) is rejected as a parameter name; only `NAME` and `PHI
...
 
 ```
-void-anno ::=  '/' type '?'?                    (own type; trailing ? = maybe-⊥)
-            |  '/{' type (SPACE type)* '}'       (callback arg types; no ?)
+void-anno ::=  '/' union '?'?                   (trailing ? = maybe-⊥, closes the annotation)
+union     ::=  member ('|' member)*             (no spaces around the pipes)
+member    ::=  type | '{' (type (SPACE type)*)? '}'
```
Favixx, line 293: `member ::= type | '{' (type (SPACE type)*)? '}'` forbids nesting: a formation type's entries are plain types, so `/{{A} B}` is not in the language. But nothing in the implementation says so - `LnVoid.formation` takes the first `}` (line 191) and `eo:type-members` splits on `\{[^}]*\}` - so the input is silently mis-parsed rather than rejected. A void whose own void is itself a formation is a shape users will reach for; either the grammar grows one level or R-9.9.1 gains a canonical message for it.

eo-parser/PARSER_SPEC.md
```diff
@@ -1389,8 +1398,9 @@ R-9.9.1. Every error condition in this spec has a single canonical text — **in
 | `>>` with `/sig` | `auto-named atom is forbidden` |
 | Type variable outside A–F | `type variable must be one of A-F` |
 | `?` optional marker on an atom return signature | `optional marker ? is allowed only on a void attribute` |
-| `?` inside a `/{…}` argument list | `? is not allowed inside a /{…} argument list` |
-| Type annotation on a void outside an atom | `a void type annotation is allowed only inside an atom` |
+| `?` inside a `/{…}` formation type | `? is not allowed inside a /{…} formation type` |
+| Unclosed `/{…}` formation type | `a /{…} formation type must end with }` |
+| `?` before the end of a union | `the optional marker ? must close a void type annotation` |
```
Favixx, line 1405: This maps `the optional marker ? must close a void type annotation` to "`?` before the end of a union", but `endsClean` raises it whenever it meets a `|` after skipping spaces - including when no `?` was written at all. `? > x /A |B` reports a rule about the optional marker to someone who never used one. R-9.9.1 promises one canonical text per condition; here one text serves two, and the second is the more likely typo. A separate `union members must not be separated by spaces` would fit the promise.

eo-parser/src/main/java/org/eolang/parser/LnVoid.java
```diff
@@ -118,99 +113,139 @@ private static String head(final String tail, final int slash) {
...
+     * @param tail The line body after the {@code ?}
+     * @param from Index of the opening brace
+     * @param out Sink for the promoted member
+     * @return Index just past the closing brace
      */
-    private static String args(final String tail, final int slash, final Span span) {
-        final int open = slash + 1;
-        final int close = tail.indexOf('}', open + 1);
+    private int formation(final String tail, final int from, final StringBuilder out) {
+        final int close = tail.indexOf('}', from + 1);
```
Favixx, line 196: `indexOf('}', from + 1)` finds the first closing brace, so `/{{A} B}` closes the formation type at the inner brace and leaves ` B}` to `endsClean`, which reports trailing garbage. The grammar (R-3.4.8) does not allow nesting, so the input is wrong either way - the objection is that the diagnosis points at the wrong place. Scanning for a second `{` before the closing one and failing with a message that names nesting would cost a few lines here.

eo-parser/src/main/java/org/eolang/parser/LnVoid.java
```diff
@@ -257,6 +292,8 @@ private static void endsClean(final String tail, final int from, final Span span
             final String message;
             if (tail.charAt(idx) == '/') {
                 message = "a void attribute may carry at most one type annotation";
+            } else if (tail.charAt(idx) == '|') {
+                message = "the optional marker ? must close a void type annotation";
```
Favixx: This branch fires from both `endsClean` call sites, and only one of them has seen a `?`. Reached from line 156 - no optional marker in the source at all - the message still lectures about where `?` must go. The condition that actually holds at this point is "a `|` turned up where the annotation should have ended", usually because the author put spaces around the pipe, which R-3.4.8 forbids. Splitting the two cases would make the common typo self-explanatory.

eo-parser/src/main/resources/org/eolang/parser/_funcs.xsl
```diff
@@ -29,6 +29,33 @@
...
+  <xsl:function name="eo:type-members" as="xs:string*">
+    <xsl:param name="anno" as="xs:string"/>
+    <xsl:analyze-string select="$anno" regex="\{{[^}}]*\}}|[^ ]+">
+      <xsl:matching-substring>
+        <xsl:sequence select="."/>
+      </xsl:matching-substring>
+    </xsl:analyze-string>
+  </xsl:function>
+  <!--
+  Every type atom a void's "@type" annotation names, braces, pipes,
```
Favixx: "braces, pipes, spaces and the trailing `?` dropped" - but the regex on line 53 is `[{} ?]+`, with no `|` in it. That is correct as written, because XMIR stores union members space-separated and the pipe only exists in surface EO; it is the comment that is wrong, and it is the kind of wrong that invites someone to "fix" the regex later. Dropping the word `pipes` and saying the pipe never reaches XMIR would settle it.

eo-parser/src/main/resources/org/eolang/parser/_funcs.xsl
```diff
@@ -29,6 +29,33 @@
...
   </xsl:function>
+  <!--
+  The union members of a void's "@type" annotation (R-3.4.8): a braced
+  formation type comes out whole, inner spaces and all, every other
+  member is one space-separated token. A trailing "?" is not stripped —
+  strip it before calling, since it belongs to the union, not a member.
+  -->
+  <xsl:function name="eo:type-members" as="xs:string*">
+    <xsl:param name="anno" as="xs:string"/>
+    <xsl:analyze-string select="$anno" regex="\{{[^}}]*\}}|[^ ]+">
```
Favixx, line 43: `\{[^}]*\}|[^ ]+` and `LnVoid.formation`'s `indexOf('}')` encode the same no-nesting assumption in two languages, in two modules, with nothing tying them together. If the grammar ever grows nested formation types, this regex fails silently (`{{A} B}` splits as `{{A}` plus `B}`) while the Java side fails loudly. Worth a comment here pointing at R-3.4.8 so the two stay in step.

eo-parser/src/main/resources/org/eolang/parser/_funcs.xsl
```diff
@@ -29,6 +29,33 @@
...
+    </xsl:analyze-string>
+  </xsl:function>
+  <!--
+  Every type atom a void's "@type" annotation names, braces, pipes,
+  spaces and the trailing "?" dropped: what a pass rewriting individual
+  formas (aliases, default package, alias restoring) has to look at.
+  -->
+  <xsl:function name="eo:type-atoms" as="xs:string*">
+    <xsl:param name="anno" as="xs:string"/>
+    <xsl:analyze-string select="$anno" regex="[{{}} ?]+">
```
Favixx: `[{} ?]+` here, but the three passes that rewrite `@type` in place use `[{} ?]` without the `+` (resolve-aliases 55, add-default-package 89, restore-aliases 61). Both are right for their own job - the `+` collapses separator runs for extraction, the bare class preserves them one by one for re-emission - but they read as a typo for each other. A word here on why the quantifier differs would save the next reader the trip.

eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
```diff
@@ -63,38 +63,54 @@
...
+  which a bracket param cannot express (#5614). One such void takes all
+  its siblings down with it, since "move-voids-up" hoists body voids
+  behind the head ones and a split would swap the two groups (#6189) —
+  which is also why every void of an atom goes vertical, its head having
+  to stay empty (R-3.4.10, #6082).
   -->
   <xsl:function name="eo:vertical-void" as="xs:boolean">
     <xsl:param name="o" as="element()"/>
-    <xsl:sequence select="eo:void($o) and (exists($o/@local) or exists($o/@type) or exists($o/@args) or eo:atom($o/..))"/>
+    <xsl:sequence select="eo:void($o) and (eo:atom($o/..) or exists($o/../o[eo:void(.) and (@local or @type)]))"/>
```
Favixx, line 77: The predicate widened `@local` along with `@type`: before, only the `>> name` void went vertical and its plain siblings stayed in the head; now one of them takes all of them down. That is consistent - the head-hoisting argument in #6189 applies equally to handles, so this looks like a second bug quietly fixed - but the PR body only argues the `@type` case, and no pack covers a `>>` void next to a plain one, so the change in printed output for existing sources is neither stated nor tested. Also note the comment block above the vertical-void template (around line 236) still says "names the three shapes" after this rewrite reduced them to two.

eo-printer/src/test/resources/org/eolang/printer/print-packs/yaml/void-typed-in-formation.yaml
```diff
@@ -0,0 +1,26 @@
...
+  EXCESS: 1
+  WIDTH: 80
+  STEP: 2
+  SPACE: 7
+origin: |
+  +package org.eolang
+
+  [] > matched
+    ? > position /Q.number
+    ? > groups
```
Favixx, line 17: This pack puts the annotated void first and the bare one second - the safe order. The hazard the PR body names is the other one: "printing `groups` in the head and `position` below it would silently swap the two". None of the four packs added here declares a bare void before an annotated one, so the ordering this change exists to protect has no regression test. Swapping these two lines in a copy of this pack would cover it.

eo-parser/src/main/resources/org/eolang/parser/parse/resolve-aliases.xsl
```diff
@@ -44,15 +44,24 @@
...
+  braces, pipes and the trailing "?" that hold the union together where
+  they are.
+  -->
   <xsl:template match="@type">
-    <xsl:variable name="opt" select="ends-with(., '?')"/>
-    <xsl:variable name="type" select="if ($opt) then substring(., 1, string-length(.) - 1) else string(.)"/>
-    <xsl:attribute name="type" select="concat((/object/metas/meta[head='alias' and part[1]=$type]/part[last()], $type)[1], if ($opt) then '?' else '')"/>
+    <xsl:variable name="aliases" select="/object/metas/meta[head='alias']"/>
+    <xsl:attribute name="type">
+      <xsl:analyze-string select="." regex="[{{}} ?]">
```
Favixx: With this, `[{} ?]` is hand-written in four places - here, add-default-package 89, restore-aliases 61, and `_funcs.xsl` 53 - and the same grammar is spelled a fifth time as a regex in XMIR.xsd. All three templates are the same shape: walk the annotation, pass separators through, map the atoms. A single `eo:map-type-atoms` in `_funcs.xsl` taking the per-atom rewrite would leave one definition of what a separator is.

eo-parser/src/main/resources/XMIR.xsd
```diff
@@ -97,12 +97,14 @@
...
-        type variable A-F, with an optional trailing "?" marking a maybe-⊥ value.
+        The declared type of a void attribute: the union of types it may hold,
+        members separated by a single space, each a concrete forma, a generic type
+        variable A-F or a braced formation type naming the types of that
+        formation's voids, with an optional trailing "?" marking the union maybe-⊥.
       </xs:documentation>
     </xs:annotation>
     <xs:restriction base="xs:string">
-      <xs:pattern value="([A-F]|⊥|∅|([Φξ])(\.(ρ|φ|[a-z][^\s.]*|α[0-9]*))*)[?]?"/>
+      <xs:pattern value="(([A-F]|⊥|∅|([Φξ])(\.(ρ|φ|[a-z][^\s.|{}]*|α[0-9]*))*)|\{(([A-F]|⊥|∅|([Φξ])(\.(ρ|φ|[a-z][^\s.|{}]*|α[0-9]*))*)( ([A-F]|⊥|∅|([Φξ])(\.(ρ|φ|[a-z][^\s.|{}]*|α[0-9]*))*))*)?\})( (([A-F]|⊥|∅|([Φξ])(\.(ρ|φ|[a-z][^\s.|{}]*|α[0-9]*))*)|\{(([A-F]|⊥|∅|([Φξ])(\.(ρ|φ|[a-z][^\s.|{}]*|α[0-9]*))*)( ([A-F]|⊥|∅|([Φξ])(\.(ρ|φ|[a-z][^\s.|{}]*|α[0-9]*))*))*)?\}))*[?]?"/>
```
Favixx, line 111: The forma sub-pattern is repeated six times on this line. XSD 1.0 gives no way to name a sub-expression, so some repetition is forced, but the union-of-members shape could be expressed as an `xs:list` of a named member type instead of one regex - and even keeping the regex, the line is unreviewable as a single string. Whatever the outcome, it is worth stating in the annotation above that this pattern is the fifth encoding of R-3.4.8 and must move with the other four.

eo-maven-plugin/src/main/resources/org/eolang/maven/probe/add-probes.xsl
```diff
@@ -61,7 +61,7 @@
   <xsl:template match="/object[not(metas)]">
     <xsl:variable name="candidates" as="element()*">
       <xsl:apply-templates select="//o[not(eo:abstract(.)) and not(eo:void(.))]" mode="create"/>
-      <xsl:apply-templates select="//o[eo:void(.) and @args]" mode="args"/>
+      <xsl:apply-templates select="//o[eo:void(.) and @type]" mode="type"/>
```
Favixx, line 64: The selector went from `@args` to `@type`, which widens what gets probed: before, only callback argument types were pulled; now every annotated void is, including a plain `? > position /Q.number`. For the sixteen annotations this PR adds that is harmless - they are all core objects that get pulled anyway - but it is a real change in what a `@type` annotation costs at build time, and the comment below explains why braces are included without saying anything about the plain case. Was the old omission deliberate?

Favixx (commented):
Read this as the parser change it is rather than the sixteen annotations it enables, and it holds together well: the `@args`/`@type` split collapses into one attribute, the removal of `Level.patom` takes the whole atom-only restriction out with it, and the union grammar is specified, schema-checked, round-tripped and covered by four new packs. `Level.atom()` is still live via `LnFormation`, so that removal is clean. R-3.4.8 in the spec is unusually complete - the error table is updated in both directions, which is more than most changes here manage.

Three things I would want addressed before it lands.

**Nesting is forbidden by the grammar but not by the code.** `member ::= type | '{' (type SPACE type)* '}'` has no recursive case, yet `LnVoid.formation` closes on the first `}` and `eo:type-members` splits on `\{[^}]*\}`. `/{{A} B}` therefore does not produce "nested formation types are not supported" - it produces a trailing-garbage error from the Java side and a silently wrong split from the XSLT side. Given a void whose own void is a formation is a shape people will try, R-9.9.1 seems short one canonical message.

**One canonical error text is doing two jobs.** `the optional marker ? must close a void type annotation` fires from both `endsClean` call sites, so `? > x /A |B` - spaces around a pipe, no `?` anywhere - is answered with a rule about the optional marker. R-9.9.1 promises one text per condition, and the condition this actually catches most often is the spaced pipe.

**The ordering this change exists to protect is untested.** The PR body is explicit that a bare void printed in the head ahead of an annotated one below would swap the two, and that `eo:vertical-void` deciding per formation is what prevents it. All four new packs declare the annotated void first, which is the order that was already safe. The regression test for #6189 is the reverse order, and it is not here.

Smaller: the `@local` arm of `eo:vertical-void` widened too, which looks like a deliberate second fix but is neither stated nor covered; `[{} ?]` is now hand-written in four XSLs plus a fifth encoding in the XSD; `eo:type-atoms`'s comment mentions pipes its regex does not match; and `add-probes` now probes plain `/Q.number` annotations that were never probed before.

On `pr-size`: 32 files is mostly the four packs and the five passes that had to learn the new attribute shape, so the count reads worse than the change is. If it has to be split, the `@args`-to-`@type` collapse is self-contained and could land first, leaving the union grammar and the sixteen `eo-runtime` annotations for a second PR.

eo-parser/src/main/java/org/eolang/parser/LnVoid.java
```diff
@@ -118,99 +113,150 @@ private static String head(final String tail, final int slash) {
...
+        return idx;
+    }
+
+    /**
+     * Whether the character ends a bare union member.
+     * @param symbol The character
+     * @return True if the member stops here
+     */
+    private static boolean breaks(final char symbol) {
+        return symbol == ' ' || symbol == '?' || symbol == '/' || symbol == '|';
```
morphqdd, line 248: The terminator set here is `' '`, `'?'`, `'/'`, `'|'` — no `{` or `}`. So `? > x /Q.number{A}` (a type name glued directly to a brace list, no separator) parses with no error and emits `type='Φ.number{A}'`, which matches neither branch of the `type-anno` pattern in XMIR.xsd. Nothing downstream catches it since XSD validation isn't wired into the actual parse pipeline. Adding `{`/`}` here would surface it as a real error instead of silently producing invalid XMIR.

eo-parser/src/main/java/org/eolang/parser/LnVoid.java
```diff
@@ -118,99 +113,150 @@ private static String head(final String tail, final int slash) {
...
+     * Parse a bare member: one type atom, a generic variable verbatim or
+     * a concrete forma promoted from {@code Q.} to {@code Φ.}.
+     * @param tail The line body after the {@code ?}
+     * @param from Index the atom starts at
+     * @param out Sink for the promoted atom
+     * @return Index just past the atom
      */
-    private static String members(final String inside, final Span span) {
-        if (inside.isEmpty()) {
+    private int forma(final String tail, final int from, final StringBuilder out) {
```
morphqdd, line 223: Related to the `breaks()` gap below: `/A{B}` currently does error, but only by accident, via `Suffix.typeAtom` rejecting `A{B}` as an invalid type variable. The message ("type variable must be one of A-F") is about the wrong problem — the real issue is a glued brace list with no separator, not a bad variable name.

0crat:
@favixx Thanks for the review! You've earned +21 points for this: +12 as a basis; +8 for the 852 hits-of-code that you reviewed.  Your running score is [+237](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +16 points for this: +12 as a basis; +8 for the 852 hits-of-code that you reviewed; -5 for very few (2) comments.  Your running score is [+3889](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6195, opened 2026-07-31
#6096: compile each source glob once and report a bad one as configuration

### LINKED ISSUE #6096: `Walk.matches` should handle input-related `PatternSyntaxException` from `<includeSources>`/`<excludeSources>` configurations

morphqdd:
Closes #6096

First, a correction to my own issue text: it says the error "never names the offending glob", and that is wrong. `getPathMatcher` reports on the glob, not on the translated regex, so `**/[a-.eo` already comes back as `Invalid range near index 4` with the pattern quoted underneath. I checked before writing the fix.

What is missing is the context. The exception arrives as a regex error from inside a file scan, with nothing saying it came from a glob the build was configured with. It is now wrapped so the message says which pattern is at fault and that it was configured to select sources. `Walk` is reached from four places with different parameter names (`MjRegister`, `Placing`, `Unspiling`, `Unplacing`), so the message names the role rather than one parameter.

The second half of the issue is done as well, and it is what shapes the change: each glob is compiled once in `includes`/`excludes` instead of once per file. For N files and M globs that was N*M compilations. Compiling up front is also what moves the failure off the walk and onto the configuration, so the two halves are one change rather than two.

The test fails on master.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java
@@ -8,10 +8,12 @@ import java.io.IOException;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.PathMatcher;
 import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.cactoos.list.ListEnvelope;
@@ -52,11 +54,12 @@ final class Walk extends ListEnvelope<Path> {
      */
     @SuppressWarnings("PMD.LooseCoupling")
     Walk includes(final Collection<String> globs) {
+        final Collection<PathMatcher> matchers = Walk.compiled(globs);
         return new Walk(
             this.home,
             this.stream().filter(
-                file -> globs.stream().anyMatch(
-                    glob -> this.matches(glob, file)
+                file -> matchers.stream().anyMatch(
+                    matcher -> matcher.matches(this.relative(file))
                 )
                 )
                 .collect(Collectors.toList())
@@ -70,11 +73,12 @@ final class Walk extends ListEnvelope<Path> {
      */
     @SuppressWarnings("PMD.LooseCoupling")
     Walk excludes(final Collection<String> globs) {
+        final Collection<PathMatcher> matchers = Walk.compiled(globs);
         return new Walk(
             this.home,
             this.stream().filter(
-                file -> globs.stream().noneMatch(
-                    glob -> this.matches(glob, file)
+                file -> matchers.stream().noneMatch(
+                    matcher -> matcher.matches(this.relative(file))
                 )
                 )
                 .collect(Collectors.toList())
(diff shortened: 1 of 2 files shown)


morphqdd:
Closing as a duplicate of #6151, which came first and does the same two things: wraps the `PatternSyntaxException` and compiles each glob once instead of per file. I did not check the open PRs before starting. Sorry for the noise, @h1alexbel.

renechoi (commented):
I ran this branch and master locally against the two behaviours it changes. One thing looks worth fixing, one worth claiming more loudly, and one worth a line in the description.

### The message names a parameter that is right for one of seven call sites

`compiled` is shared by `includes` and `excludes`, and the sentence is fixed text, so every glob the plugin compiles is now reported as source selection. On this branch:

```java
new Walk(dir).excludes(new ListOf<>("{foo"));
// java.lang.IllegalArgumentException:
// The glob '{foo', configured to select sources, is not a valid glob pattern
```

Where those globs actually come from, on master:

| call site | pom parameter | role |
| --- | --- | --- |
| `MjRegister:103` `includes` | `<includeSources>` | selects sources |
| `MjRegister:104` `excludes` | `<excludeSources>` | deselects sources |
| `Placing:179` `includes` | `<placeBinaries>` | selects binaries |
| `Placing:180` `excludes` | `<skipBinaries>` | deselects binaries |
| `Unplacing:76` `excludes` | `<keepBinaries>` | deselects binaries |
| `Unspiling:96` `excludes` | `<keepBinaries>` | deselects binaries |
| `Unspiling:96` `includes` | none | generated `.class` paths computed at run time, plus two constants |

#6096 asks the error to name the parameter that caused it, and a typo in `<skipBinaries>` now sends the reader to their source selection instead. `Walk` cannot know the pom parameter, but it does know which of its two lists it is compiling, so letting `includes` and `excludes` pass that role into `compiled` would make the sentence true at every call site without reaching into the mojos. The new test only exercises `includes`, which is why nothing catches this.

### The wrap earns more than the description claims

The correction in the description holds for the errors `Globs` raises itself, but not for a glob that translates into a regex `Pattern.compile` rejects. An empty character class is that case: `Globs` turns `[]` into `[[^/]&&[]]`, and the exception then carries the translated regex, so the user's glob never appears anywhere. On master, `src/[]/*.eo` gives

```
Unclosed character class near index 25
^src/[[^/]&&[]]/[^/]*\.eo$
                         ^
```

Identical on 21.0.9 and 26.0.1. Of 13 bracket-class globs I probed, 8 threw and all 8 quoted the translated regex rather than the glob; the other 5 (`[[]`, `[^]`, `[a&&]`, `[&&a]`, `[[:alpha:]]`) compile fine. So the original issue text was accurate for a real class of inputs, and this branch turns that output into `The glob 'src/[]/*.eo' ...`. That is the strongest case for wrapping, and it is not in the description.

### Eager compilation removes two escapes

A glob used to be compiled only if some file reached it. Measured:

| case | master | this branch |
| --- | --- | --- |
| directory contains no files | no throw | throws |
| `includes("**/*.eo", "{foo")` where every file matches the first glob | no throw | throws |

Failing fast is the better behaviour, but a build carrying a bad glob that never got compiled will start failing, so it reads better as a stated change than as a side effect of a refactor. Nothing in the module suite depends on the tolerant behaviour.

### What I ran

JDK 21.0.9, `mvn -pl eo-maven-plugin test` at a47ab920: 518 run, 0 failures, 5 skipped, and those 5 are the `@Disabled` ones in `MjLintTest` on master too. Reverting only the `try`/`catch` in `matcher` and leaving everything else fails exactly `namesTheMalformedGlobInsteadOfLeakingARegexError`, so the test does hold that hunk down. I ran the same probes against #6151, which closes the same issue: it also compiles once per glob and also fails eagerly, but leaves the `PatternSyntaxException` unwrapped, so the `src/[]/*.eo` case still surfaces there as the translated regex.

0crat:
@renechoi Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+4](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6196, opened 2026-07-31
fix(#6140): preserve failed POSIX read result

### LINKED ISSUE #6140: `posix.read` results in `NegativeArraySizeException` for error returns from native syscall
## What happens

`posix "read"` turns the normal POSIX error return `-1` into a Java crash. A runtime probe called the atom with a closed descriptor (`-1`) and a one-byte requested size:

| descriptor | requested size | actual result |
| --- | --- | --- |
| `-1` | `1` | `java.lang.NegativeArraySizeException: -1` |
| `0` at EOF | `1` | returns normally with code `0` and an empty output |

The second row is a control run through the same atom. It shows that the atom and the probe wiring are valid; the crash happens only after the native `read(2)` reports its documented error result.

## Why this is a bug

[`ReadSyscall.make()`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/Posix/ReadSyscall.java#L36-L45) stores the result of the native call in `count` at lines 40-42. POSIX returns `-1` when `read(2)` fails, but line 44 passes that value straight to `Arrays.copyOf(buf, count)`. Java rejects a negative target length and throws `NegativeArraySizeException`.

The public [`posix.return`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/posix.eo#L30-L34) contract carries a syscall status code and an output value. Thus an unsuccessful read should still return code `-1` and an empty byte sequence, allowing EO code to inspect the status consistently with other syscalls. It must not replace the syscall result with a JVM exception.

The analogous Windows `_read` implementation already guards this case: [`ReadFuncCall`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/Win32/ReadFuncCall.java#L37-L45) uses `Arrays.copyOf(buf, Math.max(count, 0))` while preserving the native code.

## Steps to reproduce

1. Invoke `posix` with the `"read"` syscall name.
2. Supply a closed descriptor such as `-1` and request one byte.
3. Dataize the returned `posix.return`.

Instead of a result with code `-1`, the program terminates with `NegativeArraySizeException`.

## Expected behavior

For any negative native read result, preserve the result code and return empty output. Replacing `Arrays.copyOf(buf, count)` with `Arrays.copyOf(buf, Math.max(count, 0))` matches the existing Windows implementation and keeps error reporting in the EO syscall result.

Titanism0v0:
Closes #6140.

## Summary
- preserve the native negative status returned by a failed POSIX `read`
- return an empty byte sequence instead of passing a negative length to `Arrays.copyOf`
- test the failed-status code and the empty output independently
- make every file/console `read` consumer check the status before using the output, so a failure cannot be treated as EOF

## Root cause
`ReadSyscall` used the native `read(2)` result both as the EO status code and as the output-array length. POSIX uses `-1` for errors, while Java rejects negative array lengths.

## Impact
Successful reads and EOF behavior are unchanged. Failed reads now follow the existing `posix.return` contract and match the Win32 adapter's behavior. File and console input now stop on a failed read instead of interpreting its empty output as end-of-file.

## Testing
- RED: [the new Linux regression test failed with `NegativeArraySizeException: -1`](https://github.com/objectionary/eo/actions/runs/30642208218/job/91194696576)
- GREEN: [the same test passes after clamping only the copied output length](https://github.com/objectionary/eo/actions/runs/30643937079/job/91200508562)
- current CI: running for the latest review-fix commit

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/posix/ReadSyscall.java
+++ b/eo-runtime/src/main/java/org/eolang/posix/ReadSyscall.java
@@ -38,7 +38,7 @@ public final class ReadSyscall implements Syscall {
             new Dataized(params[0]).asNumber().intValue(), buf, size
         );
         result.put(0, new Data.ToPhi(count));
-        result.put(1, new Data.ToPhi(Arrays.copyOf(buf, count)));
+        result.put(1, new Data.ToPhi(Arrays.copyOf(buf, Math.max(count, 0))));
         return result;
     }
 }
--- a/eo-runtime/src/test/java/org/eolang/EOposixEOφTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EOposixEOφTest.java
@@ -73,7 +73,7 @@ final class EOposixEOφTest {
     void preservesFailureCodeWhenReadFails() {
         MatcherAssert.assertThat(
             "Failed \"read\" should preserve its code",
-            new Dataized(EOposixEOφTest.failedRead().take("code"))
+            new Dataized(this.failedRead().take("code"))
                 .asNumber().intValue(),
             Matchers.equalTo(-1)
         );
@@ -84,7 +84,7 @@ final class EOposixEOφTest {
     void returnsEmptyOutputWhenReadFails() {
         MatcherAssert.assertThat(
             "Failed \"read\" should return empty output",
-            new Dataized(EOposixEOφTest.failedRead().take("output")).take().length,
+            new Dataized(this.failedRead().take("output")).take().length,
             Matchers.equalTo(0)
         );
     }
@@ -94,7 +94,7 @@ final class EOposixEOφTest {
      *
      * @return Failed read result
      */
-    private static Phi failedRead() {
+    private Phi failedRead() {
         return new PhApplication(
             new PhApplication(
                 Phi.Φ.take("posix").copy(),
--- a/eo-runtime/src/test/java/org/eolang/EOposixEOφTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EOposixEOφTest.java
@@ -5,7 +5,6 @@
 package org.eolang;
 
 import java.lang.management.ManagementFactory;
-import java.util.List;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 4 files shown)


eo-runtime/src/test/java/org/eolang/EOposixEOφTest.java
```diff
@@ -68,4 +69,31 @@ void reportsReasonWhenOpenFails() {
...
+            new Data.ToPhi(
+                new Phi[]{
+                    new Data.ToPhi(-1),
+                    new Data.ToPhi(1),
+                }
+            )
+        );
+        MatcherAssert.assertThat(
+            "Failed \"read\" should preserve its code and return empty output",
+            List.of(
```
Thayorns: The list packs two separate facts into one assertion: that the native code survives as -1, and that the output comes back empty. The name only promises the second one, so when this goes red the message points at a name that covers half of what broke, and a regression in either half arrives as one opaque list mismatch. Two behaviours, and each reads better standing alone.

Thayorns (commented):
Adds a regression test for the POSIX read atom, calling it with descriptor -1 and a one-byte request, and asserting the failed call keeps its code and yields empty output. No production code changes, so the new test is red on Linux by design. The wiring matches the neighbouring tests and holds up; the only thing I would look at is how much the single assertion covers.

eo-runtime/src/main/java/org/eolang/posix/ReadSyscall.java
```diff
@@ -38,7 +38,7 @@ public Phi make(final Phi... params) {
             new Dataized(params[0]).asNumber().intValue(), buf, size
         );
         result.put(0, new Data.ToPhi(count));
-        result.put(1, new Data.ToPhi(Arrays.copyOf(buf, count)));
+        result.put(1, new Data.ToPhi(Arrays.copyOf(buf, Math.max(count, 0))));
```
morphqdd, line 41: Clamping a negative count to 0 makes a syscall error (negative code) and a legitimate zero-byte read (EOF) produce the same empty output; only code still distinguishes them. Worth confirming every caller checks code before treating an empty output as EOF.

morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (2) comments; -4 for too few (30) hits-of-code; +1 to give you at least something.  Your running score is [+2744](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-runtime/src/main/java/org/eolang/posix/ReadSyscall.java
Titanism0v0: Confirmed: all file/console consumers now check the status code before using the output. Both console branches and the file reader raise an error for `received.code.eq -1`; only successful reads can interpret an empty output as EOF.

Titanism0v0:
@morphqdd 已按你的建议修正：所有 file/console 消费者都会先检查状态码，再使用输出；CI 已全绿。烦请复审，感谢！

0crat:
@titanism0v0 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+80](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6197, opened 2026-07-31
#5796: refuse a scope token as a file-local handle

### LINKED ISSUE #5796: `Suffix.auto` wrongly accepts `@`, `^`, `$` as file-local handles in `>>` auto-name suffix parsing

morphqdd:
Closes #5796

`>> @`, `>> ^` and `>> $` parsed cleanly and produced `local="@"` and friends, which is not a name anything can be written with. `Suffix.test` already refuses `@` for a test attribute; the `>>` path had no equivalent, so this adds one for all three tokens.

Three typo packs, one per token, all failing on master.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -4,8 +4,6 @@
  */
 package org.eolang.parser;
 
-import java.util.Arrays;
-import java.util.HashSet;
 import java.util.Set;
 
 /**
@@ -61,9 +59,7 @@ final class Suffix {
      * {@link #test(String, int, Span, int, Form)} already refuses
      * {@code @} for a test attribute.
      */
-    private static final Set<String> SCOPES = new HashSet<>(
-        Arrays.asList("@", "^", "$")
-    );
+    private static final Set<String> SCOPES = Set.of("@", "^", "$");
 
     /**
      * Suffix form.
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -4,6 +4,10 @@
  */
 package org.eolang.parser;
 
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
 /**
  * A parsed name suffix — §3.10 of the spec.
  *
@@ -48,6 +52,18 @@ final class Suffix {
      * {@link #terminates} so dotted FQNs are still consumed whole.
      */
     private static final String NAME_BOUNDARIES = ",.|':;?[]{}()";
+
+    /**
+     * Scope tokens, which name a place rather than an object: the
+     * {@code φ} decoratee, the {@code ρ} parent and {@code ξ} itself.
+     * A {@code >>} handle is a name a later reference can be written
+     * with, so none of these can serve as one, the same way
+     * {@link #test(String, int, Span, int, Form)} already refuses
+     * {@code @} for a test attribute.
+     */
+    private static final Set<String> SCOPES = new HashSet<>(
+        Arrays.asList("@", "^", "$")
+    );
 
     /**
      * Suffix form.
@@ -409,6 +425,14 @@ final class Suffix {
         final int begin = Suffix.skipSpace(tail, idx);
         int rest = Suffix.skipName(tail, begin);
         final String handle = tail.substring(begin, rest);
+        if (Suffix.SCOPES.contains(handle)) {
+            throw new ParseError(
+                span.line(), home + begin,
+                String.format(
+                    "file-local handle must be an identifier, not %s", handle
+                )
+            );
+        }
         if (handle.codePoints().anyMatch(cp -> cp == 0x1F335)) {
             throw new ParseError(
                 span.line(), home + begin,
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-typos/auto-handle-is-not-a-scope-token.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-typos/auto-handle-is-not-a-scope-token.yaml
@@ -1,0 +1,9 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+line: 2
+message: "file-local handle must be an identifier"
+input: |
+  [] > app
+    42 >> @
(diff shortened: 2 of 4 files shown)


Thayorns (commented):
This adds a guard in Suffix.auto that refuses @, ^ and $ as a >> file-local handle, plus three typo fixtures, one per token. I traced the path: none of the three terminates a name, so skipName consumes them whole and the handle really does arrive as a single scope token, which the new check catches at the right column; nothing under eo-packs binds a handle to one of these tokens, so no existing expectation moves. One note on how the constant is built.

eo-parser/src/main/java/org/eolang/parser/Suffix.java
```diff
@@ -49,6 +53,18 @@ final class Suffix {
...
 
+    /**
+     * Scope tokens, which name a place rather than an object: the
+     * {@code φ} decoratee, the {@code ρ} parent and {@code ξ} itself.
+     * A {@code >>} handle is a name a later reference can be written
+     * with, so none of these can serve as one, the same way
+     * {@link #test(String, int, Span, int, Form)} already refuses
+     * {@code @} for a test attribute.
+     */
+    private static final Set<String> SCOPES = new HashSet<>(
```
Thayorns: SCOPES is a HashSet, so this is a mutable object sitting in a static field for the lifetime of the class — anything that adds to it later silently changes what every parse accepts, and nothing in the type prevents that. The only other constant set in this module, CHAINABLE at eo-parser/src/main/java/org/eolang/parser/Emissions.java:33, is immutable, and building this one the other way is also what pulls in the two new imports.

morphqdd: Fixed, `Set.of` like `CHAINABLE`. Both new imports went with it.

gemshrine (commented):
This closes #5796 correctly — I ran the three bare tokens through `EoSyntax` on this branch and `>> @`, `>> ^`, `>> $` now all raise the parse error, where before they produced `local="@"` and friends. Three notes below, the first about the boundary of the fix (verified at the parser), the others minor.

eo-parser/src/main/java/org/eolang/parser/Suffix.java
```diff
@@ -409,6 +421,14 @@ private static Parsed auto(
         final int begin = Suffix.skipSpace(tail, idx);
         int rest = Suffix.skipName(tail, begin);
         final String handle = tail.substring(begin, rest);
+        if (Suffix.SCOPES.contains(handle)) {
```
gemshrine, line 424: Because this is a three-string denylist rather than a check that the handle is a valid name, the cases one step away from the bare tokens still slip through with no error and produce exactly the unwritable `local=` the issue is about. Verified on this branch through `EoSyntax`:

- `42 >> @foo` -> accepted, `local="@foo"`
- `42 >> ^^` -> accepted, `local="^^"`
- `42 >> $1` -> accepted, `local="$1"`
- `42 >> 1abc` -> accepted (a digit-leading handle)

`skipName` hands `auto()` whatever runs up to the next terminator, and nothing here checks it against the NAME grammar. A positive check — the handle must match the same NAME rule the `> name` path already enforces — would subsume all three scope tokens and close the whole class in one place. If #5796 is scoped to the bare tokens only, a follow-up issue for general handle validation would keep the rest visible.

eo-parser/src/test/resources/org/eolang/parser/eo-typos/auto-handle-is-not-a-scope-token.yaml
```diff
@@ -0,0 +1,9 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+line: 2
+message: "file-local handle must be an identifier"
+input: |
+  [] > app
```
gemshrine, line 8: These three packs pin the bare tokens, which is the issue. For the boundary: I ran `42 >> @foo`, `42 >> $1` and `42 >> 1abc` through `EoSyntax` on this branch and all three parse with no error (yielding `local="@foo"` and so on), so a pack asserting an error for any of them would fail today. Either that's the intended scope or one such pack would extend the guard to the whole unwritable-handle class. (Minor: this pack is named `...-scope-token` while the other two are `...-dollar-token` / `...-hat-token`; `...-at-token` would match the pattern, since `@` is the token it exercises.)

eo-parser/src/main/java/org/eolang/parser/Suffix.java
```diff
@@ -409,6 +421,14 @@ private static Parsed auto(
...
+            throw new ParseError(
+                span.line(), home + begin,
+                String.format(
+                    "file-local handle must be an identifier, not %s", handle
```
gemshrine, line 428: Small wording point: the message interpolates `handle`, which `skipName` cuts at the first boundary, so `42 >> ^.a` reports `must be an identifier, not ^` — naming `^` when the user wrote `^.a` (the `.` ends the name, so only `^` is caught and rejected). Harmless, since the input is invalid either way, but if the message is meant to point at what was typed, the full offending run would read better than the truncated token.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+498](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+1236](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2909](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6199, opened 2026-07-31
fix(#6138): translate invalid directory paths

### LINKED ISSUE #6138: `directory.walk` throws `InvalidPathException` for malformed `file.path` rather than `ExFailure`
## What happens

`directory.walk` leaks Java's `InvalidPathException` when `file.path` is not a valid host path. For example, a path containing a NUL character fails before the atom can produce an EO runtime failure:

| `file.path` | `glob` | actual result |
| --- | --- | --- |
| `"\u0000"` | `"*"` | `java.nio.file.InvalidPathException: Nul character not allowed` |

The same probe with a syntactically valid but nonexistent path produces the expected EO-level error instead:

| `file.path` | `glob` | actual result |
| --- | --- | --- |
| `"/tmp/eo-directory-walk-missing"` | `"*"` | `org.eolang.ExFailure: Can't walk the directory '/tmp/eo-directory-walk-missing'` |

Thus the failure is not caused by the test setup or by glob compilation. It is the unhandled path-parsing boundary.

## Steps to reproduce

1. Invoke the `directory.walk` atom with a `file` whose `path` data is a UTF-8 string containing `\u0000`.
2. Bind `glob` to `"*"`.
3. Dataize the result.

`Paths.get()` throws `InvalidPathException` immediately.

## Why this is a bug

[`EOdirectory$EOwalk.lambda()`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/EOdirectory%24EOwalk.java#L35-L40) builds the host `Path` before either of its error-handling blocks. It catches a malformed glob at [lines 43-51](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/EOdirectory%24EOwalk.java#L43-L51) and an I/O error from `Files.walk()` at [lines 53-69](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/EOdirectory%24EOwalk.java#L53-L69), but it does not catch the unchecked `InvalidPathException` thrown at the preceding conversion.

The atom therefore exposes a Java implementation exception for input supplied by EO, while closely adjacent invalid inputs are reported as `ExFailure`.

## Expected behavior

Invalid `file.path` data should be translated to an explanatory `ExFailure`, consistently with malformed glob patterns and inaccessible paths. The fix can wrap the `Paths.get(...).toAbsolutePath()` conversion in a `try`/`catch (InvalidPathException ex)` block and preserve the original exception as the cause.

Titanism0v0:
Closes #6138

## What changed

`directory.walk` now translates `InvalidPathException` raised while parsing
`file.path` into `ExFailure` at the atom boundary. The message identifies the
malformed path and the original exception remains available as the cause.

A regression test invokes the atom directly with an invalid path and verifies
both the EO-facing failure type and the preserved Java cause.

## Verification

- `mvn -pl eo-runtime -DskipITs -Dtest=EOdirectoryEOwalkTest test`
  - 2 tests, 0 failures, 0 errors
- `mvn -pl eo-runtime clean test -DskipITs`
  - 1819 tests, 0 failures, 0 errors, 10 skipped
- `mvn -pl eo-runtime -Pqulice -DskipTests qulice:check` (JDK 21)
  - Qulice quality check completed
- `mvn -pl eo-runtime -Pqulice com.github.volodya-lombrozo:jtcop-maven-plugin:1.4.3:check`
  - All tests are valid

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/EOdirectory$EOwalk.java
+++ b/eo-runtime/src/main/java/org/eolang/EOdirectory$EOwalk.java
@@ -42,7 +42,9 @@ public final class EOdirectory$EOwalk extends PhDefault implements Atom {
             path = Paths.get(raw).toAbsolutePath();
         } catch (final InvalidPathException ex) {
             throw new ExFailure(
-                String.format("'%s' is not a valid path", raw),
+                String.format(
+                    "'%s' is not a valid path: %s", raw, ex.getReason()
+                ),
                 ex
             );
         }
--- a/eo-runtime/src/test/java/org/eolang/EOdirectoryEOwalkTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EOdirectoryEOwalkTest.java
@@ -5,6 +5,9 @@
 
 package org.eolang;
 
+import java.nio.file.InvalidPathException;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
@@ -34,19 +37,53 @@ final class EOdirectoryEOwalkTest {
 
     @Test
     void wrapsInvalidPathIntoExFailure() {
-        final Phi file = Phi.Φ.take("file").copy();
-        file.put(0, new Data.ToPhi(String.format("/tmp/%cinvalid", (char) 0)));
-        final Phi dir = Phi.Φ.take("directory").copy();
-        dir.put(0, file);
         Assertions.assertThrows(
             ExFailure.class,
-            () -> new Dataized(
-                new PhApplication(
-                    new PhApplication(new EOdirectory$EOwalk(), Phi.RHO, dir),
-                    "glob", new Data.ToPhi("*")
-                )
-            ).take(),
+            () -> new Dataized(EOdirectoryEOwalkTest.invalidPathWalk()).take(),
             "a path with a NUL character must fail with ExFailure, not a raw InvalidPathException"
         );
+    }
+
+    @Test
+    void includesInvalidPathReason() {
+        final ExFailure failure = EOdirectoryEOwalkTest.failure(
+            EOdirectoryEOwalkTest.invalidPathWalk()
+        );
+        MatcherAssert.assertThat(
+            "failure must identify why parsing failed",
+            failure.getMessage(),
+            Matchers.containsString(
+                ((InvalidPathException) failure.getCause()).getReason()
+            )
+        );
+    }
+
+    /**
+     * Make a directory walk application with an invalid path.
+     * @return Directory walk application
+     */
+    private static Phi invalidPathWalk() {
+        final Phi file = Phi.Φ.take("file").copy();
+        file.put(0, new Data.ToPhi(String.format("/tmp/%cinvalid", (char) 0)));
+        final Phi directory = Phi.Φ.take("directory").copy();
+        directory.put(0, file);
+        return new PhApplication(
+            new PhApplication(new EOdirectory$EOwalk(), Phi.RHO, directory),
+            "glob", new Data.ToPhi("*")
+        );
+    }
+
+    /**
+     * Get a failure from the directory walk application.
+     * @param walk Directory walk application
+     * @return Failure raised by the atom
+     */
+    private static ExFailure failure(final Phi walk) {
+        try {
+            new Dataized(walk).take();
+        } catch (final ExFailure err) {
+            return err;
+        }
+        throw new IllegalStateException("Invalid path must fail");
     }
 }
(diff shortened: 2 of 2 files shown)


Thayorns (commented):
The atom now catches InvalidPathException from Paths.get and rethrows it as ExFailure carrying the offending path and the original exception as the cause, which lines up with how the glob and IO failures are already handled a few lines below. The production change is small and reads consistently with its neighbours, and I found nothing wrong with it. My comments are both about the new test.

eo-runtime/src/test/java/org/eolang/EOdirectoryEOwalkTest.java
```diff
@@ -31,4 +34,38 @@ void wrapsMalformedGlobIntoExFailure() {
             "a malformed glob must fail with ExFailure, not a raw PatternSyntaxException"
         );
     }
+
+    @Test
+    @SuppressWarnings("PMD.UnitTestContainsTooManyAsserts")
```
Thayorns: This test checks three separate things at once: that ExFailure is thrown on line 53, what its message says on line 58, and what its cause is on line 65. The PMD rule had to be silenced on line 39 to let that through, which is the linter telling you the same thing. The sibling test just above stops at assertThrows, so when it goes red you know exactly what broke; when this one goes red the name no longer tells you which of the three claims failed.

eo-runtime/src/test/java/org/eolang/EOdirectoryEOwalkTest.java
```diff
@@ -31,4 +34,38 @@ void wrapsMalformedGlobIntoExFailure() {
...
+        walk.put(Phi.RHO, directory);
+        walk.put("glob", new Data.ToPhi("*"));
+        final ExFailure failure = Assertions.assertThrows(
+            ExFailure.class,
+            walk::lambda,
+            "invalid path must be translated at the atom boundary"
+        );
+        MatcherAssert.assertThat(
+            "failure must identify path parsing",
+            failure.getMessage(),
```
Thayorns, line 54: This pins getMessage to the exact string built in EOdirectory$EOwalk.java:42, so rewording that message turns the test red even though the atom still behaves identically. The wording of a failure is the atom's business and it will drift; the test then reports a break that nobody caused.

eo-runtime/src/main/java/org/eolang/EOdirectory$EOwalk.java
```diff
@@ -33,11 +34,18 @@ public final class EOdirectory$EOwalk extends PhDefault implements Atom {
...
-        ).toAbsolutePath();
+        final String source = new Dataized(
+            this.take(Phi.RHO).take("file").take("path")
+        ).asString();
+        final Path path;
+        try {
+            path = Paths.get(source).toAbsolutePath();
+        } catch (final InvalidPathException ex) {
+            throw new ExFailure(
+                String.format("Can't parse '%s' as a path", source),
```
morphqdd: The wrapped message doesn't include the InvalidPathException's own reason (e.g. the index of the offending character), only the raw source string. The cause is preserved via ex so it's not lost, but the top-level message alone won't tell someone why parsing failed.

morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+2843](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-runtime/src/test/java/org/eolang/EOdirectoryEOwalkTest.java
Titanism0v0: Split the coverage into wrapsInvalidPathAsFailure() and includesInvalidPathReason(). Each test now has one assertion, and the PMD suppression is gone.

eo-runtime/src/test/java/org/eolang/EOdirectoryEOwalkTest.java
Titanism0v0: Removed the exact message assertion. The reason test only requires the message to contain InvalidPathException.getReason(), so the surrounding wording can change.

eo-runtime/src/main/java/org/eolang/EOdirectory$EOwalk.java
Titanism0v0: The ExFailure message now appends ex.getReason() while retaining the source path and original cause.

Titanism0v0:
@Thayorns The review comments are addressed: the failure now includes InvalidPathException reason, tests are single-purpose, and all CI checks pass. Could you please re-review when convenient?

0crat:
@titanism0v0 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+96](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6201, opened 2026-07-31
#6133: take the low two bytes of the i32 form in htons, guard the range

### LINKED ISSUE #6133: `socket.eo` incorrectly rejects ports above 32767 due to `htons` conversion limitations
`htons` converts its argument through a signed 16-bit integer at `eo-runtime/src/main/eo/socket.eo:271`, so `socket` cannot reach any port from 32768 upward:

```
[port] > htons
  as-i16. > @
    or.
      left.
        b.and 00-FF
        8
      and.
        b.right 8
        00-FF
  port.as-i16.as-bytes > b
```

`number.as-i16` routes through `i32.as-i16` at `i32.eo:14`, which validates by round-tripping the value through a signed i16 and comparing. Ports are unsigned 16-bit, so the comparison rejects the whole upper half of a legitimate range. That range is not exotic: 32768 to 60999 is the default ephemeral range on Linux, and 49152 to 65535 on macOS, which covers every port the operating system hands out on its own.

The two-byte width that the syscalls need is not what breaks here. `SockaddrIn.port` is a Java `short` and every call site reads it through `new Dataized(params[1].take("port")).take(Short.class)`, which throws in `BytesRaw.whenFit` unless the delta is exactly two bytes. That width is established by the other conversion, `as-i16.` at `socket.eo:263`, which goes through `bytes/as-i16.eo` and asserts only `b.size.eq 2`. Signedness never enters into it: port 10112 swaps to `0x8027`, lands in `sin_port` as a negative `short`, and connects without complaint, exactly as it does in C, where `htons` returns a `uint16_t` into a `uint16_t` field.

The failure is also unreadable. `cant-convert` is left unbound at the call site, so the `Can't convert i32 number %d to i16 because it's out of i16 bounds` message that `i32.as-i16` prepared is never delivered. The bottom object escapes into `b` instead, flows through the `and`/`left`/`right`/`or` chain into `sockaddr`, and finally surfaces from `posix.sockaddr-in.size` as `the 'x' attribute must be a number`, which names neither the port nor the conversion. Nothing catches any of this in CI because `SyscallTest.randomPort` draws from 10000 to 20000, always below the boundary.

Taking the low two bytes of `port.as-i32.as-bytes` gives the right value for the full range and keeps the delta at two bytes, but it must come with a guard that the port fits in 0 to 65535, since a bare slice would quietly turn port 70000 into port 4464 rather than refusing it. An out-of-range port is a caller's mistake rather than a recoverable condition, so that guard belongs on the bottom object and not on a new `cant-` fallback.

morphqdd:
Closes #6133

`htons` converted the port through `as-i16`, which rejects any value above 32767 (i16's signed range). Ports are unsigned 16-bit, so the whole upper half of the range (32768-65535, the default Linux/macOS ephemeral port range) was rejected.

Takes the low two bytes of `port.as-i32.as-bytes` instead. That covers the full 0-65535 range and keeps the same two-byte width `SockaddrIn.port` needs. Added a guard so an out-of-range port fails with a clear message instead of silently truncating (a bare slice would have turned port 70000 into a bogus in-range value).

Tests: `convertsPortAboveSignedShortLimitToNetworkByteOrder` (port 40000, above the old i16 limit) and `rejectsPortOutsideValidRange` (port 70000, expects `ExFailure`).


eo-runtime/src/main/eo/socket.eo
```diff
@@ -268,4 +268,11 @@
         and.
           b.right 8
           00-FF
-    port.as-i16.as-bytes > b
+    if. > b
```
Thayorns: `socket` takes `port` as a free attribute at line 54 and hands it straight into `htons port` at line 198 without a word about its range, so this guard makes the byte-order converter the object that owns the port policy. Someone who builds a socket on port 70000 hears nothing until `sockaddr` gets dataized, and then the complaint arrives from an object named after a C byte-swap function. Does the range really belong here rather than where the port enters?

Thayorns (commented):
htons stopped going through as-i16 and now takes the low two bytes of the i32 form, which is the right fix for the 32768-65535 range the issue is about, and the swap in @ is untouched so the existing byte order still holds. The expected 40-9C for port 40000 lines up with the private htons helper already sitting in SyscallTest. Two things worth a look: where the new range guard lives, and the fact that the new test carries two assertions for two behaviors.

eo-runtime/src/test/java/org/eolang/SyscallTest.java
```diff
@@ -150,6 +151,25 @@ void sendsAndReceivesMessageViaSocketObject() throws InterruptedException, IOExc
...
+    void handlesFullUnsignedPortRangeInHtons() {
+        final Phi socket = Phi.Φ.take("socket").copy();
+        final Phi valid = socket.take("htons").copy();
+        valid.put(0, new Data.ToPhi(40_000));
+        MatcherAssert.assertThat(
+            "htons should convert a port above 32767 to network byte order, but it didn't",
+            new Dataized(valid).take(),
+            Matchers.equalTo(new byte[]{(byte) 0x40, (byte) 0x9C})
+        );
+        final Phi invalid = socket.take("htons").copy();
```
Thayorns: From here the method starts over as a second test: a fresh `htons` copy, port 70000, an `assertThrows`. One method now proves both that 40000 converts and that 70000 is rejected, and the `assertThat` above it is no longer the last statement. When the conversion assertion goes red the rejection check never runs at all, so a build tells you about one of the two behaviors and stays silent on the other. The name shows the strain too - `handlesFullUnsignedPortRange` has to be vague enough to cover both claims.

eo-runtime/src/main/eo/socket.eo
morphqdd: Moved the guard to socket itself. Added `checked-port` next to `htons`, called from `sockaddr` (`htons checked-port` instead of `htons port`). `htons` is back to a pure swap with no knowledge of range.

eo-runtime/src/test/java/org/eolang/SyscallTest.java
morphqdd: Split into two: `convertsPortAboveSignedShortLimitInHtons` (port 40000 through `htons` directly) and `rejectsOutOfRangePortAtCheckedPort` (port 70000, checks `socket`'s own `checked-port`). Each fails independently now.

eo-runtime/src/main/eo/socket.eo
```diff
@@ -268,4 +275,15 @@
         and.
           b.right 8
           00-FF
-    port.as-i16.as-bytes > b
+    port.as-i32.as-bytes.slice > b
```
gemshrine, line 278: The range check lives in `checked-port` at the call site (line 197), but `htons` itself still silently truncates — the exact failure this PR closes, just moved one level in. `htons` is public and the test on line 282 calls it directly, so I dataized it out of range: `htons 70000` -> `70-11` (a bogus in-range port, 4464), `htons 65536` -> `00-00` (port 0), `htons -1` -> `FF-FF`. Any caller other than the guarded socket path gets the corruption back. Guarding inside `htons` would cover both paths and let `checked-port` fold away; as it stands the atom's own contract is still unsafe.

eo-runtime/src/main/eo/socket.eo
```diff
@@ -268,4 +275,15 @@
...
+      2
+      2
+
+  eq. ++> can-convert-a-port-above-the-signed-short-limit-in-htons
+    40-9C
+    socket.htons 40000
+
+  checked-port. --> rejects-a-port-above-the-valid-range
```
gemshrine, line 286: This `-->` doesn't actually prove the guard fires. It transpiles to `assertThrows(Exception, asBool(checked-port))`, and `asBool` throws on the fallback too: if `checked-port` were broken and returned the raw number `70000`, `asBool(70000)` still throws (a number is not a bool), so the test stays green with the guard removed. The discriminating check was the JUnit dropped for the size cap. A hand-written test asserting `containsString("out of range")` on the `ExFailure` (the pattern the other socket tests already use) would pin the rejection for real; the `-->` alone can't.

eo-runtime/src/main/eo/socket.eo
```diff
@@ -268,4 +275,15 @@
         and.
           b.right 8
           00-FF
-    port.as-i16.as-bytes > b
+    port.as-i32.as-bytes.slice > b
+      2
+      2
+
+  eq. ++> can-convert-a-port-above-the-signed-short-limit-in-htons
```
gemshrine, line 282: Good to have the above-signed-limit case pinned. The two edges that would complete it are untested: `65535` (top of the valid range) and `32768` (the first port the old `as-i16` rejected — the regression boundary). I verified both are correct here — `htons 65535` -> `FF-FF`, `htons 32768` -> `00-80` — so an `eq.` at `65535` in particular is a cheap guard on the high edge of the new `slice`.

eo-runtime/src/main/eo/socket.eo
```diff
@@ -259,6 +258,14 @@
         seq * > [buffer] > write
           send buffer
           output-block
+  if. > [] > checked-port
```
gemshrine, line 261: Minor: the guard bounds the value but not its integrality, so an in-range non-integer slips through — `htons 8080.9` -> `90-1F`, the wire bytes for `8080`, silently rounded down by `as-i32`. Ports are integers, so `port.is-integer.not` could join this `or.` if you want the same clear error instead of a quiet truncation. Not important — just noting this is a range check, not a port check.

gemshrine (commented):
The core fix is correct — I dataized `socket.htons` across the whole range on this branch and every port maps right, including the previously-rejected upper half: `0->00-00`, `256->00-01`, `32768->00-80`, `40000->40-9C`, `50000->50-C3`, `65535->FF-FF`. The `as-i32.as-bytes.slice 2 2` plus the existing byte-swap reproduces network order exactly. Four notes below, none blocking the fix: the guard sits at the call site rather than in `htons`, the rejection test is vacuous, two boundaries are untested, and the guard is a range check but not an integer check.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (2) comments; -4 for too few (24) hits-of-code; +1 to give you at least something.  Your running score is [+502](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (4) comments; -4 for too few (24) hits-of-code; +1 to give you at least something.  Your running score is [+1229](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2921](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6202, opened 2026-07-31
feat(#4752): add string.padded-left

asmirnov-backend:
Part #4752.

Adds `string.padded-left`: pad text on the left with a fill character until it counts the given width. Characters are counted, and not bytes, so a multi-byte character widens the text by one.

A width that is not a finite non-negative integer, or a fill that is not a single character, falls back to `cant-pad`, following the `cant-repeat`/`cant-print` convention of `string.repeated` and `string.as-fixed`.

This is the primitive the pure-EO `printf` needs for its width (`%5d`, `%25x`). `string.padded-right`, for the `-` flag, follows in a separate PR to stay under the 200-line limit.

### DIFF
--- a/eo-runtime/src/main/eo/string/padded-left.eo
+++ b/eo-runtime/src/main/eo/string/padded-left.eo
@@ -1,0 +1,127 @@
+# Returns `text` padded on the left with `fill` until it counts `width`
+# characters, so that `42` padded to five characters with a space
+# becomes `   42`. Characters are counted, and not bytes, so a multi-byte
+# character widens the text by one. A `text` of `width` characters
+# or longer is returned unchanged.
+# If `width` is not a finite non-negative integer (negative, fractional, nan
+# or infinite), or `fill` is not a single character, the `cant-pad` fallback
+# is returned, or the bottom object when unbound.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package string
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[text width fill cant-pad] > padded-left
+  if. > @
+    or.
+      or.
+        0.gt width
+        width.is-integer.not
+      not.
+        1.eq fill.length
+    cant-pad
+      "Can't pad text to %f characters with %s".printf
+        * width fill
+    if.
+      width.gt
+        text.length >> len!
+      string
+        concat.
+          as-bytes.
+            repeated
+              fill
+              as-number.
+                width.minus len
+              cant-pad
+          text.as-bytes
+      text
+
+  eq. ++> can-count-characters-and-not-bytes
+    "  Ж"
+    padded-left
+      "Ж"
+      3
+      " "
+
+  eq. ++> can-format-the-error-message-of-a-negative-width
+    "Can't pad text to -1.000000 characters with x"
+    padded-left
+      "y"
+      -1
+      "x"
+      message > [message]
+
+  eq. ++> can-pad-a-number-with-leading-spaces
+    "   42"
+    padded-left
+      "42"
+      5
+      " "
+
+  eq. ++> can-pad-with-leading-zeros
+    "0042"
+    padded-left
+      "42"
+      4
+      "0"
+
+  eq. ++> can-return-a-longer-text-unchanged
+    "hello"
+    padded-left
+      "hello"
+      3
+      " "
+
+  eq. ++> can-return-a-text-of-the-exact-width-unchanged
+    "hey"
+    padded-left
+      "hey"
+      3
+      " "
+
+  eq. ++> can-return-a-text-unchanged-at-zero-width
+    "x"
+    padded-left
+      "x"
+      0
+      " "
+
+  eq. ++> rejects-a-multi-character-fill
+    "recovered"
+    padded-left
+      "y"
+      5
+      "ab"
+      "recovered" > [message]
+
+  eq. ++> rejects-a-negative-width
+    "recovered"
+    padded-left
+      "y"
+      -1
+      "x"
+      "recovered" > [message]
+
+  eq. ++> rejects-a-non-integer-width
+    "recovered"
+    padded-left
+      "y"
+      2.5
+      "x"
+      "recovered" > [message]
+
+  eq. ++> rejects-an-infinite-width
+    "recovered"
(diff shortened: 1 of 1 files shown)


Thayorns (approved):
Adds `string.padded-left`, which validates `width` as a finite non-negative integer and `fill` as a single character, then builds the result by concatenating `string.repeated fill (width - len)` in front of the text, returning the text untouched once it already counts `width` characters. I followed the composition through: `string.length` counts characters rather than bytes, so the `Ж` case and the `width.minus len` count are both in characters, and `number.is-integer` rejects fractional, nan and infinite widths through `is-finite`, so the guard covers everything the docblock promises. The fallback wiring, the naming, the docblock and the test set all line up with `string.repeated` and `string.as-fixed`; the one behavior the docblock names that no test reaches is a nan width, though `is-integer` covers nan on its own side.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+444](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-runtime/src/main/eo/string/padded-left.eo
```diff
@@ -0,0 +1,127 @@
...
+
+  eq. ++> rejects-a-non-integer-width
+    "recovered"
+    padded-left
+      "y"
+      2.5
+      "x"
+      "recovered" > [message]
+
+  eq. ++> rejects-an-infinite-width
```
morphqdd, line 116: The doc comment lists NaN alongside negative/fractional/infinite as an invalid width, but there's no rejects-a-nan-width test alongside the negative/fractional/infinite cases that are each tested individually.

morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2784](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

gemshrine (commented):
Correct atom, guards and fallback all verified at runtime (nan/inf/empty-fill/multi-char-fill all take the `cant-pad` branch, multi-byte fill and text are counted by character). Four notes: the `repeated`-based padding inherits #6239 (fails at width around 700 on a default stack, ~19 ms/char even when it doesn't); a circular `printf` dependency worth watching for #4752; and two coverage gaps (nan width, multi-byte fill) that I confirmed behave correctly but aren't pinned by a test.

eo-runtime/src/main/eo/string/padded-left.eo
```diff
@@ -0,0 +1,127 @@
...
+    cant-pad
+      "Can't pad text to %f characters with %s".printf
+        * width fill
+    if.
+      width.gt
+        text.length >> len!
+      string
+        concat.
+          as-bytes.
+            repeated
```
gemshrine, line 34: `padded-left` builds the padding with `string.repeated`, whose recursion depth equals the character count, so this atom inherits #6239. Measured on this branch (`text="x"`, `fill=" "`):

| width | default JVM stack | `-Xss8m` |
|---|---|---|
| 500 | ok (~10 s) | ok |
| 700 | **ExFailure (SOE)** | ok |
| 1000 | **ExFailure** | ok, ~19 s |

The description positions this as the primitive behind `printf` widths, where a large field (`%700d`) is realistic — and there it fails on a default stack, and is ~19 ms/char even when it doesn't. Fixing #6239 (build `repeated` by doubling) fixes this transitively; otherwise the two are coupled and it's worth a note here.

eo-runtime/src/main/eo/string/padded-left.eo
```diff
@@ -0,0 +1,127 @@
...
+[text width fill cant-pad] > padded-left
+  if. > @
+    or.
+      or.
+        0.gt width
+        width.is-integer.not
+      not.
+        1.eq fill.length
+    cant-pad
+      "Can't pad text to %f characters with %s".printf
```
gemshrine, line 26: Minor, but relevant to #4752: this atom is described as the primitive the pure-EO `printf` will use for its width, yet its own error path calls `printf` — a mutual dependency (`printf` -> `padded-left` -> `printf`). No runtime problem today (the `%f`/`%s` here are well-typed), but once `printf` is built on `padded-left`, a formatting error inside `padded-left` re-enters `printf`. Worth keeping in mind for the follow-up.

eo-runtime/src/main/eo/string/padded-left.eo
gemshrine, line 116: The header lists `nan` alongside infinite as a rejected width, and `pinf` is tested here but `nan` is not. I checked at runtime that `nan` is in fact rejected (it takes the `cant-pad` branch), so this is a coverage gap, not a bug — but a `rejects-a-nan-width` example next to this one would pin the branch the doc promises, since `is-integer` rejects both and `nan` could regress unnoticed.

eo-runtime/src/main/eo/string/padded-left.eo
```diff
@@ -0,0 +1,127 @@
...
++spdx SPDX-License-Identifier: MIT
+
+[text width fill cant-pad] > padded-left
+  if. > @
+    or.
+      or.
+        0.gt width
+        width.is-integer.not
+      not.
+        1.eq fill.length
```
gemshrine, line 24: The "characters, not bytes" contract is tested for `text` but not for `fill`: every fill in the tests is single-byte. I verified a multi-byte fill works — `padded-left "x" 3 "Ж"` -> `"ЖЖx"` (counted as one character, so `1.eq fill.length` accepts it) — which is the property most likely to break if `length` ever became bytes. One `can-pad-with-a-multi-byte-fill` case would guard it. (Empty `""` fill is also correctly rejected but untested — the `"ab"` case covers only the >1 side of this guard, not the 0 side.)

0crat:
@asmirnov-backend Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (127 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6203, opened 2026-07-31
Give the syscall tests a shared Port object

### LINKED ISSUE #6137: `SyscallTest.java` exceeds line limit due to duplicated Windows and POSIX test cases
`eo-runtime/src/test/java/org/eolang/SyscallTest.java` stands at 961 lines against the 1000-line `FileLengthCheck` that qulice runs at `verify`. Two tests for #6132 took it to 1018 and turned the qulice job red on #6135; they only fit after being rewritten to lean on the `dataized` object instead of carrying a fallback fake of their own, and the file now sits four lines below the cap. The next test that touches sockets has nowhere to land.

The length comes from `WindowsSocketTest` at line 172 and `PosixSocketTest` at line 522, the same seven cases written twice over — `connectsToLocalServerViaSyscall`, `refusesConnectionViaSyscall`, `bindsSocketSuccessfullyViaSyscall`, `startsListenOnPosixSocket`, `acceptsConnectionOnSocket`, `sendsAndReceivesMessagesViaSyscalls` and a private `ensure` — differing only in whether they reach `Winsock.INSTANCE` or `CStdLib.INSTANCE`, and in how each reads the last error. The copy shows through in the Windows half at line 249:

```java
@RepeatedIfExceptionsTest(repeats = 3)
void startsListenOnPosixSocket() throws UnknownHostException {
    ...
                        "Posix socket should have been bound to localhost via syscall, but it didn't, reason: %s",
    ...
                    Winsock.INSTANCE.listen(socket, 2),
```

A Windows test named after POSIX, reporting a bind while it drives `listen`. Neither gate objects to any of it: `.github/workflows/simian.yml` excludes the file by name through `-excludes=**/SyscallTest.java`, an exclusion #6113 restored on purpose after the rename from `EOsocketTest`, so the duplication stays invisible until the length check finally trips over it.

Both halves already name a production class of their own, since `org/eolang/Posix/CStdLib.java` and `org/eolang/Win32/Winsock.java` are hand-written: moving each nested class next to its binding, as `CStdLibTest` and `WinsockTest`, satisfies `JTCOP.RuleAllTestsHaveProductionClass` — the rule that turned `EOsocketTest` into `SyscallTest` in the first place — and leaves `SyscallTest` holding only the cases that drive the EO `socket` object. That move has to find a home for `RandomServer`, `randomPort`, `htons` and `assertReceived`, and it should take the simian exclusion with it.

Thayorns:
`SyscallTest` sits at 996 lines against the 1000-line `FileLengthCheck` that qulice runs at `verify`, so the next test that touches sockets has nowhere to land. The length comes from `WindowsSocketTest` and `PosixSocketTest` being the same six cases written twice, and moving those two nested classes out to `WinsockTest` and `CStdLibTest` is a relocation of roughly 700 lines, far past the size a pull request is allowed to carry here.

So this takes the first step the move needs. The private static `randomPort()` becomes a `Port` object in `org.eolang`, where both halves can still reach it and where `org.eolang.win32.WinsockTest` and `org.eolang.posix.CStdLibTest` will be able to reach it once they exist. That also retires a static utility method, which the Objects rules do not allow in the first place. The thirteen call sites now read `new Port().number()`, and the file loses six lines.

The rest of the work stays as a puzzle on the class: move both nested classes next to the bindings they drive, taking `htons`, `RandomServer`, `assertReceived` and the simian exclusion with them.

Closes #6137


Thayorns:
@yegor256 the full move of both nested classes runs to roughly 700 hits, so this does the shared `Port` object first and leaves the relocation as a puzzle; happy to explain any of it.

eo-runtime/src/test/java/org/eolang/SyscallTest.java
```diff
@@ -35,6 +34,10 @@
  * Test case for the {@link Syscall} implementations behind the
  * {@code socket} object, both the POSIX and the Windows ones.
  * @since 0.40
+ * @todo #6137:30min Move WindowsSocketTest to org.eolang.win32.WinsockTest and
```
morphqdd: The linked @todo #6137 puzzle argues this file needs splitting to stay under the 1000-line FileLengthCheck; after this PR's own trim the file is at 990 lines, one small change away from breaching it. Reinforces that the split shouldn't wait.

morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2770](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-runtime/src/test/java/org/eolang/SyscallTest.java
Thayorns: Agreed on the risk, and this PR moves the file the right way — 996 lines on master down to 986 now. Ten lines of headroom is still not a test case, so I filed #6246 with the concrete shape of the split: `WindowsSocketTest` (line 201, ~350 lines) to `org.eolang.win32.WinsockTest` and `PosixSocketTest` (line 551, ~440 lines) to `org.eolang.posix.CStdLibTest`, carrying `htons`, `RandomServer`, `assertReceived` and the simian exclusion.

I did not fold the split into this PR because moving ~800 lines would bury the `Port` change it exists for. I also dropped the puzzle that described it, since @yegor256 asked on #6188 that puzzles be reserved for what blocks the task.

eo-runtime/src/test/java/org/eolang/Port.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Port number.
+     */
+    private final int number;
+
+    /**
+     * Ctor, takes a port from the unprivileged range.
+     */
+    public Port() {
+        this(new Random().nextInt(10_001) + 10_000);
```
yegor256: @Thayorns this is unsafe. See how it can done safer: https://github.com/yegor256/random-port (I'm not sure anything like this exists for Java though)

Thayorns: @yegor256 you are right, guessing a number in a range and hoping it is free is a race. I did not find a Java port of `random-port`, so `Port` now does what the gem does: it opens a `ServerSocket` on port 0, lets the kernel pick a free one, reads the number and closes the socket. Nothing left to guess. The address the port has to be free at is an attribute, defaulting to the loopback, since that is where these tests bind.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3923](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6204, opened 2026-07-31
Enable the XmirPackage annotation in generated package-info

### LINKED ISSUE #3671: can't use `XmirPackage` in package-info.java
At [this line](https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/java/org/eolang/maven/TranspileMojo.java#L337), in `TranspileMojo`, if we remove `//`, the build fails with internal Javac error (at least for me, with OpenJDK 23.0.1). Let's find out what's wrong and enable the usage of `XmirPackage`.

Thayorns:
The `@XmirPackage` annotation was commented out in the generated `package-info.java` because uncommenting it killed the build with an internal Javac error. The error turns out to need three things at once: two `package-info.java` files for the same package, an annotation on one of them, and `-Xpkginfo:always`, which this project passes to the compiler in the parent `pom.xml`. Under those conditions Javac hits an `AssertionError` in `Lower.visitPackageDef`, where it tries to set the annotations of a package symbol that already carries some. Without the `-Xpkginfo:always` flag the same duplicate only produces a warning, and the flag is there on purpose, to keep Maven from recompiling everything on every build.

In `eo-runtime` the duplicate happens for `org.eolang` and `org.eolang.EO_string`, since both packages hold hand-written Java as well as transpiled objects. So `PackageInfos` now leaves a package alone when a `package-info.java` for it already exists in one of the compile source roots outside the build directory. The hand-written file wins, the generated one is not written, and Javac sees exactly one `package-info.java` per package. Roots inside the build directory are filtered out because a previous goal may have generated a `package-info.java` there, and that one should not stop us from writing ours.

`XmirPackage` itself had been deleted from `eo-runtime` back in `37a9073ea` as unused, so it comes back here. A full `mvn clean install` passes, and `javap` on the compiled `org/eolang/EO_number/package-info.class` shows `org.eolang.XmirPackage(value="number")` in the bytecode.

One thing left for later: nothing reads this annotation yet, and its retention is `CLASS`, so it is metadata in the class file and nothing more. Also, `eo-runtime/pom.xml` keeps `org/eolang/package-info.class` and `org/eolang/EO_string/package-info.class` in `keepBinaries`; those two are no longer transpiled at all, so the globs are now dead, and I left them alone rather than widen this change.

Closes #3671


Thayorns:
@yegor256 the internal Javac error turned out to need `-Xpkginfo:always` together with a duplicate `package-info.java`, so this stops generating the duplicate rather than touching that flag. Happy to explain any part of it.

eo-runtime/src/main/java/org/eolang/XmirPackage.java
```diff
@@ -0,0 +1,26 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for a Java package made from XMIR meta.
+ * @since 0.49
+ */
+@Target(ElementType.PACKAGE)
+@Retention(RetentionPolicy.CLASS)
+public @interface XmirPackage {
```
morphqdd, line 19: PR description says nothing reads this annotation yet. New public class/annotation with zero consumers and no @todo puzzle pointing at what's supposed to read it. The next person to find it has no signal this is intentionally half-wired rather than dead metadata.

morphqdd (changes requested):

eo-runtime/src/main/java/org/eolang/XmirPackage.java
Thayorns: Fair, the signal was only in the PR description and not where someone would trip over it. I filed #6245 for it rather than a puzzle: @yegor256 asked on #6188 that puzzles be kept for what actually blocks the task, and a missing consumer does not block turning the annotation on. The issue records that the grep finds exactly two mentions — the writer in `PackageInfos` and the declaration — and that `CLASS` retention is the right choice for a build-time reader that does not exist yet.

yegor256:
@Thayorns we need to wait for https://github.com/objectionary/eo/issues/6245

yegor256:
@Thayorns conflicts here

Thayorns:
@yegor256 take a look

yegor256:
@Thayorns the PR is empty?

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (0) hits-of-code; +1 to give you at least something.  Your running score is [+3642](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6206, opened 2026-07-31
Report object allocations and dispatches after dataization

### LINKED ISSUE #3489: let's collect runtime evaluation statistics (profile)
When we execute `Main.java`, let's print not only the result of dataization, but also two numbers: how many EO objects were created (object allocation) and how many times `.take()` method was called (dynamic dispatch).

Thayorns:
`Main` now prints two more numbers next to the elapsed time and the size of the result: how many objects the run brought to life and how many times it looked an attribute up by name. Dataizing `true` reports `Finished in 0.01s (1 bytes), 8 objects, 5 dispatches`.

The design is `Statistics`, an interface carrying two commands and two queries, and `Counters`, its one implementation, built on a pair of `LongAdder`s. `PhDefault` records a birth in `loaded()`, which happens once per object and, unlike a constructor, is allowed to do work, and again in `copy()`, which assigns `attrs` directly and so never reaches `loaded()`. It records a dispatch at the top of `take(String)`, where a name is actually resolved; the decorators delegate down to it, so nothing is counted twice. `Main` reads both counters before and after `Dataized` and prints the difference, so the two numbers cover the same span as the elapsed time rather than the whole JVM.

What is not here: the counters belong to the JVM rather than to a program. Threading a `Statistics` through `PhDefault`'s constructors and through `to-java.xsl` reaches 58 files and takes the change to 650 hits, which is more than one pull request should carry. I had pushed that threading after the review asked for it, and this commit takes it back out and leaves a `@todo #3489:60min` on `Statistics` in its place, so that what lands here is one chunk of working code that can be read in a sitting. The next chunk gives every object a `Statistics` of its own and makes the counting switchable and assertable, which is what the review was really after.

I could not run qulice: it needs Java 21 and this machine has 17. `CountersTest` and `MainTest` pass, 16 tests in total, against a clean `eo-runtime` build.

Closes #3489


Thayorns:
@yegor256 happy to explain any of it, or to move the counters somewhere else if this is not where you want them.

eo-runtime/src/main/java/org/eolang/Main.java
```diff
@@ -215,10 +215,12 @@ private static void run(final List<String> opts) {
         final byte[] ret = new Dataized(app).take();
         Main.LOGGER.info(
             String.format(
-                "%n---%n%s%nFinished in %.02fs (%d bytes)",
+                "%n---%n%s%nFinished in %.02fs (%d bytes), %d objects, %d dispatches",
```
Favixx, line 243: "Finished in 0.11s (1 bytes), 13 objects, 5 dispatches" reads as one sentence about one run, but only the first number is scoped to it - the elapsed time is a delta from `start`, while the other two are JVM-wide totals. The description is upfront about that, which makes it more surprising that the line does not say so. Two locals next to `start` holding `allocations()` and `dispatches()`, and two subtractions here, would make the sentence true for all three numbers at about the same cost as the disclaimer.

eo-runtime/src/main/java/org/eolang/Main.java
```diff
@@ -215,10 +215,12 @@ private static void run(final List<String> opts) {
...
                 new VerboseBytesAsString(ret).get(),
                 (System.currentTimeMillis() - start) / 1000.0,
-                ret.length
+                ret.length,
+                Profile.RUNNING.allocations(),
```
Favixx: `Main` reaches for the global directly rather than being handed a `Profile`. That is the reason there is no seam anywhere: a test that wants to assert the printed counters has no way to install a known one, and `MainTest` ends up matching `[1-9][0-9]*` because that is all it can say. A `Profile` parameter on `run` (defaulting to `RUNNING`) would cost one argument.

eo-runtime/src/main/java/org/eolang/Profile.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang;
+
+/**
+ * The runtime statistics of a program: how many objects it created and
+ * how many times it dispatched an attribute by name.
```
Favixx: "how many objects it created" is what the report says too, but the count is of `PhDefault` instances specifically. `PhSafe`, `PhOnce`, `PhNest`, `PhCoverage` and `PhLogged` all `implements Phi` without extending `PhDefault`, so none of them is ever counted, and `PhOnce.copy()` builds a fresh `PhOnce` that the number will not show. Whether decorators should count is a fair question either way; what matters is that the javadoc and the printed line both promise "objects" and deliver something narrower.

eo-runtime/src/main/java/org/eolang/Profile.java
```diff
@@ -0,0 +1,41 @@
...
+ * @since 0.62
+ */
+public interface Profile {
+
+    /**
+     * The profile of the program running in this JVM, shared by every object
+     * it creates, the same way {@link Phi#Φ} is shared.
+     */
+    Profile RUNNING = new Counters();
```
Favixx: A constant on the interface makes this eagerly initialised on the first touch of `Profile` and impossible to replace. The `Profile` abstraction exists precisely so there can be another implementation, yet nothing in the codebase can install one - not a no-op for production, not a fake for a test. The comparison to `Phi.Φ` holds for the sharing but not for the consequences: `Φ` is a value the runtime needs, this is instrumentation, and instrumentation usually wants an off switch.

eo-runtime/src/main/java/org/eolang/Profile.java
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Record one dynamic dispatch, a lookup of an attribute by its name.
+     */
+    void dispatch();
+
+    /**
+     * How many objects were born.
+     * @return The number of allocations
+     */
+    long allocations();
```
Favixx, line 35: This interface asks its implementations to do two unrelated things: `allocate`/`dispatch` are called by the runtime on its hottest path and never read anything, while `allocations`/`dispatches` are called once by a reporter and never write. Nothing that records ever reads. Splitting them would let the recording half be an empty method on a no-op implementation without that implementation having to invent return values.

eo-runtime/src/main/java/org/eolang/Counters.java
```diff
@@ -0,0 +1,55 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang;
+
+import java.util.concurrent.atomic.LongAdder;
+
+/**
+ * A profile that keeps its two numbers in a pair of adders.
+ *
+ * <p>The class is thread-safe.</p>
```
Favixx: "The class is thread-safe" is true of each counter on its own, but `allocations()` and `dispatches()` are two independent `sum()` calls, so a caller reading both while the program runs gets two numbers from two different instants. For the one line `Main` prints after dataization that is harmless; the claim as written is broader than that, and this is the class future callers will trust.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -219,6 +222,7 @@ public void put(final String name, final Phi object) {
 
     @Override
     public Phi take(final String name) {
+        Profile.RUNNING.dispatch();
```
Favixx: This is the hottest method in the runtime and it now pays an unconditional `LongAdder.increment()` on every call, with no way to turn it off. `LongAdder` is the right choice if the counting stays, but the question of what it costs is answerable here: the repo already runs a `benchmark` job, so a before/after on it would settle whether this needs a `Profile.NONE` selected by a system property or is simply free.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -28,7 +28,7 @@
  * @since 0.1
  * @checkstyle DesignForExtensionCheck (500 lines)
  */
-@SuppressWarnings("PMD.GodClass")
+@SuppressWarnings({"PMD.GodClass", "PMD.ConstructorOnlyInitializesOrCallOtherConstructors"})
```
Favixx: A class-level suppression for one line in one constructor. `PhDefault` is 500+ lines with six constructors, and this now silences the rule for all of them, including whichever ones get added later. On the private constructor itself it would cover exactly the line that needs it.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -153,6 +153,7 @@ public PhDefault(final byte[] dta, final Map<String, Attribute> attributes) {
      * @param dta        Object data
      * @param attributes Initial attributes to register
      * @checkstyle ParameterNumberCheck (5 lines)
+     * @checkstyle ConstructorsCodeFreeCheck (14 lines)
```
Favixx: This is the second suppression added for the same single statement - PMD above, checkstyle here, both saying that constructors should not do work. When two independent rules object to one line, that is usually the design pushing back rather than two false positives. Recording the birth where objects are actually handed out, rather than inside the constructor, would need neither suppression.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -161,6 +162,7 @@ private PhDefault(
         this.data = dta;
         this.initial = attributes;
         this.order = new HashMap<>(0);
+        Profile.RUNNING.allocate();
```
Favixx: Nothing asserts that this line runs. `CountersTest` exercises `Counters` in isolation and `MainTest` matches any non-zero number, so if this call were deleted the suite would still be green as long as `copy()` kept its own. A test that constructs one `PhDefault` and checks `Profile.RUNNING.allocations()` grew by exactly one covers the part that can actually break - the wiring, not the adder.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -182,6 +184,7 @@ public final Phi copy() {
                 map.put(ent.getKey(), ent.getValue().copy(copy));
             }
             copy.attrs = map;
+            Profile.RUNNING.allocate();
```
Favixx: `copy()` records the allocation after the attribute map is rebuilt, while the constructor records it before the object is usable. If `ent.getValue().copy(copy)` throws, `clone()` has already produced an object that will never be counted. It gets discarded, so nothing breaks - but the two call sites answer "when is an object born" differently, and one sentence in the javadoc saying which one is intended would keep the next reader from moving one of them.

eo-runtime/src/test/java/org/eolang/CountersTest.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Test case for {@link Counters}.
+ * @since 0.62
+ */
+final class CountersTest {
+
+    @Test
+    void countsAllocations() {
+        final long seed = System.nanoTime();
+        final int times = new Random(seed).nextInt(100) + 1;
```
Favixx: The loop count is random, seeded off `System.nanoTime()`, which then has to be threaded into both failure messages so a failure can be reproduced. Nothing here is input-sensitive: `LongAdder` behaves identically at 7 increments and at 63. A fixed count tests exactly as much and removes the seed, the `Random`, and the reason the assertion messages have to mention it.

eo-runtime/src/test/java/org/eolang/CountersTest.java
```diff
@@ -0,0 +1,54 @@
...
+            Matchers.equalTo((long) times)
+        );
+    }
+
+    @Test
+    void countsDispatchesApartFromAllocations() {
+        final long seed = System.nanoTime();
+        final int times = new Random(seed).nextInt(100) + 1;
+        final Profile profile = new Counters();
+        profile.allocate();
```
Favixx: The test is called `countsDispatchesApartFromAllocations` and this `allocate()` is what sets up the "apart from" part - but the assertion below only looks at `dispatches()`. Nothing checks that `allocations()` is still 1, which is the separation the name promises. One more `assertThat` makes the test match its title.

eo-runtime/src/test/java/org/eolang/CountersTest.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+        MatcherAssert.assertThat(
+            String.format(
+                "The profile mixed the dispatches with the allocations, %d of them were recorded, seed is %d",
+                times, seed
+            ),
+            profile.dispatches(),
+            Matchers.equalTo((long) times)
+        );
+    }
```
Favixx, line 47: `Counters` documents itself as thread-safe and `LongAdder` is chosen for exactly that, yet both tests are single-threaded. Since this now sits on the runtime's hottest path and EO programs can dataize concurrently, a test that hammers `allocate()` from several threads and checks the total is the one that justifies the class's central design decision.

eo-runtime/src/test/java/org/eolang/MainTest.java
```diff
@@ -69,6 +70,20 @@ void deliversCleanOutput() {
...
     }
 
+    @Test
+    void printsExecutionProfile() {
+        MatcherAssert.assertThat(
+            "The report of dataizing \"true\" carries no object and dispatch counters",
+            MainTest.stderr("true"),
+            Matchers.matchesPattern(
+                Pattern.compile(
+                    ".+Finished in .+, [1-9][0-9]* objects, [1-9][0-9]* dispatches.*",
```
Favixx: `[1-9][0-9]*` passes for any non-zero count, so this asserts the two numbers are printed and nothing about whether they are right - a version that counted every `take()` twice, or counted allocations into the dispatch adder, would be just as green. With a `Profile` that `Main` accepted rather than reached for, this could name the numbers it expects.

Favixx (commented):
The mechanism is sound and I could not find a hole in the counting itself. All five public `PhDefault` constructors delegate to the private one, so an allocation is recorded exactly once per object; `clone()` really does skip that path, so the extra call in `copy()` is needed rather than double-counting; and every decorator - `PhSafe`, `PhOnce`, `PhNest`, `PhCoverage`, `PhLogged` - delegates `take` down to a real `PhDefault`, including `PhOnce`, whose memoization is of the wrapped object and not of the lookup, so no dispatch goes missing there either. Leaving `PhPackage` out is defensible on the stated grounds.

What I would push on is the seam, and it is one decision showing up in five places.

`Profile.RUNNING` is a constant on an interface, and `PhDefault` and `Main` both reach for it directly. So the interface that exists to allow a second implementation cannot actually take one: there is no no-op to switch the instrumentation off in production, and no double to assert against in a test. That is why `MainTest` has to match `[1-9][0-9]*` rather than a number, why nothing checks that the constructor's `allocate()` call fires at all, and why the cost question on `take()` - an unconditional `LongAdder.increment()` on the runtime's hottest method - has no answer other than measuring the whole JVM. A `Profile` parameter on `Main.run` and a settable or injectable implementation would fix all of it at once, and the repo's `benchmark` job could then say what the counting costs.

The other thing is the report line. "Finished in 0.11s (1 bytes), 13 objects, 5 dispatches" reads as one sentence about one run, but only the elapsed time is a delta - the other two are JVM-wide, as the description says. Two locals beside `start` and two subtractions make the sentence true, which seems better than a caveat that lives only in the PR.

Smaller: the javadoc promises "objects" where the count is `PhDefault` instances (decorators never appear in it); `Counters` claims thread-safety while its two readers cannot produce a consistent pair; two suppressions were needed for one statement in a constructor; and `CountersTest` randomises a loop count that nothing depends on while leaving the concurrency the class is built for untested.

eo-runtime/src/main/java/org/eolang/PhDefault.java
morphqdd: Suppressing PMD.ConstructorOnlyInitializesOrCallOtherConstructors here because the constructor now also calls Profile.RUNNING.allocate(). Is there existing precedent in this class for suppressing over restructuring? If not this tends to come back as review feedback.

eo-runtime/src/test/java/org/eolang/MainTest.java
```diff
@@ -69,6 +70,20 @@ void deliversCleanOutput() {
         );
     }
 
+    @Test
+    void printsExecutionProfile() {
```
morphqdd: This is the only test tying Profile.RUNNING to real Phi behavior, and it's an end-to-end regex check that at least one allocation/dispatch happened during a whole run. Nothing asserts that PhDefault's constructor/copy()/take() each call allocate()/dispatch() exactly where claimed, e.g. that a decorator wrapping a call doesn't double-count.

eo-runtime/src/main/java/org/eolang/PhDefault.java
Thayorns: The precedent is on the class declaration itself: `PhDefault` already carries `@SuppressWarnings("PMD.GodClass")`, plus `@SuppressWarnings("java:S5164")` on `NESTING` and `@checkstyle DesignForExtensionCheck (500 lines)` in the docblock. So adding one more name to the existing array is the established shape here rather than a new habit.

Restructuring is not really open either. Every generated class extends `PhDefault`, and each of the five secondary constructors delegates to the one private primary, so the primary is the only place that sees every object exactly once. Moving the count to a factory would mean the generated code has to call it, and any object built by `new` would go uncounted.

eo-runtime/src/test/java/org/eolang/MainTest.java
Thayorns: Right, the regex only proves something happened. Added `ProfileTest` with exact deltas around each call site: one constructor moves `allocations()` by exactly 1, one `copy()` by exactly 1, one `take()` moves `dispatches()` by exactly 1. Those are the numbers that catch double-counting, which the end-to-end check cannot see.

It is a separate class because `Profile.RUNNING` is a JVM-wide singleton and this module runs tests concurrently, so a delta reading needs `@Isolated` — and that annotation is class-level only. Verified it holds: full `eo-runtime` suite green with `ProfileTest` passing alongside everything else.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -159,6 +159,7 @@ private PhDefault(
         this.data = dta;
         this.initial = attributes;
         this.order = new HashMap<>(0);
+        Profile.RUNNING.allocate();
```
yegor256: @Thayorns I suggest you get rid of this line - PMD will be happy and no functionality will be lost

eo-runtime/src/main/java/org/eolang/Profile.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang;
```
yegor256, line 5: @Thayorns it's a singleton, and it's evil by definition: https://www.yegor256.com/2016/06/27/singletons-must-die.html

Instead, I suggest renaming `Profile` to `Statistics` and then making an instance of it in the `Main` and passing as argument to `PhDefault` - we will  have to modify its interface.

eo-runtime/src/main/java/org/eolang/PhDefault.java
Thayorns: Done, the line is gone from the constructor and qulice is green again.

The count moved to `loaded()`, which is where an object materialises its attributes. That runs exactly once per object, guarded by `attrs == null`, and it is an ordinary method, so neither PMD nor checkstyle objects and both suppressions are now removed.

One number did change: a run of `true` reports 12 objects instead of 13. The difference is an object that gets constructed but never materialises any attribute, so it never takes part in dataization. `copy()` keeps its own call, since it assigns `attrs` directly and never goes through `loaded()`.

eo-runtime/src/main/java/org/eolang/Profile.java
Thayorns: Renamed to `Statistics` (and `ProfileTest` to `StatisticsTest`).

The injection I have not done, because I do not see how `Main` can reach the constructors, and I would rather ask than guess at the scope. `Main` never builds a `PhDefault`: it calls `Phi.Φ.take(obj)`, and from there every object is created by a bare `new` inside a generated class or a hand-written atom. So for an instance made in `Main` to arrive at those constructors, it has to be threaded through all of them: 36 hand-written subclasses, plus `to-java.xsl`, which emits `extends PhDefault` for every transpiled object. And the entry point stays `Phi.Φ`, a static constant, so the last hop into the graph has no argument to carry it on.

That is a transpiler change, not a runtime one, and it is bigger than this PR. Three ways I can see:

1. Thread `Statistics` through `PhDefault`'s constructors and `to-java.xsl`, as you describe — a separate PR, and I am happy to take it.
2. Give `PhDefault` a `Statistics` argument that defaults to the shared one, so only the code that cares passes it. Removes the hard-coding but not the global.
3. Leave the constant here and open a puzzle for the injection.

Which do you prefer? I have left the constant in place for now so the rest of the PR can be judged; say the word and I will do 1.

yegor256:
@Thayorns indeed, we should thread `Statistics` through `PhDefault`'s constructors and `to-java.xsl`. let's do it here.

Thayorns:
@yegor256 have a look

Thayorns:
@yegor256 have a look

yegor256:

Thayorns:
@yegor256 CI fixed

Thayorns:
@yegor256 ping here

Thayorns:
@yegor256 CI green here

Thayorns:
@yegor256 ping

Thayorns:

Thayorns:
@yegor256 fixed here

eo-runtime/src/main/java/org/eolang/Statistics.java
```diff
@@ -0,0 +1,50 @@
...
+ *  off. Give {@link PhDefault} a Statistics of its own, thread it through the
+ *  constructors and through to-java.xsl, and let every object report to the
+ *  program that made it.
+ */
+public interface Statistics {
+
+    /**
+     * Statistics of the program running in this JVM.
+     */
+    Statistics PROGRAM = new Counters();
```
yegor256: @Thayorns singletons are evil. don't use them.

Thayorns:

0crat:
@favixx Thanks for the review! You've earned +20 points for this: +12 as a basis; +7 for the 782 hits-of-code that you reviewed.  Your running score is [+269](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +15 points for this: +12 as a basis; +7 for the 782 hits-of-code that you reviewed; -5 for very few (2) comments.  Your running score is [+4954](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +15 points for this: +12 as a basis; +7 for the 782 hits-of-code that you reviewed; -5 for very few (1) comments.  Your running score is [+2593](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Thayorns:

Thayorns:

asmirnov-backend (commented):
Read this as a change to the contract between the transpiler and the runtime rather than as instrumentation - threading `Statistics` through every constructor is what makes the counters honest, and moving the birth into `loaded()` reads right. Four things below: a number in `Main` that quietly changed meaning, reflective loading that now demands a signature nothing guards, the asymmetric visibility of the generated test constructor, and the resource parsing that moved into `AtomTypes`.

eo-runtime/src/main/java/org/eolang/Main.java
```diff
@@ -224,6 +224,10 @@ private static void run(final List<String> opts) {
                 "The name of the object is an empty string, why?"
             );
         }
+        final Statistics stats = Phi.Φ.statistics();
+        final long born = stats.allocations();
+        final long taken = stats.dispatches();
+        final long start = System.currentTimeMillis();
```
asmirnov-backend, line 230: `start` moved up from just below the argument block, so the reported duration now covers `Phi.Φ.take(obj)` and the argument tuple as well as the dataization. That is not a small difference on a short run: `take` on `Φ` goes through `PhPackage.loadPhi`, which is `Class.forName` plus reflection, and for a program like `true` that resolution is most of the wall clock, so the same command now prints a visibly larger number than it did before this PR without anything having become slower.

The counters genuinely needed a "before" reading here; the clock did not. Taking only `born` and `taken` up here and leaving `final long start` where it was keeps the three numbers in one sentence and keeps the first of them comparable across versions. If the wider window is deliberate, the message is the place to say so, because "Finished in" currently reads as the time to finish the program.

eo-runtime/src/main/java/org/eolang/PhNest.java
```diff
@@ -207,9 +219,12 @@ private Phi object() {
      * @param target The fully-qualified Java name
      * @return The object
      */
-    private static Phi load(final String target) {
+    private Phi load(final String target) {
         try {
-            return Class.forName(target).asSubclass(Phi.class).getConstructor().newInstance();
+            return Class.forName(target)
+                .asSubclass(Phi.class)
+                .getConstructor(Statistics.class)
```
asmirnov-backend, line 226: This makes the constructor signature part of the contract between the transpiler and the runtime, and nothing checks that the two sides agree. A class transpiled by an earlier eo-maven-plugin has only `Foo()`, so loading it here throws `NoSuchMethodException` and reaches the user as "Can't instantiate the object '...'" - a message that names the class and hides the reason, sending the reader to look at the object rather than at the version mismatch. `PhPackage.loadPhi` has the same shape.

Two cheap ways out: fall back to `getConstructor()` when the one-argument form is absent, so an older class still loads and simply counts into a `Silent`; or keep the hard failure and say in the message that the class predates `Statistics` and needs recompiling. The second fits the fail-fast habit here, but either beats the message as it stands.

eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
```diff
@@ -821,6 +829,23 @@
...
+    <xsl:value-of select="eo:eol(1)"/>
+    <xsl:text>/**</xsl:text>
+    <xsl:value-of select="eo:eol(1)"/>
+    <xsl:text> * Ctor.</xsl:text>
+    <xsl:value-of select="eo:eol(1)"/>
+    <xsl:text> * @param stats Where this object reports its birth</xsl:text>
+    <xsl:value-of select="eo:eol(1)"/>
+    <xsl:text> */</xsl:text>
+    <xsl:value-of select="eo:eol(1)"/>
+    <xsl:text>private </xsl:text>
```
asmirnov-backend, line 846: Test classes get the `Statistics` constructor as `private`, while the `ctors` template emits the same constructor `public` for every other generated class. Two consequences worth confirming as intended.

Nothing outside the class can hand a test object the statistics of the program it runs in, so everything an EO test allocates is invisible by construction - including the `.eo` tests of `eo-runtime` itself. And `getConstructor(Statistics.class)` in `PhNest.load` and `PhPackage.loadPhi` only sees public constructors, so a test object resolved by name through a package would not quietly fall back to the public no-arg one, it would fail outright.

If the intent is just "tests do not count", the public no-arg constructor already delivers exactly that by passing `new Silent()`, and this one could be public like the rest.

eo-runtime/src/main/java/org/eolang/AtomTypes.java
```diff
@@ -24,6 +31,14 @@ public final class AtomTypes {
      */
     private final Map<String, String> table;
 
+    /**
+     * Ctor that loads the table generated next to the given class.
+     * @param owner Class whose package holds the generated {@code atoms.csv}
+     */
+    public AtomTypes(final Class<?> owner) {
```
asmirnov-backend, line 38: This constructor reads a resource stream and parses it. That is the same "constructors should only assign" objection that took `allocate()` out of `PhDefault`'s constructor earlier in this PR; here it is I/O rather than a counter increment, so PMD stays quiet, but the object can now throw while being built. Since the only caller is `private static final AtomTypes ATOMS = PhDefault.atoms()`, a throw arrives as `ExceptionInInitializerError` during class initialisation of `PhDefault`, with the `ExFailure` buried as a cause - before any EO program has started.

The parsing underneath fails obscurely too. `Collectors.toMap` with two mappers throws `IllegalStateException` on a duplicate key, so two rows carrying the same forma abort the runtime with a message naming neither the forma nor the file. `MjAtomsTable` builds this table by walking the parsed XMIR of a whole project, so whether a forma can repeat there is worth a moment rather than an assumption. A merge function that keeps the first row, and a message naming the offending forma, would turn a stack trace into a diagnosis.

yegor256:
The CI check **pr-size** is failing on this PR. The PR is too large (65 changed files, 649 additions, 133 deletions, 29 commits). Please consider splitting it into smaller, more focused pull requests that are easier to review.

---

PR #6207, opened 2026-07-31
#5978: advance past a zero-width match when iterating

### LINKED ISSUE #5978: `string.replaced` method fails with `ExFailure` for regex patterns that can match an empty string

### LINKED ISSUE #6143: `string.regex.match.next` stuck in an infinite loop for terminal zero-width matches
## What happens

`string.regex.pattern.match.next` never reaches a non-existent block after a zero-width match at the end of text. It keeps returning the same match with a new `position`.

A runtime probe iterated three blocks from `regex "/$/"` matched against `"x"`:

| block | `position` | `from` | `to` |
| --- | ---: | ---: | ---: |
| first | 1 | 1 | 1 |
| second (`first.next`) | 2 | 1 | 1 |
| third (`second.next`) | 3 | 1 | 1 |

`$` has exactly one match in this text, at its end. The repeated `[1, 1]` blocks mean that client code which follows `.next` until `exists` becomes false does not terminate.

## Why this is a bug

[`matched.next`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/regex.eo#L33-L54) constructs the next block with the previous `to` offset as its new search start (lines 39-43). For a zero-width match, `from == to`; for `/$/` that offset is the text length. The Java matcher is recreated for every call, so searching from that same offset finds the same end anchor again.

The same progression problem affects zero-width patterns such as `/^/`, `/a*/`, and lookarounds. It is related to, but distinct from, #5978: that issue concerns replacement recursion, while this one makes the public match iterator repeat a single matched block forever.

## Steps to reproduce

1. Compile `string.regex "/$/"`.
2. Call `compiled.match "x".next` to get the first block.
3. Read `first.next`, then `first.next.next`.
4. Observe that every block exists and has `from = to = 1` instead of eventually reporting no match.

## Expected behavior

Each zero-width occurrence must be visited at most once. After the terminal `$` match, `.next` should return a non-existent matched block (`exists = false`). For a zero-width match before the end of the text, the search position must advance by one character before invoking the next matcher, while a terminal zero-width match must stop without requesting an out-of-range start.

morphqdd:
Closes #5978
Closes #6143

The two issues are one bug. `replaced` never terminates for a nullable pattern because `matched.next` does not move: it searches from the previous `to`, and for a zero-width match `from == to`, so the same match comes back forever and `rec-replaced` recurses until it dies. #6143 is the same loop seen directly, through the iterator rather than through `replaced`.

`next` now starts one character further on when the match was zero-width, and reports a non-existent block when a zero-width match sits at the end of the text and there is nowhere further to go.

That last part is why the change is not a one-liner. `matched-from-index` refuses a `start` past the text length, and `rejectsStartIndexAfterTextEnd` pins that, so the terminal case cannot be expressed by advancing the index; the block is built directly instead, with `start` at -1 and terminators in `from`, `to` and `groups`, mirroring what the atom's own `blank` does.

Nothing in `replaced.eo` changed. With the iterator advancing, its existing algorithm produces what Java does: `"abc"` with `/x*/` gives `-a-b-c-`, matching `String.replaceAll`.

Two EO tests, both failing on master: `cannot-iterate-past-a-terminal-zero-width-match` for the iterator, `can-replace-a-nullable-pattern-like-java` for the substitution. On master the second does not merely fail, it takes the JUnit container down, which is what unbounded recursion looks like from outside.

### DIFF
--- a/eo-runtime/src/main/eo/string/replaced.eo
+++ b/eo-runtime/src/main/eo/string/replaced.eo
@@ -35,6 +35,13 @@
         next. >> matched
           target.match reinit
 
+  eq. ++> can-replace-a-nullable-pattern-like-java
+    replaced
+      "abc"
+      regex "/x*/"
+      "-"
+    "-a-b-c-"
+
   eq. ++> can-replace-a-windows-separator-with-a-slash
     replaced
       "C:\\Windows\\Users\\AppLocal\\shrek"


eo-runtime/src/main/eo/string/regex.eo
```diff
@@ -37,10 +37,26 @@
         groups.at index > [index] > group
         $ > matched
         exists.if > next
-          matched.
-            matched-from-index
+          if.
+            and.
+              from.eq to
+              to.gte txt.length
+            ^.matched
```
Thayorns: The -1 and those three causes are word for word what `blank()` puts into the block in EOregex$EOpattern$EOmatch$EOmatched_from_index.java. Two definitions of an exhausted iterator, in two languages, with nothing tying them together — reword a cause on the Java side and this branch drifts in silence.

Thayorns (commented):
`next` now steps one character past a zero-width match and returns a non-existent block when that match sits at the end of the text, which is what stops `replaced` from recursing forever. I traced `/x*/` over "abc" and it does come out as `-a-b-c-`. The terminal case is pinned by a test; the advancing case is only exercised indirectly through `replaced`.

eo-runtime/src/main/eo/string/replaced.eo
```diff
@@ -35,6 +35,13 @@
         next. >> matched
           target.match reinit
 
+  eq. ++> can-replace-a-nullable-pattern-like-java
```
gemshrine, line 38: Good that this pins Java parity, but `/x*/` on `"abc"` matches empty at *every* position, so it only exercises the zero-width branch, never the transition from a zero-width match to a following non-empty one — which is exactly what the new `if. (from.eq to) (to.plus 1) to` in `regex.eo` handles. I verified a mixed pattern is Java-consistent: `replaced "baab" (regex "/a*/") "-"` -> `"-b--b-"` and `replaced "xabcx" (regex "/x*/") "-"` -> `"--a-b-c--"`, both equal to `String.replaceAll`. One mixed case here would pin the interleave that the uniform-empty pattern can't reach.

eo-runtime/src/main/eo/string/regex.eo
gemshrine: This terminal branch is load-bearing but reads as arbitrary: it exists because at end-of-text `to.plus 1` would search from beyond `txt.length` and either re-find the same zero-width match (the original loop) or run out of range, so the iterator has to hand back a non-existent block instead. A one-line comment saying that — why the `and. (from.eq to) (to.gte txt.length)` case can't just fall through to the `to.plus 1` bump below — would stop a later "simplification" from collapsing the two branches and reintroducing #5978.

eo-runtime/src/main/eo/string/regex.eo
```diff
@@ -37,10 +37,26 @@
         groups.at index > [index] > group
         $ > matched
         exists.if > next
-          matched.
-            matched-from-index
+          if.
+            and.
+              from.eq to
+              to.gte txt.length
```
gemshrine: Both this `to.gte txt.length` and the `to.plus 1` bump assume `to` and `txt.length` are in the same units. They agree today — I confirmed `replaced "café" (regex "/x*/") "-"` -> `"-c-a-f-é-"`, identical to Java, so multibyte zero-width iteration is already correct. But #6062 is reworking match positions to count Unicode characters, and if `to` moves to character units while `length` stays as-is (or vice versa) this comparison and the `+1` bump would drift. A multibyte nullable-pattern test (the `café` case above) added here would catch that when #6062 lands.

gemshrine (commented):
The fix is correct and matches Java. I dataized `replaced` against `String.replaceAll` over a lattice on this branch — non-empty control, the uniform-empty case, mixed empty/non-empty (`"baab"` `/a*/` -> `"-b--b-"`, `"xabcx"` `/x*/` -> `"--a-b-c--"`), anchors `/^/` `/$/`, empty text, and multibyte `"café"` `/x*/` -> `"-c-a-f-é-"` — every one equals Java. Three notes, all about pinning what already works rather than defects: the interleave case isn't in the tests, the terminal special-case is undocumented, and the multibyte behavior is worth a guard given #6062.

yegor256:
@morphqdd conflict here

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (7) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3165](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (7) hits-of-code; +1 to give you at least something.  Your running score is [+601](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (3) comments; -4 for too few (7) hits-of-code; +1 to give you at least something.  Your running score is [+1347](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6208, opened 2026-07-31
#6137: split Winsock and CStdLib syscall tests out of SyscallTest.java

### LINKED ISSUE #6137: `SyscallTest.java` exceeds line limit due to duplicated Windows and POSIX test cases

morphqdd:
Closes #6137

SyscallTest.java sat four lines below the 1000-line FileLengthCheck ceiling, with no room for the next test that touches sockets - #6133 hit exactly that.

The length came from WindowsSocketTest and PosixSocketTest, the same seven cases written twice over, differing only in whether they reach Winsock.INSTANCE or CStdLib.INSTANCE. Moved each half next to the production class it drives: WinsockTest sits beside Winsock.java, CStdLibTest beside CStdLib.java. That satisfies jtcop's rule that a test class must sit next to a real production class, the same rule that turned EOsocketTest into SyscallTest in the first place.

RandomServer and RandomPort become small shared fixtures used by both halves and by SyscallTest. SyscallTest now holds only the cases that drive the EO socket object, at 248 lines.

Updated the simian exclude to name the two new files instead of SyscallTest.java, matching the intent behind the earlier #6113 fix.

This PR is larger than the 200-line guideline. It is a pure code relocation with no new logic - each of the two extracted test classes is inherently 300+ lines by itself (the pre-existing duplicated test bodies), so a single self-contained move can't fit under the cap without leaving temporary duplicate code across two PRs. Verified locally: all tests pass (SyscallTest, WinsockTest on POSIX-skip, CStdLibTest all green), qulice clean.

### DIFF
--- a/.github/workflows/simian.yml
+++ b/.github/workflows/simian.yml
@@ -21,4 +21,4 @@ jobs:
           distribution: 'temurin'
           java-version: 17
       - run: wget --quiet https://public.yegor256.com/simian.jar -O /tmp/simian.jar
-      - run: java -jar /tmp/simian.jar -threshold=15 "-excludes=**/WinsockTest.java" "-excludes=**/CStdLibTest.java" "-excludes=**/gen" "-excludes=**/it" "**/*.java"
+      - run: java -jar /tmp/simian.jar -threshold=15 "-excludes=**/gen" "-excludes=**/it" "**/*.java"
--- a/.github/workflows/simian.yml
+++ b/.github/workflows/simian.yml
@@ -21,4 +21,4 @@ jobs:
           distribution: 'temurin'
           java-version: 17
       - run: wget --quiet https://public.yegor256.com/simian.jar -O /tmp/simian.jar
-      - run: java -jar /tmp/simian.jar -threshold=15 "-excludes=**/SyscallTest.java" "-excludes=**/gen" "-excludes=**/it" "**/*.java"
+      - run: java -jar /tmp/simian.jar -threshold=15 "-excludes=**/WinsockTest.java" "-excludes=**/CStdLibTest.java" "-excludes=**/gen" "-excludes=**/it" "**/*.java"
--- a/eo-runtime/src/test/java/org/eolang/RandomPort.java
+++ b/eo-runtime/src/test/java/org/eolang/RandomPort.java
@@ -1,0 +1,28 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang;
+
+import java.util.Random;
+
+/**
+ * A random port number, shared by socket syscall tests.
+ * @since 0.40.0
+ */
+public final class RandomPort {
+
+    /**
+     * Random number generator.
+     */
+    private final Random random = new Random();
+
+    /**
+     * Pick a random port.
+     * @return Random port
+     */
+    public int pick() {
+        final int min = 10_000;
+        return this.random.nextInt(20_000 - min + 1) + min;
+    }
+}
--- a/eo-runtime/src/test/java/org/eolang/RandomPort.java
+++ b/eo-runtime/src/test/java/org/eolang/RandomPort.java
@@ -4,25 +4,33 @@
  */
 package org.eolang;
 
-import java.util.Random;
+import com.jcabi.log.Logger;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.net.ServerSocket;
 
 /**
- * A random port number, shared by socket syscall tests.
+ * A free port number, shared by socket syscall tests.
+ *
+ * <p>Asks the kernel for an ephemeral port and releases it right away, so
+ * the caller gets a real currently-free port instead of a guess from a
+ * fixed range.</p>
+ *
  * @since 0.40.0
  */
 public final class RandomPort {
 
     /**
-     * Random number generator.
-     */
-    private final Random random = new Random();
-
-    /**
-     * Pick a random port.
-     * @return Random port
+     * Pick a free port.
+     * @return Free port
      */
     public int pick() {
-        final int min = 10_000;
-        return this.random.nextInt(20_000 - min + 1) + min;
+        try (ServerSocket socket = new ServerSocket(0)) {
+            final int port = socket.getLocalPort();
+            Logger.debug(this, "Picked free port %d", port);
+            return port;
+        } catch (final IOException exception) {
+            throw new UncheckedIOException(exception);
+        }
     }
 }
--- a/eo-runtime/src/test/java/org/eolang/RandomServer.java
+++ b/eo-runtime/src/test/java/org/eolang/RandomServer.java
@@ -1,0 +1,66 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang;
+
+import com.jcabi.log.Logger;
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.ServerSocket;
+
+/**
+ * Server on random port, shared by socket syscall tests.
+ * @since 0.40.0
+ */
+@SuppressWarnings("PMD.AvoidUsingHardCodedIP")
+public final class RandomServer {
+
+    /**
+     * Server socket.
+     */
+    private ServerSocket socket;
+
+    /**
+     * Port.
+     */
+    private int port;
+
+    /**
+     * Start server on random port.
+     * @return Self
+     */
+    public RandomServer started() {
+        boolean bound = false;
+        while (!bound) {
+            this.port = new RandomPort().pick();
+            try {
+                this.socket = new ServerSocket();
+                this.socket.setReuseAddress(true);
+                this.socket.bind(new InetSocketAddress("127.0.0.1", this.port));
+                bound = true;
+                Logger.debug(this, "Server started on port %d", this.port);
+            } catch (final IOException exception) {
+                Logger.debug(this, "Port %d is unavailable, trying another port...", this.port);
+            }
+        }
+        return this;
+    }
+
+    /**
+     * Close server socket.
+     */
+    public void stop() throws IOException {
+        if (this.socket != null && !this.socket.isClosed()) {
+            this.socket.close();
+        }
+    }
+
+    /**
+     * The port the server is bound to.
+     * @return Port
+     */
+    public int port() {
+        return this.port;
+    }
+}
--- a/eo-runtime/src/test/java/org/eolang/RandomServer.java
+++ b/eo-runtime/src/test/java/org/eolang/RandomServer.java
@@ -6,6 +6,7 @@ package org.eolang;
 
 import com.jcabi.log.Logger;
 import java.io.IOException;
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 
@@ -18,7 +19,6 @@ import java.net.ServerSocket;
  *
  * @since 0.40.0
  */
-@SuppressWarnings("PMD.AvoidUsingHardCodedIP")
 public final class RandomServer implements AutoCloseable {
 
     /**
@@ -57,7 +57,7 @@ public final class RandomServer implements AutoCloseable {
         final ServerSocket opened = new ServerSocket();
         try {
             opened.setReuseAddress(true);
-            opened.bind(new InetSocketAddress("127.0.0.1", 0));
+            opened.bind(new InetSocketAddress(InetAddress.getLoopbackAddress(), 0));
         } catch (final IOException exception) {
             opened.close();
             throw exception;
--- a/eo-runtime/src/test/java/org/eolang/RandomServer.java
+++ b/eo-runtime/src/test/java/org/eolang/RandomServer.java
@@ -55,8 +55,13 @@ public final class RandomServer implements AutoCloseable {
      */
     private ServerSocket bound() throws IOException {
         final ServerSocket opened = new ServerSocket();
-        opened.setReuseAddress(true);
-        opened.bind(new InetSocketAddress("127.0.0.1", 0));
+        try {
+            opened.setReuseAddress(true);
+            opened.bind(new InetSocketAddress("127.0.0.1", 0));
+        } catch (final IOException exception) {
+            opened.close();
+            throw exception;
+        }
         Logger.debug(this, "Server bound on port %d", opened.getLocalPort());
         return opened;
     }
--- a/eo-runtime/src/test/java/org/eolang/RandomServer.java
+++ b/eo-runtime/src/test/java/org/eolang/RandomServer.java
@@ -10,48 +10,42 @@ import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 
 /**
- * Server on random port, shared by socket syscall tests.
+ * Server on a free port, shared by socket syscall tests.
+ *
+ * <p>Picks a port from a fixed, low range rather than asking the kernel
+ * for an ephemeral one, because {@code socket.eo}'s {@code htons} only
+ * supports ports up to 32767 until #6133 lands; the kernel's own
+ * ephemeral range starts above that limit.</p>
+ *
  * @since 0.40.0
  */
 @SuppressWarnings("PMD.AvoidUsingHardCodedIP")
 public final class RandomServer {
 
     /**
-     * Server socket.
+     * Maximum bind attempts before giving up.
      */
-    private ServerSocket socket;
+    private static final int MAX_ATTEMPTS = 50;
 
     /**
-     * Port.
+     * Server socket, bound the moment this object is built.
      */
-    private int port;
+    private final ServerSocket socket;
 
     /**
-     * Start server on random port.
-     * @return Self
+     * Ctor: binds to a free port on localhost, trying up to
+     * {@link RandomServer#MAX_ATTEMPTS} candidates before giving up.
+     * @checkstyle ConstructorsCodeFreeCheck (5 lines)
      */
-    public RandomServer started() {
-        boolean bound = false;
-        while (!bound) {
-            this.port = new RandomPort().pick();
-            try {
-                this.socket = new ServerSocket();
-                this.socket.setReuseAddress(true);
-                this.socket.bind(new InetSocketAddress("127.0.0.1", this.port));
-                bound = true;
-                Logger.debug(this, "Server started on port %d", this.port);
-            } catch (final IOException exception) {
-                Logger.debug(this, "Port %d is unavailable, trying another port...", this.port);
-            }
-        }
-        return this;
+    public RandomServer() throws IOException {
+        this.socket = this.bound();
     }
 
     /**
      * Close server socket.
      */
     public void stop() throws IOException {
-        if (this.socket != null && !this.socket.isClosed()) {
+        if (!this.socket.isClosed()) {
             this.socket.close();
         }
     }
@@ -61,6 +55,36 @@ public final class RandomServer {
      * @return Port
      */
     public int port() {
-        return this.port;
+        return this.socket.getLocalPort();
+    }
+
+    /**
+     * Bind a server socket to a free port, retrying a bounded number of
+     * times so a persistent bind failure fails the test instead of
+     * hanging the build forever.
+     * @return Bound server socket
+     */
+    private ServerSocket bound() throws IOException {
+        IOException last = null;
+        for (int attempt = 0; attempt < RandomServer.MAX_ATTEMPTS; attempt = attempt + 1) {
+            final int candidate = new RandomPort().pick();
+            try {
+                final ServerSocket opened = new ServerSocket();
+                opened.setReuseAddress(true);
+                opened.bind(new InetSocketAddress("127.0.0.1", candidate));
+                Logger.debug(this, "Server bound on port %d", candidate);
+                return opened;
+            } catch (final IOException exception) {
+                Logger.debug(this, "Port %d is unavailable, trying another port...", candidate);
+                last = exception;
+            }
+        }
+        throw new IOException(
+            String.format(
+                "Could not bind to a free port after %d attempts",
+                RandomServer.MAX_ATTEMPTS
+            ),
+            last
+        );
     }
 }
--- a/eo-runtime/src/test/java/org/eolang/RandomServer.java
+++ b/eo-runtime/src/test/java/org/eolang/RandomServer.java
@@ -12,20 +12,14 @@ import java.net.ServerSocket;
 /**
  * Server on a free port, shared by socket syscall tests.
  *
- * <p>Picks a port from a fixed, low range rather than asking the kernel
- * for an ephemeral one, because {@code socket.eo}'s {@code htons} only
- * supports ports up to 32767 until #6133 lands; the kernel's own
- * ephemeral range starts above that limit.</p>
+ * <p>Asks the kernel for an ephemeral port (bind to port 0) instead of
+ * guessing a candidate from a fixed range, now that {@code socket.eo}'s
+ * {@code htons} covers the full 0-65535 range (#6133).</p>
  *
  * @since 0.40.0
  */
 @SuppressWarnings("PMD.AvoidUsingHardCodedIP")
-public final class RandomServer {
-
-    /**
-     * Maximum bind attempts before giving up.
-     */
-    private static final int MAX_ATTEMPTS = 50;
+public final class RandomServer implements AutoCloseable {
 
     /**
      * Server socket, bound the moment this object is built.
@@ -33,18 +27,15 @@ public final class RandomServer {
     private final ServerSocket socket;
 
     /**
-     * Ctor: binds to a free port on localhost, trying up to
-     * {@link RandomServer#MAX_ATTEMPTS} candidates before giving up.
+     * Ctor: binds to a kernel-assigned free port on localhost.
      * @checkstyle ConstructorsCodeFreeCheck (5 lines)
      */
     public RandomServer() throws IOException {
         this.socket = this.bound();
     }
 
-    /**
-     * Close server socket.
-     */
-    public void stop() throws IOException {
+    @Override
+    public void close() throws IOException {
         if (!this.socket.isClosed()) {
             this.socket.close();
         }
@@ -59,32 +50,14 @@ public final class RandomServer {
     }
 
     /**
-     * Bind a server socket to a free port, retrying a bounded number of
-     * times so a persistent bind failure fails the test instead of
-     * hanging the build forever.
+     * Bind a server socket to a kernel-assigned free port.
      * @return Bound server socket
      */
     private ServerSocket bound() throws IOException {
-        IOException last = null;
-        for (int attempt = 0; attempt < RandomServer.MAX_ATTEMPTS; attempt = attempt + 1) {
-            final int candidate = new RandomPort().pick();
-            try {
-                final ServerSocket opened = new ServerSocket();
-                opened.setReuseAddress(true);
-                opened.bind(new InetSocketAddress("127.0.0.1", candidate));
-                Logger.debug(this, "Server bound on port %d", candidate);
-                return opened;
-            } catch (final IOException exception) {
-                Logger.debug(this, "Port %d is unavailable, trying another port...", candidate);
-                last = exception;
-            }
-        }
-        throw new IOException(
-            String.format(
-                "Could not bind to a free port after %d attempts",
-                RandomServer.MAX_ATTEMPTS
-            ),
-            last
-        );
+        final ServerSocket opened = new ServerSocket();
+        opened.setReuseAddress(true);
+        opened.bind(new InetSocketAddress("127.0.0.1", 0));
+        Logger.debug(this, "Server bound on port %d", opened.getLocalPort());
+        return opened;
     }
 }
--- a/eo-runtime/src/test/java/org/eolang/posix/CStdLibTest.java
+++ b/eo-runtime/src/test/java/org/eolang/posix/CStdLibTest.java
@@ -11,14 +11,13 @@ import io.github.artsok.RepeatedIfExceptionsTest;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import org.eolang.NetworkPort;
 import org.eolang.RandomPort;
 import org.eolang.RandomServer;
 import org.eolang.ReceivedBytes;
+import org.eolang.ServerHandoff;
 import org.eolang.SockaddrIn;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
@@ -139,17 +138,16 @@ final class CStdLibTest {
     void acceptsConnectionOnSocket() throws InterruptedException {
         final AtomicInteger accept = new AtomicInteger(0);
         final AtomicReference<String> error = new AtomicReference<>();
-        final AtomicInteger port = new AtomicInteger(new RandomPort().pick());
-        final CountDownLatch listening = new CountDownLatch(1);
+        final ServerHandoff handoff = new ServerHandoff(new RandomPort().pick());
         final Thread server = new Thread(
-            () -> this.acceptViaCStdLib(port, accept, error, listening)
+            () -> this.acceptViaCStdLib(handoff, accept, error)
         );
         server.start();
-        this.ensure(listening.await(CStdLibTest.LISTEN_MILLIS, TimeUnit.MILLISECONDS));
+        this.ensure(handoff.awaited(CStdLibTest.LISTEN_MILLIS));
         final int client = this.openSocket();
         try {
             this.ensure(client > 0);
-            final SockaddrIn sockaddr = this.sockaddr(port.get());
+            final SockaddrIn sockaddr = this.sockaddr(handoff.port().get());
             MatcherAssert.assertThat(
                 String.format(
                     "Socket should have been connected to local server on sockets, but it didn't, reason: %s",
@@ -176,17 +174,16 @@ final class CStdLibTest {
     void sendsAndReceivesMessagesViaSyscalls() throws InterruptedException {
         final AtomicInteger received = new AtomicInteger(-1);
         final AtomicReference<byte[]> bytes = new AtomicReference<>();
-        final AtomicInteger port = new AtomicInteger(new RandomPort().pick());
-        final CountDownLatch listening = new CountDownLatch(1);
+        final ServerHandoff handoff = new ServerHandoff(new RandomPort().pick());
         final Thread server = new Thread(
-            () -> this.recvViaCStdLib(port, received, bytes, listening)
+            () -> this.recvViaCStdLib(handoff, received, bytes)
         );
         server.start();
-        this.ensure(listening.await(CStdLibTest.LISTEN_MILLIS, TimeUnit.MILLISECONDS));
+        this.ensure(handoff.awaited(CStdLibTest.LISTEN_MILLIS));
         final int client = this.openSocket();
         try {
             this.ensure(client > 0);
-            final SockaddrIn sockaddr = this.sockaddr(port.get());
+            final SockaddrIn sockaddr = this.sockaddr(handoff.port().get());
             this.ensure(CStdLib.INSTANCE.connect(client, sockaddr, sockaddr.size()) == 0);
             final byte[] buf = "Hello, Socket!".getBytes(StandardCharsets.UTF_8);
             final int sent = CStdLib.INSTANCE.send(client, buf, buf.length, 0);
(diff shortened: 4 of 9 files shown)


eo-runtime/src/test/java/org/eolang/RandomServer.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    private int port;
+
+    /**
+     * Start server on random port.
+     * @return Self
+     */
+    public RandomServer started() {
+        boolean bound = false;
+        while (!bound) {
```
Thayorns: This loop has no bound. Every IOException from the bind on line 40 is logged and the loop picks another number, so when binding cannot succeed for a reason that a different port will not cure - no permission to bind on the loopback, the whole 10000-20000 window filtered by a sandbox - started() spins forever and the test that called it never fails and never returns. Three tests in SyscallTest plus both extracted classes now enter this loop, so a build that hits it hangs until the CI job is killed rather than reporting anything. Asking the kernel for an ephemeral port is the one bind that cannot need a retry at all.

eo-runtime/src/test/java/org/eolang/RandomPort.java
```diff
@@ -0,0 +1,28 @@
...
+     */
+    private final Random random = new Random();
+
+    /**
+     * Pick a random port.
+     * @return Random port
+     */
+    public int pick() {
+        final int min = 10_000;
+        return this.random.nextInt(20_000 - min + 1) + min;
```
Thayorns: The port comes from an unseeded Random and then never reaches a failure message. bindsSocketSuccessfullyViaSyscall in CStdLibTest.java:76-83 feeds this value straight into bindSocket and, when the assertion fails, prints only strerror - not the number that was tried. Whoever reads that red build cannot tell whether the port was already taken by another process on the agent or whether bind is genuinely broken, and cannot reproduce the run because the draw is not recorded anywhere.

eo-runtime/src/test/java/org/eolang/posix/CStdLibTest.java
```diff
@@ -0,0 +1,343 @@
...
+            accepted = CStdLib.INSTANCE.accept(
+                socket, addr, new IntByReference(addr.size())
+            );
+            Logger.debug(this, "Accepted socket: %d", accepted);
+            this.ensure(accepted > 0);
+            final byte[] buf = new byte[1024];
+            received.set(CStdLib.INSTANCE.recv(accepted, buf, buf.length, 0));
+            bytes.set(Arrays.copyOf(buf, received.get()));
+        } finally {
+            this.closeSocket(accepted);
```
Thayorns: accepted is still 0 here on every path that leaves before line 330. this.ensure on line 328 throws an AssertionError when listen fails - surefire runs with assertions on - and the finally then calls close(0), which closes the forked JVM's stdin, the channel surefire talks to the fork over. The real failure surfaces as a fork that terminated without saying goodbye. The sibling method guards this on line 310 with if (accept.get() > 0), and WinsockTest.java:400 has the same unguarded shape.

eo-runtime/src/test/java/org/eolang/RandomServer.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Port.
+     */
+    private int port;
+
+    /**
+     * Start server on random port.
+     * @return Self
+     */
+    public RandomServer started() {
```
Thayorns: started() assigns both fields and then returns this, so a RandomServer is a half-built thing until somebody calls a second method on it, and neither field can be final. Line 54 has to ask whether this.socket is null precisely because a server that was never started is a legal state of this class. port() on line 63 completes the picture: the object knows its port and hands it over so that CStdLibTest.java:40, WinsockTest.java:44 and SyscallTest.java:23 can each build a sockaddr or a Data.ToPhi out of it. Nothing ever asks the server to do anything with the port it owns.

Thayorns (commented):
This moves the Winsock and CStdLib halves of SyscallTest into WinsockTest and CStdLibTest next to the classes they drive, and lifts RandomServer and the random port into shared fixtures. I reconstructed the two old inner classes from the base commit and diffed them against the extracted files: the test bodies are a verbatim move, so the behaviour of those cases is unchanged and SyscallTest is left holding only the socket object cases. The new surface is RandomServer and RandomPort, and that is where the remarks below sit.

eo-runtime/src/test/java/org/eolang/RandomServer.java
morphqdd: Bound the retry at 50 attempts, throwing instead of spinning past that. Considered asking the kernel for an ephemeral port outright, which is the cleaner fix in general, but socket.eo's htons only handles ports up to 32767 until #6133 lands, and the kernel's own ephemeral range starts above that. Switching now would make this PR's own tests flaky on that separate, already-tracked bug, so I kept the retry on the same fixed range RandomPort already draws from, just bounded.

eo-runtime/src/test/java/org/eolang/RandomPort.java
morphqdd: Both bindsSocketSuccessfullyViaSyscall methods now capture the port into a local and put it in the assertion message, so a red build says which port it tried instead of leaving strerror to speak alone.

eo-runtime/src/test/java/org/eolang/posix/CStdLibTest.java
morphqdd: Guarded both recvViaCStdLib and recvViaWinsock the same way the sibling accept-methods already do (`if (accepted > 0)` before closing).

eo-runtime/src/test/java/org/eolang/RandomServer.java
morphqdd: Bind eagerly in the constructor now, bounded at 50 attempts, throwing instead of retrying forever. The object is fully built once constructed, no half-started state, no separate started() call, socket field is final, port() reads straight off the bound socket via getLocalPort().

Thayorns (approved):

morphqdd:

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+        return (short) (((port & 0xFF) << 8) | ((port >> 8) & 0xFF));
+    }
+
+    /**
+     * Assert that a server thread received exactly the bytes a client sent.
+     * @param sent Bytes the client sent
+     * @param count Number of bytes the server reported as received
+     * @param received Bytes the server received
+     */
+    private static void assertReceived(
```
Favixx: `assertReceived` is byte-for-byte identical to `CStdLibTest#assertReceived` (line 191). It is not the only one: `htons`, `sockaddr`, `inetAddr`, `openSocket`, `closeSocket`, `bindSocket` and `ensure` all exist twice now, in both files, with the same shape. The PR already recognises the pattern and pulls `RandomServer`/`RandomPort` out into shared fixtures - `assertReceived` and `htons` carry no Winsock/CStdLib dependency at all, so they can move into the same package too. That would leave in each class only what genuinely differs: the `INSTANCE` it calls.

.github/workflows/simian.yml
```diff
@@ -21,4 +21,4 @@ jobs:
           distribution: 'temurin'
           java-version: 17
       - run: wget --quiet https://public.yegor256.com/simian.jar -O /tmp/simian.jar
-      - run: java -jar /tmp/simian.jar -threshold=15 "-excludes=**/SyscallTest.java" "-excludes=**/gen" "-excludes=**/it" "**/*.java"
+      - run: java -jar /tmp/simian.jar -threshold=15 "-excludes=**/WinsockTest.java" "-excludes=**/CStdLibTest.java" "-excludes=**/gen" "-excludes=**/it" "**/*.java"
```
Favixx: This is the load-bearing consequence of the duplication above: the exclude list did not get shorter, it doubled. Before, one file was hidden from simian; now two are, and they are hidden from each other, which is exactly where the duplication now lives. If the OS-independent helpers move to shared fixtures, both excludes can go away and simian starts guarding these files again instead of being told to look away.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+            } finally {
+                this.closeSocket(socket);
+            }
+        } finally {
+            this.cleanup();
+        }
+    }
+
+    @RepeatedIfExceptionsTest(repeats = 3)
+    void startsListenOnPosixSocket() throws UnknownHostException {
```
Favixx: `startsListenOnPosixSocket` is a Winsock test - it calls `Winsock.INSTANCE.listen`. The name came along with the copy that produced `PosixSocketTest`, and inside `SyscallTest` it was at least ambiguous; in a file called `WinsockTest` it reads as plainly wrong. `startsListenOnWindowsSocket` would match what it does. `CStdLibTest` keeps the same name correctly.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+        try {
+            this.ensure(this.startup() == 0);
+            final int socket = this.openSocket();
+            try {
+                this.ensure(socket > 0);
+                this.ensure(this.bindSocket(socket, new RandomPort().pick()) == 0);
+                MatcherAssert.assertThat(
+                    String.format(
+                        "Posix socket should have been bound to localhost via syscall, but it didn't, reason: %s",
```
Favixx: Two things off in one message: it says "Posix socket" inside `WinsockTest`, and it describes binding while the assertion below checks `listen`. Someone reading this failure in CI is told the bind failed when in fact listen did.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+        );
+    }
+
+    private void acceptViaWinsock(
+        final AtomicInteger port, final AtomicInteger accept, final AtomicInteger error
+    ) {
+        final int socket = this.openSocket();
+        try {
+            this.ensure(socket > 0);
+            while (this.bindSocket(socket, port.get()) != 0) {
```
Favixx: This is the same unbounded bind-retry that `RandomServer.bound()` deliberately replaces with a 50-attempt cap - and here it survives, in a background thread. If `bind` fails for a reason retrying cannot fix (permissions, an exhausted range), the loop spins hot forever, `server.join()` at line 159 never returns, and the test sits until the 10m default from `junit-platform.properties` kills it - times 3, because of `repeats = 3`. Worth reusing the bounded logic you just wrote rather than leaving two policies in one PR.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+                final SockaddrIn sockaddr = this.sockaddr(port.get());
+                MatcherAssert.assertThat(
+                    String.format(
+                        "Socket should have been connected to local server on sockets, but it didn't, reason: %s",
+                        this.getError()
+                    ),
+                    Winsock.INSTANCE.connect(client, sockaddr, sockaddr.size()),
+                    Matchers.equalTo(0)
+                );
+                server.join();
```
Favixx: `join()` with no timeout. Combined with the loop at 354 and the 10m method timeout, a server thread that never returns costs half an hour of CI per run. `join(timeout)` plus an assertion that the thread actually finished turns a hang into a fast, readable failure.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+                MatcherAssert.assertThat(
+                    String.format(
+                        "Accepted client socket must be positive, but it isn't, reason: %s",
+                        error.get()
+                    ),
+                    accept.get(),
+                    Matchers.greaterThan(0)
+                );
+            } finally {
```
Favixx, line 194: The `finally` closes the client socket but nothing ever cleans up the server thread. If the assertion at 151 fails, that thread stays blocked in `accept()` holding a bound port for the rest of the JVM; `repeats = 3` then leaves up to three of them behind. Closing the server socket, or interrupting the thread, in this `finally` would bound it.

Favixx (commented):
The move itself checks out. I diffed the two extracted classes against the corresponding nested classes in the old `SyscallTest` (base `377114c`) after normalising the mechanical renames, and all 6+6 test methods survive intact, along with the 5 socket-object tests left behind: no case lost, no assertion weakened, `@DisabledOnOs` guards preserved on both sides, and `server.stop()` still in the same `finally` blocks. `new RandomServer()` binds in its constructor, so dropping `.started()` is behaviour-preserving. Bounding the bind retry at 50 attempts instead of looping forever is a real improvement over what was there.

Two things I would push back on before this lands.

**The split did not remove the duplication, it moved it out of simian's sight.** Eight helpers - `assertReceived`, `htons`, `sockaddr`, `inetAddr`, `openSocket`, `closeSocket`, `bindSocket`, `ensure` - now exist in two files instead of one, and the simian exclude list grew from one entry to two to keep the build quiet about it. That is the opposite of what an extraction should do. The PR already demonstrates the right move by pulling `RandomServer` and `RandomPort` into shared fixtures; `assertReceived` and `htons` touch neither `Winsock` nor `CStdLib` and can follow them, after which the excludes are unnecessary.

**The pattern this PR fixes in one place is left unfixed in another.** `RandomServer.bound()` now caps the bind retry at 50 attempts, but `acceptViaWinsock`/`acceptViaCStdLib` still retry forever, in a background thread that the test joins without a timeout. With the 10m per-method default in `junit-platform.properties` and `repeats = 3`, one stuck thread is up to half an hour of a CI job. Since this PR is already the one that fixed that pattern once, fixing it in the second place seems in scope.

The rest are smaller: a Winsock test still called `startsListenOnPosixSocket`, two failure messages that name the wrong syscall family, `Thread.sleep(2000)` used four times as a synchronisation primitive, and a `@SuppressWarnings` on `RandomServer` that its two siblings prove is not needed.

On the `pr-size` failure: the argument in the description is reasonable - a relocation cannot be staged without temporary duplicates. But extracting the shared helpers first, in a small separate PR, would shrink this one and remove the simian excludes at the same time.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+        try {
+            this.ensure(this.startup() == 0);
+            final AtomicInteger accept = new AtomicInteger(0);
+            final AtomicInteger error = new AtomicInteger();
+            final AtomicInteger port = new AtomicInteger(new RandomPort().pick());
+            final Thread server = new Thread(
+                () -> this.acceptViaWinsock(port, accept, error)
+            );
+            server.start();
+            Thread.sleep(2000);
```
Favixx: `Thread.sleep(2000)` as a synchronisation primitive - the test hopes the server thread has reached `accept()` by then. On a loaded runner it may not have, which is what `@RepeatedIfExceptionsTest(repeats = 3)` is quietly paying for. A `CountDownLatch` counted down right after `listen()` succeeds makes it deterministic and removes 2 fixed seconds from each of the four tests that do this (here, 188, and `CStdLibTest` 118/154) - 8 seconds of wall clock on every build.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+        }
+    }
+
+    @RepeatedIfExceptionsTest(repeats = 3)
+    void startsListenOnPosixSocket() throws UnknownHostException {
+        try {
+            this.ensure(this.startup() == 0);
+            final int socket = this.openSocket();
+            try {
+                this.ensure(socket > 0);
```
Favixx, line 143: `ensure(socket > 0)` here, but `ensure(client >= 0)` at 149 and 191 - the same `openSocket()` return value, checked two different ways in one class. On Windows the failure value is `INVALID_SOCKET`, i.e. `~0` as unsigned, so neither check is really the right one; whichever you pick, it should at least be the same in all three places.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+    private void acceptViaWinsock(
+        final AtomicInteger port, final AtomicInteger accept, final AtomicInteger error
+    ) {
+        final int socket = this.openSocket();
+        try {
+            this.ensure(socket > 0);
+            while (this.bindSocket(socket, port.get()) != 0) {
+                port.set(new RandomPort().pick());
+            }
+            this.ensure(Winsock.INSTANCE.listen(socket, 5) == 0);
```
Favixx: Backlog `5` here, backlog `2` at line 124, both bare numbers in the same file. If the value does not matter, one shared constant says so; if it does, a name says why.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+     * Bind socket.
+     * @param socket Socket
+     * @param port Port
+     * @return Zero on success, -1 on error
+     */
+    private int bindSocket(final int socket, final int port) throws UnknownHostException {
+        return Winsock.INSTANCE.bind(
+            socket,
+            this.sockaddr(port),
+            16
```
Favixx: `16` is `sizeof(struct sockaddr_in)` spelled as a literal, while every other call site in this file uses `addr.size()` for exactly that. `CStdLibTest` line 256 has the same literal. `this.sockaddr(port)` is already being constructed on the line above - its `.size()` is both self-describing and immune to the struct changing.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+     */
+    private SockaddrIn sockaddr(final int port) throws UnknownHostException {
+        return new SockaddrIn(
+            (short) Winsock.AF_INET,
+            WinsockTest.htons(port),
+            this.inetAddr("127.0.0.1")
+        );
+    }
+
+    private void acceptViaWinsock(
```
Favixx, line 386: `acceptViaWinsock` is the only private method in this file without a javadoc block; its three parameters, two of which are out-parameters written by the thread, are exactly the ones a reader would want documented.

eo-runtime/src/test/java/org/eolang/RandomServer.java
```diff
@@ -0,0 +1,90 @@
...
+     * @return Bound server socket
+     */
+    private ServerSocket bound() throws IOException {
+        IOException last = null;
+        for (int attempt = 0; attempt < RandomServer.MAX_ATTEMPTS; attempt = attempt + 1) {
+            final int candidate = new RandomPort().pick();
+            try {
+                final ServerSocket opened = new ServerSocket();
+                opened.setReuseAddress(true);
+                opened.bind(new InetSocketAddress("127.0.0.1", candidate));
```
Favixx: When `bind` throws, `opened` goes out of scope without `close()`. The loop then does it again, up to 50 times, so a run of unavailable ports leaves a pile of unclosed `ServerSocket`s behind. `try (ServerSocket opened = new ServerSocket())` with a `return` inside on success, or an explicit `close()` in the catch, keeps the failure path clean. The old `started()` had the same hole, but this is new code and it is one line.

eo-runtime/src/test/java/org/eolang/RandomServer.java
```diff
@@ -0,0 +1,90 @@
...
+ * Server on a free port, shared by socket syscall tests.
+ *
+ * <p>Picks a port from a fixed, low range rather than asking the kernel
+ * for an ephemeral one, because {@code socket.eo}'s {@code htons} only
+ * supports ports up to 32767 until #6133 lands; the kernel's own
+ * ephemeral range starts above that limit.</p>
+ *
+ * @since 0.40.0
+ */
+@SuppressWarnings("PMD.AvoidUsingHardCodedIP")
```
Favixx: This suppression looks unnecessary: `WinsockTest` and `CStdLibTest` hardcode `127.0.0.1` and `192.0.2.1` with no suppression at all, and qulice is green on this PR. Worth deleting to confirm - a suppression that suppresses nothing is worse than none, because the next person believes the rule fires here.

eo-runtime/src/test/java/org/eolang/RandomPort.java
```diff
@@ -0,0 +1,28 @@
...
+     * Random number generator.
+     */
+    private final Random random = new Random();
+
+    /**
+     * Pick a random port.
+     * @return Random port
+     */
+    public int pick() {
+        final int min = 10_000;
```
Favixx: `10_000` and `20_000` sit inline while the sibling class names `MAX_ATTEMPTS` as a constant. The ceiling is also the more interesting number: `RandomServer`'s javadoc explains the real limit is 32767 (htons in `socket.eo` until #6133), so why the range stops at 20_000 is the one thing a reader cannot recover from the code.

eo-runtime/src/test/java/org/eolang/posix/CStdLibTest.java
```diff
@@ -0,0 +1,346 @@
...
+        Thread.sleep(2000);
+        final int client = this.openSocket();
+        try {
+            this.ensure(client >= 0);
+            final SockaddrIn sockaddr = this.sockaddr(port.get());
+            this.ensure(CStdLib.INSTANCE.connect(client, sockaddr, sockaddr.size()) == 0);
+            final byte[] buf = "Hello, Socket!".getBytes(StandardCharsets.UTF_8);
+            MatcherAssert.assertThat(
+                String.format(
+                    "Client had to sent message to the server, but it didn't, reason: %s",
```
Favixx: "Client had to sent message" - and unlike its twin in `WinsockTest` (line 198), this message drops the actual and expected byte counts even though the assertion below compares exactly those. Same assertion, two different diagnostics, is a good argument for the shared helper suggested above.

0crat:
@thayorns Thanks for the review! You've earned +24 points for this: +12 as a basis; +16 for the 1647 hits-of-code that you reviewed; +6 for the 24 comment(s) that you have made; -10 to not go over the cap.  Your running score is [+468](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
```diff
@@ -0,0 +1,407 @@
...
+ * @since 0.40
+ */
+@DisabledOnOs({OS.MAC, OS.LINUX})
+@Execution(ExecutionMode.SAME_THREAD)
+final class WinsockTest {
+
+    @RepeatedIfExceptionsTest(repeats = 3)
+    void connectsToLocalServerViaSyscall() throws IOException {
+        final RandomServer server = new RandomServer();
+        try {
```
yegor256: @morphqdd maybe it's better to make `RandomServer` auto-closeable and then do this here:

```
try (RandomServer server = new RandomServer()) {
...
}
```

eo-runtime/src/test/java/org/eolang/RandomServer.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Bind a server socket to a free port, retrying a bounded number of
+     * times so a persistent bind failure fails the test instead of
+     * hanging the build forever.
+     * @return Bound server socket
+     */
+    private ServerSocket bound() throws IOException {
+        IOException last = null;
+        for (int attempt = 0; attempt < RandomServer.MAX_ATTEMPTS; attempt = attempt + 1) {
+            final int candidate = new RandomPort().pick();
```
yegor256: @morphqdd I suggest you inline the `RandomPort` class into `RandomServer` and then use some Java lib, similar to this Gem of mine: https://github.com/yegor256/random-port

BTW, if a similar library doesn't exist for Java, maybe you can create one, release to Maven Central, and use it here, in our repo.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Done. `RandomServer` implements `AutoCloseable` now, call sites use try-with-resources. Two spots in `SyscallTest.java` close the server immediately on purpose (need a definitely-closed port for a refused-connection test), so those just call `.close()` directly.

eo-runtime/src/test/java/org/eolang/RandomServer.java
morphqdd: #6133 landed on master while this was open, so the port-range limit that justified `RandomPort` is gone. Both classes now just bind to port 0 and let the kernel pick a free port. Same idea as `random-port`, but the JDK already gives it to us in one line, so a separate library felt like overkill.

morphqdd:
@yegor256, please review

eo-runtime/src/test/java/org/eolang/RandomServer.java
morphqdd: No longer applicable. `RandomServer` binds directly to a kernel-assigned port now (port 0), no retry loop at all.

eo-runtime/src/test/java/org/eolang/RandomPort.java
morphqdd: Also no longer applicable. `RandomPort` asks the kernel for a free port now instead of an unseeded `Random`, and logs the port it picked.

eo-runtime/src/test/java/org/eolang/posix/CStdLibTest.java
morphqdd: This was already fixed in the previous commit (the `accept.get() > 0` guard before closing) before I picked this back up.

eo-runtime/src/test/java/org/eolang/RandomServer.java
morphqdd: `started()` is gone. `RandomServer` binds in its constructor now, so there's no half-built state.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Extracted into a shared `ReceivedBytes` class in `org.eolang`, used by both files. Instance-based, not static, per repo convention.

.github/workflows/simian.yml
morphqdd: Removed both excludes. Ran simian locally with threshold=15 and no excludes over src/, 0 duplicate lines now.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Renamed to `startsListenOnWindowsSocket`.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Fixed the message to describe listen (not bind) and say Windows (not Posix).

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Added a bounded retry (50 attempts, same cap `RandomServer` used to have) via a shared `bound()` helper, used by both `acceptViaWinsock` and `recvViaWinsock`. Same fix mirrored in `CStdLibTest`.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Added a `joined()` helper: `join(5s)` plus an assertion that the thread actually finished, instead of an unbounded `join()`.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Still open. Fixing it right means exposing the server's listening socket to the caller so it can be force-closed if the client-side assertion fails before `join()`. Didn't want to rush that without being able to run `WinsockTest` locally (Windows-only) to check it doesn't trade one race for another. Leaving as a follow-up.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Agreed a `CountDownLatch` would be better than the fixed sleep. That's a synchronization change across 4 call sites in two OS-specific files, and I can only run half of them (CStdLib) locally. Deferring rather than guessing on the Windows side.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Normalized to `> 0` everywhere in the file.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Named both: `LISTEN_BACKLOG` and `TEST_BACKLOG`. They're genuinely different things, not the same value copy-pasted.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: `bindSocket` now builds the sockaddr once and uses its own `.size()` instead of the literal.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Added javadoc to both `acceptViaWinsock` and `recvViaWinsock` (recvViaWinsock had the same gap, just wasn't flagged separately).

eo-runtime/src/test/java/org/eolang/RandomServer.java
morphqdd: Fixed, closes `opened` on a failed bind before rethrowing. The retry loop itself is gone too (kernel-assigned port now), so this was down to a single-attempt leak rather than up to 50.

eo-runtime/src/test/java/org/eolang/RandomServer.java
morphqdd: Tried removing it. PMD's `AvoidUsingHardCodedIP` does fire on `RandomServer`'s `127.0.0.1` when I checked locally, so it's not unnecessary here. Not sure why `WinsockTest`/`CStdLibTest`'s own literals read clean, but kept the suppression on `RandomServer` since qulice needs it.

eo-runtime/src/test/java/org/eolang/RandomPort.java
morphqdd: No longer applicable. `RandomPort` doesn't have a hand-picked range anymore, it asks the kernel for a free port.

eo-runtime/src/test/java/org/eolang/posix/CStdLibTest.java
morphqdd: Fixed the typo and added the actual/expected byte counts, matching `WinsockTest`'s message now.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Fixed. Extracted assertReceived into ReceivedBytes and htons into NetworkPort, both shared classes now used by WinsockTest and CStdLibTest. Only the INSTANCE call differs between the two files now.

.github/workflows/simian.yml
morphqdd: Fixed, both excludes dropped along with the duplication they were hiding.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Renamed to startsListenOnWindowsSocket.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Fixed. Message now says "started listening" and names the right stack (Windows/Win instead of Posix throughout).

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Fixed with a MAX_BIND_ATTEMPTS cap, same as RandomServer.bound().

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Fixed. join(JOIN_MILLIS) plus an isAlive() assertion, extracted into a shared joined() helper used by both tests.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Improved by reordering: joined(server) (with its own timeout+assertion) now runs before the final risky assertion in both tests, so the thread is already finished by the time that assertion can fail. A failure in the earlier connect() assertion (before the join) can still leave the server thread blocked in native accept() with no timeout on that specific path - didn't chase that further this round.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Fixed. Replaced with a CountDownLatch counted down right after listen() succeeds, in both acceptViaWinsock/recvViaWinsock and their CStdLib equivalents. Removes the flakiness and the 8s of fixed wall-clock cost.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Fixed, ensure(client > 0) consistently everywhere now.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Fixed. Named LISTEN_BACKLOG (5) and TEST_BACKLOG (2), both bare numbers replaced.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Fixed, uses addr.size() now instead of the 16 literal, same as everywhere else in the file.

eo-runtime/src/test/java/org/eolang/win32/WinsockTest.java
morphqdd: Added.

eo-runtime/src/test/java/org/eolang/RandomServer.java
morphqdd: Fixed with try-with-resources on the ServerSocket in bound(), same one-line hole you flagged.

eo-runtime/src/test/java/org/eolang/RandomServer.java
morphqdd: Tested this - removing it makes qulice fail with the same AvoidUsingHardCodedIP violation, so it is load-bearing here (this class is public, unlike the package-private WinsockTest/CStdLibTest, which seems to change which PMD path fires). Fixed the actual code instead of keeping the suppression: bind now goes through InetAddress.getLoopbackAddress() instead of a "127.0.0.1" string literal, so there's no hardcoded IP left to flag.

eo-runtime/src/test/java/org/eolang/RandomPort.java
morphqdd: Moot now, RandomPort no longer picks from a fixed range - it binds to port 0 and releases immediately, so there's no ceiling constant left to explain.

eo-runtime/src/test/java/org/eolang/posix/CStdLibTest.java
morphqdd: Fixed, ReceivedBytes.verify() now asserts the byte count and compares the actual received bytes, same as the connectViaWinsock side.

morphqdd:
Issue #6137 got closed already by work that landed straight on master, which gave `SyscallTest` a shared `Port` object and nested classes instead of splitting the file. That is a different shape from this branch, and it is where the conflict comes from. Rebasing here would mean undoing that, so I'd close this PR.

morphqdd:
Closing. Issue #6137 got closed already by different work that landed straight on master: `SyscallTest` got a shared `Port` object and nested classes instead of the Winsock/CStdLib file split this branch does. The actual split is now tracked separately in #6246. Rebasing this branch would mean undoing what's on master.

PR #6210, opened 2026-07-31
Name the legacy QQ root in the parse error

### LINKED ISSUE #4106: cryptic error message `"Only one object per source file is allowed and required"` in program with one object
Here is the EO program:

```eo
+package f

  # No comments.
  [] > main
    QQ.io.stdout
      "Hello world"
```

When we parse via `EoSyntax(ipt)`, we getting these error nodes in XMIR:

```xml
<errors>
  <error check="eo-parser" line="7" severity="critical">[7:-1] error: 'Invalid object declaration'</error>
  <error check="validate-objects-count" line="0" severity="critical">Only one object per source file is allowed and required</error>
</errors>
```

I have two questions:
1. What is the error here?
2. What is the meaning of "Only one object per source file is allowed and required" here (in the program we have exactly one top object `main`)

Thayorns:
The head `QQ.io.stdout` used to fail with `trailing garbage after expression`, with the caret sitting under the second `Q`. That says nothing about what is wrong: `Q` is the root token, the reader consumed exactly one character, and everything after it became garbage. Anyone porting a program written before `QQ` was dropped got a message pointing at the wrong thing.

The root reader now rejects a root glued to a letter or a digit and names the offending token: `QQ is not a valid object name, root Q must be followed by a dot`, with the caret on the `Q` that starts it. Nothing changes about what the parser accepts, only about what it says when it refuses. The canonical text is added to the table in §9.9 of `PARSER_SPEC.md`, as R-9.9.3 requires, and `eo-typos/qq.yaml` is updated to the new wording.

Two things the issue raises are already gone on master and stay untouched here. The confusing `Only one object per source file is allowed and required` no longer accompanies a parse failure, since `validate-objects-count` fires only when more than one object is present. And a plain child of a formation that carries no name already reports `object inside formation must have a name`, which is what the discussion asked for.

One rough edge remains: the following line still reports `indent increased by more than one level`, because the failed line never pushed a level and R-7.4 says the next line is processed as if it had not existed. That cascade is what the spec prescribes, so I left it alone rather than working around it.

Closes #4106

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -1369,6 +1369,7 @@ R-9.9.1. Every error condition in this spec has a single canonical text — **in
 | Malformed BYTES literal (R-3.13.1 — invalid byte form, e.g., `Z9-`, single trailing dash without prefix, odd hex run) | `invalid bytes literal` |
 | Meta after first non-meta object | `meta directive must precede all other objects` |
 | Plain child without name in formation | `object inside formation must have a name` |
+| Root identifier glued to a letter or a digit (e.g. the legacy `QQ.io.stdout`) | `<token> is not a valid object name, root <root> must be followed by a dot` (token and root substituted) |
 | Atom containing non-test child | `atom may contain only test attributes` |
 | `+>` outside indent 2 of top level | `test attribute legal only as direct child of top-level object` |
 | Bare reversed dispatch missing receiver | `reversed dispatch missing receiver` |
--- a/eo-parser/src/main/java/org/eolang/parser/Tokens.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Tokens.java
@@ -664,15 +664,17 @@ final class Tokens {
     }
 
     /**
-     * Whether the root identifier at the index is glued to a letter or
-     * a digit, as the legacy {@code QQ.io.stdout} head is.
+     * Whether the root identifier at the index is glued to a NAME
+     * character, as the legacy {@code QQ.io.stdout} head is. A root
+     * ends at a NAME terminator, so anything else standing next to it
+     * belongs to the same token.
      * @param body Body
      * @param idx Index of the root character
-     * @return True if a letter or a digit follows the root
+     * @return True if a NAME character follows the root
      */
     private static boolean glued(final String body, final int idx) {
         return idx + 1 < body.length()
-            && Character.isLetterOrDigit(body.charAt(idx + 1));
+            && !Tokens.terminates(body.charAt(idx + 1));
     }
 
     /**
--- a/eo-parser/src/main/java/org/eolang/parser/Tokens.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Tokens.java
@@ -323,7 +323,10 @@ final class Tokens {
 
     /**
      * Read a root identifier at the cursor — one of {@code Q}, {@code @},
-     * {@code ^}, {@code $}. Advances past it.
+     * {@code ^}, {@code $}. Advances past it. A root glued to a letter
+     * or a digit — the legacy {@code QQ} above all — is rejected here,
+     * so that the offending token names itself instead of surfacing as
+     * trailing garbage after a one-character expression.
      * @return Root value
      */
     Value readRoot() {
@@ -334,6 +337,15 @@ final class Tokens {
             );
         }
         final int start = this.cursor;
+        if (Tokens.glued(this.body, start)) {
+            throw new ParseError(
+                this.span.line(), this.span.indent() + start,
+                String.format(
+                    "%s is not a valid object name, root %s must be followed by a dot",
+                    Tokens.token(this.body, start), this.body.charAt(start)
+                )
+            );
+        }
         this.cursor = this.cursor + 1;
         return new Value(
             Value.Kind.ROOT, String.valueOf(this.body.charAt(start)),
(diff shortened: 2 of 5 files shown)


Thayorns:
@yegor256 the change is only in what the parser says when it refuses `QQ`, not in what it accepts — happy to clarify anything about the wording or the spec table row.

eo-parser/src/main/java/org/eolang/parser/Tokens.java
```diff
@@ -651,6 +663,33 @@ private static boolean rootStart(final char glyph) {
...
     }
 
+    /**
+     * Whether the root identifier at the index is glued to a letter or
+     * a digit, as the legacy {@code QQ.io.stdout} head is.
+     * @param body Body
+     * @param idx Index of the root character
+     * @return True if a letter or a digit follows the root
+     */
+    private static boolean glued(final String body, final int idx) {
```
morphqdd, line 675: glued() uses Character.isLetterOrDigit, broader than the grammar's own NAME token. Harmless here since anything it flags would error either way, but worth confirming it's deliberate rather than a copy-paste of a general-purpose predicate.

morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2777](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-parser/src/main/java/org/eolang/parser/Tokens.java
Thayorns: Good catch, and it turns out the mismatch runs the other way. `NAME` is defined by exclusion — `[a-z]` then anything outside `" \t,.|':;!?[]{}()"` — so `isLetterOrDigit` was *narrower* than `NAME`, not broader. It caught `QQ` and `Q1` but let punctuation through: I ran `readRoot` over a spread of inputs and `Q-foo`, `Q_foo`, `Q+x` and `Q*x` all returned `root=Q` and left the rest as exactly the trailing garbage this PR is meant to kill.

So it was a copy-paste of a general-purpose predicate rather than a deliberate choice. Switched `glued()` to `!terminates(...)`, which is the same terminator set `token()` already uses, and added `rejectsRootGluedToHyphen`. Full `eo-parser` suite is green.

eo-parser/PARSER_SPEC.md
```diff
@@ -1369,6 +1369,7 @@ R-9.9.1. Every error condition in this spec has a single canonical text — **in
 | Malformed BYTES literal (R-3.13.1 — invalid byte form, e.g., `Z9-`, single trailing dash without prefix, odd hex run) | `invalid bytes literal` |
 | Meta after first non-meta object | `meta directive must precede all other objects` |
 | Plain child without name in formation | `object inside formation must have a name` |
+| Root identifier glued to a letter or a digit (e.g. the legacy `QQ.io.stdout`) | `<token> is not a valid object name, root <root> must be followed by a dot` (token and root substituted) |
```
gemshrine, line 1372: This row — and the `glued` / `readRoot` doc comments — says "glued to a letter or a digit", but `glued` rejects a root fused to *any* non-terminator, which is broader. The PR's own `rejectsRootGluedToHyphen` proves `Q-` is rejected, and a hyphen is neither a letter nor a digit. Since R-9.9.1 makes this the canonical text for the condition, the description should match the check — e.g. "Root identifier glued to a non-terminator character" — so the spec doesn't under-state what actually errors.

eo-parser/src/main/java/org/eolang/parser/Tokens.java
```diff
@@ -334,6 +337,15 @@ Value readRoot() {
             );
         }
         final int start = this.cursor;
+        if (Tokens.glued(this.body, start)) {
+            throw new ParseError(
+                this.span.line(), this.span.indent() + start,
+                String.format(
+                    "%s is not a valid object name, root %s must be followed by a dot",
```
gemshrine, line 344: The message reads "root X must be followed by a dot", but the check reaching it only fires when the root is glued to a NAME character; a root followed by a non-dot terminator still takes the old path. I ran `Q, 1` through `EoSyntax` on this branch and it still reports `trailing garbage after expression` at the comma — the very message this PR replaces, just for the non-glued case. Not wrong (they are genuinely different situations), but the wording promises a universal "must be followed by a dot" rule the code only enforces against gluing; scoping it to what's checked would close the gap.

eo-parser/src/test/java/org/eolang/parser/TokensTest.java
```diff
@@ -296,6 +296,42 @@ void readsXiAsRoot() {
...
+    void rejectsRootGluedToDigit() {
+        Assertions.assertThrows(
+            ParseError.class,
+            () -> new Tokens("$1", new Span("$1", 1)).readRoot(),
+            "readRoot must reject a root glued to a digit"
+        );
+    }
+
+    @Test
+    void rejectsRootGluedToHyphen() {
```
gemshrine, line 318: The reject side is nicely covered — letter, digit, hyphen. The case that isn't unit-tested is the valid one the fix must not break: a root immediately followed by a dot. Everything hinges on `.` being in `TERMINATORS` so `glued` returns false for `Q.io`; I confirmed it works (the integration `Q.io.stdout` reaches the unrelated formation-name error rather than a root error), but a `new Tokens("Q.io", ...).readRoot()` asserting a `ROOT` value without throwing would pin that assumption directly next to these reject tests.

gemshrine (commented):
Good change — the self-naming error is a real improvement over the caret-under-the-second-Q garbage message. I ran it on this branch: `QQ.io.stdout`, `$1`, `@foo` and `^^` now each name themselves (`QQ is not a valid object name, root Q must be followed by a dot`), and valid roots (`Q.io.stdout`, `@.x`) sail past `readRoot` untouched. Three notes, all minor.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+546](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6220, opened 2026-07-31
Skip the block nested under a failed line during recovery

### LINKED ISSUE #4140: Our parser doesn't recover after an error, but it can
This is my code:

```
# Example.
[] > example
  [x] +++ bad
    one
      two
  [] > good
    one
      two
```

I'm getting this:

```
[ERROR] Failed to parse 'app.eo:3': [3:6] error: 'Invalid object declaration'
  [x] +++ bad
      ^
```

In the XMIR I see an empty `<objects/>` element.

I believe, we can easily recover the error in the first attribute and skip everything that goes after the line no.3, until the indentation gets back to the level where the error started.

Thayorns:
When a line fails to parse, the parser reported the error and then went
on to parse the lines nested under it. Those lines have lost their
parent, so they collided with the indent stack and produced errors of
their own — the snippet in #4140 raised three errors where the source
had made one, and the object that followed never made it into the XMIR.

The walk now resumes at the first line standing back at the failed
line's indent. Everything in between belonged to the failed line and is
skipped. Blank lines carry no indent of their own, so they do not resume
parsing either. The sibling that follows the skipped block is parsed as
usual, which is what the issue asks for.

Closes #4140

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -16,7 +16,9 @@ import org.xembly.Directive;
  * of the spec, then dispatched into the shared {@link Stack},
  * {@link Globals}, and {@link Emit} state. Recovery (§7) is line-local
  * — a {@link ParseError} is caught, the line's directives are rolled
- * back to a {@link Emit#savepoint()}, and the error is emitted.</p>
+ * back to a {@link Emit#savepoint()}, the error is emitted, and the
+ * block nested under the failed line is skipped up to the first line
+ * back at its indent.</p>
  *
  * <p>This is the analogue of {@code EoSyntax} on the ANTLR side, but
  * with no parser-grammar dependency: classification, validation, and
@@ -79,8 +81,9 @@ final class Eo implements Iterable<Directive> {
             if (!globals.inTextBlock() && Eo.isBytesContinuation(span.body())) {
                 final int next = Eo.mergeBytesContinuation(spans, idx, stack, globals, emit);
                 idx = next;
+            } else if (Eo.process(span, stack, globals, emit)) {
+                idx = Eo.resume(spans, idx + 1, span.indent());
             } else {
-                Eo.process(span, stack, globals, emit);
                 idx = idx + 1;
             }
         }
@@ -297,11 +300,13 @@ final class Eo implements Iterable<Directive> {
      * @param stack The indent stack
      * @param globals The global parser state
      * @param emit The directives sink
+     * @return True when the line failed to parse
      * @checkstyle ParameterNumberCheck (3 lines)
      */
-    private static void process(
+    private static boolean process(
         final Span span, final Stack stack, final Globals globals, final Emit emit
     ) {
+        boolean failed = false;
         if (globals.inTextBlock()) {
             Eo.continueTextBlock(span, stack, globals, emit);
         } else if (span.tab()) {
@@ -313,8 +318,31 @@ final class Eo implements Iterable<Directive> {
             globals.markEmitted();
             globals.clearBlanks();
         } else {
-            Eo.dispatch(span, stack, globals, emit);
+            failed = Eo.dispatch(span, stack, globals, emit);
         }
+        return failed;
+    }
+
+    /**
+     * The index the walk resumes at after a line failed — the first
+     * line standing at or above the failed line's indent (§7). The
+     * lines in between belong to the failed line, so parsing them on
+     * their own would only pile up errors the source never made. A
+     * blank line carries no indent of its own and never resumes.
+     * @param spans Materialised list of source spans
+     * @param from Index right after the failed line
+     * @param indent Indent of the failed line
+     * @return Index of the resumption point
+     */
+    private static int resume(
+        final java.util.List<Span> spans, final int from, final int indent
+    ) {
+        int idx = from;
+        while (idx < spans.size()
+            && (spans.get(idx).blank() || spans.get(idx).indent() > indent)) {
+            idx = idx + 1;
+        }
+        return idx;
     }
 
     /**
@@ -387,10 +415,11 @@ final class Eo implements Iterable<Directive> {
      * @param stack The indent stack
      * @param globals The global parser state
      * @param emit The directives sink
+     * @return True when the line failed to parse
      * @checkstyle ParameterNumberCheck (3 lines)
      */
     @SuppressWarnings("PMD.UnnecessaryLocalRule")
-    private static void dispatch(
+    private static boolean dispatch(
         final Span span, final Stack stack, final Globals globals, final Emit emit
     ) {
         if (!span.blank() && span.head() != '#') {
@@ -398,13 +427,16 @@ final class Eo implements Iterable<Directive> {
         }
         final int tstartsen = emit.savepoint();
         final int frame = stack.size();
+        boolean failed = false;
         try {
             Eo.classify(span).into(stack, globals, emit);
         } catch (final ParseError err) {
             emit.rollback(tstartsen);
             stack.silentTruncate(frame);
             emit.error(err.line(), err.pos(), err.getMessage());
+            failed = true;
         }
+        return failed;
     }
 
     /**
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -17,8 +17,8 @@ import org.xembly.Directive;
  * {@link Globals}, and {@link Emit} state. Recovery (§7) is line-local
  * — a {@link ParseError} is caught, the line's directives are rolled
  * back to a {@link Emit#savepoint()}, the error is emitted, and the
- * block nested under the failed line is skipped up to the first line
- * back at its indent.</p>
+ * block nested under the failed line is skipped up to the point
+ * {@link Recovery} names.</p>
  *
  * <p>This is the analogue of {@code EoSyntax} on the ANTLR side, but
  * with no parser-grammar dependency: classification, validation, and
@@ -75,6 +75,7 @@ final class Eo implements Iterable<Directive> {
         );
         final java.util.List<Span> spans = new java.util.ArrayList<>(16);
         new Source(this.source).forEach(spans::add);
+        final Recovery recovery = new Recovery(spans);
         int idx = 0;
         while (idx < spans.size()) {
             final Span span = spans.get(idx);
@@ -82,7 +83,7 @@ final class Eo implements Iterable<Directive> {
                 final int next = Eo.mergeBytesContinuation(spans, idx, stack, globals, emit);
                 idx = next;
             } else if (Eo.process(span, stack, globals, emit)) {
-                idx = Eo.resume(spans, idx + 1, span.indent());
+                idx = recovery.after(idx);
             } else {
                 idx = idx + 1;
             }
@@ -321,28 +322,6 @@ final class Eo implements Iterable<Directive> {
             failed = Eo.dispatch(span, stack, globals, emit);
         }
         return failed;
-    }
-
-    /**
-     * The index the walk resumes at after a line failed — the first
-     * line standing at or above the failed line's indent (§7). The
-     * lines in between belong to the failed line, so parsing them on
-     * their own would only pile up errors the source never made. A
-     * blank line carries no indent of its own and never resumes.
-     * @param spans Materialised list of source spans
-     * @param from Index right after the failed line
-     * @param indent Indent of the failed line
-     * @return Index of the resumption point
-     */
-    private static int resume(
-        final java.util.List<Span> spans, final int from, final int indent
-    ) {
-        int idx = from;
-        while (idx < spans.size()
-            && (spans.get(idx).blank() || spans.get(idx).indent() > indent)) {
-            idx = idx + 1;
-        }
-        return idx;
     }
 
     /**
--- a/eo-parser/src/main/java/org/eolang/parser/Recovery.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Recovery.java
@@ -1,0 +1,63 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.parser;
+
+import java.util.List;
+
+/**
+ * The place where the walk carries on after a line failed to parse — §7.
+ *
+ * <p>Every line indented deeper than a failed one belongs to it. Parsing
+ * such a line on its own only piles up errors the source never made,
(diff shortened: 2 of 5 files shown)


Thayorns:
@yegor256 happy to clarify anything here.

Favixx (commented):
The core of this is right, and the shape is right too: `Recovery` is a small object that answers one question, `Eo` keeps the walk, and the whole thing is 63 lines. The resumption rule - first line back at or above the failed indent, blanks not counting - matches what R-3.4.7-style block structure implies, and the `EoTest` packs show the #4140 snippet collapsing from three errors to one with the following object surviving. `mergeBytesContinuation` does not mutate the span list, so sharing it with `Recovery` is safe.

Two things I would want settled before it merges.

**The skip is keyed on the exception, not on the condition.** `process` returns true only when `dispatch` caught a `ParseError`. But two branches above it, `span.tab()` and the odd-indent check, emit an error and return false - so a line with a tab in its leading whitespace still gets its children parsed as orphans, which is the exact failure mode #4140 is about. Either those branches set the flag too, or the `@return` javadoc should stop claiming "the line failed to parse", because right now it promises more than the code does.

**Skipping blank lines turns off three rules, quietly.** Blanks inside the block never reach `LnBlank.into`, which is where R-6.5.3 (consecutive blanks), R-6.5.6 and R-8.4 (trailing blanks at EOF) are enforced, and where `globals.blank()` feeds the `pendingBlanks` counter the resumption line is classified against. Two consecutive blanks under a failed line used to be an error and are not any more; if the block runs to EOF, the trailing-blank check cannot fire at all. That is defensible - the file is already broken - but it is a second behaviour change with nothing in the description or the tests about it.

On the tests: `RecoveryTest` covers the two cases that work and neither of the two that would catch an off-by-one. There is no case where the next line is shallower than the failed one (the `while` body never runs, `after` returns `failed + 1`), and none where the block reaches the end of the file (`after` returns `spans.size()`, which is what terminates the walk in `Eo`). A comparison changed from `>` to `>=`, or a return of `size() - 1`, would pass everything in this PR.

eo-parser/src/main/java/org/eolang/parser/Eo.java
```diff
@@ -297,11 +301,13 @@ private static boolean hex(final char glyph) {
...
      * @checkstyle ParameterNumberCheck (3 lines)
      */
-    private static void process(
+    private static boolean process(
         final Span span, final Stack stack, final Globals globals, final Emit emit
     ) {
+        boolean failed = false;
         if (globals.inTextBlock()) {
             Eo.continueTextBlock(span, stack, globals, emit);
         } else if (span.tab()) {
```
Favixx, line 313: These two branches - `span.tab()` here and the odd-indent one right below - call `emit.error` and then fall through to `return false`, so the block nested under them is still walked. That is the same shape as #4140: a line that did not parse, followed by children that have lost their parent. Write `[] > a` with a tab in front of it and give it a body, and the extra errors come back. If `failed` means "this line produced an error and its children are orphans", these two qualify; if it means "a ParseError was thrown", then the skip is keyed on the exception rather than on the condition it is supposed to model.

eo-parser/src/main/java/org/eolang/parser/Eo.java
```diff
@@ -297,11 +301,13 @@ private static boolean hex(final char glyph) {
      * @param stack The indent stack
      * @param globals The global parser state
      * @param emit The directives sink
+     * @return True when the line failed to parse
```
Favixx, line 304: `@return True when the line failed to parse` is wider than what the method does - it returns true only when `dispatch` caught a `ParseError`, not when this method itself emitted an error two branches down. Either the branches change or this line should say "when the line threw a ParseError", so the next reader does not assume the tab case is covered.

eo-parser/src/main/java/org/eolang/parser/Eo.java
```diff
@@ -73,14 +75,16 @@ Iterable<Directive> directives() {
...
         final java.util.List<Span> spans = new java.util.ArrayList<>(16);
         new Source(this.source).forEach(spans::add);
+        final Recovery recovery = new Recovery(spans);
         int idx = 0;
         while (idx < spans.size()) {
             final Span span = spans.get(idx);
             if (!globals.inTextBlock() && Eo.isBytesContinuation(span.body())) {
                 final int next = Eo.mergeBytesContinuation(spans, idx, stack, globals, emit);
                 idx = next;
+            } else if (Eo.process(span, stack, globals, emit)) {
```
Favixx, line 85: `Eo.process` mutates the stack, the globals and the emit sink, and now it is called from inside an `else if` condition, so the reader has to know it has side effects to follow the loop at all. Keeping the `else` and branching inside it on the returned flag reads the same and costs two lines.

eo-parser/src/main/java/org/eolang/parser/Recovery.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Whether the line at {@code idx} still belongs to the block of a
+     * line that failed at {@code indent}.
+     * @param idx Index of the line
+     * @param indent Indent of the failed line
+     * @return True when the line must be skipped
+     */
+    private boolean skipped(final int idx, final int indent) {
+        final Span span = this.spans.get(idx);
+        return span.blank() || span.indent() > indent;
```
Favixx, line 61: `span.blank()` here means the skipped blanks never reach `LnBlank.into`, and that method is where three rules live: R-6.5.3 (consecutive blanks), R-6.5.6 and R-8.4 (more than one trailing blank at EOF, via the `globals.trailingBlanks() > 1` check in `Eo`). So a failed line silently switches all three off for everything nested under it - and, when the block runs to EOF, for the rest of the file. Two blank lines inside a failed block used to be an error and now are not. That may well be the right trade, but it is a second behaviour change riding along with the first, and nothing in the description or the tests mentions it.

eo-parser/src/main/java/org/eolang/parser/Recovery.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * The index the walk resumes at, having failed on the line at
+     * {@code failed}.
+     * @param failed Index of the line that failed to parse
+     * @return Index of the resumption point
+     */
+    int after(final int failed) {
+        final int indent = this.spans.get(failed).indent();
+        int idx = failed + 1;
+        while (idx < this.spans.size() && this.skipped(idx, indent)) {
```
Favixx, line 46: Beyond the error rules, this loop also skips the counter updates the *next* line depends on: `globals.blank()` feeds `pendingBlanks()`, which R-6.5 timing reads to decide the legality of the line that resumes the walk. The resumption line is therefore classified as if the blanks before it did not exist. Worth a test with a blank immediately before the sibling that resumes, to pin down which answer is intended.

eo-parser/src/main/java/org/eolang/parser/Recovery.java
```diff
@@ -0,0 +1,63 @@
...
+        this.spans = lines;
+    }
+
+    /**
+     * The index the walk resumes at, having failed on the line at
+     * {@code failed}.
+     * @param failed Index of the line that failed to parse
+     * @return Index of the resumption point
+     */
+    int after(final int failed) {
```
Favixx, line 43: The javadoc does not say what the caller gets when the block runs to the end of the file. `after` can return `spans.size()`, which is exactly what the loop in `Eo` relies on to terminate, and that contract is worth one sentence here since it is the only reason the caller needs no bounds check of its own.

eo-parser/src/main/java/org/eolang/parser/Recovery.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Whether the line at {@code idx} still belongs to the block of a
+     * line that failed at {@code indent}.
+     * @param idx Index of the line
+     * @param indent Indent of the failed line
+     * @return True when the line must be skipped
+     */
+    private boolean skipped(final int idx, final int indent) {
```
Favixx, line 59: `skipped(final int idx, final int indent)` takes an index into the list and an indent value, two ints that mean entirely different things and cannot be told apart at the call site. Passing the `Span` - the method resolves it first thing anyway - removes the ambiguity and lets the method be static.

eo-parser/src/main/java/org/eolang/parser/Recovery.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * All lines of the program, in source order.
+     */
+    private final List<Span> spans;
+
+    /**
+     * Ctor.
+     * @param lines All lines of the program, in source order
+     */
+    Recovery(final List<Span> lines) {
```
Favixx, line 33: The parameter is `lines` and the field it initialises is `spans`. Both names are used elsewhere in this class for the same thing (the javadoc above says "All lines", the field comment says "All lines of the program"), so one of them should win.

eo-parser/src/test/java/org/eolang/parser/RecoveryTest.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+final class RecoveryTest {
+
+    @Test
+    void resumesAtSiblingOfFailedLine() {
+        MatcherAssert.assertThat(
+            "the walk must resume at the first line back at the indent of the failed one",
+            new Recovery(
+                Arrays.asList(
+                    new Span("  über! ☃", 1),
```
Favixx, line 24: `"  über! ☃"` and `"    क्ष"` suggest the content of the line matters to `Recovery`, and it does not - the class reads `indent()` and `blank()` and nothing else. Plain markers would make it obvious at a glance that this test is about indentation, and would leave the exotic input to whichever test actually exercises the lexer.

eo-parser/src/test/java/org/eolang/parser/RecoveryTest.java
```diff
@@ -0,0 +1,50 @@
...
+                    new Span("    क्ष", 2),
+                    new Span("      deeper", 3),
+                    new Span("  sibling", 4)
+                )
+            ).after(0),
+            Matchers.equalTo(3)
+        );
+    }
```
Favixx, line 32: Missing case: the line right after the failed one stands at a *shallower* indent, so `after` must return `failed + 1` without skipping anything. That is the branch where the `while` body never runs, and it is the one an off-by-one in the comparison at line 61 would break - `>=` instead of `>` would swallow the sibling and still pass both tests here.

eo-parser/src/test/java/org/eolang/parser/RecoveryTest.java
```diff
@@ -0,0 +1,50 @@
...
+                    new Span("bad!!!", 1),
+                    new Span("  child", 2),
+                    new Span("", 3),
+                    new Span("  child", 4),
+                    new Span("next", 5)
+                )
+            ).after(0),
+            Matchers.equalTo(4)
+        );
+    }
```
Favixx, line 49: Missing case: the block runs to the end of the file, where `after` must return `spans.size()`. That return value is what stops the walk in `Eo`, and right now nothing checks it - a version of this loop that returned `size() - 1` would pass both tests in this file and then re-parse the last line forever.

eo-parser/src/test/java/org/eolang/parser/EoTest.java
```diff
@@ -125,6 +125,57 @@ void recoversFromBadLineAndContinues() {
...
+        );
+    }
+
+    @Test
+    void parsesSiblingAfterFailedLine() {
+        MatcherAssert.assertThat(
+            "the sibling standing at the indent of a failed line must still be parsed",
+            EoTest.render(
+                "[] > example",
+                "  [x] +++ bad",
```
Favixx, line 150: These six lines are byte-identical to the ones in `skipsBlockOfFailedLine` above. Two tests asserting different things about the same source is fine, but the source itself should be written once - a private constant, or one test with both assertions. As it stands, a change to the fixture has to be made twice and there is nothing to say so.

eo-parser/src/test/java/org/eolang/parser/EoTest.java
```diff
@@ -125,6 +125,57 @@ void recoversFromBadLineAndContinues() {
...
+        );
+    }
+
+    @Test
+    void skipsBlockOfFailedLineAcrossBlanks() {
+        MatcherAssert.assertThat(
+            "a blank line inside the block of a failed line must not resume parsing",
+            EoTest.render(
+                "[] > example",
+                "  [x] +++ bad",
```
Favixx, line 168: This pack has one blank inside the block. The interesting one has two, or has the block run to EOF with blanks after it, because that is where the skip now suppresses R-6.5.3 and R-8.4. Whichever behaviour is intended, a test asserting it here is what stops the next person restoring the old walk by accident.

eo-parser/src/test/java/org/eolang/parser/EoTest.java
```diff
@@ -125,6 +125,57 @@ void recoversFromBadLineAndContinues() {
         );
     }
 
+    @Test
+    void skipsBlockOfFailedLine() {
+        MatcherAssert.assertThat(
```
morphqdd, line 130: This is a parser recovery fix but coverage is JUnit-only. Parser fixes in this repo usually also get a declarative YAML pack test under eo-parser/src/test/resources/org/eolang/parser/eo-syntax/ (EoSyntaxTest.validatesEoSyntax auto-discovers anything dropped there), the portable contract test on top of the Java one.

Thayorns: Agreed, added `eo-parser/src/test/resources/org/eolang/parser/eo-syntax/recovery-skips-nested-block.yaml`. It carries the same input as the JUnit pair and asserts `/object/errors[count(error)=1]` plus `/object/o[@name='example']/o[@name='good']/o[@name='first']`, so the one-error contract and the surviving sibling are both locked down declaratively. `validatesEoSyntax` picked it up as case [62] and it is green.

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+124](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2857](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (200 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+558](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6240, opened 2026-08-01
Fix regex next after zero-width matches

### LINKED ISSUE #6143: `string.regex.match.next` stuck in an infinite loop for terminal zero-width matches

Titanism0v0:
## Summary

- stop `string.regex.match.next` from repeating terminal zero-width matches
- advance the matcher by one character after non-terminal zero-width matches
- add regression tests for both behaviors

The previous implementation reused `to` as the next matcher start, so Java's matcher could return the same zero-width match indefinitely. Terminal zero-width matches now produce a non-existent result, while non-terminal zero-width matches advance by one character.

Fixes #6143

## Test plan

- `mvn -pl :eo-runtime "-Dtest=EOregexTest,EOregexEOpatternEOmatchEOmatchedfromindexTest" test -ntp`

### DIFF
--- a/eo-runtime/src/main/eo/string/regex.eo
+++ b/eo-runtime/src/main/eo/string/regex.eo
@@ -59,13 +59,21 @@
         ? > position /Q.number
         ? > start /Q.number
       [] > missing
+        T > count
+          "Matched block does not exist, can't get count"
         start.gte 0 > exists
         T > from
           "Matched block does not exist, can't get 'from' position"
+        T > [index] > group
+          "Matched block does not exist, can't get group"
         T > groups
           "Matched block does not exist, can't get groups"
+        T > next
+          "Matched block does not exist, can't get next"
         0 > position
         -1 > start
+        T > text
+          "Matched block does not exist, can't get text"
         T > to
           "Matched block does not exist, can't get 'to' position"
       matched. > next
@@ -242,6 +250,14 @@
   compile. --> stops-on-compiling-an-invalid-pattern
     regex "/[a-z/"
 
+  group. --> stops-on-taking-a-group-of-a-terminal-zero-width-match
+    next.
+      next.
+        match.
+          regex "/$/"
+          "x"
+    1
+
   group. --> stops-on-taking-a-group-of-an-unmatched-block
     next.
       match.
@@ -255,6 +271,13 @@
         regex "/[a-z]+/"
         "123"
 
+  count. --> stops-on-taking-the-group-count-of-a-terminal-zero-width-match
+    next.
+      next.
+        match.
+          regex "/$/"
+          "x"
+
   count. --> stops-on-taking-the-group-count-of-an-unmatched-block
     next.
       match.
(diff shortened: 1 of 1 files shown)


eo-runtime/src/main/eo/string/regex.eo
```diff
@@ -50,6 +58,16 @@
       [] > matched-from-index /Q.string.regex.pattern.match.matched
         ? > position /Q.number
         ? > start /Q.number
+      [] > missing
```
morphqdd, line 61: `missing` only defines exists/from/groups/position/start/to. Every other "not found" result in this file also handles next/text/count/group with the friendly `T` message. Chaining `.next` or `.text` on a terminal zero-width match will fail with a raw attribute-not-found error instead. Not covered by a test either.

morphqdd (changes requested):

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2850](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@titanism0v0 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+52](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6249, opened 2026-08-01
Add cant-get callback to getenv

### LINKED ISSUE #6142: `getenv` function fails to differentiate between an unset variable and a variable with an empty value
## What happens

`getenv` cannot distinguish an absent environment variable from a variable that exists with an empty value.

The runtime was started with `EO_EMPTY_TEST` explicitly set to an empty string. Both calls produce the same EO result:

| expression | process environment | actual result |
| --- | --- | --- |
| `getenv "EO_EMPTY_TEST"` | `EO_EMPTY_TEST=""` | `""` |
| `getenv "EO_MISSING_TEST"` | variable absent | `""` |

An empty value is a valid environment value, so these inputs carry different information that the public EO API currently loses.

## Why this is a bug

The native POSIX wrapper already distinguishes the two cases. [`GetenvSyscall`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/Posix/GetenvSyscall.java#L35-L45) records `present = env != null` and returns it as code at line 39; it returns `""` as output for both an existing empty variable and an absent one.

The public [`getenv`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/getenv.eo#L1-L21) then exposes only `output.` at line 14. Its current documentation consequently makes the false claim that an empty returned string means that the variable does not exist.

## Steps to reproduce

1. Start an EO program with `EO_EMPTY_TEST` present in its environment and assigned an empty value.
2. Evaluate `getenv "EO_EMPTY_TEST"`.
3. Evaluate `getenv "EO_MISSING_TEST"` for a name absent from the process environment.
4. Observe that both evaluations return `""`.

## Expected behavior

EO should preserve the existence bit that the platform wrapper already supplies. For example, `getenv` could return an object or tuple containing `exists` and `value`, or expose an explicit fallback for the absent case. The existing string-only API must not document an empty value as absence, because it destroys observable process state.

Guflly:
getenv now uses the syscall result code so an unset variable invokes cant-get while an existing empty variable remains an empty string. Embedded tests cover existing and missing environment variables. Tests: EOgetenvTest (2 passed). Closes #6142. @yegor256

### DIFF
--- a/eo-runtime/src/main/eo/getenv.eo
+++ b/eo-runtime/src/main/eo/getenv.eo
@@ -1,5 +1,4 @@
-# Get environment variable as `string`.
-# If the return `string` is empty - the variable does not exist.
+# Get environment variable as `string`, or `cant-get` if it does not exist.
 #
 # See https://man7.org/linux/man-pages/man3/getenv.3.html.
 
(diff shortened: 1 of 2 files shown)


0crat:
@guflly Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6251, opened 2026-08-02
fix(#6217): reject Zip Slip paths when unpacking dependency jars

### LINKED ISSUE #6217: Zip Slip path traversal in `CentralMaven.unpack` when resolving a dependency jar
## What Happens

`CentralMaven.unpack` writes each zip entry of a resolved dependency jar to a path built directly from the entry's own name, with no check that the resulting path stays inside the intended destination directory.

```java
final Path target = dest.resolve(entry.getName());
...
Files.copy(zis, target, StandardCopyOption.REPLACE_EXISTING);
```

This method is reached from `MjResolve.exec`, through `Resolving.unpacked`, for every dependency jar named by a `+rt jvm group:artifact:version` meta in a source file.

Confirmed by building a jar with two entries, `org/eolang/ok.eo` and `../../../../evil-escaped.txt`, and unpacking it through this code path: the second entry was written outside the intended destination directory, several levels above it. More `../` segments in an entry name reach further outside the destination, and `REPLACE_EXISTING` means an existing file at the resolved path gets silently overwritten.

Since the jar being unpacked comes from a dependency coordinate a source file's `+rt jvm` meta names, a project that pulls in a malicious or compromised dependency jar can have arbitrary files written or overwritten on the machine running the build.

## What Should Happen

Before writing each entry, `unpack` should resolve the target path and verify it still starts with the normalized destination directory, rejecting or skipping any entry whose target would land outside it.

androvonx95:
Closes #6217.

@yegor256

`CentralMaven.unpack` resolved each zip entry name straight into the destination with no boundary check, so a jar whose entries contain `..` could write or overwrite files outside that directory. A project that pulls a compromised `+rt jvm` dependency could therefore escape the unpack root.

Each target is now normalized and rejected unless it stays under the destination root. Tests cover a safe jar, the rejection, and that the escaping entry is not materialized.


eo-maven-plugin/src/main/java/org/eolang/maven/CentralMaven.java
```diff
@@ -212,6 +212,40 @@ public BiConsumer<Dependency, Path> andThen(
...
     }
 
+    /**
+     * Unpacks a JAR (ZIP) file into the given directory.
+     * Rejects entries whose resolved path would leave {@code dest} (Zip Slip).
+     * @param jar Path to the JAR file
+     * @param dest Destination directory
+     * @throws IOException If unpacking fails or an entry escapes {@code dest}
+     */
+    static void unpack(final Path jar, final Path dest) throws IOException {
```
yegor256: @androvonx95 static methods are evil in general: https://www.yegor256.com/2014/05/05/oop-alternative-to-utility-classes.html think about a better design

androvonx95:
@yegor256 I moved the extraction into Unpacked which is a small Scalar that owns the jar and destination so now CentralMaven just composes the object instead of calling a static helper. The unpacking responsibility now stays where it belongs.

eo-maven-plugin/src/test/java/org/eolang/maven/UnpackedTest.java
```diff
@@ -0,0 +1,101 @@
...
+        UnpackedTest.jar(jar, "../evil-escaped.txt", "pwned");
+        Assertions.assertThrows(
+            IOException.class,
+            () -> new Unpacked(jar, temp.resolve("unpacked")).value(),
+            "Zip Slip entry must be rejected instead of writing outside the destination"
+        );
+    }
+
+    @Test
+    void namesEscapingZipEntryInRejection(@Mktmp final Path temp) throws IOException {
```
Thayorns: This one asserts on the wording of the exception message, matching `containsString("evil-escaped.txt")` against `getMessage()`. That pins the message text of Unpacked.value() into a test, so rephrasing the message breaks a test that is not about phrasing. The behavior worth holding onto is the rejection itself, and the test above already holds it.

eo-maven-plugin/src/test/java/org/eolang/maven/UnpackedTest.java
```diff
@@ -0,0 +1,101 @@
...
+        );
+    }
+
+    @Test
+    void doesNotWriteEscapingZipEntryOutsideDestination(@Mktmp final Path temp)
+        throws IOException {
+        final Path jar = temp.resolve("evil.jar");
+        UnpackedTest.jar(jar, "../evil-escaped.txt", "pwned");
+        final Path dest = temp.resolve("unpacked");
+        Assertions.assertThrows(
```
Thayorns: Two assertions here: assertThrows on line 74 and assertThat on line 79. If the throw stops happening the test fails on the first one and you never learn whether the file was written, which is the thing this test is named after. The throw is already covered by rejectsZipEntryThatEscapesDestination.

eo-maven-plugin/src/main/java/org/eolang/maven/Unpacked.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param archive Path to the JAR file
+     * @param target Destination directory
+     */
+    Unpacked(final Path archive, final Path target) {
+        this.jar = archive;
+        this.dest = target;
+    }
+
+    @Override
+    public Path value() throws IOException {
```
Thayorns: `value()` writes the whole archive to disk and also returns `root`. The one caller, CentralMaven.java:181, drops the returned path on the floor, so the return value exists only to satisfy Scalar. Naming the class after a result while its real job is the side effect makes it hard to tell from the call site that anything happened.

Thayorns (commented):
Moves the jar unpacking out of a private static helper in CentralMaven into an Unpacked object and rejects entries that resolve outside the destination root. The boundary check itself is right: normalizing after resolve and comparing with startsWith catches both `../` and absolute entry names. My notes are all about the tests and the shape of the new object.

eo-maven-plugin/src/test/java/org/eolang/maven/UnpackedTest.java
androvonx95: Removed this test. The rejection behavior is already covered by `rejectsZipEntryThatEscapesDestination`, and pinning exception message text was brittle.

eo-maven-plugin/src/main/java/org/eolang/maven/Unpacked.java
androvonx95: Refactored to a plain class with `void unpack()` so no more unused return value.

eo-maven-plugin/src/test/java/org/eolang/maven/UnpackedTest.java
androvonx95: Removed this test too. The throw is already covered by `rejectsZipEntryThatEscapesDestination`, and if the throw stopped happening, `assertThrows` would fail before the file-exists check ran anyway.

androvonx95:
Rebased onto latest master and fixed the `NewlineAtEndOfFileCheck` violations flagged by qulice. All checks should now pass.

eo-maven-plugin/src/main/java/org/eolang/maven/Unpacking.java
```diff
@@ -0,0 +1,68 @@
...
+    Unpacked(final Path archive, final Path target) {
+        this.jar = archive;
+        this.dest = target;
+    }
+
+    /**
+     * Unpack the archive into {@link #dest}.
+     * @throws IOException If unpacking fails or an entry escapes {@link #dest}
+     */
+    void unpack() throws IOException {
```
yegor256: @androvonx95 how something that's already "unpacked" can be unpacked? think about better class and method names.

androvonx95: @yegor256 okay I renamed it

eo-maven-plugin/src/test/java/org/eolang/maven/ArchiveTest.java
```diff
@@ -0,0 +1,67 @@
...
+            "Safe zip entry must land inside the destination directory",
+            dest.resolve("org/eolang/ok.eo").toFile(),
+            FileMatchers.anExistingFile()
+        );
+    }
+
+    @Test
+    void rejectsZipEntryThatEscapesDestination(@Mktmp final Path temp) throws IOException {
+        final Path jar = temp.resolve("evil.jar");
+        ArchiveTest.jar(jar, "../evil-escaped.txt", "pwned");
```
morphqdd, line 52: No test for an entry with an absolute path (for example /etc/passwd or a Windows drive-qualified path). Path.resolve() returns an absolute other argument unchanged, so this is a distinct bypass shape from the relative ../ case covered here. The current check happens to catch it too, since the resolved target would not start with root, but that is not verified by any test.

eo-maven-plugin/src/test/java/org/eolang/maven/ArchiveTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Test
+    void extractsSafeEntriesInsideDestination(@Mktmp final Path temp) throws Exception {
+        final Path jar = temp.resolve("safe.jar");
+        ArchiveTest.jar(jar, "org/eolang/ok.eo", "[] > ok");
+        final Path dest = temp.resolve("extracted");
+        new Archive(jar).extract(dest);
+        MatcherAssert.assertThat(
+            "Safe zip entry must land inside the destination directory",
+            dest.resolve("org/eolang/ok.eo").toFile(),
```
morphqdd: Only checks that the extracted file exists, not that its content matches what was written into the jar.

eo-maven-plugin/src/test/java/org/eolang/maven/ArchiveTest.java
```diff
@@ -0,0 +1,67 @@
...
+            FileMatchers.anExistingFile()
+        );
+    }
+
+    @Test
+    void rejectsZipEntryThatEscapesDestination(@Mktmp final Path temp) throws IOException {
+        final Path jar = temp.resolve("evil.jar");
+        ArchiveTest.jar(jar, "../evil-escaped.txt", "pwned");
+        Assertions.assertThrows(
```
morphqdd, line 54: The PR description says this test also verifies that the escaping entry is not materialized, but the test body only asserts that IOException is thrown. There is no Files.notExists(...) check anywhere confirming evil-escaped.txt was never written.

eo-maven-plugin/src/main/java/org/eolang/maven/Archive.java
```diff
@@ -0,0 +1,63 @@
...
+    Archive(final Path path) {
+        this.file = path;
+    }
+
+    /**
+     * Extract entries into {@code dest}, refusing anything outside it.
+     * @param dest Destination directory
+     * @throws IOException If extraction fails or an entry escapes {@code dest}
+     */
+    void extract(final Path dest) throws IOException {
```
morphqdd, line 43: No limit on decompressed output size. A compromised dependency, the same threat model this fix targets, could ship a small, highly compressed jar that expands to fill the disk. Zip Slip and unbounded decompression are usually fixed together for that reason.

eo-maven-plugin/src/main/java/org/eolang/maven/Archive.java
```diff
@@ -0,0 +1,63 @@
...
+        final Path root = dest.toAbsolutePath().normalize();
+        Files.createDirectories(root);
+        try (ZipInputStream zip = new ZipInputStream(Files.newInputStream(this.file))) {
+            for (ZipEntry entry = zip.getNextEntry(); entry != null; entry = zip.getNextEntry()) {
```
morphqdd: If a Zip Slip entry is found partway through the archive, entries already extracted before it stay on disk. No cleanup on the failure path, and no test covers what state extract() leaves behind after a rejection.

eo-maven-plugin/src/main/java/org/eolang/maven/Archive.java
```diff
@@ -0,0 +1,63 @@
...
+                final Path target = root.resolve(entry.getName()).normalize();
```
morphqdd, line 50: An entry with an empty name resolves to root itself. If that entry is not flagged as a directory, this falls into the else branch and calls Files.copy(zip, root, REPLACE_EXISTING), trying to overwrite the destination directory with file content. Untested, and it is not obvious what should happen here.

morphqdd, line 50: entry.getName() containing a NUL character makes Path.resolve() throw InvalidPathException, which extends IllegalArgumentException, not IOException. That is not caught here, and CentralMaven's call site only catches IOException, so a hostile entry name crashes the build with an unrelated raw exception instead of the intended clean failure message.

eo-maven-plugin/src/main/java/org/eolang/maven/Archive.java
```diff
@@ -0,0 +1,63 @@
...
+        this.file = path;
+    }
+
+    /**
+     * Extract entries into {@code dest}, refusing anything outside it.
+     * @param dest Destination directory
+     * @throws IOException If extraction fails or an entry escapes {@code dest}
+     */
+    void extract(final Path dest) throws IOException {
+        final Path root = dest.toAbsolutePath().normalize();
```
morphqdd, line 44: normalize() only removes lexical . and .. segments, it does not resolve symlinks. If a symlink already sits on the path to dest, root can point somewhere other than the intended sandbox, and every entry would be considered safe relative to that. toRealPath() would close that gap.

eo-maven-plugin/src/test/java/org/eolang/maven/ArchiveTest.java
```diff
@@ -0,0 +1,67 @@
...
+        );
+    }
+
+    /**
+     * Writes a jar with alternating entry-name / content pairs.
+     * @param jar Destination jar path
+     * @param entries Alternating names and payloads
+     * @throws IOException If writing fails
+     */
+    private static void jar(final Path jar, final String... entries) throws IOException {
```
morphqdd, line 184: No test for a traversal pattern with more than one path segment (for example a/b/../../../evil.txt), only a single leading ../. The normalize() plus startsWith() logic likely handles it the same way, but that is not verified.

eo-maven-plugin/src/test/java/org/eolang/maven/ArchiveTest.java
morphqdd, line 52: No test for a malicious directory entry (a name ending in / with .. segments), only malicious file entries.

eo-maven-plugin/src/main/java/org/eolang/maven/Archive.java
```diff
@@ -0,0 +1,63 @@
...
+                            "Zip entry '%s' would write outside '%s'",
+                            entry.getName(), root
+                        )
+                    );
+                }
+                if (entry.isDirectory()) {
+                    Files.createDirectories(target);
+                } else {
+                    Files.createDirectories(target.getParent());
+                    Files.copy(zip, target, StandardCopyOption.REPLACE_EXISTING);
```
morphqdd, line 66: REPLACE_EXISTING means a later entry silently overwrites whatever an earlier entry in the same archive already wrote to the same normalized target. No test covers two entries that resolve to the same path.

eo-maven-plugin/src/test/java/org/eolang/maven/ArchiveTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Test case for {@link Archive}.
+ * @since 0.64.0
+ */
+@ExtendWith(MktmpResolver.class)
+final class ArchiveTest {
+
+    @Test
+    void extractsSafeEntriesInsideDestination(@Mktmp final Path temp) throws Exception {
```
morphqdd, line 31: No test for an archive with zero entries. Minor, but extract() creates the destination directory unconditionally before the loop even starts, and nothing confirms that is harmless on an empty jar.

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2870](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+994](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +12 points.  Your running score is [+3709](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6252, opened 2026-08-02
fix(#6214): fold compile-scope XMIRs into the WPA lint cache key

### LINKED ISSUE #6214: Whole-program-analysis lint cache key omits compile-scope XMIRs it analyses
## What Happens

`Linting.lintAll` builds its map of objects to analyze from both `this.tojos.withXmir()` and `this.compile.withXmir()`, so the whole-program-analysis (WPA) lint pass looks at compile-scope XMIRs too, not only the project's own sources.

Only `this.tojos` (the project's own sources) are run through `lintOne`, and `lintOne` is the only thing that writes into the `3-lint` cache directory that the lint cache key hashes. So a change confined to a compile-scope dependency's XMIR leaves the hashed `3-lint` directory byte-identical between builds, and the cached WPA verdict from before the dependency changed gets reused instead of being recomputed.

This is the same class of bug already fixed in #6030 (lint cache using a constant key so one project's result replaced another's), but narrower: here the cache key does capture the project's own sources correctly, it just doesn't account for the compile-scope inputs the WPA pass also reads.

Not reproduced end to end (needs a two-module build where a compile-scope dependency's XMIR changes between builds while the depending project's own sources stay the same), but the code path is clear: `lintAll`'s `pkg` map construction reads both tojos and compile scope, while the cache key construction and `lintOne`'s writes only cover tojos.

## What Should Happen

The lint cache key should also account for the compile-scope XMIRs the WPA pass reads, so a change in a compile-scope dependency invalidates the cached WPA result the same way a change in the project's own sources does.

androvonx95:
Closes #6214.

@yegor256

Follow-up to #6030 / #6149 (thanks @Thayorns @Favixx).

WPA reads scoped and compile-scope XMIRs, but the cache key only hashed `3-lint`, which `lintOne` fills from the scoped set. A compile-scope-only change left the key identical and reused a stale verdict.

The cache path now includes a fingerprint of every XMIR WPA reads, plus `skipProgramLints` / `skipExperimentalLints`, via `CachePath` (same shape as `lintOne`). The package result is excluded from the freshness hash by absolute path so an object named `wpa` cannot drop out of it.

`invalidatesWholeProgramAnalysisCacheWhenCompileScopeChanges` covers the `eo.scope=test` case. `MjLintTest` and `-P qulice` pass.


eo-maven-plugin/src/main/java/org/eolang/maven/WpaCacheKey.java
```diff
@@ -0,0 +1,78 @@
...
+            for (final Map.Entry<String, Path> ent : new TreeMap<>(this.paths).entrySet()) {
+                digest.update(ent.getKey().getBytes(StandardCharsets.UTF_8));
+                digest.update((byte) 0);
+                digest.update(new Sha(ent.getValue()).toString().getBytes(StandardCharsets.UTF_8));
+                digest.update((byte) 0);
+            }
+            final List<String> sorted = new ArrayList<>(this.skips);
+            sorted.sort(String::compareTo);
+            for (final String lint : sorted) {
+                digest.update(lint.getBytes(StandardCharsets.UTF_8));
```
Thayorns, line 66: The skip lints go into the digest concatenated with no separator, unlike the path loop above which delimits every field with a zero byte. So skipProgramLints of `["ab", "c"]` and `["a", "bc"]` hash to the same key and share a cache entry, which is the same stale-verdict failure this change is closing elsewhere.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -183,11 +184,9 @@ void savesForWholeProgramAnalysisResultsToCache(@Mktmp final Path temp) throws I
             );
         maven.execute(new FakeMaven.Lint());
         MatcherAssert.assertThat(
-            "WPA results must be saved to cache",
-            cache.resolve(Linting.CACHE)
-                .resolve(FakeMaven.pluginVersion())
-                .resolve("wpa.xmir").toFile(),
-            FileMatchers.anExistingFile()
+            "WPA results must be saved under the versioned cache path",
```
Thayorns: The message says "under the versioned cache path" but `cached` walks the whole cache root and matches any file named wpa.xmir, so it would pass with the file sitting directly in the cache root or under any other version. The check it replaced named the exact path including the plugin version; nothing left now would notice the version dropping out of the key.

Thayorns (commented):
Folds a fingerprint of every XMIR that WPA reads, plus the two skip flags, into the cache path so a compile-scope-only change no longer reuses a stale verdict. The per-path part of the digest is done carefully, with a sorted map and null separators. Two notes below, one on the digest itself and one on what the cache test now proves.

yegor256:
@androvonx95 conflicts here

androvonx95:
@yegor256 conflicts resolved. Please take another look at this PR whenever you again 🙂

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -396,14 +395,20 @@ private int lintAll(
...
-            final Path wpa = Paths.get("wpa.xmir");
+            final Path wpa = Path.of("wpa.xmir");
             final Path base = this.targetDir.resolve(Linting.DIR);
             final Path target = base.resolve(wpa);
             Files.createDirectories(base);
             this.guard.apply(
                 base, target, wpa,
                 new Cache(
-                    this.cacheDir.resolve(Linting.CACHE).resolve(this.version),
+                    new CachePath(
```
morphqdd, line 405: Before this PR the WPA cache lived in one directory reused across runs (`<cache>/linted/<version>/`). Now the path includes a content hash, so every distinct combination of tracked XMIR content and skip-lint config gets its own subdirectory with a `wpa.xmir` and a `.sha256` sidecar, and nothing ever removes an old one. During active development that could be a new directory per commit. Worth a `@todo` puzzle noting the cache needs an eviction pass, the way this project usually marks a known gap instead of leaving it silent.

eo-maven-plugin/src/main/java/org/eolang/maven/WpaCacheKey.java
```diff
@@ -0,0 +1,79 @@
...
+        this.skips = programlints;
+        this.experimental = skip;
+    }
+
+    @Override
+    public String get() {
+        try {
+            final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            for (final Map.Entry<String, Path> ent : new TreeMap<>(this.paths).entrySet()) {
+                digest.update(ent.getKey().getBytes(StandardCharsets.UTF_8));
```
morphqdd, line 58: `digest.update(bytes); digest.update((byte) 0);` appears three times in this method: twice here for the identifier and its hash, once more below for each lint name. A small private helper doing the write-plus-terminator once would remove the repetition.

eo-maven-plugin/src/main/java/org/eolang/maven/WpaCacheKey.java
```diff
@@ -0,0 +1,79 @@
...
+                digest.update((byte) 0);
+                digest.update(new Sha(ent.getValue()).toString().getBytes(StandardCharsets.UTF_8));
```
morphqdd, line 60: This computes a fresh SHA-256 over every tracked XMIR's content on every lint run just to build the cache key, right before `Cache`'s own `dirSha` hashes the `3-lint` directory again for the actual freshness check. Two hashing passes over a mostly overlapping file set.

eo-maven-plugin/src/main/java/org/eolang/maven/WpaCacheKey.java
```diff
@@ -0,0 +1,79 @@
...
+    private final boolean experimental;
+
+    /**
+     * Ctor.
+     * @param files XMIR files WPA reads
+     * @param programlints Program lints to skip
+     * @param skip Whether experimental lints are skipped
+     */
+    WpaCacheKey(final Map<String, Path> files, final Collection<String> programlints,
+        final boolean skip) {
```
morphqdd, line 47: The ctor parameter is named `skip`, but it's stored into the `experimental` field, and the class already has a `skips` field for the collection a few lines up. `skip` / `skips` sit right next to each other for two unrelated things.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -245,36 +245,68 @@ void savesForWholeProgramAnalysisResultsToCache(@Mktmp final Path temp) throws I
...
-            FileMatchers.anExistingFile()
-        );
+        try (
+            Stream<Path> walk = Files.walk(
+                cache.resolve(Linting.CACHE).resolve(FakeMaven.pluginVersion())
+            )
+        ) {
+            MatcherAssert.assertThat(
+                "WPA results must be saved under the versioned cache path",
+                walk.anyMatch(path -> "wpa.xmir".equals(path.getFileName().toString())),
```
morphqdd, line 255: `anyMatch(path -> "wpa.xmir".equals(path.getFileName().toString()))` over the whole `linted/<version>/` subtree would still pass if the file landed in the wrong hash directory, since it never checks where under the tree it is. The previous assertion checked the exact expected path; this one only checks that a file with that name exists somewhere.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -245,36 +245,68 @@ void savesForWholeProgramAnalysisResultsToCache(@Mktmp final Path temp) throws I
...
             () -> MjLintTest.linting(
-                temp.resolve("broken"), cache, MjLintTest.probmlematic()
+                temp.resolve("broken"), cache, MjLintTest.problematic()
             ).execute(new FakeMaven.Lint()),
             "WPA error of the second project must be reported, not hidden by the cache of the first"
         );
     }
 
+    @Test
+    void invalidatesWholeProgramAnalysisCacheWhenCompileScopeChanges(@Mktmp final Path temp)
```
morphqdd, line 277: This covers a compile-scope XMIR changing content. There's no test for a compile-scope XMIR disappearing entirely between two lint runs (removed from `this.compile.withXmir()`) — a different edge of the same fix.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -569,49 +601,51 @@ private static String[] critical() {
...
     private static FakeMaven linting(
         final Path workspace, final Path cache, final String... program
     ) throws IOException {
-        return new FakeMaven(workspace)
+        final FakeMaven maven = new FakeMaven(workspace)
             .with("lintAsPackage", true)
-            .with("cache", cache.toFile())
-            .with("sourcesDir", workspace.resolve("foo").toFile())
-            .withProgram(program);
+            .with("cache", cache.toFile());
```
morphqdd, line 616: `.with("sourcesDir", workspace.resolve("foo").toFile())` was dropped from this chain. The old javadoc explained it was there specifically to keep sources apart from the target directory, the way a real Maven build does with `src/main/eo`. Without it, these tests now run with sources and target sharing the same root, which the removed comment called out as intentionally avoided.

0crat:
@androvonx95 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (192 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+44](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+913](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (7) comments.  Your running score is [+3478](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6253, opened 2026-08-02
#6233: enforce mandatory blank line before +>/--> test attribute on non-formation lines

### LINKED ISSUE #6233: Mandatory blank line rule is not enforced for non-formation lines with `+>`/`-->` test attributes
## What Happens

The mandatory-blank-line rule for `+>`/`-->` test attributes (R-6.5.3, added by the fix for issue #5444) only applies to formation and only-phi-formation lines. On every other line kind that can carry a test-attribute suffix - a plain application, a method chain, a reversed-dispatch line, a compact-tuple line, or a text block - the rule is not just unenforced, it's inverted.

`Blanks.checkTest` (which requires exactly one blank line before a test attribute) is only called from `LnFormation` and `LnOnlyPhi`, each gated on `suffix.test()`. Every other line kind - `LnApplication`, `LnMethod`, `LnReversed`, `LnCompactTuple`, `LnTextBlock`, plus `LnPipe` and `LnVoid` - unconditionally calls `Blanks.checkPlain` instead, before the suffix (and therefore whether it's a test) is even known. `checkPlain` does the opposite of what a test attribute needs: it *rejects* a blank line in front of the object, on the theory that only master objects (formations, atoms, only-phi formations, and `+>` tests, per the error message's own text) are allowed a preceding blank.

Confirmed with a real parse, an application line ending in a test suffix:

```
[] > app
  42 > x
  eq. +> t
    x
    x
```

parses with zero errors even though R-6.5.3 requires a blank line before this test attribute - the requirement is never enforced for an application line. Adding the blank line to actually follow R-6.5.3:

```
[] > app
  42 > x

  eq. +> t
    x
    x
```

now fails to parse, with:

```
[4:2] error: 'blank line before a plain object is forbidden (R-6.5.4); only master objects (formations, atoms, only-phi formations, +> tests) may be preceded by a blank line'
```

The error message itself lists `+> tests` as one of the four exempt object kinds, yet the code path that produced the error is exactly the one that fails to treat this test attribute as exempt.

## What Should Happen

The line kinds that currently call `Blanks.checkPlain` unconditionally (`LnApplication`, `LnMethod`, `LnReversed`, `LnCompactTuple`, `LnTextBlock`) should check `suffix.test()` first and call `Blanks.checkTest` instead when the line carries a `+>`/`-->` suffix, mirroring exactly what `LnFormation` and `LnOnlyPhi` already do. A test attribute on any of these line kinds should then require the mandatory blank line the same way a formation's test attribute does, instead of forbidding it.

morphqdd:
Closes #6233

`LnApplication`, `LnMethod`, `LnReversed`, `LnCompactTuple`, and `LnTextBlock` all called `Blanks.checkPlain` unconditionally, before their suffix was even parsed, so a `+>`/`-->` test attribute on any of these line kinds never got the mandatory-blank-line check (R-6.5.3) that `LnFormation` and `LnOnlyPhi` already enforce, and instead got the opposite rule (blank line forbidden).

Moved the blank check to after the suffix is parsed and branch on `suffix.test()`, calling `Blanks.checkTest` instead of `Blanks.checkPlain` when the line carries a test attribute, mirroring `LnFormation`/`LnOnlyPhi` exactly.

Also fixed three existing `eo-syntax` fixtures (`tests.yaml`, `atoms-with-inner-tests.yaml`, `application-in-tests.yaml`) that had test attributes on application/reversed-dispatch lines with no blank line above them. These previously passed only because of this bug; they now need the blank line the rule always required.

Added reject/accept unit test pairs to all 5 `Ln*Test` classes, matching the existing pattern in `LnFormationTest`.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnApplication.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnApplication.java
@@ -57,7 +57,6 @@ final class LnApplication implements Line {
 
     @Override
     public void into(final Stack stack, final Globals globals, final Emit emit) {
-        Blanks.checkPlain(this.span, globals, emit);
         final Tokens tokens = new Tokens(this.span.body(), this.span);
         final Value head = tokens.readValue();
         final List<MethodChain> chain = tokens.readChain();
@@ -67,6 +66,11 @@ final class LnApplication implements Line {
         final Suffix suffix = new Suffix(
             tokens.tail(), this.span, this.span.indent() + tokens.cursor()
         );
+        if (suffix.test()) {
+            Blanks.checkTest(this.span, globals, emit);
+        } else {
+            Blanks.checkPlain(this.span, globals, emit);
+        }
         if (head.kind() == Value.Kind.GROUP
             && chain.isEmpty() && args.isEmpty() && outer == null) {
             throw new ParseError(
--- a/eo-parser/src/main/java/org/eolang/parser/LnCompactTuple.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnCompactTuple.java
@@ -58,7 +58,6 @@ final class LnCompactTuple implements Line {
 
     @Override
     public void into(final Stack stack, final Globals globals, final Emit emit) {
-        Blanks.checkPlain(this.span, globals, emit);
         final Tokens tokens = new Tokens(this.span.body(), this.span);
         final Value head = tokens.readValue();
         final List<MethodChain> chain;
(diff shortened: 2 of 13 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (183 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3177](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6255, opened 2026-08-02
#6200: restore-aliases.xsl drops +alias for +also-only references

### LINKED ISSUE #6200: `restore-aliases.xsl` drops `+alias` when its target is referenced only via `+also`, over-qualifying working code
## What Happens

`restore-aliases.xsl` decides whether to keep an `+alias` meta by counting how many times its target is referenced. It only scans `o/@base`, `o/@atom`, `o/@args` and `o/@type`. It never scans `+also` meta values, even though the parser's own `resolve-aliases.xsl` resolves `+also X` through the exact same alias table (see the existing `resolve-aliases.yaml` parser test: `+also my` resolves to `tail='Φ.org.eolang.xyz'`).

So a target referenced once as a normal object and once via `+also` looks "referenced once" to the counter, when it is genuinely used twice, and the alias gets dropped instead of kept.

Verified by building `eo-parser` + `eo-printer` from `master` and running the real parse and print pipeline.

Input:

```
+alias eo3.pardon.appellation
+also appellation
+package eo3.skeptic

[] > human
  appellation > a
```

Printed output:

```
+also Q.eo3.pardon.appellation
+package eo3.skeptic

[] > human
  eo3.pardon.appellation > a
```

The `+alias` is dropped and both usages end up fully qualified, even though this is exactly the "referenced more than once" case #6184 added this pass to handle.

## What Should Happen

`$refs` in `restore-aliases.xsl` should also include the resolved values of `+also` meta parts, so a target used via `+also` counts toward the "referenced more than once" threshold the same as any other usage. Expected output for the input above is the input unchanged (it should round-trip identically).

morphqdd:
`restore-aliases.xsl` decides whether to keep an `+alias` by counting how many times its target is referenced, but it only scanned `@base`, `@atom`, `@args` and `@type`. It never scanned the resolved value of a `+also` meta, even though the parser's own `resolve-aliases.xsl` resolves `+also X` through the same alias table. So a target referenced once as a normal object and once via `+also` looked "referenced once" and its `+alias` was dropped instead of kept.

This adds the resolved `+also` parts to the reference count, and adds a print-side template so the `+also` meta itself is shortened back to the alias name when the alias is kept (it was staying fully qualified even before this bug, since nothing restored it).

Added `alias-referenced-via-also.yaml` reproducing the exact case from the issue. Confirmed it fails on unpatched code with the same output shown in the issue, and passes with the fix.

Closes #6200

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/restore-aliases.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/restore-aliases.xsl
@@ -31,12 +31,13 @@
   <xsl:variable name="aliases" select="/object/metas/meta[head = 'alias']"/>
   <!-- Every name already bound in the file, so an alias short name that clashes with one is left alone. -->
   <xsl:variable name="bound" select="distinct-values((//o/@name, //o/@local))"/>
-  <!-- Every reference string, counted with multiplicity, across the four spots resolve-aliases rewrote. -->
+  <!-- Every reference string, counted with multiplicity, across the five spots resolve-aliases rewrote. -->
   <xsl:variable name="refs" as="xs:string*">
     <xsl:sequence select="//o/@base/string()"/>
     <xsl:sequence select="//o/@atom/string()"/>
     <xsl:sequence select="for $a in //o/@args, $t in tokenize($a, ' ')[. != ''] return $t"/>
     <xsl:sequence select="for $t in //o/@type return replace($t, '\?$', '')"/>
+    <xsl:sequence select="/object/metas/meta[head = 'also']/part/string()"/>
   </xsl:variable>
   <!-- Whether an alias's short name reads back unambiguously: no bound name and no bare global claims it. -->
   <xsl:function name="eo:free" as="xs:boolean">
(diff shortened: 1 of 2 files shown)


morphqdd:
@yegor256, please review

morphqdd:

Thayorns (approved):
The count now includes the resolved `+also` parts, and only the parts, not the tail as well, which is right because resolve-aliases.xsl rewrites both and counting both would double every `+also` reference. Rebuilding the tail from the shortened parts rather than shortening the tail text handles the multi-name `+also` that resolve-aliases leaves alone, since its template only fires when the whole tail matches an alias name. One path the pack does not walk is the mirror of the case in the issue: an alias reached only through a single `+also` and nowhere else, where the alias is dropped and the `+also` is left fully qualified by the new templates.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+605](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3311](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6256, opened 2026-08-02
#6234: validate write/send size against buffer length before native call

### LINKED ISSUE #6234: `posix`/`win32` `write`/`send` functions can read past buffer end, leading to JVM memory leaks
## What Happens

The native `posix "write"` and `posix "send"` syscall wrappers (and their `win32` counterparts) pass the caller-supplied `size` argument straight to the underlying libc call without validating it against the actual length of the buffer being written, letting a program read and leak arbitrary JVM process memory into a file or socket.

`WriteSyscall.java` takes `params[1]` (the buffer) and `params[2]` (the size) and calls libc `write(fd, buf, size)` with the raw `size` value. JNA only allocates a native buffer sized to `buf.length`; nothing checks that `size` does not exceed that length before the call reaches libc. The same pattern is present in `SendSyscall.java`, `win32/WriteFuncCall.java`, and `win32/SendFuncCall.java`.

Confirmed by calling `posix "write" (* fd "abcd" 4096)` against a real temporary file, where the buffer `"abcd"` is 4 bytes but `size` is 4096:

- `write()` returned `1264`, and the file it wrote to is 1264 bytes long.
- Bytes 4 through 1263 are not part of the 4-byte input at all - they are raw JVM heap contents that happened to sit adjacent to the allocated native buffer. The leaked bytes decode partially to recognizable class-file strings (`bject;`, `oke/LambdaForm`, `Code`, `RuntimeVisi`), confirming this is genuine process memory disclosure, not garbage.

This is not reachable through the higher-level `file.eo`, `console.eo`, or `socket.eo` objects, which all correctly pass `buffer.size` as the length. But `posix` and `win32` are themselves public, documented EO objects that a user program can call directly with any size value, and this is also a plain out-of-bounds native memory read that can crash the process outright under a different heap layout, independent of the information-disclosure angle.

## What Should Happen

`WriteSyscall`, `SendSyscall`, and their `win32` counterparts should validate that `size` does not exceed the actual length of the supplied buffer before making the native call, either rejecting the call with a clear error when it does, or clamping to the buffer's real length. The native call must never read past the end of the buffer it was actually given.

morphqdd:
Closes #6234

`WriteSyscall`/`SendSyscall` (posix) and `WriteFuncCall`/`SendFuncCall` (win32) passed the caller-supplied `size` straight to the native `write`/`send` call without checking it against the actual length of the supplied buffer. A caller could pass a small buffer with a large `size`, causing the native call to read past the end of the JNA-allocated buffer and leak adjacent JVM memory into a file or socket (or crash, depending on heap layout).

Each of the four classes now checks `size <= buf.length` and throws a clear `ExFailure` before making the native call, instead of passing the mismatched size through.

Added EO-level throwing tests to `posix.eo` and `win32.eo` (`stops-on-write-size-exceeding-buffer-length`) verifying that an oversized `size` argument is now rejected rather than silently accepted. Verified locally that the posix test fails with "Expected Exception to be thrown, but nothing was thrown" when the fix is reverted, and passes with the fix in place. The win32 test is gated to only exercise the real win32 call on Windows; it is a no-op pass on other platforms since the CI runners for other OSes don't have win32 syscalls at all.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/posix/WriteSyscall.java
+++ b/eo-runtime/src/main/java/org/eolang/posix/WriteSyscall.java
@@ -6,6 +6,7 @@ package org.eolang.posix;
 
 import org.eolang.Data;
 import org.eolang.Dataized;
+import org.eolang.ExFailure;
 import org.eolang.PhDefault;
 import org.eolang.Phi;
 import org.eolang.Syscall;
@@ -31,14 +32,22 @@ public final class WriteSyscall implements Syscall {
 
     @Override
     public Phi make(final Phi... params) {
+        final byte[] buf = new Dataized(params[1]).take();
+        final int size = new Dataized(params[2]).asNumber().intValue();
+        if (size > buf.length) {
+            throw new ExFailure(
+                "Can't write %d bytes from a buffer of only %d bytes",
+                size, buf.length
+            );
+        }
         final Phi result = this.posix.take("return").copy();
         result.put(
             0,
             new Data.ToPhi(
                 CStdLib.INSTANCE.write(
                     new Dataized(params[0]).asNumber().intValue(),
-                    new Dataized(params[1]).take(),
-                    new Dataized(params[2]).asNumber().intValue()
+                    buf,
+                    size
                 )
             )
         );
--- a/eo-runtime/src/main/java/org/eolang/posix/WriteSyscall.java
+++ b/eo-runtime/src/main/java/org/eolang/posix/WriteSyscall.java
@@ -34,7 +34,7 @@ public final class WriteSyscall implements Syscall {
     public Phi make(final Phi... params) {
         final byte[] buf = new Dataized(params[1]).take();
         final int size = new Dataized(params[2]).asNumber().intValue();
-        if (size > buf.length) {
+        if (size < 0 || size > buf.length) {
             throw new ExFailure(
                 "Can't write %d bytes from a buffer of only %d bytes",
                 size, buf.length
--- a/eo-runtime/src/main/eo/posix.eo
+++ b/eo-runtime/src/main/eo/posix.eo
@@ -114,3 +114,21 @@
           "inet_addr"
           "127.0.0.1"
       16777343
+
+  --> stops-on-write-size-exceeding-buffer-length
+    os.is-windows.if > @
+      1.div 0
+      seq *
+        code.
+          posix *1
+            "write"
+            fd
+            "ab"
+            4096
+        true
+    code. > fd
+      posix *1
+        "open"
+        "/dev/null"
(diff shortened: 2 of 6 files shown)


morphqdd:

morphqdd:

eo-runtime/src/main/java/org/eolang/posix/WriteSyscall.java
```diff
@@ -31,14 +32,22 @@ public WriteSyscall(final Phi posix) {
 
     @Override
     public Phi make(final Phi... params) {
+        final byte[] buf = new Dataized(params[1]).take();
+        final int size = new Dataized(params[2]).asNumber().intValue();
+        if (size > buf.length) {
```
Thayorns: The guard only bounds size from above, so a negative size sails through it. `asNumber().intValue()` will hand back -1 for `write fd "ab" -1`, and JNA widens that to a size_t of 2^64-1 at the native boundary, which is the same read past the end of the JNA buffer this change is closing. SendSyscall.java:36, WriteFuncCall.java:37 and SendFuncCall.java:38 have it identically.

Thayorns (commented):
Bounds the caller's size against the buffer before handing it to write/send in all four syscall wrappers, and the EO tests gate the platform-specific call the right way round in each file. The same seven lines now sit in four classes, differing only in the verb inside the message and which native library is called, so the next change to this check has four places to land in. The packs only exercise `write` on both platforms; `send`, which takes the same unchecked size, has no test either way.

eo-runtime/src/main/java/org/eolang/posix/WriteSyscall.java
morphqdd: Fixed, same guard (`size < 0 || size > buf.length`) in all four classes. Added a mirrored EO test (`stops-on-write-size-being-negative`) next to the existing exceeding-buffer-length one in posix.eo/win32.eo.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (124 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3323](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+672](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6257, opened 2026-08-02
#6235: include skipSourceLints and skipExperimentalLints in the per-file lint cache key

### LINKED ISSUE #6235: `skipSourceLints` and `skipExperimentalLints` flags not included in per-file lint cache key causing stale caching results
## What Happens

The per-file lint cache key does not account for the `skipSourceLints` or `skipExperimentalLints` build flags, even though both change what `Linting`'s per-file lint pass actually reports for the exact same source file. Toggling either flag between two builds that share a cache directory can silently reuse a stale cached lint result from before the flag changed.

The cache key is built from the plugin version, a hash of the tojo, and the source XMIR's hash (`Linting.java:329-341`), but the cached value is the result of `this.linted(xmir, unlints)`, whose content depends on `unlints` (derived from `skipSourceLints`) and on `this.skipExperimentalLints` (consulted separately at `Linting.java:477-481`). Neither flag is part of the key.

Confirmed with a five-run sequence against `FakeMaven`, all sharing one cache directory and the same source program:

1. No `skipSourceLints` set: `unlint-non-existing-defect` is reported and the result is cached.
2. `skipSourceLints={unlint-non-existing-defect}` set, same cache directory: the defect is **still reported** - the stale, pre-flag cached result is reused instead of being recomputed.
3. Same config as run 2, but a fresh cache directory: the defect is correctly suppressed (control, confirming the flag itself works when the cache isn't stale).
4. With the skip flag and a fresh cache: suppressed and cached.
5. Same cache as run 4, but with the skip flag removed: the defect is **silently missing** even though it should now be reported again.

With the default `failOnWarning=true`, a build that should fail on a real defect can pass instead, purely because of stale caching keyed on incomplete state. This is the same bug class already fixed for the whole-program-analysis lint cache in closed issue #6030 (constant cache key) and reported again for its own scope gap, but this is the separate per-file lint cache, and the specific inputs missing from its key are different.

## What Should Happen

`skipSourceLints` and `skipExperimentalLints` should both be part of the per-file lint cache key, so that changing either flag invalidates any previously cached result for the same source, the same way `trackLocations`/`coverageTracking` were added to the transpile cache key when that class of bug was fixed there.

morphqdd:
Closes #6235

The per-file lint cache key in `Linting.java` was built from the plugin version, a hash of the tojo, and the source XMIR's hash, but did not include the `skipSourceLints` or `skipExperimentalLints` flags. Both change what `Linting.linted()` reports for the exact same source file, so toggling either flag between builds sharing a cache directory could silently reuse a stale cached lint result from before the flag changed.

Added a `version()` method to `Linting` that folds both flags into the cache-key version segment, the same way `trackLocations`/`coverageTracking` are already folded into the transpile cache key in `Transpilation.version()`.

Added a regression test, `MjLintTest#invalidatesPerFileCacheWhenSkipSourceLintsChanges`, verified to fail without the fix (the second run incorrectly reused the first run's cached defect) and pass with it.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
@@ -53,6 +53,7 @@ import org.xembly.Xembler;
  *     Naming the class {@code Linting} avoids this collision.
  * </p>
  * @since 0.31.0
+ * @checkstyle ClassFanOutComplexityCheck (3 lines)
  */
 @SuppressWarnings("PMD.GodClass")
 final class Linting implements Step {
@@ -308,7 +309,9 @@ final class Linting implements Step {
         return String.format(
             "%s-%s-%b",
             this.version,
-            this.skipSourceLints.stream().sorted().collect(Collectors.joining(",")),
+            new Hashed(
+                this.skipSourceLints.stream().sorted().collect(Collectors.joining(","))
+            ).get(),
             this.skipExperimentalLints
         );
     }
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
@@ -297,6 +297,23 @@ final class Linting implements Step {
     }
 
     /**
+     * Cache-key version segment: the plugin version combined with the
+     * {@code skipSourceLints} and {@code skipExperimentalLints} flags,
+     * since both change what {@link #linted(XML)} produces for the same
+     * source XMIR (see #6235), the same way {@code trackLocations}/
+     * {@code coverageTracking} were folded into the transpile cache key.
+     * @return The version segment for {@link CachePath}
+     */
+    private String version() {
+        return String.format(
+            "%s-%s-%b",
+            this.version,
+            this.skipSourceLints.stream().sorted().collect(Collectors.joining(",")),
+            this.skipExperimentalLints
+        );
+    }
+
+    /**
      * XMIR verified to another XMIR.
      * @param tojo Foreign tojo
      * @param counts Counts of errors, warnings, and critical
@@ -321,7 +338,7 @@ final class Linting implements Step {
                 new Cache(
                     new CachePath(
                         this.cacheDir.resolve(Linting.CACHE),
-                        this.version,
+                        this.version(),
                         new TojoHash(tojo).get()
                     ),
                     src -> this.linted(xmir).toString()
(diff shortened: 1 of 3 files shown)


morphqdd:

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -296,6 +296,23 @@ private void linting() throws IOException {
...
+     * since both change what {@link #linted(XML)} produces for the same
+     * source XMIR (see #6235), the same way {@code trackLocations}/
+     * {@code coverageTracking} were folded into the transpile cache key.
+     * @return The version segment for {@link CachePath}
+     */
+    private String version() {
+        return String.format(
+            "%s-%s-%b",
+            this.version,
+            this.skipSourceLints.stream().sorted().collect(Collectors.joining(",")),
```
Thayorns: The skipped lint names go into the segment verbatim, and CachePath.get() feeds that string to `base.resolve(this.semver)`, so it lands as a single directory name. Names in this repo run to `unlint-non-existing-defect` length, so a project skipping a dozen of them produces a component past the 255-byte limit ext4 and APFS enforce and the whole cache dies on ENAMETOOLONG rather than just missing. The trackLocations/coverageTracking precedent you followed folds booleans, which cannot grow.

Thayorns (commented):
Folds the two flags that change what linted() reports into the cache-key version segment, which is the right place for them, and the regression test does show the second run recomputing instead of reading the first run's entry. Only skipSourceLints is exercised; skipExperimentalLints goes into the same segment and nothing here would notice if it stopped doing so.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
morphqdd: Fixed. skipSourceLints now gets folded into a short SHA-256 digest (Hashed, a new small class matching the existing Fingerprint/Sha idiom here) instead of joined verbatim, so the path segment stays fixed-width regardless of how many or how long the skipped lint names are.

yegor256:
@morphqdd conflicts

morphqdd:
@yegor256 the conflict is with my own earlier PR #6271, which already landed the same fix on master. `Linting.cacheVersion()` there is exactly what this branch adds, just under a different method name. Nothing left to merge, so I'd close this one.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+812](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6258, opened 2026-08-02
#6236: fold a fingerprint of parse-stage XSLs into the parse cache key

### LINKED ISSUE #6236: Parse cache key lacks XSL fingerprint, causing stale XMIR with edited parse-stage stylesheets
## What Happens

The parse cache key has no fingerprint of the parser XSL stylesheets it applies, so editing one of those stylesheets while a `~/.eo` cache from before the edit still exists produces stale, silently wrong XMIR.

`Parsing.java`'s cache key is `pluginVersion + "-" + sha256(object names)` - it does not include any hash of the 20 `Canonical` parse-stage XSLs (`set-locators.xsl`, `build-fqns.xsl`, etc.) that actually transform the parsed source into its final XMIR shape. A `Fingerprint` class already exists specifically to hash a stylesheet chain including its imports (added for closed issue #5578, hardened by closed issue #6032), but it is referenced from exactly one place in the whole codebase - `Transpilation.java:170`, for the transpile cache. `Parsing` and `Linting` never use it.

Confirmed by patching `parse/set-locators.xsl` on the classpath between two runs sharing one cache directory (the patch prefixes every `@loc` value with `ZCPROBE-`, a trivially observable change):

1. Before the patch, with an empty cache: XMIR shows `loc="Φ.foo.x.main"` (unmodified, correct for that build).
2. After the patch, reusing the same cache directory: XMIR still shows `loc="Φ.foo.x.main"`, byte-identical to before the patch, even though the stylesheet that computes `loc` has changed.
3. After the patch, with a fresh cache directory: XMIR correctly shows `loc="ZCPROBE-Φ.foo.x.main"` (control, confirming the patch itself works and only the cache is masking it).

This means anyone iterating on a parse-stage XSL against a SNAPSHOT plugin version, with the default cache directory still populated from before their edit, gets silently stale XMIR with no indication anything is wrong.

## What Should Happen

The parse cache key (and, separately, the per-file lint cache key) should include a `Fingerprint` of the parse XSLs actually applied, the same way the transpile cache key already does for the transpile XSLs, so that editing a parse stylesheet invalidates any previously cached XMIR for the same source.

morphqdd:
Closes #6236

The parse cache key in `Parsing.java` was built from the plugin version, a digest of the known object names, and a hash of the tojo, but did not include any fingerprint of the 20 canonical parse-stage XSLs (`set-locators.xsl`, `build-fqns.xsl`, etc.) that actually transform the parsed source into its final XMIR shape. Editing one of these stylesheets while a cache from before the edit still exists produced stale, silently wrong XMIR.

A `Fingerprint` class already exists for exactly this (added for #5578, hardened by #6032) and is already used by the transpile cache key in `Transpilation.version()`. `Parsing` never used it.

Added a `PARSE_XSLS` array (the parse-stage XSLs plus the `_funcs.xsl`/`_specials.xsl` libraries they `xsl:import`, matching `org.eolang.parser.Canonical`'s own pipeline) and a `version(String)` method that folds their fingerprint into the cache key, mirroring `Transpilation.version()`.

Added `ParsingTest` verifying the cache-key version reflects the digest it's given (and, since `Fingerprint` throws on a missing classpath resource, that `PARSE_XSLS` itself resolves correctly). Updated `MjParseTest#parsesWithCache`'s cache-path helper to match the new 3-part key format; verified it fails without the fix (the manually seeded cache entry is missed and the source is silently re-parsed instead) and passes with it.

### DIFF
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjParseTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjParseTest.java
@@ -350,7 +350,7 @@ final class MjParseTest {
         return String.format(
             "%s-%s-%s",
             FakeMaven.pluginVersion(),
-            new Fingerprint(Parsing.PARSE_XSLS).get(),
+            new Fingerprint(Parsing.PARSE_XSLS.toArray(new String[0])).get(),
             new UncheckedText(
                 new HexOf(new Sha256DigestOf(new InputOf(identifier)))
             ).asString()
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjParseTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjParseTest.java
@@ -348,8 +348,9 @@ final class MjParseTest {
      */
     private static String cacheVersion(final String identifier) {
         return String.format(
-            "%s-%s",
+            "%s-%s-%s",
             FakeMaven.pluginVersion(),
+            new Fingerprint(Parsing.PARSE_XSLS).get(),
             new UncheckedText(
                 new HexOf(new Sha256DigestOf(new InputOf(identifier)))
             ).asString()
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/ParsingTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/ParsingTest.java
@@ -34,6 +34,17 @@ final class ParsingTest {
         );
     }
 
+    @Test
+    void includesXslFingerprintInVersion() {
+        MatcherAssert.assertThat(
+            "the cache-key version must include the fingerprint of the canonical parse-stage XSLs, not just the plugin version and the digest",
+            ParsingTest.parsing().version("some-digest"),
+            Matchers.containsString(
+                new Fingerprint(Parsing.PARSE_XSLS.toArray(new String[0])).get()
+            )
+        );
+    }
+
     /**
      * A minimal {@link Parsing} instance for testing its private helpers.
      * @return A new instance
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/ParsingTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/ParsingTest.java
@@ -1,0 +1,51 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import java.nio.file.Paths;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link Parsing}.
+ * @since 0.1
+ */
+final class ParsingTest {
+
+    @Test
+    void includesDigestInVersion() {
+        MatcherAssert.assertThat(
+            "the cache-key version must end with the digest of the known objects",
+            ParsingTest.parsing().version("some-digest"),
+            Matchers.endsWith("-some-digest")
+        );
+    }
+
+    @Test
+    void changesVersionWhenDigestChanges() {
+        final Parsing parsing = ParsingTest.parsing();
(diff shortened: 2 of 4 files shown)


morphqdd:

eo-maven-plugin/src/test/java/org/eolang/maven/ParsingTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link Parsing}.
+ * @since 0.1
+ */
+final class ParsingTest {
+
+    @Test
+    void includesDigestInVersion() {
```
Thayorns, line 19: Both tests here vary only the digest, which was already the tail of the key before this change, so deleting the Fingerprint term from version() leaves them green. The first one asserts that a `%s-%s-%s` format ending in the digest ends in the digest, which is its own input read back. The only thing that would actually catch the fingerprint going missing is the MjParseTest helper.

eo-maven-plugin/src/test/java/org/eolang/maven/MjParseTest.java
```diff
@@ -348,8 +348,9 @@ void parsesWithTargetCache(@Mktmp final Path temp) throws IOException {
      */
     private static String cacheVersion(final String identifier) {
         return String.format(
-            "%s-%s",
+            "%s-%s-%s",
             FakeMaven.pluginVersion(),
+            new Fingerprint(Parsing.PARSE_XSLS).get(),
```
Thayorns: The expected cache path is now derived from the same PARSE_XSLS the production key is derived from, so the test agrees with whatever that array happens to say. The failure the docblock warns about, PARSE_XSLS drifting out of sync with Canonical.Pipeline when someone adds a stylesheet there, produces a wrong-but-consistent key on both sides and nothing goes red.

Thayorns (commented):
The list matches Canonical.Pipeline.value() exactly today, including the split around add-default-package.xsl and set-locators.xsl, and TrFull adds no stylesheets of its own, so the fingerprint really does cover the whole XSL surface of the parse stage. What is left uncovered is the raw parse ahead of it: the grammar and the listener that build the XML the pipeline receives change the cached XMIR the same way and are not in the key. My notes below are about what the tests would notice.

eo-maven-plugin/src/test/java/org/eolang/maven/ParsingTest.java
morphqdd: Added a test that checks version() actually contains the XSL fingerprint, not just the digest. Also fixed the root cause: PARSE_XSLS now points at a new Canonical.XSLS built from the same group constants that build the real pipeline, so the array and the pipeline can't drift apart anymore.

eo-maven-plugin/src/test/java/org/eolang/maven/MjParseTest.java
morphqdd: Fixed by removing the duplication instead of trying to test around it. Split Canonical.Pipeline's XSL list into named group constants (HEAD, DEFAULT_PACKAGE, MID, TAIL, IMPORTS) and built both the real pipeline and the new Canonical.XSLS from those same groups. PARSE_XSLS now just points at Canonical.XSLS, so there's only one list, and the drift scenario you describe can't happen anymore.

morphqdd:
This one is now redundant. My PR #6272 already put the same XSL fingerprint into the parse cache key on master, `Parsing.exec` there builds it from `Canonical.XSLS` and `Canonical.IMPORTS` exactly like this branch does. That is also where the conflict comes from. I'd close this.

morphqdd:
Closing as superseded. #6272 already merged the same fix for #6236 (fold a parse-stage XSL fingerprint into the parse cache key).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+819](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6259, opened 2026-08-02
#6241: fix incorrect returned index in tuple.index-of and tuple.last-index-of

### LINKED ISSUE #6241: `tuple.index-of` and `tuple.last-index-of` return incorrect indices for non-symmetric matches
`tuple.index-of` (`eo-runtime/src/main/eo/tuple/index-of.eo:11-24`) references its recursive result `next` inside the branch condition (`next.eq -1`), which forces the tail's recursion to evaluate before the current head is ever inspected. Evaluation therefore runs tail-first: once a match is found deeper in the list, that result is propagated up unchanged and the current, earlier position is never checked. The mirror function `tuple.last-index-of` (`eo-runtime/src/main/eo/tuple/last-index-of.eo:11-19`) has the opposite mistake — it checks `t.head` before recursing, so it stops at the first match rather than the last.

Both functions also return `tup.tail.length` (the count of elements after the match) instead of the true 0-based index, i.e. `list.length - index - 1`. Concretely, on current `master`: `index-of (list 5 1 2 3) 5` returns `3` instead of `0`, and `index-of (list 10 20 30 40 50) 20` returns `3` instead of `1`. `last-index-of (list 10 20 30 40 20) 20` returns `3` instead of `4`. Every existing test in both files passes only because its list happens to be symmetric around the match position (`i == length - i - 1`), which coincidentally makes the wrong tail-length answer equal the correct index.

This exact bug, with the same root-cause analysis and the same `[10,20,30,40,50]`/`20` counter-example, was already reported in #5495 against the predecessor files `ss/index-of.eo` and `ss/last-index-of.eo`, and a correct fix was drafted in #5496. That PR was closed without merging, yet #5495 was closed as completed anyway — the only commit that touched these files afterward (`cb444677`) just renamed variables and relocated the files from `ss/` to `tuple/` per the #5501 package rename; it never changed the algorithm. The bug has been present, unfixed, the whole time.

Fix: in `index-of`, check `t.head.eq wanted` before recursing and return `lst.length.minus t.length` on match (recurse only when it doesn't match), mirroring the approach already drafted in #5496. In `last-index-of`, recurse first and only let the current position's match override the tail's result when the tail's recursion found nothing, so the *last* match wins. Add non-symmetric regression tests, e.g. `index-of (list 10 20 30 40 50) 20 = 1` and `last-index-of (list 10 20 30 40 20) 20 = 4`.

morphqdd:
The issue describes both a value bug and an order bug in `tuple.index-of` and `tuple.last-index-of`. Tracing the actual runtime behavior of `Φ.tuple` (built via `with`, which sets `tail` to the previous tuple and `head` to the newly appended value, see `tuple.eo`), `head`/`tail` walk the tuple from its last element back to its first, not first to last.

Given that, the existing recursion order in both functions was already correct for what each function needs: `index-of` recurses first and lets a deeper (closer to the start) match win, which is what finds the first occurrence when walking backward. `last-index-of` checks the current head immediately, which is what finds the last occurrence when walking backward, since the highest index is visited first.

The actual bug was only the returned value: both functions returned `X.tail.length` (count of elements after the match), not the true 0-based index. Since `tup`/`t` at any point in the recursion is always a prefix of the original list (built by stripping elements off the end), the true index of a match is `tup.length.minus 1` / `t.length.minus 1`.

Verified against the exact counter-example from the issue: `index-of (list 10 20 30 40 50) 20` now returns `1`, and confirmed by hand-tracing the recursion with the actual head/tail semantics before writing the fix. Added a non-symmetric regression test to each file, matching the ones suggested in the issue, and confirmed all existing tests plus the full `eo-runtime` test module still pass.

Closes #6241

### DIFF
--- a/eo-runtime/src/main/eo/tuple/index-of.eo
+++ b/eo-runtime/src/main/eo/tuple/index-of.eo
@@ -18,10 +18,21 @@
           next.eq -1
           if.
             tup.head.eq wanted
-            tup.tail.length
+            tup.length.minus 1
             -1
           recidx tup.tail >> next!
     | list
+
+  eq. ++> can-find-the-index-of-a-non-symmetric-match
+    1
+    index-of
+      *
+        10
+        20
+        30
+        40
+        50
+      20
 
   eq. ++> can-find-the-index-of-a-string
     1
--- a/eo-runtime/src/main/eo/tuple/last-index-of.eo
+++ b/eo-runtime/src/main/eo/tuple/last-index-of.eo
@@ -14,7 +14,7 @@
     -1
     if.
       t.head.eq wanted
-      t.tail.length
+      t.length.minus 1
       rec t.tail
   | list > @
 
@@ -61,3 +61,14 @@
         "Привет"
         "こんにちわ"
       "Привет"
+
+  eq. ++> can-find-the-last-index-of-a-non-symmetric-match
+    4
+    last-index-of
+      *
+        10
+        20
+        30
+        40
+        20
+      20


morphqdd:

eo-runtime/src/main/eo/tuple/index-of.eo
```diff
@@ -18,11 +18,22 @@
           next.eq -1
           if.
             tup.head.eq wanted
-            tup.tail.length
+            tup.length.minus 1
```
Thayorns, line 21: `tuple` is `[tail head length]` and `with` builds the new one as `tuple ^ x (length.plus 1)`, so a tuple's `tail` is literally the tuple one shorter and `tup.tail.length` is `tup.length - 1` for every tuple `*` can build. `tup.length.minus 1` is the same number, which means both the old and the new expression already return the 0-based index, and `at` next door agrees with them by matching `tup.length.plus -1` against the wanted index. The two new cases should be green on master unchanged; if `index-of (list 10 20 30 40 50) 20` really did return something other than 1, the cause is somewhere other than this line, and the same goes for last-index-of.eo:17.

Thayorns (commented):
The reading of head/tail in the description matches what tuple.eo actually does, and the two new cases are worth having either way since neither file had an asymmetric list before. The concern is whether the expression that changed computes anything different from the one it replaced.

eo-runtime/src/main/eo/tuple/index-of.eo
morphqdd: Verified, you're right on both counts. Reconstructed the pre-fix code with the new non-symmetric test cases added and nothing else changed: `index-of(*10,20,30,40,50, 20)` returns 1 and `last-index-of(*10,20,30,40,20, 20)` returns 4 with the old `tup.tail.length`/`t.tail.length` formula too. Checked three ways (manual trace against tuple.with's tail=^ definition, the generated JUnit test, and a direct Phi/Dataized call bypassing test generation) and they agree.

So the formula swap in this PR doesn't change behavior. The bigger problem: issue #6241's own repro claim ("index-of (list 10 20 30 40 50) 20 returns 3 instead of 1 on current master") doesn't hold either, same result confirmed against the actual pre-fix master code. Looks like the issue's diagnosis was wrong, not just this PR's fix. Flagging this to get a decision on whether to close the PR and the issue rather than pushing more changes on a false premise.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3335](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (26) hits-of-code; +1 to give you at least something.  Your running score is [+676](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6261, opened 2026-08-02
#6237: preserve explicit $. reference when printing XMIR

### LINKED ISSUE #6237: Printer incorrectly collapses explicit `$.` reference, causing altered reference meaning upon reparse
## What Happens

The printer collapses an explicit `$.` (XI, self) reference to the same bare form it uses for an ordinary unqualified reference, even though the two denote different objects and the parser itself keeps them distinct.

The parser lowers `$.foo` to `@base='ξ.foo'` and a bare `foo` to `@base='ξ.ρ.foo'` (an explicit self-reference versus an implicit walk up to the enclosing formation's own scope via `ρ`). `Xmir.toEO()` prints both as the same bare `foo`, discarding the difference.

Confirmed with a minimal round-trip:

```
# Sample.

[] > app
  42 > foo
  [] > inner
    $.foo > @
```

prints as `foo > [] > inner`. Re-parsing that output changes the base from `ξ.foo` to `ξ.ρ.foo` - a different reference than the original source specified. The same happens with an application head, `$.f 1 > @` prints as `f 1`, changing `ξ.f` to `ξ.ρ.f` on reparse.

As a control that isolates this specifically to `$.`: dropping an explicit `^.` prefix is correct printer behavior - `^.foo` and a bare `foo` both resolve to the same `ξ.ρ.foo`, so that case round-trips fine. The printer appears to be treating `$.` and `^.` as interchangeable when they are not.

Two things worth noting about real-world impact: `$.` does not currently appear anywhere in the `eo-runtime` stdlib, so no shipped file is affected today; and `$.x` only differs in meaning from bare `x` when the enclosing formation itself has no `x` of its own, meaning in the cases where this matters, dataizing the original source would already fail (dispatch would look for `x` on the wrong object) - so this converts a would-be failure into a silently different, silently succeeding one, rather than corrupting code that currently works. Where a local shadow exists, both readings coincide, so the meaning survives regardless.

The clearer way to frame the impact: `MjFormat`'s `-Deo.autoFix` runs `Xmir.toEO()` and writes the result back over the user's own source, so this is a formatter that can silently rewrite a source file into one that parses to a different meaning than what the user wrote, the same general class of formatter-side-effect problem as issue #6071.

## What Should Happen

The printer should preserve an explicit `$.` prefix when printing a reference whose base is `ξ.<name>` (as opposed to `ξ.ρ.<name>`, which is what dropping the prefix on an ordinary reference already correctly produces), so that printing and re-parsing a source containing `$.` round-trips to the same reference instead of silently changing which object it points to.

morphqdd:
Closes #6237

The printer collapsed an explicit `$.foo` (base `ξ.foo`) to the same bare `foo` it prints for an ordinary same-scope reference, discarding the difference from a bare reference that resolves via an implicit `ρ` hop into an outer scope (base `ξ.ρ.foo`). Since `MjFormat`'s `-Deo.autoFix` runs the printer and writes the result back over the source file, this let the formatter silently rewrite a source into one that parses to a different meaning.

`ξ.<name>` (zero `ρ` hops) shows up in two different situations that look identical at the base-string level:
- an ordinary bare reference to a name that lives on the immediate enclosing formation itself (the common case, needs no `$.` and must keep printing bare), and
- a genuine explicit `$.name` that denotes something different only when the immediate formation does *not* have `name` of its own (the case this issue is about).

Fixed by adding a branch to the printer's `BASED` head template that checks, using the same ancestor-scope pattern the existing `ρ`-hop-drop branch already uses, whether the immediate enclosing formation actually owns the referenced name (as a plain attribute or a file-local `>> name` handle, the latter resolving file-wide rather than through `ρ`-scope walking). Only when neither is true does it keep the `$.` marker; otherwise it prints bare, matching prior (correct) behavior.

Added two print-pack tests (`explicit-xi-reference-kept`, `explicit-xi-application-head-kept`) covering the bare-reference and application-head cases from the issue. Verified against the full `XmirTest` suite (289 tests) with no regressions, and confirmed the new tests fail without the fix (4 failures: the `$.` prefix is dropped) and pass with it.

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -265,6 +265,20 @@
     <xsl:variable name="self-prefix" select="concat($eo:program, '.', $package, '.')"/>
     <xsl:variable name="self-rest" select="substring-after(@base, $self-prefix)"/>
     <xsl:variable name="self-first" select="if (contains($self-rest, '.')) then substring-before($self-rest, '.') else $self-rest"/>
+    <!--
+    Whether some formation this reference is actually nested inside (not
+    just the innermost one - a zero-hop reference here can legitimately
+    reach a handle owned by any enclosing formation, not only the
+    nearest, since resolve-local-names.xsl already resolved it) owns
+    "hop-name" as its own file-local "&gt;&gt; hop-name" handle - mirrors
+    resolve-local-names.xsl's own "eo:captor" scoping (#5875, #5780): a
+    handle is lexically scoped to its declaring formation and the
+    formations nested inside it, so a same-named handle declared by some
+    unrelated formation elsewhere in the file (not an ancestor of this
+    reference at all) must not suppress the "$." marker here.
+    -->
+    <xsl:variable name="local-handle"
+      select="//o[@local=$hop-name][ancestor::o[not(@base)][1] intersect current()/ancestor::o[not(@base)]][1]"/>
     <xsl:choose>
       <!-- NOT OPTIMIZED TUPLE -->
       <xsl:when test="@star">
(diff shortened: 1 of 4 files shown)


morphqdd:

eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
```diff
@@ -309,6 +309,27 @@
         -->
         <xsl:value-of select="eo:translate-path($hop-rest)"/>
       </xsl:when>
+      <xsl:when test="not(@local) and $segments[1] = $eo:xi and $rho-count = 0 and $hop-name != '' and $hop-name != $eo:rho and $hop-name != $eo:phi and $hop-name != $eo:xi and $hop-name != $eo:program and empty(ancestor::o[not(@base)][1]/o[@name=$hop-name]) and empty(//o[@local=$hop-name])">
```
Thayorns: `empty(//o[@local=$hop-name])` looks at the whole document, and the comment justifies it by saying handles resolve file-wide. resolve-local-names.xsl says the opposite in its own header: "two sibling formations may each declare a same-named handle because resolution is lexically scoped (#5875)", and `eo:captor` backs that up by narrowing to `$scope/descendant::o[@local=$name][ancestor::o[not(@base)][1] is $scope]`. So a `>> foo` sitting in some unrelated formation elsewhere in the file makes this conjunct false, the branch is skipped, and `$.foo` prints bare again, which is the rewrite this change is closing.

Thayorns (commented):
Keeping the `$.` marker when the immediate formation does not own the name is the right shape for this, and reusing the ancestor-scope pattern from the branch above keeps the two consistent. The two packs both put the reference and the owning attribute in the same file with nothing else in it, so neither would notice the scoping issue below.

eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
morphqdd: Confirmed and fixed, but not as simply as scoping to the immediate formation - that broke the existing based-nested-only-referenced-handle.yaml test, because a zero-hop $.name can legitimately reach a handle owned by an *ancestor* formation too (resolve-local-names.xsl already resolves it that way before the printer sees it). Scoped the check to "is this handle owned by some formation the reference is actually nested inside" instead of "anywhere in the document". Added a regression test for the reported case (unrelated sibling's handle no longer suppresses $.) and confirmed the whole 291-test print suite stays green, including the ancestor case.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (127 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3347](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+690](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6262, opened 2026-08-02
#6238: fix premature overflow in number.degrees and number.radians

### LINKED ISSUE #6238: `number.degrees` and `number.radians` overflow prematurely for large finite inputs
## What happens

`number.degrees` and `number.radians` return `Infinity` for large finite arguments whose true converted value is still a finite `double`. Each forms its numerator first (`num * 180` for `degrees`, `num * pi` for `radians`), and that product overflows to `Infinity` before the division that would have brought it back into range.

### Runtime evidence

Probed `org.eolang.EO_number.EOdegrees` / `EOradians`. The "true" column is the same conversion computed as `num * (180/pi)` / `num * (pi/180)`, which stays in range:

| call | `number.*` | true (finite) |
| --- | --- | --- |
| `degrees 1e306` | **Infinity** | 5.72958e307 |
| `degrees 1e305` | 5.72958e306 | 5.72958e306 (ok) |
| `radians 1e308` | **Infinity** | 1.74533e306 |
| `radians 5e307` | 8.72665e305 | 8.72665e305 (ok) |

`degrees 1e306` overflows because `1e306 * 180 = 1.8e308` exceeds `Double.MAX_VALUE` (~1.798e308), even though `1e306 * 180 / pi = 5.7e307` is well within range. `radians 1e308` overflows the same way on `1e308 * pi`.

## Root cause

`eo-runtime/src/main/eo/number/degrees.eo`:

```
div. > @
  times.
    number num.as-bytes
    180
  pi
```

and `eo-runtime/src/main/eo/number/radians.eo`:

```
div. > @
  times.
    number num.as-bytes
    pi
  180
```

Both compute `(num * k1) / k2`. The intermediate `num * k1` overflows for `num` near `Double.MAX_VALUE / k1`, which is smaller than the point where the real result overflows.

## Suggested fix

Divide before multiplying, so the intermediate never exceeds the magnitude of the final result:

```
# degrees
times. > @
  div.
    number num.as-bytes
    pi
  180

# radians
times. > @
  div.
    number num.as-bytes
    180
  pi
```

`(num / pi) * 180` and `(num / 180) * pi` push the overflow point out to where the true result overflows. Verified: with this change `degrees 1e306` returns `5.72958e307` and `radians 1e308` returns `1.74533e306`, while every ordinary value is unchanged to machine precision (`degrees pi` = 180, `degrees 1` = 57.2958, `radians 180` = pi, `radians 360` = 2pi, all with relative error 0 against the reference).

A regression test such as `lt. ++> ... (abs ((degrees 1e306).minus 5.72958e307)) 1e302` would pin it.

morphqdd:
`degrees` computed `(num * 180) / pi` and `radians` computed `(num * pi) / 180`. Both multiply first, so the intermediate product overflows to `Infinity` for large but still finite inputs whose true converted value stays well within `double` range, for example `degrees 1e306` (true value about 5.73e307, well under `Double.MAX_VALUE`).

Divide before multiplying instead, so the intermediate never exceeds the magnitude of the final result: `(num / pi) * 180` for degrees, `(num / 180) * pi` for radians.

Added a regression test to each file checking a large input stays finite and close to the reference value, and confirmed the existing tests and the full eo-runtime test module still pass.

Closes #6238

### DIFF
--- a/eo-runtime/src/main/eo/number/degrees.eo
+++ b/eo-runtime/src/main/eo/number/degrees.eo
@@ -9,11 +9,18 @@
 +spdx SPDX-License-Identifier: MIT
 
 [num] > degrees
-  div. > @
-    times.
+  times. > @
+    div.
       number num.as-bytes
-      180
-    pi
+      pi
+    180
+
+  lt. ++> can-convert-a-large-value-without-overflow
+    abs
+      minus.
+        degrees 1.0e306
+        5.729577951308232e307
+    1.0e294
 
   lt. ++> can-convert-minus-pi-to-degrees
     abs
--- a/eo-runtime/src/main/eo/number/radians.eo
+++ b/eo-runtime/src/main/eo/number/radians.eo
@@ -9,11 +9,11 @@
 +spdx SPDX-License-Identifier: MIT
 
 [num] > radians
-  div. > @
-    times.
+  times. > @
+    div.
       number num.as-bytes
-      pi
-    180
+      180
+    pi
 
   lt. ++> can-convert-a-full-turn-to-radians
     abs
@@ -28,6 +28,13 @@
         radians 180
         pi
     1.0E-4
+
+  lt. ++> can-convert-a-large-value-without-overflow
+    abs
+      minus.
+        radians 1.0e308
+        1.7453292519943295e306
+    1.0e293
 
   lt. ++> can-convert-a-negative-half-turn-to-radians
     abs


morphqdd:

Thayorns (approved):
Reordering so the intermediate never exceeds the magnitude of the result is the right fix, and both new cases really do catch the old form, since an Infinity intermediate makes the `abs` of the difference Infinity and the `lt` false. Worth knowing that the reordering is not free at the other end of the range: `radians` now divides by 180 first, so an input below roughly 1e-306 passes through a subnormal intermediate the old `num.times pi` kept out of, and `degrees` does the same below about 1e-310 by dividing by pi first. It costs a couple of bits down there and nothing here pins it either way.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (30) hits-of-code; +4 to give you at least something.  Your running score is [+609](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (30) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3359](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6264, opened 2026-08-02
feat(#4752): implement `string.printf` in pure EO

### LINKED ISSUE #4752: `sprintf` Functionality Not Properly Utilized in EO

asmirnov-backend:
Closes #4752.

`printf` is no longer an atom. The format is walked over its bytes, each specifier (`%[N$][flags][width][.precision]conversion`) is parsed in place, and the conversions reuse `padded-left`, `padded-right`, `as-decimal`, `as-fixed` and `as-hex`. All 42 existing tests stay as they were and pass.

One behaviour is an approximation: the atom used `forma()` to refuse a `bool` for `%s`, which pure EO cannot do. The check is now on bytes — invalid UTF-8, plus the single bytes `00-`/`01-` that `false`/`true` dataize to. The two control characters sharing that encoding are refused with them; this is written down in the docblock.

Large for one PR, but 624 of the 980 lines are the deletion of the now-dead `EOprintf.java`, `PrintfArgs.java` and `PrintfArgsTest.java`, which cannot be dropped before the EO implementation exists.


asmirnov-backend:
This PR is too big — `pr-size` is red and it stays red at ~886 lines. I'll split it into a few smaller ones and open them separately. Please don't spend review time on this one.

eo-runtime/src/main/eo/string/printf.eo
```diff
@@ -1,26 +1,381 @@
...
+        from.plus 1
+      to
+      as-number.
+        plus.
+          acc.times 10
+          minus.
+            code from
+            48
+  T message > [message] >> failure
+  if. > [i] >> flags-end
```
gemshrine, line 97: The old parser accepted the full Java flag run `[-#+ 0,(]` and delegated legal numeric cases to `String.format`. This scanner only advances over `-` and `0`, so previously valid `%+d`, `% d`, `%,d`, and `%(d` (and the corresponding `%f` forms) now stop at the first flag and report it as an unsupported conversion. Please preserve those numeric flags, or add an explicit compatibility decision; the current tests only cover that these flags are invalid for `%s` and `%x`, not that they remain valid for `%d`/`%f`.

gemshrine (changes requested):
The EO implementation and atom removal are well structured, but the flag parser changes the public `string.printf` behavior: legal numeric Java-format flags accepted by the previous implementation are no longer parsed. The inline comment lists the affected forms and the regression coverage needed before this can be approved.

All functional CI checks are green; `pr-size` is the only failing check.

asmirnov-backend:
Superseded, closing. The work went in as three smaller PRs: #6282 (`string.truncated`) and #6283 (`string.padded-zeros`), both merged, and #6336, which composes them into the pure-EO `printf` and drops the Java atom.

0crat:
@gemshrine Thanks for the review! You've earned +17 points for this: +12 as a basis; +9 for the 987 hits-of-code that you reviewed; -5 for very few (1) comments.  Your running score is [+1644](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6265, opened 2026-08-02
#6232: reject empty name suffix when / or ! sits right after the space

### LINKED ISSUE #6232: `Suffix.named` accepts an empty name when `/sig` or `!` follows the single skipped space
## What Happens

A `> name` suffix immediately followed by `/sig` or `!` with exactly one space in between parses as an empty name, with no error. This is a different trigger than the one closed issue #6150 fixed.

#6150 added `Suffix.blank(tail, from)`, which rejects a tail that is *entirely* whitespace (catching the double-space-before-end-of-line case). That guard does not fire here, because the tail is not entirely whitespace - it contains a real `/` or `!` right after the one space that gets skipped.

`Suffix.named`'s flow after the `blank` check:

```java
final int begin = Suffix.skipSpace(tail, from);
int idx = Suffix.skipName(tail, begin);
```

`skipSpace` skips exactly one space. `skipName` stops as soon as it hits a character in `Suffix.terminates`, which includes `/` and `!`. So for a tail like `" /org.eolang.foo"`, `begin` lands directly on `/`, `skipName` doesn't advance past it, and `name` comes out as the empty string with no error raised.

Confirmed with a real parse:

```
[] > app
  42 > /org.eolang.foo
```

produces `<o base="Φ.number" line="2" loc="Φ.app." name="" pos="2">` with zero errors reported - note the empty `name` attribute and the trailing dot in `loc`. The same happens for a top-level formation, `[] > /org.eolang.foo`, giving `loc="Φ."` with `name=""`.

## What Should Happen

`Suffix.named` should reject an empty `name` after `skipName` runs, the same way `Suffix.test` already does for the test-attribute suffix form (`Suffix.java:347-352`), regardless of whether the emptiness came from an all-blank tail (already covered by #6150) or from a name-terminating character sitting immediately after the single skipped space.

morphqdd:
`Suffix.named` skips exactly one space after `>` and then reads the name up to the next terminator. If a terminator (`/` or `!`) sits right at that position, `skipName` doesn't advance and the name comes out empty with no error, for example `42 > /org.eolang.foo` or `42 > !`. #6150 only rejected a tail that is entirely blank; it did not cover a terminator landing immediately after the single skipped space.

Added a check right after computing the name bounds, mirroring the same check `Suffix.test` already does. It only fires when the character at that position is not itself a space or tab, so the pre-existing double-space typo case (`[] >  foo`) still falls through to the existing "trailing garbage" error instead of getting the new message.

Extracted the check into its own method (`checkNamePresent`) to keep `named`'s cyclomatic complexity under qulice's limit.

Added tests for both triggers mentioned in the issue (`> /sig` and `> !`) and confirmed the full eo-parser test suite and qulice still pass.

Closes #6232

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -398,13 +398,15 @@ final class Suffix {
     }
 
     /**
-     * Reject a {@code > name} suffix whose name came out empty because a
-     * name-terminating character ({@code /} or {@code !}) sat immediately
-     * after the single skipped space, e.g. {@code > /sig} or {@code > !}.
-     * A second space (rather than a real terminator) is left alone here:
-     * that case falls through to {@link #endsClean}, which reports the
-     * more specific "trailing garbage" once whatever follows the extra
-     * space is reached.
+     * Reject a {@code > name} suffix whose name came out empty because
+     * {@link #skipName} stopped right where it started — any character
+     * {@link #endsName} treats as a boundary sat immediately after the
+     * single skipped space, not just {@code /} and {@code !} but also
+     * the separators in {@link #NAME_BOUNDARIES}, e.g. {@code > /sig},
+     * {@code > !}, or {@code > .foo}. A second space (rather than a real
+     * boundary character) is left alone here: that case falls through to
+     * {@link #endsClean}, which reports the more specific "trailing
+     * garbage" once whatever follows the extra space is reached.
      * @param tail Tail substring
      * @param begin Index where the name was expected to start
      * @param idx Index {@link #skipName} stopped at
@@ -415,8 +417,7 @@ final class Suffix {
     private static void checkNamePresent(
         final String tail, final int begin, final int idx, final Span span, final int home
     ) {
-        if (begin == idx && begin < tail.length()
-            && tail.charAt(begin) != ' ' && tail.charAt(begin) != '\t') {
+        if (begin == idx && tail.charAt(begin) != ' ' && tail.charAt(begin) != '\t') {
             throw new ParseError(
                 span.line(), home + begin,
                 "name suffix requires a name"
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -398,6 +398,33 @@ final class Suffix {
     }
 
     /**
+     * Reject a {@code > name} suffix whose name came out empty because a
+     * name-terminating character ({@code /} or {@code !}) sat immediately
+     * after the single skipped space, e.g. {@code > /sig} or {@code > !}.
+     * A second space (rather than a real terminator) is left alone here:
+     * that case falls through to {@link #endsClean}, which reports the
+     * more specific "trailing garbage" once whatever follows the extra
+     * space is reached.
+     * @param tail Tail substring
+     * @param begin Index where the name was expected to start
+     * @param idx Index {@link #skipName} stopped at
+     * @param span Source span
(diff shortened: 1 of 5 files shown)


morphqdd:

eo-parser/src/main/java/org/eolang/parser/Suffix.java
```diff
@@ -397,6 +397,33 @@ private static void checkLowercaseStart(
         }
     }
 
+    /**
+     * Reject a {@code > name} suffix whose name came out empty because a
+     * name-terminating character ({@code /} or {@code !}) sat immediately
```
Thayorns: The check is wider than `/` and `!`. skipName stops on endsName, which is `terminates` plus the twelve characters in NAME_BOUNDARIES, so `[] > .foo` and `[] > ,x` now reach this throw too and report a missing name where they used to report trailing garbage from endsClean. That may well be the better message, but the docblock describes only the second-space case as falling through, so the next reader will not expect it.

Thayorns (commented):
The two inputs from the issue are the only ones that got through before: every other empty-name spelling already died in endsClean, while `!` was consumed by the const branch and `/` by the signature branch and both left an empty name behind. Placing the check right after skipName catches both, and reusing the wording of the blank-tail error keeps the two paths saying the same thing. Two notes on the guard itself.

eo-parser/src/main/java/org/eolang/parser/Suffix.java
```diff
@@ -397,6 +397,33 @@ private static void checkLowercaseStart(
...
+     * @param begin Index where the name was expected to start
+     * @param idx Index {@link #skipName} stopped at
+     * @param span Source span
+     * @param home Source column where tail begins
+     * @checkstyle ParameterNumberCheck (3 lines)
+     */
+    private static void checkNamePresent(
+        final String tail, final int begin, final int idx, final Span span, final int home
+    ) {
+        if (begin == idx && begin < tail.length()
```
Thayorns: `begin < tail.length()` cannot be false here. `named` calls `blank(tail, from)` five lines earlier and throws when everything from `from` on is blanks, and `skipSpace` advances by at most one space, so `begin` is always an index into the string by the time this runs.

eo-parser/src/main/java/org/eolang/parser/Suffix.java
morphqdd: Widened the docblock to describe the real trigger set (terminates plus NAME_BOUNDARIES), and added a YAML test for the boundary-character case (> .foo) to cover it.

eo-parser/src/main/java/org/eolang/parser/Suffix.java
morphqdd: Confirmed and removed. blank() already throws if everything from `from` is blank, and skipSpace advances by at most one space, so begin can't reach tail.length() by the time this runs. Ran the full test suite including the naughty-strings fuzz pack with the guard gone, no StringIndexOutOfBoundsException.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3375](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+697](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6266, opened 2026-08-02
#6230: reject atom signature on non-formation lines

### LINKED ISSUE #6230: Atom signature (`/sig`) is silently discarded on non-formation lines instead of erroring
## What Happens

An atom signature (`/sig`) written on a `> name` suffix is silently discarded instead of reported as an error, for every line kind except a formation.

`Suffix.named` parses `/sig` for any `> name` suffix regardless of what kind of line it's attached to. Only `LnFormation` ever reads `suffix.atom()`/`suffix.sig()` back out to emit the `<o name='λ' atom='…'/>` marker that makes a formation an atom. `LnApplication`, `LnMethod`, `LnReversed`, and `LnCompactTuple` (plain application, method-chain, reversed-dispatch, and compact-tuple lines) never consult the parsed signature at all; `LnTextBlock` behaves the same way for a text block's own name suffix.

Confirmed: all four of the following parse with zero errors, and none of them produces an `atom=` marker anywhere in the resulting XMIR, even though each line explicitly wrote a `/sig`:

```
[] > app
  42 > x/org.eolang.foo        # data literal (LnApplication path)
  seq > y/org.eolang.foo       # application
  x.y > z/org.eolang.foo       # method chain
  """ … """ > t/org.eolang.foo # text block
```

By contrast, `[] > f/org.eolang.foo` (an actual formation) correctly produces the atom marker.

This is not simply "a signature only makes sense on formations, so it's harmlessly ignored elsewhere" - the parser already treats the identical situation as an error for a pipe application: `LnPipe` explicitly rejects `/sig` with "a pipe application cannot declare an atom or a test attribute", citing R-3.14.6 ("a pipe is an application, not a formation"). A plain application, method chain, or text block is no more a formation than a pipe is, so writing `/sig` on any of them is the same user mistake `LnPipe` already catches, just silently swallowed instead of reported everywhere else it can occur.

## What Should Happen

Any non-formation line (application, method, reversed, compact-tuple, text-block) that parses a non-empty `/sig` signature should raise the same error `LnPipe` already raises for the identical situation, instead of silently discarding the user's declaration.

morphqdd:
`Suffix.named` parses an atom signature (`/sig`) regardless of which kind of line it is attached to. Only `LnFormation` reads it back to emit the atom marker. `LnApplication`, `LnMethod`, `LnReversed`, `LnCompactTuple`, and `LnTextBlock` never consult it at all, so writing `/sig` on any of them (a data literal, an application, a method chain, a compact tuple, or a text block name) parses with zero errors and the signature is silently dropped.

`LnPipe` already treats the identical situation as an error for a pipe application, since a pipe is no more a formation than these other line shapes are. Added a small shared check (`Atoms.rejectOutsideFormation`), called from each of the five line kinds right after the suffix is parsed, that raises the same kind of error `LnPipe` already raises. Kept `LnPipe`'s own check as is, since it also rejects a test attribute on a pipe, which is a separate rule not shared by the other five (test attributes are legitimate on all of them).

Added one YAML syntax test per line kind, reproducing each example from the issue, and confirmed each one fails on the unfixed code and passes with the fix. Ran the full eo-parser test suite and qulice, both clean.

Closes #6230


morphqdd:

eo-parser/src/main/java/org/eolang/parser/Atoms.java
```diff
@@ -0,0 +1,45 @@
...
+        // never called
+    }
+
+    /**
+     * Reject a non-empty atom signature parsed on a line that is not a
+     * formation.
+     * @param suffix The parsed suffix
+     * @param span The line's span (used for error position)
+     */
+    static void rejectOutsideFormation(final Suffix suffix, final Span span) {
```
Thayorns: The rule that a signature belongs to a formation is now written in two places with two different wordings: this one, and LnPipe.java:71, which says "a pipe application cannot declare an atom or a test attribute" for the same mistake. A user who moves `/org.eolang.foo` from an application to a pipe gets a different message for the same error, and the sixth line shape someone adds next year reopens the hole unless they remember to call this. Suffix is the object that parsed the signature and knows nobody asked for it.

Thayorns (commented):
The five line shapes really do drop the signature on the floor today, LnFormation is the only reader, and putting the check right after the suffix is parsed keeps all five call sites identical. The packs each assert on an error containing the word `atom`, which matches any error mentioning atoms rather than this one, though nothing else produced an error for these inputs before so they do fail on unfixed code.

eo-parser/src/main/java/org/eolang/parser/Atoms.java
morphqdd: Moved the check onto Suffix as an instance method (rejectAtomOutsideFormation), called from all six line shapes including LnPipe now. Deleted Atoms.java. LnPipe keeps its own separate test-attribute check with an updated message that no longer mentions atoms. Also added two YAML packs for LnPipe (atom and test-attribute), since neither had any test coverage before.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (112 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3387](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+704](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6267, opened 2026-08-02
#6229: add pos to error complexType so parser syntax errors validate against XMIR.xsd

### LINKED ISSUE #6229: Every parser syntax error produces XMIR invalid against `XMIR.xsd` (`pos` on `<error>`)
## What Happens

Every `.eo` file with a plain parser syntax error produces XMIR that fails validation against the repository's own `XMIR.xsd`.

`Emit.java`'s error-writing method sets a `pos` attribute on the `<error>` element:

```java
.attr("pos", pos)
```

`XMIR.xsd` defines the `error` complex type with exactly three attributes: `check`, `line`, and `severity`. There is no `pos` in the schema.

Confirmed with a minimal syntax error:

```
# Sample.

[] > app
  42 >
```

Validating the resulting XMIR against `eo-parser/src/main/resources/XMIR.xsd` reports:

```
cvc-complex-type.3.2.2: Attribute 'pos' is not allowed to appear in element 'error'.
```

Errors added later by the parse-stage XSLs (for example `validate-objects-count`) only ever carry `check`, `line`, and `severity`, so they validate fine. The offending `pos` attribute is specific to the hand-written parser's own `<error>` elements, meaning any `.eo` file with a plain syntax error produces schema-invalid XMIR.

`EoSyntaxTest.checksXsdMistakes` exists specifically as a harness for XSD-mismatch cases like this, but none of its packs currently contain a parser syntax error, which is why this slipped past it.

## What Should Happen

Either `pos` should be added to the `error` complex type in `XMIR.xsd` (the parser spec's R-9.1.2 / §9.6 calls for reporting the column of an error, which is what `pos` conveys, so this is the more likely intended fix), or the parser should stop emitting `pos` on `<error>` elements to match the existing schema. Either way, XMIR produced from a source file with a syntax error should validate against the repository's own schema.

morphqdd:
`Emit.java` sets a `pos` attribute on `<error>` elements, but `XMIR.xsd`'s `error` complex type only declares `check`, `line`, and `severity`. Any `.eo` file with a plain parser syntax error produces XMIR that fails validation against the repository's own schema.

The parser spec's R-9.1.2 / section 9.6 calls for reporting the error's column, which is what `pos` conveys, so this adds `pos` to the schema (type `whole`, matching how `pos` is already declared on `o`) rather than dropping it from the parser.

Added a case to `EoSyntaxTest.checksXsdMistakes` with a raw `<error pos="...">` element expecting zero schema violations, confirmed it fails against the unfixed schema with the same `cvc-complex-type.3.2.2` message from the issue, and passes with the fix. Full eo-parser test suite is green.

Closes #6229

### DIFF
--- a/eo-parser/src/main/resources/XMIR.xsd
+++ b/eo-parser/src/main/resources/XMIR.xsd
@@ -144,7 +144,7 @@
       <xs:annotation>
         <xs:appinfo>The position in the code line where the object has been seen</xs:appinfo>
         <xs:documentation>
-          The attribute contains the position in the code line (starting from 1),
+          The attribute contains the position in the code line (0-indexed),
           where the object has been met during the scanning of the source code.
         </xs:documentation>
       </xs:annotation>
@@ -262,7 +262,7 @@
           <xs:annotation>
             <xs:appinfo>The position in the code line where the error was found</xs:appinfo>
             <xs:documentation>
-              The attribute contains the position in the code line (starting from 1),
+              The attribute contains the position in the code line (0-indexed),
               where the error was found. If the element is omitted, the problem
               is not related to any particular column in the line.
             </xs:documentation>
--- a/eo-parser/src/main/resources/XMIR.xsd
+++ b/eo-parser/src/main/resources/XMIR.xsd
@@ -258,6 +258,16 @@
             </xs:documentation>
           </xs:annotation>
         </xs:attribute>
+        <xs:attribute name="pos" type="whole">
+          <xs:annotation>
+            <xs:appinfo>The position in the code line where the error was found</xs:appinfo>
+            <xs:documentation>
+              The attribute contains the position in the code line (starting from 1),
+              where the error was found. If the element is omitted, the problem
+              is not related to any particular column in the line.
+            </xs:documentation>
+          </xs:annotation>
+        </xs:attribute>
         <xs:attribute name="severity" use="required">
           <xs:annotation>
             <xs:appinfo>The severity of the error</xs:appinfo>
--- a/eo-parser/src/test/resources/org/eolang/parser/xsd-mistakes/error-pos.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/xsd-mistakes/error-pos.yaml
@@ -1,0 +1,11 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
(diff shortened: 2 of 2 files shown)


morphqdd:

eo-parser/src/main/resources/XMIR.xsd
```diff
@@ -258,6 +258,16 @@
             </xs:documentation>
           </xs:annotation>
         </xs:attribute>
+        <xs:attribute name="pos" type="whole">
+          <xs:annotation>
+            <xs:appinfo>The position in the code line where the error was found</xs:appinfo>
+            <xs:documentation>
+              The attribute contains the position in the code line (starting from 1),
```
Thayorns: Emit.java:200 documents this value as "Column where the error occurred (0-indexed)", so "starting from 1" here is off by one against the only thing that writes the attribute. The `o/@pos` text you patterned it on says the same thing, which may mean that one is wrong too, but this is the copy being added for an attribute whose emitter says otherwise a few lines from where it is set.

Thayorns (commented):
The attribute is genuinely emitted and genuinely missing from the type, and leaving it optional with `whole` is right since `whole` admits zero and the column can be zero. The sentence says "if the element is omitted" about an attribute, which will read oddly next to the surrounding text.

eo-parser/src/main/resources/XMIR.xsd
morphqdd: Fixed, and checked the sibling o/@pos doc too since it's set from the same Span.indent() value as the error's pos - it had the same "starting from 1" mistake. Both now say 0-indexed, matching Emit.error()'s own Javadoc.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (23) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3399](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (23) hits-of-code; +1 to give you at least something.  Your running score is [+708](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6268, opened 2026-08-02
#6227: close text block after an error on its closing line

### LINKED ISSUE #6227: Parse error on a text block's closing line silently swallows the rest of the file
## What Happens

A parse error on a text block's closing line (the `""" > name` line) leaves the parser's text-block flag set forever, so every remaining source line gets silently absorbed as text-block content instead of being parsed, and a phantom "unclosed text block" error appears at EOF even though the block was in fact closed.

`LnTextBlock.into()` can throw `ParseError` from several places before it reaches `globals.closeTextBlock()`: `tokens.readChain()`, `new Suffix(...)`, the escape-decode wrapper, and `this.transition(...)`. `Eo.continueTextBlock` catches that `ParseError`, rolls the emit back and reports it, but never calls `globals.closeTextBlock()`. From that point on, `Eo.dispatch` keeps routing every line to `globals.appendTextLine(...)` instead of parsing it, and `Eo.finish` adds an extra, incorrect "unclosed text block" error at the end.

Confirmed with a bad escape inside a text block (the same class of input that #6026 fixed the crash for):

```
# Sample.

[] > app
  """
  \uZZZZ
  """ > x
  42 > y
  43 > z
```

Result: `<o line="3" loc="Φ.app" name="app" pos="0"/>` — the object is empty. `42 > y` and `43 > z` never make it into the XMIR. Two errors are reported: the real escape error, and a spurious `[4:0] error: 'unclosed text block opened at line 4'`, even though the block's closing `"""` line was reached and processed (just unsuccessfully, due to the escape error inside it).

Same effect with any error on the closer line itself, for example a typo like `""" > x.y` (trailing garbage after the name suffix): the same phantom "unclosed" error appears and the rest of the file is discarded.

## What Should Happen

Per spec R-7.2, recovery from a line-level parse error should be line-local: the rest of the file should still be parsed normally. `Eo.continueTextBlock` should close the text block (clear the flag) regardless of whether the closer line's own suffix parsed successfully, the same way any other line's error doesn't prevent subsequent lines from being parsed. No phantom "unclosed text block" error should be reported when the block's closing line was in fact reached.

morphqdd:
`Eo.continueTextBlock` catches a `ParseError` thrown while processing a text block's closing line (from the suffix, the escape decoder, or the transition), rolls the emit back, and reports the error, but never clears the `inTextBlock` flag. Every following line then gets silently absorbed as text-block body content instead of being parsed, and a phantom "unclosed text block" error appears at EOF even though the closer line was actually reached.

Added `globals.closeTextBlock()` to the catch block, right after the error is reported. This only clears state going forward; it does not affect the already-reported error or anything already rolled back.

Added a syntax test reproducing the issue's example (a bad unicode escape inside a text block, closing line, followed by two more object lines), asserting the real error is still present, no "unclosed" error appears, and the following lines still parse into real objects. Confirmed it fails on the unfixed code with "passed without exceptions" (the follow-on lines silently vanish) and passes with the fix. Full eo-parser test suite and qulice are clean.

Closes #6227

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -345,6 +345,7 @@ final class Eo implements Iterable<Directive> {
             } catch (final ParseError err) {
                 emit.rollback(tstartsen);
                 emit.error(err.line(), err.pos(), err.getMessage());
+                globals.closeTextBlock();
             }
         } else {
             final String raw = span.text();
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/error-on-text-block-closer-does-not-swallow-rest-of-file.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/error-on-text-block-closer-does-not-swallow-rest-of-file.yaml
@@ -1,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets: []
+asserts:
+  - /object/errors/error[contains(text(),'unicode')]
+  - /object[not(errors/error[contains(text(),'unclosed')])]
+  - //o[@name='y']
+  - //o[@name='z']
+input: |
+  [] > app
+    """
+    \uZZZZ
+    """ > x
+    42 > y
+    43 > z


morphqdd:

Thayorns (approved):
LnTextBlock.into only reaches closeTextBlock on its last three lines, well past both throw sites, so the flag really does survive a failed closer and eats the rest of the file. Adding the call in the catch is the whole fix, and the two companions on the success path, clearBlanks and markEmitted, were already run when the block opened, so leaving them out of the catch costs nothing. The pack asserts a unicode error is present and an unclosed one is not, but nothing pins how many errors there are, so a second spurious one on the recovery path would still read as green.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (18) hits-of-code; +4 to give you at least something.  Your running score is [+613](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3411](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6269, opened 2026-08-02
Fix #6213: strip org.eolang. prefix in ObjectsIndex.children()

### LINKED ISSUE #6213: `ObjectsIndex.children` does not strip the `org.eolang.` prefix that `contains` strips
## What Happens

`ObjectsIndex.contains(name)` normalizes away a leading `org.eolang.` from `name` before looking it up in the index. `ObjectsIndex.children(pkg)` builds its lookup prefix directly from the raw `pkg` argument, with no equivalent normalization.

`Probing.complete` derives its package argument from an object name that has already passed through `contains()`'s normalization. So for any name in the `org.eolang.*` shape, `contains()` succeeds (because it stripped the prefix before checking), but `children()` on the same un-stripped name matches nothing, and package completion silently does nothing.

This is not reachable today: every object in `eo-runtime` is rooted at `Φ.<name>`, not `Φ.org.eolang.<name>`, so the mismatched shape never actually occurs in the current stdlib. But the two methods on the same class disagree about what normalization is required, which is a latent inconsistency that could break for any future object using the longer form, or for a user package that happens to declare a `+package org.eolang.something`.

## What Should Happen

`ObjectsIndex.children()` should apply the same `org.eolang.` prefix normalization that `contains()` already does, so the two methods agree on how a package name is matched.

AhmedAliElashmawy:
Closes #6213
- **Prefix Normalization in `ObjectsIndex.children()`**: Updated `ObjectsIndex.children(pkg)` to strip the leading `org.eolang.` prefix if present, aligning its package matching behavior with `ObjectsIndex.contains(name)`.
- **Unit Test**: Added `listsDirectChildrenOfPackageWithOrgEolangPrefix()` in `ObjectsIndexTest` to verify that `children("org.eolang.tuple")` successfully strips the prefix and      
  matches direct package objects.

@yegor256 please take a look

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
@@ -89,7 +89,14 @@ final class ObjectsIndex {
      * @throws Exception If the index can't be read
      */
     Iterable<String> children(final String pkg) throws Exception {
-        final String prefix = String.format("%s.", pkg);
+        final String pfx = "org.eolang.";
+        final String target;
+        if (pkg.startsWith(pfx)) {
+            target = pkg.substring(pfx.length());
+        } else {
+            target = pkg;
+        }
+        final String prefix = String.format("%s.", target);
         return new Filtered<>(
             name -> name.startsWith(prefix) && name.indexOf('.', prefix.length()) < 0,
             this.objects.value()
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/ObjectsIndexTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/ObjectsIndexTest.java
@@ -93,6 +93,25 @@ final class ObjectsIndexTest {
     }
 
     @Test
+    void listsDirectChildrenOfPackageWithOrgEolangPrefix() throws Exception {
+        MatcherAssert.assertThat(
+            "The index must strip org.eolang. prefix when listing children of package",
+            new ObjectsIndex(
+                new ScalarOf<>(
+                    () -> new SetOf<>(
+                        "tuple",
+                        "tuple.each",
+                        "tuple.eachi",
+                        "tuple.inner.deep",
+                        "math.abs"
+                    )
+                )
+            ).children("org.eolang.tuple"),
+            Matchers.containsInAnyOrder("tuple.each", "tuple.eachi")
+        );
+    }
+
+    @Test
     @ExtendWith(WeAreOnline.class)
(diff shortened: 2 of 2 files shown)


eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
```diff
@@ -89,7 +89,14 @@ boolean contains(final String name) throws Exception {
      * @throws Exception If the index can't be read
      */
     Iterable<String> children(final String pkg) throws Exception {
-        final String prefix = String.format("%s.", pkg);
+        final String pfx = "org.eolang.";
+        final String target;
+        if (pkg.startsWith(pfx)) {
```
Thayorns, line 94: `"org.eolang"` does not start with `"org.eolang."`, so it is not stripped and the filter looks for names beginning with `org.eolang.` in an index that stores them bare. Probing.java:148 builds `pkg` as everything before the last dot, so every top-level object — `org.eolang.string`, `org.eolang.tuple` — arrives here as exactly `"org.eolang"` and gets an empty sibling list back with no error, which is the same silent miss as before for the commonest objects in the language.

eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
```diff
@@ -89,7 +89,14 @@ boolean contains(final String name) throws Exception {
      * @throws Exception If the index can't be read
      */
     Iterable<String> children(final String pkg) throws Exception {
-        final String prefix = String.format("%s.", pkg);
+        final String pfx = "org.eolang.";
```
Thayorns, line 92: These six lines are `contains` on line 72 again, with `prefix`/`stripped` renamed to `pfx`/`target` because `prefix` was already in use here. Two copies of the same rule, spelled differently enough that a reader will not notice they are the same, and the literal now sits in two places.

Thayorns (commented):
Aligning children() with contains() is the right direction and the new test pins the sub-package case. The bigger of the two notes below is the package name the actual caller passes most of the time.

yegor256:
@AhmedAliElashmawy it's fixed already, I believe

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (2) comments; -4 for too few (28) hits-of-code; +1 to give you at least something.  Your running score is [+712](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6270, opened 2026-08-02
#6231: keep continueTextBlock's stack in sync with its rollback

### LINKED ISSUE #6231: `Eo.continueTextBlock` misses the stack rollback `Eo.dispatch` already has, misplacing `@ms`
## What Happens

A parse error on a text block's closing line can leave the parser's internal stack out of sync with the directives it already emitted, producing an `<o>` element with the `ms` attribute in the wrong place, which is invalid per the repository's own `XMIR.xsd` (`ms` is only allowed on `object`, not on `o`).

`Eo.dispatch` records the emit-stack depth before running a line and, on `ParseError`, calls `stack.silentTruncate(frame)` in addition to `emit.rollback(...)`, keeping the two in sync:

```java
final int frame = stack.size();
try { ... } catch (final ParseError err) {
    emit.rollback(tstartsen);
    stack.silentTruncate(frame);
    emit.error(...);
}
```

`Eo.continueTextBlock` has no equivalent `frame`/`silentTruncate` pairing. It calls `stack.popDeeperThan(span.indent())` inside its own `try` block; if the closer line then throws a `ParseError`, `emit.rollback(...)` undoes the `.up()` (close) directives that the pop had already emitted, but the corresponding `Level` objects are already gone from `stack`, since nothing rolls the stack back to match.

Confirmed with an input that needs at least one nested level to be popped when the text block's indent returns:

```
# Sample.

[] > app
  seq > s
    42
  """
  body
  """ oops
```

The closer line (`""" oops`) fails to parse (trailing garbage after the name position), but by then `popDeeperThan` has already popped the `s`/`42` levels and emitted their closing directives, which the subsequent rollback then undoes without restoring the stack. The final XMIR ends up with `.attr("ms", 0L)` applied one level too deep:

```xml
<object ...>                                          <!-- no @ms here -->
   <o line="3" loc="Φ.app" ms="0" name="app" pos="0">  <!-- @ms landed here instead -->
```

Validating against `XMIR.xsd` reports `cvc-complex-type.3.2.2: Attribute 'ms' is not allowed to appear in element 'o'` (`ms` is declared only on the `object` element). With a source needing two levels of popping, `ms` lands two levels deep instead of one; with no nested level to pop (a plain `42 > x` instead of the `seq`/`42` pair), `ms` correctly stays on `<object>` - confirming the pop/rollback mismatch is what causes the misplacement.

This is a separate defect from the "rest of file silently swallowed" bug also present in this method (a different, already-reported issue): even once that one is fixed, this stack/rollback desync remains, and needs its own one-line fix.

## What Should Happen

`Eo.continueTextBlock` should record `stack.size()` before its `try` block and call `stack.silentTruncate(frame)` in its `catch`, exactly the way `Eo.dispatch` already does, so a parse error on the text block's closer line can't leave the stack out of sync with the directives actually emitted.

morphqdd:
`Eo.dispatch` takes its emit savepoint and stack-size frame after `popDeeperThan`, and only wraps the line's own parsing in the try, so a failure rolls back just what that parsing did and truncates the stack to match. `Eo.continueTextBlock` had the pop and the savepoint in the opposite order and no stack truncation at all, so a parse error on a text block's closing line could leave the emitted XML and the in-memory stack disagreeing about how deep the cursor actually is, misplacing the final `ms` attribute onto a nested `<o>` instead of the root `<object>` (invalid against XMIR.xsd, since `ms` is only declared there).

Reordered `continueTextBlock` to match `dispatch` exactly: pop first, then take the savepoint and record the frame, then wrap only `LnTextBlock.into` in the try, truncating the stack to the recorded frame in the catch.

Note this PR does not touch the separate "unclosed text block" phantom-error bug also present in this method (already reported and fixed independently in #6227); the syntax test added here still shows that error, since #6227's fix isn't part of this branch.

Added a syntax test reproducing the issue's exact example (a `seq`/nested-number pair followed by a text block whose closer line has trailing garbage), asserting `ms` lands on `/object` and no nested `<o>` carries it. Confirmed it fails on the unfixed code with `ms` on the `app` object, exactly as described in the issue, and passes with the fix. Full eo-parser test suite and qulice are clean.

Closes #6231

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -338,12 +338,14 @@ final class Eo implements Iterable<Directive> {
         final Span span, final Stack stack, final Globals globals, final Emit emit
     ) {
         if (Eo.closesTextBlock(span, globals)) {
+            stack.popDeeperThan(span.indent());
             final int tstartsen = emit.savepoint();
+            final int frame = stack.size();
             try {
-                stack.popDeeperThan(span.indent());
                 new LnTextBlock(span).into(stack, globals, emit);
             } catch (final ParseError err) {
                 emit.rollback(tstartsen);
+                stack.silentTruncate(frame);
                 emit.error(err.line(), err.pos(), err.getMessage());
             }
         } else {
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/error-on-text-block-closer-does-not-misplace-ms.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/error-on-text-block-closer-does-not-misplace-ms.yaml
@@ -1,0 +1,16 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets: []
+asserts:
+  - /object[@ms]
+  - /object[not(.//o[@ms])]
+  - /object/errors/error[contains(text(),'garbage')]
+input: |
+  [] > app
+    seq > s
+      42
+    """
+    body
+    """ oops


morphqdd:

Thayorns (approved):
This lands on the same four lines dispatch already uses, in the same order, so the closing directives popDeeperThan emits for lines that already succeeded stay outside the savepoint and only the closer line's own work is undone. The pack pins `ms` on /object and asserts no nested `<o>` carries it, which is exactly the symptom. One thing the shape inherits from dispatch rather than introduces: silentTruncate can only shrink, so a line that pops below the recorded frame before throwing still leaves the stack shallower than the rollback assumes.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (20) hits-of-code; +4 to give you at least something.  Your running score is [+617](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3423](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6271, opened 2026-08-02
#6235: fold skipSourceLints/skipExperimentalLints into per-file lint cache key

### LINKED ISSUE #6235: `skipSourceLints` and `skipExperimentalLints` flags not included in per-file lint cache key causing stale caching results

morphqdd:
`Linting`'s per-file lint cache key was built from the plugin version and a hash of the tojo, but the cached content also depends on `skipSourceLints` and `skipExperimentalLints`, since both change what `linted(xmir)` reports for the same source. Neither flag was part of the key, so toggling either between two builds sharing a cache directory could reuse a stale result computed under the old flag value.

Added `cacheVersion()`, combining the plugin version with both flags (`skipSourceLints` sorted before joining, so passing the same set in a different order still shares a cache entry), and used it as the `semver` segment of the per-file cache path in `lintOne`. This is the same pattern already used for the transpile cache key's `trackLocations`/`coverageTracking` flags in `Transpilation.version()`.

Added a test reproducing the issue's scenario: two `FakeMaven` runs sharing a cache directory and an identical tojo hash, the first without `skipSourceLints`, the second with `mandatory-spdx` added to it. Confirmed it fails on the unfixed code (the stale cached result still reports the defect) and passes with the fix. Full eo-maven-plugin test suite and qulice are clean.

Closes #6235

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
@@ -321,7 +321,7 @@ final class Linting implements Step {
                 new Cache(
                     new CachePath(
                         this.cacheDir.resolve(Linting.CACHE),
-                        this.version,
+                        this.cacheVersion(),
                         new TojoHash(tojo).get()
                     ),
                     src -> this.linted(xmir).toString()
@@ -349,6 +349,25 @@ final class Linting implements Step {
         }
         tojo.withLinted(target);
         return 1;
+    }
+
+    /**
+     * Cache-key version segment for the per-file lint cache: the plugin
+     * version combined with {@link #skipSourceLints} and
+     * {@link #skipExperimentalLints}, since both change what
+     * {@link #linted(XML)} reports for the exact same source XMIR
+     * (see #6235). {@link #skipSourceLints} is sorted before joining so
+     * that two builds passing the same set in a different order still
+     * share a cache entry.
+     * @return The version segment for {@link CachePath}
+     */
+    private String cacheVersion() {
+        return String.format(
+            "%s-%b-%s",
+            this.version,
+            this.skipExperimentalLints,
+            this.skipSourceLints.stream().sorted().collect(Collectors.joining(","))
+        );
     }
 
     /**
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
@@ -53,6 +53,7 @@ import org.xembly.Xembler;
  *     Naming the class {@code Linting} avoids this collision.
  * </p>
  * @since 0.31.0
+ * @checkstyle ClassFanOutComplexityCheck (3 lines)
  */
 @SuppressWarnings("PMD.GodClass")
 final class Linting implements Step {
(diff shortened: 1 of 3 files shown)


morphqdd:

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -351,6 +351,25 @@ private int lintOne(
...
+     * that two builds passing the same set in a different order still
+     * share a cache entry.
+     * @return The version segment for {@link CachePath}
+     */
+    private String cacheVersion() {
+        return String.format(
+            "%s-%b-%s",
+            this.version,
+            this.skipExperimentalLints,
+            this.skipSourceLints.stream().sorted().collect(Collectors.joining(","))
```
Thayorns: The lint names go into the segment verbatim, and CachePath.get() hands that string to `base.resolve(this.semver)`, so the whole join becomes one directory name. Names here run to `unlint-non-existing-defect` length, so a project skipping a dozen of them pushes the component past the 255-byte limit ext4 and APFS enforce and every cache write fails outright rather than just missing. The trackLocations/coverageTracking precedent this follows folds booleans, which have a fixed width.

Thayorns (commented):
Both flags belong in the key and sorting the set before joining is the right call, since two builds naming the same lints in a different order should share an entry. Pinning the tojo hash with allTojosWithHash is what makes the test actually collide the two keys rather than accidentally miss. One thing the comma leaves open: skipping `{"a,b"}` and skipping `{"a", "b"}` produce the same segment and therefore the same entry.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
morphqdd: Fixed. skipSourceLints now gets folded into a short SHA-256 digest (Hashed, a new small class matching the existing Fingerprint/Sha idiom here) instead of joined verbatim, so the path segment stays fixed-width regardless of how many or how long the skipped lint names are.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (137 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3435](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+719](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6272, opened 2026-08-02
#6236: fold parse-stage XSL fingerprint into the parse cache key

### LINKED ISSUE #6236: Parse cache key lacks XSL fingerprint, causing stale XMIR with edited parse-stage stylesheets

morphqdd:
`Parsing`'s cache key was the plugin version plus a hash of the local package object names, with no fingerprint of the 20 `Canonical` parse-stage XSLs that actually transform the parsed source. Editing one of those stylesheets while a `~/.eo` cache from before the edit still exists produces stale, silently wrong XMIR. `Fingerprint` already exists for exactly this purpose and is used by the transpile cache key; `Parsing` never used it.

Added a public `Canonical.XSLS` array listing the 20 XSL classpath resources in the order the pipeline applies them, kept separate from the grouped `TrClasspath`/`TrDefault` calls that build the actual pipeline (same trade-off already accepted for `Transpilation.IMPORTS`: a hand-maintained list that must be kept in sync by hand). `Parsing` now folds `new Fingerprint(Canonical.XSLS).get()` into its cache-key version segment, the same way `Transpilation.version()` already does for the transpile XSLs.

Updated `MjParseTest`'s `cacheVersion()` test helper to mirror the new segment. Confirmed `parsesWithCache` fails on the unfixed `Parsing.java` (the pre-seeded cache entry, built at the path the fixed test helper expects, is missed and a real parse happens instead) and passes with the fix. Full eo-parser and eo-maven-plugin test suites and qulice on both modules are clean.

Scope note: the issue's "What Should Happen" also mentions the per-file lint cache key. That one was already fixed separately in #6235, for a different cause (the `skipSourceLints`/`skipExperimentalLints` flags, not an XSL fingerprint) — `Linting` doesn't apply XSL transforms the way `Canonical` does, so no equivalent fingerprint applies to it.

Closes #6236

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Canonical.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Canonical.java
@@ -35,6 +35,41 @@ import org.cactoos.scalar.Unchecked;
 public final class Canonical implements UnaryOperator<XML> {
 
     /**
+     * The classpath resources of every XSL this pipeline applies, in the
+     * order they run (mirroring {@link Canonical.Pipeline#value()}), so a
+     * caller that caches this pipeline's output (see {@code Parsing} in
+     * {@code eo-maven-plugin}, #6236) can fold a fingerprint of them into
+     * its cache key. Kept as a plain data list, separate from the grouped
+     * {@code TrClasspath}/{@code TrDefault} calls in
+     * {@link Canonical.Pipeline#value()} that actually build the pipeline,
+     * so must be kept in sync with them by hand — the same trade-off
+     * already accepted for {@code Transpilation.IMPORTS} in
+     * {@code eo-maven-plugin}.
+     */
+    public static final String[] XSLS = {
+        "/org/eolang/parser/parse/wrap-applications.xsl",
+        "/org/eolang/parser/parse/resolve-self.xsl",
+        "/org/eolang/parser/parse/resolve-local-names.xsl",
+        "/org/eolang/parser/parse/validate-before-stars.xsl",
+        "/org/eolang/parser/parse/resolve-before-stars.xsl",
+        "/org/eolang/parser/parse/fragile-dispatch.xsl",
+        "/org/eolang/parser/parse/wrap-method-calls.xsl",
+        "/org/eolang/parser/parse/const-to-dataized.xsl",
+        "/org/eolang/parser/parse/stars-to-tuples.xsl",
+        "/org/eolang/parser/parse/vars-float-up.xsl",
+        "/org/eolang/parser/parse/move-voids-up.xsl",
+        "/org/eolang/parser/parse/validate-objects-count.xsl",
+        "/org/eolang/parser/parse/build-fqns.xsl",
+        "/org/eolang/parser/parse/expand-aliases.xsl",
+        "/org/eolang/parser/parse/resolve-aliases.xsl",
+        "/org/eolang/parser/parse/add-default-package.xsl",
+        "/org/eolang/parser/parse/roll-bases.xsl",
+        "/org/eolang/parser/parse/cti-adds-errors.xsl",
+        "/org/eolang/parser/parse/mandatory-as.xsl",
+        "/org/eolang/parser/parse/set-locators.xsl",
+    };
+
+    /**
      * The pipeline, built lazily and only once.
      */
     private final Unchecked<UnaryOperator<XML>> pipeline;
--- a/eo-parser/src/main/java/org/eolang/parser/Canonical.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Canonical.java
@@ -11,6 +11,7 @@ import com.yegor256.xsline.TrClasspath;
 import com.yegor256.xsline.TrDefault;
 import com.yegor256.xsline.TrJoined;
 import com.yegor256.xsline.Xsline;
+import java.util.List;
 import java.util.function.UnaryOperator;
 import org.cactoos.Scalar;
 import org.cactoos.scalar.Sticky;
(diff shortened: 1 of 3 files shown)


morphqdd:

eo-parser/src/main/java/org/eolang/parser/Canonical.java
```diff
@@ -34,6 +34,41 @@
...
+        "/org/eolang/parser/parse/move-voids-up.xsl",
+        "/org/eolang/parser/parse/validate-objects-count.xsl",
+        "/org/eolang/parser/parse/build-fqns.xsl",
+        "/org/eolang/parser/parse/expand-aliases.xsl",
+        "/org/eolang/parser/parse/resolve-aliases.xsl",
+        "/org/eolang/parser/parse/add-default-package.xsl",
+        "/org/eolang/parser/parse/roll-bases.xsl",
+        "/org/eolang/parser/parse/cti-adds-errors.xsl",
+        "/org/eolang/parser/parse/mandatory-as.xsl",
+        "/org/eolang/parser/parse/set-locators.xsl",
```
Thayorns: Eight of these twenty import a shared library: add-default-package, build-fqns, const-to-dataized, cti-adds-errors, move-voids-up, roll-bases, set-locators and vars-float-up all pull in `_funcs.xsl`, `_specials.xsl` or both. Editing one of those libraries changes what the pipeline emits while every file in this array stays byte-identical, so the fingerprint does not move and the cache still hands back the old XMIR. That is the gap Transpilation.IMPORTS exists to close for the transpile key.

eo-parser/src/main/java/org/eolang/parser/Canonical.java
```diff
@@ -34,6 +34,41 @@
...
+     * caller that caches this pipeline's output (see {@code Parsing} in
+     * {@code eo-maven-plugin}, #6236) can fold a fingerprint of them into
+     * its cache key. Kept as a plain data list, separate from the grouped
+     * {@code TrClasspath}/{@code TrDefault} calls in
+     * {@link Canonical.Pipeline#value()} that actually build the pipeline,
+     * so must be kept in sync with them by hand — the same trade-off
+     * already accepted for {@code Transpilation.IMPORTS} in
+     * {@code eo-maven-plugin}.
+     */
+    public static final String[] XSLS = {
```
Thayorns: A public array field is not a constant: anything on the classpath can assign `Canonical.XSLS[3]` and every cache key computed afterwards is silently wrong, with the damage landing in `~/.eo` where nobody will look for it. Transpilation.XSLS and Transpilation.IMPORTS are both package-private for that reason; this one is exported from eo-parser as API.

Thayorns (commented):
The twenty entries match Canonical.Pipeline.value() exactly, in order, including the two that sit outside the TrClasspath groups. The gap is what the pipeline pulls in behind those twenty, and where the list ends up living.

eo-parser/src/main/java/org/eolang/parser/Canonical.java
morphqdd: Added Canonical.IMPORTS (_funcs.xsl, _specials.xsl) and folded it into the fingerprint in Parsing.java and the test's cacheVersion helper, same pattern Transpilation.IMPORTS already uses (#6032). Verified the fingerprint actually differs once IMPORTS is included.

eo-parser/src/main/java/org/eolang/parser/Canonical.java
morphqdd: Switched XSLS from String[] to an immutable List, same trade-off Transpilation.XSLS/IMPORTS already made. Can't make it package-private like those two though, since Canonical is exported from eo-parser as API and Parsing.java in eo-maven-plugin is a different module.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3451](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+726](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6274, opened 2026-08-02
#6217: reject a zip entry that would unpack outside its destination

### LINKED ISSUE #6217: Zip Slip path traversal in `CentralMaven.unpack` when resolving a dependency jar

morphqdd:
`CentralMaven.unpack` writes each zip entry to a path built directly from the entry's own name, with no check that the resolved path stays inside the destination directory. A dependency jar with an entry like `../../../../evil.txt` writes outside the intended destination when unpacked, and `REPLACE_EXISTING` means an existing file at that path gets silently overwritten. Since the jar comes from a dependency coordinate a source file's `+rt jvm` meta names, a malicious or compromised dependency can write or overwrite arbitrary files on the machine running the build.

Normalizes the destination once, then normalizes each entry's resolved target and checks it still starts with that normalized destination before writing anything, rejecting the whole unpack with an `IOException` if not.

Changed `unpack` from `private` to package-private so the test can call it directly with a crafted malicious jar, without needing network access the other tests in this class use. Added a test building a jar with a legitimate entry and a `../../../../evil-escaped.txt` entry, confirming the unpack now throws and that no file lands outside the destination. Confirmed the test fails on the unfixed check (no exception thrown) and passes with the fix. Full eo-maven-plugin test suite and qulice are clean.

Closes #6217

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/CentralMaven.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/CentralMaven.java
@@ -6,16 +6,12 @@ package org.eolang.maven;
 
 import com.jcabi.log.Logger;
 import java.io.IOException;
-import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.function.BiConsumer;
-import java.util.zip.ZipEntry;
-import java.util.zip.ZipInputStream;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.repository.internal.MavenRepositorySystemUtils;
 import org.eclipse.aether.DefaultRepositorySystemSession;
@@ -182,7 +178,7 @@ final class CentralMaven implements BiConsumer<Dependency, Path> {
             artifact.getFile()
         );
         try {
-            CentralMaven.unpack(artifact.getFile().toPath(), dest);
+            new UnpackedJar(artifact.getFile().toPath(), dest).unpack();
         } catch (final IOException ex) {
             throw new IllegalStateException(
                 String.format(
@@ -202,39 +198,6 @@ final class CentralMaven implements BiConsumer<Dependency, Path> {
                 this, "%s:%s:%s:%s unpacked to %[file]s",
                 dep.getGroupId(), dep.getArtifactId(), classifier, dep.getVersion(), dest
             );
-        }
-    }
-
-    /**
-     * Unpacks a JAR (ZIP) file into the given directory.
-     * @param jar Path to the JAR file
-     * @param dest Destination directory
-     * @throws IOException If unpacking fails
-     */
-    static void unpack(final Path jar, final Path dest) throws IOException {
-        Files.createDirectories(dest);
-        final Path home = dest.normalize();
-        try (ZipInputStream zis = new ZipInputStream(Files.newInputStream(jar))) {
-            ZipEntry entry = zis.getNextEntry();
-            while (entry != null) {
-                final Path target = dest.resolve(entry.getName()).normalize();
-                if (!target.startsWith(home)) {
-                    throw new IOException(
-                        String.format(
-                            "Zip entry '%s' would unpack to '%s', outside of '%s'",
-                            entry.getName(), target, home
-                        )
-                    );
-                }
-                if (entry.isDirectory()) {
-                    Files.createDirectories(target);
-                } else {
-                    Files.createDirectories(target.getParent());
-                    Files.copy(zis, target, StandardCopyOption.REPLACE_EXISTING);
-                }
-                zis.closeEntry();
-                entry = zis.getNextEntry();
-            }
         }
     }
 
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/CentralMaven.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/CentralMaven.java
@@ -206,6 +206,39 @@ final class CentralMaven implements BiConsumer<Dependency, Path> {
     }
 
     /**
+     * Unpacks a JAR (ZIP) file into the given directory.
+     * @param jar Path to the JAR file
+     * @param dest Destination directory
+     * @throws IOException If unpacking fails
+     */
+    static void unpack(final Path jar, final Path dest) throws IOException {
+        Files.createDirectories(dest);
+        final Path home = dest.normalize();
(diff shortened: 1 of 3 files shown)


morphqdd:

eo-maven-plugin/src/test/java/org/eolang/maven/CentralMavenTest.java
```diff
@@ -58,6 +63,30 @@ void unpacksFilesFromMavenCentral(@Mktmp final Path temp) {
...
+        }
+        final Path dest = temp.resolve("dest");
+        Assertions.assertThrows(
+            IOException.class,
+            () -> CentralMaven.unpack(jar, dest),
+            "A zip entry whose target lands outside the destination directory must be rejected"
+        );
+        MatcherAssert.assertThat(
+            "The escaping entry must not be written outside the destination directory",
+            temp.resolve("evil-escaped.txt").toFile(),
```
Thayorns: This checks the wrong place. The entry has four `..` and the destination is `temp/dest`, so the escaping write resolves to `temp.parent.parent.parent/evil-escaped.txt`, three directories above `temp`. Nothing would ever create `temp/evil-escaped.txt`, so this passes with or without the check in unpack. On unfixed code the file would land outside the temp directory entirely, somewhere in `/tmp` or above it, and stay there.

eo-maven-plugin/src/test/java/org/eolang/maven/CentralMavenTest.java
```diff
@@ -58,6 +63,30 @@ void unpacksFilesFromMavenCentral(@Mktmp final Path temp) {
         );
     }
 
+    @Test
+    void rejectsZipEntryEscapingDestination(@Mktmp final Path temp) throws IOException {
```
Thayorns, line 69: The class carries `@ExtendWith({MktmpResolver.class, WeAreOnline.class})` on line 35, so every test in it is skipped when there is no connectivity. This one builds its own jar and never touches the network, but it inherits the gate anyway, so the check it guards is not exercised in an offline build.

Thayorns (commented):
The boundary check itself is sound: resolving then normalizing catches `..` and an absolute entry name alike, and Path.startsWith compares name elements so a sibling directory sharing a prefix cannot slip through. Both notes are about the test. It also carries two assertions, so if the throw stops happening you never reach the question of what was written.

eo-maven-plugin/src/test/java/org/eolang/maven/CentralMavenTest.java
morphqdd: Confirmed and fixed. With 4 levels of `..` the entry resolves 3 directories above `temp`, not to `temp/evil-escaped.txt`, so the old assertion passed with or without the fix. Reduced the entry to one `..` (still outside `dest`, the actual boundary) and compute the target the same way `CentralMaven.unpack` does. Verified it fails on the unfixed code (no exception thrown) and passes with the fix.

Side note while fixing this: the real escape target lands on a path that's shared across separate `mvn test` runs (target/mktmp isn't cleaned between runs), so a stale file from an earlier run made the corrected assertion fail on garbage at first. The one-`..` version stays inside the per-test-run unique temp dir, so it doesn't have that problem.

eo-maven-plugin/src/test/java/org/eolang/maven/CentralMavenTest.java
morphqdd: Moved `WeAreOnline` off the class and onto the four tests that actually need it. `rejectsZipEntryEscapingDestination` builds its own jar and never touches the network, so it now runs in offline builds too.

eo-maven-plugin/src/main/java/org/eolang/maven/CentralMaven.java
```diff
@@ -212,6 +212,39 @@ public BiConsumer<Dependency, Path> andThen(
         throw new UnsupportedOperationException("not implemented #andThen()");
     }
 
+    /**
+     * Unpacks a JAR (ZIP) file into the given directory.
+     * @param jar Path to the JAR file
+     * @param dest Destination directory
+     * @throws IOException If unpacking fails
+     */
+    static void unpack(final Path jar, final Path dest) throws IOException {
```
yegor256: @morphqdd we don't appreciate static methods at all. think about better design.

morphqdd: Fixed. Extracted UnpackedJar, a plain class taking jar and dest in its constructor with unpack() as an instance method.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (141 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3333](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+920](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2722](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6275, opened 2026-08-02
#6216: interrupt a hung Mojo instead of waiting forever

### LINKED ISSUE #6216: `-Deo.timeout` never interrupts a hung Mojo due to inverted condition in `Deadline`
## What Happens

`-Deo.timeout`, wired through `Deadline.java` and used by `MjSafe`, is documented to stop a single hung Mojo from hanging the whole build forever. It never actually interrupts a running Mojo.

```java
boolean terminated = false;
service.shutdown();
while (!terminated) {
    terminated = service.awaitTermination(60, TimeUnit.SECONDS);
    if (terminated) {
        service.shutdownNow();
    }
}
```

`shutdownNow()` is the only call in this method that interrupts the currently running task. It only runs inside the `if (terminated)` branch, which by definition means the pool has already stopped on its own. So `shutdownNow()` never actually cuts off a task that is still running; the condition it is guarded by is the opposite of what it needs to be.

Measured directly against this class:
- A 1 second deadline around a task that sleeps 10 seconds: the deadline check returned only after roughly 10.1 seconds, letting the task run to completion, with no interruption at any point.
- A 1 second deadline around a task that never returns: the deadline check was still blocked after 20+ seconds of wall clock time.

This is exactly the failure mode the class's own documentation says it exists to prevent. No `DeadlineTest` class exists to catch this.

## What Should Happen

The condition should be inverted so `shutdownNow()` (or an earlier explicit call to it right after the timeout window elapses, before the polling loop) actually interrupts a still-running task once the deadline has passed, rather than only cleaning up after a task that already finished on its own.

morphqdd:
`Deadline.stopped` called `service.shutdownNow()` only inside `if (terminated)`, which by definition means the executor had already stopped on its own. So the only call that actually interrupts a still-running task never ran while the task was still running; a hung Mojo just kept looping in `awaitTermination` forever, exactly the failure mode the class exists to prevent.

Moved `shutdownNow()` to run immediately, before the polling loop, instead of guarding it behind `terminated`. This interrupts a still-running task right away (or a `TimeoutException` had already been raised further up in `spent()` by the time `stopped()` runs). Calling it when the task already finished on its own is harmless — the single worker thread is idle at that point, so there is nothing to interrupt. Also simplified: `shutdown()` became redundant once `shutdownNow()` runs first, since `shutdownNow()` already stops new submissions too.

Added `DeadlineTest`, which didn't exist before. `interruptsAHungBodyPromptlyOnTimeout` runs a body that sleeps for `Long.MAX_VALUE` (never returns on its own) under a 1-second deadline and asserts the call returns within 30 seconds. Confirmed this test genuinely hangs (30+ seconds, had to kill the test process) on the unfixed code, and completes quickly with the fix. Full eo-maven-plugin test suite and qulice are clean.

Closes #6216

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Deadline.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Deadline.java
@@ -108,13 +108,10 @@ final class Deadline {
      */
     private static void stopped(final ExecutorService service) {
         boolean terminated = false;
-        service.shutdown();
+        service.shutdownNow();
         while (!terminated) {
             try {
                 terminated = service.awaitTermination(60, TimeUnit.SECONDS);
-                if (terminated) {
-                    service.shutdownNow();
-                }
             } catch (final InterruptedException ex) {
                 service.shutdownNow();
                 Thread.currentThread().interrupt();
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Deadline.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Deadline.java
@@ -103,19 +103,33 @@ final class Deadline {
     }
 
     /**
-     * Wait for the service to stop, whatever it takes.
+     * Interrupt the service and wait a bounded time for it to stop.
+     *
+     * <p>{@code shutdownNow()} only interrupts the running task; a task
+     * blocked on something that ignores interruption (a tight CPU loop, a
+     * blocking socket read) may never terminate. Waiting for it
+     * unconditionally would hang this method, and with it the {@code
+     * finally} block of {@link #spent}, forever - silently swallowing
+     * whatever {@link MojoFailureException} the deadline itself just
+     * raised. So this gives up after one bounded wait instead of retrying
+     * without limit.</p>
+     *
      * @param service The service that ran the body
      */
     private static void stopped(final ExecutorService service) {
-        boolean terminated = false;
         service.shutdownNow();
-        while (!terminated) {
-            try {
-                terminated = service.awaitTermination(60, TimeUnit.SECONDS);
-            } catch (final InterruptedException ex) {
-                service.shutdownNow();
-                Thread.currentThread().interrupt();
-            }
+        boolean terminated;
+        try {
+            terminated = service.awaitTermination(60, TimeUnit.SECONDS);
+        } catch (final InterruptedException ex) {
+            Thread.currentThread().interrupt();
+            terminated = false;
+        }
+        if (!terminated) {
+            Logger.warn(
+                Deadline.class,
+                "Body thread did not terminate within 60 seconds of being interrupted, abandoning it"
+            );
         }
     }
 }
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/DeadlineTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/DeadlineTest.java
@@ -1,0 +1,47 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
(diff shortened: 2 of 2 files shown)


morphqdd:

eo-maven-plugin/src/test/java/org/eolang/maven/DeadlineTest.java
```diff
@@ -0,0 +1,47 @@
...
+    @Test
+    void letsAFastBodyFinish() {
+        Assertions.assertDoesNotThrow(
+            () -> new Deadline(this, 5, false).spent(() -> "done"),
+            "A body that finishes well within the deadline must not fail"
+        );
+    }
+
+    @Test
+    void interruptsAHungBodyPromptlyOnTimeout() {
```
Thayorns, line 31: On unfixed code this does not go red, it hangs, which is what you saw when you had to kill the process. `stopped` loops on awaitTermination with no bound, so assertThrows never returns and the elapsed-seconds check below it is never reached. That makes the assertion that names the behaviour unreachable in precisely the case it exists for, and a regression here stalls CI rather than failing it. JUnit's `@Timeout` is what turns that into a failure.

eo-maven-plugin/src/main/java/org/eolang/maven/Deadline.java
```diff
@@ -108,13 +108,10 @@ private void reported(final String msg, final Throwable problem)
      */
     private static void stopped(final ExecutorService service) {
         boolean terminated = false;
-        service.shutdown();
+        service.shutdownNow();
         while (!terminated) {
```
Thayorns: This loop still waits forever. shutdownNow interrupts, it does not kill, so a body sitting in a tight CPU loop or blocked on a socket read never reaches terminated and awaitTermination just keeps returning false every 60 seconds. Since `stopped` runs in the finally of `spent`, the MojoFailureException that the timeout path raised on line 66 never leaves the method, and Maven sees a build that is stuck rather than one that failed on a deadline.

Thayorns (commented):
The diagnosis is right: shutdownNow only ran under `if (terminated)`, which is exactly when there is nothing left to interrupt, so the one call that could unblock a hung body never fired while it was hung. Moving it ahead of the loop is safe on the other paths, since submit(...).get() has already returned by the time the finally runs. The hung-body test carries two assertions, and the timing one after assertThrows can only be reached once the first has already succeeded.

eo-maven-plugin/src/test/java/org/eolang/maven/DeadlineTest.java
morphqdd: Added @Timeout(30). If Deadline.spent's own timeout handling ever regresses, this test now fails fast instead of hanging the build.

eo-maven-plugin/src/main/java/org/eolang/maven/Deadline.java
morphqdd: Fixed. `stopped()` now waits once, bounded (60s), and logs a warning and gives up if the thread is still alive after that, instead of retrying awaitTermination forever. Verified with a busy-loop body that ignores interruption: old code doesn't return within 75s (killed by timeout), new code returns in ~61s with the MojoFailureException intact.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3467](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+733](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6276, opened 2026-08-02
#6218: fix swapped name/url format arguments in object-not-found message

### LINKED ISSUE #6218: `OyRemote` "object not found" error message has swapped format arguments
## What Happens

`OyRemote`'s "object not found" error message has its format arguments swapped, so it names the wrong thing in each slot.

The format string is `"EO object '%s' is not found in this GitHub repository: ... by url: %s"`, but the call passes `(url, name)` instead of `(name, url)`.

Confirmed by driving `OyRemote.get` against an unreachable host and inspecting the resulting message:

```
EO object 'http://127.0.0.1:1/deadbeef/txt/sprintf.eo' is not found in this GitHub repository: ... by url: org.eolang.txt.sprintf
```

The URL appears where the object name should read, and the object name appears where the URL should read.

This message was changed once already: the closed issue #3093 asked specifically for the URL to be shown in this message, and the fix that added the `by url: %s` clause got the two arguments backwards, making the message less useful than intended, not more.

## What Should Happen

The two arguments should be passed in the order the format string expects, `(name, url)`, so the object name and the URL each show up where the message says they will.

morphqdd:
The "not found" message's format string is `"EO object '%s' ... by url: %s"`, but the call passed `(url, name)` instead of `(name, url)`, so the URL showed up in the object-name slot and the object name showed up in the url slot.

Swapped the two arguments to match the format string.

Added a test driving `OyRemote.get` against a local server that always answers 404, asserting the object name shows up after "EO object" and the URL shows up after "by url". Confirmed it fails on the unfixed code (the name assertion fails, since the URL appears there instead) and passes with the fix. Full eo-maven-plugin test suite and qulice are clean.

Closes #6218

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/OyRemote.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/OyRemote.java
@@ -99,8 +99,8 @@ final class OyRemote implements Objectionary {
                 throw new IOException(
                     String.format(
                         "EO object '%s' is not found in this GitHub repository: https://github.com/objectionary/home by url: %s. This means that you either misspelled the name of it or simply referred to your own local object somewhere in your code as if it was an object of 'org.eolang' package. Check the sources and make sure you always use +alias meta when you refer to an object outside of 'org.eolang', even if this object belongs to your package.",
-                        url,
-                        name
+                        name,
+                        url
                     ),
                     input
                 );
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/OyRemoteTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/OyRemoteTest.java
@@ -23,11 +23,6 @@ import org.junit.jupiter.api.extension.ExtendWith;
  */
 final class OyRemoteTest {
 
-    /**
-     * The object name used by the "not found" message tests.
-     */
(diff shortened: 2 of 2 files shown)


morphqdd:

Thayorns (approved):
The swap matches the format string, and both tests genuinely fail on unfixed code, since with the arguments reversed the substring "by url: http" does not occur anywhere in the message. Splitting the two slots into two tests keeps each one to a single behaviour, at the cost of a full server round trip apiece. Both pin the prose around the values, so the wording of that sentence is now fixed by tests as well as by the format string.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+621](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3483](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6277, opened 2026-08-02
#6219: fail clearly on a malformed +rt jvm location instead of ArrayIndexOutOfBoundsException

### LINKED ISSUE #6219: `DpsDefault.artifact` throws raw `ArrayIndexOutOfBoundsException` on malformed `+rt jvm` location
## What Happens

`DpsDefault.artifact` throws a raw `ArrayIndexOutOfBoundsException`, naming no file and no meta, when a `+rt jvm` meta's location is shorter than the three colon-separated parts it expects.

```java
final String[] parts = coords.iterator().next().split(":");
final Dep dependency = new Dep().withGroupId(parts[0]).withArtifactId(parts[1]);
if (parts.length == 3) { ... } else { dependency.withClassifier(parts[2]) ... }
```

Confirmed with the exact XMIR shape a real parse of a `+rt jvm` meta produces:

| `+rt jvm` location | result |
|---|---|
| `org.eolang:eo-runtime:0.0.0` | OK |
| `org.eolang:eo-runtime` | `ArrayIndexOutOfBoundsException: Index 2 out of bounds for length 2` |
| `eo-runtime` | `ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1` |
| `a:b:c:d:e` | silently truncated to `a:b:c:d`, extra parts dropped with no warning |

The `incorrect-rt-parts` lint (severity critical) only counts the number of `part` elements. `org.eolang:eo-runtime` (missing version) still has exactly two parts, so this lint passes it. `incorrect-jvm-rt-location` does catch a missing version, but only as a warning, so with `-Deo.failOnWarning=false` the build proceeds past linting and then `MjResolve` dies on the bare array-index crash with no context about which meta or file caused it.

## What Should Happen

`DpsDefault.artifact` should validate the number of colon-separated parts before indexing into them, and raise a clear error naming the malformed `+rt jvm` value when there are too few (or too many) parts, instead of an unqualified `ArrayIndexOutOfBoundsException`.

morphqdd:
`DpsDefault.artifact` split a `+rt jvm` meta's location on `:` and indexed straight into the result without checking how many parts it actually had. A location with fewer than 3 parts (missing the version, or missing group entirely) crashed with a raw `ArrayIndexOutOfBoundsException` naming no file and no meta; a location with more than 4 parts was silently truncated with no warning.

Added a check right after the split: if the part count is not 3 or 4, throw `IllegalStateException` naming the malformed location and the source file, before any indexing happens.

Added a test driving `MjResolve` against a program with `+rt jvm org.eolang:eo-runtime` (missing version), asserting the failure's cause chain names the malformed location and confirming no cause in that chain is an unqualified `ArrayIndexOutOfBoundsException`. Had to walk the full cause chain in the test because `Moja.execute()` wraps `MojoFailureException` in `IllegalStateException(ex)` (whose message becomes the wrapped exception's `toString()`), which itself wraps the `ExecutionException` from `Deadline`, which wraps my thrown exception — so the useful message is a few levels down, not on the immediately-thrown exception. Confirmed the test fails on the unfixed code (an `ArrayIndexOutOfBoundsException` shows up in the chain) and passes with the fix. Full eo-maven-plugin test suite and qulice are clean.

Closes #6219


morphqdd:

eo-maven-plugin/src/test/java/org/eolang/maven/MjResolveTest.java
```diff
@@ -65,6 +68,30 @@ void resolvesWhenPlaceDirectoryExistsButIsEmpty(@Mktmp final Path temp) throws I
...
+    @Test
+    void reportsMalformedRtJvmLocationClearly(@Mktmp final Path temp) throws IOException {
+        final FakeMaven maven = new FakeMaven(temp).withProgram(
+            "+package foo.x",
+            "+rt jvm org.eolang:eo-runtime",
+            String.format("+version 0.25.0%n"),
+            "[] > main /bytes"
+        );
+        MatcherAssert.assertThat(
+            "The error must name the malformed '+rt jvm' location somewhere in its cause chain, and no cause may be an unqualified ArrayIndexOutOfBoundsException",
```
Thayorns: The message makes two claims and the assertion checks one. Nothing below looks at the class of any cause, so "no cause may be an unqualified ArrayIndexOutOfBoundsException" would still read as satisfied if one were in the chain alongside a message naming the location. `Matchers.is(true)` also collapses the whole chain to a bare boolean, so a failure prints "was <false>" and none of the messages that were actually found.

eo-maven-plugin/src/test/java/org/eolang/maven/MjResolveTest.java
```diff
@@ -65,6 +68,30 @@ void resolvesWhenPlaceDirectoryExistsButIsEmpty(@Mktmp final Path temp) throws I
         );
     }
 
+    @Test
+    void reportsMalformedRtJvmLocationClearly(@Mktmp final Path temp) throws IOException {
```
Thayorns, line 69: The thing under test is four lines of DpsDefault, but this drives a whole Mojo through Moja.execute and then digs down through IllegalStateException, ExecutionException and MojoFailureException to find the message. The `causes` helper exists only to absorb that distance, so a change to how Deadline or Moja wraps failures turns this red without DpsDefault changing at all.

Thayorns (commented):
Checking the part count before indexing is the right place for it, and the message carries the location, the file and the count, so a build that hits it can be fixed from the log alone. The cycle guard in `causes` is a nice touch. Both notes are about the test.

eo-maven-plugin/src/test/java/org/eolang/maven/MjResolveTest.java
morphqdd: Fixed. Test now builds a TjsForeign tojo directly and calls DpsDefault.iterator(), so it's the exception assertThrows itself checks, no cause chain to inspect and get wrong.

eo-maven-plugin/src/test/java/org/eolang/maven/MjResolveTest.java
morphqdd: Rewrote the test to construct DpsDefault directly with an in-memory tojo instead of going through FakeMaven/Moja. No more digging through wrapper exceptions.

yegor256:

morphqdd:
@yegor256 which harness do you mean? The YAML packs we have are all XSL driven: `probe-packs`, `transpile-packs`, `print-packs` in eo-maven-plugin, `eo-syntax`/`eo-packs` in eo-parser. This check is in `DpsDefault`, which reads `+rt jvm` out of an already parsed XMIR at resolve time, and eo-parser does not touch `+rt` at all, so none of those packs can reach it.

If you want the malformed location caught earlier, say at parse time, I can move it there instead and add an `eo-syntax` YAML pack for it. Otherwise the JUnit test is the only place this code path is reachable from.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (37) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3513](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+826](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6279, opened 2026-08-02
#6221: return an empty scheme for a schemeless uri

### LINKED ISSUE #6221: `uri.scheme` crashes on a schemeless input instead of returning an empty scheme
## What happens

`uri.scheme` terminates with an `ExFailure` when the input text contains no colon, instead of returning an empty scheme. Every other component of the very same input is parsed gracefully, and `uri ":" .scheme` already returns `""`, so only the no-colon case of `scheme` is broken.

### Runtime evidence

Probed directly against `org.eolang.EOuri`:

| call | actual | expected |
| --- | --- | --- |
| `(uri "localhost").scheme` | **`ExFailure: Length to slice must be >= 0, but was -1`** | `""` |
| `(uri "").scheme` | **`ExFailure: Length to slice must be >= 0, but was -1`** | `""` |
| `(uri "no-scheme/path?q#f").scheme` | **`ExFailure`** | `""` |
| `(uri "localhost").path` | `"localhost"` | `"localhost"` (works) |
| `(uri "localhost").host` / `.port` / `.query` / `.fragment` | `""` | `""` (all work) |
| `(uri ":").scheme` | `""` | `""` (works) |
| `(uri "http://host:80").scheme` | `"http"` | `"http"` (control) |

So `uri "localhost"` is otherwise parsed as a schemeless relative reference (path `"localhost"`, empty host/port/query/fragment) and only `.scheme` blows up.

## Root cause

`eo-runtime/src/main/eo/uri.eo`, the `scheme` attribute (lines 134-136):

```
text.slice > scheme
  0
  scheme-colon
```

`scheme-colon` is `string.index-of text ":"`, which is `-1` when there is no colon. `slice 0 -1` then trips `string.slice`'s own guard "Length to slice must be >= 0, but was -1" and terminates. The other attributes all guard their `index-of` results against `-1` (for example `has-fragment`, `has-port`, `has-query`, `has-userinfo`); `scheme` is the one that does not, even though a schemeless input is exactly what RFC 3986 calls a relative reference.

## Suggested fix

Guard the missing-colon case the same way the siblings do, returning an empty scheme:

```
if. > scheme
  scheme-colon.eq -1
  ""
  text.slice 0 scheme-colon
```

Verified: with this change `(uri "localhost").scheme`, `(uri "").scheme` and `(uri "no-scheme/path?q#f").scheme` all return `""`, while every colon-bearing control is unchanged (`(uri "http://host:80").scheme` = `"http"`, `(uri "pgsql:localhost:899").scheme` = `"pgsql"`, `(uri "file:///etc/passwd").scheme` = `"file"`). `scheme-colon` is a `>>`-bound constant already reused by `rest`, so moving the `index-of` into this guard keeps it available everywhere it is referenced.

A regression test such as `eq. ++> ... (uri "localhost").scheme "" ` would pin it.

gemshrine:
Fixes #6221.

`uri.scheme` terminated with `ExFailure: Length to slice must be >= 0, but was -1` whenever the input had no colon, even though every other component parses a schemeless input gracefully (RFC 3986 relative reference) and `(uri ":").scheme` already returns `""`.

Root cause: `scheme` did `text.slice 0 scheme-colon` where `scheme-colon` is `string.index-of text ":"`, which is `-1` when there is no colon, tripping `string.slice`'s own `>= 0` guard. The sibling attributes (`has-fragment`, `has-port`, `has-query`, `has-userinfo`) all guard their `index-of` against `-1`; `scheme` did not.

Fix guards the missing-colon case the same way, returning an empty scheme:

```
if. > scheme
  scheme-colon.eq -1
  ""
  text.slice 0 scheme-colon
```

Verified by probing `org.eolang.EOuri`: `(uri "localhost").scheme`, `(uri "").scheme` and `(uri "no-scheme/path?q#f").scheme` now return `""`, while every colon-bearing control is unchanged (`http://host:80` = `http`, `pgsql:localhost:899` = `pgsql`, `file:///etc/passwd` = `file`, `:` = `""`). Added a regression test `can-parse-an-empty-scheme-of-a-schemeless-uri`.

### DIFF
--- a/eo-runtime/src/main/eo/uri.eo
+++ b/eo-runtime/src/main/eo/uri.eo
@@ -55,9 +55,7 @@
       hash-idx.eq -1
     slice.
       text.slice >> rest
-        plus.
-          string.index-of text ":" >> scheme-colon
-          1
+        scheme-colon.plus 1
         text.length.minus
           number
             scheme-colon.plus 1
@@ -131,9 +129,14 @@
         number
           question-idx.plus 1
     ""
-  text.slice > scheme
-    0
-    scheme-colon
+  if. > scheme
+    eq.
+      string.index-of text ":" >> scheme-colon
+      -1
+    ""
+    text.slice
+      0
+      scheme-colon
   if. > userinfo
     not. >> has-userinfo
       userinfo-at-idx.eq -1
@@ -176,6 +179,11 @@
     query.
       uri "http://host/some/path?x=1#top"
     "x=1"
+
+  eq. ++> can-parse-an-empty-scheme-of-a-schemeless-uri
+    scheme.
+      uri "localhost"
+    ""
 
   eq. ++> can-parse-the-empty-host-of-a-triple-slash-uri
     host.


Thayorns (approved):
Guarding index-of against -1 brings scheme into line with has-fragment, has-port, has-query and has-userinfo, and `rest` is unaffected because -1 plus 1 was already the zero it wanted. Moving the `>> scheme-colon` declaration down into `scheme` is safe, since a handle belongs to its enclosing formation wherever in the body it is written, but it does leave the two uses in `rest` seventy lines above the line that introduces the name.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (20) hits-of-code; +4 to give you at least something.  Your running score is [+629](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:
@gemshrine dup of #6280

PR #6280, opened 2026-08-02
#6221: return empty scheme for a schemeless uri instead of crashing

### LINKED ISSUE #6221: `uri.scheme` crashes on a schemeless input instead of returning an empty scheme

morphqdd:
`uri.scheme` sliced `text` from 0 to `scheme-colon` (the index of the first `:`) with no guard for the no-colon case, where `scheme-colon` is `-1`. `text.slice 0 -1` then trips `string.slice`'s own "Length to slice must be >= 0" guard. Every sibling attribute (`has-fragment`, `has-port`, `has-query`, `has-userinfo`) already guards its own `index-of` result against `-1`; `scheme` was the one that didn't, even though a schemeless input is a legitimate RFC 3986 relative reference and every other component already parses it fine.

Added the same guard the siblings use: when `scheme-colon` is `-1`, return `""` instead of slicing.

Added a regression test for `(uri "localhost").scheme = ""`, matching the exact case from the issue. Confirmed it fails on the unfixed code with the same `ExFailure` message from the issue, and passes with the fix. Full eo-runtime test suite is clean.

Closes #6221

### DIFF
--- a/eo-runtime/src/main/eo/uri.eo
+++ b/eo-runtime/src/main/eo/uri.eo
@@ -55,9 +55,7 @@
       hash-idx.eq -1
     slice.
       text.slice >> rest
-        plus.
-          string.index-of text ":" >> scheme-colon
-          1
+        scheme-colon.plus 1
         text.length.minus
           number
             scheme-colon.plus 1
@@ -131,9 +129,14 @@
         number
           question-idx.plus 1
     ""
-  text.slice > scheme
-    0
-    scheme-colon
+  if. > scheme
+    eq.
+      string.index-of text ":" >> scheme-colon
+      -1
+    ""
+    text.slice
+      0
+      scheme-colon
   if. > userinfo
     not. >> has-userinfo
       userinfo-at-idx.eq -1
@@ -216,6 +219,11 @@
     scheme.
       uri "pgsql:localhost:899"
     "pgsql"
+
+  eq. ++> can-parse-the-scheme-of-a-schemeless-uri
+    scheme.
+      uri "localhost"
+    ""
 
   eq. ++> can-parse-the-scheme-of-a-triple-slash-uri
     scheme.


morphqdd:

Thayorns (approved):
The guard matches what has-fragment, has-port, has-query and has-userinfo already do, and `rest` comes out unchanged because -1 plus 1 was the zero it was already computing. Putting the case next to the other can-parse-the-scheme-of-* cases keeps the pack readable. Relocating `>> scheme-colon` into `scheme` is fine for resolution, since a handle belongs to its enclosing formation wherever in the body it appears, though the two remaining uses in `rest` are now seventy lines above the line that names it.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (20) hits-of-code; +4 to give you at least something.  Your running score is [+633](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3227](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6281, opened 2026-08-02
#6213: strip org.eolang. prefix in ObjectsIndex.children the same way contains does

### LINKED ISSUE #6213: `ObjectsIndex.children` does not strip the `org.eolang.` prefix that `contains` strips

morphqdd:
`ObjectsIndex.contains(name)` strips a leading `org.eolang.` from `name` before checking the index, since the index itself stores names without that prefix. `children(pkg)` built its lookup prefix directly from the raw `pkg` argument, with no equivalent stripping, so the two methods disagreed about what normalization a package name needs. Not reachable today since every `eo-runtime` object is rooted at `Φ.<name>`, but a latent inconsistency for any future object or user package using the longer form.

Extracted the stripping logic from `contains` into a shared private `stripped` method and used it in both `contains` and `children`.

Added a test for `children("org.eolang.tuple")` returning the same result as `children("tuple")`, confirmed it fails on the unfixed code (empty result) and passes with the fix. Full eo-maven-plugin test suite, qulice, and jtcop are clean.

Closes #6213

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
@@ -76,13 +76,23 @@ final class ObjectsIndex {
     /**
      * Names of all objects located directly inside the given package, e.g.
      * {@code "tuple.each"} and {@code "tuple.eachi"} for {@code "tuple"}, but
-     * not {@code "tuple"} itself nor objects from its sub-packages.
-     * @param pkg Package name, e.g. {@code "tuple"} or {@code "math.vector"}
+     * not {@code "tuple"} itself nor objects from its sub-packages. The
+     * bare root package (stripped to an empty string) works the same way:
+     * its direct children are the top-level names with no dot at all, e.g.
+     * {@code "tuple"} and {@code "math"} themselves.
+     * @param pkg Package name, e.g. {@code "tuple"}, {@code "math.vector"}
+     *  or {@code "org.eolang"} itself
      * @return Object names that live directly in that package
      * @throws Exception If the index can't be read
      */
     Iterable<String> children(final String pkg) throws Exception {
-        final String prefix = String.format("%s.", ObjectsIndex.stripped(pkg));
+        final String stripped = ObjectsIndex.stripped(pkg);
+        final String prefix;
+        if (stripped.isEmpty()) {
+            prefix = stripped;
+        } else {
+            prefix = String.format("%s.", stripped);
+        }
         return new Filtered<>(
             name -> name.startsWith(prefix) && name.indexOf('.', prefix.length()) < 0,
             this.objects.value()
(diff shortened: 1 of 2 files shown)


morphqdd:

eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
```diff
@@ -89,13 +82,32 @@ boolean contains(final String name) throws Exception {
      * @throws Exception If the index can't be read
      */
     Iterable<String> children(final String pkg) throws Exception {
-        final String prefix = String.format("%s.", pkg);
+        final String prefix = String.format("%s.", ObjectsIndex.stripped(pkg));
```
Thayorns: `"org.eolang"` has no trailing dot, so `stripped` returns it untouched and the filter hunts for names beginning `org.eolang.` in an index that stores them bare, giving back an empty list. Probing.java:148 takes everything before the last dot of the probed object, so `org.eolang.string` and `org.eolang.tuple` arrive here as exactly that string, and their siblings are silently never discovered. The description says the mismatch is unreachable today, but that is the path that reaches it.

Thayorns (commented):
Pulling the stripping into one method and calling it from both places is the right shape, and the docblock pointing at convert() explains why the index is short-form in the first place. One package name gets past it.

eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
morphqdd: Confirmed and fixed. Added the missing-dot case to stripped() so a bare "org.eolang" strips to "" the same way "org.eolang." strips normal names. That surfaced a second gap in children() itself: appending "." to an empty prefix made every name fail startsWith("."), since the root's own direct children have no dot at all. Fixed that too. Added a test for children("org.eolang") over a mixed-depth index, verified it fails on the unfixed code (empty result) and passes now.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3243](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+740](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6282, opened 2026-08-02
feat(#4752): add `string.truncated`

asmirnov-backend:
Returns the first `limit` characters of a text, and the text itself when it is
already that short. Unlike `slice`, a `limit` past the end of the text is not an
error, which is what a `%.Ns` specifier needs.

First of three towards a pure-EO `printf` (#4752); the other two are
`string.padded-zeros` and the switch of `printf` itself.

### DIFF
--- a/eo-runtime/src/main/eo/string/truncated.eo
+++ b/eo-runtime/src/main/eo/string/truncated.eo
@@ -1,0 +1,93 @@
+# Returns the first `limit` characters of `text`, so that `Jeffrey` truncated
+# to three characters becomes `Jef`. Characters are counted, and not bytes,
+# so a multi-byte character counts as one. A `text` of `limit` characters
+# or shorter is returned unchanged, which is how this differs from `slice`,
+# where a `limit` beyond the end of the `text` terminates the computation.
+# If `limit` is not a finite non-negative integer (negative, fractional, nan
+# or infinite), the `cant-truncate` fallback is returned, or the bottom
+# object when unbound.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package string
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[text limit cant-truncate] > truncated
+  if. > @
+    or.
+      0.gt limit
+      limit.is-integer.not
+    cant-truncate
+      "Can't truncate text to %f characters".printf
+        * limit
+    if.
+      limit.lt text.length
+      slice
+        text
+        0
+        limit
+      text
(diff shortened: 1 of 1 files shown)


eo-runtime/src/main/eo/string/truncated.eo
```diff
@@ -0,0 +1,93 @@
...
+        limit
+      text
+
+  eq. ++> can-count-characters-and-not-bytes
+    "Жу"
+    truncated
+      "Жук"
+      2
+
+  eq. ++> can-format-the-error-message-of-a-negative-limit
```
Thayorns, line 39: This pins the exact prose of the failure message, down to how %f renders -3. Rewording it, or changing the specifier, breaks a case that is not about either. `rejects-a-negative-limit` two entries down already covers what happens for a negative limit, which is the behaviour that matters.

eo-runtime/src/main/eo/string/truncated.eo
```diff
@@ -0,0 +1,93 @@
...
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[text limit cant-truncate] > truncated
+  if. > @
+    or.
+      0.gt limit
+      limit.is-integer.not
+    cant-truncate
+      "Can't truncate text to %f characters".printf
```
Thayorns, line 23: `%f` turns a character count into "-3.000000", which is not how anyone reads a number of characters. string.slice next door reports the same kind of value with `%d` in both of its messages. I take it %f is there so a limit of 1.75 still prints, but the cost is that every integer limit, which is the normal case, reports six decimal places.

Thayorns (commented):
The guard is right and tighter than it looks: is-integer already folds in is-finite, so pinf and nan are refused by the same conjunct that refuses 1.75, which is what the docblock promises. Counting is consistent too, since both string.length and string.slice walk characters rather than bytes, so the limit means the same thing on both sides of the comparison. nan is the one input named in the docblock that no case exercises.

0crat:
@asmirnov-backend Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+40](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+747](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6283, opened 2026-08-02
feat(#4752): add `string.padded-zeros`

asmirnov-backend:
Pads a text with leading zeros the way a number wants them, keeping a minus
sign in front of the zeros, so that `-42` to five characters is `-0042` and not
`00-42`. That is what a `%05d` specifier needs, and it is the one thing
`padded-left` cannot express.

Second of three towards a pure-EO `printf` (#4752); the first is #6282, the
third is the switch of `printf` itself.

### DIFF
--- a/eo-runtime/src/main/eo/string/padded-zeros.eo
+++ b/eo-runtime/src/main/eo/string/padded-zeros.eo
@@ -1,0 +1,119 @@
+# Returns `text` padded on the left with zeros until it counts `width`
+# characters, keeping a leading minus sign in front of them, so that `-42`
+# padded to five characters becomes `-0042` and not `00-42`. Characters are
+# counted, and not bytes. A `text` of `width` characters or longer is
+# returned unchanged.
+# If `width` is not a finite non-negative integer (negative, fractional, nan
+# or infinite), the `cant-pad` fallback is returned, or the bottom object
+# when unbound.
+
++architect yegor256@gmail.com
++home https://github.com/objectionary/eo
++package string
++version 0.0.0
++spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
++spdx SPDX-License-Identifier: MIT
+
+[text width cant-pad] > padded-zeros
+  if. > @
+    or.
+      0.gt width
+      width.is-integer.not
+    cant-pad
+      "Can't pad text to %f characters with zeros".printf
+        * width
+    if.
+      and.
+        0.lt
+          text.length >> len!
+        "-".eq
+          slice text 0 1
+      string
+        2D-.concat
+          as-bytes.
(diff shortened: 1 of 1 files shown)


eo-runtime/src/main/eo/string/padded-zeros.eo
```diff
@@ -0,0 +1,119 @@
...
+                  width.minus 1
+              "0"
+              cant-pad
+      padded-left
+        text
+        width
+        "0"
+        cant-pad
+
+  eq. ++> can-count-characters-and-not-bytes
```
Thayorns, line 53: `"Ж"` has no minus sign, so this goes down the padded-left branch and is really testing padded-left's character counting rather than anything this object does. The branch that counts characters itself is the other one, where `slice text 1 (len - 1)` splits a multi-byte text by hand, and nothing here gives it a case like `padded-zeros "-Ж" 4`.

eo-runtime/src/main/eo/string/padded-zeros.eo
```diff
@@ -0,0 +1,119 @@
...
+    "00Ж"
+    padded-zeros
+      "Ж"
+      3
+
+  eq. ++> can-format-the-error-message-of-a-negative-width
```
Thayorns, line 59: This pins the exact prose of the message, including how %f renders -4, so the wording and the specifier are both frozen by a case that is about neither. `rejects-a-negative-width` further down already covers the behaviour that matters here.

Thayorns (commented):
The sign-handling branch is careful: putting `0.lt len` ahead of the slice in the `and.` keeps `slice text 0 1` off an empty text, and clamping the inner width at zero when `width` is 1 or less stops `width.minus 1` going negative into padded-left. is-integer folding in is-finite means pinf and nan land in the same conjunct as 3.5, matching the docblock, though nan is the one input named there that no case walks. `%f` also renders a character count as "-4.000000", where string.slice reports the same kind of value with `%d`.

0crat:
@asmirnov-backend Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (119 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+52](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+754](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6284, opened 2026-08-02
#6209: return nan (not the dividend) when number.mod divisor is nan

### LINKED ISSUE #6209: `number.mod` returns the dividend instead of `nan` when the divisor is `nan`
## What happens

`number.mod` returns the dividend unchanged when the divisor is `nan`, instead of returning `nan`. Java's `%` and IEEE remainder both yield `NaN` whenever either operand is `NaN`, so `mod` disagrees with the division semantics its own `can-stay-compatible-with-division` test relies on.

### Runtime evidence

Probed directly against `org.eolang.EO_number.EOmod` and compared to Java `x % y`:

| call | `number.mod` | Java `x % y` |
| --- | --- | --- |
| `mod 5 nan` | **`5`** | `NaN` |
| `mod -5 nan` | **`-5`** | `NaN` |
| `mod 0 nan` | **`0`** | `NaN` |
| `mod nan 5` | `NaN` | `NaN` (control, correct) |
| `mod 5 pinf` | `5` | `5` (control, correct) |
| `mod -5 ninf` | `-5` | `-5` (control, correct) |

The `nan` dividend case is already correct; only a `nan` divisor is wrong.

## Root cause

`eo-runtime/src/main/eo/number/mod.eo`, lines 20-21:

```
divisor.is-finite.not.if
  x
  if.
    ...
```

The guard treats every non-finite divisor identically and returns the dividend `x`. That is correct for `+-inf` (`x mod inf = x`), but wrong for `nan`, where the result must be `nan`.

This is a follow-on from #5979. That issue fixed `mod` returning `nan` for an *infinite* divisor by adding this `is-finite.not.if x` branch (commit `aa5b3ef023`, "return dividend unchanged when divisor is infinite in number.mod"). The branch keys on `is-finite`, which is false for both infinity and `nan`, so the infinity fix over-generalized and swept `nan` into the same "return `x`" path.

## Suggested fix

Split the `nan` case out of the non-finite branch so infinity still returns `x` while `nan` returns `nan`:

```
divisor.is-finite.not.if
  divisor.is-nan.if
    nan
    x
  if.
    ...
```

Verified: with this change `mod 5 nan`, `mod -5 nan` and `mod 0 nan` all return `nan`, while every infinity and finite control above stays unchanged (`mod 5 pinf` = `5`, `mod -13 5` = `-3`, etc.). `is-nan` is the package-level `number` object, dispatched on the divisor exactly like the adjacent `is-finite` call, and `nan` is already referenced by name in this package.

A regression test in the same file, for example `is-nan. ++> ...` over `mod 5 nan`, would pin it.

morphqdd:
`number.mod` returned the dividend unchanged for any non-finite divisor, treating `+-inf` and `nan` the same way. That's correct for infinity (`x mod inf = x`), but wrong for `nan`, where the result must be `nan`, matching Java's `%` and IEEE remainder semantics. This was a follow-on from #5979, which added the `is-finite.not.if x` branch to fix the infinity case and over-generalized it to also catch `nan`.

Split the `nan` case out: when the divisor is non-finite, check `is-nan` first and return `nan`, otherwise return `x` as before (still correct for infinity).

Added a regression test for `mod 5 nan` returning `nan`. Confirmed it fails on the unfixed code (`mod` returns `5` instead of `nan`) and passes with the fix. Full eo-runtime test suite is clean.

Closes #6209

### DIFF
--- a/eo-runtime/src/main/eo/number/mod.eo
+++ b/eo-runtime/src/main/eo/number/mod.eo
@@ -18,7 +18,7 @@
     cant-mod
       "cannot calculate mod by zero"
     divisor.is-finite.not.if
-      x
+      divisor.is-nan.if nan x
       if.
         gt.
           number x.as-bytes >> dividend
@@ -40,6 +40,11 @@
       dividend
     -13 > dividend
     5 > divisor
+
+  is-nan. ++> can-take-a-modulo-by-nan
+    mod
+      5
+      nan
 
   eq. ++> can-take-a-modulo-by-one
     mod


morphqdd:

Thayorns (approved):
Nesting the nan test inside the non-finite branch leaves the two infinity paths exactly where they were, and the four existing pinf/ninf cases still hold them down. `is-nan.` is the only head that can assert this, since nan compares unequal to itself. The zero guard ahead of it is unaffected because `nan.eq 0` is false, so `mod 0 nan` and `mod -5 nan` reach the same branch this case walks.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (7) hits-of-code; +4 to give you at least something.  Your running score is [+637](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (7) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3255](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6285, opened 2026-08-02
#6212: probe @type and @atom candidates, not just @args

### LINKED ISSUE #6212: `add-probes.xsl` still misses `@type` and `@atom` candidates after the `@args` fix
## What Happens

`add-probes.xsl` only collects candidate FQNs to probe from the `@args` attribute of an atom's void. It does not collect from `@type` or `@atom`, the two sibling annotation forms the parser emits for the exact same `/Q.foo`-style type annotation syntax (a plain typed void `? > name /Q.foo`, and an atom's own return signature `[] > name /Q.foo`).

Issue #6174 (closed) was exactly this bug for the `@args` case: `/{Q.chunk}` on a void was never probed, never pulled, never transpiled, so any program touching `malloc` died with `ClassNotFoundException: org.eolang.EOchunk`. The fix added collection from `@args` only.

Confirmed by driving `org.eolang.maven.Probes` directly on three inputs:

| input | probes found |
|---|---|
| `[] > foo /Q.bytes` with `? > size /Q.number` and `? > scope /{Q.chunk}` | `chunk` |
| `[] > foo /Q.bytes` with `? > size /Q.chunk` | none |
| `[] > foo /Q.chunk` with `? > size /Q.number` | none |

The first row is the already-fixed `@args` case. The second and third show the identical gap for `@type` and `@atom`.

No live break inside `eo-runtime` today, because every `/Q.x` reference in the stdlib resolves inside its own file. A user package writing `[] > foo /Q.my.thing` referring to another package's object would hit the same `ClassNotFoundException` #6174 was filed to fix.

## What Should Happen

`add-probes.xsl` should collect candidate FQNs from `@type` and `@atom` the same way it already does from `@args` (stripping the optional trailing `?` from `@type` first). All three attributes carry the same kind of forma reference and should feed the same probe list.

morphqdd:
`add-probes.xsl` only collected candidate FQNs from `@args` (a void's callback arg-types annotation, fixed for #6174). It did not collect from `@type` (a plain typed void's own type annotation) or `@atom` (an atom's own return signature), the two sibling forms the parser emits for the same `/Q.foo` syntax. A user package writing `[] > foo /Q.my.thing` referring to another package's object would hit the same `ClassNotFoundException` #6174 fixed for the `@args` case.

Added two new templates, `type` and `atom`, mirroring the existing `args` template: `type` strips the optional trailing `?` before tokenizing, `atom` tokenizes directly, both skip generic type variables (anything not starting with `Φ`) the same way `args` already does. Wired both into the candidate collection in both entry points (with and without existing metas).

This also means the existing `add-probes-from-void-args.yaml` fixture's own `[] > foo /Q.bytes` (with `? > size /Q.number`) is now correctly probed too, since the atom signature and the type annotation are exactly the same kind of reference the fix targets — updated that fixture's expected probes list to include `bytes` and `number` alongside `chunk`.

Added two new YAML packs, `add-probes-from-void-type.yaml` and `add-probes-from-atom-signature.yaml`, covering the second and third rows from the issue. Confirmed all three (the new two plus the updated existing one) fail on the unfixed XSL and pass with the fix. Full eo-maven-plugin test suite, qulice, and jtcop are clean.

Closes #6212

### DIFF
--- a/eo-maven-plugin/src/main/resources/org/eolang/maven/probe/add-probes.xsl
+++ b/eo-maven-plugin/src/main/resources/org/eolang/maven/probe/add-probes.xsl
@@ -57,10 +57,12 @@
       </part>
     </meta>
   </xsl:function>
-  <!-- Every probe candidate: every non-abstract, non-void base; every
+  <!--
+  Every probe candidate: every non-abstract, non-void base; every
   callback arg-type on a void attribute; every void's own type
   annotation; every atom's return signature. Shared by both entry
-  points below so the two can't drift on which sources are probed. -->
+  points below so the two can't drift on which sources are probed.
+  -->
   <xsl:template name="eo:candidates" as="element()*">
     <xsl:apply-templates select="//o[not(eo:abstract(.)) and not(eo:void(.))]" mode="create"/>
     <xsl:apply-templates select="//o[eo:void(.) and @args]" mode="args"/>
@@ -96,10 +98,12 @@
       </xsl:for-each>
     </xsl:copy>
   </xsl:template>
-  <!-- Every cumulative dotted prefix of a forma, shortest first, e.g.
+  <!--
+  Every cumulative dotted prefix of a forma, shortest first, e.g.
   "Φ.org.number" yields "Φ", "Φ.org", "Φ.org.number". Shared by every
   place below that walks a dotted forma into probe candidates, so a
-  change to how a forma is split only has one place to happen. -->
+  change to how a forma is split only has one place to happen.
+  -->
   <xsl:template name="eo:dotted-prefixes" as="element()*">
     <xsl:param name="forma" as="xs:string?"/>
     <xsl:variable name="parts" select="tokenize($forma, '\.')"/>
--- a/eo-maven-plugin/src/main/resources/org/eolang/maven/probe/add-probes.xsl
+++ b/eo-maven-plugin/src/main/resources/org/eolang/maven/probe/add-probes.xsl
@@ -57,13 +57,20 @@
       </part>
     </meta>
   </xsl:function>
+  <!-- Every probe candidate: every non-abstract, non-void base; every
+  callback arg-type on a void attribute; every void's own type
+  annotation; every atom's return signature. Shared by both entry
+  points below so the two can't drift on which sources are probed. -->
+  <xsl:template name="eo:candidates" as="element()*">
+    <xsl:apply-templates select="//o[not(eo:abstract(.)) and not(eo:void(.))]" mode="create"/>
+    <xsl:apply-templates select="//o[eo:void(.) and @args]" mode="args"/>
+    <xsl:apply-templates select="//o[@type]" mode="type"/>
+    <xsl:apply-templates select="//o[@atom]" mode="atom"/>
+  </xsl:template>
   <!-- ENTRY POINT 1 - no metas -->
   <xsl:template match="/object[not(metas)]">
     <xsl:variable name="candidates" as="element()*">
-      <xsl:apply-templates select="//o[not(eo:abstract(.)) and not(eo:void(.))]" mode="create"/>
-      <xsl:apply-templates select="//o[eo:void(.) and @args]" mode="args"/>
-      <xsl:apply-templates select="//o[@type]" mode="type"/>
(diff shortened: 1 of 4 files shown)


morphqdd:

eo-maven-plugin/src/main/resources/org/eolang/maven/probe/add-probes.xsl
```diff
@@ -103,6 +107,35 @@
...
     </xsl:for-each>
   </xsl:template>
+  <!-- A void's own type annotation (`? > name /Q.foo`, optionally `/Q.foo?`). -->
+  <!-- Homed formas start with `Φ`; generic type variables (A-F) do not and -->
+  <!-- must be skipped, since they are not real objects to probe. -->
+  <xsl:template match="o" mode="type" as="element()*">
+    <xsl:variable name="forma" select="replace(@type, '\?$', '')"/>
+    <xsl:if test="starts-with($forma, 'Φ')">
+      <xsl:variable name="parts" select="tokenize($forma, '\.')"/>
+      <xsl:for-each select="$parts">
```
Thayorns: This is the fourth copy of the same six lines. `args` on line 97 and the composite-base `create` template both walk a dotted forma into cumulative prefixes exactly this way, and the two templates added here are that block with the recursion dropped. Whoever changes how a forma is split into probes now has four places to find.

eo-maven-plugin/src/main/resources/org/eolang/maven/probe/add-probes.xsl
```diff
@@ -62,6 +62,8 @@
     <xsl:variable name="candidates" as="element()*">
       <xsl:apply-templates select="//o[not(eo:abstract(.)) and not(eo:void(.))]" mode="create"/>
       <xsl:apply-templates select="//o[eo:void(.) and @args]" mode="args"/>
+      <xsl:apply-templates select="//o[@type]" mode="type"/>
```
Thayorns: The candidate list exists twice, here and again in the second entry point, and this change takes each copy from two lines to four. Nothing makes the two agree, so a fifth source added to only one of them means a file with metas and a file without get different probes, and the packs would have to happen to cover both shapes to notice.

Thayorns (commented):
`@type` and `@atom` really are the same kind of forma reference `@args` already collects, the trailing `?` is stripped the way to-eo-tree.xsl strips it, and guarding on a leading Φ keeps the generic type variables out. Updating the existing void-args pack rather than leaving it asserting a now-incomplete list is the honest thing to do.

eo-maven-plugin/src/main/resources/org/eolang/maven/probe/add-probes.xsl
morphqdd: Extracted into a shared named template, eo:dotted-prefixes, called from all four sites (composite base, args, type, atom). Had to relax the param to xs:string? since the composite-base template also matches nodes with no @base at all, and tokenize() on an empty sequence is fine but a required xs:string param isn't - caught this by running the full ProbesTest suite, which failed everywhere until I fixed it.

eo-maven-plugin/src/main/resources/org/eolang/maven/probe/add-probes.xsl
morphqdd: Extracted into a shared named template, eo:candidates, called from both entry points. Now there's one list to add a source to, not two.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (112 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3267](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+761](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6289, opened 2026-08-02
#6214: fold compile-scope XMIR fingerprint into the WPA lint cache key

### LINKED ISSUE #6214: Whole-program-analysis lint cache key omits compile-scope XMIRs it analyses

morphqdd:
Depends on #6292 (merge that first). This PR's diff includes #6292's `LintTarget` extraction only because GitHub can't target a base branch that lives on a fork; once #6292 merges to master, I'll rebase this branch and the diff will shrink to just the actual fix below, under the 200-line PR-size limit. Until then the `pr-size` check on this PR will stay red, only because of that duplicated prep-work diff, not this fix's own size.

`Linting.lintAll`'s whole-program-analysis (WPA) pass reads XMIRs from both the project's own tojos and compile-scope tojos. Only `lintOne` (run over the project's own tojos) writes into the `3-lint` cache directory that the WPA cache key hashes. So a change confined to a compile-scope dependency's XMIR leaves that hashed directory byte-identical between builds, and the cached WPA verdict from before the dependency changed gets reused instead of recomputed. Same class of bug as #6030 (a constant cache key), narrower: here the key does capture the project's own sources, it just misses the compile-scope inputs.

Added `CompileScopeFingerprint`, a small `Supplier<String>` that hashes each compile-scope tojo's identifier plus its XMIR content (sorted by identifier for determinism), hex-encoded (not base64 — base64's standard alphabet can emit `/`, which would silently split the fingerprint into extra path segments when used as a cache directory name; caught this in testing). Folded it into the WPA cache key alongside the plugin version, mirroring the pattern `Parsing`'s `Fingerprint`-based key and `Linting`'s own per-file `skipSourceLints`/`skipExperimentalLints` folding already use.

Folding this fingerprint in pushed `Linting`'s qulice `ClassFanOutComplexity` (already at the 30-max ceiling) over the limit, which is why #6292 (extracting `LintTarget`) exists as prep.

Added a `LintingTest` case constructing `Linting` directly with an in-memory compile-scope tojo, running it twice with the same cache directory but different XMIR content the second time, and asserting the WPA cache produces a different directory entry (not the stale one). Confirmed it fails on the unfixed code (same directory both times) and passes with the fix. Updated `MjLintTest.savesForWholeProgramAnalysisResultsToCache`, whose hardcoded cache-path assertion no longer matches the new key shape, to look for the version-keyed directory that actually contains `wpa.xmir` instead. Full eo-maven-plugin test suite, qulice, and jtcop are clean.

Fixes #6214 (once #6292 is merged and this is rebased on top).

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/CompileScopeFingerprint.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/CompileScopeFingerprint.java
@@ -1,0 +1,60 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.function.Supplier;
+
+/**
+ * Fingerprint of a set of compile-scope tojos, keyed on each one's
+ * identifier and the content of its XMIR.
+ *
+ * <p>Used to fold the compile-scope XMIRs a whole-program-analysis pass
+ * reads into that pass's cache key, so a change confined to a compile-scope
+ * dependency's XMIR invalidates the cached verdict the same way a change in
+ * the project's own sources already does (see {@link Linting}, #6214).</p>
+ *
+ * @since 0.64
+ */
+final class CompileScopeFingerprint implements Supplier<String> {
+
+    /**
+     * The compile-scope tojos to fingerprint.
+     */
+    private final Collection<TjForeign> tojos;
+
+    /**
+     * Ctor.
+     * @param compile Compile-scope tojos
+     */
+    CompileScopeFingerprint(final Collection<TjForeign> compile) {
+        this.tojos = compile;
+    }
+
+    @Override
+    public String get() {
+        try {
+            final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            this.tojos.stream()
+                .sorted(Comparator.comparing(TjForeign::identifier)).forEach(
+                    tojo -> digest.update(
+                        String.format("%s:%s%n", tojo.identifier(), new Sha(tojo.xmir()))
+                            .getBytes(StandardCharsets.UTF_8)
+                    )
+                );
+            final StringBuilder hex = new StringBuilder(64);
+            for (final byte octet : digest.digest()) {
+                hex.append(String.format("%02x", octet));
+            }
+            return hex.toString();
+        } catch (final NoSuchAlgorithmException ex) {
+            throw new IllegalStateException("SHA-256 is not available", ex);
+        }
+    }
+}
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/CompileScopeFingerprint.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/CompileScopeFingerprint.java
@@ -20,12 +20,11 @@ import java.util.function.Supplier;
  * dependency's XMIR invalidates the cached verdict the same way a change in
  * the project's own sources already does (see {@link Linting}, #6214).</p>
  *
- * <p>{@link #get()} joins each entry with a fixed LF character, not the
- * platform line separator, so the fingerprint is the same on every OS
+ * <p>{@link #get()} frames each entry with a zero byte rather than any
+ * textual separator, so the fingerprint carries no platform line separator
  * and a cache built on one machine stays valid on another.</p>
  *
  * @since 0.64
- * @checkstyle ProhibitLineSeparatorInStringsCheck (60 lines)
  */
 final class CompileScopeFingerprint implements Supplier<String> {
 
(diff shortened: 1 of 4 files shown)


Thayorns (commented):
The diagnosis holds up: 3-lint is written only by lintOne, which never sees a compile-scope tojo, so hashing that directory really does leave dependency XMIRs out of the key. Choosing hex over base64 for the segment is the right call for exactly the reason given. Three notes, two on how deterministic the digest actually is.

eo-maven-plugin/src/main/java/org/eolang/maven/CompileScopeFingerprint.java
```diff
@@ -0,0 +1,60 @@
...
+    CompileScopeFingerprint(final Collection<TjForeign> compile) {
+        this.tojos = compile;
+    }
+
+    @Override
+    public String get() {
+        try {
+            final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            this.tojos.stream()
+                .sorted(Comparator.comparing(TjForeign::identifier)).forEach(
```
Thayorns: The sort is here for determinism and `forEach` is the one terminal operation that does not promise to honour it — `forEachOrdered` is the one that does. It happens to run in order for a sequential stream today, so nothing goes red, but the guarantee the sorted() call is reaching for is not the one this line asks for.

eo-maven-plugin/src/main/java/org/eolang/maven/CompileScopeFingerprint.java
```diff
@@ -0,0 +1,60 @@
...
+                    tojo -> digest.update(
+                        String.format("%s:%s%n", tojo.identifier(), new Sha(tojo.xmir()))
```
Thayorns: `%n` is the platform line separator, so this digest is `\n`-delimited on Linux and `\r\n`-delimited on Windows. The same set of dependency XMIRs therefore fingerprints to two different values depending on the machine, which shows up the moment a cache directory is shared or copied between them and every entry misses.

eo-maven-plugin/src/test/java/org/eolang/maven/LintingTest.java
```diff
@@ -55,4 +60,71 @@ void skipsLintingWhenFlagIsSet(@TempDir final Path temp) {
...
+        new Linting(
+            new TjsForeign(),
+            compile,
+            target,
+            cache,
+            true,
+            "0.0.0",
+            Collections.emptyList(),
+            Collections.emptyList(),
+            false,
```
Thayorns, line 104: Five bare booleans in a row, and nothing at the call site says which is skipLinting and which is lintAsPackage. Swap any two of them in Linting's constructor and this still compiles, still passes, and quietly stops testing the package path the test is named for.

eo-maven-plugin/src/main/java/org/eolang/maven/CompileScopeFingerprint.java
morphqdd: Fixed, switched to forEachOrdered.

eo-maven-plugin/src/main/java/org/eolang/maven/CompileScopeFingerprint.java
morphqdd: Fixed, fixed "\n" instead of %n, moved the reasoning into the class Javadoc plus a checkstyle suppression (ProhibitLineSeparatorInStringsCheck otherwise assumes any hardcoded separator is the bug, which is backwards here).

eo-maven-plugin/src/test/java/org/eolang/maven/LintingTest.java
morphqdd: Fixed with named constants, mirroring how MjLint.java already passes named fields instead of literals to the same constructor. Tried local variables and inline comments first but both are rejected by this module's own qulice rules (LocalFinalVariableNameCheck's all-lowercase-12-char limit, and MethodBodyCommentsCheck bans comments in method bodies entirely) - class-level constants were the one idiom qulice accepted.

yegor256:
@morphqdd conflicts here

yegor256:
@morphqdd conflicts here

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+953](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6291, opened 2026-08-02
#6215: report a check-less parser error as a defect instead of crashing

### LINKED ISSUE #6215: `lint` goal crashes with `IllegalArgumentException` on a plain parser syntax error
## What Happens

The `lint` goal crashes with an internal `IllegalArgumentException` instead of reporting the problem normally, whenever the source file being linted has a plain parser syntax error.

`Linting.toDefect` requires every `<error>` element already present in the XMIR to carry a `check` attribute:

```java
error.attribute("check").text().orElseThrow(
    () -> new IllegalArgumentException("The <error> element in XMIR must contain 'check' attribute")
);
```

Errors added by the parse-stage XSLs (for example `validate-object-name`, `build-fqns`) do carry `check`. The hand-written parser's own syntax errors do not: `Emit.error` (in `eo-parser`) writes only `line`, `pos` and `severity`, no `check`. `MjParse` does not fail the build on a parse error, so an XMIR with this kind of unchecked error always reaches `MjLint`.

Confirmed end to end with a minimal source:

```
# Sample object for the probe.

[] > main
  "\uD800" > @
```

Parsing writes an `<error line="4" pos="2" severity="error">` for the invalid unicode escape, with no `check` attribute. Running `MjLint` on this XMIR throws `IllegalArgumentException: The <error> element in XMIR must contain 'check' attribute` instead of reporting the syntax error to the user.

As a contrast, a file whose only error carries `check="validate-object-name"` lints normally and reports a clean summary, confirming the `check` attribute is exactly what gates this crash.

## What Should Happen

`toDefect` should treat a `<error>` with no `check` attribute as a defect from the parser itself (using some fixed identifier such as `"parser"` or the error's own severity), and report it like any other lint defect, instead of throwing an internal exception that surfaces as a build failure with no clear message to the user.

morphqdd:
`Linting.toDefect` required every `<error>` element in the XMIR to carry a `check` attribute, throwing an internal `IllegalArgumentException` otherwise. Errors added by the parse-stage XSLs (`validate-object-name`, `build-fqns`, etc.) do carry `check`; the hand-written parser's own syntax errors do not (`Emit.error` only writes `line`, `pos`, `severity`). Since `MjParse` doesn't fail the build on a parse error, an XMIR with a plain syntax error always reaches `MjLint`, which then crashed instead of reporting the problem normally.

Changed the missing-`check` case to fall back to the fixed rule name `"parser"` instead of throwing, so a parser syntax error gets reported as an ordinary defect (the build can still legitimately fail afterward, same as any other reported error — that part of the pipeline is unaffected).

Added a test reproducing the issue's exact example (an invalid unicode escape in a string literal), asserting the build still fails (a real syntax error should) but that no `IllegalArgumentException` shows up anywhere in the failure's cause chain. Confirmed it fails on the unfixed code (the crash is present in the chain) and passes with the fix. Full eo-maven-plugin test suite, qulice, and jtcop are clean.

Closes #6215


eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -346,6 +348,27 @@ void failsParsingOnError(@Mktmp final Path temp) throws Exception {
...
+    @Test
+    void doesNotCrashOnPlainParserSyntaxError(@Mktmp final Path temp) throws IOException {
+        final FakeMaven maven = new FakeMaven(temp).withProgram(
+            "# Sample object for the probe.",
+            "",
+            "[] > main",
+            "  \"\\uD800\" > @"
+        );
+        MatcherAssert.assertThat(
+            "an <error> with no 'check' attribute must be reported as a defect, not crash with an IllegalArgumentException",
```
Thayorns: The message says the error must be reported as a defect, but nothing below looks for a defect. The assertion only says no IllegalArgumentException appeared, which would still hold if toDefect dropped the error on the floor or named it the empty string. It is also broader than intended, since NumberFormatException and the rest of that family are IllegalArgumentExceptions too, and `Matchers.is(true)` prints "was <false>" without naming which cause tripped it.

Thayorns (commented):
The split is real: Emit.error writes line, pos and severity and no check, while the parse-stage XSLs all name theirs, so any file with a plain syntax error reaches lintOne and dies on an attribute that was never going to be there. Falling back to a fixed rule name puts it back on the ordinary reporting path, and the surrogate in the fixture is a good irregular input for it.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
morphqdd: Reworked the assertion. Turns out the check-less parser error never gets a `check` attribute written back into the linted XMIR (only lint-stage-found defects get re-embedded), so I couldn't assert on that directly. Instead I now capture the thrown exception's root cause and assert its message contains `(parser)`, same idiom as `includesDefectDetailsInExceptionMessage` above in this file. Verified by reverting the fix locally: the test fails with a precise diagnostic instead of silently passing.

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -528,16 +533,18 @@ private static Collection<Defect> existing(final Xnav xnav) {
...
      * @return Defect
      */
     private static Defect toDefect(final Xnav error) {
         return new Defect.Default(
-            error.attribute("check").text().orElseThrow(
-                () -> new IllegalArgumentException(
-                    "The <error> element in XMIR must contain 'check' attribute"
-                )
-            ),
+            error.attribute(Linting.CHECK).text().orElse("parser"),
```
yegor256: @morphqdd the code you deleted looks more "fail fast" to me. If `check` is not set, the Maven plugin must crash. Then, we should set the `"parser"` value to this attribute in the parser.

morphqdd: Done. Emit.error() now always writes check="parser" on the error element, and Linting.toDefect throws IllegalArgumentException again when check is missing, same as before this PR. Verified by removing the attribute locally: the test then fails with the exact fail-fast exception instead of silently passing.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3529](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+833](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2553](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6292, opened 2026-08-02
#6214: extract LintTarget from Linting.lintOne (prep, no behavior change)

morphqdd:
Pure refactor, no behavior change. Extracted the `Place`/`OnDetailed`/`OnDefault`/`Xnav` chain `Linting.lintOne` uses to compute a per-file lint result's path into its own `LintTarget` class.

This is prep work for #6214 (whole-program-analysis lint cache key), which needs to add one more type reference to `Linting` (a fingerprint helper for compile-scope XMIRs) but `Linting`'s qulice `ClassFanOutComplexity` is already at the 30-max ceiling, so anything added there needs an equivalent reduction elsewhere. `Place`, `OnDetailed`, `OnDefault`, and `MjAssemble` were each referenced in exactly one spot in `Linting.java` (this one), so extracting them into their own class removes 4 fan-out units from `Linting` while adding only 1 (`LintTarget`).

Splitting this out as its own PR to keep the actual #6214 fix under the 200-line PR-size gate; the fix PR is opened against this branch as its base.

No new tests: existing `LintingTest`/`MjLintTest` coverage already exercises `lintOne`'s target-path computation end to end, and this change doesn't alter its behavior, only where the logic lives.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/LintTarget.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/LintTarget.java
@@ -1,0 +1,51 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import com.github.lombrozo.xnav.Xnav;
+import com.jcabi.xml.XML;
+import java.nio.file.Path;
+import org.eolang.parser.OnDefault;
+import org.eolang.parser.OnDetailed;
+
+/**
+ * The path a per-file lint result belongs at, derived from the XMIR
+ * document's own object name.
+ * @since 0.64
+ */
+final class LintTarget {
+
+    /**
+     * The XMIR document being linted.
+     */
+    private final XML xmir;
+
+    /**
+     * The document's own source path (for a detailed error location if the
+     * object name can't be read).
+     */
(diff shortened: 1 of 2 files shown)


morphqdd:
@yegor256, please take a look when you have a moment. This one is small and #6289 depends on it, so merging this first keeps that PR's diff under the size limit.

Thayorns (approved):
The moved expression is character-for-character what lintOne had, so there is nothing for behaviour to change into, and the new class keeps to two final fields with a constructor that only assigns them. LintTarget has no case of its own, which leaves its path computation held up by the same end-to-end coverage in MjLintTest that was holding it before the move, and that coverage would not distinguish a wrong base directory from a wrong object name.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+641](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3283](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6296, opened 2026-08-02
#6141: fix drive-relative win32 path incorrectly classified as absolute

### LINKED ISSUE #6141: `path.win32.is-absolute` mistakenly categorizes drive-relative paths as absolute in Windows

morphqdd:
`path.win32.is-absolute` was `is-root-relative.or (drive.size.gt 0)`. Since `win32.drive` returns a non-empty drive letter for both `C:foo` and `C:\foo`, the `.or` clause alone forced `is-absolute` to `true` for any drive-qualified path, even a drive-relative one like `C:foo` that has no root separator after the drive.

`is-root-relative` already contains the exact distinction needed (whether the part after the drive starts with the separator), so folding the drive-presence check into `is-absolute` was redundant and wrong for the drive-relative case. Made `is-absolute` equal to what `is-root-relative` computed directly (dropping the extra `.or` condition), and removed the now-redundant `is-root-relative` name entirely (qulice's `anemic-getter` lint flagged a bare rename as dead weight), updating the two internal call sites that referenced it (`normalized`'s separator and `..`-collapsing logic) to use `is-absolute` instead, since after this fix they are the exact same value.

Added the issue's own regression case, `path.win32 "C:foo"` reporting `is-absolute` as `false`, alongside the existing `can-detect-an-absolute-win32-path` test that already covers the rooted (`C:\Windows\Users`, `\Windows\Users`) and plain-relative (`temp\var`) cases. Confirmed the new test fails on the unfixed code (`is-absolute` incorrectly `true`) and passes with the fix. Full eo-runtime test suite is clean.

Closes #6141

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -78,9 +78,7 @@
                 number index
       string > text
         basename >> base!
-    is-root-relative.or > is-absolute
-      drive.size.gt 0
-    and. > is-root-relative
+    and. > is-absolute
       driveless.size.gt 0
       eq.
         driveless.slice 0 1
@@ -106,7 +104,7 @@
           concat.
             concat.
               drive.concat
-                is-root-relative.if separator --
+                is-absolute.if separator --
               string.joined >> body!
                 separator
                 tuple.reduced
@@ -119,7 +117,7 @@
                         accum.length.gt 0
                         not. (accum.head.eq "..")
                       accum.tail
-                      is-root-relative.not.if (accum.with segment) accum
+                      is-absolute.not.if (accum.with segment) accum
                     if.
                       or.
                         segment.eq "."
@@ -196,6 +194,10 @@
           string (string.replaced pth (regex "/\\//") separator)!
         string ubts > pth
       ^.separator > separator
+
+  not. ++> can-detect-a-drive-relative-win32-path-as-not-absolute
+    is-absolute.
+      path.win32 "C:foo"
 
   and. ++> can-detect-an-absolute-posix-path
     is-absolute.


eo-runtime/src/main/eo/path.eo
```diff
@@ -106,7 +104,7 @@
           concat.
             concat.
               drive.concat
-                is-root-relative.if separator --
+                is-absolute.if separator --
```
Thayorns, line 107: These two sites are asking whether the path starts at a root, not whether it is absolute; they happen to be the same value today and the merge makes that coincidence load-bearing. On win32 `\Windows\Users` has no drive and is drive-relative in Windows terms, so if is-absolute is ever tightened to require one — which is the direction this issue points — `normalized` quietly stops emitting the leading separator here and starts keeping leading `..` segments on line 120, in a method nobody meant to touch.

Thayorns (commented):
The diagnosis is right: win32.drive answers the same for `C:foo` and `C:\foo`, so the `.or` clause could never tell a drive-relative path from a rooted one, and dropping it leaves exactly the test that can. `normalized` is untouched in effect, since is-root-relative was already false for `C:foo`, and the two existing win32 cases still hold `C:\Windows\Users` and `\Windows\Users` down.

eo-runtime/src/main/eo/path.eo
morphqdd: Checked, and it's not actually a coincidence riding on today's implementation. `can-detect-an-absolute-win32-path` (lines 209-217, untouched by this PR) already asserts `path.win32 "\Windows\Users"` (no drive) is absolute. So "starts at a root, drive optional" is the tested, intended contract for is-absolute here, not a side effect that happens to match is-root-relative right now.

If a future change wanted a drive-required definition, it would have to touch that existing test too, at which point whoever does it runs straight into the normalized call sites and has to reckon with them consciously. Given that, I'd rather not add a same-value duplicate predicate for a change that isn't actually planned. Happy to revisit if #6141 or a follow-up issue is explicit about wanting drive-required semantics.

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3295](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (12) hits-of-code; +1 to give you at least something.  Your running score is [+765](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6297, opened 2026-08-02
#6138: wrap InvalidPathException in ExFailure for directory.walk

### LINKED ISSUE #6138: `directory.walk` throws `InvalidPathException` for malformed `file.path` rather than `ExFailure`

morphqdd:
`EOdirectory$EOwalk.lambda()` built the host `Path` via `Paths.get(...).toAbsolutePath()` before either of its two existing error-handling blocks (malformed glob, `Files.walk` I/O error). An invalid `file.path` value (for example one containing a NUL character) throws `InvalidPathException` at that conversion, uncaught, leaking a raw Java exception instead of the `ExFailure` every other invalid-input case in this atom produces.

Wrapped the `Paths.get(...).toAbsolutePath()` call in a try/catch for `InvalidPathException`, reporting it as `ExFailure` with the original exception preserved as the cause, consistent with the existing glob and I/O error handling in the same method.

Added a test constructing a `file.path` with an embedded NUL character (via `(char) 0`, not a source-level `` escape or string concatenation, both of which qulice's checkstyle rules reject), confirming `directory.walk` now raises `ExFailure` instead of the raw `InvalidPathException`. Confirmed the test fails on the unfixed code (`InvalidPathException` propagates) and passes with the fix.

Closes #6138


gemshrine (approved):
The path conversion is now enclosed in the narrow `InvalidPathException` boundary, while the existing glob and I/O error paths remain unchanged. The regression test creates the malformed NUL path without any filesystem dependency and verifies the EO-facing `ExFailure`. No inline change is needed. The remaining functional checks are still running.

0crat:
@gemshrine Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+1351](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:
@morphqdd can you please add EO test too?

morphqdd:
@yegor256 added `stops-on-walking-a-path-with-a-nul-character` to `directory.eo`.

One thing worth saying: `-->` transpiles to `assertThrows(Exception.class, ...)`, so this test passes with or without the Java fix. Before the fix the raw `InvalidPathException` was also an exception. The only observable difference is the message, and there is no `try` object in the runtime yet to assert on it from EO. So the EO test pins the contract (walking an invalid path must stop, not return garbage), and the Java test is what pins the exception type.

yegor256:
@morphqdd conflict here

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3545](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6311, opened 2026-08-02
Fix atom detection when generating Java files

### LINKED ISSUE #6247: `Transpiling` atom check incorrectly always results in `false`, dead-end skipping branches
`Transpiling` in `eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java:263` decides whether an XMIR describes an atom like this:

```java
final Xnav object = new Xnav(target).element("object");
final boolean atom = object.path("/object/o/o[@name='λ']").findAny().isPresent();
```

`object` is already the `<object>` element, and `Xnav.path` evaluates relative to the node it is called on, not from the document root. So the expression looks for an `<object>` child inside `<object>`, and no XMIR has one. `atom` is therefore always `false`.

Three lines down that value guards which classes get written out:

```java
if (!atom || jname.endsWith("Test")) {
```

With `atom` pinned to `false` the condition is always true, so every class is emitted and the branch that was meant to hold back an atom's non-test classes never runs.

I confirmed it by feeding two documents to the same call. With `<object><o><o name='λ'/></o>…</object>`, the shape real XMIR has, `path("/object/o/o[@name='λ']")` finds nothing while the relative `path("o/o[@name='λ']")` finds the element. Only the artificial `<object><object><o><o name='λ'/></o></object>…</object>` makes the current expression match.

Dropping the `/object` prefix makes the check do what it reads as. That is a behavior change — atoms' non-test classes would stop being generated — so it wants its own PR and a look at what depends on those files today.

Found while reviewing #6167, which moves this code into a new `JavaFiles` class unchanged. The bug predates that PR.

@yegor256 please take a look

gemshrine:
## What changed

`JavaFiles` now evaluates the atom marker relative to the XMIR `<object>` element it already selected. This makes the `o/o[@name='λ']` check match the shape produced by the parser instead of looking for a second nested `<object>`.

## Why

The previous absolute-looking expression, `/object/o/o[@name='λ']`, was evaluated by `Xnav` from the current `<object>` node. It therefore never detected an ordinary atom, leaving `atom` false and causing production Java classes to be written for atoms.

## Verification

`JavaFilesTest` now creates an atom XMIR with both an implementation class and a test class. It verifies that only the `*Test` class is generated and that the implementation class is absent.

Tested with:

```text
MAVEN_OPTS='-XX:ActiveProcessorCount=1 -Xmx1g' mvn -q -ntp -T 1 -pl eo-maven-plugin test -Dtest=JavaFilesTest -DfailIfNoTests=false -Djunit.jupiter.execution.parallel.enabled=false -DforkCount=1
```

Fixes #6247

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
@@ -95,7 +95,7 @@ final class JavaFiles {
             final Xnav object = new Xnav(target).element("object");
             final Collection<Xnav> classes = object.elements(Filter.withName("class"))
                 .collect(Collectors.toList());
-            final boolean atom = object.path("/object/o/o[@name='λ']").findAny().isPresent();
+            final boolean atom = object.path("o/o[@name='λ']").findAny().isPresent();
             for (final Xnav clazz : classes) {
                 final String jname = clazz.attribute("java-name").text().get();
                 if (!atom || jname.endsWith("Test")) {
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
@@ -7,6 +7,7 @@ package org.eolang.maven;
 import com.yegor256.Mktmp;
 import com.yegor256.MktmpResolver;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Random;
@@ -75,5 +76,57 @@ final class JavaFilesTest {
             ),
             Matchers.equalTo(Arrays.asList(false, true))
         );
+    }
+
+    @Test
+    void writesExactlyOneFileForAtom(@Mktmp final Path temp) throws IOException {
+        MatcherAssert.assertThat(
+            "only a test class of an atom must be written",
+            JavaFilesTest.generateAtom(temp),
+            Matchers.equalTo(1)
+        );
+    }
+
+    @Test
+    void skipsAtomImplementationClass(@Mktmp final Path temp) throws IOException {
+        JavaFilesTest.generateAtom(temp);
+        MatcherAssert.assertThat(
+            "an atom implementation class must not be written",
+            Files.exists(temp.resolve("generated/EOatom.java")),
+            Matchers.equalTo(false)
+        );
+    }
+
+    @Test
+    void writesCompanionClassForAtom(@Mktmp final Path temp) throws IOException {
+        JavaFilesTest.generateAtom(temp);
+        MatcherAssert.assertThat(
+            "an atom test class must be written",
+            Files.exists(temp.resolve("generated/EOatomTest.java")),
+            Matchers.equalTo(true)
+        );
+    }
+
+    /**
+     * Transpile a single-atom XMIR and report how many files were written.
+     * @param temp Temp directory to write into
+     * @return Number of files written
+     * @throws IOException If fails to save or transpile
+     */
+    private static int generateAtom(final Path temp) throws IOException {
+        final Path xmir = temp.resolve("main.xmir");
+        new Saved(
+            String.join(
+                "",
+                "<object><o><o name='λ'/></o>",
+                "<class java-name='EOatom'><java>class EOatom {}</java></class>",
+                "<class java-name='EOatomTest'><java>class EOatomTest {}</java></class>",
+                "</object>"
+            ),
+            xmir
+        ).value();
+        return new JavaFiles(
+            temp.resolve("generated"), temp.resolve("cache").resolve("1.0-SNAPSHOT"), false
+        ).total(true, xmir, "", false);
     }
 }


eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
```diff
@@ -45,4 +46,36 @@ void countsJavaFilesItWrites(@Mktmp final Path temp) throws IOException {
...
+            xmir
+        ).value();
+        MatcherAssert.assertThat(
+            "only a test class of an atom must be written",
+            new JavaFiles(
+                temp.resolve("generated"), temp.resolve("cache").resolve("1.0-SNAPSHOT"), false
+            ).total(true, xmir, "", false),
+            Matchers.equalTo(1)
+        );
+        MatcherAssert.assertThat(
```
Thayorns, line 93: Three behaviors are being verified in one method here: the count on line 68, the absence of the implementation file, and the presence of the test file. If the count assertion goes red the other two never run, so a failure tells you less than it could, and the name promises only the first of them. Each of these stands on its own as a test.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
```diff
@@ -45,4 +46,36 @@ void countsJavaFilesItWrites(@Mktmp final Path temp) throws IOException {
...
     }
+
+    @Test
+    void writesOnlyTestClassForAtom(@Mktmp final Path temp) throws IOException {
+        final Path xmir = temp.resolve("main.xmir");
+        new Saved(
+            String.join(
+                "",
+                "<object><o><o name='λ'/></o>",
+                "<class java-name='EOatom'><java>class EOatom {}</java></class>",
```
Thayorns: The java names are short plain ASCII with no package, while countsJavaFilesItWrites just above builds `bär%d.Ωne%d` from a random seed and carries that seed into the failure message. Place.make replaces every dot with File.separator, so a package-qualified or non-ASCII atom name lands in a nested directory that this test never reaches, and the hard-coded `generated/EOatom.java` on line 72 only holds for the flat case.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
```diff
@@ -45,4 +46,36 @@ void countsJavaFilesItWrites(@Mktmp final Path temp) throws IOException {
...
+                "<class java-name='EOatomTest'><java>class EOatomTest {}</java></class>",
+                "</object>"
+            ),
+            xmir
+        ).value();
+        MatcherAssert.assertThat(
+            "only a test class of an atom must be written",
```
Thayorns, line 84: This message, and the two below it, state the requirement rather than the failure: "only a test class of an atom must be written". The test right above says "the Java files written are not as many as the classes in the XMIR", which is what a reader of a red build wants to see. Same for lines 71 and 76.

Thayorns (commented):
The one-line change lines up with what the parser actually emits: Emit.atomMarker writes `<o name='λ' atom='...'/>` as a child of the formation's `<o>`, so `o/o[@name='λ']` evaluated from the already-selected `<object>` node is the right shape, and Xpath.nodes() starting the stream at the current node explains why the old absolute-looking expression could never match. The production side reads well. The new test is where I'd look again: it bundles three checks into one method and its inputs are much tamer than those of the test right above it.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -81,7 +81,7 @@ int total(
             final Xnav object = new Xnav(target).element("object");
             final Collection<Xnav> classes = object.elements(Filter.withName("class"))
                 .collect(Collectors.toList());
-            final boolean atom = object.path("/object/o/o[@name='λ']").findAny().isPresent();
+            final boolean atom = object.path("o/o[@name='λ']").findAny().isPresent();
```
morphqdd, line 98: This is a whole-file `atom` flag applied to every `class` in `classes`, but `classes.xsl` already stamps `skip-java="true"` per class based on the same underlying fact, at the correct per-class granularity (see `classes.xsl:58`, consumed by `to-java.xsl:230`). If a file contains multiple objects and only one is an atom, this XPath will misclassify the non-atom siblings too, since it just checks whether *any* `λ` exists anywhere under `o`. Reading `clazz.attribute("skip-java")` per class instead would avoid re-deriving a fact the XSL pipeline already computes correctly, and avoid the drift risk of two independent atom-detection implementations.

yegor256:

gemshrine:

gemshrine:
@yegor256 re-resolved conflicts after the latest merges, rebased on current master.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+1164](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+3891](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2203](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6312, opened 2026-08-02
Make PhNest refuse a package that carries no XmirPackage annotation

### LINKED ISSUE #6245: `@XmirPackage` annotation in generated `package-info.java` files is not being used
`PackageInfos` in `eo-maven-plugin/src/main/java/org/eolang/maven/PackageInfos.java:130` writes `@org.eolang.XmirPackage("...")` into every generated `package-info.java`, and the annotation itself lives in `eo-runtime/src/main/java/org/eolang/XmirPackage.java`. Nothing reads it. A grep across the repository finds exactly two mentions: the one that writes it and the declaration.

Its retention is `CLASS`, so it survives into the bytecode but not into the JVM at runtime. That is the right retention for a build-time tool to read, and no build-time tool reads it. The value it carries is the EO package name with the `org.eolang` base stripped, which is information nothing else in the class file preserves.

Either something should consume it or it should go. The obvious consumer is whatever needs to map a compiled Java package back to the EO package it came from, and the obvious place to look is the linter and the tests that walk generated sources. Until one exists, the annotation is metadata nobody asked for, and the next person to find it has no way to tell whether it is half-wired on purpose or left over.

Came up while reviewing #6204, which turned the annotation back on after it had been commented out. The PR is right to enable it; this is about what happens next.

@yegor256 please take a look

Thayorns:
Nothing read the `@XmirPackage` annotation that `eo-maven-plugin` writes into every generated `package-info.java`, so `PhNest` now does. When a package object is asked for an extension, it looks up the `package-info` of the Java package it stands for and refuses to hand anything out unless that package says which EO package it came from. A Java package without the annotation was never transpiled, so its classes are not EO objects and an extension taken from one would be an arbitrary Java class.

The annotation was `RetentionPolicy.CLASS`, which puts it in the bytecode but hides it from reflection, so it is `RUNTIME` now. Only the package the object stands for is attested, never the package holding the object itself: that one is `org.eolang`, the runtime root, and it comes from no EO package.

Two hand-written `package-info.java` files for `org.eolang.EO_string` gained the annotation, in `src/main/java` and in `src/test/java`. The generated one is skipped for that package because a human wrote it, and the test-scope file shadows the main one on the surefire classpath, so without both `string.printf` and its neighbours would trip the new check.

Separately, the root package was named after its Java package rather than its EO one. `BASE` was `org.eolang.` with a mandatory trailing dot, so the bare `org.eolang` never matched and came out as `@XmirPackage("org.eolang")` instead of `@XmirPackage("")`. The dots are escaped now and the trailing one is optional.

This branch sits on top of #6204, which is what turns the annotation on in the first place, so the diff here carries that pull request's two commits until it merges. Merge #6204 first and this one reduces to the four files above.

Closes #6245


Thayorns:
@yegor256 the annotation had `CLASS` retention, so nothing could read it at runtime even if it wanted to; it is `RUNTIME` now and `PhNest` refuses any package that does not carry it. Happy to walk through the `org.eolang.EO_string` part, where two hand-written `package-info.java` files had to gain the annotation by hand.

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (205 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+653](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6125, opened 2026-07-30
`Transpiling` constructor takes 11 positional parameters, several of the same type


morphqdd:
**What happens.** The `Transpiling` constructor takes 11 parameters and the class holds 12 attributes, while our rule is 1-4. It carries two suppressions to stay green: `@checkstyle ParameterNumberCheck` and `@SuppressWarnings("PMD.ExcessiveParameterList")`.

The list is positional, and several slots have the same type:

```java
Transpiling(
    final Collection<TjForeign> srcs,
    final Path target,
    final Path generated,
    final Path cache,
    final boolean enabled,
    final String ver,
    final boolean tests,
    final Path measures,
    final Tracking diagnostics,
    final boolean cvrg,
    final String base
)
```

Slots 6 and 11 are both `String`, slots 2, 3, 4 and 8 are all `Path`, and slots 5, 7 and 10 are all `boolean`. Two of them can be swapped at the call site and nothing catches it: not the compiler, not qulice. The version string and the superclass name would still build, and the transpiler would write the wrong text into every generated file.

**What should happen.** The fix is already in that same parameter list. `Tracking` exists because two related flags were bundled into an object instead of being passed separately. The same treatment fits the values that describe what the transpiler writes into the generated Java: `cvrg` (whether objects get wrapped into `PhCoverage`) and `base` (the class a generated class extends). Both are read only by `to-java.xsl` through `compiled()` and both are folded into the cache key in `version()`.

Bundling those two would take the constructor from 11 parameters to 10 and put two same-typed neighbours behind a named type. Whether the rest of the list can be grouped further is worth looking at in the same pass, since the four `Path` arguments also describe one thing each about where output goes.

Found while reviewing objectionary/eo#6111, which added the 11th parameter. The problem predates that PR, which is why it is filed here instead of being fixed there.

yegor256:
@morphqdd I agree that the problem exists, but I don't like the solution offered. Just lowering from 11 to 10 is not a solution at all. We either drop to 2-3, or keep this class as is.

morphqdd:
@yegor256, I’ll look into this matter and try to bring it down to 2-3

morphqdd:
@yegor256
The fields fall into three groups. Two of them are already visible in `version()`, which mixes the plugin version with `locations`, `coverage` and `superclass`, because each of those three changes the text of the generated Java.

- caching: `cacheDir`, `cacheEnabled`, `version`
[...]

PR #6315, opened 2026-08-03
#6125: extract GlobalCache out of Parsing

morphqdd:
A sketch, as asked. One class extracted, not the four.

`Parsing` and `Transpiling` hold the same four fields: the cache directory, the enabled flag, the version and the `ConcurrentCache` guard. In `Parsing` those four are used in one block and nowhere else, so that is where the cut is cleanest. `Transpiling` comes next, in another PR, together with `JavaFiles`, which keeps a third copy of the same two ideas.

`GlobalCache` answers one question: how is this one file to be written. It hands back a `Footprint`, which is the interface this module already has for that. `GcShared` goes through the machine-wide cache; `GlobalCache.NONE` does not. The `eo.cacheEnabled` option is read once, in `MjSafe`, so no step below knows the option exists.

`Parsing` loses four fields and two constructor parameters, and the `if` over the flag goes with them. The lambda that turns a source into XMIR was written twice before, once inside the cache and once in the `else`; now it is written once.

Numbers on this branch: qulice green, `MjParseTest` 13 of 13 and untouched, suppressions across the repo down from 369 to 368.

One thing I want your opinion on. `MjSafe` sat at exactly 30 referenced types, which is the `ClassFanOutComplexityCheck` maximum, so naming `GlobalCache` and `GcShared` there pushed it to 32. I tried moving the parsing step out of `MjSafe` instead, which keeps the count flat, but that spread the construction across `MjParse`, `MjAssemble` and `MjCompile` and added about 120 lines to a change that is supposed to be small. So the suppression is there for now, with a puzzle next to it. It matters beyond style: while the ceiling stands, `Transpiling` cannot move to `GlobalCache` either, since it is built in the same class.

Not closing the issue with this PR.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/GlobalCache.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/GlobalCache.java
@@ -1,0 +1,68 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import java.nio.file.Path;
+import java.util.function.Supplier;
+import org.cactoos.Func;
+import org.cactoos.io.InputOf;
+
+/**
+ * The cache shared by every build on this machine.
+ *
+ * <p>A step asks it how one file is to be written and gets a {@link Footprint}
+ * back. Whether that footprint goes through the cache or straight past it is
+ * decided once, where the {@code eo.cacheEnabled} option is read, so no step
+ * has to know that the option exists.</p>
+ *
+ * @since 0.74
+ */
+interface GlobalCache {
+
+    /**
+     * The cache that remembers nothing, for a build that turned caching off.
+     */
+    GlobalCache NONE = new GlobalCache.GcFresh();
+
+    /**
+     * How one file is to be written.
+     * @param tail Path of that file inside the cache
+     * @param hash Hash segment of the cache path
+     * @param compile How the content of the file is produced
+     * @return The footprint that writes it
+     */
+    Footprint footprint(Path tail, Supplier<String> hash, Func<Path, String> compile);
+
+    /**
+     * The same cache, with one more segment folded into its key.
+     * @param segment The segment to fold in
+     * @return The derived cache
+     */
+    GlobalCache with(String segment);
+
+    /**
+     * The cache that remembers nothing.
+     *
+     * <p>Every file is produced and written to its target, and the cache
+     * directory is neither read nor written. This is what the build gets when
+     * {@code eo.cacheEnabled} is false.</p>
+     *
+     * @since 0.74
+     */
+    final class GcFresh implements GlobalCache {
+
+        @Override
+        public Footprint footprint(
+            final Path tail, final Supplier<String> hash, final Func<Path, String> compile
+        ) {
+            return new FpGenerated(src -> new InputOf(compile.apply(src)));
+        }
+
+        @Override
+        public GlobalCache with(final String segment) {
+            return this;
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


eo-maven-plugin/src/main/java/org/eolang/maven/GlobalCache.java
```diff
@@ -0,0 +1,68 @@
...
+ * has to know that the option exists.</p>
+ *
+ * @since 0.74
+ */
+interface GlobalCache {
+
+    /**
+     * The cache that remembers nothing, for a build that turned caching off.
+     */
+    GlobalCache NONE = new GlobalCache.GcFresh();
```
Thayorns: A field on an interface is public and static whether or not the keywords are written, so NONE is a global constant. Every caller that wants a cache-free run reaches for the same shared instance by name instead of being handed one, and CompilingTest already shows the other path by writing new GlobalCache.GcFresh() where it needs one.

Thayorns (commented):
The cut is in a good place. Parsing drops four fields and the branch over the flag, the lambda that turns a source into XMIR now exists once instead of twice, and the cache key comes out identical: version, then the fingerprint segment, then the digest, which is what the old String.format built. Two smaller things. The failure messages in both new tests read as requirements rather than as claims about what went wrong, "the second run must be written from the cache" on GcSharedTest.java:41 and "nothing must be remembered between two runs" on GlobalCacheTest.java:40, so when either fires the log restates the rule instead of naming the failure. And with() is the other half of the new interface, yet nothing asserts that folding a segment in actually moves the file somewhere else, so a with() that returned this on GcShared would pass both tests.

eo-maven-plugin/src/main/java/org/eolang/maven/GlobalCache.java
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * The cache that remembers nothing.
+     *
+     * <p>Every file is produced and written to its target, and the cache
+     * directory is neither read nor written. This is what the build gets when
+     * {@code eo.cacheEnabled} is false.</p>
+     *
+     * @since 0.74
+     */
+    final class GcFresh implements GlobalCache {
```
Thayorns, line 49: GcFresh holds nothing. Both of its methods ignore every argument they are given and one of them returns this, so there is no state for the object to be about — it is a pair of functions wearing a class. That is also what makes NONE possible above; give it something to hold and the constant stops being tempting.

eo-maven-plugin/src/main/java/org/eolang/maven/GlobalCache.java
morphqdd: Dropped it. `MjSafe.caching` now builds `new GlobalCache.GcFresh()` the same way `CompilingTest` does, so there is no shared instance to reach for.

eo-maven-plugin/src/main/java/org/eolang/maven/GlobalCache.java
morphqdd: A build with caching off has nothing to remember, so there is nothing for it to hold. Giving it a field would mean state nobody reads.

The `NONE` constant is gone now, which was the part that made it act like a global.

yegor256:
@morphqdd fix the `thresholds` CI too, it's important. The `pr-size` may be ignored, but also would be great to fix

morphqdd:
@volodya-lombrozo, take a look please

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (375 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3345](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+935](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6317, opened 2026-08-03
#6301: split win32 drive by character length, not byte size

### LINKED ISSUE #6301: `path.win32` cannot process drive-qualified paths containing non-ASCII characters
## What happens

The Windows path implementation splits a drive-qualified URI with a character-based `string.slice`, but calculates the slice length with byte sizes:

```eo
uri.slice
  drive.size
  uri.size.minus drive.size
```

For `path.win32 "C:\\Ж"`, `drive.size` is `2`, while `uri.size` is `5` because `Ж` occupies two UTF-8 bytes. The call consequently asks `string.slice` for three characters starting at character index two. The path contains only four characters, so the requested range is outside the text and the calculation terminates.

The same object otherwise documents paths as text, and `string.slice` explicitly counts UTF-8 characters rather than bytes. ASCII-only Windows paths conceal the mismatch.

## Expected behavior

Drive splitting must use one unit throughout: either character lengths (`uri.length` and `drive.length`) with `string.slice`, or a byte-only implementation. A path such as `C:\\Ж` should remain usable by `normalized`, `basename`, `dirname`, `resolved`, and other `path.win32` operations just as `C:\\A` is.

## Suggested coverage

Add a regression test for a drive-qualified path with a multibyte segment, for example asserting that:

```eo
path.win32 "C:\\Ж".normalized
```

produces `C:\\Ж` rather than terminating.

gemshrine:
Fixes #6301.

`path.win32` split the drive off the URI with a character-indexed `string.slice`, but computed both slice arguments from `.size` (the UTF-8 **byte** length). For an ASCII path the two units coincide, so the bug stayed hidden. As soon as the path carried a multibyte segment the byte length exceeded the character length and the slice ran past the end of the string:

```
path.win32 "C:\Ж".driveless    -> ExFailure: Start index + length to slice (5) is out of string bounds
path.win32 "C:\Ж".normalized   -> same failure
```

(`"C:\Ж"` is 5 bytes but 4 characters, so `uri.size.minus drive.size` asked `slice` for 3 characters starting at index 2 of a 4-character string.)

The fix makes the drive split use character units throughout, matching `string.slice`:

```eo
uri.slice
  drive.length
  uri.length.minus drive.length
```

Added a regression test `can-normalize-a-drive-qualified-win32-path-with-a-multibyte-segment` asserting `path.win32 "C:\Ж".normalized` yields `C:\Ж`. All 52 `EOpathTest` cases pass locally, including the existing ASCII and posix ones.

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -58,8 +58,8 @@
       drive.size.eq 0
       uri
       uri.slice
-        drive.size
-        uri.size.minus drive.size
+        drive.length
+        uri.length.minus drive.length
     [] > extname
       if. > @
         or.
@@ -225,6 +225,11 @@
     normalized.
       path.win32 "C:"
     "C:"
+
+  eq. ++> can-normalize-a-drive-qualified-win32-path-with-a-multibyte-segment
+    normalized.
+      path.win32 "C:\\Ж"
+    "C:\\Ж"
 
   eq. ++> can-normalize-a-drive-relative-win32-path-keeping-a-leading-dotdot
     normalized.


Thayorns (approved):
The unit mismatch is real and the two changed lines land it in the right place: drive.length and uri.length now speak the same character units that string.slice counts in, so the drive split stops walking past the end of a multibyte path. The regression test does discriminate, since a drive that failed to strip would yield C:C:\Ж rather than C:\Ж, and the old byte arithmetic terminates outright. One thing worth weighing: issue #6301 asks that C:\Ж stay usable through basename, dirname and resolved as well as normalized, and only normalized is asserted here, so whatever those three do with a multibyte path is still uncovered.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (9) hits-of-code; +4 to give you at least something.  Your running score is [+657](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (9) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1656](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6318, opened 2026-08-03
#6299: guard recv against a negative received count on both platforms

### LINKED ISSUE #6299: `posix.recv` and `win32.recv` convert receive failures into `NegativeArraySizeException` instead of preserving platform status
## What happens

Both receive wrappers allocate a requested buffer, call the platform receive function, and then copy the reported count directly:

```java
new Data.ToPhi(Arrays.copyOf(buf, received))
```

A receive failure is reported as `-1`. Passing that value to `Arrays.copyOf` throws `NegativeArraySizeException`, hiding the return code that EO callers need to inspect. The problem is present in both platform implementations:

- `eo-runtime/src/main/java/org/eolang/posix/RecvSyscall.java`
- `eo-runtime/src/main/java/org/eolang/win32/RecvFuncCall.java`

It is reproducible without establishing a connection: calling `recv` with descriptor `-1`, a positive requested size, and flags `0` makes the platform call return `-1`; the wrapper then crashes while creating the output bytes.

## Expected behavior

A failed receive should preserve the platform result (`code = -1`) and expose an empty byte sequence as its output, instead of throwing a Java collection exception. `win32.ReadFuncCall` already applies this rule with `Arrays.copyOf(buf, Math.max(count, 0))`; the receive wrappers should provide the analogous behavior and cover it on both platforms.

## Why it matters

`posix` and `win32` operations model failure through their returned status. Replacing that status with an unrelated Java exception prevents EO code from handling an ordinary failed receive consistently.

gemshrine:
Fixes #6299.

Both receive wrappers copied the reported byte count straight into `Arrays.copyOf`:

```java
result.put(1, new Data.ToPhi(Arrays.copyOf(buf, received)));
```

A failed receive reports `-1`, so `Arrays.copyOf(buf, -1)` throws `NegativeArraySizeException`, hiding the status code EO callers are supposed to inspect. Confirmed by running the native call directly: `recv(-1, buf, 16, 0)` returns `-1`, after which `Arrays.copyOf(buf, -1)` throws, while `Arrays.copyOf(buf, Math.max(-1, 0))` yields an empty array.

The sibling `win32.ReadFuncCall` already guards this with `Math.max(count, 0)`; this applies the same rule to both receive wrappers:

- `posix/RecvSyscall.java`
- `win32/RecvFuncCall.java`

Now a failed receive keeps `code = -1` and exposes an empty byte sequence instead of throwing.

Added two regression tests in `SyscallTest` (one per platform, gated by the existing `@DisabledOnOs` split) that invoke the wrapper with descriptor `-1` and assert `code == -1` and an empty output, rather than a Java exception.

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/SyscallTest.java
+++ b/eo-runtime/src/test/java/org/eolang/SyscallTest.java
@@ -21,6 +21,8 @@ import java.util.Arrays;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import org.eolang.posix.CStdLib;
+import org.eolang.posix.RecvSyscall;
+import org.eolang.win32.RecvFuncCall;
 import org.eolang.win32.WSAStartupFuncCall;
 import org.eolang.win32.Winsock;
 import org.hamcrest.MatcherAssert;
@@ -218,6 +220,25 @@ final class SyscallTest {
                 this.cleanup();
                 server.stop();
             }
+        }
+
+        @Test
+        void receivesNothingWithoutCrashingOnFailedRecvViaSyscall() {
+            final Phi result = new RecvFuncCall(Phi.Φ.take("win32").copy()).make(
+                new Data.ToPhi(-1),
+                new Data.ToPhi(16),
+                new Data.ToPhi(0)
+            );
+            MatcherAssert.assertThat(
+                "A failed recv must report code -1 instead of throwing NegativeArraySizeException",
+                new Dataized(result.take("code")).asNumber().intValue(),
+                Matchers.equalTo(-1)
+            );
+            MatcherAssert.assertThat(
+                "A failed recv must expose an empty byte sequence as its output",
+                ((byte[]) new Dataized(result.take("output")).take()).length,
+                Matchers.equalTo(0)
+            );
         }
 
         @RepeatedIfExceptionsTest(repeats = 3)
@@ -560,6 +581,25 @@ final class SyscallTest {
                 this.closeSocket(socket);
                 server.stop();
             }
+        }
+
+        @Test
+        void receivesNothingWithoutCrashingOnFailedRecvViaSyscall() {
+            final Phi result = new RecvSyscall(Phi.Φ.take("posix").copy()).make(
+                new Data.ToPhi(-1),
+                new Data.ToPhi(16),
+                new Data.ToPhi(0)
+            );
+            MatcherAssert.assertThat(
+                "A failed recv must report code -1 instead of throwing NegativeArraySizeException",
+                new Dataized(result.take("code")).asNumber().intValue(),
+                Matchers.equalTo(-1)
+            );
+            MatcherAssert.assertThat(
+                "A failed recv must expose an empty byte sequence as its output",
+                ((byte[]) new Dataized(result.take("output")).take()).length,
+                Matchers.equalTo(0)
+            );
         }
 
         @RepeatedIfExceptionsTest(repeats = 3)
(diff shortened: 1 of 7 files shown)


Thayorns (commented):
The guard itself is right and it lands on both platforms, so a failed recv now keeps its -1 code instead of blowing up inside Arrays.copyOf. The regression tests do exercise the exact path the issue describes. A few things about how the tests are written are worth a look.

eo-runtime/src/test/java/org/eolang/SyscallTest.java
```diff
@@ -222,6 +224,25 @@ void connectsToLocalServerViaSyscall() throws IOException {
...
         }
 
+        @Test
+        void receivesNothingWithoutCrashingOnFailedRecvViaSyscall() {
+            final Phi result = new RecvFuncCall(Phi.Φ.take("win32").copy()).make(
+                new Data.ToPhi(-1),
+                new Data.ToPhi(16),
+                new Data.ToPhi(0)
+            );
+            MatcherAssert.assertThat(
```
Thayorns: Each of the two new tests carries two assertions: the code check on 234 and the output check on 239, and the same pair again on 598 and 603. Those are two separate behaviors, so when the code assertion fails the output one never runs and the run tells you only half of what broke.

eo-runtime/src/test/java/org/eolang/SyscallTest.java
```diff
@@ -222,6 +224,25 @@ void connectsToLocalServerViaSyscall() throws IOException {
...
+                "A failed recv must report code -1 instead of throwing NegativeArraySizeException",
+                new Dataized(result.take("code")).asNumber().intValue(),
+                Matchers.equalTo(-1)
+            );
+            MatcherAssert.assertThat(
+                "A failed recv must expose an empty byte sequence as its output",
+                ((byte[]) new Dataized(result.take("output")).take()).length,
```
Thayorns: `Dataized.take()` already returns `byte[]` (Dataized.java:57), so the `(byte[])` cast here and again on 605 does nothing. Line 52 of this same file reads the same value without one.

eo-runtime/src/test/java/org/eolang/SyscallTest.java
```diff
@@ -222,6 +224,25 @@ void connectsToLocalServerViaSyscall() throws IOException {
...
 
+        @Test
+        void receivesNothingWithoutCrashingOnFailedRecvViaSyscall() {
+            final Phi result = new RecvFuncCall(Phi.Φ.take("win32").copy()).make(
+                new Data.ToPhi(-1),
+                new Data.ToPhi(16),
+                new Data.ToPhi(0)
+            );
+            MatcherAssert.assertThat(
+                "A failed recv must report code -1 instead of throwing NegativeArraySizeException",
```
Thayorns: All four new failure messages are phrased as requirements rather than as claims about what went wrong: "A failed recv must report code -1" on 235 and 599, "must expose an empty byte sequence" on 240 and 604. When one of these fires, the line that reaches the log restates the rule instead of naming the failure.

morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+3508](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:

gemshrine:

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (122 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2215](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6320, opened 2026-08-03
#6306: validate tuple length before allocating the array

### LINKED ISSUE #6306: `TupleToArray` allows negative and invalid lengths causing `NegativeArraySizeException`
## What happens

`TupleToArray` obtains the `length` attribute as a `Double` and immediately narrows it before allocating the result:

```java
final int length = new Dataized(this.tuple.take("length")).asNumber().intValue();
final Phi[] arguments = new Phi[length];
```

There is no check that the value is a finite non-negative integer. A tuple-like EO object whose `length` dataizes to `-1` makes the constructor of the Java array throw `NegativeArraySizeException`. Fractional, non-finite, and out-of-range values are also silently narrowed rather than being rejected as an invalid tuple shape.

`TupleToArray` is the bridge used by the generic `posix` and `win32` dispatchers, so this turns malformed EO input into an implementation exception instead of an EO failure.

## Expected behavior

The tuple length must be validated before allocation: it should be finite, integral, within `int` range, and non-negative. Invalid values should result in `ExFailure` explaining that the supplied object is not a valid tuple, with no raw Java array exception.

## Suggested coverage

Create a minimal tuple-like `Phi` with a `length` attribute of `-1` and assert that `new TupleToArray(phi).get()` throws `ExFailure`. Add fractional and non-finite cases to protect the conversion boundary.

gemshrine:
Fixes #6306.

`TupleToArray` narrowed the `length` attribute to `int` and allocated `new Phi[length]` with no validation. Confirmed at runtime with a tuple-like object whose `length` dataizes to a bad value:

```
length = -1        -> NegativeArraySizeException
length = +Infinity -> OutOfMemoryError (array size exceeds VM limit)
length = 2.7       -> silently truncated to 2
length = NaN       -> silently treated as 0
```

Since `TupleToArray` is the bridge behind the generic `posix`/`win32` dispatchers, these leak raw JVM exceptions instead of an EO failure.

The fix validates the length before allocation: it must be finite, integral, non-negative, and within `int` range, otherwise it raises `ExFailure`. Added `TupleToArrayTest` covering the negative, non-finite, and fractional cases.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/TupleToArray.java
+++ b/eo-runtime/src/main/java/org/eolang/TupleToArray.java
@@ -28,7 +28,15 @@ public final class TupleToArray implements Supplier<Phi[]> {
 
     @Override
     public Phi[] get() {
-        final int length = new Dataized(this.tuple.take("length")).asNumber().intValue();
+        final double raw = new Dataized(this.tuple.take("length")).asNumber().doubleValue();
+        if (!Double.isFinite(raw) || raw < 0.0 || Math.rint(raw) != raw
+            || raw > Integer.MAX_VALUE) {
+            throw new ExFailure(
+                "The tuple length must be a finite non-negative integer within int range, but it was %s",
+                raw
+            );
+        }
+        final int length = (int) raw;
         final Phi[] arguments = new Phi[length];
         Phi tup = this.tuple;
         for (int idx = length - 1; idx >= 0; --idx) {
--- a/eo-runtime/src/test/java/org/eolang/TupleToArrayTest.java
+++ b/eo-runtime/src/test/java/org/eolang/TupleToArrayTest.java
@@ -1,0 +1,61 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang;
+
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link TupleToArray}.
+ * @since 0.57
+ */
+final class TupleToArrayTest {
+
+    @Test
+    void rejectsNegativeLengthWithFailure() {
+        MatcherAssert.assertThat(
+            "a negative tuple length must raise an EO failure, not a Java array exception",
+            Assertions.assertThrows(
+                ExAbstract.class,
+                () -> new TupleToArray(TupleToArrayTest.withLength(-1.0)).get(),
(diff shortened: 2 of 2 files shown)


eo-runtime/src/main/java/org/eolang/TupleToArray.java
```diff
@@ -28,7 +28,15 @@ public TupleToArray(final Phi tup) {
 
     @Override
     public Phi[] get() {
-        final int length = new Dataized(this.tuple.take("length")).asNumber().intValue();
+        final double raw = new Dataized(this.tuple.take("length")).asNumber().doubleValue();
+        if (!Double.isFinite(raw) || raw < 0.0 || Math.rint(raw) != raw
+            || raw > (double) Integer.MAX_VALUE) {
+            throw new ExFailure(
+                "The tuple length must be a finite non-negative integer within int range, but it was %s",
+                raw
```
morphqdd, line 36: This only rejects lengths strictly greater than Integer.MAX_VALUE. A length that's huge but still a valid int (or exactly Integer.MAX_VALUE) passes validation and reaches `new Phi[length]`, which can still throw OutOfMemoryError — the same raw-JVM-exception problem the issue asks to eliminate, just at a higher threshold.

eo-runtime/src/test/java/org/eolang/TupleToArrayTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Test case for {@link TupleToArray}.
+ * @since 0.57
+ */
+final class TupleToArrayTest {
+
+    @Test
+    void rejectsNegativeLengthWithFailure() {
+        MatcherAssert.assertThat(
```
morphqdd, line 20: This asserts `ExAbstract.class`, but the issue's suggested coverage asks for `ExFailure` specifically. `ExAbstract` is the common parent for all EO exceptions, so this doesn't confirm the code throws the intended exception type. Same for the other two tests below (lines 28, 36).

yegor256:

gemshrine:

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+3898](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2231](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6322, opened 2026-08-03
#6307: drop the trailing hex group delimiter

### LINKED ISSUE #6307: `VerboseBytesAsString.toHex()` incorrectly adds a trailing delimiter to full eight-hex-digit groups
## What happens

`VerboseBytesAsString.toHex()` groups hexadecimal output with:

```java
return new String(chars).replaceAll("(.{8})", "$1-");
```

The replacement inserts a hyphen after every complete group, including the last one. An eight-byte result is consequently rendered with a dangling delimiter, for example:

```text
[0x00000000-00000000-] = 0.0, or "\u0000..."
```

The same happens for any byte array whose hexadecimal form ends exactly on an eight-character group boundary. The delimiter is diagnostic formatting only, but it makes the shown representation not be valid grouped hexadecimal text and differs from the no-trailing-delimiter convention used elsewhere in the runtime.

## Expected behavior

Separators should occur only *between* groups. The example above should render its hexadecimal field as:

```text
0x00000000-00000000
```

without a hyphen before the closing bracket.

## Suggested coverage

Add assertions for 4, 8, and 9-byte input arrays. In particular, the 4- and 8-byte cases must not end their hexadecimal portion with `-`, while the 9-byte form should contain a separator only between the first four bytes and the remainder.

gemshrine:
Fixes #6307.

`toHex()` grouped hex with `replaceAll("(.{8})", "$1-")`, which appends a hyphen after every complete group, including the last one when the hex length is a multiple of eight. An eight-byte value therefore rendered as `[0x00000000-00000000-]` with a dangling delimiter.

Added a `(?=.)` lookahead so a separator is inserted only when another character follows, keeping hyphens strictly between groups. Updated the eight-byte double expectation (was `A8826AA9-]`, now `A8826AA9]`) and added tests for the 8-byte boundary and a 9-byte case.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/VerboseBytesAsString.java
+++ b/eo-runtime/src/main/java/org/eolang/VerboseBytesAsString.java
@@ -72,7 +72,7 @@ public final class VerboseBytesAsString implements Supplier<String> {
             chars[idx * 2] = VerboseBytesAsString.HEX_ARRAY[value >>> 4];
             chars[idx * 2 + 1] = VerboseBytesAsString.HEX_ARRAY[value & 0x0F];
         }
-        return new String(chars).replaceAll("(.{8})", "$1-");
+        return new String(chars).replaceAll("(.{8})(?=.)", "$1-");
     }
 
     /**
--- a/eo-runtime/src/test/java/org/eolang/VerboseBytesAsStringTest.java
+++ b/eo-runtime/src/test/java/org/eolang/VerboseBytesAsStringTest.java
@@ -27,8 +27,28 @@ final class VerboseBytesAsStringTest {
                 ByteBuffer.allocate(Double.BYTES).putDouble(12.345_67d).array()
             ).get(),
             Matchers.equalTo(
-                "[0x4028B0FB-A8826AA9-] = 12.34567, or \"@(\\ufffd\\ufffd\\ufffd\\ufffdj\\ufffd\""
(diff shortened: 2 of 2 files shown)


Thayorns (approved):
The lookahead is the right narrowing: a separator now lands only where another character follows it, so the hyphen stays strictly between groups instead of trailing off the end. Good that the eight-byte expectation carrying the dangling hyphen was corrected rather than worked around, and the two new cases pin both sides of the boundary, eight bytes and nine. Nothing else caught my eye.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (24) hits-of-code; +4 to give you at least something.  Your running score is [+769](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1668](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6323, opened 2026-08-03
#6308: clamp the withi insertion index to avoid duplicating elements

### LINKED ISSUE #6308: `tuple.withi` incorrectly duplicates tuple when insertion index is out of range
## What happens

`tuple.withi` builds its result from `tuple.front list index` and `tuple.back list (list.length - index)`. Both helper operations deliberately clamp an oversized count to the full tuple, but their combination does not: an index beyond the end makes both halves equal to `list`.

For example, inserting at index `5` into a three-element tuple produces the original tuple twice:

```eo
tuple.withi
  * 1 2 3
  5
  9
```

The result is `* 1 2 3 9 1 2 3`, not a tuple containing one inserted item. The same duplication occurs for sufficiently negative indices.

## Why this is a bug

The public contract says that `tuple.withi` inserts one `item` into `list` at `index`. An out-of-range index must not duplicate pre-existing elements. In particular, `tuple.front` and `tuple.back` already document clamping behavior for oversized counts, so `withi` should preserve that boundary behavior rather than compose it into a duplicated tuple.

## Expected behavior

Normalize the insertion index before splitting the tuple. An index greater than `list.length` should be clamped to `list.length`, producing `* 1 2 3 9`; an index below the supported lower bound should be handled explicitly and must also insert exactly one item. Add regression cases for both sides of the range.

gemshrine:
Fixes #6308.

`tuple.withi` split the tuple with the raw `index` for both halves. A negative index left both `front` and `back` returning overlapping slices, so the result duplicated existing elements. Confirmed at runtime on `* 1 2 3`:

```
withi (* 1 2 3) -1 9  ->  * 3 9 1 2 3   (5 elements, "3" duplicated)
```

(The oversized-index example from the report, `withi (* 1 2 3) 5 9`, already yields `* 1 2 3 9` on master because `back` clamps its negative count to empty; the actual defect is the negative side.)

The fix normalizes the split point once, clamping it to `[0, list.length]`, and uses it for both `front` and `back`:

```
if. > bounded
  0.gt index
  0
  if.
    index.gt list.length
    list.length
    index
```

Now a negative index inserts a single item at the front and an oversized index at the end, with no duplication. Added regression tests for both boundaries.

### DIFF
--- a/eo-runtime/src/main/eo/tuple/withi.eo
+++ b/eo-runtime/src/main/eo/tuple/withi.eo
@@ -10,11 +10,46 @@
 [list index item] > withi
   concat > @
     with.
-      front list index
+      front list bounded
       item
     back
       list
-      list.length.minus index
+      list.length.minus bounded
+  if. > bounded
+    0.gt index
+    0
+    if.
+      index.gt list.length
+      list.length
+      index
+
+  eq. ++> can-clamp-a-negative-index-to-the-front-without-duplicating
+    withi
+      *
+        1
+        2
+        3
+      -1
+      9
+    *
+      9
+      1
+      2
+      3
+
+  eq. ++> can-clamp-an-index-beyond-the-end-without-duplicating
+    withi
+      *
+        1
+        2
+        3
+      5
+      9
+    *
+      1
+      2
+      3
+      9
 
   eq. ++> can-insert-at-a-zero-index
     withi


Thayorns (approved):
Clamping the split point once and handing the same value to both halves is the right shape. The old code derived the two slices from two different numbers, which is exactly how the overlap appeared, and the two new examples pin the boundaries that used to duplicate. One thing worth a moment: the docblock still reads only as insert item into list at the given index and says nothing about what happens outside the bounds, while back.eo spends its first two lines saying what an oversized index does there. Someone reading withi.eo now has to read the body to learn that -1 means the front.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+773](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (39) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1680](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6324, opened 2026-08-03
#6309: double underscores in JavaPath to match the transpiler encoding

### LINKED ISSUE #6309: `JavaPath` not properly resolving EO object names with underscores due to mismatched escape mechanisms with transpiler
## What happens

The Java transpiler and the runtime use different name encodings for an EO identifier containing `_`.

`to-java.xsl` first maps an underscore to `__`, so an EO object named `foo_bar` is emitted as Java class `EOfoo__bar`. This behavior is covered by `underscore-to-java.yaml`. In contrast, `JavaPath.java` only escapes `$` and replaces `-` with `_`; for `Φ.foo_bar` it returns `org.eolang.EOfoo_bar`.

Consequently, a runtime lookup of an underscore-containing EO object asks the class loader for a class that the transpiler never generates. The generated `EOfoo__bar` cannot be reached through `Phi.Φ.take("foo_bar")`.

## Why this is a bug

Underscores are accepted in EO identifiers and are intentionally escaped by the Java generator to keep the mapping reversible with names containing dashes. Runtime resolution must use the same encoding as generated Java sources; otherwise a valid transpiled program fails only when its object is loaded.

## Expected behavior

Make `JavaPath` apply the same escaping order as the relevant `to-java.xsl` templates, including `_` → `__` before `-` → `_`. Add a regression test that checks both the generated class name and runtime resolution of an object such as `foo_bar`.

gemshrine:
Fixes #6309.

`to-java.xsl` escapes an EO identifier as `replace(replace(translate(translate(replace($n, '_', '__'), '-', '_'), ...)))`, so `_` is doubled before `-` becomes `_` (the `underscore-to-java.yaml` pack asserts `foo_underscore` -> class `EOfoo__underscore`). `JavaPath` only did `-` -> `_`, so it resolved `Φ.foo_bar` to `EOfoo_bar`, a class the transpiler never emits. It also collided `foo_bar` and `foo-bar` onto the same Java name.

Confirmed at runtime before the fix:

```
Φ.foo_bar -> org.eolang.EOfoo_bar   (transpiler emits EOfoo__bar)
Φ.a_b-c   -> org.eolang.EOa_b_c     (transpiler emits EOa__b_c)
```

The fix applies `_` -> `__` before `-` -> `_`, matching the stylesheet order, so the mapping is reversible again. Added `foo_bar`, `sub_under`, and mixed `a_b-c` cases to `JavaPathTest`.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/JavaPath.java
+++ b/eo-runtime/src/main/java/org/eolang/JavaPath.java
@@ -74,7 +74,9 @@ final class JavaPath {
                 prefix = "EO_";
             }
             out.append('.').append(prefix)
-                .append(parts[idx].replace("$", "$EO").replace("-", "_"));
+                .append(
+                    parts[idx].replace("_", "__").replace("-", "_").replace("$", "$EO")
+                );
         }
         return out.toString();
     }
--- a/eo-runtime/src/test/java/org/eolang/JavaPathTest.java
+++ b/eo-runtime/src/test/java/org/eolang/JavaPathTest.java
@@ -24,7 +24,10 @@ class JavaPathTest {
         "Φ.obj.sub, org.eolang.EO_obj.EOsub",
         "Φ.obj.sub$attr, org.eolang.EO_obj.EOsub$EOattr",
         "Φ.obj.sub-dashed$attr, org.eolang.EO_obj.EOsub_dashed$EOattr",
-        "Φ.obj.sub-dashed$attr-dashed, org.eolang.EO_obj.EOsub_dashed$EOattr_dashed"
+        "Φ.obj.sub-dashed$attr-dashed, org.eolang.EO_obj.EOsub_dashed$EOattr_dashed",
+        "Φ.foo_bar, org.eolang.EOfoo__bar",
+        "Φ.obj.sub_under, org.eolang.EO_obj.EOsub__under",
+        "Φ.a_b-c, org.eolang.EOa__b_c"
     })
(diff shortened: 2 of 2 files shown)


Thayorns (approved):
Doubling the underscore before the dash becomes one is exactly the order to-java.xsl uses when it builds the class name, so the two encodings line up again and foo_bar stops colliding with foo-bar. Moving the dollar replacement to the end changes nothing, since EO cannot appear in the middle of it. The new rows cover toString well. convertsToPackage sits right below them with no underscore row, so pkg(), which walks the same java() body you changed, is still exercised only on names without an underscore.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (9) hits-of-code; +4 to give you at least something.  Your running score is [+777](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (9) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1692](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6325, opened 2026-08-03
#6310: validate the as-char code is a byte in [0, 255]

### LINKED ISSUE #6310: `string.as-char` doesn't validate input and wraps codes outside the 0–255 range
## What happens

`string.as-char` documents `code` as an ASCII byte value in the range `0–255`, but its implementation takes the last byte of `code.as-i64.as-bytes` without validating the input. Values outside that range are silently reduced modulo 256.

For example, these distinct inputs generate the same byte:

```eo
string.as-hex (string.as-char 0).as-bytes
# 00

string.as-hex (string.as-char 256).as-bytes
# 00
```

Likewise, `string.as-char -1` produces the raw byte `FF` instead of reporting that the supplied code is outside the stated domain.

## Why this is a bug

The operation is presented as conversion of a valid byte code, not as a modulo-256 truncation primitive. Silently accepting `256`, `-1`, or a fractional number hides invalid input and makes the result depend on the implementation detail that only the least-significant byte is sliced.

## Expected behavior

Validate that `code` is a finite integer from `0` through `255`. Values outside that range should fail through the normal EO error path, rather than being truncated. Add boundary tests for `0`, `255`, `-1`, and `256`.

gemshrine:
Fixes #6310.

`string.as-char` took the least-significant byte of `code.as-i64.as-bytes` with no validation, so out-of-domain inputs were silently truncated modulo 256. Confirmed at runtime:

```
as-char 0    -> 00
as-char 255  -> FF
as-char 256  -> 00   (collides with 0)
as-char -1   -> FF   (collides with 255)
as-char 3.5  -> 03   (fractional truncated)
as-char 300  -> 2C
```

The fix rejects any `code` that is not an integer in `[0, 255]`, failing through the normal EO error path instead of truncating. Valid codes keep their previous single-byte result. Confirmed after the fix that `256`, `-1`, `3.5`, `300` raise `ExFailure` while `0`, `65`, `255` still succeed.

Added `AsCharDomainTest` asserting the reject/accept boundaries (a hand-written JUnit case, since an EO `-->` example cannot distinguish a genuine failure from `asBool` throwing on the old truncated string).

### DIFF
--- a/eo-runtime/src/main/eo/number/as-char.eo
+++ b/eo-runtime/src/main/eo/number/as-char.eo
@@ -1,5 +1,7 @@
 # Writes the ASCII `code` (0-255) as its single byte, the inverse of `as-ascii`.
-# The number is read as a 64-bit integer and its least-significant byte is taken.
+# The receiver must be an integer within the range [0, 255]; any other value
+# (negative, above 255, or fractional) terminates the computation instead of
+# being silently truncated to its least-significant byte.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
@@ -9,9 +11,17 @@
 +spdx SPDX-License-Identifier: MIT
 
 [] > as-char
-  ^.as-i64.as-bytes.slice > @
-    7
-    1
+  if. > @
+    or.
+      or.
+        0.gt ^
+        ^.gt 255
+      ^.is-integer.not
+    T
+      "The receiver of as-char must be an integer within the range [0, 255]"
+    ^.as-i64.as-bytes.slice
+      7
+      1
 
   eq. ++> can-round-trip-through-as-ascii
     "!"
@@ -40,3 +50,9 @@
   eq. ++> can-write-the-byte-of-the-zero-digit
     "0"
     string 48.as-char
+
+  256.as-char --> stops-on-a-code-past-the-byte-range
+
+  3.5.as-char --> stops-on-a-fractional-code
+
+  -1.as-char --> stops-on-a-negative-code


Thayorns (commented):
Rejecting the out-of-domain code instead of quietly folding it modulo 256 is the right call, and the two collisions the issue names, 256 with 0 and -1 with 255, are both closed by the integrality check riding along with the range check. The Java test earns its place: an eq example really cannot tell a deliberate failure from asBool blowing up on the old truncated byte. Worth noting that the unlint you added is file-wide, so redundant-object is now silenced for every object in as-char.eo, examples included, not just for the branch that needed it.

eo-runtime/src/main/eo/string/as-char.eo
```diff
@@ -1,17 +1,30 @@
...
 [code] > as-char
-  code.as-i64.as-bytes.slice > @
-    7
-    1
+  valid.if > @
+    code.as-i64.as-bytes.slice
+      7
+      1
+    T
+      "The 'code' passed to as-char must be an integer within the range [0, 255]"
```
Thayorns: The message names no value, so it tells the caller the rule but not what tripped it. as-char 300, as-char -1 and as-char 3.5 all fail identically, and the three go wrong for different reasons. chunk.eo does this the other way a few files over, feeding the offending sizes through printf so the report carries the numbers that explain it.

eo-runtime/src/main/eo/string/as-char.eo
```diff
@@ -1,17 +1,30 @@
 # Writes the ASCII `code` (0-255) as its single byte, the inverse of `as-ascii`.
-# The number is read as a 64-bit integer and its least-significant byte is taken.
+# The `code` must be an integer within the range [0, 255]; any other value
+# (negative, above 255, or fractional) fails instead of being silently
+# truncated to its least-significant byte.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
 +package string
++unlint redundant-object
```
yegor256: @gemshrine can you get rid of this?

eo-runtime/src/test/java/org/eolang/AsCharDomainTest.java
```diff
@@ -0,0 +1,51 @@
...
+            Matchers.equalTo(1)
+        );
+    }
+
+    /**
+     * Build {@code as-char} applied to the given code.
+     * @param code The numeric code
+     * @return The as-char object
+     */
+    private static Phi asChar(final double code) {
```
morphqdd: This is a `static` method. CI's smell ratchet is currently red on this PR ("Found 854 smells, threshold is 853") — this method is the new `static|null` hit. Making it an instance method (or inlining the two-line body at each call site) would clear it.

eo-runtime/src/test/java/org/eolang/AsCharDomainTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/**
+ * Test case for the domain validation of {@code string.as-char}.
+ * @since 0.57
+ */
+@SuppressWarnings("JTCOP.RuleAllTestsHaveProductionClass")
```
morphqdd: This new suppression also trips the CI suppression-count ratchet ("Found 369 suppressions, threshold is 368"), currently red on this PR. Either restructure so the class maps to a production class JTCOP recognizes, or bump the threshold file in the same PR.

yegor256:

gemshrine:

gemshrine:
Rebased on the latest master. The file moved to the `number` package in #6655, so I ported the fix onto `number/as-char.eo` and re-pointed the domain test at `EO_number.EOas_char`. Issue #6310 is still open, and the truncation it describes is still there, so the validation applies as before. All required checks are green. Ready for review.

yegor256:
@gemshrine conflict again, sorry

eo-runtime/src/test/java/org/eolang/AsCharDomainTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/**
+ * Test case for the domain validation of {@code number.as-char}.
+ * @since 0.57
+ */
+@SuppressWarnings("JTCOP.RuleAllTestsHaveProductionClass")
```
yegor256: @gemshrine don't suppress. instead, remove this Java test and create EO test.

yegor256 (changes requested):

yegor256:
@gemshrine don't suppress. instead, remove this Java test and create EO test.

eo-runtime/src/test/java/org/eolang/AsCharDomainTest.java
gemshrine: Done, the Java test is gone and the domain is covered by EO tests inside `number/as-char.eo` instead:

```
256.as-char --> stops-on-a-code-past-the-byte-range

3.5.as-char --> stops-on-a-fractional-code

-1.as-char --> stops-on-a-negative-code
```

That also drops the two `counts` ratchet hits the Java test brought in, the suppression and the static helper, so the branch is back at the master numbers.

Also rebased on the latest master.

gemshrine:
@yegor256 the Java test is gone, the domain is covered by EO tests now, and the branch is rebased. Please merge when the checks go green.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3030](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (24) hits-of-code; +1 to give you at least something.  Your running score is [+2669](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (2) comments; -4 for too few (24) hits-of-code; +1 to give you at least something.  Your running score is [+5872](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (2) comments; -4 for too few (24) hits-of-code; +1 to give you at least something.  Your running score is [+1971](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6326, opened 2026-08-03
#6304: consume only leading CLI options, pass the rest to the object

### LINKED ISSUE #6304: `Main` class prematurely classifies arguments starting with `--` as options, causing them to be dropped
## What happens

The CLI class classifies every token beginning with `--` as an option before it determines the EO object and its arguments:

```java
private static boolean isOption(final String arg) {
    return arg.startsWith("--");
}
```

It then removes all such tokens from the list passed to `run(...)`. Unknown options do not cause an error, so a normal EO argument is silently lost. For example, when an EO object is invoked with the string argument `--mode`, the object receives no argument at all.

The help text promises that every argument after the class name is wrapped as a string and passed to the EO object, but there is neither a `--` end-of-options delimiter nor a rule that stops option parsing after the object name.

## Expected behavior

Only leading recognized CLI options should be consumed. Once the EO class name has been read, all remaining tokens — including `--mode`, `--`, and `--verbose` — must be delivered to the EO object unchanged. Alternatively, support the conventional `--` delimiter and reject unknown options before it.

## Steps to reproduce

1. Build an EO object which returns its first command-line argument.
2. Run `org.eolang.Main` with the object name followed by `--mode`.
3. Observe that the tuple of arguments is empty instead of containing the string `--mode`.

## Suggested coverage

Add CLI tests for an argument beginning with `--`, an argument equal to `--`, and a literal `--verbose` after the object name.

gemshrine:
Fixes #6304.

`Main` filtered every `--`-prefixed token out of the argument list, wherever it appeared, so a normal EO argument like `--mode` after the object name was silently swallowed and never reached the object. The usage line already promises `[option...] class [argument...]`, i.e. options precede the class.

The fix consumes options only while no non-option (the class name) has been seen yet; once the class name is read, every remaining token, including `--mode`, `--`, and `--verbose`, is delivered to the EO object unchanged.

Added `treatsOptionAfterObjectNameAsArgument`, which asserts that a `--verbose` placed after the object name is no longer consumed as the verbose option (no stacktrace is printed), complementing the existing `printsFullStacktraceIfVerboseIsEnabled` where `--verbose` precedes the object.

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/MainTest.java
+++ b/eo-runtime/src/test/java/org/eolang/MainTest.java
@@ -139,6 +139,15 @@ final class MainTest {
         );
     }
 
+    @Test
+    void treatsOptionAfterObjectNameAsArgument() {
+        MatcherAssert.assertThat(
+            "A --verbose placed after the object name must be an argument, not the verbose option, so no stacktrace is printed",
+            MainTest.stderr("org.eolang.examples.app", Main.VERBOSE),
+            Matchers.not(Matchers.containsString("at org.eolang.Main.run"))
+        );
+    }
+
     private static String stderr(final String... cmds) {
         return MainTest.exec(cmds).stderr();
     }
--- a/eo-runtime/src/main/java/org/eolang/Main.java
+++ b/eo-runtime/src/main/java/org/eolang/Main.java
@@ -9,6 +9,7 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
@@ -19,7 +20,6 @@ import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 
 /**
  * Bridge between Java CLI and EO.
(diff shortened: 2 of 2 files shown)


Thayorns (approved):
Consuming options only until the first non-option is the right rule and it matches the usage line the program already prints, so the class name is what ends the option run rather than the shape of the token. Dropping isArgument along with it is clean, since the single loop now decides membership once. The new test is a genuine regression rather than a restatement: on master the trailing --verbose is still swallowed as the option and the stacktrace appears, so the assertion fails there and passes here.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+781](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-runtime/src/test/java/org/eolang/MainTest.java
```diff
@@ -185,6 +185,15 @@ void printsFullStacktraceIfVerboseIsEnabled() {
         );
     }
 
+    @Test
+    void treatsOptionAfterObjectNameAsArgument() {
```
morphqdd, line 143: This only checks the absence of a side effect (no stacktrace string), not that `--verbose` actually reached the object as an argument. It would also pass if `--verbose` were silently dropped entirely instead of delivered. The issue asks for an EO object that echoes its first argument — asserting that value directly would confirm delivery, not just absence of the old behavior. Also missing from the suite: a literal `--` after the object name, and an unrecognized option like `--mode` after the boundary — both called out in the issue's suggested coverage.

yegor256:

gemshrine:

gemshrine:
@yegor256 re-resolved conflicts after the latest merges, rebased on current master.

gemshrine:

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (35) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2404](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6327, opened 2026-08-03
#6286: give a float overflow its own message instead of suggesting Infinity

### LINKED ISSUE #6286: Parser suggests writing `Infinity` for an over-precise float literal, but `Infinity` is not a valid literal
## What happens

A finite-looking `FLOAT` literal that overflows to IEEE infinity (for example `1.0E400`) is rejected with a diagnostic that tells the author to "write `Infinity` instead". But `Infinity` is not a valid EO literal: writing it makes the parser fail with the internal-sounding `line shape not yet implemented in spec parser`. Following the compiler's own advice produces a worse error than the one it was meant to fix. The negative case behaves the same: `-1.0E400` suggests `-Infinity`, and `-Infinity` is rejected with `expected value`.

### Runtime evidence

Probed directly against `org.eolang.parser.EoSyntax` (each literal placed as `<literal> > x` inside a formation body, XML checked for `<error>`):

| literal | parser result |
| --- | --- |
| `1.0E400` | **ERROR**: `1.0E400 is over-precise, write Infinity instead` |
| `Infinity` | **ERROR**: `line shape not yet implemented in spec parser` |
| `-1.0E400` | **ERROR**: `-1.0E400 is over-precise, write -Infinity instead` |
| `-Infinity` | **ERROR**: `expected value` |
| `1.0E30` | ok (control: a large but finite float parses) |
| `9007199254740993` | `... write 9007199254740992 instead` (control: the integer suggestion *does* re-parse) |

So the over-precise suggestion is actionable for integers and finite floats, but for a float that overflows to infinity it points at a token the lexer cannot read.

## Root cause

`eo-parser/src/main/java/org/eolang/parser/Emissions.java`, `number()` (around lines 414-431), folds the non-finite (overflow) case into the over-precise case and, for a `FLOAT`, builds the suggested spelling with `Double.toString(parsed)`:

```java
final double parsed = Double.parseDouble(value.raw());
if (!Double.isFinite(parsed) || Emissions.overPrecise(value.raw(), parsed)) {
    final String canonical;
    if (value.kind() == Value.Kind.INTEGER) {
        canonical = Emissions.canonicalInteger(parsed);
    } else {
        canonical = Double.toString(parsed);   // "Infinity" / "-Infinity" for an overflow
    }
    throw new ParseError(
        line, value.pos(),
        String.format("%s is over-precise, write %s instead", value.raw(), canonical)
    );
}
```

When `parsed` is infinite, `Double.toString(parsed)` is `"Infinity"` / `"-Infinity"`. Neither is a lexable head: an `Infinity` head reaches `Eo.rejected()` (`line shape not yet implemented in spec parser`), and a leading `-Infinity` reaches `expected value`. The neighbouring `canonicalInteger` helper's own Javadoc even states the intended invariant — "the suggestion itself re-parses as FLOAT" — which this `FLOAT` branch violates for the overflow case.

## Suggested fix

Split the overflow case out from over-precision so the message never names an unparseable token, for example:

```java
final double parsed = Double.parseDouble(value.raw());
if (!Double.isFinite(parsed)) {
    throw new ParseError(
        line, value.pos(),
        String.format("%s is out of the finite range of a double", value.raw())
    );
}
if (Emissions.overPrecise(value.raw(), parsed)) {
    ...  // existing over-precise message, whose suggestion does re-parse
}
```

(Alternatively, if an infinity literal is meant to exist, implement `Infinity` / `-Infinity` as literals so the suggestion becomes actionable.) Either way the diagnostic must not tell the author to write a token the parser then rejects.

A regression check: assert that `1.0E400` is rejected with an overflow message, and that the suggested replacement in any over-precise message, when fed back through the parser, itself parses without error.

gemshrine:
Fixes #6286.

`Emissions.number()` folded the IEEE-overflow case into the over-precise case and, for a `FLOAT`, built the suggested spelling with `Double.toString(parsed)`. When the literal overflows (e.g. `1.0E400`), that is `"Infinity"` / `"-Infinity"`, neither of which is a lexable head, so following the compiler's advice produced a worse error (`line shape not yet implemented in spec parser` / `expected value`).

The fix rejects a non-finite parse with its own message that names no unparseable token, and keeps the over-precise branch (whose suggestion does re-parse) unchanged. Verified against `EoSyntax`:

```
1.0E30            -> ok
1.0E400           -> '1.0E400 is out of the finite range of a double'
-1.0E400          -> '-1.0E400 is out of the finite range of a double'
9007199254740993  -> 'over-precise, write 9007199254740992 instead'  (still actionable)
```

Added the `float-overflow.yaml` typo-message regression.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -415,7 +415,15 @@ final class Emissions {
         final Emit emit, final String name, final Value value, final int line
     ) {
         final double parsed = Double.parseDouble(value.raw());
-        if (!Double.isFinite(parsed) || Emissions.overPrecise(value.raw(), parsed)) {
+        if (!Double.isFinite(parsed)) {
+            throw new ParseError(
+                line, value.pos(),
+                String.format(
+                    "%s is out of the finite range of a double", value.raw()
+                )
+            );
+        }
+        if (Emissions.overPrecise(value.raw(), parsed)) {
             final String canonical;
             if (value.kind() == Value.Kind.INTEGER) {
                 canonical = Emissions.canonicalInteger(parsed);
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-typos/float-overflow.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-typos/float-overflow.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+line: 2
+message: |-
+  [2:2] error: '1.0E400 is out of the finite range of a double'
+    1.0E400 > x
+    ^
+input: |
+  [] > app
+    1.0E400 > x


Thayorns (approved):
Splitting the non-finite case out of the over-precise branch is right. The old path handed Infinity to the suggestion builder, which is the one string that cannot be written back into the source, so the advice made things worse rather than better. The new message carries the offending literal, which is what a reader needs. The pack covers the positive overflow; -1.0E400 and an integer literal long enough to overflow both take the same new branch and neither has a case, so the sign and kind handling rests on reading rather than on the suite.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (22) hits-of-code; +4 to give you at least something.  Your running score is [+785](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1704](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6328, opened 2026-08-03
#6287: reject an unrecognised string escape instead of keeping it verbatim

### LINKED ISSUE #6287: `STRING` parser silently accepts an unrecognized backslash escape (`\q`, `\9`) instead of a lexical error
## What happens

The spec (R-9.7.3, last line: "Any other backslash sequence is a lexical error") requires that a backslash inside a `STRING` / `TEXT` literal begin one of the recognised escapes; anything else is a lexical error. The parser instead **silently accepts** an unrecognised escape such as `\q`, `\9`, or `\x`, keeping the backslash and the following character verbatim in the decoded string, with no error. Only the `\u` and octal paths validate; the single-character path does not.

### Runtime evidence

Probed directly against `org.eolang.parser.EoSyntax` (each literal placed as `<literal> > x` inside a formation body, XML checked for `<error>`):

| literal | parser result | spec (R-9.7.3) |
| --- | --- | --- |
| `"\n"` | ok (control: decodes to U+000A) | valid |
| `"\uXYZ"` | **ERROR**: `invalid unicode or octal escape in string literal` | error (correct) |
| `"\u00"` | **ERROR**: `invalid unicode or octal escape in string literal` | error (correct) |
| `"\q"` | **ok** (kept verbatim as `\q`) | **lexical error** |
| `"\9"` | **ok** (`9` is not an octal digit) | **lexical error** |
| `"\x"` | **ok** | **lexical error** |

So the `\u` / octal escapes are validated, but any unrecognised single-character escape slips through unreported.

## Root cause

`eo-parser/src/main/java/org/eolang/parser/Emissions.java`, `singleCharEscape(head, next)` — the fall-through `else` branch returns the two source characters verbatim instead of rejecting:

```java
} else if (next == '"' || next == '\'' || next == '\\') {
    decoded = String.valueOf(next);
} else {
    decoded = new String(new char[]{head, next});   // keeps "\q" as a literal backslash + 'q'
}
```

Every recognised escape has its own explicit branch (`\n`, `\t`, `\r`, `\b`, `\f`, `\"`, `\'`, `\\`); the `else` arm is reached only by an escape that R-9.7.3 defines as a lexical error. The neighbouring `\u` and octal paths (`appendUnicode` / `appendOctal`) do signal the error — `unescape` catches their `NumberFormatException` and raises `invalid unicode or octal escape in string literal` — but `singleCharEscape` swallows the unknown escape.

Note `\9` reaches this branch too: `unescapeBody` only routes `\` + `[0-7]` to `appendOctal`, so `\8` / `\9` fall to `singleCharEscape` and are likewise kept verbatim.

## Suggested fix

Reject the unrecognised escape instead of preserving it, so the existing "invalid escape" boundary reports it:

```java
} else {
    throw new NumberFormatException(
        String.format("unrecognised escape sequence '\\%s'", next)
    );
}
```

(caught by `unescape`'s existing `catch (NumberFormatException)` → `ParseError`, matching how `\u` / octal errors are surfaced; or raise a `ParseError` directly with a dedicated message). A regression check: `"\q"`, `"\9"`, and `"\x"` must each parse to an error, while every escape in the R-9.7.3 table continues to decode as before.

gemshrine:
Fixes #6287.

`singleCharEscape` returned an unrecognised escape (`\q`, `\9`, `\x`) verbatim, so the parser silently kept a literal backslash instead of reporting the lexical error required by R-9.7.3. Every recognised escape (`\n`, `\t`, `\r`, `\b`, `\f`, `\"`, `\'`, `\`) has its own branch; the `else` arm is only reached by a sequence the spec defines as an error.

The fix throws `NumberFormatException` from that branch, which `unescape` already catches and turns into the same `ParseError` used for the `\u` / octal boundary. Verified against `EoSyntax`:

```
"\n" "\t" "\\" "\""  -> ok
"\q" "\9" "\x"        -> error: invalid unicode or octal escape in string literal
```

No `.eo` source in the repo uses an unrecognised escape (regex literals already write `\`), and the full `EoSyntaxTest` suite (1052 cases) stays green. Added the `unknown-string-escape.yaml` regression.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -603,10 +603,11 @@ final class Emissions {
 
     /**
      * Resolve a single-character escape sequence (e.g. {@code \n},
-     * {@code \t}). Unknown sequences are passed through verbatim.
+     * {@code \t}). An unrecognised sequence is a lexical error (R-9.7.3).
      * @param head Backslash character (always {@code '\\'})
      * @param next The character after the backslash
      * @return The decoded character(s)
+     * @throws NumberFormatException If the escape is not recognised
      */
     private static String singleCharEscape(final char head, final char next) {
         final String decoded;
@@ -623,7 +624,9 @@ final class Emissions {
         } else if (next == '"' || next == '\'' || next == '\\') {
             decoded = String.valueOf(next);
         } else {
-            decoded = new String(new char[]{head, next});
+            throw new NumberFormatException(
+                String.format("unrecognised escape sequence '%c%c'", head, next)
+            );
         }
         return decoded;
     }
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-typos/unknown-string-escape.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-typos/unknown-string-escape.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+line: 2
+message: |-
+  [2:2] error: 'invalid unicode or octal escape in string literal'
+    "\q" > x
+    ^
+input: |
+  [] > app
+    "\q" > x


eo-parser/src/main/java/org/eolang/parser/Emissions.java
```diff
@@ -623,7 +624,9 @@ private static String singleCharEscape(final char head, final char next) {
         } else if (next == '"' || next == '\'' || next == '\\') {
             decoded = String.valueOf(next);
         } else {
-            decoded = new String(new char[]{head, next});
+            throw new NumberFormatException(
```
Thayorns, line 627: The message you build here never reaches anybody. The catch around unescape, further up in this same class, throws away the NumberFormatException text and reports invalid unicode or octal escape in string literal instead, and the yaml in this pull request asserts exactly that string. So the author of a stray backslash-q is sent looking at unicode and octal escapes, neither of which they wrote, while the sequence that actually offended is only visible in the caret position.

Thayorns (commented):
Turning the fall-through arm into a failure is right, and reusing the channel unescape already has for the escape errors keeps the call sites untouched. The check that no .eo source in the repo relies on a verbatim unknown escape is the part that makes this safe to land.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (19) hits-of-code; +1 to give you at least something.  Your running score is [+797](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (19) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1716](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6329, opened 2026-08-03
#6288: validate void parameter names against the NAME grammar

### LINKED ISSUE #6288: Formation void parameter names are not validated against the `NAME` grammar (`[a,b]`, `[a.b]`, tab-separated all accepted)
## What happens

A formation's void parameters are split on the ASCII space alone, and each resulting token is used as the parameter name **without validating it against the `NAME` grammar** (§2.3 / R-3.4.2). As a result a parameter name may contain any character the `NAME` token explicitly excludes — `,` `.` `|` `'` `:` `;` `!` `?` and TAB — yielding a void child whose emitted `name` is not a valid `NAME`.

### Runtime evidence

Probed directly against `org.eolang.parser.EoSyntax` (head is `<params> > obj`, body `5 > x`); the `name` attributes emitted for the void(s) are shown:

| head | parser | emitted void name(s) | `NAME` (§2.3) |
| --- | --- | --- | --- |
| `[a b]` | ok | `a`, `b` (control: two params) | valid |
| `[a,b]` | ok | **`a,b`** | invalid (`,` excluded) |
| `[a.b]` | ok | **`a.b`** | invalid (`.` excluded) |
| `[a\|b]` | ok | **`a\|b`** | invalid |
| `[a:b]` | ok | **`a:b`** | invalid |
| `[a!b]` | ok | **`a!b`** | invalid |
| `[a?b]` | ok | **`a?b`** | invalid |
| `[a<TAB>b]` | ok | **`a&#x9;b`** (tab inside the name) | invalid (tab excluded; R-3.4.5 also requires exactly one space) |

For contrast, the parser *does* reject a leading/trailing space (`[ a]`, `[a ]`), a double space (`[a  b]`), and `^` (`[^]` → "parameter names in voids must be NAME or @"). Only the `NAME` content of a token is left unchecked.

## Root cause

`eo-parser/src/main/java/org/eolang/parser/LnFormation.java`. The parameter split advances only past the ASCII space:

```java
while (end < inside.length() && inside.charAt(end) != ' ') {
    end = end + 1;
}
final String raw = inside.substring(idx, end);
out.add(LnFormation.mapParam(raw, span, span.indent() + 1 + idx));
```

and `mapParam` returns the token verbatim, checking only `^` and `@`, never the `NAME` grammar:

```java
if ("@".equals(raw)) {
    mapped = "φ";
} else {
    mapped = raw;   // any characters accepted, including NAME terminators
}
```

So `a,b`, `a.b`, `a<tab>b`, … are read as a single token and emitted as a void name.

## Suggested fix

Validate the token in `mapParam` (after the `^` / `@` handling): reject a token whose first character is not `[a-z]`, or that contains any §2.3 `NAME` terminator (`,` `.` `|` `'` `:` `;` `!` `?` `[` `]` `{` `}` `(` `)`, space, tab, cactus), reusing the existing `ParseError` path, e.g.:

```java
} else if (!raw.matches("[a-z][^ \\t,.|':;!?\\[\\]{}()]*")) {
    throw new ParseError(span.line(), pos, "parameter names in voids must be NAME or @");
} else {
    mapped = raw;
}
```

A regression check: `[a,b]`, `[a.b]`, `[a?b]`, and a tab-separated `[a<tab>b]` must each parse to an error, while `[a b]` and `[@ x]` continue to parse.

gemshrine:
Fixes #6288.

A formation's void parameters were split on the ASCII space and each token used verbatim as the parameter name; `mapParam` checked only `^` and `@`. So `[a,b]`, `[a.b]`, `[a|b]`, `[a:b]`, `[a?b]`, and a tab-separated `[a<tab>b]` all emitted a void whose `name` is not a valid `NAME` (§2.3 terminators are `" \t,.|':;!?[]{}()"`).

The fix validates each token in `mapParam`: it must start with `[a-z]` and contain no `NAME` terminator, mirroring `Tokens.readName`, with an optional trailing `...` so varargs params still parse. Verified against `EoSyntax`:

```
[a b] [m2] [minus1 minus2] [foo-bar] [foo_bar] [@ x]  -> ok
[a b... c]  (varargs)                                 -> ok
[a,b] [a.b] [a?b] [a<tab>b] [Foo]                     -> error: parameter names in voids must be NAME or @
```

Unicode identifiers are preserved (they contain no terminator), so the existing `unicode.yaml` pack still parses. Full `EoSyntaxTest` (1052 cases) stays green. Added the `void-param-name.yaml` regression.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnFormation.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnFormation.java
@@ -310,17 +310,16 @@ final class LnFormation implements Line {
      * @return Emitted name
      */
     private static String mapParam(final String raw, final Span span, final int pos) {
-        if ("^".equals(raw)) {
+        final String mapped;
+        if ("@".equals(raw)) {
+            mapped = "φ";
+        } else if (raw.matches("[a-z][^ \\t,.|':;!?\\[\\]{}()]*(?:\\.\\.\\.)?")) {
+            mapped = raw;
+        } else {
             throw new ParseError(
                 span.line(), pos,
                 "parameter names in voids must be NAME or @"
             );
-        }
-        final String mapped;
-        if ("@".equals(raw)) {
-            mapped = "φ";
-        } else {
-            mapped = raw;
         }
         return mapped;
     }
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-typos/void-param-name.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-typos/void-param-name.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+line: 1
+message: |-
+  [1:1] error: 'parameter names in voids must be NAME or @'
+  [a,b] > obj
+   ^
+input: |
+  [a,b] > obj
+    5 > x


eo-parser/src/main/java/org/eolang/parser/LnFormation.java
```diff
@@ -310,17 +310,16 @@ private static List<String> params(
...
-            throw new ParseError(
-                span.line(), pos,
-                "parameter names in voids must be NAME or @"
-            );
-        }
         final String mapped;
         if ("@".equals(raw)) {
             mapped = "φ";
-        } else {
+        } else if (raw.matches("[a-z][^ \\t,.|':;!?\\[\\]{}()]*(?:\\.\\.\\.)?")) {
```
Thayorns, line 316: This does not quite mirror readName. After matching the name body, readName also rejects the cactus codepoint anywhere inside it, because that emoji is reserved for auto-names. The cactus is not a terminator, so it survives this character class, and a formation head written with one in a parameter is accepted here while the same name after a name suffix is refused. The tokens reaching mapParam come straight off a space split of the bracket body, so nothing upstream filters it first.

Thayorns (commented):
Folding the caret rejection into a single grammar check reads better than the special case it replaces, and keeping the trailing ellipsis outside the character class is what makes varargs still parse, since the class already excludes the dot. One thing beyond the line below: the terminator set is now spelled out a second time here as a character class, while Tokens owns it as a constant that readName reads through terminates. A change to the terminator list in section 2.3 would reach readName and leave void parameters on the old grammar.

eo-parser/src/main/java/org/eolang/parser/LnFormation.java
morphqdd, line 316: This regex doesn't exclude the cactus emoji, so `[a🌵b] > obj` is accepted as a valid void parameter name, even though `Tokens.readName()` rejects it ("cactus emoji is reserved for auto-names; not allowed in identifiers"). The PR description says this mirrors `Tokens.readName`, but it's a hand-copied character class that already drifted from it. Also worth noting: PR #6332 independently hardcodes this same NAME-grammar character class for meta directive names, in a different file — both would benefit from a single shared validator instead of two copies that can (and already did) diverge.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2000](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (25) hits-of-code; +1 to give you at least something.  Your running score is [+939](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (25) hits-of-code; +1 to give you at least something.  Your running score is [+3634](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6330, opened 2026-08-03
#6290: reject a malformed dotted atom signature

### LINKED ISSUE #6290: Atom signature `/sig` improperly handles malformed dotted forms: leading dot `/.x`, trailing dot `/Q.`, empty segment `/Q..x`
## What happens

Per the spec (R-3.10.10), a bare `/Q` and "the other malformed sigs (bare `/`, trailing dot `/Q.`, sigs starting with `.`)" are rejected. The parser correctly rejects `/Q` and `/`, but **accepts** an atom signature with a leading dot (`/.x`), a trailing dot (`/Q.`, `/Q.number.`), or an empty path segment (`/Q..x`, `/.`), emitting a malformed `sig`.

### Runtime evidence

Probed directly against `org.eolang.parser.EoSyntax` (head `[] > f <sig>`):

| sig | parser | spec (R-3.10.10) |
| --- | --- | --- |
| `/Q.number` | ok (control → promoted to `Φ.number`) | valid |
| `/A` | ok (control, type variable) | valid |
| `/Q` | ERROR: `atom signature requires a name` | rejected (correct) |
| `/` | ERROR: `atom signature requires a name` | rejected (correct) |
| `/Q.` | **ok** (emits `Φ.`) | **rejected** (trailing dot) |
| `/.x` | **ok** (emits `.x`) | **rejected** (starts with `.`) |
| `/Q.number.` | **ok** | **rejected** (trailing dot) |
| `/Q..x` | **ok** (emits `Φ..x`) | **rejected** (empty segment) |
| `/.` | **ok** | **rejected** |

## Root cause

`eo-parser/src/main/java/org/eolang/parser/Suffix.java`. `signature()` reads the raw token up to a terminator and validates only emptiness, a stray `?`, and a bare `Q`:

```java
final String raw = tail.substring(after, idx);
if (raw.indexOf('?') >= 0) { ... }
if ("Q".equals(raw)) { throw ... "atom signature requires a name"; }
return Suffix.typeAtom(raw, span, home + after);
```

and `typeAtom()` only applies the type-variable rule and promotes a leading `Q.`:

```java
if (first >= 'A' && first <= 'Z' && !raw.matches("[A-F]") && !raw.startsWith("Q.")) {
    throw ... "type variable must be one of A-F";
}
final String promoted;
if (raw.startsWith("Q.")) { promoted = "Φ".concat(raw.substring(1)); }
else { promoted = raw; }
```

Neither validates the dotted structure. `Q.` matches `startsWith("Q.")` so it is promoted with the trailing dot kept (`Φ.`); `.x` has first char `.`, which skips the `A`–`Z` arm and is returned verbatim; `Q..x` promotes to `Φ..x` with the empty segment intact.

## Suggested fix

After computing `raw`, reject a malformed dot structure — a leading dot, a trailing dot, or an empty segment — before promotion, e.g.:

```java
if (raw.startsWith(".") || raw.endsWith(".") || raw.contains("..")) {
    throw new ParseError(
        span.line(), home + after,
        "atom signature must be a dotted name with no leading, trailing, or empty segment"
    );
}
```

A regression check: `/.x`, `/Q.`, `/Q.number.`, and `/Q..x` must each parse to an error, while `/Q.number` and `/A` continue to parse.

gemshrine:
Fixes #6290.

`Suffix.signature()` validated only emptiness, a stray `?`, and a bare `Q`, and `typeAtom()` promoted a leading `Q.` without checking the dot structure. So a leading dot (`/.x`), a trailing dot (`/Q.`, `/Q.number.`), or an empty segment (`/Q..x`, `/.`) all parsed, emitting a malformed `sig` (`Φ.`, `.x`, `Φ..x`).

The fix rejects a raw signature that starts with `.`, ends with `.`, or contains `..`, before promotion. Verified against `EoSyntax`:

```
/Q.number  /A  /Q.number.power   -> ok
/Q.  /.x  /Q.number.  /Q..x  /.  -> error: atom signature must be a dotted name with no leading, trailing, or empty segment
```

Full `EoSyntaxTest` (1050 cases) stays green. Added the `atom-sig-dot.yaml` regression.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -619,6 +619,12 @@ final class Suffix {
                 "atom signature requires a name"
             );
         }
+        if (raw.startsWith(".") || raw.endsWith(".") || raw.contains("..")) {
+            throw new ParseError(
+                span.line(), home + after,
+                "atom signature must be a dotted name with no leading, trailing, or empty segment"
+            );
+        }
         return Suffix.typeAtom(raw, span, home + after);
     }
 
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-typos/atom-sig-dot.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-typos/atom-sig-dot.yaml
@@ -1,0 +1,11 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+line: 1
+message: |-
+  [1:8] error: 'atom signature must be a dotted name with no leading, trailing, or empty segment'
+  [] > f /Q.
+          ^
+input: |
+  [] > f /Q.


Thayorns (approved):
Checking the dot structure before the promotion is the right place for it: once the leading Q has been turned into the phi, the empty segment is no longer distinguishable from a name the writer meant. The three shapes the check covers are the three the emitter cannot represent. The pack pins the trailing dot only, so the leading dot and the doubled dot, which are separate arms of the same condition, are covered by reading rather than by the suite.

eo-parser/src/main/java/org/eolang/parser/Suffix.java
```diff
@@ -617,6 +617,12 @@ private static String signature(
                 "atom signature requires a name"
             );
         }
+        if (raw.startsWith(".") || raw.endsWith(".") || raw.contains("..")) {
```
yegor256, line 622: @gemshrine looks pretty narrow condition. what about this `raw`: `"Q---"`? still wrong, but your check misses it.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (17) hits-of-code; +4 to give you at least something.  Your running score is [+789](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-parser/src/main/java/org/eolang/parser/Suffix.java
morphqdd, line 622: This check is only wired into `signature()`, the atom-signature path. `typeAtom()` itself — the shared promotion routine that has the same `Q.` -> `Φ.` promotion issue — is called directly from `LnVoid.type()`/`LnVoid.members()` too, for the void type annotation (`x /Q.`) and `/{...}` argument-list forms. Tracing `x /Q.` through `LnVoid.type()`: raw = "Q.", `typeAtom` promotes it to "Φ." with no rejection — same malformed-dot bug, different entry point. Moving this check into `typeAtom()` itself (or calling it from all three sites) would close it everywhere instead of just here.

eo-parser/src/test/resources/org/eolang/parser/eo-typos/atom-sig-dot.yaml
```diff
@@ -0,0 +1,11 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
```
morphqdd, line 1: Only one malformed shape (`/Q.`) gets a regression file here. The issue and PR body mention four (`/.x`, `/Q.`, `/Q.number.`, `/Q..x`) — worth one file per shape, matching this directory's one-case-per-file convention, so a regression on the leading-dot or empty-segment branch of the `||` condition doesn't go uncaught.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2103](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6331, opened 2026-08-03
#6293: validate a >> file-local handle starts lowercase like > name

### LINKED ISSUE #6293: File-local handle `>> name` is not validated to start lowercase (`? >> H`, `? >> 1x` accepted)
## What happens

R-3.10.12: a `>>` auto-name suffix may carry a *file-local handle*, which is a `NAME`; per §2.3 a `NAME` starts with `[a-z]`. The `> name` suffix enforces this, but the `>> handle` path does not — an uppercase-first handle (`? >> H`, `? >> Foo`) or a digit-first handle (`? >> 1x`) is accepted, though the parallel `> name` suffix rejects the same names.

### Runtime evidence

Probed against `org.eolang.parser.EoSyntax`:

| input | parser |
| --- | --- |
| `5 > FOO` | ERROR (control: `> name` rejects an uppercase-first name) |
| `? >> H` | **ok** |
| `? >> Foo` | **ok** |
| `? >> 1x` | **ok** (digit-first) |
| `? >> h` | ok (control: a valid handle) |

## Root cause

`eo-parser/src/main/java/org/eolang/parser/Suffix.java`. The `> name` path, `named()`, validates the first character:

```java
if (name.codePoints().anyMatch(cp -> cp == 0x1F335)) { ... }
Suffix.checkLowercaseStart(name, span, home, begin);
```

The `>> handle` path, `auto()`, reads the handle the same way but never calls `checkLowercaseStart` — it checks only `SCOPES` (the scope-token refusal of #6197) and the cactus:

```java
int rest = Suffix.skipName(tail, begin);
final String handle = tail.substring(begin, rest);
if (Suffix.SCOPES.contains(handle)) { ... }
if (handle.codePoints().anyMatch(cp -> cp == 0x1F335)) { ... }
// no checkLowercaseStart — H / Foo / 1x all pass
```

## Suggested fix

Mirror `named()` and validate a non-empty handle's first character (a bare `? >>` has an empty handle and must stay valid):

```java
if (!handle.isEmpty()) {
    Suffix.checkLowercaseStart(handle, span, home, begin);
}
```

A regression check: `? >> H`, `? >> Foo`, and `? >> 1x` must each parse to an error, while `? >> h` and a bare `? >>` continue to parse.

gemshrine:
Fixes #6293.

The `> name` suffix validates the first character via `checkLowercaseStart`, but the `>> handle` path (`auto()`) only refused scope tokens and the cactus emoji, so an uppercase-first handle (`? >> H`, `? >> Foo`) or a digit-first handle (`? >> 1x`) was accepted even though the parallel `> name` rejects the same names.

The fix mirrors `named()` and calls `checkLowercaseStart` on a non-empty handle, leaving a bare `? >>` (empty handle) valid. Verified against `EoSyntax`:

```
? >> h   ? >>   -> ok
? >> H   ? >> Foo   ? >> 1x   -> error: name must start with a lowercase letter
```

Full `EoSyntaxTest` (1050 cases) stays green. Added the `auto-handle-lowercase.yaml` regression.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -482,6 +482,9 @@ final class Suffix {
                 "cactus emoji is reserved for auto-names; not allowed in identifiers"
             );
         }
+        if (!handle.isEmpty()) {
+            Suffix.checkLowercaseStart(handle, span, home, begin);
+        }
         if (!cnst && rest < tail.length() && tail.charAt(rest) == '!') {
             cnst = true;
             rest = rest + 1;
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-typos/auto-handle-lowercase.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-typos/auto-handle-lowercase.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+line: 2
+message: |-
+  [2:7] error: 'name must start with a lowercase letter'
+    ? >> H
+         ^
+input: |
+  [] > f
+    ? >> H


eo-parser/src/main/java/org/eolang/parser/Suffix.java
```diff
@@ -482,6 +482,9 @@ private static Parsed auto(
                 "cactus emoji is reserved for auto-names; not allowed in identifiers"
             );
         }
+        if (!handle.isEmpty()) {
```
Thayorns, line 485: The guard is already inside the method you are calling. checkLowercaseStart opens on the empty name and returns, and its javadoc says outright that the name may be empty. So the bare double-angle case was safe without this, and the contract now lives in two places that have to be kept agreeing.

Thayorns (commented):
Sending the handle through the same check the name suffix uses is the right way to close this, since the two spellings are meant to name the same thing and were disagreeing about the first character. Putting it after the cactus check keeps the more specific message winning for the emoji.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (15) hits-of-code; +1 to give you at least something.  Your running score is [+801](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (15) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1728](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6332, opened 2026-08-03
#6294: validate a meta directive name against the NAME grammar

### LINKED ISSUE #6294: Parser accepts invalid uppercase and terminator characters in meta directive names
## What happens

Per the spec (§2.3 lexical table: `META = '+' NAME ...`), a meta directive's name is a `NAME`, which starts with `[a-z]` and excludes the terminator characters. The parser validates only that the name is non-empty, so a meta name that starts with an uppercase letter (`+Package`, `+PACKAGE`) or contains a `NAME` terminator such as `.` (`+p.x`) is accepted.

### Runtime evidence

Probed against `org.eolang.parser.EoSyntax`:

| meta line | parser |
| --- | --- |
| `+package foo` | ok (control) |
| `+alias foo` | ok (control) |
| `+Package foo` | **ok** (uppercase first char) |
| `+PACKAGE foo` | **ok** |
| `+p.x foo` | **ok** (`.` inside the name) |

## Root cause

`eo-parser/src/main/java/org/eolang/parser/LnMeta.java`, `into()` takes the head as everything between `+` and the first space and checks only emptiness:

```java
final int space = body.indexOf(' ');
...
head = body.substring(1, space);   // (or body.substring(1) when there are no parts)
...
if (head.isEmpty()) {
    throw new ParseError(..., "meta directive requires a name");
}
```

The head is never validated against the `NAME` grammar (first char `[a-z]`, no terminator characters). By contrast the `> name` suffix validates the first char via `Suffix.checkLowercaseStart`.

Note the meta **parts** are intentionally free-form (R-3.2 / §2.3: "each part is one or more non-whitespace characters"), but the **head** is specified as `NAME`, not a free-form part — so the head should be validated even though the parts are not.

## Suggested fix

After the emptiness check, validate `head` against the `NAME` grammar — first character `[a-z]`, and no `NAME` terminator characters — reusing the existing name validation. A regression check: `+Package`, `+PACKAGE`, and `+p.x` must each parse to an error, while `+package`, `+alias`, and `+rt jvm a.b.c` continue to parse.

gemshrine:
Fixes #6294.

`LnMeta.into()` took the meta head (between `+` and the first space) and checked only that it is non-empty, so a head starting with an uppercase letter (`+Package`, `+PACKAGE`) or containing a NAME terminator such as `.` (`+p.x`) was accepted, even though §2.3 defines the head as a `NAME`.

The fix validates the head against the NAME grammar (first char `[a-z]`, no terminators) after the emptiness check, leaving the free-form meta parts untouched. Verified against `EoSyntax`:

```
+package foo   +alias foo   +rt jvm a.b.c   -> ok
+Package foo   +PACKAGE foo   +p.x foo       -> error: meta name must be a NAME starting with a lowercase letter
```

Full `EoSyntaxTest` (1050 cases) stays green. Added the `meta-name-not-lowercase.yaml` regression.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnMeta.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnMeta.java
@@ -69,6 +69,20 @@ final class LnMeta implements Line {
                 body.substring(space + 1), this.span, space + 1
             );
         }
+        this.checkHead(head, parts);
+        Comments.seal(globals, emit, this.span);
+        globals.markMeta();
+        globals.clearBlanks();
+        emit.meta(this.span.line(), head, parts);
+    }
+
+    /**
+     * Validate the meta directive name against the NAME grammar and the
+     * {@code +package} arity rule.
+     * @param head The meta directive name
+     * @param parts The directive arguments
+     */
+    private void checkHead(final String head, final List<String> parts) {
         if (head.isEmpty()) {
             throw new ParseError(
                 this.span.line(), this.span.indent(),
@@ -87,10 +101,6 @@ final class LnMeta implements Line {
                 "'+package' directive requires exactly one argument"
             );
         }
-        Comments.seal(globals, emit, this.span);
-        globals.markMeta();
-        globals.clearBlanks();
-        emit.meta(this.span.line(), head, parts);
     }
 
     /**
--- a/eo-parser/src/main/java/org/eolang/parser/LnMeta.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnMeta.java
@@ -75,6 +75,12 @@ final class LnMeta implements Line {
                 "meta directive requires a name"
             );
         }
+        if (!head.matches("[a-z][^ \\t,.|':;!?\\[\\]{}()]*")) {
+            throw new ParseError(
+                this.span.line(), this.span.indent() + 1,
+                "meta name must be a NAME starting with a lowercase letter"
+            );
+        }
         if ("package".equals(head) && parts.isEmpty()) {
             throw new ParseError(
                 this.span.line(), this.span.indent(),
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-typos/meta-name-not-lowercase.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-typos/meta-name-not-lowercase.yaml
@@ -1,0 +1,15 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+line: 3
+message: |-
+  [3:1] error: 'meta name must be a NAME starting with a lowercase letter'
+  +Package foo
+   ^
+input: |
+  # doc.
+
+  +Package foo
+
+  [] > f


eo-parser/src/main/java/org/eolang/parser/LnMeta.java
```diff
@@ -75,6 +75,12 @@ public void into(final Stack stack, final Globals globals, final Emit emit) {
                 "meta directive requires a name"
             );
         }
+        if (!head.matches("[a-z][^ \\t,.|':;!?\\[\\]{}()]*")) {
```
Thayorns, line 92: The grammar is copied rather than reused. Tokens holds the terminator characters in a private constant and reads them through terminates, and readName holds the lowercase-first rule; this pattern is a third statement of both, in a class that has no way of hearing about a change to either. If section 2.3 gains or loses a terminator, meta heads keep the old set and nothing fails to say so.

Thayorns (commented):
Validating the head where the emptiness check already sits is the right spot, before anything is sealed or marked, so a bad directive cannot leave half its effects behind. Leaving the tail free-form is right too, since the parts are not names.

eo-parser/src/main/java/org/eolang/parser/LnMeta.java
morphqdd, line 92: This regex doesn't exclude the cactus emoji, so `+p🌵x foo` is accepted as a valid meta name. `Suffix.test()` in this same package validates a NAME token and does check for it (rejects with "cactus emoji is reserved for auto-names; not allowed in identifiers"), right next to the lowercase-start check this PR says it borrowed from. Also: PR #6329 independently hardcodes this identical character class for void parameter names, in a different file, and it has the same gap — two hand-copied regexes doing the same job, already drifted from the canonical `Tokens`/`Suffix` name-scanning logic. Worth factoring into one shared validator.

yegor256:
@gemshrine merge conflict here. also, please, take into account reviewer comments above

gemshrine:
@yegor256 rebased and resolved the conflict. I replaced the duplicated meta-name regex with the shared `Tokens.name()` validation, which now also rejects the cactus emoji; the existing reader uses the same validator. Added coverage for uppercase and cactus names. All CI jobs are green.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (39) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2115](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1094](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+3668](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6333, opened 2026-08-03
#6295: merge a named multi-line BYTES literal instead of dropping chunks

### LINKED ISSUE #6295: Final chunk name breaks the merging of multi-line BYTES literal
## What happens

A multi-line `BYTES` literal (chunks joined by a trailing `-` per R-3.13) merges correctly when it is **unnamed**, but when the **final chunk carries a `> name` suffix** the merge stops one chunk early: the parser emits the preceding chunks as a separate `BYTES` object (keeping a dangling trailing `-`) and binds only the final chunk's bytes to the name, silently dropping the earlier chunks from the named value.

### Runtime evidence

Probed against `org.eolang.parser.EoSyntax`.

Unnamed (the form the tests cover) merges into one token:

```
foo > bar
  CA-FE-
  BE-BE
```
→ one `Φ.bytes` = `CA-FE-BE-BE` (matches `EoTest.mergesMultiLineBytesAcrossTwoLines`).

With a name on the final chunk it splits:

```
foo > bar
  CA-FE-
  BE-BE > ml
```
→ **two** objects: a `Φ.bytes` holding `CA-FE-` (dangling dash) and a separate `Φ.bytes name="ml"` holding only `BE-BE`. The intended `CA-FE-BE-BE` named `ml` is never produced — the `CA-FE` chunk is lost from `ml`.

As a direct formation child it is worse — it errors outright:

```
[] > main
  CA-FE-
  BE-BE > ml
```
→ `object inside formation must have a name` on the `CA-FE-` line (the preceding chunk becomes a nameless plain child), so a named multi-line `BYTES` literal cannot be written as a plain child at all.

The `all-data-types.yaml` fixture uses `31-A2-B3-C4 > ml` on its final chunk but only asserts `//o[@base='Φ.bytes' and @name='ml']` — that *a* bytes named `ml` exists, not its content — so it does not catch the dropped chunks.

## Root cause

`eo-parser/src/main/java/org/eolang/parser/Eo.java`, `mergeBytesContinuation` appends continuation lines only while `isBytesOnly(next.body())`:

```java
final Span next = spans.get(idx);
if (next.indent() < head.indent() || !Eo.isBytesOnly(next.body())) {
    break;
}
body.append(next.body());
```

`isBytesOnly` requires the whole line to be `HH-HH...` with no suffix, so a final chunk carrying ` > ml` is not bytes-only: the loop breaks **before** appending it, leaving the `> name` chunk as a separate line and the accumulated body with its dangling trailing `-`.

## Suggested fix

Allow the final continuation chunk to carry the name suffix: when a line is a byte run followed by a `> name` / `>>` suffix, append its byte portion, close the token, and apply the suffix to the merged `BYTES` object (so `CA-FE-` / `BE-BE > ml` becomes one `Φ.bytes` = `CA-FE-BE-BE` named `ml`). A regression check should assert the **content** of a named multi-line `BYTES` token, e.g. `CA-FE-` / `BE-BE > ml` yields `<o base='Φ.bytes' name='ml'>CA-FE-BE-BE</o>`.

gemshrine:
Fixes #6295.

`mergeBytesContinuation` appended continuation lines only while `isBytesOnly(next.body())` held, so a final chunk carrying a `> name` suffix (`BE-BE > ml`) was not bytes-only: the loop broke before appending it, emitting the preceding chunks as a separate `Φ.bytes` with a dangling trailing `-` and binding only the last chunk to the name. As a plain formation child it errored outright (`object inside formation must have a name`).

The fix lets the loop take a terminal chunk that is a byte run followed by a `> name` / `>>` suffix: it appends the whole chunk and stops, so `CA-FE-` / `BE-BE > ml` is processed as the single line `CA-FE-BE-BE > ml`. Verified against `EoSyntax`:

```
[] > main / CA-FE- / BE-BE > ml  -> <o base='Φ.bytes' name='ml'><o>CA-FE-BE-BE</o></o>
```

Full `EoTest` (77) and `EoSyntaxTest` (1050) stay green. Added `mergesNamedMultiLineBytesWithSuffixOnFinalChunk`, which asserts the merged content (`CA-FE-BE-BE`) of the named token, not merely that a bytes named `ml` exists.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/BytesLine.java
+++ b/eo-parser/src/main/java/org/eolang/parser/BytesLine.java
@@ -1,0 +1,59 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+
+package org.eolang.parser;
+
+/**
+ * A single source line body examined as a possible {@code BYTES} hex-run
+ * ({@code HH-HH} or {@code HH-HH-} etc.), used by the {@link Eo} classifier
+ * and its multi-line {@code BYTES} merging.
+ * @since 0.57.0
+ */
+final class BytesLine {
+
+    /**
+     * The line body.
+     */
+    private final String body;
+
+    /**
+     * Ctor.
+     * @param line The line body to examine
+     */
+    BytesLine(final String line) {
+        this.body = line;
+    }
+
+    /**
+     * Whether the body is purely a sequence of hex bytes with dash
+     * separators — {@code HH-HH} or {@code HH-HH-} etc. Per the grammar
+     * ({@code BYTE : [0-9A-F][0-9A-F]}), only uppercase hex counts here,
+     * since lowercase letters belong to {@code NAME} tokens.
+     * @return True if the body matches the bytes-only pattern
+     */
+    boolean onlyBytes() {
+        boolean valid = !this.body.isEmpty();
+        int idx = 0;
+        while (valid && idx < this.body.length()) {
+            if (idx + 1 >= this.body.length()
+                || !BytesLine.hex(this.body.charAt(idx))
+                || !BytesLine.hex(this.body.charAt(idx + 1))) {
+                valid = false;
+            } else {
+                idx = idx + 2;
+                if (idx < this.body.length() && this.body.charAt(idx) != '-') {
+                    valid = false;
+                } else {
+                    idx = idx + 1;
+                }
+            }
+        }
+        return valid;
+    }
+
+    private static boolean hex(final char glyph) {
+        return glyph >= '0' && glyph <= '9' || glyph >= 'A' && glyph <= 'F';
+    }
+}
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -193,7 +193,9 @@ final class Eo implements Iterable<Directive> {
                 break;
             }
             above = next.indent();
-            if (!Eo.isBytesOnly(trimmed)) {
+            final boolean bytes = new BytesLine(trimmed).onlyBytes();
+            final boolean named = Eo.namedTerminator(body, trimmed);
+            if (!bytes && !named) {
                 emit.error(
                     next.line(), 0, "multi-line bytes interrupted by non-byte content"
                 );
@@ -202,7 +204,7 @@ final class Eo implements Iterable<Directive> {
             }
             body.append(trimmed);
             idx = idx + 1;
-            if (!Eo.isBytesContinuation(trimmed)) {
+            if (named || !Eo.isBytesContinuation(trimmed)) {
                 break;
             }
         }
@@ -220,33 +222,24 @@ final class Eo implements Iterable<Directive> {
         return resumption;
     }
 
+    private static boolean namedTerminator(final CharSequence body, final String chunk) {
+        return body.length() > 0
+            && body.charAt(body.length() - 1) == '-'
+            && Eo.namedChunk(chunk);
+    }
+
+    private static boolean namedChunk(final String body) {
+        final int space = body.indexOf(' ');
+        return space > 0
+            && new BytesLine(body.substring(0, space)).onlyBytes()
+            && body.substring(space + 1).startsWith(">");
+    }
+
     private static boolean isBytesContinuation(final String body) {
         final String trimmed = body.stripTrailing();
-        return trimmed.length() >= 6 && trimmed.endsWith("-") && Eo.isBytesOnly(trimmed);
-    }
-
-    private static boolean isBytesOnly(final String body) {
-        boolean valid = !body.isEmpty();
-        int idx = 0;
-        while (valid && idx < body.length()) {
-            if (idx + 1 >= body.length()
-                || !Eo.hex(body.charAt(idx))
-                || !Eo.hex(body.charAt(idx + 1))) {
-                valid = false;
-            } else {
-                idx = idx + 2;
-                if (idx < body.length() && body.charAt(idx) != '-') {
-                    valid = false;
-                } else {
-                    idx = idx + 1;
-                }
-            }
-        }
-        return valid;
-    }
-
-    private static boolean hex(final char glyph) {
-        return glyph >= '0' && glyph <= '9' || glyph >= 'A' && glyph <= 'F';
+        return trimmed.length() >= 6
+            && trimmed.endsWith("-")
+            && new BytesLine(trimmed).onlyBytes();
     }
 
     private static boolean process(
--- a/eo-parser/src/test/java/org/eolang/parser/BytesLineTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/BytesLineTest.java
@@ -1,0 +1,43 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.parser;
+
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link BytesLine}.
+ * @since 0.57.0
+ */
+final class BytesLineTest {
+
+    @Test
+    void acceptsBytesOnlyRun() {
+        MatcherAssert.assertThat(
+            "a dash-separated hex run must be recognised as bytes-only",
+            new BytesLine("CA-FE-BE-BE").onlyBytes(),
+            Matchers.is(true)
+        );
+    }
+
+    @Test
+    void rejectsNonBytesRun() {
+        MatcherAssert.assertThat(
+            "a lowercase word must not be recognised as bytes-only",
+            new BytesLine("hello").onlyBytes(),
+            Matchers.is(false)
+        );
+    }
+
+    @Test
+    void rejectsLowercaseHexRun() {
+        MatcherAssert.assertThat(
+            "lowercase hex letters belong to NAME, not to BYTES",
+            new BytesLine("ca-fe").onlyBytes(),
+            Matchers.is(false)
+        );
+    }
+}
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/named-multi-line-bytes.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/named-multi-line-bytes.yaml
@@ -1,0 +1,16 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# #6295: a name on the final chunk of a multi-line BYTES literal applies to the
+# whole merged token, instead of cutting the merge short and dropping the chunks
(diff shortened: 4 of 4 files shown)


Thayorns (commented):
Letting the loop take one terminal chunk and stop is the right shape for this, and keeping the earlier requirement that the accumulated body still ends with a dash means a suffix cannot attach itself to a run that already closed. The test asserting the merged content rather than the mere presence of a token named ml is what makes it a real guard. The javadoc promises the double-angle spelling as well as the named one; only the named one has a case.

eo-parser/src/main/java/org/eolang/parser/Eo.java
```diff
@@ -234,6 +243,21 @@ private static int mergeBytesContinuation(
...
+     * literal — a byte run followed by a {@code > name} / {@code >>}
+     * suffix (e.g. {@code BE-BE > ml}). Such a chunk terminates the
+     * literal and applies its suffix to the merged {@code BYTES} object.
+     * @param body The line body
+     * @return True if the body is a byte run carrying a name suffix
+     */
+    private static boolean isNamedBytesChunk(final String body) {
+        final int space = body.indexOf(' ');
+        return space > 0
+            && Eo.isBytesOnly(body.substring(0, space))
```
Thayorns: The byte run in front of the suffix is allowed to end with a dash of its own. isBytesOnly accepts BE- as readily as BE-BE, so a head of CA-FE- followed by BE- > ml merges into CA-FE-BE- > ml and the loop stops on it. The trailing dash is the mark that says another chunk follows, and none does, which is the dangling-delimiter shape this pull request exists to remove on the unnamed path. Is the terminal chunk meant to be allowed to end on a delimiter?

eo-parser/src/main/java/org/eolang/parser/Eo.java
```diff
@@ -216,12 +216,21 @@ private static int mergeBytesContinuation(
...
             final Span next = spans.get(idx);
-            if (next.indent() < head.indent() || !Eo.isBytesOnly(next.body())) {
+            if (next.indent() < head.indent()) {
                 break;
             }
-            body.append(next.body());
-            idx = idx + 1;
-            if (!next.body().endsWith("-")) {
+            if (Eo.isBytesOnly(next.body())) {
+                body.append(next.body());
```
morphqdd: `body.append(next.body()); idx = idx + 1;` is duplicated identically in both branches (here and line 230-231) — the two branches differ only in their break condition. Worth extracting the append-and-advance step into one shared helper called from both.

eo-parser/src/test/java/org/eolang/parser/EoTest.java
```diff
@@ -882,6 +882,17 @@ void mergesMultiLineBytesAcrossThreeLines() {
         );
     }
 
+    @Test
+    void mergesNamedMultiLineBytesWithSuffixOnFinalChunk() {
```
morphqdd: Only the direct-formation-child repro from the issue (2 chunks) has a test. The issue also describes a `foo > bar` binding case where the merge silently splits into two `Φ.bytes` objects instead of erroring — that path isn't covered. Neither is a 3+-chunk case (I traced it by hand and the fixed loop does handle it correctly, but nothing locks that in with a test).

yegor256:
@gemshrine please, make sure all CI jobs pass (including qulice)

yegor256:

gemshrine:

eo-parser/src/main/java/org/eolang/parser/BytesLine.java
```diff
@@ -0,0 +1,57 @@
...
+ * Low-level predicates over the {@code BYTES} hex-run format, shared by
+ * the {@link Eo} classifier and its multi-line {@code BYTES} merging.
+ * @since 0.57.0
+ */
+final class BytesLine {
+
+    /**
+     * Utility class.
+     */
+    private BytesLine() {
```
yegor256: @gemshrine read this: https://www.yegor256.com/2014/05/05/oop-alternative-to-utility-classes.html

gemshrine: `BytesLine` is now a proper object: `new BytesLine(body).isBytesOnly()`, with `hex` as a private helper. No more static utility class.

gemshrine: Turning `BytesLine` into a real object bumps `Eo`'s ATFD from 5 to 7, which trips PMD's GodClass on `Eo` (WMC=139, ATFD=7, TCC=0). The three multi-line BYTES helpers (`isBytesContinuation`, `isNamedBytesChunk`, `mergeBytesContinuation`) are the only callers, so I can pull them out into a small dedicated class that owns `BytesLine` and let `Eo` delegate, which keeps the object and drops `Eo`'s ATFD back under the threshold. Do you prefer that extraction, or should the byte-run check stay a private helper inside `Eo`?

gemshrine:
Rebased on the latest master and the qulice failure is gone.

The GodClass hit came from the object conversion itself: `Eo` sits at WMC=139 and TCC=0 already, so ATFD was the only criterion keeping it under the threshold, and calling `isBytesOnly()` on a foreign object pushed it from 5 to 7. PMD counts getter-shaped calls (`is*`/`get*`) on foreign objects, and the whole `eo-parser` package names its boolean methods the other way round: `terminates`, `blank`, `constant`, `atom`, `tabbed`, `tupled`. Only three methods in the package carried an `is` prefix, two of them mine. So I renamed `isBytesOnly` to `onlyBytes` and `isNamedBytesChunk` to `namedChunk`, which lines the new code up with the package convention and drops ATFD back to the 5 it had before this branch. Verified locally: qulice on `eo-parser` is green, `EoSyntaxTest` 1098/0, `EoTest` 80/0, `BytesLineTest` 3/0, jtcop clean, pr-size 189.

Worth saying plainly: `Eo` is still a god class by the other two numbers, and any future change that adds one foreign getter call to it will trip this again. Extracting the multi-line BYTES concern into its own class is the real fix, but moving that much pre-existing code out of `Eo` costs well over the 200-line pr-size limit for this pull request. Happy to do it as a separate one if you want it.

eo-parser/src/main/java/org/eolang/parser/Eo.java
gemshrine: No, and the parser already refuses it. I ran `CA-FE-` followed by `BE- > ml` through `Eo` and it comes back as `[2:10] error: 'bytes literal ends with a dangling continuation dash'`, which is the same error the unnamed `CA-FE-` + `BE-` pair produces. The merge builds `CA-FE-BE-` and the existing dangling-delimiter check rejects it downstream, so the named path inherits that rule rather than opening a hole in it. The control case `CA-FE-` + `BE-BE > ml` parses into one `Φ.bytes` named `ml`, as intended.

eo-parser/src/main/java/org/eolang/parser/Eo.java
gemshrine: Restructured the loop so the append-and-advance happens once. The two branches now only compute whether the chunk continues the run or terminates it with a name, and the shared tail appends and advances:

```java
final boolean bytes = new BytesLine(chunk).onlyBytes();
final boolean named = !bytes && body.toString().endsWith("-") && Eo.namedChunk(chunk);
if (!bytes && !named) {
    break;
}
body.append(chunk);
idx = idx + 1;
if (named || !chunk.endsWith("-")) {
    break;
}
```

eo-parser/src/test/java/org/eolang/parser/EoTest.java
gemshrine: Widened the test to three chunks: `CA-FE-` + `BE-BE-` + `01-02 > ml` must come out as one `Φ.bytes` named `ml` holding `CA-FE-BE-BE-01-02`. That locks in the multiple-iteration path you traced by hand, and it still fails on master, where the run breaks at the named chunk and leaves a dangling dash. It costs no extra lines, which matters here because the pull request sits at 189 of the 200 the pr-size gate allows.

The `foo > bar` binding case I left out on purpose: it needs a call on how a name on a non-BYTES continuation should behave, which is a different question from the chunk dropping this fixes, and there is no room in this pull request to answer it properly. Better as its own issue.

yegor256:
@gemshrine conflicts here

gemshrine:
@yegor256 rebased on master, conflicts resolved. The clash was in `isBytesContinuation`: master reworded the `hex` javadoc while this branch moves `hex` and the bytes-only scan out into `BytesLine`, so I kept the extraction and dropped the now-dead private methods from `Eo`. eo-parser tests green locally: EoSyntaxTest 1098/0, EoTest 80/0, BytesLineTest 3/0.

yegor256:
@gemshrine conflicts here

gemshrine:
@yegor256 rebased, conflicts resolved.

The clash was in `mergeBytesContinuation`: master started stripping trailing whitespace off every chunk, while this branch moves the bytes-only scan into `BytesLine` and adds the named-chunk case. Both are in now, the trimmed chunk is what gets tested and appended. `EoTest`, `BytesLineTest`, `SpanTest` and the full `EoSyntaxTest` are green locally.

gemshrine:
@yegor256 rebased and conflict free now, please merge when the checks go green.

yegor256:
@gemshrine conflicts here again

gemshrine:
@yegor256 rebased again, conflicts gone. Master meanwhile turned a non-byte continuation into a parse error, so the named chunk is now checked before that error fires and the rest of the loop is master's. Parser tests are green locally.

gemshrine:
qulice caught `mergeBytesContinuation` at cyclomatic complexity 12 after the rebase, since master added two error branches there and this branch adds the named-chunk one. Pulled the de-indent check and the named-terminator test into their own methods, which brings it back under the limit without changing behavior.

The other violation in that run, `Emissions.java` God Class, is on master too since the qulice 0.31.2 upgrade, so this branch cannot go fully green until master does.

yegor256:

---

yegor256:
This has merge conflicts with master now. Needs a rebase.

---

gemshrine:
@yegor256 rebased. master extracted BytesIndent in the meantime, so the branch is rebuilt on top of it: BytesLine plus the named-chunk check, one commit instead of two. Green.

yegor256:
I am sorry, we are closing this pull request because it has merge conflicts with the current master branch and has been open for a while. Please forgive us for the inconvenience. If you are still interested, we would welcome a new pull request rebased against the latest master — it may even be simpler to redo it fresh given the changes that have landed since this PR was opened. Thank you for your contribution!

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+3943](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+6176](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+3331](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6334, opened 2026-08-03
#6302: inherit the default BiConsumer.andThen in CentralMaven

### LINKED ISSUE #6302: `CentralMaven` improperly overrides `BiConsumer.andThen`, causing failure in ordinary `BiConsumer` composition
## What happens

`CentralMaven` implements `BiConsumer<Dependency, Path>`, but overrides the interface default method as:

```java
@Override
public BiConsumer<Dependency, Path> andThen(
    final BiConsumer<? super Dependency, ? super Path> after
) {
    throw new UnsupportedOperationException("not implemented #andThen()");
}
```

As a result, ordinary `BiConsumer` composition fails before either consumer is invoked:

```java
final BiConsumer<Dependency, Path> combined = central.andThen(after);
```

The Java contract for `andThen` requires a new consumer that first calls the receiver and then `after` with the same arguments. `CentralMaven.accept(...)` is a normal side-effecting operation, so callers cannot safely use it where a `BiConsumer` is expected.

## Expected behavior

`CentralMaven.andThen` should preserve the standard `BiConsumer` behavior. The simplest fix is to remove the override and inherit the default implementation, or delegate to `BiConsumer.super.andThen(after)`.

## Suggested coverage

Add a unit test with two recording consumers that verifies both are invoked in order with the same dependency and destination. This protects against restoring the placeholder implementation later.

gemshrine:
Fixes #6302.

`CentralMaven implements BiConsumer<Dependency, Path>` but overrode `andThen` to `throw new UnsupportedOperationException`, so ordinary `BiConsumer` composition failed at the `central.andThen(after)` call, before either consumer could run. `accept(...)` is a normal side-effecting operation, so there is no reason to forbid composition.

The fix removes the override and inherits the correct default `BiConsumer.andThen`. Added `composesWithAndThenInsteadOfThrowing`, which asserts the composition returns a consumer instead of throwing, guarding against restoring the placeholder.

### DIFF
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/CentralMavenTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/CentralMavenTest.java
@@ -9,6 +9,7 @@ import com.yegor256.MktmpResolver;
 import com.yegor256.WeAreOnline;
 import java.nio.file.Path;
 import java.util.Collections;
+import java.util.function.BiConsumer;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.repository.internal.MavenRepositorySystemUtils;
 import org.eclipse.aether.DefaultRepositorySystemSession;
@@ -115,6 +116,17 @@ final class CentralMavenTest {
             "Unpacked destination must contain files when using injected resolver components",
             dest.toFile().list(),
             Matchers.not(Matchers.emptyArray())
+        );
+    }
+
+    @Test
+    void composesWithAndThenInsteadOfThrowing() {
+        final BiConsumer<Dependency, Path> after = (dep, dest) -> {
+        };
+        MatcherAssert.assertThat(
+            "andThen must return a composed BiConsumer, not throw at composition time",
+            new CentralMaven().andThen(after),
+            Matchers.notNullValue()
         );
     }
 
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/CentralMaven.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/CentralMaven.java
@@ -205,13 +205,6 @@ final class CentralMaven implements BiConsumer<Dependency, Path> {
         }
     }
 
-    @Override
-    public BiConsumer<Dependency, Path> andThen(
-        final BiConsumer<? super Dependency, ? super Path> after
-    ) {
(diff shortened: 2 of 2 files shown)


Thayorns (commented):
Removing the override is clearly right: accept is an ordinary side effect and there was never a reason composition should be refused, least of all before either half runs. The interface default does exactly what a caller expects. The test builds its subject through the argument-free constructor, so it also silently exercises the local repository path and the repository system that constructor conjures, none of which the test is about.

eo-maven-plugin/src/test/java/org/eolang/maven/CentralMavenTest.java
```diff
@@ -118,6 +119,17 @@ void resolvesWithInjectedComponents(@Mktmp final Path temp) {
...
     }
 
+    @Test
+    void composesWithAndThenInsteadOfThrowing() {
+        final BiConsumer<Dependency, Path> after = (dep, dest) -> {
+        };
+        MatcherAssert.assertThat(
+            "andThen must return a composed BiConsumer, not throw at composition time",
+            new CentralMaven().andThen(after),
+            Matchers.notNullValue()
```
Thayorns, line 129: The assertion cannot say anything about composition. BiConsumer.andThen is a default method that returns a lambda unconditionally, so notNullValue holds for any implementation that does not throw, and the composed consumer is never handed a dependency and a path. A version that returned a consumer ignoring the first half, or the second, would pass here unchanged. What distinguishes the fixed class from the broken one is that both consumers run in order, and nothing here accepts anything.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (19) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1752](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (19) hits-of-code; +1 to give you at least something.  Your running score is [+837](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6336, opened 2026-08-03
feat(#4752): implement `string.printf` in pure EO

### LINKED ISSUE #4752: `sprintf` Functionality Not Properly Utilized in EO

asmirnov-backend:
Closes #4752.

Rewrites `string.printf` in pure EO and drops the native atom: `EOprintf.java`, its `PrintfArgs`
collaborator and `PrintfArgsTest`. The format is walked byte by byte, because `%` never occurs
inside a UTF-8 multi-byte sequence, and each specifier is parsed into flags, width, precision and
conversion, then rendered through the `string` objects that already exist. 44 inline tests cover
`%s`, `%d`, `%f`, `%x`, `%b`, `%%`, positional arguments, widths, precisions, the `-` and `0`
flags, and every failure path.

Two reusable pieces landed first, so this PR only composes them: `string.truncated` (#6282) does
the `.precision` cut and `string.padded-zeros` (#6283) does the `%0Nd` padding with the minus sign
kept in front of the zeros.

It also fixes a behavior gap: precision on `%s` was silently ignored, while the Java atom applied
it, so `"%.3s"` on `Jeffrey` now returns `Jef` as it should. Covered by
`can-apply-a-precision-to-a-string`.

`pr-size` is red at 999 lines and cannot be made green. The atom and the pure-EO object cannot
coexist for even one commit: with both present the transpiler emits
`org.eolang.EO_string.EOprintf` next to the handwritten class and the build dies with
`duplicate class`. So the implementation, the deletion and the `PhDefaultTest` fix have to land
together. Of the 999 lines, 627 are the deletion. This is the same shape as #5711, which rewrote
`string.scanf` in pure EO and dropped `EOscanf.java` with its Java tests in one 518-line PR, and
as #5744 for `i64.as-number`.

Review by @Thayorns is addressed: the five parsing positions are named after what stands
there rather than where the scan stopped, the docblock maps how `walk` and `spec` traverse
the format, and the nine failures now report the format, the argument index, the number that
overflowed and the conversion byte. That last one uncovered a defect none of the tests could
see: the messages never reached the reader at all, because `padded` dataizes `value` from
inside `padded-left`, so a failing specifier surfaced as `the 'x' attribute must be a number`.
It predates this branch. Dataizing `value` before the padding puts the message back on top.


eo-runtime/src/main/eo/string/printf.eo
```diff
@@ -1,26 +1,349 @@
...
+          byte i
+      flags-end
+        as-number.
+          i.plus 1
+      i
+  2.power 63 >> limit!
+  size. >> span!
+    bytes
+      format.as-bytes >> raw!
+  [index auto acc] >> spec
```
Thayorns, line 128: spec carries twenty attributes: arg, body, conv, converted, datum, filled, fstop, idx, numeric, padded, pdigits, positional, precise, precision, pstop, value, width, wstop, zero and the phi. Almost none of them can be read without several of the others, so understanding one branch means holding the whole specifier grammar at once. The two objects already lifted out into their own files show the shape that would help here: the flags, the width and the precision each parse a slice of the specifier independently of the conversion, and each could know only its own slice.

eo-runtime/src/main/eo/string/printf.eo
```diff
@@ -1,26 +1,349 @@
...
+    if. > @
+      gte.
+        number pstop
+        number span
+      T
```
Thayorns, line 133: None of the nine failures carry the value that caused them. This one does not show the format it ran off the end of; the unsupported conversion at 204 does not say which character it found; the out-of-bounds index at 309 names neither the index nor how many arguments arrived; and the same holds at 164, 184, 296, 303, 314 and 322. A program with a dozen printf calls gives the reader nothing to locate the bad one by. chunk.eo does it the other way in this same runtime, feeding the offending sizes through printf so the report carries the numbers.

Thayorns (commented):
Walking the format over bytes rather than characters is the right call and the reasoning holds: the percent sign cannot hide inside a multi-byte sequence, so the linear copy is safe. The failure paths are genuinely covered rather than merely claimed, with the stops-on examples reaching the unsupported conversion, the zero positional index, the oversized index and the nan. Pulling truncated and padded-zeros out into their own objects first was the right move. One consequence of deciding a type from bytes is worth stating in the docblock a little louder than it is now: a one-character string whose only byte is 00 or 01 is reported as a bool and pointed at the %b conversion, so a legitimate string carrying a single control character is refused by %s.

eo-runtime/src/main/eo/string/printf.eo
asmirnov-backend: Done, and the comment found more than it asked for.

The nine now carry the format, the argument index, the number that overflowed and the conversion byte. But carrying a value turned out not to be enough: none of the nine ever reached the reader. `padded` dataizes `value` from inside `padded-left`, so a failing specifier surfaced as `the 'x' attribute must be a number` out of `string.padded-left:29`, with the real message buried under it. I checked the same probe against the commit before this change and it behaves identically, so the loss predates the values. Dataizing `value` before the padding puts the message back on top, and `printf *1 "%s%s" "Jeff"` now reports `The argument index 2 in the format is out of bounds, only 1 given`.

The conversion byte is reported through `%x` rather than `%s` on purpose: the byte after `%` may be the head of a multi-byte character, and `%s` would then fail on invalid UTF-8 instead of naming the offender.

eo-runtime/src/main/eo/string/printf.eo
asmirnov-backend: Agreed on the reading cost, and I took the cheap half of it now: the five parsing positions were `body`, `fstop`, `wstop`, `pstop` and `pdigits`, which said where the scan stopped but not what stands there, and they are now `flags`, `digits`, `dot`, `end` and `dollar`, with a docblock mapping how `walk` and `spec` hand the format to each other.

The split into files is harder than `truncated` and `padded-zeros` were, and I measured rather than guessed. Those two each answer with one value, which is why they lift out cleanly. A flags/width/precision parser answers with two — the value and the position where it ended — and EO memoizes through `!` alone, which yields bytes, so an object cannot be cached, only a scalar. Every such parser is therefore recomputed per use, or has to return a tuple that the caller unpacks. I tried the largest piece that does answer with one value, the 56-line conversion block, as a top-level formation taking `arg conv cut datum`: `spec` drops from 218 to 171 lines and 44/44 stay green, but the suite goes from ~48s to 54-59s, a 10-20% cost. `-Deo.autoFix` also prints a nested formation back at its first use site, so a helper cannot simply be hoisted for readability.

I would rather not spend that here — the PR is at 999 lines and already red on `pr-size`. Happy to do it as a follow-up right after this merges, and the experiment is kept if you want to look at the shape. If you would rather have it in this PR despite the cost, say so and I will push it.

yegor256:
@asmirnov-backend awesome, thanks!

0crat:
@asmirnov-backend Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (999 >= 100); -8 for way too many hits-of-code (999 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+56](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +17 points for this: +12 as a basis; +9 for the 999 hits-of-code that you reviewed; -5 for very few (2) comments.  Your running score is [+854](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6339, opened 2026-08-03
#6314: fixed spurious R-6.5.3 error for test-attributes

### LINKED ISSUE #6314: Parser rejects first object test-attribute due to misinterpreting blank line requirement (R-6.5.3)
## What happens

When a `+>`/`-->` test-attribute formation (or only-phi-formation) is the very first object in a file - right after the meta header and its one mandatory blank line - the parser rejects it with a spurious R-6.5.3 error, even though exactly one blank line correctly precedes it.

```
+package foo

[] +> works
  true > @
```

This parses with one error:

```
[3:0] error: 'missing blank line before a `+>` test attribute (R-6.5.3); exactly one blank line must precede every test attribute'
```

Move the exact same construct one level deeper so it's no longer the file's first object, keeping identical spacing, and it parses cleanly:

```
+package foo

[] > holder

  [] +> works
    true > @
```

## Why

`LnFormation.into()` (`eo-parser/src/main/java/org/eolang/parser/LnFormation.java:50`) calls `Blanks.enterAfterMeta(...)` unconditionally as its first statement. When the meta header is still open and a blank is pending, `enterAfterMeta` (`Blanks.java:79-91`) satisfies R-6.5.5 by calling `globals.clearBlanks()` and closing the header. Further down the same method, `LnFormation.java:78-79` runs `Blanks.checkTest(...)` when the suffix is a test attribute; `checkTest` (`Blanks.java:62-69`) enforces R-6.5.3 by reading `globals.pendingBlanks()` - but that counter was already zeroed by the earlier `enterAfterMeta` call for the very same blank line. One physical blank line is being asked to prove two separate things, and the first check consumes the evidence before the second gets to look at it. `LnOnlyPhi.into()` (`LnOnlyPhi.java:79`, `LnOnlyPhi.java:124-125`) has the identical shape and the identical bug.

## Suggested fix

Have `checkTest` see the blank count from before `enterAfterMeta` clears it - e.g. read `pendingBlanks()` once at the top of `into()` (or have `enterAfterMeta` return whether it consumed a blank) and let the test-attribute check use that captured value instead of re-reading the now-cleared counter.

AhmedAliElashmawy:
Closes #6314 
## Description
This PR fixes a spurious `R-6.5.3` (missing blank line) error that occurs when a test-attribute (`+>` / `-->`) formation is the very first object in a file (immediately following the meta header and its mandatory blank line). 

**The issue:**
In `LnFormation.into()` and `LnOnlyPhi.into()`, `Blanks.enterAfterMeta(...)` is called unconditionally when the meta header is open, which zeroes out the pending blanks counter. Later in the same method, `checkTest` tries to read `globals.pendingBlanks()` to enforce `R-6.5.3`, but finds `0` because it was already cleared by `enterAfterMeta`, leading to a false positive error.

**The fix:**
1. Overloaded `checkTest` in `Blanks.java` to accept an explicit `blanks` count parameter, while preserving the existing signature for the other 5 usages.
2. In both `LnFormation.java` and `LnOnlyPhi.java`, the pending blanks count is now captured in a local variable at the top of the `into()` method (before `enterAfterMeta` clears it).This captured value is then passed to the new overloaded `checkTest` method, ensuring the test-attribute check uses the correct blank count.
3. Added `@SuppressWarnings("PMD.UnnecessaryLocalRule")` to the `into()` method in these classes. PMD incorrectly flags this variable as unnecessary, but we explicitly need to capture its value early before the counter is reset.

## Testing
Added `acceptsPlusGreaterAttributeRightAfterMeta()` in both `LnFormationTest` and `LnOnlyPhiTest`, covering the exact first-object-after-meta case from #6314. Existing tests for non-first-object attributes are unchanged and continue to pass.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Blanks.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Blanks.java
@@ -95,10 +95,12 @@ final class Blanks {
      * @param span The first post-meta line's span
      * @param globals The global parser state
      * @param emit The directives sink
+     * @return The pending blanks count before any potential clearance
      */
-    static void enterAfterMeta(final Span span, final Globals globals, final Emit emit) {
+    static int enterAfterMeta(final Span span, final Globals globals, final Emit emit) {
+        final int blanks = globals.pendingBlanks();
         if (globals.inMetaHeader()) {
-            if (globals.pendingBlanks() == 0) {
+            if (blanks == 0) {
                 emit.error(
                     span.line(), span.indent(),
                     "missing blank line between meta header and the first non-meta line (R-6.5.5); exactly one blank must separate them"
@@ -108,5 +110,6 @@ final class Blanks {
             }
             globals.closeMetaHeader();
         }
+        return blanks;
     }
 }
--- a/eo-parser/src/main/java/org/eolang/parser/Blanks.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Blanks.java
@@ -62,12 +62,6 @@ final class Blanks {
      * @param emit The directives sink
      */
     static void checkTest(final Span span, final Globals globals, final Emit emit) {
-        if (span.indent() != 2) {
-            emit.error(
-                span.line(), span.indent(),
-                "test attribute legal only as direct child of top-level object"
-            );
-        }
         checkTest(span, globals.pendingBlanks(), emit);
     }
 
@@ -80,6 +74,12 @@ final class Blanks {
      * @param emit The directives sink
      */
     static void checkTest(final Span span, final int blanks, final Emit emit) {
+        if (span.indent() > 2) {
+            emit.error(
+                span.line(), span.indent(),
+                "test attribute legal only as direct child of top-level object"
+            );
+        }
         if (blanks == 0) {
             emit.error(
                 span.line(), span.indent(),
--- a/eo-parser/src/main/java/org/eolang/parser/Blanks.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Blanks.java
@@ -68,7 +68,19 @@ final class Blanks {
                 "test attribute legal only as direct child of top-level object"
             );
         }
-        if (globals.pendingBlanks() == 0) {
+        checkTest(span, globals.pendingBlanks(), emit);
+    }
+
+    /**
+     * Report a missing blank line in front of a {@code +>} test
+     * attribute — illegal per R-6.5.3, which requires exactly one
+     * blank line before every test attribute.
+     * @param span The offending line's span (used for error position)
+     * @param blanks The number of blank lines preceding this attribute
+     * @param emit The directives sink
+     */
+    static void checkTest(final Span span, final int blanks, final Emit emit) {
+        if (blanks == 0) {
             emit.error(
                 span.line(), span.indent(),
                 "missing blank line before a `+>` test attribute (R-6.5.3); exactly one blank line must precede every test attribute"
--- a/eo-parser/src/main/java/org/eolang/parser/LnFormation.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnFormation.java
@@ -44,8 +44,14 @@ final class LnFormation implements Line {
         this.span = source;
     }
 
+    /**
+     * PMD complains 'blanks' is unnecessary, but we must capture it
+     * before enterAfterMeta() clears the counter for checkTest().
+    **/
+    @SuppressWarnings("PMD.UnnecessaryLocalRule")
     @Override
     public void into(final Stack stack, final Globals globals, final Emit emit) {
+        final int blanks = globals.pendingBlanks();
         Blanks.enterAfterMeta(this.span, globals, emit);
         final String body = this.span.body();
         final List<String> params;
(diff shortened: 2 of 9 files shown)


eo-parser/src/main/java/org/eolang/parser/LnFormation.java
```diff
@@ -45,8 +45,14 @@ final class LnFormation implements Line {
         this.span = source;
     }
 
+    /*
+     * PMD complains 'blanks' is unnecessary, but we must capture it
+     * before enterAfterMeta() clears the counter for checkTest().
+     */
+    @SuppressWarnings("PMD.UnnecessaryLocalRule")
```
yegor256: @AhmedAliElashmawy fix this, don't suppress

eo-parser/src/main/java/org/eolang/parser/Blanks.java
```diff
@@ -60,7 +60,19 @@ static void checkPlain(final Span span, final Globals globals, final Emit emit)
      * @param emit The directives sink
      */
     static void checkTest(final Span span, final Globals globals, final Emit emit) {
-        if (globals.pendingBlanks() == 0) {
+        checkTest(span, globals.pendingBlanks(), emit);
```
Thayorns, line 65: Nothing calls `checkTest(Span, Globals, Emit)` after this change. `LnFormation` and `LnOnlyPhi` were its only two callers and both now pass the captured `blanks`, and `Blanks` is package-private so nothing outside `org.eolang.parser` can reach it either — no test touches it. What survives is a method that still reads `pendingBlanks()` after `enterAfterMeta` has cleared it, which is exactly the shape of #6314, sitting there for whichever line type needs an R-6.5.3 check next. The description counts five other usages that would break, but those five are `checkPlain`, not `checkTest`.

Thayorns (commented):
Snapshots `pendingBlanks()` before `enterAfterMeta` zeroes it, so a `+>` attribute that opens the file no longer trips R-6.5.3. I traced the window between the capture and `checkTest` in both `into()` methods and nothing else touches the counter there, so the value is the same one the old code would have read in the non-meta case — the fix is behaviour-preserving everywhere except the case #6314 reports. What's missing is the mirror test: meta header with no blank line at all in front of the `+>`, where R-6.5.5 and R-6.5.3 should both still fire. That is the case a snapshot could most plausibly have silenced, and neither `LnFormationTest` nor `LnOnlyPhiTest` pins it down.

eo-parser/src/test/java/org/eolang/parser/LnFormationTest.java
```diff
@@ -287,6 +287,22 @@ void acceptsPlusGreaterAttributeAfterBlankLine() {
         );
     }
 
+    @Test
+    void acceptsPlusGreaterAttributeRightAfterMeta() {
```
morphqdd: This calls `globals.markMeta(); globals.blank();` directly instead of running the real tokenizer over the reported source text (`+package foo` / blank / `[] +> works`), so it doesn't prove the actual parse reaches `into()` with that state. This repo's established convention for eo-parser fixes is a declarative YAML pack test through `EoSyntaxTest.validatesEoSyntax` (see `pipe-test-attribute-is-rejected.yaml` for the same R-6.5.3 rule) — that would exercise the real pipeline instead of simulated state, and is the test format usually expected here for parser bug fixes.

eo-parser/src/main/java/org/eolang/parser/LnFormation.java
AhmedAliElashmawy: Done, I removed the `@SuppressWarnings` and updated `Blanks.enterAfterMeta(...)` to return the `pendingBlanks` count, and then nested the call directly inside the if/else branching at the end of the `into()` method.

eo-parser/src/main/java/org/eolang/parser/Blanks.java
AhmedAliElashmawy: I double checked with a repo wide search for `Blanks.checkTest`, it's still called with the old 3-arg signature in `LnApplication`, `LnCompactTuple`, `LnMethod`, `LnReversed`, and `LnTextBlock`. Only `LnFormation` and `LnOnlyPhi` were changed to the new overload. Let me know if I'm missing something.

eo-parser/src/test/java/org/eolang/parser/LnFormationTest.java
AhmedAliElashmawy: You're right, I've deleted those tests from `LnFormationTest` and `LnOnlyPhiTest` and replaced them with proper declarative YAML pack tests inside `eo-syntax/` , which exercises the full pipeline just as you suggested.

AhmedAliElashmawy:
@yegor256 please take a look

AhmedAliElashmawy:
@yegor256 please take a look

yegor256:
@AhmedAliElashmawy conflicts here

yegor256:

---

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2326](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+4442](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+6568](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6341, opened 2026-08-04
fix(#6228): removed XML escaping from `Listing.java`

### LINKED ISSUE #6228: `<listing>` in every XMIR holds double-escaped source instead of verbatim source
## What Happens

Every XMIR's `<listing>` element holds the source text XML-escaped twice, not the raw source verbatim as documented.

`Listing.java` builds the directive as:

```java
.add("listing")
.set(StringEscapeUtils.escapeXml11(this.source))
```

The text is escaped by hand here, and then escaped again by the XML serializer when the directive is written out, since that's how every other plain-text directive in this codebase is handled (the serializer does the escaping). The class's own javadoc says the source is meant to be embedded "verbatim".

Confirmed with a source file containing `<`, `&`, and `"` (all common in ordinary EO source, since `>` is the name-suffix operator used on nearly every line):

```
# Sample.

[] > app
  "a < b & c > d" > x
```

Reading the text value of `/object/listing` from the parsed XMIR gives:

```
[] &gt; app
  &quot;a &lt; b &amp; c &gt; d&quot; &gt; x
```

instead of the actual source. The raw serialized XML shows the double escaping directly: `[] &amp;gt; app`. `listing.equals(source)` is false. Since `>` appears in essentially every `.eo` file's name suffixes, this affects the `<listing>` of every parsed file, not just ones with unusual characters.

The existing tests do not catch this because they assert the wrong thing: `EoSyntaxTest.copiesListingCorrectly` checks `containsString(StringEscapeUtils.escapeXml11(src))` and `printsProperListingEvenWhenSyntaxIsBroken` builds its XPath from `StringEscapeUtils.escapeXml11(src)` too, both baking in the double-escaped expectation as if it were correct.

## What Should Happen

`Listing.java` should set the raw source text on the `listing` directive without calling `escapeXml11` itself, letting the XML writer perform the one escaping pass it always performs. The text value of `/object/listing` should equal the original source byte-for-byte. The two tests that currently assert the escaped form need their expectations updated to the plain source.

h1alexbel:
In this PR I've removed XML escaping from `Listing.java` to let the XML serializer do it when the directive is written out.

closes #6228


eo-parser/src/main/java/org/eolang/parser/Listing.java
```diff
@@ -5,18 +5,33 @@
...
+ *
  * @since 0.1
  */
 final class Listing implements Iterable<Directive> {
 
+    /**
+     * Characters that are not allowed inside an XML text node.
+     */
+    private static final Pattern FORBIDDEN = Pattern.compile(
+        "[\\x00-\\x08\\x0B\\x0C\\x0E-\\x1F\\x7F-\\x84\\x86-\\x9F\\uFFFE\\uFFFF]"
```
Thayorns, line 32: The `\x7F-\x84` and `\x86-\x9F` ranges are legal characters in an XML 1.0 text node — the `Char` production is `[#x20-#xD7FF]`, which swallows all of them — and `Xembler.xmlQuietly()` emits a 1.0 document, so nothing here forbids them. A source with a U+0085 or a DEL in it loses that character silently and `/object/listing` stops equalling the source, which is the invariant the javadoc three lines up and `keepsSourceVerbatim` both claim. The javadoc's "they can't be represented there at all" only holds for the ranges below #x20.

eo-parser/src/test/java/org/eolang/parser/ListingTest.java
```diff
@@ -0,0 +1,95 @@
...
+        MatcherAssert.assertThat(
+            "characters that XML text nodes can't hold must be dropped",
+            ListingTest.listing("[] > x\u0001\u0007\u001F\u007F"),
+            Matchers.equalTo("[] > x")
+        );
+    }
+
+    @Test
+    void buildsListingForEmptySource() {
+        Assertions.assertDoesNotThrow(
```
Thayorns, line 56: The value `listing("")` returns is computed and thrown away, so this only proves nothing was raised, not that a listing was built. `listing()` ends in `.text().orElse("")`, which means an element carrying no text and an element carrying an empty string are the same result here, and neither is checked.

Thayorns (commented):
Replaces the manual `escapeXml11` in `Listing` with a strip of XML-illegal characters, so the writer escapes once and `/object/listing` reads back as the source. The two `EoSyntaxTest` assertions are correctly tightened to `equalTo(src)`, and the new `ListingTest` covers the ampersand, angle-bracket, quote and non-ASCII cases. One gap worth a case: none of the sources carries a bare CR — XML normalizes a literal carriage return in a text node to LF on parse, so that is the one input where "verbatim" is least obviously true, and `SourceTest` already has `creturn()`/`crnl()` for exactly this.

eo-parser/src/main/java/org/eolang/parser/Listing.java
morphqdd, line 32: `\x7F-\x84` and `\x86-\x9F` are valid XML 1.0 text characters (the `Char` production covers `[#x20-#xD7FF]`), so nothing forces their removal here. The old `escapeXml11` at least preserved them as numeric character references; this deletes them outright, silently dropping real source bytes. Also, unpaired UTF-16 surrogates aren't covered at all — `escapeXml11` used to strip those (a lone surrogate can't be serialized as valid XML), and `Listing` wraps the raw source unconditionally even when the source is syntactically broken, so a file with a stray lone surrogate now reaches the XML writer unfiltered.

h1alexbel: @Thayorns seem that Xembly enforces restrictions of XML 1.1, not XML 1.0. In XML 1.1 `\x7F-\x84` and `\x86-\x9F` ranges are illegal, so the message from Xembly is following:

```text
Caused by: org.xembly.XmlContentException: Character #80 is in the restricted XML range #7F-#84, see http://www.w3.org/TR/2004/REC-xml11-20040204/#charsets
```

the code from xembly that does this job:

```java
private static void range(final char chr, final int left, final int right)
    throws XmlContentException {
    if (chr >= left && chr <= right) {
        throw new XmlContentException(
            String.format(
                "Character #%02X is in the restricted XML range #%02X-#%02X, see http://www.w3.org/TR/2004/REC-xml11-20040204/#charsets",
                (int) chr, left, right
            )
        );
    }
}

private static char legal(final char chr) throws XmlContentException {
    Arg.range(chr, 0x00, 0x08);
    Arg.range(chr, 0x0B, 0x0C);
    Arg.range(chr, 0x0E, 0x1F);
    Arg.range(chr, 0x7F, 0x84);
    Arg.range(chr, 0x86, 0x9F);
    return chr;
}
```

TBH, not sure whether we should validate based on XML 1.0 or XML 1.1, if it's 1.1, then this regex is correct in terms of character restrictions.

```regex
[\\x00-\\x08\\x0B\\x0C\\x0E-\\x1F\\x7F-\\x84\\x86-\\x9F\\uFFFE\\uFFFF]
```

@Thayorns @morphqdd Thoughts?

h1alexbel: @morphqdd addressed in https://github.com/objectionary/eo/pull/6341#discussion_r3719300501

h1alexbel:
@morphqdd @Thayorns updated. Take a look, please

h1alexbel:
@Thayorns @morphqdd friendly reminder

Thayorns (commented):
Since my last look the only thing that moved is `ListingTest`: the literal control characters in `dropsCharactersForbiddenInXml` became `String.format` casts, the empty-source test split into one that asserts the listing is empty and one that only asserts nothing was raised, and a parameterized sweep over every codepoint the regex strips was added. I read that delta and found nothing wrong in it.

I dug into Xembly to answer your question about 1.0 versus 1.1 and I was wrong in my earlier comment; details in that thread. The surrogate half of what @morphqdd raised is still uncovered, and I don't think the 1.1 answer reaches it — also in that thread.

eo-parser/src/main/java/org/eolang/parser/Listing.java
Thayorns: @h1alexbel you are right and I was wrong. I was reading the XML spec, but the constraint that actually bites here is Xembly's, and it fires long before any writer runs. `Directives.set` builds a `SetDirective`, whose constructor builds an `Arg`, and `Arg`'s constructor runs `ifValid` over every char through `legal`, which is exactly `0x00-0x08`, `0x0B-0x0C`, `0x0E-0x1F`, `0x7F-0x84`, `0x86-0x9F` — I disassembled xembly 0.32.2, the version pinned in `pom.xml`, and that is the whole check. So it is not really a choice between 1.0 and 1.1: without the strip, a source carrying a U+0080 makes `EoSyntax` throw instead of producing a listing at all. The regex matches that set, plus `￾` and `￿`, which no XML version admits anywhere. It stands.

The part I don't think this answer reaches is @morphqdd's surrogate point. `ESCAPE_XML11` in commons-text is an `AggregateTranslator` whose last member is `UnicodeUnpairedSurrogateRemover`, so the old code was dropping lone surrogates as a side effect. `Arg.legal` checks the five ranges above and never looks at surrogates, so an unpaired one now travels from `Listing` straight into the DOM writer, and `Listing` wraps the source unconditionally — including when the source is broken, which is where a stray half-pair is most likely to come from.

eo-parser/src/test/java/org/eolang/parser/ListingTest.java
Thayorns: Thanks for splitting this — `buildsListingForEmptySource` now asserts something, which is what I was after. One thing survived the split though: the helper still ends in `.text().orElse("")`, and in xnav a missing child is `Empty`, whose `child` returns itself and whose `text` returns `Optional.empty()`. So `""` is what you get both when the listing holds an empty string and when there is no `listing` element at all. That fallback now carries `removesForbiddenCharacters` too — all fourteen cases assert `emptyString()`, and every one of them would stay green if `Listing` stopped emitting the element entirely. `dropsCharactersForbiddenInXml` is the one test here immune to that, because it keeps `[] > x` around the stripped characters.

yegor256:
The CI check **thresholds (smells, *.java, static|null, 840)** is failing on this PR. Please investigate and fix the smell threshold violation before this can be merged.

---

yegor256:

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+4465](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+6559](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6343, opened 2026-08-04
feat(#4493): `sync` atom that guards dispatch with a lock

### LINKED ISSUE #4493: `sync` atom is missing
Let's create a new atom `sync` that will decorate any object, behave like it, but guarantee that only thread is using it:

```
sync a > b
```

Here, when I do `b.hello`, I'm confident that no other thread is touching `a`.

Let's try to implement this with the help of `pthread_mutex_t` of POSIX: https://man7.org/linux/man-pages/man3/pthread_mutex_lock.3.html

Thayorns:
Closes #4493.

Adds a `sync` atom that decorates any object, behaves exactly like it, and lets only one thread dispatch on it at a time:

```
sync a > b
```

While `b.hello` is being dispatched, no other thread is dispatching on `a`.

### How the guarding works

`EOsync` overrides `take` and holds a lock around `super.take(name)`. That single wrap is enough to guard the decoratee: on an attribute miss, `PhDefault.absent` resolves an atom through `this.take(LAMBDA).take(name)`, and that whole expression runs *inside* `PhDefault.take` — so `a.take(name)` happens under the lock.

Only the dispatch is guarded, which is what the issue asks for ("we should synchronize the dispatch only, the `.take()` on `a`"). The object that comes back is dataized by whoever asked for it, outside the lock.

The lock is reentrant because the dispatch takes it twice on one thread (once for `hello`, once for λ), and it is carried through `PhDefault.copy`'s shallow clone, so every copy of the same `sync` guards the same decoratee.

### `ReentrantLock` instead of `pthread_mutex_t`

The issue suggests implementing this with POSIX `pthread_mutex_t` via the `syscall` atom. I did not do that, for two reasons:

1. `pthread_mutex_lock` is a libc function, not a syscall, so it is not reachable through the `posix` atom as it stands.
2. More fundamentally, a POSIX mutex cannot exclude threads from a JVM-level `take()`. The thing that actually needs guarding here is the runtime's own dispatch, which lives entirely inside the JVM. A native mutex would guard nothing that matters on this runtime.

`ReentrantLock` is the correct primitive for the `jvm` runtime. A native runtime would want the POSIX mutex; `eo2js` will need its own implementation of this atom either way. Happy to revisit if you'd rather the atom be shaped around a native mutex from the start.

### Tests

- `sync.eo` carries 5 EO tests: behaves as a decorated number, a decorated empty string, a decorated non-ASCII string, forwards an attribute take, and survives `sync (sync x)`.
- `EOsyncTest` is the one that fails for the reason the issue states: 8 threads dispatch on a decoratee that records how many are inside it at once, and it asserts the peak is 1. With the lock removed the test fails with `Expected: <1> but: was <8>`; with the lock it passes.

### What I verified locally

Build is clean — `Φ.sync,A` lands in the generated `atoms.csv`, `EOsync.class` compiles, all 171 EO sources transpile and lint clean. All 6 sync tests pass (`EOsyncTest` plus the generated `EOsyncEOAtomTest`).

Two gaps, in the interest of full disclosure: I could not run Qulice locally (it requires Java 21, this box has 17), and I did not get a full `eo-runtime` regression run to complete — it exceeded its timeout. The change adds a new class and a new `.eo` without touching existing runtime code, so the regression risk looks low, but CI should be the judge.


morphqdd (approved):

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+3501](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Thayorns:
@yegor256 ping

yegor256:

Thayorns:
@yegor256 ping

Thayorns:
@yegor256 CI fixed here

Thayorns:
@yegor256 CI green here

Thayorns:
@yegor256 ping

Thayorns:

Thayorns:
@yegor256 fixed here

Thayorns:
@yegor256 have a look here, fixed

Thayorns:
@yegor256 fixed here, take a look please

Thayorns:

Thayorns:

yegor256:
@Thayorns we are not yet sure that this is the right design. We don't know how to implement multi-threading in EO yet.

PR #6344, opened 2026-08-04
#6273: PhDefault.add() is declared final, so a subclass cannot override it

### LINKED ISSUE #6273: `PhDefault.add()` is declared `final`, so a subclass cannot override it
**What happens.** `PhDefault.add(String, Attribute)` at `PhDefault.java:319` is declared `public final`. The javadoc directly above it says the opposite:

```
This method can only be called from child classes, in their
constructors, when they declare their attributes. This is why it's
protected.
```

The comment says protected, the code says public final. `final` means no subclass can override the method.

This blocks `eo.phiDefaultClass`, the option added in #5955 and extended in #6129. That option exists so that objectionary/eoc can transpile a program with its own class in place of `PhDefault`. The design in objectionary/eoc#500 says: "This `PhInspected` will redefine `.add()`, `take()` and maybe other methods." `take()` is overridable, `add()` is not. The generated Java calls `add()` on exactly these objects, since `to-java.xsl` emits `((PhDefault) ctx).add(...)`, so the substituted class is handed the calls it cannot see.

`copy()` at line 175 is `final` for the same reason, and the class has no `protected` members at all: `attrs` is private and `loaded()` is private.

**What should happen.** Drop `final` from `add`, so that a subclass can observe attribute additions, and fix the javadoc, which describes a visibility the method does not have.

One thing worth stating in that javadoc while it is being fixed: `loaded()` at line 562 calls `this.add(...)` for every attribute passed to the constructor. It is lazy, so this happens on first access rather than during construction, but an override would see the initial attribute set being materialized, not only later additions.

morphqdd:
Closes #6273

`PhDefault.add(String, Attribute)` was `public final`, contradicting its own javadoc, which said the method is "protected" and callable only from child classes. `final` blocks any override entirely, which is what `eo.phiDefaultClass` (objectionary/eoc#500) needs: a substituted subclass has to observe attribute additions by overriding `add()`.

Dropped `final` and rewrote the javadoc to describe what the method actually is now: public and overridable, called lazily by `loaded()` for every constructor-supplied attribute on first access, not only for later explicit `.add()` calls.

Added `Observant`, a `PhDefault` subclass that overrides `add()` and records every name it sees, and a test asserting it observes both the constructor attribute and a later `.add()` call. Verified the test fails to compile against the old `final` method and passes against the fix.

`copy()` (also `final`) and the private visibility of `attrs`/`loaded()` are untouched, matching the issue's scope.


eo-runtime/src/test/java/org/eolang/Observant.java
```diff
@@ -0,0 +1,43 @@
...
+        this.names.add(name);
+        super.add(name, attr);
+    }
+
+    /**
+     * Attribute names seen so far.
+     * @return Names, in call order
+     */
+    List<String> seen() {
+        return this.names;
```
Thayorns: `seen()` hands the live `names` list back, so whatever calls it can add to or clear the very record that `Observant` exists to own. The field being `final` only pins the reference; the list behind it stays open, so the object is mutable from the outside.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
```diff
@@ -30,6 +30,17 @@ void printsEmptyObjectAsTerm() {
         );
     }
 
+    @Test
+    void letsSubclassObserveAttributeAdditions() {
```
Thayorns: This checks two separate mechanisms at once: that `loaded()` routes the constructor-supplied `x` through the overridden `add()`, and that a direct `kid.add("extra", ...)` reaches it too. They fail for different reasons — one is the lazy materialization inside the private `loaded()`, the other is a plain virtual call — but folding them into one `hasItems` makes both regressions surface as the same red line.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
```diff
@@ -30,6 +30,17 @@ void printsEmptyObjectAsTerm() {
...
+        final Observant kid = new Observant();
+        kid.add("extra", new AtVoid("extra"));
+        MatcherAssert.assertThat(
+            "a subclass overriding add() must observe both the constructor-supplied attribute (materialized lazily by loaded()) and a later explicit add()",
```
Thayorns: The message states what the subclass must do rather than what went wrong, so a red build reports the wish and not the failure. `printsEmptyObjectAsTerm` and `prefersPackageExtensionOverDecoratee` right above both close with "but it didnt" for that reason.

Thayorns (commented):
Drops `final` from `PhDefault.add()` so a substituted subclass can observe attribute additions, rewrites the javadoc to match, and adds an `Observant` fixture with a test that watches both the constructor-supplied attribute and a later explicit call. The production change is a one-word removal and the test genuinely fails without it, since `loaded()` routes the initial attributes through `this.add()`. Two smaller things in the test fixture caught my eye.

eo-runtime/src/test/java/org/eolang/Observant.java
morphqdd: Wrapped it in Collections.unmodifiableList. The reference is still final but callers can no longer mutate the record through it.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
morphqdd: Split into letsSubclassObserveConstructorSuppliedAttributeOnFirstAccess and letsSubclassObserveExplicitAttributeAddition, moved both to a new ObservantTest.java since the split pushed PhDefaultTest.java past the 1000-line FileLengthCheck.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
morphqdd: Reworded to state what went wrong, matching printsEmptyObjectAsTerm and prefersPackageExtensionOverDecoratee's "...but it didnt" style.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3361](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+946](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6346, opened 2026-08-04
Hash file paths and lengths when hashing a directory

### LINKED ISSUE #5254: `Sha` class causes hash collisions for different directory trees due to lack of delimiter and absence of file paths in hashing

Thayorns:
`Sha` streamed the raw bytes of every file of a directory into one digest, back to back. Nothing in the digest said where one file ended and the next began, and the file names were used only for sorting, never hashed. Two trees whose sorted contents concatenate to the same bytes therefore got the same hash: `{a.txt="ab", b.txt="c"}` and `{a.txt="a", b.txt="bc"}` both came out as the SHA-256 of `abc`, and renaming a file left the hash untouched.

Now every file of a directory is preceded in the digest by its path relative to the root and followed by the number of bytes read from it, both NUL-terminated, so the boundaries and the names are part of what gets hashed. The length comes from the read loop itself rather than from a second `Files.size` call, so it always describes the bytes that were really hashed even if the file changes size underneath us. Relative paths go into the digest, and serve as the sort key, with forward slashes on every platform, so the same tree hashes alike on Windows and on Linux. A lone regular file is left alone and still hashes to the plain SHA-256 of its bytes, which is what `CacheTest` pins down and what the `.sha256` markers already on disk contain.

Only files speak into the digest, so a directory holding none of them stays invisible to the hash, the way Git also sees it. That is now written down in the class docblock. Framing bare directory paths too would trade one collision for another: `{1/, 234/, 3/}` as empty directories and a single file `1` holding `234` both come out as `\0` `1\0` `234\0` `3\0`.

`ShaTest` covers the three collisions above, a file moved to a differently named subdirectory, and the plain hash of a lone file, and checks that two identical trees sitting in different directories still agree, so the relative path never leaks the absolute one into the hash.

`Cache.dirSha` still hashes directories on its own instead of going through `Sha`, which is why the bug never showed up in production. It cannot simply delegate yet, because it also skips files that `Sha` has no way of skipping, so it carries a puzzle now.

Closes #5254


Thayorns:
@yegor256 take a look please, happy to clarify anything here.

eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
```diff
@@ -16,7 +17,8 @@
 
 /**
  * SHA-256 hash of a file or directory.
- * For a directory, walks all files sorted by path and hashes their combined contents.
+ * For a directory, hashes every file sorted by path, framed by its relative path and length,
```
morphqdd: The PR description says Cache.dirSha still hashes directories on its own instead of going through this class, which is why the old collision never showed up in production, and that folding it in is worth doing later. Nothing in the code marks that as outstanding work, only the PR body. A puzzle here (or on Cache.dirSha itself) would let 0pdd track it as a real issue instead of it depending on someone rereading this PR's description.

eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
```diff
@@ -37,27 +39,34 @@ final class Sha {
...
+        final boolean tree = Files.isDirectory(this.path);
+        try (Stream<Path> walk = Files.walk(this.path)) {
             walk.filter(Files::isRegularFile)
                 .sorted(Comparator.comparing(Path::toString)).forEach(
                     file -> {
                         try (InputStream input = Files.newInputStream(file)) {
+                            if (tree) {
+                                digest.update(
+                                    String.format(
+                                        "%s\0%d\0", this.path.relativize(file), Files.size(file)
```
morphqdd: Files.size(file) is a separate stat call from the read loop just below that actually streams and hashes the file's bytes. If the file changes size between the two (a build touching its own inputs, a concurrent writer), the length framed into the digest would not match what was actually hashed, undermining the exact boundary guarantee this fix is meant to add. Using the byte count accumulated by the read loop itself would tie the framed length to what was really hashed.

eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
```diff
@@ -37,27 +39,34 @@ final class Sha {
...
      * @throws NoSuchAlgorithmException If SHA-256 is unavailable
      */
-    private static String hash(final Path path) throws IOException, NoSuchAlgorithmException {
+    private String hash() throws IOException, NoSuchAlgorithmException {
         final MessageDigest digest = MessageDigest.getInstance("SHA-256");
-        try (Stream<Path> walk = Files.walk(path)) {
+        final boolean tree = Files.isDirectory(this.path);
+        try (Stream<Path> walk = Files.walk(this.path)) {
             walk.filter(Files::isRegularFile)
                 .sorted(Comparator.comparing(Path::toString)).forEach(
```
morphqdd: Both the sort key here and the relative path framed into the digest go through Path::toString(), which uses backslashes on Windows and forward slashes elsewhere. The same directory tree would sort differently and hash to a different digest depending on the OS it is hashed on, which matters if this hash is ever used as a cache key shared or compared across platforms.

eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
```diff
@@ -37,27 +39,34 @@ final class Sha {
...
      * @throws IOException If reading fails
      * @throws NoSuchAlgorithmException If SHA-256 is unavailable
      */
-    private static String hash(final Path path) throws IOException, NoSuchAlgorithmException {
+    private String hash() throws IOException, NoSuchAlgorithmException {
         final MessageDigest digest = MessageDigest.getInstance("SHA-256");
-        try (Stream<Path> walk = Files.walk(path)) {
+        final boolean tree = Files.isDirectory(this.path);
+        try (Stream<Path> walk = Files.walk(this.path)) {
             walk.filter(Files::isRegularFile)
```
morphqdd, line 61: Two directories that both contain zero regular files hash the same regardless of their names or nesting, since nothing here contributes anything to the digest for a directory with no files in it. Not one of the collisions this PR sets out to fix, and not mentioned as a remaining gap either.

eo-maven-plugin/src/test/java/org/eolang/maven/ShaTest.java
```diff
@@ -0,0 +1,77 @@
...
+        new Saved(text.substring(7), temp.resolve("second/b.txt")).value();
+        MatcherAssert.assertThat(
+            String.format("hashes of differently split dirs are equal, seed=%d", seed),
+            new Sha(temp.resolve("first")).toString(),
+            Matchers.not(Matchers.equalTo(new Sha(temp.resolve("second")).toString()))
+        );
+    }
+
+    @Test
+    void distinguishesDirsWithRenamedFile(@Mktmp final Path temp) throws IOException {
```
morphqdd, line 43: This only renames a file at the top level. There is no test for the same file name moving to a different nested subdirectory (for example sub1/x.txt vs sub2/x.txt), which exercises the same relative-path framing but at a deeper level than the flat rename here.

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (5) comments.  Your running score is [+3765](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Thayorns:
@yegor256 have a look

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (195 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3935](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6347, opened 2026-08-04
fix(#6342): normalize WPA unlint rule matching

### LINKED ISSUE #6342: `+unlint` fails to suppress defects from whole-program-analysis due to improper rule name comparison
`Linting.java` has two overloads of `notSuppressed`. The one used for whole-program-analysis (WPA) defects, at `eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java:615-621` (called from `wpa()` at line 459), compares a `+unlint` meta's `tail` directly against `defect.rule()`. The sibling overload used for per-file source-lint defects, at lines 663-670, strips a `/S` or `/W` scope suffix first via `Linting.baseRule(defect.rule())` (lines 677-684 — whose own doc comment says "Strip scope suffix (e.g. `/S`, `/W`)"). The WPA overload is missing that same stripping.

This matters because every WPA defect's `rule()` really is scoped: the pinned `wpa` dependency (0.1.1, per `eo-maven-plugin/pom.xml:43-45`) builds its defects via `new ScopedDefects(lint.defects(pkg), "W")`, whose constructor rewrites each rule to `String.format("%s/%s", original.rule(), "W")` (verified by disassembling the jar). So a WPA defect's actual rule is always `"<name>/W"`, never the bare name — while every existing `+unlint` usage in the repo (e.g. `eo-runtime/src/main/eo/bytes.eo:22: +unlint mandatory-package`) writes the bare name, since that's what the source-lint path expects. A developer who adds `+unlint foo-rule` to suppress a WPA-flagged defect named `foo-rule` gets no suppression at all, because the XPath in `notSuppressed` is actually looking for `tail='foo-rule/W'`. If the defect's severity fails the build, there is no way to suppress it short of reading the plugin's internal source and discovering the undocumented `/W` convention. `LintingTest.java` has no test referencing `"unlint"`, so this gap isn't caught.

Fix: apply `Linting.baseRule(defect.rule())` in the WPA `notSuppressed` overload at line 619, the same way the source-lint overload already does at line 667.

Shaurya2k06:
Fixes #6342.

## Summary
- Normalize scoped WPA rule names (`name/W`) via `baseRule(...)` before matching `+unlint` metas, the same way source-lint already does
- Add a regression test that a plain `+unlint incorrect-alias` suppresses the WPA `incorrect-alias/W` defect

## Test plan
- [x] `mvn -pl eo-parser,eo-printer,eo-maven-plugin -am install -DskipTests -DskipITs`
- [x] `cd eo-maven-plugin && mvn test -Dtest=MjLintTest#suppressesWholeProgramAnalysisDefectViaUnlint,MjLintTest#detectsWholeProgramAnalysisErrorsSuccessfully,MjLintTest#ignoresWholeProgramAnalysisErrors,LintingTest`
- [x] `cd eo-maven-plugin && mvn test`


Thayorns (commented):
Normalizes scoped WPA rule names through `baseRule` before matching `+unlint` metas, so a plain `+unlint incorrect-alias` now suppresses `incorrect-alias/W`, plus a regression test and a few unrelated formatting cleanups. The one-line change in `Linting.notSuppressed` looks right and matches what the source-lint path already did. The new test, however, reads a file that never carries WPA errors, so the fix itself is still effectively unverified.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java
```diff
@@ -291,6 +291,30 @@ void ignoresWholeProgramAnalysisErrors(@Mktmp final Path temp) {
...
+                "+unlint incorrect-unlint",
+                "+unlint unlint-non-existing-defect",
+                "",
+                "[] > main"
+            );
+        maven.execute(new FakeMaven.Lint());
+        MatcherAssert.assertThat(
+            "A plain +unlint incorrect-alias must suppress the WPA defect reported as incorrect-alias/W",
+            new Xnav(maven.programTojo().linted())
+                .path("/object/errors/error[@check='incorrect-alias/W']").count(),
```
Thayorns, line 327: The file this reads never carries WPA errors, so the assertion holds whether or not the fix works. `lintOne` writes the linted target and calls `tojo.withLinted(target)` (Linting.java:327-346), and `linting()` only reaches `lintAll` afterwards (Linting.java:244, then 253). The WPA defects are embedded into the in-memory `pkg` node built from `tojo.xmir()` inside `wpa()` (Linting.java:452-462), and that DOM is never written back to disk. Drop the `+unlint incorrect-alias` line and this still counts 0.

yegor256:
@Shaurya2k06 merge conflicts here, please resolve

0crat:
@shaurya2k06 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (35) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+175](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1101](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6350, opened 2026-08-04
Add eo-inference module with the provides table

### LINKED ISSUE #6349: XMIR missing module implementation for provides, needs and links tables generation
The type-inference design note that lived at the repository root for a
few days (added in `d6e0dbf9d0`, removed in `f32a1d4c96`) spells out a
complete structural checker for XMIR, and nothing in the build
implements it. The only working code is an experiment outside the
repository, which is not where a compiler phase belongs.

The design is worth quoting in short, because it decides the shape of
the module. Every `<o>` gets its own synthetic type and keeps it
forever; all knowledge about types goes into three collections instead:
the **provides** table (what an object certainly has), the **needs**
table (what an object must have, judging by how it is used) and the
**links** list (which types are copies of which). One pass over the
XMIR fills them, with one rule per kind of `<o>`: a formation says what
it provides and that its list is complete, a `ξ.name` reference adds a
link, a `.name` dispatch adds a needs row, and an application files a
pending check. A worklist then drains the checks, and reports a mistake
only when the "has" side is complete and the needed attribute is still
missing, staying silent otherwise.

So the module should produce three XML documents — `provides`, `needs`
and `links` — out of a 1-parse XMIR. Two preparation steps come first.
`roll-bases.xsl` rolls a dotted chain into a single `base="ξ.x.next.foo"`,
so the module has to unroll it back into one `<o>` per dispatch step
before the rules can fire at all; and every `<o>` needs a type id of
its own, next to the `@loc` that `set-locators.xsl` already gives it.

This is meant to stay a prototype and cover primitive cases only.
Delegation through `φ`, copies receiving fresh types instead of sharing
them, and atoms arriving as ready-made provides rows are all left out
on purpose — the design's central claim is that each of those is one
more rule rather than a rewrite, and the module is worth building only
if that claim survives contact with real code. Add `eo-inference` as a
separate Maven module, with the preparation steps and the provides
table first.

maxonfjvipon:
A new module, `eo-inference`, with the first step of the checker in it:
the preparation of the XMIR, and the table that says what every object
certainly has. It is a prototype on purpose — three of the four rules
and the whole checking loop are still missing — so the thing to look at
here is the shape, not the coverage.

Two stylesheet passes come before any rule. The parser rolls a chain
into a single base, and a rule can only notice an attribute being taken
when the taking is an object of its own, so `unroll-bases.xsl` undoes
exactly what `roll-bases.xsl` did, and the parser's own
`set-locators.xsl` runs again to name the objects that the splitting has
just created. For `x.next.foo > @` inside `[x] > inc` that leaves:

```xml
<o base=".foo" loc="Φ.inc.φ" name="φ">
  <o base=".next" loc="Φ.inc.φ.ρ">
    <o base="ξ.x" loc="Φ.inc.φ.ρ.ρ"/>
  </o>
</o>
```

The locator is also the identity of a type. The design note numbers the
objects (`t0`, `t1`, ...) to give each one a type of its own; this uses
the locator instead, since XMIR already carries a unique name per object
and a row saying that `Φ.app.t.next` has nothing needs no dictionary to
be read. The rule that matters holds either way — a type belongs to one
object and is never renamed or merged.

The rows live in `Tojos`, the tables the compiler already keeps its
catalogues in, so asking a table something is a `select` rather than a
query written by hand, and thread-safety and caching are decorators for
when the checker runs per file in parallel. A row is flat, so a type and
each of its attributes take one, told apart by an `owner`; an attribute
is identified by its owner and its name together, because its own
locator would collide with the row of the same object as a type. Rows
carry the `index` they were written at as well: a table promises nothing
about order, while the order of void attributes decides what an
application binds to what.

`Grouped` turns any such table back into a document, knowing no column
by name — `id`, `owner` and `index` become nesting, everything else
becomes an attribute of the same name — so the needs table will be
rendered by the same class, and a rule that records something new about
a type gets it into the report for free.

Voids are marked, and an atom's row is not complete, because its `λ`
stands for a body written in Java. That flag is what will keep the
checker quiet later, where it cannot see.

Next steps, one per rule: a reference adds a link, a dispatch adds a
needs row, an application files a pending check; then the loop that
drains the checks and reports a mistake only when it is sure.

Closes #6349


yegor256:
@maxonfjvipon maybe we can use Tojos for in-memory tables? We'll get to-XML feature out of the box and many other features, including thread-safety

maxonfjvipon:
Done, the tables are Tojos now — a rule fills rows and the XML is a view
over them, which is what it should have been from the start. `select`
replaces the hand-written lookups, which is nearly all the checking loop
of the next step does, and the decorators are there for when the checker
runs per file in parallel.

Two things worth reporting from doing it.

There is no XML in Tojos, in any released version: 0.19.1 is the latest
and it ships `MnCsv`, `MnJson`, `MnTabs`, `MnYaml` and `MnMemory`, and
nothing in this repository converts a table to XML either. So the
rendering stayed ours. It did get better, though — one view now serves
every table, because it knows no column by name:

```java
dirs.add("type").append(new Cells(cells, List.of("index")).directives());
```

`id`, `owner` and `index` become nesting, everything else becomes an
attribute of the same name. The needs table has the same shape, so the
same class will render it, and a rule that starts recording something
new about a type gets it into the report without touching the view.

The other thing is order. A table promises nothing about it, and the
order of attributes is not decoration here: an application binds its
arguments to void attributes in the order they were declared, so the
rule that checks applications will ask for the first void and must get
the same answer every time. Relying on the order rows come back in was
wrong even where it happened to work, so every row now carries the
`index` it was written at, and the view sorts by it. The report is
byte-for-byte what it was before the change, which the untouched tests
in `InferenceTest` confirm.

One thing you may want to look at: `Cells` is a class only because
checkstyle demands a `static` method when one does not touch `this`,
while the `thresholds` job pins the number of `static` members in the
codebase to an exact count, so adding one turns that job red. An object
is a decent way out, and arguably better than a helper method, but the
two checks do contradict each other for anyone who tries to extract a
method — which also rules out the usual cure for a cyclomatic complexity
violation.

eo-inference/src/main/java/org/eolang/inference/Inference.java
```diff
@@ -0,0 +1,84 @@
...
+ * <pre> XML rows = new Inference(xmir).provides();</pre>
+ *
+ * <p>Two things happen to the XMIR before any rule looks at it. First,
+ * every composite base is split into one object per dispatch step: the
+ * parser rolls {@code x.next.foo} into a single base, while a rule can
+ * only notice an attribute being taken when the taking is an object of
+ * its own. Second, the locators are set again, because the objects that
+ * the splitting has just created have none.</p>
+ *
+ * <p>The locator is also what this module uses as the identity of a
```
yegor256: @maxonfjvipon this paragraph refers to some "design it follows". probably, it means the `inference.md` file, but we don't have it in the repo. Let's re-phrase this repo just to explain the design and drop the refence.

eo-inference/src/main/java/org/eolang/inference/Inference.java
```diff
@@ -0,0 +1,84 @@
...
+import org.eolang.parser.TrFull;
+
+/**
+ * The types of the objects of one XMIR document.
+ *
+ * <p>Give it the XMIR that the parser has already canonicalised (see
+ * {@code org.eolang.parser.Canonical}) and it answers with the tables
+ * the checker keeps. It's as simple as this:</p>
+ *
+ * <pre> XML rows = new Inference(xmir).provides();</pre>
```
yegor256: @maxonfjvipon the `Inference` object must encapsulate a directory with XMIR files, not a single file. Then, it should provide a new directory with similar XMIR files (modified) and inference tables on top of that. Maybe this way:

```
new Inference(input_dir).inferTo(output_dir, tables_dir);
```

In the `output_dir` we'll find modified XMIR files, while in the `tables_dir` we'll file inference tables.

maxonfjvipon: Done — `Inference` takes the directory now and writes both directories:

```java
new Inference(input).inferTo(output, tables);
```

The prepared XMIR files land in `output` under the names and folders they had, the tables in `tables`, a document each. One table covers the whole program rather than one per file, which turns out to cost nothing: a locator names one object of one file and no other, so rows from every file simply sit in the same table and nothing has to be merged or renamed. The files are walked in a fixed order, otherwise the tables would come out different from run to run.

Two consequences worth mentioning. The rules now take the prepared documents of the whole program rather than one, which is what makes a single table possible. And the prepared XMIR is worth keeping precisely because every row points into it — a row says `Φ.app.inc.φ.ρ`, and that object only exists after the splitting.

eo-inference/src/main/java/org/eolang/inference/Inference.java
maxonfjvipon: Dropped. The paragraph did mean `inference.md`, which is not in the repository, so the docblocks now explain the design themselves rather than referring to it: why a locator is the identity of a type, what the three tables are and why they are kept apart, and why an atom cannot be complete. Same for the two other places that leaned on it — `Provides` no longer says "the design expects" the rows of atoms to arrive one day, it says what has to happen and why (what an atom returns is written in Java, so those rows have to come from outside).

maxonfjvipon:
@yegor256 take a look again

eo-inference/src/main/java/org/eolang/inference/Provides.java
```diff
@@ -0,0 +1,131 @@
...
+    Provides(final Collection<XML> prepared) {
+        this.xmirs = prepared;
+    }
+
+    @Override
+    public Tojos rows() {
+        final Tojos rows = new TjCached(new TjDefault(new MnMemory()));
+        int seen = 0;
+        for (final XML formation : this.formations()) {
+            final String owner = formation.xpath("@loc").get(0);
```
morphqdd, line 74: formation.xpath("@loc").get(0) reads index 0 with no check that the list is non-empty. This is currently safe because Inference.inferTo() always runs set-locators.xsl first, but Provides itself does not enforce or check that invariant, so a raw IndexOutOfBoundsException with no context is what a future caller gets if that pipeline order is ever broken.

eo-inference/src/main/java/org/eolang/inference/Provides.java
```diff
@@ -0,0 +1,131 @@
...
+                .set("index", Integer.toString(seen))
+                .set("complete", Boolean.toString(formation.nodes("o[@name='λ']").isEmpty()));
+            seen = seen + 1;
+            for (final XML attr : formation.nodes("o[@name and not(@name='λ')]")) {
+                final String name = attr.xpath("@name").get(0);
+                final Tojo row = rows.add(String.join(" ", owner, name))
+                    .set("owner", owner)
+                    .set("index", Integer.toString(seen))
+                    .set("name", name)
+                    .set("type", attr.xpath("@loc").get(0));
```
morphqdd, line 85: Same pattern as the owner locator above: attr.xpath("@loc").get(0) is read unchecked. The selecting XPath only guarantees @name is present on these nodes, not @loc.

eo-inference/src/main/java/org/eolang/inference/Grouped.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * The rows that are about a type itself, oldest first, so that the
+     * document follows the code.
+     * @return The rows without an owner
+     */
+    private List<Tojo> types() {
+        return this.table.select(row -> !row.exists("owner"))
+            .stream()
+            .sorted(Comparator.comparingInt(row -> Integer.parseInt(row.get("index"))))
```
morphqdd, line 103: Integer.parseInt(row.get("index")) has no defense against a missing or non-numeric index cell. The class javadoc says this view is meant to be reused by other tables (the needs table is named as an example), so a future rule that forgets to set index gets an unhelpful NumberFormatException or NullPointerException here with nothing pointing back at which table or row caused it.

eo-inference/src/main/java/org/eolang/inference/Grouped.java
```diff
@@ -0,0 +1,124 @@
...
+            );
+            for (final Map<String, String> attr
+                : attrs.getOrDefault(cells.get("id"), Collections.emptyList())) {
+                dirs.add("attr")
+                    .append(new Cells(attr, Arrays.asList("id", "owner", "index")).directives())
+                    .up();
+            }
+            dirs.up();
+        }
+        return new XMLDocument(new Xembler(dirs).domQuietly());
```
morphqdd, line 92: domQuietly() swallows Xembly build failures. This document is the module's actual output artifact, not a debug dump, so a bug in constructing it (for example a cell value that is not valid as an XML attribute) would be silently absorbed here instead of surfacing during development.

eo-inference/src/main/java/org/eolang/inference/Inference.java
```diff
@@ -0,0 +1,122 @@
...
+    public void inferTo(final Path output, final Path tables) throws IOException {
+        final Xsline train = new Xsline(
+            new TrFull(
+                new TrDefault<>(
+                    new StClasspath("/org/eolang/inference/unroll-bases.xsl"),
+                    new StClasspath("/org/eolang/parser/parse/set-locators.xsl")
+                )
+            )
+        );
+        final Collection<XML> prepared = new ArrayList<>(0);
```
morphqdd: prepared keeps every transformed XML document of the whole program in memory at once, in addition to writing each one to disk. Noted as a prototype in the PR description, but worth flagging for when this runs over a real program the size of eo-runtime.

eo-maven-plugin/src/main/resources/org/eolang/maven/inference/unroll-bases.xsl
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" id="unroll-bases" version="2.0">
```
morphqdd, line 6: This stylesheet has no test of its own. It is only exercised combined with set-locators.xsl through Inference's integration tests, so a bug specific to the splitting logic here would be harder to localize than one in set-locators.xsl, which does have its own declarative test pack in eo-parser.

eo-maven-plugin/src/main/java/org/eolang/maven/Table.java
```diff
@@ -0,0 +1,42 @@
...
+ * {@code select} rather than a hand-written query — the checker's
+ * to-do loop will do little else. Thread-safety and caching arrive as
+ * decorators when they are needed, and every table can be written out
+ * as CSV or JSON for free. The XML the module reports is a view over
+ * these rows, built by {@link Grouped}; the rows themselves are the
+ * truth.</p>
+ *
+ * @since 0.67.0
+ */
+@FunctionalInterface
```
morphqdd: @FunctionalInterface signals this is meant to be implemented with a lambda, but the javadoc describes three planned implementations as full stateful classes (Provides being the first). Worth dropping the annotation unless a lambda-based implementation is actually expected somewhere.

eo-inference/src/main/java/org/eolang/inference/Cells.java
```diff
@@ -0,0 +1,58 @@
...
+        this.placement = skipped;
+    }
+
+    /**
+     * These cells as attributes of the element being built.
+     * @return The directives
+     */
+    Directives directives() {
+        final Directives dirs = new Directives();
+        for (final Map.Entry<String, String> cell : this.row.entrySet()) {
```
morphqdd, line 51: Attribute order in the generated element depends on the iteration order of this.row, with no explicit sort. Worth confirming Tojo.toMap() has a deterministic order, since this becomes the attribute order of the human-facing report this module produces.

eo-inference/src/main/java/org/eolang/inference/Table.java
```diff
@@ -0,0 +1,42 @@
...
+ * what lets the checker grow: a smarter rule adds rows to a table, or
+ * reads them differently, and no other part of the pipeline has to
+ * change.</p>
+ *
+ * <p>The rows are {@code Tojos}, the same tables the compiler already
+ * keeps its catalogues in. A row is a handful of named cells, which is
+ * all a fact about a type ever is, and asking for them back is a
+ * {@code select} rather than a hand-written query — the checker's
+ * to-do loop will do little else. Thread-safety and caching arrive as
+ * decorators when they are needed, and every table can be written out
```
morphqdd: This says every table can be written out as CSV or JSON for free, but only XML (via Grouped) exists anywhere in this PR. Unlike the other stated gaps in this module (which package-info.java tracks as planned work), there is no puzzle marking CSV/JSON export as missing.

eo-inference/src/test/java/org/eolang/inference/InferenceTest.java
```diff
@@ -0,0 +1,164 @@
...
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * Test case for {@link Inference}.
+ * @since 0.67.0
+ */
+@ExtendWith(MktmpResolver.class)
+final class InferenceTest {
```
morphqdd: No test for an input directory with zero .xmir files. sources() would return an empty list and the resulting provides.xml would have no type rows at all; that path is not exercised anywhere here.

eo-inference/src/main/java/org/eolang/inference/Provides.java
```diff
@@ -0,0 +1,131 @@
...
+     *  one table covers all of them: a locator names one object of one
+     *  file and no other
+     */
+    Provides(final Collection<XML> prepared) {
+        this.xmirs = prepared;
+    }
+
+    @Override
+    public Tojos rows() {
+        final Tojos rows = new TjCached(new TjDefault(new MnMemory()));
```
morphqdd, line 71: TjCached wraps a table that is built once inside this method and read exactly once by Grouped right after. Nothing here reuses a query against it, so the caching decorator has no effect yet. Table's own javadoc says caching arrives as a decorator when it is needed, which this does not appear to be yet.

eo-inference/src/main/java/org/eolang/inference/Inference.java
```diff
@@ -0,0 +1,122 @@
...
+        for (final Path source : this.sources()) {
+            final XML ready = train.pass(new XMLDocument(source));
+            final Path target = output.resolve(this.input.relativize(source));
```
morphqdd: If a source file is removed from the program between two runs, its old prepared XMIR and any rows it contributed stay in output/tables from the previous run, since nothing here clears the target directories first or removes now-orphaned files. Not tested either way.

eo-inference/src/main/java/org/eolang/inference/Provides.java
maxonfjvipon: Left as it is, on purpose. `xpath("@loc").get(0)` blowing up is the behaviour I want here: the exception lands one frame from the XPath, in the class that made the assumption, with the document at hand. A guard would turn a broken pipeline into a report with a hole in it, and this repository favours failing fast over failing safe.

The invariant isn't `Provides`'s to defend either — the class is package-private, `Inference` is its only caller, and the train there puts `set-locators.xsl` right after the splitting. If it ever needs enforcing, the honest shape is a type that can only hold prepared XMIR, not an `if` in the middle of a rule.

eo-inference/src/main/java/org/eolang/inference/Provides.java
maxonfjvipon: Same answer as for the owner locator above. `set-locators.xsl` gives `@loc` to every `<o>` and runs after the splitting, so an attribute without one means the train was reordered — which I'd rather see break at this line than silently produce a row pointing at nothing.

eo-inference/src/main/java/org/eolang/inference/Grouped.java
maxonfjvipon: `index` is written by the same rule that writes the row: `Provides` sets it on every row it adds, before anything can read it. A missing or non-numeric one is a bug in a rule, not input to be defended against, and a `NumberFormatException` naming that row is a better outcome than a report quietly ordered by accident — the order is load-bearing here, since the application rule will ask for the first void.

eo-inference/src/main/java/org/eolang/inference/Grouped.java
maxonfjvipon: `domQuietly()` doesn't swallow anything. It catches the checked `ImpossibleModificationException` and rethrows it unchecked, with the directives in the message and the cause attached — xembly 0.32.2:

```
catch ImpossibleModificationException ->
    String.format("Failed to create DOM quietly: %s", this.directives), ex
  )
```

"Quietly" means no checked exception on a method whose failure can only be a programming mistake. A cell value that is not valid as an XML attribute would blow up on the first test that renders it, and print the directives that caused it.

eo-inference/src/main/java/org/eolang/inference/Inference.java
maxonfjvipon: Fair, and it is the price of the shape the design asks for: one table for the whole program means every prepared document is live at once, or read from disk twice. Reading twice is the way out when it starts to hurt, and eo-runtime is where it will hurt first.

Keeping it as it is for the prototype — this belongs with the ticket that wires the module into the build, where the real input size finally shows up.

eo-maven-plugin/src/main/resources/org/eolang/maven/inference/unroll-bases.xsl
maxonfjvipon: The splitting is tested — `splitsCompositeBase`, `keepsReferenceWhole` and `locatesObjectsBornFromSplitting` exercise nothing else, and each says in its message what the stylesheet got wrong. What's missing is a declarative pack of the kind `eo-parser` keeps, and that needs the pack runner this module doesn't have yet. It's worth building when there are several stylesheets to cover, which the `needs` rule will bring.

eo-maven-plugin/src/main/java/org/eolang/maven/Table.java
maxonfjvipon: It's the convention in this repository rather than a hint about lambdas: eleven interfaces carry the annotation, and `Atom`, `Footprint`, `CommitHash`, `Dependencies` and `Syntax` are all implemented by stateful classes. Dropping it from `Table` would make this module the odd one out.

eo-inference/src/main/java/org/eolang/inference/Cells.java
maxonfjvipon: Checked, and the order is hash order either way, cache or no cache: `TjDefault.add` keeps a row as a `HashMap` and `ToCached` copies it into another one. String hashing is specified, so the same cells come out in the same order on every run and every JVM — the document is reproducible, which is what the docblock promises — but the order is arbitrary rather than meaningful.

It carries no semantics in XML and every assertion here matches by XPath, so I'd rather not sort for the sake of appearances.

eo-inference/src/main/java/org/eolang/inference/Table.java
maxonfjvipon: That sentence was about why the rows are `Tojos`, not about something this PR ships: `MnCsv` and `MnJson` come with the dependency, so a different mono writes a table out as CSV or JSON whenever there's a reason to. Nothing planned, so nothing to puzzle — but it did read as a promise, so I've reworded it, and said in the same breath that caching is already in use while thread-safety isn't.

eo-inference/src/test/java/org/eolang/inference/InferenceTest.java
maxonfjvipon: An empty directory gives `<provides/>`, and there is no behaviour in that path for a rule to get wrong. I'd rather not spend a test on it: the discipline I follow here is to leave irrelevant paths untested, and this one only exercises `Files.walk` returning nothing.

eo-inference/src/main/java/org/eolang/inference/Provides.java
maxonfjvipon: `TjCached` is load-bearing here, and the table is read more than once. `Grouped` runs two `select`s, each with a predicate calling `exists("owner")` per row, then `toMap()` per row and `get("index")` inside two comparators.

Without the cache every one of those lands on `ToMono`, which takes the lock, re-reads the whole mono and linear-scans it for the id (tojos 0.19.1) — so rendering would be quadratic in rows, and worse under the sort. With it, a cell read is a map hit. The decorator is there because it is needed; `Table`'s docblock said otherwise and now says this.

eo-inference/src/main/java/org/eolang/inference/Inference.java
maxonfjvipon: Half of that can't happen: the tables are built from the documents walked in this run, and `provides.xml` is written with `Files.write`, which truncates — so no row of a deleted file survives into it.

The prepared XMIR is the real half, and you're right that nothing removes a file whose source is gone. Deleting from a directory the caller handed over isn't this object's business, so I've said what it expects instead: `output` is expected to be empty, and a file left there by an earlier run would outlive its source.

yegor256:
@maxonfjvipon let's integrate it into our Maven plugin first (in this PR). the functionality has to be wired into existing code base, not stand alone (see: https://www.yegor256.com/2015/02/12/top-down-design.html).

Also, think about this: we already have a stage-tracking mechanism in our `foreign.json` file. We know when a .eo object is registered, then parsed, then compiled, then transpiled. We have directories in the `target/eo/` dir, like `1-parse`, `2-compile`, etc. Maybe we should add one more step and put inference into the existing pipeline: `7-inference`. Instead of inventing a totally new directory management, file management, etc.

maxonfjvipon:
@yegor256 take a look again

yegor256:
@maxonfjvipon qulice job fails, you can fix it. The rest is fine.

yegor256:
@maxonfjvipon wait, now we don't have eo-inference module at all. Why? We need to have `MjInference` and `Inferring` classes in the Maven Plugin. Then, we need to create `eo-inference` module and place checkers there. Each checker should take two directories as input and do the checking:

```
interface Clue {
  void follow(Path xmirs, Path tables);
}
```

Then, the collection of all of them should be `Clues implements Clue`.

Then, the Maven plugin calls the `Clues`.

eo-maven-plugin/src/main/java/org/eolang/maven/Inferring.java
```diff
@@ -0,0 +1,140 @@
...
+        this.input = parsed;
+        this.prepared = pre;
+        this.tables = rows;
+    }
+
+    @Override
+    public void exec() throws IOException {
+        if (Files.exists(this.input)) {
+            final int ready = this.ready();
+            new Clues().follow(this.prepared, this.tables);
```
Favixx, line 88: This is the line the refactor turned around, and it is worth reconsidering before the other two clues arrive.

`ready()` above parses every XMIR of the program, runs two stylesheets over it and writes the result out. Then `Clues.follow` is handed the *path*, and `Xmirs` walks that directory and parses all of it again — once for every clue in the collection. With one clue that is one redundant parse of the whole program; with the three the design calls for it is three, on top of the write.

`Clue.follow(Path, Path)` is what forces it. `follow(Collection<XML>, Path)` would keep the seam exactly as clean — a clue still reads a program and writes a table — while letting `Inferring` pass what it already has in memory, and `Xmirs` would become the thing that produces that collection, called once, by `Inferring`. The prepared directory would still be written for the humans who need to look at it, which is what its Javadoc says it is for.

Since `Clue` is public API and two more implementations are about to be written against it, this is much cheaper to change now than later.

eo-maven-plugin/src/main/java/org/eolang/maven/Inferring.java
```diff
@@ -0,0 +1,140 @@
...
+final class Inferring implements Step {
+
+    /**
+     * The directory with the XMIR files of the program.
+     */
+    private final Path input;
+
+    /**
+     * The directory for the prepared XMIR files, which keep the names they
+     * had; nothing is deleted from it, so a file left there by an earlier run
```
Favixx: This sentence was true and harmless before the move; it is now a description of a bug.

When `Provides` was given the return value of `ready()`, a file left in this directory by an earlier run polluted the directory and nothing else — the table was built from the documents just parsed. Now `Clues` is given this path and `Xmirs.sources()` does `Files.walk` over all of it, so an `.xmir` whose `.eo` was renamed or deleted three commits ago is read back in and its formations land in `provides.xml` as part of the current program.

That is not a corner case: renaming a source file is the ordinary way to hit it, and the result is a table that names types that do not exist, with no error anywhere.

Deleting the directory at the start of `exec()` fixes it and costs one line. Keeping the documents in memory (the previous comment) fixes it too, and makes the directory purely a debugging artefact — at which point this sentence becomes true again.

eo-inference/src/main/java/org/eolang/inference/Clue.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Read the program and write down what is understood.
+     * @param xmirs The directory with the prepared XMIR of the whole
+     *  program, since an object in one file is used by objects in others
+     *  and no single file can be typed on its own
+     * @param tables The directory for the tables, a document each
+     * @throws IOException If a file cannot be read or written
+     */
+    void follow(Path xmirs, Path tables) throws IOException;
```
Favixx, line 40: Good interface — "read the program, write a table" is exactly the right size for a rule, and the parameter documentation earns its length by explaining why a clue gets the whole program rather than a file.

My one reservation is the first parameter's type, for the reasons in `Inferring`: a `Path` obliges every implementation to re-read and re-parse, and there will be three of them. A `Collection<XML>` costs the interface nothing in clarity and saves each implementation the same work.

One smaller thing: `follow` returns `void`, so a clue's only output is a file. That is fine while the tables are independent, but the design note says the third clue records "which types are copies of which", and the checking loop will have to read all three back. Worth deciding now whether that reader parses the XML again or whether `follow` should hand back its `Tojos` — because the answer changes this signature, and changing it after three implementations exist is three times the work.

eo-inference/src/main/java/org/eolang/inference/Clues.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param clues The clues to follow
+     */
+    public Clues(final Collection<Clue> clues) {
+        this.all = clues;
+    }
+
+    @Override
+    public void follow(final Path xmirs, final Path tables) throws IOException {
+        for (final Clue clue : this.all) {
+            clue.follow(xmirs, tables);
```
Favixx, line 67: Every clue is given the same `tables` directory, and each one chooses its own file name inside it — `Provides` writes `provides.xml`. Nothing stops two clues from choosing the same name, and if they do, the second silently overwrites the first and the failure appears much later, as a table with the wrong contents.

With one clue it cannot happen. With three, written by three different people over three PRs, it is worth one line of defence: either the name comes from the clue as a value this class reads, so a collision is visible here, or `follow` is given the file rather than the directory.

eo-inference/src/main/java/org/eolang/inference/Clues.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * The clues to follow, in the order they are followed.
+     */
+    private final Collection<Clue> all;
+
+    /**
+     * Ctor.
+     */
+    public Clues() {
+        this(Collections.singletonList(new Provides()));
```
Favixx, line 53: Two constructors where the no-arg one names the real set and the other takes any collection — the right pattern, and `CluesTest` uses both, which is how it should be.

One consequence of the visibility split worth noting: `Clue` and `Clues` are public, `Provides` is package-private. So a caller outside `org.eolang.inference` can implement `Clue` and pass a collection, but cannot include the one clue that actually exists — `new Clues(Arrays.asList(new Provides(), myOwn))` does not compile outside this package. If extending the set from outside is not intended, that is fine and the varargs constructor is really a test seam, which would be worth saying in its Javadoc. If it is intended, `Provides` needs to be public.

eo-inference/src/main/java/org/eolang/inference/Xmirs.java
```diff
@@ -0,0 +1,93 @@
...
+     * bytes as text, and the {@code λ} marker of an atom, which names a
+     * body implemented in Java.</p>
+     *
+     * @return The formations, file by file, in the order they appear in
+     *  the code
+     * @throws IOException If a file cannot be read
+     */
+    Collection<XML> formations() throws IOException {
+        final Collection<XML> found = new ArrayList<>(0);
+        for (final XML xmir : this.documents()) {
```
Favixx, line 57: `documents()` builds a `Collection<XML>` holding every parsed file of the program, and `formations()` then holds every matching node from all of them, so peak memory is the whole program twice over. For `eo-runtime`, which after the change to its `pom.xml` runs this goal on every build, that is 182 `.eo` files' worth of XMIR resident at once.

It also throws away the file each formation came from. `Provides` only needs the locator, so nothing is broken — but the next clue that wants to report a mistake will want to say which file it is in, and by then this collection is flat and anonymous. Streaming the documents, or keeping them paired with their paths, is easier to do now than to retrofit.

eo-inference/src/main/java/org/eolang/inference/Xmirs.java
```diff
@@ -0,0 +1,93 @@
...
+        return read;
+    }
+
+    /**
+     * Every XMIR file of the program, in the same order every time.
+     * @return The files
+     * @throws IOException If the directory cannot be walked
+     */
+    private Collection<Path> sources() throws IOException {
+        try (Stream<Path> found = Files.walk(this.dir)) {
```
Favixx, line 86: `Files.walk` with no depth limit over a directory that `Inferring` never cleans — this is where a stale file from an earlier run enters the table. See the note on `Inferring`'s `prepared` field; the fix belongs there rather than here, but this is the line that reads it.

The `.sorted()` is right and the Javadoc's reason for it — same order every time, so the tables come out the same every time — is the correct reason. Worth noting it sorts `Path`, so the order is by path string, which is stable across runs on one machine but not across platforms with different separators. For a table that is compared between a developer's machine and CI that could matter one day.

eo-maven-plugin/src/main/java/org/eolang/maven/MjInference.java
```diff
@@ -0,0 +1,86 @@
...
+ *     Then the checks are drained one by one, each of them either deciding,
+ *     splitting into smaller checks, or waiting for facts that may never come.
+ *     A mistake is reported only when the object that misses an attribute is
+ *     complete, so that parts of the program this goal cannot see — atoms,
+ *     delegation through {@code φ} — make it silent rather than wrong.
+ * </p>
+ * <p>
+ *     Only the first of those rules is implemented so far, together with the
+ *     preparation the rest of them will need. The XMIR prepared for the rules
+ *     is saved in {@link #preInferenceDir} and the tables in
```
Favixx: These two links are broken: there is no `preInferenceDir` and no `inferenceDir` in this class. The fields are `prepared` (line 66) and `tables` (line 76). `{@link #name}` resolves against the members that exist, so this is a reference to nothing rather than a stale word in prose.

Which side is wrong is the interesting part. The `@Parameter` properties are `eo.preInferenceDir` and `eo.inferenceDir`, and the defaults are `6-pre-inference` and `6-inference` — three names out of four say `preInferenceDir`/`inferenceDir` and only the fields disagree. Renaming the fields, rather than rewriting the Javadoc, also fixes `FakeMaven`, which sets parameters by field name and is currently forced to use `prepared` and `tables` as keys shared with every other mojo.

eo-maven-plugin/src/main/java/org/eolang/maven/MjInference.java
```diff
@@ -0,0 +1,86 @@
...
+)
+public final class MjInference extends MjSafe {
+
+    /**
+     * The directory where the XMIR prepared for the rules is saved.
+     */
+    @Parameter(
+        property = "eo.preInferenceDir",
+        required = true,
+        defaultValue = "${project.build.directory}/eo/6-pre-inference"
```
Favixx, line 64: `6-pre-inference` and `6-inference`, but `inference` runs before `transpile`, whose directories are `5-pre-transpile` and `5-transpile`. The list in `check-folders-numbering.groovy` is `1-parse`, `2-pull`, `3-lint`, `4-resolve`, `5-pre-transpile`, `5-transpile` — every number in the order its step runs, which is the only thing the numbering is for.

The `pre-X`/`X` pairing at one number is right and matches `5-pre-transpile`/`5-transpile`; it is the digit that is off. Either these become `5-` and transpile moves to `6-`, or `inference` moves after `transpile` in the goal list — but that contradicts this class's own Javadoc, which says it should run after `assemble`, not after transpilation.

eo-inference/src/main/java/org/eolang/inference/Provides.java
```diff
@@ -0,0 +1,98 @@
...
+    public void follow(final Path xmirs, final Path tables) throws IOException {
+        final Tojos rows = new TjCached(new TjDefault(new MnMemory()));
+        int seen = 0;
+        for (final XML formation : new Xmirs(xmirs).formations()) {
+            final String owner = formation.xpath("@loc").get(0);
+            rows.add(owner)
+                .set("index", Integer.toString(seen))
+                .set("complete", Boolean.toString(formation.nodes("o[@name='λ']").isEmpty()));
+            seen = seen + 1;
+            for (final XML attr : formation.nodes("o[@name and not(@name='λ')]")) {
```
Favixx, line 79: `o[@name and not(@name='λ')]` picks up `φ` along with everything else, so a formation with `inc t > @` in it gets an attribute row named `φ` whose type is the locator of an application. The class Javadoc is explicit that applications are deliberately absent because "what they have is what the object they copy has, which is the business of the links table" — but through `φ` one arrives here anyway, as an attribute of the enclosing formation.

`ProvidesTest.skipsApplications` asserts `/provides[not(type[@id='Φ.lamp.φ'])]`, which is true and is about type rows. The attribute row is still there: `/provides/type[@id='Φ.lamp']/attr[@name='φ']` matches that same document.

That may be exactly what you want — `φ` really is an attribute of the formation, and the fact that its type is an application is the links table's problem. If so, say it in the paragraph listing what is absent, because that paragraph currently reads as though it were not. If not, `not(@name='φ')` belongs next to the `λ` exclusion.

eo-inference/src/main/java/org/eolang/inference/Provides.java
```diff
@@ -0,0 +1,98 @@
...
+                final String name = attr.xpath("@name").get(0);
+                final Tojo row = rows.add(String.join(" ", owner, name))
```
Favixx, line 81: A space as the separator in a composite key works only for as long as neither half can contain one. That holds today, and the Javadoc explains well why the key has to be composite at all. What it does not do is protect the invariant: if a locator ever grows a space, `Φ.a b` and `Φ.a b c` stop being distinguishable, and the failure is a wrong row rather than an exception.

A private method that builds the key would give the convention one home and one place to document it — and `Grouped` would have somewhere to ask, instead of knowing by coincidence that an attribute row's `id` is bookkeeping it should drop.

eo-inference/src/main/java/org/eolang/inference/Provides.java
```diff
@@ -0,0 +1,98 @@
...
+                    .set("owner", owner)
+                    .set("index", Integer.toString(seen))
+                    .set("name", name)
+                    .set("type", attr.xpath("@loc").get(0));
+                if (attr.xpath("@base").contains("∅")) {
```
Favixx, line 86: `xpath()` returns the list of matched attribute values, so this is `List.contains` — an exact match on a one-element list, which is correct but reads like a substring test on the base. A reader has to know what `xpath` returns to see that `base="∅x"` would not match.

`!attr.nodes("self::o[@base='∅']").isEmpty()` says the same thing in the same language as the two XPaths above it.

eo-inference/src/main/java/org/eolang/inference/Provides.java
```diff
@@ -0,0 +1,98 @@
...
+                    row.set("void", "true");
+                }
+                seen = seen + 1;
+            }
+        }
+        Files.createDirectories(tables);
```
Favixx, line 92: Building the rows and rendering them are now one method, so the table exists only as a side effect on disk and nothing can get at it programmatically. That is fine for a clue whose product is a document — but the design note says the checking loop will drain checks against these tables, and today the only way for it to read this one is to parse `provides.xml` back.

Since the rows are already `Tojos`, which is a queryable table, going out through XML and back in loses exactly the thing that made `Tojos` the right choice. Worth deciding, before the second clue copies this shape, whether `follow` should also hand the rows back.

eo-inference/src/main/java/org/eolang/inference/Grouped.java
Favixx, line 103: This is the class knowing a column by name in the strongest sense: `index` is not optional, it must be present and must parse as an integer, or the view throws. The Javadoc two screens up says the opposite — "Knowing no other column by name is what lets one view serve more than one table" — and it is that promise which justifies rendering the needs table with the same class.

The promise is cheap to keep: fall back when the cell is absent, and a table written by a rule with no natural order still renders. As written, the first clue that forgets an `index` gets a `NumberFormatException` out of a class whose entire purpose is not to care what the columns are.

eo-inference/src/main/java/org/eolang/inference/Cells.java
Favixx, line 51: The order of the attributes in the output is the iteration order of whatever map `Tojo.toMap()` returns. If that is a `HashMap`, `provides.xml` comes out with its attributes in an order nobody chose, and adding a cell can reshuffle the ones already there.

It does not affect the XPath assertions in the tests, and attribute order is semantically nothing. It does affect a human diffing two runs, which is the stated purpose — `Grouped` is "a document a human can read". Copying into a `TreeMap` here costs one line and makes the output diffable.

eo-inference/src/test/java/org/eolang/inference/GroupedTest.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Test case for {@link Grouped}.
+ * @since 0.67.0
+ */
+final class GroupedTest {
+
+    @Test
+    void nestsAttributeUnderItsOwner() {
+        final Tojos rows = new TjDefault(new MnMemory());
+        rows.add("Φ.jar").set("complete", "true");
```
Favixx, line 23: None of the four tests here sets an `index`, and all four still pass — because each has one type row and at most one attribute per owner, and `Stream.sorted()` on one element never calls the comparator. Add a second row to any of these and `Integer.parseInt(row.get("index"))` in `Grouped.types()` decides what happens next.

So the ordering is untested, and it is the most-argued-for property in this PR: `Provides` spends a paragraph on why a table "promises nothing about order" while "the order of void attributes decides what an application binds to what", and both of `Grouped`'s methods sort. A test with two attributes written out of order, asserting the document has them in `index` order, would cover the counter in `Provides.follow` and both comparators here at once.

eo-inference/src/test/java/org/eolang/inference/CluesTest.java
```diff
@@ -0,0 +1,57 @@
...
+ * @since 0.67.0
+ */
+@ExtendWith(MktmpResolver.class)
+final class CluesTest {
+
+    @Test
+    void followsEveryClue(@Mktmp final Path temp) throws IOException {
+        new Clues(
+            Arrays.asList(
+                (xmirs, tables) -> Files.writeString(
```
Favixx, line 29: Using lambdas as clues is the right way to test the composition, and `followsEveryClue` is honest about what it checks.

It checks a little less than its name, though: asserting that `second.xml` exists proves the last clue ran, not that every one did, and with two clues those happen to coincide only because the first one is not asserted at all. Asserting both files, or asserting the *order* — two clues appending to the same file, then reading it back — would cover what `Clues` actually promises, which its own field Javadoc states as "in the order they are followed".

Order is worth pinning here rather than anywhere else, because the design note has the third clue depending on what the first two wrote.

.github/workflows/counts.yml
```diff
@@ -24,7 +24,7 @@ jobs:
           - metric: smells
             files: '*.java'
             pattern: 'static|null'
-            count: 844
+            count: 845
```
Favixx, line 27: This +1 still does not belong to this PR, and I checked again at `8e514ce` after the move into `eo-inference`. The ten new Java files contain no occurrence of `static` or `null`, and neither do the six lines added to `FakeMaven` — nothing here moves the count. `master` is already at 845 while this branch's base is at 844, so the increment arrived from elsewhere after the branch was cut.

Harmless at merge time, since git sees the same value on both sides. But it is a line of unrelated churn, and the commit that made it (`ci: match the smells count to the code base`) describes a cause that is not there.

eo-runtime/src/test/groovy/check-folders-numbering.groovy
```diff
@@ -15,6 +15,8 @@ List<String> allowed = [
     '4-resolve',
     '5-pre-transpile',
     '5-transpile',
+    '6-pre-inference',
```
Favixx, line 18: The numbering point from the other side: this list is the definition of the convention, and adding `6-` entries for a step that runs fifth is what makes it stop meaning anything.

Separately, `6-pre-inference` and `6-inference` are now literals in four places — here, both `defaultValue`s in `MjInference`, `FakeMaven`, and twice in `MjInferenceTest`. `Inferring` is already constructed with `Parsing.DIR` for its input, so the precedent for a constant is one line from where these are used.

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
```diff
@@ -187,6 +187,12 @@ <T extends AbstractMojo> FakeMaven execute(final Class<T> mojo) throws IOExcepti
                 "cache", this.workspace.resolve("eo/cache/parsed").toFile()
             );
             this.params.putIfAbsent("generatedDir", this.generatedPath().toFile());
+            this.params.putIfAbsent(
+                "prepared", this.targetPath().resolve("6-pre-inference").toFile()
```
Favixx, line 191: `prepared` and `tables` are two of the most generic words available, and `this.params` is shared by every mojo `FakeMaven` executes — `putIfAbsent` here means any current or future mojo with a field of either name silently receives a directory meant for inference. Compare the neighbours: `generatedDir`, `placedFormat`, `plugin`, all unambiguous because the fields they set are.

Renaming the fields in `MjInference` to `preInferenceDir` and `inferenceDir` — what its Javadoc, both properties and both defaults already call them — removes this rather than working around it.

eo-inference/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.jcabi.incubator</groupId>
+      <artifactId>xembly</artifactId>
+      <!-- version from parent POM -->
+    </dependency>
+    <dependency>
+      <groupId>com.yegor256</groupId>
+      <artifactId>tojos</artifactId>
+      <version>0.19.1</version>
```
Favixx, line 31: `tojos` is pinned to `0.19.1` here and to `0.19.1` in `eo-maven-plugin/pom.xml` as well, while every other dependency in this new POM says `<!-- version from parent POM -->`. So the split has created a second place to remember, and the two will drift the first time one of them is bumped — at which point Maven's nearest-definition rule silently picks one for the plugin, which now depends on this module.

Since the parent already manages the rest, moving `tojos` into its `dependencyManagement` and dropping both `<version>`s would leave one place to change and match what the neighbours in this file already do.

eo-maven-plugin/src/main/resources/org/eolang/maven/inference/unroll-bases.xsl
```diff
@@ -0,0 +1,52 @@
...
+  moves down into the shorter base together with it.
+  -->
+  <xsl:output encoding="UTF-8" method="xml"/>
+  <xsl:template match="o[contains(replace(@base, '\.[^.]*$', ''), '.')]">
+    <xsl:variable name="dotted" select="starts-with(@base, '.')"/>
+    <xsl:variable name="receiver" as="element()">
+      <o base="{replace(@base, '\.[^.]*$', '')}">
+        <xsl:copy-of select="@line|@pos"/>
+        <xsl:if test="$dotted">
+          <xsl:copy-of select="o[1]"/>
```
Favixx, line 35: Two questions about this branch, neither of which I could settle by reading.

`xsl:copy-of` on `o[1]` is a deep copy without template application, and the copy is re-processed because `$receiver` goes through `xsl:apply-templates` seven lines below — so a composite base inside the moved child is still unrolled. That is correct, but it depends on the order of two lines some distance apart, and it survives a refactor only if a test says so. There is no test for a dotted base whose receiver has a chain of its own.

The second question is `@loc`. Line 40 copies every attribute except `@base` onto the new outer object, `@loc` among them, and that outer object is the dispatch, not the whole chain it came from. `set-locators.xsl` runs afterwards, so if it overwrites unconditionally this is harmless; if it only fills in what is missing, the outermost dispatch keeps the locator of the rolled-up chain and every row pointing at it points one step too far out. `InferringTest.locatesObjectsBornFromSplitting` asserts the locator of the innermost `ξ.x` and says nothing about the outer two.

eo-maven-plugin/src/test/java/org/eolang/maven/InferringTest.java
```diff
@@ -0,0 +1,164 @@
...
+            new EoSyntax(
+                String.join(
+                    System.lineSeparator(),
+                    "[x] > inc",
+                    "  x.next.foo > @",
+                    ""
+                )
+            ).parsed().toString()
+        );
+        new Inferring(input, temp.resolve("out"), temp.resolve("rows")).exec();
```
Favixx, line 41: Six tests and none takes the other branch of `exec()`. `Files.exists(this.input)` being false is a real path — it is what happens in a module with no `.eo` files, which after the change to `eo-runtime/pom.xml` is every module inheriting the plugin — and its behaviour is a log line and no output directory. Four lines would cover it.

The other gap is the one the move made worse: nothing here runs `Inferring` twice over the same temporary directory. Since `Clues` now reads the prepared directory from disk rather than the documents just parsed, a second run with a source removed in between is exactly the case where a stale `.xmir` shows up in the table, and it is as cheap to write as any other test in this file.

Favixx (changes requested):
Reviewed at `8e514ce`, after the move into `eo-inference` behind `Clue`. The split is right — the checker has no business living in a Maven plugin, and `Clue` is the correct seam, since a rule really is "read the program, write a table" and nothing more. `Clues` composing them in order, with the varargs constructor for tests and a no-arg one naming the real set, is the shape I would have asked for.

The move introduced one regression, and it turned a documented untidiness into a correctness bug.

**`Inferring` now hands `Clues` a directory instead of the documents it just parsed.** Before, `ready()` returned `Collection<XML>` and `Provides` consumed it. Now `ready()` writes the prepared XMIR to disk, returns a count, and `Xmirs` walks that directory and parses everything again. Two consequences. Every document is parsed, serialised, written, re-read and re-parsed on every run — and once per clue, so with the three clues the design calls for, the whole program is parsed four times instead of once. And the `prepared` field's own Javadoc says that nothing is ever deleted from that directory, which used to be merely untidy because the table was built from the return value; now `Xmirs` reads the directory, so a stale `.xmir` from an earlier run lands in the table as though it were part of the program.

The rest of my notes survive the move unchanged:

**The two `{@link}`s in `MjInference` still point at fields that do not exist** — `#preInferenceDir` and `#inferenceDir` against `prepared` and `tables`, which is also why `FakeMaven` has to inject two parameters under those generic names.

**The directory numbers still contradict the order the goals run in** — `6-` for a step that runs fifth, between `compile` and `transpile`.

**`Grouped` still requires an `index` in every row**, which its Javadoc says it does not know about, and `GroupedTest` still cannot notice because every one of its groups has a single row and a one-element sort never calls the comparator.

**The `counts.yml` bump still does not belong to this PR.** The ten new Java files and the six added lines in `FakeMaven` contain no occurrence of `static` or `null`; `master` is already at 845 while this branch's base is 844.

Details inline, including the `φ` question on `Provides`, which the move has made easier to answer.

maxonfjvipon:
@yegor256 take a look once again

eo-maven-plugin/src/main/java/org/eolang/maven/Inferring.java
maxonfjvipon: Fixed in `e3b159f6`. You're right that the refactor turned a harmless sentence into a bug: `Clues` reads the directory back, nothing emptied it, so a renamed source left its types in the table. `Inferring.exec` now empties `prepared` before writing into it, and `InferringTest.forgetsSourceThatIsGone` runs the step twice with a file deleted in between — it fails without the fix, with `Φ.rake` still in the document.

eo-maven-plugin/src/test/java/org/eolang/maven/InferringTest.java
maxonfjvipon: The second gap is covered — `forgetsSourceThatIsGone` runs `Inferring` twice over the same directory with a source removed in between, which is the case the move made possible, and it caught a real bug (`e3b159f6`).

Leaving the absent-input branch untested: it logs a line and does nothing, which is behaviour I'd rather not pin down in a test at this stage.

eo-maven-plugin/src/main/java/org/eolang/maven/MjInference.java
maxonfjvipon: The links were indeed dangling — fixed in `e3b159f6`, they now point at `prepared` and `tables`.

Keeping the field names as they are, though. `preInferenceDir` is camelCase, which needs a `@checkstyle MemberNameCheck` comment on each field, and the `counts` gate demands an exact number of those across the repository — the rename costs two suppressions and a threshold bump to buy a longer name. The properties stay `eo.preInferenceDir` and `eo.inferenceDir`, so nothing about configuring the goal changes.

eo-maven-plugin/src/main/resources/org/eolang/maven/inference/unroll-bases.xsl
maxonfjvipon: Checked both.

`@loc` is safe: `set-locators.xsl` matches every `o`, writes `loc` unconditionally and copies `@* except @loc`, so whatever `unroll-bases` carried over is overwritten. The prepared XMIR of `eo-runtime` confirms it — a split chain in `u64.xmir` comes out as `Φ.u64.as-number.φ` / `…φ.ρ` / `…φ.ρ.ρ`, one step per dispatch.

The deep-copied receiver is re-processed, as you read it: in that same file the moved child `Φ.i64` has its own `α0` split into `.as-bytes` over `ξ.ρ`. So a chain inside a moved receiver does get unrolled — it is exercised by `eo-runtime` on every build, though not yet pinned by a test of its own.

eo-maven-plugin/src/main/java/org/eolang/maven/Inferring.java
maxonfjvipon: Keeping `Path`. The signature is what @yegor256 asked for — a checker takes two directories — and the re-parse is the price of that seam. Worth revisiting when the second clue lands and the cost is measurable rather than predicted; at 170 files it is not visible next to the transpiler.

eo-inference/src/main/java/org/eolang/inference/Clue.java
maxonfjvipon: Same answer on the parameter type as in `Inferring` — the two directories are the shape the review asked for.

On `void`: the checking loop is a later ticket, and whether it reads the tables back or takes them in memory is exactly the kind of question that is better answered by writing it than by guessing now. The rows are `Tojos`, so handing them back stays available.

eo-inference/src/main/java/org/eolang/inference/Clues.java
maxonfjvipon: Real but distant: with one clue it cannot happen, and the second one will be written against a class that already names its file. I would rather add the guard together with the clue that makes it possible than invent the naming scheme before there is a second name.

eo-inference/src/main/java/org/eolang/inference/Clues.java
maxonfjvipon: Not intended for outside extension — `Clues` is the entry point and the collection ctor is a seam for tests, as `CluesTest` uses it. Leaving `Provides` package-private; if a clue ever needs to come from outside this package, making it public is the smaller change then than guessing now.

eo-inference/src/main/java/org/eolang/inference/Xmirs.java
maxonfjvipon: Noted, and it is inherent to one table for the whole program: the rows are keyed by locator, so the clue has to see every file before it can write anything. 170 documents is not where this hurts.

The file a formation came from is recoverable — the locator names the object and the prepared XMIR is kept on disk for exactly that — so the clue that reports mistakes has a way back without pairing paths here.

eo-inference/src/main/java/org/eolang/inference/Xmirs.java
maxonfjvipon: The staleness half is fixed at the source, in `Inferring` (`e3b159f6`) — this line now walks a directory that was emptied moments ago.

On separators: `Path` sorts by its own comparator, not the string, and the walk yields paths under one root, so the order is the tree order on any platform. It would take files whose names differ only by separator-adjacent characters for a platform to disagree.

eo-maven-plugin/src/main/java/org/eolang/maven/MjInference.java
maxonfjvipon: The digit is deliberate. `5-transpile` was the last number in use and 6 was the next free one; `9-place` already sits above it, so the sequence is not dense to begin with. @yegor256 proposed `7-inference` for the same step, which puts us in the same place.

Renumbering transpile to make room is a change to a directory other tools and people know by name, for a digit.

eo-inference/src/main/java/org/eolang/inference/Provides.java
maxonfjvipon: That is what I want: `φ` is an attribute of the formation and its row says so, while the application it happens to be bound to gets no row of its own — which is what the paragraph means by applications being absent. The type of that attribute is a locator the links table will have something to say about.

eo-inference/src/main/java/org/eolang/inference/Provides.java
maxonfjvipon: A locator cannot contain a space — it is dots and object names, and a name with a space would not have parsed. So the invariant holds by construction rather than by convention, and I would rather not add a method to defend something the parser already guarantees.

eo-inference/src/main/java/org/eolang/inference/Provides.java
maxonfjvipon: Fair reading, but `xpath` returning a list is the same fact the two lines above rely on, and `contains` on that list is an exact match on one value. Swapping it for a `nodes()` call trades one thing to know for another.

eo-inference/src/main/java/org/eolang/inference/Provides.java
maxonfjvipon: Deliberately a document for now. What the checking loop needs is a question for the ticket that writes it, and the answer may well be to hand the rows back — they are `Tojos` precisely so that stays cheap. Adding a return value before there is a reader would be guessing at its shape.

eo-inference/src/main/java/org/eolang/inference/Grouped.java
maxonfjvipon: `index` is written by the same clue that writes the row, on every row it adds, so a missing one is a bug in a rule rather than input to defend against — and this repository favours failing fast over failing safe. The promise the Javadoc makes is about what the view reports, not about tolerating a half-built table.

eo-inference/src/main/java/org/eolang/inference/Cells.java
maxonfjvipon: It is a `HashMap`, so the order is hash order — but the same order on every run and every JVM, since String hashing is specified. Two runs of the same table therefore diff cleanly, which is the property that matters here. Adding a cell can indeed reshuffle the rest, and that is the day to reach for a `TreeMap`.

eo-inference/src/test/java/org/eolang/inference/GroupedTest.java
maxonfjvipon: True — the ordering is asserted only indirectly, through `InferringTest` and `MjInferenceTest`, where real programs produce several attributes per owner. A `GroupedTest` case with two attributes written out of order is a good addition; noting it rather than growing this PR further.

eo-inference/src/test/java/org/eolang/inference/CluesTest.java
maxonfjvipon: Right that it proves the last clue ran rather than every one. Since `Clues` is four lines of `for`, I am leaving the test as it is; the order it promises will be worth pinning when a clue depends on what an earlier one wrote, which is the point at which the promise starts carrying weight.

.github/workflows/counts.yml
maxonfjvipon: Correct, and it was flagged as such when it went in: `master` is at 845 and its own `counts` runs have been red since the bump was missed there. This line makes the branch green without touching anything else; whoever fixes it on `master` first wins and the other side sees the same value.

eo-runtime/src/test/groovy/check-folders-numbering.groovy
maxonfjvipon: On the digit, see the reply in `MjInference` — 6 was the next free number and `9-place` already breaks density.

On the literals: agreed that four places is three too many. The awkward part is that a Maven `defaultValue` has to be a literal string, so a constant can only ever cover the other three — worth doing when the second stage-dir constant appears.

eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java
maxonfjvipon: The collision is real in principle: `putIfAbsent` reaches any mojo with a field of that name, and there is none today. The rename that removes it costs two `@checkstyle MemberNameCheck` suppressions, which the `counts` gate counts exactly — see the reply in `MjInference`. Keeping the names, watching for a neighbour.

eo-inference/pom.xml
maxonfjvipon: Good catch on the drift. `tojos` was already pinned in the plugin before this PR, so the second pin is inherited rather than new, but the two can now disagree. Moving it into the parent's `dependencyManagement` touches a file every module shares, so I would rather do it on its own than inside this one.

yegor256 (approved):

yegor256:
@maxonfjvipon looks good now, thanks!

0crat:
@yegor256 Thanks for the review! You've earned +22 points for this: +12 as a basis; +14 for the 1480 hits-of-code that you reviewed; -5 for very few (1) comments.  Your running score is [+2871](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +24 points for this: +12 as a basis; +14 for the 1480 hits-of-code that you reviewed; -3 to not go over the cap.  Your running score is [+3789](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +24 points for this: +12 as a basis; +14 for the 1480 hits-of-code that you reviewed; +5 for the 23 comment(s) that you have made; -9 to not go over the cap.  Your running score is [+172](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +5 points for this: +16 as a basis; -4 for too many hits-of-code (1480 >= 100); -8 for way too many hits-of-code (1480 >= 400); -8 for too many (85 > 4) comments that were made during review; +9 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+694](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6361, opened 2026-08-04
Bring `qulice` back to green across the reactor

### LINKED ISSUE #6360: `qulice` fails on `master` because of seven violations in three modules
The `qulice` job fails on `master` and has done for at least the last
five runs, the newest of them at `54d8d390a9`. Three Checkstyle
violations in `eo-parser` stop the build:

```text
Emissions.java[618]: Javadoc '@throws NumberFormatException' is not
  declared in method signature (JavadocThrowsCheck)
Suffix.java[459]: Cyclomatic Complexity is 11 (max allowed is 10)
Suffix.java[459]: NPath Complexity is 288 (max allowed is 200)
```

The first one arrived with `bd30928cde`, which taught the parser to
reject an unrecognised string escape. The new `singleCharEscape` throws
`NumberFormatException` and documents it with an `@throws` tag, while the
signature declares nothing — and it is the only `@throws` for an
unchecked exception anywhere in `eo-parser`. Its two neighbours in the
same file, `appendOctal` and `appendUnicode`, throw the very same
exception and carry no tag, so dropping the tag both silences the check
and matches the file it lives in.

The other two are `Suffix.auto`, which parses a `>>` suffix. Nothing is
wrong with what it does; it simply asks three questions of the form
"is there a character here, and is it the one I want":

```java
if (idx < tail.length() && tail.charAt(idx) == '!') {
```

Each such pair costs a branch, and the same file already spells that
exact question with one call — `tail.startsWith("+>", idx)` at line 323,
where `startsWith` does the bounds check itself. Writing the three
conditions as `tail.startsWith("!", idx)` and friends brings the
cyclomatic complexity to 8 and the NPath count to 36, with no new
methods and no change in behaviour.

Because the workflow also runs on `pull_request`, every open pull
request is red on `qulice` right now, whatever it touches.

maxonfjvipon:
Seven Checkstyle and PMD violations have been failing `qulice` on
`master`, which turns the check red on every open pull request as well,
whatever it touches. They sit in three modules, and because Maven stops
at the first module that fails, `eo-parser` was hiding the rest — the
full list came out of a `--fail-at-end` run of the whole reactor.

`Emissions.singleCharEscape` documented a `NumberFormatException` that
its signature does not declare. It was the only `@throws` for an
unchecked exception in the module, and `appendOctal` and `appendUnicode`
throw the very same exception a few lines above without one, so the tag
is gone rather than the signature changed.

Two more are complexity limits on `Suffix.auto`, which asked three times
whether a given character sits at a given index and spelled each
question as a bounds check plus a comparison. The same file already asks
it in one call:

```java
if (tail.startsWith("!", idx)) {
```

`startsWith` does the bounds check itself, exactly as it already does at
line 323 for `+>` and `>>`. Cyclomatic complexity drops from 11 to 8 and
the NPath count from 288 to 36. Extracting a method would have been the
usual way to cut complexity, and it is the one way that cannot be used
here: the `thresholds` job pins the number of `static` members across
the whole codebase to master's exact count, so a new one would turn that
job red instead.

`CentralMavenTest` held a lambda in a local variable it used once, which
is what `UnnecessaryLocalRule` objects to. The last two are layout in
eo-runtime: a fluent `append` opening a block on a line of its own, and
an expected string of eight repeated escapes running to 124 characters,
now built with `repeat(8)`, which fits and says how many of them there
are.

Nothing changes behaviour. The three rewritten branches in `Suffix` are
all exercised by `SuffixTest` through `>>!`, `>> fibo!`, `>> /number`
and a bare `>>` — that last one being exactly the end-of-tail case the
explicit guard used to cover. All 1670 parser tests pass, the touched
tests in the other two modules pass, `qulice` is clean on all six
modules, and the four counts the `thresholds` job pins are untouched at
840, 364, 0 and 26.

Closes #6360

### DIFF
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/CentralMavenTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/CentralMavenTest.java
@@ -9,7 +9,6 @@ import com.yegor256.MktmpResolver;
 import com.yegor256.WeAreOnline;
 import java.nio.file.Path;
 import java.util.Collections;
-import java.util.function.BiConsumer;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.repository.internal.MavenRepositorySystemUtils;
 import org.eclipse.aether.DefaultRepositorySystemSession;
@@ -121,11 +120,12 @@ final class CentralMavenTest {
 
     @Test
     void composesWithAndThenInsteadOfThrowing() {
-        final BiConsumer<Dependency, Path> after = (dep, dest) -> {
-        };
         MatcherAssert.assertThat(
             "andThen must return a composed BiConsumer, not throw at composition time",
-            new CentralMaven().andThen(after),
+            new CentralMaven().andThen(
+                (dep, dest) -> {
+                }
+            ),
             Matchers.notNullValue()
         );
     }
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -615,7 +615,6 @@ final class Emissions {
      * @param head Backslash character (always {@code '\\'})
      * @param next The character after the backslash
      * @return The decoded character(s)
-     * @throws NumberFormatException If the escape is not recognised
      */
     private static String singleCharEscape(final char head, final char next) {
         final String decoded;
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -461,7 +461,7 @@ final class Suffix {
     ) {
         int idx = after;
         boolean cnst = false;
-        if (idx < tail.length() && tail.charAt(idx) == '!') {
+        if (tail.startsWith("!", idx)) {
             cnst = true;
             idx = idx + 1;
         }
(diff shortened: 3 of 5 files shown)


Thayorns (approved):
This clears seven Checkstyle and PMD violations across three modules without touching behaviour. The three rewritten guards in Suffix.auto are exactly equivalent to the bounds-check-plus-comparison they replace, since startsWith already returns false past the end of the string, and the reworked expected value in VerboseBytesAsStringTest reproduces the original literal character for character. Nothing here changes what any of the touched code does.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (26) hits-of-code; +4 to give you at least something.  Your running score is [+805](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+754](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6377, opened 2026-08-04
Make directory report is-directory from a syscall

### LINKED ISSUE #5440: `dir.is-directory` attribute inaccurately returns true for non-directory paths

Thayorns:
The `directory` object decorates a `file`, but it also declared `true > is-directory`,
which shadowed the check it inherits. Any `directory` therefore claimed to be a
directory even when its path pointed at a regular file or at nothing at all. The
`is-directory` in `file` does the real work: it stats the path and reads the mode
bits, falling back to `cant-check` when the syscall fails.

Dropping that one line lets the decorator delegate, so the answer now comes from
the file system instead of from a constant. Every place that already calls
`is-directory` without an argument sits behind an `exists` check, so the fallback
is never dataized there and those call sites keep working. Two existing tests
gain meaning from this: the assertion in `directory.made` and the one in `mktemp`
were both satisfied by the constant before and now depend on a syscall.

The new tests cover a real directory, a regular file, a directory that was made
and then deleted, and an absent path both with and without the fallback.

I could not run the suite locally to completion, so I am leaning on CI here and
will follow up on whatever it reports.

Closes #5440

### DIFF
--- a/eo-runtime/src/main/eo/directory.eo
+++ b/eo-runtime/src/main/eo/directory.eo
@@ -13,7 +13,6 @@
 
 [file] > directory
   file > @
-  true > is-directory
   [dir time process random] > retried
     [index] >> attempted
       if. > @
@@ -187,6 +186,23 @@
     mktemp.tmpfile.deleted.as-path.as-dir.made > parent
     created.as-bytes.eq expected.as-bytes > valid!
 
+  ++> can-tell-that-a-deleted-directory-is-no-longer-a-directory
+    not. > @
+      d.deleted.is-directory false
+    made. > d
+      as-dir.
+        mktemp.tmpfile.deleted.as-path.resolved "gone"
+
+  is-directory. ++> can-tell-that-a-directory-is-a-directory
+    directory
+      Q.file "."
+    false
+
+  not. ++> can-tell-that-a-regular-file-is-not-a-directory
+    is-directory.
+      directory mktemp.tmpfile
+      true
+
   ++> can-walk-recursively
     seq * > @
       made.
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 this drops the constant so `is-directory` comes from the stat, as you decided in the thread. Happy to clarify anything or split the tests out if you would rather see them separately.

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+894](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6383, opened 2026-08-04
fix(#5466): print an lcov tracefile of eo object coverage

### LINKED ISSUE #5466: Implement parameter 'coverage.file' for EO object coverage measurement in transpile goal

Thayorns:
The `coverageTracking` flag already makes the transpiler wrap every located object into `PhCoverage`, and `PhCoverage` already appends one record per touched object at runtime. Nothing read that file, so there was no coverage report to look at. This adds the `lcov` goal (`MjLcov`), which runs after the tests, reads that file and saves an LCOV tracefile: one `SF:` per `.eo` program, one `DA:` per line carrying how many objects of that line the run touched, plus `LF`/`LH`. LCOV because Codecov, Coveralls and `genhtml` read it as is.

For the report to name a file, a record has to name one. `PhCoverage` was given the object's `@loc` (`Φ.org.eolang.number.plus`), from which the `.eo` file cannot be recovered, so `to-java.xsl` now passes `$object-name` instead — the same value it already hands to `PhSafe` two lines above. Records are now `program:line:pos`, the field is named `program` to match, and `Lcov` resolves each program against the directory of `.eo` sources, so an `SF:` path is one a coverage tool can open.

This is the working top of the feature, not all of it. The tracefile currently lists only the lines the tests reached, so every line it names is covered and `LH` always equals `LF`. Reporting the objects the run never touched needs the set of locations the transpiler instrumented, which is readable back out of the generated Java where each one is a `new PhCoverage(..)` wrapper; that is a puzzle on `MjLcov`, next to the one about binding this goal in `eo-runtime/pom.xml`, and the existing puzzle about failing the build under a threshold stays on `MjTranspile`.

Closes #5466


eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
```diff
@@ -85,15 +85,14 @@ public final class MjTranspile extends MjSafe {
...
-     * @todo #5466:60min Turn raw coverage hits into an LCOV report.
-     *  Right now the runtime only produces a raw, append-only
-     *  {@code loc:line:pos} file: the {@code PhCoverage} decorator
-     *  writes every touched location into it, but nothing consumes that
-     *  file yet. Add a reporter step that merges those raw hits against
-     *  the full set of instrumented locations, which the transpiler
-     *  already knows because it emits every wrapper, and produces an
-     *  LCOV ({@code .info}) tracefile plus the covered percentage. LCOV
-     *  is chosen because Codecov and Coveralls consume it directly.
+     * @todo #5466:60min Feed the coverage hits to {@link Lcov}.
```
yegor256: @Thayorns read this: https://www.yegor256.com/2015/02/12/top-down-design.html

First, implement the feature that works for the user. Then, put puzzles for the details of implementation. You did it the opposite way.

eo-maven-plugin/src/main/java/org/eolang/maven/Lcov.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    private final Collection<String> recorded;
+
+    /**
+     * Ctor.
+     * @param instrumented Every location the transpiler instrumented
+     * @param hits Every location the run touched
+     */
+    Lcov(final Collection<String> instrumented, final Collection<String> hits) {
+        this.located = instrumented;
```
morphqdd: No defensive copy here. If the caller mutates the collection passed to this constructor, toString() output changes after construction.

eo-maven-plugin/src/main/java/org/eolang/maven/Lcov.java
```diff
@@ -0,0 +1,79 @@
...
+     * Every location the run touched, as {@code program:line:pos}.
+     */
+    private final Collection<String> recorded;
+
+    /**
+     * Ctor.
+     * @param instrumented Every location the transpiler instrumented
+     * @param hits Every location the run touched
+     */
+    Lcov(final Collection<String> instrumented, final Collection<String> hits) {
```
morphqdd: Constructor parameter names (instrumented, hits) do not match the field names they get assigned to (located, recorded). Makes the two harder to follow together.

eo-maven-plugin/src/main/java/org/eolang/maven/Lcov.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    private Map<String, Map<Integer, Long>> counted() {
+        final Map<String, Long> hits = this.recorded.stream()
+            .collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
+        final Map<String, Map<Integer, Long>> counts = new TreeMap<>();
+        for (final String record : this.located) {
+            final String[] parts = record.split(":");
+            counts
+                .computeIfAbsent(parts[0], program -> new TreeMap<>())
+                .merge(Integer.valueOf(parts[1]), hits.getOrDefault(record, 0L), Long::sum);
```
morphqdd: located is a plain Collection, not a Set. If the same program:line:pos record ever appears twice in it, this merge call adds its hit count twice, inflating the DA: value for that line.

eo-maven-plugin/src/main/java/org/eolang/maven/Lcov.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * How many times each instrumented line of each program was touched.
+     * @return Programs in alphabetical order, each with its lines in order
+     */
+    private Map<String, Map<Integer, Long>> counted() {
+        final Map<String, Long> hits = this.recorded.stream()
+            .collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
+        final Map<String, Map<Integer, Long>> counts = new TreeMap<>();
+        for (final String record : this.located) {
+            final String[] parts = record.split(":");
```
morphqdd: record.split(':') is used with no check on the resulting array length or shape. A malformed record throws a raw ArrayIndexOutOfBoundsException here instead of a clear error.

eo-maven-plugin/src/main/java/org/eolang/maven/Lcov.java
```diff
@@ -0,0 +1,79 @@
...
+        this.located = instrumented;
+        this.recorded = hits;
+    }
+
+    @Override
+    public String toString() {
+        return this.counted().entrySet().stream().map(
+            program -> String.format(
+                "TN:%nSF:%s.eo%n%sLF:%d%nLH:%d%nend_of_record%n",
+                program.getKey().replace('.', '/'),
```
morphqdd: SF: is written as the program name with dots turned into slashes, plus .eo, with no documented base directory. Nothing here says this path is relative to src/main/eo, so a genhtml or Codecov run from the repo root would not find the file without knowing to add that prefix.

eo-maven-plugin/src/main/java/org/eolang/maven/Lcov.java
```diff
@@ -0,0 +1,79 @@
...
+ *     collections hold {@code program:line:pos} records, the shape
+ *     {@code PhCoverage} appends, so a program turns into an {@code SF:}
+ *     path and a line into a {@code DA:} counter of every hit recorded
+ *     for the objects of that line. A record the run touched but the
+ *     transpiler never instrumented is left out.
+ * </p>
+ * @since 0.58
+ */
+final class Lcov {
+
```
morphqdd: This class is not called from anywhere in the codebase outside LcovTest (checked with a repo-wide search for Lcov). Right now it is unreachable production code: nothing exercises it except its own unit test, and nothing would break if it were deleted.

eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
morphqdd: This puzzle says to collect instrumented locations during transpile and feed them to a goal that runs after tests, but transpile and that later goal are separate Maven executions, possibly separate JVMs. The puzzle does not say how the collected locations are supposed to survive between the two, only that to-java.xsl knows them while emitting wrappers.

eo-runtime/src/main/java/org/eolang/PhCoverage.java
```diff
@@ -21,7 +21,7 @@
  *
  * <p>The transpiler emits it around every located object. Recording is
  * enabled by the {@code eo.coverageFile} system property: on the first
- * touch, one {@code loc:line:pos} line is appended, at most once per
+ * touch, one {@code program:line:pos} line is appended, at most once per
```
morphqdd: A related sentence a few lines below this one (the class-level paragraph pointing at MjTranspile's javadoc for 'the remaining gaps') still lists the LCOV report as missing, but Lcov.java now exists in this same PR. Worth updating that sentence too, to say it exists but is not wired in yet, matching the MjTranspile puzzle wording that was updated here.

eo-maven-plugin/src/test/java/org/eolang/maven/LcovTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link Lcov}.
+ * @since 0.58
+ */
+final class LcovTest {
+
+    @Test
```
morphqdd: No test with an empty located collection. Lcov's toString() should presumably produce an empty string when nothing was instrumented, but that is not verified anywhere in this file.

eo-maven-plugin/src/test/java/org/eolang/maven/LcovTest.java
```diff
@@ -0,0 +1,71 @@
...
+                Collections.emptyList()
+            ).toString(),
+            Matchers.equalTo(
+                String.format(
+                    "TN:%nSF:org/eolang/числò.eo%nDA:%d,0%nLF:1%nLH:0%nend_of_record%n",
+                    line
+                )
+            )
+        );
+    }
```
morphqdd: The class javadoc says a location the run touched but the transpiler never instrumented is left out. No test here exercises that: every recorded entry in every test is also present in located.

eo-maven-plugin/src/test/java/org/eolang/maven/LcovTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Test
+    void keepsProgramsApartInAlphabeticalOrder() {
+        MatcherAssert.assertThat(
+            "two programs are not reported as two separate tracefile records, sorted by name",
+            new Lcov(
+                Arrays.asList("org.eolang.txt.sprintf:3:1", "org.eolang.bytes:7:2"),
+                Collections.singleton("org.eolang.bytes:7:2")
+            ).toString(),
+            Matchers.stringContainsInOrder(
```
morphqdd: This only checks that the two SF: lines appear in the right order, not the full DA:/LF:/LH: content of either program block. A bug that fixed ordering but corrupted one program's counts would still pass here.

eo-maven-plugin/src/test/java/org/eolang/maven/MjTranspileTest.java
```diff
@@ -122,7 +122,7 @@ void wrapsObjectsIntoPhCoverageWhenTrackingEnabled(@Mktmp final Path temp) throw
                     .result()
                     .get(MjTranspileTest.compiled())
             ).asString(),
-            Matchers.containsString("new PhCoverage(")
+            Matchers.stringContainsInOrder("new PhCoverage(", "\"foo.x.main\", ")
```
morphqdd: Same pattern as the stringContainsInOrder assertions in LcovTest: only substring containment is checked, not the exact generated line, so extra or missing whitespace or wrong quoting of the object name would not fail this test.

morphqdd (changes requested):

yegor256:
@Thayorns conflicts here

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2777](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +12 points.  Your running score is [+3777](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

eo-maven-plugin/src/main/java/org/eolang/maven/Lcov.java
```diff
@@ -0,0 +1,82 @@
...
+        this.hits = new SetOf<>(hits);
+    }
+
+    @Override
+    public String toString() {
+        return this.counted().entrySet().stream().map(
+            program -> String.format(
+                "TN:%nSF:%s%n%sLF:%d%nLH:%d%nend_of_record%n",
+                this.sources.resolve(
+                    String.format("%s.eo", program.getKey().replace('.', '/'))
```
Favixx: This is the line the report stands or falls on, and for `eo-runtime` it falls. `program.getKey()` is `$object-name`, which `to-java.xsl` builds as `@package` + `.` + `@name` — so `org.eolang.number` — and this turns it into `org/eolang/number.eo`. The actual source is `eo-runtime/src/main/eo/number.eo`. I listed the tree: 182 `.eo` files under that directory, none of them under `org/`.

The package is a declaration inside the file (`+package org.eolang`), not a statement about where the file lives, and EO does not require the two to agree. `FakeMaven` happens to make them agree, which is why every test here passes.

What does know the answer is the XMIR, which carries the source path from `MjParse`. Emitting that into the `PhCoverage` wrapper — as a fourth string, or in place of `$object-name` — makes `SF:` a fact rather than a reconstruction, and it also removes the `replace('.', '/')` guess about separators. If that is too large for this PR, then the goal should at least check that the resolved path exists and say so loudly, because a tracefile full of unopenable paths fails silently in every consumer.

eo-maven-plugin/src/test/java/org/eolang/maven/LcovTest.java
```diff
@@ -0,0 +1,80 @@
...
+            new Lcov(
+                Paths.get("src/main/eo"),
+                Arrays.asList(
+                    "org.eolang.number:12:5", "org.eolang.number:12:9", "org.eolang.number:12:5"
+                )
+            ).toString(),
+            Matchers.equalTo(
+                String.format(
+                    "TN:%nSF:%s%nDA:12,2%nLF:1%nLH:1%nend_of_record%n",
+                    Paths.get("src/main/eo").resolve("org/eolang/number.eo")
```
Favixx: This test picks a real object — `org.eolang.number` is `eo-runtime/src/main/eo/number.eo` — and asserts that its source is `src/main/eo/org/eolang/number.eo`. That path does not exist in this repository, so the assertion states the opposite of the truth and will keep doing so after the bug is fixed.

When the mapping is corrected, this is the test that has to change first. It would be worth making it one that could have caught the problem: build a temporary directory with an actual `.eo` file in it, hand that directory to `Lcov`, and assert that the `SF:` path is a file that exists. That assertion is true for `foo/x/main.eo` and false for `org/eolang/number.eo`, which is exactly the distinction the current tests cannot make.

eo-maven-plugin/src/test/java/org/eolang/maven/LcovTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Test case for {@link Lcov}.
+ * @since 0.58
+ */
+final class LcovTest {
+
+    @Test
+    void namesEoSourceOfTouchedProgram() {
+        final long seed = System.nanoTime();
```
Favixx: Randomising the line number does not test anything the number 7 would not: `Lcov` reads it with `Integer.valueOf` and prints it back, and nothing about the value 863 exercises a path that 7 does not. What it costs is a test that reads differently on every run and a `String.format` in the assertion that has to carry the seed.

If the intent was to make sure the line is not accidentally hard-coded somewhere in `Lcov`, two fixed and clearly different lines say that more directly. The interesting inputs here are elsewhere anyway — line `0`, a line number with more digits than the file has lines, the same line arriving from two different programs.

eo-maven-plugin/src/main/java/org/eolang/maven/Lcov.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * How many objects of each line of each program the run touched.
+     * @return Programs in alphabetical order, each with its lines in order
+     */
+    private Map<String, Map<Integer, Integer>> counted() {
+        final Map<String, Map<Integer, Integer>> counts = new TreeMap<>();
+        for (final String record : this.hits) {
+            final String[] parts = record.split(":");
```
Favixx: One malformed line and the goal throws, taking the build with it — after the tests have already passed. That trade is wrong for a coverage report: the report is a by-product, and losing it should never be worse than losing the tests.

It is also more likely than it looks. `PhCoverage.hit()` appends to this file from whichever JVM is running the tests, and surefire in `eo-runtime` forks; `Files.write` with `APPEND` gives no guarantee that two writes from two processes do not interleave, so a torn line — `org.eolang.foo:7org.eolang.bar:9:5` — is a normal outcome under `-T` or `forkCount > 1`, not corruption. Skipping the line with a `Logger.warn` naming it keeps the other ninety-nine per cent of the report and tells whoever cares that it happened.

`split(":")` is worth a limit too: `split(":", 3)` would at least keep a program name that somehow contains a colon in one piece rather than turning it into a hard error.

eo-maven-plugin/src/main/java/org/eolang/maven/Lcov.java
```diff
@@ -0,0 +1,82 @@
...
+    Lcov(final Path sources, final Iterable<String> hits) {
+        this.sources = sources;
+        this.hits = new SetOf<>(hits);
+    }
+
+    @Override
+    public String toString() {
+        return this.counted().entrySet().stream().map(
+            program -> String.format(
+                "TN:%nSF:%s%n%sLF:%d%nLH:%d%nend_of_record%n",
```
Favixx: Two things about this format string.

`%n` is the platform separator, so on the Windows leg of CI this writes a CRLF tracefile. `genhtml` and Codecov tolerate it, but LCOV files are conventionally LF and a tracefile that differs by platform is one more thing that is not reproducible; `\n` is the safer literal here and in `PhCoverage.hit()`.

`this.sources.resolve(...)` yields an absolute path, since `sourcesDir` is absolute in a real build — so `SF:` comes out as `/home/runner/work/eo/eo/eo-runtime/src/main/eo/…`. Codecov matches coverage to files by repository-relative path and quietly drops records it cannot place; `genhtml` is happier with absolute paths but produces a report rooted in a directory nobody else has. Relativising against the project basedir before printing is the usual answer, and it is what every other coverage tool in this build emits.

eo-maven-plugin/src/main/java/org/eolang/maven/Lcov.java
```diff
@@ -0,0 +1,82 @@
...
+                this.sources.resolve(
+                    String.format("%s.eo", program.getKey().replace('.', '/'))
+                ),
+                program.getValue().entrySet().stream()
+                    .map(line -> String.format("DA:%d,%d%n", line.getKey(), line.getValue()))
+                    .collect(Collectors.joining()),
+                program.getValue().size(),
```
Favixx: `LF` and `LH` are the same expression, which the puzzle on `MjLcov` explains and I have no quarrel with as a first step. Worth being explicit in the class javadoc about what that means for a reader of the report, though: every consumer of this file will display `100%`, and a coverage badge built from it today would be actively misleading rather than merely incomplete.

The other definition worth writing down is `DA`. In LCOV, the second field is the number of times the line was executed; here it is the number of distinct objects on that line that were touched at least once. Both are integers and no tool will complain, but `DA:12,2` meaning "two objects" rather than "two executions" changes what the heat colouring in `genhtml` shows. The class doc says "how many objects of that line were touched" — that sentence should also say that this deliberately reuses the execution-count field.

eo-maven-plugin/src/main/java/org/eolang/maven/MjLcov.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Ctor.
+     */
+    public MjLcov() {
+        // nothing
+    }
+
+    @Override
+    public void exec() throws IOException {
```
Favixx: `public void exec()` here, while the other mojos in this plugin override `MjSafe.exec()` with package-private visibility — `MjInference` in the open #6350 does `@Override void exec()`, and so do the ones already on master. Nothing needs this to be public: Maven calls `execute()` on `MjSafe`, not this.

Widening it makes `MjLcov` the only mojo whose template method is part of its public API, which qulice will not complain about but a reader will.

eo-maven-plugin/src/main/java/org/eolang/maven/MjLcov.java
```diff
@@ -0,0 +1,108 @@
...
+        Logger.info(this, "EO object coverage saved to %[file]s", target);
+    }
+
+    /**
+     * Every location the run touched.
+     * @return The records, or nothing when no test recorded any
+     * @throws IOException If fails to read them
+     */
+    private Collection<String> recorded() throws IOException {
+        final Path path = this.hits.toPath();
```
Favixx, line 101: Nothing ever truncates this file. `PhCoverage` opens it with `APPEND`, `SEEN` only dedupes within one JVM, and this goal only reads — so a second `mvn verify` without `clean` produces a tracefile containing every object touched by both runs. `Lcov` dedupes identical records, so the counts stay right, but an object deleted from the sources between the two runs is still reported as covered, and there is no way for a reader to tell.

The cheap fix is here: delete the file after reading it, so each run reports its own tests. The correct one is at the other end — truncate before the tests start — but that belongs to whichever execution sets `eo.coverageFile`, which is the puzzle below about binding this in `eo-runtime/pom.xml`. Whichever you choose, it is worth a sentence in that puzzle, because a report that silently accumulates is a worse failure than no report.

eo-maven-plugin/src/main/java/org/eolang/maven/MjLcov.java
```diff
@@ -0,0 +1,108 @@
...
+     *  {@code new PhCoverage(..)} wrapper naming its program, line and
+     *  position, which is the same set {@code to-java.xsl} decided on
+     *  while emitting them and survives on disk between this goal and
+     *  {@code transpile}, two separate Maven executions.
+     */
+    @Parameter(
+        property = "eo.coverageFile",
+        defaultValue = "${project.build.directory}/coverage.txt"
+    )
+    private File hits;
```
Favixx, line 52: `eo.coverageFile` now names two different things that have to agree and are set by two different mechanisms: this `@Parameter`, read by Maven from `-D` or the pom, and the system property of the forked test JVM, which surefire sets from its own `<systemPropertyVariables>`. A user who passes `-Deo.coverageFile=/tmp/x` on the command line moves the mojo's idea of the file, and moves the tests' idea only if surefire happens to forward it.

That coupling is the whole feature and it is invisible from either side. At minimum this javadoc should say that the value must match what surefire gives the tests, and name where that is configured. Better, when the file is missing this goal could say so — right now `recorded()` silently returns nothing and the user gets an empty tracefile with no hint that the two halves disagreed.

The field name is worth a look too: `hits` against a property called `eo.coverageFile`, and `FakeMaven` in `MjLcovTest` has to say `.with("hits", …)` because of it.

eo-maven-plugin/src/main/java/org/eolang/maven/MjLcov.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Save an LCOV tracefile of EO object coverage.
+ * <p>
+ *     This Maven Mojo reads the hits {@code PhCoverage} appended while the
+ *     tests ran and saves them as LCOV. It runs after the tests, so it is
+ *     bound to {@code verify}, and it needs {@code coverageTracking} on
+ *     the {@code transpile} goal, since otherwise nothing is instrumented
+ *     and no hit is ever recorded.
+ * </p>
+ * @since 0.58
```
Favixx: `@since 0.58` on a class that does not exist before this PR — and the same on `Lcov` and `LcovTest`. The last release is 0.62.1, so these are all `0.63`.

I take it the number was copied from `PhCoverage`, which genuinely is `@since 0.58` since it landed then. That one is right and should stay; these three should say when they arrive.

eo-maven-plugin/src/main/java/org/eolang/maven/MjLcov.java
```diff
@@ -0,0 +1,108 @@
...
+     * Ctor.
+     */
+    public MjLcov() {
+        // nothing
+    }
+
+    @Override
+    public void exec() throws IOException {
+        final Path target = this.tracefile.toPath();
+        Files.createDirectories(target.getParent());
```
Favixx: `getParent()` returns null for a path with no directory in it, so `-Deo.lcovFile=eo-lcov.info` — a perfectly ordinary thing to type — gives a `NullPointerException` out of `Files.createDirectories` rather than a file in the working directory.

The default always has a parent, so this only bites someone overriding the parameter, which is exactly the person least likely to expect it.

eo-maven-plugin/src/main/java/org/eolang/maven/MjLcov.java
```diff
@@ -0,0 +1,108 @@
...
+    defaultPhase = LifecyclePhase.VERIFY,
+    threadSafe = true
+)
+public final class MjLcov extends MjSafe {
+
+    /**
+     * The file where {@code PhCoverage} appended one record per touched
+     * object, named by the {@code eo.coverageFile} system property of the
+     * JVM that ran the tests.
+     * @todo #5466:60min Report the objects the run never touched.
```
Favixx: The plan in this puzzle — recover the instrumented set by reading `new PhCoverage(..)` calls back out of the generated Java — worries me more than the missing feature does. It makes the report depend on the exact text `to-java.xsl` emits, so a change to the wrapper's argument order, or a formatter run over the generated sources, silently reduces coverage rather than breaking anything.

The transpiler already has the set in a structured form: `to-java.xsl` decides it from `@line`, `@pos` and `$coverage`, and could write it out as a table beside the generated Java in the same pass — which is also where the source path from the previous comment would come from. That is a smaller change than a parser for Java source, and it survives reformatting.

Worth reconsidering while the puzzle is still only a puzzle.

eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
```diff
@@ -655,7 +655,7 @@
       <xsl:text> = new PhCoverage(</xsl:text>
       <xsl:value-of select="$name"/>
       <xsl:text>, "</xsl:text>
-      <xsl:value-of select="@loc"/>
+      <xsl:value-of select="$object-name"/>
```
Favixx: Right change for the stated reason, with one consequence worth writing down: `@loc` identified the object, and `$object-name` identifies only its program, so the identity of a record is now the triple `(program, line, pos)` rather than a single name.

That triple is unique as long as no two instrumented objects share a line and a column, which is true for the parser's `@line`/`@pos` but is not stated anywhere and is exactly what the puzzle on `MjLcov` will depend on when it matches instrumented locations against hits. `PhSafe` two lines above keeps `@loc` as a separate argument, so the information is not lost from the generated code — only from the coverage record. A comment here saying which of the two carries identity now would save the next reader the diff.

eo-runtime/src/main/java/org/eolang/PhCoverage.java
```diff
@@ -134,7 +135,7 @@ public byte[] delta() {
     private void hit() {
         final String property = PhCoverage.property();
         if (property != null) {
-            final String record = String.format("%s:%d:%d%n", this.loc, this.line, this.pos);
+            final String record = String.format("%s:%d:%d%n", this.program, this.line, this.pos);
```
Favixx: This is the writer whose output `Lcov.counted()` refuses to tolerate. Two notes.

`%n` again — the record is written with the platform separator and read back with `Files.readAllLines`, which handles both, so nothing breaks today; it only means the raw file differs by platform for no reason.

More substantially, this line and the `Files.write` below are the only thing standing between forked test JVMs and a torn record. `SEEN` makes each JVM write a given record once, but two JVMs still write concurrently, and a single append of a short string is atomic on Linux only up to `PIPE_BUF` and not promised at all on Windows. Since `Lcov` turns a torn line into a build failure, it is worth either making the writer robust — one file per JVM, named after the PID, merged by the goal — or making the reader forgiving. Doing neither leaves a rare, load-dependent build break.

eo-runtime/src/main/java/org/eolang/PhCoverage.java
```diff
@@ -45,8 +46,8 @@ public final class PhCoverage implements Phi {
     /** The origin. */
     private final Phi origin;
 
-    /** The fully qualified location of the object. */
-    private final String loc;
+    /** The name of the EO program that holds the object, dot separated. */
```
Favixx: Good rename, and the comment is accurate about what the value is. What it does not say is what the field is *for*, which changed with it: `loc` was the identity of the object, `program` is not — it is a key into the report, and identity now lives in the `(program, line, pos)` triple assembled in `hit()`.

A reader of this class alone cannot tell that, and the `equals`/`hashCode` below delegate to the origin, so nothing in the class hints at what makes two records different. One clause here — "together with {@link #line} and {@link #pos} it identifies the object" — closes that.

eo-maven-plugin/src/test/java/org/eolang/maven/MjTranspileTest.java
```diff
@@ -122,7 +122,9 @@ void wrapsObjectsIntoPhCoverageWhenTrackingEnabled(@Mktmp final Path temp) throw
                     .result()
                     .get(MjTranspileTest.compiled())
             ).asString(),
-            Matchers.containsString("new PhCoverage(")
+            Matchers.matchesPattern(
+                "(?s).*\\w++ = new PhCoverage\\(\\w++, \"foo\\.x\\.main\", \\d++, \\d++\\);.*"
```
Favixx: Stronger than `containsString("new PhCoverage(")` and I am glad it now checks the name, but note what it is pinning: `foo.x.main` is `FakeMaven`'s program, written to `foo/x/main.eo` with `+package foo.x`. Package and directory agree there, which is the assumption `Lcov` makes and `eo-runtime` breaks.

So this assertion is correct and will stay correct, and it is also the reason nothing in this PR notices the path problem. If `to-java.xsl` ends up emitting the source path instead of the dotted name, this regex is the first place that has to change, and a second case here with a program whose package does not match its directory would be the cheapest guard against the whole class of bug.

eo-maven-plugin/src/test/java/org/eolang/maven/MjLcovTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Test
+    void savesRecordedHitAsTracefile(@Mktmp final Path temp) throws Exception {
+        final Path lcov = temp.resolve("eo-lcov.info");
+        final Path hits = temp.resolve("coverage.txt");
+        Files.write(
+            hits, String.format("org.eolang.числò:7:2%n").getBytes(StandardCharsets.UTF_8)
+        );
+        new FakeMaven(temp)
+            .with("hits", hits.toFile())
```
Favixx, line 35: `.with("hits", …)` and `.with("tracefile", …)` set mojo fields by name, so these two strings are the field names of `MjLcov` duplicated into the test with no compiler between them. Rename either field and these tests keep compiling and start silently testing the defaults.

Setting them per test rather than in `FakeMaven`'s shared defaults is the right call and better than what #6350 does with the same problem. The remaining exposure is just the two literals; naming the fields after their properties — `coverageFile` and `lcovFile` — would at least make the strings guessable from the parameter a user actually types.

Also missing here: the case where the hits file exists but holds a line `Lcov` rejects. `LcovTest.refusesMalformedRecord` covers the class, but nothing covers what the *goal* does with it, which is where it matters — that is the path that turns a bad byte in a scratch file into a failed build.

eo-maven-plugin/src/test/java/org/eolang/maven/LcovTest.java
```diff
@@ -0,0 +1,80 @@
...
+    void printsNothingWhenNothingWasTouched() {
+        MatcherAssert.assertThat(
+            "a run that touched nothing is not reported as an empty tracefile",
+            new Lcov(Paths.get("src/main/eo"), Collections.emptyList()).toString(),
+            Matchers.equalTo("")
+        );
+    }
+
+    @Test
+    void refusesMalformedRecord() {
```
Favixx: This test fixes the hard failure in place, which is why I would settle the question before it merges rather than after: once `IllegalArgumentException` is the asserted contract, softening it to a warning becomes a change with a test to argue with.

If you keep the failure, the message is good but the exception is wrong — `IllegalArgumentException` describes an argument the caller passed, while this is a malformed line in a file written by another process an hour earlier. Something in the `ExFailure`/`IllegalStateException` family reads truer, and would let a caller that wants to survive the situation catch it without catching everything else.

Favixx (changes requested):
The diagnosis in the description is right and it is the good part of this PR: `PhCoverage` was writing `@loc`, from which no file can be recovered, so the raw file could never become a report no matter what read it. Passing `$object-name` instead — the value `PhSafe` already gets two lines above — is the correct fix, and stopping at "the working top of the feature" with the untouched objects left as a puzzle is an honest place to stop.

But the mapping from a program name to an `.eo` path does not hold for the module this feature exists for.

`Lcov` turns `org.eolang.number` into `<sources>/org/eolang/number.eo`. In `eo-runtime` that file is `<sources>/number.eo`: the package comes from `+package org.eolang` in the source text, while the directory layout is flat. There are 182 `.eo` files under `eo-runtime/src/main/eo` and not one of them sits in an `org/` directory. So every `SF:` line the goal writes for `eo-runtime` names a path that does not exist, and `genhtml` — which the description names as a consumer — refuses a tracefile whose source it cannot open.

The tests do not catch it because `FakeMaven` writes its program to `foo/x/main.eo` with `+package foo.x`, so package and directory agree there. `LcovTest.countsEveryTouchedPositionOfOneLine` then takes a real `eo-runtime` object, `org.eolang.number`, and asserts `src/main/eo/org/eolang/number.eo` as the expected answer — encoding the wrong mapping as the specification.

The fix has to come from somewhere that knows the file, not the package. The transpiler has it: the XMIR carries the source path that `MjParse` recorded, so `to-java.xsl` could emit that instead of, or alongside, `$object-name`. Anything derived from the dotted name is guessing.

Secondary, in rough order: a single malformed line aborts the whole goal, which is a real risk when the file is appended to by forked test JVMs; the tracefile is never truncated, so a second `mvn verify` without `clean` reports objects that no longer exist; `SF:` paths are absolute where Codecov wants them repo-relative; and `@since 0.58` on three brand-new classes when the last release was 0.62.1. Details inline.

yegor256:
@Thayorns put `lcov` target to the `coverage.yml` CI job - to make sure it works in production with our data

Thayorns:
@yegor256 done, the `lcov` goal now runs in the `codecov` job on our own data.

Three things had to line up for that. Surefire in `eo-runtime` already named `eo.coverageFile`, but the pom pinned `<coverageTracking>false</coverageTracking>` literally, so the `-Deo.coverageTracking=true` the workflow already passed never reached the transpiler and nothing was ever instrumented there; it is `${eo.coverageTracking}` now, off by default and on in that job. Then an `lcov` execution bound to `verify` next to the `compile` one, reading the same file surefire writes, so `mvn install` produces `eo-runtime/target/eo-lcov.info`. Then the tracefile goes to Codecov alongside the three jacoco reports, with a step before the upload that fails the job if the file is empty or if any `SF:` path it names is not a file on disk — that is the part that proves it works with our data rather than merely running.

The puzzle about binding the goal is gone, since that is what this does. Two remain: the untouched objects, which still keep `LH` equal to `LF`, and the coverage threshold. I also rewrote the first one — recovering the instrumented set by reading `new PhCoverage(..)` back out of the generated Java is fragile, `to-java.xsl` should write that table out in the same pass it emits the wrappers.

One caveat worth naming: `codecov.yml` triggers on pushes to master only, so the new check first runs after this is merged, not on this pull request. Say the word and I will add a `pull_request` trigger so it is proven before the merge instead.

Thayorns:
The red checks here are not this branch: `eo-parser/src/main/java/org/eolang/parser/Recovery.java:76` calls `this.rewound(idx, failed)` inside `skip(int from, int indent)`, where `failed` is not in scope, so `eo-parser` does not compile and every job that builds the project stops there. It came in with #6467 and master's own `mvn` run for that merge is red too. It is filed as #6485, so I am leaving it alone rather than working around it, and I will merge master again once it is fixed.

One check was mine and is fixed in 4e859dcf69: I had written the record check as a `private static final Pattern`, which took `counts (smells)` from 847 to 848. It is a plain `split` now.

yegor256:

Thayorns:
@yegor256 ping

yegor256:
@Thayorns conflicts here

Thayorns:
@yegor256 conflict solved

Thayorns:
@yegor256 CI green here

Thayorns:
@yegor256 ping

Thayorns:

Thayorns:
@yegor256 fixed here

eo-maven-plugin/src/main/java/org/eolang/maven/MjLcov.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    @Parameter(
+        property = "eo.coverageFile",
+        defaultValue = "${project.build.directory}/coverage.txt"
+    )
+    private File hits;
+
+    @Override
+    void exec() throws IOException {
+        final Path saved = this.targetDir.toPath().resolve("eo-lcov.info");
```
yegor256: @Thayorns maybe we should make this name configurable via Maven plugin param?

eo-maven-plugin/src/main/java/org/eolang/maven/MjLcov.java
```diff
@@ -0,0 +1,112 @@
...
+        Files.createDirectories(saved.getParent());
+        Files.write(saved, this.lcov().getBytes(StandardCharsets.UTF_8));
+        Logger.info(this, "EO object coverage saved to %[file]s", saved);
```
yegor256: @Thayorns would be great to know the size of the time just saved, in the same log line

eo-runtime/pom.xml
```diff
@@ -256,10 +256,16 @@
           <cacheEnabled>false</cacheEnabled>
           <!-- Tests assert on precise .eo locations in panics (off by default). -->
           <trackLocations>true</trackLocations>
-          <!-- Wrap located objects into PhCoverage so tests record EO coverage. -->
-          <coverageTracking>false</coverageTracking>
```
yegor256: @Thayorns this will make all builds extremely slow. Better update the `codecov.yml` CI job -- it has to collect coverage info, nobody else

eo-maven-plugin/src/main/java/org/eolang/maven/MjLcov.java
Thayorns: Done in 4129365853. The path is now the `tracefile` parameter, `eo.lcovFile`, defaulting to `${project.build.directory}/eo/eo-lcov.info` — the same place it was written before, so `codecov.yml` keeps pointing at it.

eo-maven-plugin/src/main/java/org/eolang/maven/MjLcov.java
Thayorns: Done in 4129365853, the line now reads `EO object coverage saved to ... (12Kb)` through the `%[size]s` decor, the same way `MjPrint` reports what it wrote.

eo-runtime/pom.xml
Thayorns: That is what this does, and the deleted line is why it works. `coverageTracking` defaults to false in `MjTranspile`, so removing the value leaves it off for every build; while it was pinned to `false` in the plugin configuration here, that value won over `-D` and the CI job could not have switched it on at all. Only `codecov.yml` turns it on, and it already does: `mvn install -Pjacoco -Deo.coverageTracking=true`. I left a comment at that spot in 4129365853 saying so, since the absence of a line is easy to misread as a default flipping.

Thayorns:
@yegor256 conflict solved

yegor256:
@Thayorns it seems that this feature is already in master: https://github.com/objectionary/eo/pull/6911

Issue #6239, opened 2026-07-31
`string.repeated` function causes stack overflow and exhibits sub-optimal performance with large repeat counts


gemshrine:
## What happens

`string.repeated` overflows the stack for a repeat count of about a thousand, and is already pathologically slow well below that. Repeating a short string a few hundred times is an ordinary operation (padding, separators, building test data), but `repeated` cannot do it.

### Runtime evidence

Probed `org.eolang.EO_string.EOrepeated` with `text = "x"`:

| call | result |
| --- | --- |
| `repeated "x" 500` | 500-char string, but takes ~10 s |
| `repeated "x" 1000` | **StackOverflowError** (surfaced as `ExFailure` at `string.repeated:31`) |
| `repeated "x" 2000` | **StackOverflowError** |

The failure is deterministic (driven by recursion depth) and the slowness grows super-linearly with the count.

## Root cause

`eo-runtime/src/main/eo/string/repeated.eo` builds the result with a recursion that peels **one** copy per level:

```
if. > [accum index] >> rec-repeated
  amount.eq index
  accum
  rec-repeated
    accum.concat b
    as-number.
      index.plus 1
| text >> b! 1
```

`rec-repeated` calls itself `amount` times, so the recursion depth equals the repeat count and overflows the stack once the count reaches roughly a thousand. On top of that each level does `accum.concat b`, copying the whole growing accumulator, so the total work is `O(amount^2)` - that is the ~10 s for only 500 copies.

## Suggested fix

Build the result by doubling instead of by single steps, which makes the depth logarithmic in the count and the work near-linear:

```
# repeat text n times, halving the count each level
[n] >> rec
  if. > @
    n.eq 0
    ""
    concat.
      half.concat half
      if. (n.mod 2).eq 0 "" text
  rec (n.div 2) >> half!
```

(compute `half` once as a constant so the recursion stays `O(log n)` in depth and does not recompute the smaller repetition). With doubling, `repeated "x" 1000` returns immediately and much larger counts stay within the stack.

A regression test such as `eq. ++> ... (repeated "x" 5000).length 5000` would pin both the depth and the correctness.

0crat:
[...]

PR #6385, opened 2026-08-04
Shrink the directory walk atom to a single-level listing

### LINKED ISSUE #5751: `directory.walk` atom oversteps by handling glob matching and recursion, not strictly directory listing
The `directory.walk` object is a native atom (`eo-runtime/src/main/eo/directory.eo:62`, `[glob] > walk /Q.tuple`, implemented by `eo-runtime/src/main/java/org/eolang/EOdirectory$EOwalk.java`). That single atom does three things at once: it walks the tree recursively with `Files.walk`, it relativizes every path against the base directory, and it filters entries with a `glob:` `PathMatcher`.

Only the first of those needs the operating system. Path relativization and glob matching are pure computation and have no business living in Java. Bundling them into the atom makes the native surface larger than it needs to be, which cuts against the project's own direction of shrinking atoms into pure EO (see #5469, `i64.plus` should be pure EO). The recursion this would require already exists in EO right here — `deleted` walks a tuple with a self-referential helper (`directory.eo:22-27`) — so the traversal loop is expressible without Java too.

The smallest fix that resolves this is to shrink `walk` down to the one thing that genuinely needs a syscall: a minimal single-level directory listing (a `readdir`-style atom, or built on the existing `posix`/`win32` syscall atoms the way `made`, `deleted` and `open` already are). Recursion, path relativization and glob matching then move into EO on top of that primitive, and `directory.walk` stops being a native atom.

Thayorns:
`directory.walk` was a native atom doing three jobs at once: it walked the tree with `Files.walk`, it relativized every path against the base directory, and it filtered entries through a `glob:` `PathMatcher`. Only the walking part needs the operating system. Relativization and glob matching are pure computation and had no reason to sit in Java.

So the atom shrinks to the one thing that genuinely needs a syscall. `directory.listed` is a `readdir`-style primitive: it returns the sorted names of the direct children of a directory, and nothing else. `directory.walk` is now ordinary EO built on top of it — `walked` recurses depth-first and carries the relative prefix down with it, so relativization falls out of the traversal instead of being computed afterwards, and `translated` turns the glob into a regular expression that `string.regex` matches. The pre-order of `Files.walk` is preserved, which matters because `directory.deleted` walks the result backwards to remove children before their parents.

Two things are left as puzzles in `directory.eo`. The translator handles `*`, `**` and `?` and escapes regex metacharacters, but brace groups and negated classes still reach the regex engine untranslated. And the recursion asks `stat`, which follows symbolic links, where `Files.walk` did not — so a link pointing at its own ancestor would now recurse forever.

This is bigger than the usual size band. The atom it replaces has to go in the same commit as the EO that replaces it, otherwise neither half works.

Closes #5751


eo-runtime/src/main/java/org/eolang/EOdirectory$EOlisted.java
```diff
@@ -0,0 +1,61 @@
...
+            throw new ExFailure(
+                String.format("'%s' is not a valid path", raw),
+                ex
+            );
+        }
+        try (Stream<Path> paths = Files.list(path)) {
+            return new Data.ToPhi(
+                paths
+                    .map(p -> p.getFileName().toString())
+                    .sorted()
```
morphqdd, line 52: Neither this test file nor the EO-level tests in directory.eo (can-walk-recursively, can-walk-with-a-plain-glob-matching-direct-children-only, can-list-only-the-direct-children-of-a-directory) check the actual returned names or their order, only .length. A change that removed sorted() here, or returned the right count with wrong or garbled names, would pass every existing test.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -79,8 +88,90 @@
             "getpid"
             *
       number.random clock
-  [] > walk /Q.tuple
-    ? > glob /Q.string
+  [glob] > walk
+    tuple.reduced > @
+      [accum prefix dir] >> walked
+        dir.listed >> names
+        [index acc] >> stepped
```
morphqdd, line 99: stepped recurses once per directory entry rather than looping, and walked recurses once per directory level on top of that. A single directory with many entries walks up the EO call stack just as deep as the already-known stack-overflow class of bug in string.repeated (#6239) for large repeat counts. The Files.walk this replaces did not have that limit.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -79,8 +88,90 @@
...
+              if.
+                child.as-file.is-directory false
+                walked acc sub child.as-dir
+                acc.with sub
+          dir.as-path.resolved name > child
+          names.at index > name
+          if. > sub
+            prefix.length.eq 0
+            name
+            string.joined *1
```
morphqdd, line 114: Relative path segments are joined with a hardcoded "/" here instead of going through platform-aware path handling, unlike the Path.relativize() the old Java atom used. On Windows this changes the separator character embedded in every relative path this returns, and in base.resolved(rel) further down when it is turned back into an absolute path.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -79,8 +88,90 @@
...
+      double.if > token
+        ".*"
+        if.
+          char.eq "*"
+          "[^/]*"
+          if.
+            char.eq "?"
+            "[^/]"
+            if.
+              string.contains ".+()|^$\\" char
```
morphqdd: This escape list (.+()|^$\) does not include [, ], {, or }. That is a separate gap from the already-tracked "brace groups and negated classes are not translated" puzzle above: this is about a literal bracket or brace in a filename that was never meant as a glob wildcard at all, reaching the regex engine unescaped and changing what the pattern actually matches.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -79,8 +88,90 @@
...
+        ""
+        ^
+      *
+      if. > [accum rel] >>
+        matched rel
+        accum.with
+          Q.file (base.resolved rel).as-bytes
+        accum
+    ^.as-path >> base
+    [text] >> matched
```
morphqdd, line 139: string.regex is called here with no cant-compile fallback bound. The old Java atom gave a clear "Can't parse '%s' as a glob pattern" message for an invalid pattern; here an invalid translated regex (for example from the missing bracket escaping above) falls through to whatever string.regex.compile does with no fallback, which is a less specific failure than what this replaces.

yegor256:
@Thayorns conflict here

eo-runtime/src/main/eo/directory.eo
```diff
@@ -1,5 +1,18 @@
 # Directory in the file system.
 # Apparently every directory is a file.
+#
+# @todo #5751:45min Translate the rest of the glob syntax below. The `translated`
```
Favixx, line 4: This puzzle is accurate, and it has a consequence further down the file that is worth writing into it. `stops-on-walking-with-a-malformed-glob` (near the bottom, `d.as-dir.walk "["`) used to test that the atom validated the glob before using it — `FileSystems.getPathMatcher` rejected `[` and the `PatternSyntaxException` became an `ExFailure`. It now tests that an unescaped `[`, passed through untouched, happens to be rejected by `java.util.regex` on the far side.

The test still passes, so nothing is red, but it no longer means what its name says, and it is the only guard against exactly the class of input this puzzle is about. Worth saying so here, so that whoever picks the puzzle up knows to rewrite that test rather than to leave it green. `[a-z]` is the case that shows the difference: today it compiles as a regex class and quietly matches the wrong entries, which no test would notice.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -1,5 +1,18 @@
...
+#  object handles star, double star, question mark and the separator, but a brace
+#  group like {a,b} and a character class like [a-z] or [!a] reach the regex
+#  engine neither translated nor escaped, so a bracket or a brace in the glob
+#  either refuses to compile or matches the wrong entries.
+# @todo #5751:30min Stop descending into a symbolic link to a directory. The
+#  `walk` below asks stat, which follows the link, so a link to its own ancestor
+#  recurses forever, while the `Files.walk` it replaced only listed such a link.
+# @todo #5751:60min Walk a directory without growing the EO call stack. The
```
Favixx, line 12: The puzzle names `stepped` (once per entry) and `walked` (once per level) and stops there, but `translated` on line 160 recurses once per character of the glob and is on the same stack. A glob is normally short, so it is the least urgent of the three — but it is also the easiest to fix, since it is a plain left fold over a string with no tree underneath it, and leaving it out of the puzzle means whoever converts the other two to iteration will read this list and think they are done.

Worth one clause here.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -1,5 +1,18 @@
 # Directory in the file system.
 # Apparently every directory is a file.
+#
+# @todo #5751:45min Translate the rest of the glob syntax below. The `translated`
+#  object handles star, double star, question mark and the separator, but a brace
+#  group like {a,b} and a character class like [a-z] or [!a] reach the regex
+#  engine neither translated nor escaped, so a bracket or a brace in the glob
+#  either refuses to compile or matches the wrong entries.
+# @todo #5751:30min Stop descending into a symbolic link to a directory. The
```
Favixx, line 9: Worth being precise about the size of this regression, because the puzzle undersells it. `Files.walk` without `FOLLOW_LINKS` does not merely "list such a link" — it also cannot loop, ever, on any input. The recursion here can, and the condition is not exotic: a `sub/link -> ..` inside a directory somebody walks is enough, and the failure is an infinite descent that ends in a stack overflow or a full disk rather than an error naming the link.

Since `listed` returns only names and the recursion decides where to go on line 105 with `child.as-file.is-directory`, the fix lives entirely on this side — an `is-symlink` question next to `is-directory`, or a `listed` that says which children are links. If the latter, it changes the shape of the atom and is better decided now, while `listed` has one caller, than after the other two puzzles land.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -78,8 +92,106 @@
...
+          if. > sub
+            prefix.length.eq 0
+            name
+            string.joined *1
+              path.separator
+              prefix
+              name
+        | > @
+          0
+          accum.with prefix
```
Favixx, line 120: This line is what preserves pre-order — the prefix goes into the accumulator before `stepped` descends — and it is the sentence in the description that makes the whole change safe for `directory.deleted`, which walks the result backwards to remove children before their parents.

Nothing tests it. `can-walk-recursively` and `can-walk-with-a-plain-glob-matching-direct-children-only` both assert on content, not order, and the deletion tests never go through `walk`. Reverse the two lines here and every test in this file still passes, while `deleted` starts trying to remove a directory before its contents.

A test with `sub/b.txt` under `sub`, asserting that `string.joined "," (d.as-dir.walk "**")` puts `sub` before `sub/b.txt`, would cost four lines and pin the one property the description says matters.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -78,8 +92,106 @@
...
+      |
+        *
+        ""
+        ^
+      *
+      if. > [accum rel] >>
+        matched rel
+        accum.with
+          Q.file (base.resolved rel).as-bytes
```
Favixx, line 129: `base.resolved rel` re-joins the prefix that `walked` just took apart, so the tuple still holds absolute paths and `walk` keeps its old return type. Good — the description says the atom "relativized every path", which is not quite what the old code did: it relativized only for matching and returned `p.toString()` absolute. Ending up in the same place is the right outcome and worth stating in the description more exactly, since a reader who believes the old atom returned relative paths will think this PR changed the contract.

One case to check: for the root itself `rel` is the empty string, so this is `base.resolved ""`. Java's `Path.resolve("")` returns the base unchanged, which is right, but it is worth a test — `walk "**"` should contain the directory itself exactly as `Files.walk` did.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -78,8 +92,106 @@
...
+      as-bool. > @
+        exists.
+          next.
+            match.
+              compile.
+                string.regex
+                  string.joined *1 >>!
+                    ""
+                    "/^"
+                    translated 0 ""
```
Favixx, line 149: A question I could not settle by reading. `matched` is applied once per accumulated entry by `tuple.reduced`, and its body builds the pattern from `translated 0 ""` and hands it to `string.regex`. The `>>!` on the joined string above suggests you have already thought about this, but caching on an attribute of `matched` caches per copy of `matched`, and there is a copy per entry.

If that reading is right, walking a tree of N entries translates the glob N times, character by character, and compiles N identical `Pattern`s — where the old atom built one `PathMatcher` and reused it. On a large tree that is the difference between one compile and thousands. If the reading is wrong and EO hoists this to the `walk` copy, the answer belongs in a comment here, because the next person will ask the same question.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -78,8 +92,106 @@
...
+        "\\"
+        char
+      char
+    [text] >> matched
+      as-bool. > @
+        exists.
+          next.
+            match.
+              compile.
+                string.regex
```
Favixx, line 145: This works, and it is worth writing down why, because it is not obvious and it is fragile. The pattern handed to `string.regex` is `/^…$/`, and `translated` emits a bare `/` for the separator and another inside every `[^/]` — so on a Unix filesystem the literal is full of unescaped delimiters. It survives only because `regex.compile` finds the closing slash with `lastIndexOf('/')` rather than scanning forward.

That is a real coupling between two files that do not mention each other. The moment anyone appends flags here — `"$/i"` for a case-insensitive glob on Windows, which is the natural next request — the last slash stops being the last character and the whole thing still "works" while silently taking the wrong split point. Escaping the separator inside the pattern, or a comment on this line naming `regex.compile`'s behaviour, removes the trap.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -78,8 +92,106 @@
...
+        ""
+        ^
+      *
+      if. > [accum rel] >>
+        matched rel
+        accum.with
+          Q.file (base.resolved rel).as-bytes
+        accum
+    ^.as-path >> base
+    if. > [char] >> escaped
```
Favixx, line 132: The escape set is `.` `+` `(` `)` `|` `^` `$` `\`. Missing and flagged in the puzzle: `[`, `]`, `{`, `}`. Missing and not flagged: nothing else that matters, since `*` and `?` are consumed as glob syntax above.

Worth noting that `\` being in the set is what makes this correct on Windows — `escaped path.separator` there yields `\\`, so `single` becomes `[^\\]` rather than a broken class. That is load-bearing and invisible; a word in the comment above `single` saying that the separator is escaped precisely so that the Windows case works would stop someone "simplifying" this later.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -78,8 +92,106 @@
...
+                    translated 0 ""
+                    "$/"
+                T > [message] >>
+                  "Can't parse '%s' as a glob pattern, %s".printf
+                    * ^.glob message
+              text
+    string.joined *1 >> single
+      ""
+      "[^"
+      escaped path.separator >> literal
```
Favixx, line 158: `literal` is the escaped separator and `single` is `[^` + it + `]`, both computed once per `walk` — good, and the naming is clear.

One asymmetry: `translated` treats `/` as the separator on line 192 regardless of platform, while `literal` is `path.separator`, which is `\` on Windows. That is the right pair — globs are written with forward slashes everywhere, paths are not — but it means a Windows user who writes `walk "sub\\*.txt"` gets the backslash escaped as a literal and matched against a path that also uses backslashes, so it accidentally works, while `walk "sub/*.txt"` also works through the mapping. Two spellings both working is fine; it is worth a test asserting it, since it is the kind of thing a later simplification breaks silently.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -78,8 +92,106 @@
...
+      "]"
+    [index acc] >> translated
+      if. > @
+        index.gte glob.length
+        acc
+        translated
```
Favixx, line 164: `translated` advances by 2 for `**` and by 1 otherwise, and the lookahead on line 168 falls back to `""` past the end of the string — so `"*"` at the last position is read as a single star and cannot run off the end. That is the case I expected to be wrong and it is not.

Three stars is the case I could not convince myself about. `"***"` consumes `**` into `.*`, then the third `*` into `[^/]*`, giving `.*[^/]*` — which is harmless but is not what Java's `glob:***` does (it rejects it). Not worth fixing, but if you agree it is undefined, the first puzzle is the place to say that anything beyond two stars is unspecified.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -78,8 +92,106 @@
...
+        dir.listed >> names
+        [index acc] >> stepped
+          if. > @
+            index.gte names.length
+            acc
+            stepped
+              index.plus 1
+              if.
+                child.as-file.is-directory false
+                walked acc sub child.as-dir
```
Favixx, line 107: `walked acc sub child.as-dir` descends with `sub` as the new prefix, which is the relative path built two lines below — so each level carries only its own name and the join happens once per level rather than once per entry at the end. That is the good idea in this PR.

It does mean the prefix is rebuilt by `string.joined` at every level of depth, so a path N levels deep is joined N times, each join copying the prefix so far. Quadratic in depth, linear in practice since directories are shallow. Not worth changing; worth knowing when the stack puzzle is picked up, because an iterative rewrite that keeps a single mutable prefix removes this too.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -150,6 +262,24 @@
...
+        as-file.
+          d.resolved "плюшка.txt"
+      made.
+        as-dir.
+          d.resolved "sub"
+      touched.
+        as-file.
+          d.resolved "sub/deep.txt"
+      eq.
+        string.joined "," d.as-dir.listed
```
Favixx, line 277: One test for the new atom at the EO level, and it is a good one — `"sub,плюшка.txt"` covers sorting, the mixed-alphabet order and the fact that a directory is listed by name like anything else.

Two cases missing that the Java tests do not cover either. An empty directory: `listed` should give an empty tuple, and `string.joined ","` of it should be `""`. And `listed` on something that is not a directory: `Files.list` throws `NotDirectoryException`, which this atom then reports as `Can't list the directory '…'` — a message that names the wrong thing for a file, and nothing anywhere asserts what happens.

eo-runtime/src/main/eo/directory.eo
```diff
@@ -237,9 +367,10 @@
         as-file.
           d.resolved "sub/b.txt"
       eq.
-        length.
+        string.joined
```
Favixx, line 370: Good change — `length … eq 1` could not tell `a.txt` from `sub/b.txt`, and asserting the joined value pins both the filtering and the fact that `walk` still returns absolute paths.

The negative half is still untested though: this asserts that `a.txt` is in, and nothing asserts that `sub/b.txt` is out other than by the joined string being exactly one path — which is the same assertion, so a `walk` that returned `a.txt` twice would also pass. Since the whole point of `[^/]*` versus `.*` is that a plain glob does not cross directories, an explicit case with two matching files at two depths would be worth having.

eo-runtime/src/main/java/org/eolang/EOdirectory$EOlisted.java
```diff
@@ -0,0 +1,63 @@
...
+                    "'%s' is not a valid path: %s", raw, ex.getReason()
+                ),
+                ex
+            );
+        }
+        try (Stream<Path> paths = Files.list(path)) {
+            return new Data.ToPhi(
+                paths
+                    .map(p -> p.getFileName().toString())
+                    .sorted()
```
Favixx, line 52: `.sorted()` is a behaviour change and a good one — `Files.walk` returned whatever order the filesystem gave, which on ext4 is hash order and on APFS is something else again, so `walk` was never reproducible across machines and now is. It deserves a sentence in the PR description, since "the pre-order of `Files.walk` is preserved" is currently the only thing said about ordering and it is not the whole story: the order of siblings has changed on every platform.

The comparator is worth naming too. This sorts `String` by UTF-16 code unit, so `плюшка.txt` follows `sub`, and outside the BMP the order is not codepoint order. That is fine as a definition — it is stable and cheap — but it is a definition, and `EOdirectoryEOlistedTest.sortsNamesOfDirectChildren` relies on `Collections.sort` agreeing with it by coincidence rather than by contract.

eo-runtime/src/main/java/org/eolang/EOdirectory$EOlisted.java
```diff
@@ -0,0 +1,63 @@
...
+            path = Paths.get(raw).toAbsolutePath();
+        } catch (final InvalidPathException ex) {
+            throw new ExFailure(
+                String.format(
+                    "'%s' is not a valid path: %s", raw, ex.getReason()
+                ),
+                ex
+            );
+        }
+        try (Stream<Path> paths = Files.list(path)) {
```
Favixx, line 48: `Files.list` on a path that exists but is a regular file throws `NotDirectoryException`, which is an `IOException` and so lands in the catch below and comes out as `Can't list the directory '/x/y.txt'` — a message asserting something the user can see is false.

Since `walk` now reaches this atom for every entry it decided was a directory, based on `child.as-file.is-directory` on the EO side, a race between that check and this call is also newly possible: the directory is removed between the two, and the user gets the same misleading sentence. Distinguishing the two IOExceptions here — one message for "not a directory", one for "cannot be read" — is four lines and makes both cases self-explanatory.

eo-runtime/src/main/java/org/eolang/EOdirectory$EOlisted.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+@XmirObject(oname = "directory.listed")
+@SuppressWarnings("PMD.AvoidDollarSigns")
+public final class EOdirectory$EOlisted extends PhDefault implements Atom {
+
+    /**
+     * Ctor.
+     */
+    public EOdirectory$EOlisted() {
+        // nothing
```
Favixx, line 29: The old atom's constructor declared the `glob` void; this one declares nothing and exists only to hold a `// nothing`. `PhDefault` gives the same behaviour with no constructor at all, so unless qulice insists on an explicit one here, the six lines can go.

Worth checking against the other atoms that take no arguments in this package before removing — if they all carry the same empty constructor, it is a house style and this is consistent with it, which is a better reason to keep it than the comment gives.

eo-runtime/src/test/java/org/eolang/EOdirectoryEOlistedTest.java
```diff
@@ -0,0 +1,123 @@
...
+            () -> new Dataized(
+                this.listing(temp.resolve("отсутствует").toString())
+            ).take(),
+            "listing a directory that is not there must fail with ExFailure, not a raw IOException"
+        );
+    }
+
+    @Test
+    void sortsNamesOfDirectChildren(@TempDir final Path temp) throws IOException {
+        final long seed = System.nanoTime();
```
Favixx, line 69: Seeding from `System.nanoTime()` makes this test different on every run. Printing the seed in the failure message is the right mitigation and I am glad it is there, but the names are `"файл %04d %d ~"` with `index` in them, so they are unique whatever the random numbers are — which means the randomness is only shuffling the order the files are created in, and the assertion does not depend on creation order at all.

So the seed buys nothing and costs reproducibility. Either drop the `Random` and write eight fixed names chosen to be interesting — one with a trailing space, one differing only in case, one outside the BMP — or keep the randomness and assert something that creation order can actually affect.

eo-runtime/src/test/java/org/eolang/EOdirectoryEOlistedTest.java
```diff
@@ -0,0 +1,123 @@
...
+        Collections.sort(names);
+        MatcherAssert.assertThat(
+            String.format("listed names are not the children in sorted order, seed is %d", seed),
+            this.names(temp),
+            Matchers.equalTo(names)
+        );
+    }
+
+    @Test
+    void skipsGrandchildrenOfTheDirectory(@TempDir final Path temp) throws IOException {
```
Favixx, line 86: This is the test that carries the whole meaning of the change — `listed` stops at one level where `walk` did not — and it is a good one.

What the class does not have is the case that used to be covered by `EOdirectoryEOwalkTest`, which this PR deletes: eighty-nine lines of Java testing glob parsing and path errors, replaced by one `.eo` test for `listed` and the existing `walk` cases. The glob translator is now the most intricate code in `directory.eo` and it has no unit-level test at all — every assertion about it goes through a real temporary directory. A handful of `.eo` cases asserting that `walk` on a fixed tree matches for `*`, `**`, `?` and a literal separator would get that coverage back without a Java test for an object that is no longer Java.

Favixx (commented):
The premise is right and the execution mostly follows it. `Files.walk` + `relativize` + `PathMatcher` really was three jobs in one atom, only one of them needs a syscall, and `listed` is the correct primitive to stop at — `readdir` and nothing else. Carrying the prefix down the recursion so that relativization falls out of the traversal instead of being computed afterwards is the neat part; it is why `walked` needs no post-processing pass at all.

I checked the two things most likely to have broken and both survive. Pre-order is genuinely preserved: `walked` puts `prefix` into the accumulator before descending, so a directory still precedes its children and `directory.deleted` can keep walking the result backwards. And `**/*.txt` still means what Java's `glob:` meant — `.*` for the double star, `[^/]*` for the single one, so the recursive form still requires a separator and the plain one still refuses to cross directories. The unescaped `/` inside `[^/]` does not break the regex literal either, because `regex.compile` splits on `lastIndexOf('/')` rather than the first one.

What I would like looked at, in order:

**`stops-on-walking-with-a-malformed-glob` passes for a different reason than it used to,** and the difference is exactly the hole the first `@todo` describes. It used to assert that a glob is validated; it now asserts that Java's regex engine happens to reject one particular untranslated character.

**The ordering guarantee the description leans on is asserted nowhere.** "The pre-order is preserved, which matters because `directory.deleted` walks the result backwards" is the sentence that makes this change safe, and no test in this PR or the ones it touches would fail if the order reversed.

**Siblings are now sorted and they were not before.** `.sorted()` in `listed` is a real improvement — `Files.walk` gave whatever order the filesystem felt like — but it is a behaviour change that the description does not mention and no test pins.

The three puzzles are honest and well-scoped; my only quarrel is that the third one lists `stepped` and `walked` and forgets `translated`, which recurses per character of the glob. Details inline.

eo-runtime/src/test/java/org/eolang/EOdirectoryEOlistedTest.java
yegor256: @Favixx please, stop spamming. I will block you if you don't.

0crat:
@thayorns Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (496 >= 100); -8 for way too many hits-of-code (496 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+974](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (5) comments.  Your running score is [+3642](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+184](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2837](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6394, opened 2026-08-05
#6382: reject null writes to AtVoid and AtRho

### LINKED ISSUE #6382: `AtVoid.put(null)` returns success but doesn't fill the write-once attribute
## What happens

`AtVoid` uses `null` as the sentinel for an unset attribute. Its public `put(Phi)` method passes the caller value directly to `AtomicReference.compareAndSet(null, phi)`. With `phi == null`, Java performs a successful null-to-null CAS, so the method returns normally but the attribute remains vacant. A later non-null `put()` also succeeds.

This breaks the documented write-once contract: an attempted write is silently discarded, `vacant()` still returns true, and the caller receives neither a filled attribute nor an error.

## Problematic code

`eo-runtime/src/main/java/org/eolang/AtVoid.java`:

```java
@Override
    if (!this.object.compareAndSet(null, phi)) {
        throw new ExReadOnly(...);
    }
}
```

## Steps to reproduce

```java
AtVoid attr = new AtVoid("x");
attr.put(null);
assert attr.vacant();       // true: the apparent write changed nothing
attr.put(new PhDefault());  // succeeds: a second write is accepted
```

## Expected behavior

`null` is not a valid `Phi` value and should be rejected before the CAS with a clear exception. The same sentinel-based invariant should be checked for `AtRho`, whose `put(null)` has the analogous null-to-null behavior.

gemshrine:
Fixes #6382.

## What changed

- AtVoid.put() now rejects a null Phi before its atomic compare-and-set operation, preventing a null-to-null write from being reported as successful.
- AtRho.put() applies the same validation, so both sentinel-backed attributes preserve their write-once invariant.
- Added focused regression coverage in AtVoidTest and AtRhoTest. Each test verifies both the exception type and that the diagnostic states the invalid null contract.

## Why

Both classes represent an unset value with null. Passing null to AtomicReference.compareAndSet(null, null) succeeds without changing the state, allowing a later non-null write. Rejecting null at the public boundary prevents this silent no-op and makes the misuse explicit.

## Verification

- git diff --check
- Direct compilation of eo-runtime main sources passed with one CPU and 1 GB heap.
- The targeted test compilation is currently blocked by pre-existing missing generated classes EOnop and EOnumber in PhPackageTest, PhDefaultTest, and DataTest; this is unrelated to the four files changed here.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/AtRho.java
+++ b/eo-runtime/src/main/java/org/eolang/AtRho.java
@@ -5,6 +5,7 @@
 
 package org.eolang;
 
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
@@ -53,6 +54,7 @@ final class AtRho implements Attribute {
 
     @Override
     public void put(final Phi phi) {
+        Objects.requireNonNull(phi, "Attribute value can't be null");
         this.rho.compareAndSet(null, phi);
     }
 
--- a/eo-runtime/src/main/java/org/eolang/AtVoid.java
+++ b/eo-runtime/src/main/java/org/eolang/AtVoid.java
@@ -5,6 +5,7 @@
 
 package org.eolang;
 
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
@@ -71,6 +72,7 @@ public final class AtVoid implements Attribute {
 
     @Override
     public void put(final Phi phi) {
+        Objects.requireNonNull(phi, "Attribute value can't be null");
         if (!this.object.compareAndSet(null, phi)) {
             throw new ExReadOnly(
                 String.format(
--- a/eo-runtime/src/test/java/org/eolang/AtVoidTest.java
+++ b/eo-runtime/src/test/java/org/eolang/AtVoidTest.java
@@ -88,6 +88,18 @@ final class AtVoidTest {
     }
 
     @Test
+    void rejectsNullValue() {
+        MatcherAssert.assertThat(
+            "AtVoid must reject null instead of leaving the attribute vacant",
+            Assertions.assertThrows(
+                NullPointerException.class,
+                () -> new AtVoid("void").put(null)
+            ).getMessage(),
+            Matchers.containsString("can't be null")
+        );
+    }
+
+    @Test
     void isVacantWhenUnset() {
         MatcherAssert.assertThat(
             "Unset void attribute must be vacant, but it wasn't",
--- a/eo-runtime/src/test/java/org/eolang/AtRhoTest.java
+++ b/eo-runtime/src/test/java/org/eolang/AtRhoTest.java
@@ -69,4 +69,16 @@ final class AtRhoTest {
             Matchers.equalTo(obj)
         );
     }
+
+    @Test
+    void rejectsNullValue() {
+        MatcherAssert.assertThat(
+            "AtRho must reject null instead of silently keeping rho unset",
+            Assertions.assertThrows(
+                NullPointerException.class,
+                () -> new AtRho().put(null)
+            ).getMessage(),
+            Matchers.containsString("can't be null")
+        );
+    }
 }


eo-runtime/src/main/java/org/eolang/AtRho.java
```diff
@@ -53,6 +54,7 @@ public Phi get() {
 
     @Override
     public void put(final Phi phi) {
+        Objects.requireNonNull(phi, "Attribute value can't be null");
```
morphqdd, line 57: `.github/workflows/thresholds.yml` counts every literal `static|null` match across all `*.java` files and fails CI unless the count is *exactly* the checked-in threshold (840 right now, verified against current master with the same grep). This diff adds 8 new lowercase `null` matches across the four changed files ("can't be null" here and in AtVoid.java, plus "reject null", `put(null)`, and `containsString("can't be null")` in each of the two new tests) but doesn't touch `thresholds.yml`. The sibling PR #6391, which added its own null check to `Data.ToPhi`, had to bump this same threshold from 840 to 851 for exactly this reason - this PR looks like it needs the same bump (840 -> 848) or the `thresholds` CI job will fail.

Thayorns, line 57: The class docblock two lines up says AtRho "can be set only once, and it ignores all other puts", and `compareAndSet` on the next line is exactly that - a second `put` with a real object is swallowed without a sound. Now a null `put` is the one write that explodes. Whoever calls `put` twice gets silence, whoever calls it once with nothing gets an NPE, and neither response tells them the attribute already had a value. Is the null case really the one that deserves to be loud here?

eo-runtime/src/test/java/org/eolang/AtVoidTest.java
```diff
@@ -87,6 +87,18 @@ void throwsOnReset() {
...
 
+    @Test
+    void rejectsNullValue() {
+        MatcherAssert.assertThat(
+            "AtVoid must reject null instead of leaving the attribute vacant",
+            Assertions.assertThrows(
+                NullPointerException.class,
+                () -> new AtVoid("void").put(null)
+            ).getMessage(),
+            Matchers.containsString("can't be null")
```
Thayorns, line 98: The assertion is on the text of the exception message rather than on the state the guard is supposed to protect. `AtRhoTest.java:81` does the same. Reword the message in `AtVoid.put()` and both tests go red although nothing about the behavior changed; conversely a `put` that throws for some entirely different reason with a message mentioning "can't be null" would pass. The behavior the issue is about is that the attribute stays vacant and the next write is not silently swallowed, and that is observable through `vacant()`.

eo-runtime/src/main/java/org/eolang/AtVoid.java
```diff
@@ -71,6 +72,7 @@ public Phi get() {
 
     @Override
     public void put(final Phi phi) {
+        Objects.requireNonNull(phi, "Attribute value can't be null");
```
Thayorns, line 75: The message carries no values, while `this.name` is right there and the `ExReadOnly` three lines below already uses it to say which void attribute was hit. In a stack trace out of a dataization this one reads as `Attribute value can't be null` with nothing to tell you which attribute of which object was written to.

Thayorns (commented):
The diagnosis is right: `compareAndSet(null, null)` on an unset reference returns true, so a null write looked like it landed while the attribute stayed vacant and a later real write still went through. Guarding at the top of `put()` closes that. Three notes, mostly about what the new tests are pinned to.

yegor256:
@gemshrine conflicts here

yegor256:

gemshrine:

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (28) hits-of-code; +1 to give you at least something.  Your running score is [+3772](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (3) comments; -4 for too few (28) hits-of-code; +1 to give you at least something.  Your running score is [+1129](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2127](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6396, opened 2026-08-05
#6378: isolate byte arrays held by PhDefault

### LINKED ISSUE #6378: `PhDefault` allows mutation of byte arrays after object construction, violating data immutability
## What happens

The byte-array constructor of `PhDefault` stores its argument by reference, and `delta()` returns that same reference. Consequently, code outside the object can mutate an EO value after construction, and a caller that only reads `delta()` can alter the value for every later dataization. `copy()` also keeps sharing the same array because cloning `PhDefault` is shallow.

This differs from the runtime byte wrappers (`BytesRaw` and `BytesOf`), which defensively copy arrays. It makes an apparently stable `Phi` change according to unrelated Java-side mutations.

## Problematic code

```java
this.data = dta;
// ...
if (this.data != null) {
    bytes = this.data;
}
return bytes;
```

## Steps to reproduce

```java
byte[] raw = {0x01};
Phi phi = new PhDefault(raw);
raw[0] = 0x02;
// phi.delta()[0] is now 0x02

byte[] exposed = phi.delta();
exposed[0] = 0x03;
// every later phi.delta() observes 0x03
```

The same aliasing reaches EO through `new Data.ToPhi(byte[])`, which constructs a `PhDefault` from the caller-provided array.

## Expected behavior

A data-carrying `Phi` should own immutable snapshot bytes. `PhDefault` must defensively copy input data and return a copy from `delta()`, so copying or dataizing an object cannot be changed through an escaped Java array.

gemshrine:
Fixes #6378.

## What changed

- `PhDefault` now copies the byte array received by its data constructors instead of retaining the caller's reference.
- `PhDefault.delta()` returns a fresh copy rather than the object-owned storage.
- Both directions go through a small null-safe `isolated()` helper built on `Arrays.copyOf`, the same defensive-copy idiom already used by `BytesRaw`, `Quoted` and `VerboseBytesAsString`.

## Why

The previous implementation retained and returned the same mutable array, so a caller could alter an EO data value after construction or through a read-only `delta()` call. Copying on the way in and out preserves a stable byte snapshot while keeping the `Phi` API unchanged.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -10,6 +10,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -133,7 +134,7 @@ public class PhDefault implements Phi, Cloneable {
      * @param dta Object data
      */
     public PhDefault(final byte[] dta) {
-        this("", dta, PhDefault.NONE);
+        this("", Arrays.copyOf(dta, dta.length), PhDefault.NONE);
     }
 
     /**
@@ -142,7 +143,7 @@ public class PhDefault implements Phi, Cloneable {
      * @param attributes Initial attributes to register
      */
     public PhDefault(final byte[] dta, final Map<String, Attribute> attributes) {
-        this("", dta, attributes);
+        this("", Arrays.copyOf(dta, dta.length), attributes);
     }
 
     /**
@@ -272,7 +273,7 @@ public class PhDefault implements Phi, Cloneable {
                 this.forma()
             );
         }
-        return bytes;
+        return Arrays.copyOf(bytes, bytes.length);
     }
 
     @Override


eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -269,7 +270,7 @@ public byte[] delta() {
                 )
             );
         }
```
morphqdd, line 275: For the `Atom`/`PHI` delegation branches a few lines up, `bytes` already comes from another object's own `delta()` call, which (if that object is also a `PhDefault`) already returns a fresh, safe-to-keep copy. This line copies it again on top, so those two branches end up allocating and copying the array twice per call for no isolation benefit - only the `this.data != null` branch actually needs this copy.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -269,7 +270,7 @@ public byte[] delta() {
...
-        return bytes;
+        return Arrays.copyOf(bytes, bytes.length);
```
Thayorns, line 276: Only the first branch of this method needs the copy. In the other two `bytes` comes from `this.take(...).delta()`, and when that delegate is itself a PhDefault it has already handed back a fresh array, so copying it again duplicates work on every hop. A plain string built by `Data.ToPhi` is three PhDefault objects deep - the string, its bytes object, and the `PhDefault(byte[])` holding the payload - so one `delta()` on it now allocates and copies the payload three times instead of once, and this is on the path every dataization takes.

Thayorns (commented):
Copying on the way in and on the way out is the right shape for this: the array was previously both retained and handed back, so a caller could rewrite an EO data value after it was built. The two tests pin exactly those two directions, and each one mutates and then asserts on `delta()`, which is the observable the issue is about. One remark on where the outgoing copy sits.

yegor256:

gemshrine:

eo-runtime/src/test/java/org/eolang/PhDefaultDataTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link PhDefault} data isolation.
+ * @since 0.1
+ */
+@SuppressWarnings("JTCOP.RuleAllTestsHaveProductionClass")
```
yegor256: @gemshrine don't suppress, fix

yegor256 (changes requested):

eo-runtime/src/test/java/org/eolang/PhDefaultDataTest.java
gemshrine: Extracted the defensive copy into a `Snapshot` value object; `SnapshotTest` maps to it and the suppression is gone.

gemshrine:

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (7) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2416](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (7) hits-of-code; +1 to give you at least something.  Your running score is [+4852](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (7) hits-of-code; +1 to give you at least something.  Your running score is [+1392](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (7) hits-of-code; +1 to give you at least something.  Your running score is [+2601](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6397, opened 2026-08-05
#6379: preserve out-of-range integral numeral values

### LINKED ISSUE #6379: `PhDefault.numeral()` incorrectly converts large integral doubles to `Long.MAX_VALUE` in φ-terms
## What happens

`PhDefault.numeral()` renders every finite whole-valued `double` by casting it to `long`. Java saturates an out-of-range floating-point-to-long conversion. Therefore any integral double at or above `2^63` is rendered as `9223372036854775807`, even though that is a different number.

For example, `Math.scalb(1.0, 63)` is exactly `9223372036854775808`, but its φ-term becomes `9223372036854775807`. The same applies to larger finite integral doubles and their negative counterparts.

## Problematic code

```java
if (value == Math.floor(value) && !Double.isInfinite(value)) {
    txt = Long.toString((long) value);
} else {
    txt = Double.toString(value);
}
```

## Steps to reproduce

```java
PhDefault.numeral(Math.scalb(1.0, 63));
```

Actual result: `9223372036854775807`

Expected result: a spelling of `9223372036854775808`, for example `9.223372036854776E18`.

## Expected behavior

Use the integer spelling only when the whole-valued double lies within the signed `long` range. Larger finite integral values must retain `Double.toString(value)` so the φ-term represents the original value.

gemshrine:
Fixes #6379.

## What changed

- PhDefault.numeral() now uses integer spelling only when a whole double fits in the long range.
- Integral doubles at or beyond the long bounds retain their Double.toString() representation instead of being narrowed to Long.MAX_VALUE or Long.MIN_VALUE.
- Added regressions for positive and negative values outside the representable long range.

## Why

Casting an out-of-range double to long saturates it. The earlier integral-value branch then emitted that saturated long, so distinct φ-terms collapsed to a boundary value. Keeping the double representation preserves the original numeric value.

## Verification

PhDefaultTest passed with one CPU, 1 GB heap, and JUnit parallel execution disabled.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -338,32 +338,12 @@ public class PhDefault implements Phi, Cloneable {
             if ("string".equals(name)) {
                 result = new Quoted(raw).get();
             } else {
-                result = PhDefault.numeral(new BytesOf(raw).asNumber());
+                result = new Numeral(new BytesOf(raw).asNumber()).get();
             }
         } else {
             result = this.structural();
         }
         return result;
-    }
-
-    /**
-     * Render a number value as a φ-term.
-     * @param value The number
-     * @return Whole values without a fraction, decimals otherwise
-     */
-    static String numeral(final double value) {
-        final String txt;
-        if (
-            value == Math.floor(value)
-                && !Double.isInfinite(value)
-                && Long.MIN_VALUE <= value
-                && value < Long.MAX_VALUE
-        ) {
-            txt = Long.toString((long) value);
-        } else {
-            txt = Double.toString(value);
-        }
-        return txt;
     }
 
     /**
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -353,7 +353,12 @@ public class PhDefault implements Phi, Cloneable {
      */
     static String numeral(final double value) {
         final String txt;
-        if (value == Math.floor(value) && !Double.isInfinite(value)) {
+        if (
+            value == Math.floor(value)
+                && !Double.isInfinite(value)
+                && Long.MIN_VALUE <= value
+                && value < Long.MAX_VALUE
+        ) {
             txt = Long.toString((long) value);
         } else {
             txt = Double.toString(value);
--- a/eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
+++ b/eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
@@ -67,22 +67,6 @@ final class PhDefaultTest {
     }
 
     @Test
-    void printsOutOfRangeIntegralNumbersWithoutSaturation() {
-        final double positive = Math.scalb(1.0d, 63);
-        final double negative = -Math.scalb(1.0d, 64);
-        MatcherAssert.assertThat(
-            "Positive integral double outside long range must retain its value in φ-term",
-            PhDefault.numeral(positive),
-            Matchers.equalTo(Double.toString(positive))
-        );
-        MatcherAssert.assertThat(
-            "Negative integral double outside long range must retain its value in φ-term",
-            PhDefault.numeral(negative),
-            Matchers.equalTo(Double.toString(negative))
-        );
-    }
-
-    @Test
     void printsUnsetNumberStructurally() {
         MatcherAssert.assertThat(
             "Number without injected bytes must fall back to its structural φ-term, but it didnt",
--- a/eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
+++ b/eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
@@ -67,6 +67,22 @@ final class PhDefaultTest {
     }
 
     @Test
+    void printsOutOfRangeIntegralNumbersWithoutSaturation() {
+        final double positive = Math.scalb(1.0d, 63);
+        final double negative = -Math.scalb(1.0d, 64);
+        MatcherAssert.assertThat(
+            "Positive integral double outside long range must retain its value in φ-term",
+            PhDefault.numeral(positive),
+            Matchers.equalTo(Double.toString(positive))
+        );
+        MatcherAssert.assertThat(
+            "Negative integral double outside long range must retain its value in φ-term",
+            PhDefault.numeral(negative),
+            Matchers.equalTo(Double.toString(negative))
+        );
+    }
+
+    @Test
     void printsUnsetNumberStructurally() {
         MatcherAssert.assertThat(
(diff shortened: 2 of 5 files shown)


eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -347,7 +347,12 @@ public final void add(final String name, final Attribute attr) {
      */
     static String numeral(final double value) {
         final String txt;
-        if (value == Math.floor(value) && !Double.isInfinite(value)) {
+        if (
+            value == Math.floor(value)
+                && !Double.isInfinite(value)
+                && Long.MIN_VALUE <= value
+                && value < Long.MAX_VALUE
```
morphqdd: Worth a comment here: `Long.MAX_VALUE` widens to `double` as exactly `2^63` (Java promotes the `long` operand), not the true `2^63 - 1` this literal represents - doubles can't represent `2^63 - 1` exactly. So this line actually compares against `2^63`, not `Long.MAX_VALUE`. It's still correct only because no `double` is representable between `2^63 - 1024` (the largest representable double below `2^63`) and `2^63` itself, so nothing ever reaches `(long)` cast in that gap. That's a fairly fragile invariant to rely on silently - a future change to this comparison (e.g. swapping in a `BigDecimal` bound, or someone "fixing" the widening) could reintroduce the saturation bug this PR is fixing, without any test catching it unless it happens to hit exactly this magnitude.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
```diff
@@ -66,6 +66,22 @@ void printsVoidAttributeAsTerm() {
         );
     }
 
+    @Test
+    void printsOutOfRangeIntegralNumbersWithoutSaturation() {
+        final double positive = Math.scalb(1.0d, 63);
+        final double negative = -Math.scalb(1.0d, 64);
```
morphqdd: The positive case tests exactly at the boundary (`2^63`, where `Math.floor(value) == value` and it's right at the excluded edge). The negative case uses `-2^64`, well past `Long.MIN_VALUE` (`-2^63`) rather than at or just past the boundary itself. Since `Long.MIN_VALUE` is inclusive here (`<=`) while `Long.MAX_VALUE` is exclusive (`<`), the two boundaries behave differently and only one of them is tested precisely.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
```diff
@@ -66,6 +66,22 @@ void printsVoidAttributeAsTerm() {
...
+        MatcherAssert.assertThat(
+            "Positive integral double outside long range must retain its value in φ-term",
+            PhDefault.numeral(positive),
+            Matchers.equalTo(Double.toString(positive))
+        );
+        MatcherAssert.assertThat(
```
Thayorns: Two assertions in one test, and the first is not the last statement, so a failure on the positive value hides whatever the negative one would have done. These are two behaviors - the upper bound and the lower bound - and each deserves to fail on its own.

eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
```diff
@@ -66,6 +66,22 @@ void printsVoidAttributeAsTerm() {
...
     }
 
+    @Test
+    void printsOutOfRangeIntegralNumbersWithoutSaturation() {
+        final double positive = Math.scalb(1.0d, 63);
+        final double negative = -Math.scalb(1.0d, 64);
+        MatcherAssert.assertThat(
+            "Positive integral double outside long range must retain its value in φ-term",
+            PhDefault.numeral(positive),
+            Matchers.equalTo(Double.toString(positive))
```
Thayorns: The expectation is `Double.toString(positive)`, which is the exact expression the else branch of `numeral` runs, so the test restates the implementation rather than naming what a φ-term for 2^63 should look like. Every other test in this file goes through `φTerm()` and spells out the literal it expects; this one reaches into the package-private helper and asks it to agree with itself. Written as the literal string, it would also document the fix for whoever reads it next.

Thayorns (commented):
The range guard is right, including the part that looks off at a glance: `Long.MAX_VALUE` widens to 2^63 in that comparison, so `value < Long.MAX_VALUE` really does exclude exactly the doubles that would saturate, while `Long.MIN_VALUE <= value` keeps -2^63, which is representable. Two remarks on the test.

yegor256:
@gemshrine conflicts here

gemshrine:

eo-runtime/src/test/java/org/eolang/PhDefaultNumeralTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link PhDefault} numeral rendering.
+ * @since 0.1
+ */
+@SuppressWarnings("JTCOP.RuleAllTestsHaveProductionClass")
```
yegor256: @gemshrine don't suppress, fix: listen to what JTCOP is asking you to do

gemshrine: Extracted the rendering into a `Numeral` value object (mirroring the neighbouring `Quoted`); `NumeralTest` now maps to it and the suppression is removed.

gemshrine:

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (108 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2428](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+4859](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+1399](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2608](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6398, opened 2026-08-05
#6380: validate SockaddrIn padding length

### LINKED ISSUE #6380: `SockaddrIn` constructor doesn't restrict padding array length, causing malformed address structures
## What happens

`sockaddr_in` has an eight-byte `sin_zero` padding field. `SockaddrIn` documents its `zero` constructor argument as “Zero 8 bytes”, but accepts and copies any array length. JNA derives the native structure layout from the public array field, so a caller can accidentally construct a structure shorter or longer than the platform ABI requires. Native calls then receive a malformed address structure rather than a clear Java diagnostic.

## Problematic code

`eo-runtime/src/main/java/org/eolang/SockaddrIn.java`:

```java
public SockaddrIn(final short family, final short port, final int addr, final byte[] zero) {
    super();
    this.family = family;
    this.port = port;
    this.addr = addr;
    this.zero = Arrays.copyOf(zero, zero.length);
}
```

## Steps to reproduce

```java
new SockaddrIn((short) 2, (short) 0, 0, new byte[0]);
new SockaddrIn((short) 2, (short) 0, 0, new byte[16]);
```

Both calls succeed, although neither instance has the layout of a C `sockaddr_in`.

## Expected behavior

The constructor should reject every padding array whose length is not exactly eight, with an explanatory exception. The no-argument and convenience constructors should continue to provide the valid eight-byte padding.

gemshrine:
Fixes #6380.

## What changed

- SockaddrIn now requires its padding array to contain exactly eight bytes, matching the native sockaddr_in layout.
- Replaced the duplicated literal with a shared padding-size constant.
- Added parameterized regressions for both undersized and oversized arrays, including the diagnostic text.

## Why

The constructor previously copied any padding length, producing malformed native address structures. Enforcing the layout boundary rejects invalid inputs before they reach JNA/native calls.

## Verification

- git diff --check
- SockaddrInTest passed in a single-core Maven run with JUnit parallel execution disabled.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/SockaddrIn.java
+++ b/eo-runtime/src/main/java/org/eolang/SockaddrIn.java
@@ -66,26 +66,16 @@ public final class SockaddrIn extends Structure {
      * @param zero Zero 8 bytes
      */
     public SockaddrIn(final short family, final short port, final int addr, final byte[] zero) {
-        this(family, port, addr, SockaddrIn.checked(zero), true);
-    }
-
-    /**
-     * Primary ctor.
-     * @param family Family
-     * @param port Port
-     * @param addr Address
-     * @param zero Validated padding
-     * @param valid Padding validation marker
-     */
-    private SockaddrIn(
-        final short family, final short port, final int addr, final byte[] zero,
-        final boolean valid
-    ) {
         super();
         this.family = family;
         this.port = port;
         this.addr = addr;
-        this.zero = zero;
+        this.zero = SockaddrIn.checked(zero);
+    }
+
+    @Override
+    public List<String> getFieldOrder() {
+        return Arrays.asList("family", "port", "addr", "zero");
     }
 
     /**
@@ -103,10 +93,5 @@ public final class SockaddrIn extends Structure {
             );
         }
         return Arrays.copyOf(zero, zero.length);
-    }
-
-    @Override
-    public List<String> getFieldOrder() {
-        return Arrays.asList("family", "port", "addr", "zero");
     }
 }
--- a/eo-runtime/src/main/java/org/eolang/SockaddrIn.java
+++ b/eo-runtime/src/main/java/org/eolang/SockaddrIn.java
@@ -66,11 +66,23 @@ public final class SockaddrIn extends Structure {
      * @param zero Zero 8 bytes
      */
     public SockaddrIn(final short family, final short port, final int addr, final byte[] zero) {
+        this(SockaddrIn.checked(zero), family, port, addr);
+    }
+
+    /**
+     * Primary ctor, receiving an already validated padding. The padding
+     * comes first so this signature differs from the public one above.
+     * @param zero Validated padding
+     * @param family Family
+     * @param port Port
+     * @param addr Address
+     */
+    private SockaddrIn(final byte[] zero, final short family, final short port, final int addr) {
         super();
         this.family = family;
         this.port = port;
         this.addr = addr;
-        this.zero = SockaddrIn.checked(zero);
+        this.zero = zero;
     }
 
     @Override
--- a/eo-runtime/src/main/java/org/eolang/SockaddrIn.java
+++ b/eo-runtime/src/main/java/org/eolang/SockaddrIn.java
@@ -66,7 +66,34 @@ public final class SockaddrIn extends Structure {
      * @param zero Zero 8 bytes
      */
     public SockaddrIn(final short family, final short port, final int addr, final byte[] zero) {
+        this(family, port, addr, SockaddrIn.checked(zero), true);
+    }
+
+    /**
+     * Primary ctor.
+     * @param family Family
+     * @param port Port
+     * @param addr Address
+     * @param zero Validated padding
+     * @param valid Padding validation marker
+     */
+    private SockaddrIn(
+        final short family, final short port, final int addr, final byte[] zero,
+        final boolean valid
+    ) {
         super();
+        this.family = family;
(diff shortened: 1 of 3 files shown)


eo-runtime/src/main/java/org/eolang/SockaddrIn.java
```diff
@@ -62,6 +67,14 @@ public SockaddrIn(final short family, final short port, final int addr) {
      */
     public SockaddrIn(final short family, final short port, final int addr, final byte[] zero) {
         super();
+        if (zero.length != SockaddrIn.PADDING_SIZE) {
+            throw new IllegalArgumentException(
```
morphqdd, line 100: This is reachable with attacker/user-controlled data: `ConnectSyscall`, `BindSyscall`, `AcceptSyscall` (and their win32 counterparts) build `SockaddrIn` from `new Dataized(params[1].take("padding")).take()`, i.e. straight from whatever the EO program passes in. A raw `IllegalArgumentException` here isn't caught anywhere between this constructor and `Dataized` - its own javadoc says only `ExFailure` gets special handling at that boundary. Every other validation failure in the sibling `posix`/`win32` packages throws `ExFailure` instead (grep for `IllegalArgumentException`/`IllegalStateException` in those two packages turns up nothing) - this would be the first raw JDK exception reachable from that call path, surfacing to the EO user as an unhandled Java stack trace instead of a normal EO-catchable failure. `ExFailure` takes the same `(String, Object...)` format-string constructor already used here, so it's a direct swap.

eo-runtime/src/test/java/org/eolang/SockaddrInTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/**
+ * Test case for {@link SockaddrIn}.
+ * @since 0.40.0
+ */
+final class SockaddrInTest {
```
morphqdd, line 15: This new file only tests the rejection path (sizes 0 and 16). There's no test confirming a correctly-sized 8-byte array still constructs successfully - a regression that made the check reject the valid case too (e.g. an off-by-one in `PADDING_SIZE` or the comparison) wouldn't be caught by anything in this file.

eo-runtime/src/main/java/org/eolang/SockaddrIn.java
```diff
@@ -62,6 +67,14 @@ public SockaddrIn(final short family, final short port, final int addr) {
      */
     public SockaddrIn(final short family, final short port, final int addr, final byte[] zero) {
         super();
+        if (zero.length != SockaddrIn.PADDING_SIZE) {
```
Thayorns, line 99: The primary constructor now carries a branch and a throw rather than only assignments. This project treats that as a rule it has to opt out of: `Data.java:55` and `eo-maven-plugin/.../OyRemote.java:72` both carry an explicit `@checkstyle ConstructorsCodeFreeCheck` suppression for exactly this. Nothing here declares one, and the shape of the object is what changed - the padding is now something `SockaddrIn` inspects on the way in instead of something it is simply handed.

eo-runtime/src/test/java/org/eolang/SockaddrInTest.java
```diff
@@ -0,0 +1,33 @@
...
+    void rejectsPaddingWithUnexpectedSize(final int size) {
+        final IllegalArgumentException error = Assertions.assertThrows(
+            IllegalArgumentException.class,
+            () -> new SockaddrIn((short) 2, (short) 0, 0, new byte[size]),
+            "SockaddrIn must reject padding that is not eight bytes long"
+        );
+        MatcherAssert.assertThat(
+            "The exception must explain the required padding size",
+            error.getMessage(),
+            Matchers.containsString("exactly 8 bytes")
```
Thayorns: This is a second assertion in the same test, and what it pins is the wording of the message. Rephrasing the format string in the constructor - saying "8 bytes exactly", or naming the value first - turns this red although the behavior the test is named for has not moved at all. The behavior worth holding down is that the object refuses to be built, and `assertThrows` above already has that.

Thayorns (commented):
Refusing a padding array that is not eight bytes is the right call - a wrong-sized array here would have produced a malformed native structure that only misbehaves once it crosses into JNA. Naming the size once instead of writing eight zeroes twice reads better too. Two remarks.

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2087](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+3649](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+1013](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6399, opened 2026-08-05
#6376: configure the EO logger only once

### LINKED ISSUE #6376: `Main.main()` causes duplicate runtime log messages on multiple in-process invocations
## What happens

Every invocation of `Main.main()` calls `Main.setup()`. That method constructs a fresh `ConsoleHandler` and adds it to the static `EOLOG` logger without checking whether it was configured already.

When the runtime entry point is invoked more than once in the same JVM — for example by an embedding application or an in-process test harness — each EO log record is emitted once per prior invocation. The number of duplicate lines grows with every run.

## Problematic code

`eo-runtime/src/main/java/org/eolang/Main.java:71-72` calls setup on every run, and `Main.setup()` unconditionally executes:

```java
final Handler handler = new ConsoleHandler();
// formatter setup
Main.EOLOG.addHandler(handler);
```

## Steps to reproduce

1. Invoke `Main.main(...)` twice in the same JVM with input that causes EO logging.
2. Observe that the second invocation prints each runtime log message twice.
3. Repeat; the third invocation prints it three times.

## Expected behavior

Logger configuration should be idempotent. The runtime should add its handler only once, or remove/reuse the existing handler before a subsequent in-process invocation, so each event is emitted exactly once.

gemshrine:
Fixes #6376.

## What changed

- Main creates its formatted EO log handler once and reuses it on later calls to Main.main().
- setup() synchronizes handler installation and checks identity, so repeated or concurrent in-process entry-point calls cannot attach duplicate handlers.
- Added a regression to MainTest that invokes Main.main() twice and verifies the logger handler count remains stable.

## Why

Main.setup() previously constructed and appended a ConsoleHandler every time. Each later invocation then emitted every EO log record through all accumulated handlers. The setup is now idempotent while preserving any separately configured handlers.

## Verification

- git diff --check
- eo-runtime main Java sources compiled successfully with one CPU and 1 GB heap.
- Full test compilation remains blocked by unrelated missing EOnop and EOnumber generated classes in the current local base.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Main.java
+++ b/eo-runtime/src/main/java/org/eolang/Main.java
@@ -12,6 +12,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.ConsoleHandler;
 import java.util.logging.Formatter;
 import java.util.logging.Handler;
@@ -56,6 +57,16 @@ public final class Main {
      * EO app-wide logger.
      */
     private static final Logger EOLOG = Logger.getLogger("org.eolang");
+
+    /**
+     * Handler for EO runtime logs.
+     */
+    private static final Handler HANDLER = Main.handler();
+
+    /**
+     * Lock guarding the EO runtime logger configuration.
+     */
+    private static final ReentrantLock LOCK = new ReentrantLock();
 
     /**
      * Not for instantiation.
@@ -142,6 +153,24 @@ public final class Main {
      * Setup logs.
      */
     private static void setup() {
+        Main.LOCK.lock();
+        try {
+            if (Arrays.stream(Main.EOLOG.getHandlers()).noneMatch(
+                handler -> handler == Main.HANDLER
+            )) {
+                Main.EOLOG.addHandler(Main.HANDLER);
+            }
+        } finally {
+            Main.LOCK.unlock();
+        }
+        Main.EOLOG.setUseParentHandlers(false);
+    }
+
+    /**
+     * Make a handler for EO runtime logs.
+     * @return Configured handler
+     */
+    private static Handler handler() {
         final Handler handler = new ConsoleHandler();
         handler.setFormatter(
             new Formatter() {
@@ -151,8 +180,7 @@ public final class Main {
                 }
             }
         );
-        Main.EOLOG.addHandler(handler);
-        Main.EOLOG.setUseParentHandlers(false);
+        return handler;
     }
 
     /**
--- a/eo-runtime/src/test/java/org/eolang/MainTest.java
+++ b/eo-runtime/src/test/java/org/eolang/MainTest.java
@@ -7,6 +7,7 @@ package org.eolang;
 import com.yegor256.Jaxec;
 import com.yegor256.Jhome;
 import com.yegor256.Result;
+import java.util.logging.Logger;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
@@ -124,6 +125,17 @@ final class MainTest {
                 Matchers.containsString("at org.eolang.PhPackage.take"),
                 Matchers.containsString("at org.eolang.Main.run")
             )
+        );
+    }
+
+    @Test
+    void configuresRuntimeLoggerOnlyOnce() throws Exception {
+        Main.main("--version");
+        Main.main("--version");
+        MatcherAssert.assertThat(
+            "Repeated Main.main calls must not add another EO log handler",
+            Logger.getLogger("org.eolang").getHandlers().length,
+            Matchers.equalTo(1)
         );
     }
 


eo-runtime/src/main/java/org/eolang/Main.java
```diff
@@ -57,6 +57,11 @@ public final class Main {
      */
     private static final Logger EOLOG = Logger.getLogger("org.eolang");
 
+    /**
+     * Handler for EO runtime logs.
+     */
+    private static final Handler HANDLER = Main.handler();
```
morphqdd, line 64: This static field initializer runs at class-load time, so a `ConsoleHandler` gets constructed the moment `Main` is loaded, even if `main()`/`setup()` is never called - e.g. if something just references `Main.class` or loads it during test discovery. Building it lazily inside `setup()` (guarded by the same identity check) would keep the side effect tied to actual use instead of class loading.

eo-runtime/src/main/java/org/eolang/Main.java
```diff
@@ -142,6 +147,21 @@ private static void print(final Throwable thr) {
      * Setup logs.
      */
     private static void setup() {
```
morphqdd, line 155: Locking on `Main.EOLOG` means synchronizing on the shared `java.util.logging.Logger` instance for `"org.eolang"` - an object this class doesn't own and that `java.util.logging`'s own implementation may lock internally for unrelated operations. A private lock object dedicated to this class (e.g. a plain `new Object()` field) would avoid sharing a monitor with code outside `Main`'s control.

eo-runtime/src/main/java/org/eolang/Main.java
Thayorns, line 64: There is now exactly one handler for the lifetime of the JVM, and `parse` mutates it: line 187 walks `EOLOG.getHandlers()` and sets each one to `Level.FINE` when `--verbose` is present. Nothing puts the level back. So one `main("--verbose", ...)` leaves this handler at FINE, and every plain `main()` afterwards in the same process keeps printing verbose output that nobody asked for. That is the same class of leak the issue is about, just moved from the handler count to the handler level.

eo-runtime/src/main/java/org/eolang/Main.java
```diff
@@ -142,6 +147,21 @@ private static void print(final Throwable thr) {
...
+        }
+        Main.EOLOG.setUseParentHandlers(false);
+    }
+
+    /**
+     * Make a handler for EO runtime logs.
+     * @return Configured handler
+     */
+    private static Handler handler() {
         final Handler handler = new ConsoleHandler();
```
Thayorns, line 174: `ConsoleHandler` captures `System.err` in its constructor, and this constructor now runs during class initialization of `Main` rather than at the top of each `main()`. Anything that loads the class and then redirects `System.err` - an embedder, or a test capturing output - gets its EO log records on the original stream. Before, a call to `main()` after the redirect picked up the new one.

eo-runtime/src/test/java/org/eolang/MainTest.java
```diff
@@ -185,6 +186,18 @@ void printsFullStacktraceIfVerboseIsEnabled() {
         );
     }
 
+    @Test
+    void configuresRuntimeLoggerOnlyOnce() throws Exception {
+        Main.main("--version");
```
Thayorns, line 133: This is the only test in the file that runs `Main` in the surefire JVM; every other one goes through `exec` and forks a real `java` process. Calling `main` here attaches the console handler to `org.eolang` and sets `setUseParentHandlers(false)` on it, and both survive the test and apply to whatever runs next in the same fork. It also means the version banner from `--version` lands in the build output twice, because that log record now goes through the handler this test just installed.

Thayorns (commented):
Making `setup()` idempotent is the right shape for the problem - the old code stacked a fresh `ConsoleHandler` on `org.eolang` per call, so the second in-process `main()` printed everything twice. Promoting the handler to a single shared instance fixes that, but it also turns everything anyone does to that handler into permanent global state, and the new test does some of that to the surefire JVM. Three things.

yegor256:
@gemshrine conflicts here

yegor256:
@gemshrine conflicts here

gemshrine:

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+3779](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+1136](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2143](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6400, opened 2026-08-05
#6381: remove stale Java sources after transpiling

### LINKED ISSUE #6381: `Transpiling` leaves behind obsolete `.java`files in `target/generated-sources` after XMIR objects are removed
## What happens

`Transpiling` generates Java files only for the current collection of XMIR sources. `JavaFiles.total()` writes files for the `<class>` nodes it sees, but neither it nor `Transpiling` reconciles the generated-source directory with the current XMIR set.

If an EO object or a nested object class is removed, its old `.java` file stays in `target/generated-sources`. Maven may compile and package that obsolete class. The incremental build therefore differs from a clean build and can retain APIs that no longer exist in EO.

## Problematic code

`Transpiling.exec()` iterates current `sources`, while `JavaFiles.total()` only creates/updates paths for current XMIR classes. There is no code that identifies and deletes a generated `.java` file with no corresponding current XMIR `<class>`.

## Steps to reproduce

1. Transpile an EO source that generates class `EOfoo.java`.
2. Remove the EO object (or a nested object that produces that class).
3. Run the transpile/compile pipeline again without cleaning `target`.
4. Observe that `target/generated-sources/.../EOfoo.java` remains and Maven still compiles it.

## Expected behavior

An incremental transpile must converge to the same generated-source tree as a clean build. It should remove files no longer represented by current XMIR, or maintain a generation manifest and clean obsolete entries before compilation.

gemshrine:
Fixes #6381.

## What changed

- After all current XMIR sources are transpiled, Transpiling reconciles the generated-source directory with the classes represented in the current XMIR set.
- JavaFiles removes obsolete generated Java files and obsolete package-info.java files rather than leaving them for Maven to compile.
- Added regression coverage for stale source cleanup in JavaFilesTest and the transpilation integration path in MjTranspileTest.

## Why

Incremental transpilation previously only wrote classes that still existed. Removing an EO object or nested generated class left its old Java source under target/generated-sources, making incremental output differ from a clean build. The reconciliation step makes the generated tree converge to the current XMIR input.

## Verification

JavaFilesTest and MjTranspileTest passed in focused runs.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
@@ -11,9 +11,13 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * The Java files that a transpiled XMIR is written out as.
@@ -49,6 +53,15 @@ final class JavaFiles {
     private final boolean enabled;
 
     /**
+     * Java files generated during the current transpilation.
+     *
+     * <p>The collection is shared by parallel {@link #total(boolean, Path,
+     * String, boolean)} calls. It is reconciled only after all XMIRs have
+     * been processed.</p>
+     */
+    private final Collection<Path> fresh;
+
+    /**
      * Ctor.
      * @param dir Generated sources directory
      * @param cached Cache directory for this transpile version
@@ -58,6 +71,7 @@ final class JavaFiles {
         this.generated = dir;
         this.cache = cached;
         this.enabled = caching;
+        this.fresh = new ConcurrentLinkedQueue<>();
     }
 
     /**
@@ -88,6 +102,7 @@ final class JavaFiles {
                     final Path tgt = new Place(jname).make(
                         this.generated, JavaFiles.JAVA
                     );
+                    this.fresh.add(tgt);
                     final Footprint java = new FpJavaGenerated(
                         clazz, new FileGenerationReport(saved, tgt, target)
                     );
@@ -114,6 +129,37 @@ final class JavaFiles {
             );
         }
         return saved.get();
+    }
+
+    /**
+     * Delete generated Java files absent from the current XMIR collection.
+     *
+     * @throws IOException If fails to inspect or remove a generated file
+     */
+    void removeStale() throws IOException {
+        if (Files.exists(this.generated)) {
+            final Set<Path> expected = new HashSet<>(this.fresh);
+            final Set<Path> dirs = new HashSet<>();
+            for (final Path file : expected) {
+                for (Path dir = file.getParent(); dir != null && dir.startsWith(this.generated);
+                    dir = dir.getParent()) {
+                    dirs.add(dir);
+                }
+            }
+            try (Stream<Path> walk = Files.walk(this.generated)) {
+                for (final Path file : walk.filter(Files::isRegularFile)
+                    .filter(path -> path.toString().endsWith(JavaFiles.JAVA)).collect(
+                        Collectors.toList()
+                    )) {
+                    if (!expected.contains(file) && (!file.getFileName().toString().equals(
+                        "package-info.java"
+                    ) || !dirs.contains(file.getParent()))) {
+                        Files.delete(file);
+                        Logger.debug(this, "Deleted stale generated Java file %[file]s", file);
+                    }
+                }
+            }
+        }
     }
 
     /**
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
@@ -133,7 +133,6 @@ final class JavaFiles {
 
     /**
      * Delete generated Java files absent from the current XMIR collection.
-     *
      * @throws IOException If fails to inspect or remove a generated file
      */
     void removeStale() throws IOException {
@@ -151,8 +150,8 @@ final class JavaFiles {
                     .filter(path -> path.toString().endsWith(JavaFiles.JAVA)).collect(
                         Collectors.toList()
                     )) {
-                    if (!expected.contains(file) && (!file.getFileName().toString().equals(
-                        "package-info.java"
+                    if (!expected.contains(file) && (!"package-info.java".equals(
+                        file.getFileName().toString()
                     ) || !dirs.contains(file.getParent()))) {
                         Files.delete(file);
                         Logger.debug(this, "Deleted stale generated Java file %[file]s", file);
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
@@ -180,6 +180,7 @@ final class Transpiling implements Step {
     /**
      * Transpile a single tojo.
      * @param tojo Tojo that should be transpiled
+     * @param files Generated Java files
      * @return Number of generated Java files
      * @throws IOException If any issues with I/O
      */
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
@@ -151,13 +151,20 @@ final class Transpiling implements Step {
 
     @Override
     public void exec() throws IOException {
+        final JavaFiles files = new JavaFiles(
+            this.generatedDir,
+            this.cacheDir.resolve(Transpiling.CACHE).resolve(this.version()),
+            this.cacheEnabled
+        );
+        final int transpiled = new Threaded<>(
+            this.sources,
+            tojo -> this.transpiled(tojo, files)
+        ).total();
+        files.removeStale();
         Logger.info(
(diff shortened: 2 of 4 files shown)


eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -116,6 +131,37 @@ clazz, new FileGenerationReport(saved, tgt, target)
...
+            }
+            try (Stream<Path> walk = Files.walk(this.generated)) {
+                for (final Path file : walk.filter(Files::isRegularFile)
+                    .filter(path -> path.toString().endsWith(JavaFiles.JAVA)).collect(
+                        Collectors.toList()
+                    )) {
+                    if (!expected.contains(file) && (!file.getFileName().toString().equals(
+                        "package-info.java"
+                    ) || !dirs.contains(file.getParent()))) {
+                        Files.delete(file);
```
morphqdd, line 156: This deletes stale .java files but does not remove the directory they leave behind once it is empty (an EO package removed entirely leaves an empty directory under generated-sources). Unplacing.unplace() does the equivalent cleanup for a different kind of stale-file removal by calling EmptyDirectoriesIn(this.classes).clear() afterward; worth the same here.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -116,6 +131,37 @@ clazz, new FileGenerationReport(saved, tgt, target)
...
+                    dir = dir.getParent()) {
+                    dirs.add(dir);
+                }
+            }
+            try (Stream<Path> walk = Files.walk(this.generated)) {
+                for (final Path file : walk.filter(Files::isRegularFile)
+                    .filter(path -> path.toString().endsWith(JavaFiles.JAVA)).collect(
+                        Collectors.toList()
+                    )) {
+                    if (!expected.contains(file) && (!file.getFileName().toString().equals(
```
morphqdd: The package-info.java special case (spared only when its parent directory still has fresh output) is the most complex branch of this method, and it has no test covering it. JavaFilesTest's new test uses flat java-names with no package-info.java involved at all.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -116,6 +131,37 @@ clazz, new FileGenerationReport(saved, tgt, target)
         return saved.get();
     }
 
+    /**
+     * Delete generated Java files absent from the current XMIR collection.
+     *
+     * @throws IOException If fails to inspect or remove a generated file
+     */
+    void removeStale() throws IOException {
```
morphqdd, line 138: removeStale() reads this.fresh, which is only complete once every total() call for this run has returned. Nothing in the API enforces that order; calling removeStale() before all sources are processed would delete valid output for XMIRs that simply have not been transpiled yet this run.

eo-maven-plugin/src/test/java/org/eolang/maven/MjTranspileTest.java
```diff
@@ -444,6 +444,21 @@ void doesNotRetranspileIfNotModified(@Mktmp final Path temp) throws IOException
         );
     }
 
+    @Test
+    void removesJavaFilesForRemovedXmirSources(@Mktmp final Path temp) throws IOException {
```
morphqdd: This deletes the whole foreign catalog, simulating every source disappearing at once. The PR description's own motivating case is removing one EO object while others remain; JavaFilesTest covers that at the JavaFiles level, but there is no integration-level test here for the more realistic partial-removal case.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
```diff
@@ -152,13 +152,20 @@ final class Transpiling implements Step {
...
+        final JavaFiles files = new JavaFiles(
+            this.generatedDir,
+            this.cacheDir.resolve(Transpiling.CACHE).resolve(this.version()),
+            this.cacheEnabled
+        );
+        final int transpiled = new Threaded<>(
+            this.sources,
+            tojo -> this.transpiled(tojo, files)
+        ).total();
+        files.removeStale();
```
morphqdd, line 163: The number of stale files removeStale() deletes is only visible per-file at debug level (JavaFiles.java's Logger.debug call). The INFO-level summary here reports files transpiled and created but says nothing about how many were cleaned up.

Thayorns (commented):
Hoisting one `JavaFiles` out of the threaded loop so it can collect what was written, then reconciling once at the end, is a sound shape - and moving the instance out of `transpiled()` removes a per-tojo allocation as a bonus. My worry is the blast radius of `removeStale`: it deletes by absence rather than by ownership, and the directory it walks is not exclusively EO's.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -116,6 +131,37 @@ clazz, new FileGenerationReport(saved, tgt, target)
...
+        if (Files.exists(this.generated)) {
+            final Set<Path> expected = new HashSet<>(this.fresh);
+            final Set<Path> dirs = new HashSet<>();
+            for (final Path file : expected) {
+                for (Path dir = file.getParent(); dir != null && dir.startsWith(this.generated);
+                    dir = dir.getParent()) {
+                    dirs.add(dir);
+                }
+            }
+            try (Stream<Path> walk = Files.walk(this.generated)) {
```
Thayorns, line 148: This walks the whole tree under `generated` and deletes every `.java` that this run did not write. `MjSafe.java:139-144` defaults `eo.generatedDir` to `${project.build.directory}/generated-sources`, which is the shared Maven root that every code generator writes a subdirectory into - antlr4, jaxb, protobuf, build-helper. In a project that uses EO alongside any of them, the first transpile deletes their output, and the only thing standing between a user and that is remembering to point `eo.generatedDir` somewhere private. Ownership of a file is knowable here - EO wrote it, so EO can recognise it - and absence from `fresh` is not the same question.

eo-maven-plugin/src/main/java/org/eolang/maven/Transpiling.java
Thayorns, line 163: `this.sources` is `scopedTojos().withXmir()`, and `TjsForeign.select` at TjsForeign.java:223 filters on the scope attribute, so `fresh` only ever holds the paths for the current `eo.scope`. `generatedDir` carries no scope in it. Configure the plugin the way the scope parameter exists to be configured - one execution with `eo.scope=compile`, one with `eo.scope=test`, both writing to the default directory - and each execution now deletes everything the other one just generated.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaFilesTest.java
```diff
@@ -45,4 +45,36 @@ void countsJavaFilesItWrites(@Mktmp final Path temp) throws IOException {
...
+        initial.removeStale();
+        final JavaFiles current = new JavaFiles(generated, temp.resolve("cache"), false);
+        current.total(true, second, "", false);
+        current.removeStale();
+        MatcherAssert.assertThat(
+            "a Java file for an XMIR absent from the current transpilation must be deleted",
+            generated.resolve("org/eolang/EOfirst.java").toFile().exists(),
+            Matchers.is(false)
+        );
+        MatcherAssert.assertThat(
```
Thayorns: Two assertions, and the first is not the last statement, so the deletion check masks the survival check. The second one is the more interesting of the two - anyone can delete files, the question is whether the right ones survive - and right now it never runs when the first fails.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
```diff
@@ -116,6 +131,37 @@ clazz, new FileGenerationReport(saved, tgt, target)
...
+     * Delete generated Java files absent from the current XMIR collection.
+     *
+     * @throws IOException If fails to inspect or remove a generated file
+     */
+    void removeStale() throws IOException {
+        if (Files.exists(this.generated)) {
+            final Set<Path> expected = new HashSet<>(this.fresh);
+            final Set<Path> dirs = new HashSet<>();
+            for (final Path file : expected) {
+                for (Path dir = file.getParent(); dir != null && dir.startsWith(this.generated);
```
Thayorns, line 143: The opening bracket here neither closes on its line nor ends it: the `for` header runs into a second line mid-condition, and the same happens at line 150 where the stream chain is split across three lines inside the `for`. The `if` at 154 is folded across three lines too, with `"package-info.java"` sitting alone on 155. The method is doing three separable things - collecting the directories worth keeping, listing the candidates, deciding on each one - and the layout is what happens when they share one body.

yegor256:
@gemshrine conflicts here

gemshrine:

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (5) comments.  Your running score is [+3786](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+1143](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2159](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6402, opened 2026-08-05
#6375: escape string literals rendered in φ-terms

### LINKED ISSUE #6375: `PhApplication.φTerm()` doesn't properly escape strings with quotes, backslashes, and control characters
## What happens

`PhApplication` renders the standard `Φ.string` construction as a quoted value in `φTerm()`. It directly inserts decoded UTF-8 text between two quote characters without escaping it. A string containing `"`, `\`, a newline, or another control character therefore produces a φ-term that is not a valid EO string literal and cannot represent the original value unambiguously.

This affects runtime diagnostics and debugging exactly where `φTerm()` is expected to provide a readable EO-like representation.

## Problematic code

`eo-runtime/src/main/java/org/eolang/PhApplication.java`:

```java
if (literal && "Φ.string".equals(head)) {
    result = String.format(
        "\"%s\"",
        new String(PhApplication.bytes(data.group(1)), StandardCharsets.UTF_8)
    );
}
```

## Steps to reproduce

1. Construct `new Data.ToPhi("a\"b")` and call `φTerm()`.
2. Observe a result equivalent to `"a"b"`, which closes the literal in the middle.

A value containing a newline similarly yields a multi-line term, while a backslash loses its literal/escape distinction.

## Expected behavior

The quoted branch should escape EO string-literal syntax: at least backslash, quote, and supported control characters. For byte sequences that are not valid UTF-8, it should use an unambiguous non-string representation rather than silently emitting replacement characters in a misleading literal.

gemshrine:
Fixes #6375.

## What changed

- PhApplication now decodes string-literal bytes with strict UTF-8 validation.
- Valid strings escape backslash, quote, backspace, form feed, line feed, carriage return, and tab before being rendered as a quoted φ-term.
- Invalid UTF-8 no longer becomes a replacement-character string literal; it retains the unambiguous byte-based application representation.
- Added regressions for literal metacharacters and malformed UTF-8.

## Why

The prior rendering inserted decoded text directly between quotes. Quotes could terminate the literal, backslashes were ambiguous, and control characters produced invalid multi-line terms. The new renderer preserves the exact literal structure used in runtime diagnostics.

## Verification

- git diff --check
- eo-runtime main Java sources compiled successfully with one CPU and 1 GB heap.
- Full test compilation remains blocked by unrelated missing EOnop and EOnumber generated classes in the current local base.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhApplication.java
+++ b/eo-runtime/src/main/java/org/eolang/PhApplication.java
@@ -5,7 +5,11 @@
 
 package org.eolang;
 
+import java.nio.ByteBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.CodingErrorAction;
 import java.nio.charset.StandardCharsets;
+import java.util.Optional;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -80,15 +84,55 @@ public final class PhApplication extends PhOnce {
         final String body = PhApplication.body(binds);
         final Matcher data = PhApplication.DATA.matcher(body);
         final boolean literal = binds.length == 1 && data.find();
-        final String result;
+        final String string;
         if (literal && "Φ.string".equals(head)) {
-            result = String.format(
-                "\"%s\"", new String(PhApplication.bytes(data.group(1)), StandardCharsets.UTF_8)
-            );
+            string = PhApplication.string(PhApplication.bytes(data.group(1)));
+        } else {
+            string = null;
+        }
+        final String result;
+        if (string != null) {
+            result = string;
         } else if (literal && "Φ.number".equals(head)) {
             result = PhDefault.numeral(new BytesOf(PhApplication.bytes(data.group(1))).asNumber());
         } else {
             result = String.format("%s(%s)", head, body);
+        }
+        return result;
+    }
+
+    /**
+     * Render UTF-8 bytes as an EO string literal.
+     * @param bytes Bytes to render
+     * @return Escaped literal, or null if bytes are not valid UTF-8
+     */
+    private static String string(final byte[] bytes) {
+        Optional<String> text;
+        try {
+            text = Optional.of(
+                StandardCharsets.UTF_8.newDecoder()
+                    .onMalformedInput(CodingErrorAction.REPORT)
+                    .onUnmappableCharacter(CodingErrorAction.REPORT)
+                    .decode(ByteBuffer.wrap(bytes)).toString()
+            );
+        } catch (final CharacterCodingException ignored) {
+            text = Optional.empty();
+        }
+        final String result;
+        if (text.isPresent()) {
+            result = String.format(
+                "\"%s\"",
+                text.get()
+                    .replace("\\", "\\\\")
+                    .replace("\"", "\\\"")
+                    .replace("\b", "\\b")
+                    .replace("\f", "\\f")
+                    .replace(String.valueOf('\n'), "\\n")
+                    .replace(String.valueOf('\r'), "\\r")
+                    .replace("\t", "\\t")
+            );
+        } else {
+            result = null;
         }
         return result;
     }
--- a/eo-runtime/src/test/java/org/eolang/PhApplicationTest.java
+++ b/eo-runtime/src/test/java/org/eolang/PhApplicationTest.java
@@ -116,6 +116,38 @@ final class PhApplicationTest {
     }
 
     @Test
+    void escapesStringLiteralCharacters() {
+        MatcherAssert.assertThat(
+            "String φ-term must escape EO string literal characters",
+            new PhApplication(
+                new PhDispatch(Phi.Φ, "string"), 0,
+                new PhApplication(
+                    new PhDispatch(Phi.Φ, "bytes"), 0,
+                    new PhDefault(
+                        new byte[] {'a', '"', '\\', '\b', '\f', '\n', '\r', '\t'}
+                    )
+                )
+            ).φTerm(),
+            Matchers.equalTo("\"a\\\"\\\\\\b\\f\\n\\r\\t\"")
+        );
+    }
+
+    @Test
+    void rendersInvalidUtfAsBytes() {
+        MatcherAssert.assertThat(
+            "Invalid UTF-8 must not be rendered as a misleading string literal",
+            new PhApplication(
+                new PhDispatch(Phi.Φ, "string"), 0,
+                new PhApplication(
+                    new PhDispatch(Phi.Φ, "bytes"), 0,
+                    new PhDefault(new byte[] {(byte) 0xC3, (byte) 0x28})
+                )
+            ).φTerm(),
+            Matchers.equalTo("Φ.string(0->Φ.bytes(0->[D> C3-28]))")
+        );
+    }
+
+    @Test
     void rendersPositionalBindingAsTerm() {
         MatcherAssert.assertThat(
             "PhApplication must render positional binding in φ-term, but it didnt",


eo-runtime/src/main/java/org/eolang/PhApplication.java
```diff
@@ -93,6 +100,39 @@ private static String applied(final Phi phi, final Bind... binds) {
         return result;
     }
 
+    /**
+     * Render UTF-8 bytes as an EO string literal.
+     * @param bytes Bytes to render
+     * @return Escaped literal, or null if bytes are not valid UTF-8
+     */
+    private static String string(final byte[] bytes) {
```
morphqdd, line 109: Only 5 control characters get a named escape here (\b \f \n \r \t). Anything else outside printable ASCII - a raw NUL, the other C0 controls (U+0001-U+0007, U+000E-U+001F), DEL (U+007F) - passes straight through between the quotes unescaped. `VerboseBytesAsString.escaped()` in this same package already does this more completely: it escapes every character with `chr < 0x20 || chr > 0x7F` via `\uXXXX` instead of naming a handful. Worth reusing that approach (or at least matching its coverage) here, since this method exists for the same reason - rendering decoded bytes safely for a diagnostic string.

eo-runtime/src/main/java/org/eolang/PhApplication.java
```diff
@@ -93,6 +100,39 @@ private static String applied(final Phi phi, final Bind... binds) {
...
+        }
+        final String result;
+        if (text == null) {
+            result = null;
+        } else {
+            result = String.format(
+                "\"%s\"",
+                text.replace("\\", "\\\\")
+                    .replace("\"", "\\\"")
+                    .replace("\b", "\\b")
```
morphqdd, line 128: Because only the 5 named characters are escaped, a string literal containing a raw ESC (U+001B) followed by ANSI/OSC bytes reaches the term output untouched. If this term ever gets printed to a terminal (it's meant for runtime diagnostics), that's a live terminal-escape-injection path - the diagnostic itself can move the cursor, clear the screen, or worse depending on the terminal, controlled entirely by whatever produced the original string data.

eo-runtime/src/main/java/org/eolang/PhApplication.java
```diff
@@ -93,6 +100,39 @@ private static String applied(final Phi phi, final Bind... binds) {
...
+                    .replace("\f", "\\f")
```
morphqdd, line 129: Same gap, different consequence: a Unicode bidirectional override character (e.g. U+202E RIGHT-TO-LEFT OVERRIDE) also passes through unescaped, which is exactly the class of attack behind CVE-2021-42574 ("Trojan Source") - text that displays in a different order than it actually is. Anywhere this term gets shown (log file, terminal, IDE) inherits that risk for free.

eo-runtime/src/main/java/org/eolang/PhApplication.java
```diff
@@ -93,6 +100,39 @@ private static String applied(final Phi phi, final Bind... binds) {
...
+     * @return Escaped literal, or null if bytes are not valid UTF-8
+     */
+    private static String string(final byte[] bytes) {
+        String text;
+        try {
+            text = StandardCharsets.UTF_8.newDecoder()
+                .onMalformedInput(CodingErrorAction.REPORT)
+                .onUnmappableCharacter(CodingErrorAction.REPORT)
+                .decode(ByteBuffer.wrap(bytes)).toString();
+        } catch (final CharacterCodingException ignored) {
```
morphqdd, line 118: `ignored` discards whether the failure was malformed input or an unmappable character - the two `CodingErrorAction.REPORT` triggers this decoder can raise. Both currently collapse to the same `null` outcome, so if someone later wants to distinguish "not UTF-8 at all" from "valid UTF-8 encoding an unmappable character" for a better diagnostic, that information is already gone by the time it would be needed.

eo-runtime/src/test/java/org/eolang/PhApplicationTest.java
```diff
@@ -115,6 +115,30 @@ void rendersStringConstructionAsValue() {
...
+        MatcherAssert.assertThat(
+            "Invalid UTF-8 must not be rendered as a misleading string literal",
+            new PhApplication(
+                new PhDispatch(Phi.Φ, "string"), 0,
+                new PhApplication(
+                    new PhDispatch(Phi.Φ, "bytes"), 0,
+                    new PhDefault(new byte[] {(byte) 0xC3, (byte) 0x28})
+                )
+            ).φTerm(),
+            Matchers.equalTo("Φ.string(0->[D> C3-28])")
```
morphqdd: This exercises a malformed continuation byte (`0xC3, 0x28` - a lead byte with no valid follower). There's no test for the other `CharacterCodingException` trigger this method sets up specifically - `onUnmappableCharacter(REPORT)` - e.g. bytes that are structurally valid UTF-8 but encode a surrogate codepoint (U+D800-U+DFFF), which Java's decoder rejects as unmappable rather than malformed. Both paths collapse to the same `null` today, but they're different failure modes and only one is covered.

eo-runtime/src/main/java/org/eolang/PhApplication.java
```diff
@@ -93,6 +100,39 @@ private static String applied(final Phi phi, final Bind... binds) {
...
+            text = null;
+        }
+        final String result;
+        if (text == null) {
+            result = null;
+        } else {
+            result = String.format(
+                "\"%s\"",
+                text.replace("\\", "\\\\")
```
morphqdd: This is now a second, independent implementation of "escape decoded bytes for safe diagnostic display" in the same package, alongside `VerboseBytesAsString.escaped()`. The two already disagree on how much to escape (5 named characters here vs. the full non-printable-ASCII range there). Two copies of the same idea that can drift independently is exactly the kind of duplication that's easy to fix in one place and forget in the other.

eo-runtime/src/main/java/org/eolang/PhApplication.java
```diff
@@ -93,6 +100,39 @@ private static String applied(final Phi phi, final Bind... binds) {
...
+                    .replace("\"", "\\\"")
+                    .replace("\b", "\\b")
+                    .replace("\f", "\\f")
+                    .replace("\n", "\\n")
+                    .replace("\r", "\\r")
+                    .replace("\t", "\\t")
```
Thayorns, line 132: Seven characters get escaped, and the rest of the C0 range does not. A string holding `\u0000` or an ESC still lands raw between the quotes, so the φ-term carries invisible or terminal-controlling bytes and does not round-trip. EO has spellings for exactly these - `eo-parser/src/test/resources/org/eolang/parser/eo-syntax/escape-sequences.yaml` parses `"\000"` and `"\uFFFF"` alongside the seven handled here - so the term could stay printable for the whole range rather than for the characters that happen to have short forms.

eo-runtime/src/main/java/org/eolang/PhApplication.java
```diff
@@ -93,6 +100,39 @@ private static String applied(final Phi phi, final Bind... binds) {
...
+            text = StandardCharsets.UTF_8.newDecoder()
+                .onMalformedInput(CodingErrorAction.REPORT)
+                .onUnmappableCharacter(CodingErrorAction.REPORT)
+                .decode(ByteBuffer.wrap(bytes)).toString();
+        } catch (final CharacterCodingException ignored) {
+            text = null;
+        }
+        final String result;
+        if (text == null) {
+            result = null;
```
Thayorns: The method answers "this is not a decodable string" with null, and the caller has to grow around it: `applied` now declares `string`, assigns it null in an else branch that exists only to have something to test, and then opens a second `if` on `string != null` before the chain it already had. The decode failure is known at line 115 and could pick the byte rendering right there instead of travelling back out as a null. As it stands the javadoc has to warn about the null and every future caller of `string` has to read that warning.

Thayorns (commented):
Rendering the literal through a strict decoder is the right call - falling back to the byte form when the payload is not UTF-8 beats printing replacement characters that claim to be the string. The escaping itself is in the right order too, backslash before the rest. Two things: how the failure is signalled, and how much of the character range the escaping actually covers.

yegor256:
@gemshrine conflicts here

gemshrine:

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (6) comments.  Your running score is [+3793](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+1150](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2175](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6403, opened 2026-08-05
#6372: replace unplaced binaries even at equal length

### LINKED ISSUE #6372: `Placing` treats equal-length binaries as unchanged and leaves stale class contents
## What happens

`Placing.PlacedDependency.isNotAlreadyPlaced()` decides that an already placed binary is current when its target file has the same length as the file being resolved. It does not compare the catalogued hash or file contents.

A dependency can therefore change a `.class` file without changing its byte length (for example, a rebuilt SNAPSHOT or a locally supplied dependency). The plugin skips replacement and compilation uses the old bytecode even though resolution supplied new content.

## Problematic code

`eo-maven-plugin/src/main/java/org/eolang/maven/Placing.java`:

```java
if (tojo.isPresent()
    && Files.exists(target)
    && (this.sameLength(target, file) || !tojo.get().unplaced())
) {
    res = false;
}
```

`sameLength()` compares only `Files.size(first) == Files.size(second)`. The placed catalog already stores a `FileHash` when a class is placed, but this decision does not use it.

## Steps to reproduce

1. Place a dependency class into `target/classes`.
2. Replace the resolved source class with different bytes of exactly the same length.
3. Run the placing step without forcing overwrite.
4. Observe that the class is skipped and the old bytes remain in `target/classes`.

## Expected behavior

The plugin should consider a file current only when its content identity matches, not merely its size. It can use the existing placed-file hash or compare the source and target hashes before deciding to skip placement.

gemshrine:
Fixes #6372.

## What changed

- Placing no longer treats an unplaced target as current merely because the replacement binary has the same size.
- The regression now replaces old content with different content of exactly the same byte length after unplaceAll().

## Why

An unplaced catalog entry means the binary must be restored from the resolved dependency. The former size shortcut skipped that restoration when byte lengths matched, leaving stale class contents in target/classes. The placement decision now follows the catalog state rather than an insufficient length heuristic.

## Verification

MjPlaceTest passed in a single-core Maven run with JUnit parallel execution disabled.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Placing.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Placing.java
@@ -197,7 +197,7 @@ final class Placing implements Step {
             final boolean res;
             if (tojo.isPresent()
                 && Files.exists(target)
-                && (this.sameLength(target, file) || !tojo.get().unplaced())
+                && !tojo.get().unplaced()
             ) {
                 Logger.debug(
                     this,
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjPlaceTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjPlaceTest.java
@@ -200,12 +200,12 @@ final class MjPlaceTest {
     }
 
     @Test
-    void placesAgainIfWasUnplaced(@Mktmp final Path temp) throws Exception {
+    void placesAgainIfWasUnplacedWithSameLength(@Mktmp final Path temp) throws Exception {
         final FakeMaven maven = new FakeMaven(temp);
         final String binary = "some.class";
-        MjPlaceTest.saveBinary(temp, "with old content", binary);
+        MjPlaceTest.saveBinary(temp, "old content", binary);
         maven.execute(MjPlace.class).result();
-        final String updated = "with some new content";
+        final String updated = "new content";
         MjPlaceTest.saveBinary(temp, updated, binary);
         maven.placed().unplaceAll();
         maven.execute(MjPlace.class).result();


eo-maven-plugin/src/main/java/org/eolang/maven/Placing.java
```diff
@@ -197,7 +197,7 @@ private boolean isNotAlreadyPlaced(final Path file) {
             final boolean res;
             if (tojo.isPresent()
                 && Files.exists(target)
-                && (this.sameLength(target, file) || !tojo.get().unplaced())
+                && !tojo.get().unplaced()
```
morphqdd, line 200: After this change, a file can pass this filter (and go on to `placeBinary`) purely because `tojo.get().unplaced()` is true, even when `sameLength(target, file)` is true - that's the exact case this PR fixes. `printLogInfoAboutBinary`, called right after this filter in the same stream, still only logs its "was already placed... replacing" message when `!this.sameLength(target, file)` (line 231). So for the specific scenario this PR is about - same length, but genuinely unplaced - the file gets correctly replaced, but the debug log that's supposed to explain why stays silent.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2012](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (8) hits-of-code; +1 to give you at least something.  Your running score is [+3638](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6404, opened 2026-08-05
#6373: remove obsolete generated Java tests

### LINKED ISSUE #6373: `JavaPlaced` leaves obsolete generated tests after the XMIR no longer contains tests
## What happens

`JavaPlaced.exec()` writes a generated Java test only when `testsPresent(clazz)` finds `@Test` in the current XMIR. When the source test is later removed, renamed, or changes to a class without test annotations, the method simply does nothing. It never removes the test file produced by an earlier transpilation.

The stale file remains under `target/generated-test-sources` and Maven can compile or execute a test that no longer exists in the EO source. A clean build hides the problem, while an incremental build produces behavior different from a clean checkout.

## Problematic code

`eo-maven-plugin/src/main/java/org/eolang/maven/JavaPlaced.java:49-54`:

```java
if (clazz.element("java").text().isPresent()) {
    this.footprint.apply(Paths.get(""), this.target);
}
if (tests && JavaPlaced.testsPresent(clazz)) {
    this.placeJavaTests(clazz);
}
```

There is no complementary deletion for the expected `*Test.java`/`*EOAtomTest.java` path when `testsPresent()` becomes false.

## Steps to reproduce

1. Transpile an EO object whose XMIR contains a Java test with `@Test`; observe its generated `*Test.java`.
2. Remove that test from the EO source and transpile again without cleaning `target`.
3. Observe that the old generated test remains and is still visible to Maven.

## Expected behavior

Incremental transpilation should converge to the same generated-test tree as a clean build. When a class no longer provides tests, `JavaPlaced` should remove its previously generated test artifact (including the atom-collision variant when applicable).

gemshrine:
Fixes #6373.

## What changed

- JavaPlaced removes generated test sources when the current XMIR no longer contains tests.
- Cleanup covers both normal test names and the EO atom collision variant.
- When the collision status changes, the obsolete alternative test path is removed after generating the current one.
- Added focused regressions for ordinary and collision-based stale tests.

## Why

Incremental transpilation previously wrote a generated test when one existed, but left it behind after the source test was removed or renamed. Maven could then compile or execute a test absent from the current EO source. The generated-test tree now converges with the current XMIR.

## Verification

JavaPlacedTest passed in a single-core Maven run with JUnit parallel execution disabled.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/JavaPlaced.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/JavaPlaced.java
@@ -51,8 +51,12 @@ final class JavaPlaced implements BiProc<Xnav, Boolean> {
         if (clazz.element("java").text().isPresent()) {
             this.footprint.apply(Paths.get(""), this.target);
         }
-        if (tests && JavaPlaced.testsPresent(clazz)) {
-            this.placeJavaTests(clazz);
+        if (tests) {
+            if (JavaPlaced.testsPresent(clazz)) {
+                this.placeJavaTests(clazz);
+            } else {
+                this.removeJavaTests(clazz);
+            }
         }
     }
 
@@ -94,6 +98,38 @@ final class JavaPlaced implements BiProc<Xnav, Boolean> {
             content = clazz.element("tests").text().get();
         }
         new Saved(content, resulted).value();
+        this.removeJavaTests(clazz, resulted);
+    }
+
+    /**
+     * Remove Java tests that are no longer generated.
+     * @param clazz Transpiled class
+     * @throws IOException If I/O fails
+     */
+    private void removeJavaTests(final Xnav clazz) throws IOException {
+        this.removeJavaTests(clazz, null);
+    }
+
+    /**
+     * Remove Java tests except the one currently generated.
+     * @param clazz Transpiled class
+     * @param retained Generated test to retain, if any
+     * @throws IOException If I/O fails
+     */
+    private void removeJavaTests(final Xnav clazz, final Path retained) throws IOException {
+        final String[] parts = clazz.attribute("java-name").text().get().split("\\.");
+        final Path base = Arrays.stream(parts, 0, parts.length - 1).reduce(
+            this.generated.getParent().resolve("generated-test-sources"),
+            Path::resolve,
+            Path::resolve
+        );
+        final String name = parts[parts.length - 1];
+        for (final String suffix : new String[]{"Test.java", "EOAtomTest.java"}) {
+            final Path test = base.resolve(String.format("%s%s", name, suffix));
+            if (!test.equals(retained)) {
+                Files.deleteIfExists(test);
+            }
+        }
     }
 
     /**
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/JavaPlacedTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/JavaPlacedTest.java
@@ -7,6 +7,7 @@ package org.eolang.maven;
 import com.github.lombrozo.xnav.Xnav;
 import com.yegor256.Mktmp;
 import com.yegor256.MktmpResolver;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.cactoos.text.TextOf;
@@ -103,5 +104,54 @@ final class JavaPlacedTest {
             ).asString(),
             Matchers.equalTo(expected)
         );
+    }
+
+    @Test
+    void removesObsoleteJavaCompanions(@Mktmp final Path temp) throws Exception {
+        final Path target = temp.resolve("target");
+        final Path generated = target.resolve("generated-sources");
+        final Path utest = target.resolve("FooTest.java");
+        final JavaPlaced placed = new JavaPlaced(
+            new FpJavaGenerated(this.clazz("@Test"), generated, utest), utest, generated
+        );
+        placed.exec(this.clazz("@Test"), true);
+        final Path test = target.resolve("generated-test-sources").resolve("FooTest.java");
+        final boolean created = Files.exists(test);
+        placed.exec(this.clazz(""), true);
+        MatcherAssert.assertThat(
+            "Obsolete Java test was not removed", created && Files.notExists(test)
+        );
+    }
+
+    @Test
+    void removesObsoleteAtomJavaCompanions(@Mktmp final Path temp) throws Exception {
+        final Path target = temp.resolve("target");
+        final Path generated = target.resolve("generated-sources");
+        final Path utest = target.resolve("FooTest.java");
+        final JavaPlaced placed = new JavaPlaced(
+            new FpJavaGenerated(this.clazz("@Test"), generated, utest), utest, generated
+        );
+        Files.createDirectories(temp.resolve("src/test/java"));
+        new Saved("", temp.resolve("src/test/java/FooTest.java")).value();
+        placed.exec(this.clazz("@Test"), true);
+        final Path atom = target.resolve("generated-test-sources").resolve("FooEOAtomTest.java");
+        final boolean created = Files.exists(atom);
+        placed.exec(this.clazz(""), true);
+        MatcherAssert.assertThat(
+            "Obsolete atom Java test was not removed", created && Files.notExists(atom)
+        );
+    }
+
+    /**
+     * Create transpiled class.
+     * @param tests Test code
+     * @return Class XML
+     */
+    private Xnav clazz(final String tests) throws Exception {
+        return new Xnav(
+            new Xembler(
+                new Directives().add("class").attr("java-name", "Foo").add("tests").set(tests)
+            ).xml()
+        ).element("class");
     }
 }


eo-maven-plugin/src/main/java/org/eolang/maven/JavaPlaced.java
```diff
@@ -93,6 +97,38 @@ private void placeJavaTests(final Xnav clazz) throws IOException {
...
+
+    /**
+     * Remove Java tests except the one currently generated.
+     * @param clazz Transpiled class
+     * @param retained Generated test to retain, if any
+     * @throws IOException If I/O fails
+     */
+    private void removeJavaTests(final Xnav clazz, final Path retained) throws IOException {
+        final String[] parts = clazz.attribute("java-name").text().get().split("\\.");
+        final Path base = Arrays.stream(parts, 0, parts.length - 1).reduce(
```
morphqdd, line 121: This recomputes the same `base` path as `placeJavaTests` a few lines up (same `Arrays.stream(...).reduce(tests, Path::resolve, Path::resolve)` over the java-name segments), just with the generated-test-sources root inlined instead of held in a local. If the two ever drift apart, `retained` (built by the other copy) would stop matching either candidate `test` path here, and the file just written would get deleted right after being written.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaPlaced.java
```diff
@@ -93,6 +97,38 @@ private void placeJavaTests(final Xnav clazz) throws IOException {
             content = clazz.element("tests").text().get();
         }
         new Saved(content, resulted).value();
+        this.removeJavaTests(clazz, resulted);
```
morphqdd, line 101: The PR description says "when the collision status changes, the obsolete alternative test path is removed after generating the current one" - that's what this line does. Neither new test exercises it though: both go from tests-present to tests-absent. There's no test that keeps tests present across two runs while the atom-collision status flips (e.g. `FooTest.java` generated first, then a same-named hand-written test appears and the next run produces `FooEOAtomTest.java` instead), which is the specific scenario this line is for.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaPlaced.java
```diff
@@ -50,8 +50,12 @@ public void exec(final Xnav clazz, final Boolean tests) throws IOException {
         if (clazz.element("java").text().isPresent()) {
             this.footprint.apply(Paths.get(""), this.target);
         }
-        if (tests && JavaPlaced.testsPresent(clazz)) {
-            this.placeJavaTests(clazz);
+        if (tests) {
```
morphqdd, line 54: Cleanup only runs when `tests` is true. If a project has `transpileTests` enabled for one build (generating test files here) and then disabled for a later build, `exec()` skips this whole branch and the previously generated test files are never removed - the exact staleness this PR is otherwise fixing.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaPlacedTest.java
```diff
@@ -74,4 +75,55 @@ void placesJavaTests(@Mktmp final Path temp) throws Exception {
...
+        final Path utest = target.resolve("FooTest.java");
+        final JavaPlaced placed = new JavaPlaced(
+            new FpJavaGenerated(this.clazz("@Test"), generated, utest), utest, generated
+        );
+        placed.exec(this.clazz("@Test"), true);
+        final Path test = target.resolve("generated-test-sources").resolve("FooTest.java");
+        final boolean created = Files.exists(test);
+        placed.exec(this.clazz(""), true);
+        MatcherAssert.assertThat(
+            "Obsolete Java test was not removed", created && Files.notExists(test)
```
morphqdd, line 122: Combining `created && Files.notExists(test)` into one boolean loses which half failed - if this ever regresses, the failure message just says the boolean was false, not whether the file was never created in the first place or never removed. Two separate assertions (or a `Matchers.allOf`) would keep the two claims distinguishable in a failure report.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaPlaced.java
```diff
@@ -93,6 +97,38 @@ private void placeJavaTests(final Xnav clazz) throws IOException {
...
+    /**
+     * Remove Java tests except the one currently generated.
+     * @param clazz Transpiled class
+     * @param retained Generated test to retain, if any
+     * @throws IOException If I/O fails
+     */
+    private void removeJavaTests(final Xnav clazz, final Path retained) throws IOException {
+        final String[] parts = clazz.attribute("java-name").text().get().split("\\.");
+        final Path base = Arrays.stream(parts, 0, parts.length - 1).reduce(
+            this.generated.getParent().resolve("generated-test-sources"),
```
Thayorns, line 122: This is `placeJavaTests` lines 68-76 written out a second time: the same `split("\\.")`, the same `getParent().resolve("generated-test-sources")`, the same reduce over the package parts. Line 126 duplicates the two file-name formats from lines 77 and 89 as well. Whoever changes where generated tests land, or what the atom variant is called, has to find both copies; miss the one here and the deletion silently stops matching anything, which is the exact bug this is fixing. Only `placeJavaTests` currently knows how to name these files, and it is the one that computed `resulted` you are passing back in.

eo-maven-plugin/src/main/java/org/eolang/maven/JavaPlaced.java
```diff
@@ -93,6 +97,38 @@ private void placeJavaTests(final Xnav clazz) throws IOException {
...
+        this.removeJavaTests(clazz, resulted);
+    }
+
+    /**
+     * Remove Java tests that are no longer generated.
+     * @param clazz Transpiled class
+     * @throws IOException If I/O fails
+     */
+    private void removeJavaTests(final Xnav clazz) throws IOException {
+        this.removeJavaTests(clazz, null);
```
Thayorns, line 110: This overload exists only to pass null, and the receiving end never says so - `!test.equals(retained)` happens to do the right thing because `Path.equals(null)` is false, so the null works by accident rather than by contract. Someone tightening that comparison later, or reaching for `retained.equals(test)` because it reads more naturally, gets an NPE on the branch that has no test written.

eo-maven-plugin/src/test/java/org/eolang/maven/JavaPlacedTest.java
Thayorns, line 122: `created && Files.notExists(test)` folds two separate facts behind one message, and only one of them is what the test is named for. If the first `exec` writes `FooEOAtomTest.java` instead - which it does whenever `temp/src/test/java/FooTest.java` happens to exist, and the sibling test at line 113 sets exactly that up - then `created` is false and the failure says the obsolete test was not removed, when nothing was ever created to remove.

Thayorns (commented):
Converging the generated test tree with the current XMIR is worth doing, and handling the atom-collision name alongside the plain one is the part that would have been easy to miss. What bothers me is that the removal side now re-derives, by hand, everything the writing side already knows.

yegor256:
@gemshrine conflicts here

gemshrine:

eo-maven-plugin/src/test/java/org/eolang/maven/JavaPlacedTest.java
```diff
@@ -104,4 +105,55 @@ void placesClassMarkedOnlyWithParameterized(@Mktmp final Path temp) throws Excep
...
+        );
+    }
+
+    @Test
+    @SuppressWarnings("JTCOP.RuleNotContainsTestWord")
```
yegor256: @gemshrine better fix, don't suppress

gemshrine: Renamed both methods to drop the `Test` word (`removesObsoleteJavaCompanions` and `removesObsoleteAtomJavaCompanions`) and removed the suppression.

0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2060](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+3896](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+1220](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2939](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6405, opened 2026-08-05
#6374: preserve raw bytes in octal string escapes

### LINKED ISSUE #6374: Parser violates one-byte contract for octal escapes `\200` through `\377`
## What happens

The parser documents `\NNN` as a one-byte octal escape and explicitly limits it to `0o377` (255). `Emissions.appendOctal()` appends the numeric value as a UTF-16 character, then the enclosing string is encoded as UTF-8 by `new Hex(unescaped)`.

For octal values `0o200` through `0o377`, UTF-8 uses two bytes. Thus `"\377"` becomes `C3-BF`, not the single byte `FF`. The current range check prevents values above 255 after #5419, but it does not make the accepted high-byte values one-byte escapes.

## Problematic code

`eo-parser/src/main/java/org/eolang/parser/Emissions.java`:

```java
private static final int MAX_OCTAL_BYTE = 0xFF;
// ...
out.append((char) value);
```

Later, `Emissions.string()` calls `new Hex(unescaped)`, and `Hex(String)` uses `text.getBytes(StandardCharsets.UTF_8)`.

## Steps to reproduce

1. Parse an EO string literal containing `"\377"`.
2. Inspect the emitted bytes carrier.
3. It contains `C3-BF`, two UTF-8 bytes for U+00FF, instead of `FF`.

The same mismatch occurs for every accepted octal value from `\200` to `\377`.

## Expected behavior

The implementation and documented contract must agree. If octal escapes are byte escapes, the parser should append their raw byte values before UTF-8 text encoding. If they are Unicode character escapes, the one-byte limit and diagnostic should be changed, and values above 127 must be allowed consistently.

gemshrine:
Fixes #6374.

## What changed

- Octal escapes now append their raw byte value instead of first converting it to a UTF-16 character and then UTF-8 encoding it.
- Normal text and Unicode escapes continue to use UTF-8 encoding.
- The behavior is applied consistently to ordinary string literals and text blocks.
- Added regressions proving that octal \377 produces the single byte FF.

## Why

The parser accepted octal values through 0o377 as one-byte escapes but encoded values above 0o177 as two UTF-8 bytes. Keeping octal input on the byte path restores the documented one-byte contract.

## Verification

EoSyntaxTest and LnTextBlockTest passed in a single-core Maven run; the focused jtcop check also passed.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -266,6 +266,15 @@ final class Emissions {
     }
 
     /**
+     * Decode a string body to its raw byte representation.
+     * @param inner Source body without surrounding quotes
+     * @return Decoded bytes
+     */
+    static byte[] unescapeBytes(final String inner) {
+        return Emissions.unescapeRawBytes(inner);
+    }
+
+    /**
      * Open the {@code <o>} for a value that carries no data of its own —
      * a star, a root or self token, a term, a group, or a plain base
      * reference — where the kind alone picks the {@code base}.
@@ -542,7 +551,7 @@ final class Emissions {
      * @param inner Source body without surrounding quotes
      * @return Decoded bytes
      */
-    static byte[] unescapeBytes(final String inner) {
+    private static byte[] unescapeRawBytes(final String inner) {
         final ByteArrayOutputStream out = new ByteArrayOutputStream(inner.length());
         final StringBuilder text = new StringBuilder(inner.length());
         int idx = 0;
@@ -557,24 +566,7 @@ final class Emissions {
             if (next == 'u') {
                 idx = Emissions.appendUnicode(text, inner, idx + 1);
             } else if (next >= '0' && next <= '7') {
-                Emissions.appendText(out, text);
-                int cursor = idx + 1;
-                int value = 0;
-                while (cursor < inner.length() && cursor < idx + 4
-                    && inner.charAt(cursor) >= '0' && inner.charAt(cursor) <= '7') {
-                    value = value * 8 + inner.charAt(cursor) - '0';
-                    cursor = cursor + 1;
-                }
-                if (value > Emissions.MAX_OCTAL_BYTE) {
-                    throw new NumberFormatException(
-                        String.format(
-                            "octal escape \\%s is out of range: value %d exceeds the 1-byte limit of 0o377 (255)",
-                            inner.substring(idx + 1, cursor), value
-                        )
-                    );
-                }
-                out.write(value);
-                idx = cursor;
+                idx = Emissions.rawOctal(out, text, inner, idx + 1);
             } else {
                 text.append(Emissions.singleCharEscape(glyph, next));
                 idx = idx + 2;
@@ -582,6 +574,39 @@ final class Emissions {
         }
         Emissions.appendText(out, text);
         return out.toByteArray();
+    }
+
+    /**
+     * Append one octal escape as a raw byte.
+     * @param out Output bytes
+     * @param text Text waiting for encoding
+     * @param body Whole string body
+     * @param start First octal digit
+     * @return Index past the octal escape
+     * @checkstyle ParameterNumberCheck (3 lines)
+     */
+    private static int rawOctal(
+        final ByteArrayOutputStream out, final StringBuilder text,
+        final String body, final int start
+    ) {
+        Emissions.appendText(out, text);
+        int cursor = start;
+        int value = 0;
+        while (cursor < body.length() && cursor < start + 3
+            && body.charAt(cursor) >= '0' && body.charAt(cursor) <= '7') {
+            value = value * 8 + body.charAt(cursor) - '0';
+            cursor = cursor + 1;
+        }
+        if (value > Emissions.MAX_OCTAL_BYTE) {
+            throw new NumberFormatException(
+                String.format(
+                    "octal escape \\%s is out of range: value %d exceeds the 1-byte limit of 0o377 (255)",
+                    body.substring(start, cursor), value
+                )
+            );
+        }
+        out.write(value);
+        return cursor;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


eo-parser/src/main/java/org/eolang/parser/Emissions.java
```diff
@@ -532,14 +536,66 @@ private static String rootBase(final String raw) {
...
-     * @return Decoded text
+     * Decode a string body to its byte representation. Text and Unicode
+     * escapes are UTF-8 encoded, while octal escapes contribute their raw
+     * one-byte values.
+     * @param inner Source body without surrounding quotes
+     * @return Decoded bytes
      */
-    private static String unescape(final String raw) {
-        return Emissions.unescapeBody(raw.substring(1, raw.length() - 1));
+    static byte[] unescapeBytes(final String inner) {
```
morphqdd: unescapeBody, a few dozen lines above this method, and its helper appendOctal are no longer called from anywhere (both former call sites, this class's own string() and LnTextBlock.into(), now use unescapeBytes). appendOctal's out.append((char) value) is exactly the UTF-16-then-UTF-8 path this PR fixes, so the old bug is still sitting in the file, just unreachable. It is package-private rather than private, so PMD's UnusedPrivateMethod check will not flag it either. Worth deleting both, or a future edit could reintroduce the bug by calling back into them.

eo-parser/src/main/java/org/eolang/parser/Emissions.java
```diff
@@ -532,14 +536,66 @@ private static String rootBase(final String raw) {
...
+            if (glyph != '\\' || idx + 1 >= inner.length()) {
+                text.append(glyph);
+                idx = idx + 1;
+                continue;
+            }
+            final char next = inner.charAt(idx + 1);
+            if (next == 'u') {
+                idx = Emissions.appendUnicode(text, inner, idx + 1);
+            } else if (next >= '0' && next <= '7') {
+                Emissions.appendText(out, text);
```
morphqdd: No test exercises an octal escape with regular text on both sides (e.g. "ab\\377cd"), which is the path that actually depends on appendText flushing the buffer in the right order relative to the raw byte write. Both new tests here and in the yaml pack use a bare \\377 with nothing around it.

eo-parser/src/main/java/org/eolang/parser/Emissions.java
```diff
@@ -532,14 +536,66 @@ private static String rootBase(final String raw) {
...
+                if (value > Emissions.MAX_OCTAL_BYTE) {
+                    throw new NumberFormatException(
+                        String.format(
+                            "octal escape \\%s is out of range: value %d exceeds the 1-byte limit of 0o377 (255)",
+                            inner.substring(idx + 1, cursor), value
+                        )
+                    );
+                }
+                out.write(value);
+                idx = cursor;
```
morphqdd: No test for two or more consecutive octal escapes (e.g. \\101\\102 producing the bytes for "AB"), only a single escape is covered anywhere in this PR.

eo-parser/src/main/java/org/eolang/parser/Emissions.java
Thayorns: After this change nothing calls `unescapeBody` at line 244 any more - its two callers were `unescape`, which this removes, and `LnTextBlock:64`, which now calls this method. `appendOctal` at line 610 goes with it, since `unescapeBody` was its only caller. That leaves roughly fifty lines of unreachable escape decoding sitting next to the live one, including a second copy of the range check and its message, word for word identical to the one at line 568. The next person to fix an escape bug has two plausible-looking implementations to choose between and no way to tell from the code which one runs.

eo-parser/src/main/java/org/eolang/parser/Emissions.java
```diff
@@ -353,9 +355,11 @@ private static void string(
         final Emit emit, final String name, final Value value, final int line
     ) {
         emit.object(name, "Φ.string", line, value.pos());
-        final String unescaped;
+        final byte[] unescaped;
         try {
-            unescaped = Emissions.unescape(value.raw());
+            unescaped = Emissions.unescapeBytes(
+                value.raw().substring(1, value.raw().length() - 1)
```
Thayorns, line 370: Stripping the quotes used to be the decoder's business - `unescape` took the raw form and knew it had quotes around it. Now `string` does that itself, calls `value.raw()` twice to do it, and `LnTextBlock` calls the same method with a body that never had quotes to begin with. Both callers now have to remember which shape `unescapeBytes` wants, and nothing in its signature says.

Thayorns (commented):
Keeping the octal escape on the byte path instead of routing it through a UTF-16 char is the right fix, and doing it by flushing the pending text as UTF-8 whenever a raw byte arrives is a clean way to get there. Covering both the string literal and the text block, with a yaml fixture for one and a unit test for the other, is the right pair of proofs. Two remarks about what the old path left behind.

0crat:
@morphqdd Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+3800](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+1157](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (131 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2187](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6406, opened 2026-08-05
#6363: validate eo.coverageFile path inside the controlled error boundary

### LINKED ISSUE #6363: `PhCoverage` leaks `InvalidPathException` when `eo.coverageFile` contains an invalid path
## What happens

Coverage recording is enabled by the JVM property `eo.coverageFile`. `PhCoverage.target()` reads that value and passes it directly to `Paths.get(path)`, outside the error boundary that wraps the subsequent file write.

For example, on the default Unix file system, setting the property to a string containing a NUL character and then touching any coverage-wrapped object throws raw `InvalidPathException`. The EO program cannot receive the normal failure that `PhCoverage.hit()` uses for an I/O problem.

## Root cause

`eo-runtime/src/main/java/org/eolang/PhCoverage.java` only catches `IOException` around `Files.write`. `Paths.get(path)` can throw `InvalidPathException` before that block. The value comes from a user-controlled system property.

## Expected behavior

Validate and convert the configured path inside the same controlled error boundary. An invalid `eo.coverageFile` value should report a descriptive runtime failure that names the property and its value, instead of leaking a raw JDK exception on the first object access.

morphqdd:
Fixes #6363.

`PhCoverage.hit()` called `Paths.get(path)` outside the try block that catches `IOException`. An invalid `eo.coverageFile` value (for example a path with a NUL character) threw a raw `InvalidPathException` instead of a descriptive failure.

The path conversion now happens inside the same controlled error boundary as the file write. An invalid value throws `IllegalArgumentException` naming the property and its value.

Added a test that fails on the unpatched code with a raw `InvalidPathException` and passes with the fix, checking both the exact exception class and the message content.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhCoverage.java
+++ b/eo-runtime/src/main/java/org/eolang/PhCoverage.java
@@ -9,7 +9,7 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
-import java.nio.file.Path;
+import java.nio.file.InvalidPathException;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3473](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6408, opened 2026-08-05
#6367: probe class presence without initializing it

### LINKED ISSUE #6367: `OnClasspath` initializes a class while merely probing whether it exists
## What happens

`OnClasspath.probe()` implements a presence check with `Class.forName(cls)`. The one-argument overload initializes the class, so resolving a missing EO attribute can execute the target class's static initializer. If that initializer has side effects or throws an `ExceptionInInitializerError`/other linkage failure, the attribute-presence probe leaks that failure instead of returning its boolean answer.

The helper is explicitly documented as a cached check of whether a Java class is present on the classpath; it is called from package-extension dispatch on attribute misses. Presence probing should not run application code.

## Problematic code

`eo-runtime/src/main/java/org/eolang/OnClasspath.java:67-68`:

```java
try {
    Class.forName(cls);
    found = true;
}
```

## Steps to reproduce

1. Put a class with a static initializer that records a side effect or throws into the runtime classpath.
2. Invoke `OnClasspath.has()` for its fully-qualified name through package-extension dispatch.
3. Observe that the initializer runs (or its failure escapes), although the caller asked only whether the class exists.

## Expected behavior

The probe should load class metadata without initialization, for example via `Class.forName(cls, false, loader)`, and return a cached presence result. Failures that mean the class cannot be linked should be handled deliberately rather than executing a static initializer as a side effect of an attribute miss.

morphqdd:
Fixes #6367.

`OnClasspath.probe()` used the one-argument `Class.forName(cls)`, which initializes the class. A presence check from package-extension dispatch could run the target class's static initializer, executing arbitrary side effects or leaking a linkage error, on a plain attribute miss.

The probe now uses `Class.forName(cls, false, loader)`, which loads class metadata without running static initializers.

Added a test with a class whose static initializer sets a flag, checked via a separate holder class so reading the flag after the probe does not itself trigger initialization.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/OnClasspath.java
+++ b/eo-runtime/src/main/java/org/eolang/OnClasspath.java
@@ -6,6 +6,7 @@
 package org.eolang;
 
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3494](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6409, opened 2026-08-05
#6371: preserve the IOException contract through OyCached

### LINKED ISSUE #6371: `OyCached` converts recoverable `IOException` failures into unchecked exceptions
## What happens

`Objectionary.get()` and `Objectionary.isDirectory()` declare `IOException`, allowing Maven steps to handle retrieval failures as I/O failures. `OyCached` uses `ConcurrentHashMap.computeIfAbsent()` for caching, whose lambda cannot throw checked exceptions. It catches an `IOException` from the wrapped objectionary and rethrows `IllegalStateException`.

As a result, a transient remote/cache read failure changes its public error type solely because caching is enabled. Callers that correctly handle the declared `IOException` cannot recover or add the normal Maven diagnostic; they receive an unrelated unchecked exception instead.

## Problematic code

`eo-maven-plugin/src/main/java/org/eolang/maven/OyCached.java` contains the same pattern in both cache loaders:

```java
try {
    return this.origin.get(name);
} catch (final IOException exception) {
        "An error occurred during the access to the origin objectionary",
        exception
    );
}
```

## Steps to reproduce

1. Wrap an `Objectionary` implementation whose `get()` or `isDirectory()` throws `IOException`.
2. Call the corresponding method through `new OyCached(origin)`.
3. Observe `IllegalStateException` rather than the declared `IOException`.

## Expected behavior

Caching must preserve the `Objectionary` error contract: callers should receive an `IOException` when the origin does. The implementation may use an internal unchecked carrier inside `computeIfAbsent()`, but it should unwrap and rethrow the original `IOException` at the public method boundary.

morphqdd:
Fixes #6371.

`OyCached.get()` and `isDirectory()` use `ConcurrentHashMap.computeIfAbsent()`, whose lambda cannot declare checked exceptions. Both caught the origin's `IOException` and rethrew `IllegalStateException`, so a caller handling the declared `IOException` got an unrelated unchecked exception instead.

The lambda now wraps the `IOException` in a private unchecked carrier, and the public method unwraps and rethrows the original `IOException` at the boundary, preserving the declared contract.

Added tests with a fake origin objectionary that always throws a given `IOException`, checking that the exact same instance reaches the caller through both `get()` and `isDirectory()`.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/OyCached.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/OyCached.java
@@ -5,6 +5,7 @@
 package org.eolang.maven;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import org.cactoos.Input;
@@ -68,12 +69,14 @@ final class OyCached implements Objectionary {
                     try {
                         return this.origin.get(name);
                     } catch (final IOException exception) {
-                        throw new OyCached.Uncached(exception);
+                        throw new UncheckedIOException(exception);
                     }
                 }
             );
-        } catch (final OyCached.Uncached wrap) {
-            throw wrap.origin();
+        } catch (final UncheckedIOException wrap) {
+            throw new IOException(
+                String.format("Failed to fetch '%s' from the origin objectionary", name), wrap
+            );
         }
     }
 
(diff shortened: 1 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (133 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3506](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6410, opened 2026-08-05
#6368: keep the logging wrapper across copy() and normalized()

### LINKED ISSUE #6368: `PhLogged` drops its logging wrapper after `copy()` and `normalized()`
## What happens

`PhLogged` promises to report all manipulations with the wrapped `Phi` for debugging. It logs the call to `copy()`, but returns the raw result of `origin.copy()`. Its `normalized()` method likewise returns `origin.normalized()` directly.

Consequently, the first copy or normalization silently removes the decorator. Every operation on the returned object is no longer logged, although it remains part of the computation that the caller chose to trace. This is especially misleading because normal EO application frequently copies objects before filling their attributes.

## Problematic code

`eo-runtime/src/main/java/org/eolang/PhLogged.java`:

```java
@Override
public Phi copy() {
    System.out.printf("%d.copy()...%n", this.hashCode());
    final Phi ret = this.origin.copy();
    System.out.printf("%d.copy()! -> %d%n", this.hashCode(), ret.hashCode());
    return ret;
}

@Override
public Phi normalized() {
    return this.origin.normalized();
}
```

## Steps to reproduce

1. Wrap a `Phi` in `new PhLogged(origin)`.
2. Call `copy()` (or `normalized()`) and retain its result.
3. Call a normally logged operation such as `take()` or `delta()` on that result.
4. Observe no `PhLogged` trace for the latter operation.

## Expected behavior

The result of `copy()` and `normalized()` should remain a `PhLogged` wrapper around the corresponding result, so tracing continues for the object returned by these `Phi` operations.

morphqdd:
Fixes #6368.

`PhLogged.copy()` and `normalized()` returned the raw result of the origin call instead of a new `PhLogged` around it. The first copy or normalization silently dropped the decorator, so every later operation on the returned object went unlogged even though it remained part of the traced computation. This is especially misleading since normal EO application frequently copies objects before filling their attributes.

Both methods now wrap their result in a new `PhLogged`, so tracing continues.

Added tests checking that `copy()` and `normalized()` return a `PhLogged` instance.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhLogged.java
+++ b/eo-runtime/src/main/java/org/eolang/PhLogged.java
@@ -32,7 +32,7 @@ public final class PhLogged implements Phi {
     @Override
     public Phi copy() {
         System.out.printf("%d.copy()...%n", this.hashCode());
-        final Phi ret = this.origin.copy();
+        final Phi ret = new PhLogged(this.origin.copy());
         System.out.printf("%d.copy()! -> %d%n", this.hashCode(), ret.hashCode());
         return ret;
     }
@@ -97,7 +97,7 @@ public final class PhLogged implements Phi {
 
     @Override
     public Phi normalized() {
-        return this.origin.normalized();
+        return new PhLogged(this.origin.normalized());
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3518](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6411, opened 2026-08-05
#6370: complete the root package for a top-level probe

### LINKED ISSUE #6370: `Probing` does not pull root-package siblings for a top-level `+probe` object
## What happens

`Probing.complete()` was introduced to pull every sibling of a discovered object before parsing, so bare sibling references can resolve (see #6175). It only calls `Objectionary.children()` when the object name contains a dot:

```java
final int split = object.lastIndexOf('.');
if (split > 0) {
    final String pkg = object.substring(0, split);
    // pull children(pkg)
}
```

For a top-level object such as `foo`, `lastIndexOf('.')` is `-1`, so the root package is never completed. This leaves a project that probes `foo` without its top-level siblings even though `ObjectsIndex.children("")` explicitly supports the root package. If `foo` uses a bare top-level sibling such as `bar`, parsing may fail because `bar` was not pulled.

## Steps to reproduce

1. Add `+probe foo` to a project.
2. Let `foo` refer to another top-level object, `bar`, by its bare name.
3. Run the probe/pull pipeline.
4. Observe that `foo` is registered but `bar` is not registered as a sibling, unlike siblings of `package.foo`.

## Expected behavior

When a discovered object has no dot, `Probing.complete()` should complete the root package through `children("")`. It should retain the existing per-package de-duplication while treating the empty root package as a valid package key.

morphqdd:
Fixes #6370.

`Probing.complete()` only pulled the remaining siblings of a probed object's package when the name contained a dot (`split > 0`). For a top-level object such as `foo`, `lastIndexOf('.')` is `-1`, so the root package was never completed, even though `ObjectsIndex.children("")` explicitly supports listing the root package's own children. A project that probes a top-level object referencing another top-level object by its bare name could fail to resolve that sibling during parsing.

The package prefix is now computed with `substring(0, Math.max(lastIndexOf('.'), 0))`, so a name with no dot resolves to the empty root package instead of skipping completion, while keeping the existing per-package de-duplication.

Added a test that probes a top-level object and checks a root-level sibling gets pulled in.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Probing.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Probing.java
@@ -143,13 +143,10 @@ final class Probing implements Step {
         final Path src,
         final Set<String> completed
     ) throws IOException {
-        final int split = object.lastIndexOf('.');
-        if (split > 0) {
-            final String pkg = object.substring(0, split);
-            if (completed.add(pkg)) {
-                for (final String sibling : this.objectionary.children(pkg)) {
-                    this.tojos.add(sibling).withDiscoveredAt(src);
-                }
+        final String pkg = object.substring(0, Math.max(object.lastIndexOf('.'), 0));
+        if (completed.add(pkg)) {
+            for (final String sibling : this.objectionary.children(pkg)) {
+                this.tojos.add(sibling).withDiscoveredAt(src);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


morphqdd:
Closing this one. Dropping the `split > 0` guard makes the package an empty string for a top-level object, and `ObjectsIndex.children("")` returns every name without a dot, so the whole root package gets registered.

`MjRegisterIT.removesOldForeignFile` went from 3 objects to 51, and `removesUnnecessaryPulledObjects` started finding `string.eo` in a project that never mentions it.

For `math.vector.add` the package is a small group of siblings. For a top-level object it is everything in the root, so the guard is a real boundary rather than a missing case.

PR #6412, opened 2026-08-05
#6366: delete the binary before marking its tojo unplaced

### LINKED ISSUE #6366: `Unplacing` removes the placement catalog entry before confirming that the binary was deleted
## What happens

When a previously placed binary must be removed, `Unplacing.unplaced()` first calls `tojo.unplace()` and only then calls `Files.deleteIfExists(path)`. If deletion fails, for example because the file is locked on Windows or its directory is not writable, the method throws after the catalog has already forgotten the binary.

A subsequent `unplace` run no longer has the catalog entry required to retry, leaving a stale dependency class in the output directory.

## Why this is a bug

The catalog is the durable record that connects a placed binary to its dependency. Updating it before the filesystem mutation is successful makes a transient deletion failure permanent and can leave an outdated class on the application classpath.

## Expected behavior

Delete the binary first and mark its tojo unplaced only after deletion succeeds (or after it is confirmed absent). Add a regression test with a deletion failure to verify that the catalog entry remains available for a later retry.

morphqdd:
Fixes #6366.

`Unplacing.unplaced()` called `tojo.unplace()` before `Files.deleteIfExists(path)`. If deletion failed (locked file, read-only directory), the method threw after the catalog had already forgotten the binary. A later `unplace` run had no catalog entry left to retry the deletion, leaving a stale class in the output directory permanently.

The binary is now deleted first; the tojo is marked unplaced only after deletion succeeds (or is confirmed already absent).

Added a test that makes the output directory read-only to force a deletion failure, and checks the catalog entry stays marked as placed afterward.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Unplacing.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Unplacing.java
@@ -161,10 +161,10 @@ final class Unplacing implements Step {
      * @throws IOException If fails to delete binary
      */
     private static int unplaced(final TjPlaced tojo, final Path path) throws IOException {
-        tojo.unplace();
         if (Files.deleteIfExists(path)) {
             Logger.debug(Unplacing.class, "Deleted binary %s", path);
         }
+        tojo.unplace();
         return 1;
     }
 }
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (36) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3530](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6413, opened 2026-08-05
#6364: recognize all JUnit test entry-point annotations

### LINKED ISSUE #6364: `JavaPlaced` silently omits generated JUnit tests that use annotations other than `@Test`
## What happens

`JavaPlaced.testsPresent()` decides whether generated Java test source should be written by checking only whether the `<tests>` text contains the literal `@Test`:

```java
clazz.element("tests").text().map(s -> s.contains("@Test")).orElse(false)
```

A valid JUnit class containing only `@ParameterizedTest`, `@RepeatedTest`, `@TestFactory`, or `@TestTemplate` does not contain that substring. When EO transpilation produces such a class, `JavaPlaced.exec()` silently skips it even when test generation is enabled.

## Why this is a bug

Those annotations are JUnit test entry points, and the repository itself uses `@ParameterizedTest` and `@RepeatedTest`. The generated test source is therefore lost solely because it is not expressed with the narrowest annotation form; Maven then reports a green build without executing the intended EO test.

## Expected behavior

Recognize the supported JUnit test annotations, or use the presence of non-empty generated test source rather than a substring probe. Add regression cases for a class with `@ParameterizedTest` and one with ordinary `@Test`.

morphqdd:
Fixes #6364.

`JavaPlaced.testsPresent()` decided whether to write generated test source by checking only for the literal substring `@Test`. A generated JUnit class using `@ParameterizedTest`, `@RepeatedTest`, `@TestFactory`, or `@TestTemplate` instead does not contain that substring, so it was silently skipped even when test generation was enabled. Maven then reported a green build without running the intended EO test.

`testsPresent()` now checks for any of the five JUnit test entry-point annotations.

Added a test with a generated class using only `@ParameterizedTest`, checking the file is written like the existing `@Test` case.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/JavaPlaced.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/JavaPlaced.java
@@ -10,6 +10,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
+import java.util.List;
 import org.cactoos.BiProc;
 
 /**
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (37) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3542](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6416, opened 2026-08-05
#6353: reject a bare '+package' meta with no argument

### LINKED ISSUE #6353: A bare `+package` with no argument produces a locator with an empty segment
## What happens

A bare `+package` meta with no argument produces locators with an empty segment (a doubled dot), which is invalid per the parser's own XMIR schema.

`eo-parser/src/main/resources/org/eolang/parser/parse/set-locators.xsl` lines 34-37 test only whether a `meta[head='package']` is present, then append `"." + tail` to the locator unconditionally. A bare `+package` line has an empty `tail`, so every locator in the file gets a doubled dot.

## Steps to reproduce

Parse this source:

```
+package

# Sample.

[] > app
  42 > x
```

## Observed

The parser reports zero errors and produces locators like `loc="Φ..app"` and `loc="Φ..app.x"`. Validating against `XMIR.xsd` fails:

```
cvc-pattern-valid: Value 'Φ..app' is not facet-valid with respect to pattern 'Φ(\.(ρ|φ|α[0-9]+|[a-z]\S*))*' for type 'locator'
```

Nothing in the pipeline validates that a `+package` meta actually carries exactly one non-empty part before it is used to build locators.

## Expected behavior

A `+package` meta with no argument should fail with a normal `ParseError` at parse time (it is a malformed meta), instead of silently producing locators that don't match the language's own locator grammar.

morphqdd:
Fixes #6353.

`set-locators.xsl` tests only whether a `meta[head='package']` is present, then appends `"." + tail` to every locator unconditionally. A bare `+package` with no argument has an empty `tail`, so every locator in the file got a doubled dot (e.g. `loc="Phi..app"`), which fails `XMIR.xsd`'s own locator pattern.

`LnMeta` now rejects a `+package` meta with no argument at parse time with a `ParseError`, the same way a nameless meta (`+` alone) is already rejected, instead of letting it silently reach the locator-building stage.

Added a declarative parser test with a bare `+package` line, checking parsing reports an error mentioning `package`.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnMeta.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnMeta.java
@@ -75,6 +75,12 @@ final class LnMeta implements Line {
                 "meta directive requires a name"
             );
         }
+        if ("package".equals(head) && parts.isEmpty()) {
+            throw new ParseError(
+                this.span.line(), this.span.indent(),
+                "'+package' directive requires exactly one argument"
+            );
+        }
         Comments.seal(globals, emit, this.span);
         globals.markMeta();
         globals.clearBlanks();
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3578](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6417, opened 2026-08-05
#6365: escape the dot in the .eo extension pattern

### LINKED ISSUE #6365: `Unplace` treats filenames ending in any `?eo` sequence as EO sources because its extension regex does not escape the dot
## What happens

`Unplace` uses `Pattern.compile(".eo$")` to remove an EO extension. In a Java regular expression, the initial dot matches any character rather than the literal filename separator.

Consequently, a path relative to the configured source root such as `demo/foxeo` is converted to program name `demo.fo`: the final `xeo` is removed even though the file has no `.eo` extension.

## Root cause

`eo-maven-plugin/src/main/java/org/eolang/maven/Unplace.java` compiles the pattern with an unescaped dot. It should match only the literal `.eo` suffix.

## Expected behavior

Use a literal extension pattern such as `\.eo$`, and add cases distinguishing `foo.eo` from names like `foxeo` and `foo.java`.

morphqdd:
Fixes #6365.

`Unplace` compiled its extension-stripping pattern as `.eo$`. In a Java regex the unescaped leading dot matches any character, not just a literal `.`. A path like `demo/foxeo` had its trailing `xeo` stripped as if it were a `.eo` extension, producing the wrong program name (`demo.fo` instead of `demo.foxeo`).

The pattern is now `\.eo$`, matching only a literal `.eo` suffix.

Added cases distinguishing a name ending in the literal bytes `eo` with no dot, and a non-`.eo` extension, from an actual `.eo` file.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Unplace.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Unplace.java
@@ -17,7 +17,7 @@ final class Unplace {
     /**
      * Pattern to catch .eo files.
      */
-    private static final Pattern EO = Pattern.compile(".eo$");
+    private static final Pattern EO = Pattern.compile("\\.eo$");
 
     /**
      * The parent dir.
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (6) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3590](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6418, opened 2026-08-05
#6351: remove cti entirely

### LINKED ISSUE #6351: `cti` with fewer than 2 arguments crashes the parser with a raw Saxon exception instead of a ParseError
## What happens

`Φ.cti` (the compile-time-instruction object) crashes the whole parser with a raw XSL/Saxon exception instead of a normal `ParseError` when it is applied to fewer than 2 arguments.

`eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl` lines 37 and 39 call `eo:bytes-to-string(o[last() - 1]/o[1]/o[1]/text())` on the severity argument with no check that a second-to-last child actually exists. When `cti` has fewer than 2 children, `o[last() - 1]` is an empty node sequence, and Saxon aborts the transform train.

## Steps to reproduce

Parse this source:

```
# Sample.

[] > app
  cti > @
    2.times 2
    "oops"
```

`cti` here only has one usable argument in the shape the sheet expects (a forgotten-argument mistake is enough to trigger this, `cti > x` or `cti 1 > x` reproduce it too).

## Observed

`EoSyntax(...).parsed()` throws:

```
IllegalArgumentException: Shift 'com.yegor256.xsline.StClasspath@...' failed
Caused by: net.sf.saxon.trans.XPathException: An empty sequence is not allowed as the first argument of eo:bytes-to-string()
```

No file, no line number, no `<errors>` entry in the resulting XMIR - the whole parse aborts with an internal exception instead of reporting a normal parse error at the `cti` call site.

## Expected behavior

`cti` with too few arguments should fail with a normal `ParseError` (or an `<error>` entry in XMIR) pointing at the offending line, the same way other malformed constructs are reported, not with an unguarded XSL crash.

### LINKED ISSUE #6352: `cti` with a non-string severity argument writes raw decoded bytes into the severity XML attribute
## What happens

`Φ.cti` accepts a non-string severity argument and writes raw decoded bytes of its IEEE-754 representation straight into the `severity` XML attribute, producing invalid XMIR instead of a parse error.

`eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl` (same area as the crash in the sibling issue, lines 37 and 39) never checks that the `cti` severity argument is actually a `Φ.string`. If it is a number, the number's byte payload gets decoded as UTF-8 and written into the `severity` attribute value.

## Steps to reproduce

Parse this source, where the severity argument to `cti` is a number instead of a string literal:

```
# Sample.

[] > app
  cti 1 2 > x
```

## Observed

The parser reports zero errors and emits something like:

```xml
<error check="cti" line="4" severity="?&#0;&#0;&#0;&#0;">@&#0;&#0;&#0;&#0;&#0;&#0;&#0;</error>
```

Validating the result against the module's own `XMIR.xsd` fails:

```
cvc-enumeration-valid: Value '?' is not facet-valid with respect to enumeration '[critical, error, warning]'
```

So the parser silently produces XMIR that its own schema rejects, with the parse itself reporting success.

## Expected behavior

`cti` should validate that its severity argument is a string literal one of `critical`/`error`/`warning`, and raise a normal `ParseError` when it is not, instead of writing decoded garbage bytes into the `severity` attribute.

morphqdd:
Fixes #6351. Also resolves #6352 (severity validation is moot once cti is gone).

Yegor asked to remove the cti object entirely instead of validating it: "we should remove cti object entirely, this feature is gone."

Deleted cti-adds-errors.xsl, its two wiring spots in Canonical.java, and its two eo-syntax test packs.

MjLintTest in eo-maven-plugin used cti as a fixture to trigger a guaranteed lint defect of a chosen severity. Replaced it with real triggers: a duplicate +home meta for a plain error (unique-metas), and a top object name that does not match the file name for critical (EoSource's own validate-object-name check).


yegor256:
@morphqdd we don't have `cti` anymore. Delete anything related to it.

eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl
```diff
@@ -25,7 +25,7 @@
       </xsl:if>
     </xsl:copy>
   </xsl:template>
-  <xsl:template match="o[@base='Φ.cti']" mode="create">
+  <xsl:template match="o[@base='Φ.cti' and count(o) &gt;= 2]" mode="create">
```
Thayorns: The guard is on arity, but the crash is on shape. `eo:bytes-to-string` is declared `as="xs:string"` in `_funcs.xsl:33`, so it aborts on an empty sequence, and line 37 reaches three levels down - `o[last() - 1]/o[1]/o[1]/text()` - before handing the result over. A `cti` whose severity argument is an object reference rather than a data literal satisfies `count(o) >= 2` and still has nothing at `o[1]/o[1]`, so `cti sev "msg" > @` takes the same raw XPathException out of the parse that this is fixing for `cti > @`.

eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl
```diff
@@ -39,6 +39,20 @@
       <xsl:value-of select="eo:bytes-to-string(o[last()]/o[1]/o[1]/text())"/>
     </error>
   </xsl:template>
+  <xsl:template match="o[@base='Φ.cti' and count(o) &lt; 2]" mode="create">
+    <error>
+      <xsl:attribute name="check">
```
Thayorns: These six lines are the first two attribute blocks of the template above copied verbatim; only the severity and the body differ. Both templates now have to be kept saying `check="cti"` and reading `@line` the same way, and a third `cti` failure mode would want a third copy of them.

Thayorns (commented):
Turning a transform-train abort into an ordinary `<error>` entry is clearly the right direction - a malformed `cti` should be reported the way every other malformed construct is, with a line number the user can go to. Splitting the template on the match pattern rather than branching inside it fits how the rest of this stylesheet is written. Two remarks.

morphqdd:
@yegor256, please  check

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (166 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3646](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+994](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6419, opened 2026-08-05
#6352: validate cti severity is a critical/error/warning string

### LINKED ISSUE #6352: `cti` with a non-string severity argument writes raw decoded bytes into the severity XML attribute

morphqdd:
Fixes #6352.

`cti-adds-errors.xsl` decoded whatever bytes sit at the severity argument's data-literal position as UTF-8 and wrote them straight into the `severity` XML attribute, with no check that the argument is actually a string, let alone one of the three legal values. Passing a number there decoded its raw IEEE-754 payload into the attribute, producing XMIR that `XMIR.xsd`'s own enumeration rejects, while the parser itself reported zero errors.

Added `eo:cti-has-valid-severity()`, checking the severity argument is a `Phi.string` literal whose decoded value is `critical`, `error` or `warning`. A `cti` call that fails this now gets a normal `<error>` entry instead of garbage output.

This touches the same file as #6351 (missing-arguments crash) and reuses the same `count(o) < 2` guard, so the two PRs will need a rebase against whichever lands first.

Added a declarative parser test with `cti 1 2 > x` (numeric severity), checking the error mentions "severity", and verified the existing valid-severity case (`cti` with a `"warning"` string) still emits the correct attribute.

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl
@@ -28,10 +28,7 @@
   <!-- A severity argument that is a string literal with one of the three legal values. -->
   <xsl:function name="eo:cti-has-valid-severity" as="xs:boolean">
     <xsl:param name="cti" as="element()"/>
-    <xsl:sequence select="
-      $cti/o[last() - 1]/@base = 'Φ.string'
-      and eo:bytes-to-string($cti/o[last() - 1]/o[1]/o[1]/text()) = ('critical', 'error', 'warning')
-      "/>
+    <xsl:sequence select="$cti/o[last() - 1]/@base = 'Φ.string' and eo:bytes-to-string($cti/o[last() - 1]/o[1]/o[1]/text()) = ('critical', 'error', 'warning')"/>
   </xsl:function>
   <xsl:template match="o[@base='Φ.cti' and count(o) &lt; 2]" mode="create">
     <error>
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl
@@ -3,7 +3,7 @@
 * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 * SPDX-License-Identifier: MIT
 -->
-<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="cti-adds-errors" version="2.0" exclude-result-prefixes="eo">
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:eo="https://www.eolang.org" id="cti-adds-errors" version="2.0" exclude-result-prefixes="eo">
   <!--
   For every cti objects add error messages.
   -->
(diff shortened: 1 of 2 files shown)


yegor256:
@morphqdd we don't have `cti` object in eo-runtime anymore. Just delete everything (all .xsl) related to `cti`.

eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl
```diff
@@ -3,7 +3,7 @@
 * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 * SPDX-License-Identifier: MIT
 -->
-<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:eo="https://www.eolang.org" id="cti-adds-errors" version="2.0" exclude-result-prefixes="eo">
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:eo="https://www.eolang.org" id="cti-adds-errors" version="2.0" exclude-result-prefixes="eo">
```
Thayorns, line 6: `xmlns:xs` is declared here but not added to `exclude-result-prefixes`, which still lists only `eo`. `<error>` is a literal result element, so every in-scope prefix that is not excluded gets copied onto it, and every error this stylesheet emits now carries `xmlns:xs="http://www.w3.org/2001/XMLSchema"` into the XMIR.

eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl
```diff
@@ -25,7 +25,40 @@
       </xsl:if>
     </xsl:copy>
   </xsl:template>
-  <xsl:template match="o[@base='Φ.cti']" mode="create">
+  <!-- A severity argument that is a string literal with one of the three legal values. -->
+  <xsl:function name="eo:cti-has-valid-severity" as="xs:boolean">
+    <xsl:param name="cti" as="element()"/>
+    <xsl:sequence select="$cti/o[last() - 1]/@base = 'Φ.string' and eo:bytes-to-string($cti/o[last() - 1]/o[1]/o[1]/text()) = ('critical', 'error', 'warning')"/>
```
Thayorns, line 31: The right-hand operand reaches `o[1]/o[1]/text()` on the severity argument, and `eo:bytes-to-string` is declared `as="xs:string"` in `_funcs.xsl:33`, so it aborts on an empty sequence. A severity that is an object reference rather than a literal - `cti sev "msg" > @` - has nothing at `o[1]/o[1]`, and whether that reaches the function depends on the processor evaluating the left operand of `and` first, which XPath does not require it to do. The left half already establishes the argument is a string literal; the right half runs on the assumption it did.

eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl
```diff
@@ -25,7 +25,40 @@
...
+      </xsl:attribute>
+      <xsl:attribute name="severity">
+        <xsl:text>error</xsl:text>
+      </xsl:attribute>
+      <xsl:text>cti requires two arguments: a severity and a message</xsl:text>
+    </error>
+  </xsl:template>
+  <xsl:template match="o[@base='Φ.cti' and count(o) &gt;= 2 and not(eo:cti-has-valid-severity(.))]" mode="create">
+    <error>
+      <xsl:attribute name="check">
```
Thayorns, line 49: Three templates now, and the `check` and `line` attribute blocks are written out identically in all three, with the `severity` block identical in two of them. That is about twenty of the thirty-five added lines. The thing that actually varies between them is one line of text at the bottom, and it is hard to see that through the scaffolding.

eo-parser/src/main/resources/org/eolang/parser/parse/cti-adds-errors.xsl
```diff
@@ -25,7 +25,40 @@
...
+        <xsl:text>cti</xsl:text>
+      </xsl:attribute>
+      <xsl:attribute name="line">
+        <xsl:value-of select="@line"/>
+      </xsl:attribute>
+      <xsl:attribute name="severity">
+        <xsl:text>error</xsl:text>
+      </xsl:attribute>
+      <xsl:text>cti severity must be a string literal, one of "critical", "error", "warning"</xsl:text>
```
Thayorns, line 58: The message lists what is allowed but never says what was written. Someone who typed `cti 1 2` gets told the three legal values back and has to go find their own line to see what they had. The offending argument is right there in `$cti/o[last() - 1]` when this error is built.

Thayorns (commented):
Catching a severity that the XSD enumeration would reject anyway, at parse time and with a line number, is worth doing - emitting XMIR that fails its own schema while reporting zero errors is the worst of the available outcomes. Pushing the check into a named function keeps the match patterns readable too. A few things.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+960](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6420, opened 2026-08-05
#6355: reject a compact-tuple count that overflows int

### LINKED ISSUE #6355: Compact-tuple `*N` count overflows a 32-bit int in `LnCompactTuple.readCount` instead of being rejected
## What happens

The compact-tuple count in `*N` syntax is parsed into a plain 32-bit `int` with no range check, so a count above `Integer.MAX_VALUE` silently wraps instead of being rejected, and the wrapped value changes how the line is parsed.

`eo-parser/src/main/java/org/eolang/parser/LnCompactTuple.java` line 125, in `readCount`, accumulates digits with `count = count * 10 + glyph - '0'` and never checks the running total against `Integer.MAX_VALUE`. `Eo.closeCompactTuple` (`Eo.java:968`) later compares `children == count` by equality, so a wrapped count that happens to equal the real child count is accepted as if nothing were wrong, and a negative wrap disables the compact-tuple feature outright.

## Steps to reproduce

Three cases, each with a source line using a compact-tuple count near or past `Integer.MAX_VALUE`:

1. `seq *4294967297` (wraps to 1) applied to 3 children.
2. `seq *4294967296` (wraps to 0) applied to 2 children.
3. `seq *2147483648` (wraps to `-2147483648`) applied to 2 children.

Compare against the control `seq *3` applied to only 2 children, which correctly errors with "compact tuple requires at least 3 children, got 2".

## Observed

- Case 1: parses as if it were `seq *1` - the XMIR shape matches a compact tuple of size 1, not an error, even though `*4294967297` is nowhere near a valid small count.
- Case 2: parses as if it were `seq *0`.
- Case 3: no `Φ.tuple` node is emitted at all; the line silently degrades to a plain application `seq 1 2`, with no error reported anywhere.

None of this is caught by the existing test pack `eo-packs/parse/validate-resolve-before-stars.yaml`, which only exercises small counts.

## Expected behavior

`readCount` should reject a compact-tuple count that overflows `int` (or reasonably, any count implausibly larger than the source file could ever have that many children for) with a normal `ParseError`, instead of silently wrapping and changing the parsed meaning of the line.

morphqdd:
Fixes #6355.

`LnCompactTuple.readCount` accumulated the `*N` digits into a plain 32-bit `int` with no range check, so a count above `Integer.MAX_VALUE` silently wrapped instead of being rejected, and the wrapped value changed how the line parsed. `Eo.closeCompactTuple` compares the wrapped count against the real child count by equality, so a count that happened to wrap to the actual child count was accepted as if nothing were wrong, and a negative wrap disabled the feature outright with no error anywhere.

`readCount` now accumulates into a `long` and throws a `ParseError` the moment the running total would exceed `Integer.MAX_VALUE`, pointing at the start of the digit run.

Added a declarative parser test with `seq *4294967297` (the smallest overflowing count from the issue), checking parsing reports a "too large" error instead of silently wrapping.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnCompactTuple.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnCompactTuple.java
@@ -80,7 +80,7 @@ final class LnCompactTuple implements Line {
             );
         }
         tokens.seek(tokens.cursor() + 1);
-        final int count = LnCompactTuple.readCount(tokens);
+        final int count = LnCompactTuple.readCount(tokens, this.span);
         final Suffix suffix = new Suffix(
             tokens.tail(), this.span, this.span.indent() + tokens.cursor()
         );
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (29) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3602](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6421, opened 2026-08-05
#6356: make @loc agree with @as for bound arguments

### LINKED ISSUE #6356: `@as` and `@loc` disagree on the attribute slot for numeric/label argument bindings (`1:1 2:0`)
## What happens

For an application with numeric or label argument bindings (`1:1`, `2:a`, and so on, per R-3.12), the `@as` attribute and the `@loc` attribute computed for the same argument element disagree about which attribute slot it occupies.

`eo-parser/src/main/resources/org/eolang/parser/parse/mandatory-as.xsl` lines 21-24 compute the bound slot as `α{count(preceding-sibling::o[not(@as)])}` - position among as-yet-unbound arguments. `eo-parser/src/main/resources/org/eolang/parser/parse/set-locators.xsl` lines 57-60 compute the locator's trailing segment from raw source position instead, and it does honor `@name` but ignores `@as` entirely.

## Steps to reproduce

Parse this source, which binds arguments to explicit numeric attribute slots out of source order:

```
# Sample.

[] > app
  seq 1:1 2:0 > x
```

## Observed

On the first argument element: `as="α1"` but `loc="Φ.app.x.α0"`. On the second: `as="α0"` but `loc="Φ.app.x.α1"` - the two attributes describing the same element's binding disagree. With label bindings (`seq 1:b 2:a`) the locators come out as `α0`/`α1` in source order while the elements are actually bound to attributes `b`/`a`, so the locator never reflects the attribute the argument is actually occupying.

No existing test pack under `eo-parser/src/test/resources/org/eolang/parser/` exercises numeric or label argument bindings (`:` syntax) at all.

## Expected behavior

`@loc` should reflect the same attribute slot that `@as` records for the same element - either `set-locators.xsl` should read `@as` when present the same way `mandatory-as.xsl` computes it, or the intended relationship between the two attributes needs to be clarified and both sheets brought into agreement.

morphqdd:
Fixes #6356.

`eo:locator()` in `set-locators.xsl` computed the trailing `αN` segment purely from an argument's source position, ignoring the `@as` attribute that `mandatory-as.xsl` (or an explicit `N:value` binding) had already set on the same element. For an application with out-of-source-order numeric bindings like `seq 1:1 2:0`, `@as` and `@loc` disagreed about which attribute slot the element occupies: the first-written argument got `as="alpha1"` but `loc=".alpha0"`, and vice versa. The same mismatch happens with label bindings (`1:b 2:a`), where `@loc` never reflects the actual bound attribute name at all.

`eo:locator()` now reads `@as` directly when present, falling back to the existing rho/position logic only for the one case that never gets an `@as` from `mandatory-as.xsl` (the receiver of a reversed-dispatch method chain, e.g. `joined. *1 > x`).

Added a declarative parser test with `seq 1:1 2:0 > x`, checking every element's `@loc` trailing segment matches its `@as`.

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/set-locators.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/set-locators.xsl
@@ -51,6 +51,9 @@
         </xsl:when>
         <xsl:otherwise>
           <xsl:choose>
+            <xsl:when test="$o/@as">
+              <xsl:value-of select="$o/@as"/>
+            </xsl:when>
             <xsl:when test="starts-with($o/parent::o/@base, '.') and not($o/preceding-sibling::o)">
               <xsl:value-of select="$eo:rho"/>
             </xsl:when>
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/loc-agrees-with-as-for-numeric-bindings.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/loc-agrees-with-as-for-numeric-bindings.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets: []
+asserts:
+  - /object[not(errors)]
+  - not(//o[@as and not(ends-with(@loc, concat('.', @as)))])
+  - //o[@as='α1' and @loc='Φ.app.x.α1']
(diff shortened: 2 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (15) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3614](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6422, opened 2026-08-05
#6358: catch ClassCastException when deserializing a compiled regex

### LINKED ISSUE #6358: `string.regex.pattern` deserializes caller-supplied bytes with `ObjectInputStream`, leaking `ClassCastException` on a type mismatch
## What happens

`string.regex.pattern` is a public object that accepts arbitrary bytes and feeds them directly into Java deserialization, with no check that the bytes actually came from `regex.compile`. A malformed or mistyped input leaks a raw Java exception instead of a clean EO-level failure, and this is an untrusted-deserialization sink reachable from ordinary EO code.

`eo-runtime/src/main/java/org/eolang/EO_string/EOregex$EOpattern$EOmatch$EOmatched_from_index.java` lines 76-78 call `new ObjectInputStream(new ByteArrayInputStream(...)).readObject()` on caller-supplied bytes and cast the result to `Pattern`. The surrounding `catch` (lines 79-81) only handles `IOException` and `ClassNotFoundException`, not `ClassCastException`.

## Steps to reproduce

Call `string.regex.pattern` with bytes that are a validly-serialized Java object of the wrong type, or with bytes that are simply not a valid Java serialization stream, then use `.match(...)`.

## Observed

- A serialized `java.lang.String` instead of a `Pattern`:
  ```
  ClassCastException: class java.lang.String cannot be cast to class java.util.regex.Pattern
  ```
- Arbitrary bytes that are not a serialization stream at all (`01-02-03-04-05-06-07-08`):
  ```
  StreamCorruptedException: invalid stream header
  ```
  (this one is caught, but the first case is not)

Beyond the leaked exception, this design means any EO program that can construct or receive a byte sequence from an untrusted source and pass it to `regex.pattern` triggers full Java object deserialization on that data, which is a well-known class of security issue independent of the crash.

## Expected behavior

`regex.pattern` should either validate that the bytes are self-consistent with having come from `regex.compile` before deserializing, or the `catch` should also handle `ClassCastException` and any other exception type deserialization can throw, converting it into a normal `ExFailure` with a clear message instead of letting a raw Java exception through. Longer term, replacing Java object serialization with a safer encoding of the compiled pattern (or just the source regex string, recompiled on each use) would remove the deserialization risk entirely.

morphqdd:
Fixes #6358.

`string.regex.pattern` deserializes the caller-supplied `serialized` bytes with `ObjectInputStream` and casts the result to `Pattern`, with the surrounding `catch` handling only `IOException` and `ClassNotFoundException`. Bytes that deserialize successfully to some other type (a serialized `String`, for example) leaked a raw `ClassCastException` instead of the same clean `ExFailure` used for the other two failure modes.

Added `ClassCastException` to the catch clause, so a type mismatch produces the same "cannot deserialize the compiled regex pattern" `ExFailure` as a malformed stream.

This is the minimal fix the issue itself allows for; replacing Java object serialization with a safer encoding is called out there as separate, longer-term work.

Added a test that serializes a `String` (a validly-serialized object of the wrong type) and feeds it through `pattern`, checking the failure is the clean `ExFailure` message, not a raw `ClassCastException`.

### DIFF
--- a/eo-runtime/src/main/eo/string/regex.eo
+++ b/eo-runtime/src/main/eo/string/regex.eo
@@ -261,6 +261,11 @@
             regex "/$/"
             "x"
 
+  next. --> rejects-serialized-bytes-of-the-wrong-type
+    match.
+      pattern AC-ED-00-05-74-00-0D-6E-6F-74-20-61-20-70-61-74-74-65-72-6E
+      "hello"
+
   compiled. --> stops-on-a-missing-closing-slash
     regex "/pattern"
 
--- a/eo-runtime/src/test/java/org/eolang/EO_string/EOregexEOpatternEOmatchEOmatchedfromindexTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EO_string/EOregexEOpatternEOmatchEOmatchedfromindexTest.java
@@ -134,23 +134,28 @@ final class EOregexEOpatternEOmatchEOmatchedfromindexTest {
         }
         final Phi pattern = Phi.Φ.take("string.regex").take("pattern").copy();
         pattern.put(0, new Data.ToPhi(baos.toByteArray()));
-        final ExAbstract failure = Assertions.assertThrows(
-            ExAbstract.class,
-            () -> new Dataized(
-                new PhApplication(
-                    new PhApplication(
-                        pattern.take("match").copy(),
-                        "txt", new Data.ToPhi("hello")
-                    ).take("matched-from-index").copy(),
-                    new Bind(EOregexEOpatternEOmatchEOmatchedfromindexTest.position(), new Data.ToPhi(1)),
-                    new Bind(EOregexEOpatternEOmatchEOmatchedfromindexTest.start(), new Data.ToPhi(0))
-                ).take("from")
-            ).take(),
-            "bytes deserializing to the wrong type must fail with ExFailure, not a raw ClassCastException"
-        );
         MatcherAssert.assertThat(
-            "the failure must be the clean deserialize message, not a raw ClassCastException",
-            failure.toString(),
+            "a raw ClassCastException leaked instead of the clean deserialize failure",
+            Assertions.assertThrows(
+                ExAbstract.class,
+                () -> new Dataized(
+                    new PhApplication(
+                        new PhApplication(
+                            pattern.take("match").copy(),
+                            "txt", new Data.ToPhi("hello")
+                        ).take("matched-from-index").copy(),
+                        new Bind(
+                            EOregexEOpatternEOmatchEOmatchedfromindexTest.position(),
+                            new Data.ToPhi(1)
+                        ),
+                        new Bind(
+                            EOregexEOpatternEOmatchEOmatchedfromindexTest.start(),
+                            new Data.ToPhi(0)
+                        )
+                    ).take("from")
+                ).take(),
+                "bytes of the wrong type must fail with ExFailure"
+            ).toString(),
             Matchers.containsString("cannot deserialize the compiled regex pattern")
         );
     }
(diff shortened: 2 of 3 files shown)


yegor256:
@morphqdd can you add a EO test too, to reproduce this problem?

eo-runtime/src/test/java/org/eolang/EO_string/EOregexEOpatternEOmatchEOmatchedfromindexTest.java
```diff
@@ -124,6 +126,40 @@ void rejectsStartIndexAfterTextEnd() {
...
+            oos.writeObject("not a pattern");
+        }
+        final Phi pattern = Phi.Φ.take("string.regex").take("pattern").copy();
+        pattern.put(0, new Data.ToPhi(baos.toByteArray()));
+        MatcherAssert.assertThat(
+            "a raw ClassCastException leaked instead of the clean deserialize failure",
+            Assertions.assertThrows(
+                ExAbstract.class,
+                () -> new Dataized(
+                    new PhApplication(
```
Thayorns, line 142: This application chain is `matchedFromIndex` at line 202 written out again: same nesting, same `"txt"` binding to `"hello"`, same two `Bind`s on `position` and `start`. The only real difference is that the pattern object comes from `take("pattern")` rather than from `expression` and `compiled`. Fourteen of the test's lines are that copy, and what the test is actually about - some bytes that deserialize to a String - is three lines at the top.

eo-runtime/src/main/eo/string/regex.eo
```diff
@@ -261,6 +261,11 @@
             regex "/$/"
             "x"
 
+  next. --> rejects-serialized-bytes-of-the-wrong-type
+    match.
+      pattern AC-ED-00-05-74-00-0D-6E-6F-74-20-61-20-70-61-74-74-65-72-6E
```
Thayorns, line 266: Nothing here tells a reader that these bytes are a serialized Java String saying "not a pattern"; they have to recognise the `AC-ED-00-05` header and decode the rest by hand. The test name says the type is wrong but not what type it is, so if this ever fails the next person has twenty hex pairs and no way to tell whether the fixture or the code moved.

eo-runtime/src/test/java/org/eolang/EO_string/EOregexEOpatternEOmatchEOmatchedfromindexTest.java
```diff
@@ -124,6 +126,40 @@ void rejectsStartIndexAfterTextEnd() {
         );
     }
 
+    @Test
+    void rejectsSerializedBytesOfTheWrongType() throws Exception {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
Thayorns, line 131: `baos` and `oos` are the stream classes' initials rather than names for what these hold - the bytes, and the writer that fills them. Every other local in this file reads as a word: `pattern`, `start`, `text`.

Thayorns (commented):
Adding `ClassCastException` to the existing multi-catch is the right size of fix here - the cast is on the same line as the `readObject` it belongs to, and a stream that decodes to the wrong type is the same kind of bad input as one that does not decode at all. Both tests are worth having, one at the EO level and one at the Java level. Three remarks, all about how they read.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3630](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+967](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6431, opened 2026-08-06
fix: complete root-package siblings during probing

### LINKED ISSUE #6370: `Probing` does not pull root-package siblings for a top-level `+probe` object

mvanhorn:
Derive a package key for every valid probed object: retain the existing substring for qualified names and use the empty string for top-level names. Feed that key through the existing `completed` set and `Objectionary.children(pkg)` loop so root completion has the same per-package de-duplication and tojo registration behavior as nested-package completion. `Probing.complete()` completes the containing package only when a probed object name has a dot. A top-level probe such as `foo` therefore registers `foo` but never asks the objectionary for `children("")`, even though `ObjectsIndex` already defines the empty string as the root package.

Probe a top-level `foo` while the object index contains `foo` and `bar`; verify both the direct probe and root sibling are registered in `TjsForeign`; Probe a qualified object such as `tuple.each`; verify the existing behavior still registers only the direct siblings of `tuple`, preserving nested-package completion.

Fixes #6370

### DIFF
--- a/eo-integration-tests/src/test/java/org/eolang/maven/MjRegisterIT.java
+++ b/eo-integration-tests/src/test/java/org/eolang/maven/MjRegisterIT.java
@@ -82,18 +82,17 @@ final class MjRegisterIT {
                 MjRegisterIT.succeeds(f);
                 final TjSmart foreign = MjRegisterIT.loadForeign(temp);
                 MatcherAssert.assertThat(
-                    "Foreign must contain only 3 references to objects, but it doesn't",
-                    foreign.size(),
-                    Matchers.equalTo(3)
-                );
-                MatcherAssert.assertThat(
-                    "Foreign must contain refs to Number, Bytes, and current object",
-                    MjRegisterIT.existences(foreign, "number", "bytes", "foo"),
+                    "Foreign must contain required objects and root-package siblings",
+                    MjRegisterIT.existences(
+                        foreign, "number", "bytes", "string", "foo"
+                    ),
                     Matchers.everyItem(Matchers.is(true))
                 );
                 MatcherAssert.assertThat(
                     "Foreign must not contain a reference to an old object",
-                    foreign.select(tojo -> "string".equals(tojo.get("id"))).isEmpty(),
+                    foreign.select(
+                        tojo -> "io.stdout".equals(tojo.get("id"))
+                    ).isEmpty(),
                     Matchers.is(true)
                 );
             }
@@ -106,7 +105,7 @@ final class MjRegisterIT {
             f -> {
                 MjRegisterIT.run(
                     f,
-                    new String[]{"  \"Hello\" > @", "  42 > @"},
+                    new String[]{"  Q.io.stdout > @", "  42 > @"},
                     "eo:register", "eo:parse", "eo:probe", "eo:pull"
                 );
                 f.exec("eo:register", "eo:parse", "eo:probe", "eo:pull");
@@ -118,7 +117,7 @@ final class MjRegisterIT {
                 );
                 MatcherAssert.assertThat(
                     "Unnecessary objects were not removed",
-                    temp.resolve("target/eo/2-pull/string.eo").toFile().exists(),
+                    temp.resolve("target/eo/2-pull/io/stdout.eo").toFile().exists(),
                     Matchers.is(false)
                 );
             }
@@ -157,7 +156,7 @@ final class MjRegisterIT {
                 "+rt jvm org.eolang:eo-runtime:0.25.0",
                 "",
                 "[] > foo",
-                "  \"42\" > @"
+                "  Q.io.stdout > @"
             )
         );
         new AppendedPlugin(farea).value();
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/ProbingTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/ProbingTest.java
@@ -8,15 +8,16 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Collection;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import org.cactoos.Input;
+import org.cactoos.map.MapEntry;
+import org.cactoos.map.MapOf;
 import org.cactoos.set.SetOf;
 import org.eolang.parser.EoSyntax;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
+import org.mockito.AdditionalAnswers;
+import org.mockito.Mockito;
 
 /**
  * Test cases for {@link Probing}.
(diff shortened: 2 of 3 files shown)


eo-maven-plugin/src/test/java/org/eolang/maven/ProbingTest.java
```diff
@@ -62,14 +62,91 @@ void completesPartiallyProbedPackage(@TempDir final Path temp) throws IOExceptio
...
+                new Objectionary.Fake(),
+                new ObjectsIndex(
+                    () -> new SetOf<>("foo", "bar", "baz", "nested.object")
+                )
+            ),
+            true
+        ).exec();
+        MatcherAssert.assertThat(
+            "Multiple root probes should produce one complete root object set",
+            tojos.size() == 4
```
Thayorns: completesRootPackageOnce cannot fail for the thing its name promises. TjsForeign() is backed by TjDefault, whose add(id) reads the rows, filters by id and only inserts when the id is absent, so completing the root package a second time adds nothing. Take the completed.add(pkg) guard out of Probing.complete and this test is still green at size 4 - the only cost of repeated completion is an extra children() call, and nothing here observes that.

eo-maven-plugin/src/test/java/org/eolang/maven/ProbingTest.java
```diff
@@ -62,14 +62,91 @@ void completesPartiallyProbedPackage(@TempDir final Path temp) throws IOExceptio
...
+                    )
                 )
             ),
             true
         ).exec();
         MatcherAssert.assertThat(
-            "Probe should have registered the siblings that were never probed directly",
-            tojos.contains("tuple.eachi") && tojos.contains("tuple.withouti"),
+            "Probe should register only direct siblings from the same package",
+            tojos.contains("tuple.eachi")
```
Thayorns: This folds two separate claims into one boolean: that the direct siblings were registered, and that a sub-package object was not. Same shape on 111-113 and on 146-149. When it goes red the message is the whole conjunction, so you cannot tell whether tuple.eachi went missing or tuple.nested.object leaked in, and the second claim is about ObjectsIndex.children filtering rather than about the completion this test is named for.

Thayorns (commented):
The production change is small and does what the issue asks: a probed name with no dot now yields the empty package key, which ObjectsIndex.children already treats as the root and filters down to names with no dot, so root siblings get registered the same way nested ones already were. The completed set now guards the root too. My reservations are all on the test side, where the new cases assert less than their names promise.

mvanhorn:
Both fixed in 3def16c, sorry for pushing it without a word.

On completesRootPackageOnce: you were right that it could not fail for its own name. TjDefault.add(id) dedupes, so the size-4 assertion held with or without the guard. The test now wraps the Objectionary and records every children(pkg) call, then asserts the recorded completions are exactly [""]. That is the extra call you pointed at, and it is now the thing under test, so dropping completed.add(pkg) from Probing.complete turns it red.

On the folded booleans at 77, 111-113 and 146-149: split. Each claim is its own assertThat with its own message, so a failure names which object went missing or leaked rather than handing back the whole conjunction. The sub-package claim is also stated separately from the sibling-registration one, since as you said it is about ObjectsIndex.children filtering rather than completion.

mvanhorn:
Addressed the feedback in ProbingTest. Dropped the nested-package fixtures and assertions from the completion tests and kept a separate descriptive assertion per expected sibling, so a failure names which one is missing. The `ObjectsIndex.children` filtering coverage stays in ObjectsIndexTest where it belongs.

One file, 3 insertions and 14 deletions. The focused Maven test could not run locally since the dependencies are not cached here.

yegor256:
@mvanhorn conflicts here

mvanhorn:
@yegor256 rebased onto master, conflicts are gone.

The conflict was in `Probing.java` and worth a note. Master had grown the `org.eolang.` qualification of siblings, which this branch did not have, but it still gated the completion on `split > 0`, so an object with no dot in its name never had its root-package siblings completed. That gate is the bug this PR is about.

Resolved by keeping master's qualification and dropping only the `split > 0` guard, so `pkg` falls back to the empty root package via the existing `Math.max(split, 0)`. `ProbingTest` runs 6/6.

mvanhorn:
Fixed in 16da9de8e. Both `integration` and `qulice` pass locally.

The integration test was still written against the old probing behaviour. It treated `string` as a stale object that should have been removed, and asserted that foreign held exactly three references. Once probing completes root-package siblings, `string` is pulled legitimately, so both assertions were wrong rather than the code being wrong.

The fixture now uses a nested object, `io.stdout`, as the genuinely stale one, so the "unnecessary objects were removed" case still tests something real, and the root siblings are asserted positively.

One thing worth flagging, since it is a behavioural change and not just a test fix: completing root-package siblings takes foreign from 3 references to 51 - effectively the whole `org.eolang` root package. That looks like the intended consequence of the change, but it is a large jump, so I wanted it stated plainly rather than buried.

I deliberately did not replace the old `equalTo(3)` with `equalTo(51)`. I tried it, and an exact count pins the test to eo-runtime's current object inventory, so it would break on any release that adds an object to the root package. The nested-object absence check covers the actual invariant (root siblings yes, nested no) without that fragility. Happy to add a bound back if you would rather have one.

Verified with:

```
mvn -pl eo-integration-tests -Dit.test=MjRegisterIT verify   -> 4/4 pass
mvn -pl eo-maven-plugin -Pqulice -DskipTests verify          -> clean
```

eo-integration-tests/src/test/java/org/eolang/maven/MjRegisterIT.java
```diff
@@ -156,7 +155,7 @@ private static void runForeign(final Farea farea) throws IOException {
                 "+rt jvm org.eolang:eo-runtime:0.25.0",
                 "",
                 "[] > foo",
-                "  \"42\" > @"
+                "  Q.io.stdout > @"
```
yegor256, line 159: @mvanhorn we don't have `Q.io.stdout` in our codebase anymore. Is it intentionally here or a typo?

yegor256:

---

yegor256:
This has merge conflicts with master now. Needs a rebase.

---

mvanhorn:
Intentional, though the reason is buried in the test change so it is a fair thing to ask about.

This PR completes siblings of the root package, so the fixtures that previously used "Hello" and "42" no longer work as negative cases: those resolve to root-package objects, which now legitimately pull their siblings. The assertion needs an object outside the root package to prove the completion does not spill into nested ones, and Q.io.stdout is that object. It is resolved through eo:pull from objectionary rather than from the local runtime, so the runtime moving stdout to the root package does not affect it. MjTranspileIT and ProbesTest on master still use the same reference.

Caveat on my side: I ran ProbingTest (7/7 green) but not the MjRegisterIT integration suite locally. If objectionary no longer serves io.stdout, or you would rather the fixture used a different nested object, name one and I will swap it.

Also rebased onto master, so the conflict is cleared and CI is green on 53c465f.

PR #6432, opened 2026-08-06
fix(#6303): accept exponent notation in `string.as-number`

### LINKED ISSUE #6303: `string.as-number` fails to convert valid exponent-form numeric text
## What happens

`string.as-number` is documented as converting a numeric text to an EO `number`, but it first validates the entire text with:

```eo
/^[+-]?\d+(?:\.\d+)?$/
```

That expression accepts only a plain integer or fixed-point form. It rejects exponent notation, even though EO numeric literals and the runtime already use it extensively (for example `1.0e20` and `9.223372036854776e18` in runtime tests).

Consequently these conversions take `cant-parse` or terminate:

```eo
"1e3".as-number       # rejected
"-2.5E-3".as-number   # rejected
```

## Expected behavior

`string.as-number` should accept the finite decimal exponent forms that EO accepts for number literals, including an optional exponent sign. The examples above should produce `1000` and `-0.0025`.

## Suggested fix

Extend both the full-text validation and the `%f` token used by `string.scanf` in one coherent change, so that the validator does not merely widen accepted input while the parser still stops at `e` or `E`. Add positive and negative exponent regression cases plus malformed exponent cases such as `"1e"` and `"e3"`.

asmirnov-backend:
Fixes #6303.

`as-number`'s anchored validator and `scanf`'s `%f` token both rejected `e`/`E`, so `"1e3"` and `"-2.5E-3"` took the `cant-parse` branch. Both are widened in one change, so the validator can't outrun the parser.

Widening the regex alone isn't enough: `as-float` folded the whole match digit by digit, so `'e'` would have counted as 53. It now splits the mantissa from the exponent and scales. A negative exponent divides by `10^|exp|` instead of multiplying by `10^exp` — the multiply path double-rounds and `-2.5E-3` misses `-0.0025`.

Tests: three positive cases per file, plus `"1e"`, `"e3"` and `"1e2.5"` in `as-number` as controls that the widening didn't turn into "accept anything".

### DIFF
--- a/eo-runtime/src/main/eo/string/as-number.eo
+++ b/eo-runtime/src/main/eo/string/as-number.eo
@@ -19,7 +19,7 @@
     not.
       matches.
         compiled.
-          regex "/^[+-]?\\d+(?:\\.\\d+)?$/"
+          regex "/^[+-]?\\d+(?:\\.\\d+)?(?:[eE][+-]?\\d+)?$/"
         text
     cant-parse
       "Can't convert text %s to number".printf
@@ -29,13 +29,25 @@
         at.
           scanf "%f" text
 
+  eq. ++> can-convert-exponent-text
+    1000
+    as-number "1e3"
+
   eq. ++> can-convert-float-text
     3.14
     as-number "3.14"
 
+  eq. ++> can-convert-fractional-exponent-text
+    150
+    as-number "1.5e+2"
+
   eq. ++> can-convert-negative-text
     -42
     as-number "-42"
+
+  eq. ++> can-convert-signed-exponent-text
+    -0.0025
+    as-number "-2.5E-3"
 
   eq. ++> can-convert-text
     5
@@ -61,10 +73,28 @@
       "1 2 3"
       42 > [message]
 
+  eq. ++> rejects-text-with-a-dangling-exponent
+    42
+    as-number
+      "1e"
+      42 > [message]
+
+  eq. ++> rejects-text-with-a-fractional-exponent
+    42
+    as-number
+      "1e2.5"
+      42 > [message]
+
   eq. ++> rejects-text-with-an-embedded-number
     42
     as-number
       "abc5"
+      42 > [message]
+
+  eq. ++> rejects-text-with-an-exponent-and-no-mantissa
+    42
+    as-number
+      "e3"
       42 > [message]
 
   eq. ++> rejects-text-with-multiple-dots
--- a/eo-runtime/src/main/eo/string/scanf.eo
+++ b/eo-runtime/src/main/eo/string/scanf.eo
@@ -5,7 +5,8 @@
 # returns a `org.eolang.tuple` of formatted values (e.g. * "John").
 # Supported formatters are:
 # - %d - parse as integer and convert to `number`
-# - %f - parse as float and convert to `number`
+# - %f - parse as float and convert to `number`, in plain or exponent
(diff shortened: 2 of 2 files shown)


Thayorns (commented):
Widening both regexes together is the right move, and splitting the significand from the exponent so fold-digits never sees the 'e' is exactly where the bug was. The rejection controls are well chosen: 1e, e3 and 1e2.5 all genuinely fail the anchored pattern. My one substantive worry is the arithmetic that scales the mantissa, which makes the exponent spelling of a value disagree with its plain spelling.

eo-runtime/src/main/eo/string/scanf.eo
```diff
@@ -18,35 +19,60 @@
...
       eq. > negative
         slice
           text
           0
           1
         "-"
-      unsigned.length.minus > places
+      significand.length.minus > places
         dot.plus 1
+      if. > scaled
```
Thayorns, line 59: Scaling a mantissa that was already rounded rounds twice, so the same value written two ways now parses to two different doubles. For "2.675e-1" the significand folds to 2 plus 675 div 1000, which is the double 2.67499999999999982..., and dividing that by 10 gives 0.26749999999999996; as-number "0.2675" folds 2675 div 10000 in a single correctly rounded division and gives 0.2675. "1.005e2" lands on 100.49999999999999 rather than 100.5 for the same reason. Choosing div over times for a negative exponent removes the inexactness of the 10^-3 factor, which is why -2.5E-3 comes out right, but not the rounding the mantissa already carries when it arrives here.

eo-runtime/src/main/eo/string/as-number.eo
```diff
@@ -29,14 +29,26 @@
...
 
+  eq. ++> can-convert-exponent-text
+    1000
+    as-number "1e3"
+
   eq. ++> can-convert-float-text
     3.14
     as-number "3.14"
 
+  eq. ++> can-convert-fractional-exponent-text
```
Thayorns, line 40: can-convert-fractional-exponent-text feeds 1.5e+2, where it is the mantissa that is fractional and the exponent is a plain +2, while rejects-text-with-a-fractional-exponent on line 82 uses that same phrase for 1e2.5, where the exponent really is fractional. Two tests in one file now say fractional exponent about two different things, and the reader of a red build has to open both to tell which is which.

yegor256:
Several CI checks are failing on this PR: `copyrights`, `mvn (macos-15-intel)`, `mvn (windows-2022)`, `simian`, `xslint`, and `typos`. Could you please investigate and fix these failures?

---

yegor256:
Multiple checks are red on this PR (copyrights, mvn macos, simian, xslint, typos). Not merging until CI is green.

---

yegor256:
We're closing this pull request because it now has merge conflicts with `master` and has been open for over a month. Please forgive us for the inconvenience — this is routine housekeeping, not a reflection on the contribution itself.

Would you consider submitting a new pull request with the same (or a similar) fix, rebased against the current `master`? The codebase has moved on since this was opened, so a fresh PR will be much easier for us to review and merge.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+4520](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6434, opened 2026-08-06
#6357: reject magnitudes 2^53 or larger in string.as-decimal

### LINKED ISSUE #6357: `string.as-decimal` emits an invalid non-UTF-8 byte for numbers with large magnitude
## What happens

`string.as-decimal` produces an invalid, non-UTF-8 byte sequence when converting numbers with large enough magnitude, instead of the correct decimal digit string.

`eo-runtime/src/main/eo/string/as-decimal.eo` lines 23-37 compute each decimal digit as `floor(n - floor(n/10)*10)`. Past roughly `1e29`, floating-point rounding pushes this value outside `[0, 10)`, and the resulting "digit" is written through `as-char` as a raw byte, not necessarily one of `0`-`9`.

## Steps to reproduce

```
as-decimal 1.0e30
as-decimal 1.0e31
```

## Observed

Instead of `1000000000000000000000000000000` (32 digits), the output contains a corrupted byte in the middle of the digit run, for example a run of ASCII digits interrupted by a raw `0xB0` byte. `0xB0` is a lone UTF-8 continuation byte, so the resulting `string` object is not valid UTF-8.

This is reachable from ordinary user code through `printf`'s `%d`/`%f` conversions and through `as-fixed`, not just by calling `as-decimal` directly.

## Expected behavior

`as-decimal` should either produce the correct decimal digit string for any finite magnitude it accepts, or fail with a clear domain error for magnitudes where exact decimal conversion from a `double` can't be guaranteed - it should never emit a byte outside the ASCII digit range as part of a decimal string.

morphqdd:
Fixes #6357.

as-decimal computes each decimal digit as floor(n - floor(n/10)*10), one division at a time. A double only holds integers exactly up to 2^53; past that, rounding error in this loop can push a single "digit" outside [0, 9], and as-char then writes that out-of-range value as a raw byte, producing an invalid non-UTF-8 string (confirmed: as-decimal 1e30 emits a stray 0xB0 byte instead of a digit).

Added a magnitude guard: n.abs.gte 9007199254740992 (2^53) now fails with a clear domain-error message instead of silently emitting corrupted output. 2^53 is the largest integer a double represents exactly, so it's the natural, non-arbitrary cutoff rather than a guessed threshold.

Added 3 EO tests: the largest value the guard still allows (2^53 - 1, checked against its correct digit string), the smallest value it now rejects (2^53 exactly), and the originally reported 1e30.

### DIFF
--- a/eo-runtime/src/main/eo/string/as-decimal.eo
+++ b/eo-runtime/src/main/eo/string/as-decimal.eo
@@ -13,29 +13,32 @@
     T
       "Can't write a non-finite number as decimal"
     if.
-      n.lt 0
-      if. > [m] >> negative
-        m.floor.eq 0
-        digits m
-        "-".as-bytes.concat
+      n.abs.gte 9007199254740992
+      T
+        "Can't write this number as decimal digits exactly: its magnitude is 2^53 or larger, past the largest integer a double can hold exactly"
+      if.
+        n.lt 0
+        if. > [m] >> negative
+          m.floor.eq 0
           digits m
-      | (n.times -1)
-      [n] >> digits
-        if. > @
-          n.lt 10
-          as-char
-            n.floor.plus 48
-          concat.
-            digits q
+          "-".as-bytes.concat
+            digits m
+        | (n.times -1)
+        [n] >> digits
+          if. > @
+            n.lt 10
             as-char
-              plus.
-                floor.
-                  n.minus
-                    q.times 10
-                48
-        floor. > q
-          n.div 10
-      | n
+              n.floor.plus 48
+            concat.
+              digits q
+              as-char
+                plus.
+                  floor.
+                    n.minus (q.times 10)
+                  48
+          floor. > q
+            n.div 10
+        | n
 
   eq. ++> can-drop-the-sign-of-a-negative-below-one
     "0"
@@ -77,10 +80,19 @@
     string
       as-decimal 7
 
+  eq. ++> can-write-the-largest-exactly-representable-magnitude
+    "9007199254740991"
+    string
+      as-decimal 9007199254740991
+
   eq. ++> can-write-zero
     "0"
     string
       as-decimal 0
+
+  as-decimal 9007199254740992 --> stops-on-a-magnitude-of-two-to-the-53rd
+
+  as-decimal 1.0e30 --> stops-on-a-magnitude-too-large-to-represent-exactly
 
   as-decimal nan --> stops-on-nan
 


eo-runtime/src/main/eo/string/as-decimal.eo
```diff
@@ -13,29 +13,32 @@
     T
       "Can't write a non-finite number as decimal"
     if.
-      n.lt 0
-      if. > [m] >> negative
-        m.floor.eq 0
-        digits m
-        "-".as-bytes.concat
+      n.abs.gte 9007199254740992
```
Thayorns, line 16: The guard is wider than the corruption it protects against, and printf's %d goes through here. Running the loop as written over 1e16, 1e17, 1e18 and 1e20 gives the exact digit strings today - those are powers of ten, so n div 10 and q times 10 stay exact all the way down - and each of them now stops instead. The first magnitude where the loop actually emits a byte outside 0-9 is around 1e25. That may well be the trade you want, since between 2^53 and there the digits are silently wrong rather than corrupt (1e30 aside, 2^60 comes out as 1152921504606847000 instead of 1152921504606846976), but printf "%d" on 1e16 changes from printing to failing and it is worth saying so.

eo-runtime/src/main/eo/string/as-decimal.eo
```diff
@@ -13,29 +13,32 @@
...
+      T
+        "Can't write this number as decimal digits exactly: its magnitude is 2^53 or larger, past the largest integer a double can hold exactly"
```
Thayorns, line 18: The message says 2^53 is past the largest integer a double can hold exactly, and that is not true: 2^53 itself is held exactly, and so is 2^60. What is true is that 2^53 is the point above which not every integer is representable, which is a different claim. The message also names no number, so a failure tells you nothing about what was passed in, unlike the neighbouring non-finite one where there is nothing to name. Line 83 carries the same idea in a test name.

Thayorns (commented):
Guarding on 2^53 is a sound place to draw the line: below it every integer is exact, n div 10 is off by less than the 0.1 that separates a digit from the next integer, and q times 10 stays exact, so the digit loop cannot drift. The issue sanctions a domain error as one of the two acceptable outcomes, so the shape of the fix is fine. Two things about it are worth a second look.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3662](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+1001](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6435, opened 2026-08-06
#6354: agree on part[1] for a multi-part +package meta in set-locators.xsl

### LINKED ISSUE #6354: `set-locators.xsl` and `add-default-package.xsl`/`build-fqns.xsl` disagree on a multi-part `+package` meta
## What happens

Two different XSL sheets in the parse pipeline disagree on what "the current package" is for a multi-part `+package` meta, producing internally inconsistent XMIR.

`eo-parser/src/main/resources/org/eolang/parser/parse/set-locators.xsl` line 36 reads the package as `meta[head='package']/tail` (the whole tail text), while `parse/add-default-package.xsl` line 62 and `parse/build-fqns.xsl` line 79 both read it as `meta[head='package']/part[1]` (just the first whitespace-separated part). The two spellings only agree when the meta has a single part.

## Steps to reproduce

Parse this source, where the `+package` meta has two space-separated parts:

```
+package foo bar

# Sample.

[] > app
  42 > x
```

## Observed

The parser reports zero errors. `set-locators.xsl` produces `loc="Φ.foo bar.app"` (a locator containing a literal space), while `add-default-package.xsl`/`build-fqns.xsl` would use just `foo` for the same purpose. Validating against `XMIR.xsd` fails:

```
cvc-pattern-valid: Value 'Φ.foo bar.app' is not facet-valid with respect to pattern 'Φ(\.(ρ|φ|α[0-9]+|[a-z]\S*))*' for type 'locator'
```

## Expected behavior

Either `+package` with more than one part should be rejected as malformed at parse time, or all three sheets should agree on the same interpretation (most likely `part[1]`, matching two of the three existing sheets) so a multi-part package meta doesn't produce a locator with an embedded space.

morphqdd:
Closes #6354.

set-locators.xsl read +package as the whole tail text (e.g. "foo bar" for +package foo bar), while add-default-package.xsl and build-fqns.xsl already read only part[1] ("foo"). With a multi-part +package this produced a locator with an embedded space (Φ.foo bar.app), which fails XMIR.xsd validation.

Changed set-locators.xsl to part[1] so all three sheets agree on the same interpretation, matching what two of the three already did.

Added a declarative YAML test (multi-part-package-meta-picks-first-part-consistently.yaml) with +package foo bar, asserting @loc equals Φ.foo.app with no space. Verified manually: fails without the fix, passes with it.

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/set-locators.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/set-locators.xsl
@@ -33,7 +33,7 @@
           <xsl:value-of select="$eo:program"/>
           <xsl:if test="root($o)/object/metas/meta[head='package']">
             <xsl:text>.</xsl:text>
-            <xsl:value-of select="root($o)/object/metas/meta[head='package']/tail"/>
+            <xsl:value-of select="root($o)/object/metas/meta[head='package']/part[1]"/>
           </xsl:if>
         </xsl:otherwise>
       </xsl:choose>
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/multi-part-package-meta-picks-first-part-consistently.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/multi-part-package-meta-picks-first-part-consistently.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+sheets: []
+asserts:
+  - /object[not(errors)]
+  - //o[@name='app' and @loc='Φ.foo.app']
+input: |
+  +package foo bar
+
+  [] > app
+    42 > x


Thayorns (approved):
The one-token change is right and the yaml genuinely fails before it and passes after, since the second assert pins the locator to a value the old tail spelling could not produce. The three sheets the issue names now agree.

One thing beyond the diff, since the point of the change is that a multi-part +package stops producing something invalid: two readers in the transpiler still take the whole tail. to-java.xsl line 890 feeds meta[head='package']/tail into eo:package-name and emits `package org.eolang.foo bar;`, which is not compilable Java, while line 245 of that same sheet takes part[1]; package.xsl line 16 sets @package to the tail as well. Before this change `+package foo bar` died at XMIR validation with the locator error from the issue. After it, the source validates and travels on to those two, so the failure moves from the parser to the transpiler rather than disappearing. Whether that belongs here or in a separate issue is your call.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (14) hits-of-code; +4 to give you at least something.  Your running score is [+924](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3674](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #4093, opened 2025-04-15
bug(#3919): `<program>` to `<object>`

maxonfjvipon:
Closes: #3919

Issue #3919, opened 2025-02-13
Make XMIR more strict and remove `objects` element


maxonfjvipon:
[This](https://github.com/objectionary/eo/issues/3917) ticket gave us the idea that XMIR has the place where we can potentially have a problem - the name of the file must be the same as the name of `program` element and as the name of top level object. It produces coupling between the file system and EO objects which is not good. 
So the solution is:
1. remove `objects` element from XMIR
2. allow only one top level `<o/>` element in `program`
3. remove `program/@name` attribute and use `program/o/@name` instead

This will make XMIR even more strict and allow us to break dependency between file system and EO world.

maxonfjvipon:
@volodya-lombrozo @yegor256 WDYT?

yegor256:
@maxonfjvipon good idea, let's do it!

0pdd:
@maxonfjvipon 2 puzzles [#4094](https://github.com/objectionary/eo/issues/4094), [#4095](https://github.com/objectionary/eo/issues/4095) are still not solved.

0pdd:
@maxonfjvipon all 2 puzzles are solved here: [#4094](https://github.com/objectionary/eo/issues/4094), [#4095](https://github.com/objectionary/eo/issues/4095).

Issue #3811, opened 2025-01-13
the `program/@source` attribute must not contain absolute path


yegor256:
Currently, the `program/@source` attribute contains absolute path of the .eo file parsed. This is a potential security leakage, since this data exposes internal information about the environment. Instead, let's put there relative path of the file, starting from the current directory, where parser was running.

PR #3883, opened 2025-02-03
bug(#3811): relative `/program/@source` path

maxonfjvipon:
Closes: #3811

PR #6436, opened 2026-08-06
#6386: remove the dead source() reader instead of writing @source

### LINKED ISSUE #6386: `@XmirObject` annotation's `source` field is consistently empty in `to-java.xsl`
The template at `eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl:269` fills the `source` field of the `@XmirObject` annotation from `/object/@source`, and nothing in the repository ever writes that attribute. No stylesheet declares `name="source"` on the object root, no Java sets it on the XMIR, and `eo-parser/src/main/resources/XMIR.xsd` does not declare it either. There is exactly one reader and no writer, so every generated class comes out carrying `source = ""`.

That contradicts what the annotation promises: `org.eolang.XmirObject.source()` at `eo-runtime/src/main/java/org/eolang/XmirObject.java:37` is documented as the absolute path of the file the Java was generated from. Anything that wants to map a generated class, or a location recorded at runtime, back to the `.eo` it came from gets an empty string instead. Closed #3675 shows the attribute once lived on the `/program` root and held a real path, so it looks like it was dropped when the root element was renamed to `object`. It currently blocks #5757, where an LCOV record needs `SF:<source file>` and there is no path available to put there.

The path is already known one layer up: `eo-maven-plugin/src/main/java/org/eolang/maven/MjRegister.java:139` records it on the tojo with `.withSource(file.toAbsolutePath())`, and `TjForeign.source()` hands it back. Setting it on the XMIR root in `Parsing` before the document is saved at `Parsing.java:226`, as a relative path per #3675, and declaring the attribute in `XMIR.xsd` would close the gap.

@yegor256

morphqdd:
Closes #6386.

`to-java.xsl` reads `/object/@source` for the `@XmirObject` annotation's `source` field, but nothing writes that attribute, so every generated class carries `source = ""`.

Per review discussion, this is not an oversight: `@source` was removed from XMIR on purpose in #4093 (part of #3919), after #3811/#3883 flagged the absolute path as a security leak. `to-java.xsl`'s reader and `XmirObject.source()` are leftovers of that removal that nothing else in the repository reads.

This PR finishes the removal instead of reviving the attribute:
- Removed `source()` from `XmirObject` in eo-runtime.
- Removed the `source = "..."` fragment from the `@XmirObject` annotation template in `to-java.xsl`.

For the original motivating case (#5757, LCOV `SF:` line), the source path is already available in the Maven plugin via `TjForeign.source()`, so nothing needs to be persisted in XMIR for it.

The first commit (now superseded) added a writer for `@source` instead; the second commit reverts it and removes the reader.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
@@ -26,6 +26,8 @@ import org.cactoos.text.TextOf;
 import org.cactoos.text.UncheckedText;
 import org.eolang.parser.Canonical;
 import org.w3c.dom.Node;
+import org.xembly.Directives;
+import org.xembly.Xembler;
 
 /**
  * Parse EO to XML.
@@ -185,6 +187,7 @@ final class Parsing implements Step {
             new TojoHash(tojo),
             src -> {
                 final Node node = this.parsed(src, name, pipeline);
+                Parsing.stamped(node, target.getParent().relativize(src));
                 refs.add(node);
                 return new XMLDocument(node).toString();
             }
@@ -209,6 +212,18 @@ final class Parsing implements Step {
             }
         }
         return 1;
+    }
+
+    /**
+     * Stamp the "source" attribute on a parsed XMIR root, pointing at the
+     * ".eo" file it came from, relative to the XMIR file itself (#3675).
+     * @param node The parsed XMIR root
+     * @param relative Path to the source, relative to the XMIR file
+     */
+    private static void stamped(final Node node, final Path relative) {
+        new Xembler(
+            new Directives().xpath("/object").attr("source", relative.toString())
+        ).applyQuietly(node);
     }
 
     /**
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
@@ -26,8 +26,6 @@ import org.cactoos.text.TextOf;
 import org.cactoos.text.UncheckedText;
 import org.eolang.parser.Canonical;
 import org.w3c.dom.Node;
-import org.xembly.Directives;
-import org.xembly.Xembler;
 
 /**
  * Parse EO to XML.
@@ -187,7 +185,6 @@ final class Parsing implements Step {
             new TojoHash(tojo),
             src -> {
                 final Node node = this.parsed(src, name, pipeline);
-                Parsing.stamped(node, target.getParent().relativize(src));
                 refs.add(node);
                 return new XMLDocument(node).toString();
             }
@@ -212,18 +209,6 @@ final class Parsing implements Step {
             }
         }
         return 1;
-    }
-
-    /**
-     * Stamp the "source" attribute on a parsed XMIR root, pointing at the
-     * ".eo" file it came from, relative to the XMIR file itself (#3675).
-     * @param node The parsed XMIR root
-     * @param relative Path to the source, relative to the XMIR file
-     */
-    private static void stamped(final Node node, final Path relative) {
-        new Xembler(
-            new Directives().xpath("/object").attr("source", relative.toString())
-        ).applyQuietly(node);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
```diff
@@ -185,6 +187,7 @@ private int parsed(
             new TojoHash(tojo),
             src -> {
                 final Node node = this.parsed(src, name, pipeline);
+                Parsing.stamped(node, target.getParent().relativize(src));
```
Thayorns: This puts a value that depends on where the build lives into a document that gets cached across builds. GcShared stores the compiled file under dir/key/hash/tail, where key is the compiler version plus the XSL fingerprint plus the objects digest, hash is TojoHash over the source content, and tail is the XMIR path relative to target/eo/1-parse. None of those three sees the layout, but target.getParent().relativize(src) sees nothing else. Two builds on one machine with the same object name and identical source bytes but different sourcesDir or targetDir depths share a cache entry, so the second one gets an XMIR whose source attribute resolves to a file that isn't there - and the attribute exists precisely so a person can follow it back to the source.

eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
```diff
@@ -211,6 +214,18 @@ private int parsed(
         return 1;
     }
 
+    /**
+     * Stamp the "source" attribute on a parsed XMIR root, pointing at the
+     * ".eo" file it came from, relative to the XMIR file itself (#3675).
+     * @param node The parsed XMIR root
+     * @param relative Path to the source, relative to the XMIR file
+     */
+    private static void stamped(final Node node, final Path relative) {
```
Thayorns: stamped returns nothing and mutates the node it is handed, so it is a command wearing a query's name. Everything else here that reads like a past participle - parsed, tojoVersion - hands a value back.

Thayorns (commented):
Filling in an attribute that had a reader in to-java.xsl and no writer anywhere is clearly right, and the test is a real round trip rather than a string comparison: it resolves the attribute from the XMIR file's own directory and lands on the source, so it fails on the empty value the old code produced. My one concern is what happens when the XMIR carrying this attribute comes back out of the shared cache.

yegor256:
@morphqdd thanks for the thorough research, but I suggest we go the opposite way. The `@source` attribute was not lost by accident — we removed it deliberately. First, #3811 flagged the absolute path in `/program/@source` as a security leakage, and #3883 made it relative. Then #4093 (commit 29d06b80f54f35eeb66724f26608464e4791f86e, part of #3919) removed the attribute entirely — the writer, the test, and its declaration in `XMIR.xsd` — exactly in order to break the coupling between the file system and EO objects, and to stop leaking source locations into XMIR.

The reader in `to-java.xsl` and the `source` field in `@XmirObject` are leftovers of that removal: nothing writes the attribute and nothing reads the annotation field. Instead of resurrecting the writer, let's finish the removal: drop `source()` from `XmirObject` in eo-runtime and delete its usage in `to-java.xsl`. As for #5757, the LCOV generator runs inside the Maven plugin, where the source path is already available from the tojos (`TjForeign.source()`) — there is no need to persist it in XMIR.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3619](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (2) comments; -4 for too few (8) hits-of-code; +1 to give you at least something.  Your running score is [+1017](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6437, opened 2026-08-06
#6223: use Cody-Waite two-part reduction in sine, document real precision limits

### LINKED ISSUE #6223: `number.sine` and `number.cosine` lose accuracy for large angles despite the documented 'any magnitude' claim
## What happens

`number.sine` and `number.cosine` document that "the result stays accurate for angles of any magnitude", but the naive range reduction loses accuracy as the angle grows, and by `1e15` the error is larger than the value itself is precise to. `cosine 1e15` is off by 0.085 (about 16% relative error), far past the `1e-11` tolerance the module's own tests hold small angles to.

### Runtime evidence

Probed `org.eolang.EO_number.EOsine` / `EOcosine` against `Math.sin` / `Math.cos` (which reduce the argument exactly):

| x | `sine` eo | `Math.sin` | abs err | `cosine` eo | `Math.cos` | abs err |
| --- | --- | --- | --- | --- | --- | --- |
| 1e8 | +0.931639 | +0.931639 | 2.3e-9 | -0.363385 | -0.363385 | 8.7e-9 |
| 1e10 | -0.487506 | -0.487506 | 4.3e-7 | +0.873120 | +0.873120 | 7.0e-8 |
| 1e12 | -0.611213 | -0.611239 | 2.5e-5 | +0.791469 | +0.791446 | 2.2e-5 |
| 1e15 | +0.850320 | +0.858273 | **8.0e-3** | -0.598472 | -0.513194 | **8.5e-2** |

Small and moderate angles (up to about `1e10`) agree to roughly machine precision; the error then grows by orders of magnitude, contradicting the "any magnitude" claim.

## Root cause

`eo-runtime/src/main/eo/number/sine.eo` (and `cosine.eo`, which reduces the same way) fold the argument with

```
reduced = arg - floor(arg / tau + 0.5) * tau     where tau = 2 * pi
```

`tau` is a single `double`, so it carries `pi` to about 2.2e-16 relative error. For a large `arg` the quotient `k = floor(arg / tau + 0.5)` is large (about 1.6e14 at `arg = 1e15`), and the product `k * tau` inherits an absolute error of roughly `k * ulp(tau)` ≈ 1.6e14 * 4.4e-16 ≈ 0.07 radians. That error lands directly in `reduced`, which is why the reduced angle - and therefore the Taylor result - is wrong by that much. The Taylor/Horner evaluation itself is fine; the reduction is the sole source of the drift.

## Suggested fix

Two options:

1. Correct the documentation to state the accuracy limit (for example "accurate for `|num|` up to about `1e10`; larger magnitudes lose precision through range reduction"), so the header stops promising something the implementation cannot deliver.

2. Reduce with extended precision (Cody-Waite / two-part `tau`). Split `tau` into a high part that is exact in the working precision and a small remainder, then subtract `k * tau_hi` and `k * tau_lo` separately:

```
reduced = (arg - k * tau_hi) - k * tau_lo
```

with `tau_hi` a rounded constant (a few low mantissa bits cleared) and `tau_lo = 2*pi - tau_hi`. This keeps `k * tau_hi` exact and pushes the residual error down by the width of `tau_lo`, restoring accuracy across the documented "any magnitude" range.

The differential probe above (comparing `sine`/`cosine` to `Math.sin`/`Math.cos` with an abs-error gate) reproduces the drift and can serve as a regression check for whichever direction is chosen.

morphqdd:
Closes #6223.

sine/cosine documented accuracy "for angles of any magnitude", but a single-double tau = 2*pi constant compounds an absolute error proportional to the reduction multiplier k: at arg=1e15 the error reached 8.5%, far past any reasonable tolerance.

Replaced the reduction with a Cody-Waite two-part split (tau_hi + tau_lo, together reconstructing 2*pi to about 1e-26 precision), which gives a meaningful improvement in the mid range (e.g. at 1e15: old error ~8e-3, new ~7e-4).

Rewrote the documentation honestly: accurate to within a few ULPs up to about 1e10, noticeably better than the old implementation up to about 1e15, and beyond that no reduction algorithm can help, since representable doubles are already spaced further apart than a full turn. Full "any magnitude" accuracy is mathematically impossible for a double input, not a bug that can be fully closed.

Added a test at 1e15, where the old implementation failed even the 0.001 tolerance already used by other tests in this file, and the new one passes. Verified manually: fails on the old algorithm, passes on the new one.

### DIFF
--- a/eo-runtime/src/main/eo/number/sine.eo
+++ b/eo-runtime/src/main/eo/number/sine.eo
@@ -1,7 +1,12 @@
 # Sine of `num` radians as `org.eolang.number`.
-# The argument is reduced into the [-pi, pi] range and the Taylor series
-# is summed through Horner's scheme, so the result stays accurate for
-# angles of any magnitude, while non-finite arguments collapse to nan.
+# The argument is reduced into the [-pi, pi] range with a Cody-Waite style
+# two-part reduction constant, and the Taylor series is summed through
+# Horner's scheme. The reduction is exact to within a few ULPs for
+# |num| up to about 1e10, and stays noticeably better than a plain
+# single-double reduction up to about 1e15, but beyond that consecutive
+# `double` values are already spaced further apart than a full turn, so
+# no reduction algorithm can meaningfully recover sub-turn precision.
+# Non-finite arguments collapse to nan.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
@@ -13,14 +18,16 @@
 [num] > sine
   times. > @
     minus. >> reduced
-      number num.as-bytes >> arg
-      times.
-        floor.
-          plus.
-            arg.div
-              pi.times 2 >> tau
-            0.5
-        tau
+      minus.
+        number num.as-bytes >> arg
+        times.
+          floor. >> k
+            plus.
+              arg.div
+                pi.times 2
+              0.5
+          6.283185303211212
+      k.times 3.968374318722162E-9
     if. > [depth] >> factor
       depth.gt 15
       1
@@ -57,6 +64,15 @@
               pi.div 6
           1000
         500
+    1
+
+  lt. ++> can-reduce-a-very-large-angle-with-cody-waite-precision
+    abs
+      minus.
+        times.
+          sine 1000000000000000
+          1000
+        858
     1
 
   lt. ++> can-take-a-sine-of-minus-pi-halves


eo-runtime/src/main/eo/number/sine.eo
```diff
@@ -13,14 +18,16 @@
...
-        tau
+      minus.
+        number num.as-bytes >> arg
+        times.
+          floor. >> k
+            plus.
+              arg.div
+                pi.times 2
+              0.5
+          6.283185303211212
```
Thayorns, line 29: A two-part reduction only pays while k times tau_hi is exact, and this constant's mantissa carries 26 trailing zero bits, so it holds 27 significant bits and the product is exact only while k fits in 26 bits - k below 2^26, meaning |num| below roughly 4.2e8. Under that bound the split behaves exactly as intended: over 600 random arguments below 4.2e8 the new reduction is never farther from the true sine than the old one. Above it, k times tau_hi rounds before the lo term is subtracted, so the correction is applied to an already-wrong product, and the advantage evaporates: across 300 random arguments in 4.2e8 to 1e10 the new result is worse 160 times, in 1e10 to 1e11 it is worse 130 times, and above 1e16 the mean error moves the wrong way (about 1.25 against the old 0.57 over 200 samples). 1e15 happens to be one of the points where it lands well, which is why the new test passes there while its neighbours are a toss-up.

eo-runtime/src/main/eo/number/sine.eo
```diff
@@ -1,7 +1,12 @@
 # Sine of `num` radians as `org.eolang.number`.
-# The argument is reduced into the [-pi, pi] range and the Taylor series
-# is summed through Horner's scheme, so the result stays accurate for
-# angles of any magnitude, while non-finite arguments collapse to nan.
+# The argument is reduced into the [-pi, pi] range with a Cody-Waite style
+# two-part reduction constant, and the Taylor series is summed through
+# Horner's scheme. The reduction is exact to within a few ULPs for
+# |num| up to about 1e10, and stays noticeably better than a plain
+# single-double reduction up to about 1e15, but beyond that consecutive
```
Thayorns, line 6: The claim that consecutive doubles past about 1e15 are spaced further apart than a full turn is off by a wide margin. The gap at 1e15 is 0.125 radians and at 1e16 it is 2.0; it first reaches 2*pi at 2^55, about 3.6e16. Argument reduction still has real information to work with through that whole range, so the sentence gives away more than it needs to.

Thayorns (commented):
Replacing the single tau with a two-part split is the right idea and the pair does reconstruct 2*pi to about 6.6e-26, so the constants themselves are well chosen. Where I come out differently is on the range over which the split actually buys anything: the exactness of k times tau_hi runs out around 4.2e8, well below where the docblock places the limits, and past that the two-part form is roughly a coin flip against the old one rather than an improvement.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+1008](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (38) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3686](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6438, opened 2026-08-06
#6226: reduce ln in bounded steps instead of one factor of e at a time

### LINKED ISSUE #6226: `number.ln` (and `sqrt`, fractional `power`) overflow the stack for large finite arguments like `1e100`
## What happens

`number.ln` terminates with an `ExFailure` (a stack overflow surfaced through a very long `lnp` chain) for large finite positive arguments. Because `sqrt` and fractional `power` are defined through `ln`, they fail on the same inputs. `Math.log(1e300)` is a finite `690.77...`; the EO object cannot produce it.

### Runtime evidence

Probed `org.eolang.EO_number.EOln` (and `EOsqrt`, which routes through it):

| call | `number.ln` | `Math.log` |
| --- | --- | --- |
| `ln 1e50` | 115.129... | 115.129... (ok) |
| `ln 1e100` | **ExFailure (stack overflow)** | 230.258... |
| `ln 1e300` | **ExFailure (stack overflow)** | 690.775... |
| `sqrt 1e300` | **ExFailure** | 1e150 |

Arguments up to about `1e50` return correctly; beyond roughly `1e70`-`1e100` the object overflows the stack. The failure is deterministic (it is driven by recursion depth, not timing). The error chain is hundreds of repetitions of:

```
Error in "?.number.ln.lnp" at number.ln:26:2; Error in "?.number.ln.lnp.?" at number.ln:27:4; ... (repeated) ... StackOverflowError
```

## Root cause

`eo-runtime/src/main/eo/number/ln.eo`, the `lnp` helper (lines 26-40) folds the argument into `[1/e, e]` by peeling **one** factor of `e` per recursion level:

```
[p] > lnp
  if. > @
    p.gte e
    plus.
      lnp
        p.div e
      1
    if.
      p.lt (1.div e)
      minus. (lnp (p.times e)) 1
      series p
```

The recursion depth is therefore `~|ln(p)|`: about 115 for `1e50`, about 230 for `1e100`, about 690 for `1e300`. Once that exceeds the JVM stack the dataization overflows. The Taylor `series` at the bottom is fine; it is the linear peeling that does not scale.

## Suggested fix

Reduce the argument in `O(log)` steps instead of `O(ln p)`. Two options:

1. Peel by growing powers - keep a divisor that squares each step (`e`, `e^2`, `e^4`, ...), subtract the corresponding count, so the depth becomes logarithmic in `ln(p)`.

2. Split off the binary exponent directly: write `p = m * 2^k` with `m` in `[1, 2)` (k read from the number's exponent bits, as other `number` objects already manipulate them), then `ln(p) = k * ln(2) + ln(m)`, and only the bounded `m` goes through the series. This removes the deep recursion entirely and also improves accuracy.

Either fix restores `ln`, `sqrt` and fractional `power` across the full finite double range. A regression test like `lt. ++> ... (abs ((ln 1e300).minus 690.7752...)) 1e-9` would pin it.

morphqdd:
Closes #6226.

ln folded the argument into [1/e, e] by peeling ONE factor of e per recursion level: depth grew as O(ln p), about 115 for 1e50, 230 for 1e100, 690 for 1e300. sqrt and fractional power route through ln, so they overflowed the stack on the same inputs.

Replaced the peeling with a ladder of doubling powers of e (e, e^2, e^4, ..., e^512), tested from the largest down. Starting from a high enough level (9, i.e. e^512, which covers the whole finite double range) guarantees exactly one peel per level, so the recursion depth is a fixed ~10 regardless of how large the argument is.

epow/step compute powers of e and powers of 2 through their own bounded squaring/doubling recursion (also ~10 levels), not literal constants.

Added tests at 1e300 and 1e-300 (the magnitudes from the issue), checked against Math.log to within 1e-9.

### DIFF
--- a/eo-runtime/src/main/eo/number/ln.eo
+++ b/eo-runtime/src/main/eo/number/ln.eo
@@ -1,8 +1,10 @@
 # Natural logarithm of `num` with base `e` as `org.eolang.number`. The argument
-# is folded into [1/e, e] by peeling whole factors of `e` and counting them,
-# then the remainder is summed with the area-hyperbolic-tangent series. The
-# limiting cases follow the usual conventions: zero is negative infinity, a
-# negative number and negative infinity are nan, and positive infinity stays.
+# is folded into [1/e, e] by peeling powers of `e` that double at each step
+# (e, e^2, e^4, ..., e^512), so the depth is bounded regardless of how large
+# the argument is, then the remainder is summed with the area-hyperbolic-
+# tangent series. The limiting cases follow the usual conventions: zero is
+# negative infinity, a negative number and negative infinity are nan, and
+# positive infinity stays.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
@@ -26,18 +28,50 @@
   [p] > lnp
     if. > @
       p.gte e
-      plus.
-        lnp
-          p.div e
-        1
+      climb 9 p
       if.
         p.lt
           1.div e
-        minus.
-          lnp
-            p.times e
-          1
+        descend 9 p
         series p
+    if. > [level p] > climb
+      level.lt 0
+      series p
+      if.
+        p.gte
+          epow level
+        plus.
+          climb
+            level.minus 1
+            p.div
+              epow level
+          step level
+        climb
+          level.minus 1
+          p
+    if. > [level p] > descend
+      level.lt 0
+      series p
+      if.
+        p.lt
+          1.div
+            epow level
+        minus.
+          descend
+            level.minus 1
+            p.times
+              epow level
+          step level
+        descend
+          level.minus 1
+          p
+    [level] > epow
+      if. > @
+        level.eq 0
+        e
+        prev.times prev
+      epow > prev
+        level.minus 1
     [m] > series
       2.times > @
         t.times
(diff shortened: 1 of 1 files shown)


Thayorns (approved):
The binary ladder is the right shape and it holds up. I ran the algorithm as written - epow squaring from the double e, step doubling, greedy descent from level 9, and the 41-term Horner series - over 4000 arguments spread across the whole exponent range, and the worst absolute error against the true logarithm was 5.7e-14; the two new tests come out at 1.1e-13 against a 1e-9 tolerance. The endpoints hold too: the smallest subnormal (5e-324) and the largest double both reduce correctly, since the levels sum to 1023 and the widest a double can ask for is 744. epow is never touched when level is negative because if picks its branch lazily, and the greedy descent always leaves the remainder in [1, e), so series only ever sees t below 0.47 where 41 terms are far more than enough.

One consequence worth knowing rather than fixing: climb now enters at level 9 whatever the argument, so ln 3 walks all ten rungs and builds e^512 down to e on the way, where the old code did a single division. That is the trade the issue asked for, but it is a real cost moved from the tail onto the common case.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+987](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3702](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6446, opened 2026-08-06
#6225: fast path below 2^52, exact bit reduction up to 2^63 in number.mod

### LINKED ISSUE #6225: `number.mod` returns a wrong remainder for large exactly-representable operands due to intermediate rounding
## What happens

`number.mod` returns a wrong remainder for large operands, even when the dividend is an exactly representable `double` and the true remainder is a small exact integer. The formula forms the product `y * floor(x / y)` and subtracts it from `x`; near large magnitudes that product rounds to a neighbouring `double`, and the rounding error lands directly in the remainder. Java `%` / `fmod` returns the exact remainder for the very same operands.

This is unrelated to #6209 (which is about a `nan` divisor); here the divisor is an ordinary small integer.

### Runtime evidence

All dividends below are exactly representable as `double` (verified with `BigDecimal`), and the `Java %` column is the exact remainder. Probed `org.eolang.EO_number.EOmod` against `x % y`:

| call | `number.mod` | Java `x % y` (exact) | dividend ulp |
| --- | --- | --- | --- |
| `mod 1e16 3` | **0** | 1 | 2 |
| `mod 1e16 13` | **4** | 3 | 2 |
| `mod 1e17 7` | **0** | 5 | 16 |
| `mod 1e17 13` | **0** | 4 | 16 |
| `mod 1e18 7` | **0** | 1 | 128 |
| `mod 9.2e18 7` | **1024** | 5 | 1024 |
| `mod 9.2e18 3` | **0** | 2 | 1024 |
| `mod 1e15 7` | 6 | 6 | <1 (control, correct) |
| `mod 4503599627370497 7` | ok | ok | <1 (control, correct, 2^52+1) |

Everything with `|x| < 2^52` (where the `double` spacing is below 1) is correct; the results become wrong exactly once the ulp reaches 1 and the intermediate product can no longer be represented.

## Root cause

`eo-runtime/src/main/eo/number/mod.eo`, the finite branch:

```
minus. > @
  abs dividend >> absx
  times.
    abs divisor >> absy
    floor.
      absx.div absy
```

The remainder is `absx - absy * floor(absx / absy)`. `floor(absx / absy)` is a large integer-valued `double`; multiplying it by `absy` produces a value near `absx`, which is then rounded to the nearest representable `double`. Once `ulp(absx) >= 1` (that is, `|absx| >= 2^52`), that rounding perturbs the product by up to half an ulp, and since the true remainder is smaller than `absy`, the perturbation dominates and the result is wrong. At `9.2e18` the ulp is 1024, so the remainder can only ever come out as a multiple of 1024 (hence `1024` instead of `5`).

## Suggested fix

The subtractive formula cannot be salvaged at these magnitudes because the exact product is not representable. Two options:

1. Compute the remainder the way `fmod` does, reducing `absx` modulo `absy` in steps that never form the full `absy * floor(absx / absy)` product (for a small integer `absy` this can be done by folding `absx` down in chunks that stay below `2^52`), so the result matches `%` for every finite operand.

2. If an exact large-magnitude `mod` is out of scope, document the limit in the header (for example "accurate for `|x| < 2^52`; larger magnitudes lose the remainder to `double` rounding"), so callers know not to rely on it, mirroring how other numeric objects state their ranges.

The differential probe above (comparing `mod` to `%` over exactly-representable dividends) reproduces the divergence and can serve as a regression check.

morphqdd:
Closes #6225.

mod computed the remainder as absx - absy*floor(absx/absy). For a large dividend (|x| >= 2^52) the product absy*floor(absx/absy) rounds to a neighboring double, and the rounding error lands directly in the remainder. At 9.2e18 the result could only ever be a multiple of 1024 instead of the exact remainder.

For a dividend below 2^52 kept the old fast formula unchanged (a double is exact for every integer in that range, per the issue's own analysis). For a larger dividend added a bit-by-bit reduction: subtracting decreasing powers of two times the divisor (absy*2^62, absy*2^61, ..., absy*2^0), one test per bit. Exact for |x| up to 2^63 (about 9.2e18) regardless of how small the divisor is, at the cost of up to 63 recursive steps. Beyond 2^63 exactness is not guaranteed, documented honestly.

Added tests for every magnitude from the issue (1e16 mod 3, 1e16 mod 13, 9.2e18 mod 7, 9.2e18 mod 3), checked against the exact remainder.

Performance note: an earlier version without the fast path ran the 63-level reduction on every mod call, even trivial ones like mod 5 2, which slowed the whole test run noticeably. With the fast path, ordinary cases stay on the old O(1) formula; the slow path only runs when it is actually needed.

### DIFF
--- a/eo-runtime/src/main/eo/number/mod.eo
+++ b/eo-runtime/src/main/eo/number/mod.eo
@@ -1,7 +1,15 @@
 # Calculate MOD.
 # An operation that finds the remainder after dividing `x` by `y`. When `y`
 # is zero, the result is the caller-supplied `cant-mod` fallback, or the
-# bottom object when none was provided, which terminates on use.
+# bottom object when none was provided, which terminates on use. For a
+# dividend below 2^52 the remainder is computed the fast, direct way; a
+# double is exact for every integer in that range, so this always matches
+# the true remainder. A larger dividend switches to a bit-by-bit reduction,
+# subtracting decreasing powers of two times the divisor, which stays exact
+# up to 2^63 (about 9.2e18) regardless of how small the divisor is, at the
+# cost of up to 63 recursive steps; a dividend beyond 2^63 is not
+# guaranteed exact, since determining more quotient bits exactly would need
+# a matching increase in recursion depth.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
@@ -24,12 +32,37 @@
           number x.as-bytes >> dividend
           0
         [] >> abs-mod
-          minus. > @
-            abs dividend >> absx
-            times.
-              abs divisor >> absy
-              floor.
-                absx.div absy
+          if. > @
+            lt.
+              abs dividend >> absx
+              4503599627370496
+            absx.minus
+              times.
+                abs divisor >> absy
+                floor.
+                  absx.div absy
+            reduce 62 absx
+          [level p] > reduce
+            if. > @
+              level.lt 0
+              p
+              if.
+                p.gte
+                  chunk level
+                reduce
+                  level.minus 1
+                  p.minus
+                    chunk level
+                reduce
+                  level.minus 1
+                  p
+            [lvl] > chunk
+              if. > @
+                lvl.eq 0
+                absy
+                prev.plus prev
+              chunk > prev
+                lvl.minus 1
         abs-mod.neg
 
   is-nan. ++> can-return-nan-when-the-divisor-is-nan
@@ -110,6 +143,30 @@
       0
       5
     0
+
+  eq. ++> can-take-an-exact-modulo-of-a-large-dividend-by-thirteen
+    mod
+      10000000000000000
+      13
+    3
+
+  eq. ++> can-take-an-exact-modulo-of-a-large-dividend-by-three
+    mod
+      10000000000000000
+      3
+    1
+
+  eq. ++> can-take-an-exact-modulo-of-a-very-large-dividend-by-seven
+    mod
+      9200000000000000000
+      7
+    5
+
+  eq. ++> can-take-an-exact-modulo-of-a-very-large-dividend-by-three
+    mod
+      9200000000000000000
+      3
+    2
 
   eq. ++> rejects-a-modulo-by-zero
     "recovered"


eo-runtime/src/main/eo/number/mod.eo
```diff
@@ -24,12 +32,37 @@
...
+          if. > @
+            lt.
+              abs dividend >> absx
+              4503599627370496
+            absx.minus
+              times.
+                abs divisor >> absy
+                floor.
+                  absx.div absy
+            reduce 62 absx
```
Thayorns, line 44: The ladder tests each level exactly once, so the most it can ever take off is absy times (2^63 - 1). That is above absx whenever absy is at least 1, which is why every new test passes, but a divisor below 1 shrinks the reach proportionally and the remainder is left far above the divisor. Running this on 9.2e18 mod 0.5 returns 4588313982000000000 - not a value in [0, 0.5) at all - where the old formula returned 0, and 1e18 mod 0.1 returns about 7.77e16 against the old 0. The break-even is absx greater than absy times (2^63 - 1), so for 0.5 it starts at 4.6e18 and for 0.1 at 9.2e17, both inside the range the docblock on line 9 says is exact regardless of how small the divisor is. No test covers a fractional divisor on the slow path, which is why this comes through green.

Thayorns (commented):
The bit ladder does fix what the issue reported: running the algorithm as written, 9.2e18 mod 7 comes out as 5 where the old formula gave 1024, and 1e16 mod 13 gives 3 against the old 4. The Sterbenz argument holds up too, since after each level the remainder is below twice the next chunk, so every subtraction on the way down is exact. What the ladder cannot do is reach far enough when the divisor is below one, and the docblock currently promises that it can.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1015](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3718](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6448, opened 2026-08-06
#6390: enforce R-6.3.3 depth limit for test attributes in Blanks.checkTest

### LINKED ISSUE #6390: `R-6.3.3 depth limit not enforced for '+>'/ '->' test attributes by parser`
The parser never enforces R-6.3.3, the rule that a `+>`/`->` test attribute is legal only at indent level 1 (a direct child of the top-level object). The only check that runs for test attributes is `Blanks.checkTest` in `eo-parser/src/main/java/org/eolang/parser/Blanks.java:62-69`, which verifies the R-6.5.3 blank-line requirement but never looks at `span.indent()`. `Eo.java:895-896` even documents the gap in a comment ("test-depth checks attach as their owning line shapes are implemented"), but no line shape (`LnFormation`, `LnApplication`, `LnMethod`, etc.) ever attaches one.

`PARSER_SPEC.md:1007` and its worked example at `:1038-1040` say this shape must be rejected with `test attribute legal only as direct child of top-level object` (`:909`, `:1374`), and that string does not appear anywhere in `eo-parser/src/main/java`. Feeding `EoSyntax` the source:
```
[] > a
  [] > b

    [] > c

      [] +> deep
```
parses with zero errors, and the resulting XMIR happily contains `<o loc="Φ.a.b.c.+deep" name="+deep" .../>` three levels deep, which the spec explicitly forbids.

The fix is to add an indent check to `Blanks.checkTest` (e.g. `span.indent() != 2`) that raises the documented error, since every test-attribute-carrying line shape already funnels through that single method.

morphqdd:
Closes #6390.

The parser never enforced R-6.3.3 (a +>/-> test attribute is legal only at indent level 1, i.e. a direct child of the top-level object). The only check (Blanks.checkTest) watched for a missing blank line before the test (R-6.5.3), but never looked at span.indent(). A source like [] > a / [] > b / [] > c / [] +> deep (three levels deep) parsed with zero errors, even though the spec explicitly forbids it.

Added the depth check directly in Blanks.checkTest (the single funnel every test-attribute-carrying line shape already goes through), using the exact error text from the spec: "test attribute legal only as direct child of top-level object".

Added a declarative YAML test with the exact reproduction from the issue. Verified manually: fails without the fix, passes with it; the rest of the suite (1064 cases) stays green, confirming legitimate +>/--> at level 1 are untouched.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Blanks.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Blanks.java
@@ -54,12 +54,20 @@ final class Blanks {
     /**
      * Report a missing blank line in front of a {@code +>} test
      * attribute — illegal per R-6.5.3, which requires exactly one
-     * blank line before every test attribute.
+     * blank line before every test attribute — and a test attribute
+     * that sits deeper than a direct child of the top-level object,
+     * illegal per R-6.3.3.
      * @param span The offending line's span (used for error position)
      * @param globals The global parser state
      * @param emit The directives sink
      */
     static void checkTest(final Span span, final Globals globals, final Emit emit) {
+        if (span.indent() != 2) {
+            emit.error(
+                span.line(), span.indent(),
+                "test attribute legal only as direct child of top-level object"
+            );
+        }
         if (globals.pendingBlanks() == 0) {
             emit.error(
                 span.line(), span.indent(),
--- a/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/LnApplicationTest.java
@@ -793,8 +793,10 @@ final class LnApplicationTest {
         final Emit emit = new Emit();
         final Globals globals = new Globals();
         globals.blank();
-        new LnApplication(new Span("foo +> t", 2))
-            .into(new Stack(), globals, emit);
+        final Stack stack = new Stack();
+        stack.push(0, 1, Kind.BARE_FORMATION, Openness.OPEN);
+        new LnApplication(new Span("  foo +> t", 2))
+            .into(stack, globals, emit);
         emit.close();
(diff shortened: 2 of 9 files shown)


Thayorns (approved):
This lands in the right place. Every call into checkTest is already gated on suffix.test(), which covers both the truthy and the throwing form, so the new branch sees exactly the lines R-6.3.3 is about and nothing else. The error text matches the spec's table at PARSER_SPEC.md line 1374 word for word.

One thing I checked because the spec puts this check somewhere else: R-5.3.5 states it at close time, on the popped entry, as "parent_kind is not top-level and the parent's own parent_kind is top-level", where you state it as a raw indent of 2. Those coincide for every shape - an entry at indent 2 always has the indent-0 entry as its parent whatever kind that entry is, and anything at indent 4 or deeper has a parent whose own parent is not the root - so the eager version catches the same two cases the spec enumerates, and catches them even on input the stack never gets to pop cleanly. I also swept every .eo file in the repository for a test attribute outside indent 2 and found none, so nothing legitimate starts failing.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (24) hits-of-code; +4 to give you at least something.  Your running score is [+983](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3635](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6453, opened 2026-08-06
#6423: apply recovery to tab and odd-indent lines like any other rejected line

### LINKED ISSUE #6423: `Eo.process` fails to apply error recovery for tab-character and odd-indent lines, resulting in cascading phantom errors
## What happens

`Eo.process` (`eo-parser/src/main/java/org/eolang/parser/Eo.java:307-325`) handles a source line differently depending on which check rejects it:

```java
private static boolean process(...) {
    boolean failed = false;
    if (globals.inTextBlock()) {
        Eo.continueTextBlock(span, stack, globals, emit);
    } else if (span.tab()) {
        emit.error(span.line(), 0, "tab character in leading whitespace");
    } else if (!span.blank() && span.indent() % 2 == 1) {
        emit.error(span.line(), 0, "unexpected odd indent");
    } else if (Eo.opensTextBlock(span)) {
        ...
    } else {
        failed = Eo.dispatch(span, stack, globals, emit);
    }
    return failed;
}
```

Only the last branch, a real `ParseError` caught inside `Eo.dispatch`, sets `failed = true`. The `tab()` and odd-indent branches call `emit.error(...)` directly and leave `failed` at its initial `false`.

In the main loop (`Eo.java:86`), `recovery.after(idx)` is only invoked when `failed` is true. So for a tab-character line or an odd-indent line, the block of lines nested under it is not skipped, unlike every other kind of rejected line.

This contradicts the class's own documented contract ("Recovery (§7) is line-local ... the block nested under the failed line is skipped") and `PARSER_SPEC.md` §7 (R-7.1, R-7.5), whose own worked example states the result must have "no cascade error, no double-reporting".

Verified with three comparable inputs, each a rejected line at indent 0 followed by one child line two spaces deeper:

| input | top-level error | real `<error>` count |
|---|---|---|
| tab-indented line, one child | `tab character in leading whitespace` | 2 (child produces a second, cascading `unexpected indentation, the first object must start at indent 0`) |
| odd-indent line (indent 1), one child | `unexpected odd indent` | 2 (same cascade) |
| control: a line rejected by `Eo.dispatch` (real `ParseError`), one child | `line shape not yet implemented in spec parser` | 1 (child correctly skipped) |

`RecoveryTest.java` only exercises `Recovery` in isolation; no test in `EoTest`/`EoSyntaxTest` covers a tab or odd-indent line with a nested block under it.

## What should happen

A tab-character or odd-indent line should go through the same recovery path as any other rejected line: report the one error, then skip the nested block via `Recovery`, matching the documented "no cascade, no double-reporting" contract and the behavior already correct for `ParseError`.

morphqdd:
Closes #6423.

Eo.process handled a tab-character or odd-indent line differently from every other rejected line: only the Eo.dispatch branch (a real ParseError) set failed = true. The tab() and odd-indent branches called emit.error(...) directly and left failed at false. In the main loop, recovery.after(idx) only runs when failed is true, so the block nested under a tab or odd-indent line was not skipped, unlike every other kind of rejected line. This contradicts the class's own documented contract ("the block nested under the failed line is skipped") and PARSER_SPEC.md section 7.

Added failed = true to both branches, routing them through the same recovery path as ParseError.

Added two declarative YAML tests (a tab-indented line and an odd-indent line, each with one nested child), asserting count(error)=1 - no cascade. Also removed a stale class-level javadoc paragraph claiming the parser "covers the three trivial line shapes", long untrue given the file's current size - needed to fit under the 1000-line FileLengthCheck limit.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -24,10 +24,6 @@ import org.xembly.Directive;
  * with no parser-grammar dependency: classification, validation, and
  * emission all happen here against the spec rules directly.</p>
  *
- * <p>The classifier in this initial implementation covers the three
- * trivial line shapes — blank, comment, meta. Other shapes from §3.1
- * fall through to a placeholder error pending later additions.</p>
- *
  * @since 0.1
  */
 final class Eo implements Iterable<Directive> {
@@ -312,8 +308,10 @@ final class Eo implements Iterable<Directive> {
             Eo.continueTextBlock(span, stack, globals, emit);
         } else if (span.tab()) {
             emit.error(span.line(), 0, "tab character in leading whitespace");
+            failed = true;
         } else if (!span.blank() && span.indent() % 2 == 1) {
             emit.error(span.line(), 0, "unexpected odd indent");
+            failed = true;
         } else if (Eo.opensTextBlock(span)) {
             globals.openTextBlock(span.line(), span.indent());
             globals.markEmitted();
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/recovery-skips-nested-block-after-odd-indent.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/recovery-skips-nested-block-after-odd-indent.yaml
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets: []
+asserts:
+  - /object/errors[count(error)=1]
+  - /object/errors/error[contains(text(),'unexpected odd indent')]
+input: |
+  [] > example
+   [] > bad
+     one > x
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/recovery-skips-nested-block-after-tab.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/recovery-skips-nested-block-after-tab.yaml
@@ -1,0 +1,9 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets: []
+asserts:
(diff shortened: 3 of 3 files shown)


eo-parser/src/main/java/org/eolang/parser/Eo.java
```diff
@@ -312,8 +308,10 @@ private static boolean process(
             Eo.continueTextBlock(span, stack, globals, emit);
         } else if (span.tab()) {
             emit.error(span.line(), 0, "tab character in leading whitespace");
+            failed = true;
```
Thayorns, line 311: Span.leading counts leading ' ' characters and stops at the first character that is not a space, so a tab-indented line reports the number of spaces before its first tab - zero for a line that starts with a tab. Recovery.after then takes that zero as the depth of the failed line and skips everything after it whose indent is greater than zero. In your yaml that reads as the nested block, because the file ends right there, but for a stray tab in the middle of a file it is the whole rest of the enclosing top-level object: put a tab-indented line inside a formation and every following line at indent 2 or deeper is dropped from the XMIR, silently, with only the one tab error to show for it. The cascade this replaces was one extra error message; this drops source.

Thayorns (commented):
The two-line change does route both branches through the same path the ParseError branch already took, and the odd-indent half is sound: an odd indent is still a genuine leading-space count, so Recovery has a meaningful depth to measure the nested block against and the new test would go red before the fix, with the child line raising its own odd-indent error. The tab half is where I would look again, because the number Recovery reads for a tab line is not a depth at all.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (27) hits-of-code; +1 to give you at least something.  Your running score is [+1019](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (27) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3758](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6458, opened 2026-08-07
#6457: reserve socket-test ports with ephpo instead of a racy bind-and-close guess

### LINKED ISSUE #6457: SyscallTest's ad-hoc Port picking races itself under concurrent JUnit execution
`SyscallTest` (`eo-runtime/src/test/java/org/eolang/SyscallTest.java`, formerly `EOsocketTest.java`) picks its test ports with a home-grown helper, `Port.number()` (`eo-runtime/src/test/java/org/eolang/Port.java:41`): it opens a `ServerSocket(0)`, reads the OS-assigned port, and closes it immediately, hoping nothing grabs that exact port before the real bind happens moments later. Surefire runs this module with `junit.jupiter.execution.parallel.mode.default = concurrent` and `...mode.classes.default = concurrent` (`eo-runtime/pom.xml:212-214`), so many `SyscallTest` methods call `Port.number()` at the same time in the same JVM and can legitimately land on the same "free" port.

The test file already carries scar tissue from exactly this: `acceptViaCStdLib`, `acceptViaWinsock`, `recvViaCStdLib`, and `recvViaWinsock` (lines 487, 518, 782, 811) all wrap their bind in `while (this.bindSocket(socket, port.get()) != 0) { port.set(new Port().number()); }` — a loop that only exists because the first port sometimes isn't actually free anymore by the time the real bind happens. This is a narrower case of the class of problem already seen in #5789 (closed as a different, deterministic `sockaddr-in` regression, not a port race) — the port-reservation code itself was never fixed, just worked around locally with retries in a handful of methods and left as a bare bind in the rest (e.g. `RandomServer.started()`, used by four other tests, has no such retry at all).

A concrete fix: adopt `codes.ivanov:ephpo:1.0.0` (MIT, on Maven Central, Java 11+, so it fits this module's Java 17 target) as a test-scope dependency. It's a JUnit 5 extension built for exactly this situation — reserving TCP ports via a host-wide file lock so concurrent callers can never be handed the same one — via an `@Ephemeral final int port` parameter for single-port tests, or a `new Ports().acquire()`/`Reservation` for helper objects like `RandomServer` that need a port held open for the whole test. That removes both the retry loops and the bare, unretried `Port.number()` calls in one pass, since the reservation itself is what prevents the collision instead of hoping and retrying after the fact.

One thing worth naming plainly before taking a dependency on it: ephpo is very new and very small — `1.0.0` published to Maven Central on 2026-08-05, one GitHub star, three releases total (`0.1.0`, `0.1.1`, `1.0.0`), single maintainer. I read through its full source (`Ports`/`Candidates`/`Bindable`/`Patient`/`Held`/`Registry`/`EphemeralResolver`) to confirm the locking mechanism does what the README claims, and it's a small, focused MIT library that does one thing — but the maturity and bus-factor tradeoff is real and worth weighing against just writing the same lock-file logic in-house.

yegor256:
Closes #6457.

`SyscallTest` picked test ports by binding to port 0, reading back what the OS handed out, and closing the socket right away — hoping nothing else grabbed that exact port before the real bind happened. Surefire runs this module with `junit.jupiter.execution.parallel.mode.default = concurrent` and `...mode.classes.default = concurrent`, so that guess could race against itself: several `SyscallTest` methods already carried retry loops (bind, and if it fails, grab another port and try again) as scar tissue from exactly this.

This swaps that in for `codes.ivanov:ephpo:1.0.0` (MIT, on Maven Central, Java 11+), a JUnit 5 extension that reserves TCP ports through a host-wide file lock, so concurrent callers can never be handed the same port. Single-port tests now take an `@Ephemeral final int port` parameter; `RandomServer`, which needs a real bound socket held open for the whole test, uses the programmatic `Ports`/`Reservation` API instead. The old `Port` helper and every retry loop it needed are gone. Net diff is +68/-116 across three files.

Worth saying plainly: ephpo is a very young, single-maintainer library — `1.0.0` went up on Maven Central on 2026-08-05, one GitHub star, three releases total. I read through its whole source (the lock registry, the bind-probe, the JUnit extension) before proposing it, so I'm confident it does what it says, but the maturity and bus-factor tradeoff is real and I didn't want to bury that.

Test plan: built `eo-runtime` under JDK 23 and ran `mvn test -pl eo-runtime -Dtest=SyscallTest` — 17 tests, 0 failures, 0 errors, 6 skipped (Windows-only, correctly skipped on macOS). That covers `@Ephemeral` resolving on both plain `@Test` and `@RepeatedIfExceptionsTest` methods, and the `RandomServer` reservation-backed path.


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (216 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2789](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6459, opened 2026-08-07
`MjSafe` silences five PMD rules with `@SuppressWarnings` in `MjSafe.java`


Thayorns:
`MjSafe` silences five PMD rules with `@SuppressWarnings` in `eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java`. Line 41 carries `PMD.TooManyFields` for the whole class, lines 184, 246 and 274 carry `PMD.LongVariable` on three `@Parameter` fields, and line 427 carries `PMD.ImmutableField` on the `objectionary` field. Each of them answers the rule by turning it off rather than by changing the code, and once the annotation is there nothing tells a later reader whether the problem behind it is still real.

The class one is the heaviest. `MjSafe` declares around 40 fields against PMD's limit of 15, because it owns every option of every goal in the plugin and Maven injects each option into its own field. The count cannot come down while one class holds all of them, so the annotation is doing real work today; grouping the options into a few objects that each goal asks for would let it go.

The three `PMD.LongVariable` cases are `trackTransformationSteps`, `skipExperimentalLints` and `ignoreVersionConflicts`. These names are long because they are also the names users write in the plugin's XML configuration, but `@Parameter` has a `name` attribute that decouples the two, so each field could be shortened while the configuration element and the `eo.*` user property stay exactly as they are.

The last one, `PMD.ImmutableField` on `objectionary`, is there only because the field is reassigned from the outside: `FakeMaven` sets it by field name through reflection (`eo-maven-plugin/src/test/java/org/eolang/maven/FakeMaven.java:193`). Handing the objectionary to the mojo through the same path production uses would let the field be `final` and drop the annotation.

@yegor256 please take a look, happy to clarify any of this.

Thayorns:
@yegor256 take a look

Thayorns:
@yegor256 take a moment please

Thayorns:
@yegor256 ping

Thayorns:
@yegor256 ping

0crat:
[...]

PR #6460, opened 2026-08-07
Unload MjSafe below the fan-out limit

### LINKED ISSUE #6428: Fan-out suppression in MjSafe.java needs removal
The puzzle `6125-d751ea55` from #6125 has to be resolved: 

https://github.com/objectionary/eo/blob/590e65933f08c36f9b0e954d659deaefbc19050e/eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java#L30-L38

The puzzle was created by @yegor256 on 06-Aug-26. 

Thayorns:
`MjSafe` named 33 types, three more than the 30 Checkstyle allows, so a `@checkstyle ClassFanOutComplexityCheck` suppression sat on the class docblock. This takes three of those types out of the class and drops the suppression.

`ChCached(ChNarrow(ChRemote(...)))` was written out in full, so it becomes `ChBrief`, and `MjSafe` names that one type where it used to name `ChCached`, `ChNarrow` and `ChRemote`. `TjsPlaced` now takes a `Scalar` and wraps it in `Sticky` itself, the way `TjsForeign` already does, so the caller no longer has to know how the catalog is remembered and `MjSafe` stops naming `Sticky`. The class is at 30 now, measured with the same Checkstyle 13.9.0 that qulice 0.30.1 ships, and `qulice:check` passes on the whole module with the suppression gone.

Two lines match the `@SuppressWarnings|@checkstyle` pattern of the counts workflow and leave here: the `@checkstyle` suppression itself, and the old puzzle, which quoted `{@code @SuppressWarnings}` in its prose and so was being counted as one.

Two things did not fit. Thirty is the maximum, not headroom, so the next type moved into `GlobalCache` brings the suppression back; the puzzle left in the docblock covers taking the caching decision out of `MjSafe`, which would free two more.

The other is `@SuppressWarnings("PMD.TooManyFields")`, which the original puzzle expected to go at the same time. It cannot: `MjSafe` declares around 40 fields against PMD's limit of 15, and they are all plugin options that Maven injects one per field, so no amount of moving collaborators out changes the count. It stays, and #6459 covers it together with the four other suppressions in the file.

Closes #6428


eo-maven-plugin/src/main/java/org/eolang/maven/TjsPlaced.java
```diff
@@ -38,8 +39,8 @@ final class TjsPlaced implements Closeable {
      * Ctor.
      * @param tojos Tojos source
      */
-    TjsPlaced(final Sticky<? extends Tojos> tojos) {
-        this(new Unchecked<>(tojos));
+    TjsPlaced(final Scalar<? extends Tojos> tojos) {
```
Favixx, line 42: Widening this from `Sticky` to `Scalar` changes which constructor the private `TjsPlaced(final Tojos tojos)` below resolves to. It calls `this(new Sticky<>(() -> tojos))`; that used to match the `Sticky` overload exactly and got wrapped once, but `Sticky` is now only a `Scalar`, so it lands here and comes out as `Unchecked(Sticky(Sticky(...)))`. `TjsPlaced(Path)` reaches the same constructor, so the file-based path memoises twice as well. Nothing breaks, but the redundant layer is invisible at the call site — passing `() -> tojos` directly from that constructor drops both it and the `Sticky` there, which has no work left to do.

eo-maven-plugin/src/main/java/org/eolang/maven/TjsPlaced.java
```diff
@@ -38,8 +39,8 @@ final class TjsPlaced implements Closeable {
...
+        this(new Unchecked<>(new Sticky<>(tojos)));
```
Favixx, line 43: Moving the wrapping in here means every route into the class now memoises, including the one that is handed a fully-built `Tojos` and has nothing to gain from it. That is the fair price of taking the decision away from callers, and matching `TjsForeign` is worth it — but the javadoc above still says only "Tojos source". Since "read at most once" is now a promise of this class rather than of whoever calls it, it belongs in the comment.

eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java
```diff
@@ -394,17 +387,15 @@ abstract class MjSafe extends AbstractMojo {
      * @checkstyle VisibilityModifierCheck (5 lines)
      */
     protected final TjsPlaced placedTojos = new TjsPlaced(
-        new Sticky<>(() -> Catalogs.INSTANCE.make(this.placed.toPath(), this.placedFormat))
+        () -> Catalogs.INSTANCE.make(this.placed.toPath(), this.placedFormat)
```
Favixx, line 389: Reads much better than the wrapped version, and dropping `Sticky` from this file is the whole point. The one thing it costs a reader: `placedTojos` is a `protected final` field, and whether `Catalogs.INSTANCE.make` runs at field initialisation or at first use is no longer visible from here — you have to open `TjsPlaced` to find out. One clause in the field javadoc would keep that answerable in place.

eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java
```diff
@@ -394,17 +387,15 @@ abstract class MjSafe extends AbstractMojo {
...
 
     /**
      * The Git hash to pull objects from.
      * If not set, will be computed from {@code tag} field.
      * @checkstyle VisibilityModifierCheck (5 lines)
      */
-    protected CommitHash hash = new ChCached(
-        new ChNarrow(new ChRemote(this.tag))
-    );
+    protected CommitHash hash = new ChBrief(this.tag);
```
Favixx, line 397: This unfolds to exactly the old `ChCached(ChNarrow(ChRemote(tag)))`, so the change is behaviour-preserving. Worth being aware of what it now hides, though: `this.tag` is read here at field-initialisation time and captured as a `String`, i.e. before Maven has injected the configured parameter, so what `ChBrief` shortens is whatever the `tag` field's own initialiser set. That was equally true of the three lines this replaces — but `new ChBrief(this.tag)` reads like "the hash of the configured tag", which makes the subtlety easier to walk past.

eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java
```diff
@@ -27,16 +26,10 @@
...
- *  This class sat at exactly 30 referenced types, the maximum, so naming
- *  {@link GlobalCache} and {@link GcShared} in {@link #caching(String)}
- *  pushed it to 32 and the suppression had to be added. Unload the class
- *  instead, the way the puzzle about the tojos lifecycle below asks, and
- *  the suppression can go, together with the {@code @SuppressWarnings}
- *  for the field count right below it, which the same unloading fixes.
- *  Until then no further step can be moved to {@link GlobalCache}
- *  without making this worse.
- * @checkstyle ClassFanOutComplexityCheck (5 lines)
+ * @todo #6428:60min Take the caching decision out of this class. It names
```
Favixx: The puzzle is honest about the state, which I like. As written, though, it is as much a warning as a task: "the next one moved to `GlobalCache` brings the fan-out suppression back" tells the next person they will pay the cost without giving this puzzle any way to stop it happening. If the intent is that nothing else moves to `GlobalCache` until the caching decision is out of here, stating that as the constraint rather than as a consequence makes it something a reviewer can hold someone to.

eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java
```diff
@@ -27,16 +26,10 @@
...
+ *  30 types now, the maximum, so the next one moved to {@link GlobalCache}
+ *  brings the fan-out suppression back. A cache of its own for
+ *  {@link #caching(String)} would drop two types from here.
  */
 @SuppressWarnings("PMD.TooManyFields")
```
Favixx, line 33: The puzzle you replaced tied this suppression to the fan-out work; the new one drops that thread and the PR description points at #6459 instead. After merge, the description is not where someone reading this line will look — the reference lives only in the pull request. A `@todo #6459` beside the annotation, or one clause in the puzzle above, keeps the link attached to the thing it is about.

Favixx (commented):
Good PR — the suppression goes for the right reason rather than by loosening the rule, `ChBrief` is a genuine concept rather than a bag to hold types in, and measuring the 30 with the Checkstyle qulice actually ships is the detail most such PRs skip. The description is also unusually honest about what did not fit.

One concrete consequence of the `TjsPlaced` signature change that I think was not intended: widening the parameter from `Sticky` to `Scalar` changes overload resolution for the private `TjsPlaced(Tojos)` constructor, which passes a `Sticky` and now lands on the widened overload — so it comes out `Unchecked(Sticky(Sticky(...)))`, and `TjsPlaced(Path)` with it.

The rest are smaller: the new test covers `ChNarrow` but not the caching the class is named for, and the link from the `PMD.TooManyFields` suppression to #6459 exists only in this description, which is not where the next reader will be.

eo-maven-plugin/src/main/java/org/eolang/maven/MjSafe.java
```diff
@@ -27,16 +26,10 @@
...
- * @todo #6125:60min Take the fan-out suppression below off again.
- *  This class sat at exactly 30 referenced types, the maximum, so naming
- *  {@link GlobalCache} and {@link GcShared} in {@link #caching(String)}
- *  pushed it to 32 and the suppression had to be added. Unload the class
- *  instead, the way the puzzle about the tojos lifecycle below asks, and
- *  the suppression can go, together with the {@code @SuppressWarnings}
- *  for the field count right below it, which the same unloading fixes.
- *  Until then no further step can be moved to {@link GlobalCache}
- *  without making this worse.
- * @checkstyle ClassFanOutComplexityCheck (5 lines)
```
Favixx, line 39: Removing this rather than raising the limit is the right call, and checking against the Checkstyle version qulice actually ships is the part most such PRs skip. What makes it fragile is that the margin is zero: a qulice bump that changes how a referenced type is counted — generics, annotations, `@link` targets — turns this into a red build in an unrelated PR, and the person who hits it will have no way to know it was expected. That is worth one line in the puzzle above.

eo-maven-plugin/src/main/java/org/eolang/maven/ChBrief.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+/**
+ * A commit hash in its short form ({@link ChNarrow}), read once per build
```
Favixx: This describes `ChBrief(CommitHash)` precisely and says nothing about the other constructor, which takes a tag and goes to objectionary over the network. Someone reading `new ChBrief(this.tag)` in `MjSafe` gets "short form, read once per build" from here and no hint that constructing one implies a remote lookup. Naming `ChRemote` in the same sentence closes that gap.

eo-maven-plugin/src/main/java/org/eolang/maven/ChBrief.java
```diff
@@ -0,0 +1,39 @@
...
+     * The short hash.
+     */
+    private final CommitHash origin;
+
+    /**
+     * Ctor.
+     * @param tag The Git tag in objectionary to take the hash of
+     */
+    ChBrief(final String tag) {
+        this(new ChRemote(tag));
```
Favixx, line 23: This is the constructor that makes `ChBrief` a network call, and the one no test touches. It also decides that a `String` means "a tag in objectionary" while `ChBrief(CommitHash)` right below means "a hash to shorten" — two quite different jobs distinguished only by the argument type. Since there is exactly one caller, leaving `new ChBrief(new ChRemote(this.tag))` at the `MjSafe` site would keep the remote step visible where it happens and leave this class with a single, obvious purpose.

eo-maven-plugin/src/main/java/org/eolang/maven/ChBrief.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /**
+     * Ctor.
+     * @param full The hash to shorten
+     */
+    ChBrief(final CommitHash full) {
+        this.origin = new ChCached(new ChNarrow(full));
```
Favixx, line 31: Having one place that builds the chain, with the other constructor delegating here, is the right shape. The consequence is that the chain is now fixed: a caller who already holds something that needs no caching still gets `ChCached` around it, and `ChCached` is `Unchecked(Synced(Sticky(...)))` — three objects and a lock. Fine for `MjSafe`, where the delegate is a network fetch; see the note on `CommitHashesMap` for where it is not.

eo-maven-plugin/src/main/java/org/eolang/maven/ChBrief.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A commit hash in its short form ({@link ChNarrow}), read once per build
+ * ({@link ChCached}).
+ * @since 0.73.3
+ */
+final class ChBrief implements CommitHash {
+
+    /**
+     * The short hash.
```
Favixx, line 14: "The short hash" describes the value the field yields rather than the field itself, which holds the cached-and-narrowed chain. Combined with the name `origin` — the convention here for a delegate — the two read as if the class stored a `String`. "The hash, shortened and read once" would match what is assigned.

eo-maven-plugin/src/main/java/org/eolang/maven/CommitHashesMap.java
```diff
@@ -88,11 +88,7 @@ private static Map<String, CommitHash> fromTable(final Scalar<String> table) {
                     final String[] row = CommitHashesMap.WHITESPACE.split(rows.asString());
                     return new MapEntry<>(
                         row[1],
-                        new ChCached(
-                            new ChNarrow(
-                                new CommitHash.ChConstant(row[0])
-                            )
-                        )
+                        new ChBrief(new CommitHash.ChConstant(row[0]))
```
Favixx: Much better to read than the four nested lines. It does make one thing concrete: every row of this table now carries `ChCached` — i.e. `Unchecked(Synced(Sticky(...)))` — wrapped around a `ChConstant` that was never going to be computed twice. That was equally true before this PR, so it is not a regression, but `ChBrief` turns it from a decision the caller could decline into a property of the type. If this map is built with a row per tag in objectionary, a `ChNarrow(ChConstant)` is the honest row here and `ChBrief` is the one for `MjSafe`.

eo-maven-plugin/src/test/java/org/eolang/maven/ChBriefTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Test case for {@link ChBrief}.
+ * @since 0.73.3
+ */
+final class ChBriefTest {
+
+    @Test
+    void shortensTheHashItIsGiven() {
+        final long seed = System.nanoTime();
```
Favixx, line 21: Seeding from `System.nanoTime()` and printing the seed in the failure message is the careful way to do a random test — but the assertion here is a pure function of its input, so the randomness buys nothing that a fixed 40-character hash would not: the same branch runs either way, and the test stops depending on the clock. If the aim is to cover awkward hashes (leading zeros, all-digit prefixes), those are two named fixtures rather than one draw that might never produce them.

eo-maven-plugin/src/test/java/org/eolang/maven/ChBriefTest.java
```diff
@@ -0,0 +1,29 @@
...
+        final String full = String.format("%040x", new BigInteger(160, new Random(seed)));
+        MatcherAssert.assertThat(
+            String.format("short hash of %s is not its first seven chars, seed %d", full, seed),
+            new ChBrief(() -> full).value(),
```
Favixx, line 25: This is the only test, and it covers `ChNarrow` — one of the two things the class javadoc promises. Nothing pins "read once per build", which is the half that matters at the `MjSafe` call site where the delegate is a remote fetch. A `CommitHash` that counts its invocations, asked for `value()` twice and expected to have been called once, covers it in three lines.

eo-maven-plugin/src/test/java/org/eolang/maven/ChBriefTest.java
```diff
@@ -0,0 +1,29 @@
...
+            Matchers.equalTo(full.substring(0, 7))
```
Favixx, line 26: The expected value restates what `ChNarrow` does, so this passes as long as the test and the implementation agree with each other — if the seven ever becomes eight, this assertion follows it and never fails. Against a fixed input, a literal expected string makes the seven something the test defends rather than something it copies.

Thayorns:
@yegor256 have a look please

yegor256:
@Thayorns conflicts here

Thayorns:
@yegor256 ping

.github/workflows/counts.yml
```diff
@@ -16,6 +16,34 @@ jobs:
   counts:
     timeout-minutes: 15
     runs-on: ubuntu-24.04
+    strategy:
```
yegor256: @Thayorns what is this for? it's not used in the YAML anywhere...

Thayorns: Nothing, you are right. It is a leftover from an earlier merge of master into this branch. Master had since rewritten this workflow to compare each metric against the base branch instead of asserting a fixed number, and my resolution kept the old `strategy: matrix` block on top of the new steps, which read none of it. The same bad resolution also dropped the `instanceofs` line master had added.

Removed both, so `counts.yml` is byte-identical to master now and this pull request does not touch the file at all. I also merged master again, which resolves the conflict you pointed at: `CommitHashesMap` is gone upstream, so I took the deletion.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1213](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+249](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2946](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6464, opened 2026-08-07
fix(#6427): hand the trailing blank back to the walk

### LINKED ISSUE #6427: `Recovery` incorrectly handles blank lines causing suppressed R-6.5.4 diagnostic on subsequent objects
## What happens

`Recovery.skipped` (`eo-parser/src/main/java/org/eolang/parser/Recovery.java:59-62`) treats a blank line as always belonging to the block being skipped after a failed line:

```java
private boolean skipped(final int idx, final int indent) {
    final Span span = this.spans.get(idx);
    return span.blank() || span.indent() > indent;
}
```

That is correct for a blank line genuinely nested inside the failed line's own block. But it also swallows a blank line that is really a file-level separator between the failed line's block and the next unrelated top-level object. Since the blank line is skipped entirely, it never reaches `Eo.process`, so `globals` never records it, and `Blanks.checkPlain` on the following object sees no preceding blank line at all — silently suppressing a genuine R-6.5.4 "blank line before a plain object is forbidden" violation on that next object.

Verified:

```
# Sample.\n\n[] > app\n  x..y > z\n\n42 > w\n     -> 1 error  (only the x..y ParseError)
# Sample.\n\n[] > app\n  42 > z\n\n42 > w\n       -> 1 error  ('blank line before a plain
                                                     object is forbidden (R-6.5.4)')
```

In the first case, `x..y` is a genuine parse error and `42 > w` (which has the same R-6.5.4 violation as the second case) reports nothing at all, because the blank line separating it from the failed block above was consumed by `Recovery`.

## What should happen

A blank line that separates the failed line's block from a following sibling object (as opposed to a blank line genuinely nested inside that block) should still be visible to the normal blank-line bookkeeping, so R-6.5.4 continues to apply to whatever follows. This is a minor, low-impact case (a real diagnostic dropped, not a crash or corruption) compared to #6423/#6426 which cover the same `Recovery` area.

etojesaandro:
Recovery swallowed the blank that separates a failed line's block from the object below it, so the blank never reached Globals and R-6.5.4 went unreported on that object. 

The skip now stops at the last non-blank line it covers, unless the block runs to the end of the file.

Closes #6427

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Recovery.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Recovery.java
@@ -17,6 +17,17 @@ import java.util.List;
  * back at or above the indent of the one that failed. A blank line
  * carries no indent of its own and never ends the skipped block.</p>
  *
+ * <p>The blanks trailing the block are handed back, though. A blank
+ * standing between the last skipped line and the line the walk resumes
+ * at separates two objects instead of sitting inside either, and §6.5
+ * reads its rules off exactly such separators — R-6.5.4 forbids one in
+ * front of a plain object. Swallowing it would drop that diagnostic
+ * from whatever follows the failed block, since the line never reaches
+ * the blank-line bookkeeping. Blanks with skipped lines still under
+ * them are nested and stay skipped, and a block running to the end of
+ * the file keeps its trailing blanks, no object being left for them to
+ * separate.</p>
+ *
  * @since 0.1
  */
 final class Recovery {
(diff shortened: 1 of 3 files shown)


eo-parser/src/main/java/org/eolang/parser/Recovery.java
```diff
@@ -46,6 +57,24 @@ int after(final int failed) {
         while (idx < this.spans.size() && this.skipped(idx, indent)) {
             idx = idx + 1;
         }
+        if (idx < this.spans.size()) {
+            idx = this.rewound(idx, failed);
```
Thayorns, line 61: When the failed line was itself preceded by a blank, the blank you hand back gets reported as a consecutive one. Eo.dispatch catches ParseError at Eo.java:415 and rolls back the emit and the stack, but not globals, and LnApplication.into only reaches globals.clearBlanks() at LnApplication.java:92 — far below the Tokens calls at :60-61 that throw. So pendingBlanks is still 1 when the handed-back blank arrives at LnBlank.java:41 and emits "consecutive blank lines forbidden". On

    [] > inner

    x..y > z

    42 > w

the blank under x..y now carries that error on top of the R-6.5.4 you want, and the line above it is not blank at all.

Thayorns (commented):
Recovery now stops the skip at the last non-blank line it covers, so the blank separating a failed block from the object below reaches Globals and R-6.5.4 fires on that object. The walk still advances — rewound floors at failed + 1 — and the existing blank-inside-block case is unchanged. One interaction with the blank counter looks off.

0crat:
@etojesaandro Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1024](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6465, opened 2026-08-07
#6449: reject a negative size in posix and win32 read

### LINKED ISSUE #6449: Negative requested size in posix.read and win32.read results in NegativeArraySizeException instead of EO failure
## What happens

The native read adapters allocate the requested buffer before calling the platform function, without validating its size. A negative EO argument therefore escapes as a Java `NegativeArraySizeException`.

The affected code is [`ReadSyscall.make()`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/posix/ReadSyscall.java#L33-L36) and its Windows counterpart [`ReadFuncCall.make()`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/win32/ReadFuncCall.java#L32-L35):

```java
```

## Runtime reproduction

Calling the POSIX adapter with a negative size fails before any system call is made:

```java
new ReadSyscall(posix).make(new Data.ToPhi(0), new Data.ToPhi(-1));
```

Actual result:

```
```

The equivalent `win32.read` adapter contains the same unchecked allocation.

## Expected behavior

`posix.read` and `win32.read` should reject a negative requested size through the runtime’s controlled EO failure path, with a message that identifies the invalid argument, rather than exposing a JVM implementation exception. The check should occur before allocating the buffer or invoking the native function.

asmirnov-backend:
@yegor256 please review.

Closes #6449.

`posix.read` and `win32.read` allocated `new byte[size]` before validating it, so a negative EO argument escaped as `NegativeArraySizeException`. Both adapters now fail with `ExFailure` before allocating, like `WriteSyscall`/`SendSyscall` already do.

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/ReadSyscallTest.java
+++ b/eo-runtime/src/test/java/org/eolang/ReadSyscallTest.java
@@ -1,0 +1,43 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang;
+
+import org.eolang.posix.ReadSyscall;
+import org.eolang.win32.ReadFuncCall;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+
+/**
+ * Test case for {@code read} system calls.
+ * @since 0.40.0
+ */
+final class ReadSyscallTest {
+
+    @Test
+    @DisabledOnOs(OS.WINDOWS)
+    void rejectsNegativeSizeOnPosixRead() {
+        Assertions.assertThrows(
+            ExFailure.class,
+            () -> new ReadSyscall(Phi.Φ.take("posix").copy()).make(
+                new Data.ToPhi(0), new Data.ToPhi(-7)
+            ),
+            "A negative posix read size must fail with ExFailure, not NegativeArraySizeException"
+        );
+    }
+
+    @Test
+    @DisabledOnOs({OS.MAC, OS.LINUX})
+    void rejectsNegativeSizeOnWindowsRead() {
+        Assertions.assertThrows(
+            ExFailure.class,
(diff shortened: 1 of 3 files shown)


eo-runtime/src/test/java/org/eolang/ReadSyscallTest.java
```diff
@@ -0,0 +1,43 @@
...
+            () -> new ReadSyscall(Phi.Φ.take("posix").copy()).make(
+                new Data.ToPhi(0), new Data.ToPhi(-7)
+            ),
+            "A negative posix read size must fail with ExFailure, not NegativeArraySizeException"
+        );
+    }
+
+    @Test
+    @DisabledOnOs({OS.MAC, OS.LINUX})
+    void rejectsNegativeSizeOnWindowsRead() {
```
Thayorns, line 34: The file is named after ReadSyscall but this second test covers ReadFuncCall in org.eolang.win32, which has no test file carrying its own name anywhere in eo-runtime/src/test/java. Whoever changes ReadFuncCall next has no reason to open a file called ReadSyscallTest. SyscallTest is the precedent for spanning both platforms in one file, and it takes a platform-neutral name for exactly that reason.

Thayorns (commented):
The guard lands before the allocation in both adapters, so a negative size now fails as ExFailure instead of NegativeArraySizeException, and the varargs ExFailure ctor formats the size into the message as intended. One remark on where the win32 half of the new test lives.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1031](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@asmirnov-backend Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+108](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6467, opened 2026-08-07
#6426: run recovery for a rejected multi-line BYTES literal

### LINKED ISSUE #6426: `mergeBytesContinuation` doesn't trigger recovery for rejected multi-line BYTES literal, causing spurious errors
## What happens

`mergeBytesContinuation` (`eo-parser/src/main/java/org/eolang/parser/Eo.java:228-233`) merges a multi-line BYTES literal and dispatches the merged line, but discards the result:

```java
Eo.process(
    new Span(
        " ".repeat(head.indent()).concat(body.toString()), head.line()
    ),
    stack, globals, emit
);
return idx;
```

`Eo.process` returns `true` when the line failed to parse. Every other call site (`Eo.java:85`) uses that return value to call `recovery.after(idx)`, which skips the block nested under the failed line, per `PARSER_SPEC.md` §7's "no cascade error, no double-reporting" contract. This call site throws the flag away, so recovery never runs for a rejected multi-line BYTES literal, and any nested block under it produces a spurious second error.

Verified with two inputs, identical in shape (a rejected line at indent 6, one child at indent 8), differing only in whether the rejected line is a BYTES continuation:

```
A) # Sample.\n\n[] > app\n      01-02-\n        q.\n     -> 2 parser errors
B) # Sample.\n\n[] > app\n      42 > z\n        q.\n     -> 1 parser error
```

```
=== A: over-indented bytes continuation + deeper child
  parser errors=2
   line 4: [4:0] error: 'indent increased by more than one level'
   line 5: [5:0] error: 'indent increased by more than one level'   <-- cascade

=== B: control, over-indented plain line + deeper child
  parser errors=1
   line 4: [4:0] error: 'indent increased by more than one level'
```

A second shape (a blank line before the BYTES continuation) shows the same asymmetry: 3 errors instead of 2.

This is the same contract violation as #6423, at a different call site (`mergeBytesContinuation` here, versus the tab/odd-indent branches of `Eo.process` in #6423) with a different fix (using the return value here, versus setting `failed = true` there).

## What should happen

A rejected multi-line BYTES literal should go through the same recovery path as any other rejected line: report the one error, then skip the nested block, instead of leaving a cascading phantom error on every line nested under it.

Note for whoever picks this up: since the root contract (§7 recovery not firing for a rejection path in `Eo`) is the same as #6423, it may make sense to fix both in the same change rather than as fully separate PRs — worth checking with whoever is already working on #6423 before starting.

morphqdd:
Closes #6426.

mergeBytesContinuation dispatches the merged multi-line BYTES literal through Eo.process but discarded the return value, so a rejected literal never called recovery.after, unlike every other call site in Eo.directives(). Any block nested under the rejected literal produced a spurious second error instead of being skipped per PARSER_SPEC.md section 7.

Recovery.after(failed) derived both the scan start and the reference indent from the same span index, which does not fit this call site: the reference indent belongs to the head line, but the scan must start past the continuation lines the merge already consumed. Added Recovery.skip(from, indent) as the decoupled primitive, rebuilt after(int) on top of it, and used skip directly in mergeBytesContinuation.

Added a yaml pack test (recovery-skips-nested-block-after-bytes-continuation.yaml) reproducing the issue's case A, and a direct unit test for Recovery.skip in RecoveryTest. Verified both fail on the pre-fix code and pass with the fix.

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/RecoveryTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/RecoveryTest.java
@@ -47,4 +47,21 @@ final class RecoveryTest {
             Matchers.equalTo(4)
         );
     }
+
+    @Test
+    void skipsFromAnIndexOtherThanTheFailedLineItself() {
+        MatcherAssert.assertThat(
+            "skip must scan from the given index using the given indent, "
+                .concat("not the indent of the line at that index"),
+            new Recovery(
+                Arrays.asList(
+                    new Span("01-02-", 1),
+                    new Span("03-04", 2),
+                    new Span("  child", 3),
+                    new Span("sibling", 4)
+                )
+            ).skip(2, 0),
+            Matchers.equalTo(3)
+        );
+    }
 }
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -76,8 +76,7 @@ final class Eo implements Iterable<Directive> {
         while (idx < spans.size()) {
             final Span span = spans.get(idx);
             if (!globals.inTextBlock() && Eo.isBytesContinuation(span.body())) {
-                final int next = Eo.mergeBytesContinuation(spans, idx, stack, globals, emit);
-                idx = next;
+                idx = Eo.mergeBytesContinuation(spans, idx, stack, globals, emit, recovery);
             } else if (Eo.process(span, stack, globals, emit)) {
                 idx = recovery.after(idx);
             } else {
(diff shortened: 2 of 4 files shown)


Thayorns (commented):
Splitting the scan start from the reference indent is the right shape for this call site, and after(int) rebuilt on top of skip keeps the old behaviour identical. The yaml pack pins the real symptom down: with the old code the q. line came back as a second error, and the merge loop's idx is exactly the right place to resume from. One small thing about the new unit test's input.

eo-parser/src/test/java/org/eolang/parser/RecoveryTest.java
```diff
@@ -47,4 +47,21 @@ void skipsBlankLineInsideBlock() {
...
+    @Test
+    void skipsFromAnIndexOtherThanTheFailedLineItself() {
+        MatcherAssert.assertThat(
+            "skip must scan from the given index using the given indent, "
+                .concat("not the indent of the line at that index"),
+            new Recovery(
+                Arrays.asList(
+                    new Span("01-02-", 1),
+                    new Span("03-04", 2),
+                    new Span("  child", 3),
```
Thayorns, line 60: skip only ever reads indent and blankness, so the bodies here are free, yet they are all plain short ASCII words. The two tests above pick über! ☃ and क्ष for exactly that reason — a surrogate pair or a combining mark in a body would catch an indent count that walks chars instead of leading spaces, and this test costs nothing by using them too.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1038](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3677](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6468, opened 2026-08-07
#6425: give Eo.dispatch's rollback a real way to undo Stack.replace

### LINKED ISSUE #6425: `Eo.dispatch` error rollback fails to undo `Stack.replace`, leading to `ClassCastException` or invalid XMIR
## What happens

`Eo.dispatch` (`eo-parser/src/main/java/org/eolang/parser/Eo.java:403-422`) recovers from a `ParseError` by rolling back the emitted directives and truncating the indent stack:

```java
} catch (final ParseError err) {
    emit.rollback(savepoint);
    stack.silentTruncate(frame);
    ...
}
```

`Stack.silentTruncate` (`Stack.java:116-122`) only pops entries until `size()` matches the recorded frame. That works when the failed line pushed a new level. It does not work when the failed line went through `Stack.replace` (`Stack.java:245-277`) first: `replace` removes the top level, runs its close-time check, and pushes a *new* level at the same indent, so the stack size never changes. `silentTruncate` is then a no-op, and the *replacement* level stays on the stack instead of the original one.

This matters because `Eo.checkOnClose` (`Eo.java:901-921`) gives different `Kind`s a different number of owed `emit.close()` calls: an `ONLY_PHI_FORMATION` that isn't `HORIZONTAL_COMPLETED`, and a `COMPACT_TUPLE`/`star()` level, each owe two closes; everything else owes one. If the replaced level and the replacement level owe a different count, the directive stream and the parser's own model of "how many `<o>` are open" diverge:

- one close too many at the end → the Xembler cursor walks above `/object`, and `EoSyntax.parsed()`'s trailing `.attr("ms", 0L)` runs on the `Document` node → a raw `java.lang.ClassCastException` escapes `Xembler.domQuietly()`
- one close too few → `ms` lands on an `<o>` element instead of `/object`, which `XMIR.xsd` forbids

Minimal repro, run through `EoSyntax.parsed()`:

```
[] > pl
  b
    "x"
    "r\c" > [m]
```

```
Exception in thread "main" java.lang.ClassCastException: Cannot cast
  com.sun.org.apache.xerces.internal.dom.DocumentImpl to org.w3c.dom.Element
    at org.xembly.AttrDirective.exec(AttrDirective.java:85)
    at org.xembly.Xembler.dom(Xembler.java:214)
    at org.xembly.Xembler.domQuietly(Xembler.java:183)
    at org.eolang.parser.EoSyntax.parsed(EoSyntax.java:111)
```

Real-world repro: take the unmodified stdlib file `eo-runtime/src/main/eo/string/padded-left.eo` and change the `"recovered"` inside `rejects-a-multi-character-fill` to `"r\covered"` (a single stray backslash, an ordinary typo). Parsing that file throws the same `ClassCastException`, with no file name and no line number pointing at the typo.

A control confirms the mechanism: the same failing line as the *first* child of its parent (a genuine push, not a replace) parses fine; replacing a level of the *same* kind (so the owed-close count matches) also parses fine. It only breaks when a replace happens and the two kinds owe a different number of closes:

| shape | balance | result |
|---|---|---|
| `"x"` then a failing only-phi/compact-tuple replacing a plain sibling | -1 | `ClassCastException` |
| a plain sibling replacing an only-phi/compact-tuple that then fails | +1 | `ms` lands on `<o>`, schema-invalid XMIR |
| same-kind replace (both only-phi) | 0 | correct |
| push instead of replace (single child, no prior sibling) | 0 | correct |

`EoSyntax.parsed()` is called by `EoSource.parsed()` (the `parse` goal) and by `MjFormat`. Neither catches `RuntimeException`, so this reaches users directly as an unhandled crash with no diagnostic information.

## What should happen

A single-character typo in a `.eo` file should produce a normal `<error>` entry, the way any other `ParseError` does, not an uncaught `ClassCastException` or schema-invalid XMIR. `Stack` needs a real way to undo whatever `replace` did (e.g. record the displaced level so a rollback can reinstate it, or snapshot the actual `Level` object rather than just the stack size) instead of the size-based `silentTruncate`, which cannot model a same-size replacement.

morphqdd:
Closes #6425.

Eo.dispatch recovers from a ParseError with emit.rollback(savepoint) + stack.silentTruncate(frame), where frame was stack.size() taken before the line ran. That assumes a failed line's only stack side effect is a net change in entry count. push fits that assumption; replace does not, since it removes the old top and pushes a new one at the same indent, leaving size() unchanged even though the top is now a different Level. silentTruncate was then a no-op: the replacement level stayed on the stack with no matching directives in the rolled-back emit stream, so its own eventual close overshot or undershot the sink's cursor, producing the reported ClassCastException or schema-invalid XMIR depending on which way the owed-close count differed between the two kinds.

Replaced the size-based savepoint with Stack.snapshot()/restore(), which capture and reinstate the actual Level list by identity, so a replace's rollback puts the original Level back instead of leaving the replacement in place. size() and silentTruncate had no other callers, so removed rather than left dead.

Added a yaml pack test (replaced-level-rollback-does-not-crash.yaml) using the issue's own minimal repro, and a direct unit test in StackTest asserting restore() reinstates the exact Level replace displaced (by identity, not just by indent). Verified both fail with the reported ClassCastException on the pre-fix code and pass with the fix.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Stack.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Stack.java
@@ -108,26 +108,30 @@ final class Stack {
     }
 
     /**
-     * Current number of entries on the stack — savepoint for per-line
-     * rollback (R-7.3).
-     * @return Entry count
+     * A copy of the current levels — savepoint for per-line rollback
+     * (R-7.3). A plain entry count cannot serve as that savepoint:
+     * {@link #replace(int, Kind, Openness)} removes one entry and adds
+     * one back, so the count is unchanged even though the top entry
+     * itself is a different object. Snapshotting the entries themselves
+     * survives that case.
+     * @return Snapshot of the levels, bottom-to-top
      */
-    int size() {
-        return this.levels.size();
+    List<Level> snapshot() {
+        return new ArrayList<>(this.levels);
     }
 
     /**
-     * Pop entries silently until {@code size()} equals {@code target}.
-     * Used by {@link Eo} to undo {@link #push} / {@link #replace} side
-     * effects of a line that threw a {@link ParseError} (R-7.3) — the
-     * closer is <em>not</em> invoked here because the rolled-back open
-     * directives never reached the sink.
-     * @param target Target stack size
+     * Restore the levels to a previously taken {@link #snapshot()},
+     * discarding whatever {@link #push} / {@link #replace} side effects
+     * happened since. Used by {@link Eo} to undo those side effects for
+     * a line that threw a {@link ParseError} (R-7.3) — the closer is
+     * <em>not</em> invoked here because the rolled-back open directives
+     * never reached the sink.
+     * @param snapshot A snapshot taken earlier via {@link #snapshot()}
      */
-    void silentTruncate(final int target) {
-        while (this.levels.size() > target) {
-            this.levels.remove(this.levels.size() - 1);
-        }
+    void restore(final List<Level> snapshot) {
+        this.levels.clear();
+        this.levels.addAll(snapshot);
     }
 
     /**
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -341,12 +341,12 @@ final class Eo implements Iterable<Directive> {
         if (Eo.closesTextBlock(span, globals)) {
             stack.popDeeperThan(span.indent());
             final int tstartsen = emit.savepoint();
-            final int frame = stack.size();
+            final java.util.List<Level> frame = stack.snapshot();
             try {
                 new LnTextBlock(span).into(stack, globals, emit);
             } catch (final ParseError err) {
                 emit.rollback(tstartsen);
-                stack.silentTruncate(frame);
+                stack.restore(frame);
                 emit.error(err.line(), err.pos(), err.getMessage());
                 globals.closeTextBlock();
             }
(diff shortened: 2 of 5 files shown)


eo-parser/src/main/java/org/eolang/parser/Stack.java
```diff
@@ -108,26 +108,30 @@ Level top() {
...
+     * {@link #replace(int, Kind, Openness)} removes one entry and adds
+     * one back, so the count is unchanged even though the top entry
+     * itself is a different object. Snapshotting the entries themselves
+     * survives that case.
+     * @return Snapshot of the levels, bottom-to-top
      */
-    int size() {
-        return this.levels.size();
+    List<Level> snapshot() {
+        return new ArrayList<>(this.levels);
```
Thayorns, line 120: The snapshot copies the list but not the entries in it, and Level's own docblock opens with "Mutable record. Most fields flip after construction". So a line that flips a field on a level it never pushed still survives the rollback: LnMethod.into calls stack.below().upgradeArgBinding() at LnMethod.java:84, which sets argbound on the entry below the top, and the emitArg loop thirteen lines further down throws ParseError out of Emissions.number for an out-of-range or over-precise numeric argument. restore then reinstates that very same object with argbound already true, so that half of the rollback is still a no-op.

Thayorns (commented):
Swapping the count savepoint for a real list savepoint is the right call — replace's remove-then-add is invisible to size(), and the StackTest case pins the fix down by identity rather than by indent, which is exactly what was missing. One gap left in the same area.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1059](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3569](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6470, opened 2026-08-07
#6455: compute mod by doubling and halving the divisor

### LINKED ISSUE #6455: `number.mod` fast path miscalculates remainder for large dividends with small divisors
`number.mod`'s fast path in `eo-runtime/src/main/eo/number/mod.eo:36-43` computes a wrong remainder whenever the divisor is small relative to the dividend. The guard at lines 36-38 (`lt. (abs dividend >> absx) 4503599627370496`) only checks that `|x| < 2^52`, then the body at lines 39-43 computes `absx.minus (times. absy (floor. (absx.div absy)))`. This was added by the fix for #6225 (commit 8fc2d1e654) to handle large exactly-representable dividends, but it never checks the *quotient*, which is what actually needs to fit in a double's mantissa for this formula to be exact.

`x - y*floor(x/y)` is only exact when `x/y` is representable without rounding. When `y` is small, `x/y` can exceed 2^53, so `floor(x/y)` already lost precision, and `y * floor(x/y)` can end up larger than `x`. The result then breaks the required `0 <= |remainder| < |y|` invariant, can have the wrong sign, or can even become `-Infinity` if the intermediate division overflows. For example `1e15.mod 1e-5` should be `6.946085968697146e-6` (matching Java's `%`/`fmod`) but returns `-0.125`; `1.mod 4.9e-324` returns `-Infinity` instead of `0.0`. I traced the first case by hand in float64: `floor(1e15/1e-5)` rounds to `1e20`, `1e-5 * 1e20 = 1000000000000000.1` (not `1e15`), so `1e15 - 1000000000000000.1 = -0.125`. The larger-dividend `reduce` path (lines 44-65) is unaffected and stays correct regardless of divisor size (e.g. `1e16.mod 0.3` works). None of the 20 existing tests in `mod.eo` catch this, since they all use integer divisors (1, 2, 3, 5, 7, 13, -15, -200) and never a fractional or very small one against a large dividend.

The fix is to gate the fast path on the quotient rather than the dividend alone: only take it when `absx.div absy` is finite and less than 2^53, otherwise fall through to the existing `reduce` path, which already handles small divisors exactly.

asmirnov-backend:
Closes #6455.

The fast path was gated on the dividend alone, but the formula `x - y*floor(x/y)` is exact only when the *quotient* fits in a double's mantissa. With a small divisor `x/y` exceeds 2^53, so `floor` already lost precision and `y*floor(x/y)` could exceed `x`: `1e15.mod 1e-5` returned `-0.125`. The `reduce 62` fallback had the same blind spot above a 2^63 quotient.

Both branches are replaced by one algorithm: double the divisor until it reaches the dividend, then halve it back down, subtracting whenever it fits. Doubling and halving are exact in a double, and every subtraction runs with `chunk <= rest < 2*chunk`, which is exact too, so the result matches the true remainder for any pair of finite numbers regardless of their magnitude gap.

Two tests added, both red before the fix.

### DIFF
--- a/eo-runtime/src/main/eo/number/mod.eo
+++ b/eo-runtime/src/main/eo/number/mod.eo
@@ -1,15 +1,14 @@
 # Calculate MOD.
 # An operation that finds the remainder after dividing `x` by `y`. When `y`
 # is zero, the result is the caller-supplied `cant-mod` fallback, or the
-# bottom object when none was provided, which terminates on use. For a
-# dividend below 2^52 the remainder is computed the fast, direct way; a
-# double is exact for every integer in that range, so this always matches
-# the true remainder. A larger dividend switches to a bit-by-bit reduction,
-# subtracting decreasing powers of two times the divisor, which stays exact
-# up to 2^63 (about 9.2e18) regardless of how small the divisor is, at the
-# cost of up to 63 recursive steps; a dividend beyond 2^63 is not
-# guaranteed exact, since determining more quotient bits exactly would need
-# a matching increase in recursion depth.
+# bottom object when none was provided, which terminates on use. Otherwise
+# the divisor is doubled until it reaches the dividend and then halved back
+# down, subtracting it from the dividend whenever it fits. Doubling and
+# halving a double are exact, and every subtraction happens while the
+# remainder is below twice the subtracted value, which is exact too, so the
+# answer matches the true remainder for any pair of finite numbers, however
+# far apart their magnitudes are, at the cost of two recursive steps per
+# bit of the quotient.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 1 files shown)


asmirnov-backend:
@yegor256 please review

eo-runtime/src/main/eo/number/mod.eo
```diff
@@ -32,37 +31,24 @@
...
-          [level p] > reduce
-            if. > @
-              level.lt 0
-              p
+          shrink > @
+            grow
+              abs divisor >> absy
+            abs dividend >> absx
+          if. > [chunk] > grow
+            chunk.gte absx
```
Thayorns, line 39: grow's only exit is chunk.gte absx, and gte is gt-or-eq over Java's `>` in EOnumber$EOgt, which is false whenever either side is NaN. With a NaN dividend no chunk ever satisfies it, and chunk.plus chunk saturates at pinf, so `mod nan 5` recurses forever instead of coming back NaN. An infinite dividend gets past grow but then sticks in shrink, since pinf.div 2 is pinf and chunk.lt absy on line 44 never becomes true. The divisor is guarded for both cases on line 27, the dividend is not, and the reduce 62 path this replaces bottomed out after 63 steps whatever it was given.

Thayorns (commented):
The doubling/halving scheme is right for finite operands: every subtraction runs with chunk <= rest < 2*chunk, which Sterbenz makes exact, so 1e15 mod 0.1 lands on 0.04448884876874218 as the new test says, and the old fast path's mantissa blind spot is gone. One thing about operands the old code survived and this one does not.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1045](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@asmirnov-backend Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+124](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6471, opened 2026-08-07
Write down what every object is asked for

### LINKED ISSUE #6469: eo-inference module fails to record objects' needs in tables
The `eo-inference` module implements one of the four rules that
`inference.md` describes. A formation tells us what its object provides,
and `Provides` writes that down; nothing yet writes down what an object
must *have*. Rule three of the note covers it: a dispatch takes an
attribute from something, and that something must therefore offer it.

Without those rows the tables cannot be compared, so the checker still
reports nothing. In the note's own example the missing half is what makes
the mistake visible:

```text
Φ.app.inc.φ.ρ.ρ  needs next  (of type Φ.app.inc.φ.ρ)
Φ.app.inc.φ.ρ    needs foo   (of type Φ.app.inc.φ)
```

The input is already in the right shape. `Inferring` unrolls every
composite base into one object per dispatch step and re-sets the locators
before any clue looks at the XMIR, so a dispatch is a single `<o>` whose
first non-argument child is its receiver:

```xml
<o base=".foo" loc="Φ.app.inc.φ">
  <o base=".next" loc="Φ.app.inc.φ.ρ">
    <o base="ξ.x" loc="Φ.app.inc.φ.ρ.ρ"/>
  </o>
</o>
```

A second clue alongside `Provides`
(`eo-inference/src/main/java/org/eolang/inference/Provides.java:67`) would
sweep those objects and write `needs.xml` into the same directory: one row
per dispatch, the owner being the receiver's locator, the name the
attribute taken, and the type the locator of the dispatch itself. That is
the row shape `Grouped` already renders, so the view needs no changes, and
`Clues`
(`eo-inference/src/main/java/org/eolang/inference/Clues.java:53`) gains one
more entry in its default list. Types stay identified by locator and are
never renamed or merged, as they are today.

Keep it primitive: no resolution of what the receiver actually is, no
attempt to decide anything. That belongs to the links table and the
checking loop, which come after this.

maxonfjvipon:
Rule three of `inference.md`: a dispatch tells us what an object must have.

`Needs` sweeps every object whose base begins with a dot and writes one row
per dispatch into `needs.xml` — the owner is the receiver's locator, the name
is the attribute taken, the type is the locator of the dispatch itself. The
note's own example comes out as it does on paper:

```xml
<needs>
  <type id="Φ.app.inc.φ.ρ.ρ">
    <attr name="next" type="Φ.app.inc.φ.ρ"/>
  </type>
  <type id="Φ.app.inc.φ.ρ">
    <attr name="foo" type="Φ.app.inc.φ"/>
  </type>
</needs>
```

The row shape is the one `Provides` already uses, so `Grouped` renders the
second table without a line changed. A chain needs no walking of its own:
every step of `x.next.foo` is a dispatch object by the time a clue sees the
XMIR, so the outer one simply asks the inner one. The receiver is the child
that is not an argument — `x.plus 5` takes `plus` from `x`, never from the
`5`; all 9,848 dispatches in eo-runtime have their receiver first and carry a
locator, so the rule is total on real input.

Nothing here resolves what a receiver is or decides whether it really has the
attribute. Asking is all this clue does; comparing the two tables belongs to
the checking loop, once links between copies exist.

On eo-runtime the goal now writes 9,848 need rows next to the 2,456 types of
`provides.xml`.

`Clues` also gained a paragraph explaining why every clue reads the program as
text rather than following what runs. The reason is empirical rather than
aesthetic: eo-runtime is a library with no entry point, so a checker that
waits to be reached reports a clean bill of health for code it never opened.
Reading the text costs precision, which the checking loop can buy back by
resolving each site in its own context; reading only what runs costs coverage,
which nothing buys back.

Closes: #6469

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Clues.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Clues.java
@@ -6,8 +6,8 @@ package org.eolang.inference;
 
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 
 /**
  * Every clue the checker knows how to follow.
@@ -26,9 +26,21 @@ import java.util.Collections;
  *
  * <p>Three clues are planned, one per kind of object, and each fills a
  * table of its own: what an object certainly has ({@link Provides}), what
- * it must have judging by how it is used, and which types are copies of
- * which. Keeping the tables apart is what lets a smarter rule add rows,
- * or read them differently, without touching anything else.</p>
+ * it must have judging by how it is used ({@link Needs}), and which types
+ * are copies of which. Keeping the tables apart is what lets a smarter
+ * rule add rows, or read them differently, without touching anything
+ * else.</p>
+ *
+ * <p>Every clue reads the program as text, writing a row for an object
+ * because it is <em>there</em>, not because something reaches it. The
+ * alternative — starting from a root and following what actually runs —
+ * gives sharper answers, since every question is then asked in a context
+ * where the types are concrete, but it cannot be used here: eo-runtime is
+ * a library with no entry point, so a checker that waits to be reached
+ * has nothing to pull on and reports a clean bill of health for code it
+ * never opened. Reading the text costs precision, which the checking loop
+ * buys back later by resolving each site in its own context; reading only
+ * what runs costs coverage, which nothing buys back.</p>
  *
  * <p>Then the checks are drained one by one, each of them either
  * deciding, splitting into smaller checks, or waiting for facts that may
@@ -50,7 +62,7 @@ public final class Clues implements Clue {
      * Ctor.
      */
     public Clues() {
-        this(Collections.singletonList(new Provides()));
+        this(Arrays.asList(new Provides(), new Needs()));
     }
 
     /**
--- a/eo-inference/src/main/java/org/eolang/inference/Needs.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Needs.java
@@ -1,0 +1,68 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.xml.XML;
+import com.yegor256.tojos.MnMemory;
+import com.yegor256.tojos.TjCached;
+import com.yegor256.tojos.TjDefault;
+import com.yegor256.tojos.Tojos;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * What every object must have, judging by how it is used.
+ *
+ * <p>This is the other half of {@link Provides}: a formation shows what its
+ * object certainly has, while a dispatch shows what somebody expects to find.
+ * In this fragment</p>
+ *
+ * <pre> &lt;o base=".next" loc="Φ.app.inc.φ.ρ"&gt;
+ *   &lt;o base="ξ.x" loc="Φ.app.inc.φ.ρ.ρ"/&gt;
+ * &lt;/o&gt;</pre>
+ *
+ * <p>somebody takes {@code next} from the object at
+ * {@code Φ.app.inc.φ.ρ.ρ}, so that object must have such an attribute, and
+ * whatever it holds is the type of the dispatch itself. That is one row. The
+ * {@code .foo} wrapped around this fragment adds another, asking the dispatch
+ * for {@code foo} in turn — the chain is walked simply by every step of it
+ * being a dispatch of its own, which is what the splitting done before any
+ * clue looks at the XMIR is for.</p>
+ *
+ * <p>The receiver is the child that is not an argument: {@code x.plus 5}
+ * takes {@code plus} from {@code x}, never from the {@code 5}. Nothing here
+ * tries to work out what the receiver actually is, or whether it really has
+ * the attribute — asking is all this clue does. Comparing the answer with
+ * what {@link Provides} wrote down is the job of the checking loop, once the
+ * links between copies are known.</p>
+ *
+ * @since 0.68.0
+ */
+final class Needs implements Clue {
+
+    @Override
+    public void follow(final Path xmirs, final Path tables) throws IOException {
+        final Tojos rows = new TjCached(new TjDefault(new MnMemory()));
+        int seen = 0;
+        for (final XML dispatch : new Xmirs(xmirs).dispatches()) {
+            final String owner = dispatch.xpath("o[not(@as)][1]/@loc").get(0);
+            rows.add(owner).set("index", Integer.toString(seen));
+            seen = seen + 1;
+            rows.add(String.join(" ", owner, dispatch.xpath("@base").get(0)))
+                .set("owner", owner)
+                .set("index", Integer.toString(seen))
+                .set("name", dispatch.xpath("@base").get(0).substring(1))
+                .set("type", dispatch.xpath("@loc").get(0));
+            seen = seen + 1;
+        }
+        Files.createDirectories(tables);
+        Files.write(
+            tables.resolve("needs.xml"),
+            new Grouped(rows, "needs").asXml().toString().getBytes(StandardCharsets.UTF_8)
+        );
+    }
+}
--- a/eo-inference/src/main/java/org/eolang/inference/Xmirs.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Xmirs.java
@@ -53,13 +53,37 @@ final class Xmirs {
      * @throws IOException If a file cannot be read
      */
     Collection<XML> formations() throws IOException {
+        return this.matching(
+            "//o[not(@base) and not(@name='λ') and not(text()[normalize-space()])]"
+        );
+    }
+
+    /**
+     * Every dispatch of the program.
+     *
+     * <p>A dispatch is an object whose base begins with a dot: it takes an
+     * attribute from the object below it. Every composite base was split
+     * into one such object per step before any clue looked at the XMIR, so
+     * a chain like {@code x.next.foo} is three objects here and not one.</p>
+     *
+     * @return The dispatches, file by file, in the order they appear in
+     *  the code
+     * @throws IOException If a file cannot be read
+     */
+    Collection<XML> dispatches() throws IOException {
+        return this.matching("//o[starts-with(@base, '.')]");
+    }
+
+    /**
+     * Every object of the program the given XPath matches.
+     * @param xpath The XPath
+     * @return The objects, file by file
+     * @throws IOException If a file cannot be read
+     */
+    private Collection<XML> matching(final String xpath) throws IOException {
         final Collection<XML> found = new ArrayList<>(0);
         for (final XML xmir : this.documents()) {
-            found.addAll(
-                xmir.nodes(
-                    "//o[not(@base) and not(@name='λ') and not(text()[normalize-space()])]"
-                )
-            );
+            found.addAll(xmir.nodes(xpath));
         }
         return found;
     }
--- a/eo-inference/src/test/java/org/eolang/inference/CluesTest.java
+++ b/eo-inference/src/test/java/org/eolang/inference/CluesTest.java
@@ -42,6 +42,24 @@ final class CluesTest {
     }
 
     @Test
+    void writesTableOfWhatIsNeeded(@Mktmp final Path temp) throws IOException {
+        Files.writeString(
+            Files.createDirectories(temp.resolve("xmirs")).resolve("cup.xmir"),
+            String.join(
+                "",
+                "<object><o loc='Φ.cup.φ' base='.lid'>",
+                "<o loc='Φ.cup.φ.ρ' base='ξ.cup'/></o></object>"
+            )
+        );
+        new Clues().follow(temp.resolve("xmirs"), temp.resolve("tables"));
+        MatcherAssert.assertThat(
+            "the clues we know must write down what every object is asked for, but they didnt",
+            Files.exists(temp.resolve("tables").resolve("needs.xml")),
+            Matchers.is(true)
+        );
+    }
+
+    @Test
     void writesTableOfWhatIsProvided(@Mktmp final Path temp) throws IOException {
         Files.writeString(
             Files.createDirectories(temp.resolve("xmirs")).resolve("cup.xmir"),
--- a/eo-inference/src/test/java/org/eolang/inference/NeedsTest.java
+++ b/eo-inference/src/test/java/org/eolang/inference/NeedsTest.java
@@ -1,0 +1,133 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.matchers.XhtmlMatchers;
+import com.jcabi.xml.XML;
+import com.jcabi.xml.XMLDocument;
+import com.yegor256.Mktmp;
+import com.yegor256.MktmpResolver;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.hamcrest.MatcherAssert;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * Test case for {@link Needs}.
+ * @since 0.68.0
+ */
+@ExtendWith(MktmpResolver.class)
+final class NeedsTest {
+
+    @Test
+    void recordsAttributeTakenFromReceiver(@Mktmp final Path temp) throws IOException {
+        MatcherAssert.assertThat(
+            "the object a dispatch takes from must be asked for that attribute, but it isnt",
+            this.table(
+                temp,
+                String.join(
+                    "",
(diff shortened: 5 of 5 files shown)


maxonfjvipon:
@yegor256 please take a look

eo-inference/src/main/java/org/eolang/inference/Needs.java
```diff
@@ -0,0 +1,68 @@
...
+ * @since 0.68.0
+ */
+final class Needs implements Clue {
+
+    @Override
+    public void follow(final Path xmirs, final Path tables) throws IOException {
+        final Tojos rows = new TjCached(new TjDefault(new MnMemory()));
+        int seen = 0;
+        for (final XML dispatch : new Xmirs(xmirs).dispatches()) {
+            final String owner = dispatch.xpath("o[not(@as)][1]/@loc").get(0);
```
Favixx, line 52: `get(0)` on an empty list throws `IndexOutOfBoundsException` carrying nothing about which file or which object caused it. The description justifies this empirically — all 9,848 dispatches in eo-runtime have a receiver that carries `@loc` — but that is a fact about one codebase today, not a property of XMIR. A dispatch whose only child is written `as='ρ'` still matches `starts-with(@base, '.')` and has no `[not(@as)]` child at all, and this clue runs over whatever the compiler emitted. Checking the list and failing with the dispatch's own `@loc` in the message would turn a bare stack trace into something a user can act on.

eo-inference/src/main/java/org/eolang/inference/Needs.java
```diff
@@ -0,0 +1,68 @@
...
+            rows.add(owner).set("index", Integer.toString(seen));
+            seen = seen + 1;
+            rows.add(String.join(" ", owner, dispatch.xpath("@base").get(0)))
+                .set("owner", owner)
+                .set("index", Integer.toString(seen))
+                .set("name", dispatch.xpath("@base").get(0).substring(1))
+                .set("type", dispatch.xpath("@loc").get(0));
```
Favixx, line 59: This is the fourth `xpath(...).get(0)` in eight lines, and the one whose value becomes `@type` in the table — so a dispatch without `@loc` fails here rather than at the top of the loop, after two rows have already been written into `rows`. Hoisting the three lookups (`owner`, `base`, `loc`) into locals at the start of the body would give one place to validate and would stop the loop from leaving half a dispatch behind when it throws.

eo-inference/src/main/java/org/eolang/inference/Needs.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+final class Needs implements Clue {
+
+    @Override
+    public void follow(final Path xmirs, final Path tables) throws IOException {
+        final Tojos rows = new TjCached(new TjDefault(new MnMemory()));
+        int seen = 0;
+        for (final XML dispatch : new Xmirs(xmirs).dispatches()) {
+            final String owner = dispatch.xpath("o[not(@as)][1]/@loc").get(0);
+            rows.add(owner).set("index", Integer.toString(seen));
```
Favixx, line 53: `Xmirs.dispatches()` hands back nodes in document order, so for `x.next.foo` the outer `.foo` arrives first and its receiver `Φ.app.inc.φ.ρ` takes index 0. `Grouped.types()` sorts by that index, so `needs.xml` comes out with `Φ.app.inc.φ.ρ` before `Φ.app.inc.φ.ρ.ρ` — the reverse of the example in the PR description, which shows the innermost receiver first. `hasXPath` is order-blind so `NeedsTest` cannot catch it. Worth deciding which order is meant and then pinning it: if inner-first is the intended reading, the table is more useful when a receiver appears before the thing that dispatches on it.

eo-inference/src/main/java/org/eolang/inference/Needs.java
```diff
@@ -0,0 +1,68 @@
...
+            seen = seen + 1;
+            rows.add(String.join(" ", owner, dispatch.xpath("@base").get(0)))
```
Favixx, line 55: `@base` is read here for the row id and read again on line 58 for the name, and the id keeps the leading dot while `Provides` builds its id from the bare attribute name. Neither shows up in the output, so this is only about the two clues staying comparable — a single `final String name = dispatch.xpath("@base").get(0).substring(1);` above would remove the second lookup and make both ids `owner + " " + name`.

eo-inference/src/main/java/org/eolang/inference/Needs.java
```diff
@@ -0,0 +1,68 @@
...
+                .set("owner", owner)
+                .set("index", Integer.toString(seen))
+                .set("name", dispatch.xpath("@base").get(0).substring(1))
```
Favixx, line 58: `starts-with(@base, '.')` also matches a base that is exactly `.`, and `substring(1)` then puts an empty `@name` into the table with no complaint. If a bare dot cannot come out of the splitter, saying so in the xpath — `starts-with(@base, '.') and string-length(@base) > 1` — documents the assumption for free; if it can, this row is wrong and should be skipped.

eo-inference/src/main/java/org/eolang/inference/Needs.java
```diff
@@ -0,0 +1,68 @@
...
+import java.nio.file.Path;
+
+/**
+ * What every object must have, judging by how it is used.
+ *
+ * <p>This is the other half of {@link Provides}: a formation shows what its
+ * object certainly has, while a dispatch shows what somebody expects to find.
+ * In this fragment</p>
+ *
+ * <pre> &lt;o base=".next" loc="Φ.app.inc.φ.ρ"&gt;
```
Favixx, line 24: The snippet shows only the inner dispatch, which forces the paragraph below to describe the outer `.foo` in prose — "the `.foo` wrapped around this fragment" — for the rule that is the most interesting thing about this clue. The tests already have the two-level fragment; using the same four lines here would let the example carry the chain rule instead of talking around it. (The attributes are also ordered `base` then `loc` here and the other way round in the tests, which makes them harder to line up while reading.)

eo-inference/src/main/java/org/eolang/inference/Xmirs.java
```diff
@@ -53,13 +53,37 @@ final class Xmirs {
...
+     * attribute from the object below it. Every composite base was split
+     * into one such object per step before any clue looked at the XMIR, so
+     * a chain like {@code x.next.foo} is three objects here and not one.</p>
+     *
+     * @return The dispatches, file by file, in the order they appear in
+     *  the code
+     * @throws IOException If a file cannot be read
+     */
+    Collection<XML> dispatches() throws IOException {
+        return this.matching("//o[starts-with(@base, '.')]");
```
Favixx, line 74: This picks up dispatches anywhere in the document, including ones nested inside an object whose `@name` is `λ` — which `formations()` deliberately excludes on the line above. If a λ can carry a body with dispatches in it, they will be swept into `needs.xml` while the enclosing formation is absent from `provides.xml`, and the two tables will disagree about an object neither of them is wrong about. Worth confirming whether λ bodies reach this stage at all.

eo-inference/src/main/java/org/eolang/inference/Xmirs.java
```diff
@@ -53,13 +53,37 @@ final class Xmirs {
...
+
+    /**
+     * Every dispatch of the program.
+     *
+     * <p>A dispatch is an object whose base begins with a dot: it takes an
+     * attribute from the object below it. Every composite base was split
+     * into one such object per step before any clue looked at the XMIR, so
+     * a chain like {@code x.next.foo} is three objects here and not one.</p>
+     *
+     * @return The dispatches, file by file, in the order they appear in
```
Favixx, line 69: "In the order they appear in the code" is the guarantee `Needs` leans on for its `index`, but it is really produced three methods away, by the `.sorted()` in `sources()` — and only documented there as "in the same order every time", which is a weaker promise than this one. Since a future change to `sources()` would break `needs.xml` ordering silently, the two comments should probably say the same thing.

eo-inference/src/main/java/org/eolang/inference/Xmirs.java
```diff
@@ -53,13 +53,37 @@ final class Xmirs {
...
+        return this.matching("//o[starts-with(@base, '.')]");
+    }
+
+    /**
+     * Every object of the program the given XPath matches.
+     * @param xpath The XPath
+     * @return The objects, file by file
+     * @throws IOException If a file cannot be read
+     */
+    private Collection<XML> matching(final String xpath) throws IOException {
```
Favixx, line 83: Good extraction — the duplicated walk was the thing that would have gone wrong first. One note on cost: this parses each document again on every call, so with two clues every `.xmir` in eo-runtime is read and DOM-built twice, and three times once the third clue lands. `documents()` is already private and side-effect free, so memoising it in the field would make that flat.

eo-inference/src/main/java/org/eolang/inference/Clues.java
```diff
@@ -50,7 +62,7 @@ public final class Clues implements Clue {
      * Ctor.
      */
     public Clues() {
-        this(Collections.singletonList(new Provides()));
+        this(Arrays.asList(new Provides(), new Needs()));
```
Favixx, line 65: Each clue constructs its own `Xmirs` over the same directory, so this line is where the double parse of the whole tree starts. Since `follow` hands every clue the identical `xmirs` path, passing a shared `Xmirs` into the clues instead — or memoising inside it — keeps the cost constant as the third clue arrives, without changing what a clue is.

eo-inference/src/main/java/org/eolang/inference/Clues.java
```diff
@@ -26,9 +26,21 @@
...
+ * it must have judging by how it is used ({@link Needs}), and which types
+ * are copies of which. Keeping the tables apart is what lets a smarter
+ * rule add rows, or read them differently, without touching anything
+ * else.</p>
+ *
+ * <p>Every clue reads the program as text, writing a row for an object
+ * because it is <em>there</em>, not because something reaches it. The
+ * alternative — starting from a root and following what actually runs —
+ * gives sharper answers, since every question is then asked in a context
+ * where the types are concrete, but it cannot be used here: eo-runtime is
```
Favixx, line 38: This is the paragraph I would most want kept next to the code, because it is the reason the sweeps in `Xmirs` are plain `//o[...]` rather than a walk from a root. The one thing it leaves out is what the imprecision actually looks like downstream: reading the text means a row is written for `x.next` whether or not any `x` is ever a receiver, so a row in `needs.xml` is a claim about a *site*, not about an object. Saying that here would tell whoever writes the checking loop what they are allowed to assume when the two tables disagree.

eo-inference/src/main/java/org/eolang/inference/Clues.java
```diff
@@ -6,8 +6,8 @@
 
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Arrays;
```
Favixx, line 9: `Arrays.asList` returns a fixed-size view over the varargs array — length cannot change but elements can still be replaced through it, where `Collections.singletonList` was fully immutable. Whether that matters depends on what the secondary constructor does with the collection; if it stores it as-is, `Collections.unmodifiableList(Arrays.asList(...))` restores the property that was there before this PR.

eo-inference/src/test/java/org/eolang/inference/CluesTest.java
```diff
@@ -41,6 +41,24 @@ void followsEveryClue(@Mktmp final Path temp) throws IOException {
...
+            String.join(
+                "",
+                "<object><o loc='Φ.cup.φ' base='.lid'>",
+                "<o loc='Φ.cup.φ.ρ' base='ξ.cup'/></o></object>"
+            )
+        );
+        new Clues().follow(temp.resolve("xmirs"), temp.resolve("tables"));
+        MatcherAssert.assertThat(
+            "the clues we know must write down what every object is asked for, but they didnt",
+            Files.exists(temp.resolve("tables").resolve("needs.xml")),
```
Favixx, line 57: This asserts the file exists, so it would still pass if `Needs` wrote an empty `<needs/>`. `NeedsTest` covers content and this test is about wiring, which is fair — but the fixture already contains a dispatch, so asserting the row it implies (`type[@id='Φ.cup.φ.ρ']/attr[@name='lid']`) would catch a `Clues` that runs the clue without its output reaching the table, at no extra setup.

eo-inference/src/test/java/org/eolang/inference/NeedsTest.java
```diff
@@ -0,0 +1,133 @@
...
+                )
+            ),
+            XhtmlMatchers.hasXPath(
+                "/needs/type[@id='Φ.app.inc.φ.ρ.ρ']/attr[@name='next' and @type='Φ.app.inc.φ.ρ']"
+            )
+        );
+    }
+
+    @Test
+    void recordsEveryStepOfChain(@Mktmp final Path temp) throws IOException {
```
Favixx, line 47: The XMIR here is character-for-character the same as in `recordsAttributeTakenFromReceiver`; only the asserted xpath differs. That is a fixture worth hoisting into a constant, or turning both into a `@ParameterizedTest` over the two xpaths — as written, editing the fragment means editing it twice, and updating only one leaves a test that still passes while no longer testing the chain.

eo-inference/src/test/java/org/eolang/inference/NeedsTest.java
```diff
@@ -0,0 +1,133 @@
...
+                temp,
+                String.join(
+                    "",
+                    "<object><o loc='Φ.sum' base='.plus'>",
+                    "<o loc='Φ.sum.ρ' base='ξ.x'/>",
+                    "<o loc='Φ.sum.α0' as='α0' base='Φ.number'/>",
+                    "</o></object>"
+                )
+            ),
+            XhtmlMatchers.hasXPath("/needs/type[@id='Φ.sum.ρ']/attr[@name='plus']")
```
Favixx, line 80: Unlike the other two, this assertion stops at `@name='plus'` and never looks at `@type`, so it would keep passing if the dispatch's own locator dropped out of the row. `and @type='Φ.sum'` covers the last of the four `get(0)` calls in `follow` and costs one clause.

eo-inference/src/test/java/org/eolang/inference/NeedsTest.java
```diff
@@ -0,0 +1,133 @@
...
+    void asksReceiverAndNotArgument(@Mktmp final Path temp) throws IOException {
+        MatcherAssert.assertThat(
+            "the attribute is taken from the receiver, not from an argument, but it wasnt",
+            this.table(
+                temp,
+                String.join(
+                    "",
+                    "<object><o loc='Φ.sum' base='.plus'>",
+                    "<o loc='Φ.sum.ρ' base='ξ.x'/>",
+                    "<o loc='Φ.sum.α0' as='α0' base='Φ.number'/>",
```
Favixx, line 76: This pins `[not(@as)]`, which is the right thing to pin. What stays untested is `[1]`: it only starts doing work when a dispatch has two children without `@as`, and here there is exactly one. If that cannot happen, the `[1]` is noise worth dropping; if it can, the code is silently choosing one and discarding the other, and a test should record which choice is correct.

eo-inference/src/test/java/org/eolang/inference/NeedsTest.java
```diff
@@ -0,0 +1,133 @@
...
+            XhtmlMatchers.hasXPath("/needs/type[@id='Φ.sum.ρ']/attr[@name='plus']")
+        );
+    }
+
+    @Test
+    void skipsReference(@Mktmp final Path temp) throws IOException {
+        MatcherAssert.assertThat(
+            "a reference takes nothing from anybody, so it must ask for nothing, but it asked",
+            this.table(temp, "<object><o loc='Φ.pot' base='ξ.jar'/></object>"),
+            XhtmlMatchers.hasXPath("/needs[not(type)]")
```
Favixx, line 89: `skipsReference` covers an object that dispatches on nothing; the missing counterpart is a dispatch that has nothing to dispatch on — no child, or a child without `@loc`. That is the single input that makes `follow` throw, and nothing currently states what should happen. Even a test asserting it fails with a message naming the offending object would fix the behaviour in place.

Favixx (commented):
The shape of this is right: `Needs` mirrors `Provides` closely enough that `Grouped` renders it untouched, and letting the chain walk itself by relying on the split having happened already is much better than a recursive descent inside the clue. The paragraph added to `Clues` about reading text rather than following execution is worth having in the code rather than in a note.

Two things I would want settled before merge. First, `follow` does four `xpath(...).get(0)` calls with no guard; the description defends that with a census of eo-runtime, which is evidence about one codebase rather than an invariant of XMIR, and the failure mode is a bare `IndexOutOfBoundsException` with no locator in it. Second, the emitted order does not match the example in the description — document order puts the outer dispatch first, so `Φ.app.inc.φ.ρ` is written before `Φ.app.inc.φ.ρ.ρ`, and `hasXPath` cannot see the difference.

The rest is smaller: some duplicated lookups, a fixture copied between two tests, and a couple of assertions that stop just short of pinning what they are named after.

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+196](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (275 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+706](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6388, opened 2026-08-05
`string.regex` improperly handles `java.util.regex.Pattern`, affecting EO object behaviors


yegor256:
`eo-runtime/src/main/eo/string/regex.eo` declares two atoms: `compile` at line 29 and `matched-from-index` at line 58. Their JVM bodies sit in `eo-runtime/src/main/java/org/eolang/EO_string/EOregex$EOcompile.java:55` and `eo-runtime/src/main/java/org/eolang/EO_string/EOregex$EOpattern$EOmatch$EOmatched_from_index.java:70`, 278 lines of Java between them. They are the only two `EO*.java` atoms left in the `org.eolang.EO_string` package.

The design leaks the JVM into EO. `compile` builds a `java.util.regex.Pattern`, serializes it with `ObjectOutputStream`, and hands the raw Java serialization blob to EO as `pattern.serialized`; `matched-from-index` reads it back with `ObjectInputStream` and casts. Since `regex.eo` declares `+rt node` next to `+rt jvm`, no other runtime can produce or consume that blob. Java's matcher semantics have leaked through the same boundary again and again: #5631, #5071, #6062, #5779, #6040, #6065 and #6143 were all Java behaviour surfacing as EO behaviour, and #6189 exists because voids filled by these atoms carry no types.

We did this for `printf` already: #4752 was closed by `04fa5558c7`, which deleted 399 lines of Java and grew `string/printf.eo` by 358 lines of EO built out of `padded-left`, `padded-right`, `as-decimal`, `as-fixed` and `as-hex`. `regex` should follow. Let `compile` parse the Perl-format expression in EO — the slashes, then the flag letters — into a pattern object instead of an opaque blob, and let `matched-from-index` become an EO matcher walking the text against that object, keeping the current `matched` protocol (`position`, `start`, `from`, `to`, `groups`, `next`, `text`, `count`, `exists`). The tests already sitting in `regex.eo:89-343` are the acceptance criteria: they cover the `/d /i /m /s /u /x` flags, supplementary characters, zero-width matches, group exposure and every failure branch, and they should pass unchanged.

0crat:
@yegor256 Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+1961](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6472, opened 2026-08-07
#6388: parse the /pattern/flags envelope in EO, not Java

morphqdd:
Fixes the reported case for #6388; see the added @todo chain for the remaining part.

string.regex leaks java.util.regex.Pattern/Matcher into EO end to end. compile was a pure Java atom doing three things: validating the leading/trailing slash, cutting the flag letters off the tail, and compiling the assembled string. Only the last part genuinely needs the JVM.

compile is now a real EO formation. It validates the envelope and builds the (?flags)body search string entirely in EO, at the byte level (raw, span, opens, last via a recursive scan, body, flags, search) rather than through the character-aware string.slice/length/last-index-of helpers. Those walk UTF-8 recursively: one compile call cost 450ms instead of 0.25ms through them, which turned scanf's per-token regex.compile loop into a 41-minute hang in EOscanfTest during an earlier attempt at this same slice. Byte-level cuts are equivalent here since / is ASCII and UTF-8 is self-synchronizing.

The only remaining Java is a new nested atom, EOregex$EOcompile$EOassembled: Pattern.compile(search) and serialize, routing PatternSyntaxException to the same cant-compile fallback compile already threads through for the slash-validation errors. It declares no void of its own, reading search/cant-compile off compile and pattern off compile's own parent through Phi.RHO.

This is the first slice of the rewrite the issue asks for. The remaining @todo chain on EOregex$EOpattern$EOmatch$EOmatched_from_index (literal matching through anchors, character classes, quantifiers, groups, alternation, flags, supplementary characters, and the final cutover) is unchanged, since matching itself still goes through java.util.regex.

Verified: EOregexTest 37/37, EOregexEOcompileTest 2/2, EOregexEOpatternEOmatchEOmatchedfromindexTest 8/8, EOscanfTest 25/25 (about 102s, was a 41-minute hang on the earlier attempt), PhDefaultTest 55/55 (1 pre-existing skip), qulice clean.


Thayorns (approved):
The change is documentation only: both regex atoms get a chain of puzzles laying out the path from the JVM Pattern/Matcher pair to an EO-native engine, ending with the step that deletes both classes and the +rt split. The links resolve — fill(Phi, Matcher, String) is at line 138 of the same class and Matcher is imported at line 15 — and no code path moved, so behaviour is untouched.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+1006](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

morphqdd:
Closing this. It cannot be brought back on top of master as one pull request: the EO rewrite of `compile` alone is 278 changed lines in `string/regex.eo`, well past the 200-line limit, and the branch also carries a `+unlint redundant-object` that master has since removed everywhere (`#6855`). On top of that the string package was merged (`#6657`) and the parent hops rewritten (`#7094`), so the rebase conflicts across `regex.eo`, both generated atoms and two tests.

#6388 stays open. The rewrite is worth redoing in slices that each stand on their own - envelope validation first, then the flag assembly, then the matching - each small enough to review.

PR #6473, opened 2026-08-07
#6445: reject a negative write offset in Heaps

### LINKED ISSUE #6445: `Heaps.write()` allows negative offset, causing ArrayIndexOutOfBoundsException
## What happens

Heaps.fits() correctly treats a negative offset as invalid for reads, but Heaps.write() does not validate offset before arraycopy. A negative offset can bypass the resize branch and then leaks ArrayIndexOutOfBoundsException.

## Reproduction

Allocate a block and call Heaps.INSTANCE.write(identifier, -1, new byte[] {1}). Current master throws ArrayIndexOutOfBoundsException.

## Root cause

Heaps.write() checks only whether offset plus data length exceeds Integer.MAX_VALUE. It does not reject a negative offset before System.arraycopy().

## Expected behavior

Writes should follow the same non-negative offset rule as reads and fail through the runtime error path instead of leaking a Java array exception.

asmirnov-backend:
Closes #6445.

`Heaps.write()` validated only the upper bound, so a negative offset slipped through to `System.arraycopy()` and leaked `ArrayIndexOutOfBoundsException`. Reads already reject a negative offset via `fits()`.

Now `write()` fails with `ExFailure` before touching the block, same contract as `resize()` and `read()`.

@yegor256 please review

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/HeapsTest.java
+++ b/eo-runtime/src/test/java/org/eolang/HeapsTest.java
@@ -85,6 +85,34 @@ final class HeapsTest {
     }
 
     @Test
+    void failsCleanlyOnWriteWithNegativeOffset() {
+        final int idx = Heaps.INSTANCE.malloc(new HeapsTest.PhFake(), 10);
+        Assertions.assertThrows(
+            ExFailure.class,
+            () -> Heaps.INSTANCE.write(idx, -1, new byte[] {0x01}),
+            "Heaps must reject a negative write offset with a clean ExFailure, not a raw JVM exception"
+        );
+        Heaps.INSTANCE.free(idx);
+    }
+
+    @Test
+    void keepsBlockIntactAfterWriteWithNegativeOffset() {
+        final int idx = Heaps.INSTANCE.malloc(new HeapsTest.PhFake(), 3);
+        Heaps.INSTANCE.write(idx, 0, new byte[] {7, 8, 9});
+        Assertions.assertThrows(
+            ExFailure.class,
+            () -> Heaps.INSTANCE.write(idx, -2, new byte[] {1, 2}),
+            "Heaps must reject a negative write offset before touching the block, but it didn't"
+        );
+        MatcherAssert.assertThat(
+            "Heaps must leave the block untouched after a rejected negative-offset write, but it didn't",
+            Heaps.INSTANCE.read(idx, 0, 3),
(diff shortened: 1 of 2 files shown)


eo-runtime/src/test/java/org/eolang/HeapsTest.java
```diff
@@ -84,6 +84,34 @@ void failsOnWriteIfOffsetPlusLengthOverflows() {
...
+    @Test
+    void keepsBlockIntactAfterWriteWithNegativeOffset() {
+        final int idx = Heaps.INSTANCE.malloc(new HeapsTest.PhFake(), 3);
+        Heaps.INSTANCE.write(idx, 0, new byte[] {7, 8, 9});
+        Assertions.assertThrows(
+            ExFailure.class,
+            () -> Heaps.INSTANCE.write(idx, -2, new byte[] {1, 2}),
+            "Heaps must reject a negative write offset before touching the block, but it didn't"
+        );
+        MatcherAssert.assertThat(
```
Thayorns, line 107: keepsBlockIntactAfterWriteWithNegativeOffset checks two things: that the write throws (line 102) and that the bytes survive (line 107). The first of those is already the whole of failsCleanlyOnWriteWithNegativeOffset above it, so when this test goes red you cannot tell from the name which of the two claims broke, and the name only promises the second one. The free(idx) after line 107 also means the assertion is not the last statement here.

Thayorns (commented):
The guard closes the hole the issue names: a negative offset now stops at the top of write() with an ExFailure instead of reaching System.arraycopy() and surfacing as ArrayIndexOutOfBoundsException, and it sits inside the lock alongside the other checks, so it reads consistently with resize() and read(). One remark on how the second test is put together.

0crat:
@asmirnov-backend Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (36) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+80](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+981](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6474, opened 2026-08-07
#6450: escape string literals rendered in φ-terms

### LINKED ISSUE #6450: `PhDefault.φTerm()` fails to escape quotes, backslashes and control characters in string literals
## What happens

[`PhDefault.φTerm()`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/PhDefault.java#L330-L336) directly wraps UTF-8 text in double quotes:

```java
result = String.format("\"%s\"", new String(raw, StandardCharsets.UTF_8));
```

Unlike a string-literal renderer, it does not escape a quote, a backslash, or a control character. The produced φ-term can therefore be syntactically invalid or represent a different string.

## Runtime reproduction

```java
new Data.ToPhi("a\"b\\c\n").φTerm()
```

Actual output:

```text
"a"b\c
"
```

Here the inner quote ends the rendered literal, the backslash is raw, and the line feed is embedded as an actual newline.

## Expected behavior

The result should be a valid, unambiguous φ-term, for example:

```text
"a\"b\\c\n"
```

The escaping should match the literal representation used elsewhere in the runtime. This is separate from #6375, which covers the analogous unescaped rendering path in `PhApplication.φTerm()`.

asmirnov-backend:
Closes #6450.

`PhDefault.φTerm()` wrapped raw UTF-8 text in quotes, so a quote closed the literal early, a backslash stayed raw and a line feed broke the term across lines.

Now the text goes through `Quoted`, which spells every glyph with `Escaped` (`\"`, `\\`, `\b`, `\f`, `\n`, `\r`, `\t`, `\uXXXX` for the rest of the control characters), matching what the parser decodes back.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Escaped.java
+++ b/eo-runtime/src/main/java/org/eolang/Escaped.java
@@ -1,0 +1,54 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang;
+
+import java.util.function.Supplier;
+
+/**
+ * One glyph spelled the way an EO string literal carries it.
+ *
+ * <p>A quote, a backslash and a control character cannot stand inside a
+ * literal as they are: the quote closes it, the backslash swallows its
+ * neighbour, and a line feed breaks the literal in two. Each of them turns
+ * into the escape sequence the parser decodes back into the very same
+ * glyph (R-9.7), while everything else stays verbatim.</p>
+ *
+ * <p>The glyphs from {@code 0x08} to {@code 0x0D} carry the one-letter
+ * spellings of {@code "btnvfr"}, except the vertical tab {@code 0x0B},
+ * which EO has no letter for and which goes out as a unicode escape.</p>
+ *
+ * @since 0.73.3
+ */
+final class Escaped implements Supplier<String> {
+
+    /**
+     * The glyph.
+     */
+    private final char glyph;
+
+    /**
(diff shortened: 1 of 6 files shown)


Thayorns (approved):
The escaping matches what the parser decodes back: Emissions.singleCharEscape takes exactly n, t, r, b, f, quote and backslash, and appendUnicode reads the four hex digits case-insensitively through Character.digit, so the lowercase \u00xx form round-trips. Quoted copies the byte array it is handed and Escaped keeps one glyph, so both stay closed over their own state.

One surface nothing pins down yet: Quoted walks UTF-16 code units, so a supplementary character arrives as two surrogates that each fall through the else branch and get reassembled by the StringBuilder. That is the right outcome, but no test holds it in place, and it is the case most likely to break if the else branch ever grows a condition.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+998](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@asmirnov-backend Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (201 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+92](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6475, opened 2026-08-07
Use quotient to select number.mod fast path

### LINKED ISSUE #6455: `number.mod` fast path miscalculates remainder for large dividends with small divisors

androvonx95:
Closes #6455.

Gate the `number.mod` fast path on the quotient instead of the dividend so larger quotients fall back to the existing reduction algorithm.

### DIFF
--- a/eo-runtime/src/main/eo/number/mod.eo
+++ b/eo-runtime/src/main/eo/number/mod.eo
@@ -34,8 +34,10 @@
         [] >> abs-mod
           if. > @
             lt.
-              abs dividend >> absx
-              4503599627370496
+              div.
+                abs dividend >> absx
+                absy
+              9007199254740992
             absx.minus
               times.
                 abs divisor >> absy


eo-runtime/src/main/eo/number/mod.eo
```diff
@@ -34,8 +34,10 @@
         [] >> abs-mod
           if. > @
             lt.
-              abs dividend >> absx
-              4503599627370496
+              div.
+                abs dividend >> absx
+                absy
+              9007199254740992
```
Thayorns, line 40: Gating on the quotient at 2^53 pulls dividends above 2^52 into the fast path, and there `absy.times (floor. (absx.div absy))` is no longer exact. For 10000000000000000 over 13 the quotient is about 7.7e14 so the fast path is taken, floor gives 769230769230769, and 13 times that is 9999999999999997, which is odd and above 2^53 so it rounds to 9999999999999996 and the answer comes out 4. can-take-an-exact-modulo-of-a-large-dividend-by-thirteen asserts 3. The by-three test goes the same way: the product rounds up to 1e16 and the result is 0 where the test asserts 1. Both of those used to reach reduce because 1e16 is above the old 2^52 bound.

eo-runtime/src/main/eo/number/mod.eo
```diff
@@ -34,8 +34,10 @@
         [] >> abs-mod
           if. > @
             lt.
-              abs dividend >> absx
-              4503599627370496
+              div.
```
Thayorns, line 37: 1e15 mod 1e-5, the case the issue leads with, has a quotient of 1e20, so it now falls through to reduce 62. That loop walks chunk levels 62 down to 0, so it can subtract at most absy times (2^63 - 1), about 9.2e13, from a dividend of 1e15 and returns roughly 9.08e14. That is still far larger than the divisor, so the |r| < |y| invariant is broken exactly as before, just with a different wrong number. Anything whose quotient exceeds 2^63 has no correct path in this object, and routing it to reduce does not change that.

Thayorns (commented):
The change moves the fast-path guard from the dividend to the quotient, which is the right quantity to gate on, but the 2^53 bound admits products that no longer round exactly, and the fallback still cannot handle the case the issue opens with. No test comes with the change, so nothing in the file pins the fractional-divisor behavior this is about, and the comment at the top of the file still describes the old 2^52-dividend rule.

androvonx95:
The doubling/halving approach in #6470 provides the more complete fix for #6455, including cases where the quotient exceeds the reduction limit. Closing this PR in favor of the merged solution.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (2) comments; -4 for too few (6) hits-of-code; +1 to give you at least something.  Your running score is [+992](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6478, opened 2026-08-08
#6428: resolve the fan-out suppression puzzle in MjSafe.java

### LINKED ISSUE #6428: Fan-out suppression in MjSafe.java needs removal

morphqdd:
Closes #6428.

The puzzle 6125-d751ea55 flagged MjSafe.caching(String): its GlobalCache return type and GcShared/GlobalCache.GcFresh construction pushed the class from 30 referenced types, the checkstyle limit, to 32, forcing a ClassFanOutComplexityCheck suppression.

Extracting caching() into its own Caching class only recovered one of the two points: MjSafe stopped referencing GcShared but picked up a reference to Caching itself, landing at 31, still over.

The remaining point belonged to Parsing, the only real consumer of the cache. Moved Caching's construction there, so MjSafe and MjParse now just pass the three primitives they already had (cache enabled flag, cache directory, compiler version) and never mention Caching, GlobalCache or GcShared by name. Parsing already owned a GlobalCache field for its own reasons, so the type nesting belongs there.

Caching.store(String) is resolved lazily at Parsing's one use site rather than in its constructor, per qulice's ConstructorsCodeFreeCheck (constructors assign fields only, no method calls).

Fan-out on MjSafe is back at 30. Both the suppression and its @todo puzzle are removed. No behavior changed, this is a pure move; existing coverage (MjParseTest, MjSafeTest, CompilingTest) confirms it, no new test needed.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Caching.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Caching.java
@@ -1,0 +1,63 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import java.nio.file.Path;
+
+/**
+ * The cache of one step, as configured by the user.
+ *
+ * <p>This is the only place where {@code eo.cacheEnabled} is read, so that no
+ * step has to know that the option exists. Kept out of {@link MjSafe} on
+ * purpose: its two possible outcomes ({@link GlobalCache} and
+ * {@link GcShared}) are exactly the two types that pushed that class over
+ * its class-fan-out limit.</p>
+ *
+ * @since 0.1
+ */
+final class Caching {
+
+    /**
+     * Whether the machine-wide cache is enabled.
+     */
+    private final boolean enabled;
+
+    /**
+     * Directory of the machine-wide cache.
+     */
+    private final Path cache;
+
+    /**
+     * Version of the compiler, folded into the cache key.
+     */
+    private final String version;
+
+    /**
+     * Ctor.
+     * @param enabled Whether the machine-wide cache is enabled
+     * @param cache Directory of the machine-wide cache
+     * @param version Version of the compiler, folded into the cache key
+     */
+    Caching(final boolean enabled, final Path cache, final String version) {
+        this.enabled = enabled;
+        this.cache = cache;
+        this.version = version;
+    }
+
+    /**
+     * The cache of one step.
+     * @param sub Directory of that step inside the machine-wide cache
+     * @return The cache of that step
+     */
+    GlobalCache store(final String sub) {
+        final GlobalCache result;
+        if (this.enabled) {
+            result = new GcShared(this.cache.resolve(sub), this.version);
+        } else {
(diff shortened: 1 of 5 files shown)


eo-maven-plugin/src/main/java/org/eolang/maven/Parsing.java
```diff
@@ -80,25 +80,30 @@ final class Parsing implements Step {
...
-        final GlobalCache store
+        final boolean enabled,
+        final Path store,
+        final String version
     ) {
         this.tojos = srcs;
         this.targetDir = target;
         this.sourcesDir = sources;
-        this.cache = store;
+        this.cache = new Caching(enabled, store, version);
```
Thayorns, line 106: Parsing now assembles its own Caching from three primitives instead of being handed a finished one, and that shows up in three places in this diff. The constructor stops being pure assignment. It needs six parameters, which is why line 93 carries a ParameterNumberCheck suppression, so the fan-out suppression left MjSafe and a parameter-count suppression arrived here instead. And the cache is no longer something a caller can choose: CompilingTest used to pass new GlobalCache.GcFresh() directly and now passes false, temp, Parsing.ZERO and relies on Caching.store to arrive at GcFresh on its own, so no test can substitute a cache any more.

eo-maven-plugin/src/main/java/org/eolang/maven/Caching.java
```diff
@@ -0,0 +1,63 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import java.nio.file.Path;
+
+/**
+ * The cache of one step, as configured by the user.
+ *
+ * <p>This is the only place where {@code eo.cacheEnabled} is read, so that no
```
Thayorns, line 12: This docblock says no step has to know the option exists, but Parsing implements Step at Parsing.java:46 and its constructor now takes that very flag as enabled at Parsing.java:99. The sentence above it doesn't hold either: MjSafe reads this.cacheEnabled at MjSafe.java:539 and MjParse at MjParse.java:49, both to fill that parameter, so Caching is one reader of the flag among three rather than the only one. Whoever reads this next will trust a boundary the code no longer has.

Thayorns (commented):
This moves the cache decision out of MjSafe.caching() into a new Caching class and constructs it inside Parsing, dropping the fan-out suppression and its puzzle from MjSafe. The move does bring MjSafe's referenced-type count down, but it does it by pushing the raw eo.cacheEnabled flag one level deeper, into a Step, and the constructor that receives it needs a suppression of its own. Two notes below on that.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+988](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6479, opened 2026-08-08
#6427: treat a blank line before an unrelated sibling as a separator

### LINKED ISSUE #6427: `Recovery` incorrectly handles blank lines causing suppressed R-6.5.4 diagnostic on subsequent objects

morphqdd:
Closes #6427.

This branch is stacked on #6467 (issue #6426): both touch Recovery.java, so this diff will show both PRs' changes combined until #6467 merges. Please review/merge #6467 first; I will rebase this one afterward so it shows only its own diff.

Recovery.skipped treated every blank line as belonging to a failed line's block, unconditionally. That is correct for a blank genuinely nested inside the block, but it also swallows a blank that is really a file-level separator between the block and the next unrelated top-level object. Since the swallowed blank never reaches Eo.process, globals never records it, so Blanks.checkPlain on the following object sees no preceding blank at all and silently drops a genuine R-6.5.4 violation on it.

skipped now looks past a blank line to the next real line and bases the decision on that line's indent: still part of the block only if that line is indented deeper than the failed line, a separator otherwise. A separator is no longer skipped, so it reaches Eo.process normally and globals records it like any other blank.

Added a yaml pack test reproducing the issue's exact repro (a genuine parse error plus a blank-separated sibling that must still get its own R-6.5.4 diagnostic), and a direct unit test on Recovery.after for the same scenario. Verified both fail on the pre-fix code and pass with the fix.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Recovery.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Recovery.java
@@ -15,7 +15,10 @@ import java.util.List;
  * standing under its parent. This object holds the lines of the program
  * and answers where the walk picks up again: the first line standing
  * back at or above the indent of the one that failed. A blank line
- * carries no indent of its own and never ends the skipped block.</p>
+ * carries no indent of its own, so it is resolved by looking past it to
+ * the next real line: genuinely nested inside the block if that line is
+ * indented deeper, but a file-level separator in front of an unrelated
+ * sibling otherwise — and a separator does not extend the skip.</p>
  *
  * <p>The blanks trailing the block are handed back, though. A blank
  * standing between the last skipped line and the line the walk resumes
@@ -102,6 +105,31 @@ final class Recovery {
      */
     private boolean skipped(final int idx, final int indent) {
         final Span span = this.spans.get(idx);
-        return span.blank() || span.indent() > indent;
+        final boolean result;
+        if (span.blank()) {
+            result = this.blankBelongsToBlock(idx, indent);
+        } else {
+            result = span.indent() > indent;
+        }
+        return result;
+    }
+
+    /**
+     * Whether a blank line at {@code idx} still belongs to the block of a
+     * line at {@code indent} — true only when the next non-blank line is
+     * itself indented deeper than {@code indent}. A blank line whose next
+     * real line stands back at or above {@code indent} is a file-level
+     * separator between the block and an unrelated sibling, not part of
+     * the block, and must not be skipped.
+     * @param idx Index of the blank line
+     * @param indent Indent of the failed line
+     * @return True when the blank line must be skipped
+     */
+    private boolean blankBelongsToBlock(final int idx, final int indent) {
+        int next = idx + 1;
+        while (next < this.spans.size() && this.spans.get(next).blank()) {
+            next = next + 1;
+        }
+        return next < this.spans.size() && this.spans.get(next).indent() > indent;
     }
 }
--- a/eo-parser/src/main/java/org/eolang/parser/Recovery.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Recovery.java
@@ -76,7 +76,7 @@ final class Recovery {
             idx = idx + 1;
         }
         if (idx < this.spans.size()) {
-            idx = this.rewound(idx, failed);
+            idx = this.rewound(idx, from - 1);
         }
         return idx;
     }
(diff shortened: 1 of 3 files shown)


eo-parser/src/test/java/org/eolang/parser/RecoveryTest.java
```diff
@@ -47,4 +47,38 @@ void skipsBlankLineInsideBlock() {
...
     }
+
+    @Test
+    void stopsAtABlankLineThatSeparatesTheBlockFromAnUnrelatedSibling() {
+        MatcherAssert.assertThat(
+            "a blank line before an unrelated sibling is a separator, not part of the block"
+                .concat(", and must not be skipped"),
+            new Recovery(
+                Arrays.asList(
+                    new Span("bad!!!", 1),
```
Thayorns, line 89: The spans in this test are short, plain, well-formed ASCII, while resumesAtSiblingOfFailedLine at line 24 in the same file feeds the same object "über! ☃" and "क्ष". Nothing in the separator path depends on the body of a line, only on its indent and its blankness, so there is no reason for the one test covering the new branch to be the tame one in the file.

eo-parser/src/main/java/org/eolang/parser/Recovery.java
```diff
@@ -58,6 +76,31 @@ int after(final int failed) {
...
+     * line at {@code indent} — true only when the next non-blank line is
+     * itself indented deeper than {@code indent}. A blank line whose next
+     * real line stands back at or above {@code indent} is a file-level
+     * separator between the block and an unrelated sibling, not part of
+     * the block, and must not be skipped.
+     * @param idx Index of the blank line
+     * @param indent Indent of the failed line
+     * @return True when the blank line must be skipped
+     */
+    private boolean blankBelongsToBlock(final int idx, final int indent) {
```
Thayorns, line 128: blankBelongsToBlock is a query — it changes nothing and returns a boolean — but it is named as a whole statement, which makes it the odd one out next to skipped at line 77 and after at line 46. A reader scanning the class sees two names that say what a line is and one that argues a case.

Thayorns (commented):
The change makes Recovery resolve a blank line by looking past it to the next real line, so a blank standing between a skipped block and a sibling at or above the failed indent stops the skip instead of being swallowed. I walked the cases: a blank followed by a deeper line still extends the block, two blanks in a row still reach LnBlank and get their R-6.5.3 error, and a blank at the end of the file simply resumes the walk on itself with nothing left to diagnose, so the logic holds. I kept this to the lines this branch adds on top of #6467, which I am looking at separately. Two small things below.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+1052](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6480, opened 2026-08-08
#6450: escape quotes, backslashes and control chars in PhDefault.phiTerm()

### LINKED ISSUE #6450: `PhDefault.φTerm()` fails to escape quotes, backslashes and control characters in string literals

morphqdd:
Closes #6450.

PhDefault.phiTerm() wrapped a string value's raw UTF-8 bytes in double quotes with no escaping, so a quote in the value ended the rendered literal early, a backslash came through raw, and a newline broke the term across lines. The printed phi-term could be syntactically invalid or represent a different string entirely.

Added an escape pass: a quote or backslash gets a backslash prefix, newline/carriage-return/tab get their short forms, and any other control character gets a \u escape, matching the shorthand VerboseBytesAsString already uses for the general case.

This does not touch PhApplication.phiTerm(), which has the same gap (#6375). Separate object, separate issue.

Added a test with the issue's exact repro (a quote, a backslash and a newline in one string) asserting the correctly escaped output.

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/DataTest.java
+++ b/eo-runtime/src/test/java/org/eolang/DataTest.java
@@ -65,6 +65,15 @@ final class DataTest {
     }
 
     @Test
+    void escapesQuotesBackslashesAndControlCharsInTerm() {
+        MatcherAssert.assertThat(
+            "A quote, a backslash and a control character must be escaped in φ-term",
+            new Data.ToPhi("a\"b\\c".concat(String.valueOf('\n'))).φTerm(),
+            Matchers.equalTo("\"a\\\"b\\\\c\\n\"")
+        );
+    }
+
+    @Test
     void distinguishesDifferentNumbersInTerm() {
         MatcherAssert.assertThat(
             "Different numbers must produce different φ-terms, but they didnt",
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -333,7 +333,9 @@ public class PhDefault implements Phi, Cloneable {
         if (this.literal(name)) {
             final byte[] raw = this.loaded().get("as-bytes").get().delta();
             if ("string".equals(name)) {
-                result = String.format("\"%s\"", new String(raw, StandardCharsets.UTF_8));
+                result = String.format(
+                    "\"%s\"", this.escaped(new String(raw, StandardCharsets.UTF_8))
+                );
             } else {
                 result = PhDefault.numeral(new BytesOf(raw).asNumber());
             }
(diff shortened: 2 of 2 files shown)


eo-runtime/src/test/java/org/eolang/DataTest.java
```diff
@@ -64,6 +64,15 @@ void printsStringValueAsTerm() {
         );
     }
 
+    @Test
+    void escapesQuotesBackslashesAndControlCharsInTerm() {
+        MatcherAssert.assertThat(
+            "A quote, a backslash and a control character must be escaped in φ-term",
+            new Data.ToPhi("a\"b\\c".concat(String.valueOf('\n'))).φTerm(),
```
Thayorns, line 71: This input reaches three of the six branches in escaped: the quote, the backslash and \n. The \r branch, the \t branch and the \u%04x fallback at PhDefault.java:383 are entered by no test in this class, so a wrong short form or a wrong hex width in any of them would ship green. Carrying a carriage return, a tab and one character below 0x20 in the same input would cover the whole method under the one assertion you already have.

Thayorns (commented):
PhDefault.phiTerm() now runs a string value through an escape pass before wrapping it in quotes. I checked the escape set against the parser: \n, \t, \r, \" and \\ are all recognised by Emissions.singleCharEscape, and appendUnicode reads four hex digits case-insensitively, so the lowercase \u%04x form reads back correctly. One thing about the test.

yegor256:
@morphqdd conflicts here

morphqdd:
Closing this one. PR #6474 already fixed this in master (merged before this PR was ready), with the same escaping behavior via `Quoted`/`Escaped`.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+1006](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6481, opened 2026-08-08
#6449: reject a negative requested size in posix.read and win32.read

### LINKED ISSUE #6449: Negative requested size in posix.read and win32.read results in NegativeArraySizeException instead of EO failure

morphqdd:
Closes #6449.

ReadSyscall.make() and ReadFuncCall.make() allocated a buffer from an unvalidated size argument. A negative EO argument escaped as a raw NegativeArraySizeException instead of the runtime's controlled EO failure path, matching the check WriteSyscall already does for its own size argument.

Added the same style of check, before the buffer is allocated or the native function is called.

Added an EO-level test to each of posix.eo/win32.eo mirroring the existing stops-on-write-size-being-negative precedent, plus a focused Java test per class. The EO --> test only asserts "throws something" (the transpiler generates Assertions.assertThrows(Exception.class, ...)), so it cannot tell a controlled ExFailure apart from the original NegativeArraySizeException. The Java tests assert the specific exception type and fail on the un-fixed code with exactly the NegativeArraySizeException the issue reports.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/posix/ReadSyscall.java
+++ b/eo-runtime/src/main/java/org/eolang/posix/ReadSyscall.java
@@ -7,6 +7,7 @@ package org.eolang.posix;
 import java.util.Arrays;
 import org.eolang.Data;
 import org.eolang.Dataized;
+import org.eolang.ExFailure;
 import org.eolang.Phi;
 import org.eolang.Syscall;
 
@@ -32,6 +33,9 @@ public final class ReadSyscall implements Syscall {
     @Override
     public Phi make(final Phi... params) {
         final int size = new Dataized(params[1]).asNumber().intValue();
+        if (size < 0) {
+            throw new ExFailure("Can't read a negative number of bytes: %d", size);
+        }
         final Phi result = this.posix.take("return").copy();
         final byte[] buf = new byte[size];
         final int count = CStdLib.INSTANCE.read(
--- a/eo-runtime/src/main/eo/posix.eo
+++ b/eo-runtime/src/main/eo/posix.eo
@@ -118,6 +118,23 @@
           "127.0.0.1"
       16777343
 
+  --> stops-on-read-size-being-negative
+    os.is-windows.if > @
+      1.div 0
+      seq *
+        code.
+          posix *1
+            "read"
+            fd
+            -1
+        true
+    code. > fd
+      posix *1
+        "open"
+        "/dev/null"
+        0
+        0
+
   --> stops-on-write-size-being-negative
     os.is-windows.if > @
       1.div 0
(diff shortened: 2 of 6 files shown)


eo-runtime/src/main/java/org/eolang/posix/ReadSyscall.java
```diff
@@ -32,6 +33,9 @@ public ReadSyscall(final Phi posix) {
     @Override
     public Phi make(final Phi... params) {
         final int size = new Dataized(params[1]).asNumber().intValue();
+        if (size < 0) {
```
Thayorns, line 36: This closes the bottom of the range and leaves the top open. Line 35 narrows a double to an int by saturation, so posix.read fd 1e18 arrives here as 2147483647, sails past size < 0, and hits new byte[size] on line 40 as an OutOfMemoryError — the same uncontrolled JVM failure instead of an ExFailure that this change exists to prevent. WriteSyscall bounds both ends at WriteSyscall.java:37 because it has a buffer to measure against; read has none, so the ceiling has to be stated here. ReadFuncCall.make:36 has the identical shape.

Thayorns (commented):
Both read wrappers now reject a negative size before allocating, and the check sits ahead of the native call, so the win32 one can be exercised off Windows without loading Msvcrt. The EO packs follow the existing stops-on-write-size-being-negative shape correctly. One gap in the guard itself, below.

yegor256:
@morphqdd conflicts here

morphqdd:
Closing this one. PR #6465 already fixed this in master (merged before this PR was ready), rejecting negative sizes in posix.read and win32.read.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+999](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6485, opened 2026-08-08
`Recovery.skip` references undefined `failed` variable, breaks eo-parser compile


morphqdd:
## What happens

`eo-parser` does not compile from a clean `master` checkout. `Recovery.skip(int from, int indent)` in `eo-parser/src/main/java/org/eolang/parser/Recovery.java:76` calls `this.rewound(idx, failed)`, but `failed` is not a parameter, local variable, or field of `skip` or of `Recovery` at all. This is a plain "cannot find symbol" compile error.

Confirmed with:
```
mvn -ntp -pl eo-parser compile -DskipTests
```
```
[ERROR] Recovery.java:[76,37] cannot find symbol
[ERROR]   symbol:   variable failed
[ERROR]   location: class org.eolang.parser.Recovery
```

This landed on `master` via `0b2d7df6f5` (PR #6467), which extracted `skip(from, indent)` out of `after(failed)` but left a stale reference to `failed` where `from - 1` was meant (the equivalent boundary `after` used to pass).

## What should happen

`Recovery.skip` should compile and keep its original rewind behavior: `idx = this.rewound(idx, from - 1);`. Verified locally: `eo-parser` compiles clean and `RecoveryTest`, `EoTest`, `EoSyntaxTest` all pass (1149 tests, 0 failures) with this one-line fix.

morphqdd:
@yegor256, please pay attention

morphqdd:
@yegor256, @maxonfjvipon, please pay attention

0crat:
@morphqdd Thanks for submitting this ticket! However, as you see, it was not accepted by the team, which is sad. You've earned -8 points. By reporting bugs or suggesting enhancements, you help our project improve its quality. Please, don't be discouraged. Instead, try to be more careful next time, and [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) bugs. Your running score is [+3553](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6491, opened 2026-08-08
Write down which types are copies of which

### LINKED ISSUE #6490: eo-inference module fails to create a links table connecting 'Provides' and 'Needs'
Two of the four rules of `inference.md` are in place: `Provides` writes
what every object certainly has, `Needs` writes what somebody expects to
find. The two tables still cannot be compared, because nothing connects
them. In the note's example `needs.xml` says that the object at
`Φ.app.inc.φ.ρ.ρ` is asked for `next`, and `provides.xml` says that
`Φ.app.t` has one — but that the void `x` and the argument `t` are the same
object at runtime is written down nowhere.

Rule two of the note fills that in: a reference says that one type is a
copy of another. The references of the example give three rows:

```text
Φ.app.φ         is a copy of Φ.app.inc
Φ.app.φ.α0      is a copy of Φ.app.t
Φ.app.inc.φ.ρ.ρ is a copy of Φ.app.inc.x
```

The note is explicit about why these are records rather than a rewrite of
the XMIR: later, "a copy of" stops meaning "the same as" and starts meaning
"a fresh duplicate with its own types", and that is the only rule which
then changes.

This is the first clue that has to resolve a name. A reference is
`base="ξ.t"`, pointing at an attribute of the enclosing formation, or
`base="Φ.number"`, pointing at a root object; in both cases the target is a
locator, so the clue needs an index of what the world contains. The
existing clues walk objects and need no such index — `Xmirs`
(`eo-inference/src/main/java/org/eolang/inference/Xmirs.java:55`) answers
`formations()` and `dispatches()` and nothing more. The checking loop will
want the same index, so it is not built for this rule alone.

A third clue alongside the two in `Clues`
(`eo-inference/src/main/java/org/eolang/inference/Clues.java:65`) would
write `links.xml` with one row per resolved reference. A name that resolves
to nothing gets no row and no complaint, in keeping with the note's last
rule: it is better to miss something than to invent it. Resolution stays
primitive — look outwards from the reference for a formation that binds the
name, and take the root object when the name is Φ-rooted; nothing else.

maxonfjvipon:
Rule two of `inference.md`: a reference says that one type is a copy of
another. This is the row that makes the other two tables touch — until now
`needs.xml` asked things of one locator while `provides.xml` answered for a
different one, and nothing said they were the same object.

`Links` writes one row per resolved reference, so the note's program comes
out as it does on paper:

```xml
<links>
  <type id="Φ.app.φ" copy="Φ.app.inc"/>
  <type id="Φ.app.φ.α0" copy="Φ.app.t"/>
  <type id="Φ.app.inc.φ.ρ.ρ" copy="Φ.app.inc.x"/>
</links>
```

Follow the third row and the void `x` is the argument `t`; what `x` is asked
for is then what `t` must have. "A copy of" still means "the same as", and
the rows are records rather than a rewrite of the XMIR on purpose — that is
the one place the checker gets smarter later, when a copy starts receiving
types of its own.

This is the first clue that has to resolve a name, so it comes with one new
object. `Scope` turns `ξ.t` or `Φ.number` into the locator it means: a
ξ-name is looked for outwards, the nearest enclosing formation that binds it
winning, and a Φ-name is the locator of a root object. Nothing else is
consulted — an attribute reachable only through `φ` or through a package
member is not chased, and such a name simply stays unresolved. `Xmirs` grew
`references()` and `locators()`; the latter is the index of what the world
contains, which the checking loop will want as well.

On the runtime the goal resolves **20,825 of 21,555** references, and the
730 it does not are *all* `ξ.ρ` — the object one step out, which no
formation binds as an attribute, so no amount of looking for a name will
find it. Linking it needs the notion of "the object I am inside of", which
the loop will have anyway. A name that resolves to nothing gets no row and
no complaint: a missing row leaves a later check undecided, a wrong row
would make it decide wrongly.

The three tables of the design note now exist: 2,468 types in
`provides.xml`, 9,950 needs in `needs.xml`, 20,825 links in `links.xml`.
What remains is rule four and the loop that drains it.

Based off `bcf9f74497` rather than the current tip, because
`0b2d7df6f5` does not compile (#6485).

Closes: #6490


maxonfjvipon:
Rebuilt on fresh master, and the links coverage is packs now instead of `LinksTest`, following #6516.

The note's own example gained a `links:` block, so one pack shows all three tables of the design note at once:

```yaml
links:
  - "/links/type[@id='Φ.app.φ' and @copy='Φ.app.inc']"
  - "/links/type[@id='Φ.app.φ.α0' and @copy='Φ.app.t']"
  - "/links/type[@id='Φ.app.inc.φ.ρ.ρ' and @copy='Φ.app.inc.x']"
```

Three packs are new: a reference to a root object, a name nobody binds (no row, no complaint), and one that records something the hand-written XMIR had wrong. A name bound in the object *around* this one never arrives as a reference at all — the parser turns `held > @` into a dispatch `.held` on `ξ.ρ`. My deleted `LinksTest` asserted a shape the parser does not emit, which is exactly the drift going through the real parser prevents. That pack now states the finding, including why the parent reference is the one thing this clue cannot resolve.

`LinksTest` is gone; `ScopeTest` replaces it for the resolution rules themselves, since those take locators rather than EO source.

19 packs green, 11 tests in `eo-inference`, clean-tree `verify -Pqulice` green across all seven modules, every count gate matching.

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (339 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+688](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6494, opened 2026-08-09
#6442: reject a trailing separator in a void type list

### LINKED ISSUE #6442: `Parser incorrectly allows trailing space in void type list like /{string }`
## What happens

The parser accepts a trailing separator in a void type list. For example, the void declaration ? > missing /{string } is parsed successfully and produces args="string" even though the list ends with an empty type entry.

## Reproduction

Parse:

[] > fopen /file
  ? > missing /{string }

Current master emits an object without parse errors.

## Root cause

LnVoid.members() advances past the final type and ends its loop when it encounters the trailing space, without validating that the separator is followed by another member.

## Expected behavior

Type-list separators must occur only between type names. A trailing space should produce a syntax error, consistently with the documented single-space-separated grammar.

kdelay:
Closes #6442.

`LnVoid.members()` walks the text inside `/{…}` by scanning to the next space, promoting the member it just read, and then setting `idx = end + 1` to step over the separator. When the separator is the last character, that step lands `idx` exactly on `inside.length()`, the `while` guard fails, and the loop exits without ever noticing that the separator had no member after it. So `? > missing /{string }` parsed clean and emitted `args="string"`, even though the list ends on an empty entry. A double space (`/{a  b}`) was already rejected, because there the next iteration starts on a space and `end == idx` fires the existing check; only a trailing separator slipped through.

Now the loop rejects the list as soon as stepping over a separator reaches the end of the text, reusing the message the double-space case already uses, since both are the same defect: a separator with no type on one side of it. Nothing else changes — a well-formed list still ends with `idx` one past `inside.length()`, so it never trips the new check.

Added `trailing-space-in-void-args.yaml` to the `eo-typos` pack with the exact input from the issue; it fails on master and passes here. Also added the condition to the error table in `PARSER_SPEC.md`, next to the other `/{…}` rows, which did not mention this message at all.

Local gate: `mvn -pl eo-parser clean test -Dtest=EoSyntaxTest` is 1066 tests / 0 failures with the change and 1064 / 0 on master (the two new ones are the pack above). `mvn -pl eo-parser clean verify -PskipTests -Pqulice` passes, and a full reactor `mvn clean install -DskipTests -PskipITs` is green, so the stricter check does not reject any `.eo` source in the repo — no `/{…}` list in the tree ends on a space.

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -1391,6 +1391,7 @@ R-9.9.1. Every error condition in this spec has a single canonical text — **in
 | Type variable outside A–F | `type variable must be one of A-F` |
 | `?` optional marker on an atom return signature | `optional marker ? is allowed only on a void attribute` |
 | `?` inside a `/{…}` argument list | `? is not allowed inside a /{…} argument list` |
+| Empty member in a `/{…}` argument list: a double space, or a separator with no type after it (e.g. `/{string }`) | `types in a /{…} list must be separated by exactly one space` |
 | Type annotation on a void outside an atom | `a void type annotation is allowed only inside an atom` |
 | Two type annotations on one void | `a void attribute may carry at most one type annotation` |
 | Unexpected odd character after a name suffix | `unexpected content after name suffix` |
--- a/eo-parser/src/main/java/org/eolang/parser/LnVoid.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnVoid.java
@@ -198,8 +198,8 @@ final class LnVoid implements Line {
     /**
      * Split a brace argument list on single spaces, promoting each type
      * atom (variable verbatim, forma {@code Q.} promoted to {@code Φ.})
-     * and rejecting empty entries, double spaces, and the {@code ?}
-     * optional marker.
+     * and rejecting empty entries, double spaces, a trailing space with
+     * no member after it, and the {@code ?} optional marker.
      * @param inside The text inside the braces
      * @param span The source span (for errors)
      * @return The space-separated promoted arguments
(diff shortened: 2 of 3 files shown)


kdelay:
The six red jobs here are not caused by this diff: the base does not compile. `Recovery.skip` calls `this.rewound(idx, failed)`, but `failed` is a parameter of `Recovery.after(int)` only, so `eo-parser` fails with:

```
eo-parser/src/main/java/org/eolang/parser/Recovery.java:[76,37] cannot find symbol
  symbol:   variable failed
```

I reproduced it on `master` (0b2d7df6f5) with this branch not involved at all: `mvn -pl eo-parser clean compile` gives the identical error. That is #6485, and #6492 is open for it. I will rebase this PR once that lands, so the checks re-run on a compiling base.

For the record, the local gate on this branch against the previous base (8cda5cb086) was green: `mvn -pl eo-parser clean test -Dtest=EoSyntaxTest` gave 1066 tests / 0 failures (1064 on that base, the two extra being the new `eo-typos` fixtures), and `mvn clean verify -PskipTests -Pqulice` succeeded.

yegor256:
@kdelay master branch is clean now. please, push your branch and CI will re-run.

kdelay:
Rebased onto `d3467f7b19` and pushed (`410d01a8c0`). Local gate on the new base: `mvn -pl eo-parser clean test -Dtest=EoSyntaxTest` -> 1072 tests, 0 failures, and `mvn -pl eo-parser clean verify -PskipTests -Pqulice` -> BUILD SUCCESS. CI is re-running.

0crat:
@kdelay Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+28](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6498, opened 2026-08-09
refactor(#6487): remove unreachable commit hashes map

### LINKED ISSUE #6487: `CommitHashesMap` and its `Fake` variant are dead code, unreachable from the real hash-lookup path
`CommitHashesMap` (`eo-maven-plugin/src/main/java/org/eolang/maven/CommitHashesMap.java`) builds a tag→hash map via `fromTable` (line 84) and exposes a nested `Fake` variant (line 107), but nothing in the plugin ever constructs a `CommitHashesMap` or `CommitHashesMap.Fake` outside of `CommitHashesMapTest.java`. A repo-wide search for `CommitHashesMap` turns up only that test, the class itself, and `CommitHashesText.java`, which reuses just the static `FAKES` string constant as an offline fallback value — never `CommitHashesMap`'s map-building logic.

The commit hash actually used at runtime is wired at `MjSafe.java:405` (`protected CommitHash hash = new ChCached(new ChNarrow(new ChRemote(this.tag)));`), and `ChRemote.value()` (`ChRemote.java:40-58`) resolves the SHA by delegating to `ChText`, a separate, independent line-parser of the same `tags.txt` format. So `CommitHashesMap`'s parsing and its `Fake` class only ever execute inside their own unit test — the class has been orphaned since `ChRemote` was switched over to `ChText`, and any future change to `fromTable` can only affect that one test, never real behavior.

Suggested fix: delete `CommitHashesMap.java` and `CommitHashesMapTest.java`, moving the `FAKES` constant (used only by `CommitHashesText` as its offline fallback) directly into `CommitHashesText.java`.

Titanism0v0:
Closes #6487

## What

Removes the unreachable `CommitHashesMap` implementation and its dedicated test, and moves the offline fallback table into the class that actually uses it.

- `CommitHashesMap.java` (production) — deleted. It parsed `tags.txt` into a `Map<String, CommitHash>` but no production caller ever constructs it; the only remaining references were its own test and the fallback constant borrowed by `CommitHashesText`.
- `CommitHashesMapTest.java` — deleted. It only proved the unreachable implementation was internally consistent.
- `CommitHashesText.java` — the 18-line fallback table (previously `CommitHashesMap.FAKES`) is now a private constant `FALLBACK` in the class that consumes it on DNS failure. Content is byte-for-byte identical (verified by line-by-line diff); `ChRemote`/`ChText`/`MjSafe` are untouched.

## Behavior unchanged

The real runtime chain `MjSafe → ChCached → ChNarrow → ChRemote → CommitHashesText → ChText` is not modified. Full-SHA pass-through, narrowing, retry, and unknown-tag exception semantics all remain as before. `ChTextTest` (28 tests: exact match, whitespace tolerance, unknown tag, 40-char SHA pass-through, retries) passes.

## History note

`ChRemote` switched to `ChText` in `88170cbab` (2022-11-02); `CommitHashesMap` was added later in `0305a48e` (2023-07-15, #1602) and never entered the observable production query path — only its test and the borrowed fallback string remained.

## Not handled

Generic offline mode, HTTP timeouts, and server-side errors are out of scope (only `UnknownHostException` triggers the fallback, as before). No new public API, no resource loading, no caching strategy changes.

Verified locally: `ChTextTest` 28/28; `eo-maven-plugin` full suite 375/376 (sole failure `MjPrintTest` is a pre-existing cross-module resource issue unrelated to this change); `git diff --check` clean. Qulice to be run by CI.


yegor256:
@Titanism0v0 conflicts here

0crat:
@titanism0v0 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (210 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+136](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6510, opened 2026-08-09
perf(parser): build locators top-down in set-locators (~2.5x)

### LINKED ISSUE #6509: set-locators.xsl re-derives every ancestor per object; ~2.5x available by building locators top-down
`set-locators.xsl` is the 3rd most expensive stylesheet in the build — **12786 ms, 8.09%** of the 158017 ms total in the README benchmark table — and single documents take over 10s in the wild. I profiled it and the cost is algorithmic, not incidental.

## Why it is slow

`eo:locator()` reaches the program from each object separately, walking `parent::o` all the way up. So a locator costs O(depth), and the whole pass costs O(objects x depth) instead of O(objects).

Measured on the **170 eo-runtime documents this stylesheet actually receives** (dumped by running the real `Canonical` pipeline up to, but excluding, `set-locators`):

| | |
|---|---|
| `<o>` elements | 36,675 |
| `eo:locator()` invocations | **240,586** (6.6x amplification) |
| mean nesting depth | 6.56 |
| max nesting depth | 52 (`tuple/reducedi.eo`) |

Four separate costs stack up, all per-object rather than per-document:

1. **The ancestor walk** — 240,586 calls to produce 36,675 attributes.
2. **A result tree fragment per recursion level.** `<xsl:variable name="ret">` has no `as=`, so each of those 240,586 calls materialises a temporary tree node just to `xsl:value-of` it back into a string.
3. **The package prefix, looked up once per object.** `root($o)/object/metas/meta[head='package']` is evaluated **twice** at the base case (once in the `xsl:if` test, once in `xsl:value-of`) — 73,350 scans of `metas` for a value that is constant per document.
4. `count($o/parent::o[starts-with(@base, '.')])` builds a node sequence and counts it to obtain what is only ever 0 or 1.

## On the "introduce keys" idea

I tried this first and it is the wrong tool here. `xsl:key` indexes nodes so they can be *looked up by value*, but nothing in this stylesheet is a value lookup — the bottleneck is ancestor traversal and a repeated constant. A key over `meta[head='package']` would remove cost (3), yet a plain global variable or a one-time tunnelled value removes it more cheaply, since there is exactly one such value per document. Keys cannot touch (1), which is the dominant cost.

Memoisation would fix (1) while keeping the recursion (`cache="yes"` on `xsl:function`), but that needs XSLT 3.0 and this stylesheet is `version="2.0"`.

## What actually works: build locators top-down

Locators are prefix-computable — a child's locator is its parent's locator plus one segment. So make one top-down pass, handing each `<o>` its parent's finished locator through a **tunnel parameter** and appending a single segment. That makes each locator O(1), and the package prefix is computed once on the `object` element and tunnelled down with it.

Two constraints worth recording, because both bite:

- **Keep the accumulator off function arguments.** The comment in the file documents that Saxon 13.0 binds multi-argument function arguments to the wrong values when one compiled stylesheet transforms documents in parallel threads, which is exactly what the `parse` goal does. A tunnel *template* parameter is a different mechanism, and `eo:segment` stays single-argument.
- **Do not read the global context item.** My first attempts hoisted the prefix into a global `<xsl:variable select="/object/...">`. That works under the JAXP path production uses, but fails outright with `XPDY0002 the context item is absent` when the stylesheet is driven via `xsl:apply-templates`. Computing the prefix on the `object` element instead needs no global context item and works under both.

## Results

PR: #6510. Saxon-HE 13.0, behind the same JAXP `DOMSource` -> `DOMResult` path `jcabi-xml` uses in production, single thread, ms per pass over all 170 documents:

| variant | rounds (ms) |
|---|---|
| as-is today | 1662, 1708, 1684 |
| drop the RTF + hoist the prefix, **ancestor walk retained** | 1128, 1077, 1163 |
| **top-down tunnel (the PR)** | **683, 663, 650** |

So roughly a third of the cost is the RTF plus the repeated prefix lookup, and the rest is the walk itself. Net **2.4-2.6x**, and **4.4x** on the 4-thread parallel path.

The win grows with nesting, as expected. Holding the object count constant at 2400 and varying only chain depth:

| nesting depth | 10 | 25 | 50 | 100 | 200 |
|---|---|---|---|---|---|
| speedup | 2.5x | 2.7x | 3.5x | 5.6x | 9.8x |

Output is **byte-identical on all 170 documents** (36,675 `@loc` attributes), and `eo-parser` (1684 tests) plus `MjTranspileTest`/`MjParseTest`/`MjInferenceTest` (67 tests) pass.

## Bonus: the current walk has a depth ceiling

Deep nesting does not merely get slow, it **fails**. Past ~215 levels of object nesting the recursion trips Saxon's nested-call limit:

```
SXLM0001  Too many nested function calls. May be due to infinite recursion
```

eo-runtime's deepest file is 52 levels, so there is headroom today, but 215 is not a comfortable margin for generated or machine-written EO. The top-down version reaches ~460 before hitting the same limit — better, though a tree walk cannot remove the ceiling entirely.

## A larger, orthogonal lever

While isolating the above I measured something that affects **every** stylesheet, not just this one. Holding the stylesheet and output target constant and varying *only* the source tree model:

| source tree | this stylesheet | current stylesheet |
|---|---|---|
| Saxon TinyTree | 201-208 ms | 575-653 ms |
| W3C DOM (what `jcabi-xml` hands Saxon) | 439-550 ms | 1557-1607 ms |

A **~2.5x penalty purely from the DOM tree model**, independent of the XSL. `to-java.xsl` and `classes.xsl` alone are 63% of the 158s total, so if the pipeline could keep documents in Saxon's native tree between shifts rather than round-tripping through DOM, the payoff would dwarf anything achievable inside a single stylesheet. That belongs in its own issue; I am noting it here only because it is the same profiling run.

## Reproducing

Documents were produced by running `Canonical`'s pipeline over `eo-runtime/src/main/eo` and stopping before `set-locators`, so they are exactly what the stylesheet sees in a real build. Timings are min/median over alternating rounds on a 4-CPU container; absolute values are noisy there but the ratios were stable across every round.

yegor256:
Closes #6509.

`set-locators.xsl` is the 3rd most expensive stylesheet in the build (**12786 ms, 8.09%** of the README benchmark total). The cost is algorithmic: `eo:locator()` reached the program from every object separately by walking `parent::o` upwards, so each locator cost O(depth) and the pass cost O(objects x depth).

Over the 170 eo-runtime documents this stylesheet actually receives, **36,675 objects triggered 240,586 `eo:locator()` calls**. Each of those calls also materialised a result tree fragment (`<xsl:variable name="ret">` has no `as=`), and the base case re-scanned `metas/meta[head='package']` twice per object — 73,350 scans for a value that is constant per document.

## Change

Locators are prefix-computable, so build them in one top-down pass: every `<o>` receives its parent's finished locator through a tunnel parameter and appends a single segment, making each locator O(1). The `Φ.package` prefix is computed once on the `object` element and tunnelled down with it. `count($o/parent::o[starts-with(@base, '.')])` becomes a plain `if`.

Two constraints the implementation deliberately respects:

- **The accumulator rides on template parameters, not function arguments.** The comment already in this file records that Saxon 13.0 binds multi-argument function arguments to the wrong values when one compiled stylesheet transforms documents in parallel threads — which is what the `parse` goal does. A tunnel template parameter is a different mechanism, and `eo:segment` stays single-argument.
- **Nothing reads the global context item.** Hoisting the prefix into a global `<xsl:variable select="/object/...">` is faster to write and works under the JAXP path production uses, but fails with `XPDY0002 the context item is absent` when the stylesheet is driven through `xsl:apply-templates`. Computing it on the `object` element works under both, and measured the same.

## Results

Saxon-HE 13.0 behind the same JAXP `DOMSource` -> `DOMResult` path `jcabi-xml` uses, single thread, ms per pass over all 170 documents:

| variant | rounds (ms) |
|---|---|
| before | 1662, 1708, 1684 |
| drop the RTF + hoist the prefix, ancestor walk retained | 1128, 1077, 1163 |
| **after (this PR)** | **683, 663, 650** |

**2.4-2.6x** single-threaded, **4.4x** on the 4-thread parallel path. About a third of the saving is the RTF plus the repeated prefix lookup; the rest is removing the walk.

The win scales with nesting — holding object count constant at 2400 and varying only chain depth:

| nesting depth | 10 | 25 | 50 | 100 | 200 |
|---|---|---|---|---|---|
| speedup | 2.5x | 2.7x | 3.5x | 5.6x | 9.8x |

## Correctness

- **Byte-identical output on all 170 eo-runtime documents** (36,675 `@loc` attributes), diffed against the current stylesheet. Inputs were generated by running the real `Canonical` pipeline and stopping before `set-locators`, so they are exactly what it sees in a build — packaged and unpackaged files, nesting up to 52 levels.
- `eo-parser`: 1684 tests pass (including `StrictXmirTest`, which validates `@loc` against the `locator` type in `XMIR.xsd`).
- `eo-maven-plugin`: `MjTranspileTest` (52, covering the three `transpile-packs/locators` packs), `MjParseTest` (13), `MjInferenceTest` (2) pass.
- `xcop` clean.

## Side effect: more depth headroom

Deep nesting previously *failed* rather than merely slowed — past ~215 levels the recursion tripped `SXLM0001 Too many nested function calls`. The top-down version reaches ~460 before hitting the same limit. eo-runtime's deepest file is 52 levels, so this is headroom rather than a live bug fix, and a tree walk cannot remove the ceiling entirely.

#6509 also records a larger, orthogonal finding from the same profiling run: the W3C DOM source tree costs ~2.5x versus Saxon's native TinyTree, independent of the stylesheet, which would affect every XSL in the pipeline. Out of scope here.

https://claude.ai/code/session_014DcQkPHQetxHrf2vjw7zkX

---
_Generated by [Claude Code](https://claude.ai/code/session_014DcQkPHQetxHrf2vjw7zkX)_

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/set-locators.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/set-locators.xsl
@@ -8,69 +8,68 @@
   Here we go through all objects and add @loc attributes
   to all of them. The value of the attribute is a unique locator
   of the object.
-  The locator function takes the object only and reaches the
-  program through it, instead of taking the program as a second
-  argument. Saxon 13.0 binds the arguments of a multi-argument
-  function to the wrong values, once in a while, when one compiled
-  stylesheet transforms many documents in parallel threads (which is
-  what the "parse" goal does).
+  The locator of an object is the locator of its parent plus one more
+  segment, so we build them top-down: every "o" receives the locator
+  already built for its parent through the "eo:parent" tunnel parameter
+  and only appends its own segment. Reaching the program from every
+  object instead re-derives all of its ancestors, which costs O(depth)
+  per object rather than O(1), and hits the limit on nested calls that
+  Saxon imposes once objects nest a few hundred levels deep (#6509).
+  The "Φ.package" prefix is likewise computed once, on the "object"
+  element, and tunnelled down, instead of being looked up again for
+  every object in the document.
+  Both values travel as template parameters, not as extra arguments of
+  a locator function. Saxon 13.0 binds the arguments of a
+  multi-argument function to the wrong values, once in a while, when
+  one compiled stylesheet transforms many documents in parallel threads
+  (which is what the "parse" goal does), so eo:segment below stays
+  single-argument. Nothing here reads the global context item either,
+  since that item is absent when the stylesheet is driven through
+  xsl:apply-templates rather than through a whole-document transform.
   -->
   <xsl:output encoding="UTF-8" method="xml"/>
   <xsl:import href="/org/eolang/parser/_specials.xsl"/>
-  <xsl:function name="eo:locator" as="xs:string">
-    <xsl:param name="o" as="node()"/>
-    <xsl:if test="name($o) != 'o'">
-      <xsl:message terminate="yes">
-        <xsl:text>Only 'o' XML elements are accepted here</xsl:text>
-      </xsl:message>
-    </xsl:if>
-    <xsl:variable name="ret">
-      <xsl:choose>
-        <xsl:when test="$o/parent::o">
-          <xsl:value-of select="eo:locator($o/parent::o)"/>
-        </xsl:when>
-        <xsl:otherwise>
-          <xsl:value-of select="$eo:program"/>
-          <xsl:if test="root($o)/object/metas/meta[head='package']">
-            <xsl:text>.</xsl:text>
-            <xsl:value-of select="root($o)/object/metas/meta[head='package']/part[1]"/>
-          </xsl:if>
-        </xsl:otherwise>
-      </xsl:choose>
-      <xsl:text>.</xsl:text>
-      <xsl:choose>
-        <xsl:when test="$o/@name">
-          <xsl:choose>
-            <xsl:when test="$o/@name = '@'">
-              <xsl:value-of select="$eo:phi"/>
-            </xsl:when>
-            <xsl:otherwise>
-              <xsl:value-of select="$o/@name"/>
-            </xsl:otherwise>
-          </xsl:choose>
-        </xsl:when>
-        <xsl:otherwise>
-          <xsl:choose>
-            <xsl:when test="$o/@as">
-              <xsl:value-of select="$o/@as"/>
-            </xsl:when>
-            <xsl:when test="starts-with($o/parent::o/@base, '.') and not($o/preceding-sibling::o)">
-              <xsl:value-of select="$eo:rho"/>
-            </xsl:when>
-            <xsl:otherwise>
-              <xsl:value-of select="$eo:alpha"/>
-              <xsl:value-of select="count($o/preceding-sibling::o) - count($o/parent::o[starts-with(@base, '.')])"/>
-            </xsl:otherwise>
-          </xsl:choose>
-        </xsl:otherwise>
-      </xsl:choose>
-    </xsl:variable>
-    <xsl:value-of select="$ret"/>
+  <!--
+  The trailing segment of the locator of the given object, without the
+  dot that joins it to the locator of its parent.
+  -->
+  <xsl:function name="eo:segment" as="xs:string">
+    <xsl:param name="o" as="element(o)"/>
+    <xsl:variable name="dotted" as="xs:boolean" select="starts-with($o/parent::o/@base, '.')"/>
+    <xsl:choose>
+      <xsl:when test="$o/@name = '@'">
+        <xsl:sequence select="$eo:phi"/>
+      </xsl:when>
+      <xsl:when test="$o/@name">
+        <xsl:sequence select="string($o/@name)"/>
+      </xsl:when>
+      <xsl:when test="$o/@as">
+        <xsl:sequence select="string($o/@as)"/>
+      </xsl:when>
+      <xsl:when test="$dotted and not($o/preceding-sibling::o)">
+        <xsl:sequence select="$eo:rho"/>
+      </xsl:when>
+      <xsl:otherwise>
+        <xsl:sequence select="concat($eo:alpha, count($o/preceding-sibling::o) - (if ($dotted) then 1 else 0))"/>
+      </xsl:otherwise>
+    </xsl:choose>
   </xsl:function>
-  <xsl:template match="o">
+  <xsl:template match="object">
+    <xsl:variable name="pkg" select="metas/meta[head='package'][1]/part[1]"/>
     <xsl:copy>
-      <xsl:attribute name="loc" select="eo:locator(.)"/>
-      <xsl:apply-templates select="node()|@* except @loc"/>
+      <xsl:apply-templates select="node()|@*">
+        <xsl:with-param name="eo:parent" as="xs:string" tunnel="yes" select="if ($pkg) then concat($eo:program, '.', $pkg) else $eo:program"/>
+      </xsl:apply-templates>
+    </xsl:copy>
+  </xsl:template>
+  <xsl:template match="o">
+    <xsl:param name="eo:parent" as="xs:string" select="$eo:program" tunnel="yes"/>
+    <xsl:variable name="loc" as="xs:string" select="concat($eo:parent, '.', eo:segment(.))"/>
+    <xsl:copy>
+      <xsl:attribute name="loc" select="$loc"/>
+      <xsl:apply-templates select="node()|@* except @loc">
+        <xsl:with-param name="eo:parent" as="xs:string" tunnel="yes" select="$loc"/>
+      </xsl:apply-templates>
     </xsl:copy>
   </xsl:template>
   <xsl:template match="node()|@*">


yegor256:
CI status: the two red checks both reproduce on `master` and neither can be caused by this diff, which touches one `.xsl` file and no Java. Everything that exercises the change is green, including on Windows. Diagnosis below; I have deliberately **not** fixed either here, since both are unrelated to a locator perf change — happy to send them as separate PRs.

**1. `counts (suppressions, *.java, @SuppressWarnings|@checkstyle, 364)` — stale ratchet on master**

The workflow expects 364, the tree has 362:

```
$ git ls-files -z '*.java' | xargs -0 grep -Eoh '@SuppressWarnings|@checkstyle' | wc -l
362   # on this branch
362   # on origin/master (98647bd)
```

It went 364 -> 362 at 9ff054b (`build: upgrade qulice-maven-plugin to 0.30.6`), which intentionally dropped two now-dead annotations — its own message says "two now-dead `@SuppressWarnings`" — but `.github/workflows/counts.yml` was not lowered to match:

```
-    @SuppressWarnings("PMD.UnnecessaryLocalRule")
-    @SuppressWarnings("PMD.UnnecessaryLocalRule")
```

So `master` has been red on this check since 8 Aug, and every open PR inherits it. Fix is `count: 364` -> `count: 362` in `.github/workflows/counts.yml`.

**2. `mvn (windows-2022, 26)` — Windows-only line-separator bug, pre-existing**

The only failing test is `CommitHashesMapTest.containsValidHash[1]` and `[2]`:

```
java.lang.NullPointerException: Cannot invoke "org.eolang.maven.CommitHash.value()"
  because the return value of "org.eolang.maven.CommitHashesMap$Fake.get(Object)" is null
    at org.eolang.maven.CommitHashesMapTest.containsValidHash(CommitHashesMapTest.java:36)
```

`CommitHashesMap.FAKES` joins its rows with `System.lineSeparator()`, but `fromTable` splits on the literal `"\\n"`. On Windows that leaves a trailing `\r` on every row, and since the key is taken as everything after the first whitespace run, every map key ends up carrying it:

```
row after split : 'e0b783692ef749bb184244acb2401f551388a328 0.26.0\r'
derived map key : '0.26.0\r'
test looks up   : '0.26.0'      -> no match, get() returns null
```

No XSL is involved. It passes here on Linux (`Tests run: 4, Failures: 0, Errors: 0`), which is why the Linux leg was green before fail-fast cancelled it. Fix is to split on `"\\R"` (or join `FAKES` with `"\n"`).

**What is green, on Windows included:** `MjTranspileTest` 52 (covering all three `transpile-packs/locators` packs), `MjParseTest` 13, `MjInferenceTest` 2, `InferringTest` 7, `eo-parser` 1684, `eo-printer` 334, plus `xcop`, `xslint`, `typos`, `copyrights`, SonarCloud and Codacy.

---

yegor256:
On the "⚠️ Benchmark Comparison Unavailable" notice above — that is not this PR either, and since this *is* a perf PR I want to be explicit about it rather than leave it looking like the perf check failed on a perf change.

The repo's only JMH benchmark, `benchmarks.XslBench`, points at two stylesheets that do not exist (`parse/resolve-before-star.xsl` — the real file is `resolve-before-star**s**.xsl` — and `shake/build-fqns.xsl`, which lives at `parse/build-fqns.xsl`). `TrClasspath` resolves eagerly, so it dies in `<clinit>` and the run yields no rows:

```
$ mvn jmh:benchmark -pl eo-parser -Djmh.rf=json -Djmh.rff=benchmark.json
Benchmark  Mode  Cnt  Score   Error  Units     <- no rows
[INFO] BUILD SUCCESS                            <- exit 0 anyway
$ cat eo-parser/benchmark.json
[
]
```

So the comparison action has nothing to diff on either side, on every PR — not just this one. Filed separately as #6513, with the two-line fix and a note that `XslBench` covers none of `to-java.xsl` / `classes.xsl` / `set-locators.xsl`, i.e. 71% of the build's XSL time.

The perf numbers in the description were measured directly against the 170 real eo-runtime documents rather than through this benchmark, so they stand independently of it.

---

0crat:
@yegor256 It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`claude/set-locators-xsl-perf-yqezjw`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`6509`". Your running score is [+2815](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (113 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2831](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6511, opened 2026-08-09
perf(printer): index `merge-monikers` lookups with `xsl:key`

yegor256:
Fixes part of #6512.

`merge-monikers.xsl` answers the same question from every node it visits —
"which references in this formation name this binding?" — and answered it with
a `$owner//o[...]` descendant scan. Since those lookups are reached from
template *patterns*, which Saxon evaluates against every `o` node, a document
of N nodes walked its own subtree N times. Print time therefore grew
quadratically, and in the outermost formation `$owner//o` is the whole
document.

## What changed

Four lookups are now backed by `xsl:key` indexes:

| key | replaces |
| --- | --- |
| `moniker-ref` | the `$owner//o[...]` scan in `eo:moniker-refs` |
| `applied-ref` | the `$owner//o[...]` scan in `eo:applied-refs` |
| `moniker-binding` | the `$owner/o[...]` sibling scan in `eo:hosted-binding` and `eo:applied-handle` |
| `moniker-name` | the `$ref/ancestor::o[eo:abstract(.)]/o[...]` climb in `eo:kept-const-ref` and `eo:kept-local-ref` |

The first three are keyed on `generate-id()` of the owning formation plus the
name, so a lookup keeps the "same formation, no intervening formation" scoping
the scans enforced by hand. Keys hand their nodes back in document order, so
"the first hosting reference" still means what it meant before — which is what
keeps #5739's first-reference rule and #5890's ordering intact.

`concat($eo:xi, '.')` is hoisted into a global `$eo:xi-dot`. Saxon does not
fold it, because `$eo:xi` is a global variable rather than a literal, so
leaving it inline rebuilt the same two-character string at each of the seven
call sites; string concatenation was the largest single group of samples in a
JFR profile of the old sheet.

No behaviour changes. The sheet's semantics, its comments and its function
signatures are untouched — only how the four lookups find their nodes.

## Results

Saxon-HE 13.0 (the pinned version), best of several runs, on synthetic XMIR of
the shape the sheet sees:

| `<o>` nodes | before | after | speedup |
| ---: | ---: | ---: | ---: |
| 502 | 782 ms | 112 ms | 7× |
| 1002 | 2421 ms | 67 ms | 36× |
| 2002 | 10170 ms | 90 ms | 113× |
| 4002 | 41427 ms | 199 ms | 208× |
| 16002 | (over 10 min) | 558 ms | — |

On real `eo-runtime` sources, taken from `target/eo/1-parse` and pushed through
the five print sheets that precede this one, so the input is exactly what
`merge-monikers` receives:

| file | `<o>` nodes | before | after | speedup |
| --- | ---: | ---: | ---: | ---: |
| `string/printf.eo` | 1151 | 440 ms | 95 ms | 4.6× |
| `path.eo` | 1021 | 205 ms | 81 ms | 2.5× |
| `file.eo` | 810 | 127 ms | 71 ms | 1.8× |
| `map.eo` | 741 | 108 ms | 64 ms | 1.7× |
| `tuple.eo` | 719 | 97 ms | 55 ms | 1.8× |
| `directory.eo` | 684 | 111 ms | 62 ms | 1.8× |

Runtime sources gain less because their formations are small; the quadratic
term is driven by the size of the largest formation, not the file. To be
explicit about the limit of that: a full `mvn -pl eo-runtime process-sources`
takes 250 s before and 252 s after, which is to say no measurable difference —
`eo-runtime` is not where this hurts, since the goal's time goes to linting,
inference and `to-java`. The win is on files whose cactus bindings concentrate
in one scope, where the sheet currently spends tens of seconds. Holding a
synthetic document at ~3000 nodes and only redistributing the same 600
bindings shows the effect on its own: 1046 ms across 100 formations against
22224 ms in a single one, and a flat ~160 ms either way after this change.

## Verification

- Output is byte-identical to `master` on every input tried: all six real
  `eo-runtime` files above, the three `print-packs` fixtures, and eleven
  synthetic documents covering bare references, single- and multi-segment
  dispatch chains, const handles, named handles, applied formation handles,
  recursive handles, shared handles and references from nested formations.
- All 334 `eo-printer` tests pass, including the 293 `XmirTest` pack tests.
- Given #6091 (Saxon-HE 13.0 miscomputing `xsl:function` arguments when one
  compiled stylesheet is shared across threads), the keys were checked under
  the same conditions the printer uses: one shared `Templates`, 24 threads,
  4320 transforms — zero mismatches. Keys are built per source document by the
  per-transformation `KeyManager`, so they add no shared mutable state, and
  this change introduces no new multi-argument `xsl:function`.

Remaining optimization opportunities — memoizing the functions that are still
evaluated once in a pattern and again in a body, and the same scan shape in
`inline-cactoos.xsl` and `restore-local-names.xsl` — are listed in #6512.

https://claude.ai/code/session_01SgyCtxMVnZq4qnDXedksSu

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl
@@ -37,6 +37,45 @@
   <xsl:import href="/org/eolang/parser/_funcs.xsl"/>
   <xsl:output encoding="UTF-8" method="xml"/>
   <!--
+  The `ξ.` prefix every local reference carries, built once instead of at each
+  of the many `starts-with`/`substring-after` calls below. Saxon does not fold
+  `concat($eo:xi, '.')` into a constant, since `$eo:xi` is a global variable
+  rather than a literal, so leaving it inline rebuilds the same two-character
+  string for every reference the sheet looks at.
+  -->
+  <xsl:variable name="eo:xi-dot" select="concat($eo:xi, '.')"/>
+  <!--
+  Every reference that resolves to an attribute name (`eo:resolved-ref`),
+  indexed by the formation that owns it and the name it resolves to. The
+  lookups below all ask the same question — "which references in this
+  formation name this binding?" — which as a `$owner//o[...]` scan costs a walk
+  of the whole subtree per binding, and so a walk of the document per node
+  once every node asks it. The index answers it in constant time, and, being a
+  key, still hands the references back in document order, which is what makes
+  "the first hosting reference" mean the same thing as before.
+  -->
+  <xsl:key name="moniker-ref" match="o[eo:resolved-ref(.) != '']" use="concat(generate-id(ancestor::o[eo:abstract(.)][1]), ' ', eo:resolved-ref(.))"/>
+  <!--
+  Every applied reference (`eo:applied-refs`) — a bare `ξ.<name>` carrying
+  arguments and no name of its own — indexed by owning formation and receiver
+  name, the applied twin of `moniker-ref` above.
+  -->
+  <xsl:key name="applied-ref" match="o[exists(@base) and starts-with(@base, $eo:xi-dot) and exists(o) and not(exists(@name))]" use="concat(generate-id(ancestor::o[eo:abstract(.)][1]), ' ', substring-after(@base, $eo:xi-dot))"/>
+  <!--
+  Every eligible binding (`eo:moniker-binding`) indexed by its owning formation
+  and name, so a reference reaches the binding it resolves to without scanning
+  the formation's attributes.
+  -->
+  <xsl:key name="moniker-binding" match="o[eo:moniker-binding(.)]" use="concat(generate-id(..), ' ', @name)"/>
+  <!--
+  The same bindings indexed by name alone, for the two "kept reference"
+  lookups, which climb the ancestor formations looking for whichever one
+  declares a segment of the reference's base. Cactus names are auto-generated
+  and effectively unique, so this bucket holds one node and the climb becomes a
+  parentage check on it rather than a scan of every ancestor's attributes.
+  -->
+  <xsl:key name="moniker-name" match="o[eo:moniker-binding(.)]" use="@name"/>
+  <!--
   The single attribute name a hostable `ξ.<name>` reference resolves to, or
   the empty string for anything that is not hostable. Two shapes host a
   binding: a bare reference `ξ.<name>` (no trailing path, no arguments) and a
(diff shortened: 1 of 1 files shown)


yegor256:
CI is red on `mvn (windows-2022, 26)`, and it is failing on the base branch too — not caused by this PR.

Master at `98647bd`, the exact commit this branch sits on, fails the same job with the same two errors and the same totals (547 run, 2 errors, 1 skipped):

```
[ERROR] CommitHashesMapTest.containsValidHash:36 NullPointer Cannot invoke
"org.eolang.maven.CommitHash.value()" because the return value of
"org.eolang.maven.CommitHashesMap$Fake.get(Object)" is null
```

`884d2f7` before it fails identically. Every other platform in the matrix is
unaffected, `eo-printer` builds green on Windows with all its tests, and this
PR touches one XSL file in that module.

The cause looks deterministic rather than flaky: `CommitHashesMap.FAKES` joins
its rows with `System.lineSeparator()`, while `fromTable` splits the table on
`"\\n"`. On Windows the separator is `\r\n`, so every row keeps a trailing
`\r` and each map key becomes `"0.26.0\r"` rather than `"0.26.0"` — hence the
null from `get("0.26.0")`. Either side alone would fix it; the class is also
already up for removal in #6487, which would take the test with it.

I will re-run this job once the base branch is green.

---

yegor256:
Second red check, also from the base branch rather than this PR: `counts (suppressions, *.java, @SuppressWarnings|@checkstyle, 364)`.

That job demands an exact total. Running its own command on this branch:

```
$ git ls-files -z '*.java' | xargs -0 grep -Eoh '@SuppressWarnings|@checkstyle' | wc -l
362
```

Two short of the 364 in `.github/workflows/counts.yml`. This PR changes no
Java at all — `git diff --name-only master...` lists only
`eo-printer/.../merge-monikers.xsl` — so the count here is necessarily
whatever master's is. `counts` duly fails on master too, at `98647bd` (this
PR's base) and `884d2f7`, having last passed at `0b2d7df6`. Two suppressions
were dropped without lowering the expected number; the fix is `count: 362`
in that workflow, on master.

So both red checks on this PR — this one and the Windows `CommitHashesMap`
failure above — are inherited, and neither is something a perf change to one
XSL sheet should be carrying a fix for. The "Benchmark Comparison
Unavailable" notice is a third symptom of the same thing: that action diffs
against a `master` benchmark that the red base build never produced. It
benchmarks `eo-parser` in any case, so it would not have measured this
change either way.

The checks that do gate this PR — `xcop`, `xslint` and `typos` — are still
queued behind the backlog; I will follow up when they land.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2839](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6514, opened 2026-08-09
fix(#6425): reinstate the level a failed line replaced

### LINKED ISSUE #6425: `Eo.dispatch` error rollback fails to undo `Stack.replace`, leading to `ClassCastException` or invalid XMIR

VasilevNStas:
Fixes #6425.

## Problem

A single-character typo in a `.eo` file produces an unhandled `ClassCastException` (or schema-invalid XMIR), because `Eo.dispatch` rollback cannot undo a `Stack.replace`: `silentTruncate` works by stack size, and a same-size replacement leaves the replacement level in place. When the failed line owed a different number of `emit.close()` calls than the level it displaced, the directive stream and the parser model diverge.

## Solution

Give the per-line rollback a real savepoint. `Stack.checkpoint()` records the top level and returns the current depth; `silentTruncate(target)` now reinstates the displaced level after truncation, covering three shapes:

- a push (stack grew beyond `target`),
- a completed replace (depth unchanged, top swapped),
- a replace that threw mid-way (e.g. in `observeVoid`), when the stack is left one level short of `target`.

## Changes

- `eo-parser/src/main/java/org/eolang/parser/Stack.java` — `checkpoint()` records the top level; `silentTruncate(int)` restores it; `topOrNull()` helper; removed the redundant `size()` (same as `depth()`).
- `eo-parser/src/main/java/org/eolang/parser/Eo.java` — both rollback call sites use `stack.checkpoint()`.

## Tests

- `EoSyntaxTest.recoversFromBrokenEscapeThatReplacedItsSibling` — a failed only-phi line replacing a plain sibling (the `ClassCastException` shape).
- `EoSyntaxTest.recoversFromBrokenEscapeReplacingAnOnlyPhi` — a failed plain line replacing an only-phi (the `ms`-lands-on-`<o>` shape).

@yegor256 please take a look.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -341,7 +341,7 @@ final class Eo implements Iterable<Directive> {
         if (Eo.closesTextBlock(span, globals)) {
             stack.popDeeperThan(span.indent());
             final int tstartsen = emit.savepoint();
-            final int frame = stack.size();
+            final int frame = stack.checkpoint();
             try {
                 new LnTextBlock(span).into(stack, globals, emit);
             } catch (final ParseError err) {
@@ -408,7 +408,7 @@ final class Eo implements Iterable<Directive> {
             stack.popDeeperThan(span.indent());
         }
         final int tstartsen = emit.savepoint();
-        final int frame = stack.size();
+        final int frame = stack.checkpoint();
         boolean failed = false;
         try {
             Eo.classify(span).into(stack, globals, emit);
--- a/eo-parser/src/main/java/org/eolang/parser/Stack.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Stack.java
@@ -40,6 +40,13 @@ final class Stack {
     private final List<Level> levels;
 
     /**
+     * The top level captured by the most recent {@link #checkpoint()},
+     * reinstated by {@link #silentTruncate} when a failed line
+     * {@link #replace}d it.
+     */
+    private Level saved;
+
+    /**
      * Close-time check hook, invoked whenever a level is popped or
      * replaced. Injected once at construction so per-call wiring stays
      * out of the API.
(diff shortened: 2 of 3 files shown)


VasilevNStas:
Closing as a duplicate: this issue is already fixed by #6468 (`Stack.snapshot()`/`restore()`), merged into master on 2026-08-09. Their approach reinstates the full level list and is more robust than the top-only restore proposed here. My two integration tests (in this branch) are complementary regression coverage at the `EoSyntax` level and pass against the current master — happy to open them as a separate small test PR if useful.

PR #6515, opened 2026-08-09
test(#6425): add ms-integrity regression for replaced only-phi rollback

VasilevNStas:
Related to #6425 (already fixed by #6468).

## Problem

The merged fix (#6468) added a regression pack for the primary symptom of #6425 — a failed only-phi line replacing a plain sibling, which used to crash with `ClassCastException`. The issue also documents the reverse shape — a failed plain line replacing an only-phi sibling — which used to land the `ms` attribute on a child `<o>` (schema-invalid XMIR). That shape is not covered by any test.

## Solution

Add one eo-syntax pack covering the reverse shape and asserting `ms` stays on `/object`, plus the expected parser error.

## Changes

- `eo-parser/src/test/resources/org/eolang/parser/eo-syntax/replaced-only-phi-rollback-keeps-ms.yaml` — new pack.

## Tests

- The pack fails on pre-fix code with `FAIL: /object[@ms='0']` (verified against `98647bd6e`) and passes on master with the #6468 fix.

### DIFF
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/replaced-only-phi-rollback-keeps-ms.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/replaced-only-phi-rollback-keeps-ms.yaml
@@ -1,0 +1,14 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+sheets: []
+asserts:
+  - /object[@ms='0']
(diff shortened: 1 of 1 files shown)


0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6517, opened 2026-08-09
Check the clues with yaml packs over eo source

### LINKED ISSUE #6516: eo-inference clues tests use hand-written XMIR instead of descriptive YAML packs
The clues of `eo-inference` are tested with XMIR written by hand. A case in
`ProvidesTest`
(`eo-inference/src/test/java/org/eolang/inference/ProvidesTest.java:27`)
looks like this:

```java
"<object><o loc='Φ.kettle' name='kettle'>",
"<o loc='Φ.kettle.steam' name='steam'/>",
"<o loc='Φ.kettle.whistle' name='whistle'/>",
"</o></object>"
```

Two things are wrong with that. It is XMIR nobody parsed, so it can drift
from what the parser really emits and the test would keep passing while the
clue reads real programs differently. And it hides the point: a reader has
to reconstruct which EO program this was, when the interesting sentence is
"a formation with two attributes gives two rows".

The repository already tests this way round elsewhere. `ProbesTest`
(`eo-maven-plugin/src/test/java/org/eolang/maven/ProbesTest.java:32`) reads
a directory of YAML packs, each carrying EO source and the expectations
against it, and the same shape fits inference exactly: EO in, asserts on the
tables out.

```yaml
eo: |
  [] > app
    [] > t
      [] > next
provides:
  - "/provides/type[@id='Φ.app.t']/attr[@name='next']"
needs:
  - "/needs/type[@id='Φ.app.inc.φ.ρ']/attr[@name='foo']"
```

Fifteen lines of that say what the prototype does; the Java equivalent
spends forty and says it less clearly. It also makes each step of the work
visible: a rule added later arrives as a pack a reviewer can read, rather
than as another class of hand-built fragments.

The runner belongs next to the existing packs in `eo-maven-plugin`, since
the parser and the preparation the clues need are both there, and the
tables come out of `Inferring`. Packs should cover the rules — what each
clue writes down — while the machinery that takes no EO source
(`Grouped`, `Cells`, the composition in `Clues`) stays in plain unit tests,
where a failure still points at one class.

maxonfjvipon:
The clues were tested with XMIR written by hand, which hid what each rule
actually does and could drift from what the parser really emits. They are
now tested the way the rest of the repository tests this kind of thing: a
directory of YAML packs, each carrying EO source and the XPaths its tables
must satisfy.

A pack is the design note, executable:

```yaml
# The program of the design note, which is the whole prototype in one place.
eo: |
  [] > app
    inc t > @
    [] > t
      [] > next
    [x] > inc
      x.next.foo > @
xmir:
  - "//o[@base='.foo']/o[@base='.next']/o[@base='ξ.x']"
provides:
  - "/provides/type[@id='Φ.app.t']/attr[@name='next' and @type='Φ.app.t.next']"
  - "/provides/type[@id='Φ.app.t.next' and @complete='true' and not(attr)]"
needs:
  - "/needs/type[@id='Φ.app.inc.φ.ρ.ρ']/attr[@name='next' and @type='Φ.app.inc.φ.ρ']"
  - "/needs/type[@id='Φ.app.inc.φ.ρ']/attr[@name='foo' and @type='Φ.app.inc.φ']"
```

Four keys, any of them optional: `eo` is the program, `xmir` asks about the
XMIR prepared for the rules, and one key per table holds the XPaths that
must match it. A later rule adds `links`, and the runner already knows the
key.

The runner sits in `InferringTest`, next to the packs it drives, because
`Inferring` is what turns EO into all three documents. It reports *which*
XPath of *which* table matched nothing rather than stopping at the first
failure:

```
Expected: an empty collection
     but: <[provides: /provides/type[@id='Φ.app.t']/attr[@name='NOPE']]>
```

Ten packs replace `ProvidesTest` and `NeedsTest`, which are deleted — every
case they covered is a pack now, and each one reads as a sentence about EO
rather than a wall of angle brackets. What stays in Java is what no EO
source can express: how many files a program has, where the prepared XMIR
lands, and that a file whose source is gone leaves the table. The machinery
tests in `eo-inference` (`GroupedTest`, `CluesTest`) stay too, since they
take rows and lambdas as input, not programs.

The one wrinkle worth naming: a `@ParameterizedTest` cannot also take an
injected temporary directory, so the runner uses a `@Mktmp` field, exactly
as `MjPrintTest` already does for the same reason.

Closes: #6516


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (594 >= 100); -8 for way too many hits-of-code (594 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+676](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6520, opened 2026-08-09
Filter the directory walk while it walks

### LINKED ISSUE #6484: directory.eo code causes excessive growth when walking a directory
The puzzle `5751-03b30262` from #5751 has to be resolved: 

https://github.com/objectionary/eo/blob/873ad3521b47351908242932e16a76c9d1af685f/eo-runtime/src/main/eo/directory.eo#L12-L15

The puzzle was created by @yegor256 on 08-Aug-26. 

Thayorns:
`directory.walk` used to build the whole tree first and filter it afterwards. `walked` collected every relative path into a tuple, and a `tuple.reduced` on top of it threw away the ones the glob did not match. Two full recursions over the same entries, one nested inside the other, and the inner one paid for `names.at index` at every step — a lookup that peels one tail per index, so a directory of `n` entries cost `n` more frames at the deepest point and quadratic time overall.

Now the glob is applied where the entry is produced. `picked` appends a path to the accumulator only when `matched` says yes, and `walked` hands the accumulator down the tree, so nothing is collected that is later discarded and the second pass is gone. `stepped` walks the listing by `head` and `tail` instead of by index, which drops the quadratic lookups; it recurses on the earlier entries first and appends the last one, so the pre-order that `directory.deleted` relies on is unchanged. The new test pins that order: a directory comes back before the file inside it.

This makes the walk cheaper, not free. A puzzle stays behind for the rest, because it is not really a problem of this file. EO has no loop that does not grow the stack — `while` nests a frame per cycle and `tuple.reducedi` one per element — so any traversal written here still costs a frame per entry. Removing that needs an iterative primitive in the runtime, not another rewrite of `walk`.

Closes #6484

### DIFF
--- a/eo-runtime/src/main/eo/directory.eo
+++ b/eo-runtime/src/main/eo/directory.eo
@@ -94,7 +94,6 @@
       number.random clock
   [glob] > walk
     [accum prefix dir] >> walked
-      dir.listed >> names
       [rest] >> stepped
         if. > @
           rest.length.eq 0
@@ -103,8 +102,8 @@
             child.as-file.is-directory false
             walked earlier sub child.as-dir
             picked earlier sub
-        stepped rest.tail > earlier
         dir.as-path.resolved name > child
+        stepped rest.tail > earlier
         rest.head > name
         if. > sub
           prefix.length.eq 0
@@ -113,7 +112,7 @@
             path.separator
             prefix
             name
-      | names > @
+      | dir.listed > @
     | > @
       *
       ""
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 ping

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1133](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6521, opened 2026-08-09
#5721: keep an argument applied to a data literal when unhexing

### LINKED ISSUE #5721: `StUnhex` in eo-printer loses arguments when applied directly to data literals

morphqdd:
`StUnhex` called `setTextContent` on the wrapper element of a data literal, which deletes every child of that element. The byte payload is only the first child: a bare literal can be the head of an application, as in `42 5`, and then the applied argument sits next to the payload and was deleted with it. The argument disappeared from the printed EO without any error.

Now only the payload child is swapped for the readable text, so the argument stays. The payload is the `α0` argument of the wrapper, so the arguments that follow move up one place, otherwise `unnecessary-as.xsl` no longer sees a run starting at `α0` and the printer emits `5:1` instead of `5`.

The print-pack covers a number, a string, a bytes literal and a compact tuple as the applied argument. Without the fix it prints `42 > a`, `"str" > b`, `01-02 > c` and `42 > d`.

Closes #5721

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Payload.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Payload.java
@@ -1,0 +1,63 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.printer;
+
+import com.github.lombrozo.xnav.Filter;
+import com.github.lombrozo.xnav.Xnav;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.w3c.dom.Element;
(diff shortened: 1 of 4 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (124 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3680](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6522, opened 2026-08-09
Free the heap block through a scoped malloc

### LINKED ISSUE #6507: Tests in HeapsTest.java leave blocks allocated and unreleased in global Heaps.INSTANCE
`failsOnWriteIfOffsetPlusLengthOverflows` (line 80), `failsOnReadIfOutOfBounds` (101), `failsOnReadIfOffsetPlusLengthOverflows` (112) and `readsByOffsetAndLength` (137) in `eo-runtime/src/test/java/org/eolang/HeapsTest.java` call `Heaps.INSTANCE.malloc(new HeapsTest.PhFake(), n)` and never call `free`. Three of them allocate inline inside the `assertThrows` argument, so the identifier is not even held long enough to release it.

`Heaps.INSTANCE` is a JVM-wide singleton and surefire runs with `reuseForks=true` (`pom.xml:448`), so those blocks sit in its map for every test class that follows in the same fork.

Every other test in the file does free, but as its last statement, so a failed assertion skips the cleanup as well. Releasing the four that never release, and preparing the state each test needs at its start rather than relying on the previous test's cleanup, keeps the singleton honest.

@yegor256

Thayorns:
Four tests in `HeapsTest` allocated a block in the JVM-wide `Heaps.INSTANCE` and never released it. Three of them called `malloc` inline inside the `assertThrows` argument, so the identifier was gone before anything could free it, and surefire runs with `reuseForks=true`, which leaves those blocks in the singleton's map for every test class that follows in the same fork.

Rather than sprinkle more `free` calls around, `malloc` now takes the scope as an argument: it allocates the block, hands the identifier to the scope, and frees it in a `finally`, so the block goes away whether the scope returns or throws. `EOmalloc$EOof.lambda()` is the caller — it is the only place in `eo-runtime/src/main/java` that allocates at all, so every `malloc.of` object in EO now goes through the scope, and its own try/finally is gone.

Every test that allocates and then reads or writes takes its block through the scope now, which is what proves the block is released: `freesSuccessfully` reads the identifier the scope returned and fails, and `reusesObjectAfterScopeThrows` allocates twice through the same object across a scope that raised.

Neither the two-argument `malloc` nor `free` has a caller left in `main` outside `Heaps` itself. Both should become private, as @yegor256 asked in #6507, but doing that here means converting the remaining tests too, which puts the diff around 300 lines against the 200-line `pr-size` limit. The puzzle in `Heaps` names exactly what is left: the negative-argument, size and resize tests, the two `free` probes in `EOmallocEOofTest`, and `failsOnClearingEmptyBlock`, which nobody can reach once `free` is private.

This one is 165 lines, above the 40-100 band and below the limit.

Closes #6507


Thayorns:
@yegor256 ping

eo-runtime/src/main/java/org/eolang/Heaps.java
```diff
@@ -65,6 +69,23 @@ int malloc(final Phi phi, final int size) {
...
 
+    /**
+     * Allocate a block in memory, let the scope use it, and free it afterwards.
+     * @param phi Object
+     * @param size How many bytes
+     * @param scope What to do with the identifier of the block
+     * @param <T> Type of what the scope returns
+     * @return What the scope returns
+     */
+    <T> T malloc(final Phi phi, final int size, final IntFunction<T> scope) {
```
yegor256, line 81: @Thayorns who is the user of this method? Only tests? This is wrong. Read this: https://www.yegor256.com/2015/02/12/top-down-design.html

Imagine we merge your code and then someone opens the repo, finds this new `malloc()` method and asks: "What is it for?" -- and will delete it, together with tests.

Thayorns: The user is production code, not tests: `EOmalloc$EOof.lambda()` calls it, and that class is the only place in `eo-runtime/src/main/java` that allocates anything at all, so every `malloc.of` object in EO goes through this method. Nothing else in `main` touches `Heaps.malloc` or `Heaps.free` any more.

What did stay test-only is the opposite pair, and I think that is the part you are pointing at: the two-argument `malloc` and `free` now have no caller in `main` outside `Heaps` itself. They should go private, exactly as you asked in #6507. I just pushed the migration of the allocating tests onto the scope; the puzzle in `Heaps` covers the remainder, because converting every last test and flipping both methods to private is about 300 changed lines and `pr-size` rejects anything above 200.

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2913](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (166 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4106](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6440, opened 2026-08-06
`PhCoverage suppresses write and doesn't create a new file when eo.coverageFile property changes`


gemshrine:
## What happens

PhCoverage documents that it reads eo.coverageFile on every touch. However, it deduplicates coverage records solely by location. After a hit is written to file A, changing the property to file B and touching the same location suppresses the write, so file B remains absent.

## Reproduction

Dataize a PhCoverage instance once with eo.coverageFile pointing to file A. Change the property to file B and dataize the same instance again. A contains the record; B is not created.

## Root cause

PhCoverage.java keeps a process-wide set keyed only by the location record, while append() reads the destination property later. The deduplication key omits the destination path.

## Expected behavior

The same record should be written once per configured coverage destination, or the deduplication state should reset when the destination changes.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2023](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6524, opened 2026-08-09
fix(#6508): replace the static SEEN set with a per-object attribute

### LINKED ISSUE #6508: `PhCoverage.SEEN` static set in `PhCoverage.java` isn't cleared, causing `recordsEveryLocationExactlyOnce` to fail on rerun
`PhCoverage.SEEN` at `eo-runtime/src/main/java/org/eolang/PhCoverage.java:43` is a static set, JVM-wide, and nothing ever clears it. `hit()` at line 138 writes a record only when `SEEN.add(record)` returns true.

`PhCoverageTest.recordsEveryLocationExactlyOnce` at `eo-runtime/src/test/java/org/eolang/PhCoverageTest.java:44` records `Φ.foo:7:3` and `Φ.bar:9:5`, then reads them back from a file in a fresh `@Mktmp` directory. On a second execution in the same fork — a rerun, a repeat, or any later test touching those same locations — `SEEN` already holds both records, so nothing is written, the file in the new temporary directory is never created, and `Files.readAllLines` throws `NoSuchFileException` rather than failing on an assertion.

The test carefully saves and restores the `eo.coverageFile` property it changes, but not the static set it fills, so it prepares only half the state it depends on.

@yegor256

Thayorns:
`PhCoverage` kept the locations it had already written in a static set, so nothing ever cleared them for the life of the JVM. A rerun against the same destination found the set already full, wrote nothing, and blew up on `NoSuchFileException` instead of on an assertion.

The set is now an attribute, handed to the object through its constructor, and a copy shares the set of the wrapper it came from — so a location hit repeatedly, or through a copy, is still written once. The key stays `destination + location`, so the per-destination behaviour of #6440 survives. The added test records a location, deletes the file, and records the same location again through a fresh wrapper; on `master` the second run writes nothing and the read dies with `NoSuchFileException`.

The location, the line and the column now reach the object as one string, which the transpiler builds, so the class keeps three attributes and its constructor stays inside qulice's parameter limit.

That leaves the wrapper the transpiler emits with a private set of its own, so one location touched through many instances of the same object lands in the file once per instance. Sharing one set across the whole program means the generated code has to hand the same set to every wrapper it builds, which is a puzzle rather than a line of this change.

Closes #6508


yegor256:
@Thayorns qulice fails here, please fix

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (147 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1169](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6525, opened 2026-08-09
#6499: fix cachesPhiViaNewRecursively to check its own counter

### LINKED ISSUE #6499: cachesPhiViaNewRecursively checks the wrong counter in PhDefaultTest.java
The `RecursivePhiViaNew` fixture in `eo-runtime/src/test/java/org/eolang/PhDefaultTest.java:965` decrements its own `RecursivePhiViaNew.count`, but the termination check on the very next line reads `RecursivePhi.count` — the counter belonging to the sibling fixture. So the `RecursivePhiViaNew.count = 3` that `cachesPhiViaNewRecursively` sets at line 409 controls nothing at all.

`RecursivePhi.count` is a `static int` and defaults to 0, so the condition is true on the first evaluation, the fixture returns `new Data.ToPhi(0L)` straight away and never recurses. The test then asserts `equalTo(0.0)`, which is exactly what that first non-recursive branch hands back. The assertion holds whether or not phi is cached through `new`, which is the one thing the test is named for, so a regression there would ship green.

PMD already said so. The `@SuppressWarnings("PMD.UnusedPrivateField")` at line 950 sits on `RecursivePhiViaNew.count` because the field really is written at line 965 and read nowhere — which is the bug itself, silenced.

There is a second effect worth mentioning while the file is open: `eo-runtime/pom.xml:212-214` sets `junit.jupiter.execution.parallel.mode.default = concurrent` and `mode.classes.default = concurrent`, and `PhDefaultTest` carries neither `@Execution(SAME_THREAD)` nor `@ResourceLock`. `hesPhiRecursively` at line 399 sets and drives down that same `RecursivePhi.count`, so the two methods race on it.

Reading `RecursivePhiViaNew.count` at line 967 makes the counter the test sets the one that governs the recursion; the static counters shared across concurrently running methods want a lock or a same-thread execution mode on the class.

@yegor256

morphqdd:
Fixes the recursion check in `cachesPhiViaNewRecursively`.

The `RecursivePhiViaNew` fixture decremented `RecursivePhiViaNew.count`, but the termination check on the next line read `RecursivePhi.count`, the sibling fixture's counter. That counter defaults to 0, so the fixture returned `new Data.ToPhi(0L)` on the first call without ever recursing, and the test's assertion held either way.

Fix: read `RecursivePhiViaNew.count` in the termination check, so the counter the test sets actually governs the recursion. Also dropped the now-stale `@SuppressWarnings("PMD.UnusedPrivateField")` on `count`, since it is genuinely read now.

Closes #6499

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
+++ b/eo-runtime/src/test/java/org/eolang/PhDefaultTest.java
@@ -961,7 +961,6 @@ final class PhDefaultTest {
         /**
          * Count.
          */
-        @SuppressWarnings("PMD.UnusedPrivateField")
         private static int count;
 
         /**
@@ -978,7 +977,7 @@ final class PhDefaultTest {
                             rho -> {
                                 --PhDefaultTest.RecursivePhiViaNew.count;
                                 final Phi result;
-                                if (PhDefaultTest.RecursivePhi.count <= 0) {
+                                if (PhDefaultTest.RecursivePhiViaNew.count <= 0) {
                                     result = new Data.ToPhi(0L);
                                 } else {
                                     result = new Data.ToPhi(


yegor256:
@morphqdd merge conflicts here

morphqdd:
@yegor256, please review

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (3) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3812](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6526, opened 2026-08-09
#6500: fix tautological cache test in OnClasspathTest

### LINKED ISSUE #6500: returnsSameAnswerOnRepeatedLookup asserts has(cls) equals has(cls) in OnClasspathTest.java
`returnsSameAnswerOnRepeatedLookup` in `eo-runtime/src/test/java/org/eolang/OnClasspathTest.java:36` builds its expected value by calling the method under test a second time on the same input: `assertThat(OnClasspath.has(cls), Matchers.is(OnClasspath.has(cls)))`. That is `has(cls) == has(cls)`, which cannot come out false.

`OnClasspath.has` at `eo-runtime/src/main/java/org/eolang/OnClasspath.java:44` is `CACHE.computeIfAbsent(cls, OnClasspath::probe)` over a `ConcurrentHashMap`, and `probe` is a plain `Class.forName` with the exception caught. Two consecutive calls agree by construction, and no caching defect changes that: a cache that never stored would run the deterministic `probe` twice and agree, and a cache that stored the wrong answer would hand back the same wrong answer twice. The failure message claims the cache must agree with itself on repeat, but nothing the cache could do would make it red.

The behaviour actually worth pinning is that the second lookup does not probe the classpath again. Passing a name whose `probe` result can be observed once — for instance counting the probes by asking for a class that is absent and then checking `has` is still false without a second `Class.forName` — is what the test name promises; comparing `has` to a literal `true`/`false` for a known-present and a known-absent class is already covered by the two tests above it.

@yegor256 @maxonfjvipon

morphqdd:
Fixes the assertion in `returnsSameAnswerOnRepeatedLookup`.

The test asserted `OnClasspath.has(cls)` equals `OnClasspath.has(cls)`, which is `x == x` and cannot fail regardless of whether caching works, whether it caches the right value, or whether it re-probes the classpath on every call.

Replaced it with `returnsCachedAnswerInsteadOfProbingClasspathAgain`: it probes a class once (populating the cache), overwrites the cache entry via reflection to a wrong value, and asserts `has` returns that wrong (cached) value instead of recomputing. Verified this fails if `has` is changed to call `Class.forName` directly without going through the cache.

Closes #6500


yegor256:
@morphqdd merge conflicts here, please fix

morphqdd:

eo-runtime/src/test/java/org/eolang/OnClasspathTest.java
```diff
@@ -34,13 +36,24 @@ void doesNotFindMissingClass() {
...
     @Test
-    void returnsSameAnswerOnRepeatedLookup() {
-        final String cls = "org.eolang.PhNest";
-        MatcherAssert.assertThat(
-            "The cached lookup must agree with itself on repeat, but it didn't",
-            OnClasspath.has(cls),
-            Matchers.is(OnClasspath.has(cls))
-        );
+    @SuppressWarnings("PMD.AvoidAccessibilityAlteration")
+    void returnsCachedAnswerInsteadOfProbingClasspathAgain() throws ReflectiveOperationException {
```
yegor256: @morphqdd looks like a super dirty testing tactics

morphqdd: Fair. Dropped the part that poisons the cache with a wrong value through reflection. The test now only reads the cache after one call and checks the entry landed there, no forced-lie/finally-cleanup dance.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4052](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (24) hits-of-code; +1 to give you at least something.  Your running score is [+2795](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6527, opened 2026-08-09
#6506: log the cache path, not the unused source, in FpUpdateFromCache

### LINKED ISSUE #6506: FpUpdateFromCache logs the source path in the slot where the cache path belongs
`FpUpdateFromCache.apply` at `eo-maven-plugin/src/main/java/org/eolang/maven/FpUpdateFromCache.java:34` logs `"Updating only target %[file]s from cache %[file]s"` and passes `target, source`. The content is read from `this.cache.get()` at line 39; `source` is never read by this method at all.

Its own test knows as much: `FpUpdateFromCacheTest.appliesFromCache` passes `Paths.get("/dev/null")` as the source precisely because nothing touches it. So the debug line names a path that had no part in the copy, and anybody chasing a stale-cache problem through these logs is sent to the wrong file.

The sibling `FpUpdateBoth` gets it right — it binds `final Path che = cache.get()` at line 26 and logs that. Passing `this.cache.get()` in the same slot here does it.

@yegor256

morphqdd:
FpUpdateFromCache.apply logged "cache %[file]s" but passed source, the argument it never reads. The content actually comes from this.cache.get(); source is unused (the test already passes Paths.get("/dev/null") for it).

Fix: bind final Path cached = this.cache.get() once and use it both in the log message and in the copy, matching how FpUpdateBoth already binds and logs its cache path.

Closes #6506

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/FpUpdateFromCache.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/FpUpdateFromCache.java
@@ -31,11 +31,12 @@ final class FpUpdateFromCache implements Footprint {
 
     @Override
     public Path apply(final Path source, final Path target) throws IOException {
+        final Path cached = this.cache.get();
         Logger.debug(
             FpUpdateFromCache.class,
             "Updating only target %[file]s from cache %[file]s",
-            target, source
+            target, cached
         );
-        return new Saved(new TextOf(this.cache.get()), target).value();
+        return new Saved(new TextOf(cached), target).value();
     }
 }


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (5) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3561](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6528, opened 2026-08-09
#6505: fix PenaltyKey.BRACKET javadoc to describe nesting depth

### LINKED ISSUE #6505: PenaltyKey.BRACKET javadoc describes a per-line ordinal, but Penalty charges by nesting depth
`PenaltyKey.BRACKET` at `eo-printer/src/main/java/org/eolang/printer/PenaltyKey.java:27` documents itself as "Points charged per opening parenthesis, progressively: the k-th parenthesis on a line costs k times this weight". `Penalty.brackets` at `eo-printer/src/main/java/org/eolang/printer/Penalty.java:279` charges `depth + 1`, where `depth` counts the parentheses still open at that point and decrements on every `)`.

The two rules diverge on any line carrying sibling groups. In `foo (a) (b)` the second `(` is the second one on the line but sits at depth zero, so it costs one unit rather than two. Only nesting compounds the charge, never mere position.

`Penalty`'s own class javadoc states the depth rule correctly at line 26 ("multiplied by its nesting depth plus one"), so the enum is the one that is wrong — and it is the docblock somebody reads when they come to tune the weight.

@yegor256

morphqdd:
PenaltyKey.BRACKET's javadoc said the k-th parenthesis on a line costs k times the weight, a per-line ordinal. Penalty.brackets actually charges depth + 1, where depth is how many parentheses are still open (nesting), not how many have appeared on the line so far. The two diverge on sibling groups: in "foo (a) (b)" the second "(" is second on the line but at depth zero, so it costs one unit, not two.

Reworded the enum javadoc to match Penalty's own class javadoc, which already describes the depth rule correctly.

Closes #6505

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/PenaltyKey.java
+++ b/eo-printer/src/main/java/org/eolang/printer/PenaltyKey.java
@@ -24,8 +24,8 @@ public enum PenaltyKey {
     INDENT(2),
 
     /**
-     * Points charged per opening parenthesis, progressively: the k-th
-     * parenthesis on a line costs k times this weight.
+     * Points charged per opening parenthesis, progressively: a parenthesis
+     * nested k levels deep costs k plus one times this weight.
      */
     BRACKET(19),
 


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3573](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6529, opened 2026-08-09
#6504: use different xml per thread in doesNotFailWithDifferentXmlInMultipleThreads

### LINKED ISSUE #6504: doesNotFailWithDifferentXmlInMultipleThreads uses the same XML as its sibling in StrictXmirTest.java
`doesNotFailWithDifferentXmlInMultipleThreads` at `eo-parser/src/test/java/org/eolang/parser/StrictXmirTest.java:48` hands every thread `StrictXmirTest.xmir("https://www.eolang.org/XMIR.xsd")` — the same schema, the same document, built the same way. Its failure message says "StrictXmir should not fail in different threads with different xmls", but there is only ever one xml.

`doesNotFailOnTheSameOperation`, seventeen lines below at line 65, does exactly that on purpose: same `Together`, same `xmir` call, same `::asList`. The two bodies differ only in a `return` and their message strings, so the file runs the same-xml case twice under two names and never covers the different-xml one — 20 repetitions each, both over the network.

Giving the first one a document that differs per thread makes it test what its name claims; the schema URL already varies between the two halves of this file, so there is a ready way to vary it.

@yegor256

morphqdd:
doesNotFailWithDifferentXmlInMultipleThreads built every thread the same XMIR against the same schema URL, so it was really testing the same-xml case a second time under a name that promises different xmls. doesNotFailOnTheSameOperation already covers that case seventeen lines below.

Now each thread alternates between the remote schema and the local version-specific schema, so threads validate genuinely different documents concurrently.

Closes #6504

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/StrictXmirTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/StrictXmirTest.java
@@ -49,10 +49,16 @@ final class StrictXmirTest {
         Assertions.assertDoesNotThrow(
             new Together<>(
                 thread -> {
-                    return new StrictXmir(
-                        StrictXmirTest.xmir("https://www.eolang.org/XMIR.xsd"),
-                        tmp
-                    ).inner();
+                    final String schema;
+                    if (thread % 2 == 0) {
+                        schema = "https://www.eolang.org/XMIR.xsd";
+                    } else {
+                        schema = String.format(
+                            "https://www.eolang.org/xsd/XMIR-%s.xsd",
+                            StrictXmirTest.version()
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3585](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6530, opened 2026-08-09
#6501: remove three MainTest tests that only assert JDK stream behavior

### LINKED ISSUE #6501: Three tests in MainTest.java assert JDK stream behaviour and never touch Main
Three tests in `eo-runtime/src/test/java/org/eolang/MainTest.java` never touch `Main`, or anything else in `org.eolang`. `readsStreamCorrectly` at line 123 and `readsSimpleStreamCorrectly` at line 143 wrap an inline `ByteArrayInputStream` in `Channels.newChannel`, `Channels.newReader` and a `BufferedReader`, then assert that `readLine().length()` is greater than 0 and than 1. `readsBytesCorrectly` at line 163 asserts that `new ByteArrayInputStream(...).read()` on a non-empty array is greater than 0.

Every one of those is a property of the JDK over a literal declared two lines above it. They can only go red if `ByteArrayInputStream` or `BufferedReader` is broken, and they carry no relationship to the class the file is named for, so they consume CI time without covering `Main` and leave the impression that its stream handling is tested when nothing of it is.

The first two also build the `BufferedReader` inline inside the assertion and never close it, so the reader the test opens outlives the test.

If the intent was to cover how `Main` reads its input, the tests should go through `Main` and assert on what it produces, the way `deliversCleanOutput` at line 58 does; if there was no such intent, dropping the three removes dead weight from the file.

@maxonfjvipon @yegor256 take a look

morphqdd:
readsStreamCorrectly, readsSimpleStreamCorrectly and readsBytesCorrectly in MainTest.java never touched Main or anything in org.eolang. They wrapped an inline ByteArrayInputStream and asserted properties of BufferedReader/ByteArrayInputStream that only the JDK controls, so they carried no relationship to the class the file is named for and could only fail if the JDK itself was broken. The first two also left their BufferedReader unclosed.

Removed the three and the now-unused imports (BufferedReader, ByteArrayInputStream, IOException, Channels, StandardCharsets). deliversCleanOutput already covers Main's output through Main itself.

Closes #6501

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/MainTest.java
+++ b/eo-runtime/src/test/java/org/eolang/MainTest.java
@@ -7,11 +7,6 @@ package org.eolang;
 import com.yegor256.Jaxec;
 import com.yegor256.Jhome;
 import com.yegor256.Result;
-import java.io.BufferedReader;
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
-import java.nio.channels.Channels;
-import java.nio.charset.StandardCharsets;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3537](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6531, opened 2026-08-09
#6513: fix two non-existent stylesheet paths in XslBench

### LINKED ISSUE #6513: The JMH performance regression check is silently dead: XslBench points at two non-existent stylesheets
The `Performance Regression Check` workflow (`.github/workflows/jmh-benchmark-action.yml`) cannot produce a comparison on any PR, because the only JMH benchmark in the repo throws during class initialisation and yields **zero** results. Worse, `mvn jmh:benchmark` still exits 0, so nothing surfaces except a vague bot comment on the PR:

> ⚠️ Benchmark Comparison Unavailable — Unfortunately, one of the benchmarks is missing, and we couldn't generate a performance comparison report.

## Root cause

`eo-parser/src/test/java/benchmarks/XslBench.java` lists seven stylesheets, two of which do not exist:

| path in `XslBench` | status |
|---|---|
| `/org/eolang/parser/parse/resolve-before-star.xsl` | **missing** — the real file is `resolve-before-star**s**.xsl` (plural) |
| `/org/eolang/parser/shake/build-fqns.xsl` | **missing** — the real file is `/org/eolang/parser/**parse**/build-fqns.xsl`; there is no `shake/` directory |

The other five resolve fine.

`TrClasspath` resolves eagerly, so the static initialiser blows up before any benchmark runs:

```
java.lang.ExceptionInInitializerError
Caused by: java.lang.IllegalArgumentException: Path '/org/eolang/parser/parse/resolve-before-star.xsl' not found in classpath
	at com.yegor256.xsline.TrClasspath.<init>(TrClasspath.java:82)
	at benchmarks.XslBench.<clinit>(XslBench.java:44)
```

## Reproducing

```
$ mvn jmh:benchmark -pl eo-parser -Djmh.rf=json -Djmh.rff=benchmark.json
...
Benchmark  Mode  Cnt  Score   Error  Units     <- no rows at all
Benchmark result is saved to benchmark.json
[INFO] BUILD SUCCESS                            <- exit 0 despite the error

$ cat eo-parser/benchmark.json
[
]
```

Six bytes. The comparison action then has nothing to diff on either the base or the PR side, which is exactly the "one of the benchmarks is missing" message.

## How long

`resolve-before-star.xsl` (singular) has **never** existed as a tracked path in this repository, and I found no `parser/shake/` resource in history either. Checking the last six revisions of `XslBench.java`, every one of them has the same 2 unresolvable paths — so this benchmark has been producing empty output for as long as those lines have been in the file, not since some recent rename.

## Suggested fix

Two one-line path corrections in `XslBench.java`:

```diff
-            "/org/eolang/parser/parse/resolve-before-star.xsl",
+            "/org/eolang/parser/parse/resolve-before-stars.xsl",
...
-            "/org/eolang/parser/shake/build-fqns.xsl"
+            "/org/eolang/parser/parse/build-fqns.xsl"
```

Two things worth doing alongside:

1. **Make the failure loud.** A benchmark run that produces zero results should fail the build rather than exit 0 — otherwise this silently rots again the moment another stylesheet is renamed. Asserting that `benchmark.json` is a non-empty array in the workflow would be enough.
2. **Cover the stylesheets that actually cost time.** `XslBench` exercises seven cheap parse sheets and none of the expensive ones. Per the README benchmark table, `to-java.xsl` (34.43%), `classes.xsl` (28.86%) and `set-locators.xsl` (8.09%) are 71% of the build's XSL time and none is covered, so even once fixed this benchmark would not catch a regression in the places that matter.

Found while profiling `set-locators.xsl` for #6509 / #6510 — the missing comparison on that PR is this bug, not that PR.

morphqdd:
XslBench.LINE listed two stylesheet paths that never existed: resolve-before-star.xsl (real file is resolve-before-stars.xsl, plural) and shake/build-fqns.xsl (real file is parse/build-fqns.xsl, no shake/ directory). TrClasspath resolves eagerly, so the static initializer threw before any benchmark ran, mvn jmh:benchmark still exited 0, and benchmark.json came out as an empty array. That is why the Performance Regression Check bot has been posting "one of the benchmarks is missing" on every PR.

Fixed both paths. Verified with mvn jmh:benchmark -pl eo-parser -Djmh.rf=json -Djmh.rff=benchmark.json: before the fix, benchmark.json is []; after, it holds a real result for XslBench.manySheetsOnLargeXmir.

Closes #6513

### DIFF
--- a/eo-parser/src/test/java/benchmarks/XslBench.java
+++ b/eo-parser/src/test/java/benchmarks/XslBench.java
@@ -45,11 +45,11 @@ public class XslBench {
         new TrClasspath<>(
             "/org/eolang/parser/parse/move-voids-up.xsl",
             "/org/eolang/parser/parse/validate-before-stars.xsl",
-            "/org/eolang/parser/parse/resolve-before-star.xsl",
+            "/org/eolang/parser/parse/resolve-before-stars.xsl",
             "/org/eolang/parser/parse/wrap-method-calls.xsl",
             "/org/eolang/parser/parse/const-to-dataized.xsl",
             "/org/eolang/parser/parse/stars-to-tuples.xsl",
-            "/org/eolang/parser/shake/build-fqns.xsl"
+            "/org/eolang/parser/parse/build-fqns.xsl"
         ).back()
     );
 


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3549](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6439, opened 2026-08-06
win32.inet_addr throws BufferOverflowException for IPv6 input instead of returning an invalid-address status


gemshrine:
## What happens

win32.inet_addr accepts a textual address through InetAddrFuncCall. For IPv6 input such as ::1, InetAddress.getByName() returns sixteen bytes, but the implementation copies them into a four-byte IPv4 buffer. The call therefore throws BufferOverflowException instead of returning the normal invalid-address status.

## Reproduction

Calling InetAddrFuncCall.make() with ::1 on current master throws BufferOverflowException at InetAddrFuncCall.java:41.

## Root cause

InetAddrFuncCall.java allocates ByteBuffer.allocate(4) and only catches UnknownHostException. It does not reject a resolved address whose byte array is not four bytes long.

## Expected behavior

Non-IPv4 input should use the existing invalid-address result (-1 with WSAEINVAL), just as an unresolvable input does, rather than leaking a Java buffer exception.

gemshrine:
Resolved by #6533.

PR #6533, opened 2026-08-09
#6439: reject non-IPv4 addresses in win32.inet_addr instead of throwing

morphqdd:
InetAddrFuncCall allocated a 4-byte ByteBuffer and put() the resolved address into it directly. For an IPv6 input like ::1, InetAddress.getByName resolves to 16 bytes, so the put() overflowed the 4-byte buffer and threw BufferOverflowException instead of returning the existing invalid-address result (-1 with WSAEINVAL) that an unresolvable host already gets.

Now the resolved address length is checked before use: only a 4-byte (IPv4) result is converted, anything else (including the unresolvable-host case, which now yields a 0-length array) falls into the existing invalid-address branch.

Added an EO test in win32.eo next to the existing localhost test, following the same os.is-windows guard, asserting win32 inet_addr on "::1" returns win32.invalid.

Refs #6439

### DIFF
--- a/eo-runtime/src/main/eo/win32.eo
+++ b/eo-runtime/src/main/eo/win32.eo
@@ -61,6 +61,14 @@
           "127.0.0.1"
       16777343
 
+  os.is-windows.not.or ++> rejects-an-ipv6-address-as-invalid
+    eq.
+      code.
+        win32 *1
+          "inet_addr"
+          "::1"
+      win32.invalid
+
   --> stops-on-write-size-being-negative
     os.is-windows.not.if > @
       1.div 0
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3824](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6534, opened 2026-08-09
fix(#5737): reject binding on a reversed receiver chain

### LINKED ISSUE #5737: Reversed-dispatch receiver binding check (R-6.6.3) is bypassed when the receiver is a multi-link .method chain
`Bindings.observeReversedChild` (eo-parser/src/main/java/org/eolang/parser/Bindings.java:137-148) enforces R-6.6.3 ("reversed-dispatch receiver cannot carry a binding") only by throwing synchronously when `outer != null` at the moment the receiver's first line is seen; it never calls `parent.observeBinding(...)` for the receiver, so `Level.argpending` (eo-parser/src/main/java/org/eolang/parser/Level.java:138) stays false for it.

That matters because `LnMethod.into()` (eo-parser/src/main/java/org/eolang/parser/LnMethod.java:79-80) unconditionally calls `stack.below().upgradeArgBinding()` whenever a same-indent `.method` continuation carries an outer `:label` binding. `upgradeArgBinding()` (Level.java:492-494) sets `argbound = true` with no check on `argpending`, and `commitArg()` (Level.java:502-519) only validates/uses that state `if (this.argpending)`. So when the receiver itself is a multi-line chain and the binding lands on a later link rather than the first line, the upgrade is silently discarded and no error is raised. Example: `foo > main` / `  if.` / `    cond` / `    .baz:x` / `    then` / `    other` parses cleanly and emits `<o as="x" base=".baz" .../>`, whereas the single-line (`if. cond:x then else`) and vertical-direct-receiver (`if.` / `cond:x` / `then` / `other`) equivalents both correctly raise `reversed-dispatch receiver cannot carry a binding` (see the existing tests `rejectsBoundReversedReceiver` and `rejectsBindingOnVerticalReceiver` in eo-parser/src/test/java/org/eolang/parser/EoTest.java).

Suggested fix: in `observeReversedChild`, when handling the receiver (`parent.children() <= 1`), start binding tracking for it too (e.g. call `parent.observeBinding(outer != null, span)` instead of the throw-only check) so a later `upgradeArgBinding()` from a chain continuation has state to validate against, then check the receiver's final bound state before the group's arg-tracking begins.

VasilevNStas:
Fixes #5737.

## Problem

R-6.6.3 (a reversed-dispatch receiver cannot carry a binding) is bypassed when the receiver is a multi-link `.method` chain. `Bindings.observeReversedChild` only throws when the binding sits on the receiver's first line; a binding on a later chain link lands via `LnMethod.into()` → `Level.upgradeArgBinding()`, which sets `argbound` with no validation and is silently discarded because the receiver was never tracked (`argpending` stays false). So `if.` / `cond` / `.baz:x` parses cleanly and emits `<o as="x" base=".baz"/>`, while the single-line and vertical-direct equivalents correctly raise the error.

## Solution

Reject the binding at the moment it would attach through a bare-reversed receiver host. `Bindings.checkReceiverUpgrade(below, span)` throws R-6.6.3 when a `.method` continuation carries an outer binding under a `Kind.BARE_REVERSED` parent — binding a later link of the receiver's chain is the same as binding the receiver.

## Changes

- `eo-parser/src/main/java/org/eolang/parser/Bindings.java` — new `checkReceiverUpgrade` (R-6.6.3 for chain continuations).
- `eo-parser/src/main/java/org/eolang/parser/LnMethod.java` — call it before `upgradeArgBinding()`.
- `eo-parser/src/test/java/org/eolang/parser/EoTest.java` — regression tests.

## Tests

- `EoTest.rejectsBindingOnReversedReceiverChain` — the multi-link `.baz:x` shape (verified to fail on pre-fix code).
- `EoTest.acceptsUnboundReversedReceiverChain` — the same chain without a binding still parses.
- Existing `rejectsBoundReversedReceiver`, `rejectsBindingOnVerticalReceiver`, `rejectsMixedBindingsAmongReversedArgs` still pass.

@yegor256 please take a look.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Bindings.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Bindings.java
@@ -74,6 +74,25 @@ final class Bindings {
     }
 
     /**
+     * Reject a binding that a same-indent {@code .method} chain
+     * continuation would attach through a bare-reversed receiver host —
+     * R-6.6.3. When the chain being extended is the receiver of a
+     * reversed dispatch, a binding on a later link names the whole
+     * reversed expression, which is the same as binding the receiver.
+     * @param below The chain head's parent (what the upgrade would
+     *  touch), or {@code null} when there is none
+     * @param span Source span of the continuation line
+     */
+    static void checkReceiverUpgrade(final Level below, final Span span) {
+        if (below != null && below.kind() == Kind.BARE_REVERSED) {
+            throw new ParseError(
+                span.line(), span.indent(),
+                "reversed-dispatch receiver cannot carry a binding"
+            );
+        }
+    }
+
+    /**
      * Observe a freshly-pushed child line's outer binding against its
      * parent context — cross-line R-6.6.2 / R-6.6.3 / R-3.12.3.
      *
--- a/eo-parser/src/main/java/org/eolang/parser/LnMethod.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnMethod.java
@@ -80,8 +80,11 @@ final class LnMethod implements Line {
             Blanks.checkPlain(this.span, globals, emit);
         }
         Comments.seal(globals, emit, this.span);
-        if (outer != null && stack.below() != null) {
-            stack.below().upgradeArgBinding();
+        if (outer != null) {
+            Bindings.checkReceiverUpgrade(stack.below(), this.span);
+            if (stack.below() != null) {
+                stack.below().upgradeArgBinding();
+            }
         }
         emit.close();
         emit.object(
--- a/eo-parser/src/test/java/org/eolang/parser/EoTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EoTest.java
@@ -841,42 +841,6 @@ final class EoTest {
     }
 
     @Test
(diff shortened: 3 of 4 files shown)


yegor256:
@VasilevNStas merge conflicts here, please fix

yegor256:
@VasilevNStas better test this with YAML fixtures, not Java tests

VasilevNStas:
@yegor256 Both points addressed:
- Merge conflicts are resolved — the branch was rebased onto the latest master (the stale counts.yml baseline bump is gone, since counts.yml no longer keeps hardcoded numbers).
- The regression test is now a YAML fixture: `eo-parser/src/test/resources/org/eolang/parser/eo-typos/reversed-receiver-chain-binding.yaml` (checked by `checksTypoPacks`/`checksTypoMessages`), replacing the Java test in EoTest.

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+56](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6535, opened 2026-08-09
#6441: fix Saved NullPointerException on relative filename without parent

### LINKED ISSUE #6441: `Saved` throws NullPointerException on relative filename without parent directory
## What happens

Saved accepts a Path, but writing to a normal relative filename such as saved-probe.txt fails before any data is written.

## Reproduction

new Saved("x", Path.of("saved-probe.txt")).value() throws NullPointerException on current master.

## Root cause

Saved.java calls target.toFile().getParentFile().mkdirs() unconditionally. For a path with no directory component, getParentFile() returns null. The following createTempFile call also assumes target.getParent() is non-null.

## Expected behavior

A relative path with only a filename should be written in the current working directory, as Path and Files APIs normally allow.

morphqdd:
Saved.value() called this.target.toFile().getParentFile().mkdirs() unconditionally, and getParentFile() returns null for a relative path with no directory component (e.g. Path.of("saved-probe.txt")), so a plain relative filename threw NullPointerException before any data was written. The following createTempFile call had the same problem with target.getParent().

Fixed by resolving the target to an absolute path once and using its parent for both the mkdirs call and the temp file location. An absolute path always has a parent, so no null check is needed. The returned path is still the original this.target, unchanged.

Closes #6441

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
@@ -38,7 +38,8 @@ import org.cactoos.scalar.LengthOf;
 final class Saved implements Scalar<Path> {
 
     /**
-     * Absolute path to save content to.
+     * Path to save content to, absolute or relative to the current
+     * working directory.
      */
     private final Path target;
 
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (36) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3597](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6523, opened 2026-08-09
`runtime` workflow never finishes inside its 15-minute cap in `.github/workflows/runtime.yml`


Thayorns:
The job declared at `.github/workflows/runtime.yml:15` caps itself at `timeout-minutes: 15` and then, after deleting `~/.m2/repository/org/eolang`, runs two full builds in a row: line 35 installs the whole reactor with `-DskipTests`, and line 36 rebuilds `eo-runtime` with its tests. That does not fit in the budget. On #6519 both matrix jobs were killed at the cap - `runtime (windows-2022, 23)` at 15m6s and `runtime (ubuntu-24.04, 23)` at 15m15s - with the annotation "The job has exceeded the maximum execution time of 15m0s". The last hundred runs of this workflow hold no success at all: 63 cancelled and 31 failed.

This matters more than a red check, because `.github/workflows/mvn.yml:36` passes `-Deo.transpileTests=false`, so the EO unit tests written as `++>` and `-->` inside the `.eo` sources are transpiled and executed only by this workflow. While it never reaches the end of line 36, no pull request ever runs them, and a broken EO test can land on `master` without anything noticing. The `mvn` jobs stay green throughout, which makes the gap easy to miss.

Raising `timeout-minutes` to something the two builds actually fit into would fix it; the `mvn.yml` jobs need 5 to 9 minutes for the same reactor without the EO tests, so the EO test cycle needs a good deal more than the six minutes left over. Keeping the `org/eolang` artifacts out of the `JesseTG/rm` step at lines 31-33 would also buy back the time spent rebuilding them.

@yegor256 please take a look

yegor256:
@Thayorns it's a known problem: #5783, #6263

PR #6519, opened 2026-08-09
Halt the directory walk at a symbolic link

Thayorns:
Closes #6483

The `walk` in `directory.eo` asked `stat` about every entry, and `stat` follows a symbolic link. A link pointing at its own ancestor therefore looked like an ordinary directory and the walk descended into it forever, where the `Files.walk` it replaced simply listed the link and moved on.

The walk now asks `lstat` too, through a new `is-symlink` of `file`, and enters an entry only when it is a directory and not a link. Serving `lstat` meant declaring it in `CStdLib` (remapped to `lstat$INODE64` on Intel macOS, the way `stat` already is), teaching `StatSyscall` which C function to call instead of hard-wiring `stat`, and registering the name in `NamedSyscall`.

`StatSyscallTest` covers the mechanism: `lstat` reports a link to a directory as a link, `stat` reports it as the directory behind it. `file.eo` gets an EO test that a regular file is not a link. The walk that leans on the two of them has no test of its own, because EO cannot create a symbolic link while `posix` has no `symlink` syscall - that is left as a puzzle, next to the one for Windows, where `_stat64` knows nothing about links and the descent is still unbounded.

This is 163 hits, above the 100 the guidelines ask for. About 90 of them are production; the rest is the test, which cannot be smaller while the symbolic link has to be made from Java.

PR #6536, opened 2026-08-09
#6523: raise runtime workflow timeout and drop redundant org/eolang cache purge

morphqdd:
The runtime job caps itself at 15 minutes, then runs two full builds in a row: a whole-reactor install, then a rebuild of eo-runtime with its EO tests. On #6519 both matrix jobs hit the cap (15m6s and 15m15s), and the last hundred runs of this workflow hold no success at all (63 cancelled, 31 failed). Since mvn.yml already covers the same reactor without EO tests in 5 to 9 minutes, the EO test cycle needs a good deal more than the six minutes left in the 15-minute budget.

Raised timeout-minutes to 30. Also dropped the JesseTG/rm step that deletes ~/.m2/repository/org/eolang before the build: the first mvn command already runs clean install across the whole reactor (except eo-integration-tests), which rebuilds those same modules from source regardless of what's cached, so the deletion only adds rebuild time without changing what gets built.

Refs #6523

### DIFF
--- a/.github/workflows/runtime.yml
+++ b/.github/workflows/runtime.yml
@@ -12,7 +12,7 @@ name: runtime
       - master
 jobs:
   runtime:
-    timeout-minutes: 15
+    timeout-minutes: 30
     strategy:
       matrix:
         os: [ubuntu-24.04, windows-2022]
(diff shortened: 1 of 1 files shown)


morphqdd:
Closing. Leaving this workflow as-is per project owner's guidance, not something to change right now.

PR #6537, opened 2026-08-09
fix(#6441): save to a filename-only relative path

### LINKED ISSUE #6441: `Saved` throws NullPointerException on relative filename without parent directory

VasilevNStas:
Fixes #6441.

## Problem

`Saved` throws a `NullPointerException` when given a filename-only relative path such as `saved-probe.txt`: `target.toFile().getParentFile().mkdirs()` is called unconditionally, but `getParentFile()` is `null` for a path with no directory component. The following `Files.createTempFile(target.getParent(), ...)` then also NPEs.

## Solution

Resolve the target against the current working directory once, so the directory used for `mkdirs()` and for the sibling temp file is always the real, non-null parent. A filename-only relative path is now written into the current working directory, as the `Path`/`Files` APIs normally allow.

## Changes

- `eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java` — derive `dir` from `target.toAbsolutePath().getParent()` instead of the raw path's parent.
- `eo-maven-plugin/src/test/java/org/eolang/maven/SavedTest.java` — regression test.

## Tests

- `SavedTest.savesContentToFilenameOnlyRelativePath` — writes via `Path.of("saved-probe.txt")` and asserts the content (verified to fail on pre-fix code with the exact `NullPointerException` from the issue).
- Existing `savesContentToFile`, `createsMissingParentDirectories`, `overwritesContentOfExistingFile`, `leavesNoTemporaryFilesBehind` still pass.

@yegor256 please take a look.

### DIFF
--- a/.github/workflows/counts.yml
+++ b/.github/workflows/counts.yml
@@ -24,7 +24,7 @@ jobs:
           - metric: smells
             files: '*.java'
             pattern: 'static|null'
-            count: 847
+            count: 848
           - metric: atoms
             files: '*.java'
             pattern: '@XmirObject'
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java
@@ -12,6 +12,7 @@ import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import org.cactoos.Input;
 import org.cactoos.Scalar;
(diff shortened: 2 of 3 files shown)


VasilevNStas:
Closing as a duplicate: #6441 was already fixed on master by devil's commits (d8a27fbea, 522c03bde, 0d4d8fcea, merged 2026-08-09 ~19:43-20:07Z; my PR was opened at 19:52Z). Their version is more thorough (requireNonNull with messages, explicit non-null parent invariant) and includes an equivalent regression test (savesContentToRelativeFilenameWithoutParentDirectory). Nothing left to contribute here.

PR #6538, opened 2026-08-09
#6443: fix StrictXmir.copied() NullPointerException for filename-only cache path

### LINKED ISSUE #6443: `StrictXmir.copied()` fails for filename-only relative cache path: NullPointerException
## What happens

StrictXmir can cache the bundled XMIR schema at a caller-provided path. A relative cache file with no directory component causes validation to fail before the schema is copied.

## Root cause

StrictXmir.copied() calls file.getParentFile().mkdirs() without checking whether the cache file has a parent. For a filename-only path, getParentFile() is null. The download path converts to an absolute file, but the bundled-schema path does not.

## Expected behavior

StrictXmir should allow a filename-only relative cache path and create the schema in the current working directory, or otherwise report a clear configuration error instead of leaking NullPointerException.

morphqdd:
StrictXmir.copied() called path.toFile().getParentFile().mkdirs() on the raw (possibly relative) path. When the cache directory resolves to an empty relative path (Path.of("")), the file it produces has no parent component, and getParentFile() returns null before the schema is ever copied. The sibling downloaded() method already avoids this by resolving to an absolute file first, so copied() now does the same.

Closes #6443

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
+++ b/eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
@@ -214,11 +214,11 @@ public final class StrictXmir implements XML {
      * @return Where it was saved
      */
     private static File copied(final String uri, final Path path) {
-        final File file = path.toFile();
+        final File abs = path.toFile().getAbsoluteFile();
         StrictXmir.LOCK.lock();
         try {
-            if (!file.exists()) {
-                if (file.getParentFile().mkdirs()) {
+            if (!abs.exists()) {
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3609](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6539, opened 2026-08-09
#6497: fix PhPackage.hasRho() to check the objects map instead of always returning true

### LINKED ISSUE #6497: `PhPackage.hasRho()` always returns true, contradicting `take(RHO)` and its own test
`PhPackage.hasRho()` (`eo-runtime/src/main/java/org/eolang/PhPackage.java:58-60`) is hardcoded to `return true`, regardless of whether ρ (`Phi.RHO`) was ever bound in the package's internal `objects` map. Every other `Phi` implementation (e.g. `PhDefault`, around `PhDefault.java:189`) computes this by actually checking whether ρ is set. `PhPackage.take(name)` itself, a few lines below `hasRho()`, correctly checks `this.objects.containsKey(Phi.RHO)` and throws `ExUnset` when it's absent — proving the class already has the right logic, just not wired into `hasRho()`.

This is wrong for the global root package `Phi.Φ` (`Phi.java:42`, `new PhPackage(PhPackage.GLOBAL)`), which never gets its own ρ bound. The class's own test, `PhPackageTest.doesNotSetRhoToGlobalObject()` (`eo-runtime/src/test/java/org/eolang/PhPackageTest.java:46-55`), asserts that `Phi.Φ.take(Phi.RHO)` throws `ExUnset` with "Global object 'Φ' must not have ρ attribute." Yet `Phi.Φ.hasRho()` reports `true` for that very same object — a direct contradiction, confirmed by compiling and running against `eo-runtime/target/classes`:

```
hasRho: true
take(rho) threw ExUnset: The ρ attribute is absent in package object 'Φ'
```

The impact is real: `AtWithRho#get()` (`eo-runtime/src/main/java/org/eolang/AtWithRho.java:53-59`) relies on the documented contract ("Returns true if object has bound rho attribute", `Phi.java:50-54`) to decide whether to copy an object and bind ρ to it. For any `Phi` that resolves to a `PhPackage`, this check always says "already bound" and skips binding, even when ρ is genuinely unset — silently violating the one invariant `AtWithRho` depends on.

Suggested fix: change `hasRho()` to `return this.objects.containsKey(Phi.RHO);`, mirroring the check `take()` already performs a few lines later in the same file.

morphqdd:
PhPackage.hasRho() was hardcoded to return true regardless of whether rho was ever bound in the internal objects map. take(name) a few lines below already checks this.objects.containsKey(Phi.RHO) correctly and throws ExUnset when it's absent, so the right logic already existed, just not wired into hasRho().

This contradicted the class's own test: PhPackageTest.doesNotSetRhoToGlobalObject() asserts that Phi.Φ.take(Phi.RHO) throws ExUnset, yet Phi.Φ.hasRho() reported true for the same object. AtWithRho#get() relies on hasRho() to decide whether to copy an object and bind rho to it, so for any Phi resolving to a PhPackage this always reported "already bound" even when unset.

Fixed hasRho() to mirror take()'s check. Added two tests: hasRho() reports false on the unbound global object, and true once dispatch has bound it.

Closes #6497

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhPackage.java
+++ b/eo-runtime/src/main/java/org/eolang/PhPackage.java
@@ -56,7 +56,7 @@ final class PhPackage implements Phi {
 
     @Override
     public boolean hasRho() {
-        return true;
+        return this.objects.containsKey(Phi.RHO);
     }
 
     @Override
--- a/eo-runtime/src/test/java/org/eolang/PhPackageTest.java
+++ b/eo-runtime/src/test/java/org/eolang/PhPackageTest.java
@@ -68,6 +68,30 @@ final class PhPackageTest {
     }
 
     @Test
+    void hasRhoReportsAbsenceOnGlobalObject() {
+        MatcherAssert.assertThat(
+            String.format(
+                "hasRho() must agree with take(%s) and report the global object as unset",
+                Phi.RHO
+            ),
+            Phi.Φ.hasRho(),
+            Matchers.is(false)
+        );
+    }
+
+    @Test
+    void hasRhoReportsPresenceAfterDispatch() {
+        MatcherAssert.assertThat(
+            String.format(
+                "hasRho() must report %s as set once dispatch has bound it",
+                Phi.RHO
+            ),
+            Phi.Φ.take("nop").hasRho(),
+            Matchers.is(true)
+        );
+    }
+
+    @Test
     void findsLongClass() {
         MatcherAssert.assertThat(
             "Package should resolve class with '$' in the name, but it didn't",


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3621](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6540, opened 2026-08-09
#6476: retry sendsAndReceivesMessageViaSocketObject on the accept-timing flake

### LINKED ISSUE #6476: Socket round-trip test flakes with `Connection refused` on `macos-15-intel` despite #6458
`SyscallTest.sendsAndReceivesMessageViaSocketObject` intermittently errors with `Connection refused` on `macos-15-intel`, even after the ephpo port reservation from #6458.

## What happens

On `mvn (macos-15-intel, 26)` the socket round-trip test errors after ~113 s:

```
[ERROR] org.eolang.SyscallTest.sendsAndReceivesMessageViaSocketObject(int) -- Time elapsed: 113.5 s <<< ERROR!
org.eolang.ExFailure: Error in "Φ.socket.a…55-4.connect.Δ" at .socket:72:6; … ;
  Couldn't connect to 127.0.0.1:23059 on socket #78 because Connection refused
	at org.eolang.PhSafe.through(PhSafe.java:213)
```

The client side of the test cannot connect to the loopback port the server side is supposed to be listening on — `Connection refused`, not a timeout on our side — so the whole dataization chain terminates.

## Why this looks like an infrastructure flake, not a PR defect

It fails on two unrelated open PRs whose changes have nothing to do with sockets, and only on `macos-15-intel` — every other `mvn` platform (`ubuntu-24.04`, `windows-2022`, `macos-15`) is green on the same commits:

- #6402 (a `PhApplication` φ-term change) — https://github.com/objectionary/eo/actions/runs/31218005046/job/92995924809
- #6330 (a parser `Suffix` change) — https://github.com/objectionary/eo/actions/runs/31202001723/job/92943978796

`master` itself is green on recent runs, so the failure is intermittent. Because it turns unrelated PRs red on a required check, it blocks merges without any real fault in the changes under review.

## Relationship to #6457 / #6458

#6457 (closed by #6458, "reserve socket-test ports with ephpo") targeted the *port-picking* race in `Port.number()` — `ServerSocket(0)`, read, close, then bind later. Both PRs above already contain #6458 (their base is even with `master`), yet this test still fails with `Connection refused`. So the residual symptom is not the bind-and-close race #6458 addressed: the port is reserved, but the **client connects before the server side is actually accepting** (or the loopback connect is refused for another `macos-15-intel`-specific reason). The retry loops #6458 added guard the *bind*, not this *connect*.

## Suggested next step

Reproduce on a `macos-15-intel` runner and confirm whether the server object has reached its `accept` state before the client's `connect` runs; if there is a start/accept ordering window, gate the client `connect` on the server being ready (or retry `connect` on `Connection refused`, the mirror of the bind retry) rather than relying on timing. Distinct from #6407, which is a 10-minute *timeout* of `EOdirectoryTest`/`EOdeletedTest` on the same runner.

morphqdd:
sendsAndReceivesMessageViaSocketObject synchronizes the client connect against the server's listen with a fixed Thread.sleep(2000), guessing that two seconds is enough for the server thread to reach the actual OS-level listening state. On a loaded macos-15-intel runner that guess is sometimes wrong, and the client's connect fails with Connection refused, unrelated to the port-reservation race #6458 already fixed.

Every other socket test in this file that depends on native timing (WindowsSocketTest, PosixSocketTest) already uses @RepeatedIfExceptionsTest(repeats = 3) instead of a plain @Test for exactly this reason. This test was the one exception, still on a plain @Test. Switched it to the same idiom already established in the file, rather than inventing a new retry mechanism.

Closes #6476

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/SyscallTest.java
+++ b/eo-runtime/src/test/java/org/eolang/SyscallTest.java
@@ -113,7 +113,7 @@ final class SyscallTest {
         }
     }
 
-    @Test
+    @RepeatedIfExceptionsTest(repeats = 3)
     void sendsAndReceivesMessageViaSocketObject(@Ephemeral final int port)
         throws InterruptedException {
         final String msg = "Hello, Socket!";


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3633](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6542, opened 2026-08-09
fix(#6440): dedup coverage hits per destination

### LINKED ISSUE #6440: `PhCoverage suppresses write and doesn't create a new file when eo.coverageFile property changes`

VasilevNStas:
Fixes #6440.

## Problem

`PhCoverage` documents that it reads `eo.coverageFile` on every touch, but deduplicates coverage records solely by the `loc:line:pos` record in the process-wide `SEEN` set. Once a hit is written to file A, changing the property to file B and touching the same location suppresses the write — file B stays absent even though `append()` would have read the new destination.

## Solution

Include the destination in the deduplication key (`property + NUL + record`), so the same record is written once per configured coverage destination, matching the documented per-touch read of the property.

## Changes

- `eo-runtime/src/main/java/org/eolang/PhCoverage.java` — `hit()` keys `SEEN` by destination + record; javadoc updated.
- `eo-runtime/src/test/java/org/eolang/PhCoverageTest.java` — regression test.
- `.github/workflows/counts.yml` — smells baseline 847 → 848 (the new test adds one `null`).

## Tests

- `PhCoverageTest.writesToEachConfiguredDestinationOnce` — touches the same instance against two destinations and asserts both receive the record (verified to fail on pre-fix code: the second destination is never created).
- Existing `recordsEveryLocationExactlyOnce`, `convertsInvalidCoveragePathIntoIllegalArgumentException`, `namesThePropertyAndItsValueInAnInvalidCoveragePathFailure` still pass.

@yegor256 please take a look.

### DIFF
--- a/.github/workflows/counts.yml
+++ b/.github/workflows/counts.yml
@@ -37,7 +37,7 @@ jobs:
           - metric: nulls
             files: '*.java'
             pattern: 'null'
-            count: 251
+            count: 252
           - metric: atoms
             files: '*.java'
             pattern: '@XmirObject'
--- a/eo-runtime/src/main/java/org/eolang/PhCoverage.java
+++ b/eo-runtime/src/main/java/org/eolang/PhCoverage.java
@@ -22,7 +22,8 @@ import java.util.concurrent.ConcurrentHashMap;
  * <p>The transpiler emits it around every located object. Recording is
  * enabled by the {@code eo.coverageFile} system property: on the first
  * touch, one {@code loc:line:pos} line is appended, at most once per
- * JVM; without the property every hit is a silent no-op. The property
+ * JVM per configured destination; without the property every hit is a
+ * silent no-op. The property
  * is re-read on every touch (not cached at class load), since this
  * class is now instantiated around every located object in every EO
  * program: the very first one touched anywhere in the JVM would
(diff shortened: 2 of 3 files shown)


yegor256:
@VasilevNStas conflicts here

VasilevNStas:
@yegor256 plz re-review this PR

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+40](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6543, opened 2026-08-09
#6228: stop double-escaping the <listing> element, drop XML-forbidden characters instead

### LINKED ISSUE #6228: `<listing>` in every XMIR holds double-escaped source instead of verbatim source

morphqdd:
Listing.java called StringEscapeUtils.escapeXml11 on the source before setting it on the listing directive. The XML writer then escapes the already-escaped text a second time, so /object/listing held double-escaped text (e.g. > became &amp;gt; instead of &gt;) instead of the documented verbatim source.

Fixed by setting the raw source directly and letting the XML writer escape it exactly once. The only characters actually dropped are the ones XML text nodes cannot represent at all (C0/C1 control characters and a couple of noncharacters), since those cannot round-trip through XML regardless of escaping.

Updated the two existing tests that baked in the double-escaped expectation, added a case with the literal <, &, and characters from the reported repro, and added ListingTest.java covering the character-dropping behavior directly.

Closes #6228

### DIFF
--- a/.github/workflows/counts.yml
+++ b/.github/workflows/counts.yml
@@ -24,7 +24,7 @@ jobs:
           - metric: smells
             files: '*.java'
             pattern: 'static|null'
-            count: 847
+            count: 848
           - metric: atoms
             files: '*.java'
             pattern: '@XmirObject'
--- a/.github/workflows/counts.yml
+++ b/.github/workflows/counts.yml
@@ -24,7 +24,7 @@ jobs:
           - metric: smells
             files: '*.java'
             pattern: 'static|null'
-            count: 848
+            count: 850
           - metric: atoms
             files: '*.java'
             pattern: '@XmirObject'
--- a/eo-parser/src/main/java/org/eolang/parser/Listing.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Listing.java
@@ -5,7 +5,7 @@
 package org.eolang.parser;
 
 import java.util.Iterator;
-import org.apache.commons.text.StringEscapeUtils;
+import java.util.regex.Pattern;
 import org.xembly.Directive;
 import org.xembly.Directives;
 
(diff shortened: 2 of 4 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (184 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3645](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6447, opened 2026-08-06
`MjAtomsTable.write()` throws `NullPointerException` when `eo.atomsOutput` is a filename without a parent directory


gemshrine:
## What happens

`MjAtomsTable` accepts `eo.atomsOutput` as a Maven file parameter, but fails with a raw `NullPointerException` when it is configured as a filename in the current project directory.

The failure occurs in [`MjAtomsTable.write()`](https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/java/org/eolang/maven/MjAtomsTable.java#L126-L128):

```java
final Path target = this.atomsOutput.toPath();
Files.createDirectories(target.getParent());
```

For `Path.of("atoms.csv")`, `target.getParent()` is `null`, so `Files.createDirectories()` throws before the table is written.

## Steps to reproduce

1. Configure the `atoms-table` goal with a parentless relative path:

```xml
<plugin>
  <groupId>org.eolang</groupId>
  <artifactId>eo-maven-plugin</artifactId>
  <configuration>
    <atomsOutput>atoms.csv</atomsOutput>
  </configuration>
</plugin>
```

2. Run the `atoms-table` goal in a project with either XMIR sources or an empty `atomsInputDir`.

3. Observe `NullPointerException` from `Files.createDirectories(target.getParent())`.

## Expected behavior

A filename such as `atoms.csv` should be written into the project working directory. The goal should create parent directories only when the configured path actually has a parent, consistent with ordinary Maven file parameters.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2095](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

morphqdd:
@yegor256, PR #6690 fixing this was merged but I forgot the `Closes` keyword, so it didn't auto-close. Could you close this one manually? Sorry for the extra step.

yegor256:
[...]

PR #6546, opened 2026-08-10
#6545: resolve xslMeasures to an absolute path before taking parent

### LINKED ISSUE #6545: Transpilation.measured() throws NullPointerException for a relative eo.xslMeasuresFile with no parent directory
## What Happens

`Transpilation.measured()` calls `this.xslMeasures.getParent()` and immediately dereferences the result without checking for null:

```java
private Train<Shift> measured(final Train<Shift> base) {
    final Path parent = this.xslMeasures.getParent();
    if (parent.toFile().mkdirs()) {
        Logger.debug(this, "Directory created for %[file]s", this.xslMeasures);
    }
```

`this.xslMeasures` comes from `MjSafe.xslMeasures`, a user-configurable Maven parameter:

```java
@Parameter(
    property = "eo.xslMeasuresFile",
    required = true,
    defaultValue = "${project.build.directory}/eo/xsl-measures.json"
)
protected File xslMeasures;
```

The default value has a parent directory, so the bug is hidden in normal use. But a relative, single-segment override has no parent: `Paths.get("measures.json").getParent()` returns `null`, and the next line throws a raw `NullPointerException`.

## Steps to Reproduce

Run the `transpile` goal with the property overridden to a bare filename:

```
mvn eo:transpile -Deo.xslMeasuresFile=measures.json
```

`measured()` is called from `forSource()`, which runs once per XMIR file being transpiled, so this fails on the very first source file with no diagnostic message identifying the cause.

## What Should Happen

`measured()` should resolve `xslMeasures` to an absolute path before taking its parent (e.g. `this.xslMeasures.toAbsolutePath().getParent()`, which is always non-null), the same fix pattern already applied to the identical bug in `Saved` (#6441) and `MjAtomsTable` (#6447).

HDPark95:
`Transpilation.measured()` took `this.xslMeasures.getParent()` and dereferenced the result on the next line. `xslMeasures` comes from the `eo.xslMeasuresFile` Maven parameter, whose default value has a parent directory, so the problem stays hidden in normal use. A relative, single-segment override has no parent, though:

```
mvn eo:transpile -Deo.xslMeasuresFile=measures.json
```

`Paths.get("measures.json").getParent()` returns nothing, so the following `parent.toFile().mkdirs()` throws a raw `NullPointerException` on the first transpiled source, with no message identifying the cause.

## Fix

Resolve `xslMeasures` to an absolute path before taking its parent (`toAbsolutePath().getParent()`), which is always present. This mirrors the fix already applied to the same class of bug in `Saved` (#6441) and `MjAtomsTable` (#6447).

## Verification

- Added `TranspilationTest.buildsSourceFunctionForParentlessMeasuresPath`, which builds the source function for a parentless measures path. It fails on `master` with a `NullPointerException` at `Transpilation.forSource` and passes with this change.
- `mvn -pl eo-maven-plugin clean verify -DskipTests -Pqulice` passes locally on JDK 21.

Closes #6545

@yegor256

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Transpilation.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Transpilation.java
@@ -195,7 +195,7 @@ final class Transpilation {
      * @return Measured train
      */
     private Train<Shift> measured(final Train<Shift> base) {
-        final Path parent = this.xslMeasures.getParent();
+        final Path parent = this.xslMeasures.toAbsolutePath().getParent();
         if (parent.toFile().mkdirs()) {
             Logger.debug(this, "Directory created for %[file]s", this.xslMeasures);
         }
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/TranspilationTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/TranspilationTest.java
@@ -1,0 +1,31 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import java.nio.file.Paths;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link Transpilation}.
+ * @since 0.74
+ */
+final class TranspilationTest {
+
+    @Test
+    void buildsSourceFunctionForParentlessMeasuresPath() {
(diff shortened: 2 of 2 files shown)


0crat:
@hdpark95 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (33) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+28](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6551, opened 2026-08-10
File a pending check for every argument applied

### LINKED ISSUE #6548: eo-inference module not filing essential pending checks for applications
Three of the four rules of `inference.md` are in place. What every object
provides, what it is asked for, and which types are copies of which are all
written down; the fourth rule is not. An application files a pending check —
"this argument must fit into that void of the object being copied" — and
without those the checker has nothing to drain.

In the note's program

```eo
[] > app
  inc t > @
  [x] > inc
    x.next.foo > @
```

the application `inc t` says that `t` must fit into the first void of the
copy of `inc`, which is `x`. Follow that through the links and the two other
tables finally have something to disagree about: `x` is asked for `next`, and
whatever `t` provides has to answer it.

The XMIR of an application is an object whose children carry `as`:

```xml
<o base="ξ.inc" loc="Φ.app.φ">
  <o as="α0" base="ξ.t" loc="Φ.app.φ.α0"/>
</o>
```

So a fourth clue alongside the three in `Clues`
(`eo-inference/src/main/java/org/eolang/inference/Clues.java:65`) would
write `checks.xml`, one row per bound argument: the owner being the locator of
the object receiving it, the name being the binding as written (`α0` today —
all 21,094 applications in the runtime are positional), and the type being
the locator of the argument. That is the row shape `Grouped` already renders.

Deliberately, the clue resolves nothing. Which void `α0` lands in depends on
the object being copied and on how many voids are already bound, which is a
question for the checking loop, with the links and provides tables in hand.
Recording a check and deciding it are separate steps, and only the first one
belongs here.

maxonfjvipon:
Rule four of `inference.md`, the last of them: an application files a pending
check. With this the four tables of the note all exist.

`Checks` writes one row per bound argument — the owner being the object that
receives it, the name being the binding as written, the type being the
argument itself:

```xml
<checks>
  <type id="Φ.app.φ">
    <attr name="α0" type="Φ.app.φ.α0"/>
  </type>
</checks>
```

That row is what makes the other three tables about one question. `Needs`
knows the void `x` is asked for `next`, `Provides` knows what `t` has, `Links`
knows which is a copy of which — and this says that `t` is what went into
`x`.

Nothing is resolved here on purpose. Which void `α0` lands in depends on the
object being copied and on how many of its voids are already bound; both
answers live in other tables, so the row keeps the binding as written and
leaves the arithmetic to the checking loop. A check recorded without being
decided is the point: the facts needed to decide it may arrive later, or
never.

A dispatch counts as an application when something is applied to it, so
`x.plus 5` gives a row to two tables — `plus` is asked of `x`, and the `5` has
to fit whatever `x.plus` is. On the runtime that comes to 21,094 checks beside
2,468 types, 9,950 needs and 20,825 links, all of them positional bindings.

Now rebased on master, so the seven packs here are written the way #6553 made
possible: a program of several files, with the application in one file and the
formation it copies in another. That is the arrangement worth testing, because
a check is only worth something next to rows the other files produced —
`application-files-a-check` puts `app`, `inc` and `t` in three files and then
asks the check, the two links and the void of `inc` about the same argument.
The note's own pack grew a `checks` block too, so one file shows all four
tables for the same six lines of EO.

Two of the packs record things I did not expect and would rather have written
down than smoothed over:

  * `every-argument-files-its-own-check` applies the same `t` twice. That is
    two rows with two types, since a type here is a place in the code and not
    a name.

  * `data-literal-files-a-check` — a number is an application. `5` is bytes
    put into a copy of `number`, three layers of it, so a program with a
    number in it is never short of checks. The pack also shows what happens
    when the program says nothing about `number`: the check is filed, no link
    is written for that copy, and the loop will have to keep quiet about it
    rather than guess. Which is the shape of the whole prototype — a row is a
    question, and a question nobody can answer yet is left alone.

While adding them I found a hole in the pack runner: its list of keys was
`xmir`, `provides`, `needs` and `links`, and any other key was **silently
ignored**. So the first version of these packs passed while asking about a
table that did not exist yet, and a typo like `provide:` would have switched a
pack's assertions off without a word. The runner now reports an unknown key as
a failure:

```
Expected: an empty collection
     but: <[unknown key: provide]>
```

---

One more thing came out of this, and it is not about rule four. Adding the
fourth clue made the `inference` step on the runtime jump from 1m27s to 5m40s,
which was worth understanding rather than accepting. The rules were not the
problem — the store was. `ToMono.set()` reads the whole table and searches it
for the row before writing one cell, so a row of four cells was six passes over
everything, and a fourth clue that doubles the rows quadruples that. Splitting
one `set()` into its parts at 8,000 rows put 722ms on the id search against
about 200ms for both whole-table copies, so it was algorithmic, not a constant.

That is fixed in tojos rather than worked around here: `TjDeferred`
(objectionary/eo is now on **0.20.0**) keeps the rows in memory keyed by id, so
a row is found by a map lookup, a cell written touches that row alone, and the
mono is written once at `close()`. Building 42,000 rows of four cells went from
about 40 seconds to 16 milliseconds, and this step from **2m59s to 19s** — four
clues now cost less than three did.

Two things followed from it here. The `index` cell is gone, with both sorts and
both `Integer.parseInt` calls: it existed only because a table promised nothing
about the order of its rows, and `TjDeferred` hands them back in the order they
arrived. And `GroupedTest` grew a case with two attributes, since that order was
never actually tested — every case had one attribute, and `Stream.sorted` does
not call a comparator on one element.

The order is the one thing to keep an eye on: `Tojos.select()` does not promise
it, so yegor256/tojos#247 asks for it to be stated and pinned upstream. Until
it is, `keepsAttributesInOrderTheyWereWritten` is our own guard — if that order
ever changes, a test fails here loudly rather than the tables reordering
quietly, which would matter, since the position of a row is what says which
argument fills which void.

The four tables the runtime produces are **byte-identical** through all of
this, which is how I know none of it changed what the clues understand.

Closes: #6548


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (518 >= 100); -8 for way too many hits-of-code (518 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+689](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6429, opened 2026-08-06
Unused `Expect.I16/I32/I64` code in `Expect.java` remains after refactor


yegor256:
`Expect.I16`, `Expect.I32`, and `Expect.I64` in `eo-runtime/src/main/java/org/eolang/Expect.java:273-363` are dead code: nothing in the repository calls them (`grep -rn "Expect\.I16\|Expect\.I32\|Expect\.I64" --include=*.java .` outside `Expect.java` itself returns zero hits), and no test covers them.

They became orphaned across three commits. #3461 added `Expect.I16`/`Expect.I32` (mirroring the existing `Expect.I64`) and wired all three into `EOi16$EOas_i32.java`, `EOi32$EOas_i64.java`, and `EOi64$EOas_number.java`, plus a dedicated `EOiConvertersExpectTest.java`. #5460 then rewrote `i16.as-i32` and `i32.as-i64` in pure EO, deleting the first two Java atom classes. #5471 rewrote `i64.as-number` in pure EO too, deleting the third Java atom class and the whole `EOiConvertersExpectTest.java` (nothing left for it to test). Each rewrite cleaned up its own atom but left the validation helper behind in `Expect.java`.

Fix: delete the `I16`, `I32`, and `I64` inner classes (and their Javadoc) from `Expect.java`.

PR #6554, opened 2026-08-10
fix(#6429): remove dead code I16/I32/I64 from Expect.java

RayderXEE:
Nothing outside Expect.java calls I16, I32, or I64. They became orphaned across three commits: #3461 added them, #5460 rewrote i16.as-i32 and i32.as-i64 in pure EO deleting the atoms that used I16/I32, and #5471 rewrote i64.as-number in pure EO deleting the atom that used I64 and the test that covered all three. Each rewrite cleaned up its own atom but left the validation helper behind in Expect.java.


0crat:
@rayderxee Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6555, opened 2026-08-10
Let a pack describe a program of several files

### LINKED ISSUE #6553: Inference packs limited to single-file programs, breaking interactions across files
A pack of `inference-packs` carries one EO program and the runner writes it
to one file:

```java
Files.writeString(
    Files.createDirectories(this.dir.resolve("parsed")).resolve("main.xmir"),
    new EoSyntax(pack.map().get("eo").toString()).parsed().toString()
);
```

(`eo-maven-plugin/src/test/java/org/eolang/maven/InferringTest.java:50`)

So every pack describes a one-file program, and the most important property
of the tables cannot be expressed at all. The clues read a *world*: one table
covers the whole program, because a locator names one object of one file and
no other, and an object in one file is used by objects in others. On the
runtime that is 170 files feeding four tables. In the packs it is always one.

What is missing shows up immediately in what cannot be written today. A
reference to a root object defined in another file, a dispatch whose receiver
was built elsewhere, two files whose formations must land in the same table —
all of these are the normal case for real programs and none of them has a
pack. `reference-to-root-object.yaml` had to put both objects in one file to
say anything at all, which is not how a program is laid out.

The `eo` key should therefore name its files instead of being a bare
program:

```yaml
eo:
  app.eo: |
    [] > app
      number > @
  number.eo: |
    [] > number
links:
  - "/links/type[@id='Φ.app.φ' and @copy='Φ.number']"
```

The table keys need no change, since the tables were program-wide from the
start; only the input side is too narrow. The `xmir` key, which asks about the
XMIR prepared for the rules, is per file and should name the file it means.

maxonfjvipon:
A pack described a one-file program, because the runner wrote whatever `eo`
held into a single `main.xmir`. That hid the property the tables are built
around: the clues read a *world*, one table covers the whole program, and an
object in one file is used by objects in others. On the runtime that is 170
files feeding four tables; in the packs it was always one.

The `eo` key now names its files:

```yaml
eo:
  two.eo: |
    [] > two
      number > @
  number.eo: |
    [] > number
      [] > plus
links:
  - "/links/type[@id='Φ.two.φ' and @copy='Φ.number']"
```

The table keys are untouched — they were program-wide from the start, so
`provides`, `needs` and `links` say exactly what they said before. Only `xmir`
had to grow a name, since the XMIR prepared for the rules is per file:

```yaml
xmir:
  box.eo:
    - "//o[@base='ξ.x' and @loc='Φ.box.φ.ρ.ρ']"
```

All thirteen existing packs are converted to the named form; there is one way
to write a pack, not two.

Three packs are new, and each says something no pack could say before: two
files landing in one table, a reference reaching into another file, and a
dispatch whose receiver was built in another file — where all three tables
meet across a file boundary. `reference-to-root-object.yaml` keeps its
single-file program on purpose: a sibling object in the same file is a
different case from one next door.

Two guards worth naming. A pack that names a file the program does not have
fails rather than passing quietly:

```
Expected: an empty collection
     but: <[unknown file: nope.eo]>
```

And I mutated a cross-file assertion to confirm the new packs actually bite
before trusting them green.

Closes: #6553

### DIFF
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/InferringTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/InferringTest.java
@@ -15,7 +15,9 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import org.eolang.jucs.ClasspathSource;
 import org.eolang.parser.EoSyntax;
 import org.eolang.xax.XtSticky;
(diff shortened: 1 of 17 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (277 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+700](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6556, opened 2026-08-10
#6211: keep the Q. root when an alias claims a base's first segment

### LINKED ISSUE #6211: Printer silently rebinds a reference when `+alias` claims the first segment of an FQN
## What Happens

The printer can silently rebind a reference to a different object when a file declares an `+alias` whose short name matches the first segment of some other fully-qualified name.

`to-eo-tree.xsl`'s `eo:surface` function (line 33) strips the `Φ.` root from any base unconditionally when printing it in short form, for example `Φ.string.sprintf` becomes `string.sprintf`. The one place that keeps the explicit `Q.` root (line 276) only checks whether an in-scope attribute shadows the root segment. It never checks whether an `+alias` in the same file declares that segment as its own short name.

When the file is re-parsed, the parser resolves the leading bare segment through the alias table before rejoining the rest of the path, so the reference ends up pointing at a different object than the one it started as.

Repro:

```
+package sandbox
+alias com.example.string
+alias string.sprintf

[] > app
  string > a
  string > b
  sprintf > c
```

Parses with `c` having `base="Φ.string.sprintf"`. Printing this XMIR back to EO gives:

```
+package sandbox
+alias com.example.string

[] > app
  string > a
  string > b
  string.sprintf > c
```

Re-parsing this output gives `c` a `base="Φ.com.example.string.sprintf"`, a different object than the original `Φ.string.sprintf`. Printing the result a second time gives yet another different string, so the printer is also not idempotent on this input.

The same corruption happens with an explicit `Q.` in the source instead of a bare reference, and with other alias/name combinations, not just this specific pair.

This is not a brand new bug: the same rebinding happens on `master` before today's `restore-aliases.xsl` alias-restoration pass was added (confirmed by testing against the printer as it existed one commit before that pass landed). The new `restore-aliases.xsl` has the identical blind spot in its own `eo:free` guard, which also only rejects a short name that is an exact bound name or an exact bare global, never the first segment of some longer qualified reference.

This matters in practice because `MjFormat`'s `canonical` step runs `Xmir.toEO()` and, when `-Deo.autoFix` is set, writes the result back over the user's real source file.

## What Should Happen

A base of the form `Φ.<x>.<rest>` should keep its explicit `Q.` root whenever `<x>` is declared as an alias short name (`/object/metas/meta[head='alias']/part[1]`) anywhere in the same file, the same way it already keeps the root when an in-scope attribute shadows `<x>`. `restore-aliases.xsl`'s `eo:free` function should apply the same check before deciding a short name is safe to introduce.

morphqdd:
Fixes #6211.

`to-eo-tree.xsl` kept the explicit `Q.` root only when an in-scope attribute shadowed a base's first segment, never when a `+alias` in the same file declared that segment as its own short name. `restore-aliases.xsl`'s `eo:free` guard had the identical blind spot: it only rejected a short name that was an exact bound name or an exact bare global, never the first segment of some other, longer reference.

Both are fixed by checking `/object/metas/meta[head='alias']/part[1]` for a collision before dropping the `Q.` root or restoring a short alias name.

Added a print-pack YAML test reproducing the exact case from the issue (`+alias com.example.string` and `+alias string.sprintf` sharing the `string` segment).

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/restore-aliases.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/restore-aliases.xsl
@@ -47,7 +47,8 @@
   -->
   <xsl:function name="eo:free" as="xs:boolean">
     <xsl:param name="name" as="xs:string"/>
-    <xsl:sequence select="not($name = $bound) and not(some $r in $refs satisfies $r = concat($eo:program, '.', $name) or starts-with($r, concat($eo:program, '.', $name, '.')))"/>
+    <xsl:variable name="prefixed" select="concat($eo:program, '.', $name)"/>
+    <xsl:sequence select="not($name = $bound) and not(some $r in $refs satisfies $r = $prefixed or (starts-with($r, $prefixed) and substring($r, string-length($prefixed) + 1, 1) = '.'))"/>
   </xsl:function>
   <!-- The fully-qualified names worth restoring: a free-named alias referenced more than once. -->
   <xsl:variable name="kept" as="xs:string*" select="for $m in $aliases return if (eo:free($m/part[1]) and count($refs[. = $m/part[last()]]) gt 1) then string($m/part[last()]) else ()"/>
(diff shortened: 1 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3768](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6557, opened 2026-08-10
#6544: leave odd-length hex payloads untouched in StUnhex

### LINKED ISSUE #6544: `StUnhex.buffer()` method results in StringIndexOutOfBoundsException with odd-length hex payload
## What Happens

`StUnhex.buffer()` throws an uncaught `StringIndexOutOfBoundsException` when the hex payload has an odd number of digits.

```java
private static ByteBuffer buffer(final String txt) {
    final int len = txt.length();
    final ByteBuffer buffer = ByteBuffer.allocate(len / 2);
    for (int idx = 0; idx < len; idx += 2) {
        buffer.put((byte) Integer.parseInt(txt.substring(idx, idx + 2), 16));
    }
    buffer.position(0);
    return buffer;
}
```

The loop condition is `idx < len`, but each iteration consumes two characters via `substring(idx, idx + 2)`. When `len` is odd, the buffer is sized correctly to `len / 2` (floor division), but the loop still runs one more time at `idx == len - 1`, calling `substring(len - 1, len + 1)`, which exceeds the string length.

## Steps to Reproduce

Parse and print XMIR containing a `Φ.number` (or `Φ.string`) object whose `Φ.bytes` payload has an odd count of hex digits after dashes are stripped, for example:

```
number 40-49-0F-D > @
```

`StUnhex.undash` turns this into `"40490FD"` (7 characters, odd). `buffer()` allocates a 3-byte buffer, parses `"40"`, `"49"`, `"0F"` successfully, then at `idx = 6` calls `substring(6, 8)` on a 7-character string, throwing `StringIndexOutOfBoundsException`. This propagates out of `Xmir.toEO()` unhandled.

This same odd-length case is not guarded in the `string` branch either (`StUnhex.java` lines 55-66), which calls `buffer(...).array()` directly with no length check at all.

## What Should Happen

`buffer()` (or its callers) should check for an odd-length payload up front and treat it the same way the `number` branch already treats a wrong byte count elsewhere (issue #6038's fix): leave the `Φ.number`/`Φ.string`/`Φ.bytes` node untouched rather than letting `substring` throw.

zhaoxinyi02:
## Root cause

`StUnhex.buffer(String)` parses a hex payload two characters at a time with `txt.substring(idx, idx + 2)`, but its loop runs while `idx < len`. For a payload with an odd number of hex digits (e.g. `40-49-0F-D`, which `undash` turns into the 7-char `"40490FD"`), the buffer is sized to `len / 2 = 3` bytes correctly, yet the loop still enters one final iteration at `idx == len - 1` and calls `substring(6, 8)` on a 7-character string, throwing `StringIndexOutOfBoundsException`. The exception propagates out of `Xmir.toEO()` unhandled.

Both callers were affected:
- the `number` branch already guarded `buffer.remaining() == Double.BYTES`, but that check runs *after* `buffer()` and never gets a chance on odd input;
- the `string` branch called `buffer(...).array()` with no length check at all.

## Fix

Make `StUnhex.buffer(...)` return `Optional<ByteBuffer>`, empty when the payload has an odd length (a stray trailing nibble cannot form a complete byte). Both branches now consume it with `ifPresent`, so an odd-length payload leaves the `Φ.number`/`Φ.string`/`Φ.bytes` node untouched instead of crashing.

This is the same strategy the `number` branch already uses for a wrong byte count (the fix for #6038 in #6052), and the same strategy the `string` branch already uses for invalid UTF-8, where `decode()` returns `Optional.empty()`.

## Tests

Added two parameterized regression tests to `StUnhexTest`, mirroring the existing `keepsTooFewNumberBytesUnconverted` / `keepsMalformedBytesUnconverted` style:
- `keepsOddLengthNumberPayloadUnconverted` — `Φ.number` with `40-49-0F-D` (7 hex digits);
- `keepsOddLengthStringPayloadUnconverted` — `Φ.string` with `41-42-0` (5 hex digits).

Both crash with `StringIndexOutOfBoundsException` on the unpatched `buffer()` and pass after the fix, asserting the node is left untouched (`not(o)` with the original dashed `text()` preserved).

`mvn -pl eo-printer test` — 340 tests pass. `mvn -pl eo-printer -Pqulice verify -DskipTests` — clean (checkstyle, PMD, jtcop, license headers).

Fixes #6544

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
+++ b/eo-printer/src/main/java/org/eolang/printer/StUnhex.java
@@ -40,27 +40,28 @@ final class StUnhex extends StEnvelope {
         ),
         new StXnav(
             StUnhex.elements("number"),
-            xnav -> {
-                final ByteBuffer buffer = StUnhex.buffer(
-                    StUnhex.undash(xnav.element("o").text().orElse(""))
-                );
-                if (buffer.remaining() == Double.BYTES) {
-                    final double number = buffer.getDouble();
-                    if (!Double.isNaN(number) && !Double.isInfinite(number)) {
-                        new Payload(xnav).replace(StUnhex.number(number));
+            xnav -> StUnhex.buffer(
+                StUnhex.undash(xnav.element("o").text().orElse(""))
+            ).ifPresent(
+                buffer -> {
+                    if (buffer.remaining() == Double.BYTES) {
+                        final double number = buffer.getDouble();
+                        if (!Double.isNaN(number) && !Double.isInfinite(number)) {
+                            new Payload(xnav).replace(StUnhex.number(number));
+                        }
                     }
                 }
-            }
+            )
         ),
         new StXnav(
             StUnhex.elements("string"),
-            xnav -> StUnhex.decode(
-                StUnhex.buffer(
-                    StUnhex.undash(xnav.element("o").text().orElse(""))
-                ).array()
+            xnav -> StUnhex.buffer(
+                StUnhex.undash(xnav.element("o").text().orElse(""))
             ).ifPresent(
-                decoded -> new Payload(xnav).replace(
-                    String.format("\"%s\"", StUnhex.escape(decoded))
+                buffer -> StUnhex.decode(buffer.array()).ifPresent(
+                    decoded -> new Payload(xnav).replace(
+                        String.format("\"%s\"", StUnhex.escape(decoded))
+                    )
                 )
             )
         )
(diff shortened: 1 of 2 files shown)


0crat:
@zhaoxinyi02 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (100 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6558, opened 2026-08-10
fix(#6547): reject a compact-tuple count that overflows int in LnOnlyPhi

### LINKED ISSUE #6547: `LnOnlyPhi.starCount` allows overflow in only-phi compact-tuple count, unlike `LnCompactTuple.readCount`
## What happens

`LnOnlyPhi.starCount` (`eo-parser/src/main/java/org/eolang/parser/LnOnlyPhi.java:341-359`) accumulates the only-phi compact-tuple `*N` marker's digit count in a plain 32-bit `int` with no range check: `count = count * 10 + glyph - '0'`, then returns it unchecked. This is the exact same defect reported and fixed in #6355 (PR #6420), but that fix only patched the sibling class `LnCompactTuple.readCount` (`eo-parser/src/main/java/org/eolang/parser/LnCompactTuple.java:118-141`), which accumulates in a `long` and throws `ParseError("compact tuple count is too large")` once the total exceeds `Integer.MAX_VALUE`. `LnOnlyPhi` parses the identical `*N` grammar marker (R-3.9.1, cited in both classes' javadoc) for the "only-phi" line shape (`head *N > [params] > name`), and the fix from #6355 was never ported to it.

## Why it is wrong

A count above `Integer.MAX_VALUE` silently wraps instead of being rejected, and the wrapped value changes how the line is parsed — the same silent-misparse hazard #6355 already flagged as a defect, just in the second of two parallel implementations.

## Reproduction

Compiled `eo-parser` offline (`mvn -pl eo-parser -q -o compile`) and drove the compiled classes directly, the same way `LnOnlyPhiTest`/`LnCompactTupleTest` exercise these package-private `Line` classes:

- Control, bare form (already guarded): `new LnCompactTuple(new Span("seq *4294967297 > x", 1)).into(new Stack(), new Globals(), new Emit())` throws `ParseError: compact tuple count is too large`.
- Only-phi form, same count: `new LnOnlyPhi(new Span("seq *4294967297 > [m] > bar", 1)).into(stack, new Globals(), new Emit())` throws nothing. `4294967297` (2^32 + 1) wraps to `1`, so `stack.top().count() == 1`, `star() == true`, `openness() == OPEN` — accepted as if the source had said `*1`.
- A larger wrap, `seq *2147483648 > [m] > bar` (2^31, wraps to `Integer.MIN_VALUE`): also throws nothing. Because `starCount` now returns a negative value, `LnOnlyPhi.slot`'s `stars >= 0` check treats it as "not a compact-tuple marker at all," so the whole token is instead misread as a bare empty-tuple horizontal argument, discarding the digits with no error at all (`openness=HORIZONTAL_COMPLETED star()=false count()=0`, with a spurious `<o base="Φ.tuple" star=""/>` in the emitted XMIR).
- Sanity check, `seq *3 > [m] > bar` correctly yields `count()=3`.

## Suggested fix

Mirror `LnCompactTuple.readCount`'s pattern inside `LnOnlyPhi.starCount`: accumulate in a `long` and throw a `ParseError` as soon as the running total exceeds `Integer.MAX_VALUE`, instead of returning the plain `int` (whose overflow can collide with the existing `-1` "not-a-digit" sentinel).

zhaoxinyi02:
Fixes #6547.

`LnOnlyPhi.starCount` accumulated the `*N` digits of an only-phi compact-tuple LHS (`head *N > [params] > name`) into a plain 32-bit `int` with no range check, so a count above `Integer.MAX_VALUE` silently wrapped instead of being rejected - the same defect fixed for the sibling `LnCompactTuple.readCount` in #6355 (PR #6420), but never ported to the only-phi path. A positive wrap changed how the line parsed, and a negative wrap collided with the existing `-1` "not-a-digit" sentinel so the whole token was misread as a bare empty-tuple horizontal argument with no error at all.

`starCount` now accumulates into a `long` and throws a `ParseError` the moment the running total would exceed `Integer.MAX_VALUE`, pointing at the start of the digit run - mirroring `LnCompactTuple.readCount` for the sibling `*N` marker.

Tests:
- declarative parser test `only-phi-compact-tuple-count-overflow-is-rejected.yaml` with `seq *4294967297` (2^32 + 1, the smallest overflowing count) on an only-phi line, asserting parsing reports a "too large" error instead of silently wrapping;
- unit test `LnOnlyPhiTest.rejectsCompactTupleCountThatOverflowsInt` asserting the `ParseError` directly.

`mvn -pl eo-parser -Pqulice test` passes (1681 tests, 0 failures); `qulice:check` and `jtcop:check` pass.

@yegor256

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnOnlyPhi.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnOnlyPhi.java
@@ -155,7 +155,7 @@ final class LnOnlyPhi implements Line {
      * @return The φ token reader rewound to the head
      */
     private Tokens slot(final Stack stack, final Suffix suffix, final Span inner) {
-        final int stars = LnOnlyPhi.compactStar(inner.body());
+        final int stars = LnOnlyPhi.compactStar(inner.body(), inner);
         final Tokens tokens = LnOnlyPhi.reader(inner, stars);
         final Level level = this.transition(
             stack, suffix, stars >= 0 || LnOnlyPhi.bare(tokens)
@@ -294,14 +294,15 @@ final class LnOnlyPhi implements Line {
      * the LHS is not a compact-tuple head — its head must be a single
      * space-free token that is not a reversed dispatch (no trailing dot).
      * @param lhs The LHS body
+     * @param span Source span, for a ParseError on overflow
      * @return The N count, or -1 when not a compact-tuple head
      */
-    private static int compactStar(final String lhs) {
+    private static int compactStar(final String lhs, final Span span) {
         final int space = lhs.indexOf(' ');
         final int result;
         if (space > 0 && lhs.charAt(space - 1) != '.'
             && space + 1 < lhs.length() && lhs.charAt(space + 1) == '*') {
-            result = LnOnlyPhi.starCount(lhs, space + 2);
+            result = LnOnlyPhi.starCount(lhs, space + 2, span);
         } else {
             result = -1;
         }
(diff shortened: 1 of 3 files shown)


0crat:
@zhaoxinyi02 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+28](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6574, opened 2026-08-10
#6564: fix number.mod hanging on a non-finite dividend

### LINKED ISSUE #6564: `number.mod` loops indefinitely for non-finite dividends (`nan`, `pinf`, `ninf`) in `mod.eo`
## What Happens

`number.mod` (`eo-runtime/src/main/eo/number/mod.eo`, lines 27-51) never returns when the dividend (`x`) is `nan` or infinite. The only non-finite guard in the file checks the divisor (`y`), not `x`:

```
divisor.is-finite.not.if
  divisor.is-nan.if nan x
  if.
    gt. (number x.as-bytes >> dividend) 0
    [] >> abs-mod
      shrink > @
        grow (abs divisor >> absy)
        abs dividend >> absx
      if. > [chunk] > grow
        chunk.gte absx
        chunk
        grow (chunk.plus chunk)
      if. > [chunk rest] > shrink
        chunk.lt absy
        rest
        shrink (chunk.div 2) (...)
    abs-mod.neg
```

With `x = nan`, `absx` is `nan`. The `grow` loop's exit condition is `chunk.gte absx`, and comparing anything to `nan` is always false, so `grow` doubles forever.

With `x = pinf` (or `ninf`), `absx = pinf`. `grow` does eventually reach `pinf` (since `pinf.gte pinf` is true), but then `shrink(pinf, pinf)` never exits either: `pinf.lt absy` is false for any finite `absy`, so it recurses on `chunk.div 2`, and `pinf / 2` is still `pinf`.

I confirmed this by running the real runtime with a 240-second budget:

```
mod 13 5    = 3.0     in 169ms
mod nan 5   -> still running after 240s
mod pinf 5  -> still running after 240s
mod ninf 5  -> still running (same behavior)
```

For comparison, `Math.IEEEremainder` and `%` both return `nan` for all three cases immediately.

## Steps to Reproduce

```
nan.mod 5
```
or
```
pinf.mod 5
```

Either call hangs indefinitely instead of returning or raising an error.

## What Should Happen

`mod` should check whether the dividend is finite (the same way it already checks the divisor) and return `nan` immediately when it is `nan` or infinite, instead of looping forever.

morphqdd:
Closes #6564.

number.mod only checked whether the divisor was finite before running its abs-mod computation. When the dividend itself is nan or infinite, grow/shrink's exit conditions (chunk.gte absx, chunk.lt absy) are permanently false or permanently true against a non-finite absx, so mod never terminated, regardless of the divisor.

mod now checks the dividend's finiteness first and returns nan immediately for a non-finite dividend, matching Math.IEEEremainder/% semantics. Added two EO tests (nan and pinf dividend against a finite divisor).

### DIFF
--- a/eo-runtime/src/main/eo/number/mod.eo
+++ b/eo-runtime/src/main/eo/number/mod.eo
@@ -8,7 +8,8 @@
 # remainder is below twice the subtracted value, which is exact too, so the
 # answer matches the true remainder for any pair of finite numbers, however
 # far apart their magnitudes are, at the cost of two recursive steps per
-# bit of the quotient.
+# bit of the quotient. When the dividend is `nan` or infinite, the result is
+# `nan`, regardless of the divisor.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3771](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6575, opened 2026-08-10
#6560: restore the ?. fragile-dispatch marker when printing XMIR back to EO

### LINKED ISSUE #6560: `?.` dispatch marker is missing in printed XMIR output from `eo-printer`
## What Happens

`eo-printer`'s `to-eo-tree.xsl` handles `@args @as @atom @base @const @local @name @pipe @star @type` on an `o` element, but has no branch for `@fragile`. The parser (`Emit.java`, method `fragile()`) writes `fragile=""` onto the XMIR node for a `?.` dispatch, and that attribute survives the whole pipeline into the final XMIR, but the printer silently ignores it when turning XMIR back into `.eo` source.

Example:

```
SRC   a?.b > x
XMIR  <o base="Φ.a.b" fragile="" name="x"/>
PRINTED  a.b > x       <- the ?. is gone
```

`a?.b?.c` prints the same way, as `a.b.c`.

## Steps to Reproduce

1. Parse a `.eo` file containing a fragile dispatch, e.g. `a?.b > x`.
2. Print the resulting XMIR back to `.eo` with `Xmir.toEO()` (or run `mvn eo:format`).
3. Compare the output to the original source.

This matters because `mvn eo:format -Deo.autoFix=true` rewrites the user's `.eo` file in place using this printer output, so running the formatter on a file that uses `?.` silently converts every fragile dispatch into a hard one.

## What Should Happen

The printer should render `@fragile` back as `?.` on the base, the same way it already restores every other attribute the parser attaches to a dispatch.

morphqdd:
Closes #6560.

The parser marks a `?.` fragile dispatch with `@fragile=""` on the XMIR node, but `to-eo-tree.xsl` had no branch for that attribute, so the printer silently dropped it and turned every fragile dispatch into a plain one on the next `eo:format -Deo.autoFix`.

The BASED head template's fallback branch now runs the computed surface string through a new `eo:fragile-marked` function, which turns the last dot into `?.` when `@fragile` is set. This covers both shapes a fragile dispatch's base can take: a flattened `Φ.`-rooted dotted path (`foo?.first`) and the reversed-dispatch vertical form (`read?.` with the receiver as a nested child).

Added a print-pack YAML test for the simple case.

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -32,6 +32,18 @@
     <xsl:param name="base" as="xs:string"/>
     <xsl:sequence select="if (starts-with($base, '.')) then concat(eo:translate-path(substring($base, 2)), '.') else if (starts-with($base, concat($eo:program, '.'))) then eo:translate-path(substring-after($base, concat($eo:program, '.'))) else if (starts-with($base, concat($eo:xi, '.'))) then eo:translate-path(substring-after($base, concat($eo:xi, '.'))) else if ($base = $eo:xi) then '$' else if ($base = $eo:program) then 'Q' else eo:translate-path($base)"/>
   </xsl:function>
+  <!--
+  A surface head with its last dot turned into "?." when the dispatch it
+  ends in is fragile (R-3.5 / §9.4): "foo.first" becomes "foo?.first" and
+  the reversed "first." becomes "first?.". A surface with no dot at all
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (37) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3783](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6576, opened 2026-08-10
#6562: accept a bare ! const marker on a vertical argument line

### LINKED ISSUE #6562: `!` const marker inconsistency: Accepted in horizontal arguments, rejected in vertical ones
## What Happens

The `!` const marker on an argument is accepted when the argument is written horizontally, but rejected when the exact same argument is written vertically.

Horizontally, `readArg` in `Tokens.java` reads an optional `:binding` followed by an optional `!` and accepts it fine (this is what PR/issue #5821 added). Vertically, an argument line with no name suffix goes through `Suffix.parse` in `Suffix.java`, which only recognizes `+>`, `->`, `>>` and `>`, and throws `"trailing garbage after expression"` for anything else, including a trailing `!`.

```
a b!             -> parses fine (horizontal)
a > x
  b!             -> [line:col] error: 'trailing garbage after expression'
a > x
  b:lbl!         -> [line:col] error: 'trailing garbage after expression'
a > x
  b:lbl          -> parses fine (binding alone works vertically)
```

This is reachable from the printer too: an application like `a b:1! c:0 > x`, when printed vertically, can come out with a `!` on a body line, producing `.eo` source the parser then refuses to read back.

## Steps to Reproduce

1. Parse `a b! > x` (horizontal form) — succeeds.
2. Parse:
   ```
   a > x
     b!
   ```
   — fails with "trailing garbage after expression".

## What Should Happen

`!` should be accepted on a vertical argument line the same way it is on a horizontal one, since both represent the same argument syntax and the parser should treat them consistently.

morphqdd:
Closes #6562.

`Suffix.parse` only recognized the `!` const marker after `+>`, `->`, `>>`, or `>`, so a vertical argument line with no name suffix (`b` alone, indented under `a > x`) threw "trailing garbage after expression" for a trailing `!`, even though the equivalent horizontal `a b!` form already works (#5821).

A bare `!` now yields `Form.NONE` with the const flag set — same as every other form already does — so `ChainEmission` picks it up through its existing `suffix.constant()` check with no change needed there.

Added a JUnit test on `Suffix` plus an `eo-packs` YAML test confirming the bare-`!` case produces the same `const-to-dataized` rewrite as the horizontal form.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Suffix.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Suffix.java
@@ -309,6 +309,13 @@ final class Suffix {
 
     /**
      * Parse a suffix tail into a result struct.
+     *
+     * <p>A bare {@code !} with no name suffix — the vertical-argument
+     * counterpart of the horizontal {@code a b!} form (#5821), which
+     * {@link #named} already accepts inline but which the unnamed tail of
+     * a vertical argument line never reached (#6562) — yields
+     * {@link Form#NONE} with the const marker set.</p>
+     *
      * @param tail Tail substring
      * @param span Source span
      * @param home Source column where {@code tail} begins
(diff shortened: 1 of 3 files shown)


morphqdd:
@yegor256, please review

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3821](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6577, opened 2026-08-10
#6561: keep @as on the floated wrapper, not the object inside dataized

### LINKED ISSUE #6561: `const-to-dataized.xsl` incorrectly assigns `@as` attribute, leading to potential `α0` collision
## What Happens

`const-to-dataized.xsl` (`eo-parser/src/main/resources/org/eolang/parser/parse/const-to-dataized.xsl`, around line 67) rewrites a `!`-cached argument into a `Φ.dataized` wrapper around the original object. When copying attributes onto the rebuilt tree, it copies every attribute except `const` and `name` onto the inner object (the child of `Φ.dataized`), including `@as`, the inline binding label:

```xml
<o>
  <xsl:attribute name="base" select="'Φ.dataized'"/>
  <o>
    <xsl:for-each select="@*[name()!='const' and name()!='name']">
      <xsl:attribute name="{name()}">
```

But `@as` belongs on the outer wrapper (the object that actually occupies the argument slot after `vars-float-up` replaces it with a reference), not on the object nested inside `dataized`, where the label has no effect.

Example:

```
SRC   a b:lbl!  > x
XMIR  <o base="Φ.a" name="x">
        <o as="α0" base="ξ.a🌵4-4"/>
      </o>
      <o base=".as-bytes" name="a🌵4-4">
        <o base="Φ.dataized">
          <o as="lbl" base="Φ.b"/>
        </o>
      </o>
```

The label `lbl` ends up on the object inside `dataized`, so the outer reference falls back to positional binding `α0` instead of `lbl`. If another argument in the same application already occupies `α0`, the result is invalid XMIR with two attributes bound to the same position:

```
SRC   a b:1! c:0 > x
      <o base="Φ.a" name="x">
        <o as="α0" base="ξ.a🌵4-4"/>   <- should be α1
        <o as="α0" base="Φ.c"/>        <- collision, α1 never bound
      </o>
```

## Steps to Reproduce

1. Parse `a b:lbl! > x` and inspect the resulting XMIR.
2. Compare with the un-cached version, `a b:lbl > x`, where `@as="lbl"` correctly lands on the argument that occupies the slot.
3. For the collision case, parse `a b:1! c:0 > x` and check for two `@as="α0"` siblings.

## What Should Happen

`@as` should be copied onto the outer, newly-created reference object (the one that replaces the argument in the application), not onto the object nested inside the `dataized` wrapper.

morphqdd:
Closes #6561.

`const-to-dataized.xsl` copied every attribute except `const` and `name` onto the object nested inside the `Φ.dataized` wrapper, including `@as` - the inline binding label. That put the label on the wrong object: the one inside `dataized`, where it has no effect, instead of on the outer `.as-bytes` wrapper that actually occupies the argument slot after `vars-float-up` runs.

Two visible effects: the label was silently dropped (the floated reference fell back to a positional `αN` binding instead of keeping the original name), and when another argument already occupied that positional slot, the result was invalid XMIR with two attributes bound to the same position.

`@as` now stays on the outer wrapper and is excluded from the inner copy. Verified against both cases from the issue: a single labeled const argument now keeps its label end-to-end, and the two-argument collision case (`a b:1! c:0 > x`) now resolves to distinct `α0`/`α1` instead of colliding.

Added a print/parse-pack YAML test confirming the label lands on the outer `.as-bytes` node, not the inner `Φ.dataized` one.

### DIFF
--- a/eo-parser/src/main/resources/org/eolang/parser/parse/const-to-dataized.xsl
+++ b/eo-parser/src/main/resources/org/eolang/parser/parse/const-to-dataized.xsl
@@ -61,10 +61,13 @@
       <xsl:attribute name="name" select="$cname"/>
       <xsl:attribute name="line" select="@line"/>
       <xsl:attribute name="pos" select="@pos + 8"/>
+      <xsl:if test="@as">
+        <xsl:attribute name="as" select="@as"/>
+      </xsl:if>
       <o>
         <xsl:attribute name="base" select="'Φ.dataized'"/>
         <o>
-          <xsl:for-each select="@*[name()!='const' and name()!='name']">
+          <xsl:for-each select="@*[name()!='const' and name()!='name' and name()!='as']">
             <xsl:attribute name="{name()}">
               <xsl:value-of select="."/>
             </xsl:attribute>
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-packs/parse/const-to-dataized-keeps-as.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-packs/parse/const-to-dataized-keeps-as.yaml
@@ -1,0 +1,18 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# #6561: const-to-dataized.xsl copied every attribute except "const" and
+# "name" onto the object nested inside the "Φ.dataized" wrapper, including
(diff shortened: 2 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (23) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3779](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6578, opened 2026-08-10
#6563: parenthesize a horizontally-inlined base before a :label tail

### LINKED ISSUE #6563: `to-eo-tree.xsl` printer incorrectly moves inline `:label` binding to nested argument
## What Happens

`to-eo-tree.xsl`'s tail template (`mode="tail"`, around line 382) appends an inline binding `:label` at the end of an object's printed line, after any arguments that are already inlined onto that same line. In EO, `:label` binds to whatever token comes immediately before it, so appending the parent's label after its inlined child's text re-anchors the label onto the child instead of the parent.

Example:

```
SRC       a (b c):lbl > x
XMIR-in   <o base="Φ.a" name="x">
            <o as="lbl" base="Φ.b">
              <o as="α0" base="Φ.c"/>
            </o>
          </o>

PRINTED   a > x
            b c:lbl

XMIR-out  <o base="Φ.a" name="x">
            <o as="α0" base="Φ.b">
              <o as="lbl" base="Φ.c"/>
            </o>
          </o>
```

The label `lbl` moved from `b` to `c`, and `b` silently became a positional argument (`α0`) instead of the labeled one. The printed output reparses without any error, so nothing downstream flags the change — it is a silent rewrite of which argument the label actually names.

## Steps to Reproduce

1. Parse `a (b c):lbl > x` and note that `@as="lbl"` is on `b`, not `c`.
2. Print the XMIR back to `.eo` source.
3. Reparse the printed source and note that `@as="lbl"` has moved to `c`.

## What Should Happen

The printer should keep the binding on the object it actually belongs to, either by keeping `(b c):lbl` inline as written, or by placing `c` on its own line so the label unambiguously stays with `b`.

morphqdd:
Closes #6563.

`Pretty.horizontal` glued a node's tail straight onto the end of its inlined arguments, so a labeled object with its own argument inlined on the same line (`a (b c):lbl > x`) printed as `b c:lbl` instead of `(b c):lbl`. Since `:label` binds to whatever token sits immediately before it, that silently rebound the label from the labeled object (`b`) to its last inlined argument (`c`) on reparse.

The base and its inlined arguments are now parenthesized before a `:label` tail, so the label reads back attached to the same object it named before printing. Verified against the exact case from the issue: printing now goes through the vertical rendering (`b:lbl` / `c` on separate lines, since parenthesizing scores worse than vertical for this input), and reparsing produces the original structure back.

Added a print-pack YAML test covering the case.

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Pretty.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Pretty.java
@@ -274,6 +274,15 @@ final class Pretty {
     /**
      * Render a node horizontally on a single line, if all of its
      * arguments can be inlined.
+     *
+     * <p>A {@code :label} tail (#6563) binds to whatever token sits
+     * immediately before it, so gluing it straight onto the end of the
+     * inlined argument text would silently rebind it from this node's own
+     * head to its last inlined argument on reparse. The base and its
+     * inlined arguments are parenthesized first in that case, so the
+     * label reads back attached to the group, matching what it named
+     * before printing.</p>
+     *
      * @param node The node
      * @param indent The indentation level
      * @return The single line, or empty if inlining is impossible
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3795](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6581, opened 2026-08-10
fix(#6573): unescape %% in the scanf out-of-range message

### LINKED ISSUE #6573: Out-of-range error in `scanf` shows incorrect `%%d` instead of expected `%d`
## What Happens

`string.scanf`'s out-of-range error message (`eo-runtime/src/main/eo/string/scanf.eo`, lines 66-67) contains a doubled percent sign meant to escape to a literal `%d`, but is never passed through `.printf`, so the escaping never happens:

```
        T
          "The number doesn't fit into long range for the '%%d' conversion"
```

Every other diagnostic message in `scanf.eo`/`printf.eo` that contains `%%` is passed through `.printf` (see `printf.eo:174`, `:198`, `:220`, `:351`, `:361`), which unescapes `%%` into a single `%`. This one is handed directly to `T` (bottom), and `PhTerminator.delta()` deliberately does not format-expand the reason string (`throw new ExFailure("%s", reason)` — this was the fix for closed #6139, to stop a literal `%` in user input from being misinterpreted as a format specifier). Since this particular string is never run through `.printf` first, the `%%` is never collapsed, and the user sees the raw escaped text.

I confirmed this by triggering the error path:

```
scanf "%d" "99999999999999999999"
  -> ExFailure: The number doesn't fit into long range for the '%%d' conversion
```

Expected: `... for the '%d' conversion`.

## Steps to Reproduce

```
scanf "%d" "99999999999999999999"
```
The error message reads `'%%d'` instead of `'%d'`.

## What Should Happen

The message should be passed through `.printf` (with no substitution arguments) before being handed to `T`, the same way every other `%%`-containing diagnostic in the file already is, so it displays as `'%d'`.

etojesaandro:
The out-of-range diagnostic in `string.scanf` carried a doubled percent
meant to escape to a literal `%d`, but the string was handed straight to
`T` and never went through `.printf`, so the escaping never happened.
`PhTerminator.delta()` deliberately does not format-expand the reason
(`throw new ExFailure("%s", reason)`, the fix for #6139), which means the
collapsing has to happen at the call site.

The reason is now passed through `.printf` with an empty argument tuple,
the same way every other `%%`-carrying diagnostic in `scanf.eo` and
`printf.eo` already is.

Before:

```
scanf "%d" "99999999999999999999"
  -> ExFailure: ... for the '%%d' conversion
```

After:

```
scanf "%d" "99999999999999999999"
  -> ExFailure: The number doesn't fit into long range for the '%d' conversion
```

Verified by building `eo-runtime` and driving the error path directly:
the reason string now renders with a single `%`. All 25 of `scanf.eo`'s
own `++>`/`-->` tests pass unchanged.

Closes #6573

### DIFF
--- a/eo-runtime/src/main/eo/string/scanf.eo
+++ b/eo-runtime/src/main/eo/string/scanf.eo
@@ -64,7 +64,8 @@
                 digits-lte normalized "9223372036854775808"
                 digits-lte normalized "9223372036854775807"
         T
-          "The number doesn't fit into long range for the '%%d' conversion"
+          "The number doesn't fit into long range for the '%%d' conversion".printf
+            *
         negative.if negated folded
       [a b] > digits-lte
         if. > [index] >> rec
@@ -372,9 +373,12 @@
     "hello%"
     "hello"
 
-  scanf --> stops-on-an-oversized-integer
-    "%d"
-    "99999999999999999999"
+  eq. ++> stops-on-an-oversized-integer
+    recovered
+      head.
+        scanf "%d" "99999999999999999999"
+      42
+    42
 
   gt. ++> can-scan-a-nineteen-digit-value-at-long-max
     head.
@@ -386,17 +390,26 @@
       scanf "%d" "-9223372036854775808"
     -9000000000000000000
 
-  scanf --> stops-on-a-nineteen-digit-value-above-long-max
-    "%d"
-    "9223372036854775808"
+  eq. ++> stops-on-a-nineteen-digit-value-above-long-max
+    recovered
+      head.
+        scanf "%d" "9223372036854775808"
+      42
+    42
 
-  scanf --> stops-on-a-nineteen-digit-value-below-long-min
-    "%d"
-    "-9223372036854775809"
+  eq. ++> stops-on-a-nineteen-digit-value-below-long-min
+    recovered
+      head.
+        scanf "%d" "-9223372036854775809"
+      42
+    42
 
-  scanf --> stops-on-a-zero-padded-integer-above-long-max
-    "%d"
-    "00009223372036854775808"
+  eq. ++> stops-on-a-zero-padded-integer-above-long-max
+    recovered
+      head.
+        scanf "%d" "00009223372036854775808"
+      42
+    42
 
   scanf --> stops-on-a-wrong-format
     "%l"


eo-runtime/src/test/java/org/eolang/EO_string/ScanfDiagnosticsTest.java
```diff
@@ -0,0 +1,80 @@
...
+ * Test case for the diagnostics of {@code string.scanf}.
+ *
+ * <p>The object is written in EO and has no atom behind it, so its
+ * termination reasons cannot be asserted from {@code scanf.eo} itself:
+ * {@code recovered} intercepts a termination but exposes only the
+ * fallback, never the reason.</p>
+ *
+ * @since 0.73.3
+ */
+@SuppressWarnings("JTCOP.RuleEveryTestHasProductionClass")
```
yegor256: @etojesaandro don't create Java test for EO objects. Instead, put a new EO test into the `scranf.eo` file

etojesaandro: Fixed. New test was put

etojesaandro:
Removed Java test implemented eo one. Still issue on master as build is failed

yegor256:
The `qulice` CI check is failing on this PR. Could you please fix the qulice violations?

Also, please note that #6585 appears to address the same issue (#6573). Could the authors coordinate to avoid duplicate work?

---

yegor256:
qulice is red on this PR. Not merging until CI is green.

---

yegor256:
Please forgive us for closing your pull request. It has merge conflicts with the current master branch and has been open for over a month, so we are closing it to keep the pull request queue manageable.

Please feel free to submit a new pull request, perhaps a similar one, but rebased against the latest changes in master. We would be happy to review it.

Thank you for your contribution and understanding.

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2968](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6582, opened 2026-08-10
#6568: make regex matches backtrack across the full text

### LINKED ISSUE #6568: `string.regex.pattern.matches` wrongly interprets a successful prefix match as a full match, resulting in incorrect negatives
## What Happens

`string.regex.pattern.matches` (`eo-runtime/src/main/eo/string/regex.eo`, lines 81-87) is meant to check whether a pattern matches an entire string, the way `java.util.regex.Matcher.matches()` does:

```
    [txt] > matches
      found.exists.and > @
        and.
          found.to.eq txt.length
          found.from.eq 0
      next. > found
        match txt
```

`found` is the result of the *first* unanchored match (`matched-from-index 1 0`), which in the Java atom behind it (`EOregex$EOpattern$EOmatch$EOmatched_from_index.java`) calls `Matcher.find(...)`, not `Matcher.matches()`. `matches` then only checks that this one found span happens to start at 0 and end at the string's length. A real full match needs the regex engine to backtrack across the whole string if the first candidate match doesn't reach the end; `find()` returns the first successful match it hits and never backtracks past it, so any pattern whose first match is a proper prefix of the text is reported as not matching, even though a full match does exist.

I confirmed this by dataizing the built `eo-runtime` and comparing against `java.util.regex.Pattern.matches`/`String.matches` for the same pattern and text:

```
MISMATCH  /a|ab/                  vs "ab"      eo=false java=true
MISMATCH  /(ab|abc)/              vs "abc"     eo=false java=true
MISMATCH  /(?:foo|foobar)/        vs "foobar"  eo=false java=true
MISMATCH  /\d+|\d+\.\d+/          vs "1.5"     eo=false java=true
MISMATCH  /[a-z]+|[a-z]+[0-9]+/   vs "abc1"    eo=false java=true
MISMATCH  /a+?/                   vs "aaa"     eo=false java=true
MISMATCH  /a{1,3}?/               vs "aaa"     eo=false java=true
MISMATCH  /.*?/                   vs "abc"     eo=false java=true
ok        /[a-z]+/, /^[0-9]+$/, /colou?r/, /x*/, /a*/, /(a+)(a*)/, /[0-9]/
```

Patterns already anchored with `^...$` are not affected, since `find()` and `matches()` agree once the caller pins both ends itself, which is why the four stdlib consumers of `matches` (`is-alpha`, `is-digit`, `is-alphanumeric`, `is-ascii`) and the existing tests in `regex.eo` don't show the problem — none of them uses alternation with a shorter left branch or a lazy quantifier.

## Steps to Reproduce

```
"ab".regex.pattern "a|ab" .matches
```

Returns `false`; a genuine full match against `"a|ab"` exists (`"ab"` matches the `ab` alternative), so this should return `true`.

## What Should Happen

`matches` should perform a real anchored full match (the underlying Java engine's `Matcher.matches()`, or an EO-level equivalent that backtracks across the whole string), not just check whether the first unanchored `find()` result happens to span the entire text.

Titanism0v0:
Closes #6568

@yegor256 please take a look.

## Summary

- make `string.regex.pattern.matches` compile an end-constrained form of the original expression, so the regex engine can backtrack beyond a successful prefix
- keep the implementation in EO so JVM and Node use the same matching path
- preserve comments-mode and JVM `\Q` end-of-pattern behavior through lazy compile fallbacks
- add regressions for shorter alternation branches, lazy quantifiers, flags, comments mode, inline flag changes, embedded slashes, quoted patterns, and non-zero match starts

## Verification

- `mvn surefire:test '-Dtest=EOregexTest,EOregexEOcompileTest,EOregexEOpatternEOmatchEOmatchedfromindexTest'` — 57 tests passed
- `mvn -Pqulice com.qulice:qulice-maven-plugin:0.30.7:check` — passed
- independent code review — ready to merge, no Critical/Important findings
- `mvn verify -Pqulice` — did not complete locally: the build spent about 14 minutes in the existing `eo:inference` path (`Needs.follow -> TjDefault.add`) before it was stopped; it had not reached the test phase

### DIFF
--- a/eo-runtime/src/main/eo/string/regex.eo
+++ b/eo-runtime/src/main/eo/string/regex.eo
@@ -83,8 +83,34 @@
         and.
           found.to.eq txt.length
           found.from.eq 0
+      "/(?:%s)(?![\\s\\S])/%s".printf > anchored
+        *
+          parsed.group 1
+          parsed.group 2
+      "/(?:%s\n)(?![\\s\\S])/%s".printf > commented
+        *
+          parsed.group 1
+          parsed.group 2
       next. > found
-        match txt
+        match.
+          compile.
+            regex anchored
+            retry-commented
+          txt
+      next. > parsed
+        match.
+          compiled.
+            regex "/^\\/([\\s\\S]*)\\/([dimsux]*)$/"
+          expression
+      "/(?:%s\\E)(?![\\s\\S])/%s".printf > quoted
+        *
+          parsed.group 1
+          parsed.group 2
+      compile. > [message] > retry-commented
+        regex commented
+        retry-quoted
+      compile. > [message] > retry-quoted
+        regex quoted
 
   eq. ++> can-advance-after-a-zero-width-match
     1
--- a/eo-runtime/src/test/java/org/eolang/EO_string/EOregexEOcompileTest.java
+++ b/eo-runtime/src/test/java/org/eolang/EO_string/EOregexEOcompileTest.java
@@ -30,6 +30,13 @@ final class EOregexEOcompileTest {
     @CsvSource({
         "/a|ab/, ab, true",
         "/a+?/, aaa, true",
+        "/a|ab/i, AB, true",
+        "/([0-9]) #ignore this comment/x, 4, true",
+        "/(?x)([0-9]) #ignore this comment/, 4, true",
+        "/a(?x) #tail/, a, true",
+        "/(?x)a(?-x)/, a, true",
+        "/\\Qa|ab/, a|ab, true",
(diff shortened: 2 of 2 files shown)


0crat:
@titanism0v0 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+124](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6583, opened 2026-08-11
fix(#6569): return positive zero for zero dividend modulo

### LINKED ISSUE #6569: `number.mod` produces negative zero for zero dividend instead of expected positive zero
## What Happens

`number.mod` (`eo-runtime/src/main/eo/number/mod.eo`, lines 27-52) picks the sign of its result based on `dividend.gt 0`:

```
    if.
      gt.
        number x.as-bytes >> dividend
        0
      [] >> abs-mod
        ...
      abs-mod.neg
```

When `dividend` is exactly `0`, `dividend.gt 0` is false, so the `else` branch runs and the (zero) magnitude is negated, producing negative zero instead of positive zero. The condition should be `dividend.gte 0` so a zero dividend takes the positive branch.

I confirmed this on the built runtime:

```
0.0 mod 3.0  -> eo=-0.0 bytes=80-00-00-00-00-00-00-00   java=0.0
0.0 mod -3.0 -> eo=-0.0 bytes=80-00-00-00-00-00-00-00   java=0.0
6.0 mod 3.0  -> eo=0.0  bytes=00-00-00-00-00-00-00-00   java=0.0   (control, magnitude already 0 without negation)
7.0 mod 3.0  -> 1.0     (control, correct)
-7.0 mod 3.0 -> -1.0    (control, correct)
```

The bug is masked when comparing with `number.eq`, because `number/eq.eo` special-cases `+0`/`-0` as equal:

```
(0.mod 3).as-bytes.eq 0.as-bytes  =  false   <- visible at the byte level
(0.mod 3).eq 0                    =  true    <- masked by number.eq's own +0/-0 special case
```

So it's invisible to any test written as `eq. ++> ...` comparing against `0`, but visible in the raw bytes, in printed output, and to any caller comparing byte-for-byte instead of through `number.eq`. The file's own docstring promises "the answer matches the true remainder for any pair of finite numbers", which `-0.0` does not, strictly.

## Steps to Reproduce

```
0.mod 3
```
`.as-bytes` is `80-00-00-00-00-00-00-00` (negative zero) instead of `00-00-00-00-00-00-00-00`.

## What Should Happen

The sign-selection condition should be `dividend.gte 0` instead of `dividend.gt 0`, so a zero dividend takes the positive-zero branch.

joyheroes:
## Summary

Fixes `number.mod` returning negative zero when the dividend is zero.

The sign branch used `gt` to determine whether the result should be negated. For a zero dividend, `0 > 0` is false, so the zero magnitude was sent through `.neg`, producing `-0.0`.

The condition now uses `gte`, so zero dividends stay on the non-negative path and return `+0.0`.

Closes #6569.

## Changes

- Change the sign condition in `number.mod` from `gt` to `gte`.
- Add byte-level regression coverage for:
  - `0 mod 3`
  - `0 mod -3`
- Add normal positive/negative modulo regression checks:
  - `7 mod 3`
  - `-7 mod 3`

## Testing

Verified locally with Java 17 and Maven 3.9.16.

Passed:

- CI-equivalent Maven build:
  `mvn -ntp -PskipITs --errors --batch-mode -Deo.coverageTracking=false -Deo.transpileTests=false clean install`
- Targeted EO modulo tests:
  `mvn test -pl eo-runtime -Dtest=EOmodTest`
- Reverse verification:
  temporarily restoring `gt` reproduces the regression; restoring `gte` makes the tests pass again.

Note:

`-Pqulice` was not run locally because the profile requires Java 21+, while the available environment has Java 17. The non-qulice CI-equivalent build and targeted EO tests both passed.

@yegor256 please take a look.

### DIFF
--- a/eo-runtime/src/main/eo/number/mod.eo
+++ b/eo-runtime/src/main/eo/number/mod.eo
@@ -30,7 +30,7 @@
       divisor.is-finite.not.if
         divisor.is-nan.if nan x
         if.
-          gt.
+          gte.
             number x.as-bytes >> dividend
             0
           [] >> abs-mod
@@ -68,6 +68,16 @@
     mod
       5
       nan
+
+  eq. ++> can-return-positive-zero-when-dividend-is-zero
+    as-bytes.
+      0.mod 3
+    0.as-bytes
+
+  eq. ++> can-return-positive-zero-when-dividend-is-zero-and-divisor-is-negative
+    as-bytes.
+      0.mod -3
+    0.as-bytes
 
   ++> can-stay-compatible-with-division
     eq. > @
@@ -131,10 +141,22 @@
       7
     -1
 
+  eq. ++> can-take-a-modulo-of-minus-seven-by-three
+    mod
+      -7
+      3
+    -1
+
   eq. ++> can-take-a-modulo-of-one-by-two
     mod
       1
       2
+    1
+
+  eq. ++> can-take-a-modulo-of-seven-by-three
+    mod
+      7
+      3
     1
 
   eq. ++> can-take-a-modulo-of-sixteen-by-a-large-negative


0crat:
@joyheroes Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6589, opened 2026-08-11
Write decimal digits over i64 so printf reaches 2^63

### LINKED ISSUE #6572: `printf`'s `%d`/`%f` incorrectly rejects values in range [2^53, 2^63) contrary to documented limit
## What Happens

`string.printf`'s `%d` and `%f` conversions reject any magnitude in the range `[2^53, 2^63)`, even though `printf` itself documents `%d`'s range as up to `2^63` and both conversions are handed values that are exactly representable as a double.

`printf.eo` declares its own limit:

```
  2.power 63 >> limit!
```

and the `%d` branch only rejects a value at or beyond that:

```
        64-.eq conv
        as-decimal
          if.
            or.
              gte. (number numeric) (number limit)
              lt. (number numeric) (times. (number limit) -1)
            T
              "The number %f doesn't fit into the long range of the %%d conversion".printf
                * (number numeric)
            number numeric
```

But the value that survives this check is then handed to `as-decimal`, whose own guard is `2^53`, not `2^63`:

```
    if.
      n.abs.gte 9007199254740992
      T
        "Can't write this number as decimal digits exactly: its magnitude is 2^53 or larger, ..."
```

So a value printf's own check declares in range terminates the program anyway, with the wrong error message (`as-decimal`'s, not printf's). `%f` hits the same 2^53 ceiling through `string/as-fixed.eo`'s call to `as-decimal ip` on the integer part, regardless of the requested precision.

I confirmed this against the exact boundary:

```
printf "%d" 9007199254740991    -> "9007199254740991"                     (ok, == 2^53-1)
printf "%d" 9007199254740992    -> ExFailure: Can't write this number as decimal digits exactly...
printf "%d" 1.0e16              -> same ExFailure                          (Java: 10000000000000000)
printf "%f" 9007199254740991    -> "9007199254740991.000000"               (ok)
printf "%f" 9007199254740992    -> same ExFailure                          (Java: 9007199254740992.000000)
printf "%.2f" 1.0e16            -> same ExFailure                          (Java: 10000000000000000.00)
printf "%.0f" 1.0e16            -> same ExFailure
```

Every one of these values is exactly representable as an IEEE-754 double and prints correctly in both Java and C.

## Steps to Reproduce

```
printf "%d" 9007199254740992
```
Throws instead of printing `9007199254740992`.

## What Should Happen

`%d` and `%f` should be able to print any value within printf's own documented range (magnitude below `2^63`), not just below `2^53`. Either `as-decimal`'s guard needs to widen to match printf's actual contract, or printf needs its own accurate range check before delegating.

Thayorns:
`printf`'s `%d` documents its range as magnitudes below `2^63` and checks for exactly that, but the value it lets through is then handed to `as-decimal`, which refused anything at or above `2^53`. So `printf "%d" 9007199254740992` died with `as-decimal`'s message about doubles, even though the number is an exact integer that Java and C both print fine. `%f` hit the same ceiling through `as-fixed`, which asks `as-decimal` for the integer part.

The ceiling was there because `as-decimal` peeled digits off a double: above `2^53` the division by ten stops being exact and the remainder falls outside `0..9`. This peels them off an `i64` instead, where `i64/div` is exact long division, so the guard moves up to `2^63` and matches what `printf` already promised. The negative boundary `-2^63` stays out of range, because the object negates before converting, so `printf`'s own check now rejects it with `lte` rather than letting `as-decimal` do it with the wrong message.

`%f` still has a limit of its own that this doesn't touch: `as-fixed` scales the value by a power of ten in double arithmetic before splitting off the integer part, so a product above `2^53` is rounded and the last digit can be off. That is left as a puzzle in `as-fixed.eo`.

Closes #6572

### DIFF
--- a/eo-runtime/src/main/eo/string/as-decimal.eo
+++ b/eo-runtime/src/main/eo/string/as-decimal.eo
@@ -28,7 +28,7 @@
           digits m
           "-".as-bytes.concat
             digits m
-        | (n.times -1).as-i64
+        | (n.times -1).as-i64 failure
         [n] >> digits
           if. > @
             n.lt ten
@@ -41,9 +41,10 @@
                   as-number.
                     n.minus (q.times ten)
                   48
-          n.div ten > q!
-        | n.as-i64
-  10.as-i64 > ten!
+          n.div ten failure > q!
+        | n.as-i64 failure
+  T message > [message] >> failure
+  10.as-i64 failure > ten!
 
   eq. ++> can-drop-the-sign-of-a-negative-below-one
     "0"
--- a/eo-runtime/src/main/eo/string/as-decimal.eo
+++ b/eo-runtime/src/main/eo/string/as-decimal.eo
@@ -1,5 +1,9 @@
 # Writes the number `n` as signed decimal digits, truncating its fraction
 # towards zero, so that `42.7` becomes `42` and `-42.7` becomes `-42`.
+#
+# The digits are peeled off an `i64`, where division by ten is exact, so every
+# magnitude below `2^63` is written exactly, and not only the magnitudes below
+# `2^53` that a double divides without losing a digit.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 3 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4122](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6590, opened 2026-08-11
#6572: format signed long values exactly

### LINKED ISSUE #6572: `printf`'s `%d`/`%f` incorrectly rejects values in range [2^53, 2^63) contrary to documented limit

Titanism0v0:
## Summary

- replace the `2^53` decimal-formatting cliff with an exact signed-`i64` to UTF-8 decimal JVM atom
- preserve finite/range/truncation guards and format large positive and negative integers without precision-losing double scaling
- add regressions for `2^53`, signed-long seams and extrema, the reported counterexample, `%d`, `%f`, and `%.0f`

## Root cause

`string.as-decimal` rejected every magnitude at or above `2^53`, although many such doubles represent exact signed 64-bit integers. In addition, `string.as-fixed` scaled large positive values through double multiplication/division, which could change their integer part. Pure-EO `i64` decimal conversion variants were exact but too slow for runtime formatting, so this change keeps EO validation and uses a narrow JVM atom for the final `long` conversion.

Fixes #6572.

## Validation

- JDK 21 typing-enabled focused suite: 107 tests, 0 failures, 0 errors
- isolated signed-long minimum decimal case: 0.344 s
- positive `%f` regression verified RED as `+256` drift, then GREEN after the fix
- final diff and worktree hygiene checks pass
- wider module run completed 1,483 tests before the external 30-minute Windows limit; its only failure was the existing Windows `SyscallTest` behavior

Qulice currently reports one PMD `PreserveStackTrace` violation in `Deadline.java`; that file is byte-identical to `upstream/master` and comes from upstream commit `a5b94c92`, not this PR.

## Compatibility note

This PR implements the atom for the JVM runtime. The corresponding eo2js/Node atom is not included, so cross-runtime parity remains follow-up work and is not claimed here.

### DIFF
--- a/docs/superpowers/plans/2026-08-11-issue-6572-printf-long.md
+++ b/docs/superpowers/plans/2026-08-11-issue-6572-printf-long.md
@@ -4,7 +4,7 @@
 
 **Goal:** Make `string.printf` render finite signed-long `number` values above `2^53` correctly for `%d` and `%f`, while preserving truncation and deliberate failures.
 
-**Architecture:** Keep `string.as-decimal` as the shared integer-rendering path for `%d` and `string.as-fixed`. Convert each accepted `number` once to `i64`, extract digits with exact signed `i64` quotient/remainder operations, and never negate the full input so the signed minimum remains representable.
+**Architecture:** Keep `string.as-decimal` as the shared integer-rendering path for `%d` and `string.as-fixed`. Convert each accepted `number` once to `i64`. Render safe values below `1_000_000_000` with the original fast `number` digit loop; split larger values with exactly one signed `i64.div` by `1_000_000_000`, render the exact safe quotient and remainder through that loop, and left-pad the remainder to nine digits. Never negate the full input, so the signed minimum remains representable.
 
 **Tech Stack:** EO standard-library objects, embedded EO test declarations, Maven Surefire, Java 21.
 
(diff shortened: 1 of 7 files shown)


Titanism0v0:
One implementation note: this PR introduces a JVM atom for exact signed-long-to-decimal conversion, backed by Java's `Long.toString()`.

During implementation, I noticed that there is currently no corresponding Node/eo2js implementation for this atom. I did not include one in this PR because eo2js is maintained separately and adding cross-runtime support would expand the scope of #6572.

Would you prefer this JVM-only implementation to remain as-is, or should I open a separate issue to track Node/eo2js parity?

PR #6591, opened 2026-08-11
Clamp a negative shift in tuple.shifted

### LINKED ISSUE #6570: `tuple.shifted` method misbehaves with negative shifts, returns a tail rather than dropping elements from the front
## What Happens

`tuple.shifted` (`eo-runtime/src/main/eo/tuple/shifted.eo`) is documented as "Drop the first `shift` elements of `list`", and is implemented as:

```
slice list shift list.length
```

`tuple.slice` accepts negative indexes and wraps them (counts from the end) instead of rejecting them, and `shifted` does not validate `shift` before passing it through. So a negative `shift` does not "drop the first N elements" at all; it silently drops everything except the last `|shift|` elements.

```
shifted([1,2,3], -1)    -> [3]        instead of the whole list (or a documented rejection)
shifted([5,3,5,1], -2)  -> [5,1]
```

This is the same class of docstring/behavior mismatch already accepted in #5861 for `tuple.back`.

## Steps to Reproduce

```
(tuple 1 2 3).shifted -1
```
Returns `[3]` instead of `[1, 2, 3]` (dropping zero or a negative count of elements should return the list unchanged, or the call should be rejected).

## What Should Happen

`shifted` should either reject a negative `shift` explicitly, or treat it as "drop zero elements" (return the list unchanged), instead of silently reinterpreting it as an offset from the end of the list.

Thayorns:
`tuple.shifted` handed `shift` straight to `tuple.slice`, and `slice` wraps a negative index around to the end of the list. So `shifted` did the opposite of what its docstring promises: `(tuple 1 2 3).shifted -1` returned `[3]` rather than dropping nothing.

A negative shift is now clamped to zero, so the whole list comes back unchanged. Rejecting it outright was the other option the issue offered, but nothing in the `tuple` package raises an error for an out-of-range index — `slice`, `back` and `front` all clamp or wrap — so clamping keeps `shifted` in line with its siblings. The docstring now says so too.

Three tests cover the boundary: a shift of `-1`, a shift below the negative length, and a shift of zero, which nothing tested before.

Closes #6570

### DIFF
--- a/eo-runtime/src/main/eo/tuple/shifted.eo
+++ b/eo-runtime/src/main/eo/tuple/shifted.eo
@@ -41,7 +41,7 @@
       6
     *
 
-  eq. ++> dont-shift-by-a-negative-number
+  eq. ++> can-ignore-a-negative-shift
     shifted
       *
         "привет"
@@ -55,7 +55,7 @@
       0
       "ξ"
 
-  eq. ++> dont-shift-by-a-number-below-the-negative-length
+  eq. ++> can-ignore-a-shift-below-the-negative-length
     shifted
       *
         82.42
@@ -67,7 +67,7 @@
       ""
       "日本"
 
-  eq. ++> dont-shift-by-zero
+  eq. ++> can-ignore-a-zero-shift
     shifted
       *
         "a"
--- a/eo-runtime/src/main/eo/tuple/shifted.eo
+++ b/eo-runtime/src/main/eo/tuple/shifted.eo
@@ -16,31 +16,6 @@
       shift
     list.length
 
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 ping

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4138](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6592, opened 2026-08-11
Leave a formation that delegates through φ incomplete

### LINKED ISSUE #6565: `Provides.java` incorrectly marks formations as complete, missing attributes via `φ` delegation or package-level scope
## What Happens

`eo-inference`'s `Provides.java` (line 77) decides whether a formation is "complete" (meaning the checker has seen everything it contains, so a missing-attribute mistake reported against it is trustworthy) purely by whether the formation has a `λ` (atom) child:

```java
rows.add(owner)
    .set("index", Integer.toString(seen))
    .set("complete", Boolean.toString(formation.nodes("o[@name='λ']").isEmpty()));
```

This ignores two other ways a formation can expose attributes the `provides` table does not list: delegation through `φ` (the decoratee), and attributes that live in the formation's own package as separate top-level objects. The class's own javadoc defines "complete" as "we have seen the whole formation, so there is nothing in it besides the attributes listed", and `Clues.java` explains the point of the flag is to keep the checker silent on parts of the program it cannot see, such as atoms and delegation through `φ`. The current check only covers atoms.

Minimal example:

```eo
[] > app
  holder.extra > @
  [] > holder
    base > @
    [] > base
      [] > extra
```

produces a `provides` row for `Φ.app.holder` marked `complete="true"` listing only `φ` and `base`, plus a `needs` row asking for `extra` on the object reached through `φ`. The checker concludes a complete object is missing `extra`, even though `holder.extra` resolves fine at runtime through `φ → base → extra`.

This is not a synthetic corner case. I generated the real `provides`/`needs`/`links` tables from a full `eo-runtime` build and found, for example, `Φ.number` marked `complete="true"` while listing only 8 of its real attributes (`as-bytes`, `φ`, `as-i16`, `as-i32`, `div`, `gt`, `plus`, `times`); every other attribute (`minus`, `eq`, `lt`, `mod`, `floor`, `abs`, `sqrt`, `neg`, ...) lives as a package member (e.g. `Φ.number.minus`) and is invisible to this check. The generated `needs.xml` table already contains 1532 rows asking for `eq` and 185 asking for `minus` against this "complete" object, and `links.xml` has 4899 rows connecting some object to `Φ.number`. The same thing happens for `ρ`: every outer-name reference lowers to a `.ρ` dispatch, `needs` records a requirement for `ρ`, and no formation's `provides` row ever lists `ρ`, yet formations owning it are still marked complete.

## Steps to Reproduce

1. Build `eo-inference`'s tables (`Xmirs`/`Grouped`/`Provides`/`Needs`/`Links`, or run the `Inferring` pipeline) over the snippet above, or over the real `eo-runtime` stdlib.
2. Inspect `provides.xml` for `Φ.app.holder` (or `Φ.number`): it is `complete="true"` but its listed attributes are a strict subset of what it actually exposes through `φ` and its package.

## What Should Happen

A formation should only be marked `complete="true"` when there is no `φ` (no delegation), or after `provides` also accounts for the attributes exposed through `φ` and through package-level access, so that "complete" genuinely means "the checker has seen everything", per the class's own documented contract.

Thayorns:
`Provides` decided that a formation is complete by looking for a `λ` child alone, so any formation that delegates through `φ` was written down as complete while its row listed only the attributes it binds itself. In the snippet from the issue, `Φ.app.holder` came out complete with `φ` and `base` on it, and the checker would call `holder.extra` a mistake even though it resolves at runtime through `φ → base → extra`.

Now a formation is complete only when it binds neither `λ` nor `φ`. What sits behind the decoratee is the business of the links table, so until the checker follows those links, an object that delegates says nothing and the checker stays silent about it rather than wrong.

Two parts of the report are left as puzzles: attributes that live in the owner's package as top-level objects (`Φ.number.minus` and friends), and the `ρ` that every object has but no row lists.

Closes #6565

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Provides.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Provides.java
@@ -86,12 +86,10 @@ final class Provides implements Clue {
         int seen = 0;
         for (final XML formation : new Xmirs(xmirs).formations()) {
             final String owner = formation.xpath("@loc").get(0);
+            final boolean whole = formation.nodes("o[@name='λ' or @name='φ']").isEmpty();
             rows.add(owner)
                 .set("index", Integer.toString(seen))
-                .set(
-                    "complete",
-                    Boolean.toString(formation.nodes("o[@name='λ' or @name='φ']").isEmpty())
-                );
+                .set("complete", Boolean.toString(whole));
             seen = seen + 1;
             for (final XML attr : formation.nodes("o[@name and not(@name='λ')]")) {
                 final String name = attr.xpath("@name").get(0);
--- a/eo-inference/src/main/java/org/eolang/inference/Provides.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Provides.java
@@ -51,7 +51,9 @@ import java.nio.file.Path;
  * the checker honest later: a missing attribute is a mistake only when
  * the object that misses it is complete. An atom is not complete, since
  * its {@code λ} attribute stands for a body written in Java, which this
- * module cannot read.</p>
+ * module cannot read. Neither is a formation that binds {@code φ}: what
+ * it delegates to answers for every name it does not bind itself, and
+ * that object is the business of the links table, not of this one.</p>
  *
  * <p>Three kinds of objects are deliberately absent from this table.
  * Applications and references (anything with a {@code @base}) provide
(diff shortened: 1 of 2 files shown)


maxonfjvipon:
@Thayorns please resolve conflicts and try again, I need this change ASAP

Thayorns:
@maxonfjvipon take a look please when you have time

Thayorns:
@yegor256 ping

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4154](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6593, opened 2026-08-11
Hash directories through Sha instead of Cache

### LINKED ISSUE #6552: Incorrect hash directories implementation in Cache.java: Should use SHA instead
The puzzle `5254-6c9a15c4` from #5254 has to be resolved: 

https://github.com/objectionary/eo/blob/20f5efe3ba136f3d9f04433cc12ffd6196f82bf4/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java#L140-L144

Thayorns:
`Cache` carried its own directory hashing next to `Sha`, which already knew how to walk a tree. The two existed side by side only because `Sha` hashed every file it reached, while `Cache` had to drop the ones its filter rejected.

`Sha` now takes that filter as a second attribute and applies it during the walk, so `Cache.dirSha` is gone and `Cache.sha` hands the whole job over. Hashing a tree lives in one place again.

The digest of a directory changes shape as a result, since `Sha` frames each file by its relative path, its bytes and its length, where the old code framed it by the path and the per-file digest. Nothing persists across versions here beyond the cache itself, so the worst of it is one cold pass after the upgrade.

`CacheTest` used to spell out the old recipe byte for byte; it now asserts what `Cache` actually promises, that the marker holds the `Sha` of the source. A new test in `ShaTest` pins the filter down: a directory hashed with a filter matches a directory holding only the files that filter accepts.

Closes #6552

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
@@ -5,15 +5,9 @@
 package org.eolang.maven;
 
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
-import java.util.Base64;
-import java.util.Comparator;
 import java.util.function.Predicate;
-import java.util.stream.Stream;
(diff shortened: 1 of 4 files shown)


Thayorns:
Superseded by #6602, which landed the same fix for #6552 while this was open. Closing in its favour.

PR #6596, opened 2026-08-11
Give while a cycle that does not nest

### LINKED ISSUE #6550: Constant stack issue when walking a directory in directory.eo file
The puzzle `6484-f79facd0` from #6484 has to be resolved: 

https://github.com/objectionary/eo/blob/0844c094aecf51a996d4d8b7d750be7cdf8cda3f/eo-runtime/src/main/eo/directory.eo#L12-L15

Thayorns:
`while` was the last loop in the language that still nested. Every cycle sat inside the previous one, so a few thousand iterations blew the JVM stack, and every object built on top of it inherited the limit. That is exactly what stopped `directory.walk` from walking a large tree, and what the puzzle in `directory.eo` named as the blocker.

So `while` is an atom again. The Java side runs a real Java loop: it makes `body` for 0, 1, 2 and so on for as long as `condition` says yes about the next index, dataizes every body but the last one, and hands the last one back — or `false` when the very first index was already refused. That is the same order the recursive version used, checking the next condition before dataizing the current body, so `number.integral` and the malloc-driven loops keep their results. The new test runs ten thousand cycles, which the old `while` could not survive.

This does not finish the walk. `stepped` still costs a frame per entry, because it carries a tuple as its accumulator and `malloc` holds bytes only, so a cycle has nowhere to keep it. The puzzle stays, narrowed to that: join the matched paths into one string inside the loop and split it at the end.

Closes #6550


eo-runtime/src/main/java/org/eolang/EOwhile.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * WHILE.
+ * Makes {@code body} for the indexes 0, 1, 2 and so on, while {@code condition}
+ * says yes about the next index, dataizes all of them but the last one, and
+ * gives the last one back, or {@code false} when the first index was refused.
+ * The cycles do not nest.
+ * @since 0.74.0
+ */
+@XmirObject(oname = "while")
+public final class EOwhile extends PhDefault implements Atom {
```
yegor256: @Thayorns we don't want more atoms in our code, but only less. `while` should be an object, not an atom.

Thayorns: Agreed, the atom is gone. `while` stays an object, and it now runs its cycles in blocks that double in size — 1, then 2, then 4 — where each block is a balanced tree of cycles rather than a chain. The only state it keeps is the index of the last cycle, in eight bytes of `malloc`, so a million cycles nest about forty frames deep instead of a million. The order in which the condition and the body are dataized is unchanged, and no test moved.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1185](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2932](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6601, opened 2026-08-11
#6588: fix number.mod negative zero for infinite divisor

### LINKED ISSUE #6588: `number.mod` still returns negative zero for a negative-zero dividend when the divisor is infinite
`number.mod` (`eo-runtime/src/main/eo/number/mod.eo:31`) returns the dividend `x` completely unchanged when the divisor is infinite: `divisor.is-nan.if nan x`. Issue #6569 fixed the sign of a zero result for a *finite* divisor by changing the comparison two lines below (`gt.` → `gte.`) so a zero dividend takes `abs-mod`'s positive branch, but that fix never touches this infinite-divisor branch, which just hands back `x` verbatim.

As a result a negative-zero dividend still comes out negative when the divisor is `pinf`/`ninf`, contradicting the very invariant #6569 established. I confirmed this on the built runtime by dataizing `mod` and reading the raw IEEE-754 bytes (same technique the file's own `as-bytes.eq` tests use, since plain `number.eq` treats `+0`/`-0` as equal and would hide this):

```
mod(-0.0, 3)     -> 00-00-00-00-00-00-00-00  (+0.0, correct, fixed by #6569)
mod(-0.0, pinf)  -> 80-00-00-00-00-00-00-00  (-0.0, bug)
mod(-0.0, ninf)  -> 80-00-00-00-00-00-00-00  (-0.0, bug)
```

Fix: normalize the sign on line 31 the same way, e.g. return `x.plus 0` instead of bare `x` — IEEE-754 addition of `+0` turns `-0.0` into `+0.0` and is a no-op for every other finite value.

morphqdd:
Fixes #6588.

`number.mod` returned the dividend unchanged when the divisor is infinite. For a negative-zero dividend that meant the sign stayed negative, which contradicts the invariant #6569 already set for the finite-divisor case (zero dividend always yields positive zero).

Fix: normalize the sign with `x.plus 0` in that branch, same trick used for the finite case. IEEE-754 addition of `+0` turns `-0.0` into `+0.0` and is a no-op for every other finite value.

Added two tests that check the raw bytes of `0.neg.mod pinf` and `0.neg.mod ninf`, since plain `number.eq` treats `+0`/`-0` as equal and would hide the bug.

### DIFF
--- a/eo-runtime/src/main/eo/number/mod.eo
+++ b/eo-runtime/src/main/eo/number/mod.eo
@@ -28,7 +28,9 @@
     dividend.is-finite.not.if
       nan
       divisor.is-finite.not.if
-        divisor.is-nan.if nan x
+        divisor.is-nan.if
+          nan
+          x.plus 0
         if.
           gte.
             number x.as-bytes >> dividend
@@ -68,6 +70,16 @@
     mod
       5
       nan
+
+  eq. ++> can-return-positive-zero-for-negative-zero-dividend-and-negative-infinite-divisor
+    as-bytes.
+      0.neg.mod ninf
+    0.as-bytes
+
+  eq. ++> can-return-positive-zero-for-negative-zero-dividend-and-positive-infinite-divisor
+    as-bytes.
+      0.neg.mod pinf
+    0.as-bytes
 
   eq. ++> can-return-positive-zero-when-dividend-is-zero
     as-bytes.


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3753](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6602, opened 2026-08-11
#6552: hash directories through Sha instead of Cache's own loop

### LINKED ISSUE #6552: Incorrect hash directories implementation in Cache.java: Should use SHA instead

morphqdd:
Closes #6552.

Resolves the puzzle in `Cache.dirSha`. That method used to walk a directory, hash each file separately with `Sha`, then combine those per-file hashes with its own second-level digest. `Sha` already had its own directory-hashing algorithm, so the two lived side by side and only agreed by coincidence.

Gave `Sha` a second constructor that takes a files filter, applied only when walking a directory. `Cache.dirSha` now just delegates: `new Sha(dir, this.filter).toString()`. `Cache.java` no longer needs its own digest/walk code, so the unused imports went with it.

This changes the exact byte layout of directory hashes (single-level framing by relative path and byte length, same as `Sha` already does for the cases that went through it directly), so `CacheTest.generatesCorrectHashForEntireFolderWithSeveralFiles` now builds its expected digest the same way `Sha` frames a file, instead of hardcoding the old two-level scheme.

Ran `CacheTest`, `ShaTest`, and the other cache-consumer tests (`ConcurrentCacheTest`, `GlobalCacheTest`, `ChCachedTest`, `OyCachedTest`, `Fp*CacheTest`) locally, all green. `qulice:check -Pqulice` on `eo-maven-plugin` is clean.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
@@ -5,15 +5,9 @@
 package org.eolang.maven;
 
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
-import java.util.Base64;
-import java.util.Comparator;
 import java.util.function.Predicate;
-import java.util.stream.Stream;
 import org.cactoos.Func;
 import org.cactoos.func.UncheckedFunc;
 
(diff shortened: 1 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3769](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6603, opened 2026-08-11
#6110: collapse under-indented blank text-block lines to empty

### LINKED ISSUE #6110: `Globals.appendTextLine` incorrectly maintains whitespace on under-indented blank lines

morphqdd:
Closes #6110.

`Globals.appendTextLine` only stripped a text-block body line when its length was at least the opener's indent. A blank line shorter than that (a few stray leading spaces, no content) fell into the else branch and was appended unstripped, so the stray spaces survived into the assembled string literal.

Fix: check `raw.chars().allMatch(c -> c == ' ')` first and collapse any such line to an empty string, before falling back to the existing length/prefix check for non-blank lines.

Added a direct `GlobalsTest` case that reproduces the reported scenario (`openTextBlock` at indent 6, `appendTextLine("  ")`), and confirmed it fails without the fix and passes with it.

### DIFF
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/text-block.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/text-block.yaml
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: MIT
 ---
 # yamllint disable rule:line-length
+# yamllint disable rule:trailing-spaces
 sheets: []
 asserts:
   - //o[@base='Φ.string' and o/o[text()='48-69-2C-20-74-68-65-72-65-0A-41-64-69-C3-B3-73']]
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/text-block.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/text-block.yaml
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: MIT
 ---
 # yamllint disable rule:line-length
-# yamllint disable rule:trailing-spaces
 sheets: []
 asserts:
   - //o[@base='Φ.string' and o/o[text()='48-69-2C-20-74-68-65-72-65-0A-41-64-69-C3-B3-73']]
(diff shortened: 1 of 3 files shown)


yegor256:
@morphqdd can you add a YAML fixture to demonstrate the effect of this fix?

morphqdd:
Added: text-block.yaml now has a blank-under-indented case (2 stray spaces, shorter than the opener's indent), asserting it collapses to the same bytes as a fully-empty blank line. Confirmed it fails on the unfixed code and passes with the fix.

morphqdd:

yegor256:
@morphqdd fix CI please

morphqdd:
@yegor256, review please

eo-parser/src/test/resources/org/eolang/parser/eo-syntax/text-block.yaml
```diff
@@ -2,12 +2,14 @@
 # SPDX-License-Identifier: MIT
 ---
 # yamllint disable rule:line-length
+# yamllint disable rule:trailing-spaces
```
yegor256: @morphqdd trailing spaces will be removed by any pre-commit tool and this test will break. Find another way.

morphqdd:
Switched `input` to a quoted scalar with explicit `\n` instead of a literal block. The spaces are no longer trailing on any physical line, so no auto-trim tool can strip them. Same decoded string, verified with a round-trip through PyYAML.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3861](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2953](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6604, opened 2026-08-11
#6183: name the ArrayList capacity constant

### LINKED ISSUE #6183: Unjustified usage of magic number in `Eo.java` initialization of `spans` variable

morphqdd:
Closes #6183.

`Eo.java` initialized the `spans` list with `new java.util.ArrayList<>(16)`, a magic number with no explanation. Qulice requires an explicit capacity on every `ArrayList` construction, so the number can't just be dropped.

Replaced `16` with a named constant, `SPANS_CAPACITY`, set to `10` (the same default `ArrayList`'s own no-arg constructor uses), with a javadoc line saying so.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -41,6 +41,11 @@ final class Eo implements Iterable<Directive> {
     private static final String ROOT_TOKENS = "*(QT@^$%";
 
     /**
+     * Initial capacity of the source line buffer, {@link java.util.ArrayList}'s own default.
+     */
+    private static final int SPANS_CAPACITY = 10;
+
+    /**
      * Raw EO source text.
      */
     private final String source;
@@ -69,7 +74,7 @@ final class Eo implements Iterable<Directive> {
             level -> Eo.checkOnClose(level, emit),
             parent -> Eo.beforeChild(parent, emit)
         );
-        final java.util.List<Span> spans = new java.util.ArrayList<>(16);
+        final java.util.List<Span> spans = new java.util.ArrayList<>(Eo.SPANS_CAPACITY);
         new Source(this.source).forEach(spans::add);
         final Recovery recovery = new Recovery(spans);
         int idx = 0;


eo-parser/src/main/java/org/eolang/parser/Eo.java
```diff
@@ -69,7 +74,7 @@ Iterable<Directive> directives() {
             level -> Eo.checkOnClose(level, emit),
             parent -> Eo.beforeChild(parent, emit)
         );
-        final java.util.List<Span> spans = new java.util.ArrayList<>(16);
+        final java.util.List<Span> spans = new java.util.ArrayList<>(Eo.SPANS_CAPACITY);
```
yegor256, line 77: @morphqdd maybe it's better to simply use `LinkedList` here?

morphqdd: LinkedList would make the parse loop O(n^2): `directives()` and `mergeBytesContinuation()` both do `spans.get(idx)` inside a `while (idx < spans.size())` loop, which is O(1) per call on ArrayList but O(n) on LinkedList. Keeping ArrayList, just naming the capacity instead of leaving it unexplained.

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (7) hits-of-code; +1 to give you at least something.  Your running score is [+2738](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (7) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3781](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6615, opened 2026-08-11
Drain the pending checks and say what is certainly missing

### LINKED ISSUE #6614: `eo-inference` fails to report type mistakes due to undraining pending checks
`Clues` at `eo-inference/src/main/java/org/eolang/inference/Clues.java:66` follows four clues and then stops:

```java
public Clues() {
    this(Arrays.asList(new Provides(), new Needs(), new Links(), new Checks()));
}
```

Each writes one table into `6-inference`: what every object certainly has, what it is asked for, which type is a copy of which, and what has to be checked once those three are read together. Over a full `eo-runtime` build that comes to 2,468 types, 9,950 needs, 20,825 links and 21,094 pending checks. Not one of the checks is ever decided. The module says otherwise in three places — `Clues.java:46`, `package-info.java:30` and `MjInference.java:36` all promise that "the checks are drained one by one, each of them either deciding, splitting into smaller checks, or waiting for facts that may never come" — and `Checks.java:36` justifies filing a check unresolved by saying it "leaves the arithmetic to the checking loop". There is no such loop. The `x.next.foo` program printed in `Clues`' own javadoc, where `next` has nothing at all, produces the same tables and the same green build as a correct one.

The loop is specified down to its steps by the design this module follows. Take a check; resolve its owner through `links` to the formation it copies; count that formation's voids in the order `provides` recorded them, so that "the argument bound as `α0` must fit" becomes "this argument must have everything that void needs"; decide it when the facts are there, which may split it into smaller checks, since a need answered by an attribute becomes a new check between that attribute's type and the needed one; put it back and take the next when they are not; stop when the list is empty or nothing can make progress. Every started check is remembered and never started twice, which is what keeps a self-referential object from spinning forever.

One rule is not decoration: report a mistake only when sure, meaning the type on the "has" side is `complete="true"` and the attribute is still absent. An atom, whose `λ` body is Java, an object this module cannot see, or a check that never collects enough — all of these have to keep it quiet. That makes #6565 a precondition rather than a detail: while `complete="true"` is written for formations whose attributes arrive through `φ` or through their package, the loop's first sweep of `eo-runtime` would be a flood of false complaints about `Φ.number` missing `eq`.

A fifth component that reads the four tables and drains the list is the smallest thing that closes this, writing its verdicts as a document beside them rather than failing the build yet. It should stay as primitive as its neighbours — "a copy of" still meaning "the same as", `φ` not followed when asking whether a type has an attribute, atoms contributing no rows — because each of those is a growth point the design names, one place to change later. The worked example is the first pack to write: `x.next.foo` reaches the mistake in three steps. While in those files, the line at `Clues.java:51` and `package-info.java:42` claiming only the first rule is implemented should go, since four of them are.

maxonfjvipon:
The four clues wrote down what they understood and decided nothing, so every
pending check stayed pending forever. Three javadocs promised otherwise —
`Clues`, `package-info` and `MjInference` all said the checks "are drained one
by one" — and `Checks` justified filing a check unresolved by leaving "the
arithmetic to the checking loop". This is that loop.

`Concluded` follows the clues and then reads their four tables together:

```xml
<problems>
  <type id="Φ.app.t.next">
    <attr name="foo" asked="Φ.app.inc.φ"/>
  </type>
</problems>
```

which says that the object at `Φ.app.t.next` is asked for `foo`, at
`Φ.app.inc.φ`, and will never have it. On the program of the design note that
takes the three steps the note describes: the `t` fills the void `x`, the `x`
is asked for `next`, and the `next` it finds is asked for `foo`. A check that
succeeds is rarely the end of it — whatever the attribute holds must in turn
have everything that taking it is asked for, which is a smaller check of the
same shape, so the list grows while it is drained.

The honesty rule is the point of the whole thing: a mistake is written down
only when the object that misses the attribute has been seen whole. An atom, an
object nobody in the program describes, an argument whose void cannot be found
— all of these leave a check undecided and nothing is said. Four of the seven
new packs are about exactly that silence, since it is what the checker will be
doing most of the time.

Three things I decided along the way and would rather have written down than
left implicit:

  * **No putting a check back.** The note has the loop park a check it cannot
    decide and try again later. Here it drops it, because all four tables are
    written before the loop starts, so nothing new is ever learned while it
    runs: a check nobody can decide now will not become decidable later. The
    list only grows by splitting, of which there are finitely many, so
    remembering the checks already started is about work and not about
    termination. When a copy starts receiving types of its own, facts will
    appear mid-loop and the parking comes back.

  * **The package is a second place to look.** `Φ.number` binds eight
    attributes and answers to forty; the rest are objects of their own in the
    same package, `Φ.number.eq` among them, and a locator names both kinds the
    same way — so an attribute that no row lists is looked for at
    `Φ.number.eq` before anything is said. The note names delegation through
    `φ` as the same sort of growth point; that one is not looked into yet.

  * **A chain of copies is one name.** `Same` collapses `Φ.app.φ.α0` →
    `Φ.app.lid` → `Φ.jar` into `Φ.jar`, so every question about a copy is
    answered by the formation it comes from. This is the single place the
    checker gets smarter later, and nothing else will have to change.

`Ungrouped` and `Row` are the inverse of `Grouped` and `Cells`: they read a
table back into flat rows under the name each owner goes by. Neither knows a
column by name, so a rule that starts writing a new one has it read back for
free.

---

On the runtime the loop drains **25,960** pending checks in about **5s** (the
`inference` step alone goes from 19.9s to 24.9s) beside 2,482 types, 10,090
needs and 21,068 links.

It reports 62 mistakes, and **every one of them is false**. That is worth
stating plainly rather than hiding, because they all have one cause and it is
not this loop: `complete="true"` is written for objects whose attributes arrive
through `φ` or through their package, which is #6565. Applying that ticket's
one-line rule locally — a formation that binds `φ` is not complete, as PR #6592
proposes — the same sweep reports **nothing at all**. So the loop is as honest
as the table it trusts, and it stays quiet on the whole runtime the moment
that table stops lying. I left `Provides` alone here rather than duplicating an
open PR.

Nothing fails the build. The verdicts land as a document beside the tables,
which is what a prototype's opinion is worth for now.

Closes: #6614

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Clues.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Clues.java
@@ -24,13 +24,13 @@ import java.util.Collection;
  *   [x] &gt; inc
  *     x.next.foo &gt; @</pre>
  *
- * <p>Three clues are planned, one per kind of object, and each fills a
- * table of its own: what an object certainly has ({@link Provides}), what
- * it must have judging by how it is used ({@link Needs}), and which types
- * are copies of which ({@link Links}), and what has to be checked once
- * those three are read together ({@link Checks}). Keeping the tables apart is what lets a smarter
- * rule add rows, or read them differently, without touching anything
- * else.</p>
+ * <p>There is a clue per kind of object, and each fills a table of its
+ * own: what an object certainly has ({@link Provides}), what it must have
+ * judging by how it is used ({@link Needs}), which types are copies of
+ * which ({@link Links}), and what has to be checked once those three are
+ * read together ({@link Checks}). Keeping the tables apart is what lets a
+ * smarter rule add rows, or read them differently, without touching
+ * anything else.</p>
  *
  * <p>Every clue reads the program as text, writing a row for an object
  * because it is <em>there</em>, not because something reaches it. The
@@ -43,12 +43,10 @@ import java.util.Collection;
  * buys back later by resolving each site in its own context; reading only
  * what runs costs coverage, which nothing buys back.</p>
  *
- * <p>Then the checks are drained one by one, each of them either
- * deciding, splitting into smaller checks, or waiting for facts that may
- * never come. A mistake is reported only when the object that misses an
- * attribute is complete, so that parts of the program the checker cannot
- * see — atoms, delegation through {@code φ} — make it silent rather than
- * wrong. Only the first clue is here so far.</p>
+ * <p>No clue decides anything, which is why they can be read one after
+ * another in any order. Draining the checks the last of them files is the
+ * work of {@link Concluded}, which follows all of them first and then reads
+ * their tables together.</p>
  *
  * @since 0.67.0
  */
--- a/eo-inference/src/main/java/org/eolang/inference/Concluded.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Concluded.java
@@ -1,0 +1,82 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.xml.XMLDocument;
+import com.yegor256.tojos.MnMemory;
+import com.yegor256.tojos.TjDeferred;
+import com.yegor256.tojos.Tojos;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
+
+/**
+ * The clues, followed and then read together.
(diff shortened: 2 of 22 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1049 >= 100); -8 for way too many hits-of-code (1049 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+662](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6618, opened 2026-08-11
#6610: apply Sha's filter only when hashing a directory

### LINKED ISSUE #6610: `Sha` applies its filter to a lone file in `Sha.java`, against its own Javadoc
The `filter` attribute of `Sha` in `eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java:36` is documented as "Files filter, applied only when hashing a directory", and the constructor parameter at `Sha.java:51` repeats the same promise. The `hash()` method does not keep it. The `.filter(this.filter)` step at `Sha.java:78` sits in the walk unconditionally, and `Files.walk` over a regular file yields exactly that one file, so the filter is consulted for a lone file just as it is for every file inside a directory.

A lone file that the filter rejects therefore hashes to `47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=`, the digest of no input at all, instead of the digest of its own bytes, and every rejected input collides on that one value. `Cache.sha` at `eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java:116` happens to reach the filtered constructor only for directories, taking the unfiltered `new Sha(any)` branch for files, so nothing miscomputes today. The class still hands out a contract its callers cannot rely on, and the next caller that passes a filter alongside a file path gets a silently wrong hash rather than an error.

Either guard the step with `Files.isDirectory(this.path)` so the Javadoc holds, or drop "applied only when hashing a directory" from both lines and let the filter govern the whole walk. Whichever way it goes, `ShaTest` never constructs a `Sha` with a filter, so deleting the `.filter(this.filter)` line breaks no test at all; a case pinning the filtered behaviour down belongs alongside the fix. Both the constructor and the step arrived in #6602.

morphqdd:
Closes #6610.

`Sha.hash()` applied its filter unconditionally, even for a lone file, contradicting the field's own Javadoc ("applied only when hashing a directory"). A lone file the filter rejects hashed to the digest of no input at all instead of its own bytes.

Guarded the filter with `Files.isDirectory(this.path)`, so a lone file always hashes its own bytes regardless of what filter was passed in, matching the documented contract.

Added `ignoresTheFilterForALoneFile` to `ShaTest`, constructing `Sha` with a filter that rejects everything and checking the hash still matches the file's raw bytes.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Sha.java
@@ -73,9 +73,15 @@ final class Sha {
      */
     private String hash() throws IOException, NoSuchAlgorithmException {
         final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+        final Predicate<Path> active;
+        if (Files.isDirectory(this.path)) {
+            active = this.filter;
+        } else {
+            active = any -> true;
+        }
         try (Stream<Path> walk = Files.walk(this.path)) {
             walk.filter(Files::isRegularFile)
-                .filter(this.filter)
+                .filter(active)
                 .sorted(Comparator.comparing(this::relative))
                 .forEach(file -> this.feed(digest, file));
         }
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/ShaTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/ShaTest.java
@@ -99,6 +99,27 @@ final class ShaTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (29) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3833](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6619, opened 2026-08-11
#6611: collapse Cache.dirSha into Cache.sha

### LINKED ISSUE #6611: `Cache.dirSha` is a one-line delegator with stale Javadoc in `Cache.java`
Since #6602 moved directory hashing into `Sha`, the `dirSha` method in `eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java:135` is a single line, `return new Sha(dir, this.filter).toString();`, and nothing more. Its Javadoc one line above still describes the loop it used to run, "Calculate SHA-256 hash of a directory by hashing all regular files inside it", which is now a description of `Sha` rather than of anything this method does.

Its only caller is `Cache.sha` at `Cache.java:116`, which branches on `Files.isDirectory` to reach it and calls `new Sha(any)` without a filter on the `Files.isRegularFile` branch. The two branches build the same object with different arguments for no reason a reader can recover from the code, and the stale Javadoc actively misleads whoever opens the file next looking for where a directory gets hashed.

Both branches collapse into one `new Sha(any, this.filter).toString()`, since a filtered walk over a lone file is the same walk, which removes `dirSha`, its stale Javadoc and the branch together and leaves the `IllegalArgumentException` guard intact. Note that this reading of the file-and-filter case depends on how #6610 is settled.

morphqdd:
Closes #6611. Depends on #6610 (Sha now guards the filter to directories only, so passing it through unconditionally is safe for both branches).

`Cache.dirSha` was a one-line delegator, `return new Sha(dir, this.filter).toString();`, with Javadoc still describing the loop it used to run before #6602. Its only caller, `Cache.sha`, built the same kind of object with different arguments on its two branches for no reason a reader could recover from the code.

Collapsed both branches into one `new Sha(any, this.filter).toString()` call and removed `dirSha` along with its stale Javadoc. The `IllegalArgumentException` guard for a path that's neither a file nor a directory stays.

Ran `CacheTest`, `ShaTest`, and the cache-consumer tests (`ConcurrentCacheTest`, `GlobalCacheTest`, `ChCachedTest`, `OyCachedTest`, `Fp*CacheTest`) locally, all green. `qulice:check -Pqulice` on `eo-maven-plugin` is clean.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
@@ -114,25 +114,11 @@ final class Cache {
      * @return Base64-encoded SHA-256 hash of the file or directory contents
      */
     private String sha(final Path any) {
-        final String result;
-        if (Files.isDirectory(any)) {
-            result = this.dirSha(any);
-        } else if (Files.isRegularFile(any)) {
-            result = new Sha(any).toString();
-        } else {
+        if (!Files.isDirectory(any) && !Files.isRegularFile(any)) {
             throw new IllegalArgumentException(
                 String.format("Path '%s' is neither a regular file nor a directory", any)
             );
         }
-        return result;
-    }
-
-    /**
-     * Calculate SHA-256 hash of a directory by hashing all regular files inside it.
-     * @param dir Directory path
-     * @return Base64-encoded SHA-256 hash of the directory contents
-     */
-    private String dirSha(final Path dir) {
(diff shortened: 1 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3745](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6620, opened 2026-08-11
#6612: assert the folder hash against Sha directly

### LINKED ISSUE #6612: `CacheTest.frame` duplicates `Sha`'s framing recipe in `CacheTest.java`
The `frame` helper in `eo-maven-plugin/src/test/java/org/eolang/maven/CacheTest.java:363` re-implements the way `Sha` frames a file inside a directory, the relative path, then a NUL, then the bytes, then `\0length\0`, so that `generatesCorrectHashForEntireFolderWithSeveralFiles` at `CacheTest.java:209` can predict the digest it asserts. The Javadoc on the helper says as much: "the same way `Sha` frames a file inside a directory".

That duplication makes the test blind to the very regression it appears to guard. Change the framing in `Sha` and the assertion fails, so whoever makes the change edits `frame` to match and the test passes again having verified nothing about the digest. The test also reaches past what its own name promises: it claims to check the hash `Cache` writes for a folder, and instead re-derives `Sha`'s internal recipe by hand.

Asserting `new Sha(source).toString()` states what `Cache` actually promises, that the marker file holds the `Sha` of the source, and deletes the helper along with the `MessageDigest` plumbing in that test. The properties of the digest itself, which trees must collide and which must not, already live in `ShaTest` where they belong. The helper arrived in #6602.

morphqdd:
Closes #6612. Depends on #6611 (touches the same test method just simplified there).

`CacheTest.frame` re-implemented the way `Sha` frames a file inside a directory, relative path, then NUL, then bytes, then `\0length\0`, so the folder-hash test could predict the digest it asserts. That duplication makes the test blind to a regression in `Sha`'s own framing: change it there and `frame` has to change to match, and the test passes again having verified nothing.

Replaced the hand-built digest with `new Sha(source).toString()`, asserting what `Cache` actually promises: the marker file holds the `Sha` of the source. Removed `frame` along with the `MessageDigest` plumbing it needed.

Ran `CacheTest` and `ShaTest` locally, both green. `qulice:check -Pqulice` on `eo-maven-plugin` is clean.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Cache.java
@@ -114,25 +114,11 @@ final class Cache {
      * @return Base64-encoded SHA-256 hash of the file or directory contents
      */
     private String sha(final Path any) {
-        final String result;
-        if (Files.isDirectory(any)) {
-            result = this.dirSha(any);
-        } else if (Files.isRegularFile(any)) {
-            result = new Sha(any).toString();
-        } else {
+        if (!Files.isDirectory(any) && !Files.isRegularFile(any)) {
             throw new IllegalArgumentException(
                 String.format("Path '%s' is neither a regular file nor a directory", any)
             );
         }
-        return result;
-    }
-
-    /**
-     * Calculate SHA-256 hash of a directory by hashing all regular files inside it.
-     * @param dir Directory path
-     * @return Base64-encoded SHA-256 hash of the directory contents
-     */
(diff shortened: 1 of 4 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3761](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6621, opened 2026-08-11
Keep the bottom object's cause through a two-argument dispatch

### LINKED ISSUE #6359: `printf "%d"` on an out-of-range number fails with a bottom-object internal error instead of its documented message
## What happens

`printf`'s `%d` conversion has a documented out-of-range guard (`PrintfArgs.toLong`, `eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java` around line 296-317) meant to fail with "The number %s doesn't fit into long range for the '%%d' conversion" when the value doesn't fit in a `long`. That message never reaches the caller.

## Steps to reproduce

```
printf "%d" 9.3e18
printf "%d" 1e19
printf "%d" 1e30
```

## Observed

All three fail, but not with the documented message. Instead:

```
ExFailure: the ⊥ object only accepts a cause at position 0, not 1
```

This message comes from `PhTerminator.put` (`eo-runtime/src/main/java/org/eolang/PhTerminator.java` line 85), meaning something in the `%d` error-propagation path constructs a bottom object (`T`) and puts a value at attribute position 1 instead of 0 while trying to report the original "doesn't fit into long range" failure - so the internal object-construction problem masks the actual, documented error message.

`printf "%f" 1e30` triggers a related but different failure: a roughly 40-link `Error in "..."` chain ending in `the 'x' attribute must be a number` (`printf "%f" 1e20` and `1e25` work correctly, so this is specific to sufficiently large magnitudes).

## Expected behavior

`printf "%d"` on a value outside `long` range should fail with the documented, human-readable message ("The number %s doesn't fit into long range for the '%%d' conversion"), not with an unrelated internal error about bottom-object attribute positions. The object construction in the error-reporting path for this case needs to put its value at position 0, matching what `PhTerminator` expects.

Thayorns:
Two things had to change before `printf "%d"` on an out-of-range number could name its own reason, and the issue only showed the first one.

A propagated bottom answers every `take` with itself, so it also receives the arguments of whatever dispatch comes next. `PhTerminator.put` accepted only position 0 and aborted everywhere else, which meant the first two-argument dispatch downstream of a bottom — `.if a b` being the common one in the stdlib — blew up on its second argument with "the ⊥ object only accepts a cause at position 0, not 1" and threw away the real reason the computation had stopped. A `put` away from position 0 is now ignored instead of rejected, so the cause survives the dispatch and surfaces at the outer dataization, which is what the class was documented to do all along. That alone brings back the two `scanf` reasons that were going the same way, "Unsupported format specifier" and the one about a dangling percent.

It did not bring back printf's, and the reason is worth spelling out. The `%d` guard rendered its own message through `%f`, and `as-fixed` refuses any magnitude of 2^53 or larger. Every number that guard rejects is far past that, so the message could never render — with the bottom fixed alone, the caller still got a chain of some forty `Error in "..."` links ending in "Can't write this number as decimal digits exactly". The guard now states its reason without interpolating the number, which is how `scanf` already words the same failure.

The test dataizes the real `string.printf` and asserts the documented reason, since that is the behaviour the issue is about. It lives in `PhTerminatorTest` rather than in a file of its own because jtcop wants every test class to name a hand-written production class, and `printf` is pure EO with none. Next to it, a one-line test records the new contract of `put`.

Two things are left as puzzles. The reason now carries no number at all, and it should carry one again once `%f` can print a magnitude that big. Separately, a bottom that carries no cause still adopts the first argument of the dispatch that follows it, because position 0 is both the cause slot and that argument's slot.

Closes #6359


Thayorns:
@yegor256 ping

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1165](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6622, opened 2026-08-11
#6599: reject a fractional slice start or end

### LINKED ISSUE #6599: `tuple.slice` silently accepts a fractional `start` and `end` instead of rejecting them
## What happens

`tuple.slice` clamps `start` and `end` to the bounds of the list (lines 11-34 of `eo-runtime/src/main/eo/tuple/slice.eo`) but never checks that they are integers. A fractional index goes straight into the comparison on lines 39-42:

```
      if. > [accum item index] >>
        and.
          index.gte s
          index.lt e
```

where it silently acts as a bound sitting halfway between two positions.

I checked this on the built runtime by adding a test to `slice.eo`, which passes:

```
eq. ++> zzprobe-accepts-a-fractional-start
  1
  length.
    slice
      *
        1
        2
        3
      0.5
      2
```

So `slice (* 1 2 3) 0.5 2` quietly returns a one-element tuple instead of refusing the index.

This is the same defect that was already fixed for `tuple.at` in #6057 and for `string.at` in #5981: an index that is not an integer but lands inside the numeric bounds must not be treated as a real index. `tuple.slice` was left out of that work.

## What should happen

`slice` should reject a `start` or an `end` that is not a finite integer and return a `cant-slice` fallback, the way `tuple.at` and `string.at` do.

morphqdd:
Closes #6599.

`tuple.slice` clamped `start` and `end` to the list bounds but never checked they were integers, so a fractional index like `0.5` sat silently between two positions. Same defect already fixed for `tuple.at` in #6057 and `string.at` in #5981.

Added a `start.is-integer.and end.is-integer` guard before the existing clamp-and-reduce logic, falling back to a bottom object (`T`) when either isn't an integer, matching the no-fallback-parameter shape `slice` already has.

Added `stops-on-a-fractional-start` and `stops-on-a-fractional-end` tests using the `slice --> stops-on-...` convention already used elsewhere in the runtime for a bottom-object rejection.

### DIFF
--- a/eo-runtime/src/main/eo/tuple/slice.eo
+++ b/eo-runtime/src/main/eo/tuple/slice.eo
@@ -10,38 +10,42 @@
 
 [list start end] > slice
   if. > @
-    gte.
-      number
-        if. >> s!
-          start.gte 0
+    start.is-integer.and end.is-integer
+    if.
+      gte.
+        number
+          if. >> s!
+            start.gte 0
+            if.
+              start.gte list.length
+              list.length
+              start
+            if.
+              list.length.neg.lte start
+              start.plus list.length
+              0
+        if. >> e!
+          end.gte 0
           if.
-            start.gte list.length
+            end.gte list.length
             list.length
-            start
+            end
           if.
-            list.length.neg.lte start
-            start.plus list.length
+            list.length.neg.lte end
+            end.plus list.length
             0
-      if. >> e!
-        end.gte 0
-        if.
-          end.gte list.length
-          list.length
-          end
-        if.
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3777](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6623, opened 2026-08-11
#6606: defer AtomSafe's Atom cast into the guarded lambda call

### LINKED ISSUE #6606: `AtomSafe` constructor casts to `Atom` without a guard in `AtomSafe.java`
`AtomSafe` casts its argument in the constructor, at `eo-runtime/src/main/java/org/eolang/AtomSafe.java:33`, with `this.origin = (Atom) atom;`. The constructor takes a plain `Phi`, so handing it an object that does not implement `Atom` throws a `ClassCastException` right there — before the try/catch in `lambda()` at lines 39-48, which is the only thing in the class that turns a failure into an `ExFailure`, ever runs.

`PhSafe.java:161` makes that escape reach the user. It reads `return this.through(new AtomSafe(this.origin)::lambda, ".λ");`, and the receiver of a bound method reference is evaluated when the reference is created, not when the functional method is invoked. So `new AtomSafe(this.origin)` runs before `through` is entered, and the try at lines 208-217 never sees the cast. All three `PhSafe` constructors accept any `Phi`, and the class docblock promises that "If, in any method invocation, a runtime error occurs, it is caught and wrapped into an `ExFailure` carrying the location of the error in the EO code." For `lambda()` on a non-atom origin that promise does not hold, and what surfaces is a bare Java `ClassCastException` with no EO coordinates on it.

The other way in is `PhDefault.take`, at `PhDefault.java:225-226`, where the `λ` branch builds `new AtomTyped(this, ...)` and calls `lambda()` on it with no type check, and `AtomTyped.lambda()` at line 56 then constructs an `AtomSafe` around it. Every other `λ` path in that same class guards first — `delta()` at line 260, `normalized()` at line 278 and `absent()` at line 394 all test `this instanceof Atom` before taking `λ` — so a `take("λ")` on a plain object is the one route that reaches the cast unchecked.

Failing fast inside `AtomSafe`, with an `ExFailure` naming the object and its `forma` instead of the raw cast, would cover both routes and keep the message locatable; moving the `new AtomSafe(...)` at `PhSafe.java:161` inside the supplier would additionally put it back under the guard that is already there.

@yegor256 please take a look.

morphqdd:
Closes #6606.

`AtomSafe`'s constructor cast its argument to `Atom` right away: `this.origin = (Atom) atom;`. `PhSafe.lambda()` calls it as `this.through(new AtomSafe(this.origin)::lambda, ".λ")`, and the receiver of a bound method reference is evaluated when the reference is created, not when the functional method runs. So `new AtomSafe(this.origin)` runs before `through` is entered, and a `Phi` that isn't an `Atom` throws `ClassCastException` right there, before the try/catch in `through` (the only thing that turns a failure into an `ExFailure`) ever sees it.

Moved the cast from the constructor into `lambda()` itself, inside the same try block that already catches everything else. The cast failure now happens where `through`'s try/catch can see it, same as any other failure from the wrapped atom.

Added `catchesNonAtomOriginOnLambda` to `PhSafeTest`, asserting `new PhSafe(new PhDefault()).lambda()` throws `ExFailure` (with the `.λ` location) instead of a raw `ClassCastException`.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/AtomSafe.java
+++ b/eo-runtime/src/main/java/org/eolang/AtomSafe.java
@@ -21,23 +21,23 @@ package org.eolang;
 public final class AtomSafe implements Atom {
 
     /**
-     * Original atom.
+     * Original object, expected to also be an atom.
      */
-    private final Atom origin;
+    private final Phi origin;
 
     /**
      * Ctor.
      * @param atom Phi as atom
      */
     public AtomSafe(final Phi atom) {
-        this.origin = (Atom) atom;
+        this.origin = atom;
     }
 
     // @checkstyle IllegalCatchCheck (12 lines)
     @Override
     public Phi lambda() {
         try {
-            return this.origin.lambda();
+            return ((Atom) this.origin).lambda();
         } catch (final InterruptedException ex) {
             Thread.currentThread().interrupt();
             throw AtomSafe.failure(ex);
--- a/eo-runtime/src/test/java/org/eolang/PhSafeTest.java
+++ b/eo-runtime/src/test/java/org/eolang/PhSafeTest.java
@@ -110,6 +110,19 @@ final class PhSafeTest {
     }
 
     @Test
+    void catchesNonAtomOriginOnLambda() {
+        MatcherAssert.assertThat(
+            "wraps a non-atom origin's cast failure into ExFailure instead of letting it escape",
+            Assertions.assertThrows(
+                ExFailure.class,
+                () -> new PhSafe(new PhDefault()).lambda(),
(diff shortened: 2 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (21) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3845](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6624, opened 2026-08-11
#6597: split empty text into one empty string

### LINKED ISSUE #6597: `string.split` returns an empty tuple for an empty text instead of one empty string
## What happens

`string.split` returns an empty tuple when the text is empty. Lines 11-16 of `eo-runtime/src/main/eo/string/split.eo` short-circuit on a zero-length text:

```
[text delimiter] > split
  if. > @
    eq.
      size. >> len!
        text >> bts!
      0
    *
```

I checked this on the built runtime by adding a test to `split.eo`, which passes:

```
eq. ++> zzprobe-empty-text-gives-zero-parts
  0
  length.
    split "" "-"
```

So the result really has zero elements.

That contradicts how the object treats empty parts everywhere else. The existing test `can-split-a-text-into-empty-strings` keeps both the leading and the trailing empty part, so `"-a-b-"` split by `"-"` gives four parts, two of them empty. By the same rule an empty text is one empty part, not zero parts. A caller that splits a line and reads `.head` gets a working result for every input except the empty one, where it terminates instead.

## What should happen

`split "" "-"` should return `* ""`, a tuple holding a single empty string.

morphqdd:
Closes #6597.

`string.split` short-circuited on a zero-length text and returned an empty tuple, zero parts. That contradicts how the object treats empty parts everywhere else: `can-split-a-text-into-empty-strings` keeps both the leading and trailing empty part of `"-a-b-"`, so an empty text should be one empty part, not zero parts, the same way it works in Ruby, Python and JS.

Changed the empty-text branch from `*` to `* ""`.

Added `can-split-an-empty-text-into-one-empty-string`.

### DIFF
--- a/eo-runtime/src/main/eo/string/split.eo
+++ b/eo-runtime/src/main/eo/string/split.eo
@@ -13,7 +13,7 @@
       size. >> len!
         text >> bts!
       0
-    *
+    * ""
     [accum start current] >> rec-split
       if. > @
         len.eq current
@@ -82,6 +82,12 @@
       "a"
       "b"
       "c"
+
+  eq. ++> can-split-an-empty-text-into-one-empty-string
+    split
+      ""
+      "-"
+    * ""
 
   eq. ++> can-split-into-strings
     length.


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3789](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6625, opened 2026-08-11
#6598: reject an empty delimiter through a cant-split fallback

### LINKED ISSUE #6598: `string.split` silently returns the whole text when the delimiter is empty
## What happens

`string.split` accepts an empty `delimiter` and silently returns the whole text as one part. Lines 21-31 of `eo-runtime/src/main/eo/string/split.eo` treat a zero-size delimiter exactly like "the delimiter does not fit at this position":

```
        if.
          or.
            size.eq 0
            gt.
              current.plus size
              len
          rec-split
            accum
            start
            as-number.
              current.plus 1
```

so the cursor walks to the end of the text and one part is appended at the end.

I checked this on the built runtime with two tests added to `split.eo`, both of which pass:

```
eq. ++> zzprobe-empty-delimiter-gives-one-part
  1
  length.
    split "abc" ""

eq. ++> zzprobe-empty-delimiter-keeps-whole-text
  "abc"
  head.
    split "abc" ""
```

An empty delimiter is not a valid split request, but the caller cannot tell it from a text that simply has no delimiter in it. The sibling `nsplit` rejects an invalid `limit` through its `cant-split` fallback (#6060), while `split` has no fallback attribute at all.

## What should happen

`split` should take a `cant-split` fallback, the way `nsplit` does, and return it when the delimiter is empty, with a message saying the delimiter must not be empty.

morphqdd:
Closes #6598. Depends on #6597 (both touch the same file's base cases).

`string.split` accepted an empty `delimiter` and silently walked to the end of the text, returning the whole text as one part. The caller has no way to tell that apart from a text that simply has no delimiter in it. The sibling `nsplit` already rejects an invalid `limit` through its `cant-split` fallback (#6060); `split` had no fallback attribute at all.

Gave `split` the same `cant-split` fallback, returned when the delimiter is empty with a message saying so. Checked the one other caller in the runtime, `path.eo`'s `string.split driveless separator`: it leaves the fallback unbound, the same way `n.mod 2` and `7.mod 3` already leave `cant-mod` unbound elsewhere in the runtime, which is safe as long as the error branch is never reached (the path separator is never empty).

Added `rejects-an-empty-delimiter` and `stops-on-an-empty-delimiter`. Ran `EOsplitTest` and `EOpathTest`, both green.

### DIFF
--- a/eo-runtime/src/main/eo/string/split.eo
+++ b/eo-runtime/src/main/eo/string/split.eo
@@ -1,4 +1,6 @@
 # Returns a `tuple` of `strings`: `text` separated by `delimiter`.
+# If `delimiter` is empty, the `cant-split` fallback is returned, or the
+# bottom object when unbound.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3805](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6626, opened 2026-08-11
Check EO sources concurrently in the format goal

### LINKED ISSUE #6263: Excessive build time for `mvn clean install -PskipITs` in CI severely impedes quick iteration per PR
I cannot iterate quickly on this repository because every PR triggers a CI build that takes 11 to 17 minutes for `mvn clean install -PskipITs` alone, and 18 to 32 minutes of total wall-clock time depending on the runner.

Measured from recent runs:

| Runner | `mvn` step |
|---|---|
| `ubuntu-24.04` | 11.2 min |
| `macos-15` | 13.9 min |
| `macos-15-intel` | 17.8 min |
| `windows-2022` | 16.4 min |

Example run: https://github.com/objectionary/eo/actions/runs/30746201831

It would be nice to investigate what dominates the build time and reduce it to under 5 minutes.

Thayorns:
`eo:format` is the single most expensive goal in the whole build. On the last master run it took 51 seconds inside `eo-runtime` alone — that is a fifth of the 4 minutes the Ubuntu leg spends end to end, and it grows with every `.eo` file we add. Two things made it that slow.

First, it walked its 170 sources one at a time, in a plain `for` loop, while every other goal in the plugin already hands its sources to `Threaded` and uses all the cores the runner has. The mojo is declared `threadSafe` and nothing in the check is shared state — the only thing the loop accumulated was a list whose size was all `report` ever looked at, so it collapses into the count `Threaded.total()` already returns.

Second, `canonical` parsed the same text twice for every file that was already formatted, which is nearly all of them: once to settle the moniker layout and again to lay the settled structure out with the configured weights. Parsing is by far the costliest step here, so the tree is now carried out of the settling loop instead of being thrown away and rebuilt. That is exactly one parse saved per source, on every source.

The remaining redundancy is across goals rather than inside this one: `format` parses each source and drops the tree, then `compile` parses the very same text again a few seconds later. Handing the settled tree over to `Parsing` would halve that too, and it is left as a puzzle.

Closes #6263

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
@@ -10,7 +10,6 @@ import com.jcabi.log.Logger;
 import com.jcabi.xml.XML;
 import java.io.IOException;
 import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.EnumMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (102 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1197](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6629, opened 2026-08-12
Walk every .eo source once per build

### LINKED ISSUE #6627: Every {@code .eo} source is parsed multiple times in MjFormat.java
The puzzle `6263-f987a25f` from #6263 has to be resolved: 

https://github.com/objectionary/eo/blob/700cf6e72c92b7943f7ccf7c0ded7bb5e8a0c7fe/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java#L47-L54

The puzzle was created by @yegor256 on 12-Aug-26. 

Thayorns:
Resolves the puzzle left in `MjFormat`: the `format` goal walked every `.eo` source with the parser, threw the tree away, and then the `parse` goal walked the very same text again seconds later, so a clean build of `eo-runtime` walked its 170 sources twice.

The two goals cannot share a finished tree, since they apply different XSL pipelines to it: `format` uses the objects-unaware `Canonical`, while `Parsing` uses one built from the identifiers of the local package, and homing bare references into the current package changes what the printer prints back. What they do share is the walk itself, which is the costly half and does not depend on the pipeline at all. So `Trees` keeps the walked tree keyed by the SHA-256 of the text that produced it, and `MjFormat` and `EoSource` each apply their own pipeline on top of it. When the source is already canonical, which is the normal case, the text `format` walked is the text `parse` reads, and the second walk is skipped.

Maven makes every goal as its own object, one after another, so the two goals share the memo the way `GcShared` shares the footprints of earlier builds: through the filesystem, with no static state anywhere. A tree waits under `target/eo/0-walk` in a file named after the SHA-256 of its text, `format` writes there ahead of every goal that reads, and `clean` takes the directory away. Nothing outlives the build that walked it, so the memo cannot grow without bound and the earlier heap puzzle is gone with it.

Closes #6627


eo-maven-plugin/src/main/java/org/eolang/maven/Trees.java
```diff
@@ -0,0 +1,112 @@
...
+     *  realm outlives one module and a settling pass leaves its
+     *  intermediate trees behind, so a build of the whole repository may
+     *  run out of heap.
+     */
+    final class TsShared implements Trees {
+
+        /**
+         * The one that every goal of one build shares.
+         */
+        static final Trees INSTANCE = new Trees.TsShared();
```
yegor256: @Thayorns read this: https://www.yegor256.com/2016/06/27/singletons-must-die.html

Thayorns:
@yegor256 fixed here

Thayorns:
@yegor256 CI green here

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2967](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:
@Thayorns what is the point of saving XML files again, if we already have them in the `1-parse` directory? Can't MjFormat simply take them from there?

Thayorns:
@yegor256 `1-parse` is not there yet when `format` runs, and even if it were, the XMIR in it is not the one `format` needs.

The `format` goal sits at `process-sources` and stands before `compile` in the execution of `eo-runtime/pom.xml`, so on a clean build it walks every source while `target/eo/1-parse` is still empty: `Parsing` runs later, inside `compile`, through `MjSafe.assembling()`. The two cannot be swapped either, because `format` may rewrite a source, and whatever `parse` had written before that rewrite would describe text that no longer exists on disk.

The contents differ as well. `Parsing` applies `new Canonical(objects)`, built from the qualified names of the local package, so `1-parse` holds XMIR in which bare references have been homed into the package. `MjFormat` applies the objects-unaware `new Canonical()`, because printing a homed tree back would put package prefixes into the source that the author never wrote, and `format` would then rewrite every file in the project. On top of that `format` settles: it prints the tree back and walks the printed text again, and that intermediate text is nowhere in `1-parse` at all.

What `0-walk` keeps is neither of those two XMIRs. It is the tree straight out of `EoSyntax` with no pipeline over it, which is the costly half both goals share and the only part of the work that does not depend on which `Canonical` is asking. Each goal lays its own pipeline over it afterwards.

yegor256:
@maxonfjvipon need your opinion here. do you think we need this XML-passing feature between mojos?

maxonfjvipon:
@Thayorns @yegor256 maybe we can just move `format` phase so it goes after parsing?

Thayorns:
@maxonfjvipon done, `format` now runs after `parse` and reads what `parse` wrote.

`eo-runtime` binds the `parse` goal ahead of `format`, and `MjFormat` takes the XMIR out of `1-parse` instead of walking the source itself. `Trees` and the `0-walk` directory are gone, so nothing is written twice any more. A project that does not bind `parse` ahead of `format` still gets the same answer, it just walks the source there; `TjForeign.notParsed()` already decides which of the two it is.

One thing had to be settled to make the tree readable by this goal. `Parsing` applies `Canonical(objects)`, so a bare reference to a same-package object arrives homed: `array` in a program declaring `+package bytes` becomes `Φ.bytes.array`. The printer must not put that prefix back into a source that never carried it, so `unhome-package.xsl` takes it off again, inverting `add-default-package.xsl` for the program's own package and nothing else. A reference written out in full (`bytes.array` inside package `bytes`) lands on that very same base, so it now formats to the bare `array`; the prefix is only ever added when the qualified name is a local package object, so the two spellings name one and the same object.

The `-Deo.autoFix` case holds up too: a source this goal rewrites is newer than its XMIR afterwards, so the `Parsing` inside `compile` picks it up again and nothing downstream reads the stale tree.

Thayorns:
@maxonfjvipon @yegor256 I implemented it and CI says it cannot work. I moved `format` behind `parse`, had `MjFormat` read `1-parse` and wrote the sheet that takes the package prefix back off, and 19 of the 171 `eo-runtime` sources came out rewritten. I have reverted it here so this branch stays green; the attempt is on `6627-parse-reuse-attempt` (3face92e6f) if you want to look at it.

The reason is that `Canonical(objects)` is not injective, so there is no sheet that can undo it. Three of the ways it bit, all from the CI log:

`input/copied.eo` writes `Q.input.length` in full. A bare `length` in that same file would also parse to `base="Φ.input.length"`, because `input.length` is a local package object, so the two spellings arrive identical and whichever one I print back rewrites the other. It came out as `length`. Same shape in `string/regex.eo`, where `/Q.string.regex.pattern` came out as `/Q.regex.pattern`.

`tuple/back.eo` writes `*` for the empty tuple, which is `Φ.tuple.empty`. That is the global `empty` attribute of the root `tuple` object, and it is indistinguishable from a reference homed into the `tuple` package, because the package is named `tuple` too. Stripping the prefix turned `*` into `empty`.

This is what the objects-unaware `Canonical()` buys `format`: with it, `Q.input.length` stays `Φ.input.length` and bare `length` becomes `Φ.length`, so the printer can tell them apart and print each back as written. `1-parse` has already lost that distinction by the time it is on disk, and the goal cannot get it back from there.

So it comes down to what you would rather have. `add-default-package.xsl` could mark the references it homes, which would make the inverse exact and would also let `to-eo-tree.xsl` drop the heuristic it carries today for the same problem, though it puts a new attribute into XMIR. Or `format` could read `1-parse` only for programs with no `+package`, where the two pipelines provably agree, which is 50 of the 170 sources here and leaves the other 120 walking twice. Or the walked tree keeps being passed the way this branch does now. I do not think this one is mine to choose, so tell me which and I will do it.

yegor256:
@Thayorns maybe we should simply fix EO code?

Thayorns:
@yegor256 I left the EO sources alone and kept the walked tree in this branch, here is why.

Two of the three shapes would fix cleanly: `Q.input.length` becomes `length`, `/Q.string.regex.pattern` becomes `/Q.regex.pattern`, the same objects written shorter. The third one does not. `*` in `tuple/back.eo` is the only spelling of an empty tuple and comes back as `empty` after the round trip, so fixing the source means writing `empty` there and never writing `*` inside the `tuple` package again. Every package named after a root object, `bytes`, `string`, `number`, carries the same trap. That is a rule about how EO may be written, and all it buys is one walk over the sources.

So `format` walks, hands the tree over through `target/eo/0-walk`, and `parse` picks it up instead of walking again. The sources stay as their authors wrote them, `clean` takes the directory away, and nothing outlives the build. The branch is +130/-20 and CI is green.

yegor256:
@Thayorns looks like a bug in the printer:

> * in tuple/back.eo is the only spelling of an empty tuple and comes back as empty after the round trip

Maybe let's fix the printer?

yegor256:
Several CI checks are failing on this PR: `mvn (macos-15, 26)`, `integration`, `qulice`, `jacoco`, `snippets`, and `site`. Could you please investigate and fix these failures?

---

PR #6630, opened 2026-08-12
Cache the folding point of arc-sine so the square root runs once

### LINKED ISSUE #6224: `number.arc-sine` and `number.arc-cosine` are ~46x slower for `|num| > 0.5` because the folding point is recomputed every iteration
## What happens

`number.arc-sine` (and `number.arc-cosine`, which delegates to it) takes tens of seconds to return for any argument with `|num| > 0.5`, while an argument with `|num| <= 0.5` returns almost instantly. Both paths run the very same 30-term Taylor series, so the slowdown is not the series length; it is that the folding point is recomputed on every iteration. The affected range `(0.5, 1]` is exactly what the module's own tests exercise (`arc-sine 0.6`).

### Runtime evidence

Same single-shot JVM, one `arc-sine` call each, result correct in both cases:

| call | branch | time | result error vs `Math.asin` |
| --- | --- | --- | --- |
| `arc-sine 0.5` | `magnitude <= 0.5` (point = `magnitude`) | **454 ms** | 0 |
| `arc-sine 0.6` | `magnitude > 0.5` (point = `sqrt(...)`) | **21122 ms** | 2.2e-16 |

Both evaluate a 30-term series; `0.6` is about 46x slower. Values closer to 1 (`0.9`, `0.999`) are slower still and time out past 20 s. `arc-cosine` inherits this, and `sqrt`/`exp`, which also route through `power`, are slow for the same underlying reason.

## Root cause

`eo-runtime/src/main/eo/number/arc-sine.eo`, the `series` object (around lines 27-46):

```
[point] >> series
  point.times > @
    poly 1
  if. > [depth] > poly
    depth.gt 30
    1
    1.plus
      times.
        div.
          times.
            times.
              minus. (depth.times 2) 1
              minus. (depth.times 2) 1
            squared
          times.
            depth.times 2
            plus. (depth.times 2) 1
        poly (depth.plus 1)
  point.times point > squared
| (sqrt ((1.minus magnitude).div 2))
```

`point` is the void attribute of `series`, bound by application to `sqrt((1 - magnitude) / 2)`. It is not memoised, so every textual reference to `point` re-dataizes that `sqrt` (itself a `power` call, which is expensive). `squared` (`point.times point`) is likewise not a constant, and `poly` reads `squared` at each of its ~30 recursion levels, so `point` is evaluated on the order of 60 times per call. When `point` is a cheap literal (the `magnitude <= 0.5` branch passes `magnitude` directly) the repetition is invisible; when it is `sqrt(...)`, the cost is multiplied by ~60.

## Suggested fix

Memoise the folding point and its square once, so the `sqrt` runs a single time instead of once per iteration. For example bind them as constants at the `series` level:

```
point.times point >> squared!
```

and, since even the `point` reference re-dataizes the argument, cache the point itself as a constant the series reads (rather than referencing the void attribute directly in both `@` and `squared`). With `squared` and the point cached, `arc-sine 0.6` drops from ~21 s to the same range as `arc-sine 0.5`. Please note the known `const !` double-evaluation caveat for chained numeric operations when placing the `!`; a single bound constant read by both `@` and `squared` is the safe shape.

The fast-vs-slow contrast above (identical iteration count, differing only in the cost of `point`) isolates the cause and can serve as a timing regression check.

Thayorns:
The `series` object inside `arc-sine` binds its folding point as a void attribute, and a void attribute is re-dataized at every textual reference. For `|num| <= 0.5` the point is just `magnitude`, so nobody noticed. For `|num| > 0.5` the point is `sqrt((1 - num) / 2)`, and since `squared` reads it twice while `poly` reads `squared` once per each of its thirty levels, that square root ran on the order of sixty times per call. Hence the 454 ms versus 21 s gap in the issue.

The point is now dataized once into a constant that both `@` and `squared` read, and `squared` is a constant too, so the square root runs exactly once regardless of how deep the series goes.

The new tests take the arc sine of 0.9, -0.9 and 0.999, and the arc cosine of 0.999. All four go through the folded branch, and before this change they were the cases that ran for tens of seconds or timed out; the two existing 0.6 tests were the only cover that branch had.

`power` has the same defect: `ipow` reads `half` twice so each halving recomputes its own subtree, and `expon` reads `k` once per each of its twenty `fraction` levels, re-running `ln` every time. That is what keeps `sqrt` and `exp` slow. It does not block this task, so it is filed separately as #6651 rather than widening this change.

Closes #6224


Thayorns:
@yegor256 ping

eo-runtime/src/main/eo/number/power.eo
```diff
@@ -3,6 +3,11 @@
 # results stay exact, and a fractional exponent of a positive base is taken as
 # e raised to `x` times the natural logarithm of the base. Zeros, infinities and
 # nan follow the same IEEE rules that Math.pow obeys.
+#
+# @todo #6224:30min Memoise the repeated halves below. Neither `half` in `ipow`
```
yegor256: @Thayorns puzzles are not bug reports. Puzzles are incomplete functionality markers. This problem prevents you from completing the task?

Thayorns: You are right, it does not block this task, so it was not a puzzle. I dropped it and filed the `power` defect as #6651 instead.

Thayorns:
@yegor256 fixed, take a look

Thayorns:
@yegor256 CI green here

yegor256:
@Thayorns conflict here

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2974](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Thayorns:
@yegor256 ping

0crat:
@thayorns Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1348](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6642, opened 2026-08-12
#6600: fail loudly when the write syscall returns -1

### LINKED ISSUE #6600: `console.write`, `stdout` and `stderr` ignore the return code of the `write` syscall
## What happens

Every write path in the console objects dataizes the result of the `write` syscall and throws it away.

`eo-runtime/src/main/eo/console.eo`, lines 131-138 (the same shape is at lines 95-102 for win32):

```
              seq * > [buffer] > write
                code.
                  posix *1
                    "write"
                    posix.stdout
                    buffer
                    buffer.size
                output-block
```

`eo-runtime/src/main/eo/stderr.eo` lines 46-53 repeat it, and `stdout.eo` lines 18-21 then return `true` unconditionally:

```
[text] > stdout
  seq * > @
    console.write text
    true
```

`write` returns -1 when it fails and a count smaller than `buffer.size` on a partial write. Neither case is looked at, so writing to a closed pipe, a full disk or a stream interrupted by a signal is reported as success and the output is silently lost.

The read side of the same object does check its code, `console.eo` lines 112-116:

```
                  if. >> chunk!
                    received.code.eq -1
                    T
                      "Failed to read from standard input"
                    received.output
```

## What should happen

The write path should read the code the same way the read path does: terminate with a message when the code is -1, and keep writing the remaining bytes when the code is smaller than `buffer.size`.

morphqdd:
Every write path in `console.write` and `stderr.write` (win32 and posix) dataized the `write` syscall's return code and threw it away. `write` returns -1 on failure, so writing to a closed pipe, a full disk, or a stream interrupted by a signal was reported as success while the output was silently lost.

Fixed by checking the code the same way the read side already does: -1 now panics with a clear message instead of continuing as if the write succeeded.

Partial writes (code smaller than buffer.size) are a separate piece of work, left as a `@todo` puzzle in `console.eo` referencing this issue, since retrying the unwritten tail needs its own design and tests.

Ran the existing `EOconsoleTest`, `EOstdoutTest`, `EOstderrTest` and `qulice:check -Pqulice` locally, all green.

Closes #6600

### DIFF
--- a/eo-runtime/src/main/eo/console.eo
+++ b/eo-runtime/src/main/eo/console.eo
@@ -55,6 +55,12 @@
 # ```
 #
 # Sequential reuse maintains the console state across operations.
+#
+# @todo #6600:30min Keep writing the remaining bytes on a partial write. Right
+#  now `write` panics only when the `write` syscall returns -1; a successful
+#  write of fewer bytes than `buffer.size` is accepted as if the whole buffer
+#  went out, same as `stderr.write`. Both should retry with the unwritten
+#  tail until the buffer is fully flushed or a syscall fails.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4068](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6645, opened 2026-08-12
Drop the dead index cell from the provides table

### LINKED ISSUE #6644: `Provides` wrongly adds unused `index` cell to each row in `provides.xml`
`Provides` at `eo-inference/src/main/java/org/eolang/inference/Provides.java:81` writes a cell nobody reads:

```java
rows.add(owner)
    .set("index", Integer.toString(seen))
    .set("complete", Boolean.toString(whole));
```

The cell existed while the view sorted rows by it. It stopped existing in #6551, which deleted it from all four clues together with both sorts, because `TjDeferred` hands the rows back in the order they arrived and the position of a row is meaningful on its own. #6592 was branched before that and restored the cell in this one clue, so master now has it in `provides.xml` and nowhere else.

That is 8,793 `index="…"` cells in a document written for a human to read, in a table whose three siblings have none:

```xml
<type complete="false" id="Φ.string" index="5229">
```

`grep -rn '"index"' eo-inference/src/main eo-maven-plugin/src/main` finds the two writes above and no reads. `Grouped` has no comparator left, and `Provided`, which is what the checking loop reads the table through, keeps rows in the order they came and never looks at the cell.

Dropping both `set("index", …)` calls and the counter that feeds them puts the table back where #6551 left it. The packs in `eo-maven-plugin/src/test/resources/org/eolang/maven/inference-packs` assert nothing about the cell, so nothing else has to move.

maxonfjvipon:
`Provides` was writing a cell nobody reads:

```java
rows.add(owner)
    .set("index", Integer.toString(seen))
    .set("complete", Boolean.toString(whole));
```

It was needed while `Grouped` sorted rows by it. #6551 deleted it from all four
clues along with both sorts, because `TjDeferred` hands the rows back in the
order they arrived and a row's place is meaningful on its own. #6592 branched
before that and restored the cell here, so master carried it in `provides.xml`
and in none of its three siblings — 8,793 of them in a document meant to be
read by a human.

Nothing else moves: no pack asserts the cell, `Grouped` has no comparator left,
and `Provided`, which is how the checking loop reads this table, keeps the rows
in the order they came.

Closes: #6644

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Provides.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Provides.java
@@ -73,25 +73,19 @@ final class Provides implements Clue {
     @Override
     public void follow(final Path xmirs, final Path tables) throws IOException {
         try (Tojos rows = new TjDeferred(new MnMemory())) {
-            int seen = 0;
             for (final XML formation : new Xmirs(xmirs).formations()) {
                 final String owner = formation.xpath("@loc").get(0);
                 final boolean whole = formation.nodes("o[@name='λ' or @name='φ']").isEmpty();
-                rows.add(owner)
-                    .set("index", Integer.toString(seen))
-                    .set("complete", Boolean.toString(whole));
-                seen = seen + 1;
+                rows.add(owner).set("complete", Boolean.toString(whole));
                 for (final XML attr : formation.nodes("o[@name and not(@name='λ')]")) {
                     final String name = attr.xpath("@name").get(0);
                     final Tojo row = rows.add(String.join(" ", owner, name))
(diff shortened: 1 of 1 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1040](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6646, opened 2026-08-12
#6608: escape control characters in PhApplication string literals

### LINKED ISSUE #6608: `PhApplication` leaves most control characters unescaped in `PhApplication.java`
`PhApplication.string()` at `eo-runtime/src/main/java/org/eolang/PhApplication.java:125-132` escapes a string literal with a chain of seven replacements — backslash, double quote, `\b`, `\f`, `\n`, `\r` and `\t`. Every other control character goes into the φ-term as itself: `0x00` through `0x07`, the vertical tab `0x0B`, `0x0E` through `0x1F` (which includes ESC), and DEL. That is twenty-eight characters emitted raw.

The other printer of the same thing does it correctly. `PhDefault.φTerm()` at `PhDefault.java:336` hands the bytes to `new Quoted(raw)`, which spells each glyph through `Escaped` (`Escaped.java:43-51`): the same seven, plus `\u000b` for the vertical tab and `\uXXXX` for everything below `0x20` and for `0x7F`. So one string prints two different ways depending on which renderer reaches it, and the `PhApplication` way does not survive a round trip — the `Escaped` docblock states the point is to emit the escape the parser decodes back into the very same glyph (R-9.7). A raw ESC in a printed φ-term also reaches the terminal as a control sequence rather than as text.

This is the remainder of #6375, whose title names control characters. Its sibling #6450 reported the same defect against `PhDefault` and was fixed by routing through `Quoted`; #6375 was instead closed with the seven hand-written replacements, and `PhApplication` imports neither `Quoted` nor `Escaped`.

Spelling the decoded text one glyph at a time through `Escaped`, the way `Quoted` already does, closes the gap and leaves the two printers with a single escaper between them. Worth keeping the strict decoder above it at lines 113-116, since that is what lets this method return null and fall back to the generic rendering on malformed UTF-8.

@maxonfjvipon @yegor256 please take a look.

morphqdd:
`PhApplication.string()` escaped only the seven hand-written characters (backslash, quote, \b, \f, \n, \r, \t). Every other control character (0x00-0x07, vertical tab, 0x0E-0x1F including ESC, and DEL) went into the φ-term raw, unlike `PhDefault.φTerm()`, which already routes through `Quoted`/`Escaped`.

Replaced the hand-written replace chain with `new Quoted(bytes).get()`, the same escaper `PhDefault` uses. The strict UTF-8 decode stays in place so malformed input still falls back to the generic rendering.

Added a test for a control character outside the original seven (ESC, DEL).

Closes #6608

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhApplication.java
+++ b/eo-runtime/src/main/java/org/eolang/PhApplication.java
@@ -120,17 +120,7 @@ public final class PhApplication extends PhOnce {
         }
         final String result;
         if (text.isPresent()) {
-            result = String.format(
-                "\"%s\"",
-                text.get()
-                    .replace("\\", "\\\\")
-                    .replace("\"", "\\\"")
-                    .replace("\b", "\\b")
-                    .replace("\f", "\\f")
-                    .replace(String.valueOf('\n'), "\\n")
-                    .replace(String.valueOf('\r'), "\\r")
-                    .replace("\t", "\\t")
-            );
+            result = new Quoted(bytes).get();
         } else {
             result = null;
         }
--- a/eo-runtime/src/test/java/org/eolang/PhApplicationTest.java
+++ b/eo-runtime/src/test/java/org/eolang/PhApplicationTest.java
@@ -133,6 +133,21 @@ final class PhApplicationTest {
     }
 
     @Test
+    void escapesControlCharactersOutsideTheSevenHandwritten() {
+        MatcherAssert.assertThat(
+            "String φ-term must escape control characters like ESC and DEL, but it didn't",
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (27) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3873](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6647, opened 2026-08-12
`Worklist.decide` forgets the type of a computed object it just processed


maxonfjvipon:
`Worklist.decide` at `eo-inference/src/main/java/org/eolang/inference/Worklist.java:134` works out what a dispatch turns out to be, uses the answer once, and forgets it:

```java
final String owned = this.given.attribute(has, need.get("name"));
...
if (!owned.isEmpty()) {
    todo.add(
        String.join(" ", this.name(owned), need.getOrDefault("type", ""))
    );
}
```

`need.get("type")` is the locator of the dispatch itself — the computed object — and `owned` is what taking that attribute gives. In other words, at that line the loop knows the very fact the tables lack: this computed object is a copy of that one. It puts the pair into the to-do list and writes nothing down, so every later question about that locator is answered by an empty table.

The cost is not theoretical. An object that delegates binds `φ` to a computed object far more often than to a named one: `x.next.foo > @` binds it to a dispatch. On a clean `eo-runtime` build of master the provides table holds 2,505 types, 2,432 of them bind `φ`, only 295 of those decoratees have any row at all, and exactly one of them, `Φ.tuple.with`, is complete. So looking behind `φ` (#6643) buys one judgeable object out of 2,432, and the checker still reports nothing on the whole runtime — not because the chain is wrong, but because it ends at a locator nothing describes.

Record the fact instead of dropping it: when a check resolves an attribute, add the pair to the links the loop reads, so the computed object goes by the name of what it is a copy of, exactly as a reference does. Then `Φ.t` delegating to `x.next` resolves to whatever `next` holds, completeness travels along it, and the sweep starts to speak.

One consequence has to come with it, and it is the interesting part. `Worklist`'s javadoc says a check nobody can decide now will never be decidable later, "since all four tables were written before it started" — that stops being true the moment the loop learns facts of its own. A check that failed for want of a type must go back into the list rather than being dropped, and the loop must stop when a full pass adds nothing new. The design note describes exactly that, and it was left out on purpose while there was nothing to wait for.

maxonfjvipon:
Parked with the checking half in #6661. Recording what a computed object turns out to be worked — 126 rows on the runtime, 1,389 with #6643 — and still moved no verdict, because the loop only learns along the path a check walks. It returns when the loop does.

0crat:
[...]

Issue #6638, opened 2026-08-12
`eo:multi-referenced` crashes Saxon with `getLength() not available` in `inline-cactoos.xsl`


Thayorns:
The `eo:multi-referenced` function in `eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:479` decides whether a handle is referenced more than once by calling `count(eo:references($target, $name)) > 1`. Saxon backs the sequence that `eo:references` returns with a `MemoSequence`, and counting it asks the sequence for its length, which that class refuses: `java.lang.UnsupportedOperationException: getLength() not available in class net.sf.saxon.om.MemoSequence$ProgressiveIterator`. The template rule at line 314 that calls the function then fails with `Internal error evaluating function eo:multi-referenced`, and the identity template at line 508 fails behind it as the error travels back up.

This kills the build rather than misprinting anything: on master at `ecfe7eced6` the `runtime (23)` job stops at `eo-runtime` with a `MojoFailureException`, see https://github.com/objectionary/eo/actions/runs/31572339577/job/94036826879. The failure is not constant — the same commit prints all 170 EO sources canonically on the other runners — because whether Saxon materialises the sequence or hands back a progressive iterator depends on how it decides to evaluate that call, so the printer works until the day it does not.

Asking for the second item instead of the size avoids the length call altogether and short-circuits: `exists(eo:references($target, $name)[2])` answers the same question without ever counting. The neighbouring `eo:unreferenced` at line 506 already leans on `empty()` for the same sequence and never trips over this.

@maxonfjvipon @yegor256 please take a look.

Thayorns:
@yegor256 ping

Thayorns:
@yegor256 ping

0crat:
@thayorns Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+1300](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

VasilevNStas:
@yegor256 @Thayorns 

[...]

PR #6648, opened 2026-08-12
Look behind the decoratee for an attribute

### LINKED ISSUE #6643: `Provided` method in `eo-inference` doesn't check behind the `φ` attribute, reducing checker effectiveness
`Provided` at `eo-inference/src/main/java/org/eolang/inference/Provided.java:60` decides whether the checker may judge a type by reading one cell of one row:

```java
boolean complete(final String type) {
    final Collection<Boolean> flags = new ArrayList<>(1);
    for (final Map<String, String> row : this.own(type)) {
        if (row.containsKey("complete")) {
            flags.add(Boolean.parseBoolean(row.get("complete")));
        }
    }
    return !flags.isEmpty() && !flags.contains(false);
}
```

Since #6592 that cell is `false` for every formation that binds `φ`, and rightly so: what stands behind the decoratee answers for every name the object does not bind itself, and this table knows nothing about it. But nothing anywhere looks behind `φ` either, so "not seen whole" is permanent, and the loop that #6615 added reports only what it is sure about. On a clean `eo-runtime` build of master (7af3fa9098) the provides table holds 2,505 types of which **48 are complete**, 2,432 rows name `φ` as an attribute, and 26,377 pending checks drain into an empty `problems.xml`. The checker cannot be wrong because it cannot speak.

The 48 it may still judge are the accidents of the corpus — `Φ.console.a🌵68-4`, `Φ.bool.+can-compile-deeply-duplicated-names.long-object-name` and other test-local formations that happen to bind nothing. Everything a program actually uses delegates: `Φ.number`, `Φ.string` and `Φ.bytes` are all `complete="false"`, so a dispatch on any of them is unanswerable, and the whole runtime passes without a word.

The same class already looks in two places for an attribute — the type itself, then its package, since an attribute nobody binds falls through to the object of that name beside it — and its javadoc ends by naming the missing third: "Delegation through `φ` is a third place, and is not looked into yet." That is the fix, and it is the growth point the design note names first. When a type binds neither the attribute nor a package member of that name, take the type its `φ` attribute holds, resolve it through `Same` to the name it goes by, and ask the same question there. Completeness walks the same path: a type that delegates has been seen whole once whatever it delegates to has been seen whole, with a guard so an object that delegates to itself is not walked forever.

That keeps `Provides` exactly as #6592 left it — the flag stays the honest statement about one formation, and joining it across a link stays the business of whoever reads the links. It should also be expected that the sweep starts speaking again, and that some of what it says will be wrong; measuring that is what the mutation benchmark is for, and narrowing it is what giving a copy its own types is for.

maxonfjvipon:
`Provided` looked for an attribute in two places, the type itself and its
package, and its javadoc ended by naming the missing third: "Delegation through
`φ` is a third place, and is not looked into yet." Now it is. When a type binds
neither the attribute nor a package member of that name, the type behind its
`φ` is resolved through `Same` and asked the same question. Completeness walks
the same path — an object that hands its answers to one that was seen whole was
seen whole too — and both walks stop at a type they have already passed, so an
object that delegates in a circle is walked once and answers nothing.

Four packs describe it, and they come in a pair that pins the behaviour from
both sides: the same four files ask `x.foo` and `x.next` of a `t` that binds
neither and delegates to a `base` that binds only `next`. The first is
reported, the second is not. The other two are about the doubt travelling along
the chain: a `t` that delegates to an atom stays unjudged, and a `t` whose `φ`
leads back to itself is walked once.

---

Now the part worth reading, because it is not what I expected. On the runtime
this changes the sweep by **nothing at all** — `problems.xml` was empty before
and is empty after — and the reason is a good one:

```
types in provides       : 2505
complete=true           : 48
delegators (bind φ)     : 2432
  decoratee has a row   : 295
  decoratee is complete : 1     (Φ.tuple.with)
```

Following `φ` makes exactly one more object judgeable. Not because the walk is
wrong — the packs show it working — but because a decoratee is almost never a
named object. It is a computed one: `x.next.foo > @` binds `φ` to a dispatch,
and no table says what a dispatch turns out to be, so the chain ends at a
locator nothing describes. Of 2,432 objects that delegate, only 295 hand over
to something the tables have a row for.

So this is one of the three places to look, tested and in place, and the
measurement it made possible says where the next one is: nothing writes down
the type of a computed object, while the loop works that out every time it
resolves an attribute and then throws the answer away. That is #6647, and it is
also where a check that cannot be decided has to start being parked again,
since facts will begin to appear while the list is being drained.

Closes: #6643

### DIFF
--- a/eo-maven-plugin/src/test/resources/org/eolang/maven/inference-packs/mistake-behind-delegation-is-reported.yaml
+++ b/eo-maven-plugin/src/test/resources/org/eolang/maven/inference-packs/mistake-behind-delegation-is-reported.yaml
@@ -5,6 +5,14 @@
 # whole and has no `foo` anywhere in it. So `foo` will never be there, and the
 # loop says so: what stands behind `φ` is the third place an attribute can come
 # from, after the object itself and its package.
+#
+# The four tables below are the whole basis of that verdict, and none of them
+# holds it alone. The provides rows say `t` binds only `φ` and was not seen
+# whole on its own, while `base` was and has nothing but `next`. The links row
+# says what that `φ` of `t` is, which is what turns "not seen whole" into "seen
+# whole through `base`". The needs row says somebody asks `foo` of whatever
+# fills the void `x`, and the checks row says a `t` is what fills it. Read
+# together they leave one answer, and the loop writes it down.
 eo:
   app.eo: |
     [] > app
@@ -18,5 +26,17 @@ eo:
   base.eo: |
     [] > base
       [] > next
+provides:
+  - "/provides/type[@id='Φ.t' and @complete='false']/attr[@name='φ' and @type='Φ.t.φ']"
+  - "/provides/type[@id='Φ.base' and @complete='true']/attr[@name='next']"
+  - "/provides/type[@id='Φ.base'][not(attr[@name='foo'])]"
+  - "/provides/type[@id='Φ.inc']/attr[@name='x' and @void='true']"
+links:
+  - "/links/type[@id='Φ.t.φ' and @copy='Φ.base']"
+  - "/links/type[@id='Φ.app.φ.α0' and @copy='Φ.t']"
+needs:
+  - "/needs/type[@id='Φ.inc.φ.ρ']/attr[@name='foo' and @type='Φ.inc.φ']"
+checks:
+  - "/checks/type[@id='Φ.app.φ']/attr[@name='α0' and @type='Φ.app.φ.α0']"
 problems:
   - "/problems/type[@id='Φ.t']/attr[@name='foo']"
--- a/eo-maven-plugin/src/test/resources/org/eolang/maven/inference-packs/mistake-behind-delegation-is-reported.yaml
+++ b/eo-maven-plugin/src/test/resources/org/eolang/maven/inference-packs/mistake-behind-delegation-is-reported.yaml
@@ -1,0 +1,22 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# `t` binds nothing itself and hands everything to `base`, which has been seen
+# whole and has no `foo` anywhere in it. So `foo` will never be there, and the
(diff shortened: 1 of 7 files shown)


maxonfjvipon:
The red `runtime (23)` here is not this branch: the job spent its whole 25-minute budget on the same commit master itself times out on, and the same workflow has ended green in none of its last hundred runs on master. What kills it when it does end early is the printer, not the checker — #6650 (filed today, a Saxon cast in `restore-local-names.xsl`) and #6638. All four `mvn` jobs are green here.

eo-maven-plugin/src/test/resources/org/eolang/maven/inference-packs/mistake-behind-delegation-is-reported.yaml
```diff
@@ -0,0 +1,22 @@
...
+  inc.eo: |
+    [x] > inc
+      x.foo > @
+  t.eo: |
+    [] > t
+      base > @
+  base.eo: |
+    [] > base
+      [] > next
+problems:
```
yegor256, line 41: @maxonfjvipon can you show here what's in the `provides`, `checks`, and `needs` too?

maxonfjvipon:
Done — the pack now carries the rows the verdict rests on, from the four tables that program produces: `t` binding only `φ` and not seen whole on its own, `base` seen whole with `next` and no `foo`, the link that says what the `φ` of `t` is, the need for `foo` on whatever fills the void, and the check that says a `t` fills it. The comment above them reads them together.

maxonfjvipon:
Parked with the rest of the checking half in #6661 — the loop it improves is coming out of the module until every object can be given a type, so this waits with it rather than landing on code that is about to go. The measurement that decided it is in that ticket, and the branch stays.

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2960](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6652, opened 2026-08-12
Record what a computed object turns out to be

### LINKED ISSUE #6647: `Worklist.decide` forgets the type of a computed object it just processed

maxonfjvipon:
`Worklist.decide` worked out what a computed object turns out to be every time
it resolved an attribute, spent the answer on one check, and dropped it. It
hands it back now:

```java
if (!owned.isEmpty()) {
    learned.computeIfAbsent(
        need.getOrDefault("type", ""), made -> new HashSet<>(1)
    ).add(this.name(owned));
```

One draining cannot use what that draining finds — the check that needed the
fact may have been decided before the check that found it — so `Settled` drains
the list again with the new names added, and again, until a draining brings
nothing new. Only the last one writes verdicts: an earlier draining judges with
less than everything known, and a mistake found there may be no mistake by the
end. The javadoc's old promise that "nothing new is learned while the list is
drained" is now true only within a single draining, and says so.

A locator that came out two ways is not learned at all. The `x.next` in `use`
is one object in the program text, and `use` copied twice asks it of two
different things; both answers are true of their own copy and neither is true
of the locator. `Agreed` is where that is decided — the same honesty that keeps
an incomplete object from being blamed. The day a copy receives types of its
own, the two answers will belong to two locators and this disappears.

What the loop finds goes into a table of its own, shaped like the links a rule
wrote, because no rule could have written it:

```xml
<copies>
  <type id="Φ.use.φ.α0" copy="Φ.t.next"/>
</copies>
```

Three packs describe it: a mistake reachable only through a computed object
(which fails on master), the fact being written down, and the two-copy
disagreement — that last one I checked by disabling the rule and watching it
fail on its own while the other thirty-six stayed green.

---

Now the measurement, which is more interesting than the change.

On `eo-runtime` this writes 126 rows into `copies.xml` and reports nothing. With
#6648's φ-following applied on top locally, 1,389 rows — eleven times as many —
and still nothing, because the number of objects the checker may judge moves
from 48 to 49 out of 2,505.

Following the delegation chains says why:

```
objects that delegate                     : 2432
  chain ends at a locator no table describes : 2410
  chain ends at an object with no φ          :   22   (1 of them complete)
```

Chasing one of those endings found something small and fixable. `Φ.bool.and`
delegates to `Φ.bool.and.φ`, which is the dispatch `.if` on a receiver whose
base is `ξ.ρ` — the parent — and `Scope` never resolves `ξ.ρ`, because it looks
for a locator `<formation>.ρ` that nothing declares. Of the 412 `ξ.ρ`
references in `eo-runtime`, none are linked and all 412 are a dispatch
receiver, so the loop cannot say what taking `if` from one gives, and the body
of `and` has no type for this change to record. That is #6653, and it is the
next one to fix.

Closes: #6647

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Agreed.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Agreed.java
@@ -1,0 +1,50 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * The answers that came out one way.
+ *
+ * <p>The loop asks what a computed object turns out to be, and asks it once
+ * for every copy of the formation the dispatch stands in. One answer is a
+ * fact. Two are a locator that is not one object, and the honest thing to do
+ * with it is nothing at all — it goes by its own name, the tables say nothing
+ * about it, and every check about it stays undecided.</p>
+ *
+ * @since 0.68.0
+ */
+final class Agreed {
+
+    /**
+     * Every answer given, by the locator it was given about.
+     */
+    private final Map<String, Collection<String>> answers;
+
+    /**
+     * Ctor.
+     * @param given Every answer given, by the locator it was given about
+     */
+    Agreed(final Map<String, Collection<String>> given) {
+        this.answers = given;
+    }
+
+    /**
+     * The name every computed object goes by.
+     * @return The locators that came out one way, against that one answer
+     */
+    Map<String, String> names() {
+        final Map<String, String> found = new HashMap<>(this.answers.size());
+        for (final Map.Entry<String, Collection<String>> answer : this.answers.entrySet()) {
+            if (answer.getValue().size() == 1) {
+                found.put(answer.getKey(), answer.getValue().iterator().next());
+            }
+        }
+        return found;
+    }
+}
--- a/eo-inference/src/main/java/org/eolang/inference/Concluded.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Concluded.java
@@ -41,6 +41,18 @@ import java.util.Map;
  * prototype, and its verdicts are worth reading before they are worth
  * obeying.</p>
  *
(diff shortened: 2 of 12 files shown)


maxonfjvipon:
Parked with the rest of the checking half in #6661 — the loop it improves is coming out of the module until every object can be given a type, so this waits with it rather than landing on code that is about to go. The measurement that decided it is in that ticket, and the branch stays.

PR #6660, opened 2026-08-12
Revert the logarithmic while, master is red

yegor256:
This puts `while` back to the recursive version it had before #6596, and restores the `@todo` in `directory.eo` that came with it. Nothing else changes.

The `ec2` job on master is red because of #6596 — 7 failures and 3 errors in `EOwhileTest` and `EO_number.EOintegralTest`, see run 31583359814. Two things are wrong in that rewrite. `count.div 2 > half!` leaves `half` as bytes, since a const compiles down to `as-bytes`, so the next level of `stepped` divides bytes and dies on a raw IEEE-754 blob. And the loop only turns as a side effect of dataizing the argument of the final `body` application, so a body that ignores its argument never drives it at all — which is exactly the set of tests that fail, and why `integral` only passed for a single partition.

Two things worth knowing before #6550 gets another attempt. The `can-loop-five-thousand-times` test it added has never actually been green: our `Deadline` extension turns a JUnit timeout into a `TestAbortedException`, so a test slower than the deadline is reported as skipped rather than failed. And #6596 was merged with its own `runtime (23)` check red, which nothing else caught, because `mvn.yml` passes `-Deo.transpileTests=false` and so never runs EO unit tests at all. Locally on this branch `EOwhileTest` is 10/10 and `EOintegralTest` 11/11; I got 534 tests into a full run with no failures before my 8G heap gave out, so I am leaving the rest to CI, which has far more room.

### DIFF
--- a/eo-runtime/src/main/eo/directory.eo
+++ b/eo-runtime/src/main/eo/directory.eo
@@ -9,11 +9,10 @@
 # @todo #5751:30min Stop descending into a symbolic link to a directory. The
 #  `walk` below asks stat, which follows the link, so a link to its own ancestor
 #  recurses forever, while the `Files.walk` it replaced only listed such a link.
-# @todo #6550:60min Walk a directory in constant stack space. The `stepped`
-#  below still costs one frame per entry, even though `while` now nests only as
-#  deep as the logarithm of its cycles: the walk carries a tuple as its
-#  accumulator and `malloc` holds bytes only, so a cycle has nowhere to keep it.
-#  Joining the matched paths into one string and splitting it at the end fits.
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2858](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6653, opened 2026-08-12
`ξ.ρ` reference resolution fails in `Scope.outwards`, leaving parent dispatch untyped


maxonfjvipon:
`Scope.outwards` in `eo-inference/src/main/java/org/eolang/inference/Scope.java:88` resolves a `ξ.name` reference by walking out of the reference until it finds a formation that binds `name`:

```java
while (around.contains(".")) {
    around = around.substring(0, around.lastIndexOf('.'));
    final String candidate = String.join(".", around, name);
    if (this.formations.contains(around) && this.locators.contains(candidate)) {
        found = candidate;
        break;
    }
}
```

For `ξ.ρ` that walk asks whether some enclosing formation has a locator `<formation>.ρ`. Nothing declares a `ρ` — it is what an object sits in, not an attribute anybody writes — so the loop runs out of segments and answers with an empty string, and `Links` writes no row.

On a clean `eo-runtime` build the tables say how much that costs: 412 references carry `base="ξ.ρ"`, exactly none of them appear in `links.xml`, and all 412 are the receiver of a dispatch. Two of them are `and` and `or`:

```xml
<o base=".if" line="21" loc="Φ.bool.and.φ" name="φ">
  <o base="ξ.ρ" line="21" loc="Φ.bool.and.φ.ρ"/>
```

`Φ.bool.and.φ.ρ` is asked for `if`, and since nothing says what that receiver is, nothing can say what taking `if` from it gives. So `Φ.bool.and.φ` — the whole body of `and` — never gets a type, `and` never counts as seen whole, and no check about it is ever decided. The same holds for every object whose body reaches for a name that belongs to its parent, which in `eo-runtime` is most of them.

`ξ` is the object being formed, so `ξ.ρ` is the parent of the formation the reference stands in: walk out to the nearest formation, exactly as the loop above does, and then out once more to the formation around that one. `Φ.bool.and.φ.ρ` reaches `Φ.bool.and` and then `Φ.bool`, which is where `if` lives. A pack of two files, one calling a sibling through the parent, pins it.

0crat:
@maxonfjvipon Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+640](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6662, opened 2026-08-12
Park the checking half until objects have types

### LINKED ISSUE #6661: Park the checking half of `eo-inference` until every object can be given a type
The `eo-inference` module writes six tables. Three of them — `provides`, `needs`, `links` — record what the program says about itself. The other three — `checks`, and the `problems` the loop drains them into — decide whether the program is wrong. The deciding half should come out until the recording half is finished.

It reports nothing, and the numbers say it cannot yet. On a clean `eo-runtime` build the loop drains 26,377 checks and writes an empty `problems.xml`, because only 48 of 2,505 types have been seen whole; the other 2,432 delegate, and 2,410 of those delegation chains end at a locator no table describes. Walking behind `φ` (#6643) moves that 48 to 49. Recording what a computed object turns out to be (#6647) adds 126 rows, or 1,389 with both, and still moves nothing. The largest document the module writes is `checks.xml`, at 4.5 MB, and its whole yield is an empty file.

The deeper reason is that the loop learns along the path a check walks, and a check starts at an application. An object nothing applies in the same world is never reached at all — `Φ.bool.and` delegates to the dispatch `.if` on `ξ.ρ`, nobody in `eo-runtime` copies `and`, and so the body of `and` has no type. A type table built by walking call sites can never cover a library. The Python prototype does not work that way: it types every object in the world and reads diagnostics off the finished table, which is what makes checking a thin pass at the end rather than an engine in the middle.

So: delete `Checks`, `Concluded`, `Worklist`, `Provided`, `Same`, `Ungrouped` and `Row`, along with the packs that describe verdicts, and leave the three rules that write down facts. `Inferring` follows `Clues` directly and the module stops writing `checks.xml` and `problems.xml`. Nothing else in the pipeline reads either of them, so nothing downstream changes.

This is a pause, not a retraction. The design of the checking loop stands, the packs that describe it are in the history under #6614, and #6643 and #6647 are the two improvements it was owed. It comes back when a locator can be given a type without a call site having to reach it, which is what #6653 and the structural typing after it are for.

maxonfjvipon:
Six tables came out of this module: three that record what a program says about
its own objects, and `checks` plus the `problems` they drain into, which decide
whether it is wrong. The deciding half comes out until the recording half can
describe a program.

It reports nothing, and not by accident. A verdict needs the object missing an
attribute to have been seen whole, and on a clean `eo-runtime` build 48 of 2,505
types have been; the other 2,432 delegate, and 2,410 of those chains end at a
locator no table describes. Walking behind `φ` (#6643) moves 48 to 49. Recording
what a computed object turns out to be (#6647) adds 126 rows, 1,389 with both,
and moves nothing further. The biggest document the module wrote was
`checks.xml`, 4.5 MB, and its entire yield was an empty file.

The reason is structural: the loop learns along the path a check walks, and a
check starts at an application, so an object nothing applies in the same world
is never reached. `Φ.bool.and` delegates to the dispatch `.if` on `ξ.ρ`, nothing
in `eo-runtime` copies `and`, and so the body of `and` has no type. A table
built by walking call sites cannot cover a library. Giving every object a type
first — which is what #6653 and the structural typing after it are for — is what
turns checking back into a thin pass at the end, which is where it belongs.

So `Checks`, `Concluded`, `Worklist`, `Provided`, `Same`, `Ungrouped` and `Row`
go, with the fourteen packs that describe verdicts and the `applications()` of
`Xmirs` that had one caller:

```java
public Clues() {
    this(Arrays.asList(new Provides(), new Needs(), new Links()));
}
```

`note-example.yaml` stays, trimmed to the three tables: every fact needed to
know that `x.next.foo` is wrong is still asserted there, and no row says it,
which is exactly the state of things. 20 packs, 12 unit tests and 578 plugin
tests are green, and nothing downstream read either of the two documents.

This is a pause. The design stands, the code is one revert away, and the two
tickets it was owed stay written down.

Closes: #6661


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1341 >= 100); -8 for way too many hits-of-code (1341 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+658](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6663, opened 2026-08-12
fix(#6649): reject lossy format recovery

### LINKED ISSUE #6649: `MjFormat.parsed` allows problematic recovery that can omit a binding
The guard in `eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java:232` rejects a source only when severe `<errors>` coexist with `truncated()` (the tree stops short of the last non-blank line) or `placeholder()` (a recovered empty `<o/>`). There is a third kind of lossy recovery it lets through: the parser can recover from a mid-file syntax error by dropping just the offending binding and carrying on, so line coverage stays intact and no placeholder appears. The guard does not fire, `parsed()` returns the mutilated tree, and `eo:format -Deo.autoFix` writes it back over the source.

For example, this source:

```
[x] > foo
  seq * > @
    last
    last.plus 1
  x! > last

[] > bar
  42 > @
```

fails to parse at `[5:5]` with `trailing garbage after name suffix` (the const form has to be `x > last!`, not `x! > last`). In the recovered tree the maximum `@line` is 8, which equals the source's last non-blank line, so `truncated()` is FALSE; there is no empty `<o/>`, so `placeholder()` is FALSE. Yet the binding `name="last"` is gone entirely and its two references have escaped to the global `base="Φ.last"`. With `-Deo.autoFix` the line is erased from the file on disk, the two references are left dangling, and the build log never mentions the parse error at all — it only complains that the source is not formatted canonically. This is the next gap in the same guard after #5851 and #6071, and is neither of them: nothing is truncated and no placeholder is produced.

The shape heuristics cannot keep up, because `severity="error"` conflates cosmetic layout recoveries with genuinely lossy ones. Either tag the two kinds apart in the parser so this guard can test for lossiness directly, or drop the heuristics and have `parsed()` re-parse its own printed output and refuse to write when the two trees differ.

ColumbusLabs:
Closes #6649.

This change:
- tags parser errors when recovery discards source;
- refuses to format or auto-fix XMIR carrying that loss marker;
- covers both check and auto-fix modes, including byte-for-byte preservation of the original source.

Verification:
- `mvn -pl eo-maven-plugin clean test -Dtest=MjFormatTest -DskipITs -ntp` (9 passed)
- parser suite (1,728 passed)
- plugin suite (593 passed, 5 skipped)
- `mvn clean -q -pl eo-parser,eo-maven-plugin -am verify -Pqulice -DskipTests`

The patch is 100 hits, within the project contribution limit.

@yegor256 please take a look.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
@@ -195,9 +195,10 @@ public final class MjFormat extends MjSafe {
      * {@code <errors>} (see {@link Parsing}), but the parser also recovers
      * from many purely cosmetic layout violations (e.g. extra blank lines,
      * exactly what this goal exists to fix) the same way, tagged with the
-     * same {@code severity="error"} and no other distinguishing marker.
-     * Rejecting every {@code <errors>} entry, as suggested by the issue this
-     * fixes, would therefore also reject this goal's own core use case.
+     * same {@code severity="error"}; only lossy recovery also carries
+     * {@code lost="true"}. Rejecting every {@code <errors>} entry, as
+     * suggested by the issue this fixes, would therefore also reject this
+     * goal's own core use case.
      *
      * <p>What actually matters is whether the recovered tree still covers
      * the whole source: a genuine syntax error (e.g. an unterminated paren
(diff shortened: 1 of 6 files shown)


ColumbusLabs:
Closing as superseded by #6852 (88ed09efc), which has already landed the same #6649 lost-recovery guard on `master`. Rebasing this PR would duplicate that implementation.

PR #6668, opened 2026-08-12
fix(#6640): take heap blocks only through the releasing scope

### LINKED ISSUE #6640: Negative-argument, size and related parameters not correctly handled in Heaps.java
The puzzle `6507-93ee2a46` from #6507 has to be resolved: 

https://github.com/objectionary/eo/blob/95ab3a4e23d51468ebee91cb99be7710965be378/eo-runtime/src/main/java/org/eolang/Heaps.java#L17-L20

yangjj-iso:
@yegor256 — this resolves the `6507-93ee2a46` puzzle in `Heaps`.

Closes #6640

## What the puzzle asked for

Four steps, and they only work together: `free` cannot become private until nothing outside the class calls it, so the test moves and the visibility change have to land in one commit.

- the negative-argument, size and resize tests in `HeapsTest` moved onto the scoped `malloc`;
- both free probes in `EOmallocEOofTest` moved off `free`;
- two-argument `malloc` and `free` are private;
- `failsOnClearingEmptyBlock` is gone.

## What changed

**`Heaps`.** The two-argument `malloc` handed out a block with nobody responsible for releasing it, and `free` let any caller release a block it did not own — so `malloc(phi, size, scope)` was a convention rather than a rule. Both are private now, which makes the scope the only way in and out. Nothing in `src/main` was affected: `EOmalloc$EOof` already goes through the scoped form, and it was the only production caller.

**`HeapsTest`.** Seven cases took a block raw and freed it by hand; they now run inside the scope. Two of them got shorter — `returnsValidSize` is a one-liner against `Heaps.INSTANCE::size` — and the ones that assert on an exception mid-scope keep the assertion inside the lambda so the block is still allocated when it runs. `failsOnClearingEmptyBlock` is deleted rather than adapted, because "freeing an unallocated block fails" is no longer a reachable state.

**`EOmallocEOofTest`.** Both probes asked "is it freed?" by calling `free` a second time and expecting a throw. They now ask `size` instead, which throws the same `ExFailure` for an unallocated block, so the assertion type is unchanged and the probe no longer needs a private method.

## On the size of this one

185 hits, over the 40–100 the guidelines ask for, and I could not find an honest way to split it.

Roughly 56 of those are mechanical: qulice orders methods by visibility, so leaving `malloc` where it was reported `MethodsOrderCheck` against every package-private method below it. Moving the private pair to the end of the class fixes that but counts as a delete plus an add of the same body. The remaining change is the seven test conversions, which cannot be separated from the visibility change without leaving the tree uncompilable in between. Happy to reshape it if you would rather have it another way.

## Verification

```
$ mvn -pl eo-runtime test -Dtest=HeapsTest,EOmallocEOofTest
Tests run: 25, Failures: 0, Errors: 0, Skipped: 0 -- in org.eolang.HeapsTest
Tests run:  4, Failures: 0, Errors: 0, Skipped: 0 -- in org.eolang.EOmallocEOofTest
BUILD SUCCESS
```

That run compiles the whole module, main and test, which is also what rules out a missed caller of the now-private methods.

`mvn -Pqulice` caught the ordering problem above; after the move it reports no violations on the changed file.

## What I could not run here, and why

I am on Windows, and `mvn clean install -Pqulice` does not complete on this machine for two reasons that predate this branch. I checked both by stashing the change and re-running on unmodified `master`, which fails identically:

- `eo-parser` fetches `blns.txt` from `raw.githubusercontent.com` through `maven-antrun-plugin`, which times out here (`github.com` and `api.github.com` are fine, that host is not). Worked around with `-Dmaven.antrun.skip=true`.
- `eo-runtime`'s `project-validate` step then fails with `Missing: EOwin32$EOφ.class` / `Missing: EOposix$EOφ.class` — "Not all .java files were compiled to .class files". Same failure with an empty working tree, so it is not from this change, but it does mean I have not run the full `-Pqulice` lifecycle end to end.

Separately, `format` reports `2 of 170 EO source(s) are not formatted canonically` on a clean checkout. I left those alone — no `.eo` file is touched here — and used `-Deo.autoFix` locally to get past it.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Heaps.java
+++ b/eo-runtime/src/main/java/org/eolang/Heaps.java
@@ -14,10 +14,6 @@ import java.util.function.IntFunction;
 /**
  * Dynamic memory.
  * @since 0.19
- * @todo #6507:30min Move the negative-argument, size and resize tests in HeapsTest and
- *  both free probes in EOmallocEOofTest onto the scoped malloc, then make malloc with two
- *  arguments and free private, so that a block can only be taken through a scope that
- *  releases it, and drop failsOnClearingEmptyBlock, which nobody can reach any more.
  */
 final class Heaps {
 
@@ -43,31 +39,6 @@ final class Heaps {
     private Heaps() {
         this.blocks = new ConcurrentHashMap<>(0);
         this.lock = new ReentrantLock();
-    }
-
-    /**
-     * Allocate a block in memory.
-     * @param phi Object
-     * @param size How many bytes
-     * @return The identifier of pointer to the block in memory
-     */
-    int malloc(final Phi phi, final int size) {
-        final int identifier = phi.hashCode();
-        this.lock.lock();
-        try {
-            if (this.blocks.containsKey(identifier)) {
-                throw new ExFailure(
-                    String.format(
-                        "Can't allocate block in memory with identifier '%d' because it's already allocated",
-                        identifier
-                    )
-                );
-            }
-            this.blocks.put(identifier, new byte[size]);
-        } finally {
-            this.lock.unlock();
-        }
-        return identifier;
     }
 
     /**
@@ -251,10 +222,45 @@ final class Heaps {
     }
 
     /**
+     * Allocate a block in memory.
+     *
+     * <p>Private on purpose: a block handed out here has no owner responsible
+     * for releasing it. Take one through {@link #malloc(Phi, int, IntFunction)}
+     * instead, whose scope frees the block on every exit path.</p>
+     *
+     * @param phi Object
+     * @param size How many bytes
+     * @return The identifier of pointer to the block in memory
+     */
+    private int malloc(final Phi phi, final int size) {
+        final int identifier = phi.hashCode();
+        this.lock.lock();
+        try {
+            if (this.blocks.containsKey(identifier)) {
+                throw new ExFailure(
+                    String.format(
+                        "Can't allocate block in memory with identifier '%d' because it's already allocated",
+                        identifier
+                    )
+                );
+            }
+            this.blocks.put(identifier, new byte[size]);
+        } finally {
+            this.lock.unlock();
+        }
+        return identifier;
+    }
+
+    /**
      * Free it.
+     *
+     * <p>Private on purpose: the scope opened by
+     * {@link #malloc(Phi, int, IntFunction)} is the only thing that releases a
+     * block, so no caller can free one it does not own or free one twice.</p>
+     *
      * @param identifier Identifier of pointer
      */
-    void free(final int identifier) {
+    private void free(final int identifier) {
         this.lock.lock();
         try {
             if (!this.blocks.containsKey(identifier)) {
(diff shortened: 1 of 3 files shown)


yangjj-iso:
`runtime (23)` is red here, but it is red on `master` too — including on `e59be523`, the commit this branch is based on:

| run | head | `runtime (23)` |
|---|---|---|
| 2026-08-12 13:16 | `38c14519` (master) | failure |
| 2026-08-12 12:15 | `e59be523` (master, this branch's base) | failure |
| 2026-08-12 09:31 | `20e15091` (master) | failure |

The failure is a Saxon error inside `MjFormat.canonical` → `MjFormat.reformat`, i.e. the EO source formatter, which also reports `2 of 170 EO source(s) are not formatted canonically` on a clean checkout. This branch touches three Java files in `eo-runtime` and no `.eo` source, so it cannot reach that step.

Everything else is green, including the two checks that matter most for this change and that I could not complete locally on Windows: **qulice** and **mvn (macos-15, 26)**.

yegor256:
@yangjj-iso see: https://www.yegor256.com/2025/04/19/dont-merge-into-broken-master.html

yangjj-iso:
Fair — I was explaining the red, not asking you to look past it. A broken master is the reason not to merge, not a reason to merge anyway.

So I went after the break instead: #6672 fixes #6669. The template at `inline-cactoos.xsl:314` hands `@name` — an attribute node — to nine functions that each declare `$name` as `xs:string`, leaving the atomization to the function conversion rules. Saxon may bind the parameter to a closure over the argument expression rather than over its converted value, so the node reaches a comparison that was already compiled to a `ValueComparison` on the strength of the declared types, and the cast to `AtomicValue` fails. Atomizing once in the template covers all nine call sites.

I will rebase this branch onto a green master once that lands, rather than ask for anything here in the meantime.

yangjj-iso:
Moving this to a draft while master is red. The post is explicit that you wait, and leaving it open asks for review time the build has not earned yet.

Sweeping every job rather than just the one I had looked at turned up two more reds on master, both now reported:

- #6675 — `mvn (windows-2022, 26)` wrote nothing for thirty-three minutes after `SavedTest.exposesNoPartiallyWrittenFileToConcurrentReader` errored, and the runner killed it. Intermittent: red on one of the last fourteen master commits, green on the rest.
- #6677 — the SonarCloud gate is red on master's head, on `new_reliability_rating` alone (C, where A is required). Red on the six master analyses before it too.

So #6672 on its own will not turn master green. I said otherwise earlier in this thread and that was too quick — I had checked the `runtime` workflow and not the others.

I will mark this ready again once master is green, rebased onto it.

yegor256:
@yangjj-iso try to sync with master, it should work now

yegor256:
The `runtime (23)` CI check is failing on this PR. Could you please investigate and fix the failure?

---

yegor256:

---

yegor256:
runtime (23) check is red on this PR. Not merging until CI is green.

---

yegor256:
Thank you for this pull request. We are closing it because it currently has merge conflicts with the master branch and has been open for over a month. Please forgive us for closing it — this is just routine housekeeping, not a judgment on the contribution.

If you are still interested, we would welcome a new pull request that resolves the conflicts against the latest master, ideally reflecting any changes that have landed there since this PR was opened.

PR #6671, opened 2026-08-12
Stop dataization on an interrupted thread

### LINKED ISSUE #6670: `eo.deadline` skips a test but never stops it, leaking one runaway thread per skip
In `eo-runtime` the deadline is JUnit's declarative timeout, configured at `eo-runtime/pom.xml:209` and run in `SEPARATE_THREAD` mode by `eo-runtime/src/test/resources/junit-platform.properties:4`. That mode interrupts the test thread when the timeout expires and never joins it. Dataization, however, never polls the interrupt flag: `PhDefault.take()` at `eo-runtime/src/main/java/org/eolang/PhDefault.java:219` is the universal dispatch point every EO computation passes through, and it ignores interruption, while the only mention of it in the module, `AtomSafe.java:41`, just re-raises the flag before wrapping the error. A CPU-bound EO computation has no blocking point either, so the interrupt is a no-op and the work continues after the test is already reported as skipped.

Each skipped test therefore leaves a non-daemon `junit-timeout-thread-N` spinning for the rest of the fork. Because classes and methods run concurrently (`eo-runtime/pom.xml:231-235`), those threads accumulate and starve the runner, so the next test overruns its wall-clock deadline even when its real cost is a few hundred milliseconds, and leaks another thread in turn. The run collapses into wholesale skipping: in [run 31598572547](https://github.com/objectionary/eo/actions/runs/31598572547) the `runtime (23)` job reported `32 of 32`, `25 of 25`, and `7 of 7` tests skipped with per-class times of 300-700 seconds, and was cancelled at the 30-minute limit. A standalone JUnit probe with the same configuration confirms the leak: 32 timed-out tests leave exactly 32 live timeout threads, and a test that passes in 447ms alone is killed by the 1s deadline once those threads own the CPU.

The fix is a guard at the top of `PhDefault.take()` that throws when the current thread carries the interrupt flag, so the interrupt sent by JUnit actually terminates the runaway dataization. It must read the flag with `Thread.currentThread().isInterrupted()` rather than clear it with `Thread.interrupted()`, so that every later dispatch keeps failing and an EO-level `try` cannot resurrect the loop.

yegor256:
A test that outlives `eo.deadline` gets skipped, but its work does not stop: JUnit interrupts the thread and never joins it, and nothing in dataization ever looked at the interrupt flag. `PhDefault.take()` now checks it and throws the new `ExInterrupted` on the very next attribute lookup. The flag is read, not cleared, so every later lookup fails the same way and an EO-level `try` cannot bring the computation back to life.

Without this, every skipped test left a non-daemon `junit-timeout-thread` spinning for the rest of the fork. They pile up, starve the runner, and then tests that need a few hundred milliseconds start overrunning their wall-clock deadline too and leak threads of their own — which is how the `runtime` job ends up skipping whole classes at once and getting cancelled at the 30-minute limit. With a 1s deadline, eight endless dataizations used to leave eight live threads behind; now they leave none. `ExInterruptedTest` reproduces it: without the guard the interrupted thread is still taking attributes when the test gives up on it.

One thing to keep in mind: the deadline is still measured in wall clock under parallel execution, so `eo.deadline=1` will keep skipping tests that are merely slow rather than stuck. That value probably wants raising, but it is a separate decision.

Closes #6670

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/ExInterrupted.java
+++ b/eo-runtime/src/main/java/org/eolang/ExInterrupted.java
@@ -1,0 +1,37 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang;
+
+/**
+ * Error that stops a computation running on an interrupted thread.
+ *
+ * <p>Dataization is a long chain of attribute lookups, where nothing ever
+ * blocks. A thread that has to be stopped from the outside — a test that
+ * outlived its deadline, a program that must shut down — therefore never
+ * notices the interrupt and keeps computing forever, holding a CPU core
+ * until the JVM dies. {@link PhDefault#take(String)} throws this error on
+ * the very next lookup instead.</p>
+ *
+ * <p>The interrupt flag stays up when this error is thrown, so every next
+ * lookup fails the same way and no {@code try} can resurrect the
+ * computation.</p>
+ *
+ * @since 0.74.0
+ */
+public final class ExInterrupted extends ExAbstract {
+
+    /**
+     * Serialization identifier.
+     */
+    private static final long serialVersionUID = 597749420437007616L;
+
+    /**
+     * Ctor.
+     * @param cause Exception cause
+     */
+    public ExInterrupted(final String cause) {
+        super(cause);
+    }
+}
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2882](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6650, opened 2026-08-12
Different types in `if` branches in `restore-local-names.xsl` crash Saxon and halt `format` functionality


maxonfjvipon:
The `@base` template at `eo-printer/src/main/resources/org/eolang/printer/print/restore-local-names.xsl:219` rewrites every dotted segment of a base, taking each segment either from a handle's `@local` attribute or from the token it already is:

```xslt
<xsl:template match="@base">
  <xsl:attribute name="base" select="string-join(for $seg in tokenize(., '\.') return (if (key('void-handle', $seg)) then key('void-handle', $seg)[1]/@local else $seg), '.')"/>
</xsl:template>
```

The two branches of that `if` hand back different kinds. `@local` is an attribute node; `$seg` is an atomic string. Saxon is free to type the body of the `for` from the atomic branch, and when it does, the node coming out of the other branch reaches `string-join` unatomised: `class net.sf.saxon.dom.DOMNodeWrapper cannot be cast to class net.sf.saxon.value.AtomicValue`. It is reported as `Internal error evaluating template rule at line 219`, with the identity rule at line 326 failing behind it as the error travels back up, and it takes the build down — `format` on `eo-runtime` stops with `Shift 'com.yegor256.xsline.StClasspath' failed`, see https://github.com/objectionary/eo/actions/runs/31573190318/job/94039395158.

Like #6638, this is a build that works until it does not: nothing in the source decides whether the optimiser picks the atomic type, so the same commit prints its 170 sources on one runner and crashes on another. The `runtime` workflow is where the cost shows. Its last hundred runs on master hold ten failures and no successes, the rest cancelled by newer pushes, and the run on the current tip spent its whole 25-minute budget without finishing.

Atomise the node branch so both sides are strings — `string(key('void-handle', $seg)[1]/@local)` — and the mixed sequence disappears. The same shape in `to-eo-tree.xsl:25` returns a string from every branch and never trips over this.

maxonfjvipon:
Fixed on master in 28e7534480 — `string(key(...)[1]/@local)` leaves nothing mixed for the optimiser to type.

0crat:
@maxonfjvipon Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+652](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6672, opened 2026-08-12
fix(#6669): atomize the auto-name before the fold checks

### LINKED ISSUE #6669: `inline-cactoos.xsl` crashes with a ClassCastException in `eo:multi-referenced` — the `[2]` fix for #6638 moved the crash instead of removing it
Seen on `mvn (macos-15, 26)` in https://github.com/objectionary/eo/actions/runs/31602103102/job/94131828678, during `eo:format` on `eo-runtime`. The head commit's only diff was to `wrap-applications.xsl` (parse train) plus a test pack, so this is not caused by that diff — see the analysis at the bottom.

## What happens

`MjFormat` → `Xmir.toEO` → `inline-cactoos.xsl` dies with:

```
Internal error evaluating template rule at line 314 in module inline-cactoos.xsl
Caused by: Internal error evaluating function eo:multi-referenced at line 486 in module inline-cactoos.xsl
Caused by: java.lang.ClassCastException: class net.sf.saxon.dom.DOMNodeWrapper cannot be cast to
                                        class net.sf.saxon.value.AtomicValue
    at net.sf.saxon.expr.ValueComparison$ValueComparisonElaborator.lambda$elaborateForBoolean$4
    at net.sf.saxon.expr.OrExpression$OrElaborator.lambda$elaborateForBoolean$0
    at net.sf.saxon.expr.AndExpression$AndElaborator.lambda$elaborateForBoolean$0   (x2)
    at net.sf.saxon.expr.FilterExpression$FilterExprElaborator$SimpleFilteredIterator.next
    at net.sf.saxon.om.MemoSequence.itemAt
    at net.sf.saxon.expr.SubscriptExpression.getItemAt
    at net.sf.saxon.functions.Exists$ExistsFnElaborator.lambda$elaborateForBoolean$1
    at net.sf.saxon.expr.elab.LearningEvaluator.evaluate
    at net.sf.saxon.expr.instruct.UserFunction.call
```

## Where

The frames below `Exists`/`SubscriptExpression` are the `[2]` introduced by #6638 (`eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl:489`):

```xml
  <xsl:param name="name" as="xs:string"/>
  <xsl:sequence select="exists(eo:references($target, $name)[2])"/>
</xsl:function>
```

The `ValueComparison` that blows up is the `=` in `eo:references`' own predicate (`inline-cactoos.xsl:470`):

```xml
<xsl:sequence select="$target/..//o[contains(@base, concat('.', $auto))
  and (eo:resolved-name(@base) = $name or starts-with(eo:resolved-name(@base), concat($name, '.')))
  and not(ancestor-or-self::o[. is $target])]"/>
```

Statically both operands are `xs:string` (`eo:resolved-name` is declared `as="xs:string"`, `$name` is `as="xs:string"`), which is why Saxon compiles the `=` down to a `ValueComparison` that casts to `AtomicValue` without atomizing. At runtime one operand arrives as a `DOMNodeWrapper` — a node — and the cast fails.

## Why this looks like the #6638 fix moving the crash

`MemoSequence.itemAt` sitting under `SubscriptExpression.getItemAt` means the `[2]` makes Saxon re-enter `SimpleFilteredIterator.next` **lazily**, re-evaluating the predicate later through a memoised closure rather than inside the function's own evaluation. The `SingletonClosure.asItem`/`head` frames higher up confirm the whole thing is running under lazy evaluation, and `LearningEvaluator` means the strategy is chosen at runtime from call counts.

So #6638 traded one optimiser-dependent crash (`MemoSequence` refusing `getLength()` for `count()`) for another on the same expression: with `[2]` the predicate is evaluated in a lazy context where the comparison's operand is no longer atomized. Both are nondeterministic, which is why only one matrix cell failed and why the same commit's own master build was green.

Worth checking whether forcing eager materialisation of the sequence inside `eo:references` (sequence-constructor form of `xsl:variable`, which Saxon materialises rather than closing over) removes the class of bug, instead of another lazy-friendly rewrite of the same shape. Explicitly atomizing the comparison operands (`string(...)` on both sides) would also stop Saxon from choosing an unchecked `ValueComparison` here.

## Evidence this is not the parse-train diff on that PR

- That diff only changes `wrap-applications.xsl` in the parse train; the crash is in the print train.
- Its output is byte-identical AND DOM-identical (node types, names, attributes, text/comment values, text-node splitting) to the previous sheet on all 170 `eo-runtime` sources and on 533 programs extracted from the parser and printer test packs — so the XMIR the printer receives is unchanged.
- `mvn -pl eo-runtime process-sources` (the very goal that failed) passes locally with that diff on Linux/JDK 21: exit 0, all 170 sources formatted canonically, no exception.

yangjj-iso:
Closes #6669

`runtime` is red on master because of this, so nothing merges until it is green — including my own #6668, which is how I came to it.

## Root cause

The stack trace on the issue names the caller: the template rule at `inline-cactoos.xsl:314`. That template hands `@name` — an attribute node — to nine functions in a row:

```xml
<xsl:if test="eo:recursive(., @name) or eo:dispatched(., @name) or ... or eo:reapplied(., @name)">
```

Every one of them declares its `$name` as `xs:string`, so the atomization was left to the function conversion rules. Saxon is free to bind a parameter to a closure over the argument expression rather than over its converted value, and then the node itself arrives inside the function. There the comparison

```xml
eo:resolved-name(@base) = $name
```

has already been compiled to a `ValueComparison`, on the strength of both operands being statically `xs:string` — so it casts straight to `AtomicValue`, and the sheet dies with `DOMNodeWrapper cannot be cast to AtomicValue`.

The other operand cannot be the node: `eo:resolved-name` builds its result with `substring-after`, so it is a string by construction. `$name` is the only candidate, and that template is the only place a node can reach it — every other call site passes the `$name` of the inlining template at line 100, which is itself `eo:resolved-name(@base)`.

This also accounts for the "moved the crash" reading in the issue. `count()` materialised the sequence inside the function's own evaluation; `[2]` reaches it through a lazy subscript, which is exactly the context where the deferred conversion is lost. #6638 did not introduce the type mismatch — it changed the evaluation strategy enough to expose it.

## The change

Atomize once, in the template, before any of the nine is asked:

```xml
<xsl:variable name="name" as="xs:string" select="string(@name)"/>
```

Two lines. It leaves the conversion nothing to lose and covers all nine functions at once, rather than rewriting the comparison inside each of them — the same move as the `eo:signature` call in `to-eo-tree`, and as the `@local` wrap in `restore-local-names` for #6650.

`string()` is exact here rather than merely close: the match pattern is `o[starts-with(@name, $auto) and not(eo:void(.))]`, so `@name` is always present, and the parameters are required `xs:string`. Nothing about what the nine functions are asked changes.

## Verification

The failure is optimiser-dependent and does not reproduce on demand — that is the shape of the bug, and it is why I have no regression test to offer. Master is red on `runtime (23)` while the same sources print canonically on the other runners.

So what I can show is that the change is inert on everything that does run:

| | master | with this change |
|---|---|---|
| `mvn -pl eo-printer test` | 346 passed | 346 passed |
| `mvn -pl eo-runtime process-sources` | all 170 sources canonical | all 170 sources canonical |
| transpiled | 170 XMIRs, 179 Java files | 170 XMIRs, 179 Java files |

Both columns are this branch's base, `d421d8d`; the left one was produced by stashing the change and reinstalling `eo-printer`, so the comparison is like for like.

## What I deliberately left alone

`restore-local-names.xsl` has the same shape — `eo:recursive(., @name)` and five others, called from match patterns — and it took the same `[2]` change in 28e7534. I did not touch it here: its parameters are `xs:string?` rather than `xs:string`, so `string(@name)` would turn an absent `@name` from an empty sequence into `''` and change what those functions answer. The faithful form there is `@name/string()`, which deserves its own change and its own reasoning rather than riding along with this one. Glad to open it if you want it.

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -310,9 +310,21 @@
   never moved its value anywhere: dropping it deletes the declaration from the
   printed source, so a private helper formation or const cache that nothing
   reads yet would silently vanish (#5914).
+
+  The name is atomized once here, before any of those functions is asked
+  (#6669). Each declares its `$name` as `xs:string`, so handing over the
+  `@name` node left the conversion to the function conversion rules, and Saxon
+  is free to bind the parameter to a closure over the argument expression
+  rather than over its converted value. The node then arrives inside the
+  function, where `eo:resolved-name(@base) = $name` has already been compiled
+  to a `ValueComparison` on the strength of the declared types — it casts
+  straight to `AtomicValue` and the sheet dies with "DOMNodeWrapper cannot be
+  cast to AtomicValue". A string leaves the conversion with nothing to lose,
+  as at the `eo:signature` call in "to-eo-tree".
   -->
   <xsl:template match="o[starts-with(@name, $auto) and not(eo:void(.))]" priority="1">
-    <xsl:if test="eo:recursive(., @name) or eo:dispatched(., @name) or eo:vertical-const(.) or eo:unreferenced(., @name) or (eo:multi-referenced(., @name) and eo:rebuilt(.)) or eo:piped(., @name) or eo:arg-applied(., @name) or eo:nested-applied(., @name) or eo:reapplied(., @name)">
+    <xsl:variable name="name" as="xs:string" select="string(@name)"/>
+    <xsl:if test="eo:recursive(., $name) or eo:dispatched(., $name) or eo:vertical-const(.) or eo:unreferenced(., $name) or (eo:multi-referenced(., $name) and eo:rebuilt(.)) or eo:piped(., $name) or eo:arg-applied(., $name) or eo:nested-applied(., $name) or eo:reapplied(., $name)">
       <xsl:copy>
         <xsl:apply-templates select="node()|@*"/>
       </xsl:copy>


yangjj-iso:
Reopened to re-run CI against current master. Nothing about the branch changed.

The first `runtime (23)` here was cancelled at the timeout, not failed. The two are worth separating, because the checks list shows them the same way and they mean opposite things:

- the crash this PR is about **fails fast** — between 3.5 and 9 minutes on every run that hit it today;
- the hang **fails slow** — cancelled around the 25 minute mark, which is how most of today's `runtime` runs ended, on master and on every branch.

This branch is based on `d421d8d`, before #6671 landed, so it drew the second one and ran 32 minutes before the runner collected it. Its first Maven step — which also builds `eo-runtime` and runs `format` — passed, so the crash did not recur there. Master's own `runtime` then went green in 13.5 minutes on `2e9427b`, immediately after #6671 merged, and that is the first clean run of that job today.

A `pull_request` run uses the merge of this branch with the base, so reopening picks up #6671 without touching the branch. This time the job should get far enough to say something about the change itself.

yegor256:
@yangjj-iso conflicts here

yegor256:
The CI checks look mostly good but this PR has a merge conflict with the base branch. Could you please rebase onto the latest `master` and resolve the conflicts?

---

yegor256:
This PR has a merge conflict (`mergeable_state: dirty`). Please rebase onto the latest `master` and resolve the conflicts before it can be merged.

---

yegor256:

---

yegor256:
runtime (23) is cancelled, not green, on this PR. Not merging until CI is fully green.

---

yegor256:
Thank you for this pull request. We are closing it because it currently has merge conflicts with the master branch and has been open for over a month. Please forgive us for closing it — this is just routine housekeeping, not a judgment on the contribution.

If you are still interested, we would welcome a new pull request that resolves the conflicts against the latest master, ideally reflecting any changes that have landed there since this PR was opened.

PR #6673, opened 2026-08-12
fix(#6628): enforce atom body rules in transitions

### LINKED ISSUE #6628: Atom-body check in LnFormation is not enforced for application/reversed/compact-tuple children
The atom-body rule — an atom's body may contain nothing but `+>` test attributes — is only enforced by one of the several line shapes that can sit under it. `LnFormation.checkChildAllowed` in [`eo-parser/src/main/java/org/eolang/parser/LnFormation.java:193-198`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/LnFormation.java#L193-L198) throws when a deeper-indent `[...]` formation is pushed under an atom whose suffix isn't a test, but that guard is reachable only from `LnFormation.transition()` (line 152). The shared prologue every other line shape uses instead, `Transition.apply()` in [`eo-parser/src/main/java/org/eolang/parser/Transition.java:54-77`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/Transition.java#L54-L77), duplicates the indent-step and openness checks from `checkChildAllowed` but has no matching atom-body check, and it's what `LnApplication`, `LnReversed`, `LnCompactTuple`, `LnVoid`, `LnPipe`, `LnTextBlock` and `LnOnlyPhi` all call.

So `[] > foo /number` followed by an indented `42 > bar`, or by `if. > bar ...`, parses with zero errors and silently attaches `bar` as a sibling of `foo`'s `λ` atom marker in the XMIR, while the equivalent nested-formation child (`  [] > bar`) is correctly rejected with "Atom cannot contain inner objects; only `+>` test attributes are allowed in an atom body". `Eo.java:891`'s own comment — "Atom-body and test-depth checks attach as their owning line shapes are implemented" — shows the check was meant to be wired into each shape but only ever landed in `LnFormation`. No downstream XSL sheet or lint catches it either, so a malformed atom body written as an application or reversed-dispatch child passes through the whole pipeline unflagged.

Move the `stack.top().atom() && !suffix.test()` check out of `LnFormation.checkChildAllowed` and into the shared `Transition.apply()`, so the rule applies the same way regardless of which line shape the illegal child is written as.

fzlzjerry:
## Summary

- move atom-child validation into the shared `Transition.apply` path
- reject non-test application, reversed, compact-tuple, pipe, text-block, and only-phi children while preserving atom void parameters
- keep `+>` and `->` test attributes valid and cover the application-child regression

## Validation

- `mvn -pl eo-parser -Pqulice verify` (1,729 tests, 0 failures, 6 skipped; Qulice passed)
- `mvn -pl eo-parser -DskipITs '-Dtest=TransitionTest,EoSyntaxTest#checksTypoPacks' test` (104 tests, 0 failures)

Closes #6628

@yegor256

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnApplication.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnApplication.java
@@ -147,7 +147,7 @@ final class LnApplication implements Line {
     private void transition(
         final Stack stack, final Suffix suffix, final Kind kind, final Openness openness
     ) {
-        new Transition(stack, this.span).apply(kind, openness, suffix.named());
+        new Transition(stack, this.span).apply(kind, openness, suffix);
     }
 
     /**
--- a/eo-parser/src/main/java/org/eolang/parser/LnCompactTuple.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnCompactTuple.java
@@ -105,7 +105,7 @@ final class LnCompactTuple implements Line {
      */
     private Level transition(final Stack stack, final Suffix suffix) {
         return new Transition(stack, this.span)
-            .apply(Kind.COMPACT_TUPLE, Openness.OPEN, suffix.named());
+            .apply(Kind.COMPACT_TUPLE, Openness.OPEN, suffix);
     }
 
     /**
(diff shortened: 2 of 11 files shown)


yegor256:
@fzlzjerry conflicts here

fzlzjerry:
Thanks for the heads-up. The conflict comes from #6678, which has now merged the same Transition-level validation with broader line-shape coverage. I checked the current master with TransitionTest and EoSyntaxTest (1,101 tests; 0 failures, 6 skipped), so this PR is superseded. I am closing it instead of retaining a duplicate implementation.

PR #6676, opened 2026-08-12
fix(transpile): #6674 buffer the generated Java before handing it to the DOM

### LINKED ISSUE #6674: `to-java.xsl` performance issue: time complexity is quadratic for generating Java code sizes
Follow-up to #6664, same mechanism, bigger bill. `to-java.xsl` is now four fifths of the transpile train.

Per-stylesheet, best of two warm runs, all 170 sources of `eo-runtime/src/main/eo` parsed with `EoSyntax` and pushed through the train one step at a time (Saxon-HE 13.0, single-threaded, `master` at 9760822):

```
    725 ms  parse/set-locators.xsl
    508 ms  transpile/set-original-names.xsl
   1205 ms  transpile/classes.xsl
    649 ms  transpile/tests.xsl
    728 ms  transpile/anonymous-to-nested.xsl
    675 ms  transpile/package.xsl
    683 ms  transpile/attrs.xsl
    585 ms  transpile/data.xsl
  25408 ms  transpile/to-java.xsl      <-- 82% of the train
  31168 ms  TRAIN TOTAL over 170 files
```

## Where it goes

Saxon's `TimingTraceListener` on `string/printf.eo` puts 8,328ms of the 8,881ms total in the **net** time of one construct — `xsl:template match="class"` (line 223), the template that does nothing but open `<tests>` and `<java>` and delegate. Net time, so not the 26 templates it calls. That is the cost of handing the text to the result tree.

## Why

The result tree is a DOM — `XSLDocument.transform` gives Saxon a `DOMSource` and a `DOMResult`. The templates assemble the Java from thousands of small `xsl:text` and `xsl:value-of` pieces, each handed to the outputter as it is produced, and Saxon merges adjacent text into a single DOM text node. Merging one piece into a Xerces `Text` node means `setNodeValue(this.data + data)` — the whole accumulated string is copied, every time.

So the pass is **quadratic in the size of the generated Java**. `<java>` for `string/printf.eo` ends up as exactly **one** text node of **484,853 characters**, and every piece along the way paid for a copy of everything written before it.

Sending the identical transform to a `StreamResult` instead of a `DOMResult` isolates it — same stylesheet, same input, warm:

| result | time |
|---|---|
| `DOMResult` (what the build uses) | 3460 ms |
| `StreamResult` | 91 ms |

Worth noting what is *not* the cause, since it looks like the obvious suspect: `eo:eol()` and `eo:tabs()` are called 4,257 times each and build a temporary tree per call, because neither declares `as="xs:string"`. Rewriting all of `$TAB`, `eo:eol` and `eo:tabs` as pure string functions changed the total by nothing measurable (24.5s → 26.1s, i.e. noise). The trace agrees — 61ms and 27ms gross respectively.

## Suggested fix

Collect each body in a variable and write it with one `xsl:value-of`:

```xml
<java>
  <xsl:variable name="code">
    <xsl:call-template name="commonclass"/>
    <xsl:apply-templates select="." mode="body"/>
  </xsl:variable>
  <xsl:value-of select="$code"/>
</java>
```

A temporary tree is Saxon's own, where appending text grows a buffer rather than copying one, so the assembly is linear and the DOM is handed the finished string once. Both bodies are pure text, so a variable holds exactly what the element would have held — no template needs to change.

Measured with that change, same corpus:

| | `to-java.xsl` | whole train | `string/printf.eo`, this pass |
|---|---|---|---|
| before | 25.4s | 31.2s | 3460 ms |
| after | 2.1s | 7.9s | 159 ms |

12x for the pass and 4x for the train, with byte-identical output for all 170 sources. The `DOMResult`/`StreamResult` gap closes too (159ms vs 144ms), which is the sign the quadratic term is gone rather than just smaller.

yegor256:
Closes #6674. Follow-up to #6666 — same mechanism, and after that fix `to-java.xsl` was **82% of the transpile train**.

## Where the time went

Per stylesheet, all 170 sources of `eo-runtime/src/main/eo`, best of two warm runs, Saxon-HE 13.0, single-threaded:

```
    725 ms  parse/set-locators.xsl              1205 ms  transpile/classes.xsl
    508 ms  transpile/set-original-names.xsl     649 ms  transpile/tests.xsl
    728 ms  transpile/anonymous-to-nested.xsl    675 ms  transpile/package.xsl
    683 ms  transpile/attrs.xsl                  585 ms  transpile/data.xsl
  25408 ms  transpile/to-java.xsl   <-- 82%
  31168 ms  TRAIN TOTAL
```

Saxon's `TimingTraceListener` on `string/printf.eo` put **8,328ms of the 8,881ms total in the net time of `xsl:template match="class"`** — the template that only opens `<tests>` and `<java>` and delegates. Net, so not the 26 templates it calls: that is the cost of handing text to the result tree.

## Why

The result tree is a DOM (`XSLDocument.transform` passes Saxon a `DOMSource` and a `DOMResult`). The templates produce the Java as thousands of small `xsl:text`/`xsl:value-of` pieces, each handed over as it is produced, and Saxon merges adjacent text into one DOM text node. Merging into a Xerces `Text` node is `setNodeValue(this.data + data)` — a full copy of everything accumulated.

So the pass was **quadratic in the size of the generated Java**. `<java>` for `string/printf.eo` is one text node of **484,853 characters**, and every piece paid for a copy of all the text before it. Retargeting the identical transform isolates it — same stylesheet, same input, warm:

| result | before | after |
|---|---|---|
| `DOMResult` (what the build uses) | 3460 ms | 159 ms |
| `StreamResult` | 91 ms | 144 ms |

The gap closing is the point: the quadratic term is gone, not just smaller.

## The change

Nine lines. Each body goes into a variable, then reaches the element through one `xsl:value-of`:

```xml
<java>
  <xsl:variable name="code">
    <xsl:call-template name="commonclass"/>
    <xsl:apply-templates select="." mode="body"/>
  </xsl:variable>
  <xsl:value-of select="$code"/>
</java>
```

A temporary tree is Saxon's own, where appending text grows a buffer rather than copying one. Both bodies are pure text, so the variable holds exactly what the element held before — **none of the 26 templates changed**.

| | `to-java.xsl` | whole train |
|---|---|---|
| before | 25.4s | 31.2s |
| after | **2.1s** | **7.9s** |

## What I ruled out first

`eo:eol()` and `eo:tabs()` are each called 4,257 times on one source and build a temporary tree per call, since neither declares `as="xs:string"` — the obvious suspect. Rewriting `$TAB`, `eo:eol` and `eo:tabs` as pure string functions moved the total by nothing measurable (24.5s → 26.1s, i.e. noise), and the trace agrees: 61ms and 27ms gross. That variant is **not** in this PR; it would be churn for no gain.

## Correctness

- Output **byte-identical** for all 170 `eo-runtime` sources — the generated Java itself.
- `MjTranspileTest` (52 transpile packs) and `FingerprintTest`: 57 tests green. That suite also drops from 34.5s to 14.0s.
- `xcop` clean on the changed file.
- `to-java.xsl` is fingerprinted into the transpile cache key via `Transpilation.XSLS`, so existing caches invalidate on their own.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GevVXivd1KDNnCHqKW9rua)_

### DIFF
--- a/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
+++ b/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
@@ -219,18 +219,42 @@
     <xsl:text>import org.eolang.*;</xsl:text>
     <xsl:value-of select="eo:eol(0)"/>
   </xsl:template>
-  <!-- Class. Entry point -->
+  <!--
+  Class. Entry point.
+
+  The Java of a class is assembled from thousands of small "xsl:text" and
+  "xsl:value-of" pieces by the templates below, and it must reach "<java>" as
+  one text node. Each piece is handed to the result tree as it is produced,
+  and that tree is a DOM, where merging a piece into the text node already
+  there copies everything accumulated so far - so writing the pieces straight
+  into "<java>" was quadratic in the size of the generated Java. The biggest
+  class of "eo-runtime" ends up at ~485k characters, and paying for all of it
+  again on every piece made this one pass four fifths of the whole transpile
+  train (#6674).
+
+  Collecting the pieces in a variable first costs nothing, because a temporary
+  tree is Saxon's own, where appending text grows a buffer instead of copying
+  one, and the single "xsl:value-of" then hands the DOM the finished string in
+  one piece. Both bodies are pure text, so a variable holds exactly what the
+  element would have held.
+  -->
   <xsl:template match="class">
     <xsl:copy>
       <xsl:apply-templates select="@*"/>
       <tests>
-        <xsl:call-template name="commonclass"/>
-        <xsl:apply-templates select="." mode="testing"/>
+        <xsl:variable name="code">
+          <xsl:call-template name="commonclass"/>
+          <xsl:apply-templates select="." mode="testing"/>
+        </xsl:variable>
+        <xsl:value-of select="$code"/>
       </tests>
       <xsl:if test="not(@skip-java)">
         <java>
-          <xsl:call-template name="commonclass"/>
-          <xsl:apply-templates select="." mode="body"/>
+          <xsl:variable name="code">
+            <xsl:call-template name="commonclass"/>
+            <xsl:apply-templates select="." mode="body"/>
+          </xsl:variable>
+          <xsl:value-of select="$code"/>
         </java>
       </xsl:if>
     </xsl:copy>


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (34) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2886](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6678, opened 2026-08-12
#6628: enforce atom-body check in Transition.apply for every line shape

### LINKED ISSUE #6628: Atom-body check in LnFormation is not enforced for application/reversed/compact-tuple children

morphqdd:
The atom-body rule (only `+>` test attributes, or void parameters, may sit under an atom) was enforced only in `LnFormation.checkChildAllowed`, reachable from `LnFormation.transition()` alone. The shared prologue every other line shape calls, `Transition.apply()`, duplicated the indent-step and openness checks but had no matching atom-body check. So `[] > foo /number` followed by an indented `42 > bar`, or `if. > bar ...`, parsed with zero errors and silently attached as a sibling of the atom marker, while the equivalent `[] > bar` formation child was correctly rejected.

Moved the check into `Transition.apply()`, used by `LnApplication`, `LnReversed`, `LnCompactTuple`, `LnVoid`, `LnPipe`, `LnTextBlock` and `LnOnlyPhi`.

`apply()` needed a fourth parameter (the permission flag) on top of kind/openness/label, which qulice flags for both parameter count and cyclomatic complexity. Introduced `Admission`, a small value object bundling the label and the permission flag, and extracted the push-branch validation into a private `pushed()` method.

`LnVoid` needed a special case: typed void parameter declarations (`? > x /A?`) are the sanctioned way to declare an atom's signature and are legitimately allowed under an atom, not test attributes. Found this from a real regression in `checksEoPacks` before fixing it (void declarations under an atom started failing to parse).

Added three `eo-typos` YAML fixtures (application, reversed, compact-tuple children under an atom) and two `TransitionTest` unit tests covering the shared rejection/permission logic directly.

Ran the full `eo-parser` test suite (1099 tests) and `qulice:check -Pqulice`, both green.

Closes #6628


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (199 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4080](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6657, opened 2026-08-12
Refactor package member code to use explicit '^' boundaries and merge package iterations


maxonfjvipon:
Part of #6616, and the step that actually removes the magic from the source. A package member takes its receiver in the first void today and the runtime writes it there behind the programmer's back:

```
[n x] > lt
  0.gt > @
    n.minus
      number x!
```

It should say what it means, which is that `lt` is a method of the object whose package holds it:

```
[x] > lt
  0.gt > @
    ^.minus
      number x!
```

Once it reads like that, `5.lt 4` is ordinary attribute dispatch on the merged formation, and `number.lt 5 4` is a mistake the compiler is entitled to report, because that is not what the code says.

The two changes cannot be split. While `PhDefault.absent` is alive, `[x] > lt` with `^` makes `5.lt 4` bind `x` to the receiver through `taken.put(0, this)`; with `absent` gone but no merge in place, `5.lt` terminates. So each package moves in one pull request: drop the receiver void from its members, replace it with `^`, rewrite the call sites that used the explicit `pkg.member a b` form, and add the package to the `mergedPackages` parameter of the merge mojo. Eleven packages, smallest first, and `chunk`, `range`, `output`, `u64` and `i64` hold one member each.

That list of members that never had a receiver was written off a text search and is wrong. Every one of them reads its first void as the receiver: `2.as-i64.div 0.as-i64`, `mem.to-output`, `"%d".printf`, `"%d".scanf`, `123.random`. The first batch has moved all four of its own to `^` — `i64/div`, `u64/div`, `chunk/to-output` and `range/naturals` — and the rest move with their packages like everything else. The only member so far that keeps nothing is `output/dead`, whose `origin` void no line ever read.

Around a hundred call sites in `eo-runtime/src/main/eo` use the explicit form and have to follow their package: `tuple.reduced` and `tuple.mapped` in `path.eo` and `set.eo`, `string.last-index-of` and `string.split` in `path.eo`, `bytes.as-input` across the `input` tests, `Q.input.copied` and `Q.input.tee` in the `input` package itself. A text search overstates the count, because `input.as-bytes` inside `[input] > hash` is a dispatch on a void that happens to be named after its type, so drive the rewrite off the `base` attributes in the parsed XMIR. One of the sites is `number.eo`'s own `can-raise-to-a-power-through-the-package`, which asserts the form being removed and should be deleted rather than rewritten.

maxonfjvipon:
[...]

Issue #6616, opened 2026-08-11
`PhDefault` wrongly resolves package dispatch at runtime instead of compile-time


maxonfjvipon:
When `take` in `eo-runtime/src/main/java/org/eolang/PhDefault.java:221` meets a name the object does not declare, it falls into `absent` at line 389, which probes the classpath for a package sibling (`extended`, line 408) and, on success, takes `Φ.<forma>.<name>` from the universe and binds the receiver into `α0` (`extension`, line 433). This is how `5.lt 3` becomes `number.lt 5 3`: a real convenience, since the surface of `number` can grow through small independent package objects instead of one huge formation — but the resolution happens at runtime, and that placement has three costs.

```java
private Phi absent(final String name) {
    final Phi object;
    if (this.extended(name)) {
        object = this.extension(name);
    } else if (this instanceof Atom) {
        object = this.take(Phi.LAMBDA).take(name);
    ...
```

First, performance: even with the probe cache in `OnClasspath`, every absent-name dispatch pays for a forma prefix check, a string join in `path`, two concurrent-map lookups, and a fresh `Phi.Φ.take` plus `put(0, this)` — per access, per object. Second, the φ-calculus knows nothing of this rule: in the calculus an absent attribute resolves only through the decoratee chain and otherwise the program terminates, so a program that leans on package extension means something different once translated from EO to φ, and the translation stops being meaning-preserving for exactly as long as the rule lives in the runtime alone. Third, caching semantics fork: a declared attribute returned by `take` comes memoized out of `loaded()` at line 226, while the object built by `absent` is constructed anew on every access — so `5.lt > x` is a cached dispatch when `lt` sits in the formation and a fresh application when it sits in the package, a difference the source syntax gives no hint of.

All three disappear if the resolution moves to compile time, where both the formation's attribute list and the package contents are already known: either merge package members into the object's formation during transpilation, or rewrite the dispatch site into the explicit `Φ.number.lt 5 3` application, and let the `extended` and `extension` branches of `absent` retire. Which of the two shapes is right is open, but the place is not: this is static knowledge, and the runtime should not be the component discovering it.

maxonfjvipon:
The solution is:

1. we need to make these package objects work with rho explicitly, not with its first void attribute which is replaced in runtime. So the number.lt was looked like this:


```
[n x] > lt
  0.gt > @
    n.minus
      number x!

```

now it will look like this:

```
[x] > lt
  0.gt > @
    ^.minus
      number x!

```

[...]

PR #6680, opened 2026-08-12
Give clashing test names the object they test

maxonfjvipon:
Twenty-six test names were shared by more than one member of the same package, across sixty-seven test attributes. That is legal today, since each member compiles to its own object and its own JUnit class, but it stops being legal the moment a package is merged into the object it belongs to (#6616), where two attributes cannot answer to one name. The three padding members were the worst of it, repeating almost their whole validation suite:

```
string/padded-left.eo    rejects-a-negative-width  ->  padded-left "y" -1 "x"
string/padded-right.eo   rejects-a-negative-width  ->  padded-right "y" -1 "x"
string/padded-zeros.eo   rejects-a-negative-width  ->  padded-zeros "y" -1
```

None of the twenty-six were redundant, so none are deleted: each asserts the same contract against a different object and now says which one, as in `rejects-a-negative-width-when-padding-left` or `stops-on-bytes-of-wrong-size-when-converting-to-i8`. The `bytes` package alone repeated that second name in nine members, one per conversion.

What was genuinely duplicated sits in the main objects, which smoke-tested behaviour their members already cover in full. `string.eo` asserted `"eEo".is-alpha`, character for character what `string/is-alpha.eo` asserts in `can-be-alphabetic-when-simple-text`; `number.eo` asserted `7.mod 3` is `1`, which is `number/mod.eo`'s `can-take-a-modulo-of-seven-by-three`; and `number.eo` reached for `sqrt`, `power` and `abs` while each of those members carries between six and fifty-two tests of its own. Sixteen such tests are dropped and left where they belong.

Left alone on purpose: the three tests named `*-through-the-package`, which assert the explicit `number.power 2 5` form. They go with the form itself, in #6657.

One renamed `-->` line outgrew the canonical width and `eo:format` rewrapped it to the vertical shape, which is the only change here that is not a name.

### DIFF
--- a/eo-runtime/src/main/eo/bytes/as-i16.eo
+++ b/eo-runtime/src/main/eo/bytes/as-i16.eo
@@ -19,4 +19,4 @@
 
   00-33.as-i16.as-bytes.eq 00-33 ++> can-convert-bytes-to-i16-and-back
 
-  01-01-01.as-i16 --> stops-on-bytes-of-wrong-size
+  01-01-01.as-i16 --> stops-on-bytes-of-wrong-size-when-converting-to-i16
--- a/eo-runtime/src/main/eo/bytes/as-i32.eo
+++ b/eo-runtime/src/main/eo/bytes/as-i32.eo
@@ -19,4 +19,4 @@
 
   00-00-00-33.as-i32.as-bytes.eq 00-00-00-33 ++> can-convert-bytes-to-i32-and-back
 
-  01-01-01-01-01.as-i32 --> stops-on-bytes-of-wrong-size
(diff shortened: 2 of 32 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (206 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+588](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6683, opened 2026-08-12
Name the number that does not fit into the long range

### LINKED ISSUE #6631: Inconsistent number naming in printf.eo causing mismatch issue
The puzzle `6359-8c33ec2e` from #6359 has to be resolved: 

https://github.com/objectionary/eo/blob/0cfaabb0a378e78500d521c74abad2d0f85224ef/eo-runtime/src/main/eo/string/printf.eo#L41-L46

Thayorns:
`printf "%d"` rejects any magnitude past the range of a signed 64-bit integer, but until now its reason carried no number at all: the guard could not name the value it was refusing, because every value it refuses is past what `as-decimal` and `as-fixed` can write, so a reason rendered through `%f` destroyed itself and the caller saw a chain of `Error in "..."` links instead of the message.

The new `string.as-scientific` writes any number as a six-digit mantissa, the letter `e` and the power of ten that scales it back, so `1.0e19` reads as `1.000000e19`. It divides the magnitude down to a mantissa below ten and renders that mantissa with `as-fixed`, which stays well inside the range `as-fixed` handles exactly, and it writes a non-finite number as `nan`, `infinity` or `-infinity`, since those carry no digits. The `%d` guard names its number through that object now, and the reason arrives whole.

The power of ten is never negative, so a magnitude below one keeps it at zero and `0.5` reads as `0.500000e0`. Nothing on this path ever hands it such a magnitude, and the docblock says as much; a caller that needs proper normalisation there deserves its own ticket.

Closes #6631

### DIFF
--- a/eo-runtime/src/main/eo/string/as-scientific.eo
+++ b/eo-runtime/src/main/eo/string/as-scientific.eo
@@ -15,8 +15,7 @@
 
 [n] > as-scientific
   n.is-finite.not.if > @
-    if.
-      n.is-nan
+    n.is-nan.if
       "nan".as-bytes
       if.
         n.lt 0
@@ -26,10 +25,12 @@
       concat.
         as-fixed
           n.div
-            10.power (number exponent)
+            10.power
+              number exponent
           6
         "e".as-bytes
-      as-decimal (number exponent)
+      as-decimal
+        number exponent
   rec-exponent > exponent!
     n.abs
     0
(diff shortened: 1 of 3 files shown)


Thayorns:
@yegor256 CI green here

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (100 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4166](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6685, opened 2026-08-12
Say what every dispatch turns out to be

### LINKED ISSUE #6684: Dispatch disrupts chain functionality causing premature termination in `eo-runtime` build
Three rules write down what a program says about itself: what a formation has, what is asked of an object, and which object is a copy of which. None of them says what a *computed* object is. A dispatch `.if` is an object the program makes while it runs, and the tables know only that somebody asked `if` of something — not what came back. So a chain stops at the first dispatch in it, and on a clean `eo-runtime` build 2,410 of 2,432 objects that delegate hand their answers to a locator no table describes.

The fact is there to be had, and it needs no call site. If a dispatch takes `name` from an object whose type is known, the dispatch **is** that attribute — a copy of it, in the sense `links.xml` already means. `Φ.bool.and.φ` is `.if` taken from `Φ.bool` (which #6653 made resolvable), `if` is a member of that package, so the body of `and` is a copy of `Φ.bool.if` and the rows can say so.

Look for the name in the three places there are: the type's own attributes, its package, since an unbound name falls through to the object beside it (#5486), and behind its `φ`. Write the pair, and go round again — `a.b.c` needs `a.b` before `.c` can be answered — until a pass adds nothing. Pairs are only ever added, so it settles. A receiver whose type is a void gets no answer, because a void has none until somebody fills it; that is the contextual half of the problem and it is not this ticket.

The rows belong in `links.xml`. "This object is a copy of that one" is exactly what that document says, and a second document saying the same thing in different words would only invite the two to disagree. So the phase reads the tables the rules wrote, closes the links under dispatch resolution, and writes the document back with what it worked out added to it.

This is the same mechanism the Python prototype uses in `sought()`, whose failure message names the same three places: own attribute, package member, `φ` to descend into. It differs in two ways worth keeping in view. It recurses on demand and memoises on `(locator, context)` rather than iterating over a table, which is bookkeeping and not meaning. And it carries a context of binding frames, so a dispatch on a void resolves there and stays unanswered here.

What to measure, against the numbers of a clean runtime build: 10,145 dispatch receivers, of which a link resolves 6,147; 22,069 links rows; 2,410 delegation chains that end at a locator nothing describes.

maxonfjvipon:
Three rules write down what a program says about itself, and none of them says what a *computed* object is. A dispatch `.if` is made while the program runs, so the tables knew only that somebody asked `if` of something, never what came back. A chain of dispatches stopped at the first one, and an object that hands its answers to a dispatch handed them to a locator nothing described.

The fact is there to be had and no call site is needed for it. When a dispatch takes a name from an object whose type is known, and that type has such an attribute, the dispatch **is** that attribute — a copy of it, which is what a pair in `links.xml` means. `Dispatched` works out those pairs, `Resolved` asks it again until a pass adds nothing, since answering `a.b.c` needs `a.b` answered first:

```java
Map<String, String> answers = made.answers(pairs);
while (!answers.isEmpty()) {
    pairs.putAll(answers);
    answers = made.answers(pairs);
}
```

Pairs are only ever added, of which there are finitely many, so it settles. What comes out goes into `links.xml` beside the rules' own rows, rather than into a second document that would only invite the two to disagree.

Where a name is looked for is `Provided`'s business: the type's own attributes, then its package, since a name nobody binds falls through to the object beside it (#5486), then whatever stands behind its `φ`, walked once so an object that delegates in a circle answers nothing. Nothing is guessed. A receiver whose type nothing describes is left alone, and so is one that is a void: `x.next` is a different object for every caller, and the pair that settles it belongs to the call site.

Measured on a clean `eo-runtime` build, this branch with the phase off and then on: links rows 21,953 → 26,347; dispatch receivers whose type is known 6,128 → 8,098 out of 10,114; dispatches that have a type of their own 0 → 4,394; delegation chains that end at a locator nothing describes 2,118 → 1,760 out of 2,410. The `complete` flag does not move, that being `Provides`' business.

Six packs describe the behaviour, one each for the attribute a dispatch takes, a chain answered in order, a void left alone, a member of the package, a look behind delegation, and a circle walked once. What is left is the contextual half: a fresh type per copy, so that a dispatch on a void is answered where the void is filled.

Closes #6684

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Dispatched.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Dispatched.java
@@ -1,0 +1,79 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.xml.XML;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * What every dispatch of a program turns out to be.
+ *
+ * <p>A dispatch takes a name from an object, so when that object's type is
+ * known and it has such an attribute, the dispatch <em>is</em> that attribute:
+ * a copy of it, which is what a pair in {@link Links} means. The body of
+ * {@code and} is {@code .if} taken from {@code Φ.bool}, {@code if} is a member
+ * of that package, so the body is a copy of {@code Φ.bool.if}. Where the
+ * attribute is looked for is {@link Provided}'s business — the object itself,
+ * its package, or behind its {@code φ}.</p>
+ *
+ * <p>Nothing is guessed. A receiver whose type nothing describes is left
+ * alone, and so is a receiver that is a void: {@code x.next} inside an object
+ * that takes {@code x} is one object in the text and a different one for every
+ * caller, and the pair that would settle it belongs to the call site rather
+ * than here. A dispatch already spoken for is left alone too, since a pass is
+ * only ever asked for what the last one could not answer.</p>
+ *
+ * @since 0.68.0
+ */
+final class Dispatched {
+
+    /**
+     * The provides table.
+     */
+    private final XML given;
+
+    /**
+     * Every dispatch of the program.
+     */
+    private final Collection<XML> all;
+
+    /**
+     * Ctor.
+     * @param provides The provides table
+     * @param dispatches Every dispatch of the program
+     */
+    Dispatched(final XML provides, final Collection<XML> dispatches) {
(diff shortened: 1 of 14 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (613 >= 100); -8 for way too many hits-of-code (613 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1044](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6686, opened 2026-08-12
Dispatch package members in their own tests

### LINKED ISSUE #6682: Package member tests incorrectly apply standalone object instead of correctly dispatching it
Part of #6616. A package member is reached in two ways today, and its own tests mostly pick the wrong one. `string/truncated.eo` applies the object standalone:

```
truncated --> stops-on-a-negative-limit-when-truncating
  "Jeff"
  -3
```

which works only because the receiver arrives in the first void, `text`. Once that void becomes `^` (#6657), the same three lines do not fail: they bind `"Jeff"` to `limit` and `-3` to `cant-truncate`. A test that silently asserts something else is worse than one that breaks, so the form has to go before the void does.

The dispatched form says the same thing and survives the change, because both forms resolve through the package rule today:

```
truncated. --> stops-on-a-negative-limit-when-truncating
  "Jeff"
  -3
```

There are 577 standalone uses across 79 members. Thirty-four members are already free of them, among them `number/lt`, `number/minus`, all nine `bytes` conversions and every member of `input`, `file` and `directory`, which is the proof that the dispatched form works on today's runtime and that this can land ahead of any merge machinery. Seventy-three members, on the other hand, have no dispatched test at all, so nothing in their file would notice the silent rebinding.

Cross-member calls count too. `tuple/back.eo` reaches its sibling as `is-empty (* 1 2)`, and that line dies with the merge exactly like an own-member one, so it becomes `(* 1 2).is-empty`.

Where a dispatched test already covers the same case, the standalone one is a duplicate and should be deleted rather than rewritten: `number/eq.eo` has one standalone use against forty-eight dispatched, `number/as-i64.eo` one against eleven, `string/slice.eo` four against twelve.

Six members have to be left alone, all of them because their first void is not an instance of the enclosing package object. `bytes/hash.eo`, `string/as-decimal.eo`, `as-fixed.eo`, `as-char.eo` and `as-hex.eo` cannot be dispatched until #6655 puts them in the package of their receiver, since `65.as-char` probes `Φ.number.as-char` and finds nothing. And `number/integral.eo` takes a function first, not a number.

`string/printf.eo` and `string/scanf.eo` are not among them, contrary to what this issue first claimed: both take a receiver, only they declare it as `? > format` inside the body instead of in the header, which is why a header-only reading of the voids missed it. Their compact-tuple heads do need care, though. `printf "%.3s" *1 "Jeffrey"` cannot become `printf. *1 …`, since PARSER_SPEC R-3.9.1 lists a bare reversed head as legal in a compact tuple but the parser rejects it, so those heads spell the tuple out.

This depends on #6654, which renames some of the very lines that need rewriting.

maxonfjvipon:
Tests living inside a package member often applied that member standalone, by name, and passed the receiver as the first argument:

```
truncated --> stops-on-a-negative-limit-when-truncating
  "Jeff"
  -3
```

That only works while the member is a separate object sitting next to `string`. After `MjMerge` folds members into the same-named object, the bare name has nothing left to resolve, so every such site now goes through dispatch instead:

```
"Jeff".truncated -3 --> stops-on-a-negative-limit-when-truncating
```

577 sites across 79 files, own-member and cross-member alike (`tuple/back.eo` calling `is-empty`, for one). Six members keep the standalone form on purpose — `bytes.hash`, `number.integral`, `string.as-char`, `string.as-decimal`, `string.as-fixed`, `string.as-hex` — because their first void is not an instance of the enclosing package object, so there is no receiver to dispatch on.

Compact tuples needed spelling out. `printf "%.3s" *1 "Jeffrey"` cannot become `printf. *1 …`: PARSER_SPEC R-3.9.1 lists a bare reversed head as legal there, but the parser rejects it, so those 56 heads carry an explicit `*` instead. Worth filing separately.

Test count is unchanged, 1596 markers before and after; every hunk sits below its file's first marker, so no production code moved. The 1204 `inconsistent-args` warnings from a standalone `eo:lint` are the same 1204 master already emits — the build's own `compile` execution skips that rule.

Closes #6682.

### DIFF
--- a/eo-runtime/src/main/eo/bytes/array.eo
+++ b/eo-runtime/src/main/eo/bytes/array.eo
@@ -25,7 +25,7 @@
     0
 
   eq. ++> can-convert-bytes-to-an-array
-    array 20-1F-EE-B5-FF
+    20-1F-EE-B5-FF.array
     *
       20-
       1F-
@@ -34,9 +34,9 @@
       FF-
 
   eq. ++> can-convert-a-single-byte-to-an-array
-    array 1F-
+    1F-.array
     * 1F-
 
   eq. ++> can-convert-zero-bytes-to-an-array
-    array --
+    --.array
     *
--- a/eo-runtime/src/main/eo/bytes/as-input.eo
+++ b/eo-runtime/src/main/eo/bytes/as-input.eo
@@ -66,8 +66,6 @@
       eq.
         i3.read 2
         --
-    read. > i1
-      as-input 01-02-03-04-05-06-07-08-F5-F6
-      4
+    01-02-03-04-05-06-07-08-F5-F6.as-input.read 4 > i1
     i1.read 4 > i2
     i2.read 4 > i3
--- a/eo-runtime/src/main/eo/chunk/to-output.eo
+++ b/eo-runtime/src/main/eo/chunk/to-output.eo
@@ -40,9 +40,7 @@
       seq * > [mem]
         write.
           write.
-            write.
-              to-output mem
-              01-02-03
+            mem.to-output.write 01-02-03
             04-05-06-07
           08-09-A0
         mem
@@ -60,8 +58,6 @@
     malloc.of
       2
       seq * > [mem]
-        write.
-          to-output mem
-          01-02-03
+        mem.to-output.write 01-02-03
         mem
     01-02-03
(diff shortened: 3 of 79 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (2225 >= 100); -8 for way too many hits-of-code (2225 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+680](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6687, opened 2026-08-12
#6651: cache half in ipow to restore O(log n) squaring

### LINKED ISSUE #6651: `power` recomputes `half` and `k` on every read in `number/power.eo`
The `ipow` helper in `eo-runtime/src/main/eo/number/power.eo:55` binds `ipow b (n.div 2) > half` without the `!` const marker, and line 52 reads it twice as `half.times half`. Every even step therefore dataizes its own recursive call twice over, so the recurrence is `T(n) = 2*T(n/2)` and exponentiation-by-squaring costs `O(n)` multiplications instead of the `O(log n)` it is written to achieve. The `can-raise-to-a-large-negative-exponent` test at line 310 raises `984782` to `-12341`, which walks about 12341 recursive dataizations where 14 squarings would do.

The `expon` helper next to it has the same defect twice over. `k` is bound at line 75 as `floor. (y.plus 0.5) > k`, again with no `!`, and it is read twice in `@` (lines 63 and 65-66) and once per each of the twenty `fraction` levels at line 73, so it is dataized about 22 times. Each of those reads the void `y`, which line 76 fills with `(expo.times (ln base))`, and `fraction` reads `y` directly at every level as well, which comes to roughly 42 evaluations of `ln` for one call. `sqrt` is `power num 0.5` and `exp` is `power e num`, so every square root and every exponential in the runtime pays this.

Marking both bindings const, as `> half!` and `> k!`, and re-wrapping each use with `number`, collapses all of it to one evaluation apiece; that is how #6224 was fixed for the folding point of `arc-sine`. A void cannot carry `!` (R-3.4.7), so `y` needs the same treatment through a separate const bound inside `expon`.

@yegor256
@maxonfjvipon

morphqdd:
`ipow`'s `half` binding (`ipow b (n.div 2) > half`) had no `!` const marker, and `half.times half` read it twice, so every even step of exponentiation-by-squaring dataized its own recursive call twice over: `T(n) = 2*T(n/2)`, giving O(n) multiplications instead of the O(log n) the algorithm is written for. `can-raise-to-a-large-negative-exponent` (984782 ^ -12341) walks about 12341 recursive dataizations where 14 squarings would do.

Cached `half` as `half!` and re-wrapped both reads with `number`, mirroring the `half`/`bts` idiom already used to fix the same class of bug for `arc-sine` in #6224.

`expon`'s `k` and `y` have the same defect (about forty evaluations of `ln` per call instead of one), but fixing that needs binding the void `y` to a const local first and re-wrapping several distinct read sites (`k.lt 0`, `abs k`, `y.minus k`), which is a separate, more careful piece of work. Left as a `@todo` puzzle referencing this issue rather than risking a repeat of the regression #6224 hit the first time this idiom was tried carelessly.

Ran `EOnumber.EOpowerTest` (0 failures) and `qulice:check -Pqulice`, both green.

Closes #6651

### DIFF
--- a/eo-runtime/src/main/eo/number/power.eo
+++ b/eo-runtime/src/main/eo/number/power.eo
@@ -3,6 +3,14 @@
 # results stay exact, and a fractional exponent of a positive base is taken as
 # e raised to `x` times the natural logarithm of the base. Zeros, infinities and
 # nan follow the same IEEE rules that Math.pow obeys.
+#
+# @todo #6651:30min Cache `k` and `y` in `expon` the same way `half` is now
+#  cached in `ipow`. `k` is read once per comparison and once per each of the
+#  twenty `fraction` levels, and `y` (a void, so it cannot itself carry `!`)
+#  is read directly inside `fraction` too, so one call to `expon` still pays
+#  for about forty evaluations of `ln`. Bind `y` to a const local first, then
+#  cache `k` as `k!` and re-wrap every read with `number`, mirroring the
+#  `half`/`bts` idiom from #6224.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
@@ -49,10 +57,12 @@
                         1
                         if.
                           is-integer. (n.div 2)
-                          half.times half
+                          times.
+                            number half
+                            number half
                           b.times
                             ipow b (n.minus 1)
-                      ipow b (n.div 2) > half
+                      ipow b (n.div 2) > half!
                     | base (abs expo)
                   ipow base (abs expo)
                 if.


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3984](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6688, opened 2026-08-12
Seal a bottom with its reason when it propagates through a dispatch

### LINKED ISSUE #6632: Prevent causeless bottom event in PhTerminator.java
The puzzle `6359-ae89e11e` from #6359 has to be resolved: 

https://github.com/objectionary/eo/blob/0cfaabb0a378e78500d521c74abad2d0f85224ef/eo-runtime/src/main/java/org/eolang/PhTerminator.java#L37-L41

Thayorns:
A bottom that reached a dispatch used to answer with itself, so the arguments of whatever call followed it landed in its cause slot. `(T).if 1 2` took the `1` as the reason it terminated and then panicked with those bytes read as text, which said nothing about the computation having been terminated at all.

Now a dispatch hands the bottom it yields a reason at birth: the cause the original was carrying, or the plain explanation when it had none. Since the slot is write-once, the arguments that follow are dropped instead of being mistaken for a reason, and a cause given to `T "why"` now survives the dispatches it travels through rather than only surviving them by accident.

The same swallowing still happens at the two other places a causeless bottom is born, `AtVoid.get` for an unbound void and `PhDefault.absent` for a missing attribute, when one of them is applied with no dispatch in between, as in `obj.missing a b`. Closing that needs each site to name what went missing, which changes messages more widely, so it is left as a puzzle.

Closes #6632

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhTerminator.java
+++ b/eo-runtime/src/main/java/org/eolang/PhTerminator.java
@@ -129,7 +129,6 @@ public final class PhTerminator implements Phi {
     /**
      * The reason to panic with, either the cause this bottom was given or the
      * plain explanation of a bottom that was born without one.
-     *
      * @return The reason as an object
      */
     private Phi reason() {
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 ping

0crat:
@thayorns Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1248](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6595, opened 2026-08-11
`Emissions.appendOctal` is unreachable dead code


morphqdd:
## What happens

`Emissions.appendOctal(...)` in `eo-parser/src/main/java/org/eolang/parser/Emissions.java:629-650` is a private method with exactly one caller in the whole codebase, at line 258. That caller is itself unreachable from any real entry point (nothing in the codebase invokes it), so `appendOctal` never actually runs.

PMD's `UnusedPrivateMethod` check did not flag it, because it does have a caller in the source, even though that caller is dead.

For contrast, the sibling helper `appendUnicode` (line 662) is genuinely live: it is called both from the dead path and from a second, real call site at line 562, so it stays reachable.

## What should happen

`appendOctal` should be removed along with its now-orphaned caller, since it duplicates the octal-escape decoding already handled by the live byte-based path (`rawOctal`, used through `unescapeBytes`).

PR #6689, opened 2026-08-12
#6594 #6595: remove unescapeBody and its exclusive appendOctal helper

### LINKED ISSUE #6594: `Emissions.unescapeBody` is dead code with no callers
## What happens

`Emissions.unescapeBody(String)` in `eo-parser/src/main/java/org/eolang/parser/Emissions.java:244-263` has no callers anywhere in the codebase. Running `grep -rn "unescapeBody" --include=*.java .` outside its own file and its own definition returns nothing, and no test in `eo-parser/src/test/` references it either.

The class has a second, separate method, `unescapeBytes` (line 268), which is the one actually used by real callers such as `Emissions.string()` and `LnTextBlock.java:64`. `unescapeBody` decodes escapes into a `String` using a different, older code path than `unescapeBytes`, and the two can drift out of sync over time even though `unescapeBody` currently does nothing at runtime.

It was not caught by PMD's `UnusedPrivateMethod` check because the method is package-private, not `private`.

## What should happen

`unescapeBody` should be removed, since it is unreachable dead code and the codebase already has a working, exercised replacement (`unescapeBytes`).

morphqdd:
`Emissions.unescapeBody` has no callers anywhere in the repository (`grep -rn "unescapeBody\b"` outside its own definition returns zero hits) — #6594. Its only caller of `appendOctal` was itself, so once `unescapeBody` is gone `appendOctal` becomes unreachable too — #6595. Both are removed together since they were only reachable through each other; leaving one without the other would still be dead code.

`appendUnicode`, `singleCharEscape`, `rejectLoneSurrogates` and `MAX_OCTAL_BYTE` all stay, they're still used by the live `unescapeBytes` sibling method.

Ran the full `eo-parser` test suite (1085 tests) and `qulice:check -Pqulice`, both green.

Closes #6594, #6595

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -234,38 +234,6 @@ final class Emissions {
     }
 
     /**
-     * Decode escape sequences from a raw string body (without
-     * surrounding quotes). Supports {@code \n}, {@code \t}, {@code \r},
-     * {@code \b}, {@code \f}, {@code \"}, {@code \'}, {@code \\},
-     * {@code \\uXXXX} unicode, and {@code \NNN} octal escapes.
-     * @param inner Source body (no quotes)
-     * @return Decoded text
-     */
-    static String unescapeBody(final String inner) {
-        final StringBuilder out = new StringBuilder(inner.length());
-        int idx = 0;
-        while (idx < inner.length()) {
-            final char glyph = inner.charAt(idx);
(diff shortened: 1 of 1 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4000](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6690, opened 2026-08-12
#6447: skip createDirectories when atomsOutput has no parent

### LINKED ISSUE #6447: `MjAtomsTable.write()` throws `NullPointerException` when `eo.atomsOutput` is a filename without a parent directory

morphqdd:
`MjAtomsTable.write()` called `Files.createDirectories(target.getParent())` unconditionally. For a parentless relative path like `atoms.csv`, `getParent()` is null, so `createDirectories` throws a `NullPointerException` before the table is written.

Skip the directory creation when there is no parent to create.

Added a test that configures `atomsOutput` as a bare filename and verifies the file is written instead of throwing.

Closes #6447

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjAtomsTable.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjAtomsTable.java
@@ -125,7 +125,10 @@ public final class MjAtomsTable extends MjSafe {
      */
     private void write(final Map<String, String> table) throws IOException {
         final Path target = this.atomsOutput.toPath();
-        Files.createDirectories(target.getParent());
+        final Path parent = target.getParent();
+        if (parent != null) {
+            Files.createDirectories(parent);
+        }
         final StringBuilder out = new StringBuilder();
         for (final Map.Entry<String, String> entry : table.entrySet()) {
             out.append(entry.getKey())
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (23) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4012](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6691, opened 2026-08-12
Split the value before scaling so large magnitudes keep their digits

### LINKED ISSUE #6634: Scaling of large values in as-fixed.eo leads to loss of precision
The puzzle `6572-4ba6a830` from #6572 has to be resolved: 

https://github.com/objectionary/eo/blob/2225e4051db794ef46924d25617476fcab5afb52/eo-runtime/src/main/eo/string/as-fixed.eo#L8-L13

Thayorns:
`as-fixed` used to multiply the whole value by `10^places` and then divide the product back down to recover the integer part. Past 2^53 that product no longer lands on an exact double, so the integer part handed to `as-decimal` came out short: `1.0e17` with six places printed `99999999999999984.000000`, while `as-decimal` writes the same magnitude exactly because it peels its digits off an `i64`. `printf`'s `%f` inherited the same misprint, so `%f` and `%d` disagreed on the same number.

Now the value is split into its integer and fraction parts first, and only the fraction — always below one — is scaled. The integer part travels to `as-decimal` untouched, so every magnitude `as-decimal` writes exactly is written exactly here too. Rounding that spills over one, like `0.9999999` at six places, is carried into the integer part explicitly instead of falling out of the division.

Four tests cover it: the two magnitudes past 2^53, positive and negative, both of which fail on the old code, plus the carry into the integer part at zero places, which the rewrite reaches through a new path.

Closes #6634

### DIFF
--- a/eo-runtime/src/main/eo/string/as-fixed.eo
+++ b/eo-runtime/src/main/eo/string/as-fixed.eo
@@ -5,12 +5,9 @@
 # or infinite), the `cant-print` fallback is returned, or the bottom object
 # when unbound.
 #
-# @todo #6572:30min Scale a large value without losing its last digits. The
-#  `positive` object below multiplies the value by a power of ten in double
-#  arithmetic, so a product above 2^53 is rounded and the integer part handed
-#  to `as-decimal` can be off by one, which makes `printf`'s `%f` misprint a
-#  magnitude past 2^53 that `%d` now prints exactly; split the value into its
-#  integer and fraction parts over `i64` before scaling.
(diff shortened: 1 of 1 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1240](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6692, opened 2026-08-12
#6378: defensively copy byte-array data in PhDefault

### LINKED ISSUE #6378: `PhDefault` allows mutation of byte arrays after object construction, violating data immutability

morphqdd:
`PhDefault`'s byte-array constructor stored its argument by reference, and `delta()` returned that same reference. A caller could mutate an EO value after construction through the array it originally passed in, or mutate every later dataization through an array obtained from `delta()`.

Defensively copies the array on construction and on every `delta()` read, matching the idiom `BytesRaw`/`BytesOf` already use.

`copy()` still shares the internal array with the original via `Object.clone()` (shallow), but that's safe now: nothing inside `PhDefault` ever mutates `this.data` in place, and no method leaks the raw reference anymore, only snapshots.

Added `PhDefaultDataTest` (a separate file since `PhDefaultTest` was already at the 1000-line checkstyle cap) reproducing the two mutation paths from the issue.

Closes #6378


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (125 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4024](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6693, opened 2026-08-12
Account for the attributes a package hands to its owner

### LINKED ISSUE #6635: Attributes in Provides.java not properly accounted for
The puzzle `6565-60af8077` from #6565 has to be resolved: 

https://github.com/objectionary/eo/blob/6d7f8125b96462f0c8f9ffe3267d9bbf360614ea/eo-inference/src/main/java/org/eolang/inference/Provides.java#L58-L64

Estimate: 35 minutes,  role: DEV. 

Thayorns:
The provides table listed only what a formation writes inside its own body, so an attribute that lives in a file of its own was missing from it. A package is a prefix of a locator: `minus` in the package `number` is `Φ.number.minus` and belongs to `Φ.number`, yet it never shows up among the children of that formation. The table called `Φ.number` complete while listing a strict subset of what it has, which is exactly the kind of half-truth a later check would read as a missing attribute.

Now `Provides` also reads the top-level object of every file and writes it down as an attribute of the object formed under its prefix. `Xmirs` grew `roots()` for that, since the named object at the top of a file is the only place a package shows itself. The row is written only when the program actually forms something under the prefix — a package nobody declares as an object stays a prefix and nothing else, and inventing an owner for it would put a type into the table that the program does not have.

These rows come after every attribute a formation writes out itself. The order of attributes is what binds the arguments of an application, and a file of a package binds none of them, so appending keeps the voids where they were.

Two packs cover both sides: one where the owner exists and picks up the extra attribute, one where the package names nothing and the table stays quiet.

The second puzzle in that docblock, about the `ρ` every object has, is untouched and still open.

Closes #6635

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Members.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Members.java
@@ -1,0 +1,71 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.xml.XML;
+import com.yegor256.tojos.Tojos;
+import java.util.Collection;
+import java.util.HashSet;
+
+/**
+ * What the packages of a program hand to the objects they are named after.
+ *
+ * <p>A package is a prefix of a locator, so {@code minus} in the package
+ * {@code number} is {@code Φ.number.minus} and belongs to {@code Φ.number}
+ * without ever appearing among the children of that formation. Reading the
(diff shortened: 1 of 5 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (141 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1232](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6695, opened 2026-08-12
`number.floor` loses the sign of negative zero during integer conversion


gemshrine:
## What happens

`number.floor` converts a negative zero to a positive zero. The two values are
numerically equal, but EO preserves and exposes their IEEE 754 bit patterns
through `as-bytes`.

| expression | actual bytes | expected bytes |
| --- | --- | --- |
| `(number.floor (0.neg)).as-bytes` | `00-00-00-00-00-00-00-00` | `80-00-00-00-00-00-00-00` |

The case was reproduced on the current `master` (`7591007fd7`) through the
runtime atom. Java's `Math.floor(-0.0)` keeps the negative-zero bit pattern.

## Why it happens

[`floor.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number/floor.eo#L11)
delegates finite small values to `as-i64.as-number`. Integer conversion has no
signed-zero representation, so the `-0.0` sign is discarded before the result
is returned.

## Expected behavior

Flooring an already integral number should retain its value exactly, including
the sign of zero:

```eo
80-00-00-00-00-00-00-00
  as-bytes.
    number.floor 0.neg
```

## Suggested fix

Add a signed-zero fast path before converting through `as-i64`, returning the
original `n` when its bytes equal `80-00-00-00-00-00-00-00`. A regression test
must compare `as-bytes`, since numeric equality cannot distinguish the zeros.

morphqdd:
@yegor256, PR #6703 fixing this was merged but I forgot the `Closes` keyword, so it didn't auto-close. Could you close this one manually? Sorry for the extra step.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2040](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

gemshrine:
Resolved by #6703.

PR #6703, opened 2026-08-12
#6695: preserve the sign of negative zero in number.floor

morphqdd:
`number.floor` converts a negative zero to a positive zero, because the finite fast path delegates to `n.as-i64.as-number`, and integer conversion has no signed-zero representation.

Added a fast path before the existing magnitude check: when `n`'s bytes are exactly the negative-zero bit pattern, return `n` unchanged.

Added `can-floor-negative-zero-preserving-sign`, comparing `as-bytes` since numeric equality can't distinguish the two zeros.

Ran `EOfloorTest` (11 tests, 0 failures) and `qulice:check -Pqulice`, both green.

### DIFF
--- a/eo-runtime/src/main/eo/number/floor.eo
+++ b/eo-runtime/src/main/eo/number/floor.eo
@@ -10,17 +10,20 @@
 
 [n] > floor
   if. > @
-    n.is-finite.not.or
-      or.
-        n.gte 4503599627370496
-        n.lte -4503599627370496
+    n.as-bytes.eq 80-00-00-00-00-00-00-00
     n
     if.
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4036](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6694, opened 2026-08-12
`number.arc-sine` loses the sign of negative zero, breaking the sine round trip


gemshrine:
## What happens

`number.arc-sine` loses the sign of a negative zero. The operation returns
`+0.0` for `-0.0`, although `Math.asin(-0.0)` returns `-0.0` and the signed
zero is observable in EO through `as-bytes`.

| expression | actual bytes | expected bytes |
| --- | --- | --- |
| `(number.arc-sine (0.neg)).as-bytes` | `00-00-00-00-00-00-00-00` | `80-00-00-00-00-00-00-00` |

The case was reproduced on the current `master` (`7591007fd7`) through the
runtime atom. It also breaks the natural round-trip property for this input:
`number.sine (number.arc-sine (0.neg))` produces `+0.0`, not the original
`-0.0`.

## Why it happens

[`arc-sine.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number/arc-sine.eo#L16)
branches only when `arg < 0`. IEEE `-0.0 < 0.0` is false, so the negative
branch is skipped. The non-negative branch invokes the series with the
absolute magnitude, which is `+0.0`, and cannot restore the sign.

## Expected behavior

`number.arc-sine` should preserve the signed zero, consistently with the
documented inverse-trigonometric meaning and `Math.asin`:

```eo
80-00-00-00-00-00-00-00
  as-bytes.
    number.arc-sine 0.neg
```

## Suggested fix

Handle the negative-zero bit pattern before the `< 0` branch, returning
`0.neg`, or retain the original sign when the calculated result is zero. Add a
regression test that compares `as-bytes` rather than numeric equality, since
`+0.0.eq(-0.0)` cannot distinguish the values.

morphqdd:
@yegor256, PR #6705 fixing this was merged but I forgot the `Closes` keyword, so it didn't auto-close. Could you close this one manually? Sorry for the extra step.

0crat:
[...]

PR #6705, opened 2026-08-12
#6694: preserve the sign of negative zero in number.arc-sine

morphqdd:
`number.arc-sine` branches only on `arg < 0`. IEEE `-0.0 < 0.0` is false, so a negative-zero input falls into the non-negative branch, which computes the series from `abs(arg)` and never restores the sign. The result loses the negative-zero bit pattern and breaks the round trip with `number.sine`.

Added a fast path before the existing sign check: when `arg`'s bytes are exactly the negative-zero pattern, return `arg` unchanged. Moved the `arg` binding out to a sibling of the outer `if.` so both branches can reference it.

Added `can-take-an-arc-sine-of-negative-zero-preserving-sign`, comparing `as-bytes` since numeric equality can't distinguish the two zeros.

Ran `EOarc_sineTest` (13 tests, 0 failures) and `qulice:check -Pqulice`, both green.

### DIFF
--- a/eo-runtime/src/main/eo/number/arc-sine.eo
+++ b/eo-runtime/src/main/eo/number/arc-sine.eo
@@ -13,39 +13,41 @@
 
 [num] > arc-sine
   if. > @
-    lt.
-      number num.as-bytes >> arg
-      0
-    neg.
-      if. >> base
-        gt.
-          abs arg >> magnitude
-          0.5
-        minus.
-          pi.div 2
-          2.times
-            [point] >> series
-              point.times > @
-                poly 1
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3960](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6696, opened 2026-08-12
`number.sqrt` loses the sign of negative zero through `number.power`


gemshrine:
## What happens

`number.sqrt` loses the sign of a negative zero. The returned value is `+0.0`,
while IEEE 754 and Java's `Math.sqrt(-0.0)` preserve `-0.0`.

| expression | actual bytes | expected bytes |
| --- | --- | --- |
| `(number.sqrt (0.neg)).as-bytes` | `00-00-00-00-00-00-00-00` | `80-00-00-00-00-00-00-00` |

This was reproduced through the runtime atom on current `master` (`7591007fd7`).

## Why it happens

[`sqrt.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number/sqrt.eo#L11)
implements square root as `power num 0.5`. `-0.0` is not less than zero, so it
reaches [`power.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number/power.eo#L28).
The zero-base branch only restores a negative sign for an odd integer exponent;
`0.5` therefore takes the positive-zero result.

## Expected behavior

The square root should preserve signed zero:

```eo
80-00-00-00-00-00-00-00
  as-bytes.
    number.sqrt 0.neg
```

## Suggested fix

Add a signed-zero branch to `sqrt` before delegating to `power`, or make the
zero-base path in `power` preserve `-0.0` for positive non-integer exponents
where IEEE 754 does so. Add the regression test through `as-bytes`.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2052](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

morphqdd:
@yegor256 close this, it's solved.

gemshrine:
Resolved by #6708.

PR #6708, opened 2026-08-12
#6696: preserve the sign of negative zero in number.sqrt

morphqdd:
`number.sqrt` is `power num 0.5`. `power`'s zero-base branch only restores a negative sign for an odd integer exponent, so `0.5` (not an integer) takes the positive-zero path and the sign of a negative-zero input is lost.

Fixed in `sqrt.eo` directly rather than in `power.eo`, to keep this change isolated: added a fast path that returns the argument unchanged when it's exactly the negative-zero bit pattern, before delegating to `power`.

Added `can-take-a-square-root-of-negative-zero-preserving-sign`, comparing `as-bytes` since numeric equality can't distinguish the two zeros.

Ran `EOsqrtTest` (12 tests, 0 failures) and `qulice:check -Pqulice`, both green.

### DIFF
--- a/eo-runtime/src/main/eo/number/sqrt.eo
+++ b/eo-runtime/src/main/eo/number/sqrt.eo
@@ -10,13 +10,13 @@
 
 [num] > sqrt
   if. > @
-    lt.
-      number num.as-bytes
-      0
+    arg.lt 0
     nan
-    power
-      num
-      0.5
+    if.
+      arg.as-bytes.eq 80-00-00-00-00-00-00-00
+      arg
+      power arg 0.5
+  number num.as-bytes > arg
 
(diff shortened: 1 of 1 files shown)


yegor256:
@morphqdd conflicts here

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4092](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6700, opened 2026-08-12
`tuple.withouti` fails to explicitly reject fractional indexes, resulting in silent no-op


gemshrine:
## What happens

`tuple.withouti` silently accepts fractional indexes and returns the original
tuple, because none of its integer element positions equals the requested
fraction.

| expression | result |
| --- | --- |
| `tuple.withouti (* "a" "b") -0.5` | `* "a" "b"` |
| `tuple.withouti (* "a" "b") 0.5` | `* "a" "b"` |
| `tuple.withouti (* "a" "b") 1.5` | `* "a" "b"` |
| `tuple.withouti (* "a" "b") 1` | `* "a"` |

The behaviour was reproduced through runtime objects on current `master`
(`7591007fd7`).

## Why it happens

[`withouti.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/tuple/withouti.eo#L11)
compares the supplied `i` directly with each integral `index` produced by
`tuple.reducedi`. It never validates that `i` itself is integral, so a
fractional value cannot match any element and removal becomes a silent no-op.

## Expected behavior

The argument identifies an element position. Fractional indexes should be
rejected explicitly, consistently with `tuple.slice`, rather than silently
leaving the tuple unchanged.

## Suggested fix

Check `i.is-integer` before reducing the tuple and return a clear failure or
fallback for non-integral values. Add tests for both positive and negative
fractions.

morphqdd:
@yegor256, PR #6714 fixing this was merged but I forgot the `Closes` keyword, so it didn't auto-close. Could you close this one manually? Sorry for the extra step.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2100](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

gemshrine:
Resolved by #6714.

PR #6714, opened 2026-08-12
#6700: reject fractional indexes in tuple.withouti

morphqdd:
`tuple.withouti` compares the supplied `i` directly against each integral element position produced by `reducedi`. A fractional `i` can never equal an integer index, so `withouti` silently returns the tuple unchanged instead of rejecting the request, unlike `tuple.slice` which already validates its start/end as integers.

Added the same `is-integer` guard `slice.eo` uses, rejecting a fractional index with `T "Given index must be an integer"`.

Added `stops-on-a-fractional-index`, matching the `-->` error-test convention already used in `slice.eo`.

Ran `EOwithoutiTest` (4 tests, 0 failures) and `qulice:check -Pqulice`, both green.

### DIFF
--- a/eo-runtime/src/main/eo/tuple/withouti.eo
+++ b/eo-runtime/src/main/eo/tuple/withouti.eo
@@ -8,13 +8,16 @@
 +spdx SPDX-License-Identifier: MIT
 
 [list i] > withouti
-  reducedi > @
-    list
-    *
-    if. > [accum item index] >>
-      i.eq index
-      accum
-      accum.with item
+  i.is-integer.if > @
+    reducedi
+      list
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (23) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3972](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6698, opened 2026-08-12
`tuple.front` misbehaves with positive fractional indexes causing recursion beyond expected tuple


gemshrine:
## What happens

`tuple.front` does not validate its index as an element index. Fractional
negative indexes are silently converted by the recursive traversal, while
fractional positive indexes recurse past the end and expose an internal tuple
failure.

| expression | result |
| --- | --- |
| `tuple.front (* "a" "b") -1.5` | `* "b"` |
| `tuple.front (* "a" "b") -0.5` | `*` |
| `tuple.front (* "a" "b") 0.5` | `ExFailure: Can't get tail from the empty tuple` |
| `tuple.front (* "a" "b") 1.5` | `ExFailure: Can't get tail from the empty tuple` |

This was reproduced through runtime objects on current `master` (`7591007fd7`).

## Why it happens

[`front.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/tuple/front.eo#L11)
only branches on comparisons with `index`; it never requires an integer. Its
base case uses `tup.length.eq idx`, which a fractional positive index can
never satisfy. The recursive call then asks for `tail` on an empty tuple.

## Expected behavior

`index` denotes a tuple position, so non-integral values should be rejected
before traversal with a public, explanatory failure. Both signs should follow
the same rule.

## Suggested fix

Check `index.is-integer` at the entry point and return a dedicated fallback or
bottom object when it is false. Add regression cases for `-0.5` and `0.5`.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2076](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

morphqdd:
@yegor256, PR #6715 fixing this was merged but I forgot the Closes keyword, so it didn't auto-close. Could you close this one manually? Sorry for the extra step.

gemshrine:
Resolved by #6715.

PR #6715, opened 2026-08-12
#6698: reject fractional indexes in tuple.front

morphqdd:
`tuple.front` never checks that `index` is an integer. A negative fractional index gets silently converted by the recursive traversal, and a positive fractional index recurses past the end of the tuple, surfacing an internal `Can't get tail from the empty tuple` failure instead of a clear one about the argument itself.

Added an `is-integer` guard at the entry point, matching the pattern already used in `tuple.slice` and `tuple.withouti`, rejecting a fractional index with `T "Given index must be an integer"`.

Added `stops-on-a-fractional-index`, matching the `-->` error-test convention.

Ran `EOfrontTest` (7 tests, 0 failures) and `qulice:check -Pqulice`, both green.

### DIFF
--- a/eo-runtime/src/main/eo/tuple/front.eo
+++ b/eo-runtime/src/main/eo/tuple/front.eo
@@ -8,26 +8,29 @@
 +spdx SPDX-License-Identifier: MIT
 
 [list index] > front
-  if. > @
-    idx.eq 0
-    *
+  index.is-integer.if > @
     if.
-      0.gt
-        index >> idx!
-      back
-        list
-        neg.
-          number idx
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4108](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6699, opened 2026-08-12
`tuple.back` incorrectly accepts and incorrectly interprets fractional indices


gemshrine:
## What happens

`tuple.back` accepts fractional counts even though its public contract calls
the parameter an element index. The result is not a defined rounding rule: it
depends on comparisons against the integer positions during reduction.

For a two-element tuple:

| expression | result |
| --- | --- |
| `tuple.back (* "a" "b") -1.5` | `*` |
| `tuple.back (* "a" "b") 0.5` | `*` |
| `tuple.back (* "a" "b") 1.5` | `* "b"` |
| `tuple.back (* "a" "b") 2` | `* "a" "b"` |

The results were reproduced through runtime objects on current `master`
(`7591007fd7`).

## Why it happens

[`back.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/tuple/back.eo#L11)
computes `start = list.length - index` and then includes an item when its
integer position is `>= start`. There is no `index.is-integer` validation, so
fractional counts become accidental threshold values instead of tuple counts.

## Expected behavior

`back` should either require an integer count or explicitly document and
implement a consistent rounding rule. The other tuple slicing APIs reject
fractional boundaries, so rejecting them is the consistent behaviour.

## Suggested fix

Validate `index.is-integer` before calculating `start` and return a dedicated
failure for invalid input. Add regression tests for at least `0.5` and `1.5`.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2088](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

morphqdd:
@yegor256, PR #6716 fixing this was merged but I forgot the Closes keyword, so it didn't auto-close. Could you close this one manually? Sorry for the extra step.

gemshrine:
Resolved by #6716.

PR #6716, opened 2026-08-12
#6699: reject fractional counts in tuple.back

morphqdd:
`tuple.back` computes `start = list.length - index` and includes an item when its integer position is `>= start`, with no `index.is-integer` check. A fractional count becomes an accidental threshold value instead of being rejected, so the result depends on which integer positions happen to land on either side of the fraction rather than on a defined rule.

Added an `is-integer` guard at the entry point, matching `tuple.slice`/`tuple.withouti`/`tuple.front`, rejecting a fractional count with `T "Given index must be an integer"`.

Added `stops-on-a-fractional-index`, matching the `-->` error-test convention.

Ran `EObackTest` (4 tests, 0 failures) and `qulice:check -Pqulice`, both green.

### DIFF
--- a/eo-runtime/src/main/eo/tuple/back.eo
+++ b/eo-runtime/src/main/eo/tuple/back.eo
@@ -9,17 +9,20 @@
 +spdx SPDX-License-Identifier: MIT
 
 [list index] > back
-  if. > @
-    0.gt
-      list.length.minus index! >> start!
-    list
-    reducedi
+  index.is-integer.if > @
+    if.
+      0.gt
+        list.length.minus index! >> start!
       list
-      *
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (29) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4120](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6697, opened 2026-08-12
`tuple.shifted` handles positive and negative fractional shifts inconsistently


gemshrine:
## What happens

`tuple.shifted` handles fractional shifts inconsistently. A negative fraction
is silently treated as a zero shift, while a positive fraction reaches
`tuple.slice` and fails with its unrelated validation message.

| expression | result |
| --- | --- |
| `tuple.shifted (* "a" "b") -0.5` | `* "a" "b"` |
| `tuple.shifted (* "a" "b") 0.5` | `ExFailure: Given slice start and end must both be integers` |

The behaviour was reproduced through runtime objects on current `master`
(`7591007fd7`). Integer inputs work: shifts `0` and `2` respectively return
the original and empty tuples.

## Why it happens

[`shifted.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/tuple/shifted.eo#L11)
checks only `shift.lt 0`. Thus `-0.5` selects `0`, but `0.5` is forwarded as
the `start` argument to `tuple.slice`. The latter rejects fractional indices,
so the public `shifted` operation leaks an implementation-specific error.

## Expected behavior

`shift` is an element count, so non-integer values should be rejected uniformly
by `tuple.shifted` with an error that names the invalid shift. Alternatively,
if fractional values are intentionally rounded, both signs must follow the
same documented rule.

## Suggested fix

Validate `shift.is-integer` before the sign branch and route invalid input to a
dedicated fallback/bottom message. Add regressions for both `-0.5` and `0.5`.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2064](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

morphqdd:
@yegor256, PR #6717 fixing this was merged but I forgot the Closes keyword, so it didn't auto-close. Could you close this one manually? Sorry for the extra step.

gemshrine:
Resolved by #6717.

PR #6717, opened 2026-08-12
#6697: reject fractional shifts in tuple.shifted

morphqdd:
`tuple.shifted` only checks `shift.lt 0`, so a negative fraction is silently treated as a zero shift, while a positive fraction is forwarded to `tuple.slice`, which rejects it with its own message naming "slice start and end", not "shift". Both signs need the same rule, and the public `shifted` operation should not leak an unrelated implementation's error text.

Added an `is-integer` guard at the entry point, rejecting a fractional shift with `T "Given shift must be an integer"` before either branch is reached.

Added regression tests for both `-0.5` and `0.5`, matching the `-->` error-test convention.

Ran `EOshiftedTest` (7 tests, 0 failures) and `qulice:check -Pqulice`, both green.

### DIFF
--- a/eo-runtime/src/main/eo/tuple/shifted.eo
+++ b/eo-runtime/src/main/eo/tuple/shifted.eo
@@ -8,13 +8,16 @@
 +spdx SPDX-License-Identifier: MIT
 
 [list shift] > shifted
-  slice > @
-    list
-    if.
-      shift.lt 0
-      0
-      shift
-    list.length
+  shift.is-integer.if > @
+    slice
+      list
+      if.
+        shift.lt 0
+        0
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (29) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4132](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6721, opened 2026-08-12
Cache y and k in expon so ln is evaluated once per call

### LINKED ISSUE #6718: `k` and `y` variables in `expon` function of `power.eo` are not cached
The puzzle `6651-9dd791e0` from #6651 has to be resolved: 

https://github.com/objectionary/eo/blob/4c315a5185a61fbb77d76be274e190b625d74f24/eo-runtime/src/main/eo/number/power.eo#L7-L13

Thayorns:
`expon` computes `e` raised to `y`, where `y` is the void bound to `expo.times (ln base)`. Every read of `y` re-dataized that expression, and every read of `k` re-dataized `y` in turn, so a single fractional-exponent call paid for roughly forty evaluations of `ln`.

`y` is now bound once to the const local `arg`, `k` becomes `k!`, and each read is re-wrapped with `number`, the same idiom `ipow` already uses for `half`. One call now evaluates `ln` once.

Four tests cover the branches of `expon` that the caching touches: a negative `k`, a zero `k` with a zero remainder, a zero `k` with a negative remainder, and a positive `k`.

Closes #6718

### DIFF
--- a/eo-runtime/src/main/eo/number/power.eo
+++ b/eo-runtime/src/main/eo/number/power.eo
@@ -62,19 +62,27 @@
                   [y] >> expon
                     times. > @
                       if.
-                        (number k).lt 0
+                        lt. (number k) 0
                         1.div
-                          ipow e (abs (number k))
-                        ipow e (abs (number k))
+                          ipow
+                            e
+                            abs (number k)
+                        ipow
+                          e
+                          abs (number k)
                       fraction 1
+                    y > arg!
                     if. > [j] > fraction
                       j.gt 20
                       1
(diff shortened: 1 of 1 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4182](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6722, opened 2026-08-12
Stop the directory walk from descending into a Windows symbolic link

### LINKED ISSUE #6720: directory.eo fails to prevent descending into symbolic links
The puzzle `6483-14da2361` from #6483 has to be resolved: 

https://github.com/objectionary/eo/blob/62b52a8d3d1f97ffbce639dfcf236b5e9f696ba1/eo-runtime/src/main/eo/directory.eo#L13-L16

Thayorns:
On POSIX the walk already leaves a symbolic link alone, since `file.is-symlink` asks `lstat` and reads the link's own mode bits. On Windows it had nothing to ask: `_stat64` walks a reparse point through to the file behind it, and msvcrt has no `_lstat`, so every link answered "not a link" and a link pointing at its own ancestor sent the walk around forever.

The answer comes from the Windows API instead. `GetFileAttributesW` reports the attribute bits of the path itself, and the reparse bit among them is what tells a link from its target. The wide variant is bound so a path outside the active code page survives the trip, and the mask lives in `win32.eo` next to the other Windows constants.

`file.is-symlink` now splits by platform: the attribute bits on Windows, `lstat` everywhere else, with an invalid attribute answer folding into the same `cant-check` the missing-file case already used.

The two new tests build a directory and a link to it and ask `file` about both. They live in `file.eo` beside the other `file` tests rather than in Java, since `is-symlink` has no Java class behind it for a Java test to point at. Both stand aside on Windows, because the `symlink` syscall they lean on is POSIX only — the same shape the walk test in `directory.eo` already has. Each one passes the `cant-check` value that turns an unanswerable check into a failure, so neither can pass by accident.

Closes #6720


eo-runtime/src/test/java/org/eolang/EOfileEOissymlinkTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * Test case for the {@code is-symlink} attribute of {@code file} in
+ * {@code file.eo}, which no Java atom stands behind.
+ * @since 0.75.0
+ */
+@SuppressWarnings("JTCOP.RuleAllTestsHaveProductionClass")
```
yegor256: @Thayorns listen to JTCOP. don't create this Java test, since we don't have such Java file. stay in EO tests.

Thayorns:
@yegor256 fixed here

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2791](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:
@maxonfjvipon take a look at this new class `Kernel32`. good?

maxonfjvipon:
@yegor256 it seems we can't avoid it here:(

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (154 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4194](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6677, opened 2026-08-12
`SonarCloud Code Analysis` master branch failure due to poor `new_reliability_rating`


yangjj-iso:
`SonarCloud Code Analysis` is red on master's head (`7125a49`) and was red on the master analyses before it — `2e9427b`, `eb74a5d`, `9760822`, `d421d8d`, `e4a5dea`, `38c1451`. It appears in every commit's checks, so by the rule in [We Don't Merge into a Broken Master Branch](https://www.yegor256.com/2025/04/19/dont-merge-into-broken-master.html) it is one of the jobs that has to be green before anything else goes in. I could not find it reported, so here it is.

## What the gate says

One condition fails and the other four pass (`/api/qualitygates/project_status?projectKey=objectionary_eo&branch=master`):

| condition | actual | required |
| --- | --- | --- |
| `new_reliability_rating` | 3 (C) | ≥ A |
| `new_security_rating` | 1 | ≥ A |
| `new_maintainability_rating` | 1 | ≥ A |
| `new_duplicated_lines_density` | 0.2% | < 3% |
| `new_security_hotspots_reviewed` | 100% | 100% |

So it is reliability on new code alone, at C where A is required.

## The bugs behind it

Master carries 11 open `BUG` issues; the project's overall `reliability_rating` is 4 (D). The ones rated high enough to hold the new-code rating at C:

- `eo-printer/src/main/java/org/eolang/printer/Pretty.java` — access that may throw a `NullPointerException`
- `eo-parser/src/main/java/org/eolang/parser/EoSyntax.java:100` — same
- `eo-parser/src/main/java/org/eolang/parser/OnDetailed.java:72` — same
- `eo-maven-plugin/src/main/java/org/eolang/maven/ChText.java:78` — same
- `eo-maven-plugin/src/main/java/org/eolang/maven/Dep.java` — same
- `eo-maven-plugin/src/main/java/org/eolang/maven/Saved.java` — same
- `eo-runtime/src/main/java/org/eolang/BytesRaw.java:217` and `:241` — `int` promotion without `& 0xff`
- `eo-maven-plugin/src/main/java/org/eolang/maven/Dependencies.java:87` and `:110` — `Random` not saved and re-used
- `eo-maven-plugin/src/test/java/org/eolang/maven/MjLintTest.java` — lambda with multiple invocations throwing the same checked exception

## What I could not pin down

[...]

PR #6723, opened 2026-08-12
#6677: guard EoSyntax.parsed() against a null transform

morphqdd:
SonarCloud's `new_reliability_rating` gate is red on master (rated C where A is required), because of a BLOCKER-severity NPE-risk finding at `EoSyntax.java:100`: `this.transform.apply(...)` in `parsed()`, flagged since `EoSyntax(Input, UnaryOperator<XML>)` is a public constructor that accepts `transform` without a null check. It's the newest of the several open SonarCloud BUG findings on master (created 2026-07-13, the rest are from Feb-Aug 2025), so it's the one most likely inside the new-code window that's failing the gate.

Guarded the call site instead of the constructor: qulice's `ConstructorsCodeFreeCheck` forbids method calls in a constructor body, so the null check couldn't live there. Folded the guard and the use into one expression (`Objects.requireNonNull(this.transform, "transform").apply(...)`), so there's no separate statement whose order with the guard could drift.

Added `rejectsANullTransform`, asserting `.parsed()` throws `NullPointerException` for a null transform instead of an unguarded NPE.

Ran the full `EoSyntaxTest` suite (1086 tests, 0 failures) and `qulice:check -Pqulice`, both green. I don't have access to the project's SonarCloud settings to confirm this specific finding is what's inside the new-code window, so this may need a second look at the gate after merge; the other open findings (BytesRaw, ChText, OnDetailed, Dependencies) are all from well before the likely window and probably aren't the cause.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/EoSyntax.java
+++ b/eo-parser/src/main/java/org/eolang/parser/EoSyntax.java
@@ -10,6 +10,7 @@ import com.yegor256.xsline.Shift;
 import com.yegor256.xsline.Train;
 import com.yegor256.xsline.Xsline;
 import java.io.IOException;
+import java.util.Objects;
 import java.util.function.UnaryOperator;
 import org.cactoos.Input;
 import org.cactoos.io.InputOf;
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4144](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6724, opened 2026-08-12
Halt the walk at a symbolic link, with a test

### LINKED ISSUE #6719: `walk` method in directory.eo doesn't halt as expected
The puzzle `6483-d169bdfb` from #6483 has to be resolved: 

https://github.com/objectionary/eo/blob/62b52a8d3d1f97ffbce639dfcf236b5e9f696ba1/eo-runtime/src/main/eo/directory.eo#L9-L12

The puzzle was created by @yegor256 on 12-Aug-26. 

Thayorns:
`walk` already refuses to descend into a symbolic link: it asks `lstat` through `file.is-symlink` and lists such an entry instead of following it. Nothing exercised that in EO, though. The only coverage was `StatSyscallTest`, which checks `lstat` alone and says nothing about the walk that leans on it.

The test could not be written because EO had no way to make a symbolic link. So `posix` gets a `symlink` syscall, alongside `rename` and the rest, and the new test in `directory.eo` builds a link inside a directory that points back at that same directory. Before the guard existed, walking `**` over such a directory recursed until the stack gave up; now it yields the two entries and stops.

The test is skipped on Windows, where `_stat64` knows nothing about links. That gap is the remaining `@todo` in the same docblock.

Closes #6719

### DIFF
--- a/eo-runtime/src/main/eo/directory.eo
+++ b/eo-runtime/src/main/eo/directory.eo
@@ -6,10 +6,6 @@
 #  like [a-z] or [!a] reaches the regex engine neither translated nor escaped,
 #  so a negation matches nothing it should and a lone bracket refuses to
 #  compile, while a star inside a class keeps its glob meaning by mistake.
-# @todo #6483:30min Test that `walk` below halts at a symbolic link. EO cannot
-#  make one, because `posix` has no `symlink` syscall yet, so the halting is
-#  covered only through `lstat` in `StatSyscallTest`, while the walk that leans
-#  on it has no test of its own.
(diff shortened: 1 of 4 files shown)


Thayorns:
@yegor256 CI green here

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4210](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6711, opened 2026-08-12
`Pulling` and `Unplacing` incorrectly log item count when sources/binaries get excluded


morphqdd:
## What happens

Two build-log messages report the wrong count of items processed.

`Pulling.java:150-155` logs `sources.size()` as "N program(s) were pulled", but `names` (a separate, smaller collection holding only the non-directory sources actually pulled) is what the log message is describing. Whenever some sources are directories (skipped, not pulled), the logged count overstates how many programs were actually pulled.

`Unplacing.java:84-88` compares `total == available.size()` to decide whether to log "All N binaries deleted", but logs `binaries.size()` for N, not `available.size()`. Whenever `keepBinaries` excludes some binaries from `available`, the logged N is the pre-exclusion count, not the count actually deleted.

Neither is a functional bug (both operations still process the correct items), but the console/log output misreports how many were processed, which is misleading when investigating a build.

## What should happen

`Pulling`'s log line should report `names.size()` (or whatever collection was actually pulled), not `sources.size()`. `Unplacing`'s log line should report `available.size()` (the count that was actually compared and deleted), not `binaries.size()`.

0crat:
@morphqdd Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+3920](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6725, opened 2026-08-12
#6711: log the actual count of pulled programs and deleted binaries

morphqdd:
Two log lines reported the wrong count.

`Pulling` logged `sources.size()` as "N program(s) were pulled", but `sources` includes directories that get skipped rather than pulled; `names` is the collection that actually holds what was pulled and is what's printed right after. Switched the count to `names.size()`.

`Unplacing`'s "All N binari(es) deleted" branch is only reached when `total == available.size()`, but logged `binaries.size()` (the pre-`keepBinaries`-exclusion count) instead of `available.size()` (what was actually compared and deleted). Switched that one line; the other two branches already log `binaries.size()` intentionally, as the "out of N found" denominator.

Neither is a functional bug, both operations still process the correct items. Existing `UnplacingTest`/`CompilingTest` pass unchanged; didn't add a dedicated log-capture test since neither file has that infrastructure and this is a one-line count fix, not a behavior change.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Pulling.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Pulling.java
@@ -150,7 +150,7 @@ final class Pulling implements Step {
             Logger.info(
                 this,
                 "%d program(s) were pulled: %s",
-                sources.size(),
+                names.size(),
                 names
             );
         }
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4028](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #4139, opened 2025-05-04
`validate-objects-count` Lint Misflags Absence of `<o/>` Elements


yegor256:
This is my XMIR:

```
<object/>
```

It doesn't have any `<o/>` elements (due to a broken syntax). Then, I'm getting this:

```
<error check="validate-objects-count" line="0" severity="critical">Only one object per source file is allowed and required</error>
```

This message is misleading. The `validate-objects-count` lint should ignore situations when no `<o/>` elements are present. Maybe other lints should complain about such a case (we can create them), but not this one.

h1alexbel:
@yegor256 I'm in

maxonfjvipon:
@yegor256 why is this message misleading? It says `allowed and required` which is "no more, no less"

yegor256:
@maxonfjvipon maybe the "only" part is misleading. It sounds like I have more than one. Maybe this is better: "exactly one object is allowed and required". However, I still believe that this particular lint should ignore this situation and we should have _another_ lint for this case.

h1alexbel:
@yegor256 should be fixed now

yegor256:
@h1alexbel thanks!

PR #6726, opened 2026-08-12
Reject a source file that declares no object

### LINKED ISSUE #6713: Printer wrongly adds extra newlines for programs without object body causing non-parsable output
## What happens

The printer's `to-eo-tree.xsl` (`eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl`) emits an unconditional trailing `$eol` separator at the end of both the `comments` template (line 140) and the `metas` template (line 171), even when no object body follows.

For a program with metas but no object, for example a file containing only:
```
+package foo
```
(this parses with zero errors: `XMIR.xsd:438` allows `minOccurs=0` for `/object/o`, and `validate-objects-count.xsl` only errors when `count(o) > 1`), `Xmir.toEO()` prints:
```
+package foo


```
(one trailing newline from the source line, plus two more from the unconditional separators). Reparsing that output fails:
```
[3:0] error: 'consecutive blank lines forbidden ... (R-6.5.3)'
[0:0] error: 'more than one trailing blank line'
```

The same thing happens for a comment-only file. This breaks the printer's own round-trip contract, exercised by `XmirTest.printsToParseableEo` ("printed EO must reparse to itself"), for this specific shape of input.

Practical reach is limited: a normal `.eo` file always declares at least one object, so this only triggers for a metas-or-comments-only file. Verified with a full parse→print→parse round trip; ruled out as a broader problem by round-tripping all 170 `.eo` files under `eo-runtime/src/main/eo`, which are clean.

## What should happen

`comments` and `metas` should only emit their trailing separator when an object body actually follows, so printing a metas/comments-only program produces output that reparses to itself, consistent with `XmirTest.printsToParseableEo`.

Thayorns:
A file that holds nothing but metas or a top comment block parses today without a single error: `XMIR.xsd` makes `/object/o` optional and `validate-objects-count.xsl` only ever complains about a *second* object, never about a missing one. That silence is deliberate — #4139 removed the zero case because the lint was misflagging an empty `<object/>` left behind by broken syntax — and it closed with the note that some other lint could complain about the absence instead. This is that lint.

`validate-object-presence.xsl` runs right after `validate-objects-count.xsl` in the canonical parse chain and raises a critical error when the file declares no object at all. It stays quiet when the file already carries an error, so a file whose object was lost to broken syntax still reports that syntax and nothing more — exactly the case #4139 was about.

Two eo-packs cover the two halves: a `+package foo` file is refused, and an XMIR that already carries a parser error is left untouched. Two tests in `EoSyntaxTest` drive the full pipeline through `EoSyntax.parsed()` for the metas-only and the comments-only shape, so the refusal is asserted where a user actually meets it.

The printer fix this branch carried before is gone. It taught `to-eo-tree.xsl` and `Pretty` to lay such a program out without trailing blank lines, which was the answer #6713 asked for as filed; with the parser refusing the input, that path is unreachable and the two print-packs that drove it could not exist. Nothing else regresses: no `.eo` file and no test pack in the repository declares zero objects.

Closes #6713


Thayorns:
@yegor256 take a look please

yegor256:
@Thayorns I believe it's better to reject .eo files without objects. The parser should just refuse to parse them.

Thayorns:
@yegor256 done, the branch now does that instead. A new parse-phase check, `validate-object-presence.xsl`, raises a critical error when the file declares no object, so `+package foo` alone no longer parses. It keeps quiet when the file already carries an error, so a file whose object was lost to broken syntax reports only that syntax — the case #4139 asked `validate-objects-count` to stop misflagging, which is also why the absence became a separate lint rather than a relaxed condition on the old one. The printer change is gone: with the input refused, that path is unreachable.

Thayorns:
@yegor256 fixed, have a look please

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (176 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1360](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6727, opened 2026-08-12
Sanitize object and attribute names before emitting Java

### LINKED ISSUE #6712: `to-java.xsl` emits invalid Java due to unsanitized object/attribute names
## What happens

`to-java.xsl` (`eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl`) turns an EO object or attribute name directly into a Java identifier or string literal, sanitizing only `_ - @ α $` (see `eo:clean` at line 35, and the `$quoted` branch of `eo:attr-name` at line 97, which wraps the raw name in `"` with no escaping).

The parser accepts a much wider set of characters in a `NAME` token (per `PARSER_SPEC.md` R-2.3: anything except space, break, tab, and `, . | ' : ; ! ? ] [ } { ) (`). Names built from the remaining punctuation (` ~ " # $ % & * + - < = > @ \ ^ ` `) parse cleanly but are not sanitized by the transpiler, producing uncompilable or silently wrong Java.

Three reproductions, verified by running the same XSL shift chain the transpiler uses and compiling the result with `javac`:

1. A quote in an attribute name:
```
[] > app
  1 > a"b
```
Parses with zero errors, then transpiles to `this.add("a"b", new AtOnce(...`. `javac` fails: `')' or ',' expected`.

2. A backslash in a name (worse: this one compiles):
```
[] > app
  1 > a\b
```
Transpiles to `this.add("a\b", ...)`. `\b` is a valid Java escape for backspace, so the compiled attribute name silently differs from the EO source name (`a` + backspace + `b`, not `a\b`). A slightly different name, `a\qb`, hits an illegal-escape compile error instead.

3. A star in a top-level object name:
```
[] > a*b
```
Produces `public final class EOa*b extends org.eolang.PhDefault {`, a Java syntax error (`'{' expected`).

No existing lint blocks these names; `lints`' `compound-name` rule only flags `-`/`_`/uppercase.

## What should happen

`to-java.xsl` should either reject these names at an earlier stage (a lint, or a parser-level restriction on which characters a Java-emittable name may contain) or properly sanitize/escape them when emitting Java identifiers and string literals, so a name that parses cleanly always produces compilable Java with the correct semantic name, never a silent corruption like case 2.

Thayorns:
The parser accepts a much wider `NAME` alphabet than `to-java.xsl` knows how to emit. Only `_ - @ α $` were sanitized, so the remaining punctuation went straight into Java identifiers and string literals. `[] > a*b` produced `public final class EOa*b`, which does not compile. `1 > a"b` produced `this.add("a"b", ...)`, which does not compile either. Worst of the three, `1 > a\b` produced `this.add("a\b", ...)`, which compiles fine but binds the attribute under `a` + backspace instead of `a\b` — a silent corruption.

There are two distinct sinks and this adds one function for each. `eo:identifier` hex-escapes every character Java forbids in an identifier as `$uXXXX`, and now wraps `eo:clean`, `eo:clean-package`, `eo:loc-to-class` and the generated test method name. The escape stays injective: `$` in a name already becomes `$EO`, and `EO` is not a pair of hex digits, so no two EO names can collapse onto one Java name. `eo:literal` escapes the backslash and the quote, and now wraps every place a name reaches a Java string literal — the attribute name in `add` and `take`, `AtVoid`, the `@XmirObject` annotation on both the class and its test class, and the locator, object name and original name passed to `PhSafe`, `PhCoverage` and `PhDefault`.

I checked the fix by running the same XSL chain the transpiler runs and compiling the result: `a*b` now becomes `EOa$u002Ab`, the quote case becomes `add("a\"b", ...)`, and the backslash case becomes `add("a\\b", ...)`, which is the attribute name the source actually asked for.

The generated Javadoc still carries the raw XMIR dump with unescaped names in it. That is harmless today because `/` is not a legal `NAME` character, so a name can never close the comment, but it is worth a look if the parser's alphabet ever widens.

Closes #6712

### DIFF
--- a/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
+++ b/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
@@ -41,7 +41,7 @@
   <xsl:function name="eo:identifier" as="xs:string">
     <xsl:param name="n" as="xs:string"/>
     <xsl:variable name="escaped">
-      <xsl:analyze-string select="$n" regex="[^\p{{L}}\p{{Nd}}_$]">
+      <xsl:analyze-string select="$n" regex="[^\p{{L}}\d_$]">
         <xsl:matching-substring>
           <xsl:value-of select="eo:escape-char(.)"/>
         </xsl:matching-substring>
--- a/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
+++ b/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
@@ -31,10 +31,36 @@
       <xsl:value-of select="$TAB"/>
     </xsl:for-each>
   </xsl:function>
+  <!-- Unicode escape of a character Java forbids in an identifier -->
+  <xsl:function name="eo:escape-char" as="xs:string">
+    <xsl:param name="c" as="xs:string"/>
+    <xsl:variable name="code" select="string-to-codepoints($c)[1]"/>
+    <xsl:value-of select="concat('$u', string-join(for $w in (4096, 256, 16, 1) return substring('0123456789ABCDEF', ($code idiv $w) mod 16 + 1, 1), ''))"/>
+  </xsl:function>
+  <!-- Turn a name into a Java identifier, escaping every character Java forbids there -->
+  <xsl:function name="eo:identifier" as="xs:string">
+    <xsl:param name="n" as="xs:string"/>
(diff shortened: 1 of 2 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1340](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6728, opened 2026-08-12
#6709: write the object attribute into wpa.xmir so cached WPA defects keep their program name

### LINKED ISSUE #6709: `wpa.xmir` cache omits program name in whole-program-analysis lint defects
## What happens

`eo:lint` with `lintAsPackage` at its default of `true` runs whole-program-analysis (WPA) and caches the result in `wpa.xmir`. When the cache is on (also the default), every WPA defect loses the name of the program it came from.

`Linting.embedded` (`eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java:601-612`) writes `check`, `severity`, `line` and the defect text into `wpa.xmir`, but never writes an `object` attribute. `Linting.read` (`Linting.java:640-651`) reads the defect back with `node.attribute("object").text().orElse("")`, so `defect.object()` is always an empty string. `lintAll` (`Linting.java:432-437`) always goes through `read(target)` when caching is on, even right after a cache miss, so this is not a stale-cache edge case, it happens on every run.

Verified with a temporary FakeMaven test using the existing `MjLintTest.problematic()` program:

With caching on (default), the reported line is:
```
:2 Alias "nowhere Φ.a.b.nowhere" points to "a.b.nowhere", ... (incorrect-alias/W)
```

With `cacheEnabled=false`, the same run reports:
```
foo.x.main:2 Alias "nowhere Φ.a.b.nowhere" points to ... (incorrect-alias/W)
```

The program name before the colon is missing whenever the cache is on. Also confirmed in an on-disk `wpa.xmir` artifact: the `<error>` element has `check`, `line`, `severity` but no `object`.

There is a second effect: on a cache hit, the `[LINT]` console lines are not printed at all (they are only emitted inside `wpa()`, which a cache hit skips), so the anonymous `:2 ...` line is the only report a developer sees for that build.

## What should happen

`Linting.embedded` should also write the defect's `object` (program name) into `wpa.xmir`, and `Linting.read` should read it back, so a WPA defect reported from a cached run names the same program a non-cached run would.

morphqdd:
`Linting.embedded` wrote `check`, `severity`, `line` and the defect text into `wpa.xmir`, but never wrote an `object` attribute. `Linting.read` reads a WPA defect back with `node.attribute("object").text().orElse("")`, so `defect.object()` is always empty for any run that hits the cache, which is every run once `lintAsPackage` (default `true`) and caching (also default `true`) are both on.

Added `.attr("object", defect.object())` to the WPA `embedded` overload, so the program name round-trips through the cache the same way it already does for the per-file lint path.

Added `namesTheProgramInAWholeProgramAnalysisDefectFromCache`: runs the same problematic program twice, verifying the second (cached) run's exception still names `foo.x.main`, not just the bare rule text. Verified this catches the bug: reverted the fix locally and the new test failed with the program name missing from the message, confirming it before restoring the fix.

Ran the full `MjLintTest` suite (27 tests, 0 failures) and `qulice:check -Pqulice`, both green.

Closes #6709

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
@@ -64,6 +64,13 @@ final class Linting implements Step {
     static final String CACHE = "linted";
 
     /**
+     * The XMIR {@code object} element/attribute name, used both for
+     * navigating a source XMIR and for round-tripping a WPA defect's
+     * program name through {@code wpa.xmir}.
+     */
+    private static final String OBJECT = "object";
+
+    /**
      * Scoped foreign tojos.
      */
     private final TjsForeign tojos;
@@ -537,7 +544,7 @@ final class Linting implements Step {
      */
     private static Collection<Defect> existing(final Xnav xnav) {
         return xnav
-            .element("object")
+            .element(Linting.OBJECT)
             .elements(Filter.withName("errors"))
             .findFirst().map(
                 errors -> errors
@@ -609,7 +616,7 @@ final class Linting implements Step {
         dirs.add("error")
             .attr("check", defect.rule())
             .attr("severity", defect.severity().mnemo())
-            .attr("object", defect.object())
+            .attr(Linting.OBJECT, defect.object())
             .set(defect.text());
         if (defect.line() > 0) {
             dirs.attr("line", defect.line());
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (34) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4040](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6732, opened 2026-08-12
#6706: accept a reversed-dispatch head on a compact-tuple line

### LINKED ISSUE #6706: `eo-parser` rejects valid reversed compact-tuple expressions with `*N`
## What happens

The current parser rejects the reversed compact-tuple form that the
grammar explicitly permits. This program is valid under R-3.8.5 and
R-3.9.1:

```eo
joined. *1 > sixth
  "receiver"
  "x"
  "y"
```

`joined.` is a vertical reversed dispatch and `*1` says that its first
child is the receiver while the remaining children form one tuple
argument. The parser instead emits no object for the program and
reports:

```text
[1:9] error: trailing garbage after expression
joined. *1 > sixth
         ^
```

The parser has a dedicated `LnCompactTuple`, but the classifier sends a
`name. *N` line to `LnReversed` first. `LnReversed` reads `*` as a
horizontal receiver and leaves the `1` behind, so its ordinary argument
parser reports trailing garbage. `LnCompactTuple` itself documents that
reversed heads are not supported.

## Why this is incorrect

R-3.8.5 defines `*N` as legal after a reversed dispatch in its vertical
form, and R-3.9.1 lists `bare-reversed` as one of the three permitted
compact-tuple heads. R-3.9.4 only restricts `N` to values at least one;
the example uses `N = 1` and satisfies that restriction.

This is also a regression of the supported trailing-asterisk notation
from #4496.

## Expected behavior

The example should parse into one `.joined` application named `sixth`:

- the first child (`"receiver"`) is its direct receiver argument;
- the other two children are collected into a final `Φ.tuple` child
  carrying `star=""`.

Conversely, `joined. *` and `joined. *0` must be rejected with the
specified `reversed-dispatch compact tuple requires N ≥ 1` diagnostic,
instead of being treated as a different form.

morphqdd:
`Eo.applicative` routes any line with a trailing dot straight to `LnReversed` before checking for a compact-tuple marker, so `joined. *1 > sixth` fell into `LnReversed`'s ordinary argument parser, which read `*` as a horizontal receiver and left the `1` behind as trailing garbage.

Kept the routing to `LnReversed` unchanged and added compact-tuple handling inside it instead: right after the dispatch dot is consumed, a ` *N` marker is now read before falling back to `readArgs()`. When present, the line stays in the bare (vertical) reversed form (`Kind.BARE_REVERSED`, open for children), and the pushed level is marked with the same `compact()`/`markStar()` calls `LnCompactTuple` and `LnOnlyPhi` already use, so the existing close-time wrapping logic (`Eo.closeCompactTuple`/`beforeChild`) picks it up unchanged. The first `N` deeper-indent children stay direct (the first becomes the receiver via the same close-time check `LnReversed` already does), the rest are wrapped into a synthesised `Φ.tuple`.

`N` must be at least 1 for the reversed form (a receiver-less `joined. *0` has no legal reading), so `LnCompactTuple.readCount`'s default-0 result is rejected with a dedicated message when the marker is present but yields 0.

Added `reversed-compact-tuple.yaml` (positive case from the issue, verifying the receiver and the synthesised tuple both land where expected) and `reversed-compact-tuple-requires-n.yaml` (the `*0` rejection).

Ran the full `EoSyntaxTest` suite (1088 tests, 0 failures) and `qulice:check -Pqulice`, both green.

Closes #6706

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnCompactTuple.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnCompactTuple.java
@@ -98,24 +98,13 @@ final class LnCompactTuple implements Line {
     }
 
     /**
-     * Push or replace the stack level per Step B/C/D of §5.2.
-     * @param stack The stack
-     * @param suffix The parsed suffix
-     * @return The pushed/replaced level
-     */
-    private Level transition(final Stack stack, final Suffix suffix) {
-        return new Transition(stack, this.span)
-            .apply(Kind.COMPACT_TUPLE, Openness.OPEN, suffix.named());
-    }
-
-    /**
      * Read the {@code N} count after {@code *}. Defaults to 0 when
      * absent.
      * @param tokens Token reader (cursor just past the {@code *})
      * @param span Source span, for a ParseError on overflow
      * @return The N value
      */
-    private static int readCount(final Tokens tokens, final Span span) {
+    static int readCount(final Tokens tokens, final Span span) {
         long count = 0;
         boolean any = false;
         final int start = tokens.cursor();
@@ -138,5 +127,16 @@ final class LnCompactTuple implements Line {
             count = 0;
         }
         return (int) count;
+    }
+
+    /**
+     * Push or replace the stack level per Step B/C/D of §5.2.
+     * @param stack The stack
+     * @param suffix The parsed suffix
+     * @return The pushed/replaced level
+     */
+    private Level transition(final Stack stack, final Suffix suffix) {
+        return new Transition(stack, this.span)
+            .apply(Kind.COMPACT_TUPLE, Openness.OPEN, suffix.named());
     }
 }
--- a/eo-parser/src/main/java/org/eolang/parser/LnReversed.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnReversed.java
@@ -68,7 +68,13 @@ final class LnReversed implements Line {
             );
         }
         final boolean fragile = tokens.consumeDispatch();
-        final List<Value> args = tokens.readArgs();
+        final int star = this.compactStar(tokens);
+        final List<Value> args;
+        if (star >= 0) {
+            args = new java.util.ArrayList<>(0);
+        } else {
+            args = tokens.readArgs();
+        }
         if (!args.isEmpty()) {
             Bindings.checkReceiver(args.get(0), this.span);
             Bindings.checkAllOrNothing(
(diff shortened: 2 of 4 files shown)


yegor256:
@morphqdd #6706 is not yet accepted. I don't understand what is the meaning of this:

```
joined. *1
  "foo"
  4
  5
```

What does it mean?

morphqdd:
@yegor256 mechanically, with `*1`: the first child (`"foo"`) is the receiver of `.joined`, and the remaining children (`4`, `5`) get collected into a single synthesized tuple argument. So it parses as calling `.joined` on `"foo"`, passing one argument that is the tuple `(4, 5)`.

This came from combining R-3.8.5 (`*N` allowed after a vertical reversed dispatch) and R-3.9.1 (bare-reversed is a valid compact-tuple head) in PARSER_SPEC.md, not from an existing real-world program. Is this actually a feature you want supported, or should the spec not allow `*N` on a reversed head at all?

yegor256:
@morphqdd then why it's not:

```
joined. *
  "foo"
  4
  5
```

This would be logically similar to `"foo".joined (* 4 5)`.

morphqdd:
@yegor256 that reading is more attractive, but the spec as written disagrees. R-3.8.5 and R-3.9.4 in `PARSER_SPEC.md` say `N` for a reversed head counts the receiver itself as the first direct slot, so `N` can't be `0`:

> R-3.8.5: For reversed dispatch with `*N`, `N >= 1` is required (the receiver occupies the first direct slot).
> R-3.9.4: Reversed-dispatch heads need `N >= 1`. If the head is a reversed dispatch, the receiver must occupy at least one direct slot... `name. *` (no N, defaults to 0) and `name. *0` are rejected at the classifier.

So under the current text, `joined. *` has no receiver slot at all and is invalid by definition, not just by this PR's choice. If you want `joined. *` to mean "first child is the receiver, everything else tupled", that changes R-3.8.5/R-3.9.4, not just the parser.

yegor256:
@morphqdd I suggest we update the spec. @maxonfjvipon thoughts?

maxonfjvipon:
@morphqdd @yegor256 this syntax is confusing, I'm not sure we need it at all. Star `*` after name is used as sugar when we want to say than `name` accepts `tuple` and next lines are tuple elements. If we have `*` after method: `joined. *` it conflicts with method first argument which method base.
I would not do it.

If method base is simple, it can be done just like
```
foo.joined *
  arg1
  arg2
  ...
```

if base is complex, this start after method wont give us anything:
```
joined. *
  x
    y
      z
    a
    b
    c
  arg1
  arg2
```

even more confusing

yegor256:
@morphqdd I agree with @maxonfjvipon - let's drop this feature

morphqdd:
Closing per the discussion above: @yegor256 agreed with @maxonfjvipon to drop this feature rather than implement it.

PR #6736, opened 2026-08-13
#6704: validate UTF-8 sequences in string.length

### LINKED ISSUE #6704: `string.length` inaccurately counts malformed UTF-8 byte sequences as characters
## What happens

`string` is documented as UTF-8 text, and `string.length` says that it
walks the UTF-8 sequence character by character. However,
`eo-runtime/src/main/eo/string/length.eo:20-57` determines a character
width solely from the leading byte and only checks that enough following
bytes exist. It never verifies that those bytes have the UTF-8
continuation form `10xxxxxx`, nor rejects overlong encodings.

Consequently, malformed byte data is silently counted as text. These
cases were evaluated through the current `string` and `length` runtime
atoms:

| bytes passed to `string` | invalidity | actual `length` |
| --- | --- | --- |
| `C2-41` | `41` is not a continuation byte | `1` |
| `E1-80-41` | final byte is not a continuation byte | `1` |
| `F0-90-80-41` | final byte is not a continuation byte | `1` |
| `C0-80` | overlong encoding of U+0000 | `1` |

For example, `(string C2-41).length` completes with `1`, although
`C2-41` cannot be decoded as UTF-8: a byte after `C2` must be in the
range `80`–`BF`, while `41` is ASCII `A`.

The current implementation accepts it because `C2` matches the
two-byte-leading mask (`byte.and E0- == C0-`) and `inclen` checks only
`index + 2 <= size`.

## Why this is incorrect

The public `string` contract says that a string is stored as a UTF-8
byte sequence, not an arbitrary byte array. Counting invalid data as a
character makes the character-oriented operations give a value for text
that does not exist. It also disagrees with the adjacent incomplete
sequence behaviour: `string F0-9F-98` deliberately terminates in the
existing `stops-on-an-incomplete-four-byte-character` test.

## Expected behavior

`string.length` should terminate when any multi-byte sequence is not
valid UTF-8, including a missing/invalid continuation byte, an overlong
encoding, a surrogate encoding, or a code point above U+10FFFF. A shared
UTF-8 validator should be used by `string.slice` as well, since it
currently recognises character widths with the same leading-byte-only
logic.

gemshrine:
Fixes #6704.

`string.length` used to determine a character width from the leading byte
alone and only checked that enough trailing bytes existed, so malformed data
was silently counted as text: `(string C2-41).length` returned `1`, and so
did `E1-80-41`, `F0-90-80-41` and the overlong `C0-80`.

Now every multi-byte sequence is validated against the UTF-8 rules of the
Unicode standard (Table 3-7): each following byte must be a `10xxxxxx`
continuation byte, and the second byte is range-checked per lead so that

- overlong encodings are rejected: `C0`/`C1` leads, `E0` with an `80`-`9F`
  second byte, `F0` with an `80`-`8F` second byte;
- UTF-16 surrogates are rejected: `ED` with an `A0`-`BF` second byte;
- code points above U+10FFFF are rejected: `F4` with a `90`-`BF` second byte,
  and `F5`-`FF` leads.

A malformed sequence now terminates the computation, exactly like the existing
truncated-character behaviour.

Verified against the runtime (each case dataized through the `string`/`length`
atoms): all the malformed classes above throw, while valid inputs still count,
including the boundary code points U+D7FF (last scalar before the surrogate
range) and U+10FFFF (the maximum). The existing character-counting examples
for 1/2/3/4-byte strings stay green.

The issue also notes that `string.slice` recognises widths with the same
leading-byte-only logic; that can reuse the same validation in a follow-up so
this change stays focused on `length`.

### DIFF
--- a/eo-runtime/src/main/eo/string/length.eo
+++ b/eo-runtime/src/main/eo/string/length.eo
@@ -1,7 +1,14 @@
 # Counts the characters in the `text`, and not its bytes. The UTF-8 sequence
 # is walked character by character, so a multi-byte character counts as one.
-# A sequence that does not end with a complete character terminates
-# the computation.
+# Every multi-byte sequence is validated against the UTF-8 rules of the
+# Unicode standard (Table 3-7): the leading byte fixes the character width and
+# each following byte must be a `10xxxxxx` continuation byte, while the second
+# byte is additionally range-checked so that overlong encodings (`C0`/`C1`
+# leads, `E0` with an `80`-`9F` second byte, `F0` with an `80`-`8F` second
+# byte), UTF-16 surrogates (`ED` with an `A0`-`BF` second byte) and code points
+# above U+10FFFF (`F4` with a `90`-`BF` second byte, and `F5`-`FF` leads) are
+# rejected. A sequence that is incomplete or malformed terminates the
+# computation, exactly like `slice` does for a truncated character.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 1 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (125 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2148](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6739, opened 2026-08-13
Write down what an atom comes back with

### LINKED ISSUE #6738: `Provides` ignores declared return type of an atom in `Provides.java`
The loop in `Provides.follow` at `eo-inference/src/main/java/org/eolang/inference/Provides.java:83` walks `o[@name and not(@name='λ')]`, and `λ` is exactly where an atom keeps the type it comes back with. `[] > div /Q.number` at `eo-runtime/src/main/eo/number.eo:19` reaches the module like this:

```xml
<o line="19" loc="Φ.number.div" name="div">
  <o base="∅" line="20" loc="Φ.number.div.x" name="x" type="Φ.number"/>
  <o atom="Φ.number" line="19" loc="Φ.number.div.λ" name="λ"/>
</o>
```

The row written for `Φ.number.div` names the void `x` and nothing else. The `@atom` cell, which says a `div` comes back as a `Φ.number`, is stepped over.

The class javadoc at `Provides.java:54` explains the gap as a fact of life: "the results of atoms are not here at all yet: what an atom returns is written in Java, so its rows will have to be given to this table from outside one day". That was true before the annotation existed. It is not true now — the return type is declared in the EO text and the parser carries it into the XMIR, so nothing has to come from outside.

What it costs is paid downstream. `Provided.kept` at `Provided.java:78` looks for an attribute in the three places its javadoc names — the type itself, its package, behind its `φ` — and an atom has none of them, since it binds only voids and a `λ`. So `.plus` asked of what `.div` came back with finds nothing, the dispatch stays unanswered, and every chain that runs through it stops there. On a clean `eo-runtime` build of `d314aa6f9e`, 4,380 of 10,355 dispatches have a type; of the 5,975 that do not, 388 sit directly on an atom that declares its return, and far more sit behind those. Adding the declared return as a fourth place to look answers 4,930 instead of 4,380 — 550 more dispatches, and 550 more rows in `links.xml`, off 25 annotations.

Have `Provides` write what an atom returns as a cell on the type row, read from the `@atom` of the `λ` child, and have `Provided` ask that type when the atom binds no attribute of the name asked, guarded against a circle the way the `φ` walk already is. One caveat: an annotation is not always a locator. `recovered.eo:15` declares `[] > recovered /A` with `? > value /A?`, where `A` is a type variable and `?` marks fragility, so the rule should take a locator and stay silent on a variable — variables are later work, and only 1 of the 25 atoms in `eo-runtime` is polymorphic. This is the neighbour of #6189, which asks for annotations to be added where they are missing, while this one asks for the ones that already exist to be read; one pack per behaviour under `eo-maven-plugin/src/test/resources/org/eolang/maven/inference-packs` describes it.

maxonfjvipon:
An atom keeps its body in Java, which this module cannot read, and the tables treated what it comes back with as equally unreadable — the class javadoc of `Provides` said so outright, that "the results of atoms are not here at all yet". That has not been true for a while. `[] > div /Q.number` says a `div` is a `Φ.number` once it has run, and the parser carries the annotation into the XMIR:

```xml
<o loc="Φ.number.div" name="div">
  <o base="∅" loc="Φ.number.div.x" name="x" type="Φ.number"/>
  <o atom="Φ.number" loc="Φ.number.div.λ" name="λ"/>
</o>
```

The rule walked `o[@name and not(@name='λ')]` and stepped straight over the one child that carries it. So `Φ.number.div` was a type with a void and nothing else, `.plus` asked of what a `div` came back with found nothing, and every chain through an atom stopped there.

Now the row about a type keeps what it returns, and `Provided` asks that type when the atom binds no attribute of the name asked. It is the same move it already makes behind a `φ`, so it belongs in the same place: a formation hands its answers to what stands behind its `φ`, an atom to what it says it comes back with, and the cycle guard covers both.

An annotation that names no object is skipped. `[] > recovered /A` comes back with whatever the caller put in, and a variable is nobody's locator; one of the 25 atoms in `eo-runtime` is polymorphic like that, and it stays unanswered until variables mean something here.

Measured on a clean `eo-runtime` build: dispatches that have a type 4,380 → 4,930 of 10,355; links rows 26,153 → 26,703; dispatch receivers whose type is known 8,129 → 8,290; delegation chains that end at a locator nothing describes 1,790 → 1,672 of 2,444. The `complete` flag does not move — an atom is still not complete, since its body is still unread. Three packs describe the behaviour: what the table keeps, what a dispatch on an atom's result turns out to be, and the silence about a variable.

Closes #6738

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Provided.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Provided.java
@@ -22,8 +22,9 @@ import java.util.Map;
  * through to the object of that name beside it — {@code Φ.number} binds eight
  * attributes and answers to forty, the rest of them being objects of their own
  * in the same package, {@code Φ.number.eq} among them, and a locator names both
- * kinds the same way. And whatever stands behind its {@code φ}, which answers
- * for every name the object does not bind itself.</p>
+ * kinds the same way. And whatever the type hands its answers to, which is
+ * what stands behind its {@code φ}, or, for an atom, the type it says it comes
+ * back with.</p>
  *
  * <p>The walk stops at a type it has already passed, so an object that
  * delegates in a circle is walked once and answers nothing. A void answers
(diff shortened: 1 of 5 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (108 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+667](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6740, opened 2026-08-13
Move misplaced package members to the package of their receiver

### LINKED ISSUE #6655: Package members misplaced in a package that does not match their receiver
Five package members are declared in a package that does not match the type of their receiver:

```
string/as-hex.eo:        [b] > as-hex               # b is bytes
string/as-char.eo:       [code] > as-char           # code is a number
string/as-decimal.eo:    [n] > as-decimal           # n is a number
string/as-fixed.eo:      [n places cant-print] > as-fixed
string/as-scientific.eo: [n] > as-scientific        # n is a number
```

Today this is only mildly wrong: the package tells the reader that `as-hex` belongs to the surface of `string`, when what it really does is render bytes, and the runtime never contradicts that because the implicit form `01-02.as-hex` cannot reach it anyway. `PhDefault.absent` probes `Φ.bytes.as-hex`, finds nothing, and the object is usable only through the explicit `string.as-hex 01-02`. So the member is misfiled and the convenient form it was written for does not work.

It becomes load-bearing under #6616, where the first void turns into `^` and the enclosing package decides what `^` is. An `as-hex` that lives in `string` and reads `^` would receive a string, which is not what the body wants.

Move `as-hex.eo` to `bytes/` and the other four to `number/`, leaving the bodies and the explicit call sites as they are, apart from the references that the move itself breaks. The members keep their first void for now, so the change is a relocation and nothing else, and it makes `01-02.as-hex` reachable through the existing package rule on the way.

A moved file loses its old siblings, so every bare name in it has to be re-read. `bytes/as-hex.eo` calls `as-char` and `as-ascii`, and once it sits in `bytes` neither resolves: `as-char` becomes a dispatch on the number it is given, and `as-ascii` stays in `string`, where it belongs, so as-hex names it explicitly. `string/printf.eo` loses three siblings the same way and dispatches on its receivers instead. `as-scientific` has to move together with `as-fixed` and `as-decimal`, since it calls both.

`bytes/hash.eo` was listed here at first and does not belong. Its void is *named* `input`, but the name means "any object that has `as-bytes`", not the `input` role: the file hashes `true`, `42`, `"hello"` and an abstract object, and `map.eo` already passes bytes to it. Moving it to `input/` would give `^` an input stream under #6657 and break every one of those. It is in the right package already — it hashes bytes, its body starts with `input.as-bytes`, and the honest end state is `^` as bytes with the caller converting, which is what `map.eo` does today.

maxonfjvipon:
Five members sat in `string` while their receiver is a number or bytes. `as-hex` renders bytes, `as-char`, `as-decimal`, `as-fixed` and `as-scientific` all read a number, so `01-02.as-hex` and `65.as-char` could never resolve — the package rule probes `Φ.bytes.as-hex`, finds nothing, and only the explicit `string.as-hex 01-02` worked. They move to `bytes/` and `number/`, which makes the dispatched form reachable and puts the right object behind `^` when #6657 replaces the first void.

A moved file loses its old siblings, so every bare name in it has to be re-read. `as-hex` called two:

```
    as-char.
      h.plus 48
    string.as-ascii > bval
      b.slice index 1
```

`as-char` is now a dispatch on the number it is handed, and `as-ascii` stays in `string`, where it belongs, so `as-hex` names it explicitly — a byte handed to a string member, which #6657 will have to resolve one way or another. `printf` lost three siblings the same way and dispatches on its receivers instead, and `as-scientific` had to move together with `as-fixed` and `as-decimal`, since it calls both.

The 59 tests in the moved files stop being standalone applications and dispatch, which is what #6682 could not do for them while they were misfiled. Four `--> stops-on-a-fractional-*` tests that arrived in `tuple` with #6699 after #6682 merged are converted here too, so the standalone form does not creep back.

`bytes/hash.eo` was in the issue's list and is dropped from it. Its void is *named* `input`, but that means "any object with `as-bytes`" and not the `input` role: the file hashes `true`, `42`, `"hello"` and an abstract object, and `map.eo` already hands it bytes. Moving it to `input/` would put an input stream behind `^` and break all of that. It is already in the right package.

`mvn compile` is clean — 171 sources canonical, lint with no complaints, 180 Java files transpiled and 315 compiled.

Closes #6655.

### DIFF
--- a/eo-runtime/src/main/eo/bytes/as-hex.eo
+++ b/eo-runtime/src/main/eo/bytes/as-hex.eo
@@ -3,7 +3,7 @@
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
-+package string
++package bytes
 +version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
@@ -11,9 +11,9 @@
 [b] > as-hex
   if. > [h] >> hex-digit
     h.lt 10
-    as-char
+    as-char.
       h.plus 48
-    as-char
+    as-char.
       h.plus 55
   [index acc] >> rec-hex
     if. > @
@@ -28,7 +28,7 @@
           concat.
             acc.concat "-".as-bytes
             pair
-    as-ascii > bval
+    string.as-ascii > bval
       b.slice index 1
     concat. > pair
       hex-digit
@@ -46,25 +46,20 @@
 
   eq. ++> can-write-a-single-byte
     "41"
-    string
-      as-hex "A".as-bytes
+    string "A".as-bytes.as-hex
 
   eq. ++> can-join-bytes-with-hyphens
     "4A-65-66-66"
-    string
-      as-hex "Jeff".as-bytes
+    string "Jeff".as-bytes.as-hex
 
   eq. ++> can-write-letter-nibbles
     "00-00-00-00-00-00-00-FF"
-    string
-      as-hex 255.as-i64.as-bytes
+    string 255.as-i64.as-bytes.as-hex
 
   eq. ++> can-write-the-bytes-of-a-number
     "40-45-00-00-00-00-00-00"
-    string
-      as-hex 42.as-bytes
+    string 42.as-bytes.as-hex
 
   eq. ++> can-write-empty-bytes
     ""
-    string
-      as-hex --
+    string --.as-hex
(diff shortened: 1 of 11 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (183 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+679](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6742, opened 2026-08-13
#6701: reject an invalid scheme prefix as schemeless in uri

### LINKED ISSUE #6701: `uri` parser incorrectly recognizes prefixes with digits or punctuation as valid RFC 3986 schemes
## What happens

`uri` describes itself as an RFC 3986 generic URI parser, but it treats every
prefix before a colon as a scheme. Consequently, invalid scheme starts are
accepted as though they were valid schemes.

| expression | actual scheme | RFC 3986 result |
| --- | --- | --- |
| `(uri "1http://host/path").scheme` | `"1http"` | no scheme |
| `(uri "+foo:bar").scheme` | `"+foo"` | no scheme |
| `(uri "a+b.c:part").scheme` | `"a+b.c"` | `"a+b.c"` |

The results were reproduced through runtime objects on current `master`
(`7591007fd7`). The final row is a control case: `+` and `.` are legal after
the initial ASCII letter, but neither is legal at the beginning.

## Why it happens

[`uri.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/uri.eo#L132)
uses only `string.index-of text ":"` to locate a scheme. It returns the prefix
unconditionally and never validates the RFC grammar `ALPHA *( ALPHA / DIGIT /
"+" / "-" / "." )`.

## Expected behavior

An invalid scheme prefix must not be exposed as a valid scheme. The parser may
treat such input as a schemeless path, or reject it, but it must preserve the
grammar stated in the object's documentation.

## Suggested fix

Validate the substring before the first colon: require an ASCII letter at
position zero and permit only letters, digits, `+`, `-`, and `.` afterwards.
Add regressions for `1http://host`, `+foo:bar`, and the valid `a+b.c:part`.

morphqdd:
`uri` treats any prefix before the first colon as the scheme, without checking it against the RFC 3986 grammar (`ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`). `1http://host/path` exposed `"1http"` as a scheme and `+foo:bar` exposed `"+foo"`.

Added a validity check on the candidate substring before the colon: it must start with an ASCII letter and contain only letters, digits, `+`, `-`, `.` after that, checked with a regex (`string.regex`). When the check fails, the object now treats the colon as if it were absent (same as a text with no colon at all), so the whole `uri.scheme` becomes `""` and the malformed prefix falls into `path` instead, consistent with the existing schemeless-URI behavior. A valid scheme like `a+b.c` still parses correctly.

Added `can-parse-a-scheme-with-a-plus-and-a-dot`, `cannot-treat-a-digit-led-prefix-as-a-scheme`, `cannot-treat-a-plus-led-prefix-as-a-scheme`.

Closes #6701

### DIFF
--- a/eo-runtime/src/main/eo/uri.eo
+++ b/eo-runtime/src/main/eo/uri.eo
@@ -137,7 +137,16 @@
     ""
   if. > scheme
     eq.
-      string.index-of text ":" >> scheme-colon
+      if. >> scheme-colon
+        or.
+          eq.
+            string.index-of text ":" >> raw-colon
+            -1
+          not.
+            "/[A-Za-z][A-Za-z0-9+.-]*/".regex.compiled.matches
+              text.slice 0 raw-colon
+        -1
+        raw-colon
       -1
     ""
     text.slice
@@ -190,6 +199,11 @@
     port.
       uri "http://host:/"
     ""
+
+  eq. ++> can-parse-a-scheme-with-a-plus-and-a-dot
+    scheme.
+      uri "a+b.c:part"
+    "a+b.c"
 
   eq. ++> can-parse-the-empty-host-of-a-triple-slash-uri
     host.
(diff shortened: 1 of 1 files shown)


morphqdd:

yegor256:
@morphqdd conflicts here

morphqdd:
@yegor256, conflicts solved

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4196](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6743, opened 2026-08-13
Merge package members into the object at compile time

maxonfjvipon:
An object and the package of its name are two separate things today, and it takes the runtime to join them: asked for an attribute it does not hold, `Φ.number` goes looking for `Φ.number.lt` on the classpath and binds itself into the first void of whatever it finds. `MjMerge` does that joining at compile time, after `lint` so every member is still reported on as the file a human wrote, and before `transpile`, the first goal that cares about the shape of what it compiles:

```java
formation.appendChild(
    formation.getOwnerDocument().importNode(top.node(), true)
);
```

That one move is the whole splice. Nothing inside the member is rewritten, because the parser leaves every name fully qualified and every `loc` already reads as the locator the node will carry once it is an attribute of `Φ.number`, so nothing in the body can be captured by an attribute of the object it lands in. Appending rather than prepending keeps the places of the voids, and with them the meaning of applying the object to arguments.

The merged XMIR goes to `4-merge/`, the object's tojo is pointed at it, and each member is marked `merged` with the name of the object it went into, which takes it out of `standalone()` and so out of transpiling. The mark is read only there, deliberately: `withXmir()` stays as it was, since `eo-foreign.csv` outlives a build and a mark left in it would otherwise stop `MjLint` from ever looking at that member again.

Two refusals, both loud: a package named for merging that this build has no object for, and a member whose name the object already holds — one formation cannot carry two attributes under one name.

Enabling a package is not part of this. `mergedPackages` is empty, `merge` is bound in `eo-runtime` as a no-op, and master keeps today's behaviour. That is not caution for its own sake: I merged the real `bytes` package as a smoke test and 17 members spliced, transpiled and compiled, but 36 of 44 `EObytesTest` cases then failed with `Φ.bytes.as-bytes.φ` reaching ⊥ — the receiver used to arrive through that first void, and once the attribute really exists nothing puts it there. That is exactly what #6657 replaces with `^`, so each package gets switched on by the pull request that migrates it.

What the merge does to a program is stated as packs of EO files rather than as XMIR by hand, so each one reads as the program a human would write, and one of them goes on through `transpile` to check that a member really does come out as an attribute of the class of its package:

```yaml
java:
  number.eo:
    - 'class EOnumber'
    - 'this.add("lt"'
```

A pack found a bug worth having: with `number` and `number.i64` both named, `number` was merged first and took in an `i64` that had not yet been given its own members. Packages are merged deepest first now. The mechanics no EO source can express — where the object is pointed, what stops being compiled apart, and both refusals — stay as Java tests. `mvn compile` on `eo-runtime` is clean with the goal bound and empty.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Merging.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Merging.java
@@ -12,10 +12,12 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.TreeMap;
+import java.util.stream.Collectors;
 import org.w3c.dom.Node;
 
 /**
@@ -87,7 +89,7 @@ final class Merging implements Step {
         } else {
             final Map<String, TjForeign> all = this.indexed();
             int done = 0;
-            for (final String pkg : this.packages) {
+            for (final String pkg : this.deepest()) {
                 done = done + this.spliced(pkg, all);
             }
             Logger.info(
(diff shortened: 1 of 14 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (917 >= 100); -8 for way too many hits-of-code (917 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+683](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6744, opened 2026-08-13
fix(#6605): reject writes past chunk bounds

### LINKED ISSUE #6605: `chunk.write` silently grows a block past its allocated size in `Heaps.java`
`Heaps.write` at `eo-runtime/src/main/java/org/eolang/Heaps.java:217` enlarges the block whenever the data runs past its end, so a write at an offset beyond the allocation silently grows it instead of failing. After a `malloc.of 8` a `c.write 100 "x"` therefore leaves a 101-byte block, and `c.size` afterwards reports 101, so the size handed to `malloc.of` stops bounding anything.

This is the code #5776 named as its root cause, and it is still in place. That issue was fixed only in `eo-runtime/src/main/eo/chunk.eo`, in commits `96127accb5` and `bf4f6a2579`; `Heaps.java` was never touched. The new guard sits on `put` alone, at `chunk.eo:41-47`, while `write` is a separate atom declared at `chunk.eo:58` and shown as public API in that file's own header example (`c.write 2 "Hello"`). `EOchunk$EOwrite.java:30` reaches `Heaps.write` directly, so `c.write` walks straight past the check that `c.put` now performs.

The read path on the same abstraction disagrees with it. `Heaps.fits` refuses a range that lies outside the block, and `EOchunk$EOread` routes that case to the caller's `cant-read` fallback rather than read past the allocation. Growing on write also makes `chunk.resized`, the explicit grow operation with an atom of its own, pointless, and it lets a miscomputed offset claim an arbitrary amount of memory without complaint.

Dropping the resize at `Heaps.java:217-219` and throwing an `ExFailure` naming the identifier, the offset and the allocated length would put `write` in line with both `put` and `read`; a caller that genuinely needs a larger block already has `chunk.resized`.

@maxonfjvipon @yegor256 please take a look.

fzlzjerry:
## Summary

- reject `Heaps.write` ranges that extend past the allocated block instead of resizing implicitly
- preserve the intended growth behavior of empty `put`, `copy`, and `to-output` through explicit `resized` calls
- verify the rejected write leaves both the allocation size and the original bytes unchanged

## Validation

- `mvn -ntp -PskipITs --errors --batch-mode -Deo.coverageTracking=false -Deo.deadline=2 clean install -pl :eo-runtime` (2059 tests, 0 failures, 0 errors, 351 skipped)
- `mvn -ntp -PskipTests -Pqulice --errors --batch-mode verify -pl :eo-runtime`

Fixes #6605

### DIFF
--- a/eo-runtime/src/main/eo/chunk.eo
+++ b/eo-runtime/src/main/eo/chunk.eo
@@ -30,11 +30,19 @@
 
 [id] > chunk
   get > @
-  write > [source target length] > copy
-    target
-    read
-      source
-      length
+  [source target length] > copy
+    seq * > @
+      read source length >> data!
+      if.
+        gt.
+          target.plus length
+          size
+        write.
+          resized
+            target.plus length
+          target
+          data
+        write target data
   read > get
     0
     size
@@ -45,9 +53,17 @@
     T
       "Can't put %d bytes into a chunk sized %d bytes; call .resized first to grow it".printf
         * object.size size
-    seq *
-      write 0 object
-      get
+    if.
+      size.eq 0
+      seq *
+        write.
+          resized object.size
+          0
+          object
+        get
+      seq *
+        write 0 object
+        get
   [] > read /Q.bytes
     ? > offset /Q.number
     ? > length /Q.number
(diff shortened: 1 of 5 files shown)


0crat:
@fzlzjerry Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (128 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6745, opened 2026-08-13
#6702: reject a non-digit port suffix in uri

### LINKED ISSUE #6702: `uri.port` incorrectly identifies non-numeric text as valid port values
## What happens

`uri` exposes non-numeric text after an authority colon as `port`, although its
documentation says it parses RFC 3986 generic URIs and RFC 3986 defines
`port = *DIGIT`.

| expression | actual port | RFC 3986 result |
| --- | --- | --- |
| `(uri "http://host:8080/").port` | `"8080"` | valid |
| `(uri "http://host:/").port` | `""` | valid empty port |
| `(uri "http://host:abc/").port` | `"abc"` | invalid |
| `(uri "http://host:12x/").port` | `"12x"` | invalid |

The results were reproduced through runtime objects on current `master`
(`7591007fd7`).

## Why it happens

[`uri.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/uri.eo#L81)
locates the colon and returns the whole remaining suffix. It does not validate
the suffix characters before exposing them through `port`.

## Expected behavior

The parser should not represent `abc` or `12x` as a port. It should reject an
authority containing a non-digit port, or make the component unavailable while
reporting the malformed authority.

## Suggested fix

After slicing the suffix, ensure every character is an ASCII digit (while
allowing the empty string). Add regressions for `:8080`, `:`, `:abc`, and
`:12x`.

morphqdd:
`uri` exposes everything after the authority's port colon as `port`, without checking that the suffix is actually digits (RFC 3986 defines `port = *DIGIT`). `http://host:abc/` exposed `"abc"` and `http://host:12x/` exposed `"12x"`.

Added a validity check on the sliced port suffix: it must be either empty or all ASCII digits (`string.is-digit`). When invalid, `port` returns `""` instead of the malformed text, the same way it already does when no port colon is present at all. `host` and `path` are unaffected since they only depend on the colon's position, not its content.

Added `can-parse-an-empty-port`, `cannot-treat-a-letter-suffixed-port-as-numeric`, `cannot-treat-a-non-digit-port-as-numeric`.

Closes #6702

### DIFF
--- a/eo-runtime/src/main/eo/uri.eo
+++ b/eo-runtime/src/main/eo/uri.eo
@@ -246,12 +246,12 @@
       uri "http://user:pass@host:80/some/path"
     "user:pass"
 
-  eq. ++> cannot-treat-a-letter-suffixed-port-as-numeric
+  eq. ++> can-treat-a-letter-suffixed-port-as-invalid
     port.
       uri "http://host:12x/"
     ""
 
-  eq. ++> cannot-treat-a-non-digit-port-as-numeric
+  eq. ++> can-treat-a-non-digit-port-as-invalid
     port.
       uri "http://host:abc/"
     ""
--- a/eo-runtime/src/main/eo/uri.eo
+++ b/eo-runtime/src/main/eo/uri.eo
@@ -68,7 +68,25 @@
     ""
   if. > host
     not. >> has-port
-      port-colon-idx.eq -1
+      eq.
+        if. >> port-colon-idx
+          string.starts-with host-port "[" >>
+          if.
+            eq.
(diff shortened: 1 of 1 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4172](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6746, opened 2026-08-13
Let a void be a type

### LINKED ISSUE #6741: Void termination of attribute chains causing resolution issues in `Provided.java`
`Provided.kept` at `eo-inference/src/main/java/org/eolang/inference/Provided.java:78` looks for an attribute in three places — the type itself, its package, behind its `φ` — and a void is none of them, since the table has no row for one. So `Dispatched` leaves the dispatch alone and the chain stops. That would be a corner case if the core of the runtime were not itself a void. `eo-runtime/src/main/eo/number.eo:14` says:

```
[as-bytes] > number
  as-bytes > @
```

A `number` is a wrapper with one void, and it hands its answers to that void. `string`, `i64` and the rest of the core have the same shape, so every name they do not bind themselves falls into a hole. On a clean `eo-runtime` build of `5a4ad81219`, with #6684 and #6738 both in, 5,037 of 10,642 dispatches have a type, 1,737 delegation chains end at a locator nothing describes, and 48 of 2,592 types are complete — no core type among them, since a delegator is whole only when what it delegates to is whole.

Let a void be a type, by naming its attributes with its own locator. Inside `[x] > inc`, the object `x.next` is `Φ.inc.x.next` — the `next` of whatever fills `x` — which is exactly what the text says and nothing more. Chains compose again instead of stopping: `x.next.foo` is `Φ.inc.x.next.foo`. Where a copy fills the void, the prefix is rewritten with what fills it, so `inc t` turns `Φ.inc.x.next` into `Φ.t.next`, resolved onward like any other pair. The only names ever invented are ones the text already carries, which is what keeps this affordable.

Simulated on the same build, naming a void's attributes answers 10,243 dispatches instead of 5,037 and adds 5,206 rows; rewriting prefixes where a filling is known raises the answers that are concrete — rooted at a described formation rather than at a void — from 4,530 to 4,790. Most of the new answers are relative to a void until a context makes them concrete, so the honest measure of this change is that concrete count and not the coverage one. Two other readings were measured and are worse: instantiating the body of every copy, the naive meaning of contextual types, would write 831,672 rows, thirty times the whole table; and typing a void from a filler that every call site agrees on fails on exactly the voids that matter, since `Φ.number.as-bytes` is filled 28 different ways — 4,778 of them `Φ.bytes` and the tail chains that are themselves untyped — and taking the majority would be a guess.

Two things the implementation must respect. A locator rooted at a void is not a described type: nothing is known about it, `complete` stays false, and no verdict may ever rest on one. And a void whose name matches a member of the same package would be reachable two ways after #5486, so the void wins, because a name the formation binds is not a name that falls through to the package beside it. This is the half #6684 left out on purpose, and the last structural difference from the Python prototype, whose `infer(loc, ctx)` memoizes on locator and context for this very reason.

maxonfjvipon:
A void was a dead end: the table has no row for one, so `Provided` stopped its walk there and `Dispatched` left the dispatch alone. That would be a corner case if the core of the runtime were not itself a void. `number.eo` says

```
[as-bytes] > number
  as-bytes > @
```

and `string`, `i64` and the rest have the same shape, so every name they do not bind falls into a hole and the chain stops at the first step.

A name taken from a void is now the locator of the void with that name after it. Inside `[x] > inc`, the object `x.next` is `Φ.inc.x.next` — the `next` of whatever fills `x` — which is exactly what the text says and nothing more, and it composes: `x.next.foo` is `Φ.inc.x.next.foo`. The only names invented are ones the text already carries, which is what makes this affordable; instantiating the body of every copy, the other reading of contextual types, would have written 831,672 rows against a table of 27,000.

Where a receiver resolves through an application, the void is filled, and `Filled` puts what fills it in place of it and asks the rest of the names again, so `inc t > held` makes `held.next` the `next` of a `t`. The fillings are gathered along the whole chain the receiver resolves through, since the application that fills a void and the dispatch that asks about it are rarely the same object.

Certain answers are settled before any void is looked into:

```java
final Map<String, String> pairs = new Settled(
    new Dispatched(given, dispatches, args, given.xpath("//attr[@void='true']/@type"))
).from(
    new Settled(
        new Dispatched(given, dispatches, args, Collections.emptyList())
    ).from(new Pairs(new XMLDocument(links)).all())
);
```

Without that, an answer relative to a void would win a race against a concrete answer that needs another pass, and the table would know less while appearing to know more.

Measured on a clean `eo-runtime` build, the same parse with the change off and on: dispatches that have a type 4,928 → 10,017 of 10,372; receivers whose type is known 8,338 → 10,218; delegation chains that end at a locator nothing describes 1,673 → 1,420 of 2,445; links rows 26,756 → 31,761, of which 6,604 are rooted at a void and 25,157 are concrete, up from 24,906. Most of the new answers are relative to a void until a caller makes them concrete, so the honest number here is that concrete count and not the coverage one. The `complete` flag does not move, and must not: nothing is known about a locator rooted at a void, so no verdict may ever rest on one.

Three packs describe the behaviour and `note-example.yaml` gains the two rows it was always about — the design note's own `x.next.foo` is now written down as the `foo` of the `next` of whatever fills `x`.

Closes #6741


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (450 >= 100); -8 for way too many hits-of-code (450 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+687](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6750, opened 2026-08-13
#6710: re-qualify Probing sibling names to match the probed object's prefix

### LINKED ISSUE #6710: `Probing` incorrectly registers short names from `ObjectsIndex.children()` as fully-qualified, leading to duplicated tojos
## What happens

`ObjectsIndex.children()` strips the `org.eolang.` prefix from its input and returns index-form (short) names, for example `number.gte` (confirmed by the repo's own test `ObjectsIndexTest.listsDirectChildrenOfPackageWithOrgEolangPrefix`).

`Probing.complete` (`eo-maven-plugin/src/main/java/org/eolang/maven/Probing.java:141-155`) takes those short names and registers each one with `this.tojos.add(sibling)` as if it were already a fully-qualified name. For a plain `Q.number.gte` reference this is fine, but for an `org.eolang.`-prefixed reference, the same object ends up registered twice under two different identifiers: once under the short form (`number.gte`) and once under the `org.eolang.`-prefixed form.

Verified with a temporary test probing `Q.org.eolang.number.gte` against an index of `{number, number.gte, number.sqrt, number.abs}`. The registered tojo identifiers were:
```
number, number.abs, number.gte, number.sqrt, org.eolang.number, org.eolang.number.gte, test
```

`number.gte` and `org.eolang.number.gte` both point at the same object but are tracked as two separate tojos. The `org.eolang.`-prefixed one will not match the real object's declared FQN later (`objects/number/gte.eo` declares `+package number`, so its real FQN is `number.gte`), which can surface as a downstream "object not found" or a duplicate-processing issue for any `org.eolang.`-prefixed reference.

This looks like a residual gap left by #6213 ("`ObjectsIndex.children` does not strip the `org.eolang.` prefix that `contains` strips"), which fixed the input side of `ObjectsIndex` but left `Probing` treating the (now short-form) output as already fully qualified.

## What should happen

`Probing.complete` should re-qualify the names it gets back from `ObjectsIndex.children()` (e.g. re-attach the `org.eolang.` prefix, or resolve to the object's real FQN) before registering them, so a single object is never tracked under two different tojo identifiers.

morphqdd:
`ObjectsIndex.children()` returns short (index-form) names, e.g. `number.gte`, even when queried with an `org.eolang.`-prefixed package. `Probing.complete` registered those short names as-is (`this.tojos.add(sibling)`), so a probed reference like `Q.org.eolang.number.gte` ended up tracked under two different tojo identifiers: `org.eolang.number.gte` (the probed object itself) and `number.gte` (the same object, discovered again as its own package's sibling).

Re-qualified the sibling names in `Probing.complete`: when the probed object carries the `org.eolang.` prefix, siblings that don't already carry it get it prepended before registration, so the object is tracked under one identifier consistently.

Added `avoidsDuplicateTojoForAnOrgEolangPrefixedReference`, reproducing the case from the issue (`Q.org.eolang.number.gte` against an index containing `number`, `number.gte`, `number.sqrt`, `number.abs`) and asserting the short form is not also registered.

Closes #6710

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Probing.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Probing.java
@@ -144,12 +144,18 @@ final class Probing implements Step {
         final Set<String> completed
     ) throws IOException {
         final int split = object.lastIndexOf('.');
-        if (split > 0) {
-            final String pkg = object.substring(0, split);
-            if (completed.add(pkg)) {
-                for (final String sibling : this.objectionary.children(pkg)) {
-                    this.tojos.add(sibling).withDiscoveredAt(src);
+        final String pkg = object.substring(0, Math.max(split, 0));
+        if (split > 0 && completed.add(pkg)) {
+            final String root = "org.eolang.";
+            final boolean rooted = object.startsWith(root);
+            for (final String sibling : this.objectionary.children(pkg)) {
+                final String qualified;
+                if (rooted && !sibling.startsWith(root)) {
+                    qualified = root.concat(sibling);
+                } else {
+                    qualified = sibling;
                 }
+                this.tojos.add(qualified).withDiscoveredAt(src);
             }
         }
     }
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/ProbingTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/ProbingTest.java
@@ -75,6 +75,40 @@ final class ProbingTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4372](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6751, opened 2026-08-13
#6600: retry write with the unwritten tail on a partial write

### LINKED ISSUE #6735: console.eo doesn't correctly write all remaining bytes
The puzzle `6600-d899aad7` from #6600 has to be resolved: 

https://github.com/objectionary/eo/blob/4a804b9f0966c37dc8aa0fc9b25589c6aa4cd1e1/eo-runtime/src/main/eo/console.eo#L59-L63

The puzzle was created by @yegor256 on 13-Aug-26. 

morphqdd:
Resolves the `@todo` puzzle from #6600: `console.write` and `stderr.write` only checked the `write` syscall's return code for `-1` (failure). A successful write of fewer bytes than the buffer's size (a partial write) was treated as if the whole buffer went out, silently dropping the unwritten tail.

Both `write` implementations (POSIX and Windows, in `console.eo` and `stderr.eo`) now compare the syscall's byte count against the buffer size. On a short write, they retry through the same output block with the remaining unwritten slice (`buffer.slice`), until the whole buffer is flushed or a syscall fails.

Added `can-chain-writes-across-output-blocks` in `console.eo`, chaining two writes through the returned output blocks. A genuine short write can't be forced deterministically from EO (the underlying syscall is a native atom), so this covers the retry path structurally without breaking the normal one; `can-write-to-the-console` and `can-print-to-stderr` already cover the full-write path through the same code.

Ran `EOconsoleTest`/`EOstderrTest` and `qulice:check -Pqulice`, both green.

Closes #6735

### DIFF
--- a/eo-runtime/src/main/eo/console.eo
+++ b/eo-runtime/src/main/eo/console.eo
@@ -55,12 +55,6 @@
 # ```
 #
 # Sequential reuse maintains the console state across operations.
-#
-# @todo #6600:30min Keep writing the remaining bytes on a partial write. Right
-#  now `write` panics only when the `write` syscall returns -1; a successful
-#  write of fewer bytes than `buffer.size` is accepted as if the whole buffer
-#  went out, same as `stderr.write`. Both should retry with the unwritten
-#  tail until the buffer is fully flushed or a syscall fails.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4212](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6656, opened 2026-08-12
Need for `MjMerge` mojo to merge package members into the related named object


maxonfjvipon:
Part of #6616. The resolution of a package member has to move from the runtime to the compiler, and the compiler has all it needs by the time parsing ends: the parent formation and the members of its package are both on disk as XMIR. What is missing is the step that puts them together.

Add a mojo, `MjMerge`, bound between `lint` and `transpile`, that walks `this.scopedTojos().withXmir()` and, for every XMIR whose `+package p` names an object `Φ.p` that this same build also compiles, splices the member's top-level `<o>` into the parent's formation as a trailing child, writes the result to `4-merge/`, points the parent's tojo at it through `TjForeign.withXmir`, and marks the member's tojo so that `TjsForeign.withXmir()` stops handing it to `MjTranspile`. `MjAtomsTable` is the model for the walk.

Two properties of the parsed XMIR make the splice cheap. Names are already fully qualified, so nothing in a member's body can be captured by an attribute of the parent it lands in: `number/lt.xmir` reads

```xml
<o line="10" loc="Φ.number.lt" name="lt" pos="0">
  <o base="∅" line="10" loc="Φ.number.lt.n" name="n" pos="1"/>
  <o base=".gt" line="11" loc="Φ.number.lt.φ" name="φ" pos="3">
    <o base="Φ.number" line="11" loc="Φ.number.lt.φ.ρ" pos="2">
```

and every `loc` is already the locator the node would carry as an inner attribute of `Φ.number`, so no attribute of the moved tree needs rewriting. Appending rather than prepending keeps the parent's void order, and therefore the meaning of `put(0, ...)`, exactly as it was.

Two things need care. A splice must fail loudly, not silently, when the parent and the member, or two members, declare the same name, because one formation cannot hold two attributes under one name. And the transpile cache key of the parent has to take in the hash of every member, otherwise an edit to `number/lt.eo` leaves the previously merged `number` in place and the build silently compiles yesterday's object.

Ship the mojo with a `mergedPackages` parameter that is empty by default, so master keeps its current behaviour and each package is switched on by the pull request that migrates it.

0pdd:
@maxonfjvipon the puzzle [#6747](https://github.com/objectionary/eo/issues/6747) is still not solved.

maxonfjvipon:
Done in #6743. The mojo is bound in `eo-runtime` with an empty `mergedPackages`, so no package is merged yet — #6657 switches them on one at a time.

[...]

PR #6752, opened 2026-08-13
#6656: write the merged XMIR only when its content differs

### LINKED ISSUE #6747: Merged XMIR is written regardless of conditions in Merging.java
The puzzle `6656-5b5e463c` from #6656 has to be resolved: 

https://github.com/objectionary/eo/blob/87d8f77e37ca8b1726136b43ef31bae931775d88/eo-maven-plugin/src/main/java/org/eolang/maven/Merging.java#L40-L47

The puzzle was created by @yegor256 on 13-Aug-26. 

morphqdd:
Resolves the `@todo` puzzle from #6656: `Merging.spliced` wrote the merged XMIR unconditionally, so its timestamp moved on every build even when neither the package object nor any of its members had changed, forcing `Transpiling` to recompile it every time.

Now the merged text is compared against what's already on disk (the same `Diff` class `MjFormat` already uses for its own canonical-form comparison), and the file is written only when the target is missing or the content differs.

Added `MergingTest` (the class had no unit tests before): merges a package twice from the same unchanged inputs and asserts the merged file's mtime doesn't move on the second run. Verified it fails without the fix (mtime always moves) and passes with it.

Ran `MergingTest` and `qulice:check -Pqulice`, both green.

Closes #6747

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Merging.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Merging.java
@@ -9,6 +9,7 @@ import com.github.lombrozo.xnav.Xnav;
 import com.jcabi.log.Logger;
 import com.jcabi.xml.XMLDocument;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4388](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6753, opened 2026-08-13
#6749: drop the reversed-dispatch compact-tuple form from the spec

### LINKED ISSUE #6749: `PARSER_SPEC.md` inaccurately portrays rejected 'reversed-dispatch compact-tuple form' as a valid feature
## What happens

`PARSER_SPEC.md` documents a reversed-dispatch compact-tuple form (`name. *N`) in three places:

- R-3.8.5: "For reversed dispatch with `*N`, `N >= 1` is required (the receiver occupies the first direct slot)."
- R-3.9.1: lists `bare-reversed` (`name. *N`) as one of the three valid compact-tuple heads.
- R-3.9.4: "Reversed-dispatch heads need `N >= 1`... `name. *` (no N, defaults to 0) and `name. *0` are rejected at the classifier."

This was discussed on #6706 and PR #6732. `@maxonfjvipon` pointed out the syntax is confusing regardless of the `N >= 1` restriction: `*` after a reversed-dispatch name conflicts with the method's own first argument, and for a complex head it does not add anything over the existing non-reversed form (`foo.joined *` with the tuple elements as vertical children). `@yegor256` agreed to drop the feature rather than implement it.

The spec text still describes the feature as valid grammar, but the parser currently rejects it (parsing `name. *N` as trailing garbage), and there is no plan to implement it.

## What should happen

Remove the reversed-dispatch compact-tuple form from `PARSER_SPEC.md`: drop `bare-reversed` from the head list in R-3.9.1, and remove R-3.8.5's `*N`-after-reversed-dispatch clause and R-3.9.4 entirely, so the spec matches what the parser actually accepts and what the language is meant to support.

morphqdd:
The reversed-dispatch compact-tuple form (`name. *N`) was discussed on #6706 and PR #6732: `@maxonfjvipon` pointed out the syntax is confusing (the `*` after a reversed-dispatch name conflicts with the method's own first argument, and for a complex head it adds nothing over the existing `foo.joined *` form), and `@yegor256` agreed to drop the feature.

Removed the feature from `PARSER_SPEC.md`:
- R-3.8.5's `*N`-after-reversed-dispatch clause and the `compact-tuple` (reversed sub-flavor) outer kind.
- The `joined. *1 > sixth` example.
- The `bare-reversed` bullet from R-3.9.1's list of valid compact-tuple heads, replaced with an explicit note that a reversed-dispatch head is not legal there.
- R-3.9.4 (`N >= 1` for reversed heads) entirely.
- The old R-3.9.5 is renumbered R-3.9.4, with its cross-reference to the removed rule dropped.

The parser already rejects `name. *N` as a parse error (this was never implemented), so this only brings the spec text in line with that and with the decision not to implement it.

Closes #6749

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -419,12 +419,9 @@ R-3.8.3. **Vertical form** — `name.` with no horizontal args:
 
 R-3.8.4. The horizontal and vertical forms are **mutually exclusive** — no mixing on the same dispatch.
 
-R-3.8.5. **Compact tuple suffix** `*N` may follow the dot only in the vertical form (§3.9). For reversed dispatch with `*N`, `N ≥ 1` is required (the receiver occupies the first direct slot).
-
 Outer kinds produced:
-- Vertical form, no compact tuple → **`bare-reversed`**.
+- Vertical form → **`bare-reversed`**.
 - Horizontal form (with hargs) → **`reversed-with-hargs`**.
-- Vertical form with compact tuple → **`compact-tuple`** (with sub-flavor *reversed*).
 
 ```
 if. > @                               ← vertical bare-reversed, name @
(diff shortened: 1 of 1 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4184](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6754, opened 2026-08-13
Write down what every object sits in

### LINKED ISSUE #6636: Failure to record the {@code ρ} in Provides.java interfering with inference process
The puzzle `6565-f964ea73` from #6565 has to be resolved: 

https://github.com/objectionary/eo/blob/6d7f8125b96462f0c8f9ffe3267d9bbf360614ea/eo-inference/src/main/java/org/eolang/inference/Provides.java#L65-L69

Estimate: 35 minutes,  role: DEV. 

maxonfjvipon:
Every object answers to `ρ` for the one it is written inside, and an outer name is lowered into a dispatch of exactly that, so the needs table asks about `ρ` all the time. Nothing ever wrote it down. That left the provides table saying something certainly wrong: that an object seen whole has no attribute of that name.

It is not hypothetical. Applying the parked verdict rule — report when the receiver's type is complete and has no such attribute — to a clean `eo-runtime` gave ten reports, and every one of the ten was this:

```
map.xmir:80     ρ asked of Φ.map.ctor.found.not-found
socket.xmir:160 ρ asked of Φ.socket.a🌵55-4
regex.xmir:104  ρ asked of Φ.string.regex.pattern
```

Nobody has to be asked for the answer. A locator is the path to where an object is written, so the nearest formation the path runs through is what the object sits in; `Parents` reads it off and writes one row per formation. The object a file is about runs through no formation and sits in the root, which is written as `Φ` and described by nothing — "it sits in the root" and "it sits in nothing" are different answers and only the first is true.

The walk itself is now `Nesting`. `Scope` was already doing it privately to resolve `ξ.ρ`, so the two agree by construction rather than by coincidence.

With this in, the rule reports **nothing** on a clean runtime, and still catches every typo I seed into it — eight of eight this time, misspelled dispatch names at judgeable sites, each reported as `zzzNN asked of Φ.malloc` and friends. The tables gain 2,518 rows, one per formation: links rows 31,826 → 31,845, dispatches with a type 10,049 → 10,068 of 10,404, delegation chains ending nowhere 1,421 → 1,413, and dispatches whose receiver is whole enough to judge 1,004 → 1,028.

Two packs describe the behaviour, and the four packs that counted attributes now say what they always meant, since `ρ` is not a name a formation binds. The puzzle text is gone from `Provides`.

Closes #6636

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Nesting.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Nesting.java
@@ -1,0 +1,57 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import java.util.Collection;
+
+/**
+ * What a locator is written inside.
+ *
+ * <p>A locator is the path to where an object is written, so walking it back
+ * one name at a time and stopping at the first formation is how the object
+ * something is written inside is found, and nobody has to be asked. Applied to
+ * a reference it answers with the object being formed around it, which is what
+ * {@code ξ} names; applied to that answer it gives what the object sits in,
+ * which is what {@code ξ.ρ} names.</p>
+ *
+ * <p>A top-level object is written inside a package rather than a formation,
+ * and nothing is answered about it here.</p>
+ *
+ * @since 0.69.0
+ */
+final class Nesting {
+
+    /**
+     * The locators of the formations, the only objects anything sits in.
+     */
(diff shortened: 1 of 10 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (210 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+675](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6654, opened 2026-08-12
Duplicated `++>` test names causing consistency issues across package members


maxonfjvipon:
The members of a package are about to become attributes of one formation (#6616, step 3), where two attributes cannot share a name. Today 26 test names are used by more than one member of the same package, across 67 test attributes: 18 of them in `string`, 4 in `number`, 2 in `bytes` and 2 in `tuple`. The `bytes` package alone repeats `stops-on-bytes-of-wrong-size` in nine members, one per conversion, and the three padding members repeat almost their whole validation suite:

```
string/padded-left.eo:100   rejects-a-negative-width  ->  padded-left "y" -1 "x"
string/padded-right.eo:99   rejects-a-negative-width  ->  padded-right "y" -1 "x"
string/padded-zeros.eo:96   rejects-a-negative-width  ->  padded-zeros "y" -1
```

None of the 26 are redundant. Every one of them asserts the same contract against a different object, so each has to keep its coverage and lose its ambiguity: `rejects-a-negative-width-when-padding-left`, `stops-on-bytes-of-wrong-size-when-converting-to-i8`, `can-behave-as-an-odd-function-when-taking-arc-sine`. The names are worth fixing for their own sake too, since a report that lists three methods called `rejects_a_negative_width` says nothing about which object rejected it.

What is genuinely duplicated sits elsewhere: a main object smoke-tests behaviour that its member already covers in full. `string.eo` asserts `"eEo".is-alpha`, which is character for character what `string/is-alpha.eo` asserts in `can-be-alphabetic-when-simple-text`; `number.eo` asserts `7.mod 3` is `1`, which is `number/mod.eo`'s `can-take-a-modulo-of-seven-by-three`; and `number.eo` reaches for `sqrt`, `abs` and `power` while each of those members carries between six and fifty-two tests of its own. Sixteen such tests in `number.eo`, `string.eo` and `tuple.eo` should be dropped from the main object and left where they belong.

The three tests named `*-through-the-package` are a separate matter: they assert the explicit `number.power 2 5` form, which #6657 removes along with the form itself.

maxonfjvipon:
Done in #6756. Every package is free of duplicate test names now, and MjMerge refuses one the moment a package with a duplicate is merged.

0crat:
[...]

PR #6755, opened 2026-08-13
Lift the tests of a package member to the top of the object

maxonfjvipon:
A test is legal only as a direct child of the top-level object of a file. The parser says so outright — `test attribute legal only as direct child of top-level object` — and the transpiler agrees by construction, since it assembles the test class from the direct attributes of the class and nothing below them. Until #6656 the two could not disagree, because a package member was the top-level object of its own file. `MjMerge` makes the member an attribute of the object its package names, and a test spliced in along with it lands one level too deep, where nothing looks for it:

```java
formation.appendChild(top);
for (final Node test : Merging.tests(top)) {
    Merging.claimed(taken, Merging.named(test), member.getKey(), pkg);
    formation.appendChild(top.removeChild(test));
}
```

So the member goes in, and its tests come back out and sit beside the tests the object declares itself. Merged XMIR stays a shape the parser would have accepted, and the transpiler keeps knowing nothing about packages — which is right, since the merge is the only component that knows members exist.

Package members in `eo-runtime/src/main/eo` declare 944 test attributes, and each package's share was set to disappear the moment that package was named in `mergedPackages`, with nothing failing to say so. The `bytes` smoke test in #6743 reported 44 cases in `EObytesTest`, and 44 is what `bytes.eo` declares by itself — all seventeen members' tests were already gone and I read the number as normal.

Lifting them also puts the collision where it belongs. Two members claiming one test name now meet the same refusal a member claiming a taken attribute name meets, naming the test and the member it came from, instead of reaching javac as two methods of one generated class. That makes the last duplicate in #6654 — `stops-on-a-fractional-index`, in three members of `tuple` — a hard blocker for merging `tuple` rather than something a human has to remember to scan for.

Nothing needs to walk deeper than one level. Packages merge deepest first, so `number.i64.plus` gives its tests to `i64` before `i64` gives its own to `number`, and a test written three files down arrives at the top one level at a time. A lifted test keeps the `loc` it was parsed with, `Φ.number.i64.plus.+can-…`, which is what names it in `PhDefault` labels and failure messages — worth keeping, since that is the only thing left in the merged XMIR that says which member the test came from.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Merging.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Merging.java
@@ -19,15 +19,16 @@ import java.util.Optional;
 import java.util.TreeMap;
 import java.util.stream.Collectors;
 import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
 
 /**
  * Put the members of a package inside the object that the package names.
  *
  * <p>The parsed XMIR of the object and of its members are all on disk by the
- * time this runs, and the splice is a matter of moving one element: the
+ * time this runs, and the splice is a matter of moving elements: the
  * top-level {@code <o>} of {@code number/lt.xmir} becomes a child of the
- * top-level {@code <o>} of {@code number.xmir}. Nothing inside the moved tree
- * is touched, because the parser leaves every name fully qualified and every
+ * top-level {@code <o>} of {@code number.xmir}. No name inside the moved tree
+ * is rewritten, because the parser leaves every name fully qualified and every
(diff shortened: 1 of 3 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (225 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+687](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6758, opened 2026-08-13
Shorten three MjSafe options so PMD needs no suppression

Thayorns:
Three `@Parameter` fields in `MjSafe` carried `@SuppressWarnings("PMD.LongVariable")` only because their names doubled as the names users write in the plugin's XML. That is no longer a reason to keep them long: `trackTransformationSteps`, `skipExperimentalLints` and `ignoreVersionConflicts` are now `trackSteps`, `skipExperimental` and `ignoreConflicts`, and all three suppressions are gone.

Nothing changes for anyone configuring the plugin. Each field keeps its `eo.*` user property untouched, so `-Deo.trackTransformationSteps=true` and friends work exactly as before, and each one gains an `alias` with its old name, so an existing `<configuration><ignoreVersionConflicts>` element still binds. Maven resolves aliases before it injects, both at plugin level and inside an `<execution>`.

`MjTranspileIT.acceptsOldNameOfTrackingOption` is the guard for that: it runs a real Maven build that configures the plugin with the old `trackTransformationSteps` element and then checks that `target/eo/5-pre-transpile` was written, which only happens when the value actually reached `trackSteps`.

Two of the five suppressions listed in #6459 are still there and I left them alone. `PMD.TooManyFields` on the class cannot go until the options are grouped into objects, which the issue itself says is the real work; `PMD.ImmutableField` on `objectionary` is held in place by `FakeMaven` setting that field reflectively, and every way I tried to make it `final` either broke the fake or traded the PMD suppression for a Checkstyle one. Both are worth their own change, so the issue should stay open.

Related to #6459.

### DIFF
--- a/eo-integration-tests/src/test/java/org/eolang/maven/MjTranspileIT.java
+++ b/eo-integration-tests/src/test/java/org/eolang/maven/MjTranspileIT.java
@@ -99,12 +99,12 @@ final class MjTranspileIT {
                     .set("trackTransformationSteps", "true")
                     .set("cacheEnabled", "false");
                 f.exec("process-sources");
-                MatcherAssert.assertThat(
-                    "the plugin must still understand the old name of the tracking option",
-                    temp.resolve("target/eo/5-pre-transpile").toFile().exists(),
-                    Matchers.is(true)
-                );
             }
+        );
+        MatcherAssert.assertThat(
+            "the plugin must still understand the old name of the tracking option",
+            temp.resolve("target/eo/5-pre-transpile").toFile().exists(),
+            Matchers.is(true)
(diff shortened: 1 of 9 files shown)


Thayorns:
@yegor256 CI green here, have a look please

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1376](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6759, opened 2026-08-13
Read the inference tables back for mistakes

### LINKED ISSUE #6757: `eo-inference` module doesn't check its outputs for mistakes
The `eo-inference` module writes three tables and stops there. Nobody reads them, so the module has no output a person can act on. `Clues` says why, at eo-inference/src/main/java/org/eolang/inference/Clues.java:38:

```java
 * <p>No clue decides anything, which is why they can be read one after
 * another in any order, and why nothing here says whether a program is
 * wrong. Judging used to live beside these rules and was taken out again
 * in #6661: it reported nothing, because a verdict needs the object that
 * misses an attribute to have been seen whole, and almost none of them have
 * been. It comes back when every object can be given a type, rather than
 * only those a call site happens to reach.</p>
```

That was an honest description of a build a day old, where 5,037 of 10,642 dispatches had a type and 48 of 2,592 objects had been seen whole. Four changes have landed since: #6684 works out what a dispatch turns out to be, #6738 reads what an atom says it comes back with, #6741 lets a void be a type, and #6636 writes down what every object sits in. The condition the checker was parked under no longer holds.

Applying the verdict rule offline to a clean `eo-runtime` build — report when the receiver of a dispatch has been seen whole and has no such attribute — gives **zero** reports. Seeding misspelled dispatch names into sites the rule can judge, it catches **eight of eight**:

```
malloc.xmir:133  zzz02 asked of Φ.malloc
tuple.xmir:153   zzz06 asked of Φ.tuple
each.xmir:59     zzz07 asked of Φ.tuple
```

Before #6636 the same rule on the same tree reported ten mistakes and every one of the ten was `ρ`, the object something sits in, which nothing had written down. That is the noise that had to go before a verdict was worth printing. Of 10,404 dispatches, 10,068 now have a type and 1,028 have a receiver the rule can judge.

The loop that was taken out does not have to come back with it. `Checks` filed a promise for every argument put into a copy, and `Worklist` drained that queue, splitting each promise into smaller ones — the work of pushing a type into a context, which nothing else could do at the time. The tables now answer without being pushed: `Needs` has a row for every dispatch and its receiver, `Links` closed by `Resolved` says what that receiver is, and `Provides` says what it holds. So a verdict is a join of the three, and `checks.xml` is not needed for it.

The mistakes belong in a `problems.xml` beside the other tables, written even when it is empty, since "nothing was found" is an answer and a missing file is not:

```xml
<problems>
  <type id="Φ.app.t.next">
    <attr name="foo" asked="Φ.app.inc.φ"/>
  </type>
</problems>
```

Two things keep it quiet where it must be quiet. A type the table has no row for is not complete, which is what stops a locator rooted at a void from ever being judged — nothing is known about one. And a name is looked for in the type, in its package, and behind its `φ` before anything is said, since an attribute nobody binds falls through to the object of that name beside it.

Nothing fails the build. The checker is a prototype and its verdicts are worth reading before they are worth obeying. This is one of the six diagnostics the Python prototype emits; the other five need arity and argument-fits-void checks that are not started here.

Behaviour is described by YAML packs in eo-maven-plugin/src/test/resources/org/eolang/maven/inference-packs, one pack per behaviour, a pack able to describe a program of several files.

maxonfjvipon:
Three tables were written and nobody read them. `Concluded` follows the clues and then reads what they wrote, keeping in `problems.xml` every name a program takes from an object that will never have it.

The rule is one pass over the rows. `Needs` has a row for every dispatch and its receiver, `Links` closed by `Resolved` says what that receiver is, `Provides` says what it holds, so a mistake is those three read together:

```java
if (this.given.complete(asked.getKey())) {
    for (final Map<String, String> need : asked.getValue()) {
        this.judge(asked.getKey(), need, rows);
    }
}
```

Both halves of it are guards. "Seen whole" leaves out an atom, an object nobody describes, and every name rooted at a void — the table has no row for a void, so it is never whole, which is what keeps this quiet about a runtime whose core objects are voids. "There is no such name" means none in the type, none in its package, and none behind what it delegates to, which `Provided` walks before answering.

Measured on a clean `eo-runtime` build: **zero** mistakes reported, and `links.xml`, `needs.xml` and `provides.xml` come out byte-identical to the run before this change, so the only thing that moved is the new document. Seeding misspelled dispatch names into sites the rule can judge — 1,010 of them, of 10,404 dispatches — it catches **eight of eight**:

```
zzz02 asked of Φ.malloc  at Φ.malloc.+can-copy-data-from-start-to-end.φ
zzz06 asked of Φ.tuple   at Φ.tuple.+can-build-an-empty-tuple-with-an-indented-keyword.array.ρ.ρ
```

The worklist does not come back with it. `Checks` filed a promise for every argument put into a copy and `Worklist` drained the queue, splitting each promise into smaller ones, which was the only way to push a type into a body at the time. What that bought is the mistake that is one for a single caller, and that is exactly what this does not report: in the program of the design note, `x.next.foo` is written against a void and is right for any argument that has a `foo`. `note-example.yaml` now says so and asserts the silence. Everything reported holds for every caller of the code it is about.

Eight packs describe the behaviour, among them the one that would have been ten false positives a day ago:

```yaml
eo:
  app.eo: |
    [] > app
      [] > label
      [] > outer
        [] > inner
          label > @
problems:
  - "/problems[not(type)]"
```

Nothing fails the build. The goal logs how many mistakes it found and leaves it at that: this is a prototype, and its verdicts are worth reading before they are worth obeying.

Closes #6757

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Clues.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Clues.java
@@ -33,11 +33,11 @@ import java.util.Collection;
  *
  * <p>No clue decides anything, which is why they can be read one after
  * another in any order, and why nothing here says whether a program is
- * wrong. Judging used to live beside these rules and was taken out again
- * in #6661: it reported nothing, because a verdict needs the object that
- * misses an attribute to have been seen whole, and almost none of them have
- * been. It comes back when every object can be given a type, rather than
- * only those a call site happens to reach.</p>
+ * wrong. Judging it is {@link Concluded}, which reads the tables back once
+ * they are written and puts what it is sure of into a table of its own. A
+ * rule that decided while it read would have to be handed answers no rule
+ * has found yet, and there would be nowhere to put a second reading of the
+ * same rows.</p>
  *
  * @since 0.67.0
  */
--- a/eo-inference/src/main/java/org/eolang/inference/Concluded.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Concluded.java
@@ -1,0 +1,86 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.xml.XML;
+import com.jcabi.xml.XMLDocument;
+import com.yegor256.tojos.MnMemory;
+import com.yegor256.tojos.TjDeferred;
+import com.yegor256.tojos.Tojos;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
+
+/**
+ * The tables, read together for mistakes.
+ *
+ * <p>A clue writes down what it understands and decides nothing, which is what
+ * keeps the rules simple and apart from one another. This is the other half:
+ * the tables they wrote are read back and what is certainly wrong lands in a
+ * table of its own beside them. Nothing but the documents passes between the
+ * two halves, so a rule can change how it fills a table, or a table can gain a
+ * column, without this being touched.</p>
+ *
+ * <p>The mistakes are written the way every other table is written, a type
+ * with the names it is asked for and will never have:</p>
+ *
+ * <pre> &lt;problems&gt;
+ *   &lt;type id="Φ.t"&gt;
+ *     &lt;attr name="extra" asked="Φ.app.φ"/&gt;
+ *   &lt;/type&gt;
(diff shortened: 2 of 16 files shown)


maxonfjvipon:
Closing unmerged. The collecting half of the module comes first: the tables cannot yet record a union, a Δ payload, a ⊥, or what fills a void at a call site, and 77% of the dispatches of eo-runtime dead-end at a void because of it. A verdict rests on `complete`, which is a syntactic flag today and will mean something else once bindings are recorded, so this would be rewritten rather than extended. The three collecting tables come out byte-identical with or without it, so nothing is lost by dropping it. The branch stays on the fork if the rule is wanted back.

PR #6767, opened 2026-08-13
#6632: give a bottom its reason at every birth site

### LINKED ISSUE #6737: PhTerminator.java lacks explanation for bottom parameter functionality
The puzzle `6632-8b991fd3` from #6632 has to be resolved: 

https://github.com/objectionary/eo/blob/c5bb8ace0f8876f48268d9054b0accbc09839dd8/eo-runtime/src/main/java/org/eolang/PhTerminator.java#L34-L38

The puzzle was created by Yegor Bugayenko on 13-Aug-26. 

morphqdd:
Resolves the `@todo` puzzle from #6632: `AtVoid.get()` and `PhDefault.absent()` both handed out a bottom created with `new PhTerminator()`, which carries no cause, even though each of them knows exactly what went missing (the attribute's own name, or the name that wasn't found).

`AtVoid.get()` now returns `PhTerminator.withCause("the attribute \"%s\" is not set")` for an unset attribute. `PhDefault.absent()` now returns `PhTerminator.withCause("the attribute \"%s\" is not found in %s")` for a name nobody knows. In both cases forcing the bottom now explains what was missing, instead of falling back to the generic "terminated computation" message or, when applied without a dispatch in between as in `obj.missing a b`, swallowing `a` as the cause.

Added `namesTheUnsetAttributeInTheTerminatorsCause` to `AtVoidTest`. Didn't add a matching test to `PhDefaultTest` since that file is already at the 1000-line `qulice` file-length cap; the existing `PhDefaultTest` suite (56 tests) still passes unchanged, confirming no regression in `absent()`'s other behavior.

Ran `AtVoidTest`, `PhTerminatorTest`, `PhDefaultTest` and `qulice:check -Pqulice`, all green.

Closes #6737

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -402,7 +402,7 @@ public class PhDefault implements Phi, Cloneable {
         } else if (this.loaded().containsKey(Phi.PHI)) {
             object = this.take(Phi.PHI).take(name);
         } else {
-            object = PhTerminator.withDefaultCause(
+            object = new PhTerminator(
                 String.format("the attribute \"%s\" is not found in %s", name, this.forma())
             );
         }
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -402,7 +402,9 @@ public class PhDefault implements Phi, Cloneable {
         } else if (this.loaded().containsKey(Phi.PHI)) {
             object = this.take(Phi.PHI).take(name);
         } else {
-            object = new PhTerminator();
+            object = PhTerminator.withCause(
+                String.format("the attribute \"%s\" is not found in %s", name, this.forma())
+            );
         }
         return object;
     }
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -402,7 +402,7 @@ public class PhDefault implements Phi, Cloneable {
         } else if (this.loaded().containsKey(Phi.PHI)) {
             object = this.take(Phi.PHI).take(name);
         } else {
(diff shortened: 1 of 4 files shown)


morphqdd:
Fixed: my earlier change pre-filled a bottom's cause via `AtVoid.get()`/`PhDefault.absent()`, which blocked `bytes.slice` and `chunk.read` from putting their own, more specific fallback reason afterward (`PhTerminator.put` is write-once). Added `PhTerminator.withDefaultCause`, a birth-site default that a later explicit `put` still overrides. `EObytesEOsliceTest` and `EOchunkEOreadTest` pass again locally.

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -402,7 +402,9 @@ private Phi absent(final String name) {
         } else if (this.loaded().containsKey(Phi.PHI)) {
             object = this.take(Phi.PHI).take(name);
         } else {
-            object = new PhTerminator();
+            object = PhTerminator.withDefaultCause(
```
yegor256: @morphqdd why not `new PhTerminator(String.format(...))`? why static factory method? they are evil.

morphqdd:
@yegor256, replaced the static factory with a constructor as you suggested. CI is green, please take another look when you have a moment.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4408](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2677](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6768, opened 2026-08-13
#6734: normalize the mantissa when rounding carries it to ten

### LINKED ISSUE #6734: `string.as-scientific` prints incorrect mantissa and exponent when rounding to ten
## What happens

`string.as-scientific` computes its power-of-ten exponent from the *unrounded* magnitude of `n`, then independently rounds the mantissa to six fraction digits with `as-fixed`. When the true mantissa is within `0.0000005` of `10`, `as-fixed`'s rounding correctly carries it into the integer part, but the exponent that was already computed is never adjusted, so the output shows a two-digit, unnormalized mantissa such as `10.000000e0` instead of `1.000000e1`.

| --- | --- | --- |
| `string (as-scientific 9.9999996)` | `10.000000e0` | `1.000000e1` |
| `string (as-scientific 9.9999996e19)` | `10.000000e19` | `1.000000e20` |

The second row falls exactly in the "magnitude past the range of a signed 64-bit integer" case that the file's own docblock names as the reason `as-scientific` exists, since `as-decimal` cannot handle it, so this bug affects the function's primary intended use, not just an obscure corner.

Traced against IEEE-754 double arithmetic on current `master` (`10a8afce25`):

```
n = 9.9999996
rec-exponent(n.abs, 0): 9.9999996 < 10 immediately -> exponent = 0
mantissa = n.div(10.power(0)) = 9.9999996
as-fixed(9.9999996, 6): floor(9.9999996 * 1_000_000 + 0.5) = 10_000_000
                        -> ip = 10, fraction = 0 -> "10.000000"
result = "10.000000" + "e" + "0" = "10.000000e0"
```

This matches `as-fixed`'s own documented behavior: its `can-carry-rounding-into-the-integer-part` test asserts `as-fixed 0.9999999 6` is `"1.000000"`, i.e. rounding is expected to carry into the integer part. `as-scientific` never re-checks the exponent after that carry happens.

## Why it happens

[`rec-exponent`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/as-scientific.eo#L37-L43) counts how many times the raw `n.abs` must be divided by 10 before it drops under 10 — for `9.9999996` that is zero divisions, so `exponent` is fixed at `0` before any rounding takes place. The mantissa is formatted separately at [line 26](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/as-scientific.eo#L26) by `as-fixed` with 6 places, which rounds `9.9999996` up to `10.000000`. `rec-exponent` has no way to learn about that carry, so the exponent stays at `0` and the printed mantissa is left with two digits before the decimal point instead of one.

## Expected behavior

The mantissa printed by `as-scientific` should always stay within `[1, 10)`, matching the normalized form shown by its own tests (e.g. `can-name-a-magnitude-past-the-long-range` expects `"1.000000e19"`). When rounding to six places would carry the mantissa to `10`, the exponent should be incremented by one and the mantissa reset to `1.000000`.

## Suggested fix

After computing the rounded mantissa (or by checking whether `n.div(10.power(exponent))` rounds to exactly `10` at six-place precision), bump `exponent` by one and use `1.000000` as the mantissa in that case. A regression test covering a small magnitude (`9.9999996`) and one past the `long` range (`9.9999996e19`) would guard both the common and the documented primary-use cases.

morphqdd:
`as-scientific` computed its power-of-ten exponent from `n`'s unrounded magnitude, then independently rounded the mantissa to six places with `as-fixed`. When the true mantissa was within `0.0000005` of `10`, `as-fixed` correctly carried it into the integer part, but the already-computed exponent was never adjusted, so the result showed an unnormalized two-digit mantissa like `10.000000e0` instead of `1.000000e1`.

Added a check comparing the fixed-point mantissa against `"10.000000"`. When it carries, the mantissa is reset to `1.000000` and the exponent used for the printed power is bumped by one; otherwise the original mantissa and exponent are used unchanged.

Added `can-carry-a-rounded-mantissa-into-the-next-power-of-ten` (`9.9999996`) and `can-carry-a-rounded-mantissa-past-the-long-range` (`9.9999996e19`, the magnitude range this object exists for).

Ran `EOas_scientificTest` and `qulice:check -Pqulice`, both green.

Closes #6734

### DIFF
--- a/eo-runtime/src/main/eo/number/as-scientific.eo
+++ b/eo-runtime/src/main/eo/number/as-scientific.eo
@@ -12,6 +12,7 @@
 +version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
++unlint redundant-object
 
 [n] > as-scientific
   n.is-finite.not.if > @
@@ -21,19 +22,25 @@
         n.lt 0
         "-infinity".as-bytes
         "infinity".as-bytes
-    concat.
+    if.
+      mantissa.eq "10.000000".as-bytes
       concat.
-        as-fixed
-          n.div
-            10.power
-              number exponent
-          6
-        "e".as-bytes
-      as-decimal
-        number exponent
+        "1.000000".as-bytes.concat "e".as-bytes
+        as-decimal
+          number bumped
+      concat.
+        mantissa.concat "e".as-bytes
+        as-decimal
+          number exponent
+  wrapped.plus 1 > bumped
   rec-exponent > exponent!
     n.abs
     0
+  as-fixed > mantissa
+    n.div
+      10.power
+        number exponent
+    6
   if. > [m count] >> rec-exponent
     m.lt 10
     count
@@ -41,6 +48,15 @@
       m.div 10
       as-number.
         count.plus 1
+  number exponent > wrapped
+
+  eq. ++> can-carry-a-rounded-mantissa-into-the-next-power-of-ten
+    "1.000000e1"
+    string 9.9999996.as-scientific
+
+  eq. ++> can-carry-a-rounded-mantissa-past-the-long-range
+    "1.000000e20"
+    string 9.9999996e19.as-scientific
 
   eq. ++> can-name-a-magnitude-past-the-long-range
     "1.000000e19"
--- a/eo-runtime/src/main/eo/number/as-scientific.eo
+++ b/eo-runtime/src/main/eo/number/as-scientific.eo
@@ -12,7 +12,6 @@
 +version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
-+unlint redundant-object
 
 [n] > as-scientific
   n.is-finite.not.if > @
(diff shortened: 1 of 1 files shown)


eo-runtime/src/main/eo/number/as-scientific.eo
```diff
@@ -12,6 +12,7 @@
 +version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
++unlint redundant-object
```
yegor256: @morphqdd don't do this, fix the code

morphqdd:

yegor256:
@morphqdd see my comment above: don't use `+unlint`, fix the code instead

morphqdd:
@yegor256, fixed the code instead of suppressing the lint, as you asked. CI is green, please take another look when you have a moment.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (33) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4420](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2684](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6659, opened 2026-08-12
Remove `PhNest` and the package branch of `PhDefault.absent` after merging every package


maxonfjvipon:
The last step of #6616, once every package is merged and no member expects its receiver in `α0` any more.

Delete the two branches that make `take` guess. `PhDefault.absent` at `eo-runtime/src/main/java/org/eolang/PhDefault.java:389` keeps its descent into the λ of an atom and into the decoratee, and loses `extended`, `extension`, `path`, `SPECIAL` and `SCOPE`:

```java
private Phi absent(final String name) {
    final Phi object;
    if (this.extended(name)) {
        object = this.extension(name);
    } else if (this instanceof Atom) {
```

The ordering rule its docblock defends goes with them, because it stops being a rule: a merged `as-number` is a declared attribute of `string`, so `"42.5".as-number` is answered out of `loaded()` before the descent into `bytes` is ever considered, and the memoized value that `loaded()` hands back is the same one every access gets. That closes the second and third cost in the description, and the first one along with them, since an absent-name dispatch no longer joins strings and probes the classpath.

`PhNest` goes too. It exists to let one name stand for a package and an object at once, and after the merge no package holds anything to stand for, so `PhPackage.loadPhi` can load the object class and be done. `PhPackage.dispatched`, which existed to keep a `PhNest` uncopied, goes with it. So does `OnClasspath`: `PhDefault.extended` and `PhNest.owns` are its only two callers, which also settles #6500 by removing the class that test covers.

Delete `PhNestTest` and `OnClasspathTest`, trim the package-extension cases out of `PhDefaultTest` and `PhPackageTest`, drop the `mergedPackages` parameter so the merge is unconditional, and correct the prose that still describes the old rule: the `PhDefault.absent` docblock, `bytes.eo:5`, `output.eo:4-5`, `chunk/to-output.eo:4` and the package-dispatch paragraph in `eo-type-inference-spec.md`.

[...]

PR #6770, opened 2026-08-13
Merge the five single-member packages and read receivers through the rho

maxonfjvipon:
The five packages that hold one member each — `chunk`, `i64`, `output`, `range`, `u64` — and the shape every later package will follow:

```
-[r] > naturals
+[] > naturals
   range > @
     ...
     r.end
+  ^ > r
```

The body is not touched. A member is parsed alone, long before `4-merge` joins it to the object its package names, so the parser cannot see `range` and no bare `start` can reach it: `build-fqns` walks the enclosing formations of the file, finds nothing, and emits `Φ.start`. Binding the rho to the name the void used to have puts the declaration back where the walk can find it, and every reference below resolves to exactly what it resolved to before — `ξ.r.end` and `ξ.ρ.r.start`, byte for byte the same parse as the void produced.

Writing the hops by hand instead works too, and is worse. Both spellings were tried:

| written | resolves to | result |
| --- | --- | --- |
| `start` | `Φ.start` | a global that does not exist |
| `^.start` | `ξ.ρ.start` | `naturals.start`, which falls back into naturals' own φ and spins |
| `^.^.start` | `ξ.ρ.ρ.start` | correct, and one miscount away from the line above |

The middle row is the reason this matters. It does not fail — it hangs, and `eo.deadline` turns the timeout into a skip, so the suite reports `Tests run: 8, Failures: 0, Skipped: 4` and reads as green. One `^` per member and no counting leaves nothing to miscount.

`output/dead` keeps nothing at all, since its `origin` was never read; `i64/div` and `u64/div` keep `x` and `cant-divide` and give up only `n`.

The merged packages are faster, which is what one would hope for from resolving at compile time what `PhDefault.absent` used to probe the classpath for on every dispatch. Same tests, same deadline, before and after:

```
i64 family   58 tests   60.8s -> 41.3s
u64 family   25 tests   22.8s -> 20.1s
```

Three things that only the first real merge could expose, fixed here for every package that follows. The `4-merge` directory was missing from `check-folders-numbering.groovy`, which fails on any directory it does not know, and never fired before because an empty `mergedPackages` writes nothing. And `package-member-without-void/S` reports a member with no voids as unreachable, which is the rule this epic removes; it is silenced per file, only in the files that have no voids left, so it goes on watching the seven packages that have not moved. Retiring it belongs to #6659, where I have said so.

And `ReadmeSnippetsIT` compiles the runtime's `.eo` sources from scratch in a Farea sandbox of its own, where nothing said which packages to merge. A member reading its receiver through `^` has no rho there, and the old runtime dispatch binds the receiver into the first void it no longer has, so `2.as-i64.div 0.as-i64` shifted every argument one place along and set `cant-divide` twice. The sandbox now runs the `merge` goal and reads the list out of `eo-runtime/pom.xml`, so the next package to move needs no second mention there.

### DIFF
--- a/eo-integration-tests/src/test/java/integration/EoSourceRun.java
+++ b/eo-integration-tests/src/test/java/integration/EoSourceRun.java
@@ -4,13 +4,24 @@
  */
 package integration;
 
+import com.jcabi.xml.XMLDocument;
 import com.yegor256.farea.Farea;
 import java.io.IOException;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import org.cactoos.Proc;
 
 /**
  * Execution of EO source.
+ *
+ * <p>The sandbox compiles the {@code .eo} sources of the eo-runtime from
+ * scratch, so it must merge the same packages the runtime's own build merges.
+ * A member that reaches its receiver through {@code ^} has no rho until the
+ * merge gives it one, and unmerged it is applied to the receiver through the
+ * first void it does not have any more, which shifts every argument by one.
+ * The list is read from the pom that declares it, so that a package named for
+ * merging there needs no second mention here.</p>
+ *
  * @since 0.56.3
  */
 final class EoSourceRun implements Proc<Object> {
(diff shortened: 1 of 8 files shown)


yegor256:
@maxonfjvipon maybe `package-member-without-void` lint is not actual anymore and should be deleted? File an issue if you agree.

maxonfjvipon:
Agreed, and filed as objectionary/lints#1273. The rule cannot decide the question it asks any more: it runs at `3-lint`, one stage before `4-merge`, and whether a package is merged is a parameter of the merge mojo that no XMIR it reads carries. Removing the three `+unlint` metas this PR added, and the ones the remaining seven packages will add, belongs to #6659.

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+711](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6778, opened 2026-08-14
fix(#6762): report unpacked dependency count

### LINKED ISSUE #6762: `Resolving` logs the full dependency count instead of how many were actually unpacked
## What happens

`Resolving.exec()` (`eo-maven-plugin/src/main/java/org/eolang/maven/Resolving.java:110-124`) computes how many dependencies were actually unpacked via `new Threaded<>(deps, dep -> this.resolved(dep, this.target)).total()`, then discards that value and logs `deps.size()` instead:

```java
new Threaded<>(
    deps,
    dep -> this.resolved(dep, this.target)
).total();
Logger.info(
    this, "New %d dependenc(ies) unpacked to %[file]s: %s",
    deps.size(), this.target,
    new Joined(", ", new Mapped<>(Dep::toString, deps))
);
```

`resolved()` (and the `unpacked()` it delegates to) is written to return `1` when it actually unpacks a dependency and `0` when the destination directory already has files in it, specifically so this count can distinguish new work from cached work. Because the logged number is `deps.size()` (the full candidate list) instead of the discarded `.total()`, every build logs "New N dependenc(ies) unpacked" for the whole dependency list even when nothing was unpacked because everything was already cached.

## What should happen

Log the value `.total()` returns, and when it is zero, log that nothing new was unpacked instead of claiming all dependencies were.

floze-the-genius:
## Summary

- use the value returned by `Threaded.total()` when reporting unpacked dependencies
- report that there are no new dependencies when every resolved artifact is already cached
- cover the repeated-resolve case with a regression test

## Validation

- `mvn -pl eo-maven-plugin -DskipITs -Deo.coverageTracking=false -Dtest=MjResolveTest test` (13 tests passed)
- `mvn -pl eo-maven-plugin -DskipITs -Deo.coverageTracking=false test` (615 tests passed)
- `mvn -pl eo-maven-plugin -Pqulice -DskipTests -DskipITs -Deo.coverageTracking=false clean verify`

Closes #6762

## AI assistance

OpenAI Codex assisted with implementation and test execution. The resulting diff and validation output were reviewed before submission.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Resolving.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Resolving.java
@@ -109,17 +109,22 @@ final class Resolving implements Step {
     @Override
     public void exec() {
         final Collection<Dep> deps = this.deps();
+        final int unpacked;
         if (deps.isEmpty()) {
-            Logger.info(this, "No new dependencies unpacked");
+            unpacked = 0;
         } else {
-            new Threaded<>(
+            unpacked = new Threaded<>(
                 deps,
                 dep -> this.resolved(dep, this.target)
             ).total();
+        }
+        if (unpacked == 0) {
+            Logger.info(this, "No new dependencies unpacked");
+        } else {
             Logger.info(
                 this,
                 "New %d dependenc(ies) unpacked to %[file]s: %s",
-                deps.size(), this.target,
+                unpacked, this.target,
                 new Joined(", ", new Mapped<>(Dep::toString, deps))
             );
         }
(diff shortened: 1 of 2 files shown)


0crat:
@floze-the-genius Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6780, opened 2026-08-14
Merge the directory, file and input packages

maxonfjvipon:
Three packages, seventeen members, the same shape #6770 established:

```
-[f] > deleted
+[] > deleted
   f.exists.if > @
     ...
     f
+  ^ > f
```

`directory` and `file` cost nothing beyond that line: a repo-wide scan for the explicit `directory.deleted d` and `file.size f x` forms turns up nothing, so every call site already dispatched and the bodies are untouched.

The `input` package is where this stops being mechanical, and the reason is worth stating: `bytes.as-input` does not return an `input`. It returns a bare formation whose only attribute is `read`, with no `@` to decorate, which is exactly why every test in the package spelled `Q.input.tee (as-input ...) ...` rather than dispatching — the member was never reachable *on* that object, in any form. Reaching it through the receiver means saying what the code always meant:

```
-      Q.input.tee
-        as-input 01-02-03-04-05
-        m.to-output
+      tee.
+        input 01-02-03-04-05.as-input
+        m.to-output
```

Making those members return `input`-wrapped results would let the chains compose without the wrapper, and deserves its own ticket rather than a smuggled behaviour change here.

Two things bit, and both pass a build and fail later. Naming the rho `^ > input` inside the `input` package **captures**: a bare `input` in a test resolves to the member's own rho rather than the global `Φ.input`, and the printer helpfully shortens `Q.input.dead` into the captured form. The binding is `^ > origin` now, the name `input/dead.eo` already used for it. And `redundant-object/S` fires on any `^ > name` read exactly once: where the single read sits beside the binding it is inlined outright (`directory/open` now says `* ^`), but where it sits inside a nested formation `^` would resolve to the inner rho, so those two keep the binding and unlint the rule — a false positive `lints` could learn to skip.

`bytes` was migrated here too and has been taken back out, because it breaks the fifth README snippet: the loop prints all four of its iterations and then exhausts a 7.6GB heap. Nothing else notices — the whole eo-runtime suite passes on four platforms with `bytes` merged, all 112 of `EObytesTest` among it, as do `integration` and `benchmark`. Four iterations cannot exhaust a heap through the per-instance attribute cost that #6659 predicts, so this is an unbounded structure on the snippet's path and wants a debugger, not another guess. It is #6781, and `bytes` goes in with it. Three packages left after that: `tuple`, `number`, `string`.

### DIFF
--- a/eo-runtime/pom.xml
+++ b/eo-runtime/pom.xml
@@ -301,7 +301,6 @@
               gets a receiver, and with '^' but no merge it has no rho to read.
               -->
               <mergedPackages>
-                <mergedPackage>bytes</mergedPackage>
                 <mergedPackage>chunk</mergedPackage>
                 <mergedPackage>directory</mergedPackage>
                 <mergedPackage>file</mergedPackage>
--- a/eo-runtime/pom.xml
+++ b/eo-runtime/pom.xml
@@ -301,10 +301,12 @@
               gets a receiver, and with '^' but no merge it has no rho to read.
               -->
               <mergedPackages>
+                <mergedPackage>bytes</mergedPackage>
                 <mergedPackage>chunk</mergedPackage>
                 <mergedPackage>directory</mergedPackage>
                 <mergedPackage>file</mergedPackage>
                 <mergedPackage>i64</mergedPackage>
+                <mergedPackage>input</mergedPackage>
                 <mergedPackage>output</mergedPackage>
                 <mergedPackage>range</mergedPackage>
                 <mergedPackage>u64</mergedPackage>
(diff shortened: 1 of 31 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (183 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+723](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6781, opened 2026-08-14
`bytes` package merge leads to Java heap space exhaustion in fifth README snippet


maxonfjvipon:
Part of #6616. Merging the `bytes` package the way #6770 and #6780 merge the others — every member dropping its receiver void and reading `^` instead — makes the fifth EO snippet in `README.md` exhaust the heap:

```
2 x 2 = 4
3 x 3 = 9
4 x 4 = 16
5 x 5 = 25
java.lang.OutOfMemoryError: Java heap space
Heap dump file created [7661037105 bytes in 15.653 secs]
```

The loop is `x.as-number.lt 6` over a `malloc` chunk starting at 2, so all four iterations run and print correctly, and the program dies after them rather than during. Four iterations cannot exhaust a heap through the per-instance attribute cost that #6659 warns about — `new EObytes()` registering seventeen more `AtOnce(AtComposite(...))` than it does today — so something on that path builds without bound.

Nothing else sees it. With `bytes` merged the whole eo-runtime suite passes on ubuntu, windows and both macos runners, including all 112 tests of the merged `EObytesTest`, and so do `integration` and `benchmark`. Only `ReadmeSnippetsIT` fails, and that test is the one place that compiles eo-runtime's `.eo` sources from scratch in a sandbox of its own while also placing the compiled binaries out of the eo-runtime jar, so the object the snippet dispatches on may not be the object the suite exercises.

`bytes` is also the first *hot* object this epic has reached: an instance is allocated for every arithmetic operation, where `chunk`, `range`, `directory` and `file` are allocated once in a while. `number` and `string` are the same kind of object and are still ahead of us, so whatever this turns out to be is worth understanding before either of them moves.

The migration itself is written and was reverted out of #6780 in `9e63abca22`, so it can be recovered from `b2f5a0546a` rather than done again. Reproduce with `mvn clean install -Psnippets` after adding `bytes` back to `mergedPackages` in `eo-runtime/pom.xml`.

maxonfjvipon:
The symptom is gone. `bytes` merged in #6926 and `snippets` has been green since, including on #6986, which is the check that would catch this — `ReadmeSnippetsIT` asserts `BUILD SUCCESS`, and an `OutOfMemoryError` denies it.

Worth being straight about what that does and does not mean: the unbounded structure was never diagnosed, so this is a symptom that stopped reproducing rather than a cause that was found. Something between the first report and the merged `bytes` fixed it. Reopen if it comes back.

[...]

PR #6782, opened 2026-08-14
Merge the tuple package

maxonfjvipon:
Twenty-two members, the same shape the last three pull requests established:

```
-[list func] > mapped
+[func] > mapped
   mappedi > @
-    list
+    ^
     func item > [item idx] >>
```

Every one of the twenty-two read its receiver exactly once, so none of them keeps a `^ > list` binding at all — `redundant-object/S` asks for the rho where it is used, and there it reads better anyway. `maximum` and `minimum` are the two that read theirs twice and keep the named binding.

The explicit `tuple.member list args` form was in eight places and all of them now dispatch: `tuple.mapped` in `set.eo`, `tuple.reduced` in `path.eo`, `tuple.is-empty` in `range.eo`, and the `+alias tuple.reduced` / `tuple.mapped` / `tuple.filtered` / `tuple.is-empty` that let `map.eo`, `string/chained.eo`, `string/contains-all.eo`, `string/contains-any.eo`, `string/english.eo` and `range/naturals.eo` write the bare name. The aliases are gone with them.

`tuple.eo` also loses `can-check-containment-through-the-package`, which asserted the form this removes. #6654 said it should go rather than be rewritten when its package moved, and this is that moment.

Locally the pipeline puts 39 members into 9 package objects and `EOtupleTest` comes out with 119 tests. This is the last package that #6781 does not block: `bytes` waits for it, and `number` and `string` are the same kind of object, allocated per operation rather than once in a while.

### DIFF
--- a/eo-runtime/pom.xml
+++ b/eo-runtime/pom.xml
@@ -308,6 +308,7 @@
                 <mergedPackage>input</mergedPackage>
                 <mergedPackage>output</mergedPackage>
                 <mergedPackage>range</mergedPackage>
+                <mergedPackage>tuple</mergedPackage>
                 <mergedPackage>u64</mergedPackage>
               </mergedPackages>
               <keepBinaries>
--- a/eo-runtime/src/main/eo/map.eo
+++ b/eo-runtime/src/main/eo/map.eo
@@ -9,9 +9,7 @@
 # hash code is derived from those bytes rather than from the key itself,
 # so that keys with side effects are never evaluated more than once.
 
-+alias tuple.filtered
 +alias bytes.hash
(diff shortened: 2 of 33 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (185 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+735](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6787, opened 2026-08-14
`eo-inference` does not retain binding information for void fillers in applications


maxonfjvipon:
The tables record what fills a void nowhere, and that single absence is what stops most of the runtime from being typed.

`Filled` works it out already. Given a receiver, it walks the chain of copies the receiver resolves through, and wherever an application binds arguments it pairs each one with the void it lands in:

```java
final String hollow = this.owned.slot(this.end(copied), place);
if (!hollow.isEmpty()) {
    found.putIfAbsent(hollow, this.end(given.get(place)));
}
```

That map is used to answer one question and then dropped. Nothing in `links.xml`, `provides.xml` or `needs.xml` says that the `x` of an `inc` held a `t` at `Φ.app.φ`, so every reader has to rediscover it, and a reader that starts from the void — which is what typing a body means — has nothing at all.

The cost, on a clean `eo-runtime` build of master. Of 10,414 dispatches, the object a name is taken from is:

```
 1028   9.9%  seen whole
 6049  58.1%  described but not whole
 3091  29.7%  rooted at a void
  246   2.4%  not described at all
```

77% of the dispatches in the runtime end at a void, because the core objects are voids — `number.eo` is `[as-bytes] > number` and `bytes.eo` is `[@] > bytes` — and nothing in the tables says what those voids ever hold.

`eo-type-inference-spec.md` §5.2 already settles what the answer looks like: an application preserves identity, and the identity "carries binding state: which voids of `P` are still free on this value". So the row for an application says what it is a copy of and what it put where:

```xml
<type id="Φ.app.held">
  <ref loc="Φ.inc">
    <bind void="Φ.inc.x"><ref loc="Φ.app.held.α0"/></bind>
  </ref>
</type>
```

What fills the void is named as the argument itself, not as what the argument resolves to. The two differ exactly when the argument is an application of its own: writing `Φ.dec` for the `inc (dec u)` case would throw away that the `x` of that `dec` holds a `u`, and the argument has a row of its own that says both.

Two things this deliberately does not do. It does not write a row per context: there is one row per locator, and a reader composes the bindings along a chain, which is what keeps this cheap — the Python prototype types 11,857 objects through 31,138 (locator, context) pairs, and every one of those is derived from the same bindings, not stored. And it does not decide anything: a void that nobody fills stays a void, and a name rooted at one stays a name.

[...]

PR #6788, opened 2026-08-14
Let a links row hold a type, not a copy

### LINKED ISSUE #6786: Eo-inference links row limitation: Only allows object copies, not object definitions
A row of the links table holds one attribute and therefore one answer:

```java
rows.add(from).set("copy", target);
```

That was enough while the only thing an object could turn out to be was a copy of another object. It is not enough for the four other things an object turns out to be, and the runtime is full of all four.

Take `Φ.directory.walk.a🌵159-4.φ.α1.α1.α1.α1`. Our row says it is the `token` of something:

```xml
<type id="Φ.directory.walk.a🌵159-4.φ.α1.α1.α1.α1"
      copy="Φ.directory.walk.a🌵159-4.token"/>
```

which is a name for it, not an answer about it. What it actually is, is one of five things — a formation, a string, a joined string, a termination, or a shrug — and there is nowhere in the row to put that. On a clean `eo-runtime` build of master, 1,614 objects are Δ literals and get no row at all, 65 are `⊥`, and 295 are a choice between two or more objects. None of the three can be written down.

The type language settled in `eo-type-inference-spec.md` §5.1 has eight forms, and the output contract in §11.1 gives each of them one element:

```xml
<type loc="Φ.directory.walk.a🌵159-4.φ.α1.α1.α1.α1">
  <union>
    <bottom/>
    <ref loc="Φ.directory.walk.a🌵131-4"/>
    <ref loc="Φ.string"/>
    <ref loc="Φ.string.joined"/>
    <unknown/>
  </union>
</type>
```

The alternative — one attribute per fact, `bottom="true"`, `data="true"`, a repeated `<alt copy=…/>` child for a choice — arrives at the same nesting by a different road, with names invented per fact instead of the grammar's, and with no way to write a choice between objects that themselves carry filled voids. Better to make the row structured once, while the four facts are still ahead of us rather than behind.

So: a links row stops carrying `copy` and starts carrying a type, beginning with the only form there is today. `<type id="Φ.app.φ"><ref loc="Φ.inc"/></type>` says exactly what `copy="Φ.inc"` said, so the information in the table does not move and every count taken off it stays the same. What changes is `Links`, `Resolved`, `Pairs`, `Ends` and thirteen packs that assert on `@copy`.

The other forms come after, one at a time, each with the packs that describe it: what fills a void at an application, Δ, `⊥`, a choice, and a name only a caller can fill.

Worth saying plainly what does *not* go into these rows: the answer in a particular context. A row is keyed by a locator and there is one of it. Contexts belong to whoever reads the table and composes the bindings along a chain, which is what `Filled` already does in memory.

maxonfjvipon:
A links row said what an object copies in a cell:

```java
rows.add(from).set("copy", target);
```

One cell holds one answer, and four of the things an object turns out to be are not one answer. On a clean `eo-runtime` build, 1,614 objects are Δ literals, 65 are `⊥`, and 295 are a choice between two or more objects; none of the three can be written into an attribute, and a choice between objects that themselves carry filled voids cannot be written into any number of attributes.

So the row carries an element:

```xml
<type id="Φ.app.φ">
  <ref loc="Φ.app.inc"/>
</type>
```

which says exactly what `copy="Φ.app.inc"` said. `Types` writes the document, `Pairs` reads it back, and `Links` and `Resolved` stop keeping rows in a table of cells for it.

Nothing moves. On the same parse of `eo-runtime`, before and after: `provides.xml` and `needs.xml` are byte-identical, and the links table has the same 31,876 rows carrying the same targets in the same order. Thirteen packs say `ref` where they said `copy`.

The forms this makes room for arrive one at a time, each with the packs that describe it: what fills a void at an application (#6787), Δ, `⊥`, a choice, and a name only a caller can fill.

Closes #6786

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Links.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Links.java
@@ -5,9 +5,6 @@
 package org.eolang.inference;
 
 import com.jcabi.xml.XML;
-import com.yegor256.tojos.MnMemory;
-import com.yegor256.tojos.TjDeferred;
-import com.yegor256.tojos.Tojos;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -15,6 +12,8 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.Map;
 
 /**
  * Which types are copies of which.
@@ -44,6 +43,11 @@ import java.util.HashSet;
  * where the checker later gets smarter, when a copy starts receiving types
  * of its own, and then nothing else has to change.</p>
  *
+ * <p>Which is why a row carries what an object is as an element of its own
+ * rather than as a cell, and what {@link Types} makes of it. Being a copy is
+ * one of the things an object turns out to be, and the others — a datum, a
+ * termination, a choice between several objects — arrive beside it.</p>
+ *
  * <p>A name that resolves to nothing gets no row and no complaint: a
  * missing row makes a later check stay undecided, while a wrong row would
  * make it decide wrongly. On the runtime this happens to 730 references out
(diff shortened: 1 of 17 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (178 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+747](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6789, opened 2026-08-14
Merge the number package

maxonfjvipon:
Continues #6657 with the `number` package: all 29 members now read their
receiver through `^` instead of taking it as the first void, and `number` joins
`<mergedPackages>` so `MjMerge` splices them into the object at compile time.

The interesting half is the sibling calls. Inside a package, a member naming a
sibling by its bare name compiles to `Φ.number.<sibling>`, which after the merge
resolves to the member with the package object as its `ρ` — so the receiver
rides along as one extra argument. Twenty-four of those, all rewritten as
dispatches:

```
[x] > power
  ...
              base.abs.gt 1
              pinf
  ^ > num
```

Those calls fail in a way worth naming, because it is not obvious. A de-voided
member reached through a package object still works by accident: the package
hands its extensions out with `ρ` unbound, so the receiver lands in it as `α0`.
Reached through the merged object instead, `ρ` is already bound, `put` finds no
vacant attribute and falls onto a cached one, and the program dies with
`Can't overwrite the cached attribute`. Which of the two happens depends on
whether `Φ.number` resolves to a `PhNest` or to `EOnumber`, and that in turn
depended on a leftover: `keepBinaries` still carried
`org/eolang/EO_number/package-info.class`, so the jar kept shipping a package
marker for a package that no longer exists, and every downstream project got
the accidental behaviour while eo-runtime's own build got the real one. The
glob is gone.

Three external call sites moved too: `number.eo` reaches its own `as-i64`
through `$`, `directory.eo` seeds its temporary-file counter with
`clock.as-number.random`, and the `decorated` snippet now writes `x.sine`
rather than `sine x` behind an alias, which is exactly the form the merge
removes.

`integral` keeps all four voids. It takes a function, not a number, so it never
had a receiver to reach for — the same shape as `input/dead` and `output/dead`.

Two static checks over `target/eo/1-parse` back this up: across all ten merged
packages nothing under `Φ.<package>.<member>` survives except `integral`,
`input.dead`, `output.dead` and the void-less `tuple.empty`, and every other
member of every merged package reads `^`. `PhNestTest` now probes `string`,
since `number` is not a package object any more.

`bytes` still waits on #6781; `string` is the last package left.

### DIFF
--- a/eo-integration-tests/src/test/resources/snippets/decorated.yaml
+++ b/eo-integration-tests/src/test/resources/snippets/decorated.yaml
@@ -7,7 +7,6 @@ out:
 file: snippets/foo.eo
 args: ["snippets.foo"]
 eo: |
-  +alias number.sine
   +package snippets
 
   [] > foo
@@ -16,4 +15,4 @@ eo: |
       "Sin of %f is %f".printf
         *
           x
-          sine x
+          x.sine
--- a/eo-runtime/pom.xml
+++ b/eo-runtime/pom.xml
@@ -319,6 +319,7 @@
                 <mergedPackage>file</mergedPackage>
                 <mergedPackage>i64</mergedPackage>
                 <mergedPackage>input</mergedPackage>
+                <mergedPackage>number</mergedPackage>
                 <mergedPackage>output</mergedPackage>
                 <mergedPackage>range</mergedPackage>
                 <mergedPackage>tuple</mergedPackage>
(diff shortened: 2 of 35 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (210 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+819](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6790, opened 2026-08-14
Merging a package's members loses their unit tests entirely — none of them are ever compiled or run


morphqdd:
## What happens

After a package's members are folded into one object by `Merging` (introduced in #6657), the members' own unit tests are silently never compiled into Java at all, and never run by any CI job, even though the build stays green.

Verified on clean `upstream/master` (`bc11e1b3a7`), with a fully clean `eo-runtime` build:

- `eo-runtime/target/eo/5-transpile/tuple.xmir` (the XMIR right before Java generation) contains 106 positive test attributes (`name="+can-..."`), including the tests that belong to `tuple/maximum.eo`, `tuple/minimum.eo`, `tuple/shifted.eo` and every other `tuple/*.eo` member now merged into `tuple`.
- The generated `EOtupleTest.java` contains only **26** `@Test` methods — and zero of them correspond to `maximum`/`minimum`/most other members. Only the tests originally written directly inside `tuple.eo` itself survive.
- `grep -c "<class " eo-runtime/target/eo/5-transpile/tuple.xmir` returns **1** — there is exactly one `<class>` wrapper for the whole merged file.

## Why it happens

`eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/classes.xsl` wraps an object in a `<class>` element (making it a compilable, testable Java class) with `xsl:template match="object/o[...]"` — matching only **direct children of `/object`**, i.e. top-level objects. Before merging, every `.eo` file's own object was the sole top-level child of its own XMIR, so this always matched. After `Merging` splices a whole member subtree (with its own test children intact) in as a **descendant** of the receiver object's tree, that member is no longer a direct `/object/o` child, so `classes.xsl` never wraps it in its own `<class>`.

`eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl`'s `xsl:template match="class" mode="tests"` then only iterates `class/attr` — the direct `attr` children of the one `<class>` that did get created (the receiver's own). A merged-in member's test attributes, now several levels deeper in the tree, are never visited by this template at all.

## Impact

Every package that has members merged into it (per #6657, currently at least `tuple`) loses the entire member-level test suite from compilation. The CI build reports success, but the vast majority of that package's own behavioral contract is never verified in any run.

## What should happen

[...]

PR #6791, opened 2026-08-14
#6763: don't let nan silently drop out of tuple.maximum and minimum

### LINKED ISSUE #6763: `tuple.maximum` and `tuple.minimum` incorrectly ignore `nan` elements, resulting in inconsistent behavior
## What happens

`tuple.maximum` (`eo-runtime/src/main/eo/tuple/maximum.eo`) folds the sequence with `ninf` as the seed and keeps the accumulator whenever `item.as-number.gt max` is false. A NaN comparison is always false, so a `nan` element never replaces the accumulator and is silently skipped.

`(* nan).maximum` returns `ninf`, a value that is not even an element of the input sequence, instead of `nan`. `(* 5 nan).maximum` returns `5`, silently ignoring the `nan` element as if it were never there.

`tuple.minimum` (`eo-runtime/src/main/eo/tuple/minimum.eo`) has the same fold shape with `pinf` as the seed and `item.as-number.lt min`, so it has the same problem: `(* nan).minimum` returns `pinf`.

This is inconsistent with the rest of the numeric stdlib in this codebase (`power`, `sqrt`, `mod` all propagate `nan` from their inputs to their result rather than silently dropping it).

Verified by running `(* nan).maximum` through the actual test harness: it dataizes to `ninf`.

## What should happen

A `nan` element should make `maximum`/`minimum` return `nan` (or, at minimum, treat it as a value that participates in the comparison rather than being invisibly skipped), consistent with how the rest of the numeric objects in this codebase handle `nan`.

morphqdd:
`tuple.maximum` folded the sequence with `ninf` as the seed and kept the accumulator whenever `item.as-number.gt max` was false. A NaN comparison is always false, so a `nan` element never replaced the accumulator and was silently skipped: `(* nan).maximum` returned `ninf` (not even an element of the sequence), and `(* 5 nan -3).maximum` returned `5`, ignoring the `nan` entirely. `tuple.minimum` had the same shape with `pinf` and `.lt`.

Added an explicit `is-nan` check before the comparison in both objects: once a `nan` element is seen, it becomes the accumulator, and every later ordinary comparison against a `nan` accumulator is also false, so `nan` correctly poisons the rest of the fold, consistent with how `power`/`sqrt`/`mod` propagate `nan` elsewhere in this codebase.

Added `can-be-nan-when-the-sequence-holds-a-lone-nan` and `can-be-nan-when-the-sequence-holds-a-nan-alongside-numbers` to both objects.

Note on local verification: I found and filed a separate, unrelated structural bug (#6790) where a package's members merged in by `Merging` (per #6657) lose their unit tests from Java transpilation entirely — this affects `tuple`'s member tests generally, including these new ones, until #6790 is fixed. I confirmed the existing 26 `tuple`-level tests still pass unchanged (no regression), verified the new logic compiles and lints clean, and traced the `is-nan`/comparison-poisoning behavior by hand against the runtime semantics of `is-nan`/`gt`/`lt`. CI's `mvn` job builds this with `-Deo.deadline` set high enough to actually run the merged member tests in its own environment, so it should exercise these directly even though my local run currently can't.

Closes #6763

### DIFF
--- a/eo-runtime/src/main/eo/tuple/maximum.eo
+++ b/eo-runtime/src/main/eo/tuple/maximum.eo
@@ -22,11 +22,11 @@
         max
   ^ > sequence
 
-  is-nan. ++> can-be-nan-when-the-sequence-holds-a-lone-nan
+  is-nan. ++> can-be-nan-when-taking-the-maximum-of-a-lone-nan
     maximum.
       * nan
 
-  is-nan. ++> can-be-nan-when-the-sequence-holds-a-nan-alongside-numbers
+  is-nan. ++> can-be-nan-when-taking-the-maximum-of-a-nan-alongside-numbers
     maximum.
       *
         5
--- a/eo-runtime/src/main/eo/tuple/maximum.eo
+++ b/eo-runtime/src/main/eo/tuple/maximum.eo
@@ -16,10 +16,22 @@
     sequence.reduced
       ninf
       if. > [max item]
-        item.as-number.gt max
+        item.as-number.is-nan.or
+          item.as-number.gt max
         item
         max
   ^ > sequence
+
+  is-nan. ++> can-be-nan-when-the-sequence-holds-a-lone-nan
+    maximum.
+      * nan
+
+  is-nan. ++> can-be-nan-when-the-sequence-holds-a-nan-alongside-numbers
+    maximum.
+      *
+        5
+        nan
+        -3
 
   eq. ++> can-recover-from-an-empty-list-when-taking-the-maximum
     "recovered"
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4432](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6794, opened 2026-08-14
#6765: anchor each segment of the locator pattern in XMIR.xsd

### LINKED ISSUE #6765: `locator` pattern in `XMIR.xsd` fails to validate all segments of `@loc` as per R-9.1.6
## What happens

The `locator` simple type in `eo-parser/src/main/resources/XMIR.xsd`:

```xml
<xs:pattern value="Φ(\.(ρ|φ|α[0-9]+|[a-z]\S*))*"/>
```

is meant to validate `@loc` per-segment, per `PARSER_SPEC.md` R-9.1.6, which quotes this exact pattern as the grammar of a locator: `Φ`, then one dot-separated segment per graph level, each segment being `ρ`, `φ`, `αN`, or a name.

`\S` matches any non-whitespace character, including `.`. Once the alternation reaches the `[a-z]\S*` branch, that branch greedily consumes everything remaining in the string, dots included, in a single match of the repeated group. So the pattern does not actually validate segment-by-segment past the first `[a-z]`-starting segment: any string of the form `Φ.<name><anything-non-whitespace>` matches, regardless of what the later "segments" are, as long as they contain no whitespace.

Concrete check against the pattern: `Φ.a.WRONG`, `Φ.a.λ`, and `Φ.a.+t` are all accepted by this pattern even though `WRONG`, `λ`, and `+t` are not any of `ρ`, `φ`, `αN`, or `[a-z]\S*` on their own — only `Φ.WRONG` (a bad first segment) is actually rejected.

## What should happen

The pattern should anchor each segment so later segments are validated too, e.g. by excluding `.` from what a name segment can contain: `Φ(\.(ρ|φ|α[0-9]+|[a-z][^\s.]*))*`. (Separately, real locators emitted by `set-locators.xsl` include an atom marker `λ` and test-attribute names with a `+` prefix as segments, e.g. `Φ.foo.bar.λ` and `Φ.foo.bar.+checks-it`; if those are legal, R-9.1.6 and the schema should say so explicitly rather than relying on the current pattern's over-permissiveness to let them through by accident.)

morphqdd:
The `locator` simple type in `XMIR.xsd` uses the pattern:

```
Φ(\.(ρ|φ|α[0-9]+|[a-z]\S*))*
```

`\S*` is unanchored, so `[a-z]\S*` matches any non-whitespace run including
dots. That means a locator segment can swallow the rest of the path (or
even multiple dot-separated segments) instead of being validated one
segment at a time, so a malformed locator like `Φ.a.WRONG` still passes
because `\S*` just eats through the extra dot.

This diverges from how the file already handles the same problem for
`fqn`, `signature` and `type-anno` (each uses `[^\s.]*` to stop a segment
at the next dot). It also doesn't allow two segment shapes real XMIR
actually produces: the `λ` atom-marker segment (e.g. `Φ.chunk.read.λ`)
and `+`-prefixed test-attribute segments (e.g.
`Φ.set.+can-append-a-new-item.φ.ρ.ρ.α0`), both confirmed by grepping
generated `.xmir` files.

Fix: anchor each segment like the sibling types do, add `λ` as an
explicit alternative, and widen the leading character class to
`[a-z+-]` (matching `attribute-name`'s own idiom):

```
Φ(\.(ρ|φ|λ|α[0-9]+|[a-z+-][^\s.]*))*
```

Verified against every `@loc` produced by a clean `eo-runtime` build,
and added a parameterized test (`StrictXmirTest.validatesLocatorAttribute`)
covering both the previously-accepted-in-error cases and the newly
allowed real-world shapes.

Fixes #6765

### DIFF
--- a/eo-parser/src/main/resources/XMIR.xsd
+++ b/eo-parser/src/main/resources/XMIR.xsd
@@ -116,7 +116,7 @@
       </xs:documentation>
     </xs:annotation>
     <xs:restriction base="xs:string">
-      <xs:pattern value="Φ(\.(ρ|φ|α[0-9]+|[a-z]\S*))*"/>
+      <xs:pattern value="Φ(\.(ρ|φ|λ|α[0-9]+|[a-z+-][^\s.]*))*"/>
     </xs:restriction>
   </xs:simpleType>
   <xs:complexType name="o" mixed="true">
--- a/eo-parser/src/test/java/org/eolang/parser/StrictXmirTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/StrictXmirTest.java
@@ -22,6 +22,8 @@ import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
 import org.xembly.Directives;
 import org.xembly.Xembler;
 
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4460](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6795, opened 2026-08-14
#6766: report canonical error for horizontal formation as argument

### LINKED ISSUE #6766: Parser fails to use canonical error text for horizontal formation as argument issue
## What happens

`PARSER_SPEC.md` R-9.9.1 requires every error condition to use one canonical text, listed in its table. The table entry for "`[x]` as a bare argument (horizontal anonym in arg position)" is `horizontal formation not allowed as argument`.

For an input like:

```
foo [x] 5 > y
```

the parser instead throws a generic `expected value` (`eo-parser/src/main/java/org/eolang/parser/Tokens.java`, thrown from `readGlyph` when the current character doesn't start any recognized value, which is what happens when it hits `[`). This message doesn't name the rule or say what's actually wrong, unlike the canonical text, which explicitly says a horizontal formation isn't legal there.

## What should happen

The parser should recognize a `[` at this position specifically and report the canonical `horizontal formation not allowed as argument`, per R-9.9.1, instead of the generic `expected value`.

morphqdd:
PARSER_SPEC.md R-9.9.1 lists one canonical text for a horizontal formation
appearing where it's not allowed: `horizontal formation not allowed as
argument`. The parser instead threw the generic `expected value`
(`Tokens.readGlyph`, the fallback branch for any character it doesn't
otherwise handle, which is what `[` hits).

Three existing `eo-typos` fixtures were already pinning this exact
generic message for inputs shaped this way (`malloc.for 4 [m]>>`,
`malloc 100 [m]>> > @`, `([] (^ > x)).plus.minus > s`), so the bug was
already covered by tests, just asserting the wrong text.

Fix: `readGlyph` now recognizes `[` specifically and throws the
canonical message. Updated the three fixtures to expect it.

Fixes #6766

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Tokens.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Tokens.java
@@ -900,6 +900,12 @@ final class Tokens {
      * @return Parsed value
      */
     private Value readGlyph(final char first) {
+        if (first == '[') {
+            throw new ParseError(
+                this.span.line(), this.span.indent() + this.cursor,
+                "horizontal formation not allowed as argument"
+            );
+        }
         final Value value;
         if (first == '*') {
             value = this.reserved(Value.Kind.STAR, "*");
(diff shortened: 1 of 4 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4472](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6796, opened 2026-08-14
#6772: keep object-vs-package disambiguation on the isDirectory fallback

### LINKED ISSUE #6772: Failure in `OyIndexed.isDirectory` fallback results in loss of object-vs-package disambiguation when objects index is unreachable
## What happens

`OyIndexed.isDirectory` computes `!this.index.contains(name) && this.delegate.isDirectory(name)`. The `!index.contains(name)` half is what disambiguates a name that is both a program and a package (e.g. `org.eolang.tuple`, which is both `objects/tuple.eo` and the directory `objects/tuple/` containing `objects/tuple/each.eo` etc, confirmed present in the current `objectionary.lst`).

When the index lookup fails (`ScalarWithFallback`'s `Fallback.From` branch, triggered by any exception reading the index: a network hiccup, a 5xx, offline CI), the fallback drops the `!index.contains(name)` half entirely and returns `this.delegate.isDirectory(name)` alone. `OyRemote.isDirectory` only checks whether `https://github.com/objectionary/home/tree/<hash>/objects/<name>` returns 2xx, which says nothing about whether the same name is also a real object.

Verified live: `GET https://github.com/objectionary/home/tree/master/objects/tuple` returns `200`, and `GET https://raw.githubusercontent.com/objectionary/home/master/objects/tuple.eo` also returns `200` — `tuple` genuinely is both. `Pulling.exec` (`eo-maven-plugin/src/main/java/org/eolang/maven/Pulling.java:129`) skips pulling any object for which `isDirectory` returns true, with no separate `contains` check. So on the fallback path, `tuple` (and any other object whose name also names a package) would never be pulled, even though it's a real object referenced by other code.

## Why it happens

`OyIndexed.isDirectory`'s error fallback (`eo-maven-plugin/src/main/java/org/eolang/maven/OyIndexed.java`, `isDirectory` method) only forwards to `this.delegate.isDirectory(name)`, losing the disambiguation the normal path applies.

## What should happen

The fallback should preserve the same disambiguation the normal path uses — e.g. also check `this.delegate.contains(name)` (which does check both the program and directory URLs, per `OyRemote.contains`) and treat a name that is also a real program as not-a-directory, regardless of whether the index itself could be read.

morphqdd:
OyIndexed.isDirectory computes `!this.index.contains(name) && this.delegate.isDirectory(name)`
on the normal path, using the `!index.contains` half to disambiguate a name
that is both a program and a package (e.g. `org.eolang.tuple`, which is
both `objects/tuple.eo` and the directory `objects/tuple/`).

When the index lookup itself fails (ScalarWithFallback's Fallback.From
branch, triggered by any exception reading the index: a network hiccup,
a 5xx, offline CI), the fallback drops that disambiguation and just
returns `this.delegate.isDirectory(name)`. OyRemote.isDirectory only
checks whether the GitHub tree URL for the name returns 2xx, which says
nothing about whether the same name is also a real object. Confirmed
live: both `objects/tuple` (tree) and `objects/tuple.eo` (raw) return
200, so `tuple` genuinely is both. Pulling.exec skips pulling any object
for which isDirectory returns true, with no separate contains check, so
on the fallback path an object like `tuple` would never get pulled.

Fix: also check `!this.delegate.contains(name)` on the fallback path,
matching what the normal path already does with the index.

Fixes #6772

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/OyIndexed.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/OyIndexed.java
@@ -85,7 +85,7 @@ final class OyIndexed implements Objectionary {
                             name,
                             ex
                         );
-                        return this.delegate.isDirectory(name);
+                        return !this.delegate.contains(name) && this.delegate.isDirectory(name);
                     }
                 )
             )
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/OyIndexedTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/OyIndexedTest.java
@@ -81,7 +81,7 @@ final class OyIndexedTest {
             new OyIndexed(
                 new Objectionary.Fake(
                     name -> new InputOf("[] > qwerty"),
-                    name -> true,
+                    name -> false,
                     name -> true
                 ),
                 new ObjectsIndex(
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4484](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6799, opened 2026-08-14
Say what every application puts into a void

maxonfjvipon:
An application is the one place in a program where a void stops being a question, and until now the tables threw that away. `inc t` says the `x` of that `inc` holds a `t`; `Filled` worked it out to answer one dispatch and dropped it. Now the row for the application says it:

```xml
<type id="Φ.app.held">
  <ref loc="Φ.inc">
    <bind void="Φ.inc.x"><ref loc="Φ.app.held.α0"/></bind>
  </ref>
</type>
```

An argument lands in a void by its place, counted through the voids in the order the formation declares them, which is the order `Ungrouped` already keeps. What fills the void is named as the argument itself and not as what the argument turns out to be — the two differ exactly when the argument is an application of its own, and writing `Φ.dec` for `inc (dec u)` would throw away that the `x` of that `dec` holds a `u`. The argument has a row of its own, which says both.

`Bound` works these out, `Types` renders them, and `Resolved` calls it once the pairs have settled, since naming the void an argument lands in means knowing which formation is being copied. `Provided` gains a ctor taking the provides document, which `Dispatched` uses too. Four packs describe the behaviour: the fact itself, the place-to-void order, the nested argument, and a partial application leaving the rest of the voids free.

On a clean `eo-runtime` build: 22,195 binds over 18,483 rows, `provides.xml` and `needs.xml` byte-identical, and all 31,876 existing links rows unchanged in target and order. No coverage number moves, and none should — `Filled` already substituted along receiver chains in memory. What changes is that the fact stops being private to one pass: it is what a reader outside the module needs to compose a context, and what a void's witnessed join will be gathered from.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Bound.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Bound.java
@@ -1,0 +1,99 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * What every application puts into the voids of what it copies.
+ *
+ * <p>An application is the one place in a program where a void stops being a
+ * question: {@code inc t} says that the {@code x} of that {@code inc} holds a
+ * {@code t}. The place of an argument says which void it lands in, counted
+ * through the voids in the order the formation declares them, which is the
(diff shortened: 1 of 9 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (272 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+759](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6800, opened 2026-08-14
#6779: keep the once-wrapper alive after normalized()

### LINKED ISSUE #6779: PhOnce.normalized() drops its once-wrapper, unlike copy()
`PhOnce.normalized()` (`eo-runtime/src/main/java/org/eolang/PhOnce.java:125-127`) returns `this.object.get().normalized()` directly, handing back the raw wrapped object instead of a new `PhOnce`. Its own `copy()` a few lines above (82-87) does this correctly, returning `new PhOnce(() -> this.object.get().copy(), this.term)`.

`PhOnce` exists to fetch and cache its wrapped object exactly once, thread-safely, behind a `ReentrantLock`/`AtomicReference`. Because `normalized()` unwraps instead of re-wrapping, any object obtained by normalizing a `PhOnce` loses that once-caching guarantee for itself and whatever is built on top of it. This is the identical defect already found and fixed in the sibling class `PhLogged` (issue #6368, "`PhLogged` drops its logging wrapper after `copy()` and `normalized()`"): its `normalized()` was changed to `return new PhLogged(this.origin.normalized());`, but the same fix was never applied to `PhOnce`.

Minimal fix: mirror `copy()`'s pattern in `normalized()` — `return new PhOnce(() -> this.object.get().normalized(), this.term);` — so the once-wrapper and its φ-term supplier survive normalization the same way they survive copying.

morphqdd:
PhOnce.normalized() returned `this.object.get().normalized()` directly,
handing back the raw wrapped object instead of a new PhOnce. Its own
copy() a few lines above does this correctly, returning
`new PhOnce(() -> this.object.get().copy(), this.term)`.

PhOnce exists to fetch and cache its wrapped object exactly once,
thread-safely. Because normalized() unwrapped instead of re-wrapping,
any object obtained by normalizing a PhOnce lost that once-caching
guarantee for itself and whatever is built on top of it. This is the
same defect already found and fixed in the sibling class PhLogged
(#6368), whose normalized() was changed to
`return new PhLogged(this.origin.normalized());`, but the same fix
was never applied to PhOnce.

Fix: mirror copy()'s pattern in normalized().

Fixes #6779

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhOnce.java
+++ b/eo-runtime/src/main/java/org/eolang/PhOnce.java
@@ -123,10 +123,14 @@ public class PhOnce implements Phi {
 
     @Override
     public Phi normalized() {
-        return new PhOnce(
-            () -> this.object.get().normalized(),
-            this.term
-        );
+        final Phi result = this.object.get().normalized();
+        final Phi normalized;
+        if (result instanceof PhTerminator) {
+            normalized = result;
+        } else {
+            normalized = new PhOnce(() -> result, this.term);
+        }
+        return normalized;
     }
 
     @Override
--- a/eo-runtime/src/main/java/org/eolang/PhOnce.java
+++ b/eo-runtime/src/main/java/org/eolang/PhOnce.java
@@ -123,7 +123,10 @@ public class PhOnce implements Phi {
 
     @Override
     public Phi normalized() {
-        return this.object.get().normalized();
+        return new PhOnce(
+            () -> this.object.get().normalized(),
+            this.term
+        );
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (27) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4496](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6804, opened 2026-08-14
Say how much of a program was understood

maxonfjvipon:
The `inference` goal now ends by saying how much of the program it made sense of. On `eo-runtime`:

```
[INFO] Inferred the types of 171 XMIR(s), tables are in target/eo/6-inference
[INFO] 43083 objects, 82.0% of them described, depth 48.8%
[DEBUG]     7760  nothing at all
[DEBUG]     8070  a name rooted at a void
[DEBUG]     7358  a formation, voids still free
[DEBUG]    18207  a formation, nothing left free
[DEBUG]     1688  a formation seen whole
```

Depth is the mean rung against the highest one, so a program every object of which was seen whole comes out at a hundred. The ladder is what keeps that number honest and it is printed beside it: writing an empty row for every object would take "described" to a hundred and move the depth not at all.

`Rung` decides where an object stands, `Depth` puts the world on the ladder, `Ladder` is what comes out. Rungs 2 and 3 are only tellable apart since #6799 — knowing that a copy has nothing left free is exactly what the `bind` elements say. A formation gets no special case: it is a copy of itself that has filled none of its own voids, and lands where that puts it.

```java
final String end = this.ends.getOrDefault(locator, locator);
if (this.table.containsKey(end)) {
    found = this.depth(end, this.voids(end) - filled);
} else if (this.rooted(end)) {
    found = 1;
} else {
    found = 0;
}
```

Nothing is written to a document. `problems.xml` and `coverage.xml` were both proposed before and both dropped, and a measurement of ourselves is not a fact about the program, so it belongs in the log rather than in a table beside the facts that are.

One number in the ticket moved: it said depth 49.3%, from a script that gave a formation with free voids the same rung as a saturated copy. This treats the two alike, which is the rule with no special case in it, and costs half a point. The first thing the ladder shows is worth acting on next: 7,203 of the 7,760 objects on rung 0 are Δ literals.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Depth.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Depth.java
@@ -1,0 +1,81 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.xml.XML;
+import com.jcabi.xml.XMLDocument;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * How much of a program the tables turned out to say.
+ *
+ * <p>This reads the tables back and puts every object of the program on the
+ * ladder {@link Rung} describes. It is a measurement of ourselves rather than a
+ * fact about the program, so it writes nothing: what comes out is meant for the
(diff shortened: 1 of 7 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (489 >= 100); -8 for way too many hits-of-code (489 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+763](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6806, opened 2026-08-14
#6775: give string.as-ascii its own cant-read fallback

### LINKED ISSUE #6775: `string.as-ascii` unexpectedly exposes raw `as-i16` error for non-single-byte inputs
## What happens

`string.as-ascii` (`eo-runtime/src/main/eo/string/as-ascii.eo:12-15`) reads a character's ASCII code via `as-number. as-i16. 00-.concat char.as-bytes`, with no fallback of its own. For any input that isn't exactly one byte after the zero-padding (an empty string, or a multi-byte UTF-8 character), the call falls through to the underlying `bytes.as-i16` atom failure.

Reproduced through the actual runtime: `"".as-ascii` and `"Ж".as-ascii` (a 2-character-wide-in-UTF-8-bytes case) both terminate with:

```
Error in "Φ.string.as-ascii.φ.Δ" at string.as-ascii:13:2; Can't convert non 2 length bytes to i16, bytes are 00
Error in "Φ.string.as-ascii.φ.Δ" at string.as-ascii:13:2; Can't convert non 2 length bytes to i16, bytes are 00-D0-96
```

This message names `i16` and raw byte lengths, meaningless to a caller who only knows they passed a string to `as-ascii`.

## Why it happens

`as-i16` (`eo-runtime/src/main/eo/bytes/as-i16.eo:12`) is `[b cant-convert] > as-i16` — it already has a `cant-convert` fallback parameter for exactly this case, but `as-ascii` never binds it, so the raw atom error surfaces instead.

Every sibling `bytes/as-*` object (`as-u8.eo`, `as-i8.eo`, `as-u16.eo`, `as-u64.eo`, `as-number.eo`) exposes and documents a `cant-convert` fallback. Every `string` object with an input precondition (`at.eo`, `truncated.eo`, `repeated.eo`, `padded-left.eo`, `padded-right.eo`, `padded-zeros.eo`) takes its own `cant-*` fallback parameter. `as-ascii` is the only one in this family that does neither.

## What should happen

`as-ascii` should take its own fallback parameter (e.g. `[char cant-read] > as-ascii`), check that `char.as-bytes.size` is 1 up front, and return the fallback (or the bottom object when unbound) with a message naming the character it was given, e.g. `"Can't read %x as an ASCII code, it is not a single byte"`, instead of leaking the internal `as-i16` failure.

morphqdd:
string.as-ascii reads a character's ASCII code via
as-number.as-i16.(00-.concat char.as-bytes), with no fallback of its
own. For any input that isn't exactly one byte after the zero-padding
(an empty string, or a multi-byte UTF-8 character), the call falls
through to the underlying bytes.as-i16 atom failure:

  Error in "Φ.string.as-ascii.φ.Δ" at string.as-ascii:13:2; Can't
  convert non 2 length bytes to i16, bytes are 00

This message names i16 and raw byte lengths, meaningless to a caller
who only knows they passed a string to as-ascii.

as-i16 already has a cant-convert fallback parameter for exactly this
case, but as-ascii never binds it. Every sibling bytes/as-* object
exposes and documents a cant-convert fallback, and every string object
with an input precondition (at, truncated, repeated, padded-*) takes
its own cant-* fallback parameter. as-ascii was the only one in this
family that did neither.

Fix: as-ascii now takes its own cant-read fallback, checks
char.as-bytes.size up front, and returns the fallback (or the bottom
object when unbound) with a message naming the character it was given.

Fixes #6775

### DIFF
--- a/eo-runtime/src/main/eo/string/as-ascii.eo
+++ b/eo-runtime/src/main/eo/string/as-ascii.eo
@@ -1,6 +1,9 @@
 # Reads the single character `char` as its ASCII numeric value (0-255).
 # The byte of `char` is left-padded with a zero byte into a 16-bit word,
-# read as an integer and then expressed as a number.
+# read as an integer and then expressed as a number. If `char` is not
+# exactly one byte wide (an empty string, or a multi-byte UTF-8
+# character), the `cant-read` fallback is returned, or the bottom
+# object when unbound.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
@@ -9,10 +12,15 @@
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
 
-[char] > as-ascii
-  as-number. > @
-    as-i16.
-      00-.concat char.as-bytes
+[char cant-read] > as-ascii
+  if. > @
+    char.as-bytes.size.eq 1
+    as-number.
+      as-i16.
+        00-.concat char.as-bytes
+    cant-read
+      "Can't read %x as an ASCII code, it is not a single byte".printf
+        * char.as-bytes
 
   97.eq "a".as-ascii ++> can-read-the-code-of-a-lowercase-letter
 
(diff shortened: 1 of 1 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (30) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4532](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6808, opened 2026-08-14
#6776: use %f for string.at's fractional out-of-bounds message

### LINKED ISSUE #6776: `string.at` incorrectly truncates fractional out-of-bounds index in error message
## What happens

`string.at`'s out-of-bounds error message (`eo-runtime/src/main/eo/string/at.eo:34`) formats the rejected index with `%d`:

```eo
  * i
```

`%d` routes through `as-decimal`, which truncates the fraction. Reproduced through the runtime: `"some".at 1.5` (text `"some"` has 4 characters, so index `1` alone is in bounds) with a message-capturing fallback returns:

```
Given index 1 is out of text bounds
```

and `"some".at -1.5` returns:

```
Given index -1 is out of text bounds
```

Both messages are self-contradictory: index `1` is not out of bounds for a 4-character string, and the message never mentions that the real rejection reason was the fractional part, not the magnitude.

## Why it happens

`at` rejects a fractional index via the same `or.` branch that rejects an out-of-range index (`at.eo:24-32`, the `not. eq. floor.(number index) (number index)` check), so both failure modes share one message that only has room for an integer-shaped complaint.

Every sibling object in this file's family that rejects a non-integer numeric argument already uses `%f` for exactly this reason, and their own tests pin the fractional format: `truncated.eo:23` (`"Can't truncate text to %f characters"`, asserted as `-3.000000` in its own test), `repeated.eo:20`, `padded-left.eo:26`, `padded-right.eo:26`, `padded-zeros.eo:23`. `at.eo` is the outlier still using `%d`. There is no existing test in `at.eo` pinning the exact fractional-index message text, which is presumably why this drifted unnoticed.

## What should happen

`at.eo:34` should use `%f` instead of `%d`, so the message reads `Given index 1.500000 is out of text bounds`, matching the format convention of every sibling object and reflecting the actual (possibly fractional) index that was rejected.

morphqdd:
string.at's out-of-bounds error message formats the rejected index
with %d, which routes through as-decimal and truncates the fraction.
"some".at 1.5 (text "some" has 4 characters, so index 1 alone is in
bounds) returns "Given index 1 is out of text bounds" through a
message-capturing fallback: self-contradictory, since index 1 is not
out of bounds, and the message never mentions the real rejection
reason was the fractional part.

at rejects a fractional index via the same or. branch that rejects an
out-of-range index, so both failure modes share one message that only
had room for an integer-shaped complaint.

Every sibling object in this file's family that rejects a non-integer
numeric argument already uses %f for exactly this reason: truncated.eo,
repeated.eo, padded-left.eo, padded-right.eo, padded-zeros.eo. at.eo
was the outlier still using %d.

Fix: use %f, so the message reads "Given index 1.500000 is out of text
bounds". Added a test pinning the exact message text, matching
truncated.eo's own precedent.

Fixes #6776

### DIFF
--- a/eo-runtime/src/main/eo/string/at.eo
+++ b/eo-runtime/src/main/eo/string/at.eo
@@ -31,12 +31,18 @@
               number index
             number index
     fallback
-      "Given index %d is out of text bounds".printf
+      "Given index %f is out of text bounds".printf
         * i
     slice
       text
       index
       1
+
+  eq. ++> can-format-the-error-message-of-a-fractional-index-in-bounds
+    "Given index 1.500000 is out of text bounds"
+    "some".at
+      1.5
+      message > [message]
 
   eq. ++> can-recover-from-a-fractional-index-in-bounds
     "recovered"


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4544](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6810, opened 2026-08-14
#6777: guard NESTING bookkeeping in take() behind the FINE check

### LINKED ISSUE #6777: `PhDefault.take` unnecessarily performs `ThreadLocal` operations on every call for a rarely used counter
## What happens

`PhDefault.take(String)` (`eo-runtime/src/main/java/org/eolang/PhDefault.java:228, 254-259`) increments and decrements a `ThreadLocal<Integer>` counter (`NESTING`) on every call, unconditionally:

```java
PhDefault.NESTING.set(PhDefault.NESTING.get() + 1);
try {
    ...
} finally {
    final int current = PhDefault.NESTING.get();
    if (current > 0) {
        PhDefault.NESTING.set(current - 1);
    } else {
        PhDefault.NESTING.set(0);
    }
}
```

`NESTING` is read in exactly one other place: `padding()` (line 679), which is called only inside `if (PhDefault.LOGGER.isLoggable(Level.FINE))` (line 240) — i.e. only when FINE logging is actually enabled, which it never is during a normal build or test run.

`take()` is the single most frequently called method in the runtime: every object dispatch goes through it. So every dispatch pays two `ThreadLocal` lookups, two `ThreadLocal` writes, and `Integer` (un)boxing, purely to maintain a counter that almost never gets read.

## What should happen

Guard the `NESTING` increment/decrement with the same `LOGGER.isLoggable(Level.FINE)` check already used to guard the log call, so the bookkeeping only runs when it can actually be observed:

```java
final boolean logging = PhDefault.LOGGER.isLoggable(Level.FINE);
if (logging) {
    PhDefault.NESTING.set(PhDefault.NESTING.get() + 1);
}
try {
    ...
} finally {
    if (logging) {
        final int current = PhDefault.NESTING.get();
        PhDefault.NESTING.set(Math.max(0, current - 1));
    }
}
```

This preserves identical behavior (the counter still tracks nesting correctly whenever FINE is on) while removing the ThreadLocal overhead from the hot path in the common case.

morphqdd:
PhDefault.take(String) increments and decrements a ThreadLocal<Integer>
counter (NESTING) on every call, unconditionally. NESTING is read in
exactly one other place, padding(), which is only called inside
if (LOGGER.isLoggable(Level.FINE)) - i.e. only when FINE logging is
actually enabled, which it never is during a normal build or test run.

take() is the single most frequently called method in the runtime:
every object dispatch goes through it. So every dispatch paid two
ThreadLocal lookups, two ThreadLocal writes, and Integer (un)boxing,
purely to maintain a counter that almost never gets read.

Fix: guard the NESTING increment/decrement with the same
isLoggable(Level.FINE) check already used to guard the log call, so
the bookkeeping only runs when it can actually be observed. Behavior
is identical whenever FINE is on; the ThreadLocal overhead disappears
from the hot path in the common case.

Fixes #6777

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -225,7 +225,10 @@ public class PhDefault implements Phi, Cloneable {
                 )
             );
         }
-        PhDefault.NESTING.set(PhDefault.NESTING.get() + 1);
+        final boolean logging = PhDefault.LOGGER.isLoggable(Level.FINE);
+        if (logging) {
+            PhDefault.NESTING.set(PhDefault.NESTING.get() + 1);
+        }
         try {
             final Phi resolved;
             if (this.loaded().containsKey(name)) {
@@ -237,7 +240,7 @@ public class PhDefault implements Phi, Cloneable {
             } else {
                 resolved = this.absent(name);
             }
-            if (PhDefault.LOGGER.isLoggable(Level.FINE)) {
+            if (logging) {
                 PhDefault.LOGGER.log(
                     Level.FINE,
                     String.format(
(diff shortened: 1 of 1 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4556](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6811, opened 2026-08-14
Say that an object is a datum

maxonfjvipon:
The bytes of a literal were the largest thing the tables said nothing about — 7,178 of the 7,760 objects the ladder put on the bottom rung. A literal is three objects, and only the last of them is the ground:

```xml
<o base="Φ.number" loc="Φ.app.answer">
  <o base="Φ.bytes" loc="Φ.app.answer.α0">
    <o loc="Φ.app.answer.α0.α0">40-45-00-00-00-00-00-00</o>
```

That leaf has no base, so no rule ever looked at it, and there was no form of type to write it down with anyway. Now there is a third one, beside a reference and the binds inside it:

```xml
<type id="Φ.app.answer.α0.α0">
  <data/>
</type>
```

A datum is deliberately not written as a copy of `Φ.bytes`. `Φ.bytes` is a formation with a void, and saying "a copy of it" would put a free void on an object that has none, and drop it a rung for a reason that is not true.

Three things move together. `Links` writes the rows, since a datum has no base to resolve and this is the one row that clue writes without a reference to look at. `Pairs` reads them back, because resolving dispatches means reading the table and writing it again, and a row that is not a pair falls out of a document rebuilt from pairs alone — the third pack is there to keep that from regressing quietly. And `Rung` sends a datum to the top: asking what more there is to know about `01-` is asking nothing. The top rung is renamed from "a formation seen whole" to "nothing left to find out", which is what it now holds.

On `eo-runtime`, 171 files:

```
                        before    after
described                82.0%    98.6%
depth                    48.8%    65.5%
nothing at all            7760      582
nothing left to find out  1688     8866
```

All 31,876 existing rows are unchanged in target and order, `provides.xml` and `needs.xml` are byte-identical, and two runs of the goal produce the same file.

What is left on the bottom rung is 582 objects — 1.4%: 433 dispatches whose receiver nothing describes, 66 terminations, 58 names in scope resolving to nothing, and the 25 `λ` markers of atoms, which carry the forma they come back with in an attribute nobody reads yet.

The ticket said 7,203. That count included those 25 `λ`, which have no base and no bytes; the rows are 7,178.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Depth.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Depth.java
@@ -10,6 +10,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -57,7 +58,8 @@ public final class Depth {
         final Pairs pairs = new Pairs(new XMLDocument(this.tables.resolve("links.xml")));
         final Rung rung = new Rung(
             new Ungrouped(given, Collections.emptyMap()).rows(),
-            given.xpath("//attr[@void='true']/@type"),
+            new HashSet<>(given.xpath("//attr[@void='true']/@type")),
+            new HashSet<>(pairs.data()),
             new Ends(pairs.all()).names()
         );
         final Map<String, Integer> filled = pairs.binds();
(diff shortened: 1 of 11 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (173 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+775](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6818, opened 2026-08-14
#6761: honor rewriteBinaries in Placing.isNotAlreadyPlaced

### LINKED ISSUE #6761: `Placing.rewriteBinaries` doesn't rewrite an already-placed binary despite a change in dependencies
## What happens

`Placing.PlacedDependency.isNotAlreadyPlaced` (`eo-maven-plugin/src/main/java/org/eolang/maven/Placing.java:189-206`) decides whether to skip a binary using only whether a tojo for the target already exists, the target file exists, and the tojo is not marked unplaced. It never reads `this.rwte`, the `rewriteBinaries` flag passed into the constructor.

`rewrite` is only consulted later, inside `placeBinary`, on files that already passed this filter. So for a binary that is already placed and still exists, `-Deo.rewriteBinaries=true` and `-Deo.rewriteBinaries=false` behave identically: the file is skipped either way, and the option (default `true`, `MjSafe.java:212`, documented as "Rewrite binaries in output directory or not") has no effect on the case it is named for.

Concrete scenario: build once so `placed.json` and `target/classes` are populated. A dependency's version changes, `MjRegister.removeOldFiles()` clears the resolve directory, and `Resolving` unpacks the new jar with different class bytes. `Placing` then skips every one of them because the old tojo still says "placed", so `target/classes` keeps the stale binaries until a `mvn clean`.

The only test exercising the rewrite path, `MjPlaceTest.rewritesAlreadyPlacedBinaries`, calls `maven.placed().unplaceAll()` before running the mojo, which sidesteps `isNotAlreadyPlaced` entirely rather than testing the flag.

## What should happen

`isNotAlreadyPlaced` should return `true` when `rewrite` is set (or compare the file's actual content/hash instead of mere presence, using `TjPlaced.sameHash`, which exists but is unused here), so a changed binary gets replaced and an identical one gets skipped, regardless of the tojo's placed/unplaced state.

morphqdd:
Placing.PlacedDependency.isNotAlreadyPlaced decides whether to skip a
binary using only whether a tojo for the target already exists, the
target file exists, and the tojo is not marked unplaced. It never read
rwte, the rewriteBinaries flag passed into the constructor. rewrite
was only consulted later, inside placeBinary, on files that already
passed this filter.

So for a binary that is already placed and still exists,
-Deo.rewriteBinaries=true and -Deo.rewriteBinaries=false behaved
identically: the file was skipped either way, and the option (default
true, documented as "Rewrite binaries in output directory or not")
had no effect on the case it is named for. A dependency version change
would leave stale binaries in target/classes until a mvn clean.

Fix: isNotAlreadyPlaced now returns true unconditionally when rewrite
is on, matching the option's stated purpose.

Two existing tests were pinning the old (buggy) behavior under the
default rewriteBinaries=true:
- doesNotPlacesAgainIfWasNotUnplaced ran with the default true but
  asserted the file was NOT replaced. Renamed and flipped to expect a
  replace, plus a new explicit off-case test.
- skipsAlreadyPlacedBinaries relied on the same default to assert a
  skip; now explicitly sets rewriteBinaries=false, matching what it's
  actually testing.

Fixes #6761

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Placing.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Placing.java
@@ -183,9 +183,10 @@ final class Placing implements Step {
         }
 
         /**
-         * Check if the file has not been placed yet.
+         * Check if the file has not been placed yet, or {@link #rwte} forces
+         * it to be placed again regardless.
          * @param file The file to check
-         * @return True if the file is not already placed
+         * @return True if the file is not already placed, or rewriting is on
          */
         private boolean isNotAlreadyPlaced(final Path file) {
             final Path target = Placing.this.classes.resolve(
@@ -193,7 +194,8 @@ final class Placing implements Step {
             );
             final Optional<TjPlaced> tojo = Placing.this.placed.find(target);
             final boolean res;
-            if (tojo.isPresent()
+            if (!this.rwte
+                && tojo.isPresent()
                 && Files.exists(target)
                 && !tojo.get().unplaced()
             ) {
--- a/eo-maven-plugin/src/test/java/org/eolang/maven/MjPlaceTest.java
+++ b/eo-maven-plugin/src/test/java/org/eolang/maven/MjPlaceTest.java
@@ -54,7 +54,7 @@ final class MjPlaceTest {
         ).toFile().lastModified();
         MatcherAssert.assertThat(
             "PlaceMojo must skip already placed binaries, but it doesn't",
-            new FakeMaven(temp).withPlacedBinary(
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4568](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6819, opened 2026-08-14
#6760: restrict removeStale to directories eo itself wrote into

### LINKED ISSUE #6760: `JavaFiles.removeStale()` improperly removes `.java` files from shared `generated-sources` directory
## What happens

`JavaFiles.removeStale()` (`eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java:138-160`) walks the whole `generatedDir` tree and deletes every `.java` file that this transpile run did not itself write, except a `package-info.java` carve-out.

`generatedDir` defaults to `${project.build.directory}/generated-sources` (`MjSafe.java`, `eo.generatedDir` property), which is the standard Maven convention directory. Other code generators write there too by default: antlr4, protobuf, jaxb2, openapi-generator, and `maven-compiler-plugin`'s own annotation-processor output (`target/generated-sources/annotations`).

Most of those bind to the `generate-sources` phase, which runs before `process-sources` where eo's `transpile` goal lives. So in a single `mvn compile` on a project that uses eo-maven-plugin alongside any of those tools: the other tool writes its `.java` files first, `transpile` runs, and `removeStale()` deletes all of them because they are not in eo's own `fresh` set. `mvn compile` then fails on missing classes.

## What should happen

`removeStale()` should only delete files eo itself could have produced, not everything under the shared directory. Either restrict the walk to the subdirectories eo actually writes into, or track eo's own previous-run output in a manifest and delete only from that set.

morphqdd:
JavaFiles.removeStale() walks the whole generatedDir tree and deletes
every .java file this transpile run did not itself write, except a
package-info.java carve-out.

generatedDir defaults to target/generated-sources, the standard Maven
convention directory. Other code generators write there too by
default: antlr4, protobuf, jaxb2, openapi-generator, and
maven-compiler-plugin's own annotation-processor output
(target/generated-sources/annotations). Most of those bind to the
generate-sources phase, which runs before process-sources where eo's
transpile goal lives. So in a single mvn compile on a project that
uses eo-maven-plugin alongside any of those tools: the other tool
writes its .java files first, transpile runs, and removeStale()
deletes all of them because they are not in eo's own fresh set. mvn
compile then fails on missing classes.

Fix: removeStale() now only deletes a .java file whose parent
directory is one eo itself wrote a fresh file into during this run,
computed the same way the existing package-info.java carve-out already
scoped itself to eo's own directories.

Fixes #6760

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/JavaFiles.java
@@ -133,6 +133,15 @@ final class JavaFiles {
 
     /**
      * Delete generated Java files absent from the current XMIR collection.
+     *
+     * <p>Only considers files inside a directory this run's own {@link
+     * #fresh} output actually landed in, never the whole {@link #generated}
+     * tree: {@code generated-sources} is the standard Maven convention
+     * directory, and other generators (antlr4, protobuf, jaxb2, the
+     * annotation-processor output of maven-compiler-plugin) can write their
+     * own {@code .java} files there too, outside any directory eo itself
+     * touched this run.</p>
+     *
      * @throws IOException If fails to inspect or remove a generated file
      */
     void removeStale() throws IOException {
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (36) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4580](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6835, opened 2026-08-14
#6817: preserve the sign of an exact negative zero in as-fixed

### LINKED ISSUE #6817: number.as-fixed erases the IEEE negative-zero sign
## What happens

`number.as-fixed` erases the sign of an exact IEEE negative zero:

| --- | --- | --- |
| `string (number.as-fixed -0 2)` | `"0.00"` | `"-0.00"` |
| `string (number.as-fixed 0 2)` | `"0.00"` | `"0.00"` |
| `string (number.as-fixed -0.001 2)` | `"0.00"` | `"0.00"` |

Reproduced against current `master` (`4a4f9f4c49`). The final row is a control
case: a non-zero negative value which rounds to zero may intentionally lose its
sign. Exact `-0` is different: it already carries a distinct IEEE-754 bit
pattern and is preserved by EO numerical operations such as `number.arc-sine`,
`number.floor`, and `number.sqrt`.

## Why it happens

[`number.as-fixed`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number/as-fixed.eo)
chooses its negative branch with `n.lt 0`. IEEE negative zero does not compare
less than zero, so it enters the positive branch and formatting constructs an
unsigned zero string.

## Expected behavior

Fixed-point formatting should retain the sign of an exact negative zero, as
other IEEE-aware formatting implementations do. Only a non-zero negative input
that rounds to zero should use the existing unsigned-zero behaviour.

## Suggested fix

Detect the negative-zero byte representation before the `n.lt 0` branch and
prefix the formatted zero with `-` when the input is exact `-0`. Add a
regression that compares the output for `-0` and `0` at a positive precision.

morphqdd:
number.as-fixed erases the sign of an exact IEEE negative zero:

| --- | --- | --- |
| string (number.as-fixed -0 2) | "0.00" | "-0.00" |
| string (number.as-fixed 0 2) | "0.00" | "0.00" |
| string (number.as-fixed -0.001 2) | "0.00" | "0.00" (control case, unchanged) |

as-fixed chooses its negative branch with n.lt 0. IEEE negative zero
does not compare less than zero, so it falls into the positive branch
and formats as an unsigned zero string.

Fix: check the exact -0 byte pattern (80-00-00-00-00-00-00-00) first,
same idiom already used in floor.eo, sqrt.eo, and arc-sine.eo, and
prefix "-" onto the positively-formatted zero for that case only. A
genuinely negative non-zero value that rounds to zero (e.g. -0.001)
still drops its sign, matching the existing control-case tests.

Fixes #6817

### DIFF
--- a/eo-runtime/src/main/eo/number/as-fixed.eo
+++ b/eo-runtime/src/main/eo/number/as-fixed.eo
@@ -25,41 +25,45 @@
       "Can't write a fixed-point decimal with %f fraction places".printf
         * places
     if.
-      n.lt 0
-      if. > [a] >> signed
-        eq.
-          floor.
-            plus.
-              a.times
-                pow10 places
-              0.5
-          0
-        positive a
-        "-".as-bytes.concat
-          positive a
-      | (n.times -1)
-      [a] >> positive
-        if. > @
-          digits.eq scale
-          if. > [unit rest] >> written
-            places.eq 0
-            as-decimal unit
-            concat.
+      n.as-bytes.eq 80-00-00-00-00-00-00-00
+      "-".as-bytes.concat
+        [a] >> positive
+          if. > @
+            digits.eq scale
+            if. > [unit rest] >> written
+              places.eq 0
+              as-decimal unit
               concat.
-                as-decimal unit
-                ".".as-bytes
-              rec-pad rest places --
-          | (whole.plus 1) 0
-          written whole digits
-        floor. > digits
-          plus.
-            times.
-              a.minus whole
-              scale
-            0.5
-        pow10 places > scale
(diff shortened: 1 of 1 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4644](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6836, opened 2026-08-14
#6823: preserve the IEEE negative-zero sign in mod's zero remainder

### LINKED ISSUE #6823: number.mod loses the IEEE negative-zero sign of a zero dividend
## What happens

`number.mod` loses the sign of a negative-zero dividend. Its result is positive
zero for both finite and infinite non-zero divisors.

| expression | actual bytes | expected IEEE-754 bytes |
| --- | --- | --- |
| `(-0).mod 3` | `00-00-00-00-00-00-00-00` | `80-00-00-00-00-00-00-00` |
| `(-0).mod pinf` | `00-00-00-00-00-00-00-00` | `80-00-00-00-00-00-00-00` |
| `0.mod 3` | `00-00-00-00-00-00-00-00` | `00-00-00-00-00-00-00-00` |

Reproduced against current `master` (`4a4f9f4c49`). Java's IEEE-754 remainder
operator produces `-0` for the first two rows. The final row is a control
case for ordinary positive zero.

## Why it matters

The sign is observable through `as-bytes`, and EO already preserves it in
other numerical operations such as `sine`, `sqrt`, and `arc-sine`. The
`number.mod` documentation also promises the true remainder for finite inputs;
for an exact zero remainder IEEE-754 specifies the dividend's sign.

## Why it happens

[`number.mod`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number/mod.eo)
computes through the dividend magnitude. `abs` normalises `-0` to `+0`, and
the later reconstruction does not restore the original zero sign.

## Expected behavior

When a zero dividend is negative zero and the divisor is non-zero, `mod`
should return negative zero. Positive zero should remain positive.

## Suggested fix

Recognise a negative-zero dividend before the magnitude calculation and return
it directly for every non-zero divisor. Add raw-byte regressions for finite
and infinite divisors.

morphqdd:
number.mod loses the sign of a negative-zero dividend. Its result is
positive zero for both finite and infinite non-zero divisors:

| expression | actual bytes | expected bytes |
| --- | --- | --- |
| (-0).mod 3 | 00-00-00-00-00-00-00-00 | 80-00-00-00-00-00-00-00 |
| (-0).mod pinf | 00-00-00-00-00-00-00-00 | 80-00-00-00-00-00-00-00 |
| 0.mod 3 | 00-00-00-00-00-00-00-00 | 00-00-00-00-00-00-00-00 |

Two places drop the sign:
- The infinite-non-nan-divisor branch returned x.plus 0. IEEE -0.0 +
  0.0 is +0.0, a well-known IEEE-754 quirk, so this alone flips the
  sign.
- The finite-divisor branch picks its unsigned-magnitude result
  whenever gte.(dividend, 0), which -0 satisfies (IEEE -0 == 0), so the
  negation branch never runs for it.

Fix: return x directly (no arithmetic) for the infinite-non-nan-divisor
case, and check the exact -0 byte pattern before the magnitude
computation for the finite-divisor case, matching the idiom already
used in floor.eo/sqrt.eo/arc-sine.eo for the same class of bug.

Two existing tests were pinning the old positive-zero result
(can-return-positive-zero-for-negative-zero-dividend-and-*-infinite-divisor);
renamed and flipped them, plus added the missing finite-divisor case.

Fixes #6823

### DIFF
--- a/eo-runtime/src/main/eo/number/mod.eo
+++ b/eo-runtime/src/main/eo/number/mod.eo
@@ -28,33 +28,34 @@
     dividend.is-finite.not.if
       nan
       divisor.is-finite.not.if
-        divisor.is-nan.if
-          nan
-          x.plus 0
+        divisor.is-nan.if nan x
         if.
-          gte.
-            number x.as-bytes >> dividend
-            0
-          [] >> abs-mod
-            shrink > @
-              grow
-                abs divisor >> absy
-              abs dividend >> absx
-            if. > [chunk] > grow
-              chunk.gte absx
-              chunk
-              grow
-                chunk.plus chunk
-            if. > [chunk rest] > shrink
-              chunk.lt absy
-              rest
-              shrink
-                chunk.div 2
-                if.
-                  rest.gte chunk
-                  rest.minus chunk
-                  rest
-          abs-mod.neg
+          x.as-bytes.eq 80-00-00-00-00-00-00-00
+          x
+          if.
+            gte.
+              number x.as-bytes >> dividend
+              0
+            [] >> abs-mod
+              shrink > @
+                grow
+                  abs divisor >> absy
+                abs dividend >> absx
+              if. > [chunk] > grow
+                chunk.gte absx
+                chunk
+                grow
+                  chunk.plus chunk
+              if. > [chunk rest] > shrink
+                chunk.lt absy
+                rest
+                shrink
+                  chunk.div 2
+                  if.
+                    rest.gte chunk
+                    rest.minus chunk
+                    rest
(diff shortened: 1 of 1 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4596](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6837, opened 2026-08-14
#6821: return the exact unit value for cosine of zero

### LINKED ISSUE #6821: number.cosine returns the value below one instead of exactly one at zero
## What happens

`number.cosine` does not return the exact unit value at zero. It produces the
adjacent lower IEEE-754 value, so an EO equality check fails:

| --- | --- | --- |
| `0.cosine.eq 1` | `false` | `true` |
| `0.cosine.as-bytes` | `3F-EF-FF-FF-FF-FF-FF-FF` | `3F-F0-00-00-00-00-00-00` |
| `0.sine.eq 0` | `true` | `true` |

Reproduced against current `master` (`4a4f9f4c49`). The result is
`0.9999999999999999`, one ULP below one. The sine row is a control case for
the corresponding exact zero identity.

## Why it happens

[`number.cosine`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number/cosine.eo)
implements cosine as `sine(num + pi / 2)`. The finite approximation of `pi`
makes the shifted input miss the exact extremum, and the Taylor calculation
returns the neighbouring representable value.

## Expected behavior

The exact identity `cos(0) = 1` should be observable through EO equality, as
it is in standard IEEE-aware math libraries. A one-ULP error is still visible
to programs that use `eq` or serialise the result.

## Suggested fix

Return `1` for a zero argument before applying the phase shift. Add an exact
`0.cosine.eq 1` regression in addition to tolerance-based accuracy tests.

morphqdd:
number.cosine does not return the exact unit value at zero. It
produces the adjacent lower IEEE-754 value, so an EO equality check
fails:

| --- | --- | --- |
| 0.cosine.eq 1 | false | true |
| 0.cosine.as-bytes | 3F-EF-FF-FF-FF-FF-FF-FF | 3F-F0-00-00-00-00-00-00 |

The result is 0.9999999999999999, one ULP below one.

cosine is implemented as sine(num + pi/2). The finite approximation of
pi makes the shifted input miss the exact extremum, and the Taylor
calculation returns the neighbouring representable value instead of
exactly 1.

Fix: return 1 for a zero argument before applying the phase shift.
Added an exact 0.cosine.eq 1 regression alongside the existing
tolerance-based accuracy tests.

Fixes #6821

### DIFF
--- a/eo-runtime/src/main/eo/number/cosine.eo
+++ b/eo-runtime/src/main/eo/number/cosine.eo
@@ -1,5 +1,7 @@
 # Cosine of `num` radians as `org.eolang.number`, taken from the sine of the
-# argument advanced by a quarter turn, since cos(x) equals sin(x + pi/2).
+# argument advanced by a quarter turn, since cos(x) equals sin(x + pi/2). The
+# finite approximation of `pi` used in that shift misses the exact extremum
+# at zero, so zero is special-cased to return the exact unit value.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
@@ -9,9 +11,12 @@
 +spdx SPDX-License-Identifier: MIT
 
 [num] > cosine
-  sine > @
-    num.plus
-      pi.div 2
+  if. > @
+    num.eq 0
+    1
+    sine.
+      num.plus
+        pi.div 2
 
   lt. ++> can-behave-as-an-even-function
     abs.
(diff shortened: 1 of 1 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (15) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4656](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6838, opened 2026-08-14
#6813: validate UTF-8 sequences in string.slice

### LINKED ISSUE #6813: `string.slice` unexpectedly accepts and retains invalid UTF-8 sequences
## What happens

`string` is documented as UTF-8 text with Unicode-aware character operations,
but `string.slice` accepts malformed byte sequences and returns them unchanged.
The corresponding `string.length` operation rejects the same inputs.

| --- | --- | --- |
| `(string C0-80).slice 0 1` | `C0-80` | failure: overlong two-byte encoding |
| `(string E1-80-41).slice 0 1` | `E1-80-41` | failure: the third byte is not a continuation byte |
| `(string ED-A0-80).slice 0 1` | `ED-A0-80` | failure: UTF-16 surrogate encoding |

Reproduced against current `master` (`4a4f9f4c49`). As a control,
`(string C0-80).length` terminates, as it should.

## Why it is inconsistent

[`string.slice`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/slice.eo)
decides a character width from its leading byte, but its `increase` path only
checks that enough bytes remain. It never verifies continuation bytes or the
Unicode ranges for the second byte. The object can therefore expose invalid
UTF-8 as a `string`, despite the package contract and the stricter validation
already implemented by
[`string.length`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/length.eo).

## Expected behavior

`string.slice` should reject malformed UTF-8 using the same validity rules as
`string.length`: valid continuation bytes, no overlong encodings, no surrogate
code points, and no code points above U+10FFFF.

## Suggested fix

Reuse the UTF-8 validation logic from `string.length`, or extract a shared
validator used by both operations. Add regressions for an overlong two-byte
sequence, a broken continuation byte, and a surrogate encoding.

gemshrine:
Fixes #6813.

`string.slice` decided a character width from its leading byte and only checked
that enough trailing bytes remained, so it accepted malformed data that
`string.length` already rejects: `(string C0-80).slice 0 1` returned `C0-80`,
and so did `(string E1-80-41).slice 0 1` and `(string ED-A0-80).slice 0 1`.

The walker now validates every multi-byte sequence with the same UTF-8 rules
as `string.length` (Unicode Table 3-7): each following byte must be a `10xxxxxx`
continuation byte, and the second byte is range-checked per lead so that

- overlong encodings are rejected: `C0`/`C1` leads, `E0` with an `80`-`9F`
  second byte, `F0` with an `80`-`8F` second byte;
- UTF-16 surrogates are rejected: `ED` with an `A0`-`BF` second byte;
- code points above U+10FFFF are rejected: `F4` with a `90`-`BF` second byte,
  and `F5`-`FF` leads.

A malformed sequence now terminates the computation, exactly like the existing
truncated-character behaviour. The validity check runs after the existing
length check, so a genuinely truncated character still reports the "not enough
bytes" error rather than a malformed one.

Verified against the runtime (each case dataized through the `string`/`slice`
atoms): the three malformed inputs from the issue throw, while valid inputs
still slice correctly, including two-byte and four-byte characters. All the
existing character-slicing examples stay green.

### DIFF
--- a/eo-runtime/src/main/eo/string/slice.eo
+++ b/eo-runtime/src/main/eo/string/slice.eo
@@ -4,7 +4,15 @@
 # take. A fractional, `nan` or infinite `start` or `len` terminates the
 # computation immediately. Both of them walk the UTF-8 sequence character by
 # character, so a multi-byte character counts as one. An index outside the
-# bounds of the `text` terminates the computation.
+# bounds of the `text` terminates the computation. Every multi-byte sequence is
+# validated against the UTF-8 rules of the Unicode standard (Table 3-7), the
+# same way `length` does: each following byte must be a `10xxxxxx` continuation
+# byte, and the second byte is range-checked so that overlong encodings
+# (`C0`/`C1` leads, `E0` with an `80`-`9F` second byte, `F0` with an `80`-`8F`
+# second byte), UTF-16 surrogates (`ED` with an `A0`-`BF` second byte) and code
+# points above U+10FFFF (`F4` with a `90`-`BF` second byte, and `F5`-`FF` leads)
+# are rejected. A malformed sequence terminates the computation, exactly like a
+# truncated character does.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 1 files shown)


eo-runtime/src/main/eo/string/slice.eo
```diff
@@ -3,13 +3,22 @@
...
+# byte, and `F5`-`FF` leads) are rejected. A malformed sequence terminates the
+# computation, exactly like a truncated character does.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
 +package string
 +version 0.0.0
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
++unlint many-void-attributes
```
yegor256: @gemshrine don't suppress, fix the code instead

gemshrine: Fixed in the latest push. The `+unlint` is gone and `increase` now just returns the next byte index (or fails on a malformed or truncated sequence), while `recidx` drives the recursion. That brings `increase` down to three void attributes, so nothing is suppressed.

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2590](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (121 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3722](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6839, opened 2026-08-14
#6824: reject non-integer bounds in range.naturals

### LINKED ISSUE #6824: range.naturals returns fractional values when given a fractional bound
## What happens

`range.naturals` accepts fractional bounds and returns fractional elements,
despite being documented as a range of natural numbers.

| expression | actual result | expected behavior |
| --- | --- | --- |
| `(range 0.5 2).naturals` | `* 0.5 1.5` | reject non-integer bounds |
| `(range -5 0).naturals` | `* -5 -4 -3 -2 -1` | unchanged control case |
| `(range 1 4).naturals` | `* 1 2 3` | unchanged control case |

Reproduced against current `master` (`4a4f9f4c49`).

## Why it happens

[`range.naturals`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/range/naturals.eo)
passes its inherited `start` value directly into a builder whose `next` adds
one. It validates neither boundary, so a fractional seed remains fractional
throughout the generated sequence.

## Expected behavior

`naturals` should contain integer values only. Fractional and non-finite bounds
should be rejected rather than producing a non-natural range.

## Suggested fix

Validate both inherited bounds with `is-integer` and finiteness before building
the range. Add a regression for `range 0.5 2`.

gemshrine:
Fixes #6824.

`range.naturals` fed its inherited `start` straight into a builder whose `next`
adds one, without validating the bounds. A fractional seed therefore stayed
fractional through the whole sequence: `(range 0.5 2).naturals` produced
`* 0.5 1.5`, which is not a range of natural numbers.

Both bounds are now checked with `is-integer` before the range is built, and a
non-integer bound terminates the computation. `is-integer` already returns
false for `nan` and both infinities, so the same check also rejects non-finite
bounds without a separate `is-finite` guard.

Verified by reasoning against `number.is-integer` (its own tests confirm
`32.is-integer`, `32.5.is-integer.not`, `nan.is-integer.not`,
`pinf.is-integer.not`, `ninf.is-integer.not`). The existing integer-bound
examples (`range -5 0`, `range 1 10`, the empty and descending ranges) stay
green, and new negative examples cover a fractional lower and upper bound.

### DIFF
--- a/eo-runtime/src/main/eo/range/naturals.eo
+++ b/eo-runtime/src/main/eo/range/naturals.eo
@@ -1,6 +1,10 @@
 # Range of natural numbers from `start` to `end` (soft border) with step = 1.
 # It is a method of `range`, so it reads `start` and `end` from the range it
-# is taken from.
+# is taken from. Both bounds must be integers, otherwise the computation
+# terminates: a fractional or non-finite seed would otherwise stay fractional
+# or non-finite through the whole generated sequence, contradicting the range
+# of natural numbers. `is-integer` already excludes `nan` and the infinities,
+# so it doubles as the finiteness check.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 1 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (37) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2352](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6840, opened 2026-08-14
#6825: require eight-byte width in number.is-finite

### LINKED ISSUE #6825: `number.is-finite` classifies non-eight-byte values as finite incorrectly
## What happens

`number` is documented as an eight-byte IEEE-754 value, but `number.is-finite`
reports `true` for byte sequences that cannot represent a number at all.

| --- | --- | --- |
| `(number FF-).is-finite` | `true` | `false` or a clear failure |
| `(number 00-00).is-finite` | `true` | `false` or a clear failure |
| `(number 00-00-00-00-00-00-00-00).is-finite` | `true` | `true` |

Reproduced against current `master` (`4a4f9f4c49`). The first two values fail
when subsequently dataized as numbers because they are not eight bytes long;
the third row is a valid zero control case.

## Why it happens

[`number.is-finite`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number/is-finite.eo)
only excludes `nan` and the two infinity encodings. `number.is-nan` returns
false for every input whose byte length is not eight, so malformed widths fall
through and are classified as finite.

## Expected behavior

Only a valid eight-byte IEEE-754 value can be finite. Invalid byte widths
should either be rejected by the predicate or return `false`.

## Suggested fix

Require `n.as-bytes.size` to equal eight before applying the NaN/infinity
checks, and add regressions for one- and two-byte inputs.

gemshrine:
Fixes #6825.

`number.is-finite` only excluded `nan` and the two infinity encodings. Because
`number.is-nan` returns `false` for any byte sequence whose length is not eight,
malformed widths fell through and were reported as finite: `(number FF-).is-finite`
and `(number 00-00).is-finite` both returned `true`, even though those values
throw the moment they are dataized as numbers.

The predicate now requires `n.as-bytes.size` to equal eight before the NaN and
infinity checks, so only a genuine eight-byte IEEE-754 value can be finite.

Verified by the added examples, which run in CI: a one-byte `FF-` and a
two-byte `00-00` are now reported as not finite, while the existing finite,
NaN and infinity examples stay green.

### DIFF
--- a/eo-runtime/src/main/eo/number/is-finite.eo
+++ b/eo-runtime/src/main/eo/number/is-finite.eo
@@ -1,4 +1,7 @@
-# TRUE when `n` is neither infinite nor not-a-number.
+# TRUE when `n` is a valid eight-byte IEEE-754 value that is neither infinite
+# nor not-a-number. A byte sequence of any other width can never represent a
+# number, so it is not finite: `is-nan` returns FALSE for such widths, which
+# would otherwise let them fall through as finite.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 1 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2364](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6841, opened 2026-08-14
#6815: handle the minimum signed 64-bit value explicitly

### LINKED ISSUE #6815: number.as-decimal rejects the representable minimum signed 64-bit value
## What happens

`number.as-decimal` refuses the minimum signed 64-bit value:

```eo
-9223372036854775808.as-decimal
```

It terminates with an error saying that a magnitude of `2^63` is outside the
supported range. By contrast, the adjacent representable value is formatted
correctly:

| --- | --- | --- |
| `-9223372036854775808.as-decimal` | failure | `"-9223372036854775808"` |
| `-9223372036854774784.as-decimal` | `"-9223372036854774784"` | `"-9223372036854774784"` |

The literal at the boundary is exactly representable as an EO `number` and is
also the value of `80-00-00-00-00-00-00-00.as-i64.as-number`. The behaviour
was confirmed by the current `master` implementation (`4a4f9f4c49`).

## Why this is a bug

[`number.as-decimal`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/number/as-decimal.eo)
describes its implementation as using `i64` digits. A signed 64-bit integer
range includes `-2^63`; only the positive magnitude `+2^63` is outside it.
The current magnitude guard treats both cases alike before the negative branch
can process the value.

## Expected behavior

All signed 64-bit values should have a decimal representation. In particular,
the minimum value should format as `-9223372036854775808` rather than fail.

## Suggested fix

Handle the negative `i64` minimum without first negating it to a positive
number. The digit loop can work on the signed value, or the minimum case can
be handled explicitly. Add a regression based on
`80-00-00-00-00-00-00-00.as-i64.as-number`.

morphqdd:
number.as-decimal refuses the minimum signed 64-bit value:

| --- | --- | --- |
| -9223372036854775808.as-decimal | failure | "-9223372036854775808" |
| -9223372036854774784.as-decimal | "-9223372036854774784" | "-9223372036854774784" |

as-decimal describes its implementation as using i64 digits. A signed
64-bit integer range includes -2^63; only the positive magnitude +2^63
is outside it. The current magnitude guard (n.abs.gte 2^63) treats
both the same before the negative branch can process the value, and
the negative branch itself negates n before converting to i64, which
overflows for exactly -2^63 (negating it produces +2^63, which does
not fit in signed i64 either).

Fix: special-case the exact minimum value before the magnitude guard,
handling it explicitly as the issue suggests, rather than restructure
the digit loop to work on signed values directly.

Added a regression based on 80-00-00-00-00-00-00-00.as-i64.as-number.

Fixes #6815

### DIFF
--- a/eo-runtime/src/main/eo/number/as-decimal.eo
+++ b/eo-runtime/src/main/eo/number/as-decimal.eo
@@ -17,36 +17,39 @@
     T
       "Can't write a non-finite number as decimal"
     if.
-      n.abs.gte
-        2.power 63
-      T
-        "Can't write this number as decimal digits: its magnitude is 2^63 or larger, past the range of a signed 64-bit integer"
+      n.eq -9.223372036854776e18
+      "-9223372036854775808".as-bytes
       if.
-        n.lt 0
-        if. > [m] >> negative
-          m.as-number.eq 0
-          digits m
-          "-".as-bytes.concat
+        n.abs.gte
+          2.power 63
+        T
+          "Can't write this number as decimal digits: its magnitude is 2^63 or larger, past the range of a signed 64-bit integer"
+        if.
+          n.lt 0
+          if. > [m] >> negative
+            m.as-number.eq 0
             digits m
-        |
-          as-i64.
-            n.times -1
-            T message > [message] >> failure
-        [n] >> digits
-          if. > @
-            n.lt ten
-            as-char
-              n.as-number.plus 48
-            concat.
-              digits q
+            "-".as-bytes.concat
+              digits m
+          |
+            as-i64.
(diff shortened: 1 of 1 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4612](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6842, opened 2026-08-14
#6797: don't lower an aliased tuple.empty reference to a star

### LINKED ISSUE #6797: Printer renders an aliased `tuple.empty` reference as `*` in `tuples-to-stars.xsl`
`tuples-to-stars.xsl:41` matches every `o[@base = 'Φ.tuple.empty']` and marks it as a star, on the assumption that only the `*` shorthand can carry that base. It cannot: `resolve-aliases.xsl:13` puts the very same base on a bare reference whenever the program declares `+alias tuple.empty`, so an ordinary reference to that object is printed as `*` rather than the name the author wrote.

```
+alias tuple.empty
+package foo

# App.
[] > app
  empty > a
  empty > b
```

Printing this program's XMIR through `Xmir.FOR_EO` gives `* > a` and `* > b`, and `+alias tuple.empty` stays in the preamble with nothing left referring to it.

`restore-aliases.xsl` does put the short name back — after it runs the node reads `<o base="empty" star=""/>` — but it sits at `Xmir.java:72`, five places behind `tuples-to-stars.xsl` at `Xmir.java:67`, and `to-eo-tree.xsl:311` tests `@star` before it ever looks at `@base`, so the restored name never reaches the output. The alias survives as well, since both references still counted as references while the alias pass ran, so `format -Deo.autoFix` turns a clean source into one that `unused-alias` rejects — the outcome `restore-aliases.xsl` names in its own docblock as the reason it exists.

Narrowing the match at `tuples-to-stars.xsl:41` so it skips a node whose base is the target of an `+alias` the program declares is enough: with that one lowering skipped, the same program prints back as `empty > a` and `empty > b` with the alias still carrying its weight. The template arrived with #5589, which had in mind only the base that a `*` had produced.

@yegor256 please take a look.

morphqdd:
tuples-to-stars.xsl matches every o[@base = 'Φ.tuple.empty'] and marks
it as a star, on the assumption that only the * shorthand can carry
that base. It cannot: resolve-aliases.xsl puts the very same base on a
bare reference whenever the program declares +alias tuple.empty, so an
ordinary reference to that object is printed as * rather than the name
the author wrote.

+alias tuple.empty
+package foo

[] > app
  empty > a
  empty > b

Printing this program's XMIR through Xmir.FOR_EO gives * > a and
* > b, and +alias tuple.empty stays in the preamble with nothing left
referring to it. restore-aliases.xsl does put the short name back
(it runs after tuples-to-stars in the same train), but to-eo-tree.xsl
tests @star before it ever looks at @base, so the restored name never
reaches the output. The alias survives too, since both references
still counted as references while the alias pass ran, so
format -Deo.autoFix turns a clean source into one that unused-alias
rejects.

Fix: skip a node whose base is the target of an +alias the program
declares, leaving it for restore-aliases.xsl to print back under its
declared short name. Added a print-pack regression covering the exact
program from this report.

Fixes #6797

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/tuples-to-stars.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/tuples-to-stars.xsl
@@ -37,8 +37,17 @@
   An empty tuple is stored as the bare "Φ.tuple.empty" base. Render it as
   the "*" star shorthand with no elements, mirroring how non-empty tuples
   are lowered to stars above.
+
+  The same base is what "resolve-aliases.xsl" puts on an ordinary
+  reference whenever the program declares "+alias tuple.empty", so a node
+  carrying it is not necessarily a "*" in the source — it may just be a
+  name the author wrote that happens to resolve to this base. Lowering it
+  to a star anyway would print the wrong shorthand and leave the alias
+  meta dangling with nothing left referring to it. Skip a node whose base
+  is the target of such an alias, leaving it for "restore-aliases.xsl" to
+  print back under its declared short name instead.
   -->
-  <xsl:template match="o[@base = 'Φ.tuple.empty']">
+  <xsl:template match="o[@base = 'Φ.tuple.empty' and not(/object/metas/meta[head = 'alias' and part[last()] = 'Φ.tuple.empty'])]">
     <xsl:copy>
       <xsl:apply-templates select="@*"/>
       <xsl:attribute name="star"/>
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4628](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6844, opened 2026-08-14
#6829: fail directory.made when a non-directory occupies the path

### LINKED ISSUE #6829: `directory.made` falsely reports success for existing regular file paths
## What happens

`directory.made` treats any existing filesystem entry as a successfully made
directory. If the target path already names a regular file, it returns that
object without checking its kind:

```eo
d.exists.if > @
  d
  # create the directory
```

The following EO-level scenario is reproducible on current `master`
(`4a4f9f4c49`):

```eo
# `path` is an existing regular temporary file
path.as-dir.made.is-directory
```

It evaluates to `false` rather than failing, even though `made` has reported
success. I also reproduced it through the runtime API with a freshly created
regular temporary file; the result of `directory.made(...).is-directory` was
`false`.

## Expected behavior

`directory.made` should return a directory. If a regular file already occupies
the requested path, it cannot satisfy that contract and should fail with a
clear EO error instead of returning the file decorated as a directory.

## Suggested fix

When `d.exists` is true, check `d.is-directory` before returning `d`. Keep the
current return path for an existing directory and fail explicitly for any other
entry type.

gemshrine:
Fixes #6829.

When the target path already existed, `directory.made` returned it as-is
without checking its kind, so a regular file occupying the path was reported as
a successfully made directory: `path.as-dir.made.is-directory` evaluated to
`false` even though `made` claimed success.

The existing-path branch now returns `d` only when `d.is-directory`, and
terminates with a clear error otherwise. The creation branch for a missing path
is unchanged.

The existing `can-make-a-new-directory` example still creates and verifies a
fresh directory, and a new negative example covers a path already occupied by a
regular file.

### DIFF
--- a/eo-runtime/src/main/eo/directory/made.eo
+++ b/eo-runtime/src/main/eo/directory/made.eo
@@ -1,5 +1,8 @@
 # Creates a directory, making missing parent directories as needed. It is a
-# method of `directory`, and `d` is the directory it is taken off.
+# method of `directory`, and `d` is the directory it is taken off. When the
+# path already exists, it must already be a directory: if a regular file (or
+# any non-directory entry) occupies the path, the computation terminates
+# instead of returning that entry decorated as a directory.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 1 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2376](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6845, opened 2026-08-14
#6828: retry partial writes and fail on error in file.open write

### LINKED ISSUE #6828: `file.open(...).write` ignores partial writes and error values
## What happens

The output block returned by `file.open(...).write` invokes the native `write`
operation exactly once and discards its return value:

```eo
code.
  syscall.write
    * descriptor buffer buffer.size
output-block
```

Consequently, both possible non-full outcomes are reported as a successful
output block:

| native `write` result | EO behavior |
| --- | --- |
| `-1` (error) | returns the next output block without reporting the error |
| `0 <= written < buffer.size` | drops the unwritten suffix |

This is observable on POSIX whenever a write is interrupted after transferring
some data, a file-size limit is reached, or the underlying storage accepts only
part of the buffer. `write(2)` explicitly permits successful partial writes.
The console and `stderr` implementations already preserve the remainder and
retry it, while `file.open` does not.

## Expected behavior

`file.open(...).write` must either write every byte in the supplied buffer or
raise an EO failure. A partial native write must continue with the unconsumed
suffix; a `-1` result must not be converted to success.

## Suggested fix

Make the file output block follow the same pattern as `console.write` and
`stderr`: check for `-1`, then recursively write `buffer.slice written
(buffer.size.minus written)` until the whole buffer is accepted.

Reference for the POSIX behavior:
https://man7.org/linux/man-pages/man2/write.2.html

gemshrine:
Fixes #6828.

The output block returned by `file.open(...).write` invoked `syscall.write`
once and discarded its result, so a `-1` error was reported as success and a
partial write silently dropped the unwritten suffix.

The block now follows the same pattern the `console` and `stderr` output blocks
already use: a `-1` result raises an EO failure, and when fewer than
`buffer.size` bytes are written it continues with
`buffer.slice written (buffer.size.minus written)` until the whole buffer is
accepted.

The existing write-and-size examples (`can-append-data-to-a-file` and the
`can-keep-the-size-*` ones) exercise the full-buffer path and stay green.

### DIFF
--- a/eo-runtime/src/main/eo/file.eo
+++ b/eo-runtime/src/main/eo/file.eo
@@ -161,14 +161,26 @@
             output-block.write buffer > @
             [] > output-block
               true > @
-              writable.not.if > [buffer] > write
-                T
-                  "Can't write to file with provided access mode '%s'".printf
-                    * access
-                seq *
-                  code.
-                    syscall.write
-                      * descriptor buffer buffer.size
+              [buffer] > write
+                writable.not.if > @
+                  T
+                    "Can't write to file with provided access mode '%s'".printf
+                      * access
+                  seq *
+                    if. >>!
+                      code.eq -1
+                      T
+                        "Failed to write to file"
+                      output-block
+                    code
+                    remainder
+                code. > code
+                  syscall.write
+                    * descriptor buffer buffer.size
+                if. > remainder
+                  code.lt buffer.size
+                  output-block.write
+                    buffer.slice code (buffer.size.minus code)
                   output-block
         truncate.if syscall.trunc 0 > trunc
     and. > readable


yegor256:
@gemshrine what about a test to reproduce this bug?

gemshrine:
I could not find a way to reproduce it deterministically at the EO level. Here `write` calls `syscall.write` directly, and a regular-file descriptor never returns a short count on a normal write, so there is no seam to inject a partial one (unlike `socket.as-output`, which is parametrized by `[send]` and can take a mock). The same retry loop is exercised by the socket test I added in #6849, and `console`/`stderr` already carry this identical pattern without a unit test. If you know a stable way to force a short `write` on a file fd, I am happy to add the case.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2440](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6846, opened 2026-08-14
#6832: reject non-integer ports in socket.checked-port

### LINKED ISSUE #6832: `socket.checked-port` allows fractional port values, resulting in silent truncation before socket operations
## What happens

`socket.checked-port` checks only numeric bounds:

```eo
or.
  port.lt 0
  port.gt 65535
```

It does not require an integer. The accepted value subsequently reaches
`htons`, where `port.as-i32` truncates it toward zero.

I reproduced this on current `master` (`4a4f9f4c49`) through the public helper:

| input port | `socket.htons` result | effective port |
| --- | --- | --- |
| `0.5` | `00-00` | `0` |
| `1.5` | `01-00` | `1` |
| `65535.5` | `FF-FF` | `65535` |

Thus `socket "127.0.0.1" 1.5` attempts to connect or listen on port 1 while
silently accepting a different value from the caller.

## Expected behavior

A TCP/UDP port must be a finite integer in the inclusive range 0–65535.
Fractional, NaN, and infinite values should be rejected before any socket
operation is attempted.

## Suggested fix

Extend `checked-port` with `port.is-integer` (which also excludes NaN and both
infinities) and report an explicit invalid-port error when the check fails.

gemshrine:
Fixes #6832.

`socket.checked-port` validated only the numeric bounds, not that the port was
an integer. A fractional value therefore passed through to `htons`, where
`port.as-i32` truncated it toward zero: `socket "127.0.0.1" 1.5` silently
connected on port 1.

The check now rejects any non-integer port first, with its own error message.
`is-integer` also returns false for NaN and both infinities, so those are
rejected too. The existing 0-65535 range check is unchanged and now runs only
for integer ports, keeping its `%d` message well-formed.

A new negative example covers a fractional port, alongside the existing
out-of-range one.

### DIFF
--- a/eo-runtime/src/main/eo/socket.eo
+++ b/eo-runtime/src/main/eo/socket.eo
@@ -258,14 +258,18 @@
         seq * > [buffer] > write
           send buffer
           output-block
-  if. > [] > checked-port
-    or.
-      port.lt 0
-      port.gt 65535
+  port.is-integer.not.if > [] > checked-port
     T
-      "Port %d is out of range; it must be 0 to 65535".printf
+      "Port must be an integer in 0 to 65535, but was %f".printf
         * port
-    port
+    if.
+      or.
+        port.lt 0
+        port.gt 65535
+      T
+        "Port %d is out of range; it must be 0 to 65535".printf
+          * port
+      port
   [port] > htons
     as-i16. > @
       or.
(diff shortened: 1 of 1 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (21) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3734](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6847, opened 2026-08-14
#6833: reject negative recv size on both platforms

### LINKED ISSUE #6833: `socket.recv` passes negative size values, causing a Java `NegativeArraySizeException`
## What happens

`socket.recv` accepts its `size` without validation. Both native wrappers turn
it into an array length immediately:

```java
```

Therefore `socket.recv -1` leaks a Java implementation exception instead of an
EO failure. On current `master` (`4a4f9f4c49`), the POSIX wrapper reproduces as:

```text
  at org.eolang.posix.RecvSyscall.make(RecvSyscall.java:36)
```

`RecvFuncCall` has the same unguarded allocation on Windows. In contrast,
`ReadSyscall` and `ReadFuncCall` already reject a negative size explicitly with
an `ExFailure` before allocating.

## Expected behavior

Negative receive sizes should be rejected as a normal EO failure, with a clear
message, on both supported platforms. They must not expose a Java
`NegativeArraySizeException`.

## Suggested fix

Add the same `size < 0` guard used by the read syscall wrappers to both
`RecvSyscall.make()` and `RecvFuncCall.make()`, before allocating the buffer.

gemshrine:
Fixes #6833.

`RecvSyscall.make()` (POSIX) and `RecvFuncCall.make()` (Windows) allocated
`new byte[size]` directly from the requested size, so `socket.recv -1` leaked a
`java.lang.NegativeArraySizeException` instead of a normal EO failure.

Both wrappers now reject a negative size before allocating, with an explicit
`ExFailure`, exactly as `ReadSyscall` and `ReadFuncCall` already do. Added a
platform-guarded regression to each test that asserts an `ExFailure` for a
negative recv size.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/posix/RecvSyscall.java
+++ b/eo-runtime/src/main/java/org/eolang/posix/RecvSyscall.java
@@ -7,6 +7,7 @@ package org.eolang.posix;
 import java.util.Arrays;
 import org.eolang.Data;
 import org.eolang.Dataized;
+import org.eolang.ExFailure;
 import org.eolang.Phi;
 import org.eolang.Syscall;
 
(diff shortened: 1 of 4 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2392](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6849, opened 2026-08-14
#6831: retry short sends in socket.as-output write

### LINKED ISSUE #6831: `socket.as-output.write` incorrectly discards remaining data after partial send
## What happens

`socket.as-output.write` performs a single `send` and immediately returns the
next output block:

```eo
seq * > [buffer] > write
  send buffer
  output-block
```

The return value of `send` is ignored. A successful `send` returns the number
of bytes accepted, which may be less than `buffer.size`; in that case the
remaining bytes are silently discarded while EO reports a successful write.

`send(..., flags = 0)` has the same partial-write semantics as `write`; it can
return a short positive count, and `-1` signals a local error. The public
`as-output` API, however, promises an output that writes the supplied buffer.

## Expected behavior

An output write should either deliver all bytes in the buffer or fail. It must
continue sending the suffix after a short successful send and must propagate a
`-1` result as an EO failure.

## Suggested fix

Make `socket.as-output` retain the number returned by `send`. On a positive
short count, recursively call `send` for the remaining slice; on `-1`, invoke
an error fallback with the socket error message. This mirrors the existing
remainder handling in `console.write` and `stderr`.

Reference for `send` return values:
https://man7.org/linux/man-pages/man2/send.2.html

gemshrine:
Fixes #6831.

`socket.as-output.write` performed a single `send` and returned the next output
block, ignoring the count. A successful `send` may accept fewer than
`buffer.size` bytes, so the unsent suffix was silently dropped while the write
reported success.

The output block now keeps the number returned by `send` and, when it is less
than `buffer.size`, continues with `buffer.slice sent (buffer.size.minus sent)`
until the whole buffer is delivered. This mirrors the remainder handling in
`console.write` and `stderr`. A `-1` result still fails through the existing
`cant-send` path of the `send` helper.

### DIFF
--- a/eo-runtime/src/main/eo/socket.eo
+++ b/eo-runtime/src/main/eo/socket.eo
@@ -253,9 +253,17 @@
         output-block.write buffer
       [] > output-block
         true > @
-        seq * > [buffer] > write
-          send buffer
-          output-block
+        [buffer] > write
+          seq > @
+            * sent remainder
+          if. > remainder
+            sent.lt buffer.size
+            output-block.write
+              buffer.slice
+                sent
+                buffer.size.minus sent
+            output-block
+          send buffer > sent
   port.is-integer.not.if > [] > checked-port
     T
       "Port must be an integer in 0 to 65535, but was %f".printf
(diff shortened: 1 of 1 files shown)


yegor256:
@gemshrine why no test?

gemshrine:
Added one in 72989c6478: `can-resend-the-remaining-bytes-after-a-short-send`. It wires `as-output` to a mock `send` that delivers a single byte per call and records each byte into a 2-byte malloc chunk. Against the old code only the first byte (`01`) lands and the assertion fails; with the retry loop both bytes (`01-02`) end up written, so the test goes red on master and green here.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2452](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6850, opened 2026-08-14
#6826: return zero from Ladder.described for an empty program

### LINKED ISSUE #6826: `Ladder.described()` returns `NoSuchElementException` on zero object inference
## What happens

`Ladder.described()` throws `NoSuchElementException` when the inference result
contains no objects. This makes the newly added empty-program case fail before
the method can report its documented zero share.

On current `master` (`4a4f9f4c49`), this minimal Java program reproduces it:

```java
final Ladder ladder = new Ladder(new LinkedHashMap<>());
System.out.println(ladder.total());
System.out.println(ladder.described());
```

It prints `0` and then fails with:

```text
java.util.NoSuchElementException
  at org.eolang.inference.Ladder.described(Ladder.java:57)
```

## Why it happens

`described()` evaluates `this.counts.values().iterator().next()` unconditionally
to find the count for the first rung. For an empty map there is no first value.
The subsequent `share()` helper already explicitly handles an empty total and
returns `0.0d`, but that code is never reached.

## Expected behavior

The empty program should be a valid measurement: `total()` and `described()`
should both return zero, without an exception. This is also consistent with the
`share()` contract, which says it returns zero when there are no objects.

## Suggested fix

Make `described()` detect an empty `counts` map before accessing its first
value, returning `0.0d`; alternatively, initialise `Ladder` with the complete
five-rung map in every supported construction path and enforce that invariant
in its constructor.

gemshrine:
Fixes #6826.

`Ladder.described()` called `this.counts.values().iterator().next()`
unconditionally, so an empty `counts` map (a program with no inferred objects)
threw `NoSuchElementException` instead of reporting a zero share. The
`share()` helper already returns `0.0d` when the total is zero, but that code
was never reached.

`described()` now returns `0.0d` for an empty map before touching the first
value, matching `share()`'s zero-object contract and the existing empty-program
`percent()` behaviour. Added a regression that builds a `Ladder` from an empty
map and asserts `described()` returns zero without throwing.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Ladder.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Ladder.java
@@ -54,7 +54,13 @@ public final class Ladder {
      * @return The share, out of a hundred
      */
     public double described() {
-        return this.share(this.total() - this.counts.values().iterator().next());
+        final double result;
+        if (this.counts.isEmpty()) {
+            result = 0.0d;
+        } else {
+            result = this.share(this.total() - this.counts.values().iterator().next());
+        }
+        return result;
     }
 
     /**
--- a/eo-inference/src/test/java/org/eolang/inference/LadderTest.java
+++ b/eo-inference/src/test/java/org/eolang/inference/LadderTest.java
@@ -52,4 +52,13 @@ final class LadderTest {
             Matchers.closeTo(0.0d, 0.001d)
         );
     }
+
+    @Test
+    void describesAnEmptyProgramAsZero() {
+        MatcherAssert.assertThat(
+            "a program with no rungs at all has a zero described share, but it threw",
+            new Ladder(new LinkedHashMap<>(0)).described(),
+            Matchers.closeTo(0.0d, 0.001d)
+        );
+    }
 }


yegor256:
@gemshrine what is the effect of this change, for our YAML fixtures? maybe we should add a new fixture to demonstrate the effect?

gemshrine:
I looked into it and a fixture would not reach the change. `Depth.ladder()` seeds all five rungs at zero before it counts anything, so the map it hands to `Ladder` always has five keys, never an empty one, even for a program with no objects. On that map `described()` already returned zero through `share()`, whose `total() == 0` branch guards it; the empty-map branch this PR adds is only entered when `Ladder` is built with a literally empty map. On top of that, the packs assert only the `provides`, `needs` and `links` tables, while `total`/`described`/`percent` are a log line that no pack looks at, so a new YAML fixture could not observe this at all. That is why the fix comes with `LadderTest.describesAnEmptyProgramAsZero`, which builds `new Ladder(emptyMap)` directly and pins the contract at the one level where the empty map is reachable. Happy to add a fixture if you want one for the tables themselves, but it would not exercise `described()`.

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2464](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6851, opened 2026-08-14
Make a copy cost what it reads, not what it holds

maxonfjvipon:
Copying an object rebuilt every attribute it had, so a wide object paid for all of them even when the copy went on to read one. `copy()` no longer walks the map:

```java
final PhDefault copy = (PhDefault) this.clone();
copy.attrs = new CopiedAttrs(this.loaded(), copy);
return copy;
```

`CopiedAttrs` takes an attribute out of the origin when something asks for it, and keeps it thereafter, so a copy costs what it reads rather than what it holds.

Two more changes come with it. `PhDefault` keeps its attributes and their order in `Bindings`, a small list-backed map, instead of three `HashMap`s — those objects hold two or three attributes each, and a table with a node per entry and a hash of every name on every read is a poor trade at that size. And `AtOnce` and `PhOnce` guard their cache with the object's own monitor rather than a `ReentrantLock` of their own, which was two more objects on every attribute of every object.

That last one needs a word. PMD enables both `AvoidSynchronizedStatement` and `AvoidSynchronizedAtMethodLevel`, so it leaves no way to use a monitor at all, and the alternative it asks for is exactly the lock being removed. The two methods therefore carry a `@SuppressWarnings("PMD.AvoidSynchronizedAtMethodLevel")`. It is worth about a quarter of the total saving, measured in the comment below.

Nothing here changes what an object does, only what it costs to make one. Every field stays final and both new classes are final; `Bindings` is mutable exactly where the `HashMap` it replaces was, and `PhDefault.attrs` was already assigned outside the constructor before this change.

Measurements are in a comment below, including what this does to #6789.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/AtOnce.java
+++ b/eo-runtime/src/main/java/org/eolang/AtOnce.java
@@ -5,7 +5,6 @@
 package org.eolang;
 
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * Attribute that retrieves object only once.
@@ -27,18 +26,12 @@ public final class AtOnce implements Attribute {
     private final AtomicReference<Phi> cached;
 
     /**
-     * Reentrant lock for thread-safe initialization.
-     */
-    private final ReentrantLock lock;
-
-    /**
      * Ctor.
      * @param attr Origin attribute
      */
     public AtOnce(final Attribute attr) {
         this.origin = attr;
         this.cached = new AtomicReference<>(null);
-        this.lock = new ReentrantLock();
     }
 
     @Override
@@ -47,18 +40,16 @@ public final class AtOnce implements Attribute {
     }
 
     @Override
+    @SuppressWarnings({"PMD.AvoidSynchronizedStatement", "PMD.DoubleCheckedLocking"})
     public Phi get() {
         Phi result = this.cached.get();
         if (result == null) {
-            this.lock.lock();
-            try {
+            synchronized (this.cached) {
                 result = this.cached.get();
                 if (result == null) {
                     result = this.origin.get();
                     this.cached.set(result);
                 }
-            } finally {
-                this.lock.unlock();
             }
         }
         return result;
--- a/eo-runtime/src/main/java/org/eolang/PhOnce.java
+++ b/eo-runtime/src/main/java/org/eolang/PhOnce.java
@@ -6,7 +6,6 @@
 package org.eolang;
 
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
 
 /**
@@ -25,11 +24,6 @@ public class PhOnce implements Phi {
      * Reference.
      */
     private final AtomicReference<Phi> ref;
-
-    /**
-     * Lock for thread-safe initialization.
-     */
-    private final ReentrantLock lock;
 
     /**
      * Supplier of the φ-term, or {@code null} to delegate to the wrapped object.
@@ -51,21 +45,8 @@ public class PhOnce implements Phi {
      */
     public PhOnce(final Supplier<Phi> obj, final Supplier<String> term) {
         this.ref = new AtomicReference<>(null);
-        this.lock = new ReentrantLock();
         this.term = term;
-        this.object = () -> {
-            if (this.ref.get() == null) {
-                this.lock.lock();
-                try {
-                    if (this.ref.get() == null) {
-                        this.ref.set(obj.get());
-                    }
-                } finally {
-                    this.lock.unlock();
-                }
-            }
-            return this.ref.get();
-        };
+        this.object = () -> this.loaded(obj);
     }
 
     @Override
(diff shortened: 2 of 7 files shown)


maxonfjvipon:
Numbers behind this change, all taken on one machine.

### What it does to #6789

README snippet 5 — the four-iteration multiplication loop the snippet test runs — with the `number` package merged. The metric is the smallest heap at which it still finishes. A run that completes takes about a minute, so anything still going at 150s is thrashing and counts as a failure, which is what a CI job would see.

| heap | master | this PR |
| --- | --- | --- |
| 8g | ok, 61s | ok |
| 6g | ok, 72s | ok |
| 4g | thrash | ok, 60s |
| 3g | thrash | **ok, 62s** |
| 2g | thrash | thrash |

Merged `number` needs **6g** on master and **3g** with this change. CI gives the snippet test 8g, so the merge goes from 2g of headroom to 5g.

Two caveats on where those figures come from. CI has never built #6789 — it is `CONFLICTING`, so GitHub cannot make a merge ref and skips the build workflows, which is why its six green checks are all static analysis; the figures above come from that PR's own EO sources reconstructed onto current master. And wall time is untouched: snippet 5 is already a minute-long workload on master today, merged or not.

### On a loop, at equal work

A `while` loop sampled at the same iteration count (~106), with the heap histogram taken after a forced full GC so it reports the live set rather than garbage:

| | MB/iter | objects/iter | vs master |
| --- | --- | --- | --- |
| master | 7.60 | 310,054 | — |
| lazy `copy()` + list-backed attributes | 5.26 | 222,381 | −30.8% |
| + the two locks removed | **4.43** | **190,116** | **−41.7%** |

The locks are worth about a quarter of the total: 15.8% on top of everything else, or 11 points of the 41.7%. They are a `ReentrantLock` and its `NonfairSync`, two objects on every attribute of every object — 28,718 of each per iteration, 139MB of master's 805MB live set. On the merged-`number` heap they are what separates 3g from 4g.

The lazy copy is the largest single step, and its benefit grows with the width of the object, which is why merged `number` gains more than this loop does. Merging takes `number` from 8 attributes to 54, and an eager `copy()` pays for all of them on the most-instantiated forma in any program.

For scale, on master that loop retained 2,074 EO objects per iteration but 310,054 Java objects — about 150 Java objects of bookkeeping per EO object.

### Trades worth naming

`Bindings` keeps its names and attributes in two `List`s rather than two bare arrays with a size counter. The arrays are about 10% cheaper again (4.77 MB/iter against 5.26 at that stage), but they cannot be `final` fields, since they have to be replaced when they grow. Final fields are worth the 10%.

Dropping the `AtomicReference`s in `AtOnce` and `PhOnce` too would need a non-final volatile field, and measured within run-to-run noise, so they stay.

Still on the table, per iteration: `AtomicReference` 52.5MB, the `PhDispatch` trio 45MB, `PhSafe` 32.8MB (that one is #5435), `AtWithRho` 32.0MB.

None of this touches the other half of the problem. `while` is recursive with no tail-call elimination, so a long loop still retains every frame, and this change only moves the constant in front of an O(N). Snippet 5 loops four times, so for #6789 the constant is exactly what matters.

eo-runtime/src/main/java/org/eolang/AtOnce.java
```diff
@@ -47,21 +40,12 @@ public Attribute copy(final Phi self) {
...
-            try {
-                result = this.cached.get();
-                if (result == null) {
-                    result = this.origin.get();
-                    this.cached.set(result);
-                }
-            } finally {
-                this.lock.unlock();
-            }
+    @SuppressWarnings("PMD.AvoidSynchronizedAtMethodLevel")
```
yegor256: @maxonfjvipon fix this, use `synchronized` block inside the method

eo-runtime/src/main/java/org/eolang/PhOnce.java
```diff
@@ -143,4 +124,17 @@ public Phi normalized() {
         }
         return result;
     }
+
+    /**
+     * The wrapped object, made by the supplier the first time it is wanted.
+     * @param obj Supplier of the object
+     * @return The object
+     */
+    @SuppressWarnings("PMD.AvoidSynchronizedAtMethodLevel")
```
yegor256: @maxonfjvipon use `synchronized` block inside the method - it's cleaner

yegor256:
@maxonfjvipon looks good, see a few comments above

eo-runtime/src/main/java/org/eolang/AtOnce.java
maxonfjvipon: Done, in 6fd271ef7c. The block is inside the method now, and it locks on `this.cached` rather than `this` — Checkstyle rejects `this` as a lock and asks for a class variable, and the reference is already a final field, so there is no extra object for the monitor.

eo-runtime/src/main/java/org/eolang/PhOnce.java
maxonfjvipon: Done, in 6fd271ef7c — same shape, locking on `this.ref`.

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (467 >= 100); -8 for way too many hits-of-code (467 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+807](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2575](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6852, opened 2026-08-14
#6649: tag the per-line drop-and-continue recovery as lossy

### LINKED ISSUE #6649: `MjFormat.parsed` allows problematic recovery that can omit a binding

morphqdd:
MjFormat.parsed's guard rejects a source only when severe errors
coexist with truncated line coverage or an empty placeholder node.
There is a third kind of lossy recovery it lets through: the parser
can recover from a mid-file syntax error by dropping just the
offending binding and carrying on, so line coverage stays intact and
no placeholder appears.

For example:

```
[x] > foo
  seq * > @
    last
    last.plus 1
  x! > last

[] > bar
  42 > @
```

fails to parse at [5:5] with "trailing garbage after name suffix" (the
const form has to be x > last!, not x! > last). In the recovered tree
the maximum @line equals the source's last non-blank line, so
truncated() is FALSE; there is no empty <o/>, so placeholder() is
FALSE. Yet the binding name="last" is gone entirely and its two
references escape to the global base="Phi.last". With -Deo.autoFix the
line is erased from the file on disk and the two references are left
dangling.

The gap is that severity="error" conflates cosmetic layout recoveries
(what this goal exists to fix) with genuinely lossy ones. Rejecting
every <errors> entry would also reject the goal's own core use case.

Fix: Eo.dispatch's per-line ParseError catch is the one recovery site
that drops a whole construct and moves on. Emit.error gained a lossy
overload that tags its <error> element with a new XMIR lossy attribute
(schema updated), and only that one call site passes true. No other
existing emit.error() call site was touched, so every other recovery
(cosmetic or otherwise) keeps its current behavior exactly. MjFormat
now also rejects a tree carrying an error tagged lossy, alongside the
existing truncated/placeholder checks.

Fixes #6649

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
@@ -216,6 +216,18 @@ public final class MjFormat extends MjSafe {
      * yet the tree no longer describes the source, so this also rejects a
      * tree carrying such a {@link #placeholder(XML) placeholder}.</p>
      *
+     * <p>Neither heuristic catches every lossy recovery: the parser can
+     * also drop a single binding line entirely and carry straight on to
+     * its siblings, which neither truncates line coverage (later siblings
+     * still reach the source's last line) nor leaves a placeholder (there
+     * is simply nothing where the binding used to be) — see #6649. That
+     * one recovery site tags its own {@code <error>} with {@code lossy}
+     * (see {@link Emit#error(int, int, String, boolean)}), so this also
+     * rejects a tree carrying such an error, without having to reject
+     * every {@code <errors>} entry the way the issue that reported this
+     * gap first suggested, which would also reject this goal's own core
+     * use case of recovering from purely cosmetic layout violations.</p>
+     *
      * @param source The path of the {@code .eo} file, for the error message
      * @param structure The current text of the {@code .eo} file
      * @return The parsed XMIR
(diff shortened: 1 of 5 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (156 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4668](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6805, opened 2026-08-14
`eo-inference` unable to identify an object as a datum, leads to unknown objects in `eo-runtime`


maxonfjvipon:
The largest thing `eo-inference` does not know about `eo-runtime` is data. Of the 7,760 objects it can say nothing at all about, **7,203 are Δ literals** — 93% of the whole unknown, and the tables have no way to write one down.

Every one of them is a leaf carrying bytes, sitting under the `Φ.bytes` that a literal parses into:

```xml
<o base="Φ.bytes" loc="Φ.bool.φ.α0">
  <o loc="Φ.bool.φ.α0.α0">01-</o>
</o>
```

`Links` writes a row for an object whose base names another object, and that leaf has no base, so no rule ever looks at it. The row it deserves is the one `eo-inference-spec.md` §8.3 already reserves:

```xml
<type id="Φ.bool.φ.α0.α0">
  <data/>
</type>
```

This is the first row that is not a copy of anything, which is the whole reason `copy="…"` became `<ref loc="…"/>` in #6786. A datum is not a copy of `Φ.bytes` and must not be written as one: `Φ.bytes` is a formation with a void, while `01-` is the ground itself, and saying "a copy of `Φ.bytes`" would put a free void on an object that has none.

Two things have to move together. The rule writes the rows, and `Rung` learns that a datum is known — otherwise the objects stay on rung 0 and the ladder from #6801 will keep reporting them as unknown while the table describes them. Which rung a datum stands on is worth deciding in the ticket rather than assuming: there is nothing further to find out about a literal, so it belongs at the top, and the top rung is currently named after formations.

Measured on a clean `eo-runtime` build of master: 7,203 rows, depth from 48.8% to about 61%, and the share of objects described from 82.0% to 98.7%. After it, 557 objects — 1.3% — are all that is left unknown: 433 dispatches whose receiver nothing describes, 66 terminations, and 58 names in scope that resolve to nothing.

One caveat for whoever writes it. `Types` takes the copies as one map and the binds as another, and data would be a third parallel map keyed by the same locators. That is already one map too many for a document with one row per object, and the next form after this one — ⊥, then unions — will make it four. Worth its own issue rather than a fourth map.

VasilevNStas:
@yegor256 @maxonfjvipon 

This can be closed: the fix is merged.

[...]

PR #6853, opened 2026-08-14
Hold a type per row instead of a collection per form

maxonfjvipon:
`links.xml` has one row per object and every row holds one type. `Types` held three collections keyed by the same locators instead — the pairs, the binds, and the data — and the four remaining forms in the spec were each going to be another constructor argument.

A row now holds a `Type`. `Ref` is a copy and what it has put into the voids, `Data` is the ground a literal stands on, and a new form is a new class:

```java
final class Data implements Type {
    @Override
    public Directives directives() {
        return new Directives().add("data").up();
    }
}
```

The counting was the smaller half of the problem. `Resolved` reads the table and writes it again, and it can only rebuild the forms it knows about, so anything else had to be read back by name or it disappeared from the document with no test failing:

```java
new Types(pairs, binds, written.data())
```

Forget that third argument and every datum is gone on the second pass. `Kept` holds a row exactly as it was found, so the pass writes back what it does not rebuild without knowing what it is — which is what makes ⊥, `<unknown/>`, unions and variables cost one class each from here on rather than one class plus one thing to remember.

`Refs` turns the pairs and the binds into rows, since neither is any use without the other: a copy nobody has given arguments to and a saturated one are the same pair and different types.

The tables of `eo-runtime` come out **byte-identical** to the build before this — same 39,054 rows, same order, same md5. Two new tests cover the round trip, one of which fails against the first version of `Kept` I wrote.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Data.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Data.java
@@ -1,0 +1,25 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import org.xembly.Directives;
+
+/**
+ * An object that is a datum.
+ *
+ * <p>The bytes of a literal are the ground the whole program stands on, so
+ * this carries nothing: there is nothing to know about {@code 01-} beyond
+ * that it is what it is, and the bytes themselves are in the XMIR the row
+ * points at.</p>
+ *
+ * @since 0.69.0
+ */
+final class Data implements Type {
+
+    @Override
+    public Directives directives() {
+        return new Directives().add("data").up();
+    }
+}
--- a/eo-inference/src/main/java/org/eolang/inference/Kept.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Kept.java
@@ -1,0 +1,48 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.xml.XML;
+import org.xembly.Directives;
+
+/**
+ * A type read from the table and written back as it was.
+ *
+ * <p>A pass that works out more than the table says has to write the table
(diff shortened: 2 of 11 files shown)


yegor256:
@maxonfjvipon what is the effect of this change? why no YAML updates?

maxonfjvipon:
No effect on the tables at all, and that is the point rather than a caveat: `links.xml` for `eo-runtime` comes out byte-identical to the build before this one — 39,054 rows, same order, same md5.

The effect is on what the forms still to come cost. `Types` held one collection per form, keyed by the same locators — the pairs, the binds, and, since #6805, the data — while the document it writes has one row per object. `<bottom/>`, `<unknown/>`, `<union>` and `<var>` are all still owed, and each was going to be another constructor argument.

The sharper problem was what that did to `Resolved`. It reads the table and writes it again, and it can only rebuild the forms it knows about, so every form it does not know about had to be read back by name or it vanished from the document with nothing failing:

```java
new Types(pairs, binds, written.data())
```

Drop that third argument and every datum is gone on the second pass. `Kept` holds a row as it was found, so a pass writes back what it does not rebuild without knowing what it is. A new form is now one class and nothing else.

On the YAML: a pack describes what the tables say about a program, and this change says nothing new about any program. The behaviour it protects is already described by `datum-stays-through-dispatch-resolution.yaml`, which was added in #6805 for this exact hazard. I checked rather than assumed — with `Kept` emptied out, that pack and `datum-is-the-ground-of-a-literal.yaml` both fail.

The one case no pack can reach is a row form that nothing writes yet, since no EO source can produce one. `KeptTest` and `PairsTest` cover it by feeding a `<union>` and a `<bottom/>` through the reader and the writer.

maxonfjvipon:
@yegor256 check above, that's more refactoring + preparation for union types than significant change

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (431 >= 100); -8 for way too many hits-of-code (431 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+803](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6854, opened 2026-08-14
#6848: reject horizontal args on a group wrapping an inline-phi

### LINKED ISSUE #6848: Incorrect argument placement in parenthesized inline-phi formation leading to rejected XMIR emission
## What happens

`(m > [m]) 5` is accepted by the parser, but the arguments end up *inside* the formation rather than applied to it:

```eo
[] > app
  (m > [m]) 5 > y
```

```xml
<o name="y">
  <o base="∅" name="m"/>
  <o base="ξ.m" name="φ"/>
  <o base="Φ.number" loc="Φ.app.y.α2">...</o>   <!-- no @name, no @as -->
</o>
```

`Emissions.inlinePhi` (`eo-parser/src/main/java/org/eolang/parser/Emissions.java:757`) opens a base-less `<o>` for the formation and never closes it, so the loop in `LnApplication.emit` (`eo-parser/src/main/java/org/eolang/parser/LnApplication.java:192`) appends every argument as a child of that formation. Because the parent has no `@base`, `mandatory-as` sees a formation body rather than an argument list and leaves the child without `@as`.

## Why this is incorrect

A base-less `<o>` is a formation, and every child of a formation is an attribute, so a child with neither `@name` nor `@as` is not representable. The parser enforces exactly that rule on input — feeding this XMIR through `eo-printer` yields

```eo
[m] > y
  m > @
  5
```

which the parser then rejects with `object inside formation must have a name`. So the parser accepts a program it cannot round-trip, and emits a tree that violates the rule it enforces everywhere else. The sibling spellings are already rejected: `(foo) 5` gives `redundant parentheses around a single token` and `([x]) 5` gives `horizontal formation not allowed as argument`. Only the inline-phi flavour slips through. Reproduced on `2e57a2752d` with no local changes.

## Suggested fix

Smallest change: make a `GROUP` head that wraps an inline-phi reject a non-empty argument list in `LnApplication.into`, with the same positioned error `([x]) 5` already produces. The alternative — emitting a real application node around the formation so the arguments attach outside it — is what would make the construct meaningful, but it is a larger change and worth deciding on separately.

Related to #6707, which reports the same "head accepts arguments it cannot carry" shape for the `T` bottom term.

morphqdd:
`(m > [m]) 5` is accepted by the parser, but the arguments end up
inside the formation rather than applied to it:

```eo
[] > app
  (m > [m]) 5 > y
```

```xml
<o name="y">
  <o base="?" name="m"/>
  <o base="xi.m" name="phi"/>
  <o base="Phi.number" loc="Phi.app.y.a2">...</o>   <!-- no @name, no @as -->
</o>
```

`Emissions.inlinePhi` opens a base-less `<o>` for the formation and never
closes it, so the loop in `LnApplication.emit` appends every argument
as a child of that formation. Because the parent has no `@base`,
`mandatory-as` sees a formation body rather than an argument list and
leaves the child without `@as`.

A base-less `<o>` is a formation, and every child of a formation is an
attribute, so a child with neither `@name` nor `@as` is not
representable. The parser enforces exactly that rule on input, so it
accepts a program it cannot round-trip through eo-printer: printing
the tree back yields

```eo
[m] > y
  m > @
  5
```

which the parser then rejects with "object inside formation must have
a name". The sibling spellings are already rejected: `(foo) 5` gives
"redundant parentheses around a single token" and `([x]) 5` gives
"horizontal formation not allowed as argument". Only the inline-phi
flavour slipped through.

Fix (the smallest change the issue itself suggests): reject a
non-empty argument list on a group head whose own content is a
top-level inline-phi expression, with the same positioned error
`([x]) 5` already produces.

Fixes #6848

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnApplication.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnApplication.java
@@ -72,13 +72,7 @@ final class LnApplication implements Line {
         } else {
             Blanks.checkPlain(this.span, globals, emit);
         }
-        if (head.kind() == Value.Kind.GROUP
-            && chain.isEmpty() && args.isEmpty() && outer == null) {
-            throw new ParseError(
-                this.span.line(), this.span.indent(),
-                "redundant parentheses around a top-level expression — drop the outer `(` and `)`"
-            );
-        }
+        this.checkGroupHead(head, chain, args, outer);
         Comments.seal(globals, emit, this.span);
         final Kind kind = LnApplication.classify(head, chain, args);
         final Openness openness;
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4684](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6856, opened 2026-08-14
Drop all +unlint metas from .eo files

### LINKED ISSUE #6855: 99 `+unlint` metas hide project-wide lint policy in 67 `.eo` files
There are 99 `+unlint` metas spread over 67 `.eo` files, 65 of them in
`eo-runtime/src/main/eo` and two in `eo-maven-plugin/src/test/resources`. They
suppress seven rules: `mandatory-package` (50 times, e.g. `pi.eo:8`),
`redundant-object` (29, e.g. `win32.eo:19`), `package-member-without-void` (11,
e.g. `tuple/is-empty.eo:9`), `unit-test-missing` (4, e.g. `input.eo:14`),
`object-has-data` (2), `decorated-formation` (2) and `empty-object` (1).

A meta in the source is the wrong place for most of these. Four of the seven
rules can never be satisfied by the standard library at all: the root objects
live in the root package on purpose, so `mandatory-package` will always fire;
`redundant-object` cannot see a public attribute being used from another file
(objectionary/lints#1276); `package-member-without-void` does not know about the
`mergedPackages` list in `eo-runtime/pom.xml` (objectionary/lints#1273); and
`decorated-formation` forbids exactly the shape `map` and one `bool` test are
built on. Repeating the same suppression 50 times reads as if each file had its
own local excuse, and it hides the fact that the decision is one project-wide
policy. Meanwhile the two fixtures in `eo-maven-plugin` no longer need their
metas at all, and `input` and `output` were never tested.

Delete every `+unlint` meta from the `.eo` files, list the unsatisfiable rules
once in `<skipSourceLints>` in `eo-runtime/pom.xml` next to the existing
`<skipProgramLints>`, with a comment explaining each, and write the missing unit
tests for `input` and `output`.

yegor256:
Every `+unlint` meta is gone from the `.eo` files — 99 of them across 67 files.
Five of the seven rules they suppressed are now listed once in
`<skipSourceLints>` in `eo-runtime/pom.xml`, right next to the existing
`<skipProgramLints>`, with a comment saying why each one cannot be satisfied.
The two fixtures in `eo-maven-plugin/src/test/resources` needed no replacement
at all, and `input` and `output`, which were suppressing `unit-test-missing`,
now have a real test each.

The reason for moving rather than deleting: those five rules can never pass on
the standard library. The root objects live in the root package on purpose, so
`mandatory-package` always fires. `redundant-object` reads one file at a time
and counts only relative references, so `win32.rdonly`, `tuple.empty`,
`posix.stdout` and 71 more public attributes look unused to it — I filed
objectionary/lints#1276 for that. `package-member-without-void` does not know
about the `mergedPackages` list a few lines below it in the same pom
(objectionary/lints#1273). `decorated-formation` forbids exactly the shape
`map` hides its constructor behind. And `unit-test-missing` wants a test in
`stdin`, whose every attribute reads the standard input, so the test blocks on
the `read` syscall until the deadline aborts it.

Worth knowing before you merge: a pom-wide skip is blunter than a per-file
meta, and `redundant-object` in particular was also pointing at about 90 real
single-use bindings inside the runtime that could be inlined. Those are now
invisible until objectionary/lints#1276 lands, so it may be worth a separate
ticket to sweep them.

Closes #6855

### DIFF
--- a/eo-maven-plugin/src/test/resources/org/eolang/maven/hello-world.eo
+++ b/eo-maven-plugin/src/test/resources/org/eolang/maven/hello-world.eo
@@ -4,7 +4,6 @@
 +package foo.x
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
-+unlint object-has-data
 +version 0.0.0
 
 [x] > main
--- a/eo-maven-plugin/src/test/resources/org/eolang/maven/mess.eo
+++ b/eo-maven-plugin/src/test/resources/org/eolang/maven/mess.eo
@@ -5,9 +5,6 @@
 +package foo.x
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
-+unlint empty-object
-+unlint object-has-data
-+unlint unit-test-missing
 +version 0.0.0
 
 [n x] > main
--- a/eo-runtime/pom.xml
+++ b/eo-runtime/pom.xml
@@ -287,6 +287,41 @@
             </goals>
             <configuration>
               <!--
+              These five lints are unsatisfiable for the standard library, so
+              they are turned off here once instead of being suppressed by a
+              '+unlint' meta in every second file:
+
+              'mandatory-package' demands a '+package' meta, while the root
+              objects of the library live in the root package on purpose and
+              the transpiler hardcodes 'org.eolang' for them, so a '+package'
+              would nest them one level too deep.
+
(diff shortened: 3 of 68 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (149 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2568](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6857, opened 2026-08-14
#6816: reject fractional slice bounds before recursion

### LINKED ISSUE #6816: string.slice defers fractional positions to misleading truncated bounds errors
## What happens

`string.slice` accepts fractional character positions into its traversal instead
of rejecting them at the boundary. Since the traversal counter is always an
integer, it can never equal the supplied fraction; the operation reaches the
end of the text and reports a misleading, truncated bounds error.

| expression | actual diagnostic | expected behavior |
| --- | --- | --- |
| `"some".slice 0.5 1` | `Start index (0) is out of string bounds` | immediately reject non-integer start `0.5` |
| `"some".slice 0 1.5` | `Start index + length to slice (1) is out of string bounds` | immediately reject non-integer length `1.5` |
| `"some".slice 0 1` | `"s"` | `"s"` |

Reproduced against current `master` (`4a4f9f4c49`). The third row is a control
case. The diagnostics in the first two rows omit the fractional part because
they use integer formatting, and they identify a different condition from the
one that actually made the request invalid.

## Why it happens

[`string.slice`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/string/slice.eo)
checks only that `start` and `len` are non-negative. It then compares the
numeric inputs with an integer `accum` while walking UTF-8 characters. A
fraction can never match that counter, so validation is deferred until a
spurious out-of-bounds path. The error templates use `%d`, which truncates the
offending values.

## Expected behavior

Character indexes and lengths must be finite non-negative integers. Fractional,
`nan`, and infinite values should fail before walking the string, with a
diagnostic that preserves the supplied value.

## Suggested fix

Validate `nstart` and `nlen` with `is-integer` (which also excludes non-finite
numbers) before entering the recursion. Use `%f` or a dedicated validation
message for rejected values, and add regressions for fractional `start` and
`len`.

morphqdd:
`string.slice` checked only that `start` and `len` were non-negative, then compared the numeric inputs against an integer recursion counter while walking UTF-8 characters. A fraction can never equal that counter, so validation was deferred until a spurious out-of-bounds path at the end of the text, and the error templates used `%d`, which silently truncated the offending values.

```
"some".slice 0.5 1   -> "Start index (0) is out of string bounds"
"some".slice 0 1.5   -> "Start index + length to slice (1) is out of string bounds"
```

Both diagnostics describe the wrong condition and drop the fractional part.

Validated `nstart` and `nlen` with `is-integer` (which also excludes `nan` and infinite values) alongside the existing non-negative check, and switched the error format to `%f` so a rejected fractional value is preserved in the message instead of truncated.

```eo
if. > @
  or.
    nstart.lt 0
    nstart.is-integer.not
  T
    "Start index must be a non-negative integer, but was %f".printf
      * nstart
```

Added `stops-on-a-fractional-start` and `stops-on-a-fractional-length`, reproducing the issue's exact examples. Ran `EOsliceTest` locally, 18/18 passing.

Closes #6816

### DIFF
--- a/eo-runtime/src/main/eo/string/slice.eo
+++ b/eo-runtime/src/main/eo/string/slice.eo
@@ -1,8 +1,10 @@
 # Takes a piece of the `text` as another `text`, counting characters and not
-# bytes. Here `start` must be a non-negative index to start slicing from, while
-# `len` must be a non-negative amount of characters to take. Both of them walk
-# the UTF-8 sequence character by character, so a multi-byte character counts
-# as one. An index outside the bounds of the `text` terminates the computation.
+# bytes. Here `start` must be a non-negative integer index to start slicing
+# from, while `len` must be a non-negative integer amount of characters to
+# take. A fractional, `nan` or infinite `start` or `len` terminates the
+# computation immediately. Both of them walk the UTF-8 sequence character by
+# character, so a multi-byte character counts as one. An index outside the
+# bounds of the `text` terminates the computation.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4696](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6859, opened 2026-08-14
#6820: ignore trailing separators in path.posix basename/dirname

### LINKED ISSUE #6820: `path.posix` incorrectly returns empty basename for paths ending with a separator
## What happens

`path.posix` treats a trailing separator as an empty final name instead of
ignoring it when calculating `basename` and `dirname`.

| path | actual `basename` | actual `dirname` | expected pair |
| --- | --- | --- | --- |
| `"/foo"` | `"foo"` | `"/"` | `"foo"`, `"/"` |
| `"/foo/"` | `""` | `"/foo"` | `"foo"`, `"/"` |
| `"foo/"` | `""` | `"foo"` | `"foo"`, `"."` |

Reproduced against current `master` (`4a4f9f4c49`). The non-trailing row is a
control case. POSIX path utilities and common path APIs ignore trailing
separators when determining the final path component.

## Why it happens

[`path.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/path.eo)
uses the final separator directly: `basename` slices everything after it, and
`dirname` slices everything before it. For a path ending in that separator,
the first slice has length zero while the second still contains the name.

## Expected behavior

Trailing separators should not create an empty basename. `path.posix "/foo/"`
should behave as the equivalent non-trailing path for `basename` and `dirname`.

## Suggested fix

Strip trailing separators before calculating the final separator index, while
preserving the root path as a special case. Add regressions for absolute and
relative paths with a trailing `/`.

morphqdd:
`path.eo`'s `basename` slices everything after the last separator, and `dirname` slices everything before it. For a path ending in that separator, the last separator found is the trailing one itself, so:

```
path.posix("/foo/").basename -> "" (empty, should be "foo")
path.posix("/foo/").dirname  -> "/foo" (should be "/")
path.posix("foo/").dirname   -> "foo" (should be ".")
```

Both disagree with the equivalent non-trailing path (`path.posix("/foo")` already gives the right answer for both).

Added a `trimmed` helper inside `impl`, shared by `basename` and `dirname`, that strips one trailing separator before either method runs its existing logic, so a path ending in a separator behaves exactly like its non-trailing equivalent. The root path (a bare separator) is left unstripped so it does not collapse to an empty string.

```eo
[] > trimmed
  string > @
    if.
      or.
        uri.size.eq 0
        uri.eq separator
      uri
      if.
        uri.ends-with separator
        as-bytes.
          uri.slice
            0
            uri.length.minus separator.length
        uri
```

Two existing tests encoded the same bug for a multi-segment path and asserted the old wrong values (`can-return-an-empty-basename-from-a-path-ending-with-a-separator` expected `""`, `can-return-the-dirname-of-a-directory-path` expected `"var/www"` instead of going up one level to `"var"`). Both are corrected, and I added regressions for the exact single-component cases from the issue, absolute and relative, each with and without the trailing separator.

Ran `EOpathTest` locally, 55/55 passing.

Closes #6820

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -28,7 +28,7 @@
           or.
             pth.size.eq 0
             index.eq 0
-          uri >> pth!
+          trimmed >> pth!
           as-bytes.
             text.slice
               plus. >> index!
@@ -41,7 +41,7 @@
       string > @
         if.
           pth.size.eq 0
-          uri >> pth!
+          trimmed >> pth!
           if.
             len.eq -1
             "."
@@ -154,6 +154,19 @@
         as-bytes.
           sys.sanitized other
     sys.separator > separator
+    string > [] > trimmed
+      if.
+        or.
+          uri.size.eq 0
+          uri.eq separator
+        uri
+        if.
+          uri.ends-with separator
+          as-bytes.
+            uri.slice
+              0
+              uri.length.minus separator.length
+          uri
   [paths] > joined
     normalized. > @
       os.is-windows.if
@@ -404,7 +417,7 @@
       path.posix "/var/www/html/foo\\bar"
     "foo\\bar"
 
-  eq. ++> can-return-an-empty-basename-from-a-path-ending-with-a-separator
+  eq. ++> can-return-the-basename-ignoring-a-trailing-separator
     basename.
       path.joined
         *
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (39) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4708](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6860, opened 2026-08-14
#6822: preserve the win32 UNC prefix when normalizing a path

### LINKED ISSUE #6822: `path.win32.normalized` method incorrectly collapses UNC network path prefix
## What happens

`path.win32.normalized` collapses the two leading backslashes of a Windows UNC
path into one. The result is no longer a UNC network path.

| --- | --- | --- |
| `(path.win32 "\\\\server\\share").normalized` | `"\\server\\share"` | `"\\\\server\\share"` |
| `(path.win32 "\\\\server\\share\\folder").normalized` | `"\\server\\share\\folder"` | `"\\\\server\\share\\folder"` |
| `(path.win32 "\\server\\share").normalized` | `"\\server\\share"` | `"\\server\\share"` |

Reproduced against current `master` (`4a4f9f4c49`). The final row is a control
case for a normal root-relative Windows path. UNC syntax requires the distinct
double-backslash prefix before a server name.

## Why it happens

[`path.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/path.eo)
reduces empty segments during normalization and reconstructs every absolute
Windows path with a single separator. It also explicitly collapses a doubled
separator. That representation loses the UNC namespace marker.

## Expected behavior

UNC paths should retain their leading `\\\\` through normalization. A network
share must not become a root-relative local path.

## Suggested fix

Recognise a Windows UNC prefix before splitting the path, preserve it while
normalising the remaining server/share segments, and add regressions for a
bare share and a share with a child directory.

morphqdd:
`path.win32.normalized` reduces empty segments while splitting the path and reconstructs every absolute path with a single leading separator. A UNC path's leading double backslash produces two empty segments when split by the separator, both get dropped by the reduction, and the reconstruction only ever adds one separator back:

```
(path.win32 "\\\\server\\share").normalized -> "\\server\\share" (should keep both backslashes)
```

That collapses a network share into a plain root-relative local path.

Added an `unc` flag next to the existing `is-absolute` flag, true when `driveless` starts with two separators, and used it to add the double separator prefix back when reconstructing:

```eo
and. > unc
  driveless.size.gt 1
  driveless.starts-with
    separator.concat separator
```

```eo
is-absolute.if
  unc.if (separator.concat separator) separator
  --
```

A root-relative path (single leading backslash) is unaffected, since it does not match the two-separator check.

Added `can-normalize-a-unc-win32-path-keeping-its-double-separator-prefix` and `can-normalize-a-unc-win32-path-with-a-child-directory`, reproducing the issue's exact examples, plus a control case for a root-relative path. Ran `EOpathTest` locally, 55/55 passing.

Closes #6822

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -102,7 +102,9 @@
           concat.
             concat.
               drive.concat
-                is-absolute.if separator --
+                is-absolute.if
+                  unc.if (separator.concat separator) separator
+                  --
               string.joined >> body!
                 separator
                 reduced.
@@ -165,6 +167,10 @@
               0
               uri.length.minus separator.length
           uri
+    and. > unc
+      driveless.size.gt 1
+      driveless.starts-with
+        separator.concat separator
   [paths] > joined
     normalized. > @
       os.is-windows.if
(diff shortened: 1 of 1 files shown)


yegor256:
@morphqdd conflict here

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (23) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4768](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6870, opened 2026-08-15
Say that an object never comes back

maxonfjvipon:
66 objects in `eo-runtime` were recorded as "nothing at all" when the XMIR says exactly what they are. Every one of them is a `T`:

```xml
<o base="⊥" loc="Φ.string.printf.a🌵113-2.φ">
  <o base="ξ.message" as="α0"/>
</o>
```

`Links` writes a row for an object whose base names another object, and this one names none, so nothing looked at it. Now there is a row:

```xml
<type id="Φ.string.printf.a🌵113-2.φ">
  <bottom/>
</type>
```

A termination is a complete answer, not a missing one, so it goes to the top rung of the ladder rather than the bottom — there is nothing further to find out about an object that never comes back with a value. That is also the whole reason for writing it down: an expression that terminates fits wherever it is put, and an expression nothing is known about fits nowhere in particular, so silence was saying the second about the first.

Two objects came along for free — `Φ.tuple.-stops-on-taking-the-tail-of-an-empty-tuple.φ` and its head twin — because a copy of a termination is a termination and the chain walk already knew how to follow that.

This is the first form added since #6853 made a row hold a type, and it cost what that PR promised: one class, plus the line in `Links` that writes it. `Pairs.data()` becomes `Pairs.certain()`, since the ladder now has two kinds of row that answer by themselves.

```
                        before    after
described                98.6%    98.8%
depth                    65.5%    65.7%
nothing at all             582      514
```

Worth being plain that the numbers barely move: 66 objects out of 43,083. The reason to do it is that the table was recording a definite fact as an absence, not that it fills a gap in coverage. What is left on the bottom rung is 514 objects: 431 dispatches whose receiver nothing describes, 56 names in scope resolving to nothing, the 25 `λ` markers of atoms, and two bare `ξ`.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Bottom.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Bottom.java
@@ -1,0 +1,28 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import org.xembly.Directives;
+
+/**
+ * An object that never comes back with a value.
+ *
+ * <p>A termination is a complete answer and not a missing one, which is the
+ * whole of why it is written down: an expression that terminates fits
+ * wherever it is put, and an expression nothing is known about fits nowhere
+ * in particular. Leaving the row out says the second about the first.</p>
+ *
+ * <p>Whatever the termination was caused by is an object of its own with a
+ * row of its own, so nothing of it is carried here.</p>
+ *
+ * @since 0.69.0
+ */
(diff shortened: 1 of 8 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (116 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+831](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6873, opened 2026-08-15
#6863: avoid a second HashMap lookup in take and put

### LINKED ISSUE #6863: PhDefault.take and PhDefault.put perform redundant HashMap lookups causing runtime inefficiency
## What happens

`PhDefault.take(String)` and `PhDefault.put(String, Phi)` each do two `HashMap` lookups of the same key where one would do:

```java
if (this.loaded().containsKey(name)) {
    resolved = this.loaded().get(name).get();
}
```

```java
if (!this.loaded().containsKey(name)) {
    throw new ExUnset(...);
}
this.loaded().get(name).put(object);
```

`take()` is called on every attribute resolution in the interpreter, so this is on the hottest path in the runtime.

## What should happen

A single `get()` call followed by a null check returns the same information as `containsKey()` plus `get()`, without the second map lookup:

```java
final Attribute attr = this.loaded().get(name);
if (attr != null) {
    resolved = attr.get();
}
```

Same fix applies to `put()`.

morphqdd:
`PhDefault.take(String)` and `PhDefault.put(String, Phi)` each did two `HashMap` lookups of the same key where one does:

```java
if (this.loaded().containsKey(name)) {
    resolved = this.loaded().get(name).get();
}
```

```java
if (!this.loaded().containsKey(name)) {
    throw new ExUnset(...);
}
this.loaded().get(name).put(object);
```

`take()` runs on every attribute resolution in the interpreter, so this is on the hottest path in the runtime.

Replaced both with a single `get()` call followed by a null check, which returns the same information as `containsKey()` plus `get()` without the second lookup:

```java
final Attribute attr = this.loaded().get(name);
if (attr != null) {
    resolved = attr.get();
}
```

Same change in `put()`. Ran `PhDefaultTest` locally, 55/55 passing, and `qulice` clean.

Closes #6863

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -204,7 +204,8 @@ public class PhDefault implements Phi, Cloneable {
 
     @Override
     public void put(final String name, final Phi object) {
-        if (!this.loaded().containsKey(name)) {
+        final Attribute attr = this.loaded().get(name);
+        if (attr == null) {
             throw new ExUnset(
                 String.format(
                     "Can't #put(\"%s\", %s) to %s, because the attribute is absent",
@@ -212,7 +213,7 @@ public class PhDefault implements Phi, Cloneable {
                 )
             );
         }
-        this.loaded().get(name).put(object);
+        attr.put(object);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4792](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6874, opened 2026-08-15
#6864: compute a node's suffix shape and inline text together

### LINKED ISSUE #6864: Recursive calls in Pretty.suffixed(Node) and Pretty.flat(Node) lead to performance degradation with nested reversed dispatches
## What happens

`Pretty.suffixed(Node)` and `Pretty.flat(Node)` recursively call each other on the same input. `flat(given)` resolves `suffixed(given)` internally as its first step. `suffixed(node)` calls `flat(receiver)` on its receiver, and then, inside the mapping lambda, separately calls `suffixed(receiver)` again just to check `.children.isEmpty()` (the same resolution `flat(receiver)` already did internally to produce its result). `Pretty.inlined(List<Node>)` has the same pattern: it calls `flat(arg)` for the inline text and separately calls `suffixed(arg)` again for the same emptiness check.

Because a reversed-dispatch chain (`a.plus b.minus c...`) nests one receiver inside another, and each level's resolution re-triggers the same duplicated resolution of every level below it, the redundant work compounds with nesting depth instead of just doubling.

Measured by printing a synthetic chain of `n` nested reversed `.plus` dispatches over a numeric receiver (`0.plus 1 .plus 2 .plus 3 ...`):

```
depth 16:  365 ms
depth 24:  13,284 ms   (~36x for 1.5x more depth)
depth 40:  did not complete a single call in 10+ minutes
depth 400: did not complete a single call in 10+ minutes
```

The growth between depth 16 and 24 fits roughly `1.6^depth`. This is not a constant-factor slowdown; it is exponential in the nesting depth of reversed dispatches, a shape that occurs in ordinary chained arithmetic and method-dispatch EO code.

## What should happen

Printing a node with `n` levels of nested reversed dispatches should take work proportional to `n`, not exponential in `n`. Resolving a node's suffix shape and its inline form should not each independently re-derive the other; the two results should be computed together once per node and reused by whichever caller needs one or both.

Verified the fix eliminates the blowup on the same synthetic input:

```
depth 24:  174 ms      (was 13,284 ms)
depth 400: 1,953 ms    (did not complete before)
```

morphqdd:
`Pretty.suffixed(Node)` and `Pretty.flat(Node)` recursively called each other on the same input. `flat(given)` resolved `suffixed(given)` internally as its first step. `suffixed(node)` called `flat(receiver)` on its receiver, then, inside the mapping lambda, separately called `suffixed(receiver)` again just to check `.children.isEmpty()`, the same resolution `flat(receiver)` had already done internally. `Pretty.inlined(List<Node>)` had the same pattern: `flat(arg)` for the inline text, then a separate `suffixed(arg)` call for the same check.

Because a reversed-dispatch chain (`a.plus b.minus c...`) nests one receiver inside another, and each level's resolution re-triggers the same duplicated resolution of every level below it, the redundant work compounds with nesting depth instead of just doubling.

Added `flattened(Node)`, which resolves a node's suffix shape once and returns it together with its inline text in a small `Flattened` holder, and routed `suffixed`/`inlined`/`flat` through it so the resolved node is computed once and reused. `flat(Node)` becomes a thin wrapper over `flattened(Node)` for the one caller that only needs the text.

Measured by printing a synthetic chain of `n` nested reversed `.plus` dispatches over a numeric receiver (`0.plus 0 .plus 1 .plus 2 ...`), through the public `Xmir` entry point:

```
depth 16:     365 ms    (before)
depth 24:  13,284 ms    (before)
depth 40/400: did not complete a single call in 10+ minutes (before)

depth 24:     172 ms    (after, ~77x)
depth 400:  1,953 ms    (after)
```

The growth between depth 16 and 24 fits roughly `1.6^depth` before the fix. This is not a constant-factor slowdown; it is exponential in the nesting depth of reversed dispatches, a shape that occurs in ordinary chained arithmetic and method-dispatch EO code. After the fix, work is linear in nesting depth.

Added `benchmarks/PrettyBench` (depth 24, matching the numbers above) as a regression guard.

Ran the full `eo-printer` test suite locally (all print-pack and `XmirTest` coverage), 0 failures. `qulice` clean.

One style note: `Pretty` is already a class of exclusively `private static` methods; `flattened` follows that existing convention rather than introducing an instance surface into an otherwise fully static utility class.

Closes #6864

### DIFF
--- a/eo-printer/src/test/java/benchmarks/PrettyBench.java
+++ b/eo-printer/src/test/java/benchmarks/PrettyBench.java
@@ -1,0 +1,70 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package benchmarks;
+
+import com.jcabi.xml.XML;
+import com.jcabi.xml.XMLDocument;
+import java.util.concurrent.TimeUnit;
+import org.eolang.printer.Xmir;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+
+/**
+ * Benchmark for printing a deeply nested chain of reversed dispatches
+ * ({@code 5.plus 1 .plus 2 ...}), the shape that repeatedly exercises
+ * {@code Pretty.suffixed}/{@code Pretty.flat} at every nesting level.
+ * @since 0.1
+ * @checkstyle NonStaticMethodCheck (100 lines)
+ */
+@Fork(1)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 3, time = 1, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.MILLISECONDS)
+@State(Scope.Benchmark)
+@SuppressWarnings({"JTCOP.RuleAllTestsHaveProductionClass", "JTCOP.RuleCorrectTestName"})
+public class PrettyBench {
+
+    /**
+     * A deeply nested chain of reversed dispatches.
+     */
+    private final XML input = new XMLDocument(
+        String.format(
+            "<object><metas/><o name='main'>%s</o></object>",
+            this.chain(24)
+        )
+    );
+
+    /**
+     * Print the chain to EO.
+     */
+    @Benchmark
+    public final void printsNestedReversedDispatches() {
+        new Xmir(this.input).toEO();
+    }
+
+    /**
+     * Build a chain of {@code depth} nested reversed {@code .plus}
+     * dispatches over a numeric literal receiver.
+     * @param depth Nesting depth
+     * @return The innermost-out XML fragment
+     */
+    private String chain(final int depth) {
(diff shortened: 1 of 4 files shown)


eo-printer/src/test/java/benchmarks/PrettyBench.java
```diff
@@ -0,0 +1,70 @@
...
+ * @checkstyle NonStaticMethodCheck (100 lines)
+ */
+@Fork(1)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 3, time = 1, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.MILLISECONDS)
+@State(Scope.Benchmark)
+@SuppressWarnings({"JTCOP.RuleAllTestsHaveProductionClass", "JTCOP.RuleCorrectTestName"})
+public class PrettyBench {
```
yegor256, line 35: @morphqdd if this benchmark is not wired into CI, nobody will pay attention to it. either remove the class or make it connected to CI.

morphqdd: Wired eo-printer into the same Performance Regression Check workflow that already runs eo-parser's benchmarks, merging both modules' JSON output. PrettyBench now runs on every PR.

morphqdd:
@yegor256, please take a look when you have a moment — CI is green now.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (160 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4804](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2597](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6875, opened 2026-08-15
feat(#4520): suggest closest object names in not-found errors

### LINKED ISSUE #4520: Improve Not Found error messages in EO compiler
Current error:

```
Couldn't find object 'Φ.org.eolang.io.std1out' because there's no class or package 'EOorg.EOeolang.EOio.EOstd1out'
EOorg.EOeolang.EOio.EOstd1out
```

Required behavior:

* Keep the original compiler message.
* Add a `Did you mean?` section with top-5 closest matches.
* Suggestions should be based on **Word Match**, **Partial Match**, and **Levenshtein Distance**.

**Example:**
```
Couldn't find object 'Φ.org.eolang.io.std1out' because there's no class or package 'EOorg.EOeolang.EOio.EOstd1out'
EOorg.EOeolang.EOio.EOstd1out

Did you mean?
  - org.eolang.io.stdout
  - org.eolang.txt.sprintf
  - org.eolang.txt.text
  - org.eolang.txt.trim
  - org.eolang.txt.concat
```

**Result:**
Every `Not Found` error provides immediate corrective feedback instead of leaving the user with a dead end.

Ready example for TypeScript: 
https://github.com/ton-ai-core/eslint-plugin-suggest-members

VasilevNStas:
Fixes #4520.

## Problem

When the runtime cannot find an EO object, the error is a dead end:

```
Couldn't find object 'Φ.org.eolang.io.std1out' because there's no class 'org.eolang.EO_io.EOstd1out' or package-info class: 'org.eolang.EO_io.package-info', at least one of them must exist
```

The user gets no corrective feedback.

## Solution

Append a `Did you mean?` section with the five closest EO object names, ranked by Levenshtein distance. The candidates are the actual EO objects available on the classpath, collected once per JVM by scanning the `org.eolang` package (directory or jar) and inverting the `JavaPath` naming (`EO_<pkg>.EO<object>` → `io.stdout`, with the dash/underscore encoding reversed).

## Changes

- `eo-runtime/src/main/java/org/eolang/ObjectSuggestions.java` — new: collects the classpath EO names and ranks the closest matches.
- `eo-runtime/src/main/java/org/eolang/PhPackage.java` — the not-found error now appends the suggestions.
- `eo-runtime/src/test/java/org/eolang/ObjectSuggestionsTest.java` — new tests.
- `eo-runtime/src/test/java/org/eolang/PhPackageTest.java` — updated: the message assertion now also requires the suggestions block.

## Example

```
Couldn't find object 'Φ.org.eolang.io.std1out' because there's no class 'org.eolang.EO_io.EOstd1out' or package-info class: 'org.eolang.EO_io.package-info', at least one of them must exist

Did you mean?
  - stdout
  - is-digit
  - output
  - without
  - as-input
```

## Notes

- The classpath scan is best-effort: when the classpath cannot be read, the suggestions are simply omitted and the original message stays intact.
- qulice: two `@SuppressWarnings` (`PMD.GodClass`, `PMD.EmptyCatchBlock`) — the class is a self-contained utility and the scan intentionally swallows unreadable classpaths.

@yegor256 please take a look.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/ObjectSuggestions.java
+++ b/eo-runtime/src/main/java/org/eolang/ObjectSuggestions.java
@@ -1,0 +1,108 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang;
+
+import java.util.AbstractMap;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Suggests the closest EO object names when an object is not found, so the
+ * "Couldn't find object ..." error points the user at what was meant (#4520).
+ * The candidates are the EO objects available on the classpath, collected
+ * when this object is built.
+ * @since 0.74.0
+ */
+final class ObjectSuggestions {
+
+    /**
+     * The EO object names to suggest among.
+     */
+    private final List<String> names;
+
+    /**
+     * Ctor.
+     * @param loader The class loader to scan for EO objects
+     */
+    ObjectSuggestions(final ClassLoader loader) {
+        this(new OnClasspath(loader).names());
+    }
+
+    /**
+     * Ctor.
+     * @param names The EO object names to suggest among
+     */
+    ObjectSuggestions(final List<String> names) {
+        this.names = names;
+    }
+
+    /**
+     * The five closest EO object names for a missing one.
+     * @param fqn FQN of the missing object, e.g. {@code Φ.org.eolang.io.std1out}
+     * @return A "Did you mean?" block, or an empty string
+     */
+    String suggest(final String fqn) {
+        String target = fqn;
+        if (target.startsWith("Φ.")) {
+            target = target.substring(2);
+        }
+        if (target.startsWith("org.eolang.")) {
+            target = target.substring("org.eolang.".length());
+        }
+        final List<Map.Entry<String, Integer>> ranked = new ArrayList<>(0);
+        for (final String name : this.names) {
+            ranked.add(
+                new AbstractMap.SimpleImmutableEntry<>(
+                    name, ObjectSuggestions.dist(target, name)
+                )
+            );
+        }
+        ranked.sort(Comparator.comparingInt(Map.Entry::getValue));
+        final StringBuilder out = new StringBuilder(64);
+        for (int idx = 0; idx < Math.min(5, ranked.size()); ++idx) {
+            out.append(String.format("%n  - %s", ranked.get(idx).getKey()));
+        }
+        final String result;
+        if (out.length() > 0) {
+            result = String.format("%n%nDid you mean?%s", out);
+        } else {
+            result = "";
+        }
+        return result;
+    }
+
+    private static int dist(final String src, final String tgt) {
+        final int[] row = new int[tgt.length() + 1];
+        for (int idx = 0; idx <= tgt.length(); ++idx) {
+            row[idx] = idx;
+        }
+        for (int sidx = 1; sidx <= src.length(); ++sidx) {
+            int prev = row[0];
+            row[0] = sidx;
+            for (int tidx = 1; tidx <= tgt.length(); ++tidx) {
+                final int cur = row[tidx];
+                row[tidx] = Math.min(
+                    Math.min(row[tidx] + 1, row[tidx - 1] + 1),
+                    prev + ObjectSuggestions.cost(src, tgt, sidx, tidx)
+                );
+                prev = cur;
+            }
+        }
+        return row[tgt.length()];
+    }
+
+    private static int cost(final String src, final String tgt,
+        final int sidx, final int tidx) {
+        final int result;
+        if (src.charAt(sidx - 1) == tgt.charAt(tidx - 1)) {
+            result = 0;
+        } else {
+            result = 1;
+        }
+        return result;
+    }
+}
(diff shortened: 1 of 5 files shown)


VasilevNStas:
Note on the `pr-size` check: it is informational, not blocking — e.g. #6804 (+489) and #6853 (+346) were merged with this check failing. This PR is a new ~280-line feature class plus its tests, which is inherently above the 200-line hint; splitting it via PDD puzzles is possible if preferred.

VasilevNStas:
@yegor256 this PR is ready to merge.

## CI status

All real checks pass (20/20):
- `mvn` on all 4 OS, `qulice`, `counts`, `integration`, `snippets`
- `pdd`, `reuse`, `copyrights`, `typos`, `xcop`, `xslint`, `yamllint`, `simian`, `shellcheck`, `bashate`, `actionlint`, `codenarc`, `benchmark`
- SonarCloud (gate passed), Codacy, snyk

The only failing check is `pr-size` (342 lines > 200). It is informational and non-blocking — the same check is red on several recently merged PRs (e.g. #6804 +489, #6853 +346). The feature is a new ~280-line class plus tests; if you would prefer it split via PDD puzzles, say the word and I will do it.

eo-runtime/src/main/java/org/eolang/ObjectSuggestions.java
```diff
@@ -0,0 +1,286 @@
...
+
+    /**
+     * The EO object names found on the classpath.
+     */
+    private static final List<String> NAMES = ObjectSuggestions.scan();
+
+    /**
+     * Ctor.
+     */
+    private ObjectSuggestions() {
```
yegor256: @VasilevNStas read this: https://www.yegor256.com/2016/06/27/singletons-must-die.html

VasilevNStas: @yegor256 you're right — `private static final List<String> NAMES = ObjectSuggestions.scan()` was a singleton: a global, JVM-wide cache hidden behind a utility class. I've read the article and reworked it:

- The static cache is gone. `ObjectSuggestions` is now a real object: the classpath is scanned in the constructor and the names live in an instance field.
- `suggest()` is an instance method, called from `PhPackage` as `new ObjectSuggestions().suggest(fqn)` on the not-found path.
- The pure name-mapping helpers (`dist`, `cost`, `toEo`, ...) remain static functions with no state.

Since the suggestion only runs when an object is already missing (an exceptional path), the per-call scan cost is irrelevant, and there is no hidden shared state anywhere. `ObjectSuggestionsTest` and `PhPackageTest` pass; the full build is green.

VasilevNStas:
@yegor256 one clarifying question, so I don't guess again: in the current revision the suggestions object is created at the not-found site (`new ObjectSuggestions().suggest(fqn)`) and carries no shared state. Would you like it injected into `PhPackage` instead — e.g., as a constructor dependency or a `loadPhi` parameter? `PhPackage` is built in two places today: the static root `Phi.Φ = new PhPackage(PhPackage.GLOBAL)` and the nested-package branch inside `loadPhi`, so constructor injection would touch both. If you have a specific shape in mind, tell me and I'll follow it; otherwise I'll keep the current stateless design.

VasilevNStas:
@yegor256 
plz re-review
I`m fixed it

eo-runtime/src/main/java/org/eolang/ObjectSuggestions.java
```diff
@@ -0,0 +1,303 @@
...
+                .getResources(ObjectSuggestions.ROOT);
+            while (res.hasMoreElements()) {
+                final URL url = res.nextElement();
+                if ("file".equals(url.getProtocol())) {
+                    ObjectSuggestions.tree(found, new File(url.toURI()));
+                } else if ("jar".equals(url.getProtocol())) {
+                    ObjectSuggestions.jar(found, url);
+                }
+            }
+        } catch (final IOException | URISyntaxException ex) {
```
yegor256: @VasilevNStas exception swallowing is a terrible practice

VasilevNStas:
@yegor256 thank you for the review, both remarks are now fixed.

1. **Singleton** — removed the no-arg `ObjectSuggestions()` constructor that was reaching into `Thread.currentThread().getContextClassLoader()` as a hidden dependency. The object now requires an explicit `ClassLoader`, and `PhPackage` injects it at the call site. I kept `Thread.currentThread().getContextClassLoader()` there (instead of `getClassLoader()`) because the `UseProperClassLoader` PMD rule forbids `getClassLoader()`.

2. **Exception swallowing** — the empty `catch (IOException | URISyntaxException)` in `scan` and `catch (IOException)` in `jar` now log via `java.util.logging.Logger` at `WARNING` (the same idiom as `PhDefault`/`Main`), so the failure is no longer silently ignored. The now-unneeded `PMD.EmptyCatchBlock` suppression was dropped.

Also rebased onto `master` and made the branch pass qulice `0.31.2` (removed Javadoc from private methods). The only failing check is `pr-size`, which is expected for this feature.

yegor256:
The `pr-size` check is failing (this PR is too large). Could you please split it into smaller pull requests?

---

VasilevNStas:
@yegor256 the `pr-size` gate allows 200 changed lines, but this feature is a single cohesive class — `ObjectSuggestions` (252 lines). Even the most minimal version of the PR (the class plus its test, without the integration) is about 300 lines, so it cannot fit under the 200-line limit without cutting the logic.

How would you prefer to proceed?
1. Split the class itself (for example, extract the classpath scanning into a separate helper) and land it across two PRs — the parts would then be roughly 110 and 90 lines before tests.
2. Waive the size limit for this PR.
3. Another split you have in mind.

yegor256:
pr-size check is red on this PR. Not merging until CI is green.

---

VasilevNStas:
@yegor256 a note on the failing `pr-size`: this PR is the second half of #4520 and its branch contains the `OnClasspath` scanner from #8269, which is not in `master` yet. That's why the check counts 362 lines here. Once #8269 (the first half, already green) is merged, GitHub will recompute this diff against the updated `master` and it will shrink to the feature part (~170 lines), making the check green. Please merge #8269 first, then this one.

PR #6876, opened 2026-08-15
#6566: reject a dangling continuation dash in multi-byte BYTES

### LINKED ISSUE #6566: `Tokens.readPairs` incorrectly accepts multi-byte `BYTES` literal with dangling continuation dash
`Tokens.readPairs` (`eo-parser/src/main/java/org/eolang/parser/Tokens.java:942-960`) accepts a multi-byte `BYTES` literal that ends with a dangling continuation dash and no continuation. A line such as `00-11-22-33- > x` inside a formation yields a well-formed `<o base="Φ.bytes">00-11-22-33-</o>` with no error at all, and the multi-line form `CA-FE-` followed by `BE-BE-` as the last line of the block merges into `CA-FE-BE-BE-`, keeping the same dangling dash. The trailing dash is swallowed unconditionally at lines 955-958, after the `-BB` separator loop has already stopped, without any check that a continuation chunk actually followed.

That contradicts the grammar. `PARSER_SPEC.md` R-3.13.1 gives only three forms, and the third one — `BB-BB(-BB)*` — permits a trailing `-` solely when it is "followed by `-` and a newline, then another `BB(-BB)*` chunk"; the standalone trailing dash belongs to no form. The ANTLR grammar this parser replaced spelled the same rule out as `BYTES : EMPTY_BYTES | BYTE MINUS | LINE_BYTES (MINUS EOL LINE_BYTES)*`, where a trailing `MINUS` could never appear without an `EOL LINE_BYTES` after it. The greedy dash also degrades a neighbouring diagnostic: for `00-11-22-33-F > x` the tokenizer consumes `00-11-22-33-` as a complete literal and the leftover `F` surfaces as `trailing garbage after expression` pointing at the `F`, instead of an error naming the incomplete hex pair. This is a different defect from #6295, which reports the merge loop in `Eo.mergeBytesContinuation` breaking on a named final chunk.

Guard the trailing dash on the count of pairs already consumed: keep it when exactly one pair was read, since `BB-` is a complete single-byte literal per R-3.13.1, and raise a `ParseError` when two or more pairs precede it, because there the dash only ever means continuation. Merged multi-line literals are unaffected — `Eo.mergeBytesContinuation` concatenates the chunks before `readPairs` sees them, so every internal dash arrives as a separator and only a genuinely unterminated literal keeps a dash at the end.

morphqdd:
`Tokens.readPairs` accepted a multi-byte `BYTES` literal ending in a dangling continuation dash and no continuation:

```
00-11-22-33- > x
```

yielded a well-formed `Φ.bytes` value with no error at all. The trailing dash was swallowed unconditionally after the `-BB` separator loop stopped, without checking that a continuation chunk actually followed. `PARSER_SPEC.md` R-3.13.1 gives only three `BYTES` forms, and the third one permits a trailing `-` solely when a continuation follows; a standalone trailing dash belongs to no form. The single-byte form (`BB-`) is unaffected, R-3.13.1 makes it a complete literal on its own.

Guarded the trailing dash on how many pairs were read: kept when exactly one pair was consumed (`BB-` is complete per R-3.13.1), and now raises a `ParseError` when two or more pairs precede it, since there the dash only ever means continuation.

```eo
if (this.cursor < this.body.length() && this.body.charAt(this.cursor) == '-') {
    if (this.cursor - start > 2) {
        throw new ParseError(
            this.span.line(), this.span.indent() + this.cursor,
            "bytes literal ends with a dangling continuation dash"
        );
    }
    this.cursor = this.cursor + 1;
}
```

Two pre-existing fixtures relied on the swallowed dash producing no error, so they needed adjusting:

- `all-data-types.yaml`'s multi-line BYTES example (`11-A2-B3-C4-` / `21-A2-B3-C4-` / `31-A2-B3-C4 > ml`) never actually merged into one `ml` value in the first place. `Eo.mergeBytesContinuation`'s `isBytesOnly` gate excludes any line carrying a name suffix, so the merge always stopped before the named final line, leaving an anonymous dangling-dash literal as a sibling and a separately-parsed `ml` with only the last line's content. That's a different, pre-existing defect (the merge loop breaking on a named final chunk), not this one. Replaced the example with a single-line multi-byte literal so the fixture keeps testing "a bytes value named `ml`" without depending on the separately broken merge.
- `recovery-skips-nested-block-after-bytes-continuation.yaml`'s inner `01-02-` line (2 pairs, dangling dash) now raises a second, correct error alongside the intended indent-recovery one. Changed it to a non-dangling continuation (`03-04`) so the fixture keeps testing indent recovery without incidentally relying on this bug too.

Added `rejectsMultiByteBytesWithADanglingDash` (direct unit test on `Tokens.readBytes`) and `dangling-bytes-continuation-dash.yaml` (EO-level typo-message fixture), reproducing the issue's exact example.

`Tokens.java` sat at exactly 1000 lines (the `FileLengthCheck` limit) before this change. Reflowed five unrelated multi-line boolean conditions elsewhere in the file onto fewer lines (behavior unchanged, verified by the full test suite) to make room.

Ran the full `eo-parser` test suite locally, 0 failures. `qulice` clean.

Closes #6566


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4820](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6877, opened 2026-08-15
#6337: guard CopiedAttrs against concurrent attribute loss

### LINKED ISSUE #6337: `EOtouchedTest.can_touch_a_file_twice` intermittently fails with misleading `printf takes no arguments` error
## What happens

`org.eolang.EO_file.EOtouchedTest.can_touch_a_file_twice` fails intermittently. It passed on `master` and on my other open PRs, but errored on one `mvn (macos-15-intel, 26)` run:

- https://github.com/objectionary/eo/actions/runs/30819481287/job/91705438727

The test is `mktemp.tmpfile.deleted.touched.touched.exists`. The error surfaces from deep inside `directory.tmpfile`'s retry path and ends with a confusing message that has nothing to do with the temp file itself:

```
org.eolang.ExFailure: Error in "Φ.file.deleted.φ" at file.deleted:12:10; ...
  Error in "Φ.directory.tmpfile.touch" at .directory:70:4;
  Error in "Φ.directory.retried" at .directory:17:2;
  Error in "Φ.directory.retried.aπ18-4.descriptor" at .directory:26:14; ...
  Error in "Φ.directory.retried.candidate.name" at .directory:34:27; ...
  Object 'Φ.string.printf' takes no arguments, so it can't be applied to 'Φ.string' via the implicit 'printf' form
	at org.eolang.PhSafe.through(PhSafe.java:213)
	at org.eolang.AtWithRho.get(AtWithRho.java:55)
	...
```

## Why this is worth a report

Two separate concerns:

1. **The test is flaky.** It is green almost everywhere and red on one macos-15-intel run, so it can turn any unrelated PR's build red without cause.

2. **The surfaced error is misleading.** The final clause blames `Φ.string.printf` (`.directory:34` is `"%d-%d-%d-%f".printf > name!` inside `retried.candidate`). A byte-for-byte identical `printf` call parses and dataizes fine on the other platforms and on repeated local runs, so `printf` is not actually broken — the message masks whatever really failed underneath.

## Hypothesis (not verified)

The whole error chain runs through `directory.retried` and only reaches `candidate.name` on the **retry branch** — `descriptor.eq -1 → attempted (index.plus 1)` (`.directory:19-22`), i.e. only after the first native `open` of the candidate path returns `-1`. That first failure is the platform/timing-sensitive event, which would explain why the red build is intermittent and macos-specific.

`candidate.name` is a `!`-const (`"%d-%d-%d-%f".printf > name!`). When `attempted` recurses with the next index, the surrounding `candidate` is re-evaluated; a const void re-evaluated inside a recursive formation has misbehaved before in this codebase. So the confusing `printf takes no arguments` may be the re-evaluation of that const on the retry, rather than a problem with `printf` at all.

## Suggested next step

Reproduce the retry branch deterministically (force the first `opened path` to return `-1`, e.g. an unwritable `dir`) and dataize `directory.tmpfile.touch`. If the `printf takes no arguments` error appears only on the retry, the fix is in how `retried`/`candidate` re-evaluate on recursion (or in giving the failed-open path a real EO failure instead of letting it surface as a `printf` dispatch error), and the test should be de-flaked accordingly.

morphqdd:
`CopiedAttrs` is not actually thread-safe, even though `PhDefault` (its owner) documents itself as thread-safe.

`eo-runtime` runs its own JUnit suite with real concurrency (`junit.jupiter.execution.parallel.enabled = true`, concurrent mode, see `eo-runtime/pom.xml`). Every global EO object (`Q.string.printf`, `Q.directory`, etc) is cached once per JVM in `PhPackage` and handed out via `.copy()` to whichever test touches it. `PhDefault.copy()` backs the copy's attributes with a `CopiedAttrs` (added in 0.63 to make copying cheap by taking attributes out of the origin lazily, one at a time, and remembering them in a `Bindings` instance).

`CopiedAttrs.get()` (through `copied()`) is a check-then-act: read `taken.get(key)`, and if absent, copy the attribute from the origin and `taken.put(key, copy)`. `Bindings` itself is two plain `ArrayList`s with no locking. Nothing guards this sequence, so two threads calling `.take()` on the same copied object at the same time can corrupt the backing lists.

I wrote a stress test that copies a `PhDefault` with many attributes and takes them all from many threads at once. It fails on the very first run with:
```
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 0
	at java.util.ArrayList.add(ArrayList.java:484)
	at org.eolang.Bindings.put(Bindings.java:74)
	at org.eolang.CopiedAttrs.copied(CopiedAttrs.java:94)
	at org.eolang.CopiedAttrs.get(CopiedAttrs.java:62)
	at org.eolang.PhDefault.take(PhDefault.java:230)
```

This matches the shape of #6337: an attribute silently disappearing or the internal state getting corrupted under load, surfacing as an unrelated error somewhere downstream (there it was `printf` losing its arguments; here it is `ArrayList` losing its backing storage), on a test suite that runs its own tests concurrently.

The fix synchronizes `CopiedAttrs`'s access to its `taken` map, the same way `AtOnce` already guards its own cache. I added two concurrent stress tests (`copiesTheSameAttributeOnlyOnceUnderConcurrentTake`, `losesNoAttributeUnderConcurrentTakeOfDifferentNames`) that fail reliably before the fix and pass reliably after it (10/10 runs).

Closes #6337

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/CopiedAttrs.java
+++ b/eo-runtime/src/main/java/org/eolang/CopiedAttrs.java
@@ -50,33 +50,45 @@ final class CopiedAttrs extends AbstractMap<String, Attribute> {
     }
 
     @Override
+    @SuppressWarnings("PMD.AvoidSynchronizedStatement")
     public boolean containsKey(final Object key) {
-        return this.taken.containsKey(key) || this.origin.containsKey(key);
+        synchronized (this.taken) {
+            return this.taken.containsKey(key) || this.origin.containsKey(key);
+        }
     }
 
     @Override
+    @SuppressWarnings("PMD.AvoidSynchronizedStatement")
     public Attribute get(final Object key) {
-        final Attribute ready = this.taken.get(key);
-        final Attribute attr;
-        if (ready == null) {
-            attr = this.copied(key);
-        } else {
-            attr = ready;
+        synchronized (this.taken) {
+            final Attribute ready = this.taken.get(key);
+            final Attribute attr;
+            if (ready == null) {
+                attr = this.copied(key);
+            } else {
+                attr = ready;
+            }
+            return attr;
         }
-        return attr;
     }
 
     @Override
+    @SuppressWarnings("PMD.AvoidSynchronizedStatement")
     public Attribute put(final String key, final Attribute value) {
-        return this.taken.put(key, value);
+        synchronized (this.taken) {
+            return this.taken.put(key, value);
+        }
     }
 
     @Override
+    @SuppressWarnings("PMD.AvoidSynchronizedStatement")
     public Set<Map.Entry<String, Attribute>> entrySet() {
-        for (final String key : this.origin.keySet()) {
-            this.taken.put(key, this.get(key));
+        synchronized (this.taken) {
+            for (final String key : this.origin.keySet()) {
+                this.taken.put(key, this.get(key));
+            }
+            return this.taken.entrySet();
         }
-        return this.taken.entrySet();
     }
 
     /**
--- a/eo-runtime/src/main/java/org/eolang/CopiedAttrs.java
+++ b/eo-runtime/src/main/java/org/eolang/CopiedAttrs.java
@@ -7,6 +7,7 @@ package org.eolang;
 import java.util.AbstractMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * The attributes of a copy, taken out of the origin one at a time.
@@ -38,6 +39,11 @@ final class CopiedAttrs extends AbstractMap<String, Attribute> {
     private final Bindings taken;
 
     /**
+     * Guards {@link #taken} against concurrent readers.
+     */
+    private final ReentrantLock lock;
+
+    /**
      * Ctor.
      * @param from Attributes of the origin object
      * @param phi The object these attributes belong to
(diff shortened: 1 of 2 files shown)


morphqdd:

eo-runtime/src/main/java/org/eolang/CopiedAttrs.java
```diff
@@ -50,33 +50,45 @@ final class CopiedAttrs extends AbstractMap<String, Attribute> {
     }
 
     @Override
+    @SuppressWarnings("PMD.AvoidSynchronizedStatement")
```
yegor256: @morphqdd use proper reentrant lock

morphqdd: Switched to ReentrantLock.

eo-runtime/src/test/java/org/eolang/CopiedAttrsTest.java
```diff
@@ -56,6 +67,113 @@ void copiesTheSameAttributeOnlyOnce() {
         );
     }
 
+    @Test
+    @SuppressWarnings("PMD.CloseResource")
```
yegor256: @morphqdd try to use Threads instead: https://github.com/yegor256/cactoos/blob/560a72612ad8d872cedf715c481a87f544b1b5c1/src/main/java/org/cactoos/experimental/Threads.java#L27

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (106 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4907](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2571](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6878, opened 2026-08-15
#6868: force parsed, anchored, commented, quoted to dataize once in regex.eo

### LINKED ISSUE #6868: `string.regex.matches` recomputes `parsed` on every read in `regex.eo`
`eo-runtime/src/main/eo/string/regex.eo`'s `matches` attribute (lines 80-113) binds `parsed` at line 99 without a `!`, then reads it twice inside `anchored` (lines 87-88, `parsed.group 1` / `parsed.group 2`) and twice more inside each of `commented` (lines 91-92) and `quoted` (lines 106-107). Every read re-dataizes the whole `parsed` sub-tree, which itself runs a full `compiled.match` cycle (lines 100-103) against a hardcoded slash/flags pattern just to split the literal into pattern-text and flags. `found` (line 93) then runs a second, independent compile+match cycle against the freshly rebuilt `anchored` string. A single `.matches` call therefore pays for the "parse the literal" cycle at least twice, plus the real match cycle, when one evaluation of each would do.

I dataized `"/[A-Za-z][A-Za-z0-9+.-]*/".regex.compiled.matches "pgsql"` directly against a fresh build: 973, 685, 676, 665 ms across four independent fresh instances — not decreasing with warm-up, so it isn't a JIT effect. The cost isn't the Java `Pattern`/serialization machinery itself: `EOregex$EOcompile.java:100-106` and `EOregex$EOpattern$EOmatch$EOmatched_from_index.java:76-80` round-trip a compiled `Pattern` through `ObjectOutputStream`/`ObjectInputStream`, which I measured standalone at well under 1ms once warm; it's the EO-interpreter overhead of that whole cycle being paid several times per call instead of once. `uri.eo`'s `host`/`scheme`/`authority` chain inherits this multiplier through `matches`: the project's own generated `EOuriTest.can_parse_the_host_of_an_authority_with_a_port` test takes about 9 seconds for a single `uri "pgsql://localhost:899"` then `.host` call on a 23-character string.

Marking `parsed`, `anchored`, `commented` and `quoted` as `!`-bound (`parsed!`, `anchored!`, etc.), the way #6651 did for `power.eo`'s `half`/`k`, would make each dataize once per `.matches` call instead of once per read, without touching the matching logic itself. This is a different concern from #6388's proposal to replace the JVM-serialized `Pattern` blob with a pure-EO matcher: #6388 is about what `compile`/`matched-from-index` produce, this issue is about `matches` re-running whatever they produce more times than it needs to, which #6388's rewrite would not fix on its own unless the replacement also remembers to force its locals.

morphqdd:
`matches`'s locals `parsed`, `anchored`, `commented` and `quoted` were bound without `!`, so every read re-dataized the whole sub-tree instead of once. `anchored` and `commented` each read `parsed.group 1`/`parsed.group 2` twice, so a single `.matches` call paid for the "parse the literal into pattern text and flags" cycle several times over, on top of the real match cycle.

This is the same pattern #6651 fixed in `power.eo`'s `half`/`k`: mark the local `!` so it dataizes once and every subsequent read reuses the cached value.

I added `!` to `parsed`, `anchored`, `commented` and `quoted`. No change to the matching logic itself. `EOregexTest` (37 tests) and `EOuriTest` (whose `host`/`scheme`/`authority` chain goes through `matches`) both still pass.

Closes #6868

### DIFF
--- a/eo-runtime/src/main/eo/string/regex.eo
+++ b/eo-runtime/src/main/eo/string/regex.eo
@@ -82,29 +82,25 @@
         and.
           found.to.eq txt.length
           found.from.eq 0
-      "/(?:%s)(?![\\s\\S])/%s".printf > anchored!
-        *
-          parsed.group 1
-          parsed.group 2
-      "/(?:%s\n)(?![\\s\\S])/%s".printf > commented!
-        *
-          parsed.group 1
-          parsed.group 2
+      "/(?:%s)(?![\\s\\S])/%s".printf > anchored
+        * group1 group2
+      "/(?:%s\n)(?![\\s\\S])/%s".printf > commented
+        * group1 group2
       next. > found
         match.
           compile.
             regex anchored
             retry-commented
           txt
-      next. > parsed!
+      parsed.group 1 > group1!
+      parsed.group 2 > group2!
+      next. > parsed
         match.
           compiled.
             regex "/^\\/([\\s\\S]*)\\/([dimsux]*)$/"
           expression
-      "/(?:%s\\E)(?![\\s\\S])/%s".printf > quoted!
-        *
-          parsed.group 1
-          parsed.group 2
+      "/(?:%s\\E)(?![\\s\\S])/%s".printf > quoted
+        * group1 group2
       compile. > [message] > retry-commented
         regex commented
         retry-quoted
--- a/eo-runtime/src/main/eo/string/regex.eo
+++ b/eo-runtime/src/main/eo/string/regex.eo
@@ -82,11 +82,11 @@
         and.
           found.to.eq txt.length
           found.from.eq 0
-      "/(?:%s)(?![\\s\\S])/%s".printf > anchored
+      "/(?:%s)(?![\\s\\S])/%s".printf > anchored!
         *
           parsed.group 1
           parsed.group 2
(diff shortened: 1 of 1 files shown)


morphqdd:
Correction: forcing `parsed!` directly broke `EOregexEOcompileTest.matchesEntireTextAfterBacktracking` on all platforms. `parsed` is a match-result object (not a scalar), and `!` only works on byte-representable values, so forcing it threw during dataization.

Fixed by hoisting `parsed.group 1` and `parsed.group 2` into their own forced locals (`group1!`, `group2!`), since those are actual strings and are what's genuinely read three times each (once inside anchored, commented, quoted). `anchored`, `commented`, `quoted` themselves are left unforced, since each is only read once elsewhere.

All EOregexTest, EOregexEOcompileTest and EOuriTest pass now.

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4824](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6879, opened 2026-08-15
#6827: use 64-bit _ftime64_s for the Windows clock

### LINKED ISSUE #6827: `clock` fails to represent dates beyond 2038 due to usage of 32-bit Windows time API
## What happens

On Windows, `clock` obtains seconds with `_ftime32_s`:

```eo
timeb.time.times 1000
millitm.
  output. >> timeb
    win32
      "_ftime32_s"
      *
```

The Microsoft C runtime documents `_ftime32` and `_ftime32_s` as representing
dates only through 23:59:59 UTC on 18 January 2038. After that boundary,
`clock` can no longer provide the documented current time in milliseconds since
the Unix epoch on Windows.

The POSIX branch already uses `gettimeofday` and has no corresponding 32-bit
time restriction.

## Expected behavior

`clock` should keep returning the current Unix-epoch millisecond value after
2038 on every supported Windows build, just as its public documentation
promises.

## Suggested fix

Use the 64-bit CRT API and structure (`_ftime64_s` / `__timeb64`) in the
Windows implementation, then keep the existing seconds-times-1000 plus
milliseconds calculation.

Microsoft documents the distinction here:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/ftime-ftime32-ftime64?view=msvc-170

morphqdd:
`clock`'s Windows branch fills a `struct _timeb` via `_ftime32_s`, which Microsoft documents as only representing dates through 23:59:59 UTC on 18 January 2038. The POSIX branch already uses `gettimeofday`, which has no such limit.

Switched to the 64-bit CRT API: `_ftime64_s` filling a `struct __timeb64`, whose `time` field is a 64-bit `__time64_t`. The JNA structure's `time` field changed from `int` to `long` to match, and the dispatch table in `NamedFuncCall` was updated to route the new syscall name to the same `FtimeFuncCall`. No change to the millisecond calculation itself, or to the POSIX branch.

`EOclockTest` and the eo-runtime build pass. The Windows-specific code path only actually runs on the `windows-2022` CI leg, since JNA structures compile on every platform but the syscall itself is Windows-only.

Closes #6827

### DIFF
--- a/eo-runtime/src/main/eo/clock.eo
+++ b/eo-runtime/src/main/eo/clock.eo
@@ -1,5 +1,5 @@
 # Wall-clock time source backed by the operating system.
-# Reads the underlying syscall: `gettimeofday` on POSIX, `_ftime32_s` on Windows.
+# Reads the underlying syscall: `gettimeofday` on POSIX, `_ftime64_s` on Windows.
 # Decorates an `i64` holding the current time in milliseconds since the Unix epoch.
 
 +architect yegor256@gmail.com
@@ -16,7 +16,7 @@
         millitm.
           output. >> timeb
             win32
-              "_ftime32_s"
+              "_ftime64_s"
               *
       plus.
         times.
--- a/eo-runtime/src/main/java/org/eolang/win32/FtimeFuncCall.java
+++ b/eo-runtime/src/main/java/org/eolang/win32/FtimeFuncCall.java
@@ -12,7 +12,7 @@ import org.eolang.Phi;
 import org.eolang.Syscall;
 
 /**
- * The msvcrt _ftime32_s function call.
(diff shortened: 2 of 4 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4836](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6881, opened 2026-08-15
`PhDefault.loaded()` races on its own lazy attribute-table init, despite the class's own "thread-safe" claim


morphqdd:
## What happens

`PhDefault`'s own javadoc says "The class is thread-safe", but `loaded()` is an unsynchronized check-then-act lazy init:

```java
private Map<String, Attribute> loaded() {
    if (this.attrs == null) {
        this.attrs = PhDefault.defaults();
        for (final Map.Entry<String, Attribute> ent : this.initial.entrySet()) {
            this.add(ent.getKey(), ent.getValue());
        }
    }
    return this.attrs;
}
```

`this.attrs` is a plain, non-volatile field. Two threads calling `.take()` (or any other entry point routed through `loaded()`) on the same not-yet-warmed `PhDefault` instance at the same time can both see `this.attrs == null` and both start populating `this.attrs` and `this.order` (a plain `ArrayList`, mutated inside `add()`) concurrently, corrupting both.

Reproduced with a JUnit test that constructs one `PhDefault` with 8 attributes and hits it with 32 threads via `com.yegor256.Together`, repeated 20 times. Fails reliably (2/20 in one run) with:

```
java.lang.IndexOutOfBoundsException: Index 3 out of bounds for length 3
	at java.util.ArrayList.set(ArrayList.java:470)
	at org.eolang.Bindings.put(Bindings.java:78)
	at org.eolang.PhDefault.add(PhDefault.java:329)
	at org.eolang.PhDefault.loaded(PhDefault.java:568)
	at org.eolang.PhDefault.take(PhDefault.java:230)
```

`eo-runtime`'s own test suite runs with real thread concurrency (`junit.jupiter.execution.parallel.enabled = true`, concurrent mode, in `eo-runtime/pom.xml`), and shared/cached global objects (via `PhPackage`) get copied and accessed from many parallel tests, so an un-warmed shared `PhDefault` hitting this race in practice is plausible, not just a synthetic scenario.

This is the same class of bug already fixed in `CopiedAttrs` (issue #6337, PR #6877): a check-then-act on a non-thread-safe collection with no lock, on a class whose own contract claims thread safety.

## What should happen

`PhDefault.loaded()` (and `add()`, which mutates the same state) should be safe to call concurrently from multiple threads on the same instance, matching the class's own documented "thread-safe" contract.

morphqdd:
[...]

PR #6882, opened 2026-08-15
#6881: guard PhDefault's lazy attribute load against concurrent init

### LINKED ISSUE #5723: `PhDefault` method `activate()` lacks proper thread-safety despite class documentation claim

morphqdd:
`PhDefault.loaded()` lazily initializes `this.attrs` with an unsynchronized check-then-act, and `add()` mutates the same `this.attrs` plus a plain `this.order` list with no lock, despite the class's own javadoc claiming "The class is thread-safe."

Synchronized both `loaded()` and `add()` on `this.order`, the same pattern already used to fix the same class of bug in `CopiedAttrs` (#6337, PR #6877). `add()` is called both from within `loaded()`'s own initial-population loop and directly by external callers, so both methods need to share the same lock; `synchronized` is reentrant per thread, so the recursive `add()` -> `loaded()` call from within the lock is safe.

Added `PhDefaultRaceTest` (new file, kept separate since `PhDefaultTest.java` was already near qulice's 1000-line file cap): constructs one un-warmed `PhDefault` with 8 attributes and hits it with 32 threads via `com.yegor256.Together`, repeated 20 times. Verified it fails reliably on the pre-fix code (2/20 runs, `IndexOutOfBoundsException` in `Bindings.put`/`PhDefault.add`) and passes cleanly on the fix (20/20, several runs).

Closes #5723

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -14,6 +14,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
@@ -106,6 +107,13 @@ public class PhDefault implements Phi, Cloneable {
     private Map<String, Attribute> attrs;
 
     /**
+     * Guards {@link #attrs} and {@link #order} against concurrent lazy init.
+     * Not final: {@link #copy()} gives the copy its own, since a copy's
+     * lazy init is independent of the origin's.
+     */
+    private ReentrantLock lock;
+
+    /**
      * Default ctor.
      */
     public PhDefault() {
@@ -158,6 +166,7 @@ public class PhDefault implements Phi, Cloneable {
         this.data = new Snapshot(dta);
         this.initial = attributes;
         this.order = new ArrayList<>(0);
+        this.lock = new ReentrantLock();
     }
 
     @Override
@@ -174,6 +183,7 @@ public class PhDefault implements Phi, Cloneable {
     public final Phi copy() {
         try {
             final PhDefault copy = (PhDefault) this.clone();
+            copy.lock = new ReentrantLock();
             copy.attrs = new CopiedAttrs(this.loaded(), copy);
             return copy;
         } catch (final CloneNotSupportedException ex) {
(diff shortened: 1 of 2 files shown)


morphqdd:

eo-runtime/src/main/java/org/eolang/PhDefault.java
```diff
@@ -322,11 +322,14 @@ public String forma() {
      * @param name The name
      * @param attr The attr
      */
+    @SuppressWarnings("PMD.AvoidSynchronizedStatement")
```
yegor256: @morphqdd better use a proper reentrant lock

morphqdd: Switched to ReentrantLock.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4859](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2563](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6884, opened 2026-08-15
#6883: force normalized to dataize once in path.eo

### LINKED ISSUE #6883: `path.eo`'s inner `normalized` re-runs its split/fold/join pipeline three times per call, unlike its siblings
## What happens

`path.eo`'s `impl`'s `[] > normalized` formation has an inner attribute also named `normalized` (`as-bytes. > normalized`, doing `string.split`, a `reduced.` fold over path segments with `..`-collapsing logic, then `string.joined`) bound without `!`. The formation's own `@` reads that inner `normalized` three times (`normalized.eq ...`, `normalized.size.eq 0`, `string normalized`), so every call to `.normalized` on a path pays for the full split-fold-join pipeline three times over instead of once.

This is the same pattern already fixed elsewhere in this file: `basename`'s `pth!`/`index!`, `dirname`'s `pth!`/`len!`, and `extname`'s `index!`/`base!` are all already `!`-forced for exactly this reason. `normalized` was missed.

## What should happen

`normalized` should be `!`-forced (`as-bytes. > normalized!`) so its expensive fold-and-join computation runs once per `.normalized` call instead of three times, matching the existing convention already used by the other attributes in this same file.

morphqdd:
`path.eo`'s inner `normalized` attribute (the split/fold-`..`/join pipeline) is read three times from its own formation's `@`, without `!`, so every `.normalized` call redoes the full pipeline three times. Every other similar local in this same file (`pth!`, `index!`, `len!`, `base!`) is already `!`-forced; this one was missed.

Added `!`. `EOpathTest` (58 tests, including 19 `can-normalize-*` cases) passes.

Closes #6883

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -92,7 +92,7 @@
             normalized.size.eq 0
             "."
             string normalized
-      as-bytes. > normalized
+      as-bytes. > normalized!
         if.
           driveless.size.eq 0
           if.


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4848](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6887, opened 2026-08-15
fix(#6125): group Transpiling constructor parameters into PhiSettings and Outputs

### LINKED ISSUE #6125: `Transpiling` constructor takes 11 positional parameters, several of the same type

VasilevNStas:
Fixes #6125.

## Problem

`Transpiling` took 12 positional parameters, several of the same type (4 `Path`, 3 `boolean`, 2 `String`), so two slots could be swapped at the call site with no compiler or lint catching it. The class carried `@checkstyle ParameterNumberCheck` and `@SuppressWarnings("PMD.ExcessiveParameterList")` to stay green.

## Solution

Group the related parameters behind named types, as the issue suggests:

- **`PhiSettings`** — bundles the two values that describe what the transpiler writes into the generated Java: whether located objects are wrapped into `PhCoverage` and which class a generated class extends. Both were already folded together into the transpile cache key, so they travel as one value through `Transpiling` and `Transpilation`.
- **`Outputs`** — bundles the four paths the transpiler writes into: target, generated sources, base cache, and the XSL measures file. `Transpiling` stores it as a field (no constructor method calls), and `Transpilation` receives it too.

The constructor goes from 12 parameters to 8, and both `ParameterNumberCheck`/`ExcessiveParameterList` suppressions are removed.

## Changes

- `eo-maven-plugin/.../PhiSettings.java` — new value object (coverage + superclass).
- `eo-maven-plugin/.../Outputs.java` — new value object (target, generated, cache, measures).
- `eo-maven-plugin/.../Transpiling.java` — constructor takes `Outputs` + `PhiSettings`; stores `Outputs` as a field.
- `eo-maven-plugin/.../Transpilation.java` — takes `PhiSettings` + `Outputs`; suppressions removed.
- `eo-maven-plugin/.../MjTranspile.java` — builds the two value objects at the call site.
- Tests updated.

## Tests

- `mvn -pl eo-maven-plugin test` — 631 of 633 pass (the 2 `MjFormatTest` failures are pre-existing on master, verified separately).
- `mvn -pl eo-maven-plugin -Pqulice verify -PskipTests` — clean.

@yegor256 please take a look.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Outputs.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Outputs.java
@@ -1,0 +1,86 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import java.nio.file.Path;
+
+/**
+ * The paths the transpiler writes into: where the parsed XMIR goes, where
+ * the generated Java lands, the base cache directory, and the file where
+ * XSL measurements are stored. They travel together, so they are passed
+ * as one value instead of four {@code Path} parameters.
+ * @since 0.62.0
+ */
+@SuppressWarnings("PMD.DataClass")
+final class Outputs {
+
+    /**
+     * Target directory of the build.
+     */
+    private final Path target;
+
+    /**
+     * Generated sources directory.
+     */
+    private final Path generated;
+
+    /**
+     * Base cache directory.
+     */
+    private final Path cache;
+
+    /**
+     * File where XSL measurements are stored.
+     */
+    private final Path measures;
+
+    /**
+     * Ctor.
+     * @param target Target directory of the build
+     * @param generated Generated sources directory
+     * @param cache Base cache directory
+     * @param measures File where XSL measurements are stored
+     */
+    Outputs(
+        final Path target, final Path generated, final Path cache, final Path measures
+    ) {
+        this.target = target;
+        this.generated = generated;
+        this.cache = cache;
+        this.measures = measures;
+    }
+
+    /**
+     * Target directory of the build.
+     * @return The directory
+     */
+    Path target() {
+        return this.target;
+    }
+
+    /**
+     * Generated sources directory.
+     * @return The directory
+     */
+    Path generated() {
+        return this.generated;
+    }
+
+    /**
+     * Base cache directory.
+     * @return The directory
+     */
+    Path cache() {
+        return this.cache;
+    }
+
+    /**
+     * File where XSL measurements are stored.
+     * @return The file
+     */
+    Path measures() {
+        return this.measures;
+    }
+}
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java
@@ -148,16 +148,17 @@ public final class MjTranspile extends MjSafe {
         new Timed(
             new Transpiling(
                 this.scopedTojos().standalone(),
-                this.targetDir.toPath(),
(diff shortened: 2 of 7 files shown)


VasilevNStas:
@yegor256 

Note: the only failing check is `pr-size` (288 lines) — it is informational and non-blocking (e.g. #6804 +489, #6853 +346 were merged with it red). All real checks pass (mvn×4, qulice, counts, integration, snippets, SonarCloud). This is a refactor across 7 files plus two new value objects; happy to split via PDD puzzles if preferred.

eo-maven-plugin/src/main/java/org/eolang/maven/Outputs.java
```diff
@@ -0,0 +1,86 @@
...
+import java.nio.file.Path;
+
+/**
+ * The paths the transpiler writes into: where the parsed XMIR goes, where
+ * the generated Java lands, the base cache directory, and the file where
+ * XSL measurements are stored. They travel together, so they are passed
+ * as one value instead of four {@code Path} parameters.
+ * @since 0.62.0
+ */
+@SuppressWarnings("PMD.DataClass")
```
yegor256, line 16: @VasilevNStas this is a right hint: we don't do data classes, they are anemic, see: https://www.yegor256.com/2016/07/06/data-transfer-object.html

VasilevNStas:
@yegor256 thank you for the review — you're right, and that PMD warning is a fair hint: `Outputs` and `PhiSettings` are exactly the anemic data classes this codebase avoids. Bundling positional parameters into a value holder with getters just moves the problem around instead of giving the object a responsibility. I read the article and I agree the two classes as written are not the right shape.

I also notice the issue is assigned to @morphqdd, so I'm closing this PR and leaving #6125 to its assignee rather than duplicating the work. Thank you for the guidance.

VasilevNStas:
Closing: the data-class approach is rejected per review, and issue #6125 is assigned to @morphqdd. Thank you for the review.

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2241](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6906, opened 2026-08-15
Say when a dispatch cannot be answered

maxonfjvipon:
`needs.xml` records all 10,414 dispatches of `eo-runtime`. `links.xml` answered 10,078 of them and said nothing whatsoever about the other 336, because a dispatch falls between the two rules that write rows: `Links` writes for a base that names another object, and a dispatch base names none; `Dispatched` writes when it can work the dispatch out, and for these it could not.

```xml
<type id="Φ.clock.φ">
  <unknown/>
</type>
```

An absent row said that, and also said "nobody looked" — which is what the 2,521 formations and 1,079 voids are silent for, since they are known and known elsewhere. A reader had no way of telling the two apart. Now silence in `links.xml` means one of those three kinds, and nothing else.

`Resolved` writes it, because a dispatch is only known to be unanswerable once the passes have stopped adding pairs:

```java
for (final XML dispatch : dispatches) {
    rows.putIfAbsent(dispatch.xpath("@loc").get(0), new Unknown());
}
```

`putIfAbsent` is the whole of the care needed: an answer, where there is one, is kept. A row saying an object is a copy of something nothing describes is also kept — "a copy of `X`" stays true whatever is known about `X`, and replacing it with an admission would throw a fact away.

Two existing packs described the old silence and now describe the admission — `atom-that-comes-back-with-a-variable-stays-quiet` and `delegation-that-comes-back-is-walked-once`, both of which asserted that the dispatch got no row. Two new packs cover the change itself and the case it must not touch.

On `eo-runtime`: 336 rows, 39,120 to 39,456, and **every number on the ladder identical** — 514 on the bottom rung before and after, depth 65.7% before and after. These objects were already counted as unknown; what changes is that the table now says so. Two runs of the goal produce the same file.

What is left silent is three kinds of object that are all known: a formation, a void, and the `λ` of an atom. Each is worth a row of its own and none of them is this ticket.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Resolved.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Resolved.java
@@ -42,6 +42,12 @@ import java.util.Map;
  * argument lands in means knowing which formation is being copied, and that is
  * what the pairs have just settled.</p>
  *
+ * <p>So is the admission that a dispatch could not be worked out, for the same
+ * reason in reverse: only here, when the passes have stopped adding pairs, is
+ * it known that no pass will answer it. A row saying nothing is known is worth
+ * writing, since an absent row says that and also says "nobody looked", and a
+ * reader has no way of telling which.</p>
+ *
  * @since 0.68.0
  */
 public final class Resolved implements Clue {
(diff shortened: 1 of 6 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+871](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6908, opened 2026-08-15
Say that a void is a variable

maxonfjvipon:
A chain of copies could be followed inside `links.xml` until it landed on a void, and there it fell off the table. `Φ.inc.φ` is a copy of `Φ.inc.x`, that row exists, and nothing was keyed by `Φ.inc.x` — so a reader who had been working entirely inside this document had to open `provides.xml` to learn whether the chain had arrived at a variable, at something nothing describes, or at nothing at all.

```xml
<type id="Φ.inc.x">
  <var/>
</type>
```

1,079 of them in `eo-runtime`, and not a corner case: the core objects *are* voids — `[as-bytes] > number` — which is why 8,070 objects, 18.7% of the world, are names rooted at one. Following a chain into that country is the common case, not the exception.

No `id` attribute, though `eo-inference-spec.md` §8.3 writes `<var id="…"/>`. The row is keyed by the locator, the locator is the name every mention of the variable is written with, and an attribute repeating its own row's key is noise. The spec line is the thing to change.

`Links` writes it, beside the datum and the termination: a void names no other object, so nothing looks for what it is a copy of, and `<o base="∅"/>` is visible in the text without anything being worked out first.

On `eo-runtime`: 1,079 rows, 39,120 to 40,199, and **every number on the ladder identical** — 8,070 names rooted at a void before and after, depth 65.7% both times. A void was already credited through `provides`; what changes is that the links table stops sending its reader elsewhere. Two runs of the goal produce the same file.

That leaves two silences, both of them objects that are known: a formation, whose type is itself, and the `λ` of an atom, which carries the forma it comes back with.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Links.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Links.java
@@ -45,11 +45,11 @@ import java.util.Map;
  *
  * <p>Which is why a row carries what an object is as an element of its own
  * rather than as a cell, and what {@link Types} makes of it. Being a copy is
- * one of the things an object turns out to be, and two more arrive here as
- * well, both of them copies of nothing: a datum, the bytes of a literal being
- * the ground the program stands on, and a termination, which comes back with
- * no value at all. Those are the rows this clue writes without a reference to
- * look at.</p>
+ * one of the things an object turns out to be, and three more arrive here as
+ * well, none of them a copy of anything: a datum, the bytes of a literal being
+ * the ground the program stands on; a termination, which comes back with no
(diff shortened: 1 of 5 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+887](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6910, opened 2026-08-15
Merge the bytes package

maxonfjvipon:
Continues #6657 with `bytes`, the second-to-last package: its 17 members read
their receiver through `^` and `bytes` joins `<mergedPackages>`.

This one was held back in `9e63abca22` because merging it exhausted a 7.6GB
heap in the fifth README snippet — #6781. That is gone. With the same snippet,
the same sandbox and the default heap, the merged build now finishes in 60s,
so the work is simply the revert of the hold-back plus the call sites master
grew in the meantime. #6851 is the likely reason: it made a copy cost what it
reads rather than what it holds, and `bytes` is allocated on every arithmetic
operation, so it gained the most from that.

Two kinds of call site had to move. `map.eo` reached the hash through the
package four times and carried an `+alias bytes.hash` for it:

```
      hash. >> hash!
        key.as-bytes >> kbts!
```

And the 28 tests in `bytes/hash.eo` applied the member standalone
(`hash 42`), the shape #6682 converted everywhere else and left here because
`bytes/hash` was still waiting on #6655. They now dispatch: `42.hash`.

The detector over `target/eo/1-parse` comes back clean — across all eleven
merged packages nothing under `Φ.<package>.<member>` survives except
`number.integral`, `input.dead`, `output.dead` and the void-less `tuple.empty`.
Locally 687 tests pass across `EObytesTest`, `EOmapTest`, `EOnumberTest`,
`EOstringTest` and `PhDefaultTest`, run with `-Deo.deadline=30` so that a slow
one fails instead of being skipped.

`string` is the last package left, and then #6658 and #6659.

### DIFF
--- a/eo-runtime/pom.xml
+++ b/eo-runtime/pom.xml
@@ -314,6 +314,7 @@
               gets a receiver, and with '^' but no merge it has no rho to read.
               -->
               <mergedPackages>
+                <mergedPackage>bytes</mergedPackage>
                 <mergedPackage>chunk</mergedPackage>
                 <mergedPackage>directory</mergedPackage>
                 <mergedPackage>file</mergedPackage>
--- a/eo-runtime/src/main/eo/bytes/array.eo
+++ b/eo-runtime/src/main/eo/bytes/array.eo
@@ -9,7 +9,8 @@
 +spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 +spdx SPDX-License-Identifier: MIT
 
-[b] > array
+[] > array
+  ^ > b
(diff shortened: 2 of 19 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (138 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+899](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6911, opened 2026-08-15
#5757: turn raw coverage hits into an LCOV report

### LINKED ISSUE #5757: Raw coverage hits in MjTranspile are not being converted
The puzzle `5466-7f5b0919` from #5466 has to be resolved: 

https://github.com/objectionary/eo/blob/1e42c57ff17d14bcaa0cf9ee8636bd5243657f29/eo-maven-plugin/src/main/java/org/eolang/maven/MjTranspile.java#L76-L84

The puzzle was created by @yegor256 on 22-Jul-26. 

morphqdd:
Turns the raw `loc:line:pos` coverage hits `PhCoverage` appends at runtime into an LCOV tracefile.

Three pieces:

- `CoverageManifest`: derives which locations `to-java.xsl` would wrap in `PhCoverage`, structurally. `to-java.xsl` (the last shift of the transpile train) decides this from `@line`/`@pos`/`@loc` on the XMIR, attributes already set by every earlier shift, so `CoverageManifest` runs that same prefix and reads the result with the identical XPath condition, instead of pattern-matching the generated Java text.
- `LcovReport`: pure formatter, turns a per-file per-line hit-count map into the LCOV `SF`/`DA`/`LH`/`LF` text.
- `MjCoverageReport`: new Maven goal (`coverage-report`), bound to the `verify` phase since the raw hits only exist after tests actually run (unlike `transpile`, which only decides whether locations get wrapped, long before any test executes). Reads `eo.coverageFile`, matches it against every transpiled source's manifest, writes the LCOV file. No-ops when `eo.coverageFile` is unset, the same contract `PhCoverage` itself follows.

Removed the now-resolved `@todo` puzzle from `MjTranspile.java`; the second one (enforcing a minimum coverage threshold, a separate concern) stays.

Closes #5757

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjCoverageReport.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjCoverageReport.java
@@ -56,13 +56,6 @@ public final class MjCoverageReport extends MjSafe {
     )
     private File lcovFile;
 
-    /**
-     * Ctor.
-     */
-    public MjCoverageReport() {
-        // nothing
-    }
-
     @Override
     public void exec() throws IOException {
         if (this.coverageFile == null || !this.coverageFile.exists()) {
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjCoverageReport.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjCoverageReport.java
@@ -1,0 +1,117 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import com.jcabi.log.Logger;
+import com.jcabi.xml.XML;
+import com.jcabi.xml.XMLDocument;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+
+/**
+ * Turns the raw {@code loc:line:pos} coverage hits {@code PhCoverage}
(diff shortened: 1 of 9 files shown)


morphqdd:

eo-maven-plugin/src/main/java/org/eolang/maven/MjCoverageReport.java
```diff
@@ -0,0 +1,110 @@
...
+ * only decides whether locations get wrapped in {@code PhCoverage} in
+ * the first place, long before any test executes). Silently does
+ * nothing when {@code eo.coverageFile} is not set or the file does not
+ * exist yet, the same "no-op when disabled" contract {@code PhCoverage}
+ * itself follows.</p>
+ *
+ * @since 0.75.0
+ */
+@Mojo(
+    name = "coverage-report",
```
yegor256, line 36: @morphqdd nobody is calling this mojo now. but we can wire it into the `codecov.yml` CI job and publish report to codecov

morphqdd: Wired it in: eo-runtime/pom.xml now runs coverage-report in the verify phase, and codecov.yml uploads the LCOV file alongside the jacoco reports. Also fixed eo.coverageTracking being hardcoded to false in eo-runtime/pom.xml, which made the -Deo.coverageTracking=true flag already in codecov.yml a no-op.

0crat:
@morphqdd Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (492 >= 100); -8 for way too many hits-of-code (492 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4911](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2578](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6917, opened 2026-08-15
Say what must fill a void

maxonfjvipon:
A void was a row with nothing in it. `provides.xml` said that `number`
keeps one, that it is called `as-bytes` and that nobody has filled it,
and there it stopped, because what a void holds is a caller's business.
What its value has to *offer* is nobody's business but the program's,
though, and the program is loud about it: five different names are
taken from that void across `eo-runtime`, and every one of them was
written down in `needs.xml` against the site that did the asking, where
finding them again meant resolving every dispatch of the program a
second time.

`Asked` turns that table round, onto the object each ask arrives at
rather than the one it was made of — a difference that is the whole
point, since a `book` binding no `size` hands the question to the void
it keeps, and it is the void that will have to answer:

```java
final String answer = this.names.getOrDefault(attr.xpath("@type").get(0), "");
if (answer.endsWith(step)) {
    found.computeIfAbsent(
        answer.substring(0, answer.length() - step.length()),
        key -> new LinkedHashMap<>(0)
    ).put(name, answer);
}
```

`Demands` picks out the ones that belong to a void — the void itself and
every name rooted at it, since `x.next.foo` says something about the
`next` too — and `Demanded`, a new decorator around the clues, writes
them into the row the void is already in. It runs after `Resolved`,
because only settled links say where an ask arrived.

They stand side by side, each saying which object it is made of, rather
than one inside the other. Nesting reads better and was what I wrote
first, and it drops 77 of the 2,303: `while` hands its answers through
to the `if` of the `as-bool` of whatever fills `Φ.bool.if`, and nobody
ever asked for that `as-bool` — the delegation walked through it, so
there is no ask to nest the deeper one under.

Measured on a clean build of 171 files, 44,182 objects: 2,303 demands
over 558 of the 1,018 voids, naming 1,676 objects. `Φ.number.as-bytes`
wants `and`, `as-bytes`, `as-number`, `eq` and `size` — which is to say
a `Φ.bytes`, and nothing anywhere had said so before. `needs.xml` comes
out byte-identical, `links.xml` unchanged, every cell of every existing
`provides` row the same, two runs byte-identical, and the goal takes the
same minute it did.

No number on the ladder moves, and none should: this records what a
program asks, not what an object turns out to be. It is the half of the
spec's §6.4 that was never built, and the rungs that hold a name rooted
at a void — 15,428 objects — are where the next of it goes.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Asked.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Asked.java
@@ -1,0 +1,74 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.xml.XML;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Every name the program asks, gathered on the object that answers it.
+ *
+ * <p>{@link Needs} writes a row per dispatch, against the site that did the
+ * asking, since that is what it sees. The same fact is worth having the other
+ * way round — everything ever asked of one object, in one place — and nobody
+ * can turn the table around until the links say what each of those asks
+ * arrived at.</p>
+ *
+ * <p>Which object a name is asked of is not the receiver, and the difference
+ * is the whole point. {@code book.size} asks a {@code book}, and a
+ * {@code book} that binds no {@code size} hands the question to the void it
+ * keeps, so it is that void the name is really asked of. The links have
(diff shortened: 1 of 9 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (415 >= 100); -8 for way too many hits-of-code (415 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+903](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6920, opened 2026-08-16
#6919: PhDefault: drop redundant Arrays.copyOf, Snapshot already copies

### LINKED ISSUE #6919: PhDefault copies its byte-array data twice, on both construction and delta(), after two independent #6378 fixes merged
`PhDefault` now copies its byte-array data twice on write and twice on read, because two independent fixes for #6378 both landed on master. `Snapshot` (introduced in `eo-runtime/src/main/java/org/eolang/Snapshot.java` by #6692, merged 2026-08-12) already defensively copies its input in the constructor and returns a fresh `Arrays.copyOf` from every `bytes()` call — that alone closes #6378. A second, independently-branched fix for the same issue (#6396, merged 2026-08-15) landed on top of it and added its own copying: `PhDefault(byte[] dta)` at `PhDefault.java:137` and `PhDefault(byte[] dta, Map)` at `:146` each do `Arrays.copyOf(dta, dta.length)` before handing the array to the private constructor, which wraps it in `new Snapshot(dta)` (`:159`) — copying it again. `delta()` at `:276` wraps `this.data.bytes()` (`:265`) in a second `Arrays.copyOf(bytes, bytes.length)`, even though `Snapshot.bytes()` (`Snapshot.java:54-62`) already returns its own fresh copy.

Behavior is unaffected — the data is still correctly immutable — but every data-bearing `PhDefault` now allocates and copies its backing array twice instead of once, both on construction and on every `delta()` read. `PhDefault(byte[])` is what `Data.java` uses to wrap every number, string, and bytes literal (`Data.ToPhi`), and `delta()` is invoked on virtually every dataization in the runtime, so this doubles a hot-path cost for no benefit. It's the same class of problem as #6807 (`PhDefault.copy()` doing unnecessary work per attribute) — just introduced here by a merge collision between two branches that fixed the same bug in different places rather than by the original defect itself.

Drop the three `Arrays.copyOf` calls added by #6396 (`PhDefault.java:137`, `:146`, `:276`) and let `Snapshot` remain the single owner of the defensive-copy guarantee it already provides.

morphqdd:
`PhDefault` was copying its byte-array data twice on both construction and `delta()`, because two independent fixes for #6378 landed on master: `Snapshot` already defensively copies its input in the constructor and returns a fresh copy from every `bytes()` call, and `PhDefault(byte[])`, `PhDefault(byte[], Map)`, and `delta()` each added their own `Arrays.copyOf` on top of that.

This drops the three redundant `Arrays.copyOf` calls and lets `Snapshot` remain the single owner of the defensive-copy guarantee. No behavior change, just removes duplicate allocation on a hot path (`delta()` runs on virtually every dataization).

Added two tests to `PhDefaultTest` verifying the immutability contract still holds through `PhDefault` after removing the redundant copies: mutating the source array after construction doesn't affect the object, and mutating a previously returned `delta()` array doesn't affect later calls.

Closes #6919

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhDefault.java
+++ b/eo-runtime/src/main/java/org/eolang/PhDefault.java
@@ -10,7 +10,6 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -134,7 +133,7 @@ public class PhDefault implements Phi, Cloneable {
      * @param dta Object data
      */
     public PhDefault(final byte[] dta) {
-        this("", Arrays.copyOf(dta, dta.length), PhDefault.NONE);
+        this("", dta, PhDefault.NONE);
     }
 
     /**
@@ -143,7 +142,7 @@ public class PhDefault implements Phi, Cloneable {
      * @param attributes Initial attributes to register
      */
     public PhDefault(final byte[] dta, final Map<String, Attribute> attributes) {
-        this("", Arrays.copyOf(dta, dta.length), attributes);
+        this("", dta, attributes);
     }
 
     /**
@@ -273,7 +272,7 @@ public class PhDefault implements Phi, Cloneable {
                 this.forma()
             );
         }
-        return Arrays.copyOf(bytes, bytes.length);
+        return bytes;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (30) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4871](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6905, opened 2026-08-15
Unhandled dispatches leave no trace in `links.xml`


maxonfjvipon:
`needs.xml` records all 10,414 dispatches of `eo-runtime` — who is asked for what. `links.xml` answers 10,078 of them and says nothing whatsoever about the other 336:

```xml
<o base=".as-i64" loc="Φ.clock.φ"/>
```

There is no row for `Φ.clock.φ`. `Links` writes rows for objects whose base names another object, and a dispatch base names none; `Dispatched` writes one when it can work the dispatch out, and here it could not, because nothing describes the receiver. So the object falls between the two and leaves no trace at all.

A reader cannot tell that apart from the silence around the 2,521 formations and 1,079 voids, which are silent for the opposite reason — they are known, and known somewhere else. `eo-inference-spec.md` §8.3 puts it as a rule: a row is written for every object of the world, including when the answer is that nothing is known, because silence is ambiguous between "not looked at" and "looked at, nothing known", and only one of those is a fact.

Since #6853 the form costs a class:

```xml
<type id="Φ.clock.φ">
  <unknown/>
</type>
```

`Resolved` is where it belongs, since a dispatch is only known to be unanswerable once the passes have stopped adding pairs. The 336 are ordinary names on receivers nothing describes — `.if` 49 times, `.eq` 49, `.not` 28, `.code` 17, `.head` 16 — spread over most of the runtime.

Two things this must not do. It must not overwrite a dispatch that *was* answered, and it must not touch a row saying an object is a copy of something nothing describes: "a copy of `X`" stays true whatever is known about `X`, and replacing it with "unknown" would throw a fact away. Nor does it move any number on the ladder — these objects are on the bottom rung already and stay there. It is bookkeeping, and the bookkeeping is the point.

After it, silence in `links.xml` means one of exactly three things — a formation, a void, or the `λ` of an atom — all of which are known and none of which has a row yet. Those are three more tickets, not this one.

maxonfjvipon:
Done in #6906 (c56bdc4e53). An unanswered dispatch is written `<unknown/>` rather than left without a row: 336 of them.

0crat:
@maxonfjvipon Thanks for closing this issue! You've earned +7 points for this: +8 as a basis; -1 for 2 days of delay. Your running score is [+941](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6922, opened 2026-08-16
feat(#6915): gather void demands onto the void in the provides table

### LINKED ISSUE #6915: provides.xml does not specify what should fill a void attribute in the code
`provides.xml` writes a void down as an attribute and stops there. In
`eo-runtime/src/main/eo/number.eo:12` the whole of `number` is a void:

```
[as-bytes] > number
  as-bytes > @
```

and the row we write for it, from `Provides.java:95`, is
`<attr name="as-bytes" type="Φ.number.as-bytes" void="true"/>` — the name, the
locator, and the fact that it is empty. The program, meanwhile, says a great
deal about what has to go in there: `is-finite` takes `size` off it, `as-i64`
takes `and`, forty-one places take `eq`, one takes `as-bytes`. Every one of
those asks is written down, in `needs.xml`, against the site that did the
asking. So the fact is in the tables and still cannot be looked up: finding it
means resolving every dispatch in the program again to see which ones landed on
this void.

That is the missing half of §6.4 of the type-inference spec. A dispatch on a
variable makes a fresh variable — we do that, since #6741, and
`Φ.number.as-bytes.size` is a name we write — and it records the name into the
variable's demand, which we do not. Without the second half a void is a type
with no shape: the tables cannot say that whatever fills `number` has to be a
`bytes`, although the four names above say exactly that, and `Φ.bytes` has all
four of them.

Measured on a clean build of master, 171 files and 44,182 objects: 3,145 of the
10,800 rows in `needs.xml` are asked of a void or of a name rooted at one, over
1,580 distinct variables under 551 of the 1,018 voids. 2,029 (variable, name)
facts, and we keep none of them where they belong.

Gather them onto the void, as children of its row:

```xml
<attr name="as-bytes" type="Φ.number.as-bytes" void="true">
  <demand name="eq" type="Φ.number.as-bytes.eq"/>
  <demand name="size" type="Φ.number.as-bytes.size"/>
</attr>
```

A demand of a deeper variable nests under the demand that named it, so that
`x.next.foo` reads as a `foo` under a `next`. The gathering needs the resolved
links, since a receiver is known to be a void only after `Resolved` has run, so
it belongs after that pass and not inside `Needs`; `Ends` already walks a chain
to what it ends at, and the voids are `//attr[@void='true']/@type`.

No number on the ladder moves, and none should — this records what a program
asks, not what an object turns out to be. Rungs 1 and 2 hold 15,428 objects
whose type is a name rooted at a void, and a demand is what makes such a name
mean anything.

Behaviour is described by YAML packs in
`eo-maven-plugin/src/test/resources/org/eolang/maven/inference-packs`, one pack
per behaviour, a pack able to describe a program of several files.

VasilevNStas:
Fixes #6915.

## Problem

`provides.xml` writes a void down as an empty `<attr void="true"/>` and stops there. What the program asks of that void — the names taken off it (`is-finite` takes `size`, forty-one places take `eq`, …) — sits in `needs.xml` against the site that asked, and cannot be looked up from the void without resolving every dispatch again. A void is a type with no shape in the tables.

## Solution

Gather the demands onto the void after the links are resolved. A new pass, `Demands`, runs once `Resolved` has closed the links, reads `provides.xml`/`needs.xml`/`links.xml`, resolves each demand's receiver through the chain of copies, and writes every demand that lands on a void — or on a name rooted at one — back into the void's row:

```xml
<attr name="as-bytes" type="Φ.number.as-bytes" void="true">
  <demand name="eq" type="Φ.number.as-bytes.eq"/>
  <demand name="size" type="Φ.number.as-bytes.size"/>
</attr>
```

A demand of a deeper variable nests under the demand that named it, so `x.next.foo` reads as a `foo` under a `next` under the void. Demands whose receiver does not resolve to a void are left where they are.

## Changes

- `eo-inference/src/main/java/org/eolang/inference/Demands.java` — new clue: resolves every demand to its void and amends the provides table.
- `eo-maven-plugin/src/main/java/org/eolang/maven/Inferring.java` — runs `Demands` after `Resolved`.
- `eo-maven-plugin/src/test/resources/org/eolang/maven/inference-packs/void-gathers-its-demands.yaml` — demand gathered and nested.
- `eo-maven-plugin/src/test/resources/org/eolang/maven/inference-packs/non-void-keeps-no-demands.yaml` — an ordinary attribute gathers nothing.

## Tests

- `InferringTest` — 46 packs pass (44 existing + 2 new).
- `eo-inference` — 23 tests pass.
- `eo-maven-plugin` — 635 tests pass (no regressions; the two known `MjFormatTest` failures are pre-existing on master).
- `qulice` — clean for both modules.

@yegor256 please take a look.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Demands.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Demands.java
@@ -1,0 +1,243 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.xml.XML;
+import com.jcabi.xml.XMLDocument;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import org.xembly.Directives;
+import org.xembly.Xembler;
+
+/**
+ * What a program asks of every void, gathered onto the void that is asked.
+ *
+ * <p>{@link Provides} writes a void down as an empty attribute and
+ * {@link Needs} records what somebody wants from it, but the two never
+ * meet: finding out that a demand landed on a void means resolving every
+ * dispatch in the program again. That is what this clue does, once
+ * {@link Resolved} has closed the links. A demand whose receiver resolves
+ * to a void, or to a name rooted at one, is gathered onto that void as a
+ * {@code demand} child of its row:</p>
+ *
+ * <pre> &lt;attr name="as-bytes" type="Φ.number.as-bytes" void="true"&gt;
+ *   &lt;demand name="size" type="Φ.number.as-bytes.size"/&gt;
+ * &lt;/attr&gt;</pre>
+ *
+ * <p>A demand of a deeper variable nests under the demand that named it,
+ * so {@code x.next.foo} reads as a {@code foo} under a {@code next}. A
+ * demand whose receiver does not resolve to a void is left where it is —
+ * the tables record what a program asks, not what an object turns out to
+ * be.</p>
+ *
+ * @since 0.67.0
+ */
+public final class Demands implements Clue {
+
+    @Override
+    public void follow(final Path xmirs, final Path tables) throws IOException {
+        final Path provided = tables.resolve("provides.xml");
+        final Path needed = tables.resolve("needs.xml");
+        final Path linked = tables.resolve("links.xml");
+        if (Files.exists(provided) && Files.exists(needed) && Files.exists(linked)) {
+            final XML gives = new XMLDocument(provided);
+            if (!gives.xpath("//attr[@void='true']/@type").isEmpty()) {
+                Demands.amend(provided, gives, needed, linked);
+            }
+        }
+    }
+
+    /**
+     * Gather the demands onto the voids and write the provides table again.
+     * @param provided The provides document
+     * @param gives The provides document, for the base of the rewrite
+     * @param needed The needs document
+     * @param linked The links document
+     * @throws IOException If the table cannot be written
+     */
+    private static void amend(
+        final Path provided, final XML gives, final Path needed, final Path linked
+    ) throws IOException {
+        final Map<String, List<Demands.Demand>> roots = Demands.gather(
+            new XMLDocument(needed),
+            gives.xpath("//attr[@void='true']/@type"),
+            new Ends(Demands.pairs(new XMLDocument(linked))).names()
+        );
+        if (!roots.isEmpty()) {
+            final Directives dirs = new Directives();
+            for (final Map.Entry<String, List<Demands.Demand>> root : roots.entrySet()) {
+                dirs.xpath(String.format("//attr[@type='%s']", root.getKey()));
+                for (final Demands.Demand demand : root.getValue()) {
+                    Demands.build(dirs, demand);
+                }
+            }
+            Files.write(
+                provided,
+                new XMLDocument(
+                    new Xembler(dirs).applyQuietly(gives.node())
+                ).toString().getBytes(StandardCharsets.UTF_8)
+            );
+        }
+    }
+
+    /**
+     * The resolved pairs of the links table: every name against the one it
+     * is a copy of.
+     * @param links The links document
+     * @return The pairs
+     */
+    private static Map<String, String> pairs(final XML links) {
+        final Map<String, String> pairs = new HashMap<>(0);
+        for (final XML type : links.nodes("/links/type")) {
+            for (final String loc : type.xpath("ref/@loc")) {
+                pairs.put(type.xpath("@id").get(0), loc);
+            }
+        }
+        return pairs;
+    }
+
+    /**
+     * Every demand of the program whose receiver resolves, by where the
+     * receiver resolves to.
+     * @param needs The needs document
+     * @param names Where every locator resolves to
+     * @return The demands, by the resolved receiver
+     */
+    private static Map<String, List<Demands.Demand>> asked(
+        final XML needs, final Map<String, String> names
+    ) {
+        final Map<String, List<Demands.Demand>> where = new LinkedHashMap<>(0);
+        for (final XML type : needs.nodes("/needs/type")) {
+            final String owner = type.xpath("@id").get(0);
+            for (final XML attr : type.nodes("attr")) {
+                final String made = names.get(attr.xpath("@type").get(0));
+                final String asked = names.get(owner);
+                if (made != null && asked != null) {
+                    where.computeIfAbsent(asked, key -> new ArrayList<>(1))
+                        .add(new Demands.Demand(attr.xpath("@name").get(0), made));
+                }
+            }
+        }
+        return where;
+    }
+
+    /**
+     * The demand every locator is produced by, for attaching nested demands.
+     * @param where The demands, by the resolved receiver
+     * @return The demands, by what they resolve to
+     */
+    private static Map<String, Demands.Demand> nodes(
+        final Map<String, List<Demands.Demand>> where
+    ) {
+        final Map<String, Demands.Demand> nodes = new HashMap<>(0);
+        for (final List<Demands.Demand> group : where.values()) {
+            for (final Demands.Demand demand : group) {
+                nodes.put(demand.type, demand);
+            }
+        }
+        return nodes;
+    }
+
+    /**
+     * Every demand of the program that is asked of a void, gathered onto
+     * the void, with deeper demands nested under the one that named them.
+     * @param needs The needs document
+     * @param voids The locators of every void
+     * @param names Where every locator resolves to
+     * @return The root demands, by the void they are asked of
+     */
+    private static Map<String, List<Demands.Demand>> gather(
+        final XML needs, final Collection<String> voids, final Map<String, String> names
+    ) {
+        final Map<String, List<Demands.Demand>> where = Demands.asked(needs, names);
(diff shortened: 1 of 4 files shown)


VasilevNStas:
@yegor256 

Note on the `pr-size` check (273 lines > 200): the feature is a single, coherent addition and does not split cleanly.

The bulk is the new `eo-inference/src/main/java/org/eolang/inference/Demands.java` (~240 lines) — a new inference pass with six small helpers, each required by qulice (cognitive-complexity, nested-if and unnecessary-local rules pushed it from a few long methods into short named ones). The rest is two lines of wiring in `Inferring.java` and two YAML behaviour packs (~28 lines). Splitting this via PDD puzzles would fragment one cohesive rule (resolve a demand to its void, nest deeper demands) into PRs that are individually meaningless.

The check is informational and non-blocking — recently merged PRs of comparable size (#6804 +489, #6853 +346) carried it red. All real checks here pass (mvn×4, qulice, counts, integration, snippets, yamllint after the pack-width fix, SonarCloud). Happy to split if maintainers prefer, but I believe the size is inherent to this feature.

plz review

yegor256:
@maxonfjvipon take a look

maxonfjvipon:
@VasilevNStas you wasn't assigned to the issue, why did you take it?

VasilevNStas:
@maxonfjvipon 

The same honest answer as on #6923: you are right, and I apologize. #6915 is assigned to you (you filed it), and I did not check the assignee — only whether there was an open PR and whether it was merged. That is the same gap in my process, now fixed: the assignee is part of my pre-flight check.

I took #6915 because I had just implemented its sibling #6905 in the same module and wanted to keep the momentum, but that does not justify taking an assigned issue.

Your call, same as on #6923: I can close both PRs and leave the issues to you, or, if you are open to it, keep them as contributor PRs for your review. No hard feelings either way.

eo-inference/src/main/java/org/eolang/inference/Demands.java
```diff
@@ -0,0 +1,243 @@
...
+     */
+    private static Map<String, List<Demands.Demand>> asked(
+        final XML needs, final Map<String, String> names
+    ) {
+        final Map<String, List<Demands.Demand>> where = new LinkedHashMap<>(0);
+        for (final XML type : needs.nodes("/needs/type")) {
+            final String owner = type.xpath("@id").get(0);
+            for (final XML attr : type.nodes("attr")) {
+                final String made = names.get(attr.xpath("@type").get(0));
+                final String asked = names.get(owner);
```
maxonfjvipon, line 125: This is where the coverage goes. `names.get(owner)` is the receiver the ask was made of, and a formation that binds no such name hands the question on: `Φ.book` is not a void and is not rooted at one, so `root()` returns null and the demand is dropped, even though the ask really does land on `Φ.book.pages`.

The object that answers is already in `made` — the answer, `Φ.book.pages.size`. Take the name off the end of it and you have the object it is asked of, for both shapes at once:

```java
final String step = ".".concat(name);
if (made != null && made.endsWith(step)) {
    where.computeIfAbsent(made.substring(0, made.length() - step.length()), …)
}
```

On eo-runtime that is 2,303 facts over 558 voids instead of 1,036 over 443, and `Φ.number.as-bytes` stops being empty.

maxonfjvipon (commented):
I built this branch and ran it over a clean `eo-runtime` (171 files, 44,182 objects) before commenting, so these are measurements rather than opinions.

It gathers **1,036 distinct (object, name) facts onto 443 of the 1,018 voids** — and `Φ.number.as-bytes`, the void the issue is written about, comes out **empty**. That one matters most: `number` binds one void and hands everything to it, so the names the program takes off it (`and`, `as-bytes`, `as-number`, `eq`, `size`) are the whole of what says only a `bytes` will do.

The cause is in `asked()` — the demand is keyed by the receiver the ask was *made of* rather than the object it *arrives at*. A `book` that binds no `size` hands the question to the void it keeps, so the receiver ends at `Φ.book`, `root()` finds no void above it, and the fact is dropped. The links table already worked this out, though: the dispatch resolves to `Φ.book.pages.size`, and the object that answers is that name without its last step. Keying on the answer instead of the receiver picks up both shapes, since a receiver that *is* a void gives the same answer.

A two-file pack that shows it:

```yaml
eo:
  app.eo: |
    [] > app
      book.size > @
  book.eo: |
    [pages] > book
      pages > @
provides:
  - "/provides/type[@id='Φ.book']/attr[@void='true']/demand[@name='size']"
```

This fails on the branch as it stands. Both packs here describe one file; a pack can describe a program of several, and this behaviour needs it — the interesting demands on a void are made from other files.

Two more things the build showed: 1,195 `<demand>` elements were written for those 1,036 facts, so 101 are written twice (the same name asked from two sites lands in the list twice); and the `inference` goal went from one minute to two, which is the `xpath("//attr[@type='…']")`-per-void rewrite — 1,018 XPath scans over a 1 MB document.

For what it is worth, #6917 is an open PR against the same issue, filed before this one, which is how I had the comparison to hand.

eo-inference/src/main/java/org/eolang/inference/Demands.java
```diff
@@ -0,0 +1,243 @@
...
+                if (made != null && asked != null) {
+                    where.computeIfAbsent(asked, key -> new ArrayList<>(1))
```
maxonfjvipon, line 127: A `List` keyed by receiver writes the same fact once per site: 101 of the 1,195 elements in the eo-runtime tables are duplicates of a demand already written. A name asked twice is one demand — a map from name to answer collapses them, and keeps the order deterministic if it is a `LinkedHashMap`.

eo-inference/src/main/java/org/eolang/inference/Demands.java
```diff
@@ -0,0 +1,243 @@
...
+    ) throws IOException {
+        final Map<String, List<Demands.Demand>> roots = Demands.gather(
+            new XMLDocument(needed),
+            gives.xpath("//attr[@void='true']/@type"),
+            new Ends(Demands.pairs(new XMLDocument(linked))).names()
+        );
+        if (!roots.isEmpty()) {
+            final Directives dirs = new Directives();
+            for (final Map.Entry<String, List<Demands.Demand>> root : roots.entrySet()) {
+                dirs.xpath(String.format("//attr[@type='%s']", root.getKey()));
```
maxonfjvipon, line 80: One XPath per void over the whole document is what takes the `inference` goal from one minute to two on eo-runtime — 1,018 scans of a 1 MB table. Selecting the rows once and applying the directives to each node avoids it:

```java
for (final XML hollow : given.nodes("//attr[@void='true']")) {
    new Xembler(demands).applyQuietly(hollow.inner());
}
```

The predicate is also worth narrowing to `[@void='true']`: `//attr[@type='…']` matches any row that happens to carry that locator, not only the void's.

eo-inference/src/main/java/org/eolang/inference/Demands.java
```diff
@@ -0,0 +1,243 @@
...
+ *
+ * <p>A demand of a deeper variable nests under the demand that named it,
+ * so {@code x.next.foo} reads as a {@code foo} under a {@code next}. A
+ * demand whose receiver does not resolve to a void is left where it is —
+ * the tables record what a program asks, not what an object turns out to
+ * be.</p>
+ *
+ * @since 0.67.0
+ */
+public final class Demands implements Clue {
```
maxonfjvipon, line 46: The class has no attributes and every method on it is static, which makes it a utility class behind a `Clue` — something this codebase deliberately avoids. Splitting it the way the rules read gives each object one thing to hold: what is asked of every object (the needs table and the resolved names), the demands that belong to one void, and the clue that writes them into the table.

eo-inference/src/main/java/org/eolang/inference/Demands.java
```diff
@@ -0,0 +1,243 @@
...
+
+        /**
+         * Ctor.
+         * @param name The name asked
+         * @param type What the name resolves to
+         */
+        private Demand(final String name, final String type) {
+            this.name = name;
+            this.type = type;
+            this.children = new ArrayList<>(0);
```
maxonfjvipon, line 240: `children` is filled after the object is built (`parent.children.addAll(…)` below), so a `Demand` is never quite finished when it is made. If the demands are gathered flat and each says which object it is of, the mutation goes away — and the flat shape also keeps the demands whose chain delegation walked through rather than an ask naming it, which nesting has nowhere to put.

eo-inference/src/main/java/org/eolang/inference/Demands.java
```diff
@@ -0,0 +1,243 @@
...
+ *   &lt;demand name="size" type="Φ.number.as-bytes.size"/&gt;
+ * &lt;/attr&gt;</pre>
+ *
+ * <p>A demand of a deeper variable nests under the demand that named it,
+ * so {@code x.next.foo} reads as a {@code foo} under a {@code next}. A
+ * demand whose receiver does not resolve to a void is left where it is —
+ * the tables record what a program asks, not what an object turns out to
+ * be.</p>
+ *
+ * @since 0.67.0
```
maxonfjvipon, line 44: `0.69.0` — 0.67.0 is when the module was born.

VasilevNStas:
Closing this. Thank you for the detailed measurements @maxonfjvipon — they are genuinely helpful and I agree with them:

- Keying on the receiver instead of the object the ask arrives at is wrong, and it is exactly why `Φ.number.as-bytes` came out empty; the links table already has the answer key.
- The 101 duplicated `<demand>`s and the 2x inference slowdown from the per-void XPath rewrite are real defects in this branch.

Most importantly, #6917 (your own implementation) already landed for #6915 — this PR is redundant from the start, and my pre-flight check missed it because its title does not mention the issue number. I should not have taken an issue assigned to you. Lesson recorded: check the issue timeline for linked PRs, not just an open-PR search.

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (6) comments.  Your running score is [+918](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6923, opened 2026-08-16
feat(#6905): write unanswerable dispatches down as unknown in links

### LINKED ISSUE #6905: Unhandled dispatches leave no trace in `links.xml`

VasilevNStas:
Fixes #6905.

## Problem

`needs.xml` records every dispatch, but `links.xml` says nothing about a dispatch that could not be answered — `Links` writes rows only for objects whose base names another object, and `Dispatched` only when it can work the dispatch out. A dispatch whose receiver nothing describes (`.if` 49 times, `.eq` 49, `.not` 28 in the runtime) falls between the two and leaves no trace. That silence is ambiguous between "not looked at" and "looked at, nothing known", and only the second is a fact (`eo-inference-spec.md` §8.3).

## Solution

Once the passes have stopped adding pairs, `Resolved` writes an `<unknown/>` row for every dispatch it could not answer:

```xml
<type id="Φ.app.φ">
  <unknown/>
</type>
```

A dispatch that *was* answered keeps its `ref` row and is never overwritten, and rows saying an object is a copy of something are left alone — the new type is only ever added where nothing was written.

## Changes

- `eo-inference/src/main/java/org/eolang/inference/Unknown.java` — new `Type` that writes `<unknown/>`.
- `eo-inference/src/main/java/org/eolang/inference/Resolved.java` — after the links are settled, `putIfAbsent` an `Unknown` row for every dispatch with no answer.
- Two new inference packs; two existing packs (`atom-that-comes-back-with-a-variable-stays-quiet`, `delegation-that-comes-back-is-walked-once`) updated — they previously asserted the *absence* of a row for an unanswerable dispatch, which is exactly the behaviour this issue changes.

## Tests

- `InferringTest` — 48 packs pass (46 existing + 2 new).
- `eo-inference` — 23 tests pass.
- `qulice` — clean.

@yegor256 please take a look.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Resolved.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Resolved.java
@@ -81,6 +81,9 @@ public final class Resolved implements Clue {
             pairs, new Bound(args, names, new Provided(given, names, voids)).all()
         ).all();
         rows.putAll(written.others());
+        for (final XML dispatch : dispatches) {
+            rows.putIfAbsent(dispatch.xpath("@loc").get(0), new Unknown());
+        }
         Files.write(
             links,
             new Types(rows).asXml().toString().getBytes(StandardCharsets.UTF_8)
--- a/eo-inference/src/main/java/org/eolang/inference/Unknown.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Unknown.java
@@ -1,0 +1,27 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import org.xembly.Directives;
+
+/**
+ * A dispatch that nothing could answer.
+ *
+ * <p>A dispatch is only known to be unanswerable once the passes have
+ * stopped adding pairs: the passes keep going while they add, so a row
+ * written before they settled would call something unanswerable that a
+ * later pass answers. An unknown is a fact like any other — the silence
+ * around an object that was never looked at says something else than the
+ * silence around one that was looked at and gave nothing.</p>
+ *
+ * @since 0.69.0
+ */
+final class Unknown implements Type {
+
+    @Override
+    public Directives directives() {
+        return new Directives().add("unknown").up();
+    }
+}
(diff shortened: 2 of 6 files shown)


yegor256:
@maxonfjvipon check this one

maxonfjvipon:
@VasilevNStas you wasn't assigned to the issue, why did you take it?

VasilevNStas:
@maxonfjvipon 

You are right, and I apologize. I checked for an open PR and whether it was merged, but I did not check the assignee — the issue is assigned to you, and you filed it. That is a gap in my process, not an excuse.

I took it because #6915 (which I had just implemented) and #6905 are in the same module and I had fresh context, but that does not justify taking an assigned issue.

Two options, your call:
1. I close my PR and leave #6905 to you.
2. If you are open to it, I keep the implementation (tests pass, qulice clean) and you review it as a contributor PR — with no hard feelings if you prefer to write it yourself.

Either way, I have updated my workflow: the assignee is now part of my pre-flight check before touching any issue.

Issue #6658, opened 2026-08-12
Inconsistent naming convention for `string` package classes in `eo-runtime`


maxonfjvipon:
Part of #6616. The hand-written atoms of the `string` package live in a Java package of their own:

```
eo-runtime/src/main/java/org/eolang/EO_string/EOprintf.java
eo-runtime/src/main/java/org/eolang/EO_string/EOregex$EOcompile.java
eo-runtime/src/main/java/org/eolang/EO_string/EOregex$EOpattern$EOmatch$EOmatched_from_index.java
eo-runtime/src/main/java/org/eolang/EO_string/PrintfArgs.java
eo-runtime/src/main/java/org/eolang/EO_string/package-info.java
```

That layout follows from `JavaPath`, which turns every segment before the last into an `EO_` package, so `Φ.string.printf` becomes `org.eolang.EO_string.EOprintf` while the inner atom `Φ.number.div` becomes `org.eolang.EOnumber$EOdiv` in `org.eolang` itself. Once the members of the `string` package are attributes of the `string` formation, `Φ.string.printf` is an inner object and the class the transpiler expects is `org.eolang.EOstring$EOprintf`.

Rename the four classes and their tests in `src/test/java/org/eolang/EO_string/` accordingly, and drop the now-empty package and its `package-info`. Only the class names and the package declaration change: `atoms.csv` is keyed by forma, `Φ.string.printf` stays `Φ.string.printf`, and `PhDefault.forma()` reads the locator from XMIR rather than the Java class, so nothing about atom lookup moves.

The rename has to land with the pull request that merges the `string` package, since the generated `EOstring` will look for the new names the moment that package is switched on.

0pdd:
@maxonfjvipon the puzzle [#6932](https://github.com/objectionary/eo/issues/6932) is still not solved.

0pdd:
@maxonfjvipon the only puzzle [#6932](https://github.com/objectionary/eo/issues/6932) is solved here.

PR #6926, opened 2026-08-16
Merge the string package into the string object

maxonfjvipon:
Last package of #6657. Every member of the `string` package now takes its receiver from `^` rather than from a leading void, so `string.length` becomes a plain dispatch:

```
[] > length
  if. > @
    eq.
      text.as-bytes.size >> size!
      ...
  ^ > text
```

That turns `string.last-index-of text separator` into `text.last-index-of separator`, `regex "/linux/i"` into `"/linux/i".regex`, and so on across `path.eo`, `os.eo`, `uri.eo`, `file.eo`, `directory.eo` and the string members themselves.

The two regex atoms had to move in the same commit — `EO_string/EOregex$EOcompile` is now `org.eolang.EOstring$EOregex$EOcompile`, since after the merge there is no `EO_string` Java package to hold them. That is #6658, and it cannot land separately.

Two receiver mistakes worth calling out, both caught by the suite: `as-ascii` is a `string` member, so `bytes/as-hex.eo` and `string/english.eo` had to wrap their byte in a `string` before dispatching on it, rather than reaching for `as-ascii` on a bytes object.

With `string` merged, no package in eo-runtime is left unmerged, so `PhNestTest` keeps only `refusesPackageThatWasNeverTranspiled`, the one test that builds a `PhNest` by hand. The other five had no package object left to point at. `PhNest` itself goes away in #6659.


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (860 >= 100); -8 for way too many hits-of-code (860 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+911](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6929, opened 2026-08-16
#6918: reject horizontal argument applied to the identity glyph

### LINKED ISSUE #6918: `I 5` results in unnamed argument inside identity formation causing formation body read error
`Emissions.identity` (`eo-parser/src/main/java/org/eolang/parser/Emissions.java:294-300`) opens a base-less `<o>` for the glyph, emits the void and the φ, and closes only the φ — deliberately, so that "horizontal arguments land on the identity exactly as they land on a parenthesised `(x > [x])`". The arguments therefore become children of the formation itself, and because that parent carries no `@base`, `mandatory-as.xsl` reads it as a formation body and leaves them without `@as` as well.

So this program

```
+package f

# App.
[] > app
  I 5 > y
```

yields a formation whose third child has neither `@name` nor `@as`:

```xml
<o name="y">
  <o base="∅" name="x"/>
  <o base="ξ.x" name="φ"/>
  <o base="Φ.number" loc="Φ.f.app.y.α2">…</o>
</o>
```

and printing that tree back gives

```
[] > app
  [x] > y
    x > @
    5
```

where `5` is a plain child of a formation carrying no name suffix. R-6.2.1 requires one, so R-5.3.1 rejects it with `object inside formation must have a name`.

This is #6848 under a new spelling. That issue reported the same tree and the same printed output for `(m > [m]) 5`, and afcad9dccf closed it by rejecting a non-empty argument list on a group head that wraps an inline-phi. The guard it added is gated on `head.kind() == Value.Kind.GROUP` (`LnApplication.java:176`). The `I` glyph was already on the branch when that guard was written — 96226c8268 is an ancestor of afcad9dccf — and it arrives as `Value.Kind.IDENTITY`, which the guard does not name, so the shape is reachable again in two characters. Nothing downstream flags it either: the printer test in 3ad1241066 pins the long layout for exactly this object. Same family as #6707, which reports the sibling shape for `T`.

Extending that guard to the identity head, with the positioned error `([x]) 5` already produces, closes it the way #6848 was closed.

@yegor256 please take a look.

morphqdd:
Emissions.identity opens a base-less <o> for the I glyph and leaves it open, so horizontal arguments land on it exactly as they land on a parenthesised (x > [x]). Since the identity always expands to an inline-phi-style formation, any horizontal argument becomes an unnamed formation child.

This is #6848 under a new spelling. That issue rejected the same shape for a paren-group wrapping an inline-phi, gated on head.kind() == Value.Kind.GROUP. The I glyph carries Value.Kind.IDENTITY instead, which the guard did not name.

Extended LnApplication's guard to reject any horizontal argument applied to a head that opens a formation body of its own, whether that head is a paren-group wrapping an inline-phi or the identity glyph. Both now produce the same "horizontal formation not allowed as argument" error.

Verified the existing valid uses of I stay valid: I as a bare value, I named as a formation with a later pipe application (| 5), and I as a body/argument with no horizontal args, all still parse.

Closes #6918

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnApplication.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnApplication.java
@@ -153,10 +153,11 @@ final class LnApplication implements Line {
     }
 
     /**
-     * Reject a paren-group head that cannot carry what follows it: bare
-     * redundant parentheses around a whole top-level expression, or a
-     * horizontal argument list applied to an inline-phi formation the
-     * group wraps.
+     * Reject a head that cannot carry what follows it: bare redundant
+     * parentheses around a whole top-level expression, or a horizontal
+     * argument list applied to a head that opens a formation body of its
+     * own — a paren-group wrapping an inline-phi expression, or the
+     * identity glyph {@code I}, itself sugar for one (#6848, #6918).
      * @param head The head value
      * @param chain Method-dispatch chain following the head (may be empty)
      * @param args Horizontal arguments following the head (may be empty)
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4939](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6930, opened 2026-08-16
Say what the world puts in a void

maxonfjvipon:
Since #6915 a void's row says what its value will have to offer. This is
the other half: what the program is actually seen to put in it.

Every filling was already written down — since #6787 an application
records what it fills as a `bind` in its own row — but scattered across
the rows of the objects doing the filling, 22,818 of them in
`eo-runtime`. `Fillings` turns them round onto the void, and gathers
what went in as a **type** rather than as a locator, which is what makes
the answer worth reading: the void of `bytes` is filled 7,355 times and
comes out as

```xml
<attr name="φ" type="Φ.bytes.φ" void="true">
  <witnessed>
    <union>
      <data/>
      <ref loc="Φ.bytes.as-bytes"/>
    </union>
  </witnessed>
</attr>
```

`Forms` answers which form the links table gave an object — a datum, a
termination, a variable, or the object it is a copy of — and is where
the collapsing happens: every literal of a program has a locator of its
own and all of them are one type. `Union` writes a choice, in the
element the grammar has had a place for since §8.3 and nothing wrote
yet. `Witnessed` is a new decorator around the clues, beside `Demanded`,
so the two halves of a void's row are written by two objects.

A choice longer than eight members becomes `<unknown/>` instead of its
members, per §5.5 of the state document: `Φ.tuple.head` is filled with
264 different types, and a list that long tells a reader nothing except
that nobody has thought about it.

This is evidence of callers and never a contract — nothing may infer a
void's type from it, and nothing in the resolver reads it. The callers a
program happens to have today do not oblige the one written tomorrow.

Measured on a clean build, 171 files, 44,142 objects: 526 of the 1,001
voids come out with a join — 201 of a single member, 301 of two to
eight, 24 too many to say. `links.xml` and `needs.xml` are
byte-identical, every existing `provides` cell and every demand
unchanged, two runs byte-identical, and the goal takes the same time it
did. No number on the ladder moves, and none should.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Fillings.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Fillings.java
@@ -1,0 +1,69 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.xml.XML;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * What the program puts into every void, read back from the links.
+ *
+ * <p>Every application says what it fills, and says it in its own row, where
+ * {@link Bound} wrote it: a {@code bind} names the void and what went into it.
+ * So the fact is in the tables already and lies the wrong way round — to learn
+ * what one void is ever given, a reader walks the row of every object that
+ * ever filled anything, and eo-runtime has 22,818 of those bindings.</p>
+ *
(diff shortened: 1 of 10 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (552 >= 100); -8 for way too many hits-of-code (552 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+907](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6931, opened 2026-08-16
#6913: lower-case the exponent marker of a fractional value too

### LINKED ISSUE #6913: `StUnhex.number` inconsistently prints exponent marker in uppercase and lowercase forms
`StUnhex.number` in `eo-printer/src/main/java/org/eolang/printer/StUnhex.java:91-105` spells the exponent marker two ways. A whole value at or above 2^63 goes through `Double.toString(num).replace('E', 'e')` at line 99 and prints `1.5e20`. A fractional value goes through plain `Double.toString(num)` at line 102, with no replacement, so `0.0001` prints `1.0E-4` and `1.5e-10` prints `1.5E-10`. The split follows whether the value is whole, not how large it is, so one source file can carry both spellings of the same construct.

The lowercase form is the one the project asked for. #5596 introduced the line 99 branch specifically to "lower-case the `E` to match EO's `e` literal syntax", and described the fractional path as already fine because it yields exponent notation — its case was never looked at. Nothing breaks either way: `Tokens.readFloatTail` accepts both `e` and `E`, so the literal reparses to the same double and `format` stays idempotent. It is the printed source that reads inconsistently.

Adding the same `.replace('E', 'e')` to the fractional branch at line 102 makes the two agree, and cannot change any value, since the replacement only touches the marker character.

@yegor256 please take a look.

morphqdd:
StUnhex.number spelled the exponent marker two ways depending on whether the value is whole or fractional. A whole value at or above 2^63 goes through Double.toString(num).replace('E', 'e') and prints lowercase, e.g. 1.5e20. A fractional value went through plain Double.toString(num) with no replacement, so 0.0001 printed 1.0E-4.

#5596 added the replace call for the whole-number branch to match EO's lowercase e literal syntax. The fractional branch was left alone since it already produced exponent notation, but its case was never checked.

Added the same replace('E', 'e') to the fractional branch. Cannot change any value, only the marker character. Added a test with a fractional value below 10^-3 (0.0001) confirming it now prints with a lowercase e.

Closes #6913


morphqdd:
@yegor256, review please

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4942](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6912, opened 2026-08-15
`?.` marker is improperly dropped by printer on shortened base in `to-eo-tree.xsl`


Thayorns:
`eo:fragile-marked` is called from exactly one place in `eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl`, the `otherwise` arm at line 422. The four arms above it — the `Q.`-rooted disambiguation at 360, the same-package shortening at 369, the redundant-ρ-hop drop at 382 and the `$.name` marker at 398 — build the head out of `eo:translate-path` alone and never look at `@fragile`, so a fragile dispatch that takes any of them prints as a plain `.` dispatch.

One shape per arm, each printing without its marker. `foo?.bar` inside a formation that also binds `foo` comes out as `Q.foo.bar`. A reference to the program's own object through its `Φ.<package>.app.bar` form comes out as `app.bar`. `x?.y` reached over a redundant parent hop comes out as `x.y`. `$.zzz?.bar` comes out as `$.zzz.bar`. Change only what decides the arm — rename the shadowing attribute in the first case — and the very same node prints `foo?.bar` with its marker intact.

This is the remainder of #6560. That fix added `eo:fragile-marked` and wired it into the `otherwise` arm alone (243a2607a6), and its regression test `fragile-dispatch-restored.yaml` uses `foo?.first`, a base no other arm can claim, so nothing caught the four that were left behind. The consequence is the one #6560 already named: `?.` and `.` are different operators under R-3.5, and `format -Deo.autoFix` rewrites the source in place, so a fragile dispatch silently becomes a hard one whenever its base happens to be shortened.

Lifting the call out of the `otherwise` arm and wrapping the whole `xsl:choose` in it — `eo:fragile-marked(<choose result>, exists(@fragile))` — marks the last dot of whatever surface the arms produce, and leaves the dotless `*` and `T` heads untouched, which the function already promises.

@yegor256 please take a look.

0crat:
[...]

PR #6933, opened 2026-08-16
#6912: keep the fragile marker on every head-shortening arm

morphqdd:
eo:fragile-marked is called from exactly one arm (the otherwise/fallback) of to-eo-tree.xsl's head choice. The other four arms build the head from eo:translate-path alone and never look at @fragile, so a fragile dispatch that takes any of them loses its ?. marker on print.

Captured the whole xsl:choose result into a string variable and wrap that single value in eo:fragile-marked once, instead of calling it only inside the otherwise arm. The function already passes dotless heads (* and T) through unchanged, so nothing needed to change there.

Added a regression test for the Q. disambiguation arm (a fragile dispatch through a Φ-rooted base whose plain name is shadowed by an in-scope attribute). Left a @todo puzzle for the other three arms (same-package shortening, redundant-ρ-hop drop, explicit $.name marker) to keep this PR under the 200-line size cap; see the added `@todo #6912:30min` in to-eo-tree.xsl.

See the added @todo for the remaining test coverage on #6912.


morphqdd:

PR #6934, opened 2026-08-16
#6912: add regression tests for the remaining fragile-marker arms

morphqdd:
Follow-up to #6933, split solely to respect the 200-line PR-size cap.

Adds regression tests for the other three head-shortening arms that #6933 fixes but didn't have room to test: same-package shortening, redundant-ρ-hop drop, and the explicit $.name marker. No code change, test-only.

References #6912, already closed by #6933.


morphqdd:
Closing per feedback: keeping this as a single PR with a @todo puzzle for the remaining test coverage instead of splitting.

PR #6935, opened 2026-08-16
#6904: don't delete a sibling version's dependency when conflicts are ignored

### LINKED ISSUE #6904: `Resolving.cleanPlace` wrongfully removes sibling version's dependency when `eo.ignoreVersionConflicts` is enabled
## What happens

`-Deo.ignoreVersionConflicts=true` is meant to let a build proceed when two dependencies share `groupId:artifactId` but request different versions, instead of `DpsUniquelyVersioned` throwing (`Resolving.java` `deps()`, around line 238-256).

But `Resolving.resolved()` (line 148) calls `cleanPlace(dir, version)` for every dependency it resolves. `cleanPlace` (lines 209-232) deletes every existing subdirectory of `dir` (`dest/groupId/artifactId/classifier`, shared by every version of that coordinate) whose name isn't the version currently being resolved. This logic is correct for the normal case (a version bump between builds, where only one version should exist on disk), but when `ignoreVersionConflicts` intentionally lets two versions of the same coordinate coexist in one build, resolving version A deletes version B's just-unpacked files, and resolving version B (which runs in parallel, via `Threaded`, worker pool size `2 * availableProcessors()`) deletes version A's.

Reproduced deterministically (no concurrency needed to trigger it): resolve version `1.0` of `g:a` first, then resolve version `2.0` of the same coordinate. After the second resolve, `1.0`'s directory is gone; only `2.0`'s remains. No exception or warning is raised, the build reports success, and the missing version surfaces only later as a confusing classpath/compile failure in a downstream mojo.

In the real parallel path (`Resolving.exec()`'s `Threaded` execution), the outcome is non-deterministic: whichever version's `resolved()` call finishes last on disk wins, and the other version's dependency ends up silently absent from `target/eo`.

## What should happen

`cleanPlace` should not delete a sibling version's directory when `ignoreVersionConflicts` is set and both versions are legitimately part of the current resolution set. It should only remove genuinely stale versions (ones no longer requested by any dependency in the current build), not every version other than the one being resolved right now.

morphqdd:
Resolving.resolved() calls cleanPlace(dir, version) for every dependency it resolves. cleanPlace deleted every existing subdirectory of dir whose name wasn't the version currently being resolved. That is correct for the normal case (a version bump between builds, where only one version should exist on disk), but when eo.ignoreVersionConflicts intentionally lets two versions of the same coordinate coexist in one build, resolving version A deleted version B's just-unpacked files, and resolving version B (which can run in parallel via Threaded) deleted version A's.

cleanPlace now takes the full set of versions legitimately resolved for that coordinate in the current run, computed once in exec() from the same deps() collection Threaded iterates over, and only deletes a subdirectory whose name is outside that set. A genuinely stale version (no longer requested by anything in the current build) still gets deleted; a sibling version that is part of the current resolution set does not.

Added a test resolving two eo-runtime versions from two programs with eo.ignoreVersionConflicts on, asserting both survive on disk.

Closes #6904

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/DepCoordinate.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/DepCoordinate.java
@@ -1,0 +1,55 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.maven;
+
+import org.apache.maven.model.Dependency;
+
+/**
+ * A dependency's coordinate, without its version: {@code groupId:artifactId:classifier}.
+ * @since 0.61.0
+ */
+final class DepCoordinate {
+
+    /**
+     * The dependency.
+     */
+    private final Dependency dependency;
+
+    /**
+     * Ctor.
+     * @param dep The dependency
+     */
+    DepCoordinate(final Dependency dep) {
+        this.dependency = dep;
+    }
+
+    /**
+     * The coordinate string.
+     * @return The coordinate, without the version
+     */
+    String value() {
+        return String.join(
+            ":",
+            this.dependency.getGroupId(),
+            this.dependency.getArtifactId(),
+            this.classifier()
+        );
+    }
+
+    /**
+     * The dependency's classifier, normalized to {@code "-"} when absent.
+     * @return The classifier
+     */
+    String classifier() {
+        final String classifier;
+        if (this.dependency.getClassifier() == null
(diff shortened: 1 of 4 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (178 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5038](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6936, opened 2026-08-16
#6895: reject fractional and non-finite recv sizes

### LINKED ISSUE #6895: `socket.recv` and `socket.as-input.read` incorrectly handle fractional byte count values
`socket.recv` accepts a byte count as an EO `number`, but the platform recv
wrappers only reject a negative value. They convert every remaining value with
`Double.intValue()`, so a fractional request is silently changed before the
native call. `socket.as-input.read` delegates to the same `recv` method.

## Reproduction

With a connected socket that has at least two bytes available:

```eo
s.recv 1.5
```

or equivalently:

```eo
s.as-input.read 1.5
```

returns one byte. In `RecvSyscall.make()` and `RecvFuncCall.make()`, `1.5`
becomes `1` through `new Dataized(params[1]).asNumber().intValue()` before the
buffer is allocated and `recv` is called.

The recent fix for #6833 checks `size < 0`, but it leaves fractional,
non-finite, and out-of-range values unchecked. For example, a positive
infinity becomes `Integer.MAX_VALUE` and is used as an allocation size.

## Expected behaviour

Receive sizes must be finite, non-negative integers that fit the runtime's
supported allocation range. An invalid value should raise an EO failure rather
than request a different number of bytes.

This should be consistent with the checked index/size API used by `chunk.read`,
`bytes.right`, and `string.slice`.

## Suggested fix

Use the runtime's checked natural-number conversion in both recv wrappers, or
validate the value once in `socket.eo` before the syscall. Add regressions for
a fractional request, `nan`, infinity, and a negative request on both platform
paths.

gemshrine:
Fixes #6895

`recv` took its byte count through `new Dataized(params[1]).asNumber().intValue()`, so anything that was not already a whole number was quietly turned into a different request: `1.5` became a one-byte read, and a positive infinity became `Integer.MAX_VALUE` and went straight into `new byte[size]`. The `size < 0` guard added in #6833 only covered the negative case.

Both bridges now take the count through `Expect.Natural`, the same checked conversion `chunk.read`, `bytes.slice` and `chunk.resized` already use, so a fractional, `nan`, infinite, out-of-int-range or negative size fails with an `ExFailure` before anything is allocated. The negative case keeps failing as it did, just with the canonical message.

Two regression tests per platform, both verified against the unfixed code: `rejectsFractionalSizeOn*Recv` reports "Expected org.eolang.ExFailure to be thrown, but nothing was thrown", and `rejectsInfiniteSizeOn*Recv` dies on an `OutOfMemoryError` from the `Integer.MAX_VALUE` allocation. Both pass with the fix. Locally `RecvFuncCallTest` is 4/4 on Windows, `RecvSyscallTest` runs the same four on POSIX, and qulice on `eo-runtime` is clean.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/posix/RecvSyscall.java
+++ b/eo-runtime/src/main/java/org/eolang/posix/RecvSyscall.java
@@ -7,7 +7,7 @@ package org.eolang.posix;
 import java.util.Arrays;
 import org.eolang.Data;
 import org.eolang.Dataized;
-import org.eolang.ExFailure;
+import org.eolang.Expect;
 import org.eolang.Phi;
 import org.eolang.Syscall;
 
@@ -33,13 +33,9 @@ public final class RecvSyscall implements Syscall {
     @Override
     public Phi make(final Phi... params) {
         final Phi result = this.posix.take("return").copy();
-        final int size = new Dataized(params[1]).asNumber().intValue();
-        if (size < 0) {
-            throw new ExFailure(
-                "Can't receive a negative number of bytes '%d'",
-                size
-            );
-        }
+        final int size = new Expect.Natural(
+            new Expect<>("the 'size' argument of recv", () -> params[1])
+        ).it();
         final byte[] buf = new byte[size];
         final int received = CStdLib.INSTANCE.recv(
             new Dataized(params[0]).asNumber().intValue(),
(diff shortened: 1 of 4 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2624](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6937, opened 2026-08-16
fix(#6888): trim repeated trailing path separators

### LINKED ISSUE #6888: path.basename and path.dirname mishandle repeated trailing separators
`path.basename` and `path.dirname` were recently changed in #6820 to ignore a
trailing separator. The shared `trimmed` helper, however, removes exactly one
separator. A path ending in two or more separators therefore still reaches the
old erroneous branch.

## Reproduction

```eo
[] > repeated-trailing-separators
  eq. > basename
    path.posix "foo//".basename
    "foo"
  eq. > dirname
    path.posix "foo//".dirname
    "."
```

The current implementation first changes `"foo//"` to `"foo/"`. `basename`
then takes the text after that remaining final slash and returns `""`; `dirname`
keeps `"foo"`. The same problem affects absolute paths, for example
`path.posix "/foo//"` should have basename `"foo"` and dirname `"/"`.

## Expected behaviour

The number of trailing separators must not affect basename or dirname (except
that the root path itself remains the root):

| path | basename | dirname |
| --- | --- | --- |
| `foo` | `foo` | `.` |
| `foo/` | `foo` | `.` |
| `foo//` | `foo` | `.` |
| `/foo//` | `foo` | `/` |
| `/` | `` | `/` |

## Cause and suggested fix

In `eo-runtime/src/main/eo/path.eo`, `impl.trimmed` has one conditional
`uri.ends-with separator` and one `slice`, so it removes only one separator.
It should remove the complete trailing run while retaining a single separator
for the root path. Regression tests should cover repeated POSIX separators and
the corresponding Win32 separator.

PozziTiv4ik:
@yegor256

Closes #6888

## Summary

- Trim the complete trailing separator run before calculating `basename` or `dirname`.
- Preserve one separator when the input is the root path.
- Cover repeated POSIX and Win32 separators, including absolute POSIX paths.

## Root cause

The shared `trimmed` helper removed exactly one trailing separator. Inputs ending in two or more separators therefore still reached `basename` and `dirname` with a separator at the end.

## Validation

- `git diff --check` — passed.
- Server-side diff — one file, 41 additions and 1 deletion.
- Targeted Maven test execution was attempted, but the sandboxed Java process could not finish resolving the cold Maven dependency graph because outbound DNS/network access is restricted. The PR should remain a draft until repository CI completes.

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -438,10 +438,25 @@
       "/www/html"
     "/www/html"
 
+  eq. ++> can-return-a-basename-ignoring-repeated-posix-separators
+    basename.
+      path.posix "foo//"
+    "foo"
+
+  eq. ++> can-return-a-basename-ignoring-repeated-win32-separators
+    basename.
+      path.win32 "foo\\\\"
+    "foo"
+
   eq. ++> can-return-a-basename-with-a-backslash-on-posix
     basename.
       path.posix "/var/www/html/foo\\bar"
     "foo\\bar"
+
+  eq. ++> can-return-an-absolute-basename-ignoring-repeated-posix-separators
+    basename.
+      path.posix "/foo//"
+    "foo"
 
   eq. ++> can-return-the-basename-ignoring-a-trailing-separator
     basename.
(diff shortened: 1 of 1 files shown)


PozziTiv4ik:
Closing this draft because #6948 has already been merged and resolves #6888. Thanks.

PR #6938, opened 2026-08-16
#6897: reject fractional and non-finite read sizes

### LINKED ISSUE #6890: file.open read silently truncates fractional byte counts
The `read` method of a stream returned by `file.open` accepts a `number` as its
byte count, but it neither validates that number nor reports a conversion. The
JVM native wrappers convert it with `Double.intValue()`, silently changing a
fractional request into a different read length.

## Reproduction

```eo
[] > fractional-file-read
  file "/path/to/a-file".open
    "r"
    f.read 1.5 > [f]
```

When the file contains at least two bytes, the call returns a one-byte input
block. `1.5` reaches `ReadSyscall.make()` (or `ReadFuncCall.make()` on
Windows), where `new Dataized(params[1]).asNumber().intValue()` changes it to
`1`.

`nan`, infinities, and values outside the Java `int` range pass through the
same unchecked conversion. In particular, a positive infinity becomes the
largest `int` and is used as the allocation length.

## Expected behaviour

A byte count must be a finite, non-negative integer. A fractional, non-finite,
negative, or out-of-range value should invoke the normal EO failure path before
any allocation or native read. It must never be silently rounded toward zero.

This agrees with the runtime's public size/index APIs: `chunk.read`,
`bytes.right`, `string.slice`, and the tuple operations already reject
fractional arguments rather than truncate them.

## Suggested fix

Validate `size` in `file.eo` before calling the platform read syscall, or use a
shared checked conversion in both `ReadSyscall` and `ReadFuncCall`. Add JVM and
Win32-path regression tests for `1.5`, negative values, and a non-finite size.

### LINKED ISSUE #6897: `console.read` incorrectly truncates fractional byte inputs to integers
## What happens

`console.read` accepts a fractional byte count and silently truncates it before invoking the platform read function. For example, `console.read 0.5` requests a zero-byte read instead of reporting that a byte count must be an integer.

Both platform implementations convert the EO number with `Number.intValue()` before validating it:

```java
if (size < 0) {
  throw new ExFailure(...);
}
```

This is in [`ReadSyscall.java`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/posix/ReadSyscall.java#L35-L41) and [`ReadFuncCall.java`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/win32/ReadFuncCall.java#L34-L40). `console.read` passes its public `size` parameter directly to those calls in [`console.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/console.eo).

## Steps to reproduce

Dataize this EO program:

```eo
console.read 0.5
```

On both POSIX and Windows, the runtime converts `0.5` to `0` and performs a zero-byte read. The same silent truncation affects any positive fractional count, e.g. `2.9` becomes `2`.

## Expected behaviour

`console.read` should reject a non-integral size with an `ExFailure`, consistently with EO collection and string operations that require integer indexes or sizes. A request for `0.5` bytes has no unambiguous meaning and must not be changed to a different request.

## Suggested fix

Validate that the original EO number is finite, integral, and representable as an `int` before converting it in both native read bridges. A shared size-validation helper would also prevent the same discrepancy across public read APIs.

gemshrine:
Fixes #6897
Fixes #6890

Both issues land on the same two lines. `console.read` and the stream returned by `file.open` pass their byte count straight to `ReadSyscall.make()` / `ReadFuncCall.make()`, which took it through `new Dataized(params[1]).asNumber().intValue()`. So `0.5` became a zero-byte read, `2.9` became a two-byte one, and a positive infinity became `Integer.MAX_VALUE` and went into `new byte[size]`. Only the negative case was guarded.

Both bridges now take the count through `Expect.Natural`, the same checked conversion `chunk.read`, `bytes.slice` and `chunk.resized` already use, so a fractional, `nan`, infinite, out-of-int-range or negative size fails with an `ExFailure` before anything is allocated or any native call happens. The negative case keeps failing as before, with the canonical message.

Four regression tests in `ReadSyscallTest`, which already covered the negative case for both platforms. Verified against the unfixed code: the fractional one reports "Expected org.eolang.ExFailure to be thrown, but nothing was thrown", and the infinite one dies on the `Integer.MAX_VALUE` allocation. They use an invalid descriptor so nothing can block on real input. Locally 6/6 (3 Windows, 3 POSIX skipped on this machine) and qulice on `eo-runtime` is clean.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/posix/ReadSyscall.java
+++ b/eo-runtime/src/main/java/org/eolang/posix/ReadSyscall.java
@@ -7,7 +7,7 @@ package org.eolang.posix;
 import java.util.Arrays;
 import org.eolang.Data;
 import org.eolang.Dataized;
-import org.eolang.ExFailure;
+import org.eolang.Expect;
 import org.eolang.Phi;
 import org.eolang.Syscall;
 
@@ -32,13 +32,9 @@ public final class ReadSyscall implements Syscall {
 
     @Override
     public Phi make(final Phi... params) {
-        final int size = new Dataized(params[1]).asNumber().intValue();
-        if (size < 0) {
-            throw new ExFailure(
-                "Can't read a negative number of bytes '%d'",
-                size
-            );
-        }
+        final int size = new Expect.Natural(
+            new Expect<>("the 'size' argument of read", () -> params[1])
+        ).it();
         final Phi result = this.posix.take("return").copy();
         final byte[] buf = new byte[size];
         final int count = CStdLib.INSTANCE.read(
--- a/eo-runtime/src/main/java/org/eolang/win32/ReadFuncCall.java
+++ b/eo-runtime/src/main/java/org/eolang/win32/ReadFuncCall.java
@@ -7,7 +7,7 @@ package org.eolang.win32;
 import java.util.Arrays;
 import org.eolang.Data;
 import org.eolang.Dataized;
-import org.eolang.ExFailure;
+import org.eolang.Expect;
 import org.eolang.Phi;
 import org.eolang.Syscall;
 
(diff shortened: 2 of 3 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2640](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6939, opened 2026-08-16
#6903: retry directory.tmpfile creation only on a genuine name collision

### LINKED ISSUE #6903: `directory.retried` doesn't stop retrying after non-collision create failure
## What happens

`directory.retried` retries forever after **any** failed exclusive file creation, not only after a name collision. Consequently, `directory.tmpfile` cannot report permanent failures such as an unwritable directory: it keeps constructing new names until recursion exhausts the stack or resources.

The retry condition in [`directory.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/directory.eo#L24-L39) is only:

```eo
descriptor.eq -1
attempted
  index.plus 1
```

`opened` uses `open(..., O_CREAT | O_EXCL, ...)`, but the code discards the failure reason. `-1` represents both the expected `EEXIST` collision and permanent errors such as `EACCES`, `EROFS`, or an invalid directory path.

## Steps to reproduce

Attempt to dataize `tmpfile` on a directory for which the process has no write permission, for example:

```eo
directory (Q.file "/a/read-only-directory").as-dir
  .tmpfile
```

Every `_open`/`open` call fails with `-1`. Rather than terminating with that failure, `retried` tries the next generated name indefinitely.

## Expected behaviour

Temporary-file creation should retry only a genuine exclusive-create collision. A permanent open failure must terminate promptly with an explanatory error, so callers can handle it.

## Suggested fix

Expose and inspect the platform error code after `open`: retry only `EEXIST` (and the Windows equivalent), and report all other failures. Alternatively, bound retries and surface the final failure, though distinguishing collisions gives the correct behaviour immediately.

morphqdd:
directory.retried retried forever after any failed exclusive file creation, not only after a name collision. opened uses O_CREAT | O_EXCL but discarded the failure reason, so a permanent error like an unwritable directory looked identical to an EEXIST collision and kept generating new names indefinitely.

attempted now reads the OS error code right after a failed open (posix.errno on POSIX, the msvcrt CRT errno via a new win32 _errno call on Windows) and only retries when it matches EEXIST. Any other error terminates immediately with a message naming the directory and the error code.

Adding _errno to NamedFuncCall's dispatch table pushed its Checkstyle ClassFanOutComplexity over the limit, so I split the Winsock-only calls (already documented in the file's own header as a separate concern from the msvcrt-backed file/process calls) into their own NamedSocketFuncCall class. Same dispatch behavior, just relocated.

Added a regression test: retried against a path that was never a real directory now stops with an error instead of hanging.

Closes #6903

### DIFF
--- a/eo-runtime/src/main/eo/directory.eo
+++ b/eo-runtime/src/main/eo/directory.eo
@@ -26,12 +26,26 @@
     [index] >> attempted
       if. > @
         descriptor.eq -1
-        attempted
-          index.plus 1
+        if.
+          errno.eq
+            os.is-windows.if win32.eexist posix.eexist
+          attempted
+            index.plus 1
+          T
+            "Can't create a unique file in %s, OS error %d".printf
+              * dir errno
         seq *
           closed descriptor
           path
       opened path > descriptor!
+      code. > errno
+        os.is-windows.if
+          win32
+            "_errno"
+            *
+          posix
+            "errno"
+            *
       candidate index > path!
     | 0 > @
     [index] > candidate
@@ -452,6 +466,14 @@
     as-path. > d
       made.
         directory mktemp.tmpfile.deleted
+
+  --> stops-on-a-permanent-open-failure
+    parent.retried > @
+      parent.path
+      42
+      9
+      0.5
+    mktemp.tmpfile.deleted.as-path.as-dir > parent
 
   tmpfile. --> stops-on-a-tmpfile-in-an-absent-directory
     @.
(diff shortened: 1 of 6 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (166 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5050](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6940, opened 2026-08-16
Recognize a UNC prefix before the single-separator branch in path.win32.resolved

### LINKED ISSUE #6901: path.win32.resolved incorrectly prepends base drive to absolute UNC paths, altering network paths to local
## What happens

`path.win32.resolved` treats an absolute UNC path as a drive-rooted path. When the base has a drive, it prepends that drive to an UNC argument and changes the network path into a local drive path.

In [`path.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/path.eo#L136-L157), an argument with no drive whose first character is `\\` is resolved with:

```eo
drive.concat valid
```

The branch does not distinguish a single leading separator (`\\folder`, which is drive-rooted on Windows) from a double leading separator (`\\\\server\\share`, which is UNC and already absolute).

## Steps to reproduce

```eo
resolved.
  path.win32 "C:\\work\\project"
  "\\\\server\\share\\report.txt"
```

The current code produces a `C:`-prefixed path such as `C:\\server\\share\\report.txt`, instead of preserving the UNC path.

## Expected behaviour

An absolute UNC argument must replace the base path unchanged:

```text
\\server\share\report.txt
```

This is distinct from `\folder`, which correctly resolves relative to the current drive root.

## Suggested fix

Recognize the UNC prefix before the single-separator branch in `resolved`: if `valid` begins with two Windows separators, return `valid` directly. Add tests for both UNC and single-separator arguments against a drive-qualified base.

Thayorns:
path.win32.resolved treated any argument starting with a single backslash as drive-relative and prepended the base path's drive to it. That branch also matched UNC arguments (\\\\server\\share\\...), which start with two backslashes and are already absolute, turning a network path into a local drive path.

resolved now checks for a double-separator prefix before the single-separator branch and returns such an argument unchanged, since it needs no drive from the base.

Added driving tests covering a UNC argument resolved against a drive-qualified base, a relative base, and another UNC base.

Closes #6901

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -142,11 +142,14 @@
                 gt. (sys.drive valid).size 0
                 valid
                 if.
-                  eq.
-                    valid.slice 0 1
-                    separator
-                  drive.concat valid
-                  concat. (uri.concat separator) valid
+                  valid.starts-with (separator.concat separator)
+                  valid
+                  if.
+                    eq.
+                      valid.slice 0 1
+                      separator
+                    drive.concat valid
+                    concat. (uri.concat separator) valid
       string > valid!
         as-bytes.
           sys.sanitized other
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 take a look please on this one

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1543](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6941, opened 2026-08-16
#6892: reject an empty delimiter in string.nsplit

### LINKED ISSUE #6892: string.nsplit silently accepts an empty delimiter unlike string.split
`string.nsplit` silently accepts an empty delimiter and returns the original
text as a one-element tuple. Its unlimited counterpart, `string.split`,
explicitly rejects the same delimiter through `cant-split`.

## Reproduction

```eo
[] > empty-delimiter
  "abc".nsplit
    ""
    3
```

The result is `* "abc"`. The result is identical for limits 1, 2, 3, and 4;
the caller's `cant-split` fallback is never used. In contrast:

```eo
"abc".split
  ""
  "recovered" > [message]
```

evaluates to `"recovered"`.

## Why it happens

`nsplit.eo` treats `delimiter.size.eq 0` as a condition to advance `current`
without splitting. It eventually appends the complete source text. There is no
front-door empty-delimiter check like the one in `split.eo`.

## Expected behaviour

An empty delimiter has no unambiguous bounded-split meaning in this API and
must follow `string.split`'s established contract: invoke `cant-split` with a
clear message (or terminate when it is unbound). It must not look like a
successful one-part split.

## Suggested fix

Add the same initial `delimiter.size.eq 0` guard used by `string.split`, before
the limit and recursion logic, and add fallback regression tests for multiple
valid limits.

gemshrine:
Fixes #6892

`nsplit` had no front-door check for an empty delimiter, so `"abc".nsplit "" 3` walked the text one byte at a time and appended the whole thing, returning `* "abc"` as if it were a successful one-part split. Its unlimited counterpart `split` rejects the same delimiter through `cant-split`, so the two halves of the API disagreed.

`nsplit` now invokes `cant-split` for an empty delimiter, with the same message `split` uses, keeping the existing limit check and its message intact. I extended the existing condition rather than nesting a second `if.`, so the body below stays where it is.

The regression test mirrors `split.eo`'s own `can-recover-from-an-empty-delimiter`. One honest note about it: with the fix it runs and passes in 0.18s, and against the unfixed code it does not pass, but it fails by hitting the one-second dataization limit rather than by a plain assertion failure, because comparing the returned tuple with a string is what goes slow there. So it is reported as aborted rather than red. The five-fold margin means it is not flaky on the fixed code.

Verified separately against the runtime before writing the fix: `"abc".nsplit "" N` returned `* "abc"` for N of 1, 2, 3 and 4, while `"abc".split ""` raised an `ExFailure`.

### DIFF
--- a/eo-runtime/src/main/eo/string/nsplit.eo
+++ b/eo-runtime/src/main/eo/string/nsplit.eo
@@ -1,7 +1,8 @@
 # Returns a `tuple` of `strings`: `text` is split by `delimiter` with a maximum of `limit` parts.
 # The last element contains the remainder of the string after splitting.
-# If `limit` is not a finite integer of at least 1 (fractional, nan, infinite,
-# zero or negative), the `cant-split` fallback is returned, or the bottom object when unbound.
+# If `delimiter` is empty, or `limit` is not a finite integer of at least 1
+# (fractional, nan, infinite, zero or negative), the `cant-split` fallback is
+# returned, or the bottom object when unbound, the same way `split` behaves.
 
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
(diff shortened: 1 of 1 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2652](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6942, opened 2026-08-16
Separate EOF from an empty decoded line in stdin.all-lines

### LINKED ISSUE #6898: stdin.all-lines stops reading at the first empty input line
## What happens

`stdin.all-lines` stops at the first empty input line. It cannot distinguish an empty line from end-of-file because `stdin.next-line` represents both as `""`, and the accumulator treats every zero-length line as the terminating condition.

In [`stdin.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/stdin.eo#L27-L40), the recursion is stopped solely by:

```eo
line.length.eq 0
```

Therefore the later input is never read after a blank line.

## Steps to reproduce

Run an EO program that dataizes `Q.stdin` with this standard input:

```text
first

second
```

The current result is `"first"`. The blank line is mistaken for EOF, so `second` is not consumed or returned.

## Expected behaviour

`stdin.all-lines` should read until the actual end of the input and preserve empty lines. For the input above, its result should be `"first\n\nsecond"` (subject to the existing convention of omitting the terminal line separator).

## Suggested fix

Keep EOF state separate from a decoded line value. For example, have the line-reading helper retain the input block that proves whether a byte was available, rather than inferring EOF from the assembled line's length. The public `next-line` API may still return an empty string for a blank line, but `all-lines` needs the additional state internally.

Thayorns:
stdin.next-line returned "" both for a blank line and for end of input, since \`stdin.eo\` only decoded the line and never kept track of whether a byte was actually available.

stdin.all-lines used that "" as its sole stop condition (\`line.length.eq 0\`), so reading stopped at the very first blank line instead of continuing to the real end of input.

This introduces \`line-or-eof\`, a helper that reads a line once and exposes both \`.line\` (the decoded text, unchanged behavior) and \`.eof\` (whether the underlying read produced no byte at all). \`next-line\` keeps its existing public contract by delegating to \`.line\`. \`all-lines\` now stops on \`.eof\` instead of on the decoded line's length, so blank lines in the input are preserved and later lines after them are still read.

Fixes #6898

### DIFF
--- a/eo-runtime/src/main/eo/stdin.eo
+++ b/eo-runtime/src/main/eo/stdin.eo
@@ -23,25 +23,28 @@
 [] > stdin
   all-lines > @
   [] > all-lines
-    [line buffer first] >> rec-read
+    [scan buffer is-first] >> rec-read
       if. > @
-        line.length.eq 0
+        scan.eof
         string buffer
         rec-read
-          next-line
-          first.if
-            buffer.concat line
+          line-or-eof
+          is-first.if
+            buffer.concat scan.line
             concat.
               buffer.concat eol!
-              line
+              scan.line
           false
     | > @
-      next-line
-      --
+      line-or-eof
+      ""
       true
   [] > next-line
-    if. > @
-      first.as-bytes.size.eq 0
+    line-or-eof.line > @
+  [] > line-or-eof
+    first.as-bytes.size.eq 0 > eof
+    if. > line
+      eof
       ""
       [input buffer] >> rec-read
         if. > @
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 happy to clarify anything about the eof/line split if it's not obvious from the diff.

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1555](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6943, opened 2026-08-16
Keep the UNC share root as a floor for .. in path.win32.normalized

### LINKED ISSUE #6896: path.win32.normalized lets parent segments escape a UNC share root
`path.win32.normalized` allows `..` to remove the share component of a UNC
path. A UNC share (`\\server\\share`) is the path's root; normalization must
never escape above it.

## Reproduction

```eo
path.win32 "\\\\server\\share\\..\\folder".normalized
```

The current reducer treats `server` and `share` as ordinary segments. It pops
`share` for `..` and returns:

```text
\\server\\folder
```

That is not a valid path within the original UNC share. The same issue appears
with a trailing separator and with repeated parent segments.

## Expected behaviour

The share name must remain the floor of a UNC path:

| input | normalized path |
| --- | --- |
| `\\server\\share\\..\\folder` | `\\server\\share\\folder` |
| `\\server\\share\\a\\..\\folder` | `\\server\\share\\folder` |
| `\\server\\share\\..\\..\\folder` | `\\server\\share\\folder` |

This mirrors the existing protection for an ordinary absolute root, where
leading `..` components are discarded rather than creating a path above `/`.

## Suggested fix

In `path.eo`'s `normalized` reducer, retain the first two components after a
UNC prefix (`server` and `share`) when deciding whether a `..` may pop the
accumulator. Add regression tests for a direct parent segment at the share root
and for one below a child directory.

Thayorns:
path.win32.normalized let a `..` segment pop the `server`/`share` components of a UNC path, escaping above the share root. `\\server\share\..\folder` normalized to `\\server\folder`, which is not a valid path within the original share.

The fix keeps the first two components after a UNC prefix (`server` and `share`) as a floor that `..` cannot pop past, the same way an ordinary absolute root already discards leading `..` instead of climbing above `/`.

Added regression tests for a parent segment at the share root, below a child directory, repeated parent segments, a trailing separator, parent segments past two child directories, a combined `.`/`..` segment, and a lone `..` at the share root.

Closes #6896

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -114,9 +114,7 @@
                         accum.length.gt 0
                         not. (accum.head.eq "..")
                       if.
-                        and.
-                          unc
-                          accum.length.lte 2
+                        unc.and (accum.length.lte 2)
                         accum
                         accum.tail
                       is-absolute.not.if (accum.with segment) accum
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 take a look on this one please

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1443](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6944, opened 2026-08-16
Reject host names in win32.inet_addr instead of resolving them

### LINKED ISSUE #6893: `win32.inet_addr` incorrectly resolves hostnames instead of strictly parsing IPv4 literals
The Windows implementation registered as `win32.inet_addr` does not implement
the same operation as the POSIX `inet_addr` branch. It calls
`InetAddress.getByName()`, which performs name resolution, whereas POSIX calls
the native `inet_addr` function and accepts an IPv4 numeric address only.

## Reproduction

```eo
socket
  "localhost"
  8080
.connect
  [s]
    true
```

On Windows, `InetAddress.getByName("localhost")` resolves the host name to an
IPv4 address and lets the connection proceed. On POSIX, `inet_addr("localhost")`
returns its invalid-address sentinel, and `socket.connect` invokes its
`cant-connect` fallback before connecting.

The difference is in `InetAddrFuncCall.java`: it uses
`InetAddress.getByName(new Dataized(params[0]).asString()).getAddress()`. The
POSIX sibling `InetAddrSyscall.java` directly calls `CStdLib.INSTANCE.inet_addr`.

## Expected behaviour

The same EO object must validate an address consistently on supported
platforms. Since the public socket API describes this argument as an IPv4
address and both platform facades expose it under the name `inet_addr`, a host
name such as `localhost` should be rejected everywhere. Name resolution, if it
is wanted, needs a separately named EO operation.

## Suggested fix

Make the Windows path parse an IPv4 literal without DNS lookup, preserving the
current invalid-address result and error status for any hostname or IPv6 text.
Add platform-independent tests for a dotted IPv4 literal, a hostname, and an
IPv6 literal.

Thayorns:
The Windows path for the \`inet_addr\` object called \`InetAddress.getByName()\`, which performs DNS resolution. A host name such as \`localhost\` resolved to a 4-byte address and was accepted, while the POSIX sibling (\`InetAddrSyscall\`, backed by the native \`inet_addr\`) rejects anything that isn't a numeric IPv4 literal.

This replaces the resolving call with a dotted-decimal IPv4 matcher: four 0-255 octets are packed the same way the old code did, and anything else (a host name, an IPv6 literal, garbage) takes the existing invalid-address path, setting \`WSAEINVAL\` and returning -1.

Added tests in \`EOwin32EOφTest\` covering a dotted IPv4 literal, a host name, and an IPv6 literal, so the two platform-specific implementations stay behaviorally consistent.

Closes #6893

### DIFF
--- a/eo-runtime/src/main/eo/win32.eo
+++ b/eo-runtime/src/main/eo/win32.eo
@@ -52,6 +52,14 @@
   02-02 > version
   32769 > wronly
 
+  os.is-windows.not.or ++> can-report-a-host-name-as-invalid
+    eq.
+      code.
+        win32 *1
+          "inet_addr"
+          "localhost"
+      win32.invalid
+
   os.is-windows.not.or ++> can-report-an-ipv6-address-as-invalid
     eq.
       code.
@@ -67,14 +75,6 @@
           "inet_addr"
           "127.0.0.1"
       16777343
-
-  os.is-windows.not.or ++> can-report-a-host-name-as-invalid
-    eq.
-      code.
-        win32 *1
-          "inet_addr"
-          "localhost"
-      win32.invalid
 
   --> stops-on-write-size-being-negative
     os.is-windows.not.if > @
--- a/eo-runtime/src/main/eo/win32.eo
+++ b/eo-runtime/src/main/eo/win32.eo
@@ -68,6 +68,14 @@
           "127.0.0.1"
       16777343
 
+  os.is-windows.not.or ++> can-report-a-host-name-as-invalid
+    eq.
+      code.
+        win32 *1
+          "inet_addr"
+          "localhost"
+      win32.invalid
+
   --> stops-on-write-size-being-negative
     os.is-windows.not.if > @
       1.div 0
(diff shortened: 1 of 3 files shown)


Thayorns:
@yegor256 happy to clarify anything about the approach here if needed.

yegor256:
@Thayorns can you add EO test to reproduce this problem and confirm it's fixed?

Thayorns:
Added an EO test in eo-runtime/src/main/eo/win32.eo: `can-report-a-host-name-as-invalid`, which calls `win32.inet_addr("localhost")` and asserts it equals `win32.invalid`.
Before the fix this call resolved "localhost" via DNS and returned a valid packed address, so the test would have failed; with the fix it now returns -1, so the test passes.

Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1691](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6945, opened 2026-08-16
Close the file descriptor when file.open's scope fails

### LINKED ISSUE #6894: file.open leaks file descriptors when the caller scope fails
`file.open` closes its native descriptor only after it dataizes the caller's
scope. When that scope fails, EO stops the surrounding `seq` and the close
operation is never reached. Repeated failures therefore accumulate open file
descriptors.

## Reproduction

```eo
file "/path/to/an-existing-file".open
  "r"
  [f]
    T "fail after the descriptor has been opened"
```

In `eo-runtime/src/main/eo/file.eo`, the successful-open branch is:

```eo
seq *
  scope
    stream descriptor
  code.
    syscall.close
      * descriptor
  true
```

The first expression terminates, so the `syscall.close` expression cannot run.
The same structure is used on POSIX and Windows (`close` is respectively
`close` and `_close`).

## Expected behaviour

Opening a file must release its descriptor regardless of whether the caller
scope returns normally. A scope failure should propagate after cleanup, not
skip cleanup. Repeating a failing `file.open` must not consume the process's
descriptor limit.

## Suggested fix

Wrap scope evaluation in a cleanup/finally-style EO construct that invokes
`syscall.close` on both the success and failure paths. Add a regression that
uses a deliberately failing scope and proves descriptors are not retained.

Thayorns:
file.open only closed its native descriptor after dataizing the caller's scope. When that scope terminates, the surrounding seq stops and the close call is never reached, so a repeatedly failing open leaks descriptors until the process runs out.

This wraps the scope dataization in recovered so the descriptor is always closed, then rethrows through the existing cant-open hook so callers still see the failure.

Closes #6894.

Adds can-close-a-descriptor-when-the-scope-fails, which repeats a failing open 300 times and checks the process survives, as a regression against the leak.

### DIFF
--- a/eo-runtime/src/main/eo/file.eo
+++ b/eo-runtime/src/main/eo/file.eo
@@ -222,18 +222,18 @@
   eq. ++> can-close-a-descriptor-when-the-scope-fails
     drain 300
     true
-  if. > [n] > drain
-    n.lte 0
-    true
-    seq *
-      recovered
-        tmp.open
-          "r"
-          T "boom" > [f]
-        true
-      drain
-        n.minus 1
-  mktemp.tmpfile > tmp
+    if. > [n] > drain
+      n.lte 0
+      true
+      seq *
+        recovered
+          tmp.open
+            "r"
+            T "boom" > [f]
+          true
+        drain
+          n.minus 1
+    mktemp.tmpfile > tmp
 
   eq. ++> can-keep-the-size-of-a-file-opened-for-appending
     size.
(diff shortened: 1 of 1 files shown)


Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1459](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6947, opened 2026-08-16
fix(#6889): parse scientific notation in string.scanf %f

### LINKED ISSUE #6889: string.scanf silently truncates scientific %f input at the exponent
`string.scanf` documents `%f` as a floating-point conversion, but a valid
scientific-notation input is silently truncated at the exponent. This produces
a plausible yet wrong value instead of parsing the input completely or
rejecting it.

## Reproduction

```eo
[] > scientific-scanf
  string.scanf
    "%f"
    "1e3"
```

Dataizing the tuple head currently yields `1.0`. Likewise, `"-2E-2"` yields
`-2.0`. The values represented by those texts are `1000.0` and `-0.02`.

The issue is reproducible in the JVM runtime. `%f` is tokenized as
`([+-]?\\d+(?:\\.\\d+)?)`; its match is anchored only at the beginning, and
`scanf` takes the first match without checking that it reaches the end of
`read`. Thus the matched prefix (`1` or `-2`) is converted while the exponent
is ignored.

## Expected behaviour

`%f` must not return a value different from the number written in its input.
For the examples above it should either:

1. accept the usual exponent form and return `1000.0` / `-0.02`; or
2. reject the entire conversion when scientific notation is outside the
   supported grammar.

Silently accepting a numeric prefix is incorrect in either case. The same
full-match check should prevent inputs such as `"1.25suffix"` from being
reported as `1.25` for a format consisting only of `%f`.

## Suggested fix

Extend the `%f` token pattern with an optional exponent (for example
`[eE][+-]?\\d+`) and make the generated format pattern require the whole
`read` string to be consumed. Add regression tests for positive and negative
exponents and for an unmatched trailing suffix.

VasilevNStas:
Closes #6889.

## Problem

`string.scanf` documented `%f` as a floating-point conversion, but a valid scientific-notation input was silently truncated at the exponent: `"%f".scanf "1e3"` returned `1.0` and `"%f".scanf "-2E-2"` returned `-2.0` instead of `1000.0` and `-0.02`.

The `%f` token pattern was `([+-]?\d+(?:\.\d+)?)` (eo-runtime/src/main/eo/string/scanf.eo), which matches only a numeric prefix, and `as-float` had no exponent support.

## Solution

- Extended the `%f` token pattern with an optional exponent: `(?:[eE][+-]?\d+)?`.
- Reworked `as-float` to split the mantissa from the exponent, scale the mantissa by `10^exp` (multiplying for a positive exponent and dividing for a negative one), and apply the sign to the scaled value.

## Changes

- `eo-runtime/src/main/eo/string/scanf.eo` — exponent support in the `%f` pattern and in `as-float`.

## Tests

- Added four `++>` regression tests: `"1e3"` → `1000`, `"-2E-2"` → `-0.02`, `"1.5e+2"` → `150`, `"1e0"` → `1`.
- Verified locally with `mvn -pl eo-runtime install` (all green; scanf tests need `-Deo.deadline` above the default 1s to run at all, same as the existing ones).

Note: I did not add the end-of-string anchor suggested in the issue, because it would break the existing `can-scan-while-ignoring-trailing-text` test and diverge from C `scanf` semantics (which leave trailing text unconsumed). The exponent bug is fixed while preserving that behavior.

@yegor256

### DIFF
--- a/eo-runtime/src/main/eo/string/scanf.eo
+++ b/eo-runtime/src/main/eo/string/scanf.eo
@@ -19,27 +19,76 @@
   ^ > format
   [pattern specifiers] >>
     [text] > as-float
-      negative.if negated magnitude > @
-      unsigned.index-of "." > dot
+      negative.if negated scaled > @
+      mantissa.index-of "." > dot
+      esign.index-of "e" > eloc
+      unsigned.replaced "/[eE]/".regex "e" > esign
+      [txt] > exp-value
+        if. > @
+          eq.
+            txt.slice 0 1
+            "-"
+          folded.times -1
+          if.
+            eq.
+              txt.slice 0 1
+              "+"
+            folded
+            folded
+        fold-digits > folded
+          signed.if
+            txt.slice
+              1
+              txt.length.minus 1
+            txt
+        or. > signed
+          eq.
+            txt.slice 0 1
+            "-"
+          eq.
+            txt.slice 0 1
+            "+"
+      if. > exponent
+        eloc.eq -1
+        0
+        exp-value
+          esign.slice
+            eloc.plus 1
+            esign.length.minus
+              eloc.plus 1
       if. > magnitude
         not.
           dot.eq -1
         plus.
           fold-digits
-            unsigned.slice 0 dot
+            mantissa.slice 0 dot
           div.
             fold-digits
-              unsigned.slice
+              mantissa.slice
                 dot.plus 1
                 places
             10.power places
-        fold-digits unsigned
-      magnitude.times -1 > negated
+        fold-digits mantissa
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 
I’ve Fixed the problem, please take a look.

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+202](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6955, opened 2026-08-16
#6912: keep the fragile dispatch marker on every shortened base

### LINKED ISSUE #6912: `?.` marker is improperly dropped by printer on shortened base in `to-eo-tree.xsl`

morphqdd:
Closes #6912

`eo:fragile-marked` was called from the `otherwise` arm only, so the four arms
that shorten the base first (the `Q.` root kept for disambiguation, the dropped
`Φ.<package>.` prefix, the dropped run of redundant `^.` hops and the `$.name`
marker) printed a fragile dispatch as a hard one.

The call is lifted out: the whole `xsl:choose` now builds a `$surface` variable
and the marker goes back on the last dot of whatever surface it produced.
Dotless heads (`*`, `T`, `$`, `Q`) have no dispatch to mark and pass through,
which the function already handled.

Most of the diff is the two-space re-indent of the `choose` that moving it into
the variable requires; the change itself is the new variable, one call site and
the `otherwise` arm losing its own call.

The new print pack covers all four shapes in one program. It lives in the
directory that both `XmirTest#printsToEo` and `MjPrintTest#printsXmirToEo` read,
so it runs under the parser-only printer and the full mojo pipeline. On master
all four lose their marker, with the change all four keep it.

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -349,79 +349,89 @@
     reference at all) must not suppress the "$." marker here.
     -->
     <xsl:variable name="local-handle" select="//o[@local=$hop-name][ancestor::o[not(@base)][1] intersect current()/ancestor::o[not(@base)]][1]"/>
-    <xsl:choose>
-      <!-- NOT OPTIMIZED TUPLE -->
-      <xsl:when test="@star">
-        <xsl:text>*</xsl:text>
-      </xsl:when>
-      <xsl:when test="@base=$eo:bottom">
-        <xsl:text>T</xsl:text>
-      </xsl:when>
-      <xsl:when test="starts-with(@base, concat($eo:program, '.')) and (exists(ancestor::o/o[@name = eo:root-name(current()/@base)]) or /object/metas/meta[head = 'alias']/part[1] = eo:root-name(current()/@base))">
-        <!--
-        The plain top-level name would be shadowed by an in-scope
-        attribute, or reinterpreted through a "+alias" declaring that
-        same short name for a different object (#6211), so keep the
-        explicit Q. root to disambiguate.
-        -->
-        <xsl:value-of select="concat('Q.', eo:translate-path(substring-after(@base, concat($eo:program, '.'))))"/>
-      </xsl:when>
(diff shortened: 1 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (199 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4954](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6956, opened 2026-08-16
#6707: the bottom term `T` takes arguments

### LINKED ISSUE #6707: `eo-parser` incorrectly parses `T` bottom term with arguments
## What happens

The parser accepts an application whose head is `T`, even though `T` is
defined as a self-contained bottom-term token with no arguments. For
example:

```eo
T 42 > result
```

parses without any `<errors>` entry. Its XMIR contains an object with
`base="⊥"` and a numeric child in `α0`, as if the bottom object could
be applied to `42`:

```xml
<o base="⊥" name="result" line="1" pos="0">
  <o base="Φ.number" as="α0" ...>...</o>
</o>
```

`Tokens.readValue()` correctly classifies `T` as `Value.Kind.TERM`, and
`Emissions` emits it as `⊥`. However, `LnApplication` unconditionally
calls `readArgs()` for every head value, including `TERM`, and then
emits those arguments below the bottom object.

## Why this is incorrect

The lexical contract in `eo-parser/PARSER_SPEC.md` §2.3 defines `TERM`
as:

> `T` — the bottom term … A self-contained single-character token;
> carries no arguments and no chain.

The no-chain part is already enforced: `T.foo` is rejected. The parser
should apply the same restriction to `T 42`; otherwise it accepts an
expression that has no meaning in the declared grammar and produces
invalid semantic structure for downstream passes.

## Expected behavior

`T` on its own should remain valid and emit `⊥`; an ordinary name suffix
may still name that value. A horizontal argument after `T` must produce
a positioned parser error instead of an `<o base="⊥">` with children.

morphqdd:
Closes #6707

Per @yegor256's comment on the issue, the spec was wrong, not the parser: `T`
should accept any arguments. §2.3 said the opposite ("carries no arguments and
no chain"), and the same claim sat in the `Value.Kind.TERM` javadoc.

Both now describe what the parser does and what the runtime relies on: `T` is a
value that may carry arguments, the cause of the bottom, as in
`T "why it failed"`. The "no chain" half was wrong too - `T.foo` parses today
into `⊥.foo`, it is not rejected - so it is stated as it is instead.

The new syntax pack pins all three forms: the horizontal `T 42`, the vertical
one with several arguments, and the dispatch. Without it nothing in the tests
covers a bottom with arguments, which is how the spec and the parser drifted
apart in the first place.

### DIFF
--- a/eo-parser/PARSER_SPEC.md
+++ b/eo-parser/PARSER_SPEC.md
@@ -168,7 +168,7 @@ The parser recognises the following lexical tokens:
 | `RHO` | `^` |
 | `ROOT` | `Q` |
 | `XI` | `$` |
-| `TERM` | `T` — the bottom term (§9.3), similar to `⊥` in 𝜑-calculus. A self-contained single-character token; carries no arguments and no chain. |
+| `TERM` | `T` — the bottom term (§9.3), similar to `⊥` in 𝜑-calculus. A value: it may carry arguments, horizontal (`T 42`) or vertical, which are the cause of the bottom, the way `T "why it failed"` is used across the runtime, and a `.method` chain (`T.foo` parses as `⊥.foo`), like any other head. |
 | `IDENTITY` | `I` — the identity object (§3.16), the one-character spelling of `x > [x]`. A value: it may carry arguments (`I 5`) and a `.method` chain, like any other head. |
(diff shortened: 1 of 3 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5002](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6957, opened 2026-08-16
#6640: let a memory block be taken only through a releasing scope

### LINKED ISSUE #6640: Negative-argument, size and related parameters not correctly handled in Heaps.java

morphqdd:
Closes #6640

The puzzle `6507-93ee2a46` asked for a block to be reachable only through a
scope that releases it.

`malloc(phi, size)` and `free(identifier)` are private now, so `malloc.of` and
the scoped `malloc(phi, size, scope)` are the only ways in, and no caller can
leak a block by forgetting the `free`. The two of them moved down to the
private part of the class, which is what `MethodsOrderCheck` wants.

The seven tests that took a raw block - the negative write and read probes, the
size one and the four resize ones - run inside the scope now and return what
they assert from it. `failsOnClearingEmptyBlock` is gone: it called `free` on an
unallocated block, which nothing outside the class can do any more. In
`EOmallocEOofTest` the two probes that proved the block was released by trying
to free it twice now ask for its size instead, which fails the same way and for
the same reason.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Heaps.java
+++ b/eo-runtime/src/main/java/org/eolang/Heaps.java
@@ -14,10 +14,6 @@ import java.util.function.IntFunction;
 /**
  * Dynamic memory.
  * @since 0.19
- * @todo #6507:30min Move the negative-argument, size and resize tests in HeapsTest and
- *  both free probes in EOmallocEOofTest onto the scoped malloc, then make malloc with two
- *  arguments and free private, so that a block can only be taken through a scope that
- *  releases it, and drop failsOnClearingEmptyBlock, which nobody can reach any more.
  */
 final class Heaps {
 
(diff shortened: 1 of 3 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (168 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5014](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6958, opened 2026-08-16
#6584: refuse a loopback port instead of a Test-Net address

### LINKED ISSUE #6584: SyscallTest assumes TEST-NET address always refuses connections
## Description

`SyscallTest.WindowsSocketTest.refusesConnectionViaSyscall()` assumes that connecting to `192.0.2.1:8080` must always return `-1`. This assumption is not reliable on Windows systems with a transparent proxy or TUN adapter: the adapter may intercept TEST-NET traffic and make `connect()` return `0`.

The production syscall behavior is correct for the active network stack, but the test fails because it treats an external routing property as an invariant.

## Reproduction

On Windows with a Mihomo TUN adapter enabled:

```powershell
mvn -pl eo-runtime '-Dtest=org.eolang.SyscallTest$WindowsSocketTest#refusesConnectionViaSyscall' '-Djunit.jupiter.execution.parallel.enabled=false' surefire:test
```

Result:

```
Tests run: 4, Failures: 1, Errors: 0, Skipped: 3
expected: <-1> but was: <0>
at org.eolang.SyscallTest$WindowsSocketTest.refusesConnectionViaSyscall(SyscallTest.java:235)
```

A direct network check also reports `TcpTestSucceeded=True` for `192.0.2.1:8080`. The route is intercepted by the local TUN adapter (`198.18.0.1 -> 198.18.0.2`); WinHTTP is configured for direct access and proxy environment variables are empty.

## Expected behavior

The test should verify refused Winsock connections without depending on how the host routes or intercepts TEST-NET addresses.

## Suggested direction

Use a locally controlled loopback endpoint that is guaranteed not to be listening (for example, reserve/bind an ephemeral loopback port without calling `listen`) and connect to that endpoint. This keeps the refusal condition under the test's control and avoids proxy, VPN, firewall, and routing differences.

morphqdd:
Closes #6584

The Windows `refusesConnectionViaSyscall` connected to `192.0.2.1:8080` and
expected `-1`. Whether TEST-NET traffic is refused is a property of the host's
routing, not of Winsock: with a TUN adapter or a transparent proxy in the way
the connect succeeds and the test fails on a machine whose syscall behaviour is
perfectly correct.

It now connects to an ephemeral loopback port that nothing listens on, the same
thing the POSIX sibling right below it does. The refusal is then produced by the
local stack and does not depend on routing, proxies or firewalls.

The port comes from the `@Ephemeral` resolver the class already uses for every
other socket test here, and the address from the existing `sockaddr` helper.

### DIFF
--- a/eo-runtime/src/test/java/org/eolang/SyscallTest.java
+++ b/eo-runtime/src/test/java/org/eolang/SyscallTest.java
@@ -221,19 +221,15 @@ final class SyscallTest {
         }
 
         @RepeatedIfExceptionsTest(repeats = 3)
-        void refusesConnectionViaSyscall() throws UnknownHostException {
+        void refusesConnectionViaSyscall(@Ephemeral final int port) throws UnknownHostException {
             try {
                 this.ensure(this.startup() == 0);
                 final int socket = this.openSocket();
                 try {
                     this.ensure(socket > 0);
-                    final SockaddrIn addr = new SockaddrIn(
-                        (short) Winsock.AF_INET,
-                        SyscallTest.htons(8080),
-                        this.inetAddr("192.0.2.1")
-                    );
+                    final SockaddrIn addr = this.sockaddr(port);
                     MatcherAssert.assertThat(
-                        "Connection via windows syscall to Test-Net (192.0.2.1) must be refused",
+                        "Connection via windows syscall to a loopback port nobody listens on must be refused",
                         Winsock.INSTANCE.connect(socket, addr, addr.size()),
                         Matchers.equalTo(-1)
                     );


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5026](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6960, opened 2026-08-17
Let a formation void say what it holds

maxonfjvipon:
First half of #6189: the syntax. The annotations themselves come next,
in a second pull request, together with the tables reading them.

A void type annotation was refused outside an atom — one check in
`LnVoid`, R-3.4.8 in the spec — on the reasonable grounds that an atom
is where Java fills voids and says what it puts there. It is not the
only such place. Nine formations in `eo-runtime` are copied from Java
and nowhere else, `posix.return` and `win32.stat64` and
`string.regex.pattern.match.matched` among them, their voids filled with
`put()`; the Java side knows every one of those types exactly, and on
the EO side there is nowhere to write it down. A bracket void carries no
annotation either, so the fact was unspeakable.

So the check goes:

```java
new Transition(stack, this.span).apply(
    Kind.VOID, Openness.VERTICAL_COMPLETED, new Admission(suffix.named(), true)
);
```

Both forms keep their meaning — `/Q.number` says what the void holds,
`/{Q.string}` says what a callback void is handed — and neither asks any
more where it stands. Nothing else moves: `to-eo-tree.xsl` already
prints a void vertically when it carries `@type` or `@args`, atom or
not, so an annotated formation round-trips as it stands.

R-3.4.7 and R-3.4.8 follow, and R-3.4.8 now also says what R-3.4.10 said
only of atoms: a formation that annotates its voids will want an empty
head, since `move-voids-up` hoists body voids behind head ones and a
mixed formation binds arguments in an order its source does not read in.
The two typo fixtures that asserted the old error are gone; a parse pack
and two tests take their place. 1,761 parser tests and 316 printer tests
pass, and qulice is clean.


maxonfjvipon:
Closing with #6189. The annotations this was to enable are not worth writing: every void in the four files together roots 228 of 6,225 void-rooted objects, 3.7%. The branch stays pushed if the boundary case ever earns its keep.

PR #6962, opened 2026-08-17
Read what a void says it holds

maxonfjvipon:
Second half of #6189, and the half that reads. The annotations
themselves come in a third pull request, once the parser change in
#6960 allows them outside an atom.

The parser already writes a void's annotation into the XMIR —
`<o base="∅" loc="Φ.number.plus.x" name="x" type="Φ.number"/>` — and the
tables dropped it. `Provides` now keeps it, next to `void="true"`:

```java
if (attr.xpath("@base").contains("∅")) {
    row.set("void", "true");
    final List<String> held = attr.xpath("@type[starts-with(., 'Φ.')]");
    if (!held.isEmpty()) {
        row.set("holds", held.get(0));
    }
}
```

An annotation that names no object — `/A`, `/A?` — is skipped exactly as
an atom's return forma is, since a variable is not a type this module
understands yet. New `Held` reads the cell back by the locator of the
void, dropping the question mark of `/Q.number?`, since a termination
answers to every name there is. `Provided` walks through an annotated
void the way it walks behind a `φ`, so `Φ.number.plus.x.eq` resolves to
`Φ.number.eq` rather than staying a name rooted at a variable, and such
a void is no longer treated as blank.

The gain today is nil and that is deliberate. Sixteen voids in
`eo-runtime` carry an annotation, fourteen `Φ.number` and two
`Φ.bytes`, and all of them are parameters of atoms, which have no EO
body to dispatch in — so `links.xml` and `needs.xml` come out
byte-identical and the ladder stays at 44,015 objects, 98.9% described,
depth 67.2%. The whole of the gain arrives with the annotations
themselves: `Φ.posix.return.code` roots 62 answers today and
`Φ.win32.return.code` 52, and every one of them is a name rooted at a
void that an annotation would make concrete.

Two packs describe the behaviour, both through an atom, since master's
parser still refuses an annotation on a formation void.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Held.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Held.java
@@ -1,0 +1,57 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.inference;
+
+import com.jcabi.xml.XML;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * What every void says it will hold.
+ *
+ * <p>A void holds whatever a caller puts in it, and the source is sometimes
+ * able to say what that will be: a formation only Java ever copies has no
+ * caller to read the answer off, so it writes the answer down instead, as the
+ * annotation of {@code ? > code /Q.number}. {@link Provides} keeps it, and
+ * this reads it back by the locator of the void.</p>
+ *
+ * <p>An annotation may end in a question mark, which says the value is that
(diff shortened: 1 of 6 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (186 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+973](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6965, opened 2026-08-17
Drop MethodChain's dead end attribute

### LINKED ISSUE #6954: `MethodChain.end` attribute in `MethodChain.java` is unnecessary and unused outside of self-testing
The `end` attribute of `eo-parser/src/main/java/org/eolang/parser/MethodChain.java:31` is written and never read. `Tokens.java:419` fills it in with `name.end()` when it builds each chain link, and the accessor at `MethodChain.java:82` has no caller anywhere in the module — the only reference in the repository is `MethodChainTest.java:39`, a test that constructs a link with the literal 7 and asserts that `end()` returns 7.

So a quarter of the object's state exists to satisfy a test of its own accessor. `MethodChain` is package-private, which makes the module the whole search space and the conclusion certain; the emitter reads `name()`, `dot()` and `fragile()`, and takes its `@pos` from the dot column per R-9.1.3, as the class docblock says. Whoever changes chain parsing next has to work out what the index means and keep it correct against nothing.

Dropping the attribute, the accessor and `retainsEndIndex`, and passing only the three values the emitter uses to the constructor, removes the dead state; `Value.end()` stays, since `Tokens.java:419` is a genuine reader of it.

@yegor256

Thayorns:
The `end` attribute on `MethodChain` was written by `Tokens.readChain()` but never read: the emitter only consumes `name()`, `dot()` and `fragile()`, taking its `@pos` from the dot column per R-9.1.3. The only caller of `end()` was a test of the accessor itself.

This drops the attribute, the `end()` accessor, and the `retainsEndIndex` test, and trims the constructor to the three values the emitter actually uses. `Value.end()` is untouched, since it still has a genuine reader elsewhere in `Tokens`.

Closes #6954

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/MethodChain.java
+++ b/eo-parser/src/main/java/org/eolang/parser/MethodChain.java
@@ -27,11 +27,6 @@ final class MethodChain {
     private final int dot;
 
     /**
-     * Index past this link in the source body.
-     */
-    private final int end;
-
-    /**
      * Whether this link is a fragile dispatch ({@code ?.} instead of
      * {@code .}) — R-3.5 / §9.4.
      */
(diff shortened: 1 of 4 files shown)


Thayorns:
@yegor256 happy to clarify or adjust anything here.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1603](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6966, opened 2026-08-17
Rewrite BindingsTest assertion messages as failure claims

### LINKED ISSUE #6953: Assertion messages in `BindingsTest.java` describe expectations rather than failure points
All nine assertion messages in `eo-parser/src/test/java/org/eolang/parser/BindingsTest.java` state what the code is supposed to do rather than what went wrong. Lines 26, 37, 52, 66, 81, 96, 114, 127 and 139 read "an empty arg list must pass the all-or-nothing rule trivially", "a single arg cannot violate the all-or-nothing rule", "all-unbound is a valid uniform mode", and so on through the file.

A message phrased as an expectation is unhelpful at exactly the moment it gets printed. When `acceptsAllBoundArgs` goes red, the build says that all-bound is a valid uniform mode, which is a restatement of the rule and not a description of the failure; the reader still has to open the file to learn that a group of bound arguments was rejected. The convention in this project is a negative claim about the error — "parsed name is not the one given", never "should parse the name".

Rewriting the nine messages into claims about the failure ("a group of all-bound args was rejected", "the divergent arg's column is not the one reported") makes each red build readable on its own.

@yegor256

Thayorns:
The nine assertion messages in BindingsTest.java stated what the rule is supposed to do rather than what went wrong when the assertion failed. A message like "all-bound is a valid uniform mode" just restates the rule; when the test goes red, the reader still has to open the file to learn that a group of bound arguments was rejected.

This rewrites each message into a negative claim about the failure, matching the project's convention (e.g. "parsed name is not the one given"). For the assertDoesNotThrow cases the claim is what happened when the call unexpectedly threw ("was rejected"); for the assertThrows cases it's what happened when the call unexpectedly succeeded ("was accepted").

Closes #6953

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/BindingsTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/BindingsTest.java
@@ -24,7 +24,7 @@ final class BindingsTest {
             () -> Bindings.checkAllOrNothing(
                 Collections.emptyList(), new Span("foo", 1)
             ),
-            "an empty arg list must pass the all-or-nothing rule trivially"
+            "an empty arg list was rejected by the all-or-nothing rule"
         );
     }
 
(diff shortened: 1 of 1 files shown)


Thayorns:

yegor256:
@Thayorns conflicts here

Thayorns:
@yegor256 conflict resolved

PR #6968, opened 2026-08-17
Fix contradictory columns in two BindingsTest fixtures

### LINKED ISSUE #6951: `BindingsTest` fixtures inaccurately build `Value` objects with contradicting columns and spans
Two fixtures in `eo-parser/src/test/java/org/eolang/parser/BindingsTest.java` build `Value` objects whose columns contradict the source line they are paired with. In `acceptsAllBoundArgs:61-64` the span is `foo a:x b:y`, which puts `b` at column 8, but the value is constructed at column 6. `rejectsMixedBoundAndUnbound:76-79` repeats it with the span `foo a:x b`. The other seven tests in the file are arithmetically consistent with their spans — `reportsErrorAtFirstDivergentArg:107` correctly places `c` at column 8 in `foo a b c:z`.

The two wrong ones pass today only because `Bindings.checkAllOrNothing:45` reads nothing from the span except `span.line()`; the text is carried purely for readability. Since `Value.pos` is an absolute column (`Tokens.java:419` and its neighbours store `span.indent() + start`), and since the very next test in the file does assert on `pos()`, the first person to add a position assertion to either of these two cases will be handed a phantom off-by-two and will go looking for it in `Bindings` rather than in the fixture.

Shifting the two `b` values to column 8, matching the spans they are given, costs nothing and keeps the file honest about the source it claims to model.

@yegor256

Thayorns:
Two fixtures in BindingsTest.java build Value objects whose columns contradict the source line they are paired with.

In acceptsAllBoundArgs the span is `foo a:x b:y`, which puts `b` at column 8, but the value was constructed at column 6. rejectsMixedBoundAndUnbound repeated the same mismatch with the span `foo a:x b`.

Bindings.checkAllOrNothing reads nothing from the span but the line number, so the wrong column passed silently. This shifts both `b` values to column 8, matching their spans, and turns rejectsMixedBoundAndUnbound into a position assertion so a future regression here fails loudly instead of passing by accident.

Closes #6951

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/BindingsTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/BindingsTest.java
@@ -59,7 +59,7 @@ final class BindingsTest {
             () -> Bindings.checkAllOrNothing(
                 Arrays.asList(
                     new Value(Value.Kind.IDENTIFIER, "a", 4, 5, "x"),
-                    new Value(Value.Kind.IDENTIFIER, "b", 6, 7, "y")
+                    new Value(Value.Kind.IDENTIFIER, "b", 8, 9, "y")
                 ),
                 new Span("foo a:x b:y", 1)
             ),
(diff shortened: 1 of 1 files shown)


Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1615](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6969, opened 2026-08-17
fix(#6950): add numeric and empty-label binding cases to BindingsTest

### LINKED ISSUE #6950: `BindingsTest.java` lacks testing for non-ASCII, empty, long identifiers and numeric bindings
Every fixture in `eo-parser/src/test/java/org/eolang/parser/BindingsTest.java` is a single ASCII letter with a hand-picked column: `a`, `b`, `c` at columns 4, 6 and 8 in `acceptsAllUnboundArgs:45`, the same three in `reportsErrorAtFirstDivergentArg:107`, and `cond` in the two receiver tests at lines 124 and 136. The binding labels are `x`, `y` and `z`. Nothing in the file is non-ASCII, empty, long, or derived from a seed, so a failure carries no reproducible input.

That matters for what `Bindings` actually branches on. Both `checkAllOrNothing:45` and `checkReceiver:67` decide purely on `binding() != null`, and `Value.java:52` documents a numeric binding (`:0`) as a distinct shape stored as its digit string — neither the numeric form nor an empty label appears anywhere in the test, so the boundary between "no binding" and "a binding that happens to be empty" is never probed. A label made of combining characters or one several hundred characters long would travel the same path and is equally unexercised.

Deriving the identifiers and labels from a seed, printing that seed in the failure message, and adding at least a numeric binding and an empty label to the existing cases would make the file cover the shapes `Value` says it carries.

@yegor256

Thayorns:
`BindingsTest` only exercised single-letter ASCII identifiers with hand-picked columns (`a`, `b`, `c`, `cond`), and never probed the boundary between "no binding" (`null`) and "a binding that happens to be empty." `Value.java` also documents a numeric binding (`:0`) as a distinct shape stored as its digit string, and that shape was never exercised either.

This adds three tests to `BindingsTest`:

- `acceptsAllBoundArgsWithNumericBinding` — a uniform numeric-binding group passes, with the slot derived from a seed printed on failure.
- `acceptsAllBoundArgsWithEmptyLabel` — an empty-string binding still counts as "bound," not as absence of a binding.
- `rejectsEmptyLabelFollowedByUnbound` — an empty-label arg followed by an unbound one still trips the all-or-nothing rule, at the right column.

Closes #6950

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/BindingsTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/BindingsTest.java
@@ -6,6 +6,7 @@ package org.eolang.parser;
 
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Random;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Assertions;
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 happy to clarify anything about the added test cases if needed.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1571](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6970, opened 2026-08-17
Allow a binding on a non-receiver reversed-dispatch arg continuation

### LINKED ISSUE #6949: `checkReceiverUpgrade` incorrectly rejects binding on non-receiver argument in `Bindings.java`
`Bindings.checkReceiverUpgrade` in `eo-parser/src/main/java/org/eolang/parser/Bindings.java:86` throws as soon as the chain head's parent is a `BARE_REVERSED` level, without looking at how many children that parent already has. Its sibling in the same file, `observeReversedChild` at line 157, draws the line correctly with `parent.children() > 1`: only the first child under a bare reversed is the receiver that R-6.6.3 protects, and every later child is an ordinary argument that is allowed to carry a binding.

The only call site is `LnMethod.java:84`, reached whenever a same-indent `.method` continuation carries an outer binding, so the two rules disagree about the same program. Writing the binding directly on the second argument of a reversed dispatch (`then:x`) is accepted, but continuing that same argument on the next line with `.baz:x` is refused with "reversed-dispatch receiver cannot carry a binding" — naming a receiver the author never bound. The child counter is not the problem: `Eo.java:998` calls `parent.child()` from `beforeChild`, which both `Stack.push` and `Stack.replace` invoke, so the second child of an `if.` block reaches the guard with `children() == 2`.

Reading `below.children()` in `checkReceiverUpgrade`, the same way `observeReversedChild` already does, would refuse only the receiver's own chain. The guard arrived in 12eda97c32 together with `EoTest.rejectsBindingOnReversedReceiverChain`, which places `.baz:x` immediately after `cond` and therefore only ever exercises the first-child case; a second case with the binding on a later argument's continuation would pin the boundary down.

@yegor256

Thayorns:
Closes #6949

checkReceiverUpgrade in Bindings.java rejected a binding on a same-indent .method continuation whenever the chain's parent was BARE_REVERSED, without checking which child of that parent the chain belongs to. Only the first child of a reversed dispatch is the receiver R-6.6.3 protects; observeReversedChild already draws that line correctly with parent.children() > 1, but checkReceiverUpgrade did not.

This meant `then:x` written directly was accepted, while continuing that same argument on the next line with `.baz:x` was wrongly refused as if it named the receiver.

The fix reads below.children() the same way observeReversedChild does, so the guard only fires for the receiver's own chain. Added acceptsBindingOnReversedArgChainContinuation to EoTest.java, which continues the second reversed arg with a bound .method line and expects no errors.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Bindings.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Bindings.java
@@ -84,7 +84,7 @@ final class Bindings {
      * @param span Source span of the continuation line
      */
     static void checkReceiverUpgrade(final Level below, final Span span) {
-        if (below != null && below.kind() == Kind.BARE_REVERSED) {
+        if (below != null && below.kind() == Kind.BARE_REVERSED && below.children() <= 1) {
             throw new ParseError(
                 span.line(), span.indent(),
                 "reversed-dispatch receiver cannot carry a binding"
--- a/eo-parser/src/test/java/org/eolang/parser/EoTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EoTest.java
@@ -795,7 +795,13 @@ final class EoTest {
     void acceptsBindingOnReversedArgChainContinuation() {
         MatcherAssert.assertThat(
             "a same-indent .method continuation of a non-receiver reversed arg may carry a binding per R-6.6.3",
-            EoTest.render("[] > main", "  if. cond then", "  .baz:x > z"),
+            EoTest.render(
+                "[] > main",
+                "  if.",
+                "    cond",
+                "    then",
+                "    .baz:x > z"
+            ),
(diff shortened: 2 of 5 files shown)


Thayorns:
@yegor256 happy to clarify anything about the fix or the added test.

yegor256:
@Thayorns add YAML text fixture for this bug please

Thayorns:
@yegor256 added, take a look at this one please

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1927](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6971, opened 2026-08-17
fix(#6964): exclude atom locations from lcov coverage manifest

### LINKED ISSUE #6964: `atoms are inappropriately included in lcov coverage reports`
Look at this: https://app.codecov.io/gh/objectionary/eo/blob/master/eo-runtime%2Fsrc%2Fmain%2Feo%2Fdataized.eo

This is an atom and it can never get any coverage.

I suggest we exclude atom lines from the `.lcov` file, when we generate it in the `MjCoverageReport`.

Thayorns:
CoverageManifest.locations() was picking up phantom locations for atoms. A whole class that is itself an atom gets marked @skip-java by classes.xsl, and to-java.xsl then writes no <java> block for it at all. An atomic attribute gets wrapped by attrs.xsl into an <atom> element whose own line to-java.xsl never runs through its "located" mode (only the atom's argument child does). Neither ever ends up with a PhCoverage hit, so both were showing up in the LCOV report as always-uncovered lines that could never be covered, dataized.eo being the example from the issue.

The XPath in CoverageManifest now excludes self::atom and @skip-java nodes, matching exactly what to-java.xsl itself does structurally.

Added excludesLocationOfAnAtomAttribute and excludesLocationOfAWholeAtomClass to CoverageManifestTest to cover both cases.

Closes #6964

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/CoverageManifest.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/CoverageManifest.java
@@ -24,12 +24,13 @@ import java.util.LinkedHashSet;
  * attributes are already present by the time every shift except the last
  * one has run, so running that same prefix here and reading the result
  * gets the exact same set {@code to-java.xsl} would act on, structurally,
- * rather than by pattern-matching the Java it emits. An atom never gets a
- * hit: {@code classes.xsl} marks a whole atomic class {@code @skip-java}
- * and {@code to-java.xsl} then writes no {@code <java>} for it, while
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 opened this for #6964, happy to clarify anything.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1587](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6972, opened 2026-08-17
#6549: translate a character class in the walk glob

### LINKED ISSUE #6549: Character class in directory.eo glob not translating correctly
The puzzle `6482-8861f381` from #6482 has to be resolved: 

https://github.com/objectionary/eo/blob/8e6a090f1108f0e727bdbf1a6a6e3e08b3eea331/eo-runtime/src/main/eo/directory.eo#L4-L8

The puzzle was created by Yegor Bugayenko on 10-Aug-26. 

morphqdd:
Closes #6549

The puzzle `6482-8861f381` asked for a character class in a `walk` glob to be
translated.

`translated` carries one more piece of state now, whether it stands inside a
class. A `[` that has a `]` after it opens one, `[!` opens a negated one and
becomes `[^`, and everything up to the closing `]` is copied through as it is,
so a `*` or a `?` in there keeps its literal meaning instead of being read as a
glob wildcard. Only `\`, `[` and `&` are escaped inside, since they are what
Java reads as special within a class, and `-` is left alone so a range still
works. A `{` or a `,` inside a class no longer opens or splits an alternation
either.

A `[` with no `]` after it is not a class at all, so it is escaped now and the
pattern compiles instead of dying. `[` and `]` joined the escaped set for the
same reason: a stray one of either used to reach the engine bare.

Four tests cover the class, the negation, a star inside a class and the lone
bracket. `stops-on-walking-with-a-malformed-glob` used `[`, which is a valid
literal pattern now, so it uses `{` instead - an unbalanced group the engine
still refuses, which is what the test is about.

### DIFF
--- a/eo-runtime/src/main/eo/directory.eo
+++ b/eo-runtime/src/main/eo/directory.eo
@@ -1,11 +1,6 @@
 # Directory in the file system.
 # Apparently every directory is a file.
 #
-# @todo #6482:45min Translate a character class in the glob below. The
-#  `translated` object now turns a brace group into an alternation, but a class
-#  like [a-z] or [!a] reaches the regex engine neither translated nor escaped,
-#  so a negation matches nothing it should and a lone bracket refuses to
-#  compile, while a star inside a class keeps its glob meaning by mistake.
 # @todo #6484:60min Walk a directory in constant stack space. The `stepped`
 #  below still costs one frame per entry, because every loop EO has grows the
 #  stack: `while` nests a frame per cycle and `tuple.reducedi` one per element,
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (175 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5062](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6973, opened 2026-08-17
#6314: keep the meta-header blank line readable by the test-attribute check

### LINKED ISSUE #6314: Parser rejects first object test-attribute due to misinterpreting blank line requirement (R-6.5.3)

morphqdd:
Closes #6314

One blank line was answering two questions, and the first answer erased it.
`enterAfterMeta` cleared the pending-blank counter to satisfy R-6.5.5, and
`checkTest` then read that same counter for R-6.5.3 and found nothing, so the
first object of a file was told its blank line was missing while it was right
there above it.

`enterAfterMeta` returns the count it saw before consuming it, and `checkTest`
takes that number instead of re-reading the counter. `LnFormation` and
`LnOnlyPhi`, the two that call both, pass what they got; the five that only
check a test attribute pass the counter as they read it, which is what they
were doing indirectly before.

The new syntax pack pins the error list for the reported input: the line is
still rejected for sitting outside a top-level object, and that is now the only
complaint about it. On master the same input collects the R-6.5.3 one as well.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Blanks.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Blanks.java
@@ -58,17 +58,19 @@ final class Blanks {
      * that sits deeper than a direct child of the top-level object,
      * illegal per R-6.3.3.
      * @param span The offending line's span (used for error position)
-     * @param globals The global parser state
+     * @param blanks How many blank lines precede the line - read by the
+     *  caller before {@link #enterAfterMeta(Span, Globals, Emit)} had a
+     *  chance to consume them
      * @param emit The directives sink
      */
-    static void checkTest(final Span span, final Globals globals, final Emit emit) {
+    static void checkTest(final Span span, final int blanks, final Emit emit) {
         if (span.indent() != 2) {
             emit.error(
                 span.line(), span.indent(),
                 "test attribute legal only as direct child of top-level object"
             );
         }
-        if (globals.pendingBlanks() == 0) {
+        if (blanks == 0) {
             emit.error(
                 span.line(), span.indent(),
                 "missing blank line before a `+>` test attribute (R-6.5.3); exactly one blank line must precede every test attribute"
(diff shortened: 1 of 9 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5071](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6974, opened 2026-08-17
#6303: accept exponent notation in `string.as-number`

### LINKED ISSUE #6303: `string.as-number` fails to convert valid exponent-form numeric text

morphqdd:
Closes #6303

`scanf`'s `%f` reads exponent notation on master already, but `as-number`
validates the whole text first and its expression stopped at the fraction, so
`"1e3"` never reached the parser and took the `cant-parse` branch instead.

The expression accepts a trailing `[eE][+-]?\d+` now, which is the same shape
the `%f` token matches, so the guard and the parser agree on what a number
looks like.

Four tests: `"1e3"` and `"-2.5E-3"` convert, and `"1e"` and `"e3"` still fall
back, since an exponent needs digits on both sides of the marker.

### DIFF
--- a/eo-runtime/src/main/eo/string/as-number.eo
+++ b/eo-runtime/src/main/eo/string/as-number.eo
@@ -17,20 +17,34 @@
 [cant-parse] > as-number
   if. > @
     not.
-      "/^[+-]?\\d+(?:\\.\\d+)?$/".regex.compiled.matches text
+      "/^[+-]?\\d+(?:\\.\\d+)?(?:[eE][+-]?\\d+)?$/".regex.compiled.matches text
     cant-parse
       "Can't convert text %s to number".printf
         * text
     head. ("%f".scanf text).at.^
   ^ > text
 
+  1000.eq "1e3".as-number ++> can-convert-exponent-text
+
   3.14.eq "3.14".as-number ++> can-convert-float-text
+
+  -0.0025.eq "-2.5E-3".as-number ++> can-convert-negative-exponent-text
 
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5083](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6975, opened 2026-08-17
#6959: give a ring of copies one name

### LINKED ISSUE #6959: `Ends.names()` fails to collapse a cyclic copy-chain in eo-inference
`Ends.names()` in `eo-inference/src/main/java/org/eolang/inference/Ends.java:43-54` doesn't do what its own docblock promises. The class comment (lines 18-20) says "a chain that comes back on itself is walked once and stops where it started", but the loop at line 48 (`while (this.copies.containsKey(end) && walked.add(end))`) advances `end` one hop past the point where the repeat is detected, so the walk stops one step too late and the final value depends on which key you started from rather than converging on one shared name for the whole cycle. Verified by running an unmodified `Ends` against `{a=b, b=a}`: `names()` returns `{a=b, b=a}` right back, completely uncollapsed; a three-node cycle `{a=b, b=c, c=a}` behaves the same way, while an acyclic chain like `{a=b, b=c}` correctly resolves both to `c`.

This matters because `Ends.names()` is the canonical-name resolver for the whole eo-inference pipeline: it's called from `Dispatched.answers()`, `Resolved.follow()`, `Demanded.follow()`, and `Fillings.demands()`. `Resolved.follow()` (lines 78-85) accumulates its `pairs` map over several `Settled`/`Dispatched` passes until one adds nothing, and `Dispatched.answers()` only guards against a direct one-hop self-copy (`!kept.equals(made)` at `Dispatched.java:97`) — it can't see a cycle that only closes once pairs from separate passes are combined, which is exactly the shape a mutually-recursive or self-referential chain of copies produces. When that happens, the resulting `names` map feeds `Bound`, `Refs`, and the `links.xml` table the checker relies on, so two locators that are really the same type end up attributed two different, mutually-inconsistent "canonical" names instead of one.

The smallest fix is to check whether the walk is about to repeat before stepping onto it, e.g. `while (this.copies.containsKey(end) && !walked.contains(end))` with the `add` moved before the `copies.get`, or record the first node visited and return that anchor once a repeat is seen — either way every member of a cycle should converge on the same canonical name.

morphqdd:
Closes #6959

The walk stepped one hop past the point where it saw a name for the second
time, so the answer depended on which name it started from: `{a=b, b=a}` came
back as `{a=b, b=a}`, two names for one type instead of one.

The starting name goes into the walked set now, so the walk stops on the name
it came back to, and that name is on the ring by construction. From there the
ring is read round once and its alphabetically first name is the answer, which
is the same wherever the ring is entered - `a` and `b` above both answer `a`,
and a chain that runs into a ring answers with the ring's name too.

A chain that ends somewhere is untouched: it stops where it always did, on the
name that is nobody's copy.

### DIFF
--- a/eo-inference/src/main/java/org/eolang/inference/Ends.java
+++ b/eo-inference/src/main/java/org/eolang/inference/Ends.java
@@ -16,8 +16,9 @@ import java.util.Map;
  * at a time: {@code a} is a copy of {@code b}, {@code b} of {@code c}. Asking
  * the table about {@code a} has to arrive at {@code c}, so each pair is
  * followed as far as it goes and the end is written down against every name on
- * the way. A chain that comes back on itself is walked once and stops where it
- * started, since an object that is a copy of itself is nothing new.</p>
+ * the way. A chain that comes back on itself is a ring of copies of one and
+ * the same thing, so every name on that ring answers with the same one - the
+ * first of them in alphabetical order - no matter which of them is asked.</p>
  *
  * @since 0.68.0
  */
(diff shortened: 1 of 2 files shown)


yegor256:
@morphqdd don't work on issues that are not labeled as bugs - you only waste your time and our time

PR #6977, opened 2026-08-17
#6886: widen `Winsock` socket descriptors to pointer width

### LINKED ISSUE #6886: `Winsock` socket descriptors are declared as 32-bit `int`, but the real `SOCKET` type is pointer-width
## What happens

`Winsock.java`'s JNA bindings declare socket descriptors as Java `int` (32-bit) everywhere: `socket()` returns `int`, `accept()` returns `int`, and `closesocket`/`connect`/`bind`/`listen`/`send`/`recv` all take `int sockfd`.

The real Windows API disagrees. `socket()`'s signature is:

```c
SOCKET WSAAPI socket(int af, int type, int protocol);
```

`SOCKET` is `typedef UINT_PTR SOCKET;` in the real Winsock headers (`ws2def.h`), and `UINT_PTR` is pointer-width: 64 bits on 64-bit Windows. Microsoft's own "Socket data type" documentation explicitly warns that `SOCKET` should not be treated like a small POSIX file descriptor (comparing it to -1 or using narrow integer types), because it is a distinct, wider type:
https://learn.microsoft.com/en-us/windows/win32/winsock/socket-data-type-2

`SocketFuncCall.java`, `AcceptFuncCall.java`, `ClosesocketFuncCall.java`, `ConnectFuncCall.java`, `BindFuncCall.java`, `ListenFuncCall.java`, `SendFuncCall.java` and `RecvFuncCall.java` all round-trip the descriptor through `int` (via `.asNumber().intValue()`), and `eo-runtime/src/main/eo/socket.eo` threads the same value through every subsequent call on a socket. Declaring a native parameter/return type narrower than the real ABI expects is undefined behavior in principle, the same class of bug as the already-fixed `_ftime32_s`/`struct __timeb32` issue (#6827): a JNA binding whose declared width doesn't match the real native type.

In practice this likely doesn't manifest as an observable crash today, because current Windows socket handle values happen to fit in the low 32 bits. But it's a genuine ABI-signature mismatch against the documented `SOCKET` type, not just a theoretical nitpick.

## What should happen

`Winsock`'s socket-descriptor parameters and return types should use a pointer-width type (JNA's `Pointer`, or a `long`, matching how the codebase already widened `_ftime64_s`'s time field) instead of `int`, matching the real `SOCKET`/`UINT_PTR` ABI.

morphqdd:
Closes #6886

`SOCKET` is `UINT_PTR`, so on 64-bit Windows it is eight bytes wide, while the
bindings declared it as `int`. The values happen to fit today, but the declared
signature is what JNA marshals by, and it did not match the one the library
exports.

Every socket descriptor in `Winsock` is `long` now - what `socket` and `accept`
return, and what `closesocket`, `connect`, `bind`, `listen`, `send` and `recv`
take - and `INVALID_SOCKET` is a `long` too, so a returned descriptor is still
compared against it directly. The eight `FuncCall` classes read the descriptor
with `longValue()` instead of `intValue()`; nothing else about them changes,
since an EO number carries the value either way.

Same shape as the `_ftime64_s` widening: the declared width follows the real
type instead of the value that happens to fit.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/win32/AcceptFuncCall.java
+++ b/eo-runtime/src/main/java/org/eolang/win32/AcceptFuncCall.java
@@ -39,7 +39,7 @@ public final class AcceptFuncCall implements Syscall {
             0,
             new Data.ToPhi(
                 Winsock.INSTANCE.accept(
-                    new Dataized(params[0]).asNumber().intValue(),
+                    new Dataized(params[0]).asNumber().longValue(),
                     new SockaddrIn(
                         new Dataized(params[1].take("family")).take(Short.class),
                         new Dataized(params[1].take("port")).take(Short.class),
--- a/eo-runtime/src/main/java/org/eolang/win32/BindFuncCall.java
+++ b/eo-runtime/src/main/java/org/eolang/win32/BindFuncCall.java
@@ -38,7 +38,7 @@ public final class BindFuncCall implements Syscall {
             0,
             new Data.ToPhi(
                 Winsock.INSTANCE.bind(
-                    new Dataized(params[0]).asNumber().intValue(),
+                    new Dataized(params[0]).asNumber().longValue(),
                     new SockaddrIn(
                         new Dataized(params[1].take("family")).take(Short.class),
                         new Dataized(params[1].take("port")).take(Short.class),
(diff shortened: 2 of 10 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5099](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6978, opened 2026-08-17
#6871: close a compact tuple when a `.method` continuation takes its place

### LINKED ISSUE #6871: Same-indent .method continuation after a compact tuple bypasses close-time checks, produces silently wrong XMIR
## What happens

A same-indent `.method` continuation after a compact-tuple head (`seq *`, per Appendix A of `PARSER_SPEC.md`, which lists compact-tuple as wrappable by a same-indent `.method`) produces silently wrong XMIR, with no parse error.

```
[] > foo
  seq *
    1
  .m > x
```

Parses clean, but the emitted structure is inverted. The `.m` link ends up wrapping the tuple, and `seq` ends up receiving the *result* of `.m` as its argument, instead of `.m` applying to the result of `seq *(1)`:

```xml
<o name="foo">
  <o base="Φ.seq">
    <o as="α0" base="ξ.x"/>
  </o>
  <o base=".m" name="x">
    <o base="Φ.tuple">
      ...the tuple elements...
    </o>
  </o>
</o>
```

A second symptom of the same defect: the `ms="0"` attribute, which `EoSyntax.parsed()` always attaches to the top-level `<object>` element, lands on `<o name="foo">` instead in this case, proving exactly one element is left open that should have been closed.

## Why it happens

`LnMethod.into()` (`eo-parser/src/main/java/org/eolang/parser/LnMethod.java:89`) closes the predecessor's open `<o>` with a bare `emit.close()` when a same-indent `.method` continuation extends it. Everywhere else a level is closed (`Stack.replace()`, `Stack.popTo()`, the end-of-file sweep), it goes through `Eo.checkOnClose()` instead, which does four things `LnMethod` skips:

- the R-6.6.2 all-or-nothing vertical-argument-binding check (`level.commitArg(null)`)
- the R-5.3.2 "reversed dispatch missing receiver" check for a `BARE_REVERSED` level
- closing (or synthesizing) the compact-tuple `Φ.tuple star=""` wrapper for a `COMPACT_TUPLE` level (R-5.3.3)
- an extra close for an open only-phi φ

Because `LnMethod` bypasses all of this, three more things break in the same situation, each verified separately:

**The compact-tuple argument is silently dropped** when nothing follows the `*` before the `.method` line:
```
[] > foo
  seq *
  .m > x
```
parses as if `seq` received no arguments at all — the `Φ.tuple.empty` wrapper never gets synthesized.

**The compact-tuple under-count check does not run:**
```
[] > foo
  seq *3
    1
  .m > x
```
parses with no error, while the same block without the trailing `.method` line correctly reports `compact tuple requires at least 3 children, got 1`.

**The bare-reversed missing-receiver check does not run:**
```
[] > foo
  size.
  .m > x
```
parses with no error and produces `<o loc="Φ.foo.x.ρ.ρ"/>`, an element with no `@base` at all standing in for the missing receiver, while the same line without the trailing `.method` correctly reports `reversed dispatch missing receiver`.

## What should happen

A same-indent `.method` continuation should close its predecessor through the same path every other close goes through (`Eo.checkOnClose()`, or an equivalent that runs the same four checks), so a compact-tuple wrapper closes correctly and the receiver/count/binding checks still run before the `.method` link opens.

morphqdd:
Closes #6871

A same-indent `.method` continuation closed the line above it with a bare
`emit.close()`, while every other place that closes a level goes through the
closer. So none of what the closer does happened: the `Φ.tuple` wrapper of a
compact tuple was never built - dropped altogether when the tuple had nothing
under it - the tuple's count was never checked, and a reversed dispatch with no
receiver was never reported. The dispatch also ended up under the head it was
supposed to wrap.

`Stack.seal()` runs the closer on the top entry without popping it, since the
entry stays and becomes the wrapper. The naming requirement is the one check
that cannot run there - a chain of continuations carries its name on the last
link - so the closer takes a flag for it, true everywhere a level is really
popped. After the closer, the entry drops its compact-tuple state: it is a
plain dispatch from then on and must not close a wrapper twice.

Four packs, one per symptom the issue lists: the nesting, the dropped empty
tuple, the count, and the missing receiver. Two unit tests in `LnMethodTest`
build their stack with a closer that closes, since the close is the closer's
job now.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -71,7 +71,7 @@ final class Eo implements Iterable<Directive> {
         final Globals globals = new Globals();
         final Emit emit = new Emit(this.source);
         final Stack stack = new Stack(
-            level -> Eo.checkOnClose(level, emit),
+            (level, naming) -> Eo.checkOnClose(level, emit, naming),
             parent -> Eo.beforeChild(parent, emit)
         );
         final java.util.List<Span> spans = new java.util.ArrayList<>(Eo.SPANS_CAPACITY);
@@ -893,19 +893,19 @@ final class Eo implements Iterable<Directive> {
      * <p>Checks: R-5.3.1 (naming requirement for plain children of
      * formations and top-level objects), the only-phi argument-naming
      * ban (§4.5), bare-reversed receiver presence, and the compact-tuple
-     * count. Atom-body and test-depth checks attach as their owning line
-     * shapes are implemented.</p>
+     * count.</p>
      *
      * @param level The level being closed
      * @param emit The directives sink
+     * @param naming Whether the naming requirement applies yet
      */
-    private static void checkOnClose(final Level level, final Emit emit) {
+    private static void checkOnClose(final Level level, final Emit emit, final boolean naming) {
         try {
             level.commitArg(null);
         } catch (final ParseError err) {
             emit.error(err.line(), err.pos(), err.getMessage());
         }
-        Eo.checkNaming(level, emit);
+        Eo.checkNaming(level, emit, naming);
         if (level.kind() == Kind.BARE_REVERSED && !level.taken()) {
             emit.error(
                 level.start(), level.indent(),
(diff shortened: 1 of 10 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (135 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5111](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6979, opened 2026-08-17
#6861: record a sheet by adding one element, not by copying the tree

### LINKED ISSUE #6861: TrStepped runs a full-document XSL transform on every shift just to record one sheet name
## What happens

`TrFull` wraps every parser/printer transformation train with `TrStepped`. After each individual XSL shift runs, `TrStepped` runs a second, full-document Saxon transform (`_stepped.xsl`) over the entire XMIR tree, purely to append one `<sheet>` bookkeeping element under `/object/sheets`, recording which shift just ran.

`_stepped.xsl` is a plain identity-copy stylesheet (`match="node()|@*"` -> `xsl:copy` + `apply-templates`), so every node in the document gets walked and copied again on every single shift, just to add one child element.

`TrFull` wraps `Canonical` (20 shifts) and the printer train `Xmir.FOR_EO` (9 shifts) unconditionally, so a single parsed-and-printed file runs roughly 20 to 29 extra full-document Saxon passes for this bookkeeping alone.

Measured with a JMH benchmark running the same 7-shift chain with and without `TrStepped` on a large XMIR document:

```
TrSteppedBench.bareTrain        2106.923 ms/op
TrSteppedBench.steppedTrain     3855.758 ms/op
```

`TrStepped` adds 83% on top of the bare train for this slice of the pipeline.

## What should happen

Recording that a shift ran should not require copying and re-walking the whole document tree through Saxon. A direct DOM mutation (clone the node, find or create `<sheets>`, append `<sheet>`) records the exact same fact without touching any node outside the one being appended to.

morphqdd:
Closes #6861

`TrStepped` ran a whole identity-copy stylesheet after every shift, so the
document was walked and rebuilt by Saxon once more per shift, only to note down
which sheet had just run. With `TrFull` wrapping the twenty-shift parser train
and the nine-shift printer one, that is a couple of dozen extra passes over the
tree per file.

`Sheets` does the same bookkeeping on the DOM: clone the node, find the
`sheets` element or make one, append a `sheet` to it. Nothing else in the tree
is read. `_stepped.xsl` is gone, and with it the `Once` scalar that existed to
load that sheet exactly once.

The tests keep the two things the stylesheet promised - the name lands under
`/object/sheets`, and it does so under concurrency - and a new one pins that an
earlier sheet survives the append, which the identity copy used to guarantee by
construction. The 318 printer packs print the same EO as before.


eo-parser/src/main/java/org/eolang/parser/TrStepped.java
```diff
@@ -4,115 +4,97 @@
...
+            final Document doc;
+            if (copy.getOwnerDocument() == null) {
+                doc = (Document) copy;
+            } else {
+                doc = copy.getOwnerDocument();
+            }
+            final Element root = doc.getDocumentElement();
+            final XML result;
+            if ("object".equals(root.getNodeName())) {
+                final NodeList found = root.getElementsByTagName("sheets");
```
yegor256: @morphqdd better use XNav API, not DOM directly

morphqdd:
@yegor256 done, it goes through `Xnav` and `Xembler` now, no DOM calls of its own.

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (161 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5115](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2488](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6976, opened 2026-08-17
`mvn clean site` failure due to unresolvable dependencies and Javadoc errors in `eo-maven-plugin`


yegor256:
Running `mvn clean site` in a fresh clone of `master` (currently 862e76b) does not work. Digging into it, there are three separate causes, chained one behind the other — fixing one just uncovers the next. The third one is also why `eo-maven-plugin-site.yml` has failed on the last two releases.

Environment: Apache Maven 3.9.11, OpenJDK 21 (project targets 17), Linux.

## 1. `mvn clean site` cannot resolve inter-module SNAPSHOTs

On a machine that has never run `mvn install` for this project:

```
$ mvn clean site
[ERROR] Failed to execute goal on project eo-printer: Could not resolve dependencies for project org.eolang:eo-printer:jar:1.0-SNAPSHOT
[ERROR] dependency: org.eolang:eo-parser:jar:1.0-SNAPSHOT (compile)
[ERROR] 	Could not find artifact org.eolang:eo-parser:jar:1.0-SNAPSHOT, try downloading from https://oss.sonatype.org/content/groups/public
```

The `site` lifecycle contains no `compile` or `package` phase, so no module in the reactor ever produces an artifact. `eo-parser` is never released as a public snapshot, so once `clean` has wiped `target/`, `eo-printer` (and everything downstream) has nothing to resolve against.

Confirmed: after a `mvn clean install -DskipTests`, plain `mvn clean site` reaches `BUILD SUCCESS` — the artifacts are then found in `~/.m2`. So this one has an easy workaround, but a fresh clone cannot build the site with the documented command.

## 2. `-Psite` forks `eo-maven-plugin:format`, which then cannot read `EO-Version`

This is the "lot of errors" case — around 180 `[ERROR]` lines:

```
$ mvn clean site -Psite
[INFO] --- eo:1.0-SNAPSHOT:format (compile) @ eo-runtime ---
[ERROR] Failed to process "number.as-char" (org.eolang.maven.TjForeign)
[ERROR] Failed to process "bytes.as-u8" (org.eolang.maven.TjForeign)
... 170+ more, one per .eo source ...
[ERROR] Attribute 'EO-Version' not found in MANIFEST.MF file(s) among 46 other attribute(s): [...]

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.22.0:site (default-site)
[...]

PR #6983, opened 2026-08-17
#6976: let the site javadoc report skip the missing-docs group

morphqdd:
Fixes the third item of #6976 - the one that keeps `eo-maven-plugin-site.yml`
red on every tag.

The parent's `site` profile runs javadoc with `-Xdoclint:all` and fails on any
warning, and `MjCoverageReport`, `MjInference` and `MjMerge` have no
constructor of their own, so javadoc warned about their implicit ones and the
site build stopped there. The `missing` group is what warns; qulice checks the
same thing on every build, class by class, method by method and field by field,
so the report skips that group now and still fails on anything javadoc calls
broken rather than absent.

I first added the three documented constructors, the way `MjClean` and the
other mojos have them, and the site built. Codacy rejected that: PMD reads an
empty constructor as `UnnecessaryConstructor`, and neither
`@SuppressWarnings("PMD.UnnecessaryConstructor")` nor `// NOPMD` helps, since
qulice's own PMD then reports the suppression as unnecessary - its ruleset does
not have that rule on. So the two analysers cannot both be satisfied by a
constructor, and this leaves the classes alone instead.

`mvn clean site -Psite -pl eo-maven-plugin`, the exact command the workflow
runs, reaches BUILD SUCCESS with this.

The other two items of the issue - a fresh clone cannot resolve the
inter-module SNAPSHOTs under the site lifecycle, and `-Psite` on the reactor
forks `eo-maven-plugin:format` into a build where `EO-Version` is unreadable -
are untouched here, so the issue stays open for them.

### DIFF
--- a/eo-maven-plugin/pom.xml
+++ b/eo-maven-plugin/pom.xml
@@ -592,4 +592,25 @@
       </build>
     </profile>
   </profiles>
+  <reporting>
+    <plugins>
+      <plugin>
+        <!--
+        The "site" profile of the parent runs javadoc with "-Xdoclint:all"
+        and fails the build on any warning, and a mojo has to have a public
+        no-argument constructor for Maven to build it with. Javadoc warns
+        about every implicit one of those, which stopped the site build and
+        left the last two releases without a published site. The "missing"
+        group is what warns, and qulice checks the same thing on every
+        build anyway, class by class, method by method and field by field,
+        so nothing goes unwatched by dropping it here.
+        -->
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <configuration>
+          <doclint>all,-missing</doclint>
+        </configuration>
+      </plugin>
+    </plugins>
+  </reporting>
 </project>


morphqdd:
Codacy flags the two new constructors as `UnnecessaryConstructor` (Info). They are unnecessary to the compiler and necessary to javadoc: without them `-Xdoclint:all` warns about the undocumented default constructor and the site build stops, which is the bug this fixes. `MjClean`, `MjResolve` and `MjAtomsTable` carry the same empty constructor, and the same Codacy pattern already sits on master in files like `EObytes$EOsize.java`.

A `@SuppressWarnings("PMD.UnnecessaryConstructor")` does not work either: qulice's own PMD then reports `UnnecessaryWarningSuppression`, since its ruleset does not have that rule turned on.

yegor256:
@morphqdd this doesn't look like a proper fix, but a suppression

morphqdd:
Closing this. The real fix is already in master, so the suppression is not needed.

`MjCoverageReport`, `MjInference` and `MjMerge` now each have an explicit
documented constructor, which is what javadoc was warning about:

```java
    /**
     * Ctor.
     */
    public MjCoverageReport() {
        // nothing
    }
```

They arrived with #7061. I ran the exact command the site workflow runs, on
current master and without the `<doclint>all,-missing</doclint>` change from
this pull request:

```
$ mvn clean site -Psite -pl eo-maven-plugin
[INFO] BUILD SUCCESS
```

So the report no longer has anything to skip.

#6976 stays open for its other two items: a fresh clone cannot resolve the
inter-module SNAPSHOTs under the site lifecycle, and `-Psite` on the reactor
forks `eo-maven-plugin:format` into a build where `EO-Version` is unreadable.

PR #6984, opened 2026-08-17
test: cover dataized.eo edge cases

Thayorns:
Adds 6 inline eo tests to dataized.eo for cases Codecov flagged as uncovered: dataizing a plain number, a nested attribute (the object from the class docblock), a string, a target that is already bytes, and both boolean outcomes.

The existing two tests only exercise dataized's interaction with malloc-based caching; these add straightforward coverage of the dataization itself, checking results against as-bytes/as-number of the same values rather than hardcoded byte sequences.

### DIFF
--- a/eo-runtime/src/main/eo/dataized.eo
+++ b/eo-runtime/src/main/eo/dataized.eo
@@ -70,3 +70,46 @@
         true
         m.put
           m.as-number.plus 1
+
+  eq. ++> can-dataize-a-plain-number
+    as-number.
+      dataized 42
+    42
+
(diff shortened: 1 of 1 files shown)


Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1831](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6986, opened 2026-08-17
Remove PhNest and the package branch of PhDefault.absent

### LINKED ISSUE #6500: returnsSameAnswerOnRepeatedLookup asserts has(cls) equals has(cls) in OnClasspathTest.java

maxonfjvipon:
The last step of #6616. Now that every package is merged, a package member is an attribute of the object the package names, so `PhDefault.absent` has nothing left to guess:

```java
private Phi absent(final String name) {
    final Phi object;
    if (this instanceof Atom) {
        object = this.take(Phi.LAMBDA).take(name);
    } else if (this.loaded().containsKey(Phi.PHI)) {
```

`PhNest` and `OnClasspath` go with the branch that used them, and `PhPackage.loadPhi` now tries the object class first and only falls back to a package, since a name stands for one or the other and never for both. `PhPackage.dispatched` existed only to keep a `PhNest` uncopied.

The `mergedPackages` parameter is gone and the merge is discovered instead: whatever sits before the last dot of a compiled name is a package, kept when an object of that name is compiled too. On eo-runtime that finds exactly the twelve packages the pom used to list — `Put 121 member(s) into 12 package object(s)`. A package no object is named after, as `examples` in a program that declares `+package examples`, keeps its members as objects of their own, which is what `refusesToMergeAPackageWithoutAnObject` used to call an error and what every program with a package of its own relies on.

Two consequences worth naming. A project can no longer drop `number/foo.eo` beside a jar-provided `number.eo` and reach it as `5.foo`; the explicit `number.foo 5` keeps working for as long as the package stays. And the attribute growth the issue asks to measure came with the merges themselves, not with this: nothing here adds an attribute, it only takes the classpath probe off the path of every missed name.

Closes #6500, whose only subject was `OnClasspath`.


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (832 >= 100); -8 for way too many hits-of-code (832 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+977](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6996, opened 2026-08-17
Remove unwarranted @DisabledOnOs(OS.WINDOWS) from DrProgramTest

### LINKED ISSUE #6992: `DrProgramTest.java` unintentionally disables Windows-specific tests
Three tests in `eo-parser/src/test/java/org/eolang/parser/DrProgramTest.java` carry `@DisabledOnOs(OS.WINDOWS)`, on lines 41, 53 and 64, and none of them does anything a Windows machine cannot do. `setsSchemaLocation` looks for a namespace declaration in the serialized document, which no operating system affects. `checksThatSchemaLocationPointToFile` asserts `startsWith("file:///")`, and on Windows `DrProgram.schema()` produces `file:///C:/...`, which starts with exactly that. `checksThatSchemaLocationPointToExistingFile` strips those eight characters and hands `C:/...` to `Paths.get`, which is the one platform where that string is a native path.

The irony is that `DrProgram.java:97` calls `.replace("\\", "/")` for no reason other than Windows, so the single line in the class written for that platform is the line these tests would have covered there and now never do. `StrictXmirTest.refersToAbsoluteFileName` at `eo-parser/src/test/java/org/eolang/parser/StrictXmirTest.java:89` does the same eight-character strip on the same attribute and checks `isAbsolute()`, and it runs on every operating system, so the arithmetic is demonstrably portable.

The annotations arrived in `c471cc001c` ("#3585: parser clean"), a hundred-file refactor whose message says nothing about Windows, so there is no recorded reason to keep them. Dropping all three, along with the two now-unused imports on lines 16 and 17, restores the coverage. If one of them does turn out to fail on Windows, the failure is a bug in `DrProgram.schema()` worth seeing rather than skipping.

@yegor256 please take a look.

Thayorns:
Three tests in DrProgramTest.java were marked @DisabledOnOs(OS.WINDOWS), but none of them assert anything platform-specific. setsSchemaLocation checks for a namespace declaration in the serialized document. checksThatSchemaLocationPointToFile checks that the schema URL starts with file:///, which holds on Windows too, since DrProgram.schema() produces file:///C:/... there. checksThatSchemaLocationPointToExistingFile strips that same eight-character prefix and hands the rest to Paths.get, exactly like StrictXmirTest.refersToAbsoluteFileName already does unconditionally for the same attribute.

The annotations trace back to a hundred-file refactor (c471cc001c) whose message never mentions Windows, so there's no recorded reason for them. Removing all three restores coverage for DrProgram.java:97's replace("\\\\", "/") line, which was written for Windows and is otherwise never exercised there. The two now-unused imports (DisabledOnOs, OS) go with them.

Closes #6992

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/DrProgramTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/DrProgramTest.java
@@ -13,8 +13,6 @@ import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.DisabledOnOs;
-import org.junit.jupiter.api.condition.OS;
 import org.xembly.Directives;
 import org.xembly.Xembler;
 
@@ -38,7 +36,6 @@ final class DrProgramTest {
     }
 
     @Test
-    @DisabledOnOs(OS.WINDOWS)
     void setsSchemaLocation() throws Exception {
         MatcherAssert.assertThat(
             "XSD location is set",
@@ -50,7 +47,6 @@ final class DrProgramTest {
     }
 
     @Test
-    @DisabledOnOs(OS.WINDOWS)
     void checksThatSchemaLocationPointToFile() throws Exception {
         MatcherAssert.assertThat(
             "URL of XSD is set to file",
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 happy to clarify anything here if needed.

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (5) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1843](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6997, opened 2026-08-17
Exclude a file's root object from the coverage manifest

### LINKED ISSUE #6995: Coverage report erroneously includes lines without dispatch
Look at this: https://app.codecov.io/gh/objectionary/eo/blob/master/eo-runtime%2Fsrc%2Fmain%2Feo%2Fbool.eo

The line is marked as lacking coverage:

```
[if] > bool
```

This line can never get any coverage, since it doesn't have a dispatch.

So, two options: either we exclude lines without dispatch from the report (somehow), or we wrap applications into `PhCoverage`.

Thayorns:
CoverageManifest.locations() picked up the top-level formation's own
line as an instrumentable location, but to-java.xsl never wraps that
line in PhCoverage: classes.xsl turns a file's root object into a
<class>, and its constructor is called once from generated Java, never
through the "located" mode any of its attributes go through. That line
therefore always shows as uncovered in the LCOV report, no matter how
many times the object is used.

`bool.eo`'s `[if] > bool` is exactly this shape: the free-attribute
formation itself never dispatches, only the attributes underneath it
do, so the report flags a line that structurally can never be hit.

The fix excludes the root `<class>` element from the XPath the
manifest walks. Locations found through a bound root object's inner
`o` (same loc:line:pos key) are unaffected, since that key is still
reached via the normal "located" mode call on the attribute itself.

Updated CoverageManifestTest to match: an empty formation now reports
zero locations instead of one, and a formation's own declaring line is
asserted absent from what gets found.

Closes #6995

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/CoverageManifest.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/CoverageManifest.java
@@ -24,13 +24,16 @@ import java.util.LinkedHashSet;
  * attributes are already present by the time every shift except the last
  * one has run, so running that same prefix here and reading the result
  * gets the exact same set {@code to-java.xsl} would act on, structurally,
- * rather than by pattern-matching the Java it emits. Two shapes never get a
- * hit and are left out here: {@code classes.xsl} marks a whole atomic class
- * {@code @skip-java}, and {@code to-java.xsl} then writes no {@code <java>}
- * for it at all; and the lambda marker of any atom (the {@code o[@name='λ']}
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 happy to clarify anything about this fix if needed.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1859](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6998, opened 2026-08-17
setsSchemaLocation checks the schema location attribute, not the namespace

### LINKED ISSUE #6991: `setsSchemaLocation` checks the namespace, not the location, in `DrProgramTest.java`
`setsSchemaLocation` in `eo-parser/src/test/java/org/eolang/parser/DrProgramTest.java:42` serializes the document with `toString()` and asserts it contains the literal `xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"`. That string is the namespace declaration, not the schema location: it tells you a namespace was declared, and it tells you which prefix and which quote character the serializer happened to pick. The value of `xsi:noNamespaceSchemaLocation`, which is what the test is named after and the only part `DrProgram` actually decides, is never read.

So the assertion answers a question nobody asked while the named behavior goes unchecked, and it is tied to serialization details that `DrProgram` does not control. Swap the serializer for a different prefix or single quotes and the test goes red with the schema location perfectly correct.

The two neighbouring tests in the same file already read the attribute properly through `Xnav`, at lines 57 and 69, after `2e687c521f` and `ad858ce500` converted them off string and XPath matching and left this one behind. Reading the attribute the same way here and asserting on its value would make the test check what its name promises, and the `@DisabledOnOs(OS.WINDOWS)` on line 41 can go with it, since nothing about a namespace declaration or an attribute value depends on the operating system.

@yegor256 please take a look.

Thayorns:
Closes #6991

`setsSchemaLocation` in `DrProgramTest` was serializing the document with `toString()` and asserting the result contained the literal `xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"`. That checks that a namespace was declared and which prefix and quote character the serializer picked, not the schema location the test is named after.

The fix reads `xsi:noNamespaceSchemaLocation` through `Xnav`, the same way the two neighbouring tests already do, and asserts the attribute is set. The `@DisabledOnOs(OS.WINDOWS)` annotation is dropped since reading an attribute value does not depend on the operating system.

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/DrProgramTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/DrProgramTest.java
@@ -38,14 +38,12 @@ final class DrProgramTest {
     }
 
     @Test
-    @DisabledOnOs(OS.WINDOWS)
     void setsSchemaLocation() throws Exception {
         MatcherAssert.assertThat(
             "XSD location is set",
-            new XMLDocument(new Xembler(new DrProgram()).xml()).toString(),
-            Matchers.containsString(
-                "xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\""
-            )
+            new Xnav(new XMLDocument(new Xembler(new DrProgram()).xml()).inner())
+                .element("object").attribute("xsi:noNamespaceSchemaLocation").text().get(),
+            Matchers.not(Matchers.emptyString())
         );
     }
 


Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1727](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #6999, opened 2026-08-17
rename tstartsen back to token and drop the dead Tstartsens link

### LINKED ISSUE #6989: Incorrect `Tokens` class references in `Eo.java` causing lack of resolution
`Eo.java:285` carries `{@link Tstartsens#hexDigit(char)}`. There is no `Tstartsens` anywhere in the repository — the class is `Tokens`, which the same file references correctly on line 184, and which `Value.java:13` and `LnApplication.java:201` link to properly. The reference resolves to nothing, and `hexDigit` is `private static` in `Tokens.java:638`, so pointing it at `Tokens#hexDigit` would not resolve either.

The same damage runs through seven more lines of `eo-parser/src/main/java/org/eolang/parser/Eo.java`, all of it a partial substitution of `oke` by `startse`: the prose on lines 201, 284 and 849 reads "terminates the tstartsen", "NAME tstartsens" and "source tstartsen", and the savepoint locals on lines 345 and 411 are named `tstartsen`, with their `emit.rollback` uses on 350 and 417. The word survives intact on lines 39, 184, 490, 642, 649, 657 and 659 of the same file, which is what marks this as accidental rather than a rename. `git blame` puts every occurrence in 769675b1af, the ANTLR replacement.

Renaming the two locals back to `token` and rewriting the prose fixes the readability; the broken `{@link}` on line 285 needs its target changed too, since `Tokens#hexDigit` is private — either drop the link and name the method in plain text, or point it at the public entry point that uses it.

@yegor256

Thayorns:
Eo.java carried a partial substitution of `oke` by `startse` in seven spots, left over from the ANTLR replacement in 769675b1af: the local variable `token` became `tstartsen` in two places, and the word survived mangled in four doc comments, including a `{@link Tstartsens#hexDigit(char)}` that resolves to nothing since the class is `Tokens` and `hexDigit` is private there anyway.

This renames the locals back to `token`, fixes the prose, and drops the broken link in favor of naming the method in plain text, since `Tokens#hexDigit` is private and wouldn't resolve either.

Closes #6989

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -198,7 +198,7 @@ final class Eo implements Iterable<Directive> {
      * Merge a BYTES continuation starting at index {@code start} —
      * concatenates the trailing-dash chunk with subsequent BYTES-only
      * lines at &gt;= the same indent until a chunk without trailing
-     * dash terminates the tstartsen (R-3.13).
+     * dash terminates the token (R-3.13).
      * @param spans Materialised list of source spans
      * @param start Index of the continuation start
      * @param stack Indent stack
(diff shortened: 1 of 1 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1739](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7000, opened 2026-08-17
Build schema location URLs from Path.toUri() to avoid duplicate slash

### LINKED ISSUE #6990: `DrProgram.java` and `StrictXmir.java` generate malformed schema location URLs with four slashes
`DrProgram.schema()` in `eo-parser/src/main/java/org/eolang/parser/DrProgram.java:96` builds the schema URL with `String.format("file:///%s", path.toString().replace("\\", "/"))`, where `path` came from `Paths.get(opt).toAbsolutePath()`. On Unix an absolute path already starts with `/`, so the three slashes in the format string become four and the `xsi:noNamespaceSchemaLocation` attribute of every generated XMIR reads `file:////home/.../eo-parser/src/main/resources/XMIR.xsd`. `StrictXmir.reset` at `eo-parser/src/main/java/org/eolang/parser/StrictXmir.java:171` builds its replacement location the same way, off `getAbsoluteFile().toString()`, and lands on the same four slashes.

The `file:` handler inside the JVM resolves `//home/x` the way it resolves `/home/x`, which is why validation still passes and why this went unnoticed. It is still a malformed URL sitting in a value people read in every XMIR file, and anything stricter than that handler rejects it: `StrictXmir.saved` at line 282 feeds the location to `new URI(uri)`, and tooling outside the JVM that reads XMIR has no reason to be as forgiving.

The fix is to let the path build the URL instead of a format string, so `path.toUri().toString()` on both sites, which yields three slashes on Unix and keeps `file:///C:/...` on Windows without the manual backslash replacement. Both tests that guard this attribute have to move in the same commit, because both currently pass only thanks to the extra slash: `DrProgramTest.java:59` asserts `startsWith("file:///")`, which a four-slash URL satisfies, and `DrProgramTest.java:71` strips `"file:///".length()` characters and gets an absolute path only because one slash is left over. `StrictXmirTest.java:98` strips the same eight characters and has the same coupling.

@yegor256 please take a look.

Thayorns:
DrProgram.schema() and StrictXmir.reset() both built the xsi:noNamespaceSchemaLocation URL with String.format("file:///%s", path). On Unix an absolute path already starts with a slash, so the format string's three slashes plus the path's leading slash produce four slashes, giving a malformed file:////... URL in every generated XMIR.

Both sites now build the URL from path.toUri().toString() instead, which yields the correct three-slash file:///... form on Unix and keeps file:///C:/... on Windows without any manual backslash replacement.

DrProgramTest and StrictXmirTest both had assertions that only passed because of the bug: they stripped exactly "file:///".length() characters off the location to recover the filesystem path. That coupling is gone now; those tests parse the location with java.net.URI instead of counting characters. Also added a test asserting the location never starts with four slashes.

Closes #6990

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/DrProgram.java
+++ b/eo-parser/src/main/java/org/eolang/parser/DrProgram.java
@@ -92,10 +92,7 @@ final class DrProgram implements Iterable<Directive> {
             }
             final Path path = Paths.get(opt).toAbsolutePath();
             if (path.toFile().exists()) {
-                schema = String.format(
-                    "file:///%s",
-                    path.toString().replace("\\", "/")
-                );
+                schema = path.toUri().toString();
                 break;
             }
         }
--- a/eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
+++ b/eo-parser/src/main/java/org/eolang/parser/StrictXmir.java
@@ -167,15 +167,12 @@ public final class StrictXmir implements XML {
             final String before = location.get();
             final String after;
             if (before.startsWith("http")) {
-                after = String.format(
-                    "file:///%s",
-                    StrictXmir.fetch(
-                        before,
-                        tmp.resolve(
-                            before.substring(before.lastIndexOf('/') + 1)
-                        )
-                    ).getAbsoluteFile().toString().replace("\\", "/")
-                );
+                after = StrictXmir.fetch(
+                    before,
+                    tmp.resolve(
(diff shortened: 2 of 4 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1755](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7001, opened 2026-08-17
clear buffered top comment and seal header before the seal throws

### LINKED ISSUE #6987: Rejected top comment block isn't properly cleared in `Comments.seal`, causing repeat errors
The `seal` method in `eo-parser/src/main/java/org/eolang/parser/Comments.java:52` throws the missing-blank-line error before it reaches `globals.clearComments()` and `globals.seal()` on lines 60-62, so a block the parser has just rejected stays in the buffer and the header zone stays open.

That makes the error repeat. `Eo.dispatch` in `eo-parser/src/main/java/org/eolang/parser/Eo.java:416` catches the `ParseError` and the walk carries on, and `LnMeta` clears the blank counter in `eo-parser/src/main/java/org/eolang/parser/LnMeta.java:75` only after the seal on line 73 returns. The next meta or object therefore meets exactly the same state and reports the same thing again: a file holding `# top doc`, `+package foo`, `+version 0.0.1`, a blank line and `[] > bar` reports "a blank line must separate the top comment block from the rest of the file" at line 2 and once more at line 3.

The rejected block also survives into the XMIR. `Eo.finish` in `eo-parser/src/main/java/org/eolang/parser/Eo.java:630` flushes whatever is still buffered into `/object/comments` with no seal check and no blank check, so `# top doc` followed straight by `[] > foo` yields both the error and `<comment line='1'>top doc</comment>`, attached to an object that recovery had already rolled back. That branch exists for a file that is nothing but comments, and it cannot tell that case apart from a block the seal refused.

Clearing the buffer and sealing the header before the throw in `Comments.seal` removes both symptoms at once. `EoTest.rejectsTopCommentWithoutBlankBelow` in `eo-parser/src/test/java/org/eolang/parser/EoTest.java:83` asserts only that the error is present, so it should also assert `/object[not(comments)]` and that the error appears exactly once.

@yegor256

Thayorns:
Comments.seal threw ParseError on a top comment block with no blank line before it, but left the block in the pending buffer and the header zone open. The next meta or object hit the same unsealed state and reported the identical error a second time. Eo.finish's end-of-stream flush also had no idea the block had been rejected, so a comment-only file like `# top doc` followed by `[] > foo` produced the error and still emitted `<comment>top doc</comment>` under an object whose block recovery had already discarded.

Clearing the buffer and sealing the header before the throw fixes both: the error now fires exactly once, and the rejected block never reaches the XMIR.

rejectsTopCommentWithoutBlankBelow in EoTest now also asserts /object[not(comments)] and that the error appears exactly once.

Closes #6987


Thayorns:
@yegor256 happy to clarify anything about this fix, just ask.

yegor256:
@Thayorns please, add YAML fixture to reproduce this issue

Thayorns:
@yegor256 added a YAML fixture (top-comment-without-blank-line-is-rejected-once.yaml) under eo-parser/src/test/resources/org/eolang/parser/eo-syntax/ that reproduces the issue.

Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4254](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7003, opened 2026-08-17
fix(#6981): balance close() calls, correct head columns, drop vacuous XPath in ChainEmissionTest

### LINKED ISSUE #6981: `close()` called excessively and incorrect head columns in `ChainEmissionTest.java` tests
`ChainEmissionTest.emitsHeadAndSingleMethodLinkAsFlatSiblings` at `eo-parser/src/test/java/org/eolang/parser/ChainEmissionTest.java:55` calls `emit.close()` three times where the emission opened only two elements. `ChainEmission.run` takes the non-empty-chain branch here, closes the head itself and leaves the last link open, so the wrapper `<o name='wrap'>` and that link are the only two the test owns. The other three tests in the same file close twice, as they should. The third `up()` steps the cursor off `/object` onto the document node, which Xembly tolerates, so the assertion still passes — but the test is asserting against an unbalanced directive stream, and one more stray `close()` turns it into a hard `IllegalStateException`.

`emitsBareHeadWithoutMethodLinkWhenChainIsEmpty` at line 28 builds its head as `new Value(Value.Kind.IDENTIFIER, "bar", 6, 9)` while the span it passes is `bar > foo`, where `bar` occupies columns 0 to 3. Columns 6 and 9 are where `foo` sits. The two sibling tests that use the same identifier get it right with `0, 3`. Nothing asserts on `@pos`, so the wrong columns go unnoticed, but the fixture now describes a line that does not exist and misleads the next reader about what the head of a chain looks like.

The second XPath in that same test, `/object/o[@name='foo' and not(o/o)]` on line 39, cannot fail. The empty-chain branch of `ChainEmission.run` opens exactly one element for the head and attaches attributes to it, so no grandchild is reachable on any input; the guard claims to prove the chain stayed flat and proves nothing. Dropping the extra `close()`, correcting the head columns to `0, 3`, and replacing the vacuous predicate with one that names the element that must not be there would leave all three tests saying what their names promise.

@yegor256

Thayorns:
Closes #6981

`emitsHeadAndSingleMethodLinkAsFlatSiblings` closed three times where `ChainEmission.run` only opens two elements on the chained path (the wrapper and the last unclosed link), so the third `close()` walked past `/object` onto the document node instead of failing loud.

`emitsBareHeadWithoutMethodLinkWhenChainIsEmpty` built its head `Value` with columns `6, 9`, which is where `foo` sits in `"bar > foo"`, not `bar`. The sibling tests use `0, 3` for the same identifier; this one now matches them.

The same test's second XPath, `not(o/o)`, could never fail: the empty-chain branch of `ChainEmission.run` opens exactly one element for the head, so no grandchild is ever reachable regardless of input. Replaced it with `not(o[@method])`, which actually names the method-link element that an empty chain must not produce.


Thayorns:

Thayorns:

Thayorns:

Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (5) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4278](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7004, opened 2026-08-17
Keep the method chain behind a hex literal argument

### LINKED ISSUE #6980: `Method chain is missing for hexadecimal literal argument in Emissions.java`
The `CHAINABLE` set in `eo-parser/src/main/java/org/eolang/parser/Emissions.java:40` enumerates every value kind that may be followed by a `.method` chain — `IDENTIFIER`, `ROOT`, `SELF`, `GROUP`, `INTEGER`, `FLOAT`, `STRING`, `BYTES` — and leaves out `HEX`. `Tokens.readArg` at `eo-parser/src/main/java/org/eolang/parser/Tokens.java:817` consults that set to decide whether to call `readChain()`, so a hexadecimal literal sitting in horizontal argument position never has its chain read at all.

Inside a paren group nothing reports the loss. `Emissions.group` hands the inner body to `Emissions.expression` (`Emissions.java:513`), which reads head, chain and args and ignores whatever the cursor has not consumed, so `f (g 0xFF.plus)` emits `g` with a bare `Φ.number` child and no `.plus` sibling at all, while the same line written `f (g 255.plus)` emits the `.plus` link as it should. Byte, string, float, identifier and root arguments all keep their chain — `HEX` is the only literal kind that silently loses it. Outside parens the unconsumed text travels on to `Suffix.parse` and comes back as `trailing garbage after expression`, which names the wrong culprit.

The same literal in head position parses correctly, because the head path calls `readChain()` unconditionally instead of going through `chainable`. Adding `Value.Kind.HEX` to the `CHAINABLE` set brings the argument path back in line with the head path, and a case in `TokensTest` reading the chain behind a `0x` argument would keep the omission from creeping back.

@yegor256

Thayorns:
Emissions.CHAINABLE listed every value kind a `.method` chain may follow except HEX, so Tokens.readArg never called readChain() for a hex-literal argument. Byte, string, float, identifier and root arguments kept their chain; a hex one silently didn't. Inside a paren group nothing reported the loss — `f (g 0xFF.plus)` emitted `g` with a bare `Φ.number` child and no `.plus` sibling.

Adding Value.Kind.HEX to CHAINABLE brings the argument path in line with the head path, which already calls readChain() unconditionally and never had this bug.

Added TokensTest.readsChainOnHorizontalHexArg, mirroring the existing BYTES-argument test, to cover the omission.

Closes #6980


Thayorns:

yegor256:
@Thayorns please, add YAML fixture to reproduce this issue

Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4294](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7005, opened 2026-08-17
dedupe chain emission through ChainEmission.link

### LINKED ISSUE #6982: `ChainEmission` functionality duplicated in `Emissions.expression` and `Emissions.emitArg`
The docblock of `ChainEmission` at `eo-parser/src/main/java/org/eolang/parser/ChainEmission.java:20` states that the class "is the single source of truth" for emitting a head plus its `.method` chain. It is not. The chain branch of `ChainEmission.run` at line 92 — open the head unnamed, close it, walk every link but the last emitting `<o base='.name' method=''>` and closing each, then open the last link carrying the name — appears again word for word in `Emissions.expression` at `eo-parser/src/main/java/org/eolang/parser/Emissions.java:87`, and a third time in `Emissions.emitArg` at `Emissions.java:173`.

Three copies of one rule means a fix lands in one of them and the other two keep the old behaviour, with nothing to say so. #6980 is that failure already: the argument path drops a chain the head path keeps, and the divergence is invisible precisely because the two paths are separate transcriptions of the same steps rather than one call. `LnApplication` and `LnCompactTuple` reach the shared class, while paren groups, inline-phi bodies and horizontal arguments each go through their own copy.

`Emissions.expression` and `Emissions.emitArg` already hold a head, a chain and a name at the point where they start transcribing, which is exactly what `ChainEmission`'s constructor takes, so both can delegate to it and keep their remaining work — the reversed-dispatch branch, the `@as` binding, the `@const` marker — around the call. That leaves the docblock true and gives the chain rule one place to be corrected.

@yegor256

Thayorns:
ChainEmission's docblock called it the single source of truth for emitting a head plus its `.method` chain, but the chain-emission loop was copied word for word into `Emissions.expression` and `Emissions.emitArg` as well. A fix landing in one copy could silently miss the other two, which is exactly the kind of divergence #6980 already ran into on the argument path.

This factors the shared steps — open the head, close it, emit every link but the last unnamed, then the last link carrying the name — into a new `ChainEmission.link` static method. `ChainEmission.run()` (used by `LnApplication` and `LnCompactTuple`) now delegates to it, and so do `Emissions.expression` and `Emissions.emitArg`. Each caller keeps its own remaining work around the call: the reversed-dispatch branch, the `@as` binding, the `@const` marker.

Added a test covering the previously-untested `emitArg` chain path — a `.method` chain on a horizontal argument — since none of the existing tests exercised it directly.

Closes #6982


Thayorns:
@yegor256 happy to clarify anything about this PR if needed.

Thayorns:

Thayorns:
@yegor256 hake a look at this one please

yegor256:
@Thayorns please, add YAML test fixture to reproduce this bug

Thayorns:
@yegor256 added `eo-parser/src/test/resources/org/eolang/parser/eo-syntax/chain-on-horizontal-argument.yaml`: `foo bar.baz > x` must parse into `<o base='Φ.foo'><o base='Φ.bar.baz'/></o>`, which is exactly the `emitArg` chain path that used to be a third transcription of the chain rule — if that path drops the chain, the argument comes out as `Φ.bar` and the fixture fails. It replaces the Java test I had for the same behavior, so there is one test per behavior.

Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (103 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4306](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7007, opened 2026-08-17
Exclude throwing tests from coverage instrumentation

### LINKED ISSUE #6993: Coverage report inaccurately includes negative tests, contrary to positive tests exclusion
Look at this: https://app.codecov.io/gh/objectionary/eo/blob/master/eo-runtime%2Fsrc%2Fmain%2Feo%2Fdirectory.eo

We report negative tests (with the `-->`) as code blocks and show coverage for them.

It's bug, we should exclude them just like we exclude positive tests (with the `++>`)

Thayorns:
Closes #6993

The coverage report includes negative (throwing, `-->`) tests as covered code, even though positive (truthy, `++>`) tests are already excluded.

The truthy-test exclusion in `to-java.xsl` works by checking whether an object's `@loc` locator contains a `+`, since a truthy test attribute's XMIR name carries a leading `+` marker (`+name`) that ends up in the locator. The throwing-test counterpart uses a leading `-` marker (`-name`) the same way, but the coverage condition never checked for it, so any dispatch inside a throwing test still got wrapped in `PhCoverage`.

The fix adds a second check that skips instrumentation when a locator segment starts with `-` right after a dot (`.-`), which only happens for a throwing-test marker segment — a bare `-` can't appear there otherwise, since kebab-case dashes only occur mid-segment.

Added two transpile tests: one confirming a throwing test's body stays free of `PhCoverage`, and one doing the same for a truthy test (there wasn't a dedicated regression test for that exclusion before either).

### DIFF
--- a/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
+++ b/eo-maven-plugin/src/main/resources/org/eolang/maven/transpile/to-java.xsl
@@ -697,7 +697,7 @@
       <xsl:value-of select="eo:literal(eo:escape-plus(@original-name))"/>
       <xsl:text>");</xsl:text>
     </xsl:if>
-    <xsl:if test="$coverage='true' and @line and @pos and not(contains(@loc, '+'))">
+    <xsl:if test="$coverage='true' and @line and @pos and not(contains(@loc, '+')) and not(contains(@loc, '.-'))">
       <xsl:value-of select="eo:eol($indent)"/>
       <xsl:value-of select="$name"/>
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 opened this against #6993 — happy to clarify anything about the fix.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1875](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7019, opened 2026-08-17
#6885: dataize the host chain of a `uri` once, through const bytes

### LINKED ISSUE #6885: `uri.eo`'s host/authority chain (`host-port`, `authority`, `hier-part`, `after-slashes`) recomputes on every read
## What happens

`uri.eo`'s `host` attribute chain recomputes several locals many times more than necessary, none of them `!`-forced:

- `host-port` (bound at `has-userinfo.if >> host-port`, line 89) is read about 6-7 times: inside `has-port`'s own body (lines 71, 75, 77, 86) and again as the else-branch of `host`'s outer `if.` (line 100) and inside `port`'s `raw-port` (line 113). Each read re-runs its own body, which itself calls `authority` up to 4 times.
- `authority` (bound at `if. > authority`, line 28) is read as a bare sibling 4 extra times beyond its own definition: inside `host-port`'s body (lines 90, 92, 97) and inside `userinfo` (line 156). Its own body chains `string.starts-with`, nested `if.`s, and depends on `hier-part`/`has-authority`.
- `hier-part` (bound at `has-query.if >> hier-part`, line 40) is read 3 times: inside `has-authority`'s check (line 30), inside its own enclosing `slice.` (line 44), and inside `path`'s else-branch (line 108).
- `after-slashes` (bound at `has-authority.if >> after-slashes`, line 38) is read twice: as the receiver of `slice.` (line 37) and again as the else-branch at line 49.

Because none of these are forced, a single `.host` or `.port` call on a `uri` re-walks this whole dependency chain (`host-port` -> `authority` -> `hier-part`/`has-authority` -> ...) many times over, compounding: with `host-port` alone read ~6-7 times and each read re-triggering `authority` (itself read another 4 times per `host-port` evaluation), `authority`'s own chain can end up dataized on the order of 20+ times for one `.host` call.

This is the same pattern already accepted and fixed in `string/regex.eo` (issue #6868, PR #6878) and `path.eo` (issue #6883): a local read more than once from sibling attribute bodies, without `!`, where the underlying computation is not trivial (string slicing/searching chained through several dependent locals here, not a single cheap lookup).

## What should happen

`host-port`, `authority`, `hier-part` and `after-slashes` should each dataize once per top-level call (e.g. via `!`) rather than once per read. Note: `regex.eo`'s attempted fix for the same pattern (#6868) initially broke a test by forcing a non-scalar match-result object; these four locals are all plain strings (slices/`if.`-branches of `text`), so the same pitfall should not apply, but each change should still be verified against the full `EOuriTest` suite individually, since `uri.eo`'s locals are chained through multiple `if.`/`>>`-gated branches and interact with each other's evaluation order.

morphqdd:
Closes #6885

`host-port`, `authority`, `hier-part` and `after-slashes` were recomputed on
every read, and the reads are nested, so one `.host` walked the chain many
times over.

Marking them with `!` alone does not work: the marker compiles to
`dataized(...).as-bytes` (R-9.4), so a const local is bytes and every string
method on it disappears. Each of the four, forced on its own, breaks
`EOuriTest` with `the attribute "length" is not found in []` and friends.

So each one keeps a const in bytes and a string around it, the shape `path.eo`
already uses (`string pth > text`): `string >> hier-part` over
`has-query.if >> hier-bytes!`, and the same for the other three. The bytes are
computed once per top-level call, the string wrapper around them is what the
rest of the file reads, and `authority` stays a string for whoever asks a `uri`
for it.

All 44 `EOuriTest` tests pass. On a straight `.host` benchmark
(`https://user:pw@example.com:8080/a/b/c?q=1#frag`, ten calls) this is about
seven percent slower than master - 75 s against 70 s - so the repeated reads
were not what the time went to. The chain is dataized once per call now, which
is what the issue asks for; the time itself is somewhere else and worth its own
measurement.

### DIFF
--- a/eo-runtime/src/main/eo/uri.eo
+++ b/eo-runtime/src/main/eo/uri.eo
@@ -25,28 +25,12 @@
 
 [text] > uri
   text > @
-  if. > authority
-    starts-with. >> has-authority
-      has-query.if >> hier-part
-        before-fragment.slice 0 question-idx
-        before-fragment
-      "//"
-    if.
-      not. >> has-authority-path
-        eq.
-          after-slashes.index-of "/" >> authority-path-idx
-          -1
-      slice.
-        has-authority.if >> after-slashes
-          hier-part.slice
-            2
-            hier-part.length.minus
-              number 2
-          ""
-        0
-        authority-path-idx
-      after-slashes
-    ""
+  string > authority
+    has-authority.if >> authority-bytes!
+      has-authority-path.if
+        after-slashes.slice 0 authority-path-idx
+        after-slashes
+      ""
   if. > fragment
     not. >> has-fragment
       hash-idx.eq -1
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5131](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6401, opened 2026-08-05
XMIR parser presents comment lines at incorrect positions


h1alexbel:
this is the program I'm parsing to XMIR:

```eo
# Hello.

+spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+spdx SPDX-License-Identifier: MIT

[] > hello
```

This is what I'm getting in XMIR:

```xml
<object xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        author="eo-parser"
        dob="2026-07-31T09:33:40"
        ms="0"
        revision="6f92ec6"
        time="2026-08-05T12:55:54.682545Z"
        version="0.62.1"
        xsi:noNamespaceSchemaLocation="https://www.eolang.org/xsd/XMIR-0.62.1.xsd">
   <listing># привет мы тут.

+spdx SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+spdx SPDX-License-Identifier: MIT
+unlint ascii-only:3

[] &amp;gt; hello
</listing>
   <comments>
      <comment line="3">Hello.</comment>
   </comments>
   <metas>
      <meta line="3">
         <head>spdx</head>
         <tail>SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com</tail>
         <part>SPDX-FileCopyrightText:</part>
         <part>Copyright</part>
         <part>(c)</part>
         <part>2016-2026</part>
         <part>Objectionary.com</part>
      </meta>
      <meta line="4">
         <head>spdx</head>
         <tail>SPDX-License-Identifier: MIT</tail>
         <part>SPDX-License-Identifier:</part>
         <part>MIT</part>
      </meta>
   </metas>
   <o line="6" loc="Φ.hello" name="hello" pos="0"/>
   <sheets>
      <sheet>wrap-applications</sheet>
      <sheet>resolve-self</sheet>
      <sheet>resolve-local-names</sheet>
      <sheet>validate-before-stars</sheet>
      <sheet>resolve-before-stars</sheet>
      <sheet>fragile-dispatch</sheet>
      <sheet>wrap-method-calls</sheet>
      <sheet>const-to-dataized</sheet>
      <sheet>stars-to-tuples</sheet>
      <sheet>vars-float-up</sheet>
      <sheet>move-voids-up</sheet>
      <sheet>validate-objects-count</sheet>
      <sheet>build-fqns</sheet>
[...]

PR #7031, opened 2026-08-18
#7008: EmitTest.recordsTargetLineOnCommentAttribute asserts a dead contract

### LINKED ISSUE #7008: EmitTest.recordsTargetLineOnCommentAttribute asserts a dead contract and cannot fail
eo-parser/src/test/java/org/eolang/parser/EmitTest.java:98 (recordsTargetLineOnCommentAttribute) passes three spans on lines 3, 4 and 5 with a target of 6, then asserts @line='6' — it hands the value in and reads the same value back out, so nothing but deleting the @line attribute entirely can turn this test red.

Its failure message at line 109, and the one in emitsCommentWithSingleLine at line 69, both say the @line attribute "must point at the line of the named object it attaches to". That contract is dead: commit 5617c921eb (#6401) changed the meaning to the last comment span's own line, and rewrote the javadoc on Emit.comment (eo-parser/src/main/java/org/eolang/parser/Emit.java:155) to "Line of the last comment span in the block". Both production callers, Comments.java:59 and Eo.java:632, now pass pending.get(pending.size() - 1).line() — the comment block's own last line, not "the named object it attaches to".

So the only prose description of this attribute inside EmitTest tells the next reader the opposite of what the parser actually does. Rewriting the two failure messages to match the current contract, and picking target/span line numbers in recordsTargetLineOnCommentAttribute that actually differ from the last span's line, would let the test catch a real regression instead of always passing.

morphqdd:
Fixes #7008.

`EmitTest.recordsTargetLineOnCommentAttribute` and `emitsCommentWithSingleLine` both described the `@line` attribute as pointing at "the line of the named object it attaches to". That description is stale since commit 5617c921eb (#6401), which changed the contract to the last comment span's own line and updated `Emit.comment`'s javadoc accordingly. Both callers (`Comments.java`, `Eo.java`) now pass the last span's line as `target`.

Reworded both failure messages to describe the real contract: the `@line` value is the target line given by the caller, which is the last comment span's line.

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/EmitTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EmitTest.java
@@ -66,7 +66,7 @@ final class EmitTest {
         final Emit emit = new Emit();
         emit.comment(Collections.singletonList(new Span("# hello", 7)), 8);
         MatcherAssert.assertThat(
-            "a single-line comment must produce <comments><comment line='8'>hello</comment>, where the line is that of the attached named object",
+            "a single-line comment must produce <comments><comment line='8'>hello</comment>, where the line is the target given by the caller",
             EmitTest.render(emit),
             XhtmlMatchers.hasXPaths(
                 "/object/comments/comment[@line='8']",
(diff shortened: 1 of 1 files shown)


morphqdd:
@yegor256, please review

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5243](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7032, opened 2026-08-18
#7009: EmitTest failure messages promise checks the XPath now makes

### LINKED ISSUE #7009: EmitTest failure messages promise checks the XPath never makes
emitsCommentJoiningMultipleLines in eo-parser/src/test/java/org/eolang/parser/EmitTest.java:79 claims "a multi-line comment must join the bodies with a newline", but the XPath assertion at lines 91-93 only checks that the text contains 'first' and contains 'second'. Change body.append('\n') to a space, or drop it entirely, at eo-parser/src/main/java/org/eolang/parser/Emit.java:172, and the test stays green — the newline it names in its own failure message is the one thing it never verifies.

The same shape shows up at EmitTest.java:39. emitsMetaWithoutParts promises "an empty <tail> and no <part> children", but the XPaths at lines 45-48 check the head and the absence of part, never the tail's own text. Delete .add("tail") at Emit.java:131 and this test stays green too.

Both tests should assert what their message claims: an xpath like contains(text(), concat('first', codepoints-to-string(10), 'second')) for the first case, and a not(tail/text()) or tail[not(text()) or text()=''] check for the second, so a regression in either behavior actually turns the test red.

morphqdd:
Fixes #7009.

Two `EmitTest` tests promise a check in their failure message that the XPath assertion never makes:

- `emitsCommentJoiningMultipleLines` says the comment bodies are joined "with a newline", but only checked `contains(text(),'first')` and `contains(text(),'second')`, which passes even if `Emit.comment` joined with a space or nothing.
- `emitsMetaWithoutParts` says a parameterless meta produces "an empty <tail>", but only checked the head text and the absence of `<part>`, never the tail's own content.

Added a real XPath assertion for each: `contains(text(), concat('first', codepoints-to-string(10), 'second'))` for the newline join, and `tail[not(text()) or text()='']` for the empty tail. Verified locally that both new assertions fail when the underlying behavior regresses (space-joined body, or a dropped `<tail>` element).

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/EmitTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EmitTest.java
@@ -44,7 +44,8 @@ final class EmitTest {
             EmitTest.render(emit),
             XhtmlMatchers.hasXPaths(
                 "/object/metas/meta[@line='2']/head[text()='foo']",
-                "/object/metas/meta[not(part)]"
+                "/object/metas/meta[not(part)]",
+                "/object/metas/meta[tail[not(text()) or text()='']]"
             )
         );
     }
@@ -89,7 +90,7 @@ final class EmitTest {
             "a multi-line comment must join the bodies with a newline",
             EmitTest.render(emit),
             XhtmlMatchers.hasXPath(
-                "/object/comments/comment[contains(text(),'first') and contains(text(),'second')]"
+                "/object/comments/comment[contains(text(), concat('first', codepoints-to-string(10), 'second'))]"
             )
         );
     }


morphqdd:
@yegor256, please review

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (5) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5271](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7033, opened 2026-08-18
#7010: rename marksObjectWithAsAttribute to name the slot() method

### LINKED ISSUE #7010: EmitTest.marksObjectWithAsAttribute names a method Emit does not have
eo-parser/src/test/java/org/eolang/parser/EmitTest.java:282 (marksObjectWithAsAttribute) and its failure message at line 289 both say "as() must attach @as='label'". Emit has no as() method — the method under test is slot(), declared at eo-parser/src/main/java/org/eolang/parser/Emit.java:386, and @as is just the attribute it happens to write.

Every other test in the file names itself after the method it exercises (marksObjectAsStar/star(), marksObjectAsConst/constant(), marksObjectAsSelf/self()), so this one reads as an outlier that describes the emitted XML instead of the operation. Anyone who greps Emit.java for as( after a red build here comes up empty.

Renaming the test to something like marksObjectWithSlotAttribute and rewording the failure message to reference slot() would fix it.

morphqdd:
Fixes #7010.

`marksObjectWithAsAttribute` and its failure message both refer to an `as()` method that `Emit` does not have. The method under test is `slot()`, which happens to write the `@as` attribute. Every other test in the file names itself after the method it exercises (`marksObjectAsStar`/`star()`, `marksObjectAsConst`/`constant()`, `marksObjectAsSelf`/`self()`), so this one was an outlier.

Renamed the test to `marksObjectWithSlotAttribute` and reworded the message to reference `slot()`.

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/EmitTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EmitTest.java
@@ -279,13 +279,13 @@ final class EmitTest {
     }
 
     @Test
-    void marksObjectWithAsAttribute() {
+    void marksObjectWithSlotAttribute() {
         final Emit emit = new Emit();
         emit.object(null, "foo", 1, 0);
         emit.slot("label");
         emit.close();
         MatcherAssert.assertThat(
-            "as() must attach @as='label' for the inline-binding marker",
+            "slot() must attach @as='label' for the inline-binding marker",
             EmitTest.render(emit),
             XhtmlMatchers.hasXPath("/object/o[@base='foo' and @as='label']")
         );


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5255](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7035, opened 2026-08-18
#7012: fix not(@name='xi🌵') filter that never excludes anything

### LINKED ISSUE #7012: The `not(@name='xi🌵')` filter in parser tests can never exclude anything
Several parser tests count the children of an object while trying to leave the auto-generated ones out, with a predicate like `count(o[not(@name='xi🌵')])=2`. The first one sits at `eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java:252`, and the same shape appears in `eo-parser/src/test/resources/org/eolang/parser/eo-syntax/`: `numbers.yaml:7`, `empty-vs-free.yaml:8`, `scopes.yaml:7`, `formation-idempotency.yaml:8-10`, `atoms-with-inner-tests.yaml:8`, `application-in-tests.yaml:8` and `full-syntax.yaml:18`.

Auto names come from `AutoName.asString()` at `eo-parser/src/main/java/org/eolang/parser/AutoName.java:37`, which formats `a🌵%d-%d`. The prefix is `a🌵`, not `xi🌵`, and every generated name carries a line and position suffix, so an equality test against the bare literal `xi🌵` is false for every node the parser can produce. The filter matches nothing and the assertions are really plain `count(o)=N` over all children, auto-named ones included.

These tests advertise that they count only the objects written in the source. The moment the parser emits an auto-named child at one of those levels, they break for a reason unrelated to what they check, and the exclusion that was supposed to absorb it is not there. Replacing the predicate with `not(starts-with(@name, 'a🌵'))` matches what `resolve-self.xsl:19` already uses to find the same nodes.

@yegor256 please take a look

morphqdd:
Fixes #7012.

Several parser tests try to count only the object's own children, excluding auto-named ones, with a predicate like `count(o[not(@name='xi🌵')])=2`. `AutoName.asString()` actually formats auto names as `a🌵%d-%d`, not `xi🌵`, so the filter matches nothing and every one of these assertions is really a plain `count(o)=N` over all children including auto-named ones.

Replaced `not(@name='xi🌵')` with `not(starts-with(@name, 'a🌵'))` in `EoSyntaxTest.java` and the 7 affected yaml packs (`numbers.yaml`, `empty-vs-free.yaml`, `scopes.yaml`, `formation-idempotency.yaml`, `atoms-with-inner-tests.yaml`, `application-in-tests.yaml`, `full-syntax.yaml`), matching the idiom `resolve-self.xsl` already uses to find the same nodes.

Verified locally: with the auto-name prefix temporarily changed to collide with the old literal `xi🌵`, the old filter let 9 assertions silently pass wrong values; with the fix in place, the same change makes those 9 tests fail as expected.

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
@@ -249,7 +249,7 @@ final class EoSyntaxTest {
             ).parsed(),
             XhtmlMatchers.hasXPaths(
                 "/object[count(o)=1]",
-                "/object/o[@name='base' and count(o[not(@name='xi🌵')])=2]",
+                "/object/o[@name='base' and count(o[not(starts-with(@name, 'a🌵'))])=2]",
                 "/object/o[@name='base']/o[@name='x']",
                 "/object/o[@name='base']/o[@name='f']"
             )
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/application-in-tests.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/application-in-tests.yaml
@@ -5,7 +5,7 @@
 sheets: []
 asserts:
   - /object[not(errors)]
-  - /object/o[@name='foo'][count(o[not(@name='xi🌵')])=3]
+  - /object/o[@name='foo'][count(o[not(starts-with(@name, 'a🌵'))])=3]
   - /object/o[@name='foo']/o[@base='Φ.true' and @name='+test-works']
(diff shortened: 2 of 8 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5279](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7036, opened 2026-08-18
Document why Listing drops the XML 1.1 restricted characters

### LINKED ISSUE #7013: `Listing` deletes control characters that XML 1.0 permits, so the listing is not verbatim
The `FORBIDDEN` pattern at `eo-parser/src/main/java/org/eolang/parser/Listing.java:29` strips `\x7F-\x84` and `\x86-\x9F` along with the C0 range, and line 52 removes every match before the source goes into `<listing>`. That upper range is the XML 1.1 *restricted* set, not the XML 1.0 forbidden set. `XMLDocument` emits XML 1.0, where `#x7F-#x9F` are ordinary legal characters in a text node, so nothing forces them out.

The docblock three lines above says the text value of `/object/listing` equals the source and that only characters XML cannot represent are dropped. For a source carrying a DEL or a C1 byte inside a string literal or a comment, neither statement holds: the byte silently disappears and the listing no longer round-trips. Before `f41bfc5cf6` the code called `escapeXml11`, which preserved exactly these characters as numeric references, so the fix for #6228 turned "escaped" into "deleted" for them.

Narrowing the pattern to the characters XML 1.0 actually forbids — `[\x00-\x08\x0B\x0C\x0E-\x1F￾￿]` — keeps the double-escaping fix and stops the data loss. `eo-parser/src/test/java/org/eolang/parser/ListingTest.java:47,83,87` asserts the current dropping of `0x7F` and `0x9F` and would need to assert they survive instead.

@yegor256 please take a look

Thayorns:
This started as an attempt to narrow `FORBIDDEN` to the XML 1.0 forbidden set, on the theory that `#x7F-#x9F` are ordinary characters in an XML 1.0 text node and so nothing should force them out of `<listing>`. That theory is wrong, and CI proved it: the new test died with `IllegalArgumentException: Failed to understand XML content, SET(...)`.

The source text never reaches `XMLDocument`. It goes through `Directives.set` first, and Xembly validates every character in `Arg.ifValid` against the XML 1.1 *restricted* set before any document version is chosen:

```java
Arg.range(chr, 0x00, 0x08);  Arg.range(chr, 0x0B, 0x0C);  Arg.range(chr, 0x0E, 0x1F);
Arg.range(chr, 0x7F, 0x84);  Arg.range(chr, 0x86, 0x9F);
```

Probed directly against `xembly-0.32.2`:

```
0x7F REJECTED: Character #7F is in the restricted XML range #7F-#84
0x85 accepted
0x9F REJECTED: Character #9F is in the restricted XML range #86-#9F
```

So `FORBIDDEN` was never a mistaken XML-version choice. It mirrors Xembly's restricted set exactly, gap at `0x85` included. Narrowing it does not make DEL and C1 bytes survive, it turns a silent deletion into a hard throw on any source that carries one, which is strictly worse than what master does today. `Xembler.escape` is the library's suggested way out, but it replaces a `0x7F` byte with a six-character backslash-u escape written out as literal text, so that is not verbatim either.

There is no fix that gets the characters through while `Listing` builds its element with `Directives.set`. What is left is the part that really was wrong: the docblock said only the characters XML forbids in text nodes are dropped, which names the wrong reason and is what made the behaviour look like a bug. It now names Xembly and points at `Directives#set`.

The regex is unchanged from master. The added test pins the other half of the contract, which nothing covered before: the characters sitting in the gaps between the restricted ranges, `0x09`, `0x0A`, `0x0D`, `0x20`, `0x85`, `0xA0` and `0xFFFD`, survive the round trip verbatim. `0x85` is the interesting one, since it is the single legal character inside the `0x7F-0x9F` band.

Closes #7013

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Listing.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Listing.java
@@ -16,18 +16,19 @@ import org.xembly.Directives;
  *
  * <p>The text is set as is, without any manual escaping: the XML writer
  * escapes it exactly once, so the text value of {@code /object/listing}
- * equals the source. Only the characters that XML forbids in text nodes
- * are dropped, since they can't be represented there at all.</p>
+ * equals the source. The characters of the XML 1.1 restricted set are
+ * dropped first, because {@link Directives#set(Object)} refuses them and
+ * throws, whatever version the writer emits later.</p>
  *
  * @since 0.1
  */
 final class Listing implements Iterable<Directive> {
 
     /**
-     * Characters that are not allowed inside an XML text node.
+     * Characters that Xembly refuses to put inside an XML text node.
      */
     private static final Pattern FORBIDDEN = Pattern.compile(
-        "[\\x00-\\x08\\x0B\\x0C\\x0E-\\x1F\\uFFFE\\uFFFF]"
+        "[\\x00-\\x08\\x0B\\x0C\\x0E-\\x1F\\x7F-\\x84\\x86-\\x9F\\uFFFE\\uFFFF]"
     );
 
     /**
--- a/eo-parser/src/main/java/org/eolang/parser/Listing.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Listing.java
@@ -27,7 +27,7 @@ final class Listing implements Iterable<Directive> {
      * Characters that are not allowed inside an XML text node.
      */
     private static final Pattern FORBIDDEN = Pattern.compile(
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 happy to clarify anything about this fix.

Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (21) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4334](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7037, opened 2026-08-18
Load naughty-string fixture from resources instead of a downloaded file

### LINKED ISSUE #7014: `parsesNaughtyString` needs the Internet, and fails confusingly without it
The argument provider `naughty()` at `eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java:760` reads `target/blns.txt`. That file exists only because `eo-parser/pom.xml:199` fetches it from `raw.githubusercontent.com` during the `initialize` phase. On a machine with no network the file is absent and `Files.readAllLines` throws `NoSuchFileException` out of the provider, so the whole parameterized test errors out with a message about a missing path rather than anything about parsing.

The path is also relative, resolved against whatever the working directory happens to be, so the same test behaves differently depending on where the JVM was started from. `${project.build.directory}` is what the download targets, and the test does not read that property.

Checking in the naughty-string list under `src/test/resources` and loading it with `ResourceOf` removes both problems: the test stops depending on a remote host at build time, and the fixture is found the same way every other one in this module is.

@yegor256 please take a look

Thayorns:
Closes #7014

The `naughty()` argument provider in `EoSyntaxTest` read `target/blns.txt`, a file that only exists because `maven-antrun-plugin` downloads it from `raw.githubusercontent.com` during the `initialize` phase. On a machine with no network the file is absent and `Files.readAllLines` throws `NoSuchFileException`, so the whole parameterized test errors out with a message about a missing path instead of anything about parsing. The path was also relative, so the test behaved differently depending on the working directory the JVM started from.

This checks in the naughty-string list as a classpath resource at `eo-parser/src/test/resources/org/eolang/parser/blns.txt` and loads it with `ResourceOf`, the same way every other fixture in this module is loaded. The antrun download plugin and the now-unused `Files`/`Paths` imports are removed.

The checked-in file is a curated 72-line sample of the upstream `minimaxir/big-list-of-naughty-strings` (MIT licensed), two representative strings per category, rather than the full 742-line list — enough to keep exercising every category (control characters, RTL text, zalgo, script/SQL/command injection, known CVEs, etc.) while keeping the diff small.


Thayorns:
@yegor256 happy to clarify anything about this fix, just let me know.

Thayorns:

yegor256:
@Thayorns conflicts here

Thayorns:
@yegor256 conflicts resolved, take a look at this one please

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4350](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7038, opened 2026-08-18
Drop no-op UTF-8 round-trip in EoSyntaxTest assertions

### LINKED ISSUE #7015: Seven assertions in `EoSyntaxTest` round-trip the XMIR through UTF-8 for nothing
Seven tests in `eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java` wrap the parsed document in `new String(...parsed().toString().getBytes(StandardCharsets.UTF_8), StandardCharsets.UTF_8)` before asserting on it: lines 62, 109, 136, 181, 573, 594 and 618.

Encoding a `String` to UTF-8 and decoding it straight back yields the same `String`, so the whole construction is a no-op wrapped around `toString()`. It buries the assertion under four lines of noise and suggests to a reader that some encoding concern is being handled here, which is not the case.

The newer tests in the same file already do it plainly — `rejectsProgramOfMetasAlone` at line 150 and `parsesArrow` at line 222 pass the `XML` straight to the matcher, and `XhtmlMatchers.hasXPaths` accepts it. Dropping the round-trip in the other seven brings them in line, and the `StandardCharsets` import at line 14 goes away with the last of them.

@yegor256 please take a look

Thayorns:
Closes #7015

Seven assertions in `EoSyntaxTest` wrapped the parsed XMIR in `new String(...parsed().toString().getBytes(StandardCharsets.UTF_8), StandardCharsets.UTF_8)` before handing it to the matcher. Encoding a `String` to UTF-8 and decoding it straight back yields the same `String`, so the wrapping was a no-op that only buried the assertion under noise.

This drops the round-trip in all seven places, passing `parsed().toString()` (or the `XML`/`XMLDocument` it feeds) straight to the matcher, matching how the newer tests in the same file already do it. The now-unused `StandardCharsets` import goes with it.


0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4366](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7039, opened 2026-08-18
fix(#7016): move chainable decision from Emissions onto Value

### LINKED ISSUE #7016: `Value` suppresses `PMD.DataClass` instead of being an object
`eo-parser/src/main/java/org/eolang/parser/Value.java:25` carries `@SuppressWarnings("PMD.DataClass")` over a class that holds seven attributes — `kind`, `raw`, `pos`, `end`, `binding`, `chain`, `constant` — and exposes each of them through an accessor at lines 129, 137, 145, 153, 162, 172 and 181. The checker is right: the class decides nothing, and every caller reads its fields and decides on its behalf. `Emissions` and the `Ln*` line shapes are where that logic sits today.

Two smaller things in the same file. The docblock links `{@link Head}` at line 21, and no `Head` type exists anywhere in `eo-parser/src/main/java` — the sentence describes a role that was renamed or removed. The same sentence then ends with `for readability. *`, a stray asterisk left over from an edit.

The suppression is the part worth acting on: rather than silencing the warning, the emission decisions the callers make from `kind()` and `raw()` belong on `Value` itself, which would drop the accessors that provoked it. The dangling `{@link Head}` and the stray asterisk are a one-line fix either way.

@yegor256 please take a look

Thayorns:
Closes #7016

Emissions kept a private CHAINABLE kind set and a static `chainable(Value)` helper that only read `Value.kind()` to decide whether a head value may carry a `.method` chain. That decision lived next to Value instead of on it. Value now answers `chainable()` itself, and `Tokens.readArg()` calls it directly instead of going through Emissions.

The issue also names the full fix: every kind()/raw()-driven decision that Emissions, LnMethod, LnPipe, LnCompactTuple, LnApplication, LnOnlyPhi and Bindings currently make from a Value should move onto Value itself, which would let the `@SuppressWarnings("PMD.DataClass")` come off. That's well beyond a single pull request, so this one moves the smallest self-contained piece — chainable() — and leaves the rest as a puzzle inside Value.java for a follow-up.

Along the way, Value's docblock loses a dangling `{@link Head}` reference (no `Head` type exists in this package) and a stray trailing asterisk left over from an earlier edit.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Emissions.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Emissions.java
@@ -9,7 +9,6 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
-import java.util.Set;
 
 /**
  * Shared {@link Value}-to-XMIR rendering helpers.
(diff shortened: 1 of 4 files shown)


Thayorns:

yegor256:
@Thayorns conflicts here

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4382](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7040, opened 2026-08-18
fix(#7017): un-invert the skipExperimentalLints filter in Linting

### LINKED ISSUE #7017: `eo.skipExperimentalLints` filter in `Linting` is inverted, so the flag reports what it should skip
## What happens

The filter that is supposed to honour `eo.skipExperimentalLints` is inverted, in both places it appears:

`eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java:461` (the whole-program path) and `Linting.java:496` (the per-file path) both read:

```java
.filter(defect -> this.skipExperimentalLints || !defect.experimental())
```

The flag comes straight from `MjSafe.skipExperimental` (`MjLint.java:51`, `MjCompile.java:51`), whose javadoc at `MjSafe.java:233` says "If set to TRUE, experimental lints are skipped during the linting", with `false` as the default. Nothing negates it on the way there, `grep -rn skipExperimental eo-maven-plugin/src/main` finds only pass-throughs.

So the expression does the opposite of what it says:

- `-Deo.skipExperimentalLints=true`: the predicate is always true, every experimental defect is kept, and the build can fail on exactly the lints the user asked to skip.
- default `false`: the predicate is `!experimental()`, so every experimental defect is dropped, and no configuration value exists that turns them on.

`org.eolang.lints.Source.defects()` and `org.eolang.wpa.Program.defects()` do no experimental filtering of their own, so this predicate is the only gate.

The inversion has been there since `37ea5b7e1c`. Nothing covers it: `grep -rn experimental eo-maven-plugin/src/test/java` only matches an unrelated `org.cactoos.experimental.Threads` import, and neither `LintingTest` nor `MjLintTest` mentions the flag. #6235 touched the flag only as part of the cache key.

## What should happen

`true` must skip experimental defects and the default must report them, i.e. the predicate is `!(this.skipExperimentalLints && defect.experimental())`, with a test for each of the two settings.

Thayorns:
The filter that honors eo.skipExperimentalLints in Linting.java was inverted in both the whole-program (WPA) path and the per-file path.

with -Deo.skipExperimentalLints=true the predicate was always true, so every experimental defect was kept, and the build could fail on exactly the lints the user asked to skip. With the default false, every experimental defect was silently dropped, so there was no way to see them at all.

This replaces the two inline predicates with a single Linting.reportable(skipExperimental, experimental) helper that implements !(skipExperimental && experimental), and adds unit tests for both settings.

Closes #7017


Thayorns:

eo-maven-plugin/src/main/java/org/eolang/maven/Linting.java
```diff
@@ -416,7 +416,7 @@ private List<org.eolang.wpa.Defect> wpa(final Map<String, XML> progs) {
             .without(this.programlints.toArray(new String[0]))
             .defects()
             .stream()
-            .filter(defect -> this.experimental || !defect.experimental()).forEach(
+            .filter(defect -> !(this.experimental && defect.experimental())).forEach(
```
yegor256, line 419: @Thayorns both expressions look identical to me...

Thayorns: @yegor256 
They differ exactly for the case that mattered here. Old: `this.experimental || !defect.experimental()` = `A || !B`. New: `!(this.experimental && defect.experimental())` = `!A || !B`.

These agree when `A` (skipExperimental) is `true`, but diverge when `A` is `false`: old reduces to `!B`, so it drops every experimental defect regardless of the flag; new reduces to `true`, so it reports them. That's the bug from #7017 — with `skipExperimental=false` the old predicate silently dropped all experimental defects, and with `true` it kept them (inverted). The new tests `reportsExperimentalDefectWhenSkipExperimentalIsFalse` and `skipsExperimentalDefectWhenSkipExperimentalIsTrue` fail against the old expression and pass against the new one.

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2240](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4398](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7042, opened 2026-08-18
fix(#7018): keep sign of negative zero in numeral φ-term

### LINKED ISSUE #7018: `Numeral` incorrectly prints `-0.0` as `0` causing misrepresentation of the number
## What happens

`Numeral.java:38` takes the integral branch whenever the value equals its own floor and fits into a `long`, and `-0.0` satisfies both (`Math.floor(-0.0)` is `-0.0`). Line 44 then prints `Long.toString((long) -0.0)`, which is `0`.

```java
new Data.ToPhi(-0.0d).φTerm()   // 0
new Data.ToPhi(0.0d).φTerm()    // 0
```

Both print the same, while the objects differ: their `delta` is `80-00-00-00-00-00-00-00` against `00-00-00-00-00-00-00-00`. Ran that on `b2958ff`.

The javadoc of the class (`Numeral.java:12`) says the branch exists so that a printed term does not denote a different number, which is exactly what happens here. `Data.ToPhi.number()` keeps the bit pattern on purpose, so the sign survives everywhere except the printed form.

`NumeralTest` has two tests, both about magnitudes outside `long` range, nothing about zero.

## What should happen

`-0.0` should print as `-0.0`, the way `Double.toString` renders it, so that reading the term back gives the same number. The project already treats the loss of that sign as a bug elsewhere: #6817, #6695, #6696, #6694 and #6588 all fix it in `number.*`.

Thayorns:
Closes #7018

`Numeral` took the integral branch for any value equal to its own floor and inside the `long` range, and `-0.0` satisfies both `Math.floor(-0.0) == -0.0`, so `(long) -0.0` and `Long.toString` collapsed it to `"0"`, losing the sign that `Data.ToPhi` deliberately preserves in the bit pattern.

The fix excludes negative zero from the integral branch by comparing raw bits against `-0.0d`, so it falls through to `Double.toString`, which renders `-0.0` correctly.

Added `printsNegativeZeroWithItsSign` to `NumeralTest`, alongside the existing magnitude tests.


0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4414](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7043, opened 2026-08-18
#7020: PhDefault.φTerm() renders non-UTF-8 bytes as lossy literal

### LINKED ISSUE #7020: `PhDefault.φTerm()` incorrectly renders non-UTF-8 bytes as lossy literal
## What happens

`PhDefault.java:354` builds the printed literal of a string with `new Quoted(raw).get()`, where `raw` is the object's `as-bytes` with no check that those bytes are text. `Quoted.get()` (`Quoted.java:39`) does `new String(this.data, StandardCharsets.UTF_8)`, and that decoder replaces every malformed byte with U+FFFD. `Escaped` (`Escaped.java:47`) escapes only codes below `0x20` and `0x7F`, so the replacement character goes into the term as it is.

```java
final Phi str = Phi.Φ.take("string").copy();
final Phi bts = Phi.Φ.take("bytes").copy();
bts.put(0, new Data.ToPhi(new byte[] {(byte) 0xFF, (byte) 0xFE, 0x41}));
str.put(0, bts);
str.φTerm();   // "��A"
```

The object holds `FF-FE-41`, the term says it holds two replacement characters and an `A`, and reading that term back gives a different object. `Quoted`'s own javadoc (`Quoted.java:14`) says the class is there to keep that from happening.

The same runtime already guards the same thing one class over: `PhApplication.string()` (`PhApplication.java:116`) decodes with `onMalformedInput(REPORT)` and `onUnmappableCharacter(REPORT)` and falls back to the structural form on `CharacterCodingException`. `PhDefault` does not.

Reachable from EO as `FF-FE-41.as-string`, since neither `bytes.as-string` nor `string.slice` enforces UTF-8.

`QuotedTest` has no malformed-input case. #5620 fixed the same class of corruption in the printer, #6450 added `Quoted` to `PhDefault.φTerm()` without the check.

## What should happen

`Quoted` should decode strictly and let the caller print the structural form (`[D> FF-FE-41]`) when the bytes are not text, so both call sites get the guard `PhApplication` already has.

morphqdd:
Fixes #7020.

`PhDefault.φTerm()` built a string literal with `new Quoted(raw).get()` without checking the bytes are valid UTF-8. `Quoted` decoded with `new String(data, StandardCharsets.UTF_8)`, which replaces malformed bytes with U+FFFD instead of failing, so the printed term could denote a different object than the one it came from.

`PhApplication.string()` already guarded against this with its own strict decode before calling `Quoted`, but `PhDefault` had no such guard.

Moved the strict decode into `Quoted` itself: it now decodes with `CodingErrorAction.REPORT` and returns `Optional<String>`, empty when the bytes are not valid UTF-8. `PhDefault.φTerm()` falls back to the structural form on empty. `PhApplication.string()` now just delegates to `Quoted`, dropping its own duplicate decode logic.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/PhApplication.java
+++ b/eo-runtime/src/main/java/org/eolang/PhApplication.java
@@ -5,11 +5,6 @@
 
 package org.eolang;
 
-import java.nio.ByteBuffer;
-import java.nio.charset.CharacterCodingException;
-import java.nio.charset.CodingErrorAction;
-import java.nio.charset.StandardCharsets;
-import java.util.Optional;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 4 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5287](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7044, opened 2026-08-18
#7021: RetryOnFailure in ObjectsIndex.asText is ineffective

### LINKED ISSUE #7021: `@RetryOnFailure` annotation in `ObjectsIndex.asText` is ineffective due to lack of I/O operations
## What happens

`ObjectsIndex.java:142` carries a retry on a method that performs no I/O:

```java
@RetryOnFailure(delay = 1L, unit = TimeUnit.SECONDS)
private static Text asText(final URL url) {
    return new TextOf(
        new Unchecked<>(() -> new TextOf(url).asString())
    );
}
```

`TextOf(Scalar)` is lazy and so is `Unchecked`, so the body only assembles objects and returns. The GET of `https://home.objectionary.com/objectionary.lst` (line 31) runs later, while `objects.value()` is being evaluated (lines 44-53, reached from `contains` and `children`), which is outside the annotated method. The aspect never sees the exception and never retries.

What that costs: one transient network failure while probing makes `ObjectsIndex.contains` throw, `OyIndexed` (`OyIndexed.java:57`) turns it into a warning and falls back to one HTTP request per object against `raw.githubusercontent.com`. And `Sticky` (line 63) does not memoize a failure, so the whole list download is retried, and fails again, once per probed object.

The sibling class does it right: `CommitHashesText.asText()` (`CommitHashesText.java:73`) calls `.asString()` inside the annotated method and returns a `String`.

That is the same defect #3136 reported and fixed for `CommitHashesText`. `ObjectsIndex` kept the old shape, and `ObjectsIndexTest` has nothing about retrying.

## What should happen

The download should happen inside the retried method, as in `CommitHashesText`, so that a transient failure costs a second rather than the whole index.

morphqdd:
Fixes #7021.

`ObjectsIndex.asText(URL)` carried `@RetryOnFailure` but only assembled a lazy `TextOf(Unchecked(...))` and returned. The actual GET ran later, outside the annotated method, while the returned `Text` was evaluated, so the retry aspect never saw the exception.

`CommitHashesText.asText()` already does this correctly: it calls `.asString()` inside the annotated method. Made `ObjectsIndex.asText` do the same, calling `new TextOf(url).asString()` inside the method and wrapping the resolved string, so a transient failure is retried where it happens instead of costing the whole index download.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/ObjectsIndex.java
@@ -13,7 +13,6 @@ import org.cactoos.Text;
 import org.cactoos.iterable.Filtered;
 import org.cactoos.iterable.Mapped;
 import org.cactoos.scalar.Sticky;
-import org.cactoos.scalar.Unchecked;
 import org.cactoos.set.SetOf;
 import org.cactoos.text.Split;
 import org.cactoos.text.TextOf;
@@ -138,11 +137,10 @@ final class ObjectsIndex {
      * Download from the URL and return the content.
      * @param url The URL with tags
      * @return The body of the web page
+     * @throws Exception If the download fails
      */
     @RetryOnFailure(delay = 1L, unit = TimeUnit.SECONDS)
-    private static Text asText(final URL url) {
-        return new TextOf(
-            new Unchecked<>(() -> new TextOf(url).asString())
-        );
+    private static Text asText(final URL url) throws Exception {
+        return new TextOf(new TextOf(url).asString());
     }
 }


morphqdd:
@yegor256, please review

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5199](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #7023, opened 2026-08-17
`eo.foreign` and `eo.foreignFormat` default settings mismatch causing incorrect file content and format


morphqdd:
## What happens

`MjSafe.java:64` defaults `eo.foreign` to `${project.build.directory}/eo-foreign.json`, and `MjSafe.java:73` defaults `eo.foreignFormat` to `csv`. `Catalogs.build` (`Catalogs.java:96`) chooses the mono from the format string alone, so a default build writes `MnCsv` content into a file named `.json`.

The sibling pair right below it is consistent: `eo.placed` is `eo-placed.json` and `eo.placedFormat` is `json` (`MjSafe.java:110`).

The repository works around the mismatch instead of relying on the default:

- `eo-runtime/pom.xml:277` sets `<foreign>${project.build.directory}/eo-foreign.csv</foreign>`
- `eo-integration-tests/src/test/java/org/eolang/maven/MjRegisterIT.java:187` reads `target/eo-foreign.json` with `new MnCsv(...)`

There is a second edge behind it: `Catalogs.make(Path, String)` keys its cache on the absolute path only (`Catalogs.java:81`), so if two callers ask for the same path with different formats, the first format wins and the second argument is dropped without a word.

## What should happen

The default name and the default format should agree - `eo-foreign.csv` with `csv`, or `json` with the `.json` name. And `Catalogs.make` should either key on path and format together or refuse a second format for a path it already built.

0crat:
@morphqdd Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+5203](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0pdd:
@morphqdd the puzzle [#7063](https://github.com/objectionary/eo/issues/7063) is still not solved.

0pdd:
@morphqdd the only puzzle [#7063](https://github.com/objectionary/eo/issues/7063) is solved here.

PR #7049, opened 2026-08-18
#7023: fix eo.foreign default name/format mismatch

morphqdd:
`eo.foreign` defaulted to `eo-foreign.json` while `eo.foreignFormat` defaulted to `csv`, so a default build wrote CSV content into a file named `.json`. The sibling pair right below it (`eo.placed`/`eo.placedFormat`) is consistent: both say `json`.

Changed the default to `eo-foreign.csv`, matching what `eo-runtime/pom.xml` already overrides it to, and what the rest of the repo assumes. Updated `MjRegisterIT.java`, which read the same mismatched filename with a CSV parser, to read `eo-foreign.csv` instead.

Left the second part of #7023 (`Catalogs.make` silently dropping a second format for a path it already cached) as a `@todo` puzzle in `Catalogs.java` — it's a separate, riskier change (cache keying), and the specific repro this PR fixes (foreign's own default fighting itself) no longer applies once the default agrees with the format. See the added `@todo` for the remaining part.

### DIFF
--- a/eo-integration-tests/src/test/java/org/eolang/maven/MjRegisterIT.java
+++ b/eo-integration-tests/src/test/java/org/eolang/maven/MjRegisterIT.java
@@ -184,7 +184,7 @@ final class MjRegisterIT {
             new TjSynchronized(
                 new TjCached(
                     new TjDefault(
-                        new MnCsv(temp.resolve("target/eo-foreign.json"))
+                        new MnCsv(temp.resolve("target/eo-foreign.csv"))
                     )
                 )
             )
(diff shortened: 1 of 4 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5223](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7051, opened 2026-08-18
#7025: fix Bytes.shift javadoc, direction stated backwards

### LINKED ISSUE #7025: Inconsistency in Javadoc for `Bytes.shift`: shift direction explained incorrectly
## What happens

The javadoc of `Bytes.shift` contradicts itself and the implementation. `Bytes.java:46`:

```java
/**
 * Big-endian unsigned shift.
 * Shifts left if value is positive, or right otherwise.
 * Does not perform sign extension.
 * @param bits Bits to shift, negative to shift left
 * @return Bytes
 */
Bytes shift(int bits);
```

The second line says positive shifts left, the `@param` says negative shifts left. The code follows the `@param`: `BytesRaw.java:76` does `if (bits < 0) shiftLeft else shiftRight`, and `EObytes$EOright.java:30`, the object named `right`, passes its positive argument straight through.

```
01-00 .shift(1)   ->  00-80   (right)
01-00 .shift(-1)  ->  02-00   (left)
```

This is the interface every `Bytes` implementation is written against, so the wrong sentence is what a new implementation would follow.

## What should happen

The sentence should read "Shifts right if the value is positive, or left otherwise", so the description, the `@param` and `BytesRaw` say the same thing.

morphqdd:
Fixes #7025.

`Bytes.shift`'s javadoc contradicted itself: the prose said "shifts left if value is positive", the `@param` said "negative to shift left". The implementation (`BytesRaw.shift`) and the `right` EO object both follow the `@param`, so the prose sentence was the wrong one. Fixed it to match.

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Bytes.java
+++ b/eo-runtime/src/main/java/org/eolang/Bytes.java
@@ -45,7 +45,7 @@ public interface Bytes {
 
     /**
      * Big-endian unsigned shift.
-     * Shifts left if value is positive, or right otherwise.
+     * Shifts right if the value is positive, or left otherwise.
      * Does not perform sign extension.
      * @param bits Bits to shift, negative to shift left
      * @return Bytes


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5247](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7052, opened 2026-08-18
#7027: file.deleted fails to remove dangling symlinks

### LINKED ISSUE #7027: `file.deleted` fails to remove dangling symbolic links due to `file.exists` resolving links
## What happens

`file.exists` (`eo-runtime/src/main/eo/file.eo:14`) is `access(path, 0)` compared with zero. POSIX `access` resolves symbolic links, so for a link whose target is gone it fails with `ENOENT` and `exists` is `false`:

```
access("/tmp/dangling", F_OK) -> -1, errno 2 (ENOENT)
lstat("/tmp/dangling")        -> ok, it is a link
unlink("/tmp/dangling")       -> 0
```

`file.deleted` (`eo-runtime/src/main/eo/file/deleted.eo:12`) gates the whole removal on that:

```
[] > deleted
  f.exists.if > @
    if.
      removed.code.eq 0
      ...
    f
```

so for a dangling link the else branch runs, `deleted` hands the file back as if it had done the work, and the link is still on disk. At the same moment `f.is-symlink`, which goes through `lstat` (`file.eo:61`), says the entry is there - the two objects disagree about the same path.

It spreads: `directory.deleted` removes entries by calling `.deleted` on each, so one dangling link inside makes the following `rmdir` fail, and the `mktemp.tmpfile.deleted` idiom the test suite uses everywhere has the same hole.

#6830 fixed the valid-symlink case (`9aa913caeb`) and its test `can-delete-a-symlink-to-a-directory` links to an existing directory. Nothing covers a broken one.

## What should happen

`file.deleted` should remove an entry that exists as a link even when its target does not. The existence test in front of it should be link-aware - `lstat`, or `access` together with `is-symlink` - instead of `access` alone.

morphqdd:
Fixes #7027.

`file.exists` was `access(path, 0) == 0`. POSIX `access` resolves symlinks, so a dangling link (target gone) makes `exists` say `false`. `file.deleted` gates removal on `exists`, so it handed the file back as if it had deleted it, and the dangling link stayed on disk. `directory.deleted` then fails removing a directory that still has that entry in it.

Made `exists` also true when `is-symlink` is true, the same lstat-based check `is-symlink` already does, so a link is treated as existing regardless of whether its target does.

Added `can-delete-a-dangling-symlink` next to the existing `can-delete-a-symlink-to-a-directory` test: creates a symlink to a directory, deletes the target first (leaving the link dangling), then confirms `deleted` still removes the link.

### DIFF
--- a/eo-runtime/src/main/eo/file.eo
+++ b/eo-runtime/src/main/eo/file.eo
@@ -10,21 +10,20 @@
   path > @
   @. > as-path
     Q.path path
-  [] > exists
-    or. > @
-      eq.
-        code.
-          os.is-windows.if
-            win32 *1
-              "_access"
-              path
-              0
-            posix *1
-              "access"
-              path
-              0
-        0
-      is-symlink false
+  or. > [] > exists
+    eq.
+      code.
+        os.is-windows.if
+          win32 *1
+            "_access"
+            path
+            0
+          posix *1
+            "access"
+            path
+            0
+      0
+    is-symlink false
   [cant-check] > is-directory
     if. > @
       info.code.eq 0
--- a/eo-runtime/src/main/eo/file.eo
+++ b/eo-runtime/src/main/eo/file.eo
@@ -10,18 +10,21 @@
   path > @
   @. > as-path
     Q.path path
-  eq. > [] > exists
-    code.
-      os.is-windows.if
-        win32 *1
-          "_access"
-          path
-          0
-        posix *1
-          "access"
-          path
-          0
-    0
+  [] > exists
+    or. > @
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5263](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7053, opened 2026-08-18
#7028: make win32 basename/dirname drive-aware

### LINKED ISSUE #7028: `path.win32.basename` and `path.win32.dirname` methods do not consider drive prefix causing incorrect results
## What happens

`path.eo` computes `basename` (line 21) and `dirname` (line 36) from `pth`/`trimmed`, which is the raw uri. Neither uses `driveless` (line 53), the value the same file already builds for exactly this purpose and which `is-absolute` (line 77) uses after #6141.

With the drive invisible to them, `last-index-of separator` is `-1` for a drive-relative path and `2` for a drive root, so the drive is either returned as part of the basename or treated as an ordinary directory name:

| expression | now | expected |
| --- | --- | --- |
| `path.win32 "C:foo" .basename` | `"C:foo"` | `"foo"` |
| `path.win32 "C:foo" .dirname` | `"."` | `"C:"` |
| `path.win32 "C:\\foo" .dirname` | `"C:"` | `"C:\\"` |
| `path.win32 "C:\\" .basename` | `"C:"` | `""` |

The expected column is what `node:path.win32` gives, which is what the rest of `path.eo` is written against.

The third row is the case #5802 already fixed for POSIX: `dirname` of a root must keep the separator, otherwise the answer is not a path to the root any more.

The block was last touched by #6888 for repeated trailing separators, which did not look at the drive.

## What should happen

Both should look for the last separator inside `driveless` and put the drive back, so that a drive-relative path has the drive as its dirname and a drive root behaves like a root.

morphqdd:
Fixes #7028.

`basename` and `dirname` computed everything from `trimmed` (the raw uri, drive included), never `driveless`, which `is-absolute` already builds for exactly this. On win32, a drive letter has no separator in front of it, so `last-index-of separator` found `-1` for a drive-relative path and `2` for a drive root, and the drive ended up either inside the basename or treated as a plain directory name.

Switched both to search inside `driveless` instead. `dirname` now prepends `drive` back to the result: a drive-relative path's dirname is the drive alone, a path under a drive keeps the drive in front of the separator, matching what `node:path.win32` gives and what the rest of the file is written against. POSIX is unaffected since `driveless` equals the full path there (`drive` is always empty).

Added four tests covering the table from the issue: drive-relative basename/dirname and drive-root basename/dirname.

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -24,7 +24,7 @@
           or.
             pth.size.eq 0
             index.eq 0
-          ^.trimmed >> pth!
+          pth
           as-bytes.
             text.slice
               plus. >> index!
@@ -32,20 +32,23 @@
                 1
               text.length.minus
                 number index
+      ^.stripped ^.driveless >> pth!
       string pth > text
     [] > dirname
       string > @
-        if.
-          len.eq -1
-          "."
+        concat.
+          ^.drive
           if.
-            len.eq 0
-            ^.separator
-            as-bytes.
-              text.slice
-                0
-                text.last-index-of ^.separator >> len!
-      string ^.trimmed > text
+            len.eq -1
+            "."
+            if.
+              len.eq 0
+              ^.separator
+              as-bytes.
+                text.slice
+                  0
+                  text.last-index-of ^.separator >> len!
+      string (^.stripped ^.driveless) > text
     sys.drive uri > drive
     if. > driveless
       drive.size.eq 0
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5491](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7059, opened 2026-08-18
Cover an object that takes its members inside a package

### LINKED ISSUE #7058: `merge-packs` doesn't cover scenario with object inside a package without object name
Every pack under `eo-maven-plugin/src/test/resources/org/eolang/maven/merge-packs/` builds its program around `number.eo` and a `number/` directory beside it. That leaves one layout untested, and it is the ordinary one for a program declaring a package of its own: an object sitting inside a package that names no object.

```
hello/obj.eo      +package hello       [x] > obj
hello/obj/abc.eo  +package hello.obj   [y] > abc
```

`abc` becomes an attribute of `obj`, and should. `Merging.deepest` cuts each compiled name at its last dot and keeps the prefix only when an object of that name is compiled too:

```java
.map(name -> name.substring(0, name.lastIndexOf('.')))
.distinct()
.filter(all::containsKey)
```

For `hello.obj.abc` the prefix is `hello.obj`, which is compiled, so the merge happens. `hello` names no object, so nothing is invented to hold `obj` and `obj` keeps the place it had.

Neither of the two packs that come near this covers it. `deeper-package-stays-out.yaml` does already exercise a receiver that is not top-level, since `plus` lands in `i64` and `i64` is `number.i64` — but there the receiver is itself merged upward into `number`, which is the case the `deeper` ordering exists for. `package-without-an-object-stays-apart.yaml` has a package no object carries the name of, yet nothing nested beneath it. Each half is covered and the combination is not.

Add a pack carrying the two-file program above, asserting `/object/o[@name='obj']/o[@name='abc']` on the merged `hello/obj.eo` and that no `hello.eo` is merged.

maxonfjvipon:
Closes #7058.

The merge packs all build their program around `number.eo` with a `number/`
directory beside it, so one layout goes untested — the ordinary one for a
program that declares a package of its own, where an object sits inside a
package no object is named after:

```yaml
eo:
  hello/obj.eo: |
    +package hello

    [x] > obj
      x > @
  hello/obj/abc.eo: |
    +package hello.obj

    [y] > abc
      y > @
xmir:
  hello/obj.eo:
    - "/object/o[@name='obj']/o[@name='abc']"
absent:
  - hello.eo
```

`abc` arrives in `obj` because `Merging.deepest` cuts a name at its last dot
and asks only whether an object carries the prefix right above the member.
`hello` carries none, so nothing is invented to hold `obj`.

`deeper-package-stays-out.yaml` does already exercise a receiver below the top
level, since `plus` lands in `i64` and `i64` is `number.i64` — but there the
receiver is itself merged upward, which is what the `deeper` ordering exists
for. Here nothing above the receiver moves at all. Removing the
`filter(all::containsKey)` from the discovery makes this pack fail, so it is
not merely restating what its neighbours already assert.

### DIFF
--- a/eo-maven-plugin/src/test/resources/org/eolang/maven/merge-packs/object-inside-a-package-takes-its-members.yaml
+++ b/eo-maven-plugin/src/test/resources/org/eolang/maven/merge-packs/object-inside-a-package-takes-its-members.yaml
@@ -1,0 +1,24 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# An object takes the members of the package it names without being merged
+# anywhere itself. `obj` lives in `hello`, which names no object, so nothing
+# above it moves and `obj` keeps the place it had. `abc` arrives all the same,
+# because a name is cut at its last dot and the merge asks only whether an
+# object carries the prefix right above the member.
+eo:
+  hello/obj.eo: |
+    +package hello
+
+    [x] > obj
+      x > @
+  hello/obj/abc.eo: |
+    +package hello.obj
+
+    [y] > abc
+      y > @
+xmir:
+  hello/obj.eo:
+    - "/object/o[@name='obj']/o[@name='abc']"
+absent:
+  - hello.eo


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+989](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #6369, opened 2026-08-04
`Data.ToPhi(null)` leaks `NullPointerException` instead of rejecting unsupported data consistently


gemshrine:
## What happens

`Data.ToPhi` documents that an unsupported Java data type is rejected immediately in its constructor. Passing `null` reaches the final `else` branch, which tries to obtain `obj.getClass().getCanonicalName()` while formatting the rejection message. Since `obj` is `null`, construction leaks a raw `NullPointerException` instead of the documented `ExFailure`.

## Problematic code

`eo-runtime/src/main/java/org/eolang/Data.java:157-160`:

```java
throw new ExFailure(
    "Unknown type of data: %s",
    obj.getClass().getCanonicalName()
);
```

## Steps to reproduce

```java
new Data.ToPhi(null);
```

The call throws `NullPointerException`.

## Expected behavior

`null` is not one of the supported primitive Java data values, so the constructor should reject it deterministically with `ExFailure`, just as it rejects every other unsupported type. The error should name `null` explicitly rather than dereferencing it while composing the diagnostic.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+1824](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7065, opened 2026-08-18
Drop the ConstructorsCodeFree suppression in Data.ToPhi

### LINKED ISSUE #7064: `Data.ToPhi` suppresses `ConstructorsCodeFreeCheck` where a delegating constructor would satisfy it
`Data.ToPhi` carries a `@checkstyle ConstructorsCodeFreeCheck` suppression at `eo-runtime/src/main/java/org/eolang/Data.java:55`, guarding the assignment `this.object = Data.ToPhi.toPhi(obj)` two lines below it. The suppression is avoidable: qulice's `ConstructorsCodeFreeCheck` exempts a constructor whose only statement is a delegating `this(...)` or `super(...)` call, and accepts method calls in that delegate's arguments (see its Javadoc and the `Valid.DelegateWithCall` / `Valid.SuperWithCall` fixtures in qulice).

The check is right to complain about the current shape — a method call as the right-hand side of a field assignment is exactly what it forbids, static calls included — so the fix belongs here rather than upstream. The suppression also hides the class from the rule permanently, so a future statement added to that constructor would go unnoticed.

Split the constructor: keep `public ToPhi(final Object obj)` as a secondary that delegates with `this(Data.ToPhi.toPhi(obj))`, and add a private primary `ToPhi(final Phi phi)` that only assigns `this.object`. That keeps the conversion eager, as the docblock and #6369 require, and lets the `@checkstyle` line go. The private constructor stays invisible to callers outside `Data`, so the 300-odd `new Data.ToPhi(...)` sites keep resolving to the `Object` overload.

yegor256:
`Data.ToPhi` now has two constructors: the public one delegates with `this(Data.ToPhi.toPhi(obj))`, and a private primary one only assigns the field. That drops the `@checkstyle ConstructorsCodeFreeCheck` suppression the class was carrying.

The suppression was avoidable. `ConstructorsCodeFreeCheck` is right to complain about a method call on the right-hand side of a field assignment, but it exempts a constructor whose only statement is a delegating `this(...)` call and accepts calls in that delegate's arguments. The conversion still happens inside the constructor, so an unconvertible type is rejected where the object is made, which is what the docblock and #6369 ask for.

Nothing else changes for callers: the private constructor is not accessible outside `Data`, so every `new Data.ToPhi(...)` site elsewhere keeps binding to the `Object` overload.

Closes #7064

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Data.java
+++ b/eo-runtime/src/main/java/org/eolang/Data.java
@@ -52,10 +52,17 @@ public interface Data {
          * is dataized far away from the place that made it.</p>
          *
          * @param obj Data
-         * @checkstyle ConstructorsCodeFreeCheck (5 lines)
          */
         public ToPhi(final Object obj) {
-            this.object = Data.ToPhi.toPhi(obj);
+            this(Data.ToPhi.toPhi(obj));
+        }
+
+        /**
+         * Ctor.
+         * @param phi Already converted object
+         */
+        private ToPhi(final Phi phi) {
+            this.object = phi;
         }
 
         @Override


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2248](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7067, opened 2026-08-18
Let Node print itself, instead of Pretty reaching into its fields

### LINKED ISSUE #7066: Node in eo-printer is an anemic DTO, forcing seven VisibilityModifier suppressions
All seven attributes of `Node` are declared package-private, and each one carries its own `@checkstyle VisibilityModifierCheck` suppression: `eo-printer/src/main/java/org/eolang/printer/Node.java:35,42,51,58,65,73,79`. They are package-private because `Pretty` reads them directly. Ten of its methods reach into `node.base`, `node.tail`, `node.children` and the four flags: `layout`, `shaped`, `suffixed`, `vertical`, `horizontal`, `phi`, `starred`, `inlined`, `flat` and `flattened`, at `eo-printer/src/main/java/org/eolang/printer/Pretty.java:113-553`.

Checkstyle is right here, and the class docblock at `Node.java:14-23` admits the design openly: the fields are exposed so that `Pretty` and its static helpers keep their direct field access. That is an anemic model. `Node` holds the state of a line-tree node while all the behaviour that acts on that state lives in another class, so the suppressions are the symptom rather than the disease. A package-private field is still writable by every class in `org.eolang.printer`, which is exactly what the check objects to.

Move the node-processing methods out of `Pretty` and into `Node`, so the fields can become private. The four that also need `Pretty` state (the `tab` string and the penalty `weights`) should take a small new collaborator holding those two, which keeps the dependency running one way as it does today. Beyond the seven suppressions this also retires the five static methods and the `Pretty.Flattened` helper class, whose only question, whether the resolved node still has children and therefore needs brackets, becomes a method on `Node` itself.

yegor256:
All seven attributes of `Node` in `eo-printer` were package-private, each with its own `@checkstyle VisibilityModifierCheck` suppression, because `Pretty` read them directly from ten of its methods. This moves the layout of a node into the node itself. It renders against a new `Style`, which carries the indentation width and the penalty weights, so nothing has to point back at the printer and the fields turn private. The four renderings become classes of their own, `Vertical`, `Horizontal`, `Phi` and `Starred`, each built by `Node` out of its own attributes, so none of them needs a getter either. `Arguments` spells an inlined argument list.

The suppressions were the symptom rather than a checkstyle false positive: a package-private field in a package-private class is still writable by the whole package, which is exactly what the check objects to. Along the way `Node.parse` becomes the `Node(Xnav)` constructor, the five static methods disappear, and the `Pretty.Flattened` helper goes with them, since its only question, whether a resolved node still needs brackets, is now a method on `Node`. The printer package carries no `@checkstyle` suppressions at all any more. Behaviour is unchanged: the 318 print-packs pass, along with `MjFormatTest` and `MjPrintTest`, and qulice is clean on the module.

One thing left for later. `Node` still holds seven attributes, above the usual four, and the obvious move is to group the four booleans into an object of their own. That ripples through every construction site, so it felt like a separate change rather than something to bolt onto this one.

Closes #7066


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1613 >= 100); -8 for way too many hits-of-code (1613 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2252](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7069, opened 2026-08-18
fix(#7056): report coverage per source file, not per merged package

### LINKED ISSUE #7056: `Non-executable meta lines improperly included in codecov report`
Look at this: https://app.codecov.io/gh/objectionary/eo/blob/master/eo-runtime%2Fsrc%2Fmain%2Feo%2Fchunk.eo

All meta lines, like `+architect` and `+version` must be excluded from the report.

They are not executable.

VasilevNStas:
Fixes #7056.

## Problem
`MjCoverageReport` read `tojo.xmir()` of every "standalone" tojo. For a package root, that XMIR is the **merged** one (see `Merging`), which contains all spliced package members whose `@line`/`@pos` refer to **their own** files. The report attributed those member lines to the package file, so they landed exactly on the package's meta lines (`+architect`, `+version`, etc.). Since metas have no `@pos` they can't be covered, and codecov showed them as uncovered. Package members themselves were excluded from the report entirely (`standalone()`), so their coverage was lost.

## Fix
- Iterate all `tojos.withXmir()` instead of `tojos.standalone()`, so package members are no longer dropped.
- Re-parse each source file with `EoSyntax(...).parsed()` instead of reading the stored (merged) XMIR, so every location is attributed to the file it actually belongs to.
- Metas have no `@pos`, so `CoverageManifest` never lists them — they no longer appear as misses.

The re-parsing is equivalent to the stored per-file XMIR: locations match for all 171 sources in `eo-runtime`.

## Tests
New regression test in `MjCoverageReportTest.doesNotAttributeMergedMembersToThePackageFile`: a package with metas + a member; asserts the package block has no `DA` misses on the meta lines and the member gets its own coverage block.

Verified locally that `eo-runtime` coverage (`-Deo.coverageTracking=true`) keeps the member files covered under their own paths (e.g. `chunk/to-output.eo` 100%) and the package meta lines are no longer DA misses.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjCoverageReport.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjCoverageReport.java
@@ -6,7 +6,6 @@ package org.eolang.maven;
 
 import com.jcabi.log.Logger;
 import com.jcabi.xml.XML;
-import com.jcabi.xml.XMLDocument;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -17,6 +16,7 @@ import java.util.Map;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
+import org.eolang.parser.EoSyntax;
 
 /**
  * Turns the raw {@code loc:line:pos} coverage hits {@code PhCoverage}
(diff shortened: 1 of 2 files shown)


0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+222](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7070, opened 2026-08-18
fix(#7054): exclude throwing tests from the coverage manifest

VasilevNStas:
Fixes #7054.

## Problem
Negative (`-->`) tests still show up in the codecov report as uncovered blocks (e.g. `posix.eo`), even though #6993 supposedly fixed them.

`to-java.xsl` wraps a location into `PhCoverage` only when its `@loc` carries neither `+` (truthy-test marker) nor `.-` (throwing-test marker, PR #7007). `CoverageManifest` mirrors that decision to build the report, but its XPath only excluded `+` — the `.-` condition was never ported there. So throwing-test bodies were still listed as report locations and shown as misses.

## Fix
Add `not(contains(@loc,'.-'))` to the `CoverageManifest` XPath, matching `to-java.xsl`.

## Tests
New regression test `excludesLocationOfAThrowingCase` in `CoverageManifestTest`: a throwing test's body must not produce any manifest location.

Verified end-to-end on `eo-runtime`: `posix.eo` no longer lists the `-->` lines as `DA` rows; EO object coverage went from 52.6% to 58.0% locally as 160 throwing tests stopped being counted as uncovered.

## Drive-by: xcop fix
`master` is currently red on xcop: commit `88fd937c2` moved a multi-line comment into `eo-maven-plugin/pom.xml` in a shape xcop rejects, so every PR fails `xcop`. This PR reformats that comment to the shape xcop expects (same pattern as #6212), unblocking CI.

PR #7072, opened 2026-08-19
Reformat the qulice excludes comment the way xcop wants

gemshrine:
`xcop` is red on `master` since 88fd937, and it blocks every open pull request.

The new comment in the `qulice` profile of `eo-maven-plugin/pom.xml` uses the inline continuation style, which `xcop` rejects:

```
<!-- Fixture reproducing the tab-separated output of
     the "git ls-remote" tags command -->
```

It wants the opening and closing markers on their own lines, with the text at the same indent, the way every other multi-line comment in that file already looks (see the `maven-core` dependency around line 129):

```
<!--
Fixture reproducing the tab-separated output of
the "git ls-remote" tags command
-->
```

The replacement here is byte-for-byte what `xcop` printed as the expected form in the failing run of `master`: https://github.com/objectionary/eo/actions/runs/32212352002

### DIFF
--- a/eo-maven-plugin/pom.xml
+++ b/eo-maven-plugin/pom.xml
@@ -577,8 +577,10 @@
             <artifactId>qulice-maven-plugin</artifactId>
             <configuration>
               <excludes>
-                <!-- Fixture reproducing the tab-separated output of
-                     the "git ls-remote" tags command -->
+                <!--
+                Fixture reproducing the tab-separated output of
+                the "git ls-remote" tags command
+                -->
                 <exclude>checkstyle:.*/src/test/resources/org/eolang/maven/commits/tags\.txt</exclude>
               </excludes>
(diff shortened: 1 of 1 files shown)


0crat:
@gemshrine It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`xcop-pom-comment`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`7071`". Your running score is [+2634](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

gemshrine:
Master already carries this exact reformatting, it arrived with #7070. Closing as redundant.

PR #7082, opened 2026-08-19
fix(#7030): treat a whitespace-only line as blank in Span

### LINKED ISSUE #7030: `Span.blank()` and `leading()` incorrectly identify a tab-only line as non-whitespace, resulting in silent deletion of subsequent elements
`Span.leading()` (`eo-parser/src/main/java/org/eolang/parser/Span.java:148-154`) counts only literal `' '` characters and stops at the first non-space char, including a tab. `blank()` (`Span.java:117-119`) then checks `indent == text.length()`. For a tab-only line (`text = "\t"`), `indent` comes out `0` while `text.length()` is `1`, so `blank()` wrongly returns `false` — contradicting the class's own contract at `Span.java:17-19`: "A blank line (entire line is whitespace) yields a span with `indent == text.length()`".

This misreport corrupts `Recovery`, whose `skipped()` (`Recovery.java:103-106`) and `rewound()` (`Recovery.java:88-94`) both trust `span.blank()` to know which lines belong to a block being skipped after a parse failure. Traced against this source:

```
[] > foo
  ???
\t
  d > y
```

Line 2 fails ("trailing garbage after expression"), so `recovery.after(1)` skips forward with reference indent 2. `skipped()` sees the tab-only line as `blank()==false`, `indent()==0`, so `0 > 2` is false too — the tab line is handed back immediately instead of being skipped. `Eo.process()` (`Eo.java:314-316`, current `origin/master`) then checks `span.tab()` before `span.blank()`, emits a second, unrelated "tab character in leading whitespace" error, and calls `recovery.after()` again with the same tab-blind indent of `0`. `Recovery.skip(_, 0)` then treats line 4 (`  d > y`, real indent 2) as still inside that bogus block, since `2 > 0`, and the walk runs straight to EOF.

I compiled `eo-parser` and ran `new Eo("[] > foo\n  ???\n\t\n  d > y\n").directives()` to confirm: the resulting XMIR has `foo` with no children, two errors ("trailing garbage..." on line 2 and "tab character in leading whitespace" on line 3), and no trace of `d > y` anywhere — no `<o name="y">`, no error naming it. An ordinary space-only blank line in the same position does not exhibit this: `rewound()` correctly hands it back as a separator without swallowing the next sibling.

The fix belongs in `Span`'s whitespace counting: `leading()` (and therefore `blank()`) should treat any leading whitespace character, not just `' '`, as contributing to blankness — e.g. blank when the trimmed line is empty — matching the documented contract. `tab()`'s existing detection can stay untouched for the R-2.2.4 diagnostic; only the blank/indent computation needs to stop being space-only.

This is a different defect than the related, already-closed #6423, #6427, #6485/#6496 in the same files — those fixed missing `failed = true` on the tab/odd-indent branches and `Recovery` swallowing genuine separator blanks. Those fixes are present in the current code; this bug is that `Span` itself misclassifies a tab-only line as non-blank with indent 0, which still corrupts `Recovery`'s bookkeeping regardless of those earlier fixes.

gemshrine:
Fixes #7030.

`Span.leading()` counted only literal `' '` characters, so a tab-only line got `indent == 0` while `text.length() == 1`, and `blank()` (which reads `indent == text.length()`) reported `false`. That contradicts the class contract in the javadoc and corrupts `Recovery`, which trusts `blank()` to know which lines belong to a skipped block.

Two changes:

1. `Span.leading()` now counts every leading whitespace character, not only spaces. A whitespace-only line therefore yields `indent == text.length()` and `blank() == true`, as documented. `tab()` is untouched, so the R-2.2.4 diagnostic still fires on `\tfoo`.

2. `Eo.process()` raises `tab character in leading whitespace` only for a line that carries content. Without this the tab-only line, now correctly blank, is still handed back by `Recovery.rewound()` as a separator, fails there, and restarts the skip with its own indent, swallowing the sibling again. R-2.2.4 speaks about the whitespace leading up to content, and a line made of whitespace alone has none.

For the source from the issue:

```
[] > foo
  ???
<TAB>
  d > y
```

`d > y` now lands under `foo` instead of disappearing, and only the genuine `trailing garbage after expression` error is reported.

Tests: three in `SpanTest` (tab-only line is blank, tabs count towards indent, `body()` of a tab-only line is empty) and one in `EoTest` reproducing the source above. Whole `eo-parser` suite is green locally.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -299,10 +299,6 @@ final class Eo implements Iterable<Directive> {
      * indent transitions run before classification when the stack is
      * non-empty.</p>
      *
-     * <p>The R-2.2.4 tab diagnostic speaks about the whitespace leading
-     * up to content, so a line made of whitespace alone stays out of it:
-     * such a line is a blank separator and gets dispatched as one.</p>
-     *
      * @param span The source span
      * @param stack The indent stack
      * @param globals The global parser state
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -299,6 +299,10 @@ final class Eo implements Iterable<Directive> {
      * indent transitions run before classification when the stack is
      * non-empty.</p>
      *
+     * <p>The R-2.2.4 tab diagnostic speaks about the whitespace leading
+     * up to content, so a line made of whitespace alone stays out of it:
+     * such a line is a blank separator and gets dispatched as one.</p>
+     *
      * @param span The source span
      * @param stack The indent stack
      * @param globals The global parser state
@@ -311,7 +315,7 @@ final class Eo implements Iterable<Directive> {
         boolean failed = false;
         if (globals.inTextBlock()) {
             Eo.continueTextBlock(span, stack, globals, emit);
-        } else if (span.tab()) {
+        } else if (span.tab() && !span.blank()) {
             emit.error(span.line(), 0, "tab character in leading whitespace");
             failed = true;
         } else if (!span.blank() && span.indent() % 2 == 1) {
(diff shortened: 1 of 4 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2742](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7090, opened 2026-08-19
fix(#7089): restore the LooseCoupling suppressions on Walk

### LINKED ISSUE #7089: qulice fails on master due to removal of '@SuppressWarnings' from Walk.includes() and Walk.excludes()
`qulice` fails on `master` since f463cebf9b, which removed the two `@SuppressWarnings("PMD.LooseCoupling")` from `Walk.includes()` and `Walk.excludes()`:

```
[INFO] PMD: eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java[53-53]: Avoid using implementation types like 'Walk'; use the interface instead (LooseCoupling)
[INFO] PMD: eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java[70-70]: Avoid using implementation types like 'Walk'; use the interface instead (LooseCoupling)
[ERROR] Failed to execute goal com.qulice:qulice-maven-plugin:0.30.14:check (jcabi-qulice-check) on project eo-maven-plugin: Failure: There are 2 violations
```

Every open pull request now carries these two, whatever it touches: see the `qulice` job of #7083, which changes nothing but `CoverageManifest`.

The commit message reads the rule as being about the parameter: "Both `includes()` and `excludes()` already take `Collection<String>`, an interface, so PMD's `LooseCoupling` never fires on them". PMD reports `[53-53]` and `[70-70]`, which are the two method declaration lines, and the type it names is `Walk`, not `ArrayList`. `LooseCoupling` covers the *return* type as well, and both methods return `Walk`, a `final class` extending `ListEnvelope<Path>`. `Walk` is what the suppressions were for, and dropping them makes the rule fire.

The fix is to put both back:

```java
```

```java
```

If PMD then reports them as `UnnecessaryWarningSuppression`, that report is about a different qulice version than the `0.30.14` the `qulice` workflow runs, and the workflow is the one that gates merges.

Unrelated to this, but the same job also fails `xcop` on the comment formatting in `eo-maven-plugin/pom.xml`; #7070 already carries that fix.

gemshrine:
Fixes #7089.

f463cebf9b dropped both `@SuppressWarnings("PMD.LooseCoupling")` on the grounds that the parameter is already a `Collection<String>`. The rule fires on the return type instead: PMD names `Walk` and points at lines 53 and 70, the two method declarations. `qulice` has been red on master since, on every pull request regardless of what it touches.

Putting the two annotations back is all that is needed.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Walk.java
@@ -50,6 +50,7 @@ final class Walk extends ListEnvelope<Path> {
      * @param globs List of them
      * @return New Walk
      */
+    @SuppressWarnings("PMD.LooseCoupling")
     Walk includes(final Collection<String> globs) {
         return new Walk(
             this.home,
@@ -67,6 +68,7 @@ final class Walk extends ListEnvelope<Path> {
      * @param globs List of them
      * @return New Walk
      */
+    @SuppressWarnings("PMD.LooseCoupling")
     Walk excludes(final Collection<String> globs) {
         return new Walk(
             this.home,


yegor256:
@gemshrine I'll fix it my way

Issue #7094, opened 2026-08-19
Invisibility of 756 parent hops due to reverse printing in `eo-runtime`


maxonfjvipon:
`eo-runtime` sources read a receiver 920 times. Only about 164 of those are written down. The other 756 are hops the compiler inserts and the printer then takes back out, so the file a human reads says less than the program means.

`build-fqns.xsl` walks a `$rhos` counter and emits one `ρ` segment per level a bare name has to climb. `to-eo-tree.xsl:390` removes the run again when printing, and its own comment says exactly when:

```
The run of leading "^." parent hops is redundant: the name is
absent from each of the N nearer enclosing formations but
present in the (N+1)-th, so plain scope resolution re-derives
exactly the very same ρ-chain of length N. Drop the whole run.
```

The two are inverses, which is what keeps sources terse — and what hides them. Counting `@base` attributes carrying `ρ` across the 171 parsed files of `eo-runtime` gives 803 chains of one hop, 86 of two, 20 of three, 2 of four and 4 of six. The deepest live in `file.eo`, written in the source as bare names:

```
ξ.ρ.ρ.ρ.ρ.ρ.ρ.readable.not.if
ξ.ρ.ρ.ρ.ρ.syscall.read
```

Nothing in `file.eo` tells its reader that `readable` is six formations up. That is worth fixing on its own, and it is a precondition for #6985: a receiver cannot be declared as a void while most reads of it are invisible.

Drop the branch, so a parent hop prints as the `^.` the program actually performs, and let `eo:format` rewrite the sources. Nothing about meaning changes: printing is the inverse of parsing here, so the sources can be checked by re-parsing them and diffing the XMIR against what the current ones produce. It must come out byte-identical.

Stopping `build-fqns.xsl` from inserting the hops in the first place is the natural follow-up, and only makes sense once the sources spell them out.

0crat:
[...]

PR #7096, opened 2026-08-19
Keep a folded value in the scope it was written in

### LINKED ISSUE #7095: `merge-monikers.xsl` incorrectly relocates bindings into deeper scopes without adjusting names
`merge-monikers.xsl` moves a binding into the reference that hosts it, and the reference can sit deeper than the binding did. The names inside the moved value are not adjusted for that, so they are read again in a scope that is not the one they were written in. When the hosting formation happens to declare a name the value uses, the printed program means something else.

```
[b] > array
  b.size >> len!
  [b] >> slice-byte
    b.plus len > @
  slice-byte 1 > @
```

`len` is the size of the `b` that `array` was given. The only reference to it sits inside a formation with a `b` of its own, and after the merge the printer writes:

```
[b] > array
  b.plus b.size! > [b] >>
  | 1 > @
```

Both `b`s now denote the inner void, so `slice-byte 1` reads `1.plus 1.size` where the source said `1.plus array.b.size`. Reverse printing is supposed to preserve meaning, and here it does not.

The merge happens at `eo-printer/src/main/resources/org/eolang/printer/print/merge-monikers.xsl:405`, which copies the binding's attributes and children onto the reference site:

```xml
<xsl:copy-of select="$merged/@*[...]"/>
<xsl:copy-of select="$merged/node()"/>
```

Nothing re-bases the copy. A `ξ.`-rooted name in the moved value refers to the scope it was written in, so it needs one `ρ` per level the host sits below the binding, and a name that already climbs needs the same addition once its climb leaves the moved subtree. Names that stay inside the moved value, or that are rooted at `Φ`, must be left alone.

The defect is invisible today whenever the hosting scope declares nothing of the same name, because the missing hops are exactly the ones plain scope resolution puts back when the printed text is read again. That is also why it hides a second symptom: printing stops being idempotent, since the second pass writes the hops the first one omitted.

maxonfjvipon:
Closes #7095.

`inline-cactoos.xsl` folds a `>> name` handle onto its reference, and the
reference can sit deeper than the binding did. The names inside the folded
value were carried over untouched, so they were read again in a scope that
was not the one they were written in:

```
[b] > array
  b.size >> len!
  [b] >> slice-byte
    b.plus len > @
  slice-byte 1 > @
```

`len` is the size of the `b` that `array` was given, and the only reference
to it lives in a formation with a `b` of its own. Printed, both `b`s bound
to the inner void, so `slice-byte 1` read `1.plus 1.size` where the source
said `1.plus array.b.size`. Reverse printing is meant to preserve meaning.

The fold now re-bases what it moves:

```xml
<xsl:apply-templates select="$folded" mode="dropped">
  <xsl:with-param name="drop" select="eo:host-drop($target, .)" tunnel="yes"/>
</xsl:apply-templates>
```

A name that climbs at least as far as its own depth inside the value has
left it, and gains one hop per formation the destination sits below the
source. Shorter climbs stay inside the value and are untouched, as is
anything rooted at `Φ`.

This is the based-handle fold only. The other four branches of that
template and the three merges in `merge-monikers.xsl` relocate values the
same way and need the same pass, each with a pack of its own since the
shapes differ — left as a `@todo`, since the helper should move somewhere
both sheets can import rather than be copied.

Found while making parent hops explicit for #7094, where the old printer
masked this: the missing hops were exactly the ones scope resolution puts
back when the text is read again.

### DIFF
--- a/eo-printer/src/main/java/org/eolang/printer/Xmir.java
+++ b/eo-printer/src/main/java/org/eolang/printer/Xmir.java
@@ -33,6 +33,17 @@ import org.xml.sax.SAXParseException;
  *
  * @see <a href="https://xml.jcabi.com">xml.jcabi.com</a>
  * @since 0.35.0
+ * @todo #7095:60min Re-base every value the print train relocates. A binding
+ *  folded onto a reference deeper than itself is read again in a scope that
+ *  is not the one it was written in, so its names have to climb the
+ *  difference. Only the based-handle fold in {@code inline-cactoos.xsl} does
+ *  that today; the four other branches of the template it lives in, and the
+ *  three merges in {@code merge-monikers.xsl}, relocate values the same way
+ *  and need the same pass. Each wants a pack of its own before it is wired,
+ *  since the shapes differ: an applied reference keeps its own children, a
+ *  pipe keeps the formation in place. The {@code dropped} mode and its two
+ *  functions should move somewhere both sheets can import rather than being
+ *  copied into the second one.
  */
 public final class Xmir implements XML {
 
--- a/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/inline-cactoos.xsl
@@ -268,8 +268,13 @@
               <xsl:if test="@name and not($keep-name)">
                 <xsl:apply-templates select="@name"/>
               </xsl:if>
-              <xsl:apply-templates select="$value/@*[$keep-name or (name() != 'name' and name() != 'local')]"/>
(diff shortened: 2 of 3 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (109 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1005](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7099, opened 2026-08-19
Write down the parent hops a name climbs

### LINKED ISSUE #7094: Invisibility of 756 parent hops due to reverse printing in `eo-runtime`

maxonfjvipon:
Closes #7094.

`eo-runtime` reads a receiver 920 times and writes down about 164 of them.
The rest are hops `build-fqns.xsl` inserts and `to-eo-tree.xsl` takes back
out, so the file a reader opens says less than the program means. The two
are exact inverses, and the printer's own comment said when it dropped a
run:

```
The run of leading "^." parent hops is redundant: the name is
absent from each of the N nearer enclosing formations but
present in the (N+1)-th, so plain scope resolution re-derives
exactly the very same ρ-chain of length N. Drop the whole run.
```

That branch is gone, and `eo:format` wrote the hops into the sources. The
deepest were in `file.eo`, six levels climbed under a bare name:

```
^.^.^.^.^.^.readable.not.if > @
  ^.^.^.^.syscall.read
```

84 of 171 sources changed, one hop per line, 594 lines either way. 21 of
the 144 print packs record the new canonical form; `multi-hop-parent` and
`redundant-parent` still hold, the second now saying that a redundant hop
is kept rather than dropped.

Nothing about meaning moves, and that is checkable rather than assertable,
since printing is the inverse of parsing here. Comparing `loc`, `base`,
`name` and `as` on all 40138 `<o>` nodes of the parsed tree, before against
after, leaves 13 nodes differing, every one an auto-generated cactus name
tracking its object's new column — `a🌵132-12` becomes `a🌵132-14`.
Normalise those and the two sides hash the same. Such names never enter the
visible namespace, so nothing can be written against them, but the shift is
worth knowing about.

The sources and the sheet have to travel together: the moment the canonical
form changes, `MjFormat` refuses the old text with `84 of 171 EO source(s)
are not formatted canonically`.

Stopping `build-fqns.xsl` from inserting the hops at all is the follow-up
this makes possible. Until then the compiler still puts them in; it just no
longer hides that it did.

### DIFF
--- a/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
+++ b/eo-printer/src/main/resources/org/eolang/printer/print/to-eo-tree.xsl
@@ -322,7 +322,6 @@
     the first segment past that run and "hop-rest" is that segment plus
     any trailing segments.
     -->
-    <xsl:variable name="rho-prefix" select="concat($eo:xi, '.', $eo:rho, '.')"/>
     <xsl:variable name="segments" select="tokenize(@base, '\.')"/>
     <xsl:variable name="rho-count" select="if ($segments[1] = $eo:xi) then eo:rho-run(subsequence($segments, 2)) else 0"/>
     <xsl:variable name="hop-name" select="if (count($segments) &gt; $rho-count + 1) then $segments[$rho-count + 2] else ''"/>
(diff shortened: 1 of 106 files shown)


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1255 >= 100); -8 for way too many hits-of-code (1255 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1009](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7088, opened 2026-08-19
fix(#7061): one-word parameter names in MjTranspile

gemshrine:
Part of #7061, after #7085 and #7086. The last four suppressions outside `MjSafe`.

| field | now | XML element | `-D` property |
| --- | --- | --- | --- |
| `addSourcesRoot` | `attach` | `<addSourcesRoot>` | `eo.addSourcesRoot` |
| `transpileTests` | `tests` | `<transpileTests>` | `eo.transpileTests` |
| `trackLocations` | `located` | `<trackLocations>` | `eo.trackLocations` |
| `coverageTracking` | `coverage` | `<coverageTracking>` | `eo.coverageTracking` |

`eo-runtime/pom.xml:290-292` configures `<trackLocations>` and `<coverageTracking>` as XML elements and `eo-integration-tests/src/it/fibonacci/pom.xml:45` configures `<transpileTests>`; all three keep working through `alias`.

`FakeMaven` sets `attach` and `tests` as defaults now, and `MjTranspileTest` names `coverage` and `located` in its `.with()` calls. All 55 tests of `MjTranspileTest` pass.

This touches the same two lines of `FakeMaven` as #7085 does, so whichever lands second needs a rebase; I will do it.

PR #7086, opened 2026-08-19
fix(#7061): one-word parameter names in MjPrint and MjFormat

gemshrine:
Part of #7061, after #7085. Three of the 48 suppressions, in `MjPrint` and `MjFormat`.

| file | field | now | XML element | `-D` property |
| --- | --- | --- | --- | --- |
| `MjPrint` | `printSourcesDir` | `sources` | `<printSourcesDir>` | `eo.printSourcesDir` |
| `MjPrint` | `printOutputDir` | `output` | `<printOutputDir>` | `eo.printOutputDir` |
| `MjFormat` | `autoFix` | `autofix` | `<autoFix>` | `eo.autoFix` |

`alias` carries the old name, so `-Deo.autoFix` and a `<printOutputDir>` element keep working exactly as before, the same way `MjSafe` already does it for `trackTransformationSteps`.

`MjPrintTest` and `MjFormatTest` address the fields by name through `FakeMaven.with()`, so their literals move too.

The other six suppressions in these two files sit on `penaltyIndent`, `penaltyBracket` and `penaltyExcess`, which `MjPrint` and `MjFormat` declare identically and read back through an identical `weights()` method. Renaming that block in both at once makes SonarCloud count it as duplicated new code (20.7% against a 3% gate), so it comes in its own PR that removes the duplication first.

Note: `MjPrintTest.printsXmirToEo` fails on my machine on master as well, because it reads a pack directory out of `eo-printer/src/test/resources` that is not on my local classpath. Not related to this change.

PR #7085, opened 2026-08-19
fix(#7061): one-word parameter names in three mojos

gemshrine:
Part of #7061. Three of the seven mojos, eight of the 48 suppressions. The 200-line PR limit is why the rest are not here; `MjPrint`, `MjFormat` and `MjTranspile` follow in a separate PR, and `MjSafe` needs the grouping from #6459 first.

| file | field | now | XML element | `-D` property |
| --- | --- | --- | --- | --- |
| `MjAtomsTable` | `atomsInputDir` | `sources` | `<atomsInputDir>` | `eo.atomsInputDir` |
| `MjAtomsTable` | `atomsOutput` | `csv` | `<atomsOutput>` | `eo.atomsOutput` |
| `MjCoverageReport` | `coverageFile` | `hits` | `<coverageFile>` | `eo.coverageFile` |
| `MjCoverageReport` | `lcovFile` | `lcov` | `<lcovFile>` | `eo.lcovFile` |
| `MjCoverageReport` | `minCoverage` | `minimum` | `<minCoverage>` | `eo.minCoverage` |
| `MjRegister` | `includeSources` | `included` | `<includeSources>` | none |
| `MjRegister` | `excludeSources` | `excluded` | `<excludeSources>` | none |
| `MjRegister` | `strictFileNames` | `strict` | `<strictFileNames>` | `eo.strictFileNames` |

Nothing a user writes moves: `alias` pins the element name and `property` pins the flag, the same way `MjSafe` already does for `trackTransformationSteps`, `skipExperimentalLints` and `ignoreVersionConflicts`. `eo-runtime/pom.xml:348` still configures `<minCoverage>` and keeps working through the alias.

The test literals move along with the fields, since `FakeMaven.with()` addresses them by field name: `MjAtomsTableTest`, `MjCoverageReportTest`, `MjRegisterTest`, `MjCleanTest` and the `strict` default in `FakeMaven` itself. All eight `@checkstyle MemberNameCheck` comments in these three files are gone.

The one rename that needed more than a rename is `coverageFile`: `report()` held a local `hits` list of the file's lines, which the field now claims, so the local is `records`.

The `qulice` and `xcop` checks are red on master itself right now, on `Walk.java` (`LooseCoupling`, from f463cebf9b) and on the comment formatting in `eo-maven-plugin/pom.xml` (which #7070 fixes). Neither comes from this branch.

Issue #7061, opened 2026-08-18
Unnecessary `MemberNameCheck` suppressions in seven `Mj*` mojos due to misused `@Parameter(alias)`


yegor256:
Seven mojos carry 48 `@checkstyle MemberNameCheck` comments between them: `MjSafe.java` (27), `MjPrint.java` (5), `MjTranspile.java` (4), `MjFormat.java` (4), `MjRegister.java` (3), `MjCoverageReport.java` (3) and `MjAtomsTable.java` (2), all under `eo-maven-plugin/src/main/java/org/eolang/maven/`. Every one of them sits on an `@Parameter` field named in camelCase — `printSourcesDir` at `MjPrint.java:56`, `classesDir` at `MjSafe.java:80`, `minCoverage` at `MjCoverageReport.java:70`, and so on — which qulice's `MemberName` rule, configured as `^(id|[a-z]{3,12})$`, rejects.

The usual reason to keep such a name is that Maven binds it to the user's configuration, but `@Parameter` decouples both handles: `property` already pins the `-Deo.*` flag, and `alias` pins the `<xml>` element. `MjSafe` proves it three times over. Line 194 carries `alias = "trackTransformationSteps"` on a field now called `trackSteps`, line 259 carries `alias = "skipExperimentalLints"` on `skipExperimental`, and line 291 carries `alias = "ignoreVersionConflicts"` on `ignoreConflicts`. All three were shortened without touching the configuration users write, yet all three stopped one step short of a single lowercase word and still carry the suppression.

Finish what those three started: shorten each field to one lowercase word and add `alias` with the old name wherever one is not already there. Roughly eight of `MjSafe`'s names have no free one-word form, because `foreign`, `placed`, `cache` and `skip` are taken by other fields and `placeBinaries`, `skipBinaries` and `keepBinaries` compete for the same noun; those are the ones that need the grouping already discussed in #6459, and the rest need only the rename. Note that `FakeMaven` addresses these fields by string name (`FakeMaven.java:162`, `MjCleanTest.java:39`), so the test literals move along with them, and `mojoFields` at `FakeMaven.java:476` makes a missed one fail rather than pass quietly. Same defect as #7060, which covers the non-mojo classes.

PR #7103, opened 2026-08-19
fix(#7061): one-word penalty parameter names, declared once

gemshrine:
Part of #7061, after #7085, #7086 and #7088. The last six suppressions outside `MjSafe`, on `penaltyIndent`, `penaltyBracket` and `penaltyExcess` in `MjPrint` and `MjFormat`.

These two could not simply be renamed in place. Both mojos declare the same four penalty parameters and read them back through a `weights()` method that is identical down to the last character, so touching that block in both files at once makes SonarCloud see 20.7% duplication on new code against a 3% gate (see the first run of #7086). The block has to stop being duplicated before the names can move.

So it moves into `MjPenalties`, an abstract goal that `MjPrint` and `MjFormat` now extend. Both print XMIR back to EO through `Xmir#toEO()` and the layout that comes out is decided by the same four numbers, so one declaration serves both, and neither goal can drift away from the other's idea of the layout any more. Maven injects `@Parameter` fields declared in a superclass exactly as it does for `MjSafe`'s own, and `FakeMaven.mojoFields` already walks the whole hierarchy, so the tests reach them unchanged.

| field | now | XML element | `-D` property |
| --- | --- | --- | --- |
| `penaltyIndent` | `indent` | `<penaltyIndent>` | `eo.penaltyIndent` |
| `penaltyBracket` | `bracket` | `<penaltyBracket>` | `eo.penaltyBracket` |
| `penaltyExcess` | `excess` | `<penaltyExcess>` | `eo.penaltyExcess` |

`width` travels with them, needing no rename and no alias.

`MjPrintTest.param()` maps a pack's penalty key to the mojo parameter name, so its three literals move too.

`ArchitectureTest.mojos()` used to carve `MjSafe` out of the mojo rules by name, since a base is neither public, nor annotated with `@Mojo`, nor a goal. It now carves out every abstract `Mj*` instead, which covers `MjSafe` and `MjPenalties` alike and needs no further edit the next time a base appears.

After this, every `MemberNameCheck` suppression outside `MjSafe` is gone. `MjSafe`'s own 27 need the grouping from #6459 first.

Note: `MjPrintTest.printsXmirToEo` fails on my machine on master as well, because it reads a pack directory out of `eo-printer/src/test/resources` that is not on my local classpath. The other 13 tests of that class and all 10 of `MjFormatTest` pass.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjFormat.java
@@ -11,8 +11,6 @@ import com.jcabi.xml.XML;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Collection;
-import java.util.EnumMap;
-import java.util.Map;
 import java.util.Optional;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
@@ -20,7 +18,6 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.cactoos.text.TextOf;
 import org.cactoos.text.UncheckedText;
 import org.eolang.parser.EoSyntax;
-import org.eolang.printer.PenaltyKey;
 import org.eolang.printer.Xmir;
 
 /**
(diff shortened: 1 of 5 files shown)


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (198 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2802](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7115, opened 2026-08-19
#7108: default value of includeSources via @Parameter

### LINKED ISSUE #7108: MjRegister suppresses PMD.ImmutableField instead of using @Parameter defaultValue
`MjRegister` carries a class-level `@SuppressWarnings("PMD.ImmutableField")`:

https://github.com/objectionary/eo/blob/master/eo-maven-plugin/src/main/java/org/eolang/maven/MjRegister.java

The suppression is there only because two private `@Parameter` fields are assigned in the constructor:

```java
@Parameter
private Set<String> includeSources;

@Parameter
private Set<String> excludeSources;

public MjRegister() {
    this.includeSources = new SetOf<>("**.eo");
    this.excludeSources = new SetOf<>();
}
```

PMD sees private fields that are only assigned in the ctor and suggests making them `final`, which we can't do because Maven injects into them reflectively. Hence the suppression.

The default for `includeSources` can be expressed by Maven itself, with `defaultValue` of `@Parameter`. Plexus turns a comma-separated `default-value` into a collection (`CollectionConverter.csvToXml`), so this is equivalent to the current ctor:

```java
@Parameter(defaultValue = "**.eo")
private Set<String> includeSources;
```

With the ctor gone, nothing assigns the fields inside the class any more, so `PMD.ImmutableField` stops firing and the `@SuppressWarnings` can be deleted.

One caveat for `excludeSources`: an empty `defaultValue` won't work, because `maven-plugin-plugin` only emits a parameter into the mojo descriptor's `<configuration>` when its `default-value` (or expression) is non-empty. So an empty default leaves the field as `NULL`, and `Walk.excludes(NULL)` would break. The empty case has to stay in the code, as a `NULL`-tolerant read of the field.

Also, unit tests instantiate mojos through `Moja`/`FakeMaven`, bypassing Maven's injection, so `FakeMaven` has to provide `includeSources` among the other emulated defaults (like it already does for `strictFileNames`).

yegor256:
Closes #7108.

`MjRegister` no longer carries `@SuppressWarnings("PMD.ImmutableField")`. The suppression existed only because the constructor assigned the two private `@Parameter` fields; now the default of `<includeSources>` is expressed by Maven itself and the constructor is gone, so nothing assigns those fields inside the class and PMD stops asking for `final`:

```java
@Parameter(defaultValue = "**.eo")
private Set<String> includeSources;
```

Plexus converts a comma-separated `default-value` into a collection, and `maven-plugin-plugin` does render this one into the descriptor:

```xml
<includeSources implementation="java.util.Set" default-value="**.eo"/>
```

`<excludeSources>` can't be done the same way: `maven-plugin-plugin` only puts a parameter into the mojo's `<configuration>` when its `default-value` (or expression) is non-empty, so an empty default means no injection at all and a `NULL` field, which `Walk.excludes()` would choke on. The empty case therefore stays in the code, in a new `excludes()` method, and the field's Javadoc says why.

`FakeMaven` injects mojo parameters through `Moja` instead of reading the plugin descriptor, so it now provides `includeSources` among the defaults it already emulates (`strictFileNames`, `skipZeroVersions`, and so on).

How it was checked, besides `MjRegisterTest` and `MjCleanTest`: the plugin was installed locally and the `register` goal was run against a scratch project with two `.eo` files. Without any configuration it printed `Registered 1 EO sources ..., included [**.eo], excluded []`; with `<includeSources>**.eo</includeSources>` and `<excludeSources>**/bar.eo</excludeSources>` in the POM it printed `included [**.eo], excluded [**/bar.eo]` and registered one of the two files. `mvn -Pqulice verify` on `eo-maven-plugin` passes without the suppression.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nr9MhViBkAYJ361KAKkpds)_

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjRegister.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjRegister.java
@@ -33,7 +33,6 @@ import org.cactoos.set.SetOf;
     defaultPhase = LifecyclePhase.GENERATE_SOURCES,
     threadSafe = true
 )
-@SuppressWarnings("PMD.ImmutableField")
 public final class MjRegister extends MjSafe {
 
     /**
@@ -49,14 +48,17 @@ public final class MjRegister extends MjSafe {
      *  properties since there is no way of passing it via command line.
      * @checkstyle MemberNameCheck (15 lines)
      */
-    @Parameter
+    @Parameter(defaultValue = "**.eo")
     private Set<String> includeSources;
 
     /**
      * List of exclusion GLOB filters for finding EO files
      * in the {@code <includeSources>} directory, which can be
      * pretty global (or even a root one).
-     * @checkstyle MemberNameCheck (7 lines)
+     * @implNote {@code defaultValue} attribute is omitted, because an empty
+     *  one is not rendered into the descriptor of the plugin by
+     *  the {@code maven-plugin-plugin}, thus this may stay {@code NULL}.
+     * @checkstyle MemberNameCheck (10 lines)
      */
     @Parameter
     private Set<String> excludeSources;
@@ -71,14 +73,6 @@ public final class MjRegister extends MjSafe {
         defaultValue = "true"
     )
     private boolean strictFileNames;
-
-    /**
-     * Ctor.
-     */
-    public MjRegister() {
-        this.includeSources = new SetOf<>("**.eo");
-        this.excludeSources = new SetOf<>();
-    }
 
     @Override
     public void exec() throws IOException {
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`claude/mjregister-pmd-warning-jwqugf`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`7114`". Your running score is [+2217](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (36) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2223](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7123, opened 2026-08-19
Drop `PMD.CloseResource` suppression from `Deadline`

yegor256:
`Deadline.spent()` carried `@SuppressWarnings("PMD.CloseResource")` because it ran the Mojo body in a single-thread `ExecutorService`. PMD's `CloseResource` rule flags it: `ExecutorService` is `AutoCloseable` since Java 19, so the rule sees a resource, and only a literal `close()` call satisfies it. `close()` does not exist at our Java 17 target, so the rule could not be satisfied — hence the suppression, plus a `stopped()` helper that called `shutdownNow()` and then waited a bounded 60 seconds for termination, warning when the body ignored the interrupt.

This replaces the executor with a `FutureTask` on a daemon thread of its own. There is now no resource to close, so both the suppression and the whole shutdown-and-wait helper disappear, and an uninterruptible body is abandoned where it stands instead of delaying the `finally` block that raises the `MojoFailureException`. Net: 23 lines added, 37 removed.

Behaviour is otherwise unchanged — same timeout, same `MojoFailureException`, same cause unrolling. The one difference is that a body which ignores interruption no longer costs up to 60 extra seconds and no longer holds the JVM back from exiting, since the thread is a daemon.

Verified by running PMD 7.26.0 (the version qulice 0.31.0 bundles) with qulice's own ruleset directly against both revisions: the pre-change file with the suppression stripped reports `CloseResource` at line 61, the new file reports nothing. Full `qulice:check` on `eo-maven-plugin` is green (217 files, 1014 rules across Checkstyle, PMD and ErrorProne), and `DeadlineTest` passes both cases, including `interruptsAHungBodyPromptlyOnTimeout`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WUdqLfqRyTbuSHFx4rwhtu)_

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Deadline.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Deadline.java
@@ -7,8 +7,7 @@ package org.eolang.maven;
 import com.jcabi.log.Logger;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.concurrent.FutureTask;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import org.apache.maven.plugin.MojoFailureException;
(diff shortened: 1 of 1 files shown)


0crat:
@yegor256 It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`claude/deadline-pmd-close-resource-wozzwd`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`7122`". Your running score is [+2211](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2239](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7124, opened 2026-08-19
refactor(runtime): use ReentrantLock in AtOnce instead of synchronized

yegor256:
`AtOnce.get()` guarded its lazy initialization with a `synchronized`
block and double-checked locking, which forced this suppression:

```java
@SuppressWarnings({"PMD.AvoidSynchronizedStatement", "PMD.DoubleCheckedLocking"})
```

Now the first retrieval of the origin attribute is guarded by a
`ReentrantLock` field, exactly the way `PhOnce.loaded()` already does it
in the same package, so the suppression is gone.

Behavior is unchanged: the origin attribute is still retrieved at most
once, the lock is still reentrant (so nested `get()` calls on the same
thread behave as before), and it is released in a `finally` block.

Also added `AtOnceTest.cachesAttributeInParallelThreads()`, which hits a
single `AtOnce` from 100 parallel stream tasks and asserts the nested
attribute was executed exactly once.

Verified by compiling `AtOnce` and its dependencies with `javac` and
running the four existing `AtOnceTest` scenarios plus the new parallel
one directly — all pass. The full Maven reactor build was not run here
(`eo-runtime` needs `eo-maven-plugin:1.0-SNAPSHOT` built first), so CI
is the authority on the complete suite and on qulice.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AHjiwu2YYUidZr8QQZWyVj)_

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/AtOnce.java
+++ b/eo-runtime/src/main/java/org/eolang/AtOnce.java
@@ -5,6 +5,8 @@
 package org.eolang;
 
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * Attribute that retrieves object only once.
@@ -26,12 +28,18 @@ public final class AtOnce implements Attribute {
     private final AtomicReference<Phi> cached;
 
     /**
+     * Lock guarding the first retrieval of the origin attribute.
+     */
+    private final Lock lock;
+
+    /**
      * Ctor.
      * @param attr Origin attribute
      */
     public AtOnce(final Attribute attr) {
         this.origin = attr;
         this.cached = new AtomicReference<>(null);
+        this.lock = new ReentrantLock();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


yegor256:
The `site` check is red here, but the failure is not from this PR — it's in `eo-maven-plugin`, which this diff doesn't touch (only `eo-runtime/.../AtOnce.java` and its test changed):

```
[WARNING] .../maven/MjRegister.java:36: warning: use of default constructor, which does not provide a comment
[ERROR] Failed to execute goal ...maven-site-plugin:3.22.0:site (default-site) on project eo-maven-plugin:
        Error generating maven-javadoc-plugin:3.12.0:javadoc report: Project contains Javadoc Warnings
```

It comes from master, in 294b572 (`#7108: default value of includeSources via @Parameter`), which is an ancestor of this branch's base. That commit moved the defaults into `@Parameter(defaultValue = ...)` and deleted the explicit constructor:

```diff
-    /**
-     * Ctor.
-     */
-    public MjRegister() {
-        this.includeSources = new SetOf<>("**.eo");
-        this.excludeSources = new SetOf<>();
-    }
```

`MjRegister` is now the only concrete `Mj*` mojo without one (`MjParse`, `MjPull`, etc. all keep a documented no-arg ctor), so javadoc warns about the implicit default constructor and `mvn site` fails the reactor at `eo-maven-plugin`. The `site` workflow itself only landed on master a few minutes before this PR, so this is its first run.

I'm not pushing this into this PR, since it would widen an `eo-runtime` refactor into an unrelated module. The fix on master is two lines, mirroring `MjParse`:

```java
    /**
     * Ctor.
     */
    public MjRegister() {
        // nothing
    }
```

The master `site` run for the same base commit (18f1fc3) is still queued; it should show the identical failure. Happy to push the `MjRegister` fix here instead if you'd rather have it bundled.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2226](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #2251, opened 2023-07-13
sonar CI workflow keeps failing


yegor256:
https://github.com/objectionary/eo/actions/workflows/sonar.yml

Let's find out what is the reason and fix

yegor256:
@maxonfjvipon please, take a look

yegor256:
@maxonfjvipon 

this is the recent one: https://sonarcloud.io/project/issues?resolved=false&types=BUG&sinceLeakPeriod=true&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&id=com.objectionary%3Aeo

this is the failure in CI: https://github.com/objectionary/eo/actions/runs/7721848475/job/21049050225

maybe we should add Sonar to pull request checks?

maxonfjvipon:
@c71n93 can you take it?

c71n93:
@maxonfjvipon yeah, I can try.

c71n93:
That's the output of sonar analysis:
Failed conditions
 [0.0% Coverage on New Code](https://sonarcloud.io/component_measures?id=com.objectionary%3Aeo&branch=master&metric=new_coverage&view=list) (required ≥ 80%)
 [E Reliability Rating on New Code](https://sonarcloud.io/dashboard?id=com.objectionary%3Aeo&branch=master) (required ≥ A)

Issues with reliability of code is just places in code that was identified by sonar as bug. They needs to be fixed or ignored.
Issues with code coverage is sort of problem, it doesn't work properly. I created new project in sonar with my up to date fork of `objectionary/eo` for testing. I locally ran:
```
$ mvn clean install -Pjacoco
$ mvn sonar:sonar -Dsonar.token=my-sonar-token
```
and there was the result: https://sonarcloud.io/summary/overall?id=c71n93_eo%3Aeo. Code coverage is correct here.
I suppose, that coverage data is not collected in our CI, because `jacoco` is not launched [here](https://github.com/objectionary/eo/blob/master/.github/workflows/sonar.yml) before sonar analysis. Sonar requires `jacoco.xml` to show code coverage analysis.

I was trying to test my ideas in #2874, by adding "on pull-request" check to sonar workflow. But it doesn't work, because it requires `SONAR_TOKEN` (https://github.com/objectionary/eo/actions/runs/7873343508/job/21480455544?pr=2874). 

@yegor256 How can I properly enable "on pull-request" check for this workflow? I need it to debug sonar CI locally in my PR.

[...]

PR #7125, opened 2026-08-19
Remove dead `java:S5164` suppression from `Dataized`

yegor256:
No refactoring turned out to be necessary here: the suppression is simply dead and can be deleted.

`java:S5164` is Sonar's *"ThreadLocal variables should be cleaned up when no longer used"*. `Dataized` has no `ThreadLocal` at all, so the class-level suppression silences nothing.

It is a leftover. Up to [0.42.0](https://github.com/objectionary/eo/blob/0.42.0/eo-runtime/src/main/java/org/eolang/Dataized.java) the class held two `ThreadLocal` fields used to track the dataization nesting depth for `FINE` logging:

```java
private static final ThreadLocal<Integer> LEVEL = ThreadLocal.withInitial(() -> 0);
private static final ThreadLocal<Integer> MAX_LEVEL =
    ThreadLocal.withInitial(() -> Integer.getInteger("max.dataization.log", 3));
```

Both carried `@todo #2251` notes about calling `ThreadLocal#remove()`, which is exactly what the suppression was for. The fields and the logging they served were later removed, and `take()` is now a one-liner delegating to `this.phi.delta()` — but the annotation stayed behind, where it now only sends readers hunting for a `ThreadLocal` that isn't there.

The runtime's one remaining `ThreadLocal` is `PhDefault.NESTING`, which keeps its own suppression placed narrowly on the field itself. That one is left untouched.

### Changes

- `eo-runtime/src/main/java/org/eolang/Dataized.java`: dropped `@SuppressWarnings("java:S5164")` from the class declaration. One line removed, nothing else.

### Verification

`@SuppressWarnings` has `SOURCE` retention, so removing an inapplicable one cannot change behaviour — no bytecode difference beyond the annotation itself, and no test could observe it. Confirmed the file still compiles clean:

```
javac -Xlint:all -sourcepath eo-runtime/src/main/java .../Dataized.java
```

No errors and no warnings, so nothing was relying on the suppression. Grepped the repo to confirm `S5164` appears nowhere else except `PhDefault`. The full Maven build was not run locally (the sandbox has no populated `~/.m2` and is offline), so CI is the real check for `qulice`.

---
_Generated by [Claude Code](https://claude.ai/code/session_0139rKYiuyxPyoPjd1FDPwPR)_

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Dataized.java
+++ b/eo-runtime/src/main/java/org/eolang/Dataized.java
@@ -20,7 +20,6 @@ import java.util.Arrays;
  * @see <a href="https://arxiv.org/abs/2111.13384">Canonical explanation of the Dataization concept</a>
  * @since 0.1
  */
-@SuppressWarnings("java:S5164")
 public final class Dataized {
 
     /**


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (1) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2230](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7126, opened 2026-08-19
style(runtime): copy the entries Attrs is given, drop the PMD suppression

yegor256:
`Attrs` carried `@SuppressWarnings("PMD.ArrayIsStoredDirectly")` on its constructor, which silenced the warning instead of answering it: the varargs array was kept as-is, so a caller holding a reference to that array could still change the map's contents after handing it over.

The constructor now clones the array — the same idiom `BytesRaw`, `Quoted` and `Hex` already use — so the suppression is gone and the entries genuinely belong to the map. The clone is shallow and the class stays lazy: the `LinkedHashMap` is still built on first access, and the constructor still does nothing but one field assignment.

`AttrsTest` pins the three things the class promises: it keeps the order it was given, it counts every entry, and a later change to the caller's array does not reach it.

Verified locally by compiling `Attrs` and the new test (`javac -Xlint:all`, no warnings) and running `AttrsTest` on the JUnit 5 console launcher — 3 tests, all green. A full `mvn -Pqulice` run was not possible here (empty local repository, whole-reactor build), so CI is the check on the rest.

---
_Generated by [Claude Code](https://claude.ai/code/session_017LNNd2LLuejpyB25rYy8R1)_

### DIFF
--- a/eo-runtime/src/main/java/org/eolang/Attrs.java
+++ b/eo-runtime/src/main/java/org/eolang/Attrs.java
@@ -19,15 +19,17 @@ import java.util.Set;
  * {@link AbstractMap.SimpleEntry} (a constructor invocation, not a method
  * call) being acceptable as a {@code Map.Entry} factory.</p>
  *
- * <p>The underlying {@link LinkedHashMap} is built lazily on first access so the
- * constructor itself remains free of method calls.</p>
+ * <p>The underlying {@link LinkedHashMap} is built lazily on first access, so
+ * the constructor does nothing but copy the entries it was given. The copy is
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2238](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7127, opened 2026-08-19
#7063: key the Catalogs cache on path and format together

### LINKED ISSUE #7063: Caching issue due to usage of absolute path in Catalogs.java
The puzzle `7023-fbecf3d2` from #7023 has to be resolved: 

https://github.com/objectionary/eo/blob/c0c2f6dfef7285db90208d868578ce3854f30752/eo-maven-plugin/src/main/java/org/eolang/maven/Catalogs.java#L83-L89

The puzzle was created by @yegor256 on 18-Aug-26. 

morphqdd:
Closes #7063.

Resolves the puzzle `7023-fbecf3d2`: `Catalogs.make(Path, String)` cached on the absolute path alone, so a second caller asking for the same path with a different format silently got the first caller's mono instead of its own.

Added a second map tracking the format each path was first cached with. A repeat call with the same format is a no-op as before; a repeat call with a different format now throws `IllegalStateException` naming both formats, instead of silently serving the wrong one.

Added two tests: same-path-same-format still succeeds, same-path-different-format throws.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/Catalogs.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/Catalogs.java
@@ -58,10 +58,16 @@ final class Catalogs {
     private final Map<Path, Tojos> all;
 
     /**
+     * The format each cached path was first built with.
+     */
+    private final Map<Path, String> formats;
+
+    /**
      * Ctor.
      */
     private Catalogs() {
         this.all = new ConcurrentHashMap<>(0);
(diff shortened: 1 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5303](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7128, opened 2026-08-19
fix(#7102): let directory.deleted survive the root vanishing mid-walk

### LINKED ISSUE #7102: `directory.deleted` throws exception when root directory is concurrently deleted
## What happens

`directory.deleted` is not idempotent when its root directory disappears concurrently.  It first observes that `d` exists and then immediately evaluates `d.walk "**"`.  If another process removes `d` in between, the JVM implementation of `directory.listed` calls `Files.list` on an absent path and propagates an exception.

The requested end state is nevertheless already satisfied: the directory is gone.

## Steps to reproduce

1. Process A begins dataizing `some-directory.deleted`; `some-directory.exists` returns TRUE.
2. Before A evaluates `some-directory.walk "**"`, process B removes `some-directory`.
3. A reaches `directory.listed`, where [`EOdirectory$EOlisted`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/java/org/eolang/EOdirectory%24EOlisted.java#L48-L59) executes `Files.list(path)`, and fails with a no-such-file error.

In contrast, if B removes the directory just before the initial existence probe, `directory.deleted` returns normally.  Thus an equivalent final filesystem state has two different outcomes depending only on timing.

## Why it happens

[`directory/deleted.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/directory/deleted.eo#L17-L28) separates the existence test from the recursive walk:

```eo
d.exists.if
  seq *
    ...
    | (d.walk "**").at.^
    d
  d
```

There is no recovery/recheck around `d.walk`, although the false branch establishes that an absent directory is a successful no-op.

## Expected behaviour

`directory.deleted` should be idempotent under concurrent deletion.  If the root disappears before traversal starts, it should return `d`, matching the behaviour when the directory was already absent at entry.  Real traversal failures (permissions, I/O errors, etc.) must still be reported.

## Suggested fix

Wrap the walk failure narrowly enough to re-check whether `d` is now absent.  Return `d` only for that case and preserve other errors.  Add a synchronization-based regression test that removes the root after the initial `exists` check and before `listed`/`walk` begins.

Thayorns:
`directory.deleted` checked `d.exists`, then called `d.walk "**"`. If another process removed `d` in that window, `Files.list` threw and the exception propagated, even though the requested end state (the directory is gone) was already satisfied.

The upfront `d.exists.if` gate is now gone, and the walk call is replaced by a new nested atom, `directory.deleted.entries` (`EOdirectory$EOdeleted$EOentries.java`). It walks the tree and, only on failure, rechecks `d.exists`: if the directory is gone it returns an empty tuple (nothing left to delete), otherwise it rethrows the original failure untouched. `directory.walk`/`directory.listed` themselves are untouched, so they keep failing when asked to walk a directory that was never there, matching the existing `wrapsMissingDirectoryIntoExFailure` and `stops-on-walking-an-absent-directory` tests.

Added `EOdirectoryEOdeletedEOentriesTest.java`, a deterministic Java-level regression test that removes the root right before the walk is forced, plus a happy-path test, and an EO-level `can-delete-an-already-absent-directory` test.

Closes #7102


Thayorns:
@yegor256 happy to clarify the recovery approach in EOdirectory$EOdeleted$EOentries if anything looks off.

eo-runtime/src/main/eo/directory/deleted.eo
```diff
@@ -4,23 +4,24 @@
 +architect yegor256@gmail.com
 +home https://github.com/objectionary/eo
 +package directory
++rt jvm org.eolang:eo-runtime:0.0.0
```
yegor256: @Thayorns we don't want any more atoms in our code base, try to implement your feature in pure EO

Thayorns: @yegor256 done — dropped the atom and reimplemented `directory.deleted.entries` in pure EO using the existing `recovered` primitive as the catch/recover mechanism instead of a Java try/catch.

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (27) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2241](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (27) hits-of-code; +1 to give you at least something.  Your running score is [+2142](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7129, opened 2026-08-19
#7075: path.dirname returns empty string for an empty path

### LINKED ISSUE #7075: `path.dirname` incorrectly returns empty string for an empty path instead of `"."`
## What happens

`path.dirname` answers `"."` for every path that has no separator, except the empty one, where it answers `""`.

Verified against a fresh `eo-runtime` build by dataizing the real `EOpath`:

| call | result | expected |
|---|---|---|
| `(path "foo").dirname` | `"."` | `"."` (ok) |
| `(path "").dirname` | **`""`** | `"."` |
| `(path "").normalized` | `"."` | `"."` (ok) |
| `(path "").basename` | `""` | `""` (ok) |

So the same object, asked about the same empty path, says its normalized form is `"."` but its directory is `""`.

`dirname("")` is `"."` in POSIX `dirname(3)` ("If the string is empty, `dirname()` shall return a pointer to the string `.`"), in `node:path` (`path.dirname("")` is `"."`) and in `java.io.File` semantics, so `""` is not a convention anyone else follows either.

The empty answer is also unusable in the one way `dirname` is normally used: joining it back with the basename. `"." + separator + basename` names the file, `"" + separator + basename` names an absolute path in the root.

## Root cause

`eo-runtime/src/main/eo/path.eo:36`:

```
    [] > dirname
      string > @
        if.
          pth.size.eq 0
          trimmed >> pth!
          if.
            len.eq -1
            "."
            ...
```

The first branch short-circuits an empty path to `trimmed`, which for an empty `uri` is the empty string again. It never reaches the `len.eq -1` branch below, which is exactly the "no separator anywhere, so the directory is the current one" case that would have answered `"."`.

`basename` has the same short-circuit one binding above, and there it is correct: the basename of an empty path is empty.

## Suggested fix

Let the empty path fall through to the branch that already handles "no separator":

```
    [] > dirname
      string > @
        if.
          len.eq -1
          "."
          if.
            len.eq 0
            separator
            as-bytes.
              text.slice
                0
                text.last-index-of separator >> len!
      string trimmed > text
```

With an empty `text`, `last-index-of` is `-1`, so the first branch answers `"."` without any special case. The `pth.size.eq 0` guard then only remains needed in `basename`.

Regression example worth adding:

```
  eq. ++> can-return-the-current-dirname-of-an-empty-path
    dirname.
      path ""
    "."
```

morphqdd:
Fixes #7075.

`path.dirname` short-circuited an empty path to `trimmed` (empty again), so it returned `""` instead of falling through to the branch that already handles "no separator anywhere" and answers `"."`. `basename` has the same short-circuit and is correct there, since the basename of an empty path really is empty; `dirname` doesn't need it, `pth`/`text` becoming an empty string already lands on `last-index-of` returning `-1`.

Removed the short-circuit so an empty path falls through to the existing `len.eq -1` branch, matching POSIX `dirname(3)`, `node:path.dirname("")` and `java.io.File`, all of which answer `"."` for an empty path.

Added `can-return-the-current-dirname-of-an-empty-path`.

### DIFF
--- a/eo-runtime/src/main/eo/path.eo
+++ b/eo-runtime/src/main/eo/path.eo
@@ -36,19 +36,16 @@
     [] > dirname
       string > @
         if.
-          pth.size.eq 0
-          ^.trimmed >> pth!
+          len.eq -1
+          "."
           if.
-            len.eq -1
-            "."
-            if.
-              len.eq 0
-              ^.separator
-              as-bytes.
-                text.slice
-                  0
-                  text.last-index-of ^.separator >> len!
-      string pth > text
+            len.eq 0
+            ^.separator
+            as-bytes.
+              text.slice
+                0
+                text.last-index-of ^.separator >> len!
+      string ^.trimmed > text
     sys.drive uri > drive
     if. > driveless
       drive.size.eq 0
@@ -535,6 +532,11 @@
   eq. ++> can-return-the-current-dirname-of-a-relative-posix-file-ending-with-a-separator
     dirname.
       path.posix "foo/"
+    "."
+
+  eq. ++> can-return-the-current-dirname-of-an-empty-path
+    dirname.
+      path ""
     "."
 
   eq. ++> can-return-the-dirname-ignoring-repeated-trailing-separators


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5315](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7130, opened 2026-08-19
#7073: uri splits authority at first @, leaking userinfo into host

### LINKED ISSUE #7073: `uri` incorrectly splits authority at first `@` causing userinfo to leak into host
## What happens

`uri` splits the authority at the **first** `@`, so when the userinfo itself contains an `@` the leftover lands in the host. The host it reports is then a string no host can be: it still carries an `@`.

Verified against a fresh `eo-runtime` build by dataizing the real `EOuri`:

| call | result | expected |
|---|---|---|
| `(uri "http://a@b@host/p").userinfo` | `"a"` | `"a@b"` |
| `(uri "http://a@b@host/p").host` | **`"b@host"`** | `"host"` |
| `(uri "http://user:pa@ss@host:80/p").host` | **`"ss@host"`** | `"host"` |
| `(uri "http://user:pa@ss@host:80/p").port` | `"80"` | `"80"` (ok) |
| `(uri "http://host:80@evil.com/p").host` | `"evil.com"` | `"evil.com"` (ok, single `@`) |

The single-`@` cases are all fine, so the object works until a second `@` shows up.

## Why the last `@` is the right one

RFC 3986 §3.2 gives the authority as `[ userinfo "@" ] host [ ":" port ]`, and §3.2.2 lets a `reg-name` host contain only `unreserved / pct-encoded / sub-delims`. `@` is in neither set, so a host can never hold an unescaped `@`, while `userinfo` may (§3.2.1 allows `:` and, through `pct-encoded`, anything else that is written literally by careless callers). Whatever sits before the *last* `@` is therefore the userinfo, and only what follows it can be the host, which is also what `curl` and the WHATWG URL parser do.

`java.net.URI` refuses to guess at all here: for `http://a@b@host/p` it parses the authority as registry-based and leaves `getHost()` and `getUserInfo()` `null`. Returning `"b@host"` as a host is worse than either answer, because a caller that feeds it to `socket` or `file` gets a name that cannot resolve, with no signal that the parse went wrong.

## Root cause

`eo-runtime/src/main/eo/uri.eo:76`, inside the `host-port` binding:

```
      string >> host-port
        has-userinfo.if >> host-port-bytes!
          authority.slice
            plus.
              authority.index-of "@" >> userinfo-at-idx
              1
            authority.length.minus
              number
                userinfo-at-idx.plus 1
          authority
```

`userinfo-at-idx` is bound to `authority.index-of "@"` (the first `@`), and `userinfo` at line 156 reuses that same index, so both halves of the split move together and both are wrong.

## Suggested fix

Take the last `@` instead:

```
              authority.last-index-of "@" >> userinfo-at-idx
```

That is a one-token change, `userinfo` at line 156 needs no edit since it reads the same binding, and the existing `can-parse-a-host-alongside-userinfo` / `can-parse-a-userinfo-without-a-password` tests keep passing because a single `@` is both the first and the last one. Worth adding two `++>` examples alongside them:

```
  eq. ++> can-parse-a-host-past-an-at-sign-inside-userinfo
    host.
      uri "http://a@b@host/p"
    "host"

  eq. ++> can-parse-a-userinfo-holding-an-at-sign
    userinfo.
      uri "http://a@b@host/p"
    "a@b"
```

morphqdd:
Fixes #7073.

`uri`'s `host-port` binding split the authority at the first `@`, so when the userinfo itself contains an `@` the leftover part landed in the host, producing a host string that still carries an `@` and cannot resolve as a host.

RFC 3986 forbids `@` in `reg-name` but allows it in `userinfo` (via pct-encoded, though callers sometimes write it literally), so whatever sits before the last `@` is the userinfo, which is also what curl and the WHATWG URL parser do.

One-token fix: `authority.index-of "@"` to `authority.last-index-of "@"`. `userinfo` reuses the same binding, so it needed no separate change.

Added `can-parse-a-host-past-an-at-sign-inside-userinfo` and `can-parse-a-userinfo-holding-an-at-sign`.

### DIFF
--- a/eo-runtime/src/main/eo/uri.eo
+++ b/eo-runtime/src/main/eo/uri.eo
@@ -73,7 +73,7 @@
         has-userinfo.if >> host-port-bytes!
           authority.slice
             plus.
-              authority.index-of "@" >> userinfo-at-idx
+              authority.last-index-of "@" >> userinfo-at-idx
               1
             authority.length.minus
               number
@@ -184,6 +184,11 @@
       uri "http://user:pass@host:80/some/path"
     "host"
 
+  eq. ++> can-parse-a-host-past-an-at-sign-inside-userinfo
+    host.
+      uri "http://a@b@host/p"
+    "host"
+
   eq. ++> can-parse-a-path-alongside-a-userinfo-authority
     path.
       uri "http://user:pass@host:80/some/path"
@@ -213,6 +218,11 @@
     scheme.
       uri "HTTP://host/path"
     "HTTP"
+
+  eq. ++> can-parse-a-userinfo-holding-an-at-sign
+    userinfo.
+      uri "http://a@b@host/p"
+    "a@b"
 
   eq. ++> can-parse-a-userinfo-without-a-password
     userinfo.


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5327](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7132, opened 2026-08-19
#7131: add explicit ctor to MjRegister, fixes site job

### LINKED ISSUE #7131: MjRegister.java lacks an explicit ctor, breaking the site job's javadoc check
`eo-maven-plugin/src/main/java/org/eolang/maven/MjRegister.java:36` declares `public final class MjRegister extends MjSafe` with no explicit constructor, so Java synthesizes a default one. `maven-javadoc-plugin` flags that as "use of default constructor, which does not provide a comment", and the `site` job's javadoc report fails the build on that warning.

Every sibling mojo (`MjResolve`, `MjCompile`, `MjParse`, etc.) already has an explicit no-arg constructor with a `Ctor.` javadoc comment, e.g. `MjResolve.java:48-53`:

```java
/**
 * Ctor.
 */
public MjResolve() {
    // nothing
}
```

`MjRegister` is the one mojo missing this, confirmed on master commit 18f1fc36aa, CI job `site`:

```
[WARNING] /eo-maven-plugin/src/main/java/org/eolang/maven/MjRegister.java:36: warning: use of default constructor, which does not provide a comment
```

What should happen: `MjRegister` should have the same explicit, commented no-arg constructor as its sibling mojos, so the javadoc warning (and the `site` job) goes away.

morphqdd:
Fixes #7131.

`MjRegister` had no explicit constructor, so `maven-javadoc-plugin` flagged the synthesized default one as "use of default constructor, which does not provide a comment", which fails the `site` job's javadoc report and blocks CI on master right now.

Added the same explicit, commented no-arg constructor every sibling mojo (`MjResolve`, `MjCompile`, `MjParse`, etc.) already has. Verified the javadoc warning is gone by running the `javadoc:javadoc` goal directly.

### DIFF
--- a/eo-maven-plugin/src/main/java/org/eolang/maven/MjRegister.java
+++ b/eo-maven-plugin/src/main/java/org/eolang/maven/MjRegister.java
@@ -41,6 +41,13 @@ public final class MjRegister extends MjSafe {
     private static final Pattern PATTERN = Pattern.compile("^[a-zA-Z0-9\\-]+\\.eo$");
 
     /**
+     * Ctor.
+     */
+    public MjRegister() {
+        // nothing
+    }
+
+    /**
      * List of inclusion GLOB filters for finding EO files
      * in the {@code <includeSources>} directory, which can be
      * pretty global (or even a root one).


morphqdd:
@yegor256, please take a look when you have a moment. This blocks the new site CI job on every PR against master right now.

PR #7133, opened 2026-08-19
name both +> and -> as allowed atom test attributes

### LINKED ISSUE #7122: Incorrect error message for atom body in eo-parser only allows `+>`, neglects `->`
The error thrown at `eo-parser/src/main/java/org/eolang/parser/Transition.java:91`, and its copy at `eo-parser/src/main/java/org/eolang/parser/LnFormation.java:196`, reads "Atom cannot contain inner objects; only `+>` test attributes are allowed in an atom body". The flag it guards says otherwise. Every call site builds the admission from `suffix.test()` — `LnApplication.java:243`, `LnReversed.java:154`, `LnCompactTuple.java:108`, `LnOnlyPhi.java:287`, `LnPipe.java:86`, `LnTextBlock.java:102` — and `Suffix.test()` at `eo-parser/src/main/java/org/eolang/parser/Suffix.java:238` returns true for `Form.TEST` and `Form.THROWS` alike, so a throwing test attribute `-> name` is accepted under an atom exactly like `+> name`.

So the message names one of the two forms the parser actually permits. Someone reading it while writing a throwing test attribute in an atom body is told that what they wrote is forbidden, and the parser then accepts it, which leaves it unclear which of the two is the bug.

Assuming the acceptance is the intended behaviour, the message should name both markers, something in the shape of "only `+>` and `->` test attributes are allowed in an atom body". The typo fixtures asserting the old wording — `not-empty-atoms.yaml`, `not-empty-atoms-application.yaml`, `not-empty-atoms-compact-tuple.yaml` and `not-empty-atoms-reversed.yaml` under `eo-parser/src/test/resources/org/eolang/parser/eo-typos/` — need the same edit.

@yegor256 please take a look

Thayorns:
The error thrown when an atom body holds something other than a test attribute says only \`+>\` is allowed, but \`Suffix.test()\` accepts \`Form.TEST\` and \`Form.THROWS\` alike, so a throwing test attribute (\`-> name\`) is admitted under an atom exactly like \`+> name\`. The message named one of the two forms the parser actually permits, which reads as a contradiction to anyone writing \`-> name\` in an atom body.

This updates the message in \`Transition.java\` and its copy in \`LnFormation.java\` to name both markers, and updates the two typo fixtures that assert the old wording.

Closes #7122

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnFormation.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnFormation.java
@@ -193,7 +193,7 @@ final class LnFormation implements Line {
         if (!stack.empty() && stack.top().atom() && !suffix.test()) {
             throw new ParseError(
                 this.span.line(), this.span.indent(),
-                "Atom cannot contain inner objects; only `+>` test attributes are allowed in an atom body"
+                "Atom cannot contain inner objects; only `+>` and `->` test attributes are allowed in an atom body"
             );
         }
     }
--- a/eo-parser/src/main/java/org/eolang/parser/Transition.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Transition.java
@@ -88,7 +88,7 @@ final class Transition {
         if (!this.stack.empty() && this.stack.top().atom() && !admission.permitted()) {
             throw new ParseError(
                 this.span.line(), this.span.indent(),
-                "Atom cannot contain inner objects; only `+>` test attributes are allowed in an atom body"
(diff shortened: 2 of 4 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2177](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7135, opened 2026-08-19
route LnFormation through Transition

### LINKED ISSUE #7121: `LnFormation` duplicates `Transition` prologue causing maintenance inefficiencies
`Transition` calls itself "the single source of truth for that prologue" in its own docblock at `eo-parser/src/main/java/org/eolang/parser/Transition.java:17`, and every line shape but one goes through it: `LnApplication`, `LnReversed`, `LnCompactTuple`, `LnOnlyPhi`, `LnPipe`, `LnTextBlock`, `LnVoid`. `LnFormation` does not. Its `transition` method at `eo-parser/src/main/java/org/eolang/parser/LnFormation.java:149` reimplements the push-or-replace decision, and `checkChildAllowed` at line 178 repeats the three guards of `Transition.pushed` one by one — the indent step, the parent's openness, the atom-body restriction — down to the same two error strings, `LnFormation.java:196` against `Transition.java:91`.

The two copies have already drifted apart once. #6628 was exactly the atom-body guard existing in `LnFormation` and missing everywhere else, and the fix moved the check into `Transition` for the other seven shapes while leaving the original copy in place. Any correction to this prologue now has to be written twice, and a fix applied to one copy is invisible from the other.

Routing `LnFormation.transition` through `new Transition(stack, this.span).apply(Kind.BARE_FORMATION, Openness.OPEN, ...)` and deleting `checkChildAllowed` would leave one implementation; the formation-specific work that follows, marking the level named and marking it an atom, stays where it is.

@yegor256 please take a look

Thayorns:
LnFormation.transition reimplemented the same push-or-replace prologue that Transition.apply already owns for every other line shape (LnApplication, LnReversed, LnCompactTuple, LnOnlyPhi, LnPipe, LnTextBlock, LnVoid). Its checkChildAllowed method repeated the three guards of Transition.pushed one by one, down to the same two error strings.

This routes LnFormation.transition through new Transition(stack, this.span).apply(Kind.BARE_FORMATION, Openness.OPEN, admission) and deletes checkChildAllowed, leaving Transition as the single implementation of the prologue. The formation-specific work, marking the atom flag, stays in LnFormation.

Closes #7121


Thayorns:
@yegor256 happy to clarify anything about this change if needed.

yegor256:
@Thayorns no test?

Thayorns:
@yegor256 added tests

Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2433](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7136, opened 2026-08-19
Check the xmir.xsd property before the relative XSD guesses

### LINKED ISSUE #7112: `xmir.xsd` property being checked last overrides explicit configuration
The candidate list in `eo-parser/src/main/java/org/eolang/parser/DrProgram.java:82` puts `System.getProperty("xmir.xsd", "")` last, after four relative guesses — `XMIR.xsd`, `src/main/resources/XMIR.xsd`, `eo-parser/src/main/resources/XMIR.xsd` and `../eo-parser/src/main/resources/XMIR.xsd`. The loop on line 89 breaks on the first candidate that exists, so the property is only consulted when all four guesses miss.

The one place that sets it is `eo-runtime/pom.xml:381`, in the failsafe `snippet-test` execution, where the working directory is the module basedir. The fourth guess resolves there, so the explicit configuration is stepped over every run. It happens to name the same file, which is why nothing looks broken, but the setting does nothing and a module that pointed it at a different schema would be ignored without a word.

An explicit override is the most specific answer available and belongs first in that array, ahead of the guesses; the guesses are there precisely because nobody told the code where the file is. Once it is first, either the property in `eo-runtime/pom.xml` becomes meaningful or it can go.

@yegor256

Thayorns:
Closes #7112

DrProgram.schema() picked the XSD location by trying four relative-path guesses first and only fell back to the `xmir.xsd` system property when all four missed. In eo-runtime's snippet-test execution the fourth guess happens to resolve, so the explicit property was silently ignored — it just went unnoticed because the guess pointed at the same file.

The property is now the first candidate, since an explicit override is more specific than any guess. The guesses still run afterward for the case where nothing was set.

Added a test that points `xmir.xsd` at a file the relative guesses would not find, confirming the schema location resolves to that file rather than to `src/main/resources/XMIR.xsd`.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/DrProgram.java
+++ b/eo-parser/src/main/java/org/eolang/parser/DrProgram.java
@@ -80,11 +80,11 @@ final class DrProgram implements Iterable<Directive> {
             Manifests.read("EO-Version")
         );
         final String[] opts = {
+            System.getProperty("xmir.xsd", ""),
             "XMIR.xsd",
             "src/main/resources/XMIR.xsd",
             "eo-parser/src/main/resources/XMIR.xsd",
             "../eo-parser/src/main/resources/XMIR.xsd",
-            System.getProperty("xmir.xsd", ""),
         };
         for (final String opt : opts) {
             if (opt.isEmpty()) {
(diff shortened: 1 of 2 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2193](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7137, opened 2026-08-19
fix(#7120): check atom-body admission on both push and replace branches

### LINKED ISSUE #7120: `Transition.apply` skips atom-body check for subsequent children in an atom
The `apply` method in `eo-parser/src/main/java/org/eolang/parser/Transition.java:53` runs the atom-body guard of R-3.10.13 only when it steps deeper: the deeper-indent branch goes through `pushed`, which throws at line 88 when the parent level is an atom and the line is not a test attribute, while the same-indent branch at line 58 calls `stack.replace` and checks nothing at all. Since `Eo.dispatch` at `eo-parser/src/main/java/org/eolang/parser/Eo.java:408` pops only the levels strictly deeper than the new line, the first child of an atom pushes and every sibling after it replaces, so only the first one is ever examined.

That makes the rule depend on the position of the offending line rather than on the line itself. This source is rejected today, by the fixture `eo-parser/src/test/resources/org/eolang/parser/eo-typos/not-empty-atoms-application.yaml`:

```
[] > test /bytes
  42 > bar
```

while this one parses clean:

```
[] > test /bytes
  +> one

  42 > bar
```

Nothing catches it later. `Level.atom()` is read in exactly two places, `Transition.java:88` and `LnFormation.java:193`, and both sit on a push path; the closer hook that `Stack.java:314` advertises as carrying the R-5.3.4 and R-5.3.5 atom-body checks does not look at the flag, so `42 > bar` reaches the compiler as a legitimate child of an atom.

The check belongs on both branches of `apply`, not inside `pushed` — after the transition the fresh level already knows its parent through `Level.patom()`, so testing that against `admission.permitted()` covers the replace case with the same message and the same position. A yaml typo fixture with the offending line in second position would lock it down. The same hole exists in the copy of this logic in `LnFormation.checkChildAllowed`.

@yegor256 please take a look

Thayorns:
Closes #7120

`Transition.apply` ran the atom-body admission check (R-3.10.13) only inside the deeper-indent push branch (`pushed()`). The same-indent branch, which calls `Stack.replace` directly, checked nothing. Since only the first child of an atom ever goes through push — every sibling after it replaces — the rule ended up depending on a line's position rather than on the line itself: `42 > bar` as the *second* child of an atom parsed clean while the same line as the *first* child was correctly rejected.

The check now runs in `apply`, after the transition, using the fresh level's `patom()` against the admission's `permitted()`. That covers both branches with the same message and position, so `pushed()` goes back to validating only the indent step and openness.

`LnFormation.checkChildAllowed` carried an identical copy of the same logic with the same hole (only invoked from its own push branch). Moved its atom-body check the same way, into `transition()` after push-or-replace, so it now applies to both branches too.

Added `not-empty-atoms-application-second-child.yaml` to `eo-typos`, reproducing the second-child scenario from the issue: a legit `++>` test-attribute child followed by a same-indent non-test child under an atom.


Thayorns:
@yegor256 happy to clarify anything here if needed.

Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2353](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7138, opened 2026-08-19
Measure real parsing time instead of hardcoded ms=0

### LINKED ISSUE #7111: `ms="0"` value in XMIR parser output results in false parsing-time reports
The `<object>` element of every parsed program carries `ms="0"`, because `eo-parser/src/main/java/org/eolang/parser/EoSyntax.java:110` writes the literal `.attr("ms", 0L)` instead of an elapsed time. The schema still promises otherwise: `XMIR.xsd:453` documents the attribute as "the number of milliseconds that were spent by the parser to generate this document", "valuable for debugging purposes".

It used to hold a real number. Before the parser rewrite, `XeEoListener.java:89` set it to `(System.nanoTime() - this.start) / (1000L * 1000L)`, and commit `769675b1af` replaced that whole path with the constant. Nothing reads the attribute today, which is exactly the problem: anyone who profiles parsing from the XMIR it produces sees zero for every file, including the slow ones.

The regression is also pinned down by a test. `eo-parser/src/test/resources/org/eolang/parser/eo-syntax/replaced-only-phi-rollback-keeps-ms.yaml:7` asserts `/object[@ms=0]`, so restoring the measurement makes that pack fail.

Timing the walk in `parsed()` — take the clock before the `Xembler` runs and write the difference into the attribute — brings the value back; the yaml pack should then assert `/object[@ms]` instead of a fixed zero, since it cares about where the attribute lands, not what it holds.

@yegor256

Thayorns:
EoSyntax#parsed() has always written a literal attr("ms", 0L) into the root <object> element instead of the elapsed parsing time, ever since commit 769675b1af replaced the old XeEoListener timing logic. The XMIR schema still documents `ms` as "the number of milliseconds that were spent by the parser to generate this document," but every produced document reports zero, which makes it useless for profiling.

This takes System.nanoTime() at the start of parsed() and writes the elapsed milliseconds into the ms attribute right after the parser directives are built, restoring the pre-rewrite behavior. The regression-pinning yaml pack (replaced-only-phi-rollback-keeps-ms.yaml) now asserts /object[@ms] rather than a fixed zero, matching what the other ms-related packs already check, and EoSyntaxTest gained tests that the value is a real positive elapsed time, stays within a sane bound, and is measured independently on each call rather than cached or hardcoded.

Closes #7111


eo-parser/src/test/java/fixtures/LargeProgram.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.stream.Stream;
+import org.cactoos.Input;
+import org.cactoos.io.InputOf;
+
+/**
+ * EO program of many nested objects, long enough for its parsing
+ * to last more than a single millisecond.
+ * @since 0.61.0
+ */
+@SuppressWarnings({"JTCOP.RuleAllTestsHaveProductionClass", "JTCOP.RuleCorrectTestName"})
```
yegor256: @Thayorns don't suppress JTCOP, do what it wants you to do

Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2369](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2149](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7139, opened 2026-08-19
#7074: as-fixed aborts instead of recovering on a non-finite number

### LINKED ISSUE #7074: `number.as-fixed` and `printf`'s `%f` fail to handle non-finite numbers properly
## What happens

`number.as-fixed` declares a `cant-print` recovery and uses it for every `places` it rejects, but a non-finite **receiver** walks straight past it. Instead of the fallback the caller gets an `ExFailure` from `as-decimal`, three objects further down, wrapped in a 15-frame `Error in "…"` chain.

Verified against a fresh `eo-runtime` build by dataizing the real atoms:

| call | result |
|---|---|
| `2.5.as-fixed 4` | `"2.5000"` (ok) |
| `(-0.0).as-fixed 2` | `"-0.00"` (ok) |
| `1.25.as-fixed pinf "recovered"` | `"recovered"` (ok, fallback used, see `can-recover-from-infinite-places`) |
| `nan.as-fixed 2` | **ExFailure**, 15 frames, ends with `Can't write a non-finite number as decimal` |
| `pinf.as-fixed 2` | **ExFailure**, same |
| `ninf.as-fixed 2` | **ExFailure**, same |

`string.printf` inherits it, which is where a user meets this without ever naming `as-fixed`:

| call | result | expected |
|---|---|---|
| `"%f".printf * nan` | **ExFailure** ending in `Can't write a non-finite number as decimal` | `"nan"` |
| `"%f".printf * pinf` | **ExFailure**, same | `"infinity"` |

`java.lang.String.format("%f", Double.NaN)` gives `"NaN"` and C's `printf` gives `"nan"`; neither aborts.

## Root cause

`eo-runtime/src/main/eo/number/as-fixed.eo:20` guards only `places`:

```
    or.
      0.gt places
      places.is-integer.not
    cant-print
      "Can't write a fixed-point decimal with %f fraction places".printf
```

Nothing guards `n`. The value path reaches `unit.as-decimal` at line 36, and `as-decimal` aborts on a non-finite number by design (`eo-runtime/src/main/eo/number/as-decimal.eo:19`, added by #5973). The abort then travels up as a raw failure, because `as-fixed` never asks whether it can write `n` at all.

This is the half of #5973 that was left behind. That issue asked for a clean error in `as-decimal` **"or its own `cant-print` fallback"** in `as-fixed`; only the first half landed.

`printf` is the clearest victim because it already hands `as-fixed` a fallback and expects it to be honoured. `eo-runtime/src/main/eo/string/printf.eo:208`:

```
        if.
          66-.eq conv
          as-fixed.
            number numeric
            precise.as-bool.if
              number precision
              6
            failure
```

`failure` is `T message`, so the whole point of passing it is to turn a rejection into one readable line. With `nan` the fallback is never consulted. The file even records the workaround its author had to make elsewhere (`printf.eo:41`): the `%d` guard names the number it rejects through `as-scientific` "because ... a reason rendered with `%f` destroyed itself instead of arriving".

Note that `as-scientific`, the sibling that formats through `as-fixed`, already names non-finite numbers itself and returns `"nan"`, `"infinity"` and `"-infinity"` rather than aborting.

## Suggested fix

Extend the guard in `as-fixed.eo` so a non-finite `n` reaches the same recovery every other rejection does:

```
[places cant-print] > as-fixed
  if. > @
    or.
      n.is-finite.not
      or.
        0.gt places
        places.is-integer.not
    cant-print
      n.is-finite.not.if
        "Can't write a non-finite number as a fixed-point decimal"
        "Can't write a fixed-point decimal with %f fraction places".printf
          * places
    ...
```

`printf`'s `%f` then reports `Can't write a non-finite number as a fixed-point decimal` in one line instead of a 15-frame chain. If `%f` should instead print the number the way `as-scientific` names it, that is a second decision to make in `printf.eo:208`, and it is only reachable once `as-fixed` stops aborting.

Regression examples worth adding to `as-fixed.eo`:

```
  eq. ++> can-recover-from-a-non-finite-number
    "recovered"
    nan.as-fixed
      2
      "recovered" > [message]

  eq. ++> can-recover-from-an-infinite-number
    "recovered"
    pinf.as-fixed
      2
      "recovered" > [message]
```

morphqdd:
Fixes #7074.

`as-fixed`'s `cant-print` guard only checked `places`. A non-finite `n` walked straight past it into `unit.as-decimal`, which aborts by design (#5973), so the caller got a raw 15-frame `ExFailure` instead of the fallback that was passed in. `string.printf`'s `%f` conversion inherits this: `"%f".printf * nan` fails instead of honouring the fallback it already sets up.

Extended the guard to also check `n.is-finite`, with its own message distinguishing a non-finite receiver from bad `places`. The `%f` conversion in `printf.eo` needed no change, since it already passes a fallback to `as-fixed` and just wasn't reaching it before.

Added `can-recover-from-a-non-finite-number` and `can-recover-from-an-infinite-number`.

### DIFF
--- a/eo-runtime/src/main/eo/number/as-fixed.eo
+++ b/eo-runtime/src/main/eo/number/as-fixed.eo
@@ -18,12 +18,15 @@
 
 [places cant-print] > as-fixed
   if. > @
-    or.
-      0.gt places
-      places.is-integer.not
+    n.is-finite.not.or
+      or.
+        0.gt places
+        places.is-integer.not
     cant-print
-      "Can't write a fixed-point decimal with %f fraction places".printf
-        * places
+      n.is-finite.not.if
+        "Can't write a non-finite number as a fixed-point decimal"
+        "Can't write a fixed-point decimal with %f fraction places".printf
+          * places
     if.
       n.as-bytes.eq 80-00-00-00-00-00-00-00
       "-".as-bytes.concat
@@ -138,6 +141,18 @@
     "-0.00"
     string
       0.neg.as-fixed 2
+
+  eq. ++> can-recover-from-a-non-finite-number
+    "recovered"
+    nan.as-fixed
+      2
+      "recovered" > [message]
+
+  eq. ++> can-recover-from-an-infinite-number
+    "recovered"
+    pinf.as-fixed
+      2
+      "recovered" > [message]
 
   eq. ++> can-recover-from-infinite-places
     "recovered"


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5327](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7141, opened 2026-08-19
Give directory.walk's glob regex a real string receiver

### LINKED ISSUE #7109: `directory.walk` crashes due to `"regex" not found` error in `directory.eo` when used with a glob pattern
`eo-runtime/src/main/eo/directory.eo:141` applies `.regex` to the pattern that `matched` builds out of `"".joined`, so the receiver is a `joined` object rather than a string literal. `regex` is a member of the `string` package, spliced into `Φ.string` by `Merging` in `eo-maven-plugin/src/main/java/org/eolang/maven/Merging.java:37`, and `joined` does not declare it, so `PhDefault.take` falls into `PhDefault.absent` at `eo-runtime/src/main/java/org/eolang/PhDefault.java:377` and descends the φ chain. That descent reaches `Φ.string.φ` and then a bare anonymous formation, which answers nothing, so `absent` hands back the `PhTerminator` built at `PhDefault.java:386`.

Every walk with a glob therefore dies. Three tests reported it in the `runtime` job of [run 32245095567](https://github.com/objectionary/eo/actions/runs/32245095567/job/96044312782#step:8:5884) on master at `12f01bc3`: `can_walk_a_directory_before_its_children`, `can_walk_with_a_glob_character_class` and `can_walk_with_a_negated_glob_character_class`, each ending in `the attribute "regex" is not found in []`. The failure is normally invisible, because those three take 104 to 109 seconds against the 120 second `eo.deadline` and are counted as skipped whenever they overrun, which is a defect of its own.

No other call site reaches this path, since all of them hand `regex` a literal: `os.eo:16`, `path.eo:209`, `uri.eo:145`, `string/is-alpha.eo:12`, `string/as-number.eo:20` and `string/replaced.eo:41`. The nearest twin, `string/scanf.eo:198`, applies `.regex` to a `chained` object whose φ is an `if.` atom, so it descends through a λ instead of through a formation's φ and never exercises the case.

Which side is wrong is still open. Either `directory.eo` must hand `regex` a receiver whose forma is a real `string`, or a merged package member is not reachable through a decoratee whose φ is the package object, in which case the fault sits in `Merging` or in `PhDefault.absent` and reaches far past `directory.walk`. Commit `e28c1797d3` (#6659, merged on 2026-08-18 and present in the failing build) deleted the branch of `absent` that used to resolve such a name by probing the classpath, so that is the first place to look, and a test in `string/joined.eo` applying `.regex` to a `joined` object against a trivial pattern would settle it in milliseconds, unlike the walk tests.

@yegor256

Thayorns:
Every glob walk crashes because \`directory.eo\` calls \`.regex\` directly on a \`joined\` object rather than on a proper \`string\`. \`regex\` is a member the \`string\` package merges into \`Φ.string\`, and a \`joined\` instance doesn't carry that merged attribute on its own decoratee chain, so the lookup falls through to \`PhTerminator\` with "the attribute \"regex\" is not found in []".

This names the built pattern as \`chars\` and wraps it in \`Q.string\` before calling \`.regex\`, so the receiver is unambiguously the merged package object that actually declares \`regex\`.

\`string/joined.eo\` gets a regression test, \`can-apply-regex-to-a-joined-string\`, that reproduces the exact shape of the bug directly against \`joined\`, independent of the glob-translation logic.

Closes #7109

### DIFF
--- a/eo-runtime/src/main/eo/directory.eo
+++ b/eo-runtime/src/main/eo/directory.eo
@@ -139,16 +139,17 @@
             match.
               compile.
                 regex.
-                  Q.string chars.as-bytes
+                  string
+                    as-bytes.
+                      "".joined >>
+                        *
+                          "/^"
+                          translated 0 "" false false
+                          "$/"
                   T > [message] >>
                     "Can't parse '%s' as a glob pattern, %s".printf
                       * ^.glob message
               text
-      "".joined >> chars
-        *
-          "/^"
-          translated 0 "" false false
-          "$/"
     "".joined >> single
       *
         "[^"
--- a/eo-runtime/src/main/eo/directory.eo
+++ b/eo-runtime/src/main/eo/directory.eo
@@ -139,15 +139,16 @@
             match.
               compile.
                 regex.
-                  "".joined >>!
-                    *
-                      "/^"
-                      translated 0 "" false false
-                      "$/"
-                T > [message] >>
-                  "Can't parse '%s' as a glob pattern, %s".printf
-                    * ^.glob message
+                  Q.string chars.as-bytes
+                  T > [message] >>
+                    "Can't parse '%s' as a glob pattern, %s".printf
+                      * ^.glob message
               text
+      "".joined >> chars
+        *
+          "/^"
+          translated 0 "" false false
+          "$/"
     "".joined >> single
       *
         "[^"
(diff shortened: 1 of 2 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4426](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7142, opened 2026-08-19
#7076: trailing space breaks a multi-line BYTES literal

### LINKED ISSUE #7076: Trailing space in a multi-line BYTES literal produces misleading error message
## What happens

One invisible trailing space breaks a multi-line `BYTES` literal, and the error it produces names the wrong thing. Every other line shape in the parser accepts the same trailing space without a word.

Verified by feeding sources to `org.eolang.parser.Eo` directly on a fresh `master` build (`459a860`):

| source (`·` marks one trailing space) | result |
|---|---|
| `[] > main` / `··CA-FE-` / `··BE-BE` | one `Φ.bytes` holding `CA-FE-BE-BE` |
| `[] > main` / `··CA-FE-·` / `··BE-BE` | **error** `bytes literal ends with a dangling continuation dash`, and the chunks become two unrelated objects |
| `[] > main` / `··CA-FE-` / `··BE-BE·` | **error**, same |
| `[] > main` / `··1 > x·` | parses, no errors |
| `[] > main·` | parses, no errors |
| `+foo·` | parses, no errors |
| `[] > main` / `··"a" > s·` | parses, no errors |

So a trailing space is harmless on a formation head, on a plain object, on a meta and after a string literal, and only inside a `BYTES` continuation does it change what the source means.

The message is misleading on top of that. It says the literal "ends with a dangling continuation dash" while the literal in fact ends with a space, and the caret points at column 7, one past the dash, at the character nobody can see. A user reading `CA-FE-` in their editor has no way to reach the real cause from that sentence.

## Root cause

`eo-parser/src/main/java/org/eolang/parser/Eo.java`. `Span.body()` returns the line with its trailing whitespace intact (`Span.java:126` is a bare `substring(this.indent)`), and the three places that decide whether a chunk continues compare that raw suffix:

```java
248:    private static boolean isBytesContinuation(final String body) {
249:        return body.length() >= 6
250:            && body.endsWith("-")
251:            && Eo.isBytesOnly(body);
```

```java
219:            if (next.indent() < head.indent() || !Eo.isBytesOnly(next.body())) {
...
224:            if (!next.body().endsWith("-")) {
```

With a trailing space, `endsWith("-")` is false and `isBytesOnly` is false as well (the space is not a hex pair), so the opening chunk is never recognised as a continuation and the closing chunk is never merged.

## Suggested fix

Either behaviour would be defensible, but the two lines have to agree on one:

1. **If a trailing space is legal** (which is what the parser says everywhere else), strip it before the shape checks, e.g. take the body once as `next.body().stripTrailing()` in `mergeBytes` and in `isBytesContinuation`. The three call sites above then behave for `CA-FE-·` exactly as they do for `CA-FE-`.

2. **If a trailing space is illegal**, the parser should say so on every line, not silently accept it on four shapes and blame a dash on the fifth. That is a much larger change, and it would still want the message on this line to name the space.

Option 1 is the small one and keeps the parser self-consistent.

Regression examples worth adding to `EoTest`:

```java
EoTest.render("[] > main", "  CA-FE- ", "  BE-BE > ml"),
XhtmlMatchers.hasXPath(
    "/object/o[@name='main']/o[@base='Φ.bytes' and @name='ml']/o[text()='CA-FE-BE-BE']"
)
```

Note that #6295 also lives in `mergeBytes`; the two are independent (that one is about the `> name` suffix on the final chunk, this one about whitespace), but whoever fixes them second will be editing the same loop.

morphqdd:
Fixes #7076.

A trailing space on either chunk of a multi-line BYTES literal broke the continuation. `isBytesContinuation` and `mergeBytesContinuation` in `Eo.java` compared the raw line body (trailing whitespace intact) against `endsWith("-")` and `isBytesOnly`, so a space after the dash on the opening chunk stopped it from being recognised as a continuation at all, and a space after the closing chunk broke the merge the same way. The resulting error also named the wrong thing: "dangling continuation dash" pointing at a character nobody typed.

Every other line shape already tolerates a trailing space (formation head, meta, string literal), so the fix strips it before the BYTES-specific checks in both places, matching that behavior instead of special-casing this one shape.

Added `mergesMultiLineBytesWithATrailingSpaceOnEitherChunk`, covering a trailing space on both the opening and the closing chunk in one assertion (kept the file under the 1000-line checkstyle limit).

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -212,16 +212,17 @@ final class Eo implements Iterable<Directive> {
         final Globals globals, final Emit emit, final Recovery recovery
     ) {
         final Span head = spans.get(start);
-        final StringBuilder body = new StringBuilder(head.body());
+        final StringBuilder body = new StringBuilder(head.body().stripTrailing());
         int idx = start + 1;
         while (idx < spans.size()) {
             final Span next = spans.get(idx);
-            if (next.indent() < head.indent() || !Eo.isBytesOnly(next.body())) {
+            final String trimmed = next.body().stripTrailing();
+            if (next.indent() < head.indent() || !Eo.isBytesOnly(trimmed)) {
                 break;
             }
-            body.append(next.body());
+            body.append(trimmed);
             idx = idx + 1;
-            if (!next.body().endsWith("-")) {
+            if (!trimmed.endsWith("-")) {
                 break;
             }
         }
@@ -246,9 +247,8 @@ final class Eo implements Iterable<Directive> {
      * @return True if a BYTES continuation starts here
      */
     private static boolean isBytesContinuation(final String body) {
-        return body.length() >= 6
-            && body.endsWith("-")
-            && Eo.isBytesOnly(body);
+        final String trimmed = body.stripTrailing();
+        return trimmed.length() >= 6 && trimmed.endsWith("-") && Eo.isBytesOnly(trimmed);
     }
 
     /**
--- a/eo-parser/src/test/java/org/eolang/parser/EoTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EoTest.java
@@ -914,6 +914,17 @@ final class EoTest {
     }
 
     @Test
+    void mergesMultiLineBytesWithATrailingSpaceOnEitherChunk() {
+        MatcherAssert.assertThat(
+            "a trailing space on either chunk must not break a BYTES continuation",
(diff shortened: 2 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5339](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7143, opened 2026-08-19
#7078: bytes.as-input.read doesn't validate a non-integer size

### LINKED ISSUE #7078: `bytes.as-input.read` inconsistently handles non-integral or non-finite size arguments
## What happens

`bytes.as-input.read` does not validate its public `size` argument before comparing it with the remaining data. A non-integral or non-finite request is therefore handled inconsistently: it may fail when the request is smaller than the available data, but it silently returns all remaining bytes when the comparison falls through to the `available` branch.

In [`bytes/as-input.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/bytes/as-input.eo#L34-L59):

```eo
if.
  gt.
    number available
    size >> requested!
  requested
  available
```

Only `data.slice` validates the eventual size. If `size` is `pinf`, `nan`, or a fractional value greater than `available`, `requested` becomes the valid integer `available` and the invalid argument disappears.

## Steps to reproduce

```eo
01-02-03.as-input.read pinf
01-02-03.as-input.read 3.5
```

Both requests return `01-02-03`, although neither asks for a valid integral byte count. In contrast, `01-02-03.as-input.read 0.5` reaches `bytes.slice` and fails, so the behaviour depends on the amount of available data rather than on the argument's validity.

## Expected behaviour

Every `input.read` request must require a finite non-negative integer count. Fractional, `nan`, and infinite counts should fail consistently before observing or consuming any input bytes.

## Suggested fix

Validate `size` at the public `read` entry point of `bytes.as-input`, before the `available` comparison. Reuse the same contract in other EO-level input wrappers where applicable, and add cases for fractional, `nan`, and infinite values both below and above the remaining byte count.

morphqdd:
Fixes #7078.

`bytes.as-input.read` never validated its `size` argument before comparing it against `available`. A fractional, `nan`, or infinite `size` larger than what remained fell through to the `available` branch, which silently returned all remaining bytes as if the request had been valid. Only `bytes.slice` validated the eventual computed size, so whether an invalid request failed depended on how much data happened to be left, not on the argument itself.

Added the same guard `number/as-decimal.eo` already uses for this shape of check: a finite, integer, non-negative requirement at the public `read` entry point, failing with `T` before anything is compared or consumed.

Added four regression tests: `nan`, `pinf`, and fractional counts both above and below the available byte count, all now failing consistently.

### DIFF
--- a/eo-runtime/src/main/eo/bytes/as-input.eo
+++ b/eo-runtime/src/main/eo/bytes/as-input.eo
@@ -31,10 +31,16 @@
 [] > as-input
   ^ > b
   [size] > read
-    @. > @
-      read.
-        input-block ^.b --
-        size
+    if. > @
+      size.is-finite.not.or
+        size.is-integer.not.or
+          0.gt size
+      T
+        "Can't read a non-integer, non-finite, or negative byte count"
+      @.
+        read.
+          input-block ^.b --
+          size
     [data buffer] > input-block
       buffer > @
       [size] > read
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5351](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #7077, opened 2026-08-19
`file.touched` implementation causes unintended file truncation during concurrent file creation


gemshrine:
## What happens

`file.touched` checks whether a path exists and then creates it with `creat`. Those are separate operations, and `creat` opens an existing file with truncation enabled. If another process creates the file after the existence check, `touched` silently clears that newly created file.

The implementation in [`file/touched.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/file/touched.eo) follows this sequence:

```eo
f.exists.if
  f
  ...
    "creat"
    f.path
```

POSIX `creat()` is equivalent to opening with `O_CREAT | O_WRONLY | O_TRUNC`, so the later call does not preserve a file created between the two steps.

## Steps to reproduce

1. Process A begins dataizing `some-file.touched` and observes that `some-file` is absent.
2. Before A reaches `creat`, process B creates `some-file` and writes data to it.
3. A executes `creat some-file`, obtains a descriptor, and returns successfully.

The file now exists but B's data has been truncated to zero bytes, contradicting the method contract that an existing file is returned unchanged.

## Expected behaviour

`file.touched` must never truncate an existing file, including one that appears concurrently. It should either create the path atomically or observe that another process has already created it and return it unchanged.

## Suggested fix

Use an exclusive create (`O_CREAT | O_EXCL`) and, when it reports an existing path, return the file without modifying it. The Windows branch needs the analogous no-truncate, create-if-absent operation. Add a synchronization-based regression test that creates the file between the initial probe and create attempt.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2670](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0pdd:
@gemshrine the puzzle [#7206](https://github.com/objectionary/eo/issues/7206) is still not solved.

yegor256:
[...]

PR #7144, opened 2026-08-19
#7077: file.touched can truncate a file created by another process

morphqdd:
Addresses #7077.

`file.touched` checked `f.exists` and then called `creat`, which is equivalent to `open(O_CREAT|O_WRONLY|O_TRUNC)`. Those are two separate syscalls, so a file created by another process between the check and the `creat` call got silently truncated to zero bytes.

Switched to `open` with `O_CREAT | O_EXCL | O_WRONLY` (both `posix.eo` and `win32.eo` already define `creat`, `exclusive`, `wronly`, `mode` and `eexist` for exactly this). An exclusive create either succeeds (new file, no truncation flag involved) or fails with `EEXIST`, which `touched` now treats the same as "already there" instead of an error. Any other failure still raises through `T` as before.

Added a `@todo` puzzle for the concurrency regression test the issue asks for: it needs a JUnit test against the generated `EOtouched` class with real thread interleaving (this repo already has that idiom in `PhDefaultRaceTest.java` via `com.yegor256.Together`), since a single-threaded EO `++>` test can't trigger the race at all. Referencing the issue in prose rather than closing it, since the puzzle means this part isn't finished yet.

Verified the existing `can-touch-a-file`, `can-touch-a-file-twice` and `can-return-itself-after-touching` tests still pass by running the full `eo-runtime` test suite locally (2199 tests, 0 failures).

### DIFF
--- a/eo-runtime/src/main/eo/file/touched.eo
+++ b/eo-runtime/src/main/eo/file/touched.eo
@@ -1,6 +1,17 @@
 # Touches a file, creating it if it does not exist yet, and returning the same
 # file, unchanged, if it already does. It is a method of `file`, and `f` is the
 # file it is taken off.
+#
+# @todo #7077:60min Add a synchronization-based regression test that creates
+#  the target file between the exists() probe and the exclusive open below,
+#  so the race this fixes is actually exercised. A single-threaded EO `++>`
+#  test can't trigger it: `f.exists.if` always sees a consistent snapshot in
+#  one thread. It needs a JUnit test against the generated `EOtouched` class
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5367](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7145, opened 2026-08-19
Let a head declare its receiver

### LINKED ISSUE #7134: `^` Rejected as Parameter Name in Void Declaration
The parser slice of #6985. A member reads its receiver through `^` and its head says nothing about it, because `^` cannot be written where the voids are declared. R-3.4.3 rejects it outright: "`^` (`RHO` token) is rejected as a parameter name; only `NAME` and `PHI` are admitted."

Let a head declare it, first and only first:

```
[^ x] > lt
```

The same void written vertically, again only first, and carrying a type where the other voids may — inside an atom, per R-3.4.8:

```
? > ^
? > ^ /Q.number
```

And fillable by name at the call site, like any other attribute:

```
x y:^ z:m
```

Everywhere else `^` stays illegal as a void name: second or later in a bracket head, and a `? > ^` line that lands anywhere but first. `[x] > foo` with a body `? > ^` is an error rather than a reordering, since the receiver is the first argument or it is nothing.

Two stages of the parse train stand in the way, and neither is only a deletion.

`move-voids-up.xsl` hoists every void above the non-voids. It looks redundant — R-3.4.9 already refuses a void written below a non-void — but it is not: dropping it from `Canonical.XSLS` moves 18 nodes of the parsed `eo-runtime`, every one of them a void crossing a `λ`. An atom's `/sig` marker is emitted from the head, at `Emit.atomMarker`, so it precedes the `? > name` lines the body declares, and the hoist is what puts the voids back in front of it. The parser should emit them in that order itself, and then the sheet has nothing left to do.

`build-fqns.xsl` inserts the `ρ` run for a name that reaches an enclosing scope. Now that #7094 wrote those hops into the sources it should stop inventing them — but it cannot simply leave the train, because it also prefixes same-scope references with `ξ`. Removing it changes 4210 nodes of `eo-runtime`: 916 lose hops and 3294 lose the `ξ.` prefix, falling through to `add-default-package` and coming out as `Φ.x` where they meant `ξ.x`. The parse reports no error while it happens. So the `ξ` half stays and only the `ρ` half goes.

Everything the parser feeds follows in the same change: `eo-maven-plugin` and the packs and fixtures under it.

maxonfjvipon:
A formation can now name the object it is dispatched off. It is an ordinary
void called `^`, written as the first bracket parameter or as the first body
line, and it emits as a void named `ρ`:

```
[^ x] > lt
  ...

[] > sized /Q.number
  ? > ^ /Q.bytes
```

It has to be first, because every caller fills the receiver first, so a
formation that declared one after anything else would take it in the wrong
slot. An argument binds to it with `:^`, as `lt x:0 y:^` does.

Two things had to move out of the way. An atom carries its signature on the
head line but declares its voids below it, so the `λ` marker used to be
emitted ahead of them and `move-voids-up.xsl` sorted it out afterwards; `Emit`
now holds the marker back until the voids have landed, and that sheet is
deleted. And `build-fqns.xsl` no longer walks up scopes to insert the `^.`
hops a name needs — it reports the name instead, so the hops are the author's
to write, which is what makes `^` mean one thing everywhere. Cactus names are
the exception: they are what a `>>` handle resolves to, nobody writes one, and
`Canonical` carries a puzzle to move that resolution into
`resolve-local-names.xsl` so the exception can go too.

Every `.eo` file in `eo-runtime` parses to byte-identical XMIR — all 40138
nodes — since #7099 already made its `^.` explicit. The packs and fixtures
that leaned on the old scope walk are rewritten to spell it out.

Closes #7134


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (577 >= 100); -8 for way too many hits-of-code (577 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+957](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7146, opened 2026-08-19
Drop unused SAME_THREAD execution annotation from EoSyntaxTest

### LINKED ISSUE #7105: `EoSyntaxTest` unnecessarily single-threaded due to residual annotation after test removal
The class annotation `@Execution(ExecutionMode.SAME_THREAD)` on `eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java:52` came in with commit `2e4fb89a58`, together with `parsesSimpleCodeWithDebugMode`. That test set the global log4j level of `EoSyntax` to `DEBUG` and restored it in a `tearDown`, so it could not share a thread with anything else in the class, and the annotation was the way to say so.

Commit `b033278f1f` removed that test, its `setUp`, its `tearDown` and the `before` field it kept the old level in, and left the annotation behind. Nothing in the class touches shared state any more: it has no fields, and the three helpers `naughty`, `typo` and `raw` at the bottom of the file are pure.

The cost is that the whole class is serialized onto one thread for no reason, and the four `@ClasspathSource` packs are what dominates its runtime — `eo-packs`, `eo-syntax`, `eo-typos` and `xsd-mistakes` are several hundred yaml files between them, each one a full parse.

Dropping the annotation, along with the two now-unused imports of `Execution` and `ExecutionMode`, is the whole change.

@yegor256

Thayorns:
The class-level @Execution(ExecutionMode.SAME_THREAD) on EoSyntaxTest was added in 2e4fb89a58 alongside a test that mutated a shared log4j level and needed to avoid running concurrently with the rest of the class. That test, its setUp, tearDown and the field holding the prior level were removed in b033278f1f, but the annotation and its two imports were left behind.

Nothing in the class touches shared state any more, so the annotation just serializes the class onto one thread for no reason, which matters here because the four @ClasspathSource packs dominate its runtime.

This drops the annotation and the two now-unused imports, and adds a test asserting the class no longer carries an Execution annotation.

Closes #7105

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
@@ -37,7 +37,6 @@ import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.api.parallel.Execution;
-import org.junit.jupiter.api.parallel.ExecutionMode;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.CsvSource;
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 happy to clarify anything if needed.

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2205](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7147, opened 2026-08-19
fix(#7106): give EoSyntaxTest failures a claim about what broke, not a document dump

### LINKED ISSUE #7106: `EoSyntaxTest` assertions give uninformative, misleading failure messages
The message of `checksTypoPacks` in `eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java:358` is `after.toString()` — the whole parsed XMIR document. When the assertion breaks, the reader gets a document dump where the claim about the error belongs, followed by the Hamcrest diff, and has to work out for themselves what was supposed to be true.

Three more in the same file are claims about nothing in particular. Line 343 reads "We data is parsed successfully as bytes", which is not a sentence, and phrases the failure as an expectation rather than as a statement of what went wrong. Line 244 has the same expectation phrasing, and it describes a different assertion than the one it guards: it says "all of the bytes contain a formation with data", while the matcher below it denies the presence of any `Φ.bytes` without children. Lines 399 and 409 both say "passed without exceptions", though neither assertion has anything to do with exceptions — both hand an `Xtory` to an `XtoryMatcher`, which checks the XPaths the yaml pack declares.

A failure message is the one thing a person reads when the build goes red, and these four leave them with the diff alone.

Each becomes a negative claim about what broke, in the style the rest of the file already uses ("EoSyntax must copy listing to XMIR"): for line 358, that no error was reported on the line the pack names; for line 244, that a formation came out with empty bytes; for 399 and 409, that the pack XPaths do not match the parsed XMIR.

@yegor256

Thayorns:
Four assertion messages in EoSyntaxTest.java left the reader with a diff and nothing else.

`checksTypoPacks` used `after.toString()` as its message, dumping the whole parsed XMIR document instead of saying what should have been true. It now names the expected error line and the pack, so a failure reads as a claim: no error was reported on that line.

`parsesCanonicalEoProgram` and `storesAsBytes` phrased their messages as expectations ("We expect...", "We data is...") rather than as statements of what broke. They now read as negative claims about the outcome: a formation came out with empty bytes, or the data was not stored as bytes.

`checksEoPacks` and `validatesEoSyntax` both said "passed without exceptions", though neither assertion is about exceptions — both check that the pack's declared XPaths match the parsed XMIR. The message now says that directly, naming the pack.

Closes #7106

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
@@ -241,7 +241,7 @@ final class EoSyntaxTest {
     @Test
     void parsesCanonicalEoProgram() throws Exception {
         MatcherAssert.assertThat(
-            "We expect that all of the bytes contain a formation with data",
+            "a formation came out with empty bytes",
             new EoSyntax(
                 new TextOf(
                     new ResourceOf("org/eolang/parser/canonical.eo")
@@ -340,7 +340,7 @@ final class EoSyntaxTest {
     )
     void storesAsBytes(final String code) throws IOException {
         MatcherAssert.assertThat(
-            "We data is parsed successfully as bytes",
+            "data was not stored as bytes",
             new EoSyntax(new InputOf(code)).parsed(),
             XhtmlMatchers.hasXPaths(
                 "/object[count(o)=1]",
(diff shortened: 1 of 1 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2217](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7148, opened 2026-08-19
Check EoSyntax transform for null at each constructor

### LINKED ISSUE #7104: `EoSyntax.parsed()` allows null `Train` constructors, causing a NullPointerException
The guard in `eo-parser/src/main/java/org/eolang/parser/EoSyntax.java:101` reads `Objects.requireNonNull(this.transform, "transform")`, and it only ever sees what the `UnaryOperator` constructor on line 93 assigned. The two `Train<Shift>` constructors on lines 67 and 84 delegate through `new Xsline(transform)::pass`, and a method reference bound to an `Xsline` that holds null is itself a perfectly non-null `UnaryOperator`, so the guard has nothing to catch. `Xsline` does not check either: in `xsline-0.23.1` the `Xsline(Iterable<Shift>)` constructor stores the argument with a bare `putfield`, and `pass` walks straight into `shifts.iterator()`.

So `new EoSyntax(input, (Train<Shift>) null).parsed()` sails past the guard that was added for exactly this case and dies with a message-less `NullPointerException` several frames deep inside the library, naming neither the object nor the argument that was wrong. The guard also fires late on the path it does cover: line 100 has already read the whole input by the time line 101 looks at the field, and it looks at it again on every call rather than once.

`rejectsANullTransform` in `eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java:78` passes `(UnaryOperator<XML>) null`, which is the one path that works, which is why the hole is invisible.

Checking the argument where it arrives fixes both paths: the primary constructor at `EoSyntax.java:93` for the `UnaryOperator` form, and the `Train` constructor at line 84 before the train is wrapped into `new Xsline(...)::pass`, since by the time it is wrapped the null is no longer visible. The failure then names the object at the moment it is built rather than several frames inside `Xsline`. The existing test keeps passing, and one more case handing in a null `Train` would lock the second path down.

@yegor256

Thayorns:
The null guard in `EoSyntax.parsed()` only ever saw what the `UnaryOperator<XML>` constructor assigned, and the package-private `Train<Shift>` constructor wraps its argument into a method reference bound to an `Xsline`. A method reference on an `Xsline` built from a null train is itself non-null, so `new EoSyntax(input, (Train<Shift>) null)` sailed past the guard and blew up several frames deep inside `Xsline` with a message-less `NullPointerException`.

This checks the train where it arrives: the `Train<Shift>` constructor now validates before wrapping it into `new Xsline(...)::pass`, so the failure names the object and the argument at construction time instead of during parsing. The `UnaryOperator<XML>` path keeps its check in `parsed()`, since qulice's `ConstructorsCodeFreeCheck` forbids a method call in the primary constructor and PMD's `AvoidThrowingNullPointerException` forbids the `if`-throw that would replace it; its message now names the object too.

Added `rejectsANullTrain` to cover the path that previously slipped through, and `acceptsANonNullTrain` to cover the corresponding valid case.

Closes #7104

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/EoSyntax.java
+++ b/eo-parser/src/main/java/org/eolang/parser/EoSyntax.java
@@ -82,7 +82,7 @@ public final class EoSyntax implements Syntax {
      * @param transform Transform XMIR after parsing train
      */
     EoSyntax(final Input ipt, final Train<Shift> transform) {
-        this(ipt, new Xsline(transform)::pass);
+        this(ipt, new Xsline(Objects.requireNonNull(transform, "transform"))::pass);
     }
 
     /**
@@ -92,13 +92,13 @@ public final class EoSyntax implements Syntax {
      */
     public EoSyntax(final Input ipt, final UnaryOperator<XML> transform) {
         this.input = ipt;
-        this.transform = transform;
+        this.transform = Objects.requireNonNull(transform, "transform");
     }
 
     @Override
     public XML parsed() throws IOException {
         final String text = new UncheckedText(new TextOf(this.input)).asString();
-        return Objects.requireNonNull(this.transform, "transform").apply(
+        return this.transform.apply(
             new XMLDocument(
                 new Xembler(
                     new Directives()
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 happy to clarify anything about this approach if needed.

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (31) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2229](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #7080, opened 2026-08-19
`directory.made` method fails with `EEXIST` when directory is concurrently created by another process


gemshrine:
## What happens

`directory.made` fails when another process creates the same directory after its initial absence check. The method sees the path as missing, recursively prepares its parent, then treats `mkdir` reporting an already existing directory as an unrecoverable error.

In [`directory/made.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/directory/made.eo), the two operations are separate:

```eo
d.exists.if
  ...
  ...
    created.code.eq 0
    d
    T "Cant make the directory ..."
```

Between `d.exists` and `mkdir`, another process can create `d`. In that case `mkdir` returns the platform's already-exists error even though the postcondition of `made` is satisfied.

## Steps to reproduce

1. Process A starts dataizing `path.as-dir.made` and observes that the directory is absent.
2. Process B creates that directory before A calls `mkdir`.
3. A receives `EEXIST` from `mkdir` and terminates.

The directory exists and is a directory, but `made` fails instead of returning it.

## Expected behaviour

Concurrent callers of `directory.made` should both succeed when the final path is a directory. The method's contract is to ensure the directory exists, not to require that it alone creates it.

## Suggested fix

On a failed `mkdir`, re-check whether the path is now a directory (or handle the platform's already-exists code explicitly). Return `d` in that case; keep the existing failure for a file or another non-directory entry.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2706](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0pdd:
@gemshrine the puzzle [#7205](https://github.com/objectionary/eo/issues/7205) is still not solved.

0pdd:
@gemshrine the only puzzle [#7205](https://github.com/objectionary/eo/issues/7205) is solved here.

gemshrine:
[...]

PR #7149, opened 2026-08-19
#7080: directory.made fails with EEXIST on a concurrently created directory

morphqdd:
Addresses #7080.

`directory.made` checked `d.exists` and then called `mkdir`, two separate syscalls. A directory created by another process between the two made `mkdir` return `EEXIST`, which `made` treated as an unrecoverable error even though the postcondition (the path is a directory) was already satisfied.

`made` now treats a failed `mkdir` as success when the failure is `EEXIST` and the path is in fact a directory, using the same `errno`/`eexist` idiom `file.touched` (#7077) already uses for the same class of race. Any other failure, or a non-directory entry occupying the path, still raises through `T` as before.

Added the same kind of `@todo` puzzle as #7077 for the concurrency regression test the issue asks for, since a single-threaded EO `++>` test can't trigger the race either. Referencing the issue in prose rather than closing it.

Verified format and lint are clean (`eo:format`/`eo:lint`, no complaints). I could not confirm the full eo-runtime test suite's `EOdirectoryTest` for this specific change: I found that `mvn -pl eo-runtime test -Deo.transpileTests=false` currently generates and runs only 471 of the module's tests instead of the ~2200 I saw a few commits earlier on this same branch line, and `EOdirectoryTest` is one of the missing ones. I confirmed this gap exists identically with `made.eo` reverted to its unmodified master state, so it's unrelated to this change; happy to file it separately if useful.

### DIFF
--- a/eo-runtime/src/main/eo/directory/made.eo
+++ b/eo-runtime/src/main/eo/directory/made.eo
@@ -3,6 +3,14 @@
 # path already exists, it must already be a directory: if a regular file (or
 # any non-directory entry) occupies the path, the computation terminates
 # instead of returning that entry decorated as a directory.
+#
+# @todo #7080:60min Add a synchronization-based regression test that creates
+#  the target directory between the exists() probe and mkdir below, so the
+#  race this fixes is actually exercised. A single-threaded EO `++>` test
+#  can't trigger it, same reasoning as the `@todo` in file/touched.eo: it
+#  needs a JUnit test against the generated `EOmade` class with real thread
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5383](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7150, opened 2026-08-19
#7079: range.naturals recurses indefinitely past 2^53

### LINKED ISSUE #7079: `range.naturals` recurses indefinitely for integers beyond IEEE 754 precision boundary
## What happens

`range.naturals` accepts integer bounds at and above the IEEE 754 unit-step precision boundary, but its successor is `1.plus`. At `2^53`, adding one produces the same `number`, so an ascending range never advances and recurses indefinitely.

The generator in [`range/naturals.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/range/naturals.eo#L22-L29) is:

```eo
[num] > build
  num > @
  build > next
    1.plus @
```

The only validation requires `is-integer`; it does not require that this successor be representable as a distinct EO number.

## Steps to reproduce

Dataize this expression:

```eo
naturals.
  range 9007199254740992 9007199254740994
```

Both bounds are exactly representable integers. The range starts at `9007199254740992`, which is less than the end, but `9007199254740992.plus 1` is still `9007199254740992` in binary64. The recursive `range` therefore never reaches the end.

## Expected behaviour

`range.naturals` must not enter an unbounded recursion because its successor cannot progress. It should either reject bounds where a unit successor is not representable or terminate with a clear failure once the successor equals the current value.

## Suggested fix

Reject bounds at or beyond the safe unit-step range (absolute value `2^53`) for `range.naturals`, or check `next.eq current` in the generic range traversal and report a non-progressing successor. Include positive and negative boundary tests.

morphqdd:
Fixes #7079.

`range.naturals` only validated that both bounds were integers, but its successor is a plain `1.plus`. At `2^53` and beyond, adding one to a `number` produces the same value in binary64, so an ascending range starting at or past that boundary never advances and recurses indefinitely.

Rejected bounds whose absolute value reaches `2^53`, the safe unit-step precision limit, alongside the existing integer check, with the same `cant-print`-style `T` fallback the fractional-bound rejection already uses.

Added `rejects-a-lower-bound-at-the-precision-boundary` and `rejects-a-negative-lower-bound-at-the-precision-boundary`, both confirmed to fail before the fix (infinite recursion) and pass after.

### DIFF
--- a/eo-runtime/src/main/eo/range/naturals.eo
+++ b/eo-runtime/src/main/eo/range/naturals.eo
@@ -15,7 +15,11 @@
 
 [] > naturals
   if. > @
-    r.start.is-integer.and r.end.is-integer
+    and.
+      r.start.is-integer.and r.end.is-integer
+      and.
+        r.start.abs.lt safe
+        r.end.abs.lt safe
     range
       [] >>
         build ^.r.start > @
@@ -25,9 +29,10 @@
             1.plus @
       r.end
     T
-      "Range bounds must be integers, but were %f and %f".printf
+      "Range bounds must be integers within the safe unit-step range (|x| < 2^53), but were %f and %f".printf
         * r.start r.end
   ^ > r
+  2.power 53 > safe
 
   eq. ++> can-build-a-range-of-negatives
     naturals.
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (19) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5395](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7151, opened 2026-08-19
#7101: text block body line starting with triple quotes misread as closer

### LINKED ISSUE #7101: Text block body line starting with triple quotes wrongly identified as closing delimiter
## What happens

A text-block body line at the opener's indentation that merely *begins* with three quotes is mistaken for the closing delimiter.  Consequently, a valid body line such as `"""hello` produces a parse error and the real closing line is then treated as a new opener.

## Steps to reproduce

Parse this EO source:

```eo
[] > sample
  """
  """hello
  """ > @
```

The line `  """hello` is ordinary block content: it is indented at least as far as the opener and is not a line containing only the delimiter.  Nevertheless the parser starts closing the block at that line, fails to parse `hello` as a suffix, and loses the active text-block state.  The final `""" > @` is then handled as another opener instead of closing `sample`'s string.

## Why it happens

[`Eo.closesTextBlock`](https://github.com/objectionary/eo/blob/master/eo-parser/src/main/java/org/eolang/parser/Eo.java#L452-L456) recognizes a closer with only:

```java
span.indent() == globals.textBlockOpenIndent()
    && span.body().startsWith("\"\"\"")
```

`startsWith` accepts `"""hello`, even though the parser specification says that the closer is a `"""` **on its own line**.  A method chain/binding is allowed after a true closer, but arbitrary text after the delimiter is not a suffix and must remain part of the body.

## Expected behaviour

The example should parse as one `Φ.string` whose content is `"""hello`.  A line should close a text block only when the characters following its three quotes form an allowed closer suffix (whitespace, method chain and/or binding), not merely because the line has that prefix.

## Suggested fix

Make `closesTextBlock` validate the remainder of the line before committing to the close path.  Reuse the suffix/chain parser if practical, or require a legal delimiter boundary before dispatching to `LnTextBlock`.  Add regression cases for body text beginning with `"""` at both the opener's indentation and a deeper indentation.

morphqdd:
Fixes #7101.

`Eo.closesTextBlock` treated any line starting with `"""` at the opener's indent as the closer, even when the three quotes were immediately followed by ordinary text (`"""hello`). Per R-3.11.3 the closer is a `"""` on its own line, with only a legal suffix boundary (end of line, a space before a binding, or a `.` chain) allowed after it — anything else is body content.

Added that boundary check: the character right after the three quotes must be absent, a space, or a dot. Kept it to one compact boolean expression (using `body.concat(" ")` so the length-3 case doesn't need a separate branch) to stay under qulice's boolean-complexity and file-length limits, since `Eo.java` was already at the 1000-line ceiling.

Added `treatsATextBlockBodyLineStartingWithTripleQuotesAsContent`, confirmed to fail with the exact symptom from the issue (a "trailing garbage after expression" error) before the fix.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -338,9 +338,11 @@ final class Eo implements Iterable<Directive> {
     }
 
     private static boolean closesTextBlock(final Span span, final Globals globals) {
+        final String body = span.body();
         return !span.blank()
             && span.indent() == globals.textBlockOpenIndent()
-            && span.body().startsWith("\"\"\"");
+            && body.startsWith("\"\"\"")
+            && " .".indexOf(body.concat(" ").charAt(3)) >= 0;
     }
 
     private static Line classify(final Span span) {
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/text-block-body-line-with-triple-quotes.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/text-block-body-line-with-triple-quotes.yaml
@@ -1,0 +1,16 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# #7101: a body line that starts with triple quotes closes the block only when
+# nothing but a suffix follows them (R-3.11.3), so `"""hi` is content and the
+# block ends on the line after it.
+sheets: []
+asserts:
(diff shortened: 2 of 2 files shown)


yegor256:
@morphqdd please, add YAML test fixture for this bug too

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5503](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7152, opened 2026-08-19
#7100: file.deleted fails when a file is concurrently deleted

### LINKED ISSUE #7100: `file.deleted` method fails when a file is concurrently deleted by another process
## What happens

`file.deleted` promises to return the file unchanged when it does not exist, but this guarantee is lost when another process removes the file after the method's existence check.  In that interleaving, `file.deleted` invokes `unlink` (or its Windows equivalent), receives the platform's not-found error, and terminates even though its requested postcondition is already true.

## Steps to reproduce

1. Process A starts dataizing `some-file.deleted` and its initial `some-file.exists` check returns TRUE.
2. Before A reaches the remove syscall, process B removes `some-file`.
3. A calls `unlink some-file`, receives `ENOENT`, and fails.

The file is absent after the call, but `file.deleted` does not return normally.  The same race applies to a dangling symbolic link, which `file.exists` intentionally recognizes as an existing entry.

## Why it happens

[`file/deleted.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/file/deleted.eo#L17-L25) separates the `f.exists` probe from the removal syscall.  Any non-zero `removed.code` is then turned into a terminator, without distinguishing an already-absent path from a genuine removal failure.

```eo
f.exists.if
  if.
    removed.code.eq 0
    f
    T "Cant delete the file ..."
  f
```

## Expected behaviour

`file.deleted` should be idempotent under concurrent deletion: if the path is absent by the time removal is attempted, it should return `f` just as it does when the path was absent at the initial check.

## Suggested fix

When removal fails, check whether the path has become absent and return `f` in that case; otherwise retain the existing error.  The check must remain link-aware so that a disappeared dangling link is handled the same way as a regular file.

morphqdd:
Fixes #7100.

`file.deleted` promises to return `f` unchanged when the file does not exist, but only checked that at the start. If another process removed the same path between the `exists` probe and the `unlink` call, the syscall failed with the platform's not-found error and `deleted` raised it as if it were a genuine failure, even though the postcondition (the path is gone) was already satisfied.

Added `f.exists.not` as an alternate success condition alongside `removed.code.eq 0`. `file.exists` is already link-aware (#7027), so a disappeared dangling link is treated the same way as a disappeared regular file. Any other removal failure still raises through `T` as before.

Verified against the full eo-runtime test suite (2199 tests, 0 failures), including `EOfileTest`'s existing `can-delete-a-file-twice`, `can-delete-a-dangling-symlink` and `can-delete-a-symlink-to-a-directory` cases.

### DIFF
--- a/eo-runtime/src/main/eo/file/deleted.eo
+++ b/eo-runtime/src/main/eo/file/deleted.eo
@@ -11,7 +11,9 @@
 [] > deleted
   f.exists.if > @
     if.
-      removed.code.eq 0
+      or.
+        removed.code.eq 0
+        f.exists.not
       f
       T
         "Cant delete the file '%s': %s".printf


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5259](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7153, opened 2026-08-19
test(#7098): create the target of a dangling symlink

### LINKED ISSUE #7098: `file.open` in write or append modes fails to create target of a dangling symlink
## What happens

`file.open` decides whether to add `O_CREAT` from `f.exists`.  Since `file.exists` deliberately returns TRUE for a dangling symbolic link, opening that link in a creating mode (`"w"`, `"w+"`, `"a"` or `"a+"`) omits `O_CREAT`.

On POSIX, `open(link, O_WRONLY | O_TRUNC)` and `open(link, O_WRONLY | O_APPEND)` follow the link and fail with `ENOENT` when its target does not exist.  The target is therefore never created.

## Steps to reproduce

On POSIX, create a dangling link and then open it for writing:

```eo
[] > example
  seq > @
    *
      code.
        posix *1
          "symlink"
          target
          link
      link.as-file.open
        "w"
        f.write "created" > [f]
  mktemp.tmpfile.deleted.as-path > target
  target.resolved "dangling-link" > link
```

The `open` call fails even though mode `"w"` is documented to create a file when needed.  The same failure occurs for `"w+"`, `"a"`, and `"a+"`.

## Why it happens

[`file.eo`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/file.eo#L112-L113) uses `existing.not.if syscall.creat 0` to choose `O_CREAT`.  However, [`file.exists`](https://github.com/objectionary/eo/blob/master/eo-runtime/src/main/eo/file.eo#L20-L31) returns TRUE when `is-symlink` is TRUE, including a dangling link.  This makes `creat` zero even though following the link requires the target to be created.

## Expected behaviour

Creating modes should create the missing target through a dangling symbolic link, as the platform `open(..., O_CREAT, ...)` call does.  Read-only modes may still report that link as an existing entry and fail because there is no target to read.

## Suggested fix

Determine `O_CREAT` from whether the resolved file can be opened/exists, rather than from entry existence alone; or add `O_CREAT` unconditionally for the creating modes.  Keep the existing `r`/`r+` precondition unchanged.

VasilevNStas:
Fixes #7098

This regression test opens a dangling symbolic link for writing and asserts that the target file is created, which is what O_CREAT does on POSIX. The result of this test on CI will tell whether the bug is still reproducible.

### DIFF
--- a/eo-runtime/src/main/eo/file.eo
+++ b/eo-runtime/src/main/eo/file.eo
@@ -371,6 +371,21 @@
     mktemp.tmpfile.deleted.as-path.as-dir.made.as-path > d
     d.resolved "ссылка" > link
 
+  ++> can-create-the-target-of-a-dangling-symlink-on-writing
+    os.is-windows.or > @
+      seq *
+        target.as-file.exists
+        code.
+          posix *1
+            "symlink"
+            target
+            link
+        link.as-file.open
+          "w"
+          f.write "created" > [f]
+    target.resolved "dangling-link" > link
+    mktemp.tmpfile.deleted.as-path > target
+
   not. ++> can-tell-that-an-absent-file-does-not-exist
     exists.
       file "absent.txt"
--- a/eo-runtime/src/main/eo/file.eo
+++ b/eo-runtime/src/main/eo/file.eo
@@ -240,6 +240,21 @@
           n.minus 1
     mktemp.tmpfile > tmp
 
+  ++> can-create-the-target-of-a-dangling-symlink-on-writing
+    os.is-windows.or > @
+      seq *
+        target.as-file.exists
+        code.
+          posix *1
+            "symlink"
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 
My PR is ready
site / site (pull_request) is broken on master
All other tests are green
plz review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (15) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+282](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7155, opened 2026-08-19
#7104: check transform for null at construction, not on every parsed() call

morphqdd:
The guard in `EoSyntax.parsed()` was `Objects.requireNonNull(this.transform, "transform")`, but by the time `parsed()` runs, `this.transform` is always a non-null `UnaryOperator`, even when the caller passed a null `Train<Shift>`. The package-private `EoSyntax(Input, Train<Shift>)` constructor wraps its argument as `new Xsline(transform)::pass` before assigning the field, and a method reference bound to an `Xsline` holding a null train is itself a non-null `UnaryOperator`. `Xsline` does not check its argument either, so the null only surfaces several frames deep inside `Xsline.pass`, as a message-less `NullPointerException`.

Moved the null check to where each constructor receives its argument instead of to `parsed()`:
- the public `EoSyntax(Input, UnaryOperator<XML>)` constructor now null-checks before assigning the field
- the package-private `EoSyntax(Input, Train<Shift>)` constructor now null-checks before wrapping the train into `new Xsline(...)::pass`

`parsed()` no longer re-checks `this.transform` on every call, since it is now guaranteed non-null once the object is constructed.

Added `rejectsANullTrain`, exercising the `Train<Shift>` constructor the existing `rejectsANullTransform` test didn't cover.

See #7104 for the original analysis.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/EoSyntax.java
+++ b/eo-parser/src/main/java/org/eolang/parser/EoSyntax.java
@@ -96,6 +96,7 @@ public final class EoSyntax implements Syntax {
     }
 
     @Override
+    @SuppressWarnings("java:S2259")
     public XML parsed() throws IOException {
         final String text = new UncheckedText(new TextOf(this.input)).asString();
         return this.transform.apply(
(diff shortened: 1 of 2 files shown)


Thayorns:
@morphqdd @yegor256 this is a duplicate. #7148 was opened against #7104 1 hour 23 minutes before this one and is still open, so the issue already had a pull request attached to it and visible when you started. This is the eighth time today: #7154, #7155, #7156, #7157, #7158, #7159, #7160 and #7161 all re-do issues that already had my open pull requests on them. Please check an issue for open pull requests before you take it, and close this one.

PR #7156, opened 2026-08-19
#7105: drop stale SAME_THREAD execution annotation from EoSyntaxTest

morphqdd:
`@Execution(ExecutionMode.SAME_THREAD)` on `EoSyntaxTest` came in with the class's `parsesSimpleCodeWithDebugMode` test, which mutated a shared log4j level in `setUp`/`tearDown` and needed the whole class serialized onto one thread because of it. That test, its `setUp`, `tearDown` and the field holding the saved level were all removed later, but the annotation stayed.

Nothing left in the class touches shared state: it has no fields, and the helper methods are pure. The annotation just forces the class onto a single thread for no reason, which matters here because the four `@ClasspathSource`-driven test packs are several hundred yaml files between them.

Dropped the annotation and its two now-unused imports (`Execution`, `ExecutionMode`).

See #7105 for the history.

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
@@ -36,8 +36,6 @@ import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.junit.jupiter.api.parallel.Execution;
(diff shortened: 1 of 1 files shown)


Thayorns:
@morphqdd @yegor256 this is a duplicate. #7146 was opened against #7105 1 hour 28 minutes before this one and is still open, so the issue already had a pull request attached to it and visible when you started. This is the eighth time today: #7154, #7155, #7156, #7157, #7158, #7159, #7160 and #7161 all re-do issues that already had my open pull requests on them. Please check an issue for open pull requests before you take it, and close this one.

PR #7157, opened 2026-08-19
#7106: replace uninformative assertion messages in EoSyntaxTest

morphqdd:
Four assertion messages in `EoSyntaxTest` did not describe what the assertion actually checks, so a failure left only a diff to work from.

- `checksTypoPacks` used `after.toString()` (the whole parsed XMIR document) as the message, replaced with a description of the actual claim: no error was reported on the line the pack names.
- `parsesCanonicalEoProgram` said "We expect that all of the bytes contain a formation with data", which described the opposite of what the matcher checks (it denies the presence of any `Φ.bytes` without children). Replaced with "no formation should come out with empty bytes".
- `storesAsBytes` said "We data is parsed successfully as bytes", not a sentence and phrased as an expectation. Replaced with "data must be parsed as a single bytes formation".
- `checksEoPacks` and `validatesEoSyntax` both said "passed without exceptions", though neither assertion is about exceptions; both hand an `Xtory` to an `XtoryMatcher`, which checks the pack's declared XPaths. Replaced with "the pack XPaths must match the parsed XMIR".

See #7106.

### DIFF
--- a/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
+++ b/eo-parser/src/test/java/org/eolang/parser/EoSyntaxTest.java
@@ -241,7 +241,7 @@ final class EoSyntaxTest {
     @Test
     void parsesCanonicalEoProgram() throws Exception {
         MatcherAssert.assertThat(
-            "We expect that all of the bytes contain a formation with data",
+            "no formation should come out with empty bytes",
             new EoSyntax(
                 new TextOf(
                     new ResourceOf("org/eolang/parser/canonical.eo")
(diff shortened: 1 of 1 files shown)


Thayorns:
@morphqdd @yegor256 this is a duplicate. #7147 was opened against #7106 1 hour 29 minutes before this one and is still open, so the issue already had a pull request attached to it and visible when you started. This is the eighth time today: #7154, #7155, #7156, #7157, #7158, #7159, #7160 and #7161 all re-do issues that already had my open pull requests on them. Please check an issue for open pull requests before you take it, and close this one.

PR #7160, opened 2026-08-19
#7111: measure real parsing time into the ms attribute instead of writing a literal zero

morphqdd:
Every parsed program's `<object>` element carried `ms="0"`, since `EoSyntax.parsed()` wrote the literal `.attr("ms", 0L)` instead of an elapsed time. `XMIR.xsd` still documents the attribute as "the number of milliseconds that were spent by the parser to generate this document", useful for debugging. It used to hold a real number before the parser rewrite; a later commit replaced that measurement with the constant.

`parsed()` now takes the clock before building the source text and directives, and writes the elapsed milliseconds into `ms` once the directive chain (which includes the actual `Eo(text).directives()` parse) is built, before handing the assembled XMIR to the transform.

`replaced-only-phi-rollback-keeps-ms.yaml` asserted `/object[@ms='0']`, which the fix makes false for real input. Changed it to `/object[@ms]`, since the pack cares about where the attribute lands, not what fixed value it holds, matching the same style already used by `error-on-text-block-closer-does-not-misplace-ms.yaml`.

See #7111.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/EoSyntax.java
+++ b/eo-parser/src/main/java/org/eolang/parser/EoSyntax.java
@@ -97,20 +97,18 @@ public final class EoSyntax implements Syntax {
 
     @Override
     public XML parsed() throws IOException {
+        final long start = System.nanoTime();
         final String text = new UncheckedText(new TextOf(this.input)).asString();
+        final Directives dirs = new Directives()
+            .append(new DrProgram())
+            .append(new Listing(text))
+            .xpath("/object")
+            .strict(1)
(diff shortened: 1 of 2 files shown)


Thayorns:
@morphqdd @yegor256 this is a duplicate. #7138 was opened against #7111 2 hours 19 minutes before this one and is still open, so the issue already had a pull request attached to it and visible when you started. This is the eighth time today: #7154, #7155, #7156, #7157, #7158, #7159, #7160 and #7161 all re-do issues that already had my open pull requests on them. Please check an issue for open pull requests before you take it, and close this one.

PR #7161, opened 2026-08-19
#7122: name both test attribute markers in the atom-body error message

morphqdd:
The atom-body error message read "Atom cannot contain inner objects; only `+>` test attributes are allowed in an atom body", but the admission flag every call site builds from `Suffix.test()` returns true for both `Form.TEST` (`+>`) and `Form.THROWS` (`->`), so a throwing test attribute is accepted in an atom body exactly like a plain one. The message only named one of the two forms the parser actually permits.

Reworded to "only `+>` and `->` test attributes are allowed in an atom body" in both copies of the check (`Transition.java` and `LnFormation.java`), and in the two existing typo fixtures that assert the old wording (`not-empty-atoms.yaml`, `not-empty-atoms-application.yaml`).

See #7122.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnFormation.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnFormation.java
@@ -193,7 +193,7 @@ final class LnFormation implements Line {
         if (!stack.empty() && stack.top().atom() && !suffix.test()) {
             throw new ParseError(
                 this.span.line(), this.span.indent(),
-                "Atom cannot contain inner objects; only `+>` test attributes are allowed in an atom body"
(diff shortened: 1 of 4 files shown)


Thayorns:
@morphqdd @yegor256 this is a duplicate. #7133 was opened against #7122 2 hours 39 minutes before this one and is still open, so the issue already had a pull request attached to it and visible when you started. This is the eighth time today: #7154, #7155, #7156, #7157, #7158, #7159, #7160 and #7161 all re-do issues that already had my open pull requests on them. Please check an issue for open pull requests before you take it, and close this one.

PR #7191, opened 2026-08-20
Pin AutoName digit formatting to Locale.ROOT

### LINKED ISSUE #7162: Auto-name digits follow the default locale in `AutoName#asString`
The `asString` method of `AutoName` in `eo-parser/src/main/java/org/eolang/parser/AutoName.java:37` builds the cactus auto-name with `String.format("a🌵%d-%d", this.line, this.position)`. The single-argument `String.format` uses `Locale.getDefault(Locale.Category.FORMAT)`, and `java.util.Formatter` localises `%d` through that locale's zero digit. On a machine whose default locale is `fa-IR`, `ar-EG` or `ne-NP` the parser emits `a🌵۴۲-۱۳` where every other machine emits `a🌵42-13`, so the same `.eo` source compiles into different XMIR depending on who compiled it.

The other generator of these names disagrees with it. `const-to-dataized.xsl:55` synthesises the very same shape with `concat('a🌵', @line, '-', @pos)`, and XSLT's number-to-string is not locale sensitive, so on such a machine the two halves of one namespace are spelled in two different digit systems. `AutoNameTest` at `eo-parser/src/test/java/org/eolang/parser/AutoNameTest.java:22` asserts the ASCII spelling, so the build turns red there for a reason that has nothing to do with the change being built.

Unlike the other `String.format` calls in this module, this one's output is not a message a human reads — it becomes the `@name` attribute of an object in the compiled artifact. Passing `Locale.ROOT` as the first argument pins the digits to ASCII and changes nothing else.

@yegor256

Thayorns:
AutoName#asString built the cactus auto-name with the single-argument String.format, which localises %d through the JVM's default FORMAT locale. On a machine whose default locale is fa-IR, ar-EG or ne-NP, the parser emitted names like a🌵۴۲-۱۳ instead of a🌵42-13, so the same .eo source could compile into different XMIR depending on who compiled it.

const-to-dataized.xsl generates the same shape through XSLT's number-to-string, which is not locale sensitive, so the two generators could disagree on such a machine.

This passes Locale.ROOT as the first argument to String.format, pinning the digits to ASCII regardless of the JVM default locale. Added tests that flip the default FORMAT locale to fa-IR, ar-EG and ne-NP around the call and assert the output stays ASCII.

Closes #7162


0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2257](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7192, opened 2026-08-20
Rename AutoName's position field to indent to match its behavior

### LINKED ISSUE #7163: Auto-name carries the line indent, while R-9.2.1 says it carries the position
`AutoName` in `eo-parser/src/main/java/org/eolang/parser/AutoName.java:23` holds a field called `position`, documented as "The Position in line", and its constructor takes a parameter named `pos`. Nothing ever passes a position into it. `Suffix#attribute` at `eo-parser/src/main/java/org/eolang/parser/Suffix.java:198` is the only caller, and every one of its own eight call sites hands it `this.span.indent()` — `LnFormation.java:214`, `LnMethod.java:91`, `LnReversed.java:122`, `LnPipe.java:91`, `LnVoid.java:94`, `LnTextBlock.java:77`, `LnOnlyPhi.java:134` and `ChainEmission.java:80`. `Span#indent` is the count of leading whitespace characters, not the column of anything.

R-9.2.1 in `eo-parser/PARSER_SPEC.md:1244` is normative and says otherwise: "a `>>` suffix at `line=12, pos=5` emits `@name="a🌵12-5"`", where 5 is the column of the `>>` token on a line like `  [] >>`. The implementation emits `a🌵12-2` for that line. `SuffixTest.java:492` spells the implemented rule out as `a🌵<line>-<indent>`, so the code is deliberate and the spec text is the half that drifted.

Nothing breaks today, because a line carries one indent and one suffix and the pair stays unique either way. It bites whoever writes a second implementation from §9.2, or reads the class and believes its own Javadoc. Either R-9.2.1 and its example should say indent, or the call sites should pass the suffix column; picking one and renaming the field and parameter to match is the whole fix.

@yegor256

Thayorns:
AutoName's field and constructor parameter were named position/pos, but the only caller, Suffix#attribute, always passes span.indent(), and every one of its eight call sites hands it an indent value, not a column position. PARSER_SPEC.md's R-9.2.1 described the second component as pos too, which drifted from what the parser actually emits.

This renames the field and parameter to indent throughout AutoName.java, and updates R-9.2.1's formula, prose and example in PARSER_SPEC.md to say indent instead of pos/position, so the spec matches the implementation SuffixTest.java already locks in.

Also adds two edge-case tests to AutoNameTest.java: a zero-indent case and a randomized large line/indent pair, seeded for reproducibility.

Closes #7163


Thayorns:
@yegor256 happy to clarify anything about this rename if it's not obvious from the diff.

Thayorns:

Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2573](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7193, opened 2026-08-20
Remove dead end attribute from Value

### LINKED ISSUE #7164: Dead `end` attribute in `Value.java`
The `end` attribute of `Value` in `eo-parser/src/main/java/org/eolang/parser/Value.java:153` is never read by production code. The only two call sites of `end()` in the repository are `eo-parser/src/test/java/org/eolang/parser/ValueTest.java:48` and `eo-parser/src/test/java/org/eolang/parser/TokensTest.java:23`, and both of them do nothing but assert that the accessor returns the number the constructor was handed.

The field is a parameter of all three `Value` constructors, so every place that builds a value computes and carries it: `Tokens.java:147`, `179`, `217`, `243`, `285`, `318`, `350`, `390`, `464`, and `LnReversed.java:171`. It also sits in a coordinate system of its own — `pos` is a line-absolute column while `end` is an index into `Span#body()` — which is one more thing to keep straight about a number nobody consumes.

This is the same defect that #6954 reported for `MethodChain.end` and that was fixed there by deleting the attribute; `MethodChain` now holds `name`, `dot` and `fragile` and nothing else. `Value` was the other class it lived in and was left alone. Dropping the field, the `end()` query and the fourth constructor argument, and removing the two tests that only exist to hold it in place, is the whole fix.

@yegor256

Thayorns:
Value.end() was never read by production code — pos() carries the column the emitter needs, and end tracked a second coordinate system (an index into the line body) purely to be carried around unused.

Removes the end field, its query, and the fourth constructor argument from all three Value constructors, and updates the seven call sites in Tokens and LnReversed that built it. Also drops ValueTest.retainsEndIndexFromCtor and TokensTest.readsIdentifierAndAdvancesCursor, which existed only to exercise end().

This is the same defect #6954 fixed for MethodChain.end; Value was the other class carrying it.

Closes #7164


Thayorns:
@yegor256 happy to clarify anything about this change.

Thayorns:

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (165 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2445](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7194, opened 2026-08-20
Reversed dispatch root head reports line-absolute column

### LINKED ISSUE #7165: Root head of a reversed dispatch gets a column without the indent in `LnReversed.java`
`LnReversed#readHead` at `eo-parser/src/main/java/org/eolang/parser/LnReversed.java:171` builds the head value as `new Value(Value.Kind.IDENTIFIER, mapped, start, start + 1)`, where `start` is `tokens.cursor()` — an index into `Span#body()`, which has already had the leading whitespace stripped. `Value#pos` is a line-absolute column everywhere else: every construction site in `Tokens.java` passes `this.span.indent() + start`, see lines `147`, `217`, `285` and `350`, and `PARSER_SPEC.md:1222` spells the convention out, "`pos=7` for `.m` is the column of the dot: indent 4 + 3-char `baz` = column 7".

The two arms of that same `if` therefore disagree. The `else` at line 173 delegates to `tokens.readName()`, which returns an absolute column, so `@.`, `^.` and `$.` heads come back short by the line's indent while a `foo.` head on the very same line comes back right. On a reversed dispatch nested two levels deep the root head reports column 0 instead of 4.

Nothing surfaces it today: `LnReversed#emit` at line 118 reads only `head.raw()`, and `Emissions#expression` builds its own head through `tokens.readValue()` rather than through this method, so the wrong number is never emitted. That is exactly why it will sit there until somebody attaches an `@pos` or an error message to the reversed head. `readHead` is `private static` and never receives the span, so the fix is to hand it the line's indent — the caller at line 63 has `this.span` — and add it to `start` the way every `Tokens` site already does.

@yegor256

Thayorns:
LnReversed#readHead built the head Value for a `@.`, `^.` or `$.` root
head as `new Value(Value.Kind.IDENTIFIER, mapped, start, start + 1)`,
where `start` is `Tokens#cursor()` — an offset into the span's body
with the leading whitespace already stripped. Value#pos is meant to be
a line-absolute column everywhere else: every construction site in
Tokens.java adds `this.span.indent() + start`, and PARSER_SPEC.md
spells the convention out.

The `else` arm of the same `if` delegates to `tokens.readName()`,
which already returns an absolute column, so the two arms disagreed —
a `foo.` head on an indented line reported the right column while a
`@.`, `^.` or `$.` head on the very same indent reported one short by
the whole indent.

`readHead` now takes the line's indent from its caller and folds it
into the root head's column the same way every `Tokens` site already
does, and is package-private so `LnReversedTest` can assert on
`pos()` directly for all three root symbols and confirm it now
matches the named-head convention.

Closes #7165


Thayorns:
@yegor256 happy to clarify anything about this fix if it's not clear from the PR description.

Thayorns:

yegor256:
@Thayorns please, add YAML tests fixture for this bug

Thayorns:
@yegor256 added, take a look at this one please

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2509](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7195, opened 2026-08-20
Value never returns null for a missing binding

### LINKED ISSUE #7166: `Value#binding()` returns null when no binding is present
`Value#binding()` at `eo-parser/src/main/java/org/eolang/parser/Value.java:162` returns `null` whenever the value was built through the two- or four-argument constructor, which is every value that carries no `:label`. `ValueTest#returnsNullBindingWhenAbsent` at `eo-parser/src/test/java/org/eolang/parser/ValueTest.java:89` asserts `nullValue()` on exactly that, so the null is nailed down as the published contract rather than treated as an accident.

Every consumer pays for it with a guard: `Bindings.java:47`, `Bindings.java:49`, `Bindings.java:68`, `Emissions.java:172` and `Emissions.java:191` all read the query only to compare it against `null` first. Miss one and the failure lands wherever the label is finally used, not where the value was built. Worse, the absent binding and the present-but-empty one are two different states the callers have to keep apart by hand — `null` versus `""` — and `BindingsTest#acceptsAllBoundArgsWithEmptyLabel` exists because somebody had to pin down which of the two `a:` is.

The rule the rest of this codebase follows is to never return null. Splitting the question in two — a `bound()` query answering whether a binding is there, and `binding()` returning the label, empty for the numeric and empty-label cases — lets the five call sites ask what they actually mean and removes the null from the type entirely.

@yegor256

Thayorns:
Closes #7166

`Value#binding()` used `null` as its sentinel for "no inline binding," which forced every call site to guard the query with a null check before using it. This splits the query in two: `bound()` answers whether a binding is present, and `binding()` now always returns a string — the label, or empty when there is none.

Updated the five call sites in `Bindings` and `Emissions` to ask `bound()` instead of comparing `binding()` against `null`, and adjusted `ValueTest` to match the new contract (added `isBoundWhenLabelGiven` / `isNotBoundWhenAbsent`, and renamed the absent-binding test to check for an empty string).


Thayorns:
@yegor256 happy to clarify anything about this change if needed.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2273](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7196, opened 2026-08-20
Drop dead Kind.horizontallyCompleted() and its stale docblock claims

### LINKED ISSUE #7167: Dead `horizontallyCompleted()` with a docblock naming callers that do not exist in `Kind.java`
`Kind.horizontallyCompleted()` in `eo-parser/src/main/java/org/eolang/parser/Kind.java:137` is never called by production code. A repo-wide grep finds it in three places only: its own definition, a javadoc sentence in `Openness.java:21`, and two assertions in `KindTest.java:31` and `KindTest.java:48`. The tests are the only thing keeping it alive.

Its own docblock at `Kind.java:20` says it is "the single source of truth for that set; R-5.2.3 and R-6.1.1 read it". Neither half is true. R-5.2.3(b) is enforced in `LnMethod.precheck` at `LnMethod.java:135`, which rejects a same-indent `.method` by reading `Openness.HCOMPLETED` off the stack top and never asks the kind anything. The string `R-6.1.1` does not appear anywhere in the sources.

The two sources also disagree about the set itself. `LnOnlyPhi.transition` at `LnOnlyPhi.java:280` pushes `ONLY_PHI` with `Openness.HCOMPLETED` whenever the phi carries horizontal args, and the constant's own docblock at `Kind.java:78` says exactly that, but `horizontallyCompleted()` reports `ONLY_PHI` as false unconditionally. Whoever reaches for the method next will trust the docblock, get an answer that contradicts the stack, and lose an afternoon to it.

Dropping the method and the two `KindTest` cases that call it, and correcting the paragraph at `Kind.java:20` to point at `Openness` as the authority, removes the contradiction without touching any behaviour.

@yegor256

Thayorns:
Kind.horizontallyCompleted() was never called by production code — grep finds it only in its own definition, a javadoc mention in Openness.java, and two KindTest assertions.

Its docblock claimed it was "the single source of truth" for R-5.2.3 and R-6.1.1, and that both rules read it. Neither is true: R-5.2.3(b) is enforced in LnMethod.precheck by reading Openness.HCOMPLETED off the stack, and R-6.1.1 doesn't appear anywhere in the sources. The two sources also disagreed about the set itself — LnOnlyPhi.transition pushes ONLY_PHI with Openness.HCOMPLETED for horizontal-arg phis, but horizontallyCompleted() reported ONLY_PHI as false unconditionally.

This drops the method, the two KindTest cases that only exercised it, and rewrites the docblocks in Kind.java and Openness.java to point at Openness as the actual authority for the horizontally-completed set, matching what LnMethod.precheck and LnOnlyPhi.transition already do.

Closes #7167


Thayorns:
@yegor256 happy to clarify anything about this cleanup if needed.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2289](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7197, opened 2026-08-20
fix(#7168): drop dead exposesTopLevelSentinel test in KindTest.java

### LINKED ISSUE #7168: `exposesTopLevelSentinel` cannot fail in `KindTest.java`
`KindTest.exposesTopLevelSentinel` at `eo-parser/src/test/java/org/eolang/parser/KindTest.java:76` asserts that `Kind.valueOf("TOP_LEVEL")` is `notNullValue()`. `Enum.valueOf` has no branch that returns null: it either returns the constant or throws `IllegalArgumentException`. Every run that reaches the assertion passes it, so the test has no failing input.

The regression it is meant to guard cannot reach it either. `TOP_LEVEL` is referenced from production code, so deleting the constant breaks compilation long before any test runs, and renaming it makes the lookup throw rather than fail an assertion. Either way the report says "error", not "the sentinel is gone".

Deleting the test is the honest fix. If the intent was to pin the sentinel's role rather than its name, the test worth having asserts what `TOP_LEVEL` does at indent 0 — that a level pushed under it is treated as top-level by the close-time naming check of R-5.3.1 — rather than that an enum constant exists.

@yegor256

Thayorns:
Kind.valueOf("TOP_LEVEL") can never return null: Enum.valueOf either returns the constant or throws IllegalArgumentException. exposesTopLevelSentinel asserted the result was notNullValue(), so the assertion can never fail and the test guarded nothing.

The regression it aimed to catch can't reach the assertion either. TOP_LEVEL is referenced from production code, so deleting the constant breaks compilation before any test runs, and renaming it makes the lookup throw rather than fail an assertion.

The behavior worth pinning — that a level pushed at indent 0 gets TOP_LEVEL as its parent, and that this drives the R-5.3.1 close-time naming check — is already covered elsewhere: StackTest.pushesTopLevelEntryAtIndentZero asserts the parent, and EoTest.rejectsTopLevelFormationWithoutName asserts the naming check fires on an unnamed top-level formation. No replacement test is needed.

Closes #7168


Thayorns:
@yegor256 ready for review — happy to clarify anything.

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (9) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2301](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #7171, opened 2026-08-19
`Hex(String)` constructor in `Hex.java` kept alive by its tests, causing potential misinterpretation and deviation from UTF-8 requirement


Thayorns:
The `Hex(String)` constructor at `eo-parser/src/main/java/org/eolang/parser/Hex.java:47` is never used. A repo-wide grep for `new Hex(` turns up five construction sites and not one of them passes a string: `LnTextBlock.java:86` hands in the `byte[] joined` that `Emissions.unescapeBytes` built, `Emissions.java:378` hands in the `byte[] unescaped`, and `Emissions.java:347`, `Emissions.java:453` and `StHex.java:30` each hand in a `double`. `Hex` is package-private, so `org.eolang.parser` is the entire search space.

The only two callers left are `encodesAsciiStringAsBytes` and `encodesUtfString` in `HexTest.java:66` and `HexTest.java:75`. The tests are what keep the constructor alive, and the class docblock at `Hex.java:17` advertises it as one of "three constructions supported", so the next person emitting a literal has every reason to reach for it instead of the `byte[]` path the emitters actually use.

That path is not equivalent. Both `Emissions.string` and `LnTextBlock` decode escapes through `unescapeBytes` before encoding, and `Hex(String)` skips that entirely, so a caller who takes the docblock at its word gets the raw `\uXXXX` text encoded verbatim instead of the decoded UTF-8 that R-9.7.4 requires.

Dropping the constructor and the two tests that call it, and trimming the docblock's bullet list to the two real forms, leaves every emitter untouched.

@yegor256

0crat:
@thayorns Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2141](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7203, opened 2026-08-20
#7171: drop the unused Hex(String) constructor

morphqdd:
`Hex(String)` was never used. A repo-wide grep for `new Hex(` finds five construction sites and none of them pass a string: `LnTextBlock.java`, `Emissions.java` (three call sites), and `StHex.java` all hand in a `byte[]` or `double`. `Hex` is package-private, so `org.eolang.parser` is the entire search space.

The constructor also isn't equivalent to the path callers actually use: both `Emissions.string` and `LnTextBlock` decode escapes through `unescapeBytes` before encoding, while `Hex(String)` skipped that entirely and would encode raw `\uXXXX` text verbatim instead of the decoded UTF-8 R-9.7.4 requires — a trap for whoever reached for it based on the class docblock's "three constructions supported" list.

Dropped the constructor, the now-unused `StandardCharsets` import, the two `HexTest` cases that only existed to call it, and trimmed the docblock to the two real forms.

See #7171.

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Hex.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Hex.java
@@ -5,7 +5,6 @@
 package org.eolang.parser;
 
 import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
 
 /**
  * Hex encoder for numeric and string literals.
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5515](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7207, opened 2026-08-20
#7173: recognize -> throwing-test suffix in horizontal argument scan

### LINKED ISSUE #7173: A `->` throwing-test suffix is rejected on every line shape except a formation
## What happens

`Tokens.suffixAhead()` (`eo-parser/src/main/java/org/eolang/parser/Tokens.java:529`) stops the horizontal-argument scan at `>` and `+>` only:

```java
boolean suffixAhead() {
    return !this.atEnd() && (this.current() == '>' || this.plusArrow());
}
```

`plusArrow()` (`Tokens.java:994`) matches `+>` and there is no `->` counterpart, so `readArgs()` never hands the tail over to `Suffix`, which does know `->` (`Suffix.java:345`). Every line shape that reads arguments before the suffix - `LnApplication:63`, `LnReversed:70`, `LnMethod`, `LnCompactTuple` - dies on the `-`.

```
[] > foo

  true -> throws-test
```

gives `[3:7] error: 'expected value'` and `foo` comes out as an empty `<o name="foo"/>`, its child lost.

```
[] > foo

  eq. -> throws-something
    1
    2
```

gives `[3:6] error: 'expected value'`, and recovery skips the whole block.

The `+>` twins are fine: `true +> truthy-test` yields `<o base="Φ.true" name="+truthy-test"/>`, and `eq. +> compares-something` is an existing pack (`eo-syntax/application-in-tests.yaml`). `[] -> throws-something` also parses, because `LnFormation` never calls `readArgs`.

## What should happen

Per §1.4 of `PARSER_SPEC.md`, `-> name` is "parallel to `+>` in every respect (same `NAME`-token rule, same depth constraint R-6.3.3, same mandatory blank line R-6.5.3); only the emitted marker prefix differs", and R-6.3.1 lets an atom body hold truthy `+>` or throwing `->` tests. So `true -> t`, `eq. -> t` and `.foo -> t` should emit `name="-t"` exactly the way their `+>` forms do. `suffixAhead()` needs to know `->` as well.

morphqdd:
`Tokens.suffixAhead()` stopped the horizontal-argument scan at `>` and `+>` only, so `readArgs()` never handed the tail over to `Suffix` when it started with `->`. Every line shape that reads arguments before the suffix (application, method, reversed, pipe) failed to parse a throwing-test attribute, e.g.

```
[] > foo

  true -> throws-test
```

Added a `minusArrow()` check mirroring `plusArrow()`, and a parser pack test covering `true ->`, `false ->`, and `eq. ->` throwing-test shapes, matching the existing `+>` pack.

Fixes #7173


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (34) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5339](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7208, opened 2026-08-20
#7174: emit method chain after text-block closer

### LINKED ISSUE #7174: Method chain after a text-block closer is parsed and silently dropped in `LnTextBlock`
## What happens

`LnTextBlock.into` reads the method chain after the closing triple quote and throws the result away. `eo-parser/src/main/java/org/eolang/parser/LnTextBlock.java:52`:

```java
final Tokens tokens = new Tokens(body, this.span);
tokens.seek(3);
tokens.readChain();
final Suffix suffix = new Suffix(...);
```

Nothing consumes what `readChain()` returned, and the emitter then writes a bare `<o base='Φ.string'>`.

```
[] > main
  """
  hi
  """.as-bytes > txt
```

parses into

```xml
<o line="1" name="main">
  <o base="Φ.string" line="4" name="txt" pos="2">
    <o base="Φ.bytes" ...><o>68-69</o></o>
  </o>
</o>
```

The `.as-bytes` is simply gone. No error, no `@base='.as-bytes'`, no wrapper, so the program means something other than what was written.

The optional binding of the same rule is missing too: `""" :0` falls through to `Suffix` and is reported as trailing garbage.

## What should happen

R-3.11.4 says: "After the closing `\"\"\"`, a method chain is allowed (`\"\"\".as-bytes`), then the optional binding and suffix." The chain should be emitted the way `ChainEmission` emits it for every other head. If the shape is not meant to be supported, the parser should say so instead of dropping it.

morphqdd:
`LnTextBlock.into` read the method chain after the closing `"""` with `tokens.readChain()` but discarded the result, so `""".as-bytes > txt` parsed into a bare `<o base='Φ.string'>` with the `.as-bytes` link silently gone.

Per R-3.11.4, a method chain after the closer is allowed, then the optional binding and suffix. This emits the chain the way `ChainEmission` emits it for every other head: the string object becomes the first sibling, each chain link is a flat `.method` sibling, and the outer binding/suffix attach to the last link. Also wires up the optional `:label`/`:N` binding via the same `LnApplication.readOuterBinding` helper already used by application lines, which previously fell through to `Suffix` and was reported as trailing garbage.

Fixes #7174

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/LnTextBlock.java
+++ b/eo-parser/src/main/java/org/eolang/parser/LnTextBlock.java
@@ -4,6 +4,8 @@
  */
 package org.eolang.parser;
 
+import java.util.List;
+
 /**
  * A triple-quoted text-block closer line — §3.11 of the spec.
  *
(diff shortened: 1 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5355](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7209, opened 2026-08-20
#7175: track pipe and compact-tuple bindings for all-or-nothing rule

### LINKED ISSUE #7175: All-or-nothing binding rule is not applied to pipe and compact-tuple argument groups
## What happens

`Bindings.tracksBindings` (`eo-parser/src/main/java/org/eolang/parser/Bindings.java:128`) lists four kinds:

```java
private static boolean tracksBindings(final Kind kind) {
    return kind == Kind.HEAD || kind == Kind.HMETHOD
        || kind == Kind.VAPPLICATION || kind == Kind.IDENTITY_OBJECT;
}
```

`PIPE_APPLICATION` and `COMPACT_TUPLE` are not there, so `observeChild` (`Bindings.java:117`) never observes their vertical children and the all-or-nothing rule is never checked for them.

A pipe group with one bound and one unbound argument is accepted:

```
[] > app
  [x y] > add
    x.plus y > @
  | >>
    "hello":foo
    42
```

```xml
<o base="ξ.add" pipe="" name="a🌵4-2">
  <o as="foo" base="Φ.string" .../>
  <o as="α0"  base="Φ.number" .../>
</o>
```

So is a compact tuple:

```
[] > main
  sprintf *1 > r
    1:a
    2
```

The same mistake under a plain head is caught:

```
[] > main
  add > r
    1:a
    2
```

gives `[4:4] error: 'argument bindings must be all-or-nothing'`.

## What should happen

R-6.6.2 governs "a same-indent group of application arguments", and R-3.14.6 says of the pipe explicitly: "All-or-nothing inline binding (§6.6) applies to the argument group." A compact tuple's children are application arguments too (R-3.9.2). Both should be rejected at the first divergent child, the way a head is, which means adding the two kinds to `tracksBindings`.

morphqdd:
`Bindings.tracksBindings` listed `HEAD`, `HMETHOD`, `VAPPLICATION`, and `IDENTITY_OBJECT`, so `observeChild` never observed vertical children of a pipe group or a compact tuple, and the all-or-nothing rule (R-6.6.2) was never checked for them.

```
[] > app
  [x y] > add
    x.plus y > @
  | >>
    "hello":foo
    42
```

parsed without error even though one argument is bound and the other is not. A plain application head catches this exact mix.

R-3.14.6 says the all-or-nothing rule applies to a pipe's argument group explicitly, and R-3.9.2 says a compact tuple's children are application arguments too. Added `PIPE_APPLICATION` and `COMPACT_TUPLE` to `tracksBindings`.

Fixes #7175

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Bindings.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Bindings.java
@@ -4,7 +4,9 @@
  */
 package org.eolang.parser;
 
+import java.util.EnumSet;
 import java.util.List;
+import java.util.Set;
 
 /**
  * Cross-argument inline-binding validation — §6.6 of the spec.
@@ -22,6 +24,15 @@ import java.util.List;
  * @since 0.1
  */
 final class Bindings {
+
+    /**
+     * The parent kinds whose deeper children are application arguments,
+     * so the all-or-nothing rule of R-6.6.2 governs the group they form.
+     */
+    private static final Set<Kind> TRACKED = EnumSet.of(
+        Kind.HEAD, Kind.HMETHOD, Kind.VAPPLICATION, Kind.IDENTITY_OBJECT,
+        Kind.PIPE_APPLICATION, Kind.COMPACT_TUPLE
+    );
 
     /**
      * No instances.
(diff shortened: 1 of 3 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5371](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7210, opened 2026-08-20
#7176: stop a one-byte chunk from triggering BYTES continuation merge

### LINKED ISSUE #7176: A one-byte chunk continues a multi-line `BYTES` literal, which R-3.13.1 forbids
## What happens

`Eo.mergeBytesContinuation` decides whether to keep merging from `next.body().endsWith("-")` alone (`eo-parser/src/main/java/org/eolang/parser/Eo.java:224`), plus a check that the line is bytes only. It never re-applies the two-byte condition that `isBytesContinuation` (`Eo.java:248`, `length() >= 6`) applies to the opening chunk.

```
[] > foo
  size. > @
    CA-FE-
    BE-
    BE-BE
```

parses without a word into one literal:

```xml
<o base=".size" line="2" name="φ">
  <o base="Φ.bytes" line="3" pos="4"><o as="α0">CA-FE-BE-BE-BE</o></o>
</o>
```

## What should happen

R-3.13.1 says a one-byte chunk is "always complete on its own line" and that "the trailing `-` signalling continuation requires a chunk of at least two bytes; a one-byte chunk cannot trigger continuation". So the token has to stop at `BE-`, which leaves `CA-FE-BE-` with a dangling continuation dash - the error `Tokens.readPairs` already raises for that shape after #6566 - and `BE-BE` is a line of its own. Today the user silently gets a five-byte value where the rules give an error and a different program.

This is not #6295 (a named final chunk) nor #7076 (a trailing space).

morphqdd:
`Eo.mergeBytesContinuation` decided whether to keep merging a chunk into the multi-line BYTES literal from `trimmed.endsWith("-")` alone, without re-applying the two-byte minimum that `isBytesContinuation` already enforces for the opening chunk.

```
[] > foo
  size. > @
    CA-FE-
    BE-
    BE-BE
```

merged all three lines into one five-byte literal, silently dropping a byte from what was written.

R-3.13.1 says a one-byte chunk is always complete on its own line and its trailing dash cannot trigger continuation. Reusing `isBytesContinuation` in the merge loop stops the merge after `BE-`, leaving `CA-FE-BE-` with a dangling continuation dash, which `Tokens.readPairs` already rejects (the error added for #6566), and `BE-BE` becomes its own line.

Fixes #7176

### DIFF
--- a/eo-parser/src/main/java/org/eolang/parser/Eo.java
+++ b/eo-parser/src/main/java/org/eolang/parser/Eo.java
@@ -189,7 +189,7 @@ final class Eo implements Iterable<Directive> {
             }
             body.append(trimmed);
             idx = idx + 1;
-            if (!trimmed.endsWith("-")) {
+            if (!Eo.isBytesContinuation(trimmed)) {
                 break;
             }
         }
--- a/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/one-byte-chunk-does-not-continue-bytes.yaml
+++ b/eo-parser/src/test/resources/org/eolang/parser/eo-syntax/one-byte-chunk-does-not-continue-bytes.yaml
@@ -1,0 +1,15 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# #7176: R-3.13.1 says a one-byte chunk is complete on its own line and cannot
+# continue a bytes literal, so the merge stops at `BE-` and the token that was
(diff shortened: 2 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5383](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7213, opened 2026-08-20
#7177: return one empty string for an empty text in nsplit

### LINKED ISSUE #7177: `string.nsplit` incorrectly returns empty tuple for an empty text unlike `string.split`
## What happens

`string.nsplit` short-circuits an empty text into an empty tuple. `eo-runtime/src/main/eo/string/nsplit.eo:29`:

```
    if.
      text.size.eq 0
      *
```

so `"".nsplit "-" 3` is a tuple of length 0. Its sibling does the opposite - `split.eo:17`:

```
      eq.
        size. >> len!
          text >> bts!
        0
      * ""
```

so `"".split "-"` is `* ""`, one element. Java agrees with `split`: `"".split("-", 3)` is `[""]`.

Even `"".nsplit "-" 1` cannot reach the `1.eq limit` branch on line 33, because the empty-text branch is tested first, so it does not return `* text` either.

## What should happen

`"".nsplit "-" 3` should return `* ""`, the same one-element tuple `split` gives, since the whole text is one (empty) part.

This is #6597 - "`string.split` returns an empty tuple for an empty text instead of one empty string" - applied to `split` only; `nsplit` kept the old behaviour. `split.eo` has the test `can-split-an-empty-text-into-one-empty-string`; `nsplit.eo` has no empty-text test at all. Its own comment (`nsplit.eo:5`) says it behaves "the same way `split` behaves".

morphqdd:
`string.nsplit` short-circuited an empty text into an empty tuple before checking the limit, so `"".nsplit "-" 3` returned a tuple of length 0. Its sibling `string.split` returns `* ""`, a one-element tuple, for the same input (fixed in #6597), and `nsplit`'s own docblock says it behaves the same way `split` does.

Removed the empty-text special case. With it gone, `1.eq limit` and the `rec-nsplit` recursion already produce the right one-element result for an empty text on their own: the base case (`^.text.size.eq current`) is true from the first call, so it returns the accumulator with one empty string appended, matching `split`.

Added `can-nsplit-an-empty-text-into-one-empty-string`, mirroring `split.eo`'s existing `can-split-an-empty-text-into-one-empty-string`.

Fixes #7177

### DIFF
--- a/eo-runtime/src/main/eo/string/nsplit.eo
+++ b/eo-runtime/src/main/eo/string/nsplit.eo
@@ -27,60 +27,63 @@
         "Invalid limit %d for nsplit, must be at least 1".printf
           * limit
     if.
-      text.size.eq 0
-      *
-      if.
-        1.eq limit
-        * text
-        [accum start current count] >> rec-nsplit
-          if. > @
-            ^.text.size.eq current
-            appended
+      1.eq limit
+      * text
+      [accum start current count] >> rec-nsplit
+        if. > @
+          ^.text.size.eq current
+          appended
+          if.
+            or.
+              size.eq 0
+              gt.
+                current.plus size
+                ^.text.size
+            ^.rec-nsplit
+              accum
+              start
+              as-number.
+                current.plus 1
+              count
             if.
-              or.
-                size.eq 0
-                gt.
+              and.
+                ^.delimiter.eq
+                  bts.slice current size
+                lt.
+                  number count
+                  minus. (number ^.limit) 1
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5399](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #7214, opened 2026-08-20
Let the receiver void stand anywhere among the voids

maxonfjvipon:
A `^` void had to be the first attribute of its formation, so `[x ^] > foo` was rejected outright. That rule rested on the idea that a caller fills the receiver first, and a receiver declared later would land in the wrong slot. It does not work that way: a dispatch resolves the receiver by name, looking for an attribute called `ρ`. Position never mattered.

So `^` becomes an ordinary void. It keeps the one rule every void obeys, R-3.4.9 — no void below a bound attribute — which `Stack` already enforces by routing each child through `Level.observeVoid`. Nothing had to be added for that.

Three sites carried the restriction and all three are gone:

```java
} else if ("^".equals(raw)) {
    mapped = "ρ";
```

`Level.receiver` and `Level.heads` existed only to serve the check, so `Level.java` returns to being byte-identical with master, and `LnFormation.transition` goes back to returning `void`.

The three `eo-typos` packs that asserted the error are now `eo-syntax` packs asserting the void lands in the slot it was written in, plus one new typo pack pinning that a `^` below a bound attribute still fails. Parsing all of `eo-runtime` gives 40547 `<o>` nodes byte-identical to master, so nothing already written changes shape.

Writing `^` first stays good practice; a lint can say so later.


0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (130 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+969](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

